PR #8540, opened 2019-02-15
Validate network plugin type

cliffwoodave14:
This PR contains two changes:
1) if any VNet related option is listed but not no any network plugin is specified, then it returns a warning. 
2) If the customers specify the option pod_cidr but without specifying network_plugin as kubenet, then it returns a warning.


src/command_modules/azure-cli-acs/azure/cli/command_modules/acs/custom.py
```diff
@@ -1502,7 +1502,13 @@ def aks_create(cmd, client, resource_group_name, name, ssh_key_value,  # pylint:
...
 
     network_profile = None
-    if any([network_plugin, pod_cidr, service_cidr, dns_service_ip, docker_bridge_address, network_policy]):
+    if any([pod_cidr, service_cidr, dns_service_ip, docker_bridge_address, network_policy]):
+        if not network_plugin:
+            logger.warning('Please explicitly specify network plugin type')
+            return
+        if pod_cidr and network_plugin == "azure":
+            logger.warning('Please use kubenet as a network plugin when pod_cidr is specified')
+            return
```
tjprescott: Seems to me like these should be usage errors, not warnings, but I will defer to AKS folks on whatever they prefer.

zqingqing1:
lgtm

src/command_modules/azure-cli-acs/azure/cli/command_modules/acs/custom.py
cliffwoodave14: You're right. I will change them to raise a CLIError.

PR #8549, opened 2019-02-15
Replace yaml.load with yaml.safe_load

tjprescott:
Knack was updated to use `yaml.safe_load` instead of `yaml.load` based on discussions with @derekbekoe. This PR updates the knack dependency in CLI to 0.5.2 and implements the same fix, replacing yaml.load with yaml.safe_load and removing the CLI dependency on the beta version of PyYAML which will be removed from PyPI.

---


tjprescott:
Since Python 2.7 is being Python 2.7, I will remove this from the current milestone.

tjprescott:
Closing this for now. Was hoping it would be a quick fix but since it is being stubborn I don't want it polluting out PR list.

PR #8556, opened 2019-02-16
Cosmos DB Allow users to specify ignoreMissingVNetServiceEndpoint

shurd:
This change gives users the option to specify a boolean value for the "ignoreMissingVNetServiceEndpoint" aspect of a VirtualNetworkResource. They can specify this optional value similar to how failover priority is specified for locations in a VNetId[=ignoreMissingVNetServiceEndpoint] format.

---


src/command_modules/azure-cli-cosmosdb/azure/cli/command_modules/cosmosdb/custom.py
```diff
@@ -221,6 +222,128 @@ def cli_cosmosdb_list(client, resource_group_name=None):
...
+    cosmos_db_account = client.get(resource_group_name, account_name)
+    return cosmos_db_account.virtual_network_rules
+
+
+def _get_virtual_network_id(cmd, resource_group_name, subnet, virtual_network):
+    from azure.cli.core.commands.client_factory import get_subscription_id
+    from msrestazure.tools import is_valid_resource_id, resource_id
+    if not is_valid_resource_id(subnet):
+        if virtual_network is None:
+            raise CLIError("Provide either the subnet ID or the VNET and subnet name")
```
tjprescott: Recommend usage error syntax here:
`usage error: --subnet ID | --subnet NAME --vnet-name NAME`

src/command_modules/azure-cli-cosmosdb/azure/cli/command_modules/cosmosdb/custom.py
```diff
@@ -221,6 +222,128 @@ def cli_cosmosdb_list(client, resource_group_name=None):
...
+    virtual_network_rules = []
+    rule_already_exists = False
+    for rule in existing.virtual_network_rules:
+        if rule.id != subnet:
+            virtual_network_rules.append(
+                VirtualNetworkRule(id=rule.id,
+                                   ignore_missing_vnet_service_endpoint=rule.ignore_missing_vnet_service_endpoint))
+        else:
+            if rule.ignore_missing_vnet_service_endpoint == ignore_missing_vnet_service_endpoint:
+                raise CLIError("This rule already exists for the Cosmos DB account")
```
tjprescott: My recommendation to simplify the logic here would be to just detect that the rule already exists and then to add a warning "This rule already exists and will be overwritten." instead of then checking whether the flag changed or not.

Generally, we would like creates to be idempotent, so throwing an error because nothing has changed would make this command not idempotent and could interfere with scripting efforts.

src/command_modules/azure-cli-cosmosdb/azure/cli/command_modules/cosmosdb/custom.py
```diff
@@ -221,6 +222,128 @@ def cli_cosmosdb_list(client, resource_group_name=None):
...
+        ip_range_filter=existing.ip_range_filter,
+        enable_automatic_failover=existing.enable_automatic_failover,
+        capabilities=existing.capabilities,
+        is_virtual_network_filter_enabled=True,
+        virtual_network_rules=virtual_network_rules,
+        enable_multiple_write_locations=existing.enable_multiple_write_locations)
+
+    async_docdb_create = client.create_or_update(resource_group_name, account_name, params)
+    docdb_account = async_docdb_create.result()
+    docdb_account = client.get(resource_group_name, account_name)  # Workaround
```
tjprescott: Your PUT doesn't return the object? You *should* be able to just:

```Python
return client.create_or_update(resource_group_name, account_name, params)
```

shurd: I looked through our handler, and the only thing we currently return when the operation completes is {status: "Succeeded"}, so for the time being, the GET is needed to retrieve the object.

src/command_modules/azure-cli-cosmosdb/azure/cli/command_modules/cosmosdb/custom.py
```diff
@@ -318,7 +313,7 @@ def cli_cosmosdb_network_rule_remove(cmd,
         else:
             rule_removed = True
     if not rule_removed:
-        raise CLIError("This rule does not exist for the Cosmos DB account")
+        raise logger.warning("This rule does not exist for the Cosmos DB account")
```
tjprescott: This actually would be okay to remain an error, but the end result isn't any different. Deletes on non-existent resources throw errors.

shurd: I'll make the change so it is consistent

PR #8561, opened 2019-02-17
Add optional overwrite for duplicate kubeconfig values.

brendandburns:
@mboersma


mboersma:
I think there also needs to be a command-line way to force overwrite, so scripts can guarantee that an interactive prompt doesn't pop up. Maybe just having an `--overwrite` argument that defaults to `False`.

 We have a lot of CI tests that use the `az` client directly, for example, and this change might break them.

mboersma:
I was confused about where this code gets called--apologies for my first review comment. I checked out this branch and tested it. This behavior is definitely an improvement and should be safe by default.

I took this commit and fixed the linter errors in #8681. Please take a look.

mboersma:
I cherry-picked this commit and updated it so it's passing in #8681, so I think we can close this PR.

PR #8590, opened 2019-02-20
Fix redundant sources of image import.

### LINKED ISSUE #8451: az acr import allows redundant sources
**Describe the bug**
Commands like:
az acr import --source mysource.azurecr.io/myrepo:mytag -n mytarget --force -r /subscriptions/mysub/resourceGroups/myrg/providers/Microsoft.ContainerRegistry/registries/mysource

Specifies mysource as the source registry twice.  However the backend import service has 2 modes, one that takes a registry uri and one that takes a resource id.
**To Reproduce**
specify a registry with the source image argument and specify a registry resource id.  The command sends a request using the registry uri mode which is probably incorrect and the error will be confusing to the user.

**Expected behavior**
If the 2 sources match, use the resource id mode.  If the 2 sources don't match err saying only one source registry is allowed.

NatalieYuGong:
---

Fixes #8451


src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
```diff
@@ -73,7 +73,7 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
                        help='Default action to apply when no rule matches. Only applicable to Premium SKU.')
 
     with self.argument_context('acr import') as c:
-        c.argument('source', help="The source identifier in the format '[registry.azurecr.io/]repository[:tag]' or '[registry.azurecr.io/]repository@digest'.")
+        c.argument('source', help="The source identifier will be either a source image name or a fully qualified source.")
```
nathana1: Should we keep the examples for what is a source image or fully qualified source?  Also mention that when -r is specified it is a source image?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/import.py
```diff
@@ -36,53 +34,40 @@ def acr_import(cmd,
...
-               source_registry.lower() != registry.id.lower() and \
-               source_registry.lower() != registry.name.lower() and \
-               source_registry.lower() != registry.login_server.lower():
-                raise CLIError(REGISTRY_MISMATCH)
-            source = ImportSource(resource_id=registry.id,
-                                  source_image=source_image)
-        else:
+        slash = source_image.find('/')
+        if slash > 0:
+            registry_uri = source_image[:slash]
```
nathana1: didn't want to check if this was possibly a host name (i.e. contains a '.')?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
NatalieYuGong: Examples are given as --help

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/import.py
NatalieYuGong: what do you want to do if source registry is not provided and the string before slash does not contain a dot?

NatalieYuGong: Any valid host name could potentially work is what I had on the Team channel. /cc @sajayantony

nathana1: It should error out, a valid host name will have a dot in it.

PR #8596, opened 2019-02-21
[HDInsight] - Support ADLS Gen2 MSI

idear1203:
## Note

- Update `azure-mgmt-hdinsight` SDK version to 0.2.1
- `az hdinsight create`: added the `--storage-account-managed-identity` parameter to support ADLS Gen2 MSI.
---

### DIFF
--- a/src/command_modules/azure-cli-hdinsight/azure/cli/command_modules/hdinsight/custom.py
+++ b/src/command_modules/azure-cli-hdinsight/azure/cli/command_modules/hdinsight/custom.py
@@ -100,7 +100,7 @@ def create_cluster(cmd, client, cluster_name, resource_group_name, cluster_type,
         raise CLIError('Either the default container or the default filesystem can be specified, but not both.')
 
     # Retrieve primary blob service endpoint
-    is_wasb = True if not storage_account_managed_identity else False
+    is_wasb = not storage_account_managed_identity
     storage_account_endpoint = None
     if storage_account:
         storage_account_endpoint = get_storage_account_endpoint(cmd, storage_account, is_wasb)
@@ -207,7 +207,7 @@ def create_cluster(cmd, client, cluster_name, resource_group_name, cluster_type,
                 key=storage_account_key,
                 container=storage_default_container,
                 file_system=storage_default_filesystem,
-                resource_id=storage_account if not is_wasb else None,
+                resource_id=None if is_wasb else storage_account,
                 msi_resource_id=storage_account_managed_identity,
                 is_default=True
             )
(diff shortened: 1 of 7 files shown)


src/command_modules/azure-cli-hdinsight/azure/cli/command_modules/hdinsight/custom.py
```diff
@@ -99,13 +100,13 @@ def create_cluster(cmd, client, cluster_name, resource_group_name, cluster_type,
         raise CLIError('Either the default container or the default filesystem can be specified, but not both.')
 
     # Retrieve primary blob service endpoint
+    is_wasb = True if not storage_account_managed_identity else False
```
yugangw-msft: maybe just `is_wasb = not storage_account_managed_identity`?

src/command_modules/azure-cli-hdinsight/azure/cli/command_modules/hdinsight/custom.py
```diff
@@ -115,16 +116,20 @@ def create_cluster(cmd, client, cluster_name, resource_group_name, cluster_type,
...
     # Validate storage info parameters
-    if not _all_or_none(storage_account, storage_account_key,
-                        (storage_default_container or storage_default_filesystem)):
+    if is_wasb and not _all_or_none(storage_account, storage_account_key, storage_default_container):
         raise CLIError('If storage details are specified, the storage account, storage account key, '
-                       'and either the default container or default filesystem must be specified.')
+                       'and the default container must be specified.')
+    elif not is_wasb and not _all_or_none(storage_account, storage_default_filesystem):
+        raise CLIError('If storage details are specified, the storage account, '
+                       'and the default filesystem must be specified.')
```
yugangw-msft: please double check whether or not the error text is right. Not so sure what `_all_or_none` means really.

src/command_modules/azure-cli-hdinsight/azure/cli/command_modules/hdinsight/custom.py
```diff
@@ -202,6 +207,8 @@ def create_cluster(cmd, client, cluster_name, resource_group_name, cluster_type,
                 key=storage_account_key,
                 container=storage_default_container,
                 file_system=storage_default_filesystem,
+                resource_id=storage_account if not is_wasb else None,
```
yugangw-msft: Suggest `None if is_wasb else storage_account` to get rid of the `not`

src/command_modules/azure-cli-hdinsight/azure/cli/command_modules/hdinsight/custom.py
idear1203: Thanks for advance! I will update the logic.

src/command_modules/azure-cli-hdinsight/azure/cli/command_modules/hdinsight/custom.py
idear1203: Good catch! This logic is not quite accurate but cannot really be reached. I will make a little change to check none only.

src/command_modules/azure-cli-hdinsight/azure/cli/command_modules/hdinsight/custom.py
idear1203: Good suggestion! I am still on my way of learning Python:)

src/command_modules/azure-cli-hdinsight/azure/cli/command_modules/hdinsight/custom.py
idear1203: After a third thought, this is correct logic. Let's keep it as what it is now.

PR #8599, opened 2019-02-21
Adding existence check to role assignment creation

### LINKED ISSUE #8568: az role assignment create is not idempotent
Creating the same role assignment twice results in success the first time, but the second time results in an error: `The role assignment already exists.`

```
$ az role assignment create --role Owner --assignee "my@account.example.com" --scope /subscriptions/<redacted>/resourceGroups/<redacted>/providers/Microsoft.Storage/storageAccounts/<redacted>
The role assignment already exists.
$ echo $?
1
```

According to https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md#standard-command-types this operation should be idempotent, i.e.: creating existing role assignments should not result in an error.

This happens with azure-cli 2.0.58 (role 2.4.0) on Linux but also with azure-cli 2.0.51 (storage 2.2.5) on Windows.

marstr:
Fixes #8568

The role assignment creation REST API is not idempotent when the HTTP verb PUT is used. If an existing role is there, instead of returning a 200 it returns a 409 (even though the requests are identical, and definitely not in conflict.) For that reason I've added a list operation first to see if any exist before sending the creation request.

This is a curious thing, as other provisioning tools will have run into the same problem. For instance, how do ARM templates handle this?


marstr:
Worth noting, two additional properties are included in the response if the assignment already existed: `roleDefinitionName` and `principalName`.

tjprescott:
@marstr can you this this PR passing?

yugangw-msft:
@marstr, I will take care of this in my PR for overall idempotent support in RBAC area.
Closing

Issue #1528, opened 2016-12-12
Add wait/no-wait support to Commands


tjprescott:
PR #1463 added support for the `... wait` command for select command groups and the `--no-wait` option for several long-running operations in those groups. This issue is for identifying and tracking which commands still need this functionality exposed.

- [x] vm create/update (PR #1463)
- [x] group delete (PR #1463)
- [x] group deployment create (PR #1463)
- [x] application-gateway create/update/delete (PR #1463)
- [x] vnet-gateway create/update (PR #1463)
- [ ] local-gateway create/update/delete
- [ ] express-route create/update/delete
- [ ] vmss create/update/delete/start/stop/deallocate/...
- [ ] vm delete/deallocate/start/stop/...

derekbekoe:
This is needed for end of Sprint 12 right?
cc: @johanste, @yugangw-msft

derekbekoe:
Removed milestone as this is an additive change.

tjprescott:
I would think this is not really a GA-mandatory item. It would simply be an additive feature we can incrementally roll out over time. This issue is to track those items for which it would be nice to have.

yangl900:
Please add support for "az vm access" as well, the command took minutes to complete, it's practically not possible to manage large set of VMs without --no-wait.

jamescgibson:
I'd love to get --no-wait on `az disk delete`

mrx88:
I would really like to get --no-wait for "az network nic create" and "am vm extension set" commands.

weznagwama:
az group deployment delete --no-wait would be great, ATM there's no way to async delete a batch of deploys, have to serially do it one by one and it takes a long time.

dcbrown16:
I have a customer request for --no-wait on `az snapshot create`. Thanks!

tjprescott:
@yugangw-msft this issue is intended not to be closed. It is intended for users to comment when there are new commands they want wait/no-wait support for.

goenning:
Can we have `--no-wait` on `az container start` and `az container stop`? It is already supported when creating containers, but not when restarting an existing one.

tjprescott:
@geoning I don't see any instance when `az container stop` is a long running operation.  `az container start` can be.  Is there a particular image you use that makes container stop long-running?

goenning:
[...]

PR #8600, opened 2019-02-21
[Container] Support --no-wait on container start/restart.

tjprescott:
Issue #1528.

---


src/command_modules/azure-cli-container/azure/cli/command_modules/container/commands.py
```diff
@@ -27,6 +27,6 @@ def load_command_table(self, _):
         g.custom_command('attach', 'attach_to_container')
 
     with self.command_group('container', container_group_sdk) as g:
-        g.command('restart', 'restart')
+        g.command('restart', 'restart', supports_no_wait=True)
         g.command('stop', 'stop')
```
adewaleo: Stop doesn't need a no wait?

tjprescott: No. As far as I can tell, it is never long-running.

PR #8607, opened 2019-02-21
Get default location from resource group when not provided for creati…

QingqingYuan2:
…ng servers and add validation for retention days

---

### DIFF
--- a/src/command_modules/azure-cli-rdbms/azure/cli/command_modules/rdbms/validators.py
+++ b/src/command_modules/azure-cli-rdbms/azure/cli/command_modules/rdbms/validators.py
@@ -40,6 +40,7 @@ def password_validator(ns):
         except NoTTYException:
             raise CLIError('Please specify password in non-interactive mode.')
 
+
 def retention_validator(ns):
     if ns.backup_retention_days:
         val = ns.backup_retention_days
@@ -48,6 +49,7 @@ def retention_validator(ns):
         else:
             raise CLIError('incorrect usage: --backup_retention. Range is 7 to 35 days.')
 
+
 def retention_validator2(ns):
     if ns.backup_retention:
         val = ns.backup_retention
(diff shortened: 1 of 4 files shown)


src/command_modules/azure-cli-rdbms/azure/cli/command_modules/rdbms/validators.py
```diff
@@ -41,6 +41,24 @@ def password_validator(ns):
             raise CLIError('Please specify password in non-interactive mode.')
 
 
+def retention_validator(ns):
+    if ns.backup_retention_days:
+        val = ns.backup_retention_days
+        if val >= 7 and val <= 35:
+            pass
+        else:
+            raise CLIError('incorrect usage: --backup_retention. Range is 7 to 35 days.')
```
williexu:
```
if not 7 <= val <= 35:
    raise ...
```

src/command_modules/azure-cli-rdbms/azure/cli/command_modules/rdbms/validators.py
```diff
@@ -41,6 +41,24 @@ def password_validator(ns):
...
+
+
+def retention_validator2(ns):
+    if ns.backup_retention:
+        val = ns.backup_retention
+        if val >= 7 and val <= 35:
+            pass
+        else:
+            raise CLIError('incorrect usage: --backup_retention. Range is 7 to 35 days.')
```
williexu: You can avoid having this second validator by checking for both `backup_retention` and `backup_retention_days`. 
As a better change, it would be best to change the instances of `backup_retention` to `backup_retention_days` in the function signature being used the the custom.py file.

src/command_modules/azure-cli-rdbms/azure/cli/command_modules/rdbms/validators.py
QingqingYuan2: updated

src/command_modules/azure-cli-rdbms/azure/cli/command_modules/rdbms/validators.py
QingqingYuan2: Updated the function and get rid of the extra validator

ajlam:
Released in v2.0.60. Release notes [here](https://github.com/MicrosoftDocs/azure-docs-cli/blob/master/docs-ref-conceptual/release-notes-azure-cli.md#rdbms-1).

PR #8625, opened 2019-02-22
[AKS] Improve messages for upgrade command

lalo:
- Inform users of specific behaviours when they try to upgrade a cluster
to the current version of that cluster.

Skipped the HISTORY.rst since this is only logging, can add if prompted by maintainers (or anyone who disagrees).

---

### DIFF
--- a/src/command_modules/azure-cli-acs/azure/cli/command_modules/acs/custom.py
+++ b/src/command_modules/azure-cli-acs/azure/cli/command_modules/acs/custom.py
@@ -1735,8 +1735,8 @@ def aks_upgrade(cmd, client, resource_group_name, name, kubernetes_version, no_w
 
     if instance.kubernetes_version == kubernetes_version:
         if instance.provisioning_state == "Succeeded":
-            logger.warning("Your cluster is already on version %s and in a non-failed cluster state. No operations "
-                           "will occur on same version upgrades when cluster is in non-failure state.",
+            logger.warning("The cluster is already on version %s and is not in a failed state. No operations "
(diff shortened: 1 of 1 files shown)


lalo:
@mboersma adding you since you've done some work here before, also adding @jluk @yangl900

src/command_modules/azure-cli-acs/azure/cli/command_modules/acs/custom.py
```diff
@@ -1732,6 +1732,16 @@ def aks_scale(cmd, client, resource_group_name, name, node_count, nodepool_name=
 
 def aks_upgrade(cmd, client, resource_group_name, name, kubernetes_version, no_wait=False, **kwargs):  # pylint: disable=unused-argument
     instance = client.get(resource_group_name, name)
+
+    if instance.kubernetes_version == kubernetes_version:
+        if instance.provisioning_state == "Succeeded":
+            logger.warning("Your cluster is already on version %s and in a non-failed cluster state. No operations "
```
mboersma: "non-failed" sounds a little awkward to me. Maybe reword it like this?

"The cluster is already on version %s and is not in a failed state. No operations "
"will occur when upgrading to the same version if the cluster is not in a failed state."

jluk: Thanks @mboersma agree on change

lalo: done

mboersma:
@tjprescott could this get into the next CLI release?

PR #8628, opened 2019-02-23
Changes to have CLI get along with AutoRest4

yugangw-msft:
The deal is to propagate the operation_group along the pipeline explicitly.
Then i did a global find & replace using regular expressions with pattern `(operations_tmpl=.azure\.mgmt\..+\.operations)(\..+_operations)(#.+)(['"])` and replace with `$1$3$4`
      


yugangw-msft:
I looked through the existing profile support from Python SDK's end to see whether or not there are any new things to leverage
1.   [General Profile](https://github.com/Azure/azure-sdk-for-python/blob/master/azure-common/azure/profiles) support. This is more or less a port (half a year ago) from CLI's AZURE_API_PROFILES. It is marked as unstable.
2.   Generated clients have multi-api support e.g. [compute](https://github.com/Azure/azure-sdk-for-python/blob/master/azure-mgmt-compute/azure/mgmt/compute/compute_management_client.py), after cross checked the SDK from the new code-gen, I believe the support stays the same. In a nutshell, the support will only come after you have a client created.

For 1st one, it doesn't help CLI.
For 2nd, to leverage it, CLI can choose to initialize the client first at the very beginning the command loading, and propagate the client across the pipeline like this PR propagates the operation group. This is a more involved change and it introduces a new path to deal with multi-api, with the other one on retrieving model types. I don't think this is the right way as it doesn't simplify.
Other approaches?

Still getting ramped up in the multi-api, but my feeling is this PR might be the best we can do. Regardless of the result, CLI and SDK teams should get on the same page with how CLI's multi-api logic interacts with SDK's, and we can put a test case at the code-gen repository to ensure the promise can be kept for long run.

//CC: @lmazuel @derekbekoe  @tjprescott

derekbekoe:
@yugangw-msft There is a .whl file in your PR. Is that intended?
Also, will it still work for extensions that won't have updated to remove the `_operations` from `operations_tmpl`? i.e. Is this PR backwards compatible?

yugangw-msft:
> There is a .whl file in your PR. Is that intended?
This is intended for testing, but will be removed later. It is a private .whl produced from autorest4

   >  Also, will it still work for extensions that won't have updated to remove the `_operations` from `operations_tmpl`? i.e. Is this PR backwards compatible?

It should be fine, but i will double check.
4 situations:
1. Old SDK with old CLI operations_tmpl, it still works.
2. Old SDK with new CLI operations_tmpl, it still works since the operation module name is really optional
3. New SDK with old CLI operations_tmpl, it is broken but author will realize this when pick up the new SDKs. I believe all cli command modules are pinning to the exact SDK version. CC: @williexu to keep me honest
4. New SDK with new CLI operations_tmpl, this will work

williexu:
@yugangw-msft to respond to 3; even better than pinning the versions, we make extensions vendor their sdks 🥇

yugangw-msft:
I have cross checked 3 extensions per suggestions, `azure-devops`, `azure-iot-cli-extension`, and `azure-batch-cli-extensions`. They are all vendoring the SDKs, so we are good.

PR #8646, opened 2019-02-26
Moving select command_modules into the Azure-CLI package

### LINKED ISSUE #8592: Proposal: Stop Shipping Command Modules as Independent PyPI Packages
Today we ship a series of Python libraries on PyPI that match the pattern `azure-cli-*` which encapsulate the declaration of commands for interacting with individual services. A few examples:
- https://pypi.org/project/azure-cli-appservice/
- https://pypi.org/project/azure-cli-resource/
- https://pypi.org/project/azure-cli-vm/

We then stitch together the set of commands that we want to call the Azure CLI in another top-level PyPI package: https://pypi.org/project/azure-cli Fundamentally, there isn't any real code that ships as part of this package, it just aggregates the list of command modules that we want to include.

We are currently in the middle of discussing how we want to go about locking our dependencies and create a stronger reproducible build story. A lot of that conversation has hinged around just locking our command modules, so that we aren't too opinionated about dependencies when people are installing our package in a virtualenv that isn't exclusively dedicated to our tool. However, it's worth considering: If we only want to lock our command modules, and we will always want to lock to the version that is currently in the same branch in source control, should we be shipping these independently at all?

In a lot of ways, the easiest thing to do would be to ship all of the command modules together inside the "azure-cli" package itself. It would solve the problem we've seen customers run into while trying to install a particular version of the CLI using pip, where we need to distribute to them a  `requirements.txt` file based on their platform and version of Python. It would also mean that there were exactly two well-defined install scenarios: 1. Installing into a foreign virtualenv where we don't have control of our dependencies. 2. Installing into a virtualenv we can call our own, and use modern tools like `pipenv` to lock our dependencies and have a truly reproducible build.

What's more, in some ways, we're essentially already doing this. Consider our [`.msi`](https://github.com/Azure/azure-cli/blob/fe0f256ee877c6f88ff432213115e9a0bf82e97b/build_scripts/windows/scripts/build.cmd#L96-L102), [`.rpm`](https://github.com/Azure/azure-cli/blob/fe0f256ee877c6f88ff432213115e9a0bf82e97b/scripts/release/rpm/azure-cli.spec#L54-L55) and, [`.deb`](https://github.com/Azure/azure-cli/blob/fe0f256ee877c6f88ff432213115e9a0bf82e97b/scripts/release/debian/Dockerfile#L28-L32) packages, which install a locally built wheel of the command modules instead of reaching out to PyPI. This means that taking this approach would just bring our `pip` package into parity as a release channel with our other more platform-specific release channels.

Now, there is at least one downside: Folks who want to install a dramatically minified version of the CLI would have a harder time than they do today. Instead of creating their own shell packages akin to [`azure-cli` ](https://pypi.org/project/azure-cli) from their own repositories and taking dependencies on just the command module libraries they care about, they would have to create a fork of our repository. Truthfully, I don't _think_ there's anybody doing this today. However, even if they are, I'd be excited to see if it actually just opened up a doorway for us to merge those curated packages back into our repository to publish different SKUs of the CLI that have optimized command-sets for different tasks.

Now, all that said, if I've missed something that makes this untenable, I want to hear it! What have I missed? @yugangw-msft @tjprescott @williexu @adewaleo

marstr:
We've been wringing our hands about how to keep these modules in-sync with each other while shipping them in separate packages, without forcing deeper dependency requirements on `pip` installers. While there are many options, the easiest is to just change our criteria and not ship these core modules as separate packages.

This is related to #8592 

---


src/azure-cli/azure/cli/command_modules/role/tests/latest/test_role_commands_thru_mock.py
```diff
@@ -11,16 +11,14 @@
...
-from azure.mgmt.authorization.models import RoleDefinition, RoleAssignmentCreateParameters
+from azure.mgmt.authorization.models import RoleDefinition
 from azure.graphrbac.models import (Application, ServicePrincipal, GraphErrorException,
-                                    ApplicationUpdateParameters, GetObjectsParameters)
+                                    ApplicationUpdateParameters)
 from azure.cli.command_modules.role.custom import (create_role_definition,
                                                    update_role_definition,
                                                    create_service_principal_for_rbac,
                                                    reset_service_principal_credential,
-                                                   update_application, _try_x509_pem,
```
marstr: I need to investigate where these changes came from... not something I did manually.

marstr: Actually, digging into it, it seems like these are things the PyCharm cleaned up for me because they were unused. I'm going to go ahead and assume these changes can stay. (Also low-pri because these are just test package imports.)

src/azure-cli/azure/cli/command_modules/role/_help.py
```diff
@@ -238,6 +238,15 @@
...
 """
+helps['ad user create'] = """
+    type: command
+    short-summary: Create an Azure Active Directory user.
+    parameters:
+        - name: --force-change-password-next-login
+          short-summary: Marks this user as needing to update their password the next time they authenticate.
+        - name: --password
+          short-summary: The password that should be assigned to the user for authentication.
+"""
```
williexu: The linter rule in the past was supressing this rule exception through: https://github.com/Azure/azure-cli/blob/dev/tools/automation/cli_linter/rules/ci_exclusions.yml#L23
Please look at: https://github.com/Azure/azure-cli/blob/dev/tools/automation/cli_linter/util.py#L22

I suspect that moving the modules has bypassed the filtering mechanism used by the linter to exclude modules for certain rule exceptions.

marstr: Let's see if a516f2d handles this :)

williexu: Might also want to re-add one of the failing scenarios to test and make sure the change works :)

marstr: Good idea, I'll add one back. 

Also looks like the change is insufficient because consuming code assumes that each of the command_modules is its own package. I'm curious about how deep that rabbit hole goes, I'll spend a little more time digging.

marstr: Poking around, it looks like our linter exceptions are really tied to the packages, not modules. I opened up #8656 to track this work, but I can do it after this PR is merged.

yugangw-msft:
@marstr, on top of @tjprescott's list, make sure test all installers and adjust related building scripts (should get simpler though). I suppose the public devop CI should cover that, but please double check

marstr:
> @tjprescott says:
Also, please test the potential impact of one big shared HISTORY file by making two PRs that each add lines to that file and see if Git complains about conflicts.

Ran some experiments to see how often this will cause conflicts, seems like if two new services attempt to add their own sections in parallel it will result in a merge conflict. It's worth noting that it is a particularly trivial one to fix though. If two teams modifying existing sections in parallel, Git is able to resolve the differences correctly.

There are some things we could do, like having a change log in each of the modules, and having the release notes just be an aggregate of those. However, I stand by wanting to see it being a problem before we add much complexity here.

marstr:
Thanks for looking into this @adewaleo, it's much appreciated!

> @adewaleo says:
> This [function](https://github.com/Azure/azure-cli/blob/5593aceac52c883eef8e84a5a788d8f4d32401db/src/azure-cli-core/azure/cli/core/util.py#L69-L71) might be different now that the modules have been moved. If not, this could be a breaking change. This method was a way for clients to get the list of installed modules in the CLI. This function is used by the feedback module and core to get version info. It is used by a conversion script, I doubt any extensions use this method.

I believe this will still behave in the desired way, as it is about finding our distributed packages (not modules). After my changes we should just have a significantly reduced set of packages.

> 2\. I generated documentation (with sphinx) for the CLI before and after your changes. And it looks good. The doc_source_map info was different for some entries, which makes sense. I haven't tested the difference with help.yaml or extensions installed.

Woot! Thanks for looking into this. Anything I should watch out for with extensions? I was assuming that they'd be pretty tangential to my changes.

adewaleo:
> Thanks for looking into this @adewaleo, it's much appreciated!
> 
> > @adewaleo says:
> > 1. This [function](https://github.com/Azure/azure-cli/blob/5593aceac52c883eef8e84a5a788d8f4d32401db/src/azure-cli-core/azure/cli/core/util.py#L69-L71) might be different now that the modules have been moved. If not, this could be a breaking change. This method was a way for clients to get the list of installed modules in the CLI. This function is used by the feedback module and core to get version info. It is used by a conversion script, I doubt any extensions use this method.
> 
> I believe this will still behave in the desired way, as it is about finding our distributed packages (not modules). After my changes we should just have a significantly reduced set of packages.

Okay, I see.

> 
> > 2. I generated documentation (with sphinx) for the CLI before and after your changes. And it looks good. The doc_source_map info was different for some entries, which makes sense. I haven't tested the difference with help.yaml or extensions installed.
> 
> Woot! Thanks for looking into this. Anything I should watch out for with extensions? I was assuming that they'd be pretty tangential to my changes.

I can't see any reason why help should be different with extensions. But I can help verify, @marstr.

marstr:
I've confirmed @tjprescott's suspicion that azdev's test discovery doesn't find the moved module's tests. I'm following up on that work now.

marstr:
We chatted offline and agreed to revisit this after //build in May.

PR #8661, opened 2019-02-28
[AKS] handle older kubectl versions in "az aks browse" 

mboersma:
#8408 relied on a new `--address` argument that only appears in `kubectl` v1.13.x and later, so it effectively broke `az aks browse` for most users. This adds fallback logic to call `kubectl` without that argument if need be.

I've tested this with various versions of `kubectl`.

---

### DIFF
--- a/src/command_modules/azure-cli-acs/HISTORY.rst
+++ b/src/command_modules/azure-cli-acs/HISTORY.rst
@@ -3,6 +3,10 @@
 Release History
 ===============
 
+2.3.19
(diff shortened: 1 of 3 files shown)


mboersma:
@tjprescott could this get into the next CLI release? We basically broke `az aks browse` with the last one. 😞

marstr:
I'm not @tjprescott, but I do think this should get in and it looks good to me. I'll go ahead and merge this.

PR #8678, opened 2019-03-04
Update azure-mgmt-rdbms version to 1.7.0

elemount:
---


src/command_modules/azure-cli-rdbms/HISTORY.rst
```diff
@@ -4,6 +4,10 @@ Release History
 ===============
 * Minor fixes: Get default location from resource group when not provided for creating servers and add validation for retention days. 
 
+0.3.8
```
yugangw-msft: The entry above for the new release should be good enough, so you can remove yours

PR #8684, opened 2019-03-05
Direct users to github when unexpected error is raised.

### LINKED ISSUE #8644: Improve generic uncaught error text.
Direct uncaught exceptions (stack traces--by definition a bug) to our repo for issues.

adewaleo:
Closes #8644

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/util.py
+++ b/src/azure-cli-core/azure/cli/core/util.py
@@ -47,7 +47,13 @@ def handle_exception(ex):
             logger.error(ex)
         return 1
 
+    # Otherwise, unhandled exception. Direct users to create an issue.
+    logger.error("The command failed with an unexpected error. Here is the traceback:\n")
     logger.exception(ex)
+    logger.error("\nIf you think this is a bug, please visit: https://github.com/Azure/azure-cli to create an issue."
+                 "\nIf you would like tips for using the CLI, "
+                 "visit: https://github.com/Azure/azure-cli/blob/dev/doc/use_cli_effectively.md.")
+
     return 1
 
 
--- a/src/azure-cli-core/azure/cli/core/util.py
+++ b/src/azure-cli-core/azure/cli/core/util.py
@@ -50,9 +50,8 @@ def handle_exception(ex):
     # Otherwise, unhandled exception. Direct users to create an issue.
     logger.error("The command failed with an unexpected error. Here is the traceback:\n")
     logger.exception(ex)
-    logger.error("\nIf you think this is a bug, please visit: https://github.com/Azure/azure-cli to create an issue."
-                 "\nIf you would like tips for using the CLI, "
-                 "visit: https://github.com/Azure/azure-cli/blob/dev/doc/use_cli_effectively.md.")
+    logger.warning("\nTo open an issue, please visit: https://aka.ms/azure-cli-issues"
+                   "\nIf this command is an extension, please visit: https://aka.ms/azure-cli-extensions-issues")
 
     return 1
 
--- a/src/azure-cli-core/azure/cli/core/util.py
+++ b/src/azure-cli-core/azure/cli/core/util.py
@@ -20,7 +20,7 @@ CLI_PACKAGE_NAME = 'azure-cli'
 COMPONENT_PREFIX = 'azure-cli-'
 
 
-def handle_exception(ex):
+def handle_exception(ex, cli_ctx=None):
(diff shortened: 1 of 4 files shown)


adewaleo:
Text looks like this:
uncaught error info

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -47,7 +47,13 @@ def handle_exception(ex):
             logger.error(ex)
         return 1
 
+    # Otherwise, unhandled exception. Direct users to create an issue.
+    logger.error("The command failed with an unexpected error. Here is the traceback:\n")
     logger.exception(ex)
+    logger.error("\nIf you think this is a bug, please visit: https://github.com/Azure/azure-cli to create an issue."
+                 "\nIf you would like tips for using the CLI, "
```
yugangw-msft: I think directing people to log an issue should be good enough. We can leave out the document reference for a while as its final location is not yet decided.

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -47,7 +47,13 @@ def handle_exception(ex):
             logger.error(ex)
         return 1
 
+    # Otherwise, unhandled exception. Direct users to create an issue.
+    logger.error("The command failed with an unexpected error. Here is the traceback:\n")
     logger.exception(ex)
+    logger.error("\nIf you think this is a bug, please visit: https://github.com/Azure/azure-cli to create an issue."
```
marstr: I'd create an aka.ms link to guide people to our bug tracker. It'll do two things:
1. If we ever change where the product is hosted (because there was a nuclear apocalypse or something) we could update the link, and people with the old version of the product will still be able to file bugs.
1. We can track how many times people are clicking that link and creating issues.

adewaleo: Thanks for the tip will do @marstr

src/azure-cli-core/azure/cli/core/util.py
adewaleo: Makes sense @yugangw-msft

src/azure-cli-core/azure/cli/core/util.py
tjprescott: To make the "If you think this is... blah" more visually distinct, I would suggest using `logger.warning` here.

adewaleo: Will update.

adewaleo:
Warning now looks like this:
image

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -50,9 +50,8 @@ def handle_exception(ex):
     # Otherwise, unhandled exception. Direct users to create an issue.
     logger.error("The command failed with an unexpected error. Here is the traceback:\n")
     logger.exception(ex)
-    logger.error("\nIf you think this is a bug, please visit: https://github.com/Azure/azure-cli to create an issue."
-                 "\nIf you would like tips for using the CLI, "
-                 "visit: https://github.com/Azure/azure-cli/blob/dev/doc/use_cli_effectively.md.")
+    logger.warning("\nTo open an issue, please visit: https://aka.ms/azure-cli-issues"
+                   "\nIf this command is an extension, please visit: https://aka.ms/azure-cli-extensions-issues")
```
tjprescott: So, now my question is... we should know whether the command is an extension or not. It's how we know to display the "this command's behavior is overridden by blah extension" message. We should investigate whether we can access this metadata and simply provide the single correct link.

adewaleo: Done!

Issue link image.

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -48,10 +48,18 @@ def handle_exception(ex):
...
     # Otherwise, unhandled exception. Direct users to create an issue.
+    is_extension = False
+    if cli_ctx:
+        try:
+            if cli_ctx.invocation.data['command_extension_name']:
+                is_extension = True
+        except (AttributeError, KeyError):
+            pass
+
+    issue_url = "https://aka.ms/azcli/ext/issues" if is_extension else "https://aka.ms/azcli/issues"
```
tjprescott: ❤️

src/azure-cli-core/azure/cli/core/util.py
tjprescott: Awesome!

adewaleo:
does this look good with everyone?

PR #8739, opened 2019-03-08
[Cloud] Fix cloud set name case sensitive

### LINKED ISSUE #4663: `az cloud set -n` should not be case sensitive
### Description
While *bash* is case sensitive, and the concept makes sense to those users, I don't see a business or customer value in make this a policy.  As a result, having cloud names be case sensitive only serves to add another error condition.

```
az cloud set -n azurecloud
The cloud 'azurecloud' is not registered.
[1] ~ $ az cloud set -n AzureCloud
Switched active cloud to 'AzureCloud'.
```

limingu:
Fixes #4663
---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/cloud.py
+++ b/src/azure-cli-core/azure/cli/core/cloud.py
@@ -256,8 +256,10 @@ def get_custom_clouds(cli_ctx):
     known_cloud_names = [c.name for c in KNOWN_CLOUDS]
     return [c for c in get_clouds(cli_ctx) if c.name not in known_cloud_names]
 
+
 def _get_cloud_name(cli_ctx, cloud_name):
     return next((x.name for x in get_clouds(cli_ctx) if x.name.lower() == cloud_name.lower()), cloud_name)
+
 
 def get_clouds(cli_ctx):
     clouds = []
--- a/src/azure-cli-core/azure/cli/core/cloud.py
+++ b/src/azure-cli-core/azure/cli/core/cloud.py
@@ -256,6 +256,8 @@ def get_custom_clouds(cli_ctx):
     known_cloud_names = [c.name for c in KNOWN_CLOUDS]
     return [c for c in get_clouds(cli_ctx) if c.name not in known_cloud_names]
 
+def get_cloud_name(cli_ctx, cloud_name):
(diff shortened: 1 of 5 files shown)


src/azure-cli-core/azure/cli/core/cloud.py
```diff
@@ -256,6 +256,8 @@ def get_custom_clouds(cli_ctx):
     known_cloud_names = [c.name for c in KNOWN_CLOUDS]
     return [c for c in get_clouds(cli_ctx) if c.name not in known_cloud_names]
 
+def get_cloud_name(cli_ctx, cloud_name):
+    return next((x.name for x in get_clouds(cli_ctx) if x.name.lower() == cloud_name.lower()), cloud_name)
```
tjprescott: Let's make this private so as not to expand the public interface of core. `_get_cloud_name`

src/command_modules/azure-cli-cloud/azure/cli/command_modules/cloud/tests/latest/test_cloud.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def test_cloud_set_azureusgovernment(self):
+        self.cmd('az cloud set -n azureusgovernment')
+
+    def test_cloud_set_azuregermancloud(self):
+        self.cmd('az cloud set -n azuregermancloud')
+
+    def test_cloud_set_unregistered_cloud_name(self):
+        with pytest.raises(Exception, match =  "The cloud 'azCloud' is not registered.") :
+            self.cmd('az cloud set -n azCloud').get_output_in_json()
```
tjprescott: I'd recommend following the syntax here:
https://github.com/Azure/azure-cli/blob/dev/doc/authoring_tests.md#sample-9-assert-specific-error-occurs

This avoids needing a new dependency on pytest.

src/azure-cli-core/azure/cli/core/cloud.py
limingu: Updated

src/command_modules/azure-cli-cloud/azure/cli/command_modules/cloud/tests/latest/test_cloud.py
limingu: Updated

Issue #8715, opened 2019-03-07
Enabling network rule on storage account using az cli disables encryption on storage 


sajithvasu:
Hello,
az cli removes existing encryption on storage account when network firewall rules are added.  This is a consistent behavior with mostly all the version of az cli.  I have tested on windows and linux. 

Steps to reproduce the behavior.
a) Create Storage account
b) Enable encryption on storage by using encryption key created on keyvault.
c) run storage account network command.
az storage account update --resource-group <rg_name> --name <storageaccountname> --default-action Deny
az storage account network-rule add --resource-group <rg_name> --account-name <storageaccountname> --ip-address x.x.x.x/x

williexu:
@sajithvasu the issue is actually with the `storage account update` command, not actually the adding of network rules. 
The update command had some very faulty logic that was ignoring encryption parameters in various scenarios and replacing them with null values with a `Microsoft.Storage` key source. https://github.com/Azure/azure-cli/pull/8740 fixes this problem.

PR #8740, opened 2019-03-08
[Storage] Storage encryption

williexu:
---
Fixes: https://github.com/Azure/azure-cli/issues/8715


src/command_modules/azure-cli-storage/azure/cli/command_modules/storage/_params.py
```diff
@@ -127,7 +127,7 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals, too-many-statem
         c.extra('encryption_key_vault', help='The Uri of the KeyVault')
         c.extra('encryption_key_version', help='The version of the KeyVault key')
         c.argument('encryption_key_source',
-                   arg_type=get_enum_type(['Microsoft.Storage', 'Microsoft.Keyvault'], 'Microsoft.Storage'),
+                   arg_type=get_enum_type(['Microsoft.Storage', 'Microsoft.Keyvault']),
```
tjprescott: Intentionally removing the default here?

williexu: Yes, having the default for the parameter will overwrite the previous value when sent along the service. This is part of the issue being solved.

Issue #8731, opened 2019-03-07
az aks create -g RG -n AKS --node-count 2 --generate-ssh-keys : Error - The password must contain at least 1 special character. paramName: PasswordCredentials, paramValue:


SriHarsha001:
Following this article to create an AKS cluster in Azure CLI
https://docs.microsoft.com/en-us/cli/azure/aks?view=azure-cli-latest#az-aks-create

az aks create -g MyResourceGroup -n MyManagedCluster --node-count 2
An RSA key file or key value must be supplied to SSH Key Value. You can use --generate-ssh-keys to let CLI generate one for you

az aks create -g MyResourceGroup -n MyManagedCluster --node-count 2 --generate-ssh-keys
**The password must contain at least 1 special character. paramName: PasswordCredentials, paramValue: , objectType: Microsoft.Online.DirectoryServices.Application**

az --version
azure-cli                         2.0.59
....
....
Python location 'C:\Program Files (x86)\Microsoft SDKs\Azure\CLI2\python.exe'
Extensions directory 'C:\Users\sharsha\.azure\cliextensions'
Python (Windows) 3.6.6 (v3.6.6:4cf1f54eb7, Jun 27 2018, 02:47:15) [MSC v.1900 32 bit (Intel)]
Legal docs and information: aka.ms/AzureCliLegal
Your CLI is up-to-date.

I get the same error using cloud shell on Azure Portal as well.

akshita31:
I am also getting the same error when using the Azure Cloud Shell 
`az aks create -l eastus -n testCluster -g volunteer --generate-ssh-keys -k 1.12.4 -c 1`

SriHarsha001:
Please try the below work around - 

Create a new SP : 
	az ad sp create-for-rbac --skip-assignment

Create AKS cluster by providing credentials created above :  
	az aks create --name AKS --resource-group RG --node-count Int --generate-ssh-keys --service-principal AppID --client-secret Pwd

yevster:
When I try the workaround, I get this error:

> The request did not have a subscription or a valid tenant level resource provider.

Confirmed that `az account show` lists the correct subscription. Adding the `subscriptionId` parameter or the `aad-tenant-id` parameter had no effect. 

(CLI version 2.0.60, Ubuntu 18.04)

DivineOps:
@yugangw-msft 
I am still getting this issue running the command in cloudshell
[...]

PR #8741, opened 2019-03-08
[AKS] update SP secret to include special characters for aad

GaneshaThirumurthi:
Fixes: https://github.com/Azure/azure-cli/issues/8731

AAD has updated SP secrets to require special characters when the tenant is a Microsoft tenant. 

- A special character is added at random
- I tested creating service principals with each of these secrets and none of them throw an error on creation, however, I don't have an official list of secret characters

---

### DIFF
--- a/src/command_modules/azure-cli-acs/azure/cli/command_modules/acs/custom.py
+++ b/src/command_modules/azure-cli-acs/azure/cli/command_modules/acs/custom.py
@@ -2220,9 +2220,7 @@ def _ensure_aks_service_principal(cli_ctx,
         else:
             # Nothing to load, make one.
             if not client_secret:
-                special_chars = '!#$%&*-+_.:;<>=?@][^}{|~)('
-                special_char = special_chars[ord(os.urandom(1)) % len(special_chars)]
-                client_secret = binascii.b2a_hex(os.urandom(10)).decode('utf-8') + special_char
+                client_secret = _create_client_secret()
             salt = binascii.b2a_hex(os.urandom(3)).decode('utf-8')
             url = 'https://{}.{}.{}.cloudapp.azure.com'.format(salt, dns_name_prefix, location)
 
@@ -2249,7 +2247,7 @@ def _ensure_osa_aad(cli_ctx,
     rbac_client = get_graph_rbac_management_client(cli_ctx)
     if not aad_client_app_id:
         if not aad_client_app_secret and update:
-            aad_client_app_secret = binascii.b2a_hex(os.urandom(10)).decode('utf-8')
+            aad_client_app_secret = _create_client_secret()
         reply_url = 'https://{}/oauth2callback/Azure%20AD'.format(identifier)
 
         # Delegate Sign In and Read User Profile permissions on Windows Azure Active Directory API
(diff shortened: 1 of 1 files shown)


src/command_modules/azure-cli-acs/azure/cli/command_modules/acs/custom.py
```diff
@@ -2220,7 +2220,9 @@ def _ensure_aks_service_principal(cli_ctx,
         else:
             # Nothing to load, make one.
             if not client_secret:
-                client_secret = binascii.b2a_hex(os.urandom(10)).decode('utf-8')
```
mboersma: Maybe add a comment above this like

```
# Add a special character to satisfy AAD SP secret requirements
```

src/command_modules/azure-cli-acs/azure/cli/command_modules/acs/custom.py
```diff
@@ -2312,7 +2314,9 @@ def _ensure_service_principal(cli_ctx,
...
+                special_chars = '!#$%&*-+_.:;<>=?@][^}{|~)('
```
mboersma: Since these three lines are duplicated, maybe make this into a `_create_client_secret()` function?

src/command_modules/azure-cli-acs/azure/cli/command_modules/acs/custom.py
GaneshaThirumurthi: Good idea, will do

src/command_modules/azure-cli-acs/azure/cli/command_modules/acs/custom.py
GaneshaThirumurthi: I am open to making it into a function. However, right now the duplication is across the acs and the aks functions for ensuring sp. Those two functions have an incredible amount of overlap but I assumed there was a logical reason for keeping duplicated code. My guess (not validated) is that since we are retiring acs in 2020 we want to be able to easily remove that code.

Either way, I am open to making it a function but eventually it might only have only one caller

mboersma: I think the real reason for so much duplication is that we haven't paid attention to where we could do these kind of refactorings, we've just added AKS features and not looked back. ACS is deprecated but there's no reason we can't share client code with AKS (at least for another ten months).

But it's just a suggestion—the PR looks good to me either way.

julienstroheker:
@mboersma @GaneshaThirumurthi can you also this fix for OSA ?

https://github.com/Azure/azure-cli/blob/f20c6acecd4c2ffdfd74baf4f07755002d0e7d04/src/command_modules/azure-cli-acs/azure/cli/command_modules/acs/custom.py#L2223

Thanks a lot !

src/command_modules/azure-cli-acs/azure/cli/command_modules/acs/custom.py
GaneshaThirumurthi: Ahh makes sense. I think it is a good idea - it is now fixed

PR #8744, opened 2019-03-09
Update Homebrew Formula to install in Virtualenv

### LINKED ISSUE #8689: brew install into a virtualenv
Our homebrew formula installs `az` using system python without a virtual environment. This is causing issues for folks who have existing non-hombrew python installs, or use system Python for their projects.

We should isolate the CLI into its own virtual environment, [the way the bash install script does](https://github.com/Azure/azure-cli/blob/971ce078888f428bfd3799475dbdb1bf41a3e7cf/scripts/curl_install_pypi/install.py#L355). It'll also allow us to fix this logic: https://github.com/Azure/azure-cli/blob/971ce078888f428bfd3799475dbdb1bf41a3e7cf/scripts/release/homebrew/docker/formula_template.txt#L58-L66

Which is getting confused on systems where python3.x is available, but not where the azure-cli was installed. (Folks setting it up in a virtualenv by themselves.)

marstr:
Fixes #8689

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -196,7 +196,7 @@ jobs:
   displayName: Build Homebrew Formula
 
   dependsOn: BuildPythonWheel
-  condition: succeeded()
+  condition: and(succeeded(), in(variables['Build.Reason'], 'IndividualCI', 'BatchedCI', 'Manual'))
   pool:
     vmImage: 'ubuntu-16.04'
   steps:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -196,7 +196,7 @@ jobs:
   displayName: Build Homebrew Formula
 
   dependsOn: BuildPythonWheel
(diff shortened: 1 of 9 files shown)


marstr:
There is a known issue I'm still working around, where for some reason two versions of [wheel](https://pypi.org/project/wheel) are installed simultaneously after the installer finishes. However, I thought I'd get this open now to allow review/commenting on what I do have at the moment.

azure-pipelines.yml
```diff
@@ -196,7 +196,7 @@ jobs:
   displayName: Build Homebrew Formula
 
   dependsOn: BuildPythonWheel
-  condition: and(succeeded(), in(variables['Build.Reason'], 'IndividualCI', 'BatchedCI', 'Manual'))
+  condition: succeeded()
```
marstr: Before this PR is merged, I will revert this change so that homebrew builds are not done as part of PR validation, only CI.

marstr: Updated in 2f9cc5a

PR #8747, opened 2019-03-09
Add credential support for custom registries + Auth mode flag

shahzzzam:
---
This PR:
-Adds credential command group to add, update, remove and list Credentials for a task.
-Adds support for auth-mode flag for Source registry.

- [ x] The PR has modified HISTORY.rst describing any customer-facing, functional changes. Note that this does not include changes only to help content. (see [Modifying change log](https://github.com/Azure/azure-cli/tree/master/doc/authoring_command_modules#modify-change-log)).


src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -349,12 +355,146 @@ def acr_task_update(cmd,  # pylint: disable=too-many-locals
...
+                            registry_name,
+                            login_server=None,
+                            username=None,
+                            password=None,
+                            resource_group_name=None):
+    _, resource_group_name = validate_managed_registry(
+        cmd, registry_name, resource_group_name, TASK_NOT_SUPPORTED)
+
+    if not login_server or not username or not password:
+        raise CLIError("--login-server, --username and --password are required to add a credential")
```
djyou: Make the three params required and remove this check.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -349,12 +355,146 @@ def acr_task_update(cmd,  # pylint: disable=too-many-locals
...
+
+    TaskUpdateParameters, PlatformUpdateParameters, AgentProperties, TriggerUpdateParameters = cmd.get_models(
+        'TaskUpdateParameters', 'PlatformUpdateParameters', 'AgentProperties', 'TriggerUpdateParameters')
+    taskUpdateParameters = TaskUpdateParameters(
+        status=None,
```
djyou: Remove all the None assignment.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -349,12 +355,146 @@ def acr_task_update(cmd,  # pylint: disable=too-many-locals
...
+            login_server=login_server,
+            username=username,
+            password=password
+        )
+    )
+
+    return client.update(resource_group_name, registry_name, task_name, taskUpdateParameters)
+
+
+def acr_task_credential_update(cmd,
```
djyou: How is this different from `acr_task_credential_add`?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
```diff
@@ -375,6 +375,49 @@ def get_validate_platform(cmd, os_type, platform):
...
+            'CustomRegistryCredentials',
+            'SecretObject',
+            'SecretObjectType'
+        )
+
+        # if empty username and password, then send a Null object to have it removed
+        custom_reg_credential = None
+        if username and password:
+            custom_reg_credential = CustomRegistryCredentials(
+                user_name=SecretObject(
```
djyou: Is `user_name` modeled in the SDK as two words?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
```diff
@@ -375,6 +375,49 @@ def get_validate_platform(cmd, os_type, platform):
...
+
+    custom_registries = {}
+    if login_server is not None:
+        CustomRegistryCredentials, SecretObject, SecretObjectType = cmd.get_models(
+            'CustomRegistryCredentials',
+            'SecretObject',
+            'SecretObjectType'
+        )
+
+        # if empty username and password, then send a Null object to have it removed
```
djyou: I think it might better to error out if only one of `username` and `password` is provided.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
```diff
@@ -375,6 +375,49 @@ def get_validate_platform(cmd, os_type, platform):
...
+    :param str username: The username for custom registry
+    :param str password: The password for custom registry
+    """
+    source_registry_credentials = None
+    if auth_mode is not None:
+        SourceRegistryCredentials = cmd.get_models('SourceRegistryCredentials')
+        source_registry_credentials = SourceRegistryCredentials(login_mode=auth_mode)
+
+    custom_registries = {}
+    if login_server is not None:
```
djyou: Do you want to check for null or empty?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
```diff
@@ -186,6 +187,12 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
     with self.argument_context('acr task create') as c:
         c.argument('task_name', completer=None)
 
+    with self.argument_context('acr task credential') as c:
+        # Custom registry credentials
+        c.argument('login_server', help="The login server of the custom registry. For instance, 'myregistry.azurecr.io'")
```
djyou: nit: add a period to the help message

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
```diff
@@ -573,6 +573,47 @@
...
+short-summary: Update the credential for a task.
+examples:
+  - name: Update the credential for a task
+    text: >
+        az acr task credential update -n taskname -r registryname --login-server myregistry.docker.io --username 'myusername2' --password 'mysecret'
+"""
+
+helps['acr task credential show'] = """
+type: command
+short-summary: Display the Credentials for task.
```
djyou: credentials?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
```diff
@@ -573,6 +573,47 @@
...
+helps['acr task credential'] = """
+type: group
+short-summary: Credential for a task
+"""
+
+helps['acr task credential add'] = """
+type: command
+short-summary: Add a credential to the task
+examples:
+  - name: Enable the system-assigned credential to an existing task. This will replace all existing user-assigned identities.
```
northtyphoon: It needs to updated.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
```diff
@@ -573,6 +573,47 @@
...
+    text: >
+        az acr task credential add -n taskname -r registryname --login-server myregistry.docker.io --username 'myusername' --password 'mysecret'
+"""
+
+helps['acr task credential remove'] = """
+type: command
+short-summary: Remove credential for a task.
+examples:
+  - name: Remove a user-assigned credential from a task.
```
northtyphoon: registry login credential

`user-assigned` is for msi feature

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/commands.py
```diff
@@ -174,6 +174,10 @@ def load_command_table(self, _):  # pylint: disable=too-many-statements
         g.command('list', 'acr_task_list')
         g.command('delete', 'acr_task_delete')
         g.command('update', 'acr_task_update')
+        g.command('credential add', 'acr_task_credential_add')
+        g.command('credential remove', 'acr_task_credential_remove')
```
northtyphoon: I see in acr commands, we use `delete` instead of `remove`

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -349,12 +355,146 @@ def acr_task_update(cmd,  # pylint: disable=too-many-locals
...
+    _, resource_group_name = validate_managed_registry(
+        cmd, registry_name, resource_group_name, TASK_NOT_SUPPORTED)
+
+    if not login_server:
+        raise CLIError("--login-server is required to remove a credential")
+
+    TaskUpdateParameters, PlatformUpdateParameters, AgentProperties, TriggerUpdateParameters = cmd.get_models(
+        'TaskUpdateParameters', 'PlatformUpdateParameters', 'AgentProperties', 'TriggerUpdateParameters')
+    taskUpdateParameters = TaskUpdateParameters(
+        status=None,
```
northtyphoon: Do we have to manually set all other properties manually to None one by one? What if I miss one property? Will that cause any grief?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
mnltejaswini: Can we elaborate like adds a custom registry login credential to the task

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
```diff
@@ -573,6 +573,47 @@
...
+        az acr task credential update -n taskname -r registryname --login-server myregistry.docker.io --username 'myusername2' --password 'mysecret'
+"""
+
+helps['acr task credential show'] = """
+type: command
+short-summary: Display the Credentials for task.
+examples:
+  - name: Display the Credentials for a task.
+    text: >
+        az acr task credential show -n taskname -r registryname
```
mnltejaswini: Does this show only registry credentials or entire credential property for a task?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
```diff
@@ -186,6 +187,12 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
...
+        c.argument('username', options_list=['--username', '-u'], help='The username used to log into a custom container registry')
+        c.argument('password', options_list=['--password', '-p'], help='The password used to log into a custom container registry')
```
mnltejaswini: nit: used to login to the registry,

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -349,12 +355,146 @@ def acr_task_update(cmd,  # pylint: disable=too-many-locals
...
+        raise CLIError("--login-server, --username and --password are required to add a credential")
+
+    TaskUpdateParameters, PlatformUpdateParameters, AgentProperties, TriggerUpdateParameters = cmd.get_models(
+        'TaskUpdateParameters', 'PlatformUpdateParameters', 'AgentProperties', 'TriggerUpdateParameters')
+    taskUpdateParameters = TaskUpdateParameters(
+        status=None,
+        platform=PlatformUpdateParameters(
+            os=None
+        ),
+        agent_configuration=AgentProperties(
```
mnltejaswini: Is it CLI requirement that you need to define all Null parameters? @djyou

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
mnltejaswini: +1 to @djyou  comment, credential add can be add or update , from the server, if the same credential is passed , it will be updated.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -349,12 +355,146 @@ def acr_task_update(cmd,  # pylint: disable=too-many-locals
...
 
+def acr_task_credential_show(cmd,
+                             client,
+                             task_name,
+                             registry_name,
+                             resource_group_name=None):
+    _, resource_group_name = validate_managed_registry(
+        cmd, registry_name, resource_group_name, TASK_NOT_SUPPORTED)
+
+    return client.get_details(resource_group_name, registry_name, task_name).credentials
```
mnltejaswini: if credential command only manages custom registry login credentials, do you just need to show only them? not the entire credential object which has source registry auth mode. Isn't the auth mode not showed as part of task show ?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
```diff
@@ -375,6 +375,49 @@ def get_validate_platform(cmd, os_type, platform):
...
+    :param str login_server: The login server of custom registry
+    :param str username: The username for custom registry
+    :param str password: The password for custom registry
+    """
+    source_registry_credentials = None
+    if auth_mode is not None:
+        SourceRegistryCredentials = cmd.get_models('SourceRegistryCredentials')
+        source_registry_credentials = SourceRegistryCredentials(login_mode=auth_mode)
+
+    custom_registries = {}
```
mnltejaswini: can this be initialized to None?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/commands.py
djyou: I do see in some cases when working with a collection we use `add` entity to or `remove` entity from the collection, such as our own `az acr network-rule remove` command that follows the pattern from everyone else.

Also see https://english.stackexchange.com/a/52509 as an informal reference.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
shahzzzam: `null`. Cases where `login_server` is None is where the user only set the `auth_mode` to a value. Cases where we deal with credentials (`login_server` is never empty because it is then a required property).

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
shahzzzam: The error is handled by the CLI itself since `username` and `password` is a required property for `credential add/update`. For `delete`, the `username` and `password` is always `None`

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
shahzzzam: yes

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/commands.py
```diff
@@ -174,6 +174,10 @@ def load_command_table(self, _):  # pylint: disable=too-many-statements
         g.command('list', 'acr_task_list')
         g.command('delete', 'acr_task_delete')
         g.command('update', 'acr_task_update')
+        g.command('credential add', 'acr_task_credential_add')
+        g.command('credential update', 'acr_task_credential_add')
+        g.command('credential delete', 'acr_task_credential_delete')
```
djyou: I think `remove` is the right word here, but will defer to @northtyphoon

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/run.py
```diff
@@ -79,6 +80,10 @@ def acr_run(cmd,  # pylint: disable=too-many-locals
             os=platform_os,
             architecture=platform_arch,
             variant=platform_variant
+        ),
+        credentials=get_custom_registry_credentials(
+            cmd,
```
djyou: nit: `cmd=cmd,`

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
djyou: nit: just a minor concern since this is a public helper method with an implicit assumption.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
djyou: Could you make sure this is fixed in the next api-version? /cc @mnltejaswini

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
```diff
@@ -186,6 +187,18 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
     with self.argument_context('acr task create') as c:
         c.argument('task_name', completer=None)
 
+    for scope in ['acr task credential add', 'acr task credential update']:
+        with self.argument_context(scope) as c:
+            # Custom registry credentials
+            c.argument('login_server', help="The login server of the custom registry. For instance, 'myregistry.azurecr.io'.", required=True)
+            c.argument('username', options_list=['--username', '-u'], help='The username to login to the registry.', required=True)
```
djyou: nit: login to which registry?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
djyou: --login-server ""

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -349,12 +355,72 @@ def acr_task_update(cmd,  # pylint: disable=too-many-locals
...
+def acr_task_credential_list(cmd,
+                             client,
+                             task_name,
+                             registry_name,
+                             resource_group_name=None):
+    _, resource_group_name = validate_managed_registry(
+        cmd, registry_name, resource_group_name, TASK_NOT_SUPPORTED)
+
+    resp = client.get_details(resource_group_name, registry_name, task_name).credentials
+    return None if resp is None else resp.custom_registries
```
djyou: Is `custom_registries` a dict? If so, `return {}`, also why is the server returning `null` instead of empty dict? /cc @mnltejaswini

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -349,12 +355,72 @@ def acr_task_update(cmd,  # pylint: disable=too-many-locals
...
+
+    TaskUpdateParameters = cmd.get_models('TaskUpdateParameters')
+    taskUpdateParameters = TaskUpdateParameters(
+        credentials=get_custom_registry_credentials(
+            cmd,
+            login_server=login_server
+        )
+    )
+
+    return client.update(resource_group_name, registry_name, task_name, taskUpdateParameters)
```
djyou: The add/delete methods are returning a different obj than the list method, is this by design?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
shahzzzam: The server returns `null` if there are no credentials. I think in general in az commands if you list anything and it does not have anything, you don't get any response (`None`).

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/commands.py
shahzzzam: According to the English stackexchange link, `delete` makes sense here because we don't store that credential once they delete it. 
There are other commands in the `acr` that use `delete` except for `acr network-rule remove`

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
shahzzzam: Yes, the `client.update` method in this case returns a `msrest.polling.poller.LROPoller` object. I cannot extract out `credentials` from that.

While `list` returns a `azure.mgmt.containerregistry.v2018_09_01.models.task_py3.Task` object.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
shahzzzam: cc @mnltejaswini @northtyphoon

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
shahzzzam: cc @mnltejaswini @northtyphoon

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
shahzzzam: created workitem

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
```diff
@@ -66,6 +66,7 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('os_type', options_list=['--os'], help='The operating system type required for the build.', arg_type=get_enum_type(OsType), deprecate_info=c.deprecate(redirect='platform', hide=True))
         c.argument('platform', help="The platform where build/task is run, Eg, 'windows' and 'linux'. When it's used in build commands, it also can be specified in 'os/arch/variant' format for the resulting image. Eg, linux/arm/v7. The 'arch' and 'variant' parts are optional.")
         c.argument('target', help='The name of the target build stage.')
+        c.argument('auth_mode', help='Auth mode of the source registry.', arg_type=get_enum_type(SourceRegistryLoginMode))
```
djyou: The auth mode? /cc @jaysterp

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
```diff
@@ -186,6 +187,17 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
...
+    for scope in ['acr task credential add', 'acr task credential update']:
+        with self.argument_context(scope) as c:
+            # Custom registry credentials
+            c.argument('login_server', help="The login server of the custom registry. For instance, 'myregistry.azurecr.io'.", required=True)
+            c.argument('username', options_list=['--username', '-u'], help='The username to login to the custom registry.', required=True)
+            c.argument('password', options_list=['--password', '-p'], help='The password to login to the custom registry.', required=True)
+
+    with self.argument_context('acr task credential remove') as c:
+        # Custom registry credentials
+        c.argument('login_server', help="The login server of the custom registry. For instance, 'myregistry.azurecr.io'.", required=True)
```
djyou: Can the three new params be grouped under `acr task credential` so `login_server` doesn't need to be duplicated?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
```diff
@@ -375,6 +375,54 @@ def get_validate_platform(cmd, os_type, platform):
...
 
+def get_custom_registry_credentials(cmd, auth_mode=None, login_server=None, username=None, password=None):
+    """Get the credential object from the input
+    :param str auth_mode: The login mode for the source registry
+    :param str login_server: The login server of custom registry
+    :param str username: The username for custom registry
+    :param str password: The password for custom registry
+    """
+    source_registry_credentials = None
+    if auth_mode is not None:
```
djyou: Same question, null or empty?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/commands.py
```diff
@@ -174,6 +174,10 @@ def load_command_table(self, _):  # pylint: disable=too-many-statements
         g.command('list', 'acr_task_list')
         g.command('delete', 'acr_task_delete')
         g.command('update', 'acr_task_update')
+        g.command('credential add', 'acr_task_credential_add')
+        g.command('credential update', 'acr_task_credential_add')
```
djyou: I wonder if we need this update at all. If we have an update, should we block adding an existing credential, since it will `update` it?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -349,12 +355,80 @@ def acr_task_update(cmd,  # pylint: disable=too-many-locals
...
+def acr_task_credential_list(cmd,
+                             client,
+                             task_name,
+                             registry_name,
+                             resource_group_name=None):
+    _, resource_group_name = validate_managed_registry(
+        cmd, registry_name, resource_group_name, TASK_NOT_SUPPORTED)
+
+    resp = client.get_details(resource_group_name, registry_name, task_name).credentials
+    return {} if not resp else resp.custom_registries
```
djyou: @mnltejaswini Why are we returning null for empty dict?

shahzzzam: @djyou I created a workitem for this and we will return empty dict. This code handles both null and {}

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
shahzzzam: `acr task credential list` does not need --login-server

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/commands.py
shahzzzam: defer to @northtyphoon

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
djyou: It doesn't show unless the command method has the parameter.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/commands.py
northtyphoon: add new credential
update existing credential

what is current behavior? AddOrUpdate?

shahzzzam:
@tjprescott Please review/merge.

PR #8784, opened 2019-03-13
[vm] Added image update. Added examples.

### LINKED ISSUE #8398: az image: allow to add tags to existing image
It's not possible to add a tag to an existing image using azure-cli. The "image update" command doesn't exist in azure-cli-2.0.56-1.el7

az image update --name $IMAGE_NAME --resource-group $RESOURCE_GROUP --set "tags.$MY_TAG=$VALUE"

**Additional context**
"az image create" allows to add tags when creating a new image, however this solution doesn't work for me, since I need to be able to tag both new and existing images.

adewaleo:
Closes #8398 

---

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,6 +1,8 @@
 # See for instructions on this file https://help.github.com/articles/about-codeowners/
 
-*help.py @sptramer
+*help.py @sptramer @adewaleo
+*help.yaml @sptramer @adewaleo
+*help.yml  @sptramer @adewaleo
 
 /tools/ @tjprescott
 /scripts/ @tjprescott
@@ -28,6 +30,6 @@
 /src/command_modules/azure-cli-storage/ @williexu
 /src/command_modules/azure-cli-servicefabric/ @QingChenmsft
 /src/command_modules/azure-cli-sql/ @jaredmoo
-/src/command_modules/azure-cli-vm/ @yugangw-msft
+/src/command_modules/azure-cli-vm/ @yugangw-msft @adewaleo
 /src/command_modules/azure-cli-eventhubs/ @v-ajnava
 /src/command_modules/azure-cli-servicebus/ @v-ajnava
--- a/src/command_modules/azure-cli-vm/HISTORY.rst
+++ b/src/command_modules/azure-cli-vm/HISTORY.rst
@@ -2,6 +2,10 @@
 
 Release History
 ===============
+2.2.17
+++++++
+* Added `image update` command.
+
 2.2.16
 ++++++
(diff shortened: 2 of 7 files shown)


src/command_modules/azure-cli-vm/azure/cli/command_modules/vm/commands.py
```diff
@@ -146,6 +146,7 @@ def load_command_table(self, _):
         g.custom_command('list', 'list_images')
         g.show_command('show', 'get')
         g.command('delete', 'delete')
+        g.generic_update_command('update')
```
tjprescott: Why are you not exposing any convenience arguments?

src/command_modules/azure-cli-vm/HISTORY.rst
```diff
@@ -2,6 +2,10 @@
 
 Release History
 ===============
+2.2.17
```
marstr: Does adding a new command call for a minor bump instead of a patch bump? i.e. should this be `2.3.0`?

tjprescott: We've never required that, no.

src/command_modules/azure-cli-vm/azure/cli/command_modules/vm/_help.py
```diff
@@ -1520,24 +1520,34 @@
     short-summary: List custom VM images.
 """
 
+helps['image update'] = """
+    type: command
+    short-summary: Update custom VM images.
+    examples:
+        - name: Add or update a tag.
+          text: az image update -n ImageName -g ResourceGroup --set tags.tagName=tagValue
```
tjprescott: Seems very odd to have an example for how to set tags with generic update when you could easily add the `--tags` convenience argument, which would be more consistent with nearly all other updates in the CLI.

src/command_modules/azure-cli-vm/azure/cli/command_modules/vm/commands.py
adewaleo: None in particular came to mind, I will add tags as you've requested below. I didn't bother adding it, as vm update does not add `--tags`

.github/CODEOWNERS
```diff
@@ -28,6 +30,6 @@
 /src/command_modules/azure-cli-storage/ @williexu
 /src/command_modules/azure-cli-servicefabric/ @QingChenmsft
 /src/command_modules/azure-cli-sql/ @jaredmoo
-/src/command_modules/azure-cli-vm/ @yugangw-msft
+/src/command_modules/azure-cli-vm/ @yugangw-msft @adewaleo
```
adewaleo: What are we going to do about storage.

cc: @williexu @yugangw-msft

PR #8795, opened 2019-03-14
[Botservice] Switched default version to v4 from v3

mdrichardson:
Fixes Microsoft/botbuilder-tools#1016

Per issue, when using `az bot create...` without supplying a version number, it defaults to `v3`. This makes it default to `v4`.

Test results:

image

Recommendations:

[botservice]
Reviewer: @stevengum 	

---

### DIFF
--- a/src/command_modules/azure-cli-botservice/azure/cli/command_modules/botservice/custom.py
+++ b/src/command_modules/azure-cli-botservice/azure/cli/command_modules/botservice/custom.py
@@ -31,7 +31,7 @@ logger = get_logger(__name__)
 def create(cmd, client, resource_group_name, resource_name, kind, description=None, display_name=None,  # pylint: disable=too-many-locals
            endpoint=None, msa_app_id=None, password=None, tags=None, storageAccountName=None,
            location='Central US', sku_name='F0', appInsightsLocation='South Central US',
-           language='Csharp', version='v3'):
+           language='Csharp', version='v4'):
     """Create a WebApp, Function, or Channels Registration Bot on Azure.
 
     This method is directly called via "bot create"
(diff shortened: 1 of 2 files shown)


stevengum:
@mdrichardson this is a breaking change, if we shipped this it would break users who don’t explicitly ask for v3 bots via `az bot create -v v3`.

 We’ll be addressing it in a later, new major version of `az bot`. Thanks for giving this a go but the SDK team needs to give it more discussion before this change is shipped.

src/command_modules/azure-cli-botservice/azure/cli/command_modules/botservice/custom.py
```diff
@@ -31,7 +31,7 @@
 def create(cmd, client, resource_group_name, resource_name, kind, description=None, display_name=None,  # pylint: disable=too-many-locals
            endpoint=None, msa_app_id=None, password=None, tags=None, storageAccountName=None,
            location='Central US', sku_name='F0', appInsightsLocation='South Central US',
-           language='Csharp', version='v3'):
```
tjprescott: Is this a breaking change?  Generally, changing defaults is considered as such. Additionally, if you change the default to None will the service fail or will it choose one?  If it chooses one, you may wish to do that instead so you don't need to update your CLI code whenever the service default changes.

mdrichardson: @stevengum 	Per the issue, users *must* specify `-v v4` if they want a v4 bot. Wouldn't it make more sense to default to v4 unless the user specifies `-v v3` (which this still allows, I believe...unless there's an `msbot` command it breaks, maybe)? Regardless, if this is being addressed later, should we just close this PR and the issue?

stevengum: @mdrichardson If I create a script that calls `az bot create -n TestBot -k webapp --location westus` it creates a v3 bot. I now expect that I get a v3 bot when I call my script.

If I update Azure CLI and take a version of the CLI with this PR's changes, my old script now creates a v4 bot. This is a breaking change because my previous script no longer works the way I expect it to.

Similarly for `az bot publish`, the `--version` arg changes the way we prepare to publish a bot. Changing the default would break users.

I agree with the original issue, it _should_ default to v4. However we shouldn't take it at this time and this PR should be closed. The issue should be retained to outwardly show that we're aware of it, and I've updated the issue to show that it's in need of a sprint.

tjprescott:
Based on @stevengum's comments, I will close this PR for now.

Issue #8310, opened 2019-01-21
az app permission list --id <app id> does not work when app registration was created with azure cli.


The-First-Tiger:
**Describe the bug**
`az app permission list --id <app id>` does not work when app registration was created with azure cli.
Adding API Permissions manually in the portal and granting access makes `az app permission` working again.

**To Reproduce**
az ad app create --display-name test --homepage https://test --reply-urls https://test/auth --identifier-uris https://test
`az app permission list --id <app id>`

**Expected behavior**
`az app permission list --id <app id>` returns permission list.

**Environment summary**
macOS Mojave 10.14.3
PowerShell 6.0.2

azure-cli (2.0.55)

acr (2.1.12)
acs (2.3.14)
advisor (2.0.0)
ams (0.3.2)
appservice (0.2.11)
backup (1.2.1)
batch (3.4.1)
batchai (0.4.6)
billing (0.2.0)
botservice (0.1.4)
cdn (0.2.0)
cloud (2.1.0)
cognitiveservices (0.2.4)
command-modules-nspkg (2.0.2)
configure (2.0.20)
consumption (0.4.2)
container (0.3.11)
core (2.0.55)
cosmosdb (0.2.7)
dla (0.2.3)
dls (0.1.7)
dms (0.1.1)
eventgrid (0.2.0)
eventhubs (0.3.2)
extension (0.2.3)
feedback (2.1.4)
find (0.2.13)
hdinsight (0.2.0)
interactive (0.4.1)
iot (0.3.4)
iotcentral (0.1.5)
keyvault (2.2.9)
lab (0.1.5)
maps (0.3.3)
monitor (0.2.8)
network (2.3.0)
nspkg (3.0.3)
policyinsights (0.1.0)
profile (2.1.2)
rdbms (0.3.5)
redis (0.3.2)
relay (0.1.2)
reservations (0.4.1)
resource (2.1.8)
role (2.3.0)
search (0.1.1)
security (0.1.0)
servicebus (0.3.2)
servicefabric (0.1.11)
signalr (1.0.0)
sql (2.1.7)
storage (2.3.0)
telemetry (1.0.0)
vm (2.2.12)

Python location '/usr/local/opt/python/bin/python3.7'
Extensions directory '/Users/alex/.azure/cliextensions'

Python (Darwin) 3.7.2 (default, Jan 21 2019, 16:56:28) 
[Clang 10.0.0 (clang-1000.11.45.5)]

[...]

Issue #7177, opened 2018-08-29
How to grant extra permissions with az ad app update


tumluliu:
First of all, is it possible to do this with this command? If yes, how to do that, e.g. if I want to grant Windows Active Directory -&gt; Read Directory Data permission to a specific App? Thanks!

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 921aa920-c25f-5227-aabb-49116da38fa7
* Version Independent ID: 9b5a53ad-edf8-4d89-f8e6-cde01b7866f6
* Content: [az ad app](https://docs.microsoft.com/en-us/cli/azure/ad/app?view=azure-cli-latest#az-ad-app-update)
* Content Source: [src/command_modules/azure-cli-role/azure/cli/command_modules/role/_help.py](https://github.com/Azure/azure-cli/blob/master/src/command_modules/azure-cli-role/azure/cli/command_modules/role/_help.py)
* Service: **active-directory**
* GitHub Login: @rloutlaw
* Microsoft Alias: **routlaw**

yugangw-msft:
Not a very smooth experience, but you can do it though `az ad app update` by providing the manifest artifacts

yugangw-msft:
okay, i will expose `az ad app permission add` to get it simpler. 
Please note, in the latest release, you will see "az ad app permission grant" which does the same thing of the `grant permissions` button in the portal. The new add command will default to grant unless you opt out with the flag of `--no-grant`

tumluliu:
that would be much better! Many thanks @yugangw-msft !

tumluliu:
@yugangw-msft could you disclose which version of az cli will contain this feature? For now, it's not available yet in `2.0.49`

paolosalvatori:
Hi, I have a couple of related questions: it looks like when running:

az ad app permission grant --id 9f4f64e0-4fd4-4a22-bbd2-b0ccd814f569 --api 00000003-0000-0000-c000-000000000000

[...]

PR #8804, opened 2019-03-17
graph: support admin consent

yugangw-msft:
With this, CLI will do exactly like the "Grant Permission" button in portal and hence both "Application" and "Delegated" permissions can be handled.

This addresses the lingering ask in #7177 and #8310 

//CC @paolosalvatori @jianghaolu 

### DIFF
--- a/src/command_modules/azure-cli-role/azure/cli/command_modules/role/_help.py
+++ b/src/command_modules/azure-cli-role/azure/cli/command_modules/role/_help.py
@@ -201,7 +201,7 @@ helps['ad app permission'] = """
 helps['ad app permission grant'] = """
     type: command
     short-summary: Grant the app an API Delegated permissions
-    long-summary: for application permissions, please use "ad app permission admin-consent"
+    long-summary: for Application permissions, please use "ad app permission admin-consent"
     examples:
         - name: Grant a native application with permissions to access an existing API with TTL of 2 years
           text: az ad app permission grant --id e042ec79-34cd-498f-9d9f-1234234 --api a0322f79-57df-498f-9d9f-12678 --expires 2
--- a/src/command_modules/azure-cli-role/azure/cli/command_modules/role/_help.py
+++ b/src/command_modules/azure-cli-role/azure/cli/command_modules/role/_help.py
@@ -200,10 +200,16 @@ helps['ad app permission'] = """
 """
 helps['ad app permission grant'] = """
     type: command
-    short-summary: Grant the app an API permission
+    short-summary: Grant the app an API Delegated permissions
+    long-summary: for application permissions, please use "ad app permission admin-consent"
     examples:
         - name: Grant a native application with permissions to access an existing API with TTL of 2 years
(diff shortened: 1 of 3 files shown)


src/command_modules/azure-cli-role/azure/cli/command_modules/role/custom.py
```diff
@@ -807,6 +807,34 @@ def delete_permission(cmd, identifier, api):
...
 
 
+def admin_consent(cmd, identifier):
+    import requests
+    from azure.cli.core.cloud import AZURE_PUBLIC_CLOUD
+    from azure.cli.core._profile import Profile
+    from azure.cli.core.commands.client_factory import UA_AGENT
+    from azure.cli.core.util import should_disable_connection_verify
+    if cmd.cli_ctx.cloud.name != AZURE_PUBLIC_CLOUD.name:
+        raise CLIError('This command does not function yet in solverign clouds')
```
marstr: typo! "solverign" should be "sovereign"

nit, I'd rephrase this to "This command is not yet supported on sovereign clouds"

src/command_modules/azure-cli-role/azure/cli/command_modules/role/custom.py
```diff
@@ -807,6 +807,34 @@ def delete_permission(cmd, identifier, api):
     return graph_client.applications.patch(application.object_id, update_parameter)
 
 
+def admin_consent(cmd, identifier):
+    import requests
+    from azure.cli.core.cloud import AZURE_PUBLIC_CLOUD
+    from azure.cli.core._profile import Profile
```
marstr: Importing from private namespaces always raises my eyebrows. Is `Profile` something we should bring into the public interface?

src/command_modules/azure-cli-role/azure/cli/command_modules/role/_help.py
```diff
@@ -200,11 +200,17 @@
...
+    short-summary: Grant the app an API Delegated permissions
+    long-summary: for Application permissions, please use "ad app permission admin-consent"
     examples:
         - name: Grant a native application with permissions to access an existing API with TTL of 2 years
           text: az ad app permission grant --id e042ec79-34cd-498f-9d9f-1234234 --api a0322f79-57df-498f-9d9f-12678 --expires 2
 """
+helps['ad app permission admin-consent'] = """
+    type: command
+    short-summary: grant Application & Delegated permissions through admin-consent.
+    long-summary: you must login as a directory administrator
```
marstr: This seems like a good detail to include. When we present help text, do we always present the short summary when we present the long summary? If not, I think we should repeat the short summary there as well.

yugangw-msft: Long summary will always be present from command's help. It will not be visible from a category's help such as `az ad permission -h`.
This should be good enough

src/command_modules/azure-cli-role/azure/cli/command_modules/role/custom.py
yugangw-msft: `_profile` is okay to use from core command modules which `role`, `keyvault` command modules have done that already, because we are always releasing them together.

PR #8814, opened 2019-03-19
Enable running image without context

### LINKED ISSUE #8813: [ACR] Contextless run and task
Azure Container Registry runs an image and creates a task with some context. The issue wants to support doing so without any context. 

## Running a task with an image 

|Description|Command | Notes|
|:-----------|:----------|:------|
| Running an image | `az acr run --cmd hello-world /dev/null`| |
| Image with Args | `az acr run --cmd "bash -c 'echo hello-world'" /dev/null`| Arguments similar to docker run| 
| Image with timeout | `az acr run --timeout 100 --cmd nginx /dev/null`| Run properties come before image args |

## Creating a Task with an Image or task file without a Source or GIT context 

|Description|Command | Notes|
|:-----------|:----------|:------|
| Create a task without context | `az acr task create -n taskname -f - -c /dev/null` | Used to support interactive input.  | 
| Create a task with an image | `az acr task create -n taskname --cmd hello-world -c /dev/null` |  This can run an image filelessly.  |

Goto aka.ms/acr/design for details.

NatalieYuGong:
Fixes #8813
---


src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
```diff
@@ -137,6 +137,8 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('values', help="The task values file path relative to the source context.")
         c.argument('set_value', options_list=['--set'], help="Value in 'name[=value]' format.", action='append', validator=validate_set)
         c.argument('set_secret', help="Secret value in 'name[=value]' format.", action='append', validator=validate_set_secret)
+        c.argument('cmd', options_list=['--__cmd__'])
```
northtyphoon: why do you need to define two arguments for `--cmd`? Do you need to workaround some CLI limitations? If yes, can you add some comments to clarify it?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/run.py
```diff
@@ -40,9 +42,15 @@ def acr_run(cmd,  # pylint: disable=too-many-locals
...
         cmd, registry_name, resource_group_name, RUN_NOT_SUPPORTED)
 
+    platform_os, platform_arch, platform_variant = get_validate_platform(cmd, os_type, platform)
+
+    EncodedTaskRunRequest, FileTaskRunRequest, PlatformProperties = cmd.get_models('EncodedTaskRunRequest', 'FileTaskRunRequest', 'PlatformProperties')
+
     client_registries = cf_acr_registries(cmd.cli_ctx)
 
-    if os.path.exists(source_location):
+    if source_location == NULL_SOURCE_LOCATION:
```
northtyphoon: Is the string comparison case-sensitive?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/run.py
```diff
@@ -67,25 +75,48 @@ def acr_run(cmd,  # pylint: disable=too-many-locals
...
+        YAML_TEMPLATE = ""
+        for s in sys.stdin.readlines():
+            YAML_TEMPLATE += s
+        values_content = ""
+
+    if cmd_value:
+        YAML_TEMPLATE = "steps: \n  - cmd: {{ .Values.run_image }}\n"
+        values_content = "run_image: {0}\n".format(cmd_value)
+
+    if values_content is not None:
```
northtyphoon: Should you check YAML_TEMPLATE as well?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/run.py
```diff
@@ -67,25 +75,48 @@ def acr_run(cmd,  # pylint: disable=too-many-locals
...
-            os=platform_os,
-            architecture=platform_arch,
-            variant=platform_variant
-        ),
-        credentials=get_custom_registry_credentials(
-            cmd=cmd,
-            auth_mode=auth_mode
+    if file == "-":
+        import sys
+        YAML_TEMPLATE = ""
```
northtyphoon: I don't see other places use ALL UPPER CASE. Is it something special?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/run.py
```diff
@@ -67,25 +75,48 @@ def acr_run(cmd,  # pylint: disable=too-many-locals
...
+            encoded_task_content=base64.b64encode(YAML_TEMPLATE.encode()).decode(),
+            encoded_values_content=base64.b64encode(values_content.encode()).decode(),
+            values=(set_value if set_value else []) + (set_secret if set_secret else []),
+            source_location=source_location,
+            timeout=timeout,
+            platform=PlatformProperties(
+                os=platform_os,
+                architecture=platform_arch,
+                variant=platform_variant
+            )
```
northtyphoon: Add credentials to EncodedTaskRunRequest.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -53,34 +55,62 @@ def acr_task_create(cmd,  # pylint: disable=too-many-locals
                     resource_group_name=None,
                     target=None,
                     auth_mode=None):
+
+    registry, resource_group_name = validate_managed_registry(
+        cmd, registry_name, resource_group_name, TASK_NOT_SUPPORTED)
+
+    if context_path == NULL_CONTEXT:
```
northtyphoon: case sensitive?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/run.py
```diff
@@ -67,25 +75,48 @@ def acr_run(cmd,  # pylint: disable=too-many-locals
...
+        values_content = ""
+
+    if cmd_value:
+        YAML_TEMPLATE = "steps: \n  - cmd: {{ .Values.run_image }}\n"
+        values_content = "run_image: {0}\n".format(cmd_value)
+
+    if values_content is not None:
+        import base64
+        request = EncodedTaskRunRequest(
+            encoded_task_content=base64.b64encode(YAML_TEMPLATE.encode()).decode(),
```
northtyphoon: why do you need to decode() at the end?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -53,34 +55,62 @@ def acr_task_create(cmd,  # pylint: disable=too-many-locals
...
+        values_content = ""
+
+    if values_content is not None:
+        import base64
+        EncodedTaskStep = cmd.get_models('EncodedTaskStep')
+        step = EncodedTaskStep(
+            encoded_task_content=base64.b64encode(YAML_TEMPLATE.encode()).decode(),
+            encoded_values_content=base64.b64encode(values_content.encode()).decode(),
             context_path=context_path,
             context_access_token=git_access_token,
```
northtyphoon: Since context_path is NULL in this case, you probably don't need to pass context_access_token.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -53,34 +55,62 @@ def acr_task_create(cmd,  # pylint: disable=too-many-locals
...
+        values_content = "image: {0}\n".format(cmd_value)
+
+    if file == "-":
+        import sys
+        YAML_TEMPLATE = ""
+        for s in sys.stdin.readlines():
+            YAML_TEMPLATE += s
+        values_content = ""
+
+    if values_content is not None:
```
northtyphoon: check YAML_TEMPLATE as well

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
NatalieYuGong: --cmd is a default value for cmd in CLI. I am overwriting it with another value and freeing it for image use.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/run.py
NatalieYuGong: The string comparison is case-sensitive.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/run.py
NatalieYuGong: Was thinking the existence of values_content infers the existence of yaml_template. Not sure if it is the other around. To sum up the discussions on Teams channel, for now RP supports either task_content + values_content or task_content + values. Might make more sense to check yaml_template instead of values_content.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/run.py
NatalieYuGong: The template was initially defined as a global constant, however, VisualStudio complains about that. I will convert to lowercase.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/run.py
NatalieYuGong: ok

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/run.py
NatalieYuGong: b64encode encodes bytewise. I am decode the byte representation of the decoded value back to a string.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
NatalieYuGong: The default value of access_token is None.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
NatalieYuGong: ok

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
```diff
@@ -406,6 +406,12 @@
 type: command
 short-summary: Queues a quick run providing streamed logs for an Azure Container Registry.
 examples:
+  - name: Queue a contextless run.
+    text: >
+        az acr run -r MyRegistry --cmd bash /dev/null
+  - name: Queue a run from standard input. Either 'Ctrl + z' or 'Ctrl + d' terminates the input stream.
```
northtyphoon: 'Ctrl + z' (Windows) or 'Ctrl + d' (Linux)

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
```diff
@@ -406,6 +406,12 @@
 type: command
 short-summary: Queues a quick run providing streamed logs for an Azure Container Registry.
 examples:
+  - name: Queue a contextless run.
```
northtyphoon: Queue a run to execute a container command.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
northtyphoon: Queue a run with the task definition from  the standard input.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
```diff
@@ -456,6 +462,12 @@
 type: command
 short-summary: Creates a series of steps for building, testing and OS & Framework patching containers. Tasks support triggers from git commits and base image updates.
 examples:
+  - name: Create a contextless task.
```
northtyphoon: Create a task without the source location.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
```diff
@@ -456,6 +462,12 @@
...
+    text: >
+        az acr task create -t hello-world:{{.Run.ID}} -n hello-world -r MyRegistry --cmd bash -c /dev/null
```
northtyphoon: `-t hello-workd:{{.Run.ID}}` is to specify the image tag and is not relevant to this example.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
```diff
@@ -132,11 +132,14 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
 
     with self.argument_context('acr run') as c:
         c.argument('registry_name', options_list=['--registry', '-r'])
-        c.positional('source_location', help="The local source code directory path (e.g., './src') or the URL to a git repository (e.g., 'https://github.com/Azure-Samples/acr-build-helloworld-node.git') or a remote tarball (e.g., 'http://server/context.tar.gz').", completer=FilesCompleter())
+        c.positional('source_location', help="The local source code directory path (e.g., './src') or the URL to a git repository (e.g., 'https://github.com/Azure-Samples/acr-build-helloworld-node.git') or a remote tarball (e.g., 'http://server/context.tar.gz'). /dev/null will be its value for a contextless run.", completer=FilesCompleter())
         c.argument('file', options_list=['--file', '-f'], help="The task template/definition file path relative to the source context.")
```
northtyphoon: If '/dev/null' is specified, the value will be set to None and ignored.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/run.py
```diff
@@ -67,25 +76,52 @@ def acr_run(cmd,  # pylint: disable=too-many-locals
...
-            auth_mode=auth_mode
+    if file == "-":
+        import sys
+        yaml_template = ""
+        for s in sys.stdin.readlines():
+            yaml_template += s
+        values_content = ""
+
+    if cmd_value:
+        yaml_template = "steps: \n  - cmd: {{ .Values.run_image }}\n"
```
northtyphoon: It just occurs to me the step default timeout is 10 minutes. If timeout is passed into the command, we need to add timeout https://github.com/Azure/acr-builder/blob/master/docs/task.md#steptimeout
https://github.com/Azure/acr-builder/blob/master/docs/task.md#totaltimeout

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
northtyphoon: Can you also add a pipeline example, like cat task.yaml | az acr run -r MyRegistry -f - .

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
```diff
@@ -160,7 +163,7 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('values', help="The task values/parameters file path relative to the source context.")
 
         # common to DockerBuildStep, FileTaskStep and RunTaskStep
-        c.argument('context_path', options_list=['--context', '-c'], help="The full URL to the source code repository (Requires '.git' suffix for a github repo).")
+        c.argument('context_path', options_list=['--context', '-c'], help="The full URL to the source code repository (Requires '.git' suffix for a github repo). /dev/null will be its value for a contextless task.")
```
ankurkhemani: nit: might wanna rename based on what we are keeping
`contextless` vs `without a context path`

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
```diff
@@ -187,6 +190,9 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
 
     with self.argument_context('acr task create') as c:
         c.argument('task_name', completer=None)
+        # Overwrite default shorthand of cmd to make availability for image usage
+        c.argument('cmd', options_list=['--__cmd__'])
+        c.argument('cmd_value', help="The image to create task with.", options_list=['--cmd'])
```
ankurkhemani: Do we need to define this twice? Can't we instead move it all the way up to `az acr` group and apply to all commands? Will that break any other command .. I guess not but please validate.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
```diff
@@ -406,6 +406,12 @@
 type: command
 short-summary: Queues a quick run providing streamed logs for an Azure Container Registry.
 examples:
+  - name: Queue a contextless run.
+    text: >
+        az acr run -r MyRegistry --cmd bash /dev/null
+  - name: Queue a run from standard input. Either 'Ctrl + z' or 'Ctrl + d' terminates the input stream.
+    text: >
+        az acr run -r MyRegistry -f - .
```
ankurkhemani: How does user know `-f` take a value of `-`? We probably need to update help message for the argument `-f` here. // cc @northtyphoon

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
```diff
@@ -132,11 +132,14 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
...
+        c.positional('source_location', help="The local source code directory path (e.g., './src') or the URL to a git repository (e.g., 'https://github.com/Azure-Samples/acr-build-helloworld-node.git') or a remote tarball (e.g., 'http://server/context.tar.gz'). /dev/null will be its value for a contextless run.", completer=FilesCompleter())
         c.argument('file', options_list=['--file', '-f'], help="The task template/definition file path relative to the source context.")
         c.argument('values', help="The task values file path relative to the source context.")
         c.argument('set_value', options_list=['--set'], help="Value in 'name[=value]' format.", action='append', validator=validate_set)
         c.argument('set_secret', help="Secret value in 'name[=value]' format.", action='append', validator=validate_set_secret)
+        # Overwrite default shorthand of cmd to make availability for image usage
+        c.argument('cmd', options_list=['--__cmd__'])
+        c.argument('cmd_value', help="Indicates whether the source location be interpreted as an image name to run or a context.", options_list=['--cmd'])
```
ankurkhemani: @northtyphoon Think help message can be improved here to make it more generic. I might be a little out of context here but afair `--cmd` is to take any cmdlet value right and not just an image name as per yaml definition?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -53,34 +55,62 @@ def acr_task_create(cmd,  # pylint: disable=too-many-locals
...
                        "[--pull-request-trigger-enabled] --git-access-token must be provided.")
 
-    if file.endswith(ALLOWED_TASK_FILE_TYPES):
-        FileTaskStep = cmd.get_models('FileTaskStep')
-        step = FileTaskStep(
-            task_file_path=file,
-            values_file_path=values,
+    if cmd_value:
+        yaml_template = "steps: \n  - cmd: {{ .Values.image }}\n"
+        values_content = "image: {0}\n".format(cmd_value)
```
ankurkhemani: Should this be `cmd` and not `image`?
 `values_content = "cmd: {0}\n".format(cmd_value)`

ankurkhemani: Do we ensure that `cmd_value` is an image name? Is the help message pointing that out? // cc @northtyphoon

ankurkhemani: Also, do we support multiple images that are space separated? If yes, have we validated that scenario?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -53,34 +55,62 @@ def acr_task_create(cmd,  # pylint: disable=too-many-locals
                     resource_group_name=None,
                     target=None,
                     auth_mode=None):
+
+    registry, resource_group_name = validate_managed_registry(
```
ankurkhemani: super nit: we can move this back to where it was as it's not used before.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
sajayantony: bash -c echo should be valid.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
NatalieYuGong: Done

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
NatalieYuGong: Examples might not be conspicuous enough to catch the attention. Also added in the argument list.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/run.py
NatalieYuGong: Added timeout for each step in the yaml template.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
sajayantony: As per discussion - we should be removing this validation all together.

NatalieYuGong: Dropped the registry validation for task group.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
NatalieYuGong: Escalated the cmd from acr/task to directly under acr.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
sajayantony: This looks like an invalid example.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/run.py
northtyphoon: where did you add the timeout in the yaml template?

NatalieYuGong: Only did this for `az acr task create`. Now `az acr run` is having timeout in the yaml template too.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -53,34 +60,62 @@ def acr_task_create(cmd,  # pylint: disable=too-many-locals
...
         raise CLIError("If source control trigger is enabled [--commit-trigger-enabled] or "
                        "[--pull-request-trigger-enabled] --git-access-token must be provided.")
 
-    if file.endswith(ALLOWED_TASK_FILE_TYPES):
-        FileTaskStep = cmd.get_models('FileTaskStep')
-        step = FileTaskStep(
-            task_file_path=file,
-            values_file_path=values,
+    if cmd_value:
+        yaml_template = "steps: \n  - cmd: {{ .Values.image }}\n    timeout: {{ .Values.timeout }}\n"
```
ankurkhemani: nit: I would rather name the values for `cmd` as `command` instead of `image` as it can more than just an image and confused me.
 `- cmd: {{ .Values.command}}`

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -53,34 +60,62 @@ def acr_task_create(cmd,  # pylint: disable=too-many-locals
...
+        values_content = "image: {0}\ntimeout: {1}\n".format(cmd_value, timeout)
```
ankurkhemani: nit: I would rather rename the key `image` to `command` as it can more than just an image and confused me.
 `- cmd: {{ .Values.command}}`

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/run.py
```diff
@@ -67,25 +76,52 @@ def acr_run(cmd,  # pylint: disable=too-many-locals
...
+    if file == "-":
+        import sys
+        yaml_template = ""
+        for s in sys.stdin.readlines():
+            yaml_template += s
+        values_content = ""
+
+    if cmd_value:
+        yaml_template = "steps: \n  - cmd: {{ .Values.run_image }}\n    timeout: {{ .Values.timeout }}\n"
+        values_content = "run_image: {0}\ntimeout: {1}\n".format(cmd_value, timeout)
```
ankurkhemani: nit: rename `run_image` to `command` or `cmd`

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
```diff
@@ -406,6 +406,15 @@
 type: command
 short-summary: Queues a quick run providing streamed logs for an Azure Container Registry.
 examples:
+  - name: Queue a run to execute a container command.
+    text: >
+        az acr run -r MyRegistry --cmd MyImage /dev/null
+  - name: Queue a run with the task definition from the standard input. Either 'Ctrl + z'(Windows) or 'Ctrl + d'(Linux) terminates the input stream.
```
djyou: Ctrl + Z, Ctrl + D

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/run.py
```diff
@@ -40,9 +42,16 @@ def acr_run(cmd,  # pylint: disable=too-many-locals
     _, resource_group_name = validate_managed_registry(
         cmd, registry_name, resource_group_name, RUN_NOT_SUPPORTED)
 
+    platform_os, platform_arch, platform_variant = get_validate_platform(cmd, os_type, platform)
+
+    EncodedTaskRunRequest, FileTaskRunRequest, PlatformProperties = cmd.get_models(
+        'EncodedTaskRunRequest', 'FileTaskRunRequest', 'PlatformProperties')
```
djyou: nit: move this block back to its original place, so that if error out early, not need to make these calls.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/run.py
```diff
@@ -67,25 +76,52 @@ def acr_run(cmd,  # pylint: disable=too-many-locals
...
+        yaml_template = ""
+        for s in sys.stdin.readlines():
+            yaml_template += s
+        values_content = ""
+
+    if cmd_value:
+        yaml_template = "steps: \n  - cmd: {{ .Values.run_image }}\n    timeout: {{ .Values.timeout }}\n"
+        values_content = "run_image: {0}\ntimeout: {1}\n".format(cmd_value, timeout)
+
+    if yaml_template is not None:
```
djyou: if yaml_template? Do you want to check for null or empty?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -53,34 +60,62 @@ def acr_task_create(cmd,  # pylint: disable=too-many-locals
...
+
+    if file == "-":
+        import sys
+        yaml_template = ""
+        for s in sys.stdin.readlines():
+            yaml_template += s
+        values_content = ""
+
+    if yaml_template is not None:
```
djyou: Same question, do you want to check for null or empty?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -53,34 +60,62 @@ def acr_task_create(cmd,  # pylint: disable=too-many-locals
...
-            values_file_path=values,
+    yaml_template = None
+    if cmd_value:
+        yaml_template = "steps: \n  - cmd: {{ .Values.command }}\n    timeout: {{ .Values.timeout }}\n"
+        values_content = "command: {0}\ntimeout: {1}\n".format(cmd_value, timeout)
+
+    if file == "-":
+        import sys
+        for s in sys.stdin.readlines():
+            yaml_template += s
```
northtyphoon: The code doesn't work. Error: unsupported operand type(s) for +=: 'NoneType' and 'str'

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/run.py
```diff
@@ -67,25 +71,57 @@ def acr_run(cmd,  # pylint: disable=too-many-locals
         source_location = check_remote_source_code(source_location)
         logger.warning("Sending context to registry: %s...", registry_name)
 
+    yaml_template = None
```
northtyphoon: `az acr run -r MyRegistry -f acb.yaml /dev/null` return error `Some of the properties of 'runRequest' are invalid. InnerErrors: FileTaskRunRequest:Missing required property 'SourceLocation'.`

If the user passes `/dev/null`, EncodedTaskRunRequest should be sent.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/run.py
```diff
@@ -67,25 +71,57 @@ def acr_run(cmd,  # pylint: disable=too-many-locals
...
+    if file == "-":
+        import sys
+        for s in sys.stdin.readlines():
+            yaml_template += s
+        values_content = ""
+
+    if cmd_value:
+        yaml_template = "steps: \n  - cmd: {{ .Values.command }}\n    timeout: {{ .Values.timeout }}\n"
```
northtyphoon: `az acr run -r MyRegistry --cmd bash /dev/null` doesn't work. Error: `
failed to deserialize task and validate: yaml: unmarshal errors: line 3: cannot unmarshal !!str `None` into int`

Please notice in `acr_run` function, the default `timeout` parameter value is `None`

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/run.py
NatalieYuGong: Is it a workaround? What about enabling none for source location of FileTaskRunRequest API?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/run.py
NatalieYuGong: For the command `az acr run --timeout 400 -f - /dev/null` with TTY input
`steps: `
`  - cmd: docker ps`
`    timeout: 500` ,
which timeout takes precedence?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -53,34 +60,66 @@ def acr_task_create(cmd,  # pylint: disable=too-many-locals
...
+        yaml_template = "steps: \n  - cmd: {{ .Values.command }}\n"
+        values_content = "command: {0}\n".format(cmd_value)
+
+    if file == "-":
+        import sys
+        for s in sys.stdin.readlines():
+            yaml_template += s
+        values_content = ""
+
+    if yaml_template and timeout:
```
northtyphoon: You need to move the timeout code into `if cmd_value`. 

    if cmd_value:
        yaml_template = "steps: \n  - cmd: {{ .Values.command }}\n"
        values_content = "command: {0}\n".format(cmd_value)
        if timeout:
            yaml_template += "    timeout: {}\n".format(timeout)

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -53,34 +60,66 @@ def acr_task_create(cmd,  # pylint: disable=too-many-locals
...
-        FileTaskStep = cmd.get_models('FileTaskStep')
-        step = FileTaskStep(
-            task_file_path=file,
-            values_file_path=values,
+    yaml_template = ""
+    if cmd_value:
+        yaml_template = "steps: \n  - cmd: {{ .Values.command }}\n"
+        values_content = "command: {0}\n".format(cmd_value)
+
+    if file == "-":
```
northtyphoon: if user pass --cmd and -f at the same time, we should block and the error message should tell users they can only pass --cmd or -f in the command.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/run.py
```diff
@@ -67,25 +71,61 @@ def acr_run(cmd,  # pylint: disable=too-many-locals
         source_location = check_remote_source_code(source_location)
         logger.warning("Sending context to registry: %s...", registry_name)
 
+    yaml_template = ""
```
northtyphoon: I have some comments in tas.py which also apply here.
1. timeout settings should be moved into `cmd_value`.
2. block if users pass file and cmd_value at the same time.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
```diff
@@ -406,6 +406,15 @@
 type: command
 short-summary: Queues a quick run providing streamed logs for an Azure Container Registry.
 examples:
+  - name: Queue a run to execute a container command.
+    text: >
+        az acr run -r MyRegistry --cmd MyImage /dev/null
+  - name: Queue a run with the task definition from the standard input. Either 'Ctrl + Z'(Windows) or 'Ctrl + D'(Linux) terminates the input stream.
+    text: >
+        az acr run -r MyRegistry -f - /dev/null
+  - name: Queue a piped run.
```
northtyphoon: Queue a run to execute the tasks passed through the pipe.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
```diff
@@ -456,6 +465,15 @@
 type: command
 short-summary: Creates a series of steps for building, testing and OS & Framework patching containers. Tasks support triggers from git commits and base image updates.
 examples:
+  - name: Create a task without the source location.
+    text: >
+        az acr task create -n hello-world -r MyRegistry --cmd MyImage -c /dev/null
+  - name: Create a task with the definition from the standard input. Either 'Ctrl + Z'(Windows) or 'Ctrl + D'(Linux) terminates the input stream.
+    text: >
+        az acr task create -n hello-world -r MyRegistry -f - -c /dev/null
+  - name: Create a piped task.
```
northtyphoon: Create a task with the definition from the pipe.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
NatalieYuGong: What about timeout for interactive input, for instance, `az acr run --timeout 500 -f - /dev/null`?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
NatalieYuGong: For now, file has the default value 'acb.yaml'. There is no way to tell if -f and --cmd are passed at the same time once inside the acr_task_create or acr_run.

NatalieYuGong: Unless file removes its default value from 'acb.yaml' to None.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
NatalieYuGong: Which timeout will be honored? The timeout after --timeout or the timeout from the input stream?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/run.py
NatalieYuGong: Are you assuming timeout can be taken in differently? For --cmd, the timeout should be read from  --timeout while for piped definition, the timeout should be read from the stream. If timeouts are not used in those ways by customers, they will be skipped.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
NatalieYuGong: Are we going to support a list of `--cmd` and a list of `--timeout` for `az acr run` and `az acr task create`?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/run.py
```diff
@@ -40,9 +47,15 @@ def acr_run(cmd,  # pylint: disable=too-many-locals
     _, resource_group_name = validate_managed_registry(
         cmd, registry_name, resource_group_name, RUN_NOT_SUPPORTED)
 
+    if cmd_value and file:
+        raise CLIError(
+            "Azure Container Registry can either run a contextless command or from a stream definition, but not both.")
```
northtyphoon: The error message needs to be actionable and point out the exact error if possible. In this case, the error message should tell the users they need to either pass `--cmd <cmd> /dev/null` or `-f acb.yaml <source_location>|/dev/null` to scheduler a run but not both.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
```diff
@@ -375,6 +375,38 @@ def get_validate_platform(cmd, os_type, platform):
     return platform_os, platform_arch, platform_variant
 
 
+def get_yaml_and_values(cmd_value, timeout, file):
+    """Generates yaml template and its value content if applicable
+    :param str cmd_value: The command to execute in each step
+    :param str timeout: The timeout for each step
+    :param str file: The task definition
+    """
+    yaml_template = ""
```
northtyphoon:
```
if cmd_value:
  -- generate template and values from cmd_value and timeout
else:
  if file == None:
    -- set file to default value 'acb.yaml'

  if file == '-':
    -- generate yaml from stdin
  else:
    if file exists:
      -- read yaml from file
    else 
      -- raise file not exist error

if yaml is empty
  -- raise yaml is empty error

```

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
```diff
@@ -375,6 +375,43 @@ def get_validate_platform(cmd, os_type, platform):
...
+    values_content = ""
+    if cmd_value:
+        yaml_template = "steps: \n  - cmd: {{ .Values.command }}\n"
+        values_content = "command: {0}\n".format(cmd_value)
+        if timeout:
+            yaml_template += "    timeout: {{ .Values.timeout }}\n"
+            values_content += "timeout: {0}\n".format(timeout)
+    else:
+        if not file:
+            file = "acb.yml"
```
northtyphoon: Previously the default file name passed to `acr_run` is in `acb.yaml`. Can you change back to it?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/run.py
```diff
@@ -69,23 +84,45 @@ def acr_run(cmd,  # pylint: disable=too-many-locals
...
-        ),
-        credentials=get_custom_registry_credentials(
-            cmd=cmd,
-            auth_mode=auth_mode
+    EncodedTaskRunRequest, FileTaskRunRequest, PlatformProperties = cmd.get_models(
+        'EncodedTaskRunRequest', 'FileTaskRunRequest', 'PlatformProperties')
+
+    if source_location:
+        request = FileTaskRunRequest(
+            task_file_path=file,
```
northtyphoon: Make sure file is also default to acb.yaml in this case

guanwening:
> +    """Generates yaml template and its value content if applicable
+    :param str cmd_value: The command to execute in each step
+    :param str timeout: The timeout for each step
+    :param str file: The task definition
+    """
+    yaml_template = ""
+    values_content = ""
+    if cmd_value:
+        yaml_template = "steps: \n  - cmd: {{ .Values.command }}\n"
+        values_content = "command: {0}\n".format(cmd_value)
+        if timeout:
+            yaml_template += "    timeout: {{ .Values.timeout }}\n"
+            values_content += "timeout: {0}\n".format(timeout)
+    else:
+        if not file:
+            file = "acb.yml"
Previously the default file name passed to acr_run is in acb.yaml. Can you change back to it?

NatalieYuGong:
@tjprescott , we are ready to merge the pull request once CI passed. /cc @northtyphoon @sajayantony

NatalieYuGong:
Screen Shot 2019-04-01 at 8 26 51 PM
Screen Shot 2019-04-01 at 8 28 05 PM
Is there any update on the testSDK and the virtual environment of the travis system?

tjprescott:
@NatalieYuGong is this PR ready to be merged?

NatalieYuGong:
@tjprescott , acr run test recording file is added. 
I have a dangling question. Do you know how CI reads and writes a local file? In other words, is there any kind of path configuration the CI takes in as an environment and searches for files?

tjprescott:
@NatalieYuGong there are examples in `test_dns_commands.py` where external files are used. That may be helpful.

PR #8823, opened 2019-03-19
Invoke azdev to test command_module versions before release

### LINKED ISSUE #8182: Add pip packaging checks to release pipeline
There are two scripts that should be added to the PIP release pipeline and should gate *all* release mechanisms.

`azdev verify version` - script to ensure that the version has been bumped if the build artifact differs from the latest version on PyPI.
`azdev verify history` - checks random little things like the length of the underline, etc.

Verify history is done as part of each PR on the dev branch, but `verify version` and `verify history` must be run on the release branch since that is where the versioning is ultimately resolved before release. If these steps are missed for any reason (since this is a manual step), there is a risk of modules which *should* be published NOT being published because of the version number.

So, as a failsafe, the first step of the release pipeline should be to ensure these checks pass.

cc/ @yugangw-msft @marstr

marstr:
Fixes #8182

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -60,6 +60,15 @@ jobs:
       . ./venv/bin/activate
       azdev verify history
 
+    displayName: 'Verify History'
+
+  - bash: |
+      #!/usr/bin/env bash
+
+      set -ev
+
+      . ./venv/bin/activate
+      azdev cli check-versions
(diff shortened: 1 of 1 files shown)


marstr:
Todo's embedded here for when Azure/azure-cli-dev-tools#56 is fixed.

PR #8825, opened 2019-03-19
Aladdin generated examples for Vm

glGarg:
---


src/command_modules/azure-cli-vm/azure/cli/command_modules/vm/_help.py
```diff
@@ -629,6 +659,10 @@
 long-summary: |
     For Linux VMs, disabling encryption is only permitted on data volumes.
     For Windows VMS, disabling encryption is permitted on both OS and data volumes.
+examples:
+  - name: Disable disk encryption on the OS disk and/or data disks. (autogenerated)
```
tjprescott: This description is wrong. It is clearly disabling encryption only on data disks...

src/command_modules/azure-cli-vm/azure/cli/command_modules/vm/_help.py
```diff
@@ -857,6 +899,10 @@
 helps['vm image accept-terms'] = """
 type: command
 short-summary: Accept Azure Marketplace term so that the image can be used to create VMs
+examples:
+  - name: Accept Azure Marketplace term so that the image can be used to create VMs. (autogenerated)
+    text: az vm image accept-terms --urn publisher:offer:sku:version
```
tjprescott: `--urn publisher:offer:sku:version` this is not a real value.

src/command_modules/azure-cli-vm/azure/cli/command_modules/vm/_help.py
glGarg: I agree, we will keep this feedback in mind for getting better descriptions in future examples.

src/command_modules/azure-cli-vm/azure/cli/command_modules/vm/_help.py
glGarg: Makes sense, we are currently in process of prioritize more concrete parameter values. As per offline discussion, we will still retain such values in the AI as they contain valuable information.

JasonRShaver:
Ignore this:

Please react

PR #8826, opened 2019-03-20
refactor LICENSE.txt to make github even more happier

tariq1890:
GitHub looks for files named `LICENSE` to determine the kind of license used by a repo.

I have modified the contents for more conformance to standards.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
-The MIT License (MIT)
+MIT License
 
-Copyright (c) Microsoft Corporation. All rights reserved.
+Copyright (c) 2016 Microsoft Azure
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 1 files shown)


tariq1890:
FYI @marstr

LICENSE
```diff
@@ -1,6 +1,6 @@
-The MIT License (MIT)
+MIT License
 
-Copyright (c) Microsoft Corporation. All rights reserved.
+Copyright (c) 2016 Microsoft Azure
```
tariq1890: Added 2016 looking at the git history.

marstr: Adding the year makes sense to me, but I'm hesitant to change the name of the organization holding the copyright and removing the "All rights reserved." text. Being a fellow Microsofty, is there some official guidance you can send me offline on what this license file should hold?

PR #8828, opened 2019-03-20
Add postgresql replica commands and restart server command

grigusev:
---


src/command_modules/azure-cli-rdbms/azure/cli/command_modules/rdbms/_params.py
```diff
@@ -118,5 +121,8 @@ def _complex_params(command_group):
             c.argument('server_name', options_list=['--server-name', '-s'])
             c.argument('configuration_name', id_part='child_name_1', options_list=['--name', '-n'])
 
+    with self.argument_context('postgres server replica list') as c:
+        c.argument('server_name', options_list=['--server-name', '-s'], help='Name of the master server.')
+
     with self.argument_context('mysql server replica list') as c:
         c.argument('server_name', options_list=['--server-name', '-s'], help='Name of the master server.')
```
tjprescott: Seems you have a lot of duplication here.  You could refactor for more concise code and maintainability:

```Python
for scope in ['postgres', 'mysql']:
  with self.argument_context('{} server replica'.format(scope)) as c:
    c.argument(…)
```

This avoids having to update seomthing in multiple places if, say, you want to change the help text for source_server.

grigusev: it is good idea and there are many other places -  but we didn't  "generalize" postgres, mysql, mariadb code right now as different team are  supporting these databases (with different release cycles), it seems it's better to refactor all these places in the dedicated PR.

PR #8829, opened 2019-03-20
[wip] az feedback

### LINKED ISSUE #2048: az feedback does not request feedback
The az feedback command asks 2 questions:

1) would you recommend us 1-10. 
2) what do we do well. 

One would expect a 3rd question , what do we do "not so well" which would afford the user an opportunity to vent some frustration or report an issue. Unless you only want positive feedback, that is :-)

adewaleo:
`az feedback` now:
- directs users to helpful links
- shows last 9 commands
- prompts user to create an issue template based on selected command
- directs user to core repo or extensions repo based on command

Todo:
- Update `az login` webpage to contain getting started links / commands.
- Properly handle showing feedback commands (possibly duplicate commands). For now, we simply don't add `az feedback` to the list of recent commands.

closes #2048

---


adewaleo:
Please feel free to review away!

cc: @tjprescott, @yugangw-msft, @achandmsft.

adewaleo:
ping! (for review)
Also not sure why the Azure Pipelines test is failing
CC: @yugangw-msft  @tjprescott @marstr

src/azure-cli-core/setup.py
```diff
@@ -72,7 +72,9 @@
     'six',
     'tabulate>=0.7.7',
     'wheel',
-    'azure-mgmt-resource==2.1.0'
+    'azure-mgmt-resource==2.1.0',
+    'pyperclip==1.7.0',
+    'psutil==5.6.1'
```
marstr: Looks like this dependency is causing our ADO check to fail. I suspect that this dependency isn't playing nice inside of alpine.

src/azure-cli-core/setup.py
```diff
@@ -72,7 +72,9 @@
     'six',
     'tabulate>=0.7.7',
     'wheel',
-    'azure-mgmt-resource==2.1.0'
+    'azure-mgmt-resource==2.1.0',
```
marstr: In general, our setup.py files should contain a range of acceptable versions, instead of one pinned version. Not doing this causes issues like https://github.com/Azure/azure-cli/issues/6973.

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -24,44 +24,48 @@ def handle_exception(ex, cli_ctx=None):
...
+                if cli_ctx.invocation.data['command_extension_name']:
+                    is_extension = True
+            except (AttributeError, KeyError):
+                pass
 
-    issue_url = "https://aka.ms/azcli/ext/issues" if is_extension else "https://aka.ms/azcli/issues"
-    logger.error("The command failed with an unexpected error. Here is the traceback:\n")
-    logger.exception(ex)
-    logger.warning("\nTo open an issue, please visit: %s", issue_url)
+        issue_url = "https://aka.ms/azcli/ext/issues" if is_extension else "https://aka.ms/azcli/issues"
```
marstr: I'd rather see this as a constant declared at the top of the file instead of buried in the implementation.

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -24,44 +24,48 @@ def handle_exception(ex, cli_ctx=None):
...
+        logger.error("The command failed with an unexpected error. Here is the traceback:\n")
+        logger.exception(ex)
+        logger.warning("\nTo open an issue, please visit: %s", issue_url)
```
tjprescott: Shouldn't we direct people to use `az feedback` here?

src/azure-cli/azure/cli/__main__.py
```diff
@@ -43,9 +45,21 @@ def cli_main(cli, args):
     else:
         telemetry.set_success()
 
+    elapsed_time = timeit.default_timer() - start_time
```
tjprescott: Why do we log the time for each command?  I worry this may attract more perf related issues being raised.

src/command_modules/azure-cli-feedback/azure/cli/command_modules/feedback/custom.py
```diff
@@ -4,61 +4,557 @@
...
 from knack.log import get_logger
 from knack.prompting import prompt, NoTTYException
 from knack.util import CLIError
 
-from azure.cli.core.util import COMPONENT_PREFIX, get_installed_cli_distributions
-from azure.cli.core.telemetry import set_feedback
+from azure.cli.core.util import get_az_version_string
+from azure.cli.core.azlogging import UNKNOWN_COMMAND, CMD_LOG_LINE_PREFIX
+from azure.cli.core.util import open_page_in_browser
+import pyperclip
```
tjprescott: What is this for?

src/command_modules/azure-cli-feedback/azure/cli/command_modules/feedback/custom.py
```diff
@@ -4,61 +4,557 @@
...
+_EXTENSIONS_ISSUES_URL = "aka.ms/azcli/ext/issues"
+
+_MSG_INTR = \
+    '\nWe appreciate your feedback!\n\n' \
+    'For more information on getting started, visit: {}\n' \
+    'If you have questions, visit our Stack Overflow page: {}\n'\
+    .format(_GET_STARTED_URL, _QUESTIONS_URL)
+
+_MSG_CMD_ISSUE = "\nEnter the number of the command you would like to create an issue for. " \
+                 "\nEnter {} to create a generic issue. Enter q to quit: "
```
tjprescott: I would prefer:
```
[0] Open a generic issue.
[1] cmd foo
[2] cmd blah blah
```
This requires less reading and thus should lead to better outcomes.

src/command_modules/azure-cli-feedback/azure/cli/command_modules/feedback/custom.py
```diff
@@ -4,61 +4,557 @@
...
+
+BEGIN TEMPLATE
+===============
+**A browser has been opened to {} to create an issue.**
+**The issue template has been copied to your clipboard. You can also run `az feedback --verbose` to emit the output to stdout.**
+"""
+
+_ISSUES_TEMPLATE = """
+
+### **This autogenerated template contains command parameters. Please review and remove sensitive information, if necessary**
```
tjprescott: Why don't we simply remove the values and replace them with placeholders? This is the kind of thing we do not want to screw up. As soon as someone DOESN'T read this and posts sensitive information (which would take all of oh... 1 week?) we have a problem.

src/command_modules/azure-cli-feedback/azure/cli/command_modules/feedback/custom.py
```diff
@@ -4,61 +4,557 @@
...
+
+**Command Name**
+    `{command_name}`
+
+**Ran:**
+    `{executed_command}`
+
+
+**Command Duration**
+    {command_duration}
```
tjprescott: Why is this useful?

src/command_modules/azure-cli-feedback/azure/cli/command_modules/feedback/custom.py
```diff
@@ -4,61 +4,557 @@
...
+
+### **This autogenerated template contains command parameters. Please review and remove sensitive information, if necessary**
+
+## Describe the bug
+
+**Command Name**
+    `{command_name}`
+
+**Ran:**
+    `{executed_command}`
```
tjprescott: It is safer to put sanitized values here and invite the user to put actual values than vice-versa. @JasonRShaver @yugangw-msft you know more about the risks here as they are largely the same as the risks involved with telemetry.

src/command_modules/azure-cli-feedback/azure/cli/command_modules/feedback/custom.py
```diff
@@ -4,61 +4,557 @@
...
+
+
+**Command Duration**
+    {command_duration}
+
+**Errors:**
+{errors_string}
+
+## To Reproduce:
+
```
tjprescott: It seems like "Ran" from above could be combined here.  At a minimum, the command they ran is the last step of the repro...

src/command_modules/azure-cli-feedback/azure/cli/command_modules/feedback/custom.py
```diff
@@ -4,61 +4,557 @@
...
+    {system_name}
+    {platform}
+    {shell}
+
+**CLI Version**
+
+{cli_version}
+
+## Additional context
+Add any other context about the problem here.
```
tjprescott: I would move all the stuff they would be expected to fill in manually to the top of the auto-gen template. Otherwise they will simply expect that everything is pre-filled for them.

.github/ISSUE_TEMPLATE/Bug_report.md
```diff
@@ -4,7 +4,7 @@ about: Create a report to help us improve
 
 ---
 
-**Describe the bug**
+q**Describe the bug**
```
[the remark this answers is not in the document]
adewaleo: yes, I noticed this too, its a typo, mistake, will update it with my next round of changes.

src/azure-cli-core/azure/cli/core/azlogging.py
```diff
@@ -23,11 +23,112 @@
...
+        def _delete_old_logs(log_dir):
+            log_file_names = [file for file in os.listdir(log_dir) if file.endswith(".log")]
+            sorted_files = sorted(log_file_names, reverse=True)
+
+            if len(sorted_files) > 25:
+                for file in sorted_files[5:]:
+                    try:
+                        os.remove(os.path.join(log_dir, file))
+                    except OSError:  # FileNotFoundError introduced in Python 3
+                        continue
```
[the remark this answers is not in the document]
adewaleo: keeps files between 0 and 20 / 25. I could rename the variable to clarify that the files are in reverse order.  I wanted to make removing the files an infrequent operation that removes a bulk of unnecessary commands. Maybe I could fine tune this.

src/azure-cli-core/azure/cli/core/util.py
adewaleo: We most definitely should!

src/azure-cli-core/azure/cli/core/util.py
adewaleo: Sounds good!

src/azure-cli-core/setup.py
adewaleo: makes sense! do you think this could be related to the alpine linux problem?

src/azure-cli/azure/cli/__main__.py
adewaleo: I think this was brought up in the meeting with @yugangw-msft and  @achandmsft. I think the rational was to help customers notice / point out situations were commands are much slower than usual. I suppose we just saw it as general command related information. 

If perf really is a problem (esp. on the CLI side) moving forward then I think it doesn't hurt to give customers and ourselves more clarity regarding it. If it **really** is a problem then we eventually have to address it.

I guess we could discuss this more when I am back from the conference.

src/command_modules/azure-cli-feedback/azure/cli/command_modules/feedback/custom.py
adewaleo: It's for copying the issue template output to the user's clip board.

src/command_modules/azure-cli-feedback/azure/cli/command_modules/feedback/custom.py
adewaleo: Sounds good. I will make the update.

src/command_modules/azure-cli-feedback/azure/cli/command_modules/feedback/custom.py
adewaleo: Yes this is a tough / nuanced decision to make. On the one hand, we risk customer security by showing command arguments in the output by default, but this would often contain very helpful information.

I suppose place holders could be the happy medium. I can make this change then. 

cc: @yugangw-msft  @achandmsft

src/command_modules/azure-cli-feedback/azure/cli/command_modules/feedback/custom.py
adewaleo: It's just metadata about the command which could possible be useful. I suppose it doesn't hurt.

cc: @yugangw-msft @achandmsft

src/command_modules/azure-cli-feedback/azure/cli/command_modules/feedback/custom.py
adewaleo: That was the goal, but @yugangw-msft suggested moving "ran..." up near the top of the issue so users are less likely to post the issue without checking the command's arguments for sensitive info.

However, if we go with the placeholder approach, then I could move this command info back down to this section..

src/azure-cli-core/azure/cli/core/azlogging.py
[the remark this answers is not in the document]
tjprescott: Add comments so we understand the logic.

src/command_modules/azure-cli-feedback/azure/cli/command_modules/feedback/custom.py
tjprescott: Well, it would be helpful to have all the data in telemetry as well... problem is, that would be illegal. :)

adewaleo: fair enough, lol. I will confirm with team and go ahead to replace with placeholders after scrum...

src/azure-cli-core/azure/cli/core/azlogging.py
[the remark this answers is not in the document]
adewaleo: @tjprescott, you were right, there was a bug. I have fixed it.

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -20,48 +20,42 @@
...
-            if cli_ctx.invocation.data['command_extension_name']:
-                is_extension = True
-        except (AttributeError, KeyError):
-            pass
-
-    issue_url = "https://aka.ms/azcli/ext/issues" if is_extension else "https://aka.ms/azcli/issues"
-    logger.error("The command failed with an unexpected error. Here is the traceback:\n")
-    logger.exception(ex)
-    logger.warning("\nTo open an issue, please visit: %s", issue_url)
+        if isinstance(ex, (CLIError, CloudError)):
```
adewaleo: this section should be exactly the same, except that we use the command logger context and we no longer print the issues url but direct users to use az feedback.

Dockerfile
```diff
@@ -34,7 +34,7 @@ LABEL maintainer="Microsoft" \
 # pip wheel - required for CLI packaging
 # jmespath-terminal - we include jpterm as a useful tool
 RUN apk add --no-cache bash openssh ca-certificates jq curl openssl git zip \
- && apk add --no-cache --virtual .build-deps gcc make openssl-dev libffi-dev musl-dev \
+ && apk add --no-cache --virtual .build-deps gcc make openssl-dev libffi-dev musl-dev linux-headers \
```
marstr: Could you point me to where you found this secret-sauce? It looks promising (seeing as we haven't historically been too sensitive to image size) but I want to make sure that I understand the consequences of adding this dependency.

adewaleo: @marstr, this is the [github issue / comment](https://github.com/giampaolo/psutil/issues/872#issuecomment-271986653) that helped me out.

tjprescott:
@marstr ping.

PR #8832, opened 2019-03-20
Add  --hub-name for show-connection-string command

anusapan:
---

### DIFF
--- a/src/command_modules/azure-cli-iot/azure/cli/command_modules/iot/_params.py
+++ b/src/command_modules/azure-cli-iot/azure/cli/command_modules/iot/_params.py
@@ -172,5 +172,6 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
                    help='The number of partitions of the backing Event Hub for device-to-cloud messages.', type=int)
 
     with self.argument_context('iot hub show-connection-string') as c:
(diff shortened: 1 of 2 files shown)


src/command_modules/azure-cli-iot/azure/cli/command_modules/iot/_params.py
```diff
@@ -172,5 +172,6 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
                    help='The number of partitions of the backing Event Hub for device-to-cloud messages.', type=int)
 
     with self.argument_context('iot hub show-connection-string') as c:
+        c.argument('hub_name', options_list=['--hub-name', '--name', '-n'])
```
yugangw-msft: For consistency, the argument should be just `--name`/`-n`

anusapan: Agreed, but we are trying to maintain the CLI consistency and partially restoring the behavior of show-connection-string command which we removed from IoT cli-extension. That command accepts --hub-name as a parameter for IoT hub-name, So now allowing to use --hub-name/--name/-n for the hub-name parameter.

PR #8834, opened 2019-03-20
BotService 0.1.9: add prepare-deploy and update az bot create

stevengum:
- Adds `az bot prepare-deploy` _once_ which is to be used before deploying bots using `az webapp`
- Add tests for `az bot prepare-deploy`
- Cleaned up imports order in `test_bot_commands.py`
---

### DIFF
--- a/src/command_modules/azure-cli-botservice/azure/cli/command_modules/botservice/custom.py
+++ b/src/command_modules/azure-cli-botservice/azure/cli/command_modules/botservice/custom.py
@@ -63,7 +63,7 @@ def create(cmd, client, resource_group_name, resource_name, kind, description=No
     bot_kind = 'bot'
     webapp_kind = 'webapp'
     function_kind = 'function'
-    show_password = True if not password else False
+    show_password = False
 
     if resource_name.find(".") > -1:
         logger.warning('"." found in --name parameter ("%s"). "." is an invalid character for Azure Bot resource names '
@@ -87,7 +87,7 @@ def create(cmd, client, resource_group_name, resource_name, kind, description=No
     if not msa_app_id:
 
         logger.info('Microsoft application id not passed as a parameter. Provisioning a new Microsoft application.')
-
+        show_password = True
         msa_app_id, password = ConvergedApp.provision(resource_name)
         logger.info('Microsoft application provisioning successful. Application Id: %s.', msa_app_id)
 
--- a/src/command_modules/azure-cli-botservice/azure/cli/command_modules/botservice/custom.py
+++ b/src/command_modules/azure-cli-botservice/azure/cli/command_modules/botservice/custom.py
@@ -63,6 +63,7 @@ def create(cmd, client, resource_group_name, resource_name, kind, description=No
     bot_kind = 'bot'
     webapp_kind = 'webapp'
     function_kind = 'function'
+    show_password = True if not password else False
 
     if resource_name.find(".") > -1:
         logger.warning('"." found in --name parameter ("%s"). "." is an invalid character for Azure Bot resource names '
@@ -99,7 +100,7 @@ def create(cmd, client, resource_group_name, resource_name, kind, description=No
 
         # Registration bot specific validation
         if not endpoint:
-            raise CLIError('Endpoint is required for creating a registration bot.')
+            endpoint = ''
         if not msa_app_id:
             raise CLIError('Microsoft application id is required for creating a registration bot.')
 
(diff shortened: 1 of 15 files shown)


stevengum:
@tjprescott @yugangw-msft, the latest command I've added is a pre-az webapp deployment step which we need to streamline the process of publishing a bot to azure when using az webapp. It doesn't perform any network calls to Azure (except a call to a blob).

Do you know how to fix or work around the error message that popped up in the tests, about needing to call [`az login`](https://travis-ci.org/Azure/azure-cli/jobs/509110593#L6183)?

src/command_modules/azure-cli-botservice/azure/cli/command_modules/botservice/custom.py
```diff
@@ -63,6 +63,7 @@ def create(cmd, client, resource_group_name, resource_name, kind, description=No
     bot_kind = 'bot'
     webapp_kind = 'webapp'
     function_kind = 'function'
+    show_password = True if not password else False
```
yugangw-msft: Simplify `show_password = not password`

src/command_modules/azure-cli-botservice/azure/cli/command_modules/botservice/custom.py
```diff
@@ -118,11 +119,14 @@ def create(cmd, client, resource_group_name, resource_name, kind, description=No
...
 
-        return client.bots.create(
+        result = client.bots.create(
             resource_group_name=resource_group_name,
             resource_name=resource_name,
             parameters=parameters
-        )
+        ).as_dict()
+        if show_password:
+            result['password'] = password
```
yugangw-msft: you do understand that the `password` could be None, and is the command fine with that?

stevengum: I've changed the show_password value to be set to True during the call to create a Converged App.

This way if a password is created on behalf of the user, the user will know what the password is. (Fixes: Microsoft/botbuilder-tools#1021)

If the password is already provided, the user has the password and doesn't need the information again.

The password can only be None if the user provides an appid, in which case they can access the portal(s) or use `az ad app update` to create a password.

Thanks for catching that!

src/command_modules/azure-cli-botservice/azure/cli/command_modules/botservice/custom.py
stevengum: @yugangw-msft resolving via comment [below](https://github.com/Azure/azure-cli/pull/8834/files?file-filters%5B%5D=.py#r267631987) and that this comment is now out of date.

hansmbakker:
Does this PR break publishing when publishing a webapp bot that was provisioned without the `SCM_DO_BUILD_DURING_DEPLOYMENT` setting?

I'm getting warnings in my Azure DevOps pipeline since recently:
` WARNING: Didn't detect SCM_DO_BUILD_DURING_DEPLOYMENT or its value was "false" in App Service's Application Settings. Build may not commence during deployment. To learn how to trigger a build when deploying to your App Service, see https://github.com/projectkudu/kudu/wiki/Deploying-from-a-zip-file`

hansmbakker:
Is it needed to run `az bot prepare-deploy` now before `az bot publish`? I didn't see this documented in the Enterprise Bot Template for example.

stevengum:
@hansmbakker, this PR (and the other one you commented on) doesn't break the method of using `az bot publish` to publish bots, but it does add a warning message that you can most likely ignore. 

We are now recommending that users use [`az webapp deployment`](https://docs.microsoft.com/en-us/cli/azure/webapp/deployment/source?view=azure-cli-latest) to deploy their bot's code. One of the deployment capabilities in this command group is for deploying via the [zipdeploy API](https://docs.microsoft.com/en-us/cli/azure/webapp/deployment/source?view=azure-cli-latest#az-webapp-deployment-source-config-zip) (which we use in `az bot publish`. 

The recommendation for the native `az webapp deployment` is because it gives the user more control over their deployment process. 

`az bot publish` rolled a lot of separate steps into one command. `prepare-deploy` creates the necessary web.config for Node.js bots to work with IIS.

For .netcore bots, it creates the .deployment file that is created in `az bot publish` that tells Kudu to build the project. If `SCM_DO_BUILD_DURING_DEPLOYMENT`  is true, `prepare-deploy` should be run by the user to create the .deployment file before you publish your bot to your web app.

After running `az bot prepare-deploy` (if necessary), the user should then zip their code up and use the `az webapp deployment source config-zip` to deploy their code.


Cross-linking to [issue](https://github.com/Microsoft/botbuilder-tools/issues/985#issuecomment-484912981) in BotBuilder-tools for more context.

PR #8848, opened 2019-03-22
releases: bump up all versions

yugangw-msft:
---


azure-pipelines.yml
```diff
@@ -71,7 +71,7 @@ jobs:
       set -ev
 
       # TODO: once Azure/azure-cli-dev-tools#56 is fixed, remove grep component of check
-      azdev cli check-versions | grep "OK!"
+      azdev cli check-versions 2>&1 >/dev/null | grep "OK!"
```
yugangw-msft: //CC @marstr @tjprescott,  the [OK!](https://github.com/Azure/azure-cli-dev-tools/blob/master/azdev/operations/pypi.py#L212) goes to stderr, hence the redirecting is needed

tjprescott: Sounds good. I've fixed the exit code in azdev so we will be able to simplify this very soon.

PR #8862, opened 2019-03-23
role assignment delete: prompt when delete all assignments under the subscription unconditionally

### LINKED ISSUE #2933: Default behaviour of 'az role assignment delete' is to delete ALL role assignments on the subscription. This is wildly dangerous and breaks everything
### Environment summary

**Install Method:** How did you install the CLI? (e.g. pip, interactive script, apt-get, Docker, MSI, nightly)  
Answer here: pip

**CLI Version:** What version of the CLI and modules are installed? (Use `az --version`)  
Answer here:
azure-cli (2.0.2)
acr (2.0.0)
acs (2.0.2)
appservice (0.1.2)
batch (2.0.0)
cloud (2.0.0)
component (2.0.0)
configure (2.0.2)
container (0.1.2)
core (2.0.2)
documentdb (0.1.2)
feedback (2.0.0)
find (0.0.1b1)
iot (0.1.2)
keyvault (2.0.0)
lab (0.0.1)
monitor (0.0.1)
network (2.0.2)
nspkg (2.0.0)
profile (2.0.2)
redis (0.1.1b3)
resource (2.0.2)
role (2.0.1)
sql (2.0.0)
storage (2.0.2)
vm (2.0.2)

**OS Version:** What OS and version are you using?  
Answer here: Windows 10

**Shell Type:** What shell are you using? (e.g. bash, cmd.exe, Bash on Windows)  
Answer here: cmd.exe

---

### Description
The default behaviour of 'az role assignment delete' is to delete all role assignments on the subscription. No other command has such a wildly dangerous default parameterless behaviour. I typed this and have completely wiped out everyone's access to my subscription. :D Luckily it was in dev.

When 'az role assignment delete' is entered, it should prompt for a parameter. The current default behaviour is nuts.

yugangw-msft:
Fix #2933 

### DIFF
--- a/src/command_modules/azure-cli-role/azure/cli/command_modules/role/custom.py
+++ b/src/command_modules/azure-cli-role/azure/cli/command_modules/role/custom.py
@@ -405,8 +405,8 @@ def _get_displayable_name(graph_object):
     return graph_object.display_name or ''
 
 
-def delete_role_assignments(cmd, ids=None, assignee=None, role=None,
-                            resource_group_name=None, scope=None, include_inherited=False):
+def delete_role_assignments(cmd, ids=None, assignee=None, role=None, resource_group_name=None,
+                            scope=None, include_inherited=False, yes=None):
     factory = _auth_client_factory(cmd.cli_ctx, scope)
     assignments_client = factory.role_assignments
     definitions_client = factory.role_definitions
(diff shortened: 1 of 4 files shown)


src/command_modules/azure-cli-role/azure/cli/command_modules/role/custom.py
```diff
@@ -417,6 +417,14 @@ def delete_role_assignments(cmd, ids=None, assignee=None, role=None,
         for i in ids:
             assignments_client.delete_by_id(i)
         return
+    if not any([ids, assignee, role, resource_group_name, scope, assignee, yes]):
+        from knack.prompting import prompt_y_n, NoTTYException
+        try:
+            msg = 'You are deleting all role assignments under the subscription. Are you sure to continue?'
```
tjprescott: Knack already handles the "NoTTYException" so you shouldn't have to. On wording, I would recommend: "This will delete all role assignments under the subscription. Are you sure?"

PR #8874, opened 2019-03-26
[Storage] Add param help for public-access

### LINKED ISSUE #8205: Document that --public-access in create is configurable post-creation using set-permission
I was trying to find how to set `--public-access` on an already existing container, the Azure Portal UI doesn't really guide one in the right direction with showing this under "Access Policy", so I was very surprised to find that creating two accounts, one with default and one with `blob` and querying their policies using `az storage container policy list`resulted in `{}` in both cases.

I figured out I have to use `az storage container set-permission`, but it would be nice to say that in the description of the `--public-access` argument in the `az storage container create` section.

---
#### Document Details

* ID: adcba5ed-e693-6345-2c1b-c17422d7b8f9
* Version Independent ID: 58297400-b777-6868-1687-0af3a512a16f
* Content: [az storage container](https://docs.microsoft.com/en-us/cli/azure/storage/container?view=azure-cli-latest)
* Content Source: [src/command_modules/azure-cli-storage/azure/cli/command_modules/storage/_help.py](https://github.com/Azure/azure-cli/blob/dev/src/command_modules/azure-cli-storage/azure/cli/command_modules/storage/_help.py)
* Service: **storage**
* GitHub Login: @rloutlaw

limingu:
Fixes #8205
---

### DIFF
--- a/src/command_modules/azure-cli-storage/azure/cli/command_modules/storage/_params.py
+++ b/src/command_modules/azure-cli-storage/azure/cli/command_modules/storage/_params.py
@@ -381,7 +381,9 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals, too-many-statem
                    arg_type=get_enum_type(get_container_access_type_names()),
                    help='Specifies whether data in the container may be accessed publically. By default, container '
                         'data is private ("off") to the account owner. Use "blob" to allow public read access for '
-                        'blobs. Use "container" to allow public read and list access to the entire container.')
+                        'blobs. Use "container" to allow public read and list access to the entire container. You '
+                        'can configure the --public-access using `az storage container set-permission -n '
+                        'CONTAINER_NAME --public-access blob/container/off`')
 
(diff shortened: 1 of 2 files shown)


src/command_modules/azure-cli-storage/azure/cli/command_modules/storage/_params.py
```diff
@@ -381,7 +381,9 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals, too-many-statem
                    arg_type=get_enum_type(get_container_access_type_names()),
                    help='Specifies whether data in the container may be accessed publically. By default, container '
```
tjprescott: This is way too long for "short help". IMO this help should be moved to `_help.py` and the language starting with "By default..." should be moved to the long summary. 

Nit: "publically" should really be "publicly".

limingu: Updated,

PR #8886, opened 2019-03-26
webapp: Updating AppService to use the new Python SDK version

panchagnula:
---


src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/custom.py
```diff
@@ -1162,7 +1162,7 @@ def update_app_service_plan(instance, sku=None, number_of_workers=None,
 
     instance.sku = sku_def
     if admin_site_name is not None:
-        Logger.warning("The 'admin_site_name' property has been deprecated")
+        logger.warning("The 'admin_site_name' property has been deprecated")
```
tjprescott: This should use the CLI's built-in deprecation mechanisms.

panchagnula: @tjprescott can you point me to an example of this when deprecating a property - the examples i saw were all where the command was being deprecated. thank you!

panchagnula:
Also @tjprescott  i see this error in the build -> "azure-cli-appservice 0.2.17 has requirement azure-mgmt-web==0.41.0, but you'll have azure-mgmt-web 0.40.0 which is incompatible." though i updated the setup.py as a part of my PR.  I don't understand this error?

yugangw-msft:
@panchagnula, the reason is the botservice  also references the azure-mgmt-web, so do update their [command module](https://github.com/Azure/azure-cli/blob/dev/src/command_modules/azure-cli-botservice/setup.py#L39) as well

src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/custom.py
stevengum: @panchagnula, please see https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#deprecating-commands-and-arguments

panchagnula: thank you.

panchagnula:
@yugangw-msft & @tjprescott , removed one flaky test that was failing while recording filed the following to rack & fix this issue https://github.com/Azure/azure-cli/issues/8969 going to try & have this by tomorrow. Just trying to unblock other PRs that depend on this change as well.

tjprescott:
@yugangw-msft look good to you?

src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/tests/latest/test_webapp_commands.py
```diff
@@ -1186,10 +1186,6 @@ def test_webapp_assign_system_identity(self, resource_group):
             self.cmd('webapp identity show -g {} -n {}'.format(resource_group, webapp_name), checks=[
                 self.check('principalId', result['principalId'])
             ])
-        self.cmd('role assignment list -g {} --assignee {}'.format(resource_group, result['principalId']), checks=[
```
yugangw-msft: why remove this?

panchagnula: Two reasons
1. the check did not seem to add any value to this specific test for MSI assignment for webapps
2. This call for role assignment list is returning an empty object -> even without my changes , ex. if i run this on PROD CLI  
image

I was trying not to take dependency on non webapp calls here - also honestly i don't understand the role assignment call & what the expected value should be here. If you believe this is valuable for the test i can create a bug to investigate & add this back, but i don't think this is blocking this PR. what do you think?

yugangw-msft: We have a big regression here, if `role assignment list` returns nothing. Note, `'webapp identity assign` is supposed to create such a role assignment for you. Please do cross check here.

panchagnula: Cross checked & if this is a regression i will start an investigation with the backend folks on this, but this is the behavior i am seeing on PROD as well. Can we keep this investigation orthogonal to the PR?

yugangw-msft: // CC: @praries880 who is working on the Java SDK support on this. Please post update here.

PR #8896, opened 2019-03-27
Pin wheel version.

tjprescott:
CLI extensions rely on wheel files containing metadata and thus require they be built using wheel 0.29.0 or 0.30.0.  Recent releases are not compatible and updating the CLI results in wheel 0.31.1.  This PR pins the version.

---

### DIFF
--- a/src/azure-cli-core/setup.py
+++ b/src/azure-cli-core/setup.py
@@ -71,7 +71,7 @@ DEPENDENCIES = [
     'requests>=2.20.0',
     'six',
     'tabulate>=0.7.7',
-    'wheel',
+    'wheel==0.30.0',
     'azure-mgmt-resource==2.1.0'
 ]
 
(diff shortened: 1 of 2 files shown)


yugangw-msft:
I assume this won't impact the homebrew fix

tjprescott:
https://github.com/Azure/azure-cli/pull/8744
@marstr for comment. The reason for pinning to 0.30.0 was not solely due to the WHEEL_INFO_RE. It is more fundamentally to do with https://github.com/pypa/wheel/issues/195 which is called out in the error message you will receive if your extension is built with anything other than 0.29.0 or 0.30.0.

Right now, the unpinning of this is giving extension developers a version of wheel that the CI will reject and also broke the extension docs CI. I think we should revert this change until we see how much work is involved in making our entire extensions infrastructure compatible with the latest versions of wheel.

Issue #5575, opened 2018-02-13
Misbehaving `--dryrun` flag in `az storage blob delete-batch`


elibarzilay:
(Ubuntu, azure-cli 2.0.27, and showing storage as 2.0.25.)

I've tried playing with the new `delete-batch`, and could not figure out
how to use `--if-modified-since` or `--if-unmodified-since` -- no matter
what I did, it seems to always insist on deleting *everything*.  After
lots of fiddling around, I used a new container just to experiment with,
and to my surprise, it looks like `--dryrun` just ignores these flags,
but when it's not used, there are not ignored (but have other, unrelated
issues).

If this is correct then this is a very bad problem, since it is common
to use a dry run flag to see what will be deleted, and ignoring these
flags means that I can end up deleting blobs unexpectedly.

williexu:
The parameters are used to limit the operations according to whether the resource has been modified or not. 
None of the batch commands take these parameters into account upon `--dryrun`.

`--dryrun` will simply list all blobs under **consideration** for deletion/upload.
This will always be a superset of the actual blobs that would be deleted/uploaded.
Only once the actual command is run will the properties be applied on a blob by blob basis.

elibarzilay:
@williexu This is pretty bad in two ways:

1. It is a definition of "dry run" that is very different from any other
   use of the term.  This should be made very clear in both the `-h`
   text and in the output that it produces because it's so unexpected.

2. There *should* be some way for people to do a proper dry run.
   Consider writing a script that purges old files: it is very common to
   write such a script in a dry run mode to verify that it would delete
   the correct blobs, and then just remove the dry run flag.

   If I understand it correctly, there is no way other than either
   creating a dummy container for testing (which is a major PITA, since
   there is no way to change the last-modified time, so it's a tedious
   process of copying files, waiting some time, etc).

williexu:
@seguler If we want to do this, it would have to be done through server support for batch-commands.
The CLI simply passes the `modified` parameters to the service to process and cannot be done for a dryrun.

arbaiti:
Hello,
I need to do cleanup of dump files on a blob container, and i don't know how to use this command.
I should delete dump older than 14 days.
Please if you have an example ?
Regards,

williexu:
@arbaiti if you still need an example:
```
az storage blob delete-batch --source backups --if-unmodified-since `date -d "14 days  ago" '+%Y-%m-%dT%H:%MZ'`
```

williexu:
[...]

PR #8915, opened 2019-03-28
[Storage] Add more help message for storage blob delete-batch --dryrun

### LINKED ISSUE #6113: `az storage blob delete-batch` - `--if-unmodified-since` not working
```bash
...
DATE="$(date -ud "10 years ago" '+%Y-%m-%dT%H:%MZ')"
echo "$DATE"
...
    az storage blob delete-batch \
        --account-name "$NAME" \
        --account-key "$KEY" \
        --source "$CONTAINER" \
        --if-unmodified-since "$DATE" \
        --dryrun
...
```

I have some blobs modified in 2017, and I've tried `10 years ago` (`2008-04-13T16:06Z`) and `10 years` (`2028-04-13T16:06Z`) and both produce the same output:
```
delete action: from foo
    pattern None
  container foo
      total 67
 operations
  - FILE1
  - ...
...
```

No matter the date, I can't seem to get it to change its output 🤔 

---

### Environment summary

 - python2
 - `pip2 install 'azure-cli==2.0.31'` (also affects `2.0.28`)

limingu:
Fixes #6113
---

### DIFF
--- a/src/command_modules/azure-cli-storage/azure/cli/command_modules/storage/operations/blob.py
+++ b/src/command_modules/azure-cli-storage/azure/cli/command_modules/storage/operations/blob.py
@@ -393,6 +393,10 @@ def storage_blob_delete_batch(client, source, source_container_name, pattern=Non
     source_blobs = list(collect_blobs(client, source_container_name, pattern))
 
     if dryrun:
+        if if_modified_since:
+            logger.warning('--if-modified-since argument is ignored.')
+        if if_unmodified_since:
+            logger.warning('--if-unmodified-since argument is ignored.')
         logger.warning('delete action: from %s', source)
         logger.warning('    pattern %s', pattern)
         logger.warning('  container %s', source_container_name)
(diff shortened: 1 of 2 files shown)


tjprescott:
Wouldn't it make more sense to simply make `--dryrun` actually do the comparisons so that the list produced is accurate?

limingu:
> Wouldn't it make more sense to simply make `--dryrun` actually do the comparisons so that the list produced is accurate?

It needs server-side support `--dryrun` to get the accurate list. See in #5575

tjprescott:
@limingu makes sense. Maybe we could pull down the list to get the modified times, but that sort of negates the performance benefit of dry run (and maybe that info isn't even returned). However, we should then emit a warning if the user specifies the conflicting options that the --if-modified arguments are being ignored.

src/command_modules/azure-cli-storage/azure/cli/command_modules/storage/operations/blob.py
```diff
@@ -393,6 +393,10 @@ def _delete_blob(blob_name):
     source_blobs = list(collect_blobs(client, source_container_name, pattern))
 
     if dryrun:
+        if if_modified_since:
+            logger.warning('--if-modified-since argument is ignored.')
+        if if_unmodified_since:
+            logger.warning('--if-unmodified-since argument is ignored.')
```
tjprescott: I would be more specific here. `--foo argument is ignored when using --dry-run`.

PR #8919, opened 2019-03-29
Add support for managed instance's proxy override and public data endpoint enabled

stefanb995:
---


stefanb995:
Is there a way to specify argument type to be bool?

There is currently a regression in westcentralus and I'm not able to run the test as Managed Instance creation gets stuck . Once the region is refreshed with a fix, I'll rerun the test and update test recordings.

src/command_modules/azure-cli-sql/azure/cli/command_modules/sql/_params.py
```diff
@@ -1076,6 +1076,13 @@ def _configure_security_policy_storage_params(arg_ctx):
         c.argument('collation',
                    help='The collation of the managed instance.')
 
+        c.argument('proxy_override',
+                   arg_type=get_enum_type(ServerConnectionType),
+                   help='The connection type used for connecting to the instance.')
+
+        c.argument('public_data_endpoint_enabled',
+                   help='Whether or not the public data endpoint is enabled for the instance.')
```
jaredmoo: You can add `arg_type=get_three_state_flag()` here so it will be treated like `SwitchParameter` in powershell

src/command_modules/azure-cli-sql/azure/cli/command_modules/sql/_params.py
```diff
@@ -1086,7 +1093,8 @@ def _configure_security_policy_storage_params(arg_ctx):
                 'virtual_network_subnet_id',
                 'vcores',
                 'storage_size_in_gb',
-                'collation'
+                'collation',
+                'proxy_override',
```
jaredmoo: If you add `public_data_endpoint_enabled` here, then you should not need to explicitly add it in `custom.py`

src/command_modules/azure-cli-sql/azure/cli/command_modules/sql/_params.py
stefanb995: Great. Thanks

src/command_modules/azure-cli-sql/azure/cli/command_modules/sql/_params.py
stefanb995: Done

Issue #8666, opened 2019-02-28
Az webapp up should enable logging by default


yiliaomsft:
Please refer to https://docs.microsoft.com/en-us/cli/azure/webapp/log?view=azure-cli-latest#az-webapp-log-config

tjprescott:
@panchagnula for consideration.

adewaleo:
@panchagnula moving this to sprint 57

panchagnula:
@yiliaomsft , looking more into this, turning on this will cause an app recycle & so we don't turn this on by default on Portal either. Please comment if you still think this is required.

panchagnula:
Added a --logs parameter for now as an optional parameter

PR #8920, opened 2019-03-29
webapp: az webapp up improvements

panchagnula:
1. Node default version for Linux updated to 10.14 #8914
2. Added support to enable and view streaming logs using --logs argument (thank you @yugangw-msft ) #8666
3. the default naming convention for RG & ASP uses the logged in user name instead of appsvc as the prefix
4. Updated the help text for az webapp up
5. Zip deployment automatically enables logs & displays log (not streaming) before marking the deployment as Failure
---

### DIFF
--- a/src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/_constants.py
+++ b/src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/_constants.py
@@ -3,7 +3,7 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 
-NODE_VERSION_DEFAULT = "8.11"
+NODE_VERSION_DEFAULT = "10.14"
 NETCORE_VERSION_DEFAULT = "2.0"
 DOTNET_VERSION_DEFAULT = "4.7"
 PYTHON_VERSION_DEFAULT = "3.7"
--- a/src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/custom.py
+++ b/src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/custom.py
@@ -2079,7 +2079,7 @@ def create_function(cmd, resource_group_name, name, storage_account, plan=None,
     # adding appsetting to site to make it a function
     site_config.app_settings.append(NameValuePair(name='AzureWebJobsStorage', value=con_string))
     site_config.app_settings.append(NameValuePair(name='AzureWebJobsDashboard', value=con_string))
-    site_config.app_settings.append(NameValuePair(name='WEBSITE_NODE_DEFAULT_VERSION', value='8.11.1'))
+    site_config.app_settings.append(NameValuePair(name='WEBSITE_NODE_DEFAULT_VERSION', value='10.14.1'))
 
     if consumption_plan_location is None and not is_plan_elastic_premium(plan_info):
         site_config.always_on = True
--- a/src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/custom.py
+++ b/src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/custom.py
@@ -53,7 +53,8 @@ from ._appservice_utils import _generic_site_operation
 from ._create_util import (zip_contents_from_dir, get_runtime_version_details, create_resource_group,
                            should_create_new_rg, set_location, should_create_new_asp, should_create_new_app,
                            get_lang_from_content, get_num_apps_in_asp)
-from ._constants import (NODE_RUNTIME_NAME, OS_DEFAULT, STATIC_RUNTIME_NAME, PYTHON_RUNTIME_NAME, RUNTIME_TO_IMAGE)
+from ._constants import (NODE_RUNTIME_NAME, OS_DEFAULT, STATIC_RUNTIME_NAME, PYTHON_RUNTIME_NAME,
+                         RUNTIME_TO_IMAGE, NODE_VERSION_DEFAULT)
 
 logger = get_logger(__name__)
 
@@ -78,7 +79,7 @@ def create_webapp(cmd, resource_group_name, name, plan, runtime=None, startup_fi
     if not plan_info:
         raise CLIError("The plan '{}' doesn't exist".format(plan))
     is_linux = plan_info.reserved
-    node_default_version = '8.11.1'
+    node_default_version = NODE_VERSION_DEFAULT
     location = plan_info.location
     site_config = SiteConfig(app_settings=[])
     webapp_def = Site(location=location, site_config=site_config, server_farm_id=plan_info.id, tags=tags)
@@ -288,7 +289,8 @@ def enable_zip_deploy(cmd, resource_group_name, name, src, timeout=None, slot=No
         logger.warning("Starting zip deployment")
         requests.post(zip_url, data=zip_content, headers=headers)
     # check the status of async deployment
-    response = _check_zip_deployment_status(deployment_status_url, authorization, timeout)
+    response = _check_zip_deployment_status(cmd, resource_group_name, name, deployment_status_url,
+                                            authorization, timeout)
     return response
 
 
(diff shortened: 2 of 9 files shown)


panchagnula:
TODO: 
1. Update history.rst, 
2. noticed that for Linux apps the HttpLogging is not getting enabled using the CLI command - investigating to check if this is an SDK API issue or CLI command issue, since enabling this directly from resources.azure.com is successful
3. received feedback from Linux crew to enable Always On by default for better performance on re- deployment scenarios

panchagnula:
@yugangw-msft since you have the most context of this command - can you take a look at this PR once the build is Green. Thanks!

panchagnula:
This has the SDK changes merged to this branch as well. Will rebase to DEV once the sdk PR is approved & checked-in.

src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/_constants.py
```diff
@@ -3,7 +3,7 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 
-NODE_VERSION_DEFAULT = "8.11"
+NODE_VERSION_DEFAULT = "10.14"
```
panchagnula: @yiliaomsft & @ahmedelnably should the function app create use the new Default Version for Node on Linux as well? Currently it hard-coded to use 8.11.1. Thanks!

src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2095,25 +2092,37 @@ def list_locations(cmd, sku, linux_workers_enabled=None):
...
     while num_trials < total_trials:
         time.sleep(2)
         response = requests.get(deployment_status_url, headers=authorization)
+        time.sleep(2)
         res_dict = response.json()
         num_trials = num_trials + 1
         if res_dict.get('status', 0) == 3:
-            raise CLIError("Zip deployment failed. {}".format(res_dict))
+            # Turn on diagnostic logs
+            config_diagnostics(cmd, resource_group_name=rg_name, name=name,
```
yugangw-msft: i suggest you create a shared routine for this. The ideal way is to enable the minimum and do it only it was not configured; otherwise this risks wiping out the existing configuration made by the user

src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2095,25 +2092,37 @@ def list_locations(cmd, sku, linux_workers_enabled=None):
...
+                               application_logging=True, web_server_logging='filesystem',
+                               docker_container_logging='filesystem', detailed_error_messages=True)
+            logger.warning('Please run the command ')
```
yugangw-msft: this warning text seems redundant.

src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2095,25 +2092,37 @@ def list_locations(cmd, sku, linux_workers_enabled=None):
...
         response = requests.get(deployment_status_url, headers=authorization)
+        time.sleep(2)
         res_dict = response.json()
         num_trials = num_trials + 1
         if res_dict.get('status', 0) == 3:
-            raise CLIError("Zip deployment failed. {}".format(res_dict))
+            # Turn on diagnostic logs
+            config_diagnostics(cmd, resource_group_name=rg_name, name=name,
+                               application_logging=True, web_server_logging='filesystem',
+                               docker_container_logging='filesystem', detailed_error_messages=True)
```
yugangw-msft: We should emit out a warning so users know we are turning it on for them.

src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2292,17 +2305,16 @@ def create_deploy_webapp(cmd, name, location=None, sku=None, dryrun=False):  # p
...
-                if key.upper() == 'SCM_DO_BUILD_DURING_DEPLOYMENT' and value.upper() == "FALSE":
-                    _set_build_app_setting = False
+    if do_deployment and not is_skip_build:
+        _set_build_app_setting = True
+        # setting the appsettings causes a app restart so we avoid if not needed
+        application_settings = client.web_apps.list_application_settings(rg_name, name)
+        _app_settings = application_settings.properties
+        for key, value in _app_settings.items():
+            if key.upper() == 'SCM_DO_BUILD_DURING_DEPLOYMENT' and value.upper() == "FALSE":
+                _set_build_app_setting = False
```
yugangw-msft: you can break out here

src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2322,13 +2334,17 @@ def create_deploy_webapp(cmd, name, location=None, sku=None, dryrun=False):  # p
             os.remove(zip_file_path)
         except OSError:
             pass
-    create_json.update({'app_url': url})
-    logger.warning("All done.")
-    return create_json
+    if logs:
+        logger.waring("Configure logging for a web app")
+        config_diagnostics(cmd, rg_name, name,
```
yugangw-msft: like said, try create a common routine with a bit intelligence that to avoid overwrite users' existing logging config

src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/_constants.py
panchagnula: CC: @ankitkumarr can you follow up Ahmed & check if you need this change for functionapp for linux & handle it accordingly, currently functionapp create is setting 8.11.1 as default version for both Linux & windows.

src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/custom.py
panchagnula: Good point, may be the detailed error messages is not a normal default but docker-logging is ignored in windows & needed for Linux, so will keep that.

src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/custom.py
panchagnula: yeah no idea what i was thinking when i wrote this line of code :-/

src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/_constants.py
ColbyTresness: Yeah we should now default to 10.

panchagnula: Done updated this to 10.14.1

tjprescott:
@yugangw-msft ping

src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2523,3 +2535,10 @@ def create_devops_build(cmd, functionapp_name=None, organization_name=None, proj
                                                                  organization_name, project_name, repository_name,
                                                                  overwrite_yaml, allow_force_push, use_local_settings)
     return azure_devops_build_interactive.interactive_azure_devops_build()
+
+
+def _configure_default_logging(cmd, rg_name, name):
```
yugangw-msft: Make sure do it in the follow-up PR that you should do nothing if the diagnostics are already configured

panchagnula: Just want to clarify this- the API will not change anything if these specific logs are turned on. 
With regards to a scenario where a user might have Blob stoarage enabled & we are enabling fileSystem, then this will not turn off blobStorage one. On our end we allow both of these to be enabled. For logstreaming to display messages on console the FileSystem & weblogging should be enabled. If you still think this might be a problem i will file an issue to track this - especially we get some feedback around this.

PR #8922, opened 2019-03-29
`az --version`: fail-fast on connectivity problems

### LINKED ISSUE #8762: v2.0.59 and v2.0.60 az --version hangs
**Describe the bug**
With version 2.0.59 or 2.0.60 installed executing az --version from an admin command prompt just hangs.

**To Reproduce**
Install Azure CLI
Open cmd admin prompt
Execute az --version
Hangs!

**Expected behavior**
It shouldn't hang.

**Environment summary**
Windows MSI installer v 2.0.59 or 2.0.60 downloaded from https://docs.microsoft.com/en-us/cli/azure/install-azure-cli?view=azure-cli-latest

**Additional context**

tjprescott:
Fixes #8762.

Adds a utility method `check_connectivity` to speed up the process of timing out on connectivity problems. Previously we just sent the request and waited for it to fail, but this proved far too slow. This method is designed to fail fast by default (though a connectivity failure is still slower than a success).

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/util.py
+++ b/src/azure-cli-core/azure/cli/core/util.py
@@ -88,23 +88,24 @@ def _update_latest_from_pypi(versions):
     from subprocess import check_output, STDOUT, CalledProcessError
 
     success = False
-    try:
-        cmd = [sys.executable] + '-m pip search azure-cli -vv --disable-pip-version-check --no-cache-dir'.split()
-        logger.debug('Running: %s', cmd)
-        log_output = check_output(cmd, stderr=STDOUT, universal_newlines=True)
-        success = True
-        for line in log_output.splitlines():
-            if not line.startswith(CLI_PACKAGE_NAME):
-                continue
-            comps = line.split()
-            mod = comps[0].replace(COMPONENT_PREFIX, '') or CLI_PACKAGE_NAME
-            version = comps[1].replace('(', '').replace(')', '')
-            try:
-                versions[mod]['pypi'] = version
-            except KeyError:
-                pass
-    except CalledProcessError:
-        pass
+
+    if not check_connectivity():
+        return versions, success
+
+    cmd = [sys.executable] + '-m pip search azure-cli -vv --disable-pip-version-check --no-cache-dir --retries 0'.split()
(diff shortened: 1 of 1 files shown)


src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -459,3 +460,22 @@ def find_child_collection(parent, *args, **kwargs):
     if collection is None:
         raise CLIError("collection '{}' not found".format(collection_path))
     return collection
+
+
+def check_connectivity(url='https://example.org', max_retries=0, timeout=1):
```
limingu: Is https://example.org reliable? Should we use azure.microsoft.com or other more reliable url to test the connectivity?

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -88,23 +88,24 @@ def _update_latest_from_pypi(versions):
...
-            if not line.startswith(CLI_PACKAGE_NAME):
-                continue
-            comps = line.split()
-            mod = comps[0].replace(COMPONENT_PREFIX, '') or CLI_PACKAGE_NAME
-            version = comps[1].replace('(', '').replace(')', '')
-            try:
-                versions[mod]['pypi'] = version
-            except KeyError:
-                pass
-    except CalledProcessError:
```
marstr: IMHO, I would keep this try/catch for pip failure even if we've added a fail-fast above. It'll keep the robust behavior before, with the improved fail-fast. I think the counter argument, which I can't really speak to, is that the Python idiom would be to have the exception propagate up instead of returning a bool to indicate that the check was unsuccessful.

marstr:
Just to capture this thought somewhere: This change makes us more sensitive to transient network problems than we previously were, with the benefit of not hanging during a version check. For just about any other function, we would want to continue being robust. So, this behavior shouldn't be copied by future command authors. I do think it is the right thing to do in this instance though.

tjprescott:
@marstr comments addressed. Also, I set the default on max_retries to something more generally useful and then simply override it where I use it to be 0. This partially addressed your comment on transient network issues, which I agree with.

PR #8928, opened 2019-03-29
Run CI for all branches in this repo by default.

marstr:
This will enable white-listing of branches for private-preview drops in our edge release pipeline.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -5,9 +5,7 @@ trigger:
   batch: true
   branches:
     include:
-    - master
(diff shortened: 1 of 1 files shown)


marstr:
You can review the syntax here: https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml

But basically, because we want to batch builds to save compute and be a good citizen in the ADO cluster, we need to specifically provide a trigger clause in our pipeline definition.

PR #8936, opened 2019-04-01
3842802: Automatic name server delegation in parent zone when child z…

ShwethaRaghupathi:
…one created

- Enable DNS name server delegation when a DNS zone is created whose parent zone is already hosted in Azure.
- For successful resolution of the newly added child zone, an NS record must exist in the parent containing the name servers of the newly created child zone.
- Achieve this additional step as part of zone creation when user indicates the parent zone details as part of -p or --parent-name parameter

**Current state**
- It supports parent zone present in same resource group & subscription as the child zone being created.

**Intended future state**
- Since the parent zone might reside in another subscription, we propose to accept FQDN(fully qualified domain name) of the parent and extract resource group , subscription details

**Open Question ?**
Would it be possible to create two resources in different subscriptions as part of this single cmd ?
If so how ? 

---

### DIFF
--- a/src/command_modules/azure-cli-network/azure/cli/command_modules/network/custom.py
+++ b/src/command_modules/azure-cli-network/azure/cli/command_modules/network/custom.py
@@ -1086,6 +1086,7 @@ def add_dns_delegation(cmd, created_zone, parent_zone_name, resource_group_name,
             print(ex)
             print('Could not add delegation in \'{}\'\n'.format(parent_zone_name), file=sys.stderr)
 
+
 def create_dns_zone(cmd, client, resource_group_name, zone_name, parent_zone_name=None, tags=None,
                     if_none_match=False, zone_type='Public', resolution_vnets=None, registration_vnets=None):
     Zone = cmd.get_models('Zone', resource_type=ResourceType.MGMT_NETWORK_DNS)
--- a/src/command_modules/azure-cli-network/azure/cli/command_modules/network/custom.py
+++ b/src/command_modules/azure-cli-network/azure/cli/command_modules/network/custom.py
@@ -1074,7 +1074,19 @@ def list_ddos_plans(cmd, resource_group_name=None):
 
 
 # region DNS Commands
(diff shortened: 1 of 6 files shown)


ShwethaRaghupathi:
+Team @SoSain

src/command_modules/azure-cli-network/azure/cli/command_modules/network/custom.py
```diff
@@ -1084,7 +1115,13 @@ def create_dns_zone(cmd, client, resource_group_name, zone_name, tags=None,
         zone.registration_virtual_networks = registration_vnets
         zone.resolution_virtual_networks = resolution_vnets
 
-    return client.create_or_update(resource_group_name, zone_name, zone, if_none_match='*' if if_none_match else None)
+    created_zone = client.create_or_update(resource_group_name, zone_name, zone,
+                                           if_none_match='*' if if_none_match else None)
+
+    if parent_zone_name is not None:
```
tjprescott: Best practice here would be to explicitly gate this code on api version to ensure it will not be reached on older profiles.

```Python
if cmd.supported_api_version(min_api='BLAH') and parent_zone_name is not None:
    #logic
```

ShwethaRaghupathi: @tjprescott  So next release version is 2.3.6  & assuming this change goes in as part of that release, min_api would be 2.3.6 ?

tjprescott: Min api would look something like 2019-04-01 or whatever the API version is for your service, not the CLI version.

ShwethaRaghupathi:
@tjprescott Thank you for your inputs, updated the PR as per the review comments. Please take a look.

tjprescott:
@ShwethaRaghupathi please resolve the static issue. If you have azdev, you can simply run `azdev style network` locally. If you believe the broad except is okay, you can disable it for that line.

PR #8979, opened 2019-04-03
[functionapp] fixes #8913 - allow function apps to change plan

### LINKED ISSUE #8913: Allow Function Apps to change plan
This command would allow you to change a Function App from a consumption to a premium plan.

The command would be something akin to:

Parameters:
-	SourceApp (the details for a function app already deployed)
-	DestinationPlan (the details for a plan you want to move it to - this likely needs to be in the same resource group so we don’t get into stamp incompatibility issues - or just make sure we have a warning if it fails because that plan is incompatible)

Steps:
1.	Get the resource ID of the destination plan /Microsoft.Web/serverfarm
2.	Update the Microsoft.Web/sites resource for the function app plan.
a.	Properties.serverFarmId should be set to the resource ID from step 1

Ideally the command should validate that the sourceapp plan is consumption / premium (via the SKU of the plan), and that the destination plan is consumption / premium.  Moving between ASP and consumption / premium may work but you need tom ake sure the WEBSITE_CONTENT_CONNECTIONSTRING / CONTENTSHARE settings are set before you do (by default dedicated won’t have these, you need them for consumption / premium

ankitkumarr:
@panchagnula @jeffhollan @alexkarcher-msft @ahmedelnably 

---

### DIFF
--- a/src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/custom.py
+++ b/src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/custom.py
@@ -367,6 +367,36 @@ def update_webapp(instance, client_affinity_enabled=None, https_only=None):
     return instance
 
 
+def update_functionapp(cmd, instance, plan=None):
+    client = web_client_factory(cmd.cli_ctx)
+    if plan is not None:
+        if is_valid_resource_id(plan):
+            dest_parse_result = parse_resource_id(plan)
+            dest_plan_info = client.app_service_plans.get(dest_parse_result['resource_group'],
+                                                          dest_parse_result['name'])
+        else:
+            dest_plan_info = client.app_service_plans.get(instance.resource_group, plan)
+        if dest_plan_info is None:
+            raise CLIError("The plan '{}' doesn't exist".format(plan))
+        validate_plan_switch_compatibility(client, instance, dest_plan_info)
+        instance.server_farm_id = dest_plan_info.id
+    return instance
+
+
+def validate_plan_switch_compatibility(client, src_functionapp_instance, dest_plan_instance):
+    general_switch_msg = 'Currently the switch is only allowed between a Consumption or an Elastic Premium plan.'
+    src_parse_result = parse_resource_id(src_functionapp_instance.server_farm_id)
+    src_plan_info = client.app_service_plans.get(src_parse_result['resource_group'],
+                                                 src_parse_result['name'])
+    if src_plan_info is None:
+        raise CLIError('Could not determine the current plan of the functionapp')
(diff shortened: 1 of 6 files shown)


src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/custom.py
```diff
@@ -367,12 +367,66 @@ def update_webapp(instance, client_affinity_enabled=None, https_only=None):
...
+
 def set_functionapp(cmd, resource_group_name, name, **kwargs):
     instance = kwargs['parameters']
     if 'function' not in instance.kind:
         raise CLIError('Not a function app to update')
     client = web_client_factory(cmd.cli_ctx)
-    return client.web_apps.create_or_update(resource_group_name, name, site_envelope=instance)
+    try:
+        return client.web_apps.create_or_update(resource_group_name, name, site_envelope=instance)
+    except DefaultErrorResponseException as ex:
```
panchagnula: @ankitkumarr for exception handling of this sort we have a helper in commands.py we initiate it by adding this to the command on the commands.py exception_handler=ex_handler_factory(), ex. see   g.custom_command('create', 'create_function', exception_handler=ex_handler_factory()) -> can you check if this is something you can utilize here for this method as well?

ankitkumarr: Thanks @panchagnula, I will update it to use that factor directly.

src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/custom.py
```diff
@@ -367,6 +367,39 @@ def update_webapp(instance, client_affinity_enabled=None, https_only=None):
...
+
+def validate_plan_switch_compatibility(client, src_functionapp_instance, dest_plan_instance):
+    general_switch_msg = 'Currently the switch is only allowed between a Consumption or an Elastic Premium plan.'
+    src_parse_result = parse_resource_id(src_functionapp_instance.server_farm_id)
+    src_plan_info = client.app_service_plans.get(src_parse_result['resource_group'],
+                                                 src_parse_result['name'])
+    if src_plan_info is None:
+        # if kind == functionapp, it means the app is using a Consumption plan
+        # Sometimes retrieving the exact plan fails, so it's ok as long as we know they are in Consumption
+        if src_functionapp_instance.kind.lower() != 'functionapp':
```
ankitkumarr: Need to update this. This is not a good way to check.

ankitkumarr: I verified with people involved and there's no need to perform additional checks. This should be resolved.

ankitkumarr:
This should be ready to be merged in.

PR #8981, opened 2019-04-03
Enable Homebrew Build for PR validation

marstr:
The build has been broken a number of times lately, in ways that would be caught if building the Hombrew Formula were part of PR validation. I'm turning it on to help catch these issues sooner rather than later.


tjprescott:
@marstr what's the status of this PR?

marstr:
Ugh, it pointed out a chicken-and-egg problem that we have with dependency resolution. It should be closed for now. Once we've merged all commands into a single distribution we can do this again.

PR #8995, opened 2019-04-04
Removing submodule

marstr:
Seems like this was added erroneously in #8920, please correct me if this isn't the case.


panchagnula:
Sorry & thanks for taking care of this.

PR #9015, opened 2019-04-05
[ACR] Cleanup runtime access token scope

djyou:
---

### DIFF
--- a/src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_docker_utils.py
+++ b/src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_docker_utils.py
@@ -35,7 +35,7 @@ logger = get_logger(__name__)
 
 EMPTY_GUID = '00000000-0000-0000-0000-000000000000'
 ALLOWED_HTTP_METHOD = ['get', 'patch', 'put', 'delete']
-ACCESS_TOKEN_PERMISSION = ['*', 'pull']
+ACCESS_TOKEN_PERMISSION = ['pull', 'push', 'delete', 'push,pull', 'delete,pull']
 
 AAD_TOKEN_BASE_ERROR_MESSAGE = "Unable to get AAD authorization tokens with message"
 ADMIN_USER_BASE_ERROR_MESSAGE = "Unable to get admin user credentials with message"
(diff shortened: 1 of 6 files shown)


src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_docker_utils.py
```diff
@@ -35,7 +35,7 @@
 
 EMPTY_GUID = '00000000-0000-0000-0000-000000000000'
 ALLOWED_HTTP_METHOD = ['get', 'patch', 'put', 'delete']
-ACCESS_TOKEN_PERMISSION = ['*', 'pull']
+ACCESS_TOKEN_PERMISSION = ['pull', 'push', 'delete', 'push,pull', 'delete,pull']
```
tjprescott: This should really just be `['pull', 'push', 'delete']` and access a space-separated list of permissions. The service could either validate which combinations were valid or client-side validation, but we don't allow comma-separated lists in the CLI.

djyou: We probably can just remove this list and client-side validation. The client has to send `push,pull` as one word as the service didn't provide a way to specify a list.

djyou:
@yuwaMSFT2 @sajayantony

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_docker_utils.py
tjprescott: Since we do not allow comma separated lists, you would need to expose it as a list and then use a type callable or validator to convert to the appropriate values the service accepts.

CLI
`--permissions push pull`

Service
`"push,pull"`

tjprescott: Alternatively, you could keep the values list, and just call "push,pull" as "push-pull".  Then you have a type callable that replaces "-" with "," to send the properly formatted value to the service.

djyou: I see where the confusion was. This is an internal method and we don't expose permissions to customers in the CLI.

tjprescott: Ah, okay. In that case it is fine.

tjprescott:
@djyou thanks for the clarification. @yuwaMSFT2 or @sajayantony please give it the okay before I merge.

sajayantony:
will defer to @yuwaMSFT2 on this.

yuwaMSFT2:
LGTM

PR #9020, opened 2019-04-08
Add time zone parameter to the managed instance create command.

NMijat1024:
---


src/command_modules/azure-cli-sql/HISTORY.rst
```diff
@@ -2,6 +2,9 @@
 
 Release History
 ===============
+2.2.2
++++++
+* Support for custom time zone on managed instance.
 
 2.2.1
 +++++
```
jaredmoo: I see that 2.2.1 was released without my "Allow elastic pool name..." change, I put it in the wrong version header. Can you move it into 2.2.2 ?

jaredmoo: Never mind, already merged. I'll fix it.

PR #9028, opened 2019-04-09
[core] Improve error message when get_file_json / shell_safe_json_parse fails.

### LINKED ISSUE #7740: Create/Validate deployment template commands should handle improper syntax more gracefully
**Describe the bug**

Not able to deploy an Azure App as a docker container pulled from a private Azure Container Registry.
Seeing an error message referring to some `malformed node or string` (see below).

**To Reproduce**

Run this command:

```sh
az group deployment create \
  --name "deployAzureApp" \
  --resource-group MyRG \
  --template-file ./ci/arm/function-app.json  --verbose --debug
```

Then see this at the bottom of the stdout/stderr:

```
malformed node or string: <_ast.Name object at 0x7f3cecc8add8>
```

The content of the template is this:

```
{
  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
  "contentVersion": "1.0.0.0",
  "parameters": {
    "appName": {
      "type": "String",
      "value": "<PUT_APP_NAME_HERE>"
    },
    "dockerImageName": {
      "type": "String",
      "value": "<NAMESPACE>/<IMAGE_NAME>:latest"
    },
    "dockerRegistryUrl": {
      "type": "String",
      "value": "https://<PUT_REGISTRY_HERE>.azurecr.io"
    },
    "dockerRegistryUsername": {
      "type": "String",
      "value": "<PUT_REGISTRY_USERNAME_HERE>"
    },
    "dockerRegistryPassword": {
      "type": "String",
      "value": "<PUT_REGISTRY_PSW_HERE>"
    },
    "servicePlanName": {
      "type": "String",
      "value": "<PUT_EXISTING_SERVICE_PLAN_HERE>"
    },
    "appLocation": {
      "type": "String",
      "value": "<PUT_AZURE_REGION_HERE>"
    }
  },
  "resources": [
    {
      "type": "Microsoft.Web/sites",
      "name": "[parameters('appName')]",
      "location": "[parameters('appLocation')]",
      "properties": {
        "name": "[parameters('appName')]",
        "siteConfig": {
          "linuxFxVersion": "[concat('DOCKER|', parameters('dockerImageName'))]",
          "alwaysOn": true,
          "appSettings": [
            {
              "name": "WEBSITES_ENABLE_APP_SERVICE_STORAGE",
              "value": "false"
            },
            {
              "name": "DOCKER_REGISTRY_SERVER_URL",
              "value": "[parameters('dockerRegistryUrl')]"
            },
            {
              "name": "DOCKER_REGISTRY_SERVER_USERNAME",
              "value": "[parameters('dockerRegistryUsername')]"
            },
            {
              "name": "DOCKER_REGISTRY_SERVER_PASSWORD",
              "value": "[parameters('dockerRegistryPassword')]"
            }
          ],
        },
        "serverFarmId": "[resourceId('Microsoft.Web/serverfarms', parameters('servicePlanName'))]"
      }
    }
  ]
}
```

The template details come from:

- https://docs.microsoft.com/en-us/azure/templates/microsoft.web/sites
- https://stackoverflow.com/a/52177725/1264920
- https://social.msdn.microsoft.com/Forums/aspnet/en-US/966b5126-f361-497f-97cb-e94832a74303/web-app-container-registry-arm-template

**Expected behavior**

Being able to deploy an Azure App.

**Environment summary**

- Install Method: `sudo apt-get install --only-upgrade azure-cli`
- Azure CLI version: `2.0.49`
- OS: Debian stretch 9.5
- Shell Type: Bash

adewaleo:
Closes #7740  

Note:
We should discuss sometime whether or not the CLI should accept python dict string syntax. I am not sure how often this syntax is used by CLI users. 
Moreover, it complicates the logic in `shell_safe_json_parse`. if a string uses python dict single quotes (invalid json) and also has a syntactical error common to both json and dict strings (extra comma), it is hard to know which error is more relevant to the user. As shell_safe_json_parse technically supports both syntaxes.

 
---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/util.py
+++ b/src/azure-cli-core/azure/cli/core/util.py
@@ -197,7 +197,10 @@ def get_file_json(file_path, throw_on_empty=True, preserve_order=False):
     content = read_file_content(file_path)
     if not content and not throw_on_empty:
         return None
-    return shell_safe_json_parse(content, preserve_order)
+    try:
+        return shell_safe_json_parse(content, preserve_order)
+    except CLIError as ex:
+        raise CLIError("Failed to parse {} with exception:\n    {}".format(file_path, ex))
 
 
 def read_file_content(file_path, allow_binary=False):
@@ -236,6 +239,9 @@ def shell_safe_json_parse(json_or_dict_string, preserve_order=False):
             return ast.literal_eval(json_or_dict_string)
         except SyntaxError:
             raise CLIError(json_ex)
+        except ValueError as ex:
+            logger.debug(ex)  # log the exception which could be a python dict parsing error.
+            raise CLIError(json_ex)  # raise json_ex error which is more readable and likely.
 
 
 def b64encode(s):


tjprescott:
Python dict syntax is permitted because Windows cmd.exe only allows use of double quotes, necessitating nonsense like this: `"{\"foo\":\"bar\"}"`

It allows `"{'foo': 'bar'}"` since the Linux equivalent `'{"foo":"bar"}'` is not possible.

adewaleo:
Even if the CLI continues to accept python dict strings, we should at least add the ability to restrict string parsing to json, as some clients of `shell_safe_json_parse` might expect json strings and not python dict strings.  E.g group deployment create/validate

adewaleo:
> Python dict syntax is permitted because Windows cmd.exe only allows use of double quotes, necessitating nonsense like this: `"{\"foo\":\"bar\"}"`
> 
> It allows `"{'foo': 'bar'}"` since the Linux equivalent `'{"foo":"bar"}'` is not possible.

@tjprescott, that's a fair point. I guess I was a bit concerned because of things like the following:

```
>>> will_it_parse = "{1: (2,3), 2: {1,4}}"
>>> it_did
{1: (2, 3), 2: {1, 4}}
>>> will_it_parse = "{1: (2,3), 2: {1,4}}"
>>> it_did = shell_safe_json_parse(will_it_parse)
>>> it_did
{1: (2, 3), 2: {1, 4}}
>>> is_it_json = json.loads(will_it_parse)
  File "<stdin>", line 1, in <module>
  ...
json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 1 column 2 (char 1)
>>> 
```

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -236,6 +239,9 @@ def shell_safe_json_parse(json_or_dict_string, preserve_order=False):
             return ast.literal_eval(json_or_dict_string)
         except SyntaxError:
             raise CLIError(json_ex)
+        except ValueError as ex:
+            logger.debug(ex)  # log the exception which could be a python dict parsing error.
+            raise CLIError(json_ex)  # raise json_ex error which is more readable and likely.
```
tjprescott: Why is it not sufficient to do the error detection only in `shell_safe_json_parse`?

adewaleo: @tjprescott, i guess i did so because get_json_file has file path information which can be added to error output, but shell_safe_json_parse does not. Moreover, some methods call get_json_file but others call shell_safe_json_parse.

adewaleo:
is this good to go

PR #9042, opened 2019-04-10
Aladdin generated examples for Storage

glGarg:
---


src/command_modules/azure-cli-storage/azure/cli/command_modules/storage/_help.py
```diff
@@ -516,6 +576,10 @@
     short-summary: Space-separated list of response headers allowed to be part of the cross-origin request.
   - name: --exposed-headers
     short-summary: Space-separated list of response headers to expose to CORS clients.
+examples:
+  - name: Add a CORS rule to a storage account. (autogenerated)
+    text: az storage cors add --allowed-headers 'Accept-Ranges,Content-Encoding,Content-Length,Content-Type,Range,Authorization,x-ms-blob-content-type,x-ms-blob-type,x-ms-version' --connection-string $current_env_conn_string --max-age 86400 --methods DELETE --origins '*' --services bfqt
```
tjprescott: The CLI does not accept comma-separated lists, so this should not be a valid example. These should be space-separated lists: `--foo A B C` not `--foo 'A,B,C'`

src/command_modules/azure-cli-storage/azure/cli/command_modules/storage/_help.py
```diff
@@ -651,6 +731,10 @@
   - name: --dryrun
     type: bool
     short-summary: List the files and blobs to be deleted. No actual data deletion will occur.
+examples:
+  - name: Delete files from an Azure Storage File Share. (autogenerated)
+    text: az storage file delete-batch --account-key 00000000 --pattern *.py --source /path/to/file
```
tjprescott: You probably need to quote `*.py` or it will cause problems in certain shells.

PR #9057, opened 2019-04-10
az feedback - Clean up formatting. Remove emoji.

tjprescott:
Tweaks the formatting and some wording of the generated template. Removes the "reaction" logic as it is not useful for debugging.


src/command_modules/azure-cli-feedback/azure/cli/command_modules/feedback/custom.py
```diff
@@ -364,15 +364,6 @@ def _get_parent_proc_name():
                 extension_info = "\nExtension Name: {}. Version: {}.".format(extension_name, extension_version)
                 is_ext = True
 
-            if errors:
```
adewaleo: 😮  😢

PR #9070, opened 2019-04-11
Aladdin generated examples for Network

glGarg:
---


src/command_modules/azure-cli-network/azure/cli/command_modules/network/_help.py
```diff
@@ -2147,6 +2157,10 @@
 helps['network express-route port update'] = """
 type: command
 short-summary: Update settings of an ExpressRoute port.
+examples:
+  - name: Update settings of an ExpressRoute port. (autogenerated)
+    text: az network express-route port update --ids vpnConnectionIds --set tags.CostCenter=MyBusinessGroup
```
tjprescott: vpnConnectionIds is not a valid ID. It should have placeholder syntax.

glGarg: Makes sense, I will disable those commands until our model has optimized better values for these parameters.

PR #9083, opened 2019-04-12
OSA - Remove FQDN parameters and update ensure_osa_aad flow

julienstroheker:
---


julienstroheker:
CC @jim-minter @mjudeikis @amanohar

src/command_modules/azure-cli-acs/azure/cli/command_modules/acs/custom.py
```diff
@@ -2564,8 +2568,15 @@ def openshift_create(cmd, client, resource_group_name, name,  # pylint: disable=
 
     try:
         # long_running_operation_timeout=300
-        return sdk_no_wait(no_wait, client.create_or_update,
-                           resource_group_name=resource_group_name, resource_name=name, parameters=osamc)
+        result = sdk_no_wait(no_wait, client.create_or_update,
+                             resource_group_name=resource_group_name, resource_name=name, parameters=osamc)
+        result = LongRunningOperation(cmd.cli_ctx)(result)
```
mboersma: We don't use `result` after this point in this method. Should it be `return`ed? Why are we making a `LongRunningOperation` out of it if it isn't used?

src/command_modules/azure-cli-acs/HISTORY.rst
```diff
@@ -2,6 +2,10 @@
 
 Release History
 ===============
+2.3.22
+++++++
+* Removing `--fqdn` flag on az openshift commands
```
tjprescott: This is a breaking change and cannot be done with a simple patch version. This needs to call out the breaking change and be bumped to version 2.4.0

julienstroheker:
@tjprescott can you give another look please ? Thanks

Issue #9037, opened 2019-04-09
[feedback] az feedback feature updates


adewaleo:
Feature Updates. When az feedback is run:

- [x] The CLI should include the autogenerated issue in the opened browser via the url. If necessary, any exceptions should be minified to stick with the defacto standard of [2000 characters](https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers).
- [x] The CLI should open a bug report issue if the command selected failed, otherwise it should open a feedback issue.
- [ ] It should incorporate the new landing page being worked on by @achandmsft.
- [ ] Attempt to report "logical"/brand version names like OS X or Ubuntu 18.04

cc: @yugangw-msft

adewaleo:
Concerning getting brand / os flavor names, see:

- https://unix.stackexchange.com/questions/92199/how-can-i-reliably-get-the-operating-systems-name/92218#92218
- https://github.com/easybuilders/easybuild/wiki/OS_flavor_name_version
- https://askubuntu.com/questions/459402/how-to-know-if-the-running-platform-is-ubuntu-or-centos-with-help-of-a-bash-scri

adewaleo:
closing this due to lack of bandwidth and urgency

cc: @achandmsft

PR #9088, opened 2019-04-12
[feedback] now auto-fills webpage on browser open.

adewaleo:
Users no longer need to paste the issue into the web page as it is now automatically filled.
Exceptions are minified to ensure total url length is less than 2030.

TODO: bump up history.rst.

Ideally there would be some tests for the ErrorMinifier.

see #9037

---


src/command_modules/azure-cli-feedback/azure/cli/command_modules/feedback/custom.py
```diff
@@ -167,6 +172,16 @@ def get_command_status(self):
             success_msg = "SUCCESS" if was_successful else "FAILURE"
         return success_msg
 
+    def failed(self):
+        if not self.command_data_dict:
+            return False
+
+        was_successful = self.command_data_dict.get("success", True)
+        if was_successful:
```
yugangw-msft: `return not was_successful`

PR #9152, opened 2019-04-18
Botservice 0.2.0: New minor version!

stevengum:
### Changes:
* [Breaking Change]: The default value for `--version` is now `v4`, not `v3` (except for `az bot prepare-publish`).
* [Breaking Change]: `--lang` no longer defaults to `Csharp`. If the command requires `--lang` and it is not provided, the command will error out.
* [Breaking Change]: The `--appid` and `--password` args for `az bot create` are now required and can be created via `az ad app create`.
* [Breaking Change]: `az bot create -v v4` does not create or use a Storage Account or Application Insights.
* [Breaking Change]: Instead of mapping Application Insights regions for `az bot create -v v3`, the command only accepts regions where Application Insights is creatable.
* [Breaking Change]: `az bot update` is no longer a generic update command, but instead can affect specific properties of a bot.
* [Deprecation]: All --lang flags now accept and advise users to use `Javascript` instead of `Node`. `Node` as a --lang value is deprecated.
* Add `Typescript` support to `az bot prepare-deploy`
* `az bot prepare-deploy` returns `true` if successful and has helpful verbose logging.
* Add more available Application Insights regions to `az bot create -v v3`
---


stevengum:
@tjprescott, @yugangw-msft I have a few changes that we're still planning on getting in for //Build. However, since this PR is so large already I was going to cut another PR or two in the near future to make it easier to process. 

This PR is ready for review!

src/command_modules/azure-cli-botservice/HISTORY.rst
```diff
@@ -3,6 +3,19 @@
 Release History
 ===============
 
+1.0.0
```
tjprescott: 1.0.0 is a taboo version number. This should be 0.2.0 or 2.0.0.  Given the sheer number of breaking changes, it is not appropriate to declare the module GA and I would recommend 0.2.0.

stevengum: Done, this PR is ready!

src/command_modules/azure-cli-botservice/azure/cli/command_modules/botservice/_params.py
```diff
@@ -68,6 +90,27 @@ def load_arguments(self, _):
...
+        c.argument('tags', arg_type=tags_type)
+        c.argument('app_inisghts_key', options_list=['--app-insights-key', '--ai-key'],
+                   arg_group='Bot Analytics/Application Insights',
+                   help='Azure Application Insights Key used to write bot analytics data. Provide a key if you want '
+                        'to receive bot analytics.')
+        c.argument('app_inisghts_api_key', options_list=['--app-insights-api-key', '--ai-api-key'],
+                   arg_group='Bot Analytics/Application Insights',
+                   help='Azure Application Insights API Key used to read bot analytics data. Provide a key if you want '
+                        'to view analytics about your bot in the Analytics blade.')
+        c.argument('app_inisghts_app_id', options_list=['--app-insights-app-id', '--ai-app-id'],
```
swagatmishra2007: spelling

src/command_modules/azure-cli-botservice/azure/cli/command_modules/botservice/commands.py
```diff
@@ -70,8 +64,8 @@ def load_command_table(self, _):
         g.custom_command('create', 'create_connection')
         g.custom_command('list-providers', 'get_service_providers')
 
-    with self.command_group('bot', botServices_commandType) as g:
-        g.generic_update_command('update', setter_name='update', setter_type=updateBotService_commandType)
```
swagatmishra2007: i think the recommended way is to use the generic_update_command. @tjprescott ?

stevengum: Using the generic_update_command didn't seem to work for us prior, and exposed too many properties on the bot, i.e. the `--kind`. It was also unclear when using --properties if users were allowed to change _everything_ in it, such as msaAppId, endpointVersion, configuredChannels, enabledChannel; all of which are unchangeable.

src/command_modules/azure-cli-botservice/azure/cli/command_modules/botservice/_params.py
stevengum: Fixed, thanks!

stevengum:
@tjprescott, I fixed the build break and this PR is ready to go 👍

vivekkranjan:
Reported with ticket #119042525001052  
[Breaking Change]: az bot create -v v4 does not create or use a Storage Account or Application Insights.

Thanks for the fix.

PR #9154, opened 2019-04-19
[compute] Proximity placement group support

adewaleo:
---


src/command_modules/azure-cli-vm/azure/cli/command_modules/vm/_validators.py
```diff
@@ -68,6 +68,26 @@ def validate_keyvault(cmd, namespace):
...
 
+def validate_proximity_placement_group(cmd, namespace):
+    from msrestazure.tools import parse_resource_id, is_valid_resource_id
+
+    namespace.proximity_placement_group = _get_resource_id(cmd.cli_ctx, namespace.proximity_placement_group,
+                                                           namespace.resource_group_name,
+                                                           'proximityPlacementGroups', 'Microsoft.Compute')
+
+    if not namespace.proximity_placement_group or not is_valid_resource_id(namespace.proximity_placement_group):
+        logger.debug('Proximity placement group is %s', namespace.proximity_placement_group)
```
yugangw-msft: Is this a correct trace message when the PPD is not available?

src/command_modules/azure-cli-vm/azure/cli/command_modules/vm/_validators.py
```diff
@@ -68,6 +68,26 @@ def validate_keyvault(cmd, namespace):
...
+        return
+
+    parsed = parse_resource_id(namespace.proximity_placement_group)
+
+    rg = parsed['resource_group']
+    name = parsed['name']
+
+    if not check_existence(cmd.cli_ctx, name, rg, 'Microsoft.Compute', 'proximityPlacementGroups'):
```
yugangw-msft: what will happen at service end, if we don't do the check at client end? Do we get a good error?

src/command_modules/azure-cli-vm/setup.py
```diff
@@ -34,7 +34,7 @@
 DEPENDENCIES = [
     'azure-mgmt-msi==0.2.0',
     'azure-mgmt-authorization==0.50.0',
-    'azure-mgmt-compute==4.6.1',
+    'azure-mgmt-compute==4.6.2',
```
yugangw-msft: you might need to update  other command  modules such as acs

src/command_modules/azure-cli-vm/azure/cli/command_modules/vm/_validators.py
adewaleo: This is just for debugging purpose to show what was processed for the parameter. I can reword the logging message.

src/command_modules/azure-cli-vm/azure/cli/command_modules/vm/_validators.py
adewaleo: I believe the error should be something like the following, if the ppg is in a different region or does not exist:

```
Deployment failed. Correlation ID: 260fc599-b6f4-418f-9c2e-fc800602d78f. {
  "error": {
    "code": "NotFound",
    "message": "Proximity Placement Group '/subscriptions/00977cdb-163f-435f-9c32-39ec8ae61f4d/resourceGroups/ova-test/providers/Microsoft.Compute/proximityPlacementGroups/tosinppg' cannot be found."
  }
}
```

I can remove this existence check, however the CLI does perform this check for availability set...

src/command_modules/azure-cli-vm/azure/cli/command_modules/vm/_validators.py
```diff
@@ -68,6 +68,26 @@ def validate_keyvault(cmd, namespace):
...
+        raise CLIError("Availability set '{}' does not exist.".format(name))
```
adewaleo: This should be proximity placement group not availability set

src/command_modules/azure-cli-vm/setup.py
adewaleo: Yes @yugangw-msft acs and servicefabric have been updated.

src/command_modules/azure-cli-vm/azure/cli/command_modules/vm/_validators.py
```diff
@@ -76,7 +76,7 @@ def validate_proximity_placement_group(cmd, namespace):
                                                            'proximityPlacementGroups', 'Microsoft.Compute')
 
     if not namespace.proximity_placement_group or not is_valid_resource_id(namespace.proximity_placement_group):
-        logger.debug('Proximity placement group is %s', namespace.proximity_placement_group)
+        logger.debug('namespace.proximity_placement_group is %s', namespace.proximity_placement_group)
```
tjprescott: Why is this here?

adewaleo: debug statement just in case the user enters an invalid proximity placement group. The previous lines try to parse the resource if and return None. Maybe I should actually raise a CLI error if this fails to parse...

What do you think?

src/command_modules/azure-cli-vm/azure/cli/command_modules/vm/_params.py
```diff
@@ -50,13 +51,24 @@ def load_arguments(self, _):
 
     extension_instance_name_type = CLIArgumentType(help="Name of the vm's instance of the extension. Default: name of the extension.")
 
-    if StorageAccountTypes:
-        disk_sku = CLIArgumentType(arg_type=get_enum_type(StorageAccountTypes))
+    # StorageAccountTypes renamed to DiskStorageAccountTypes in 2018_06_01 of azure-mgmt-compute
+    if StorageAccountTypes and not DiskStorageAccountTypes:
+        DiskStorageAccountTypes = StorageAccountTypes
```
yugangw-msft: maybe `DiskStorageAccountTypes  = DiskStorageAccountTypes or StorageAccountTypes`

PR #9170, opened 2019-04-22
Adopt compute 4.6.2

marstr:
/cc @lmazuel 

This is test of the azure-mgmt-compute package to make sure we're compatible with the latest version. This does not necessarily need to be merged.


marstr:
Closing after establishing that the CI failure was not related to the updated dependency.

PR #9193, opened 2019-04-24
Adding script that consolidates our existing install instructions

marstr:
/cc @achandmsft @sptramer

This consolidates the instructions that are currently found [here](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli-apt?view=azure-cli-latest) into a single script. The idea is that we can tell Debian/Ubuntu users to just execute `sudo curl -sL https://aka.ms/InstallAzureCLIDeb | bash`.

Because they will need to use `sudo` to do this, I've added a conspicuous warning that they should try to refrain from sudo executing scripts from sources other than well trusted entities.

### DIFF
--- a/scripts/release/debian/deb_install.sh
+++ b/scripts/release/debian/deb_install.sh
@@ -34,6 +34,8 @@ function assert_consent {
     fi
 }
 
+global_consent=0 # Artificially giving global consent after review-feedback. Remove this line to enable interactive mode
+
 assert_consent "Add Microsoft as a trusted package signer?" ${global_consent}
 set -v
 curl -sL https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > /etc/apt/trusted.gpg.d/microsoft.asc.gpg
@@ -41,7 +43,8 @@ set +v
 
 assert_consent "Add the Azure CLI Repository to your apt sources?" ${global_consent}
 set -v
-echo "deb [arch=amd64] https://packages.microsoft.com/repos/azure-cli/ $(lsb_release -cs) main" \
+CLI_REPO=$(lsb_release -cs)
+echo "deb [arch=amd64] https://packages.microsoft.com/repos/azure-cli/ ${CLI_REPO} main" \
     > /etc/apt/sources.list.d/azure-cli.list
 apt-get update
 set +v
--- a/scripts/release/debian/deb_install.sh
+++ b/scripts/release/debian/deb_install.sh
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+
+#######################################################################################################################
+# This script does three fundamental things:                                                                          #
+#   1. Add Microsoft's GPG Key has a trusted source of apt packages.                                                  #
+#   2. Add Microsoft's repositories as a source for apt packages.                                                     #
+#   3. Installs the Azure CLI from those repositories.                                                                #
+# Given the nature of this script, it must be executed with elevated privledges, i.e. with `sudo`.                    #
+#                                                                                                                     #
+# Remember, with great power comes great responsibility.                                                              #
+#                                                                                                                     #
+# Do not be in the habit of executing scripts from the internet with root-level access to your machine. Only trust    #
+# well-known publishers.                                                                                              #
+#######################################################################################################################
+
+set -ev
+
+apt-get install -y curl apt-transport-https lsb-release gnupg
+
+curl -sL https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > /etc/apt/trusted.gpg.d/microsoft.asc.gpg
+
+AZ_REPO=$(lsb_release -cs)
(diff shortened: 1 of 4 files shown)


scripts/release/debian/deb_install.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+
+#######################################################################################################################
```
adewaleo: Just to clarify this warning is an in script comment. 

If users run `sudo curl -sL https://aka.ms/InstallAzureCLIDeb | bash` they won't see the warning, unless they read the script, right?

marstr: Correct. By the time they're executing the script, the damage is already done. My thinking was that I'd rather make clear to the skeptical folks taking the time to read through this script, that the danger is not lost on us.

sptramer:
@marstr Which versions of Ubuntu/Debian is this script verified to work on?

marstr:
Excellect question, @sptramer. I have confirmed this with the following:

| Distro | Versions |
| :-----: | :------: |
| Ubuntu | bionic, xenial, cosmic  |
| Debian |  stretch, jessie |

By end of week, I expect to also support ubuntu:disco

scripts/release/debian/deb_install.sh
sptramer: @marstr The traditional way to get around that is to allow for `-sL` but also by default make the script interactive in some way, so that users can sign off on Y/N operations or otherwise inspect it before running. A lot of system administrators consider curling a bash script without reading it a serious security problem and may disallow use of the Azure CLI in their environments without that kind of support.

marstr: 10-4, I'll add some prompts to explain what we're doing.

For the record, I absolutely agree with the concern about sudo executing a bash script. My preferred solution would be to publish our product directly to Debian/Ubuntu repositories so that folks didn't have to muck with installing ours. However, for the moment I think that's a pipe-dream, so something like this will have to do.

sptramer: Have we considered adding a PPA? That's a decent middle ground for solving that kind of problem.

yugangw-msft:
@marstr, just a question, will this script also work on WSL?

scripts/release/debian/deb_install.sh
marstr: Oooh, that's not a bad idea! I'll toss that around a little more after build.

marstr:
>@yugangw-msft says:
> @marstr, just a question, will this script also work on WSL?

Yes! This should work on WSL. I'll have @kristapratico try this now, assume no news is good news on that front.

scripts/release/debian/deb_install.sh
marstr: consent prompts add in 294f11e

adewaleo: whats a PPA @marstr and @sptramer?

marstr: Good question :) 
https://en.wikipedia.org/wiki/Ubuntu#Package_Archives

marstr:
Closing and opening to attempt to kick DevOps into paying attention.

marstr:
For anybody who wants to try this, the command to use is this:

``` bash
curl -sL https://aka.ms/InstallAzureCLIDeb > az_install_apt.sh && sudo bash ./az_install_apt.sh
```

yugangw-msft:
@achandmsft @marstr, this change looks good to me. My only question is that the script doesn't include the prerequisite configuration e.g. `sudo apt-get install curl apt-transport-https lsb-release gnupg`. I am not very familiar with Linux community, but if those packages are common enough then skipping this step is fine. My worry is mainly on the debian distros which might not even have the `curl`. Ubuntu should be fine since it has preloaded lots of common stuff.

marstr:
I've only tested in container environments where those tools might not be present for minification purposes. You raise a good point though, I'm creating a VM based on debian:9 to check.

marstr:
Good call, @yugangw-msft. Those pre-requistes are now installed at the beginning of the script.

scripts/release/debian/deb_install.sh
```diff
@@ -0,0 +1,59 @@
...
+        exit 1
+    fi
+}
+
+global_consent=0 # Artificially giving global consent after review-feedback. Remove this line to enable interactive mode
+
+assert_consent "Add packages necessary to modify your apt-package sources?" ${global_consent}
+set -v
+apt-get update
+apt-get install -y apt-transport-https lsb-release gnupg curl
```
yugangw-msft: Do need `curl` since the whole flow starts with a `curl`? I am fine to let users pre-install curl.

marstr: I ended up including it because I figured there's a chance that people download the script using their browser to inspect it, then run the script. Because I use curl a few lines down, I figured its safest to assert its presence.

PR #9195, opened 2019-04-24
[webapp] ssh skip ssl check

patricklee2:
customer was using a self-signed cert, so ssl check fails
add option to skip ssl check

---

### DIFF
--- a/src/command_modules/azure-cli-appservice/HISTORY.rst
+++ b/src/command_modules/azure-cli-appservice/HISTORY.rst
@@ -2,6 +2,8 @@
 
 Release History
 ===============
+* webapp: webapp ssh, add option to skip ssl check
+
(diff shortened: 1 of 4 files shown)


panchagnula:
@patricklee2 I believe for CLI to skip self-signed cert check the recommended way is to use the env variable AZURE_CLI_DISABLE_CONNECTION_VERIFICATION=1
if the API's are from the management libraries Microsoft.web, CLI handles this by default, but if not your code should handle this by reading the environment variable  value and pass in a verify argument.
we recently this with zip deployment - please take a look & verify if this fixes the issue here instead having to pass skip check as a parameter.

patricklee2:
closes in favour of https://github.com/Azure/azure-cli/pull/9206

PR #9202, opened 2019-04-25
Support Ubuntu Disco Dingo

marstr:
Deleting previously commented-out Debian, Wheezy build, and adding Ubuntu, Disco build.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -511,7 +511,8 @@ jobs:
   displayName: Build Ubuntu Disco Package
 
   dependsOn: BuildPythonWheel
-  condition: and(succeeded(), in(variables['Build.Reason'], 'IndividualCI', 'BatchedCI', 'Manual'))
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -507,37 +507,35 @@ jobs:
...
-
-
-#   - task: PublishBuildArtifacts@1
-#     displayName: 'Publish Artifact: debian'
-#     inputs:
-#       ArtifactName: debian
-
+  dependsOn: BuildPythonWheel
+  # condition: and(succeeded(), in(variables['Build.Reason'], 'IndividualCI', 'BatchedCI', 'Manual'))
+  condition: succeeded()
```
tjprescott: What's the rationale for 515 over 514?  Does this mean this build will always run?

marstr: Just temporary! This is the only change in 70fb79c, and I'll revert it before merging this PR. I just wanted to actually run this code before it gets merged :)

marstr: [Looks like the build worked](https://dev.azure.com/azure-public/azcli/_build/results?buildId=5412), reverting now.

PR #9207, opened 2019-04-25
webapp add remote-connection

patricklee2:
port remote-connection to cli-core

---

### DIFF
--- a/src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/_help.py
+++ b/src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/_help.py
@@ -1480,3 +1480,8 @@ helps['webapp webjob triggered run'] = """
 type: command
 short-summary: Run a specific triggered webjob hosted on a web app.
 """
+
+helps['webapp create-remote-connection'] = """
+    type: command
+    short-summary: Creates a remote connection using a tcp tunnel to your web app
+"""
--- a/src/command_modules/azure-cli-appservice/HISTORY.rst
+++ b/src/command_modules/azure-cli-appservice/HISTORY.rst
@@ -6,6 +6,7 @@ Release History
 * appserviceplan: az appserviceplan create support for Linux FREE sku
 * webapp: az webapp up now has a 30s sleep after setting SCM_DO_BUILD_DURING_DEPLOYMENT=true appsetting to handle kudu cold start
 * functionapp: `az functionapp create` supports a runtime of `powershell` on Windows
+* webapp: add create-remote-connection command
 
 0.2.18
 ++++++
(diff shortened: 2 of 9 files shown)


patricklee2:
@panchagnula @yugangw-msft  is the ready for review

src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/_help.py
```diff
@@ -1474,3 +1474,13 @@
 type: command
 short-summary: Run a specific triggered webjob hosted on a web app.
 """
+
+helps['webapp remote-connection'] = """
+    type: group
+    short-summary: Create a remote connection using a tcp tunnel to your web app
+"""
+
+helps['webapp remote-connection create'] = """
```
yugangw-msft: will you ever have other commands under  `remote-connection` like `remove`? If not, then you should just create a command called `create-remote-connection` and get rid of the group.

src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/_help.py
```diff
@@ -1474,3 +1474,13 @@
 type: command
 short-summary: Run a specific triggered webjob hosted on a web app.
 """
+
+helps['webapp remote-connection'] = """
+    type: group
+    short-summary: Create a remote connection using a tcp tunnel to your web app
```
yugangw-msft: suggest `manage remote connections to your web app`

src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/_help.py
patricklee2: There is a lot of documentation for `remote-connection create` I think we need a plan or announcement before renaming it.

src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/_help.py
panchagnula: @patricklee2 given we are moving this to core from extension, most of the documentation needs to be updated. If we want to make any changes this would be the time to change it. 

//CC: @rramachand21

panchagnula: Also once this releases on core - can you make sure to update the command on extension with a message to start using the core command & eventually we can remove it completely from the extension.

patricklee2: I'll rename it to create-remote-connection, add a message to the extension and start changing documentation.

PR #9210, opened 2019-04-26
[configure] Add local configure support

limingu:
---

### DIFF
--- a/src/command_modules/azure-cli-configure/azure/cli/command_modules/configure/custom.py
+++ b/src/command_modules/azure-cli-configure/azure/cli/command_modules/configure/custom.py
@@ -125,6 +125,7 @@ def _handle_global_configuration(config):
             config.set_to_use_local_config(True)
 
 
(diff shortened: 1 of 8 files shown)


src/command_modules/azure-cli-configure/azure/cli/command_modules/configure/custom.py
```diff
@@ -105,8 +106,11 @@ def _handle_global_configuration(config):
         answers['modify_global_prompt'] = should_modify_global_config
     if not config_exists or should_modify_global_config:
         # no config exists yet so configure global config or user wants to modify global config
+        need_to_reset_use_local_config = True if config.use_local_config else False
```
yugangw-msft: Can be simplified `need_to_reset_use_local_config  = config.use_local_config`

limingu: Good catch. Updated.

PR #9256, opened 2019-04-30
login: redirect to the latest landing page

yugangw-msft:
Looks like the old link was tweaked by doc team hence is not quite valid any more, with that, can we just redirect to the new landing page?

//CC: @achandmsft @marstr   


yugangw-msft:
The redirecting is already happening w/o my change. I believe I hit a web page bug once that prevented the page from rendering correctly. Based on that, I am closing this PR and will apply a right fix (if needed) for this.

PR #9279, opened 2019-05-01
[storage] Add 'storage blob sync' command

limingu:
---


src/command_modules/azure-cli-storage/azure/cli/command_modules/storage/operations/azcopy.py
```diff
@@ -0,0 +1,68 @@
...
+    azcopy = _azcopy_blob_client(cmd, client)
+    storage_blob_copy(azcopy, source, _add_url_sas(destination, azcopy.creds.sas_token), recursive=recursive)
+
+
+def storage_blob_download(cmd, client, source, destination, recursive=None):
+    azcopy = _azcopy_blob_client(cmd, client)
+    storage_blob_copy(azcopy, _add_url_sas(source, azcopy.creds.sas_token), destination, recursive=recursive)
+
+
+# def storage_blob_upload_batch(cmd, client, source, destination):
```
yugangw-msft: Any reason we comment it out here? Because we will use it later?

src/command_modules/azure-cli-storage/azure/cli/command_modules/storage/azcopy/util.py
```diff
@@ -0,0 +1,138 @@
...
+import subprocess
+import datetime
+from six.moves.urllib.parse import urlparse
+from azure.cli.core._profile import Profile
+from knack.log import get_logger
+
+logger = get_logger(__name__)
+
+
+STORAGE_RESOURCE_ENDPOINT = "https://storage.azure.com"
```
yugangw-msft: should we consider the end points in other clouds env, such as China cloud?

src/command_modules/azure-cli-storage/azure/cli/command_modules/storage/azcopy/__init__.py
```diff
@@ -0,0 +1,4 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
```
yugangw-msft: This will increase the  installer payload a lot, I suggest we download per demand (which the async is invoked). Doable?
In the same way like 'az aks install-cli'

src/command_modules/azure-cli-storage/azure/cli/command_modules/storage/_params.py
```diff
@@ -387,6 +387,16 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals, too-many-statem
                    help='Name of the destination blob. If the exists, it will be overwritten.')
         c.argument('source_lease_id', arg_group='Copy Source')
 
+    with self.argument_context('storage blob sync') as c:
+        c.extra('destination_container', options_list=['--container', '-c'], required=True,
```
tjprescott: `--container` seems ambiguous to me. Does the help text make clear that this is the destination? Why is the destination a URL then?

src/command_modules/azure-cli-storage/azure/cli/command_modules/storage/operations/azcopy.py
limingu: Removed.

src/command_modules/azure-cli-storage/azure/cli/command_modules/storage/azcopy/util.py
limingu: I will confirm with service team to see if this endpoint works all the clouds.

src/command_modules/azure-cli-storage/azure/cli/command_modules/storage/azcopy/__init__.py
limingu: Updated.

src/command_modules/azure-cli-storage/azure/cli/command_modules/storage/_params.py
limingu: The helper text is clear as 'The sync destination container.'
The validate function is to validate that the url of the blob of destination container with path is valid. The url is to access the destination container with path.

src/UserslimguAppDataLocalTemptmpsroyqvhk
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
```
tjprescott: Seems like there is a bad path somewhere because a temp file like this shouldn't be committed. Looking at the path, it looks like it was using hard coded path separators instead of `os.path.join`.

limingu: Removed this file. It is accidentally checked in.

PR #9297, opened 2019-05-02
aad: new command 'az ad user update'

yugangw-msft:

### DIFF
--- a/src/azure-cli/HISTORY.rst
+++ b/src/azure-cli/HISTORY.rst
@@ -29,6 +29,10 @@ Release History
 
 * Fix for issue #5824 - added validation for parameter --rights of authorizationrule commands
 
+**RBAC**
+
+* user: support update command
+
 **RDBMS**
 
 * Add optional parameter to specify replica SKU for create replica command.
--- a/src/azure-cli/HISTORY.rst
+++ b/src/azure-cli/HISTORY.rst
@@ -2,6 +2,12 @@
 
 Release History
 ===============
+
+**RBAC**
+
+* add "user update" command
+* deprecate "--upn-or-object-id" from user related commands and introduce "--id"
+
 2.0.69
 ++++++
 
@@ -28,10 +34,6 @@ Release History
 **Eventhubs**
 
 * Fix for issue #5824 - added validation for parameter --rights of authorizationrule commands
-
-**RBAC**
-
-* user: support update command
 
 **RDBMS**
 
--- a/src/azure-cli/azure/cli/command_modules/role/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/role/_help.py
@@ -467,6 +467,11 @@ type: command
 short-summary: List Azure Active Directory users.
 """
 
+helps['ad user update'] = """
+type: command
+short-summary: Update Azure Active Directory users.
+"""
+
 helps['identity'] = """
 type: group
 short-summary: Managed Service Identities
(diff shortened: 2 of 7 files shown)


yugangw-msft:
@limingu  @adewaleo , can one of you take a look?

src/command_modules/azure-cli-role/azure/cli/command_modules/role/_params.py
```diff
@@ -127,9 +127,11 @@ def load_arguments(self, _):
     with self.argument_context('ad user') as c:
         c.argument('mail_nickname', help='mail alias. Defaults to user principal name')
         c.argument('force_change_password_next_login', arg_type=get_three_state_flag())
+        c.argument('account_enabled', arg_type=get_three_state_flag(), help='enable the user account')
+        c.argument('password', help='user password')
+        c.argument('upn_or_object_id', help='The object ID or principal name of the user for which to get information')
```
tjprescott: Is there a better name for this argument?  Why not simply have `options_list=['--upn', '--object-id']`... or break it out into two arguments with distinct help text? My gut tells me we shouldn't have "or" in our argument names.

limingu: But I can see the sdk function is using that as a parameter in the code below.
client.update(upn_or_object_id=upn_or_object_id, parameters=update_parameters)

src/command_modules/azure-cli-role/azure/cli/command_modules/role/tests/latest/test_graph.py
```diff
@@ -268,11 +276,13 @@ def test_graph_group_scenario(self):
...
             self.cmd('ad group member check -g {group} --member-id {user1_id}',
                      checks=self.check('value', True))
+
             # check user2 memebership
             self.cmd('ad group member check -g {group} --member-id {user2_id}',
-                     checks=self.check('value', True))            # list memebers
+                     checks=self.check('value', True))
+
             self.cmd('ad group member list -g {group}', checks=[
-                self.check("length([?displayName=='{user1}'])", 1),
```
limingu: We can check if the older user1 name is not valid any more. Will something like below will do the check
self.check("length([?displayName=='{user1}'])", 0),

src/command_modules/azure-cli-role/azure/cli/command_modules/role/_params.py
yugangw-msft: This is the named used by SDK, but we should not have exposed it through `az ad user show`. Since the name has been out there for 2 years, breaking it right now might be a stretch. For better usability, I can add `--id` as another option, this will get us the consistency across `app`, `sp`, and `user`. Also `upn` is a logical id for a user, so `--id` is a right term.

tjprescott: We can add the split options and deprecate the one with `or` in it.

tjprescott:
@yugangw-msft what's the deal with this? Is it still legitimate?

yugangw-msft:
@tjprescott, the PR is still legit. I have addressed your review feedback, can you please take a look?

src/azure-cli/HISTORY.rst
```diff
@@ -25,6 +25,10 @@ Release History
 
 * Fix for issue #5824 - added validation for parameter --rights of authorizationrule commands
 
+**RBAC**
+
+* user: support update command
```
tjprescott: Should add the new alias to `--id` from the deprecated option.

yugangw-msft: Done

Issue #9026, opened 2019-04-08
Please update to clarify CREATE options.


ThePhunny:
It might be useful to clarify that in az snapshot create --resource-group -g that it is the TARGET RG the snapshot will be created in not the location of the source.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: fca6f8c5-a518-04ee-08e3-4ad8d4c9516e
* Version Independent ID: 17710c8c-9cd9-57ea-b6b0-1411648ea03f
* Content: [az snapshot](https://docs.microsoft.com/en-us/cli/azure/snapshot?view=azure-cli-latest#feedback)
* Content Source: [latest/docs-ref-autogen/snapshot.yml](https://github.com/MicrosoftDocs/azure-docs-cli/blob/live/latest/docs-ref-autogen/snapshot.yml)
* Service: **virtual-machines**
* GitHub Login: @rloutlaw
* Microsoft Alias: **routlaw**

adewaleo:
@ThePhunny. Thanks for reaching out about this. This is the existing general pattern of the CLI
`--resource-group` is the destination of the resource to be created.
see the following commands and their examples:
- az vm create --help
- az storage account create --help
- az image create --help

This is why --source accepts **a name or ID** for source managed disks and snapshots. I

adewaleo:
@ThePhunny 

Is there any way you think we could make this command clearer? Could an example showing snapshot creation from a source in a different resource group help?

adewaleo:
Closing this as #9306 adds an example to `az snapshot create` that illustrates how to create a snapshot from a disk in a different resource group (via its resource ID).

Issue #8669, opened 2019-03-01
[VM Create]How to add multiple ssh keys while creating VM?


yuxisun1217:
Hi,

We wish to add multiple ssh keys while creating VM. For example, I wish to create a VM with user "azureuser" and add 2 public keys to it. The RestAPI(https://docs.microsoft.com/en-us/rest/api/compute/virtualmachines/createorupdate#create_a_vm_with_ssh_authentication.) can support this feature. Is this feature supported in Azure CLI? Thanks!

yugangw-msft:
Not supported through `vm create`, and also after a vm gets create, the `linuxConfiguration.ssh.publicKeys` becomes read-only. Based on that, before we get more votes for the support, right now, your workaround is:
1. az vm create --debug --validate" and get the template
2. Add a new ssh entry into the template file
3. Run `az group deployment create <template-file>`

yugangw-msft:
@adewaleo, let us wait on this before we get more votes. If we need to expand the `--ssh-key-value` to be a list

adewaleo:
@yugangw-msft, this might also be a good candidate for generic `--set` for create commands.

yuxisun1217:
@yugangw-msft Thanks for the workaround and looking forward to the new feature support :)

adewaleo:
@yuxisun1217, this feature has been added via #9306 with `--ssh-key-values`. `--ssh-key-value` would also work. They are now aliases for one another. 

These change should be in our upcoming release.

yuxisun1217:
Got it. Thank you so much!

Issue #5292, opened 2018-01-12
vm command modules: revisit all artifacts marked as preview


yugangw-msft:
Log this as the check-point so we can revisit routinely to remove those flags once GA, and apply necessary adjustment if the GA version has breaking change:
1. ~VMSS Rolling upgrade~
2. ~Availability Zone~
3. ~Run-Commands~
4. ~Disk Encryption~ 
5. ~OS-disk swap~

yugangw-msft:
@ejarvi, is single pass disk encryption still under feature flag for VM and VMSS?

ejarvi:
- Single VM - No
- VMSS - Yes

adewaleo:
@ejarvi just to confirm vmss disk encryption is no longer in preview right?

ejarvi:
That's right... 

- Announcement:  https://azure.microsoft.com/en-us/blog/extending-azure-security-center-capabilities/
- Windows: https://docs.microsoft.com/en-us/azure/security/azure-security-disk-encryption-windows#encrypt-virtual-machine-scale-sets 
- Linux: https://docs.microsoft.com/en-us/azure/security/azure-security-disk-encryption-linux#encrypt-virtual-machine-scale-sets

adewaleo:
Thanks a lot @ejarvi

adewaleo:
Will close this as this is more of a reminder and not an issue. We will always have features in preview. I have removed. Moreover the features mentioned in the issue have had "(PREVIEW)" removed from their help text.

adewaleo:
Some "preview" text in compute help text has been removed in #9306

PR #9306, opened 2019-05-03
[compute] Compute cli updates

### LINKED ISSUE #8402: az vm create should have a parameter to set osprofile.computerName (hostname)
Yes. I can create a VM from one of my images easily for everything including networking but for some reason there isn't a parameter to set osprofile.computerName (ie the hostname) of the new VM.
I need this almost everyday and I have to resort to a group deployment with templates and many files just to set up a simple VM 

there should be a --computer-name switch to specify the hostname of each created VM.

I have to go through the hassle of a az group deployment with JSON files for this simple thing.

**Additional context**

adewaleo:
Closes #8402 #8669 #9026 #5292 

* vm create: Added --computer-name for setting a vm's computer name.
* vm/vmss create: `--ssh-key-value` renamed to `--ssh-key-values` and can now accept multiple ssh public key values or paths.
* Minor doc updates, such as removing "preview" from some help text.
---


src/command_modules/azure-cli-vm/azure/cli/command_modules/vm/_params.py
```diff
@@ -160,6 +160,7 @@ def load_arguments(self, _):
         c.argument('vm_name', existing_vm_name)
         c.argument('size', completer=get_vm_size_completion_list)
         c.argument('name', arg_type=name_arg_type)
+        c.argument('computer_name')
```
tjprescott: Why is this here? It is doing nothing.

src/command_modules/azure-cli-vm/HISTORY.rst
```diff
@@ -6,6 +6,7 @@ Release History
 2.2.21
 ++++++
 * vm create: Added --computer-name for setting a vm's computer name.
+* vm/vmss create: `--ssh-key-value` can now accept multiple ssh public key values or paths.
```
tjprescott: Then shouldn't it be `--ssh-key-values`?  This would not be a breaking change due to the way argparse parses things.

adewaleo: Yeah, I didn't want cause confusion, even though it would work. but that concern was probably unnecessary.

adewaleo:
@tjprescott, let me know if you have any other review comments

adewaleo:
Does this look good to you @tjprescott ? I considered renaming namespace.ssh_key_value to namespace.ssh_key_values, what do
You think?

PR #9308, opened 2019-05-03
[Botservice] Allow different casings of language input into CLI

mdrichardson:
I got sick of accidentally typing CSharp instead of Csharp and having to look up whether to use Typescript or TypeScript, so I added a function to accept all casings and return an error if there's a non-case typo.

Recommend @stevengum for review. It was an easy fix for a minor annoyance; I won't be offended if you don't merge 😄 

---


stevengum:
@mdrichardson thanks for the suggestion! I folded it into #9412 to make it cleaner, and added a couple smaller items as well.

@tjprescott we already use the enum type for the `_params.py`, but the errors were being emitted from the custom logic. I changed the values to lowercase of the constants and added it to a separate file for usage by the different classes. For the entry point through the methods in `custom.py` I normalized the values for `language` by calling `str.lower()`.

tjprescott:
@stevengum thanks. Since the fix has been rolled into another merged PR, I will close this.

PR #9317, opened 2019-05-06
[ACR] Support Managed Identities for Task

jaysterp:
---


src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
```diff
@@ -506,9 +559,21 @@
 type: command
 short-summary: Add a custom registry login credential to the task
 examples:
-  - name: Add a registry login credential to a task.
+  - name: Add a registry login credential to a task using an opaque username and password.
```
shahzzzam: do we wanna use the term "opaque"?

jaysterp: Changing this to `plain text` as opaque is for internal use.

src/command_modules/azure-cli-acr/HISTORY.rst
```diff
@@ -6,6 +6,14 @@ Release History
 +++++
 * Minor fixes
 
+2.2.7
++++++
+* Support Managed Identities for Task
+
+2.2.6
++++++
+* Minor fixes
```
northtyphoon: duplicated with the top section

jaysterp: Thanks. Will fix this rebasing issue

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
```diff
@@ -495,6 +495,59 @@
...
+"""
+
+helps['acr task identity assign'] = """
+type: group
+short-summary: Update the managed service identity for a task.
+examples:
+  - name: Enable the system-assigned identity to an existing task. This will replace all existing user-assigned identities.
+    text: >
+        az acr task identity assign -n MyTask -r MyRegistry
+  - name: Assign user=assigned managed identities to an existing task. This will replace an existing system-assigned identity.
```
northtyphoon: >will replace [](start = 76, length = 12)

do you mean it will remove existing system-assigned identity?

jaysterp: Yes, `remove` is more clear. Will fix

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
```diff
@@ -182,17 +182,27 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
...
         with self.argument_context(scope) as c:
-            c.argument('username', options_list=['--username', '-u'], help='The username to login to the custom registry.', required=True)
-            c.argument('password', options_list=['--password', '-p'], help='The password to login to the custom registry.', required=True)
+            c.argument('username', options_list=['--username', '-u'], help="The username to login to the custom registry.")
+            c.argument('password', options_list=['--password', '-p'], help="The password to login to the custom registry.")
+            c.argument('password', options_list=['--password', '-p'], help="The password to login to the custom registry.")
+            c.argument('kv_username', help="The key vault secret URI of the username to login to the custom registry.")
+            c.argument('kv_password', help="The key vault secret URI of the password to login to the custom registry.")
+            c.argument('use_identity', nargs='*', help="The task managed identity used for the credential. Use '[system]' to refer to the system assigned identity or a resource id to refer to a user assigned identity.")
```
northtyphoon: >resource id [](start = 168, length = 11)

`resource id` or `client id`?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
```diff
@@ -412,44 +439,71 @@ def get_yaml_and_values(cmd_value, timeout, file):
...
+                                    password=None,
+                                    kv_username=None,
+                                    kv_password=None,
+                                    identity=None):
     """Get the credential object from the input
     :param str auth_mode: The login mode for the source registry
     :param str login_server: The login server of custom registry
     :param str username: The username for custom registry
     :param str password: The password for custom registry
+    :param str kv_username: The key vault secret URI of the username to login to the custom registry.
```
northtyphoon: >. [](start = 100, length = 1)

remove `.` at the end

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
jaysterp: Fixed this typo. Thank you for catching

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
```diff
@@ -495,6 +495,59 @@
   - name: Create a Windows task from a public GitHub repository which builds the Azure Container Builder image on Amd64 architecture.
     text: >
         az acr task create -t acb:{{.Run.ID}} -n acb-win -r MyRegistry -c https://github.com/Azure/acr-builder.git -f Windows.Dockerfile --commit-trigger-enabled false --pull-request-trigger-enabled false --platform Windows/amd64
+  - name: Create a Linux task from a public GitHub repository which builds the hello-world image with a git commit trigger and with the system-assigned managed identity
+    text: >
+        az acr task create -t hello-world:{{.Run.ID}} -n hello-world -r MyRegistry -c https://github.com/Azure-Samples/acr-build-helloworld-node.git -f Dockerfile --assign-identity
```
djyou: nit: add an ending period, same for others.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
```diff
@@ -412,44 +439,71 @@ def get_yaml_and_values(cmd_value, timeout, file):
...
+        source_registry_credentials = SourceRegistryCredentials(
+            login_mode=auth_mode)
 
+    is_remove = False
     custom_registries = None
     if login_server:
+        # if null username and password (or identity), then remove the credential
+        custom_reg_credential = None
+        # Validate if username or password are used, exactly one of each type exist is given
+        if username is not None and kv_username is not None:
```
northtyphoon: If users pass username and kv_password, is it valid?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
```diff
@@ -495,6 +495,59 @@
...
+"""
+
+helps['acr task identity'] = """
+type: group
+short-summary: Managed Service Identities for Task.
+"""
+
+helps['acr task identity assign'] = """
+type: group
+short-summary: Update the managed service identity for a task.
```
djyou: Does `assign` do create or update or both?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
```diff
@@ -495,6 +495,59 @@
...
+examples:
+  - name: Enable the system-assigned identity on an existing task. This will replace all existing user-assigned identities for that task.
+    text: >
+        az acr task identity assign -n MyTask -r MyRegistry
+  - name: Assign user-assigned managed identities to an existing task. This will remove the existing system-assigned identity.
+    text: >
+        az acr task identity assign -n MyTask -r MyRegistry --identities "/subscriptions/<SUBSCRIPTON ID>/resourcegroups/<RESOURCE GROUP>/providers/Microsoft.ManagedIdentity/userAssignedIdentities/myUserAssignedIdentitiy"
+  - name: Assign both system-assigned and user-assigned managed identities to an existing task.
+    text: >
+        az acr task identity assign -n MyTask -r MyRegistry --identities [system] "/subscriptions/<SUBSCRIPTON ID>/resourcegroups/<RESOURCE GROUP>/providers/Microsoft.ManagedIdentity/userAssignedIdentities/myUserAssignedIdentitiy"
```
djyou: Does the user have to include `[]` as part of the value in `[system]`? It seems to be more straightforward to just specify `system`. Is this consistent with other modules?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
```diff
@@ -506,9 +559,21 @@
 type: command
 short-summary: Add a custom registry login credential to the task
 examples:
-  - name: Add a registry login credential to a task.
+  - name: Add a registry login credential to a task using a plain text username and password.
     text: >
         az acr task credential add -n taskname -r registryname --login-server myregistry.docker.io -u 'myusername' -p 'mysecret'
```
djyou: Curious, can we actually reference a docker hub registry by `myregistry.docker.io`?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
```diff
@@ -182,17 +182,27 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         # Run agent parameters
         c.argument('cpu', type=int, help='The CPU configuration in terms of number of cores required for the run.')
 
+        # MSI parameter
+        c.argument('assign_identity', nargs='*', help="Assigns managed identities to the task. Use '[system]' to refer to the system assigned identity or a resource id to refer to a user-assigned identity.")
```
djyou: resource ID

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
```diff
@@ -182,17 +182,27 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
...
         # Custom registry credentials
         c.argument('login_server', help="The login server of the custom registry. For instance, 'myregistry.azurecr.io'.", required=True)
 
     for scope in ['acr task credential add', 'acr task credential update']:
         with self.argument_context(scope) as c:
-            c.argument('username', options_list=['--username', '-u'], help='The username to login to the custom registry.', required=True)
-            c.argument('password', options_list=['--password', '-p'], help='The password to login to the custom registry.', required=True)
+            c.argument('username', options_list=['--username', '-u'], help="The username to login to the custom registry.")
+            c.argument('password', options_list=['--password', '-p'], help="The password to login to the custom registry.")
+            c.argument('password', options_list=['--password', '-p'], help="The password to login to the custom registry.")
```
djyou: duplicated

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
```diff
@@ -34,20 +34,24 @@ def _arm_get_resource_by_name(cli_ctx, resource_name, resource_type):
     :param str resource_type: The type of resource
     """
     result = get_resources_in_subscription(cli_ctx, resource_type)
-    elements = [item for item in result if item.name.lower() == resource_name.lower()]
+    elements = [item for item in result if item.name.lower() ==
+                resource_name.lower()]
```
djyou: nit: this seems to be unnecessary extra work since the original line is also within the limit, but no need to revert either...

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -663,6 +786,28 @@ def _get_list_runs_message(base_message, task_name=None, image=None):
...
+    identities = identities or []
+    identity_types = []
+    if IDENTITY_GLOBAL_REMOVE in identities:
+        return IdentityProperties(type=ResourceIdentityType.none.value)
+    if not identities or IDENTITY_LOCAL_ID in identities:
+        identity_types.append(ResourceIdentityType.system_assigned.value)
+    external_identities = [x for x in identities if x != IDENTITY_LOCAL_ID]
+    if external_identities:
+        identity_types.append(ResourceIdentityType.user_assigned.value)
+    identity_types = ', '.join(identity_types)
```
northtyphoon: Do you change the type of IdentityProperties from array to string here?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
```diff
@@ -412,44 +439,71 @@ def get_yaml_and_values(cmd_value, timeout, file):
...
+        source_registry_credentials = SourceRegistryCredentials(
+            login_mode=auth_mode)
 
+    is_remove = False
     custom_registries = None
     if login_server:
+        # if null username and password (or identity), then remove the credential
+        custom_reg_credential = None
+        # Validate if username or password are used, exactly one of each type exist is given
+        if username is not None and kv_username is not None:
```
djyou: `if username and kv_username:`? I remember the same question, how would you handle empty vs. null `username`/`kv_username`?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
```diff
@@ -412,44 +439,71 @@ def get_yaml_and_values(cmd_value, timeout, file):
...
+            raise CLIError("Provide either username or kv_username.")
+        if (password is not None and kv_password is not None):
+            raise CLIError("Provide either password or kv_password.")
+
+        isIdentityCredential = False
```
djyou: is_identity_credential

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
```diff
@@ -412,44 +439,71 @@ def get_yaml_and_values(cmd_value, timeout, file):
...
+            login_mode=auth_mode)
 
+    is_remove = False
     custom_registries = None
     if login_server:
+        # if null username and password (or identity), then remove the credential
+        custom_reg_credential = None
+        # Validate if username or password are used, exactly one of each type exist is given
+        if username is not None and kv_username is not None:
+            raise CLIError("Provide either username or kv_username.")
```
djyou: ... but not both?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -398,13 +415,110 @@ def acr_task_update(cmd,  # pylint: disable=too-many-locals
...
+    _, resource_group_name = validate_managed_registry(
+        cmd, registry_name, resource_group_name, TASK_NOT_SUPPORTED)
+
+    identity = _build_identities_info(cmd, identities)
+
+    TaskUpdateParameters, PlatformUpdateParameters, AgentProperties, TriggerUpdateParameters = cmd.get_models(
+        'TaskUpdateParameters', 'PlatformUpdateParameters', 'AgentProperties', 'TriggerUpdateParameters')
+    taskUpdateParameters = TaskUpdateParameters(
+        identity=identity,
+        status=None,
```
djyou: `None` is the default value, no need to write them out.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -398,13 +415,110 @@ def acr_task_update(cmd,  # pylint: disable=too-many-locals
...
+        identities = IDENTITY_GLOBAL_REMOVE if not existingUserIdentites else list(existingUserIdentites.keys())
+        identity = _build_identities_info(cmd, identities)
+    else:
+        identity = _build_identities_info(cmd, identities, True)
+
+    TaskUpdateParameters, PlatformUpdateParameters, AgentProperties, TriggerUpdateParameters = cmd.get_models(
+        'TaskUpdateParameters', 'PlatformUpdateParameters', 'AgentProperties', 'TriggerUpdateParameters')
+    taskUpdateParameters = TaskUpdateParameters(
+        identity=identity,
+        status=None,
```
djyou: `None` is the default value.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -23,6 +23,8 @@
 TASK_NOT_SUPPORTED = 'Task is only supported for managed registries.'
 DEFAULT_TOKEN_TYPE = 'PAT'
 NULL_CONTEXT = '/dev/null'
+IDENTITY_LOCAL_ID = '[system]'
+IDENTITY_GLOBAL_REMOVE = '[all]'
```
djyou: Same question, why not just `all`? Is this also case-sensitive?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -663,6 +786,28 @@ def _get_list_runs_message(base_message, task_name=None, image=None):
     return "{}.".format(base_message)
 
 
+def _build_identities_info(cmd, identities, is_remove=None):
```
djyou: nit: the way `is_remove` is used, it could just be a bool with default value

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
northtyphoon: >None [](start = 15, length = 4)

Can you add a comment to explain why it needs to explicitly set the None value? When I first saw the code, my instinct reaction was why we need to set the unrelated value at all for identity command? Some comments will help clarify it.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
jaysterp: Yes, we combinations of either type. We just want to make sure only one username and one password is provided (ie throw an error if user gives `username` AND `kv-username`).

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -398,13 +415,110 @@ def acr_task_update(cmd,  # pylint: disable=too-many-locals
...
+                           client,
+                           task_name,
+                           registry_name,
+                           resource_group_name=None):
+    _, resource_group_name = validate_managed_registry(
+        cmd, registry_name, resource_group_name, TASK_NOT_SUPPORTED)
+
+    resp = client.get_details(resource_group_name, registry_name, task_name).identity
+
+    return {} if not resp else resp
```
northtyphoon: Curious why need to explicitly return `{}`. Can we just return resp like other commands?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
shahzzzam: can we consistently write [system] for system identities? or is this a pattern followed by other RPs as well? Cuz I see an example below has to mention [system] bc it involves multiple identities.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
shahzzzam: both

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
shahzzzam: we have to use `[system]` as other modules do the same.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
jaysterp: `--assign-identity` is default to mean the system-assigned identity for the task resource. The reason the `[system]` is used is for the case where the user wants to use the `SystemAssigned, UserAssigned` type for assigning multiple identities at once. This is an established precedent of other services, such as Azure VM: https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/qs-configure-cli-windows-vm#system-assigned-managed-identity

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
shahzzzam: Its what we do in CLI. If the RP returns Null. The experience will be that when they do az acr show something, nothing gets printed. 
v/s if you do {} they know they dont have anything to show. We do this in couple other places.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
shahzzzam: yep its an Enum in RP.

jaysterp: Yes, this is just for the `SystemAssigned, UserAssigned` case. This just checks if the user is assigning both `SystemAssigned` and `UserAssigned` identities, then the type is `SystemAssigned, UserAssigned`. The identity `Type` is a string. When there is one identity type in the list, no join is performed and it is rendered as a string during `type=identity_types`

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
jaysterp: This is the convention of other services that adopted MSI. See Azure VM https://github.com/Azure/azure-cli/blob/dev/src/command_modules/azure-cli-vm/azure/cli/command_modules/vm/custom.py

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
jaysterp: @djyou We allow both null `username` and `kv_username` when managed identity is used as the sole method for authentication. See help
```
- name: Add a registry login credential to a task that uses only the task system-assigned identity to authenticate to the registry.
    text: >
        az acr task credential add -n taskname -r registryname --login-server myregistry.docker.io --use-identity
```

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
jaysterp: @northtyphoon I was keeping consistent with the existing credential command group return. However, I can change this back. Do we want to also change the credential return? //cc @shahzzzam

shahzzzam: I remember we had a conversation about this and we settled on using `{}` or `[]` to keep consistent with other ACR commands. //cc @djyou

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
shahzzzam: it will work with `--username`, `--password`, `--vault-username` or `--vault-password`. 


Not with MSI, bc with MSI we have a special process of getting ARM token and exchanging it with ACR token and then `docker login reg -u 000000 -p $token`

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
```diff
@@ -506,9 +559,21 @@
...
+  - name: Add a registry login credential to a task using key vault secret URIs for the username and password and the task system-assigned identity.
+    text: >
+        az acr task credential add -n taskname -r registryname --login-server myregistry.docker.io -kv-username 'https://mykeyvault.vault.azure.net/secrets/secretusername' -kv-password 'https://mykeyvault.vault.azure.net/secrets/secretpassword' --use-identity
```
northtyphoon: `-kv-password` => `--kv-password`
The build validation caught it.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
```diff
@@ -506,9 +559,21 @@
...
+  - name: Add a registry login credential to a task using key vault secret URIs for the username and password and a task user-assigned identity given by its client id.
+    text: >
+        az acr task credential add -n taskname -r registryname --login-server myregistry.docker.io -kv-username 'https://mykeyvault.vault.azure.net/secrets/secretusername' -kv-password 'https://mykeyvault.vault.azure.net/secrets/secretpassword' --use-identity 00000000-0000-0000-0000-000000000000
+  - name: Add a registry login credential to a task using a plain text username and key vault secret URI for the password and the task user-assigned identity given by its client id.
+    text: >
+        az acr task credential add -n taskname -r registryname --login-server myregistry.docker.io -u 'myusername' -kv-password 'https://mykeyvault.vault.azure.net/secrets/secretpassword' --use-identity 00000000-0000-0000-0000-000000000000
```
northtyphoon: ditto

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
```diff
@@ -506,9 +559,21 @@
...
-  - name: Add a registry login credential to a task.
+  - name: Add a registry login credential to a task using a plain text username and password.
     text: >
         az acr task credential add -n taskname -r registryname --login-server myregistry.docker.io -u 'myusername' -p 'mysecret'
+  - name: Add a registry login credential to a task using key vault secret URIs for the username and password and the task system-assigned identity.
+    text: >
+        az acr task credential add -n taskname -r registryname --login-server myregistry.docker.io -kv-username 'https://mykeyvault.vault.azure.net/secrets/secretusername' -kv-password 'https://mykeyvault.vault.azure.net/secrets/secretpassword' --use-identity
+  - name: Add a registry login credential to a task using key vault secret URIs for the username and password and a task user-assigned identity given by its client id.
+    text: >
+        az acr task credential add -n taskname -r registryname --login-server myregistry.docker.io -kv-username 'https://mykeyvault.vault.azure.net/secrets/secretusername' -kv-password 'https://mykeyvault.vault.azure.net/secrets/secretpassword' --use-identity 00000000-0000-0000-0000-000000000000
```
northtyphoon: ditto

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
northtyphoon: make sense

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
jaysterp: @shahzzzam just to clarify it is `--kv-username` and `kv-password`

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
jaysterp: I had some pylint trouble with that earlier but it works now. See fix

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
djyou: I would prefer returning server response as is. It's consistent to show nothing if the result is empty. If `{}` is preferred, I would question why server side is not returning empty dict but null.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
djyou: I meant if we can just have `is_remove=False` and if `is_remove:`

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
jaysterp: To keep with the precedent of `[system]` I used `[all]`. I will make this case insensitive

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
jaysterp: There are multiple cases where Identity is null server side. For instance, if a call is made from an older API version where Identity is not defined. We sanitize the Task based on API version and this property will always be null for any lower api version than 2018-09-01 when GET is called. //cc @mnltejaswini

djyou: The CLI will only make request to the latest api-version 2018-09-01?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
jaysterp: Added a fix in RP that checks for null Task Properties. Before, RP threw a null exception when the None values were not set. `Identity` is not a part of Task Properties so we needed a null check for when just Identity is updated. The RP change is merged. 
https://msazure.visualstudio.com/AzureContainerRegistry/_git/Build/pullrequest/1656552?_a=overview

jaysterp: See new `TaskUpdateParameters` with just Identity

jaysterp: See fix. An RP changed was required before this was possible.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
jaysterp: ditto

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
jaysterp: thanks for catching this

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
jaysterp: See logic simplification. Both empty and null or blocked because python expects one argument.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -398,13 +415,83 @@ def acr_task_update(cmd,  # pylint: disable=too-many-locals
...
+            raise CLIError(
+                "Cannot specify additional identities when [all] is used in [--identities]")
+
+    identity = None
+    if not identities or IDENTITY_LOCAL_ID in identities:
+        # To remove only the system assigned identity if user-assigned identities also exist
+        # PATCH with the existing user-assigned identities
+        # If no user-assigned identities exist, set the type to None
+        existingUserIdentites = client.get_details(
+            resource_group_name, registry_name, task_name).identity.user_assigned_identities
```
djyou: Will `get_details().identity` be None?

jaysterp: Good catch. Adding a null check here in case the user tries to remove the system identity when Identity is null for the task

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -398,13 +415,83 @@ def acr_task_update(cmd,  # pylint: disable=too-many-locals
     return client.update(resource_group_name, registry_name, task_name, taskUpdateParameters)
 
 
+def acr_task_identity_assign(cmd,
+                             client,
+                             task_name,
+                             registry_name,
+                             identities=None,
```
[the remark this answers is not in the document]
jaysterp: If `--identities` is not provided during the `task identity assign` method then we assign the system identity. The system-assigned identity is the default MSI for a resource so `--identities` is optional.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
```diff
@@ -412,44 +439,68 @@ def get_yaml_and_values(cmd_value, timeout, file):
...
 
     custom_registries = None
     if login_server:
+        # if null username and password (or identity), then remove the credential
+        custom_reg_credential = None
+        # Validate if username or password are used, exactly one of each type exist is given
+        if username and kv_username:
+            raise CLIError("Provide either username or kv-username.")
+        if password and kv_password:
+            raise CLIError("Provide either password or kv-password.")
```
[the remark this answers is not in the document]
jaysterp: @djyou This is a valid case that we want to support. Say the user has no problem keeping their username in plaintext but wants to hide their passwords. Then we want to support combinations of credential types.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -398,13 +415,83 @@ def acr_task_update(cmd,  # pylint: disable=too-many-locals
...
+                             resource_group_name=None):
+    _, resource_group_name = validate_managed_registry(
+        cmd, registry_name, resource_group_name, TASK_NOT_SUPPORTED)
+
+    if identities is not None and IDENTITY_GLOBAL_REMOVE in identities:
+        if len(identities) > 1:
+            raise CLIError(
+                "Cannot specify additional identities when [all] is used in [--identities]")
+
+    identity = None
```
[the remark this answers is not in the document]
jaysterp: We pass identity to `TaskUpdateParameters`.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -663,6 +760,28 @@ def _get_list_runs_message(base_message, task_name=None, image=None):
...
+        return IdentityProperties(type=ResourceIdentityType.none.value)
+    if not identities or IDENTITY_LOCAL_ID in identities:
+        identity_types.append(ResourceIdentityType.system_assigned.value)
+    external_identities = [x for x in identities if x != IDENTITY_LOCAL_ID]
+    if external_identities:
+        identity_types.append(ResourceIdentityType.user_assigned.value)
+    identity_types = ', '.join(identity_types)
+    identity = IdentityProperties(type=identity_types)
+    if external_identities:
+        if is_remove is not None:
```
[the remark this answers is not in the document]
jaysterp: See fix

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -398,13 +415,83 @@ def acr_task_update(cmd,  # pylint: disable=too-many-locals
...
+def acr_task_identity_remove(cmd,
+                             client,
+                             task_name,
+                             registry_name,
+                             identities=None,
+                             resource_group_name=None):
+    _, resource_group_name = validate_managed_registry(
+        cmd, registry_name, resource_group_name, TASK_NOT_SUPPORTED)
+
+    if identities is not None and IDENTITY_GLOBAL_REMOVE in identities:
```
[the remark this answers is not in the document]
jaysterp: Thanks, see fix

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -663,6 +760,28 @@ def _get_list_runs_message(base_message, task_name=None, image=None):
     return "{}.".format(base_message)
 
 
+def _build_identities_info(cmd, identities, is_remove=None):
+    IdentityProperties, UserIdentityProperties, ResourceIdentityType = cmd.get_models(
+        'IdentityProperties', 'UserIdentityProperties', 'ResourceIdentityType')
+    identities = identities or []
```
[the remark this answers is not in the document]
jaysterp: If identity is None or empty then we handle it as the default case: the system-assigned identity.

jaysterp:
@tjprescott Can you review when you have a chance. Thank you.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
```diff
@@ -182,17 +182,26 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         # Run agent parameters
         c.argument('cpu', type=int, help='The CPU configuration in terms of number of cores required for the run.')
 
+        # MSI parameter
+        c.argument('assign_identity', nargs='*', help="Assigns managed identities to the task. Use '[system]' to refer to the system-assigned identity or a resource ID to refer to a user-assigned identity.")
```
tjprescott: Using '*' implies that `--assign-identity` with no value is acceptable. Is this the case?  If not, use `nargs='+'`

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
```diff
@@ -182,17 +182,26 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
...
+
     with self.argument_context('acr task create') as c:
         c.argument('task_name', completer=None)
 
+    with self.argument_context('acr task identity') as c:
+        c.argument('identities', options_list=['--identities'], nargs='*', help="Assigns managed identities to the task. Use '[system]' to refer to the system-assigned identity or a resource ID to refer to a user-assigned identity.")
```
tjprescott: Same question on `nargs='*'`

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
```diff
@@ -182,17 +182,26 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
...
         c.argument('login_server', help="The login server of the custom registry. For instance, 'myregistry.azurecr.io'.", required=True)
 
     for scope in ['acr task credential add', 'acr task credential update']:
         with self.argument_context(scope) as c:
-            c.argument('username', options_list=['--username', '-u'], help='The username to login to the custom registry.', required=True)
-            c.argument('password', options_list=['--password', '-p'], help='The password to login to the custom registry.', required=True)
+            c.argument('username', options_list=['--username', '-u'], help="The username to login to the custom registry.")
+            c.argument('password', options_list=['--password', '-p'], help="The password to login to the custom registry.")
+            c.argument('kv_username', help="The key vault secret URI of the username to login to the custom registry.")
+            c.argument('kv_password', help="The key vault secret URI of the password to login to the custom registry.")
```
tjprescott: These should simply be folded. `--username` and `--password` should accept a keyvault URI or a value. If there is a compelling reason to keep these separate, there is (to my knowledge) no precedence for the names `--kv-username` and `--kv-password`.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
```diff
@@ -340,12 +362,15 @@ def get_validate_platform(cmd, os_type, platform):
...
+        platform_arch = platform_split[1] if len(
+            platform_split) > 1 else Architecture.amd64.value
+        platform_variant = platform_split[2] if len(
+            platform_split) > 2 else None
 
     if os_type and platform:
         if os_type.lower() != platform_os.lower():
-            raise CLIError("The OS in '--platform' should exactly match the value provided in '--os'.")
+            raise CLIError(
+                "The OS in '--platform' should exactly match the value provided in '--os'.")
```
tjprescott: Why are these even allowed to be set independently then?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
jaysterp: Yes, no value indicates to assign the system-assigned managed identity, which is the default managed identity type for a resource. The user can also enter the value `[system]` but we are treating this type of identity as the default. Azure VM supports a similar behavior https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/qs-configure-cli-windows-vm#system-assigned-managed-identity

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
jaysterp: @northtyphoon @shahzzzam @sajayantony I agree with Travis here. We can simply check if either username or password contains `vault.azure.net` instead of introducing two new parameters. What do you think?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
jaysterp: //cc @shahzzzam

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
jaysterp: @tjprescott see update.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
shahzzzam: we are deprecating `--os` in favor of `--platform`. Platform is also inclusive of the architecture. For instance, --os can only be [windows,linux] while `platform` can be [linux/arm, windows/amd64, amd64/v7]. 
But before deprecating, we want to make sure users can also do `--os windows --platform amd64`.

shahzzzam: Although  @jaysterp and I discussed this offline, and we believe that its better to error out when the user uses both `--os` and `--platform` together.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
jaysterp: @tjprescott the team is concerned to combine `--kv-username` and `--kv-password` parameters with the username/password parameters. For instance, if the parameter has a keyvault domain in its raw value we will treat that input a keyvault URI credential and require a managed identity. Passwords can be anything. Should we constrain what the raw input can be?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
jaysterp: see fix

shahzzzam:
LGTM

src/command_modules/azure-cli-acr/HISTORY.rst
```diff
@@ -6,6 +6,7 @@ Release History
 2.2.7
 +++++
 * Support Managed Identities for Task
+* Upgrade azure-mgmt-storage from 3.1.1 to 3.3.0
```
tjprescott: This is not a customer facing change. Please remove.

jaysterp:
@tjprescott Do these changes look good? //cc @northtyphoon @sajayantony

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/build.py
```diff
@@ -47,6 +47,9 @@ def acr_build(cmd,  # pylint: disable=too-many-locals
     from ._client_factory import cf_acr_registries
     client_registries = cf_acr_registries(cmd.cli_ctx)
 
+    if os_type and platform:
+        raise CLIError("[--os] has been depricated. Please use [--platform] instead.")
```
tjprescott: There is a dedicated mechanism in the CLI for this situation:
https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#deprecating-commands-and-arguments

The previous logic was "both could be used but had to match".  The current logic is "--os can't be used". This is a breaking change. A more logical way would be for `--os` and `--platform` to reference the same variable, but have `--os` deprecated in favor of `--platform`.  It would look something like:

```Python
  c.argument('platform', options_list=['--platform', c.deprecate(target='--os', redirect='--platform', hide=True)])
```

In this way, --platform no longer appears, but does work, and fills the same value as using `--os` would.

jaysterp: So as opposed to introducing a separate argument `--platform` alongside `--os`, we should remove the `--os` parameter and add it to the options of `--platform` with the deprecate wrapper? What do you mean by "--platform no longer appears"? Do you mean --os no longer appears, but does work, and fills the same value as using --platform would?

tjprescott: If you use `hide=True` then `--os` won't appear when using `-h` but it will still work. It is effectively an alias to `--platform` and the user will get a warning if they use it that it is deprecated and they should use `--platform` instead.

tjprescott: If your situation is different feel free to not use it, but it seems like it fits your scenario.

jaysterp: I understand. This is definitely a better option for maintaining the code as well. Will make this change of integrating `--os` into `--platform`. //cc @shahzzzam

shahzzzam:
os/platform change LGTM

PR #9321, opened 2019-05-07
 role assignment: support scope of management group

yugangw-msft:


yugangw-msft:
@limingu, when you get time, can you please take a look?

yugangw-msft:
I will handle the test coverage on assignee-object-id through PR #9337. I will let you review that one as well when it is ready

PR #9322, opened 2019-05-07
update command to list all keys

anusapan:
---


src/command_modules/azure-cli-iot/azure/cli/command_modules/iot/_params.py
```diff
@@ -172,6 +172,7 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
                    help='The number of partitions of the backing Event Hub for device-to-cloud messages.', type=int)
 
     with self.argument_context('iot hub show-connection-string') as c:
+        c.argument('all', options_list=['--all'], help='Allow to show all shared access policies.')
```
tjprescott: You don't need options_list here at all. `--all` would be the default name of this anyways.

anusapan: replace param as this was failing the CI build

PR #9328, opened 2019-05-07
Skip loading extensions which are not compatible with azure-cli-core version.

### LINKED ISSUE #9241: Extensions should not attempt to load if they are not compatible with the current CLI core version
https://github.com/Azure/azure-cli/issues/9229
https://github.com/Azure/azure-cli/issues/9199
https://github.com/Azure/azure-cli/issues/9146

In all of these cases, a version of `az interactive` was released which was compatible with CLI 2.0.63. Folks had an older version of interactive and the CLI and then updated their core version. Updating the extension fixes the problem, but not before a stack trace is thrown. The mechanism for loading extensions should issue a warning and refuse to load extensions if they are not compatible with the current core CLI version.

tjprescott:
Fixes #9241.

While the extension add command prevents installing extensions which are not compatible with the core, if an extension is installed and then the CLI is updated such that the extension is no longer compatible, the CLI would still load that extension, resulting in a potential crash.

This PR ensures that the CLI checks the compatibility of each extension during loading and issues a warning and skips that extension if it is no longer compatible.

Example for `az interactive`:
```
(env) D:\github>az interactive
The 'interactive' extension is not compatible with this version of the CLI.
You have CLI core version 2.0.64 and this extension requires a min of 2.0.50.dev0 and max of 2.0.61.
Run `az interactive --update` and try again.
```

Added custom logic to az interactive to add the final line about running with --update. Otherwise, it would just be the warning.

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/extension/operations.py
+++ b/src/azure-cli-core/azure/cli/core/extension/operations.py
@@ -66,23 +66,6 @@ def _whl_download_from_url(url_parse_result, ext_file):
                 f.write(chunk)
 
 
-def _validate_whl_cli_compat(azext_metadata):
-    is_compatible, cli_core_version, min_required, max_required = ext_compat_with_cli(azext_metadata)
-    logger.debug("Extension compatibility result: is_compatible=%s cli_core_version=%s min_required=%s "
-                 "max_required=%s", is_compatible, cli_core_version, min_required, max_required)
-    if not is_compatible:
-        min_max_msg_fmt = "The extension is not compatible with this version of the CLI.\n" \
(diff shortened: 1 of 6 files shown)


src/azure-cli-core/azure/cli/core/extension/operations.py
```diff
@@ -66,31 +66,14 @@ def _whl_download_from_url(url_parse_result, ext_file):
                 f.write(chunk)
 
 
-def _validate_whl_cli_compat(azext_metadata):
```
tjprescott: Since this is used in commands, I renamed it and made it public.

PR #9359, opened 2019-05-09
Migrate Automation Test Runs to ADO

marstr:
TravisCI has the limitation of needing jobs to have a heartbeat and not run longer than 50 minutes. Our Automation steps have been flirting with that line for a while now, and now that we're starting to see it be a problem it's a good kick in the pants to move towards our longer term goal of moving all CI responsibilities to Azure DevOps.


azure-pipelines.yml
```diff
@@ -285,6 +385,8 @@ jobs:
     inputs:
       artifactName: pypi
 
+
+
```
yugangw-msft: can you remove these?

marstr:
> LGTM, Thanks @marstr! 2 questions since we are at it:

These are great questions :)
 
> 1. Is `/AzurePipelines run` still needed to get CI trigger'd?

tl;dr **No**

 `/AzurePipelines run` was never needed. We had just been seeing some flakiness from the triggers for the pipeline in the `azure-public` organization, and I was trying to work around it. We need to transition to the `azure-sdk` organization anyway because of new policy and access to signing jobs. I've been running CI in both for a little over a week to test our flow, and `azure-sdk` is seeming more reliable so far.

> 2. Still new to the Devops world, but will this add significant loads to the group wide shared pool? If the pool is still under stress right now, let us discuss a bit before merge this PR

I do not believe it'll add significant load. :)

azure-pipelines.yml
marstr: Fixed in 05a72c6

PR #9385, opened 2019-05-14
test: skip 2 ams tests under Python 2.7 to unblock CI

yugangw-msft:
@BrianBlum, I will log an issue to track the investigation, but right now I  need to do it to unblock other PRs which are impacted by these 2 test failures.
I am not able to reproduce it in my dev box with 2.7, and git history has not shown any related change. Same tests are passing at #9317, so more likely a flaky test? 


yugangw-msft:
We are deprecating the support on Python 2.7. Hence we are `won't fix` this.

PR #9390, opened 2019-05-14
Add Cmds For Sql Instance Failover Group

ziwa-msft:
---

- [ x] The PR has modified HISTORY.rst describing any customer-facing, functional changes. Note that this does not include changes only to help content. (see [Modifying change log](https://github.com/Azure/azure-cli/tree/master/doc/authoring_command_modules#modify-change-log)).


ziwa-msft:
Please review and merge this PR. Thanks

src/command_modules/azure-cli-sql/azure/cli/command_modules/sql/_params.py
```diff
@@ -1229,3 +1230,42 @@ def _configure_security_policy_storage_params(arg_ctx):
...
+                   options_list=['--name', '-n'],
+                   help="The name of the Instance Failover Group")
+
+        c.argument('failover_group_name',
+                   options_list=['--name', '-n'],
+                   help="The name of the Instance Failover Group")
+
+        c.argument('managed_instance',
+                   arg_type=managed_instance_param_type,
+                   options_list=['-s'])
```
tjprescott: Why does this only have a short option?

src/command_modules/azure-cli-sql/azure/cli/command_modules/sql/_params.py
```diff
@@ -1229,3 +1230,42 @@ def _configure_security_policy_storage_params(arg_ctx):
...
+
+        c.argument('partner_managed_instance',
+                   help="The name of the partner managed instance of a Instance Failover Group")
```
tjprescott: Is this an existing resource?

src/command_modules/azure-cli-sql/azure/cli/command_modules/sql/_params.py
```diff
@@ -1229,3 +1230,42 @@ def _configure_security_policy_storage_params(arg_ctx):
...
+
+        c.argument('partner_resource_group',
+                   help="The name of the resource group of the partner managed instance")
```
tjprescott: If partner managed instance is an existing resource, then this violates command guidelines:
https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#supporting-name-or-id-parameters

src/command_modules/azure-cli-sql/azure/cli/command_modules/sql/_params.py
```diff
@@ -1229,3 +1230,42 @@ def _configure_security_policy_storage_params(arg_ctx):
...
+
+        c.argument('failover_policy',
+                   help="The failover policy of the Instance Failover Group",
+                   arg_type=get_enum_type(InstanceFailoverPolicyType))
+
+        c.argument('grace_period',
+                   help='Interval in hours before automatic failover is initiated '
+                        'if an outage occurs on the primary server. '
+                        'This indicates that Azure SQL Database will not initiate '
```
tjprescott: Everything after this first line should be in a long summary, defined in help.py.

src/command_modules/azure-cli-sql/azure/cli/command_modules/sql/custom.py
```diff
@@ -2310,3 +2315,119 @@ def virtual_cluster_list(
...
+
+    managed_instance_client = get_sql_managed_instances_operations(cmd.cli_ctx, None)
+    # pylint: disable=no-member
+    primary_server = managed_instance_client.get(
+        managed_instance_name=managed_instance,
+        resource_group_name=resource_group_name)
+
+    partner_server = managed_instance_client.get(
+        managed_instance_name=partner_managed_instance_name,
+        resource_group_name=partner_resource_group)
```
tjprescott: Since these are existing resources, we do not allow something like `--partner-resource-group`. `--partner-managed-instance` should accept a name or ID. In either case, you shouldn't need to do a get for these resources as you can build the ARM ID very easily.

src/command_modules/azure-cli-sql/azure/cli/command_modules/sql/_params.py
ziwa-msft: yes this is an existing resource

src/command_modules/azure-cli-sql/azure/cli/command_modules/sql/_params.py
ziwa-msft: if we constructed this parameter called partner_managed_instance and have customers pass in the name of resource id of this managed_instance, does it imply that the server belongs in the current user subscription? Because if we only pass in the name, I assume the ARM resource id will be constructed using the current resource group/subscription id.

This is not possible because we will need to support cross resource group/subscription pairing for this feature. Any thoughts on that?

src/command_modules/azure-cli-sql/azure/cli/command_modules/sql/custom.py
ziwa-msft: @tjprescott  Can we force the customers to pass in the ARM resource ID only for --partner-managed-instance parameter? We would love to support the cross-subscription scenario here.

ziwa-msft: @tjprescott 
also we would need the partner instance location information, and that is currently being retrieved with the get.

I think our options are:

1) ask customer to provider partner managed instance resource id + location 
2) ask customer to provide partner server name, resource group and subscription.

Shall we with option 1?

src/command_modules/azure-cli-sql/azure/cli/command_modules/sql/_params.py
tjprescott: The crossRG/cross-subscription story in CLI is to use ARM ID.

src/command_modules/azure-cli-sql/azure/cli/command_modules/sql/custom.py
tjprescott: 3. Ask customer to provide partner managed instance name OR ID. If you need the location of the partner instance, then it would make sense to go ahead with the GET call. It looked to me like you were just doing it to get the id field.

ziwa-msft: @tjprescott Yeah we can ask the customer to provide partner managed instance ID. However name would not be okay for cross-subscription scenarios.

Are you suggesting to ask for ID only? I realized that GET wouldn't work for cross-subscription resource either so we HAVE TO ask for the location parameter too.

I short, I suggest we go with option 1. Please let me know if that works for you.

src/command_modules/azure-cli-sql/azure/cli/command_modules/sql/_params.py
```diff
@@ -1229,3 +1230,42 @@ def _configure_security_policy_storage_params(arg_ctx):
...
+    ###############################################
+
+    with self.argument_context('sql instance-failover-group') as c:
+        c.argument('instance_failover_group_name',
+                   options_list=['--name', '-n'],
+                   help="The name of the Instance Failover Group")
+
+        c.argument('failover_group_name',
+                   options_list=['--name', '-n'],
+                   help="The name of the Instance Failover Group")
```
jaredmoo: Why is there both instance_failover_group_name and failover_group_name? Please fix swagger that there is only one of these.

src/command_modules/azure-cli-sql/azure/cli/command_modules/sql/_params.py
jaredmoo: More importantly, why does this not line up with param names in other command groups?

```
Command
    az sql midb show : Get the details for a managed database.

Arguments

Resource Id Arguments
    --ids                   : One or more resource IDs (space-delimited). If provided, no other
                              'Resource Id' arguments should be specified.
    --managed-instance --mi : Name of the Azure SQL managed instance.
```

src/command_modules/azure-cli-sql/azure/cli/command_modules/sql/_params.py
```diff
@@ -1229,3 +1230,42 @@ def _configure_security_policy_storage_params(arg_ctx):
...
+                   options_list=['-s'])
+
+        c.argument('partner_managed_instance',
+                   help="The name of the partner managed instance of a Instance Failover Group")
+
+        c.argument('partner_resource_group',
+                   help="The name of the resource group of the partner managed instance")
+
+        c.argument('failover_policy',
+                   help="The failover policy of the Instance Failover Group",
```
jaredmoo: Please extract common type definition to top of file and use it for both managed & standalone db FG.

src/command_modules/azure-cli-sql/azure/cli/command_modules/sql/_params.py
```diff
@@ -1229,3 +1230,42 @@ def _configure_security_policy_storage_params(arg_ctx):
...
+                   arg_type=get_enum_type(InstanceFailoverPolicyType))
+
+        c.argument('grace_period',
+                   help='Interval in hours before automatic failover is initiated '
```
jaredmoo: Same comment, extract commonality between mi and standalone db FG. Don't copy paste!!

src/command_modules/azure-cli-sql/azure/cli/command_modules/sql/custom.py
```diff
@@ -2310,3 +2315,119 @@ def virtual_cluster_list(
...
+    if failover_policy is not None:
+        instance.read_write_endpoint.failover_policy = failover_policy
+
+    if instance.read_write_endpoint.failover_policy == FailoverPolicyType.manual.value:
+        grace_period = None
+        instance.read_write_endpoint.failover_with_data_loss_grace_period_minutes = grace_period
+
+    if grace_period is not None:
+        grace_period = int(grace_period) * 60
+        instance.read_write_endpoint.failover_with_data_loss_grace_period_minutes = grace_period
```
jaredmoo: This is duplicated with failover_group_update. Please extract common logic, do NOT copy paste.

src/command_modules/azure-cli-sql/azure/cli/command_modules/sql/custom.py
```diff
@@ -2310,3 +2315,119 @@ def virtual_cluster_list(
...
+
+    return instance
+
+
+def instance_failover_group_failover(
```
jaredmoo: Extract common logic

yugangw-msft:
Please reopen when you can address the review feedback

src/command_modules/azure-cli-sql/azure/cli/command_modules/sql/custom.py
ziwa-msft: failover is not a good candidate for abstraction as fg lives under location for Managed Instance.

Issue #8780, opened 2019-03-13
RPM Spec Should be updated to accomodate different distros (incl. RHEL8).


marstr:
[Our RPM Spec is only setup to build using Enterprise Linux 7 repositories](https://github.com/Azure/azure-cli/blob/e5fd28ca284aa2d2b96409555b86bf68ba643258/scripts/release/rpm/azure-cli.spec), and indeed we only publish an `*.el7.x86_64.rpm`.

However, [in our installation instructions](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli-yum?view=azure-cli-latest), we callout several RPM based Linux distros that we support. The list that we say we support is:
- RHEL 7
- Fedora 19+
- CentOS 7
- openSUSE 42.2
- SLES 12 SP2

We should decide which of these distros we want to continue supporting, and make sure that our RPM specifies the particular packages that are available in each of those ecosystems. There are some choices that should be easier than others. For instance, only Fedora Core 28 and 29 are supported now, so I believe those are the only Fedora releases we need to bother attempting to support. https://github.com/Azure/azure-cli/issues/7523 is a great example of what happens when we don't cater to each ecosystem.

Another perk of doing this, is that we may be able to continue ratcheting down the number of people installing the CLI who use Python2.

marstr:
In addition, this will help us manage Enterprise Linux 8, and it's new handling of Python.

Arnavion:
@achandmsft @marstr The issue https://github.com/Azure/azure-cli/issues/7523 that was duped against this one still repros with the latest package `azure-cli-2.0.65-1.el7.x86_64.rpm`. Should one of these be reopened?

marstr:
Good call, yes. I'll reopen #7253 for this.

PR #9394, opened 2019-05-14
Make the RPM Spec flexible to EL7 vs EL8

marstr:
RHEL 8 is changing the way Python is distributed. [See this blog post](https://developers.redhat.com/blog/2018/11/14/python-in-rhel-8/).

This change gives us the flexibility needed to build this RPM in either an EL7 or EL8 environment. To test it, I spun up a RHEL8 VM in Azure and ssh'd in, and ran the centos7 Docker image defined in the repository to make sure there wasn't a regression.

In theory, this change should also let us build and distribute Fedora specific packages that also use Python3 instead of Python2.

Related to #8780

### DIFF
--- a/scripts/release/rpm/azure-cli.spec
+++ b/scripts/release/rpm/azure-cli.spec
@@ -6,7 +6,7 @@
   %define dist .el%{?rhel}
 %endif
 
-%if 0%{?rhel} < 8
+%if 0%{?rhel} && 0%{?rhel} < 8
     %define python_cmd python2
 %else
     %define python_cmd python3
@@ -28,7 +28,7 @@ BuildArch:      x86_64
 Requires:       %{python_cmd}, %{python_cmd}-virtualenv
 
 BuildRequires:  gcc, libffi-devel, openssl-devel
-# BuildRequires:  %{python_cmd}-devel
+BuildRequires:  %{python_cmd}-devel
 
 %global _python_bytecompile_errors_terminate_build 0
 
--- a/scripts/release/rpm/azure-cli.spec
+++ b/scripts/release/rpm/azure-cli.spec
@@ -27,7 +27,7 @@ Url:            https://docs.microsoft.com/en-us/cli/azure/install-azure-cli
 BuildArch:      x86_64
 Requires:       %{python_cmd}, %{python_cmd}-virtualenv
 
-BuildRequires:  gcc, libffi-devel, openssl-devel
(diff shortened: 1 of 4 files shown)


marstr:
While the build completes okay on el8, it seems like there's some sort of package conflict that I still need to get taken care of. I threw the `Do Not Merge` label on there for the interim.

scripts/release/rpm/azure-cli.spec
```diff
@@ -2,8 +2,14 @@
...
 
 # .el7.centos -> .el7
-%if 0%{?rhel} == 7
-  %define dist .el7
+%if 0%{?rhel}
+  %define dist .el%{?rhel}
+%endif
+
+%if 0%{?rhel} < 8
+    %define python_cmd python2
```
yugangw-msft: Just a question, guess older RHEL doesn't have Python3 pre-installed?

marstr: Correct, and worse yet it's not even available in the standard repositories. You can test this out by running the following command:
``` PowerShell
docker run --rm -it centos:7 yum install -y python3
```

There are ways to get at it, but you need to use an SCL appropriate for the EL7 distro your using as seen here: https://linuxize.com/post/how-to-install-python-3-on-centos-7/

marstr: Thinking about it more, I do think I'd be able to syntactically require python3 and get the installation right. However, I just don't know enough about running an EL7 instance to know if people are open to using the SCLs or not.

yugangw-msft: I think this PR is good enough to get started. Still new to EL world, but I guess we can drop support for EL7 around the time that Python 2 gets deprecated?

marstr:
I've got the issue that I was running into figured out. Basically, because I'm including symlinks to files provided by the `python*` package, rpm thinks it's protecting me from stomping on existing files. The immediate solution is to mark the targets of those symlinks as excluded. However, this feels slightly heavy handed to me.

The longer term solution may be to take a note from @zooba and stop using a virtualenv for our isolation, and opt to simply invoke python with `-I`.

scripts/release/rpm/azure-cli.spec
marstr: IMHO, we can safely take the position that no new versions of the CLI will be supported on EL7 starting in January. However, we may need to publish security patches from time to time.

PR #9395, opened 2019-05-14
[storage] Check account_key not None for 'storage blob generate-sa'

### LINKED ISSUE #9346: az storage blob generate-sas fails with self.account_key should not be None
## Describe the bug

**Command Name**
`az storage blob generate-sas`

**Errors:**
```
self.account_key should not be None.
Temp\pip-install-v1m4qsu_\knack\knack\cli.py, ln 206, in invoke
azure\cli\core\commands\__init__.py, ln 560, in execute
azure\cli\core\commands\__init__.py, ln 618, in _run_jobs_serially
azure\cli\core\commands\__init__.py, ln 609, in _run_job
azure\cli\command_modules\storage\__init__.py, ln 240, in new_handler
azure\cli\command_modules\storage\__init__.py, ln 183, in handler
azure\cli\core\commands\__init__.py, ln 588, in _run_job
azure\cli\core\commands\__init__.py, ln 297, in __call__
azure-cli-core\azure\cli\core\__init__.py, ln 453, in default_command_handler
cli\command_modules\storage\operations\blob.py, ln 423, in generate_sas_blob_uri
multiapi\storage\v2018_03_28\blob\baseblobservice.py, ln 494, in generate_blob_shared_access_signature
multiapi\storage\v2018_03_28\common\_error.py, ln 130, in _validate_not_none
ValueError: self.account_key should not be None.
```

## To Reproduce:

- `az storage blob generate-sas -c {} -n {} --permissions {} --expiry {}`

## Environment Summary
```
Windows-10-10.0.17134-SP0
Python 3.6.6
Shell: cmd.exe

azure-cli 2.0.64

```

limingu:
Fix #9346
---

### DIFF
--- a/src/command_modules/azure-cli-storage/azure/cli/command_modules/storage/operations/blob.py
+++ b/src/command_modules/azure-cli-storage/azure/cli/command_modules/storage/operations/blob.py
@@ -417,6 +417,8 @@ def generate_sas_blob_uri(client, container_name, blob_name, permission=None,
                           protocol=None, cache_control=None, content_disposition=None,
                           content_encoding=None, content_language=None,
                           content_type=None, full_uri=False):
+    if client.account_key is None:
+        raise CLIError("Missing --account-key to generate sas")
     sas_token = client.generate_blob_shared_access_signature(
         container_name, blob_name, permission=permission, expiry=expiry, start=start, id=id, ip=ip,
         protocol=protocol, cache_control=cache_control, content_disposition=content_disposition,
--- a/src/command_modules/azure-cli-storage/HISTORY.rst
+++ b/src/command_modules/azure-cli-storage/HISTORY.rst
@@ -3,6 +3,10 @@
 Release History
 ===============
 
+2.4.3
++++++
+* `storage blob generate-sa`: check account_key  not None
+
 2.4.2
 +++++
 * `storage blob sync`: add sync command for storage blob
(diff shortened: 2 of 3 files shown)


src/command_modules/azure-cli-storage/azure/cli/command_modules/storage/operations/blob.py
```diff
@@ -417,6 +417,8 @@ def generate_sas_blob_uri(client, container_name, blob_name, permission=None,
                           protocol=None, cache_control=None, content_disposition=None,
                           content_encoding=None, content_language=None,
                           content_type=None, full_uri=False):
+    if client.account_key is None:
+        raise CLIError("Missing --account-key to generate sas")
```
tjprescott: This should probably be in usage string format, but generally, is there a logic bug?  Typically if you specify a storage account but not the key (or other credential) then the CLI will retrieve the key behind the scenes unless unable.

limingu: Yes, the logic is that CLI will retrieve the key if there is only account name.

But from the call stack,

493:        _validate_not_none('self.account_name', self.account_name)
494:        _validate_not_none('self.account_key', self.account_key)

It failed at 494 line. That mean account_key is None but account_name is not None. So it did happen somehow in such case.

tjprescott: In that case, it seems like the problem is in the logic that retrieves the account key. It would make more sense to do this logic there instead of having a special case check right here. It could also be that this command isn't hitting that logic.

PR #9431, opened 2019-05-17
[Test] Traffic is not recorded by default during RGPreparer setup.

### LINKED ISSUE #9272: test: do not record any http traffics coming from test preparer
1. This blows up almost all test recordings since `ResourceGroupPreparer` is always used 
2. This will cause a nightmare when we need to use azure-mgmt-resource with new api-version, which will come up in the near future with features which CLI need to incorporate like the real `dry-run` support

adewaleo:
Resource Group Preparer is no longer recorded. Moreover `group create` and `group delete` logic will be ignored if the test is not run live.

This can be extended to other preparers as well. It might be a bit more complicated if the preparer returns information that isn't an input. E.g. `account_key` in the case of the `StorageAccountPreparer`. But it is doable. In the case of the StorageAccountPreparer, we can solve this problem by returning the recording replacement of the account key.

closes #9272 

---


src/azure-cli-testsdk/azure/cli/testsdk/preparers.py
```diff
@@ -14,15 +14,17 @@
...
 class ResourceGroupPreparer(AbstractPreparer, SingleValueReplacer):
     def __init__(self, name_prefix='clitest.rg',
                  parameter_name='resource_group',
                  parameter_name_for_location='resource_group_location', location='westus',
                  dev_setting_name='AZURE_CLI_TEST_DEV_RESOURCE_GROUP_NAME',
                  dev_setting_location='AZURE_CLI_TEST_DEV_RESOURCE_GROUP_LOCATION',
-                 random_name_length=75, key='rg'):
-        super(ResourceGroupPreparer, self).__init__(name_prefix, random_name_length)
+                 random_name_length=75, key='rg', disable_recording=True):
+        super(ResourceGroupPreparer, self).__init__(name_prefix, random_name_length, disable_recording)
```
tjprescott: The intent with preparers is that none of the traffic for their creation is recorded. They are "assumed to exist" on playback. Thus, this fix should really be implemented in some kind of base class, because it should equally apply to StorageAccountPreparers and all the others.

adewaleo: Sure, this makes sense. The other preparers might be complicated as they sometimes rely on nondeterministic info (at least StorageAccountPreparers), but it is definitely doable.

I will make the changes.

adewaleo:
I think this should be good to go @tjprescott @yugangw-msft

PR #9456, opened 2019-05-22
Update CLI to use Knack 0.6.2 (preview mechanism)

### LINKED ISSUE #5969: Implement Preview strategies for commands/command groups/parameters
We currently have a mechanism to show that extensions are in preview, but we have no such thing for specific CLI commands, command groups, or parameters.  Even within a GA module, certain parameters may be added in preview, for example `--ddos-protection` for `vnet create`. It would be good to standardize the experience for these: perhaps issue a warning if a preview command or parameter is used so that users are warned that these things may break and thus to avoid using them in scripts. Displaying this warning or not could be a configurable setting.

---

### Environment summary

```
azure-cli (2.0.30)
```

tjprescott:
Closes #5969.

Updates the CLI to use the latest knack.
- <del>Removes the command suggestion code since it now happens in knack</del> Turns out there is CLI-specific logic in the spellchecker so I have to keep the logic in place. It just means that knack now has a limited implementation of this feature (it doesn't handle values)
- Uses the Preview status tag from the latest knack to annotate commands, groups, and arguments in preview. 
- Deprecates and hides BatchAI commands (service is being retired)
- Hides already deprecated ACS commands.

---

### DIFF
--- a/scripts/dependency/requirements.Darwin.external.txt
+++ b/scripts/dependency/requirements.Darwin.external.txt
@@ -75,7 +75,7 @@ idna==2.7
 isodate==0.6.0
 jmespath==0.9.3
 keyring==15.1.0
-knack==0.5.1
+knack==0.6.2
 mock==2.0.0
 msrest==0.5.5
 msrestazure==0.4.34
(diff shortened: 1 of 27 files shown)


tjprescott:
Adding DNM label so we can consider the strategy for existing preview modules more fully before we release this. However, now that the mechanism is done, this should unblock @marstr from his packaging work.

PR #9459, opened 2019-05-23
[ACR] Add `az acr check-health` command

ni-reis:
PR for new ACR command for self-diagnosis `az acr check-health`.

More information can be found here:
https://gist.github.com/ni-reis/a9ef51a6f08a2694482be401b068e558

Examples of the command:

img1
img2
img3
img4

---

### DIFF
--- a/src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_errors.py
+++ b/src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_errors.py
@@ -1,3 +1,11 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+# pylint: disable=line-too-long
+
+
 class ErrorClass:
     error_title = ""
     error_message = ""
--- a/src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_errors.py
+++ b/src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_errors.py
@@ -1,0 +1,117 @@
+class ErrorClass:
+    error_title = ""
+    error_message = ""
+
+    def __init__(self, title, message):
+        self.error_title = title
+        self.error_message = message
+
+    def get_error_message(self, additional_message=None):
+        if additional_message:
+            return "An error occurred: {}\n{}\n{}".format(self.error_title, self.error_message, additional_message)
+        else:
+            return "An error occurred: {}\n{}".format(self.error_title, self.error_message)
+
+    def set_error_message(self, message):
+        return ErrorClass(self.error_title, message)
+
+    def append_error_message(self, message):
+        return self.set_error_message(self.error_message + "\n" + message)
+
+    def format_error_message(self, *args):
+        return self.set_error_message(self.error_message.format(*args))
+
+
+# DOCKER ERRORS
+
+
+DOCKER_COMMAND_ERROR = ErrorClass(
+    "DOCKER_COMMAND_ERROR",
+    "Please verify if Docker client is installed and running."
+)
+
+
+DOCKER_DAEMON_ERROR = ErrorClass(
+    "DOCKER_DAEMON_ERROR",
+    "Please verify if Docker daemon is running, or try restarting it."
+)
+
+
+DOCKER_VERSION_ERROR = ErrorClass(
+    "DOCKER_VERSION_ERROR",
+    "An error occured while retrieving Docker version. Please try verifying it manually."
+)
+
+
+DOCKER_PULL_ERROR = ErrorClass(
+    "DOCKER_PULL_ERROR",
+    "An error occurred while pulling a sample image. Please validate if all necessary components to pull an image are running properly."
+)
+
+
+# HELM ERRORS
+
+
+HELM_COMMAND_ERROR = ErrorClass(
+    "HELM_COMMAND_ERROR",
+    "Please verify if Helm is installed."
+)
+
+HELM_VERSION_ERROR = ErrorClass(
+    "HELM_VERSION_ERROR",
+    "An error occurred while retrieving Helm version. Please make sure that you have the latest Azure CLI version, and that you are using the recommended Helm version."
+)
+
+
+# CONNECTIVITY ERRORS
+
+
+CONNECTIVITY_DNS_ERROR = ErrorClass(
+    "CONNECTIVITY_DNS_ERROR",
+    "Failed to reach DNS for registry '{}'. Please check if the spelling is correct, if the CLI environment is on correct cloud and if you have the right permissions on it."
+)
+
+
+CONNECTIVITY_FORBIDDEN_ERROR = ErrorClass(
+    "CONNECTIVITY_FORBIDDEN_ERROR",
+    "Looks like you don't have access to registry '{}'. Are firewalls and virtual networks enabled?"
+)
+
+
+CONNECTIVITY_CHALLENGE_ERROR = ErrorClass(
+    "CONNECTIVITY_CHALLENGE_ERROR",
+    "Registry '{}' did not issue a challenge."
+)
+
+
+CONNECTIVITY_AAD_LOGIN_ERROR = ErrorClass(
+    "CONNECTIVITY_AAD_LOGIN_ERROR",
+    "Registry '{}' does not support AAD login."
+)
+
+
+CONNECTIVITY_REFRESH_TOKEN_ERROR = ErrorClass(
+    "CONNECTIVITY_REFRESH_TOKEN_ERROR",
+    "Access to registry '{}' was denied. Response code: {}. Please try running 'az login' again to refresh permissions."
+)
+
+
+CONNECTIVITY_ACCESS_TOKEN_ERROR = ErrorClass(
+    "CONNECTIVITY_ACCESS_TOKEN_ERROR",
+    "Access to registry '{}' was denied. Response code: {}. Please try running 'az login' again to refresh permissions."
+)
+
+
+# GENERAL ERRORS
+
+
+LOGIN_SERVER_ERROR = ErrorClass(
+    "LOGIN_SERVER_ERROR",
+    "An error occurred while retrieving the login server for registry '{}'. Please check if the spelling is correct, if the CLI environment is on correct cloud and if you have the right permissions on it."
+)
+
+
+UNEXPECTED_ERROR = ErrorClass(
+    "UNEXPECTED_ERROR",
+    "An unexpected error occurred."
+)
--- a/src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_errors.py
+++ b/src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_errors.py
@@ -17,8 +17,8 @@ class ErrorClass:
     def get_error_message(self, additional_message=None):
         if additional_message:
             return "An error occurred: {}\n{}\n{}".format(self.error_title, self.error_message, additional_message)
-        else:
-            return "An error occurred: {}\n{}".format(self.error_title, self.error_message)
+
+        return "An error occurred: {}\n{}".format(self.error_title, self.error_message)
 
     def set_error_message(self, message):
         return ErrorClass(self.error_title, message)
@@ -31,8 +31,6 @@ class ErrorClass:
 
 
 # DOCKER ERRORS
-
-
 DOCKER_COMMAND_ERROR = ErrorClass(
     "DOCKER_COMMAND_ERROR",
     "Please verify if Docker client is installed and running."
@@ -53,13 +51,11 @@ DOCKER_VERSION_ERROR = ErrorClass(
 
 DOCKER_PULL_ERROR = ErrorClass(
     "DOCKER_PULL_ERROR",
-    "An error occurred while pulling a sample image. Please validate if all necessary components to pull an image are running properly."
+    "An error occurred while pulling a sample image. Please validate your network connection and verify if docker daemon is running properly."
 )
 
 
 # HELM ERRORS
-
-
 HELM_COMMAND_ERROR = ErrorClass(
     "HELM_COMMAND_ERROR",
     "Please verify if Helm is installed."
@@ -72,11 +68,9 @@ HELM_VERSION_ERROR = ErrorClass(
 
 
 # CONNECTIVITY ERRORS
-
-
 CONNECTIVITY_DNS_ERROR = ErrorClass(
     "CONNECTIVITY_DNS_ERROR",
-    "Failed to reach DNS for registry '{}'. Please check if the spelling is correct, if the CLI environment is on correct cloud and if you have the right permissions on it."
+    "Failed to reach DNS for registry '{}'. Please check if the spelling is correct, if the CLI environment is on correct cloud and your network connectivity."
 )
 
 
@@ -111,8 +105,6 @@ CONNECTIVITY_ACCESS_TOKEN_ERROR = ErrorClass(
 
 
 # GENERAL ERRORS
-
-
 LOGIN_SERVER_ERROR = ErrorClass(
     "LOGIN_SERVER_ERROR",
     "An error occurred while retrieving the login server for registry '{}'. Please check if the spelling is correct, if the CLI environment is on correct cloud and if you have the right permissions on it."
--- a/src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
+++ b/src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
@@ -830,12 +830,12 @@ examples:
 
 helps['acr check-health'] = """
 type: command
-short-summary: Gets health information on the environment and, optionally, a target registry.
+short-summary: Gets health information on the environment and optionally a target registry.
 examples:
   - name: Gets health state with target registry 'MyRegistry', skipping confirmation for pulling image.
     text: >
-        az acr check-health -n myRegistry
-  - name: Gets health state with target registry 'myRegistry', without stopping on first error.
+        az acr check-health -n MyRegistry -y
+  - name: Gets health state of the environment, without stopping on first error.
     text: >
-        az acr check-health -n myRegistry --ignore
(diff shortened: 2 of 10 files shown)


sajayantony:
@ankurkhemani please help review this.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_errors.py
```diff
@@ -0,0 +1,125 @@
...
+
+DOCKER_VERSION_ERROR = ErrorClass(
+    "DOCKER_VERSION_ERROR",
+    "An error occured while retrieving Docker version. Please try verifying it manually."
+)
+
+
+DOCKER_PULL_ERROR = ErrorClass(
+    "DOCKER_PULL_ERROR",
+    "An error occurred while pulling a sample image. Please validate if all necessary components to pull an image are running properly."
```
nathana1: Can we be more specific about what components are necessary?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
```diff
@@ -827,3 +827,15 @@
     text: >
         az acr webhook update -n MyWebhook -r MyRegistry --status disabled
 """
+
+helps['acr check-health'] = """
+type: command
+short-summary: Gets health information on the environment and, optionally, a target registry.
+examples:
+  - name: Gets health state with target registry 'MyRegistry', skipping confirmation for pulling image.
```
nathana1: nit:  capitalization of MyRegistry varies.  Probably should make consistent with other samples.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/check_health.py
```diff
@@ -0,0 +1,266 @@
...
+               output.find(DOCKER_IMAGE_UP_TO_DATE.format(IMAGE)) != -1:
+                print_pass("Docker pull of '{}'".format(IMAGE))
+            else:
+                _handle_error(DOCKER_PULL_ERROR, ignore_errors)
+
+
+# Get current CLI version
+def _get_cli_version():
+
+    acr_cli_version = "2.2.8"
```
nathana1: are people going to need to update this when they make changes?  can it be a constant somewhere (or is it already)?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_errors.py
```diff
@@ -0,0 +1,125 @@
...
+    "An error occurred while retrieving Helm version. Please make sure that you have the latest Azure CLI version, and that you are using the recommended Helm version."
+)
+
+
+# CONNECTIVITY ERRORS
+
+
+CONNECTIVITY_DNS_ERROR = ErrorClass(
+    "CONNECTIVITY_DNS_ERROR",
+    "Failed to reach DNS for registry '{}'. Please check if the spelling is correct, if the CLI environment is on correct cloud and if you have the right permissions on it."
```
nathana1: It seems like this should be about network connectivity instead of permissions

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/check_health.py
```diff
@@ -0,0 +1,266 @@
...
+
+from knack.util import CLIError
+from .custom import get_docker_command
+from ._docker_utils import _get_aad_token
+from .helm import get_helm_command
+from ._utils import get_registry_by_name
+from ._errors import ErrorClass
+
+
+DOCKER_PULL_SUCCEEDED = "Downloaded newer image for {}"
```
nathana1: does the docker cli get localized?  This won't work for non-English versions if it does.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/check_health.py
avtakkar: It's in the version file

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/check_health.py
```diff
@@ -0,0 +1,266 @@
...
+        if error.error_title != DOCKER_DAEMON_ERROR.error_title:
+            return  # We cannot proceed if the error is unexpected or with docker command
+        else:
+            docker_daemon_available = False
+
+    if docker_daemon_available:
+        print("Docker daemon status: available")
+
+    # Docker version check
+    output, stderr = _subprocess_communicate([docker_command, "--version"])
```
avtakkar: Some warnings written to stderr can potentially be ignored. Such as: `WARNING: Error loading config file: /xyz/.docker/config.json: stat /xyz/.docker/config.json: permission denied`. If `ignore_errors` is `false`, probably we should not raise an error in this case?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
```diff
@@ -827,3 +827,15 @@
     text: >
         az acr webhook update -n MyWebhook -r MyRegistry --status disabled
 """
+
+helps['acr check-health'] = """
+type: command
+short-summary: Gets health information on the environment and, optionally, a target registry.
```
ankurkhemani: nit: too many comas
`Gets health information on the environment and optionally a target registry.` should be good.

ankurkhemani: Also, do we want to add an example where we don't specify target registry?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
```diff
@@ -827,3 +827,15 @@
...
+examples:
+  - name: Gets health state with target registry 'MyRegistry', skipping confirmation for pulling image.
+    text: >
+        az acr check-health -n myRegistry
+  - name: Gets health state with target registry 'myRegistry', without stopping on first error.
+    text: >
+        az acr check-health -n myRegistry --ignore
```
ankurkhemani: As per `params` file, the argument is `--ignore-errors` ?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/check_health.py
```diff
@@ -0,0 +1,266 @@
...
+from ._docker_utils import _get_aad_token
+from .helm import get_helm_command
+from ._utils import get_registry_by_name
+from ._errors import ErrorClass
+
+
+DOCKER_PULL_SUCCEEDED = "Downloaded newer image for {}"
+DOCKER_IMAGE_UP_TO_DATE = "Image is up to date for {}"
+IMAGE = "mcr.microsoft.com/mcr/hello-world:latest"
+FAQ_MESSAGE = "\nPlease refer to https://aka.ms/acr/faq for more informations."
```
ankurkhemani: nit: information

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/check_health.py
```diff
@@ -0,0 +1,266 @@
...
+
+
+# Utilities functions
+
+
```
ankurkhemani: nit: can remove 2 extra lines

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_errors.py
```diff
@@ -0,0 +1,125 @@
...
+    def append_error_message(self, message):
+        return self.set_error_message(self.error_message + "\n" + message)
+
+    def format_error_message(self, *args):
+        return self.set_error_message(self.error_message.format(*args))
+
+
+# DOCKER ERRORS
+
+
```
ankurkhemani: nit: remove extra 2 lines after comment from all places. Usually, comment should be right above the section you are referring to.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/check_health.py
ankurkhemani: @ni-reis  This needs to be updated for every future cli version increment in `setup.py` ? Can you somehow parse it from there directly else add a comment here. I feel this can be easily missed since the CI doesn't check this.

src/command_modules/azure-cli-acr/setup.py
```diff
@@ -14,7 +14,7 @@
     logger.warn("Wheel is not available, disabling bdist_wheel hook")
     cmdclass = {}
 
-VERSION = "2.2.7"
+VERSION = "2.2.8"
```
ankurkhemani: Check if this needs to be `2.2.9` based on whether `2.2.8` is already released (I see dev already has `2.2.8`). If not, you need to merge the `history.rst` as per this [commit ](https://github.com/Azure/azure-cli/commit/a93391639134617536190e7eca57710fa6ce524b)

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/check_health.py
```diff
@@ -0,0 +1,266 @@
...
+            _handle_error(result_from_token, ignore_errors)
+            return
+
+        print_pass("Obtain access token for registry '{}'".format(login_server))
+
+        return
+
+    # If return is not of type ErrorClass, then it is the token
+    print_pass("Challenge endpoint {}".format(url))
+    print_pass("Obtain refresh token for registry '{}'".format(login_server))
```
ankurkhemani: nit: consider `Fetch` or `Get` instead of `Obtain`

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/check_health.py
```diff
@@ -0,0 +1,266 @@
...
+    print(Fore.YELLOW + str(message) + Style.RESET_ALL)
+
+
+def print_error(message):
+    from colorama import Fore, Style, init
+    init()
+    print(Fore.RED + str(message) + Style.RESET_ALL)
+
+
+def _handle_error(error, ignore_errors):
```
ankurkhemani: method can probably be renamed to `_throw_error`

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/check_health.py
ankurkhemani: Also, currently installed acr version on client can be different from the latest acr cli version? Use `az --version` to parse the current acr version?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_errors.py
ni-reis: I agree. Which should be the components? I can only think of docker daemon right now...

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/check_health.py
ni-reis: Fetch sounds good :)

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/check_health.py
ni-reis: I don't think we can get latest version if it's not installed already. I had some issues reading from the setup file on the version, but I'll try it again. If it keeps failing I will use `az --version`.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/check_health.py
ni-reis: What do you mean by localized? I don't think the docker cli is available in other languages than english

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
ni-reis: Changed to 'MyRegistry', as it is the standard being used.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
ni-reis: Agreed, and added an example without a registry.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_help.py
ni-reis: Yes :) fixed it, thank you!

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_errors.py
ni-reis: I see. Removed

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_errors.py
ni-reis: Agreed.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/check_health.py
ni-reis: fixed

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/check_health.py
ni-reis: fixed

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/check_health.py
ni-reis: Do you think we should parse specific errors, or maybe just add a parameter to ignore errors and just print the output?
Parsing errors can be hard and is vulnerable in case the error message changes. Maybe we can add a ignore parameter for simple commands like `docker --version` or `helm version --client`, and just throw on more complex commands like docker pull.

avtakkar: I think that sounds reasonable. However, on `docker pull`, the warning is still written to stderr. How about printing anything that starts with `WARNING` in stderr and ignoring it? I agree that this is a bit of a special case that may not be that common.

ni-reis: I see, so we should cover every command since this may happens everywhere.
I'm not sure if it's a good idea to print things to stderr in this command since we want to give users the every possible info, but I also agree we shouldn't just ignore it.

Considering this, we can make the function `_subprocess_communicate` similar to
```
if stderr.lower().startswith("warning"):
    warning = stderr
    stderr = None
...
return output, warning, stderr
```
and if there's a warning but no error, we use `print_warning` after printing the output.
What do you think of this?

avtakkar: I think that sounds good.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_errors.py
ni-reis: Using network connection and docker daemon by now :)

ni-reis: Yes. I will suggest network connection and docker daemon checks, it should be good by now.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/check_health.py
ni-reis: Just checked. Docker cli is only available in english, so this should be ok by now :)

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/check_health.py
ni-reis: Changed to this model.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/check_health.py
ni-reis: So, I found the code invoked by `az --version`. I adapted it to our best interest in order to optimize it. The code can be found [here](https://github.com/Azure/azure-cli/blob/c2955e842c86aa0b65d7a98745a0596eb7919f3a/src/azure-cli-core/azure/cli/core/util.py#L116). Although this involves parsing, which is not ideal, it is pretty reliable, since if there was a change on the code we would be notified and able to adapt our code as well.

src/command_modules/azure-cli-acr/setup.py
ni-reis: Updated to 2.2.9 :)

src/command_modules/azure-cli-acr/setup.py
```diff
@@ -14,7 +14,7 @@
     logger.warn("Wheel is not available, disabling bdist_wheel hook")
     cmdclass = {}
 
-VERSION = "2.2.8"
+VERSION = "2.2.9"
```
ankurkhemani: Is 2.2.8 released though? If not (https://pypi.org/project/azure-cli-acr/), you want to keep 2.2.8 and merge history files instead.

ni-reis: I see, 2.2.8 is not released yet. When is the next release? Will it be able to include this PR?

sajayantony: I'm guessing there will be no update on this section of the code when we add features to the ACR module will az version be consistent. Should we add a test to ensure this?

ni-reis: You say, add a test to ensure that the version output for `az acr check-health` is the same outputted for `az --version`?

sajayantony: Test is not required if are calling a core function.

ni-reis: So you think that's ok?

nathana1:
LGTM :shipit:

src/command_modules/azure-cli-acr/setup.py
ankurkhemani: Release history: https://github.com/Azure/azure-cli/releases
I think we can merge this with `2.2.8` if we are not going to sit on this for long.

sajayantony:
LGTM!

PR #9490, opened 2019-05-28
generic command to send out any kinds of HTTP requests

### LINKED ISSUE #4041: Functionapp:Add ability to create a Function App or Function key directly
### Description

The Azure CLI should be able to create a function-app-scoped or function-scoped key in an Azure Function App via the `functionapp` subcommand. It should also be able to retrieve existing keys.

### LINKED ISSUE #7618: `az curl` or `az rest` for custom REST support (coverage)

The Azure docs have a REST API explorer, and that is helpful for one-off testing, but isn't very helpful for scripting or for REST APIs that haven't made their way into the Azure CLI.

I'd like to propose adding a new subcommand, such as `az curl` or `az rest`, that allows a user to call a resource provider with a custom payload. I could see the feature looking something like this:

```bash
az {curl,rest} /subscriptions/{subscriptionId}/providers/{Resource.Provider}/{endpoint} \
  --api-version {version} \
  --payload '{"json": "payload"}'
```

or 

```bash
az {curl,rest} /{Resource.Provider}/{endpoint} \
  --subscription {sub-id} \
  --api-version {version} \
  --payload '{"json": "payload"}'
```
or

```bash
az {curl,rest} {endpoint} \
  --subscription {sub-id} \
  --resource-provider {Resource.Provder} \
  --api-version {version} \
  --payload '{"json": "payload"}'
```

This would allow users to pass JSON payloads for scripting or for testing and validating of the REST API for custom clients.

Currently you can `curl` the Azure REST API, but you have to pass a lot of authentication context that will persist in your shell history, which can be a security concern. This would alleviate that and allow a user to interact with the raw REST API while letting the Azure CLI handle authentication contexts on behalf of the user.

**Additional context**

You can see a good example of this with the Cloud Foundry CLI, `cf`. You can `cf curl` to interact with raw API resources the CLI doesn't support out of the box.

yugangw-msft:
Fix #7618, a community ask, which will deal with any short/mid term service gaps. ~~To dogfood the support, I used functionapp key management scenario, hence I also fix #4041, another long time community ask~~.


src/command_modules/azure-cli-resource/HISTORY.rst
```diff
@@ -2,6 +2,7 @@
 
 Release History
 ===============
+* 'az rest': new command to sent out any REST request
```
adewaleo: typo: new command to **send** out an / any REST request

maybe:

```suggestion
* 'az rest': new command for making http requests 
```

or 

```suggestion
* 'az rest': new command for making REST calls 
```

src/command_modules/azure-cli-resource/azure/cli/command_modules/resource/_help.py
```diff
@@ -1193,3 +1193,15 @@
 type: group
 short-summary: Manage resource tags.
 """
+
+helps['rest'] = """
+    type: command
+    short-summary: invoke custom request
```
adewaleo: Make / invoke **a** custom http request?

src/command_modules/azure-cli-resource/azure/cli/command_modules/resource/_params.py
```diff
@@ -276,3 +276,16 @@ def load_arguments(self, _):
     with self.argument_context('account management-group update') as c:
         c.argument('display_name', options_list=['--display-name', '-d'])
         c.argument('parent_id', options_list=['--parent', '-p'])
+
+    with self.argument_context('rest') as c:
+        c.argument('method', arg_type=get_enum_type(['head', 'get', 'put', 'post', 'delete', 'options', 'patch'], default='get'),
+                   help='HTTP request method')
+        c.argument('url', help="request url")
+        c.argument('headers', nargs='+', help="Space-separated appsettings in KEY=VALUE format or JSON string. Use @{file} to load from a file")
```
adewaleo:
```suggestion
        c.argument('headers', nargs='+', help="Space-separated headers in KEY=VALUE format or JSON string. Use @{file} to load from a file")
```

src/command_modules/azure-cli-resource/azure/cli/command_modules/resource/_params.py
```diff
@@ -276,3 +276,16 @@ def load_arguments(self, _):
...
+        c.argument('uri_parameters', nargs='+', help='Space-separated appsettings in KEY=VALUE format or JSON string. Use @{file} to load from a file')
```
adewaleo:
```suggestion
        c.argument('uri_parameters', nargs='+', help='Space-separated query parameters in KEY=VALUE format or JSON string. Use @{file} to load from a file')
```

or 

```suggestion
        c.argument('uri_parameters', nargs='+', help='Space-separated uri parameters in KEY=VALUE format or JSON string. Use @{file} to load from a file')
```

src/command_modules/azure-cli-resource/azure/cli/command_modules/resource/_params.py
```diff
@@ -276,3 +276,16 @@ def load_arguments(self, _):
...
+        c.argument('skip_authorization_header', action='store_true', help='do not auto append "Authorization" header')
+        c.argument('body', help='request body')
+        c.argument('output_file', help='save response payload to a file')
+        c.argument('resource', help='Resource which CLI would acquire token so to put on the "Authorization" header. '
```
adewaleo: This can possibly reworded / made more concise.

Also should this be called `--endpoint`, `--auth-resource` or `--resource-type` or something?
An example using this parameter would be wonderful, as this option might be one of the least straightforward, even those it seems to be simply the base url or a prefix of the request url.

```suggestion
        c.argument('resource', options_list=['--endpoint'], help='Endpoint to use when retrieving the auth token for the "AUTHORIZATION" header. Defaults to an endpoint in `az cloud show`. Values from `az cloud show --query "endpoints"`'
```

yugangw-msft:
@adewaleo, feedback addressed, please take a look.

src/command_modules/azure-cli-resource/azure/cli/command_modules/resource/_params.py
adewaleo: Proposed re-wording
```suggestion
        c.argument('resource', help='Resource for which the CLI should acquire a token to be used for authorization. The token will be place in the "Authorization" header.  Values from `az cloud show --query "endpoints"'  
```
Instead of hardcoding the values from. Maybe we should use populator commands.

adewaleo: @yugangw-msft 
Proposed re-wording
```suggestion
        c.argument('resource', help='Resource for which the CLI should acquire a token to be used for authorization. The token will be place in the "Authorization" header.  Values from `az cloud show --query "endpoints"'  
```
Instead of hardcoding the values from. Maybe we should use populator commands.

rrmistry:
Hi @yugangw-msft , not sure if this is already a known issue but I'm seeing problems with JSON bodies where property values are string type. Ref https://github.com/Azure/azure-cli/issues/13340#issuecomment-624367033

What would be the best way to resolve this?

PR #9494, opened 2019-05-28
test: fix a flaky test

yugangw-msft:


yugangw-msft:
It is a test flakiness not a service end issue, that it uses the current time to calculate the cred valid length, which is wrong.

PR #9512, opened 2019-05-30
[HDInsight] support ssh_public_key

aim-for-better:
HDInsight Fix can not support --ssh-public-key when create cluster
---

### DIFF
--- a/src/command_modules/azure-cli-hdinsight/azure/cli/command_modules/hdinsight/custom.py
+++ b/src/command_modules/azure-cli-hdinsight/azure/cli/command_modules/hdinsight/custom.py
@@ -36,7 +36,7 @@ def create_cluster(cmd, client, cluster_name, resource_group_name, cluster_type,
     from azure.mgmt.hdinsight.models import ClusterCreateParametersExtended, ClusterCreateProperties, OSType, \
         ClusterDefinition, ComputeProfile, HardwareProfile, Role, OsProfile, LinuxOperatingSystemProfile, \
         StorageProfile, StorageAccount, DataDisksGroups, SecurityProfile, \
-        DirectoryType, DiskEncryptionProperties, Tier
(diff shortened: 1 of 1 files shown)


src/command_modules/azure-cli-hdinsight/azure/cli/command_modules/hdinsight/custom.py
```diff
@@ -157,7 +157,11 @@ def create_cluster(cmd, client, cluster_name, resource_group_name, cluster_type,
         linux_operating_system_profile=LinuxOperatingSystemProfile(
             username=ssh_username,
             password=ssh_password,
-            ssh_public_key=ssh_public_key
+            ssh_profile=ssh_public_key and SshProfile(
```
tjprescott: Please add a test for this. Using `and` here does not seem like it would work. If you supply ssh_public_key it seems like you would just get two truthy values and ssh_profile would evaluate to True.

aim-for-better: > Please add a test for this. Using `and` here does not seem like it would work. If you supply ssh_public_key it seems like you would just get two truthy values and ssh_profile would evaluate to True instead of an actual SSH key value or an SSH Profile object. Did you mean `or`?

Hi tjprescott, Thanks for your comment. We have tested this.  This usage works well. If supply ssh_public_key ssh_profile is  a SshProfile object.

tjprescott: Interesting. After some investigating it seems like this is a more concise (albeit less intuitive) way of expressing:
```Python
ssh_profile=SshProfile(
  public_keys=[SshPublicKey(
    certificate_data=ssh_public_key
  )]
) if ssh_public_key else None

```

PR #9540, opened 2019-06-02
Remove line continuation from inside 'quotes' in help files of az resource

zikalino:
Line continuation inside 'quotes' cause json parsing error, so removed.
---


src/command_modules/azure-cli-resource/azure/cli/command_modules/resource/_help.py
```diff
@@ -226,10 +226,10 @@
         az deployment create --location WestUS --template-uri https://myresource/azuredeploy.json --parameters @myparameters.json
   - name: Create a deployment from a local template file, using parameters from a JSON string.
     text: |
-        az deployment create --location WestUS --template-file azuredeploy.json --parameters '{ \\
```
tjprescott: These quotes actually make the example copy-pastable in bash. Removing them breaks this. Where are you seeing a JSON parse error?

zikalino: I am using Ubuntu in Docker on Windows. I have pasted the screenshots below.

I used example from the documentation here:

https://docs.microsoft.com/en-us/cli/azure/resource?view=azure-cli-latest#az-resource-create

First screenshot shows example just pasted directly from the documentation: 

example-invalid

There's json parse error, because \ appears to be part of passed json.

Below I removed \ from within '' and now json is parsed correctly (some other error occurs):

example-valid

zikalino: + tried on Ubuntu on Windows (original example):

ubuntu

tjprescott: Sorry I mean the backslashes, not quotes.

adewaleo:
> I can also repro this on OSX, which is strange since I thought the backslash was bash-specific syntax that needed to be there. In any case, yes, you'll need to update the linter rule accordingly to make the CI pass.

Yes it seems like a linter issue. Typically for multiline commands a backslash or uncompleted quote is needed to continue the command on multiple lines.

The "faulty_help_example_parameters_rule" linter rule needs to be updated to properly handle situations where a quotation spans many lines. 

This is strange though as I cannot repro this on osx:

```tosin$  az group deployment create -g MyResourceGroup --template-file azuredeploy.json --parameters '{ \
>                 "location": { \
>                     "value": "westus" \
>                 } \
>             }'
[Errno 2] No such file or directory: 'azuredeploy.json'
```
```tosin$  az group deployment create -g MyResourceGroup --template-file azuredeploy.json --parameters '{ \\
>                 "location": { \\
>                     "value": "westus" \\
>                 } \\
>             }'

[Errno 2] No such file or directory: 'azuredeploy.json'
```

zikalino:
@adewaleo 
perhaps in your case it tries to load azuredeploy.json first and fails, so it doesn't attempt to parse parameters' json

adewaleo:
> @adewaleo
> perhaps in your case it tries to load azuredeploy.json first and fails, so it doesn't attempt to parse parameters' json

@zikalino you are right about the CLI failing fast on azuredeploy.json:

```
(cli-venv) [02:15 PM] Work-Mac:azure-cli tosin$ az group deployment create -g MyResourceGroup --parameters '{ \\
>     "location": { \\
>        "value": "westus" \\
>     } \\
> }' --template-file azuredeploy.json 
Unable to parse parameter: { \\
    "location": { \\
       "value": "westus" \\
    } \\
}
(cli-venv) [02:15 PM] Work-Mac:azure-cli tosin$ az group deployment create -g MyResourceGroup --parameters '{ \
>     "location": { \
>        "value": "westus" \
>     } \
> }' --template-file azuredeploy.json
Azure Error: InvalidRequestContent
Message: The request content was invalid and could not be deserialized: 'Required property '$schema' not found in JSON. Path 'properties.template', line 1, position 45.'.
(cli-venv) [02:15 PM] Work-Mac:azure-cli tosin$ 
```

For the second "successful" request, the following payload was sent, which is expected:
```
msrest.http_logger : Request body:
msrest.http_logger : {"properties": {"template": {"resources": []}, 
"parameters": {"location": {"value": "westus"}}, "mode": "Incremental"}}
```

So it seems the help text was valid, at least on osx / linux as the output help text from the yaml is actually the [following](https://docs.microsoft.com/en-us/cli/azure/group/deployment?view=azure-cli-latest#examples):

```
az group deployment create -g MyResourceGroup --template-file azuredeploy.json --parameters '{ \
                                "location": { \
                                    "value": "westus" \
                                } \
                            }'
```

Thanks for the updates to the linter! Could we add the same changes [here](https://github.com/Azure/azure-cli-dev-tools/blob/master/azdev/operations/linter/rules/help_rules.py#L146)?

adewaleo:
Thanks for the updates, especially to the linter! Could we add the same changes [here](https://github.com/Azure/azure-cli-dev-tools/blob/master/azdev/operations/linter/rules/help_rules.py#L146)?

We will be phasing out the core cli linter for the one in azdev...

PR #9546, opened 2019-06-03
[storage] generate-sas needs key

### LINKED ISSUE #9498: az storage container generate-sas fails when using auth-mode login
## Describe the bug

**Command Name**
`az storage container generate-sas` fails when using `auth-mode login`

**Errors:**
```
self.account_key should not be None.
Temp\pip-install-dsvrkf2x\knack\knack\cli.py, ln 206, in invoke
azure\cli\core\commands\__init__.py, ln 560, in execute
azure\cli\core\commands\__init__.py, ln 618, in _run_jobs_serially
azure\cli\core\commands\__init__.py, ln 609, in _run_job
azure\cli\command_modules\storage\__init__.py, ln 240, in new_handler
azure\cli\command_modules\storage\__init__.py, ln 183, in handler
azure\cli\core\commands\__init__.py, ln 588, in _run_job
azure\cli\core\commands\__init__.py, ln 297, in __call__
azure-cli-core\azure\cli\core\__init__.py, ln 461, in default_command_handler
multiapi\storage\v2018_03_28\blob\baseblobservice.py, ln 407, in generate_container_shared_access_signature
multiapi\storage\v2018_03_28\common\_error.py, ln 130, in _validate_not_none
ValueError: self.account_key should not be None.
```

## To Reproduce:

1. Create a service principal account
2. For an existing storage account, assign role Storage Account Contributor to this service principal
3. 'az login' as service principal
4. `az storage container generate-sas --name {} --auth-mode {} --account-name {} --debug`

## Expected Behavior

SAS generated.

Note that SAS is correctly generated if `auth-mode key` is used instead

## Environment Summary
```
Windows-10-10.0.17763-SP0
Python 3.6.6
Shell: cmd.exe

azure-cli 2.0.65

```
## Additional Context

--debug output is available if required

limingu:
Fix #9498
---

### DIFF
--- a/src/command_modules/azure-cli-storage/HISTORY.rst
+++ b/src/command_modules/azure-cli-storage/HISTORY.rst
@@ -3,6 +3,10 @@
 Release History
 ===============
 
+2.4.3
++++++
+* `storage container generate-sas`: Fix missing account key
+
 2.4.2
 +++++
 * `storage blob sync`: add sync command for storage blob
--- a/src/command_modules/azure-cli-storage/azure/cli/command_modules/storage/commands.py
+++ b/src/command_modules/azure-cli-storage/azure/cli/command_modules/storage/commands.py
@@ -212,7 +212,7 @@ def load_command_table(self, _):  # pylint: disable=too-many-locals, too-many-st
         g.storage_command_oauth('create', 'create_container',
                                 transform=create_boolean_result_output_transformer('created'),
                                 table_transformer=transform_boolean_for_table)
-        g.storage_command_oauth('generate-sas', 'generate_container_shared_access_signature')
+        g.storage_command('generate-sas', 'generate_container_shared_access_signature')
         g.storage_command_oauth('exists', 'exists', transform=create_boolean_result_output_transformer('exists'),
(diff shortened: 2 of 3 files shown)


src/command_modules/azure-cli-storage/HISTORY.rst
```diff
@@ -3,6 +3,10 @@
 Release History
 ===============
 
+2.4.3
++++++
+* `storage container generate-sas`: Fix missing account key
```
tjprescott: Does this actually fix the problem or just improve the error message?

limingu: The root cause of the problem is that generate_container_shared_access_signature need account key. So auth_mode login won't work as it only provide the token. So I removed the auth_mode of login. The user has to use the account key to generate sas.

tjprescott: So they will see a better error message?

limingu: The user has to use account key to call generate-sas. No auth_mode login is allowed and no error will show any more.

PR #9548, opened 2019-06-03
Renamed command

dorfire:


dorfire:
@sajayantony @northtyphoon

dorfire:
This new command (`acr pack`) was only introduced last week, and hasn't even been documented. I added the preview flag to it as a part of this PR.

PR #9551, opened 2019-06-04
Consolidate Primary command_modules

marstr:
Related to #8592

This PR seeks to consolidate all of the command modules that are owned by the CLI team directly into the [`azure-cli`](https://pypi.org/project/azure-cli) package.

The goals are to:
- Reduce installation time, by reducing the number of pip invocations.
- Streamline version management, by reducing the number of individual version numbers present.
- Prevent a bootstrapping issue when releasing new command_modules.

The known trade-offs:
- Makes it impossible to install a subset of the Azure CLI without source changes.
- Change log will have some conflicts.


marstr:
@yugangw-msft @tjprescott 

I've created a branch `flatten` which will accumulate all of the migrated command modules. This way, the whole change can be committed at once back to the `dev` branch when we're ready to release. With that change in strategy, this PR is ready for your eyes and approval/review. Once I've gotten sign-off I'll go ahead and merge this PR, and email our partner teams to spread awareness.

src/azure-cli/azure/cli/command_modules/role/_help.py
```diff
@@ -310,6 +310,8 @@
   - name: --create-cert
     short-summary: Create a self-signed certificate to use for the credential.
     long-summary: Use with `--keyvault` to create the certificate in Key Vault. Otherwise, a certificate will be created locally.
+  - name: --password -p
```
tjprescott: New things?

marstr: Just missing help text that I'm adding. :) Not sure why it wasn't getting flagged before.

marstr:
> @tjprescott says:
> Where will the old history entries go?

This is a good question, and I'm open to suggestions. To get the conversation started:

The way I see it there are two things that need to happen to tidy up history/project status as we do this:
1. We must mark the old command modules as deprecated, and point people looking to consume them to the `azure-cli` package. To do this, once I've merged this PR, I want to create a secondary branch before any of the command_modules were migrated that will accept their deprecation messages. We can then publish those PyPI packages for what is likely the last time.
1. We must establish the new change log format in `azure-cli`. To do this, we would start by adding an entry saying something like "The azure-cli package is now hosting all of the command_modules itself." From there, we can add all new change messages as they come in.

PR #9556, opened 2019-06-04
Adding --image-sku as a parameter for sql vm create and update command.

yareyes:
Customers are now able to create a sql vm or update a sql vm with the sql sku type. This parameter is an optional parameter, so it's not a breaking change. 

Additionally, customers are now required to provide the sql license type when creating a vm.

Added corresponding tests for this change. 

---


yareyes:
Hey @tjprescott, I have some failures while running tests. Most of the error I see are from other modules.

Examples like: 
ERROR: {'load_commands': ModuleNotFoundError("No module named 'azure.mgmt.network.v2019_02_01.operations.usages_operations'")} 

or 

FAIL: test_security_pricings (azure.cli.command_modules.security.tests.latest.test_pricings_scenario.SecurityCenterPricingsTests)

How do I get around this? Thanks.

yareyes:
@tjprescott can we have this in the sprint 63 tag? :)

src/command_modules/azure-cli-sqlvm/HISTORY.rst
```diff
@@ -5,7 +5,7 @@ Release History
 0.1.2
```
tjprescott: If you have a breaking change, you need to bump this to 0.2.

src/command_modules/azure-cli-sqlvm/azure/cli/command_modules/sqlvm/_params.py
```diff
@@ -162,7 +162,7 @@ def load_arguments(self, _):
     with self.argument_context('sql vm', arg_group='SQL Server License') as c:
         c.argument('sql_server_license_type',
                    help='SQL Server license type.',
-                   options_list=['--license-type'],
+                   options_list=['--license-type', '-b'],
```
tjprescott: Short options should be used sparingly, and there should be a clear connection between the short and long option. `-b` has no obvious connection to `--license-type`.

Issue #4194, opened 2017-08-10
[EventGrid] Required work prior to GA


tjprescott:
- [ ] Add support for update commands (with generic update support) (https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#generic-update-commands)
- [ ] Add support for IDs in general (--ids parameters) (https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#supporting-the-ids-parameter)
- [ ] Add resource specific features (similar to `az resource` and `az lock`) (something to consider, low priority)
- [x] Finalize command tree structure.
- [ ] Tab completion for names
- [ ] Table formats
- [ ] `az eventgrid topic key regenerate` `--key-name` should have a choice list of the possible values.

snobu:
Isn't EventGrid supposed to still be super secret confidential stuff?

ahamad-MS:
@tjprescott  @jfggdl  This is an old issue and many of the requests are already taken care or not applicable any more and remaining ones are being addressed currently. Here is an update for these issues. I am planning to close this one and open a task on the last issue remaining (table format) if still applicable. thanks

1. Add support for update commands (with generic update support) (https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#generic-update-commands) 

===> This has been already addressed in previous PRs and part of both extension and Stable versions.

2.  Add support for IDs in general (--ids parameters) 

===> This has been already addressed in previous PRs and part of both extension and Stable versions.

3. Add resource specific features (similar to az resource and az lock) (something to consider, low priority)

===> Per discussion with Kalyan, this was already decided as Won't fix or not applicable any more.

4.  Finalize command tree structure.

===> This has been already addressed in previous PRs and part of both extension and Stable versions.

5.  Tab completion for names

===> This has been already addressed in previous PRs and part of both extension and Stable versions.

6.  Table formats

===> This is not clear. Is this in the context of output? Can you please provide some context/info on the ask?

7.  az eventgrid topic key regenerate --key-name should have a choice list of the possible values.

===> This is currently being addressed by this PR:  https://github.com/Azure/azure-cli/pull/9578

jfggdl:
@tjprescott, would you please help us with questions 6 above?

jfggdl:
[...]

PR #9578, opened 2019-06-06
EventGrid: Add key completer for az eventgrid domain key regenerate and az eventgrid topic key regenerate commands

ahamad-MS:
EventGrid: Add key completer for az eventgrid domain key regenerate and az eventgrid topic key regenerate commands. This is one of the asks in Issue https://github.com/Azure/azure-cli/issues/4194 

---


src/command_modules/azure-cli-eventgrid/azure/cli/command_modules/eventgrid/_params.py
```diff
@@ -46,6 +49,12 @@
     arg_type=name_type,
     options_list=['--domain-topic-name'])
 
+key_name_type = CLIArgumentType(
+    help='Key name to regenerate, which can be either \'key1\' or \'key2\'.',
+    arg_type=name_type,
+    options_list=['--key-name'],
+    completer=get_keys_completion_list)
```
tjprescott: A completer isn't really need for this. The simplest way would simply be:

```Python
key_values = ['key1', 'key2']  # this *should* be an enum somewhere in your SDK

with self.argument_context(…) as c:
  c.argument('key_name', arg_type=get_enum_type(key_values))
```

Then you can cut out all of that completer code. Completers are intended for when you need to make REST API calls to get the completions, not for static lists of values.

ahamad-MS: Thanks for the suggestion. Fixed.

FYI, key1 and key2 are not enum values in swagger or sdk 

image

image

PR #9586, opened 2019-06-06
Add support to regenerate policy keys

anusapan:
---


src/command_modules/azure-cli-iot/azure/cli/command_modules/iot/_help.py
```diff
@@ -487,6 +487,15 @@
     crafted: true
 """
 
+helps['iot hub policy key-renew'] = """
```
tjprescott: This should be consistent with other "keys" commands in the CLI, such as `storage account keys`.  Is there a way to list keys?  The typical group has two commands:
```
iot hub policy keys regenerate
iot hub policy keys list
```

If there is no way to list keys, then maybe `regenerate-key` would be the canonical name, but not `renew`.

anusapan: Updated to regenerate-key

PR #9588, opened 2019-06-06
Migrating azure-cli-telemetry to azure-cli package

marstr:
Move the telemetry related code into the main package.


marstr:
This package is depended on by `azure-cli-core`. Therefor, it is not safe to merge it into `azure-cli`. A better fit may be to merge it into `azure-cli-core`. we can decide tomorrow.

PR #9593, opened 2019-06-07
Temp fix for no match for request bug. Fixes #9591

### LINKED ISSUE #9591: Re-recording an old test or creating a new test fails on next run
This is a development issue while creating and running tests.

__To Reproduce -__

1. Delete a test recording (that uses attribute `@ResourceGroupPreparer`) from a module (example- `src\command_modules\azure-cli-appservice\azure\cli\command_modules\appservice\tests\latest\test_webapp_commands.py`). 
2. Run the test, recording will be created and the test will pass. 
3. Re-run the test without deleting the recording this time. It will fail with error similar to 
`No match for the request (<Request (GET) https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourcegroups/clitest.rg000001>)`

I do notice that in my recording, the uri is something like - `https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourcegroups/clitest.rg000002?api-version=2018-05-01`

Notice the difference between `clitest.rg000002` and `clitest.rg000001`. it feels that recording it the first time assigns a number that may not be 000001. While re-running it always expects it to be 000001.

adewaleo:
Temp fix for #9591.

Because NoRecordingPreparers don't record requests during resource creation, during playback they are not aware of the random name used during recording.

The fix is to use a deterministic moniker during recording and playback. This means to get the effect of randomness a user might have to manually change the resource preparer's prefix and / or name length.

Some solutions to keep resource names random during live runs:
1. add some metadata to the recordings that indicates what deterministic names should be used during playback
2. change request matching logic and ensure that each preparer of the same type has a different prefix and name_length
3. Keep traffic recordings for preparers but only use them to determine the name of the prepared resource during playback

---

### DIFF
--- a/src/azure-cli-testsdk/azure/cli/testsdk/preparers.py
+++ b/src/azure-cli-testsdk/azure/cli/testsdk/preparers.py
@@ -34,6 +34,11 @@ class NoTrafficRecordingPreparer(AbstractPreparer):
 
         return None
 
+    @property
+    def random_name(self):
+        # override random_name so that in play-back mode the name is deterministic.
+        # non-deterministic preparers have no recording of requests involved in creating the resource
+        return self.moniker
 
 # Resource Group Preparer and its shorthand decorator
 
--- a/src/azure-cli-testsdk/azure/cli/testsdk/preparers.py
+++ b/src/azure-cli-testsdk/azure/cli/testsdk/preparers.py
@@ -40,14 +40,6 @@ class NoTrafficRecordingPreparer(AbstractPreparer):
         # non-deterministic preparers have no recording of requests involved in creating the resource
         return self.moniker
 
-    @property
-    def moniker(self):
-        # temp workaround to keep test rg's unique enough for live runs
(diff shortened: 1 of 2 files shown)


src/azure-cli-testsdk/azure/cli/testsdk/preparers.py
```diff
@@ -34,6 +34,12 @@ def live_only_execute(self, cli_ctx, command, expect_failure=False):
 
         return None
 
+    @property
+    def random_name(self):
+        # override random_name so that in play-back mode the name is deterministic.
+        # non-deterministic preparers have no recording of requests involved in creating the resource
+        return self.moniker
```
tjprescott: If this is only supposed to apply to playback mode, then should it not be something like:
```
if self.is_live:
  return random_thing
else:
  return self.moniker
```
?

adewaleo:
I will close this given the comments and #9602 fixes this issue

PR #9603, opened 2019-06-07
Test all extensions instead of testing each one

limingu:
---

### DIFF
--- a/scripts/ci/test_extensions.sh
+++ b/scripts/ci/test_extensions.sh
@@ -36,9 +36,4 @@ then
     echo "Failed to verify:" $ext
 fi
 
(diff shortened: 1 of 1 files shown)


scripts/ci/test_extensions.sh
```diff
@@ -27,12 +27,16 @@ for ext in $output; do
...
+done
+
+azdev verify load-all
+if [ $? != 0 ]
+then
+    exit_code=1
+    echo "Failed to verify:" $ext
+fi
+
+for ext in $output; do
```
tjprescott: Technically we don't need this removal block anymore

PR #9609, opened 2019-06-08
rest: simplifications on uri and add telemetry

yugangw-msft:
Add a few optimizations towards the e2e experience by starting with referencing ARM rest API doc:
1. A full uri is not required. If host is missing, assume ARM endpoint
2. Replace the  `{subscriptionId}` token with real id

Also add telemetry data.

//CC: @achandmsft  
Related issue: #7618

### DIFF
--- a/src/command_modules/azure-cli-resource/azure/cli/command_modules/resource/_help.py
+++ b/src/command_modules/azure-cli-resource/azure/cli/command_modules/resource/_help.py
@@ -1200,8 +1200,8 @@ helps['rest'] = """
     examples:
        - name: Get Audit log through Microsoft Graph
          text: >
-             az rest --method get --url https://graph.microsoft.com/beta/auditLogs/directoryAudits
-       - name: Update a Azure Active Directory Graph User's display name
+             az rest --method get --uri https://graph.microsoft.com/beta/auditLogs/directoryAudits
+       - name: Update a Azure Active Directory Graph User's display name in BASH
          text: >
-             az rest --method patch --headers "{\"Content-Type\": \"application/json\"}" --url "https://graph.microsoft.com/v1.0/users/johndoe@azuresdkteam.onmicrosoft.com" --body "{\"displayName\":\"jondoe2\"}"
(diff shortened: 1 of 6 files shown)


src/command_modules/azure-cli-resource/azure/cli/command_modules/resource/custom.py
```diff
@@ -1769,10 +1769,10 @@ def list_resource_links(cmd, scope=None, filter_string=None):
 # endregion
 
 
-def rest_call(cmd, method, url, headers=None, uri_parameters=None,
+def rest_call(cmd, method, uri, headers=None, uri_parameters=None,
```
tjprescott: I would recommend making the `az rest` command `is_preview=True` until the signature stabilizes. One thing

src/command_modules/azure-cli-resource/azure/cli/command_modules/resource/_help.py
```diff
@@ -1200,8 +1200,8 @@
     examples:
        - name: Get Audit log through Microsoft Graph
          text: >
-             az rest --method get --url https://graph.microsoft.com/beta/auditLogs/directoryAudits
-       - name: Update a Azure Active Directory Graph User's display name
+             az rest --method get --uri https://graph.microsoft.com/beta/auditLogs/directoryAudits
+       - name: Update a Azure Active Directory Graph User's display name in BASH
          text: >
-             az rest --method patch --headers "{\"Content-Type\": \"application/json\"}" --url "https://graph.microsoft.com/v1.0/users/johndoe@azuresdkteam.onmicrosoft.com" --body "{\"displayName\":\"jondoe2\"}"
+             az rest --method patch --uri "https://graph.microsoft.com/v1.0/users/johndoe@azuresdkteam.onmicrosoft.com" --body '{"displayName": "jondoe2"}'
```
tjprescott: This will not work on CMD.

adewaleo: @tjprescott because of the single quotes?

tjprescott: Yes.

PR #9610, opened 2019-06-10
Feature/container attach add no move cursor

mohdlee:
We've added --no-move-cursor to make piping easier and less messy, no duplicate lines. We needed this to pipe the container attach/log output to another application

---

### DIFF
--- a/src/command_modules/azure-cli-container/azure/cli/command_modules/container/_params.py
+++ b/src/command_modules/azure-cli-container/azure/cli/command_modules/container/_params.py
@@ -129,3 +129,6 @@ def load_arguments(self, _):
 
     with self.argument_context('container attach') as c:
         c.argument('container_name', help='The container to attach to. If omitted, the first container in the container group will be chosen')
+        c.argument('no_move_cursor', options_list=['--no-move-cursor'], help='Do not move cursor up, useful when piping', action='store_true')
+
+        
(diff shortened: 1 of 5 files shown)


mohdlee:
Travis CI is failing with `ERROR: paramiko 2.5.0 has requirement cryptography>=2.5, but you'll have cryptography 2.4.2 which is incompatible.`, should I wait for upstream fixes?

src/command_modules/azure-cli-container/azure/cli/command_modules/container/_params.py
```diff
@@ -129,3 +130,4 @@ def load_arguments(self, _):
 
     with self.argument_context('container attach') as c:
         c.argument('container_name', help='The container to attach to. If omitted, the first container in the container group will be chosen')
+        c.argument('no_move_cursor', options_list=['--no-move-cursor'], help='Do not move cursor up, useful when piping', action='store_true')
```
tjprescott: Instead of duplicating this line, you should register it once under the `container` scope.

mohdlee: But this only applies to `attach` and `logs`, not other command, is that the right way?

tjprescott: Yes, it is. Since those commands don't have `no_move_cursor`, then this metadata would be ignored.

yonzhan:
Since this PR was very old, we will close it right now. Please open another PR if you still need it.

PR #9654, opened 2019-06-13
Fixing bug doc in acr repository delete

### LINKED ISSUE #9579: Bad documentation bug in az acr repository delete
When I run `az acr repository delete -h`, the help text has this at the
top:

    az acr repository delete : Delete a repository or image in an Azure
    Container Registry.
        This command deletes all associated layer data that are not
        referenced by any other manifest in the container registry.

This made me think that removing a specific tag with the delete command
will not affect any other tags -- but turns out that this is wrong, and
all tags that refer to the same manifest get nuked as well...  I'm
assuming that this is the intended behavior, since running a plain
delete without `-y` does indeed list all of the tags that will be
removed too -- and therefore I consider this a documentation bug.

There are four parts that should be fixed:

1. That top blurb should *really* be fixed to say that it ***does***
   delete all other tags in the same repository that refer to the same
   manifest.

2. I now see that the last example at the bottom of the help text says
   what the expected behavior is too -- but it's at the bottom, so it
   could use some highlighting, maybe make "and all tags" be in all
   caps.

3. Having `-y` is helpful, but it is often used to dismiss plain
   warnings.  That was my original assumption so I used `-y` and didn't
   see the tag listing that you get without `-y`...  It would be much
   better if that list of deleted tags would be shown even if you used
   -y.

4. Finally, I think that it's kind of strange that there is nothing that
   is similar to `docker rmi`, which deletes a tag and only if there are
   no other references to the image, it will delete it too.  The ACR
   `delete` command is of course much too strong in what it deletes, and
   `unlink` is on the other hand too weak since it doesn't remove the
   manifest when it's left with no references.

   This means that to do the `docker rmi` behavior, I must implement my
   own logic that uses `untag` or `delete` depending on whether there are
   additional references to the same sha256...  It would be REALLY helpful
   if such a CLI command existed (and shouldn't be too hard...).

ElianaLiberman:
---

FIXES #9579


src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/repository.py
```diff
@@ -475,6 +472,11 @@ def _delete_manifest_confirmation(login_server,
     if filter_by_manifest:
         images = ", ".join(["'{}:{}'".format(repository, str(x)) for x in filter_by_manifest])
         message += " and all the following images: {}".format(images)
+
+    if yes:
+        print(message)
```
tjprescott: You cannot use a raw print statement. You need to use logger warning or redirect this to stderr instead of stdout.

ElianaLiberman: done

marstr:
Howdy @ElianaLiberman, I've gone ahead and merged your changes with the most recent version of dev to fix some conflicts for you. Other than having your local repository need a pull there should be no further action needed by you.

ElianaLiberman:
> Howdy @ElianaLiberman, I've gone ahead and merged your changes with the most recent version of dev to fix some conflicts for you. Other than having your local repository need a pull there should be no further action needed by you.

Thanks! Can I safely delete the branch?

marstr:
Yes @ElianaLiberman, it is now safe for you to delete the branch.

PR #9655, opened 2019-06-13
[ACR] Support Task Timer Triggers

jaysterp:
---


src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_format.py
```diff
@@ -224,9 +226,29 @@ def _get_value(item, *args):
...
+    try:
+        for arg in args:
+            item = item[arg]
+        if item:
+            schedules = []
+            for trigger in item:
+                schedules.append(trigger['schedule'])
+            return ', '.join(schedules)
+        else:
+            return ' '
```
rosanch: You don't need the else for return here. This will probably throw a lint error

jaysterp: It is fine to return empty strings for pylint. See the previous example for `_get_value`

rosanch: No, I mean 
else: 
  return ' '
Should just be the return. There is no need for the else. If it doesn't enter the if then just return. No need to type else

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
```diff
@@ -169,10 +169,11 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('set_value', options_list=['--set'], help="Task value in '--set name[=value]' format. Multiples supported by passing --set multiple times.", action='append', validator=validate_set)
         c.argument('set_secret', help="Secret task value in '--set-secret name[=value]' format. Multiples supported by passing --set-secret multiple times.", action='append', validator=validate_set_secret)
 
-        # Source Trigger parameters
+        # Trigger parameters
         c.argument('source_trigger_name', help="The name of the source trigger.")
         c.argument('commit_trigger_enabled', help="Indicates whether the source control commit trigger is enabled.", arg_type=get_three_state_flag())
         c.argument('pull_request_trigger_enabled', help="Indicates whether the source control pull request trigger is enabled.", arg_type=get_three_state_flag())
+        c.argument('schedules', nargs='+', help="Space-separated list of schedules represented as cron expressions for the timer triggers. An optional trigger name can be specified using `--schedules name:schedule` format.")
```
ehotinger: is this comment correct, i.e. it doesn't look like it will accept anything space-separated, rather it should be specified by chaining `--schedules` multiple times.

I would also prefer using `--schedule` when chaining instead of the plural

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
```diff
@@ -207,6 +208,12 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
             c.argument('password', options_list=['--password', '-p'], help="The password to login to the custom registry. This can be plain text or a key vault secret URI.")
             c.argument('use_identity', help="The task managed identity used for the credential. Use '[system]' to refer to the system-assigned identity or a client id to refer to a user-assigned identity.")
 
+    with self.argument_context('acr task timer') as c:
+        # Timer trigger parameters
+        c.argument('timer_name', help="The name of the timer trigger.", required=True)
```
ehotinger: maybe `--name` would be nicer instead of `acr task timer --timer-name`

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
```diff
@@ -207,6 +208,12 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
...
+        c.argument('schedule', help="The schedule of the timer trigger represented as a cron expression.")
+        c.argument('timer_enabled', help="Indicates whether the timer trigger is enabled.", arg_type=get_three_state_flag())
```
ehotinger: same for `enabled`

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
```diff
@@ -581,6 +590,110 @@ def acr_task_credential_list(cmd,
...
+    # Use PUT to remove the timer trigger
+
+    # Get the existing task
+    existingTask = client.get_details(resource_group_name, registry_name, task_name)
+    if existingTask.trigger:
+        # Remove the timer trigger
+        trimmed_timer_triggers = remove_timer_trigger(task_name, timer_name, existingTask.trigger.timer_triggers)
+        existingTask.trigger.timer_triggers = trimmed_timer_triggers
+
+        try:
```
ehotinger: the `try/except` here seem misleading, if you fail on network you'd get the wrong exception

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
sajayantony: Agreed singular is better.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
```diff
@@ -497,6 +497,45 @@ def get_custom_registry_credentials(cmd,
...
+    """Remove the timer trigger from the list of existing timer triggers for a task.
+    :param str task_name: The name of the task
+    :param str timer_name: The name of the timer trigger to be removed
+    :param str timer_triggers: The list of existing timer_triggers for a task
+    """
+
+    if not timer_triggers:
+        raise CLIError("No timer triggers exist for the task '{}'.".format(task_name))
+
+    for timer in timer_triggers:
```
ehotinger: maybe warning message if the timer didn't exist to begin with so users can self-correct typos or something

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
northtyphoon: is --name used for task name?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
```diff
@@ -497,6 +497,45 @@ def get_custom_registry_credentials(cmd,
...
+
+    # Provide a default name for the timer if no name was provided.
+    for index, schedule in enumerate(schedules, start=1):
+        # space?
+        split_schedule = None
+        if ':' in schedule:
+            split_schedule = schedule.split(":")
+        timer_triggers.append(
+            TimerTrigger(
+                name=split_schedule[0] if split_schedule else "defaultTimer" + str(index),
```
northtyphoon: Do you need to call strip() to remove the whitespace?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_format.py
jaysterp: Removed this else. Thanks!

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
jaysterp: The `nargs='+'` allows you to pass a space separated list of values.

jaysterp: So do we want the user to provide `--schedules "* */12 * * *" "30 9 * * *"` or `--schedule "* */12 * * *" --schedule  "30 9 * * *"` for create?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
jaysterp: Yes, name is used for task name here already.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_params.py
jaysterp: I agree

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
jaysterp: See fix

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/task.py
jaysterp: The no triggers exist message would not be displayed in this case because the `if` statement was entered  not the `else`. A network failure would cause the command request to time out and show that exception message.

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
```diff
@@ -497,6 +497,45 @@ def get_custom_registry_credentials(cmd,
...
+
+    # Provide a default name for the timer if no name was provided.
+    for index, schedule in enumerate(schedules, start=1):
+        # space?
+        split_schedule = None
+        if ':' in schedule:
+            split_schedule = schedule.split(":")
+        timer_triggers.append(
+            TimerTrigger(
+                name=(split_schedule[0] if split_schedule else "defaultTimer" + str(index)).strip(),
```
northtyphoon: defaultTimer1 => t1

jaysterp: I was thinking `timer1` or `timer2` would be less vague but I have no strong preference. t1 it is!

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/_utils.py
jaysterp: Good suggestion. See fix

jaysterp:
@yugangw-msft could you please review?

src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/tests/latest/test_acr_commands.py
```diff
@@ -84,56 +76,6 @@ def test_check_name_availability(self):
             self.check('nameAvailable', True)
         ])
 
-    @ResourceGroupPreparer()
-    @StorageAccountPreparer(parameter_name='storage_account_for_update')
-    def test_acr_create_with_new_storage(self, resource_group, resource_group_location,
```
yugangw-msft: are you removing the test on purpose? if yes, you probably should remove the recording file as well
//CC: @djyou

jaysterp: @yugangw-msft Yes this was intentional. ACR removed support for `Classic` registries in all api versions , thereby also retiring the use of the `--storage-account-name` parameter. The PR https://github.com/Azure/azure-cli/pull/9676/files addresses these changes in the CLI. Thank you. //cc @ankurkhemani

PR #9666, opened 2019-06-13
Aladdin generated examples for acr, cosmosdb, dls, network, resource, storage, vm

AladdinBI:


src/command_modules/azure-cli-vm/azure/cli/command_modules/vm/_help.py
```diff
@@ -2039,6 +2068,9 @@
   - name: Reimage VMs within a VMSS
     text: az vmss reimage --instance-id instanceId --name MyScaleSet --no-wait --resource-group MyResourceGroup
     crafted: true
+  - name: Reimage VMs within a VMSS. (autogenerated)
+    text: az vmss reimage --instance-id instanceId --name MyScaleSet --resource-group MyResourceGroup --subscription MySubscription
```
tjprescott: Instance ID should just be an integer, so this value is not correct.

marstr:
Howdy @AladdinBI, I've gone ahead and merged your changes with the most recent version of dev to fix some conflicts for you. Other than having your local repository need a pull there should be no further action needed by you.

src/command_modules/azure-cli-vm/azure/cli/command_modules/vm/_help.py
albaloo: I made a commit to fix this, since we don't want to create another PR. I've also made a change to the pipeline to resolve this issue permanently.

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -462,6 +462,24 @@
...
+type: command
+short-summary: Queues a quick build task that builds an app and pushes it into an Azure Container Registry.
+examples:
+  - name: Queue a build for the current directory with the default Oryx-based buildpack.
+    text: az acr pack build -r MyRegistry -t {{.Run.Registry}}/node-app:{{.Run.ID}} .
+  - name: Queue a build for the given GitHub repository with Heroku's buildpacks.
+    text: az acr pack build -r MyRegistry -t {{.Run.Registry}}/node-app:{{.Run.ID}} --pull --builder heroku/buildpacks:18 https://github.com/Azure-Samples/nodejs-docs-hello-world.git
+"""
+
+>>>>>>> 9feda7cfd4ddfe7a27a459eb1ef747fa979807d1
```
tjprescott: There are git diff artifacts in your code. These need to be removed.

albaloo: Yes, not sure how they got there. I resolved the conflicts with vscode and for some reason it didn't remove them. It's fixed now.

Issue #9483, opened 2019-05-27
Update `az feedback` to open issues in github repo of extension if extension is coming from a different repo


gauravsaralMs:
**Is your feature request related to a problem? Please describe.**
Currently giving feedback on any command coming from extension goes into azure cli extension repo. While a extension can even live outside of the main cli extension repo, like devops cli extension.

**Describe the solution you'd like**
Say some user wants to give feedback for command coming from devops-cli-extension, then also currently issue will be opened in cli extension repo, it will be better if issue can directly be opened in azure devops cli github repository

**Describe alternatives you've considered**
NA

**Additional context**
NA

adewaleo:
This should be possible via the extension's [project_url link in the index.json file](https://github.com/Azure/azure-cli-extensions/blob/master/src/index.json#L673). If this could be surfaced up within the CLI, it should be possible to direct a user to this link instead.

cc: @achandmsft

atbagga:
@adewaleo  This was fixed by above PR and is released already. We can close this issue.

adewaleo:
Thanks a lot @atbagga!

PR #9679, opened 2019-06-14
az feedback should open the corresponding repository issues page instead of generic extensions repo page

atbagga:

-----

az feedback can get the project url from the index and file the issue in corresponding github repo instead of azure-cli-extensions repo for all extensions.

Issue #9483


atbagga:
@tjprescott @yugangw-msft Can you take a quick look at the issue and approach?
If it looks OK. I will add the tests.

yugangw-msft:
@adewaleo, please review

src/command_modules/azure-cli-feedback/azure/cli/command_modules/feedback/custom.py
```diff
@@ -46,6 +46,8 @@
 _EXTENSIONS_ISSUES_URL = "aka.ms/azcli/ext/issues"
 _RAW_EXTENSIONS_ISSUES_URL = "https://github.com/Azure/azure-cli-extensions/issues/new"
 
+_EXTENSION_REPOS_URL_TO_ISSUES_URL = '/issues/new'
```
adewaleo: maybe rename this to something like `new_issues_suffix` or `simply new_issues_str` and inline in the only function that uses it. What do you think?

src/command_modules/azure-cli-feedback/azure/cli/command_modules/feedback/custom.py
```diff
@@ -555,7 +559,18 @@ def _get_parent_proc_name():
     return _ISSUES_TEMPLATE_PREFIX.format(pretty_url_name), formatted_issues_url, original_issue_body
 
 
-def _get_minified_issue_url(command_log_file, format_dict, is_ext, capacity):
+def  _get_extension_repo_url(ext_name, raw=False):
+    try:
+        project_url = resolve_project_url_from_index(extension_name=ext_name)
+        if 'https://github.com' in project_url:
+            return project_url.strip('/') + _EXTENSION_REPOS_URL_TO_ISSUES_URL
+    except BaseException as ex:
```
adewaleo: Why catch such a broad exception? Please catch CLI error and any other exceptions you deem necessary

atbagga:
> Nice work! Thanks for your contribution!
> 
> Please also add logic to `_prompt_issue` to inform the user via logger.info() about the regular extensions repo just in case they would like to open an issue there instead. This shouldn't be in the issue body which is used when opening a browser.

Thanks for feedback.
Please take a look at the changes.

src/command_modules/azure-cli-feedback/azure/cli/command_modules/feedback/custom.py
```diff
@@ -741,8 +747,11 @@ def _prompt_issue(recent_command_list):
     #     pyperclip.copy(original_issue)
     # except pyperclip.PyperclipException as ex:
     #     logger.debug(ex)
-
     logger.info(original_issue)
+
+    logger.info('You can also file the issue in the Azure CLI %s repository by opening %s in the browser.',
```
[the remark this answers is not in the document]
atbagga: Sounds good! Done the changes.

src/command_modules/azure-cli-feedback/azure/cli/command_modules/feedback/custom.py
```diff
@@ -41,10 +41,10 @@
 _QUESTIONS_URL = "aka.ms/azcli/questions"
 
 _CLI_ISSUES_URL = "aka.ms/azcli/issues"
-_RAW_CLI_ISSUES_URL = "https://github.com/Azure/azure-cli/issues/new"
+_RAW_CLI_ISSUES_URL = "https://github.com/azure/azure-cli/issues/new"
```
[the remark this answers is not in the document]
atbagga: Kept it this way to avoid converting to lower all the time for comparison. We only use this for browser open and not printing so it should be ok I guess.

src/command_modules/azure-cli-feedback/azure/cli/command_modules/feedback/custom.py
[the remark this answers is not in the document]
atbagga: Sample output: 
image

marstr:
Howdy @atbagga, I've gone ahead and merged your changes with the most recent version of dev to fix some conflicts for you. Other than having your local repository need a pull there should be no further action needed by you.

src/azure-cli/azure/cli/command_modules/feedback/custom.py
```diff
@@ -726,7 +748,6 @@ def _prompt_issue(recent_command_list):
     #     pyperclip.copy(original_issue)
     # except pyperclip.PyperclipException as ex:
     #     logger.debug(ex)
-
```
jonringer: Found the conflict, I deleted this commented section in #9709

atbagga:
Thanks! @jonringer @marstr
Resolved conflicts. We should be good once the test run completes.

src/azure-cli/azure/cli/command_modules/feedback/custom.py
```diff
@@ -552,10 +557,35 @@ def _get_parent_proc_name():
...
+            if not raw and raw_url == _RAW_EXTENSIONS_ISSUES_URL:
+                return _EXTENSIONS_ISSUES_URL
+            return raw_url
+    except (CLIError, NoExtensionCandidatesError) as ex:
+        # since this is going to feedback let it fall back to the generic extensions repo
+        logger.debug(ex)
+    return _RAW_EXTENSIONS_ISSUES_URL if raw else _EXTENSIONS_ISSUES_URL
+
+
+def _is_valid_github_project_url(url):
```
atbagga: @adewaleo I looked at the current project urls in the index file and found a few links like..  https://github.com/Azure/azure-cli-extensions/tree/master/src/azure-firewall which would have broken in case of my previous check. 
Making this change to strictly verify the project url for format **https://github.com/org/project** format.

adewaleo: Where you able to manually test your changes. I.e. running az feedback on:
- an azure firewall command (this special case)
- a regular extension command with its own project url
- a regular extension command with our (cli extension) project url
- regular cli command?

Just test that you can run -h with each command and open the browser to a reasonable page with the correct content.

Thanks a lot!

atbagga: Yes, tested with-
1. an azure firewall command - **az network firewall list**
2. a regular extension command with its own project url - **az devops project list** 
3. a regular extension command with our (cli extension) project url - **az alias list**
4. regular cli command - **az extension add**
5. File generic issue option - **az feedback - option 0**
6. Extension with non giithub.com url in project home page  - **az ml workspace list**

atbagga: Note that -h commands are not logged as extension commands in the metadata so they will come to azure-cli repo irrespective of my change. 
Not sure if this is a recent regression or has been like this always but this behavior can be verified in 2.0.67.

atbagga: @adewaleo I am done with code, test and manual verification from my side. Please take care of the review and merge.

adewaleo: The behavior has been like this

PR #9684, opened 2019-06-14
WIP: Create Dependency List during CI Builds

marstr:
This will help us have reproducible builds, including the ability to lock down transitive dependencies.


marstr:
Closing in favor of #9785, which takes a more prescriptive approach.

PR #9685, opened 2019-06-14
Annotate packages as supporting Python 3.7.

marstr:
They've been distributed on MacOS using Python 3.7 for quite some time. It's good empirical evidence that we have the capacity to work with Python 3.7.

Note: Because this is targeting the `flatten` branch, we have many fewer places where this notation needs to be added.


marstr:
Ping @tjprescott and @yugangw-msft

marstr:
Got approval offline from both @tjprescott and @yugangw-msft, I'm pushing this one in :)

PR #9699, opened 2019-06-17
 role assignment: expose --assignee-principal-type from create command

yugangw-msft:
This is to mitigate the error caused by MSI/ServicePrincipal not fully propagated in the graph. With this, we still let the role assignment creation accepted by RBAC


marstr:
Howdy @yugangw-msft, I've gone ahead and merged your changes with the most recent version of dev to fix some conflicts for you. Other than having your local repository need a pull there should be no further action needed by you.

PR #9701, opened 2019-06-17
support defer argument in azure-cli-extensions

wurp:
---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/commands/__init__.py
+++ b/src/azure-cli-core/azure/cli/core/commands/__init__.py
@@ -171,7 +171,7 @@ class CacheObject(object):
         model_path_regex = re.compile(r':rtype:.*(?P<path>azure.mgmt[a-zA-Z0-9_\.]*)')
         try:
             self._model_name = model_name_regex.search(doc_string).group('model')
-            if  not self._model_path:
+            if not self._model_path:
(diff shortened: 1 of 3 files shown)


src/azure-cli-core/azure/cli/core/commands/__init__.py
```diff
@@ -166,11 +166,13 @@ def _resolve_model(self):
                 break
 
         doc_string = doc_string.replace('\r', '').replace('\n', ' ')
+        doc_string = re.sub(' +', ' ', doc_string)
```
wurp: When "that returns" happens after the line break, the regex on line 170 fails due to the leading space on each line. Compressing multiple spaces to just one fixes it.

src/azure-cli-core/azure/cli/core/commands/__init__.py
```diff
@@ -222,13 +224,13 @@ def prop_dict(self):
             'group': self._resource_group
         }
 
-    def __init__(self, cmd, payload, operation):
+    def __init__(self, cmd, payload, operation, model_path=None):
```
wurp: model_path for extensions tends to be {extension}.{extension subpackage}.vendored_sdks.models instead of that specified in the model docstring. This lets us override the model_path.

PR #9704, opened 2019-06-17
Don't expose the --subscription flag for `az logout`.

### LINKED ISSUE #9694: az logout --subscription

**Describe the bug**

**To Reproduce**
Steps to reproduce the behavior.

**Expected behavior**

**Environment summary**

**Additional context**

marstr:
Fixes #9694

This is just mimicking what I saw here:
https://github.com/Azure/azure-cli/blob/a1f2f2b666a759eb2f593dd7796db6bbd0237614/src/command_modules/azure-cli-profile/azure/cli/command_modules/profile/__init__.py#L65

Testing locally it though, it seemed like `--subscription` was still showing up in the usage for this an other commands using this syntax. Have any insight there, @tjprescott?


src/azure-cli/HISTORY.rst
```diff
@@ -11,6 +11,9 @@ Release History
 
 * Update ADLS version (0.0.45).
 
+**Profile**
+* Suppress `--subscription` argument for logout command.
```
tjprescott: Should this be annotated as a breaking change?  Technically any scripts that might be using it would break, but then it didn't work as expected to begin with.

tjprescott:
> Testing locally it though, it seemed like --subscription was still showing up in the usage for this an other commands using this syntax. Have any insight there, @tjprescott?

Interesting. Let me look into this.

src/azure-cli/HISTORY.rst
marstr: Yeah, I'm not too concerned about this. The only people it would "break" haven't been working anyway. I don't think this needs any escalated attention. Thoughts, @yugangw-msft?

PR #9718, opened 2019-06-18
Fix issue where `--subscription` was not properly suppressed.

tjprescott:
Addresses this comment from PR #9704.

> Testing locally it though, it seemed like --subscription was still showing up in the usage for this an other commands using this syntax. Have any insight there, @tjprescott?

**TODO**
- [x] Add regression test

---


src/azure-cli-core/azure/cli/core/__init__.py
```diff
@@ -277,7 +277,6 @@ def load_arguments(self, command):
                 loader.load_arguments(command)  # this adds entries to the argument registries
                 self.argument_registry.arguments.update(loader.argument_registry.arguments)
                 self.extra_argument_registry.update(loader.extra_argument_registry)
-                loader._update_command_definitions()  # pylint: disable=protected-access
```
tjprescott: Has to be relocated so the CMD_TBL_LOADED event can be called and processed before the arguments are updated.

src/azure-cli/HISTORY.rst
```diff
@@ -2,6 +2,11 @@
 
 Release History
 ===============
+
+**Core**
```
marstr: For the moment, we're still distributing `azure-cli-core` as a separate library. (Could change if we can get @lmazuel what he needs else-where, say the MSAL library) Until then, I'd add this message to `src/azure-cli-core/HISTORY.rst` instead of here.

marstr: (Sorry, I know this is boring review feedback)

tjprescott: I included a modified message in both.

tjprescott:
@marstr please take a second look.

tjprescott:
Moving to Sprint 65 to spend more time addressing the performance impact of this change.

PR #9731, opened 2019-06-19
Fix functionapp slots deployment issue

ankitkumarr:
Fix a minor issue caused by slots. Backend takes care of creating the correct site config. It doesn't need to be explicitly copied by the CLI.

This should not need an update to History.rst.

\cc @panchagnula @yugangw-msft


ankitkumarr:
@yugangw-msft, @tjprescott, would love a review / approval before the end of this Sprint :)

PR #9733, opened 2019-06-19
Replace call to profile.get_refresh_token() with profile.get_raw_token()

BrianBlum:
Only internal method call change.
---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/ams/operations/mru.py
+++ b/src/azure-cli/azure/cli/command_modules/ams/operations/mru.py
@@ -10,7 +10,6 @@ import requests
 from azure.cli.core.util import CLIError
 from azure.cli.command_modules.ams._completers import get_mru_type_completion_list
 from azure.cli.core.commands.client_factory import get_subscription_id
-from azure.cli.core._profile import _ACCESS_TOKEN
 
 _rut_dict = {0: 'S1',
              1: 'S2',
@@ -62,7 +61,7 @@ class MediaV2Client():
         from msrestazure.tools import resource_id
         from azure.cli.command_modules.ams._sdk_utils import (get_media_namespace, get_media_type)
 
-        access_token = self.profile.get_raw_token()[0][2].get(_ACCESS_TOKEN)
+        access_token = self.profile.get_raw_token()[0][2].get('accessToken')
 
         media_old_rp_url = resource_id(subscription=get_subscription_id(cli_ctx),
                                        resource_group=resource_group_name,
@@ -87,7 +86,7 @@ class MediaV2Client():
 
     def _get_v2_access_token(self):
         # pylint: disable=protected-access
-        access_token = self.profile.get_raw_token(resource=self.v2_media_api_resource)[0][2].get(_ACCESS_TOKEN)
+        access_token = self.profile.get_raw_token(resource=self.v2_media_api_resource)[0][2].get('accessToken')
         return access_token
 
     def set_mru(self, account_id, count, type):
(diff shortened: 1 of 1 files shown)


src/azure-cli/azure/cli/command_modules/ams/operations/mru.py
```diff
@@ -9,7 +9,13 @@
...
 from azure.cli.core.util import CLIError
 from azure.cli.command_modules.ams._completers import get_mru_type_completion_list
-
+from azure.cli.core.commands.client_factory import get_subscription_id
+from azure.cli.core._profile import _ACCESS_TOKEN
+from azure.cli.core._profile import _REFRESH_TOKEN
+from azure.cli.core._profile import _SERVICE_PRINCIPAL
+from azure.cli.core._profile import _USER_ENTITY
+from azure.cli.core._profile import _USER_TYPE
+from azure.cli.core._profile import _USER
```
tjprescott: These should just be combined:
```Python
from azure.cli.core._profile import (
_ACCESS_TOKEN, _REFRESH_TOKEN, ...)
```

BrianBlum: I'm using those profile constants in the results of the profile calls. If I just replicate the strings in my code, and profile changes a string, I would definitely break.

src/azure-cli/azure/cli/command_modules/ams/operations/mru.py
```diff
@@ -84,24 +90,34 @@ def _get_v2_api_endpoint(self, cli_ctx, resource_group_name, account_name):
...
+            refresh_token = raw_token[_REFRESH_TOKEN]
+            return context.acquire_token_with_refresh_token(refresh_token,
+                                                            clientId,
+                                                            self.v2_media_api_resource).get(_ACCESS_TOKEN)
+        if user_type == _SERVICE_PRINCIPAL:
+            sp_authinfo = self.profile.get_sp_auth_info()
+            acq_token = context.acquire_token_with_client_credentials(self.v2_media_api_resource,
+                                                                      sp_authinfo['clientId'],
+                                                                      sp_authinfo['clientSecret'])
+            return acq_token['accessToken']
```
yugangw-msft: I believe the logics here can be replaced with one line call of `get_row_token(resource=self.v2_media_api_resource)`

BrianBlum: This statement works for both user logins and service principal: self.profile.get_raw_token(resource=self.v2_media_api_resource)[0][2].get(_ACCESS_TOKEN)
Do you recommend **NOT** using _ACCESS_TOKEN from _profile.py?

src/azure-cli/azure/cli/command_modules/ams/operations/mru.py
```diff
@@ -55,15 +56,15 @@ def __init__(self, cli_ctx, resource_group_name, account_name):
...
     def _get_v2_api_endpoint(self, cli_ctx, resource_group_name, account_name):
         from msrestazure.tools import resource_id
         from azure.cli.command_modules.ams._sdk_utils import (get_media_namespace, get_media_type)
-        from azure.cli.core.commands.client_factory import get_subscription_id
 
-        refresh_token_obj = self.profile.get_refresh_token()
-        access_token = refresh_token_obj[2]
+        raw_token_obj = self.profile.get_raw_token()
+        raw_token = raw_token_obj[0][2]
+        access_token = raw_token[_ACCESS_TOKEN]
```
yugangw-msft: should be `_, access_token, _ = self.profile.get_raw_token()`

src/azure-cli/azure/cli/command_modules/ams/operations/mru.py
tjprescott: You should not take a dependency on anything that is private (leading _) as it explicitly makes no guarantee of stability.

BrianBlum: @tjprescott Thanks Travis. I'll use a local string.

BrianBlum:
@tjprescott Thanks Travis.

BrianBlum:
@tjprescott How do I find which Milestone this fix is going in? We have a customer waiting for sp login support.
Thanks

tjprescott:
https://github.com/Azure/azure-cli/milestones

PR #9745, opened 2019-06-20
Fixing recording location after #9699

marstr:
This fix didn't quite make it where it needed to be!


marstr:
This will fix https://dev.azure.com/azure-sdk/public/_build/results?buildId=54052 too.

PR #9760, opened 2019-06-21
Adding missing package_data entries

marstr:
This was missed during initial migration


marstr:
For posterity, I sped this up significantly by using the following command to help me find all instances of `package_data` in the former command_module structure:

``` bash
 git diff azure-cli-2.0.67 upstream/dev -- **setup.py | grep package_data
```

PR #9764, opened 2019-06-21
Restore linter exclusions

marstr:


marstr:
The following script was used to speed things up a bit:

``` bash
git diff --name-only azure-cli-2.0.67 upstream/dev -- src/command_modules/*/linter_exclusions.yml | xargs -I {} git checkout azure-cli-2.0.67 -- {}
```

.travis.yml
```diff
@@ -30,10 +30,10 @@ jobs:
       env: PURPOSE='Dependency Check'
       script: ./scripts/dependency/check.sh
       python: 3.6
-#    - stage: verify
-#      env: PURPOSE='Lint Command Table and Help'
-#      script: ./scripts/ci/verify_linter.sh
-#      python: 3.6
```
tjprescott: I would leave this disabled for now and just merge the relocation of the YML files

marstr: Works for me!

marstr: Fixed in f07fb88

PR #9772, opened 2019-06-22
[ACR] Update the Samples for Timer Triggers

jaysterp:
---


jaysterp:
@yugangw-msft I bumped the SDK version here to what it was in my earlier PR https://github.com/Azure/azure-cli/pull/9655. It appears it was changed to an earlier version in the repo restructure last week. What does the syntax `~=` imply for SDK versions in new setup.py?

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -565,11 +568,12 @@
         az acr task create -t hello-world:{{.Run.ID}} -n hello-world -r MyRegistry \\
             -c https://github.com/Azure-Samples/acr-build-helloworld-node.git -f Dockerfile \\
             --assign-identity [system] "/subscriptions/<SUBSCRIPTON ID>/resourcegroups/myResourceGroup/providers/Microsoft.ManagedIdentity/userAssignedIdentities/myUserAssignedIdentitiy"
-  - name: Create a Linux task from a public GitHub repository which builds the hello-world image with a git commit trigger and a timer trigger that runs that task at noon on Mondays through Fridays using the default trigger name.
+  - name: Create a Linux task from a public GitHub repository which builds the hello-world image with a system-assigned managed identity and a timer trigger that runs that task at noon on Mondays through Fridays using the default trigger name.
```
northtyphoon: If users read the example, I guess they will wonder wonder where the system-assigned managed identity is used for a docker build task. Maybe a multiple-step task is more close to the real user case.

yugangw-msft:
>  What does the syntax ~= imply for SDK versions in new setup.py?
`azure-mgmt-containerregistry~=3.0.0rc2` means `>= 3.0.0rc2, == 3.0.*`. The mechanism has the goal on convenience that new compatible SDK releases won't incur code change

jaysterp:
> >  What does the syntax ~= imply for SDK versions in new setup.py?
> 
> azure-mgmt-containerregistry~=3.0.0rc2 means >= 3.0.0rc2, == 3.0.*. The mechanism has the goal on convenience that new compatible SDK releases won't incur code change

@yugangw-msft how does the cli resolve which version to use in this case? Is it by most recent?

yugangw-msft:
> @yugangw-msft how does the cli resolve which version to use in this case? Is it by most recent?

For your case, yes, it should be `3.0.<max>`
If you use `~=3.0` , then it should be `3.<max>`
 I referenced the doc [here](https://www.python.org/dev/peps/pep-0440/#compatible-release)

src/azure-cli/azure/cli/command_modules/acr/_help.py
jaysterp: @northtyphoon see updated context

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -654,7 +658,7 @@
 examples:
   - name: Add a timer trigger to a task.
     text: >
-        az acr task timer add -n taskname -r registryname --timer-name myTimer --schedule "30 9 * * 1-5"
+        az acr task timer add -n taskname -r registryname --timer-name t2 --schedule "30 9 * * 1-5"
```
ankurkhemani: nit: why changed to `t2`?

src/azure-cli/azure/cli/command_modules/acr/_format.py
```diff
@@ -226,22 +237,22 @@ def _get_value(item, *args):
...
     :param dict item: The dict object
     """
     try:
         for arg in args:
             item = item[arg]
+        enabled = False
         if item:
-            schedules = []
             for trigger in item:
                 if trigger['status'].lower() == "enabled":
```
ankurkhemani: nit: instead of using hard-coded `enabled`, you can use Enum's value like: `trigger_status.enabled.value` ?

src/azure-cli/azure/cli/command_modules/acr/_help.py
jaysterp: The output does not wrap when printing help. Decided to shorten the name similar to the default trigger name as opposed to putting a new line

src/azure-cli/azure/cli/command_modules/acr/_format.py
jaysterp: We do not pass a client to `_format.py` so we do not have access to the models. I prefer keeping the file more lightweight and just for json response formatting.

marstr:
@jaysterp, looks like CI is failing because only the `setup.py` was updated, and not the `requirements.*.txt` files. They likely weren't present when you started this, so it's just bad timing not something you missed. The fix is simple, please find-replace all (6) instances of `azure-mgmt-containerregistry==2.8.0` to `azure-mgmt-containerregistry==3.0.0rc2`

jaysterp:
> @jaysterp, looks like CI is failing because only the setup.py was updated, and not the requirements.*.txt files. They likely weren't present when you started this, so it's just bad timing not something you missed. The fix is simple, please find-replace all (6) instances of azure-mgmt-containerregistry==2.8.0 to azure-mgmt-containerregistry==3.0.0rc2

Hi @marstr! I started this change  after the setup.py update so I only see 3 instance of `azure-mgmt-containerregistry`. One is in src\azure-cli\setup.py and the other two are in requirements.Linux.external.txt and requirements.Darwinexternal.txt. Those two files have an quite outdated sdk version. Should I update the version is those two files as well?

marstr:
Ah, you may need to rebase on a more recent version of `dev`. There should be a series of requirements.txt files at `src/azure-cli/requirements.py3.linux.txt`, `src/azure-cli/requirements.py2.windows.txt`, etc.

jaysterp:
@marstr Thank you, that did the trick. We are ready to merge this when you are!

Issue #9767, opened 2019-06-21
Extend `az functionapp deployment config-zip` to support Linux Consumption


ankitkumarr:
Currently because `config-zip` use SCM's `api/zipdeploy` for all deployments, it does not work for Linux Consumption (no SCM site). Opening an issue to add support for deployment using storage account in Linux Consumption.

\cc @ahmedelnably

ankitkumarr:
Please assign this to me.

PR #9784, opened 2019-06-24
Add deploy support to Linux Consumption

ankitkumarr:
\cc @panchagnula @ahmedelnably @yugangw-msft 

Fixes https://github.com/Azure/azure-cli/issues/9767
The expiry tokens, zip file format etc is kept consistent with `azure-functions-core-tools` publish command.

I used `print` statements because we need to show the users information and these are not warnings. And it would be difficult to show a progress bar using `logger`.
Additionally, we don't need to show them any json.

```
Uploading...
============================== 100.0%
Upload to storage complete.
Syncing Triggers...
Deployment completed successfully.
```

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -46,7 +46,7 @@ from azure.cli.core.commands.client_factory import get_mgmt_service_client
 from azure.cli.core.commands import LongRunningOperation
 from azure.cli.core.util import in_cloud_console, shell_safe_json_parse, open_page_in_browser, get_json_object, \
     ConfiguredDefaultSetter
-from azure.cli.core import __version__
+from azure.cli.core.commands.client_factory import UA_AGENT
 
 from .tunnel import TunnelServer
 
@@ -282,21 +282,18 @@ def enable_zip_deploy_functionapp(cmd, resource_group_name, name, src, timeout=N
     app = client.web_apps.get(resource_group_name, name)
     parse_plan_id = parse_resource_id(app.server_farm_id)
     plan_info = None
-    retries = 5
     retry_delay = 10  # seconds
     # We need to retry getting the plan because sometimes if the plan is created as part of function app,
     # it can take a couple of tries before it gets the plan
-    while retries > 0:
+    for _ in range(5):
         plan_info = client.app_service_plans.get(parse_plan_id['resource_group'],
                                                  parse_plan_id['name'])
         if plan_info is not None:
             break
-        retries -= 1
         time.sleep(retry_delay)
     if is_plan_consumption(plan_info) and app.reserved:
-        upload_zip_to_storage(cmd, resource_group_name, name, src, slot)
-    else:
-        return enable_zip_deploy(cmd, resource_group_name, name, src, timeout, slot)
+        return upload_zip_to_storage(cmd, resource_group_name, name, src, slot)
+    return enable_zip_deploy(cmd, resource_group_name, name, src, timeout, slot)
 
 
 def enable_zip_deploy(cmd, resource_group_name, name, src, timeout=None, slot=None):
@@ -310,7 +307,7 @@ def enable_zip_deploy(cmd, resource_group_name, name, src, timeout=None, slot=No
     authorization = urllib3.util.make_headers(basic_auth='{0}:{1}'.format(user_name, password))
     headers = authorization
     headers['content-type'] = 'application/octet-stream'
-    headers['User-Agent'] = 'azure-cli:{}'.format(__version__)
+    headers['User-Agent'] = UA_AGENT
 
     import requests
     import os
@@ -350,13 +347,11 @@ def upload_zip_to_storage(cmd, resource_group_name, name, src, slot=None):
         filled_length = int(round(total_length * current) / float(total))
         percents = round(100.0 * current / float(total), 1)
         progress_bar = '=' * filled_length + '-' * (total_length - filled_length)
-        print('\r{} {}%'.format(progress_bar, percents), flush=True, end="")
+        progress_message = 'Uploading {} {}%'.format(progress_bar, percents)
+        cmd.cli_ctx.get_progress_controller().add(message=progress_message)
 
-    print('Uploading...')
     block_blob_service.create_blob_from_path(container_name, blob_name, src, validate_content=True,
                                              progress_callback=progress_callback)
-    # For a new line
-    print('')
 
     now = datetime.datetime.now()
     blob_start = now - datetime.timedelta(minutes=10)
@@ -369,12 +364,11 @@ def upload_zip_to_storage(cmd, resource_group_name, name, src, slot=None):
 
     blob_uri = block_blob_service.make_blob_url(container_name, blob_name, sas_token=blob_token)
     website_run_from_setting = "WEBSITE_RUN_FROM_PACKAGE={}".format(blob_uri)
-    print('Upload to storage complete.')
     update_app_settings(cmd, resource_group_name, name, settings=[website_run_from_setting])
-    print('Syncing triggers...')
     client = web_client_factory(cmd.cli_ctx)
 
     try:
+        logger.info('\nSyncing Triggers...')
         if slot is not None:
             client.web_apps.sync_function_triggers_slot(resource_group_name, name, slot)
         else:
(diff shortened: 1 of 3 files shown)


src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -299,6 +316,66 @@ def enable_zip_deploy(cmd, resource_group_name, name, src, timeout=None, slot=No
...
+
+    # https://gist.github.com/vladignatyev/06860ec2040cb497f0f3
+    def progress_callback(current, total):
+        total_length = 30
+        filled_length = int(round(total_length * current) / float(total))
+        percents = round(100.0 * current / float(total), 1)
+        progress_bar = '=' * filled_length + '-' * (total_length - filled_length)
+        print('\r{} {}%'.format(progress_bar, percents), flush=True, end="")
+
+    print('Uploading...')
```
panchagnula: Use Logger.warning instead

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -299,6 +316,66 @@ def enable_zip_deploy(cmd, resource_group_name, name, src, timeout=None, slot=No
...
+    blob_end = now + datetime.timedelta(weeks=520)
+    blob_token = block_blob_service.generate_blob_shared_access_signature(container_name,
+                                                                          blob_name,
+                                                                          permission=BlobPermissions(read=True),
+                                                                          expiry=blob_end,
+                                                                          start=blob_start)
+
+    blob_uri = block_blob_service.make_blob_url(container_name, blob_name, sas_token=blob_token)
+    website_run_from_setting = "WEBSITE_RUN_FROM_PACKAGE={}".format(blob_uri)
+    print('Upload to storage complete.')
```
panchagnula: Generic comment don't use print use Logger.warning

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -284,6 +300,7 @@ def enable_zip_deploy(cmd, resource_group_name, name, src, timeout=None, slot=No
     authorization = urllib3.util.make_headers(basic_auth='{0}:{1}'.format(user_name, password))
     headers = authorization
     headers['content-type'] = 'application/octet-stream'
+    headers['User-Agent'] = 'azure-cli:{}'.format(__version__)
```
panchagnula: Can you make sure this doesn't regress/ break any behavior on webapp since this is shared code?

panchagnula:
Also add a new test for functionapp ZipDeploy as well.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
ankitkumarr: I used print statements because we need to show the users information and these are not warnings. And it would be difficult to show a progress bar using logger.
Additionally, we don't need to show them any json.

panchagnula: logger.warning should work the same as print -> you don't need a JSON object as an input  to use logger.warning. You can see webapp up command which uses this for general text display

src/azure-cli/azure/cli/command_modules/appservice/custom.py
ankitkumarr: I used the `az webapp` command, and it worked fine. It just adds User agent `azure-cli:<version>` to all outgoing requests. This can help us understand what requests came from azure CLI when debugging.

ankitkumarr:
@panchagnula, I have a test locally, but for some reason it keeps failing. When I run the same exact command which makes the exact same request, it works perfectly.
I need to spend more time trying to figure out what's going on there. I can open a new issue for that test, but I would rather not block this PR on it.

btw, this is the exception I get `azure.common.AzureException: utf-8`. I looked deeper and it looks like Azure Storage fails with the error that zip file (byte array) has invalid utf-8 character. But when I try the same exact file, outside of debug context, the same byte array is created and it works fine. I will look deeper later today.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
ankitkumarr: Discussed offline with @panchagnula, I have print there to show the progress bar. To do that, I need to be able to print directly to stdout as well as reset the cursor in the terminal. I am unable to do that using `logger.warning`.

@yugangw-msft, we were wondering what's your take on it? Do you think it's ok to use `print` here given it's not an ARM call, and need a progress bar.

ankitkumarr:
@panchagnula / @yugangw-msft, I have opened this issue and assigned it to me - https://github.com/Azure/azure-cli/issues/9811 for the test.

Other than that, this PR is good on my side.

It would be good if we don't block this PR on that test. I am pretty sure I found a bug on `azure.storage`, trying to encode bytes using utf-8. I am doing the investigation and will complete the test by the next sprint.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -273,6 +277,28 @@ def update_azure_storage_account(cmd, resource_group_name, name, custom_id, stor
...
+def enable_zip_deploy_functionapp(cmd, resource_group_name, name, src, timeout=None, slot=None):
+    client = web_client_factory(cmd.cli_ctx)
+    app = client.web_apps.get(resource_group_name, name)
+    parse_plan_id = parse_resource_id(app.server_farm_id)
+    plan_info = None
+    retries = 5
+    retry_delay = 10  # seconds
+    # We need to retry getting the plan because sometimes if the plan is created as part of function app,
+    # it can take a couple of tries before it gets the plan
+    while retries > 0:
```
yugangw-msft: In Python, we usually do `for _ in range(5)` and avoid the temporary variable of `retries`

src/azure-cli/azure/cli/command_modules/appservice/custom.py
yugangw-msft: For setting agent, please check out https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/azure/cli/core/util.py#L499

src/azure-cli/azure/cli/command_modules/appservice/custom.py
yugangw-msft: I agree with @panchagnula,  `print` should be avoided as the the principle here is anything going to `stdout` should be data which can be scripted. 
Any reason you won't want to use CLI's built-in progress support like the [code here](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/azure/cli/core/extension/operations.py#L82)

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -299,6 +326,66 @@ def enable_zip_deploy(cmd, resource_group_name, name, src, timeout=None, slot=No
...
+    try:
+        if slot is not None:
+            client.web_apps.sync_function_triggers_slot(resource_group_name, name, slot)
+        else:
+            client.web_apps.sync_function_triggers(resource_group_name, name)
+    except CloudError as ce:
+        # This SDK function throws an error if Status Code is 200
+        if ce.status_code != 200:
+            raise ce
+    print('Deployment completed successfully.')
```
yugangw-msft: CLI goes with the principle that `silence is golden`. So you only need to raise the error if things went wrong; otherwise just exit silently or return json data if there is

ankitkumarr: @yugangw-msft, that's good to know. I was trying to match the upload behavior exactly with `azure-functions-core-tools`, that is currently used for uploading functions (keeping the messages identical). But, you are right, I will update the PR to keep it silent.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
ankitkumarr: I wasn't aware of the built-in progress :)
I have updated the PR to use that. With that, I wouldn't need the `print` statements. I got rid of those. Thank you both. Glad it worked out

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -284,6 +308,7 @@ def enable_zip_deploy(cmd, resource_group_name, name, src, timeout=None, slot=No
     authorization = urllib3.util.make_headers(basic_auth='{0}:{1}'.format(user_name, password))
     headers = authorization
     headers['content-type'] = 'application/octet-stream'
+    headers['User-Agent'] = 'azure-cli:{}'.format(__version__)
```
yugangw-msft: For setting agent consistently, please check out https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/azure/cli/core/util.py#L499

ankitkumarr:
@yugangw-msft, I have updated the PR. Thank you so much for the review!

PR #9785, opened 2019-06-24
Add requirements.*.txt

marstr:
This adds requirements.txt files for the complete matrix of OS and Python that we officially support.

This is the result of instantiating new virtualenv/venv on each OS, running a `pip install` of each distributable in our repository, ensuring that `pip check` came back clean, and finally running `pip freeze` redirected to the appropriate location.


.travis.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       python: 3.6
     - stage: precheck
       env: PURPOSE='Dependency Check'
-      script: ./scripts/dependency/check.sh
+      script: ./scripts/ci/dependency_check.sh
       python: 3.6
 #    - stage: verify
```
tjprescott: Shouldn't there be more edits to the CI where we install the CLI from the requirements file, such as before running tests?

marstr: Ah, there's a family of PRs coming as I update the tests and installers. This was just to get started before I unleash all of those, to make sure we are all on the same page in terms of general approach.

scripts/ci/dependency_check.sh
```diff
@@ -0,0 +1,12 @@
...
+set -ev
+
+REPO_ROOT="$(dirname ${BASH_SOURCE[0]})/../.."
+
+# Install everything from our repository first.
+find ${REPO_ROOT}/src -name setup.py -type f | xargs dirname | xargs pip3 install --no-deps
+
+pip3 install -r ${REPO_ROOT}/src/azure-cli/requirements.py3.linux.txt
+
+pip3 check
```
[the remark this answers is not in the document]
marstr: Fair question! You can see an example of failure here:
https://travis-ci.org/Azure/azure-cli/jobs/549989807#L800-L803

.travis.yml
tjprescott: So I think the existing script is more self-documenting in terms of what someone should do to fix the issue. I don't see where it calls out which requirements file is the offending one. Also, is this only being run for Python 3 Linux since that is what the CI runs?  What about the other requirements files? How will we know they need to be updated?

marstr: Mind if I address this in a follow-up? I think the points raised here are important, but I think it's more critical that I get started on overhauling the installers to use these dependency lists.

marstr: Track in #9795

PR #9796, opened 2019-06-25
Add dep diagnostics

marstr:
Looks like something didn't get caught Travis CI builds for #9785, and a previously unseen error is occurring for other PRs now.


marstr:
This change unblocks CI for other PRs.

PR #9797, opened 2019-06-25
role update: search all assigned scopes for existing role definitions

yugangw-msft:

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/role/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/role/custom.py
@@ -77,7 +77,7 @@ def _create_update_role_definition(cmd, role_definition, for_update):
         definitions_client = _auth_client_factory(cmd.cli_ctx, scope=role_resource_id).role_definitions
         scopes_in_definition = role_definition.get('assignableScopes', None)
         scopes = (scopes_in_definition if scopes_in_definition else
-                 ['/subscriptions/' + definitions_client.config.subscription_id])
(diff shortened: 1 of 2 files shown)


yugangw-msft:
@adewaleo, could you please review?

src/azure-cli/azure/cli/command_modules/role/custom.py
```diff
@@ -109,19 +109,22 @@ def delete_role_definition(cmd, name, resource_group_name=None, scope=None,
...
+def _search_role_definitions(cli_ctx, definitions_client, name, scopes, custom_role_only=False):
+    for scope in scopes:
+        roles = list(definitions_client.list(scope))
+        worker = MultiAPIAdaptor(cli_ctx)
+        if name:
+            roles = [r for r in roles if r.name == name or worker.get_role_property(r, 'role_name') == name]
+        if custom_role_only:
+            roles = [r for r in roles if worker.get_role_property(r, 'role_type') == 'CustomRole']
+        if roles:
+            return roles
```
adewaleo: Do we want to return roles and break out of the for loop immediately or search all scopes and keep updating roles?

yugangw-msft: The old behavior is to only update one, which is the main scenario which I would like to maintain.

yugangw-msft: We don't need a history entry since it is a bug fix towards a pretty edge scenario

Issue #8937, opened 2019-04-01
Proposal/Feature Request: Enable comments and full language support for ARM templates


lawrencegripper:
**Is your feature request related to a problem?**

Currently the following template will fail to deploy via the CLI but will succeed if sent directly to the `ARM` API endpoints. 

https://github.com/opencb/opencga/blob/0d4c942d71894a16aa30b498ecb2c77318449cd8/opencga-app/app/scripts/azure/arm/daemonvm/azuredeploy.jsonc

There are 2 reasons for this failure. 

1. The `JSON` contains comments 
2. The some forms of `ARM` language aren't valid `JSON` but ARE valid `ARM` templates. 

For example, the following is valid `ARM` but not valid `JSON` as the value splits over multiple lines. 

```json
"initCmd": "[concat('docker run  --name=opencga-init --mount type=bind,src=/media/primarynfs,dst=/opt/volume', 
            ' -e INIT_OPENCGA_PASS=',parameters('openCgaAdminPassword'), 
            ' -e INIT_HEALTH_CHECK_INTERVAL=',parameters('healthCheckInterval'), 
            ' -e INIT_HBASE_SSH_DNS=',parameters('hdInsightSshDns'), 
            ' -e INIT_HBASE_SSH_USER=',parameters('hdInsightSshUsername'), 
            ' -e INIT_HBASE_SSH_PASS=',parameters('hdInsightSshPassword')]"
```

When submitted to the `ARM` endpoint the content of `"[ here including multi-line ]"` is processed by the `ARM` language then re-inserted into the template. 

**Describe the solution you'd like**

In Powershell this is handled correctly and when talking to the `ARM` team they mentioned there is a desire to unify the approaches to give consistency. 

Looking at the code in the CLI I'd like to submit a PR moving the `ARM` template code to using [commentsjson](https://commentjson.readthedocs.io/en/latest/) to allow the processing of the `JSON` files correctly. 

So I'd update the [utils](https://github.com/Azure/azure-cli/blob/master/src/azure-cli-core/azure/cli/core/util.py#L212) function here to use `commentjson` (for this change to be handled in all resource not just ARM) or just update the ARM specific code. 

Then have a simple regex (sudo code) function which would strip out the `ARM` specific lines before using the `utils` loader. This would allow the `JSON` to be validated but not trip it up on `ARM` language elements. Then once the `JSON` is validated re-insert the `ARM` language chunks before sending to the `ARM API`.

```python
[...]

PR #9809, opened 2019-06-26
Handling comments and linebreaks in JSON / ARM templates

zikalino:
This is fix for: https://github.com/Azure/azure-cli/issues/8937

Please check discussion there. Basically I have just written simplest possible solution that will preprocess given json that may contain line breaks and comments.

I have tried to use **jsoncomment** but it has very obvious flaws which caused problems even with sample ARM template given in the issue I am solving. More information can be found in the issue I have created below:
https://github.com/vaidik/commentjson/issues/22

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/util.py
+++ b/src/azure-cli-core/azure/cli/core/util.py
@@ -232,29 +232,28 @@ def read_file_content(file_path, allow_binary=False):
 
 def preprocess_json(original):
     lines = original.splitlines()
-    for i in range(len(lines)):
-        line = lines[i]
-        if "#" in line or "//" in line:
-            quote = None
-            escape = None
-            for ci in range(len(line)):
-                cc = line[ci]
-                nc = line[ci + 1] if ci < (len(line) - 1) else None
-                if quote is None:
-                    if cc == "\"" or cc == "\'":
-                        quote = cc
-                    elif cc == "#" or (cc == "/" and nc == "/"):
-                        line = line[:ci]
-                        break
-                else:
-                    if escape is None:
-                        escape = None
-                    else:
-                        if cc == "\\":
-                            escape = cc
-                        elif line[ci] == quote:
-                            quote = None
-            lines[i] = line
+    for line_index, line in enumerate(lines):
+        if not ("#" in line or "//" in line):
+            continue
+        quote = None
+        escape = None
+        for i, cc in enumerate(line):
+            if quote is None and (cc == "\"" or cc == "\'"):
+                quote = cc
+                continue
+            
+            if quote is None:
+                nc = line[i + 1] if i < (len(line) - 1) else None
+                if cc == "#" or (cc == "/" and nc == "/"):
+                    line = line[:i]
+                    break
+            elif escape is not None:
+                escape = None
+            elif cc == "\\":
+                escape = cc
+            elif cc == quote:
+                quote = None
+        lines[line_index] = line
     return "".join(lines)
 
 
--- a/src/azure-cli-core/azure/cli/core/util.py
+++ b/src/azure-cli-core/azure/cli/core/util.py
@@ -235,23 +235,25 @@ def preprocess_json(original):
     for i in range(len(lines)):
         line = lines[i]
         if "#" in line or "//" in line:
-            outside = True
-            escape = False
+            quote = None
+            escape = None
             for ci in range(len(line)):
-                if outside:
-                    if line[ci] == "\"":
-                        outside = False
-                    elif line[ci] == "#" or (line[ci] == "/" and ci < (len(line) - 1) and line[ci + 1] == "/"):
+                cc = line[ci]
+                nc = line[ci + 1] if ci < (len(line) - 1) else None
+                if quote is None:
+                    if cc == "\"" or cc == "\'":
+                        quote = cc
+                    elif cc == "#" or (cc == "/" and nc == "/"):
                         line = line[:ci]
                         break
                 else:
-                    if escape:
-                        escape = False
+                    if escape is None:
+                        escape = None
                     else:
-                        if line[ci] == "\\":
-                            escape = True
-                        elif line[ci] == "\"":
-                            outside = True
+                        if cc == "\\":
+                            escape = cc
+                        elif line[ci] == quote:
+                            quote = None
             lines[i] = line
     return "".join(lines)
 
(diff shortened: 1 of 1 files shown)


yugangw-msft:
Please cross check the CI failure and verify no regression; also please add unit tests

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -230,11 +230,38 @@ def read_file_content(file_path, allow_binary=False):
...
+    for line_index, line in enumerate(lines):
+        if not ("#" in line or "//" in line):
+            continue
+        quote = None
+        escape = None
+        for i, cc in enumerate(line):
+            if quote is None and (cc == "\"" or cc == "\'"):
+                quote = cc
+                continue
+            if quote is None:
```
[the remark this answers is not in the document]
zikalino: I actually flattened it because sanity tests were complaining about too many levels or something like that.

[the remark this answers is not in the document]
zikalino: ah, yes, I know what you mean :-)

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -230,11 +230,38 @@ def read_file_content(file_path, allow_binary=False):
     raise CLIError('Failed to decode file {} - unknown decoding'.format(file_path))
 
 
+def preprocess_json(original):
+    lines = original.splitlines()
+    for line_index, line in enumerate(lines):
+        if not ("#" in line or "//" in line):
```
StephenWeatherford: What about /* */ comments?  Powershell allows them, so I assume the portal is handling them.

StephenWeatherford: Are you sure # works as a comment?  PowerShell doesn't allow them, although I'm not sure if the backend would allow them if they got past PowerShell.

zikalino: Well, I just followed the same idea as in **commentjson**.
So currently I allow both "//" and "#"
If PS doesnt' allow "#" then probably it just won't appear in any of the templates, so it's not a problem.

zikalino: i could add code to process "/*" comments as well after json is concatenated.

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -230,11 +230,38 @@ def read_file_content(file_path, allow_binary=False):
     raise CLIError('Failed to decode file {} - unknown decoding'.format(file_path))
 
 
+def preprocess_json(original):
+    lines = original.splitlines()
```
StephenWeatherford: What part of this code deals with multi-line breaks?  I.e. stuff like this:

    "multiline1": {
      "type": "string",
      "value": "this string has two
        lines"
    },
    "multiline2": {
      "type": "string",
      "value": "[a('')]  // this is not a comment, it's
        part of the first line
        and this is another line"
    }

Thx!

Do you need some unit tests?

src/azure-cli-core/azure/cli/core/util.py
StephenWeatherford: Lidong says the backend doesn't support # as a comment

src/azure-cli-core/azure/cli/core/util.py
zikalino: yes, it would be great if you could send me some tests!

Regarding multi-line breaks, perhaps I am doing it wrong. Basically I am concatenating all teh lines, but now when i see this example i assume there must be some special algorithm that also takes care of the space.
Is it described somewhere?

StephenWeatherford: Unfortunately I don't think so.  Basically their parser (Newtonsoft) doesn't consider \r or \n to end a string and simply collects those characters and continues until it finds the end of string.  

Assuming you need to preprocess before handing it to your parser, it seems like the best way to do this is probably to simply turn \n into the two characters "\" and "n", ditto for \r.  That I think should give you equivalence values, say if you need to grab default parameter values. If you ever give errors with line numbers, you'll need to adjust the line/column numbers appropriately.  Or add the \n\r after the string, so that code after the multi-line string still has the same line number.

Start by looking at the examples here: https://github.com/bmoore-msft/arm-template-language-samples

StephenWeatherford: BTW, you are passing the original, unaltered JSON to the backend, right, and the altered one is just used for parameter parsing, correct?  This is important.

src/azure-cli-core/azure/cli/core/util.py
ajf214: It does seem odd to support "#" comments, as this will be flagged as an error in vs code and will result in an error when it is sent to the ARM backend.

src/azure-cli-core/azure/cli/core/util.py
zikalino: hmm... that's getting complicated right now.
There are 2 scenarios in Azure CLI right now.

(1) ARM template is specified via URI.
In this case Azure CLI is doing only validation, and later only URI is passed.
So that's pretty simple to handle.

(2) ARM is loaded from the file
In this case Azure CLI is doing validation, but also json.loads / json.dumps.
That is most likely done to replace '' with "" which may sometimes happen.
Perhaps it handles something else as well.
json.loads can't handle comments or multiline strings.

StephenWeatherford: For #2, do you currently pass an altered version of the JSON to the backend? You can modify it in order to load the JSON for processing parameters, but is there a reason you can't pass the original file contents as a string to ARM?

cc @bmoore-msft

bmoore-msft: you'll want to pass the original, unaltered template to ARM.  We calculate a templateHash on the template and if you change it the user will never be able to map it back to their source.

yugangw-msft: Yes, CLI can supply the original template to the ARM service. 
The main issue is how many cases that Newton JSON  handles which the default Python JSON parser doesn't? Seems like we need to do a Python version of  Newton JSON? Or the handling 2 cases mentioned in #8937 are sufficient, which are the comment and multi-line strings?

yugangw-msft:
Let us close this and let @lmazuel submit a right one

PR #9812, opened 2019-06-26
Aladdin generated examples for acr, acs, cosmosdb, dls, network, resource, storage, vm

AladdinBI:


src/azure-cli/azure/cli/command_modules/acs/_help.py
```diff
@@ -437,7 +437,7 @@
 
 helps['aks remove-connector'] = """
 type: command
-short-summary: Remove the ACI Connector from a managed Kubernetes cluster.
```
tjprescott: It looks like this was a bad merge or bad conflict resolution. These "(PREVIEW)" labels should not appear.

PR #9815, opened 2019-06-26
Add vmrepair module (migrating from vm-repair extension)

swbae31:
**Added new command module vmrepair**

New commands:
1) az vm repair create : Create a new repair VM and attach the source VM's copied OS disk as a data disk.
2) az vm repair restore : Replace source VM's OS disk with data disk from repair VM.


azure-cli2017.pyproj
```diff
@@ -130,809 +130,896 @@
     <Compile Include="azure-cli-testsdk\azure\cli\testsdk\utilities.py" />
     <Compile Include="azure-cli-testsdk\azure\cli\testsdk\__init__.py" />
     <Compile Include="azure-cli-testsdk\setup.py" />
+    <Compile Include="azure-cli\azure\cli\command_modules\acr\build.py" />
```
swbae31: The azure-cli2017.pyproj file did not contain the new command module file directories so I added them. Let me know if this was intended and I'll remove the change.

src/azure-cli/azure/cli/command_modules/vmrepair/repair_utils.py
```diff
@@ -0,0 +1,185 @@
...
+def _call_az_command(command_string, run_async=False, secure_params=None):
+    """
+    Uses subprocess to run a command string. To hide sensitive parameters from logs, add the
+    parameter in secure_params. If run_async is False then function returns the stdout.
+    Raises AzCommandError if command fails.
+    """
+
+    tokenized_command = shlex.split(command_string)
+    # If command does not start with 'az' then raise exception
+    if tokenized_command[0] != 'az':
```
swbae31: Will add index out of range check.

swbae31:
> As was pointed out during the extension review, you cannot merge this into the core CLI with lines that subprocess call into other CLI commands. That logic will need to be rewritten.

Before I started the project I consulted with your team on how the vmrepair module would reuse the infrastructure already built in the CLI commands such as creating VMs, and copying disks. It would be inefficient and hard to maintain if I rewrite all the code which were all already heavily tested and used on the current CLI release.

I think that using subprocess to call other CLI commands is the best way to do what this module does. Do you have any specific reasons on why this should not be done given that the current CLI release does not give good methods to leverage existing commands?  @tjprescott @adewaleo

tjprescott:
Given that the team does not plan to accept this into the CLI without significant changes, I will close this PR.

PR #9816, opened 2019-06-27
[ACR] Helm version validation added

hkuadithya:
Issue https://github.com/Azure/acr/issues/243

**New changes have been tested and corresponding screenshots have been added below.**

**Default behavior**

**New changes - helm version < 2.11.0**

**New changes - helm version regex pattern changed; default output is shown**

**Note**: 
1. Azure-cli codebase was refactored. History file has not been modified.
2. Pylint ignore comment has been added to prevent merge failures.

@ni-reis 

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acr/check_health.py
+++ b/src/azure-cli/azure/cli/command_modules/acr/check_health.py
@@ -3,6 +3,7 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 
+import re
 from knack.util import CLIError
 from .custom import get_docker_command
 from ._docker_utils import _get_aad_token
@@ -15,6 +16,8 @@ DOCKER_PULL_SUCCEEDED = "Downloaded newer image for {}"
 DOCKER_IMAGE_UP_TO_DATE = "Image is up to date for {}"
 IMAGE = "mcr.microsoft.com/mcr/hello-world:latest"
 FAQ_MESSAGE = "\nPlease refer to https://aka.ms/acr/faq for more information."
+MIN_HELM_VERSION = "2.11.0"
+HELM_VERSION_REGEX = re.compile(r'SemVer:"v([.\d]+)"', re.I)
 
 
 # Utilities functions
@@ -136,6 +139,7 @@ def _get_cli_version():
 # Get helm versions
 def _get_helm_version(ignore_errors):
     from ._errors import HELM_VERSION_ERROR
+    from distutils.version import LooseVersion  # pylint: disable=import-error,no-name-in-module
 
     # Helm command check
     helm_command, error = get_helm_command(is_diagnostics_context=True)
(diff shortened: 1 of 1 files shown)


src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -136,6 +139,7 @@ def _get_cli_version():
 # Get helm versions
 def _get_helm_version(ignore_errors):
     from ._errors import HELM_VERSION_ERROR
+    from distutils.version import LooseVersion  # pylint: disable=import-error,no-name-in-module
```
ni-reis: Is it safe to disable "import-error"? Won't it break any customers using older versions of python?

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -153,7 +157,20 @@ def _get_helm_version(ignore_errors):
...
+
+    # Retrieve the helm version if regex pattern is found
+    match_obj = HELM_VERSION_REGEX.search(output)
+    if match_obj:
+        output = match_obj.group(1)
+
+    print("Helm version: {}".format(output))
+
+    # Display an error message if the current helm version < min required version
+    if match_obj and LooseVersion(output) < LooseVersion(MIN_HELM_VERSION):
```
ni-reis: What's the difference between `LooseVersion` and `StrictVersion` ?

src/azure-cli/azure/cli/command_modules/acr/check_health.py
hkuadithya: Looks like it's safe. It's been used in a few other places in the CLI codebase.

For instance:
https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/azure/cli/core/util.py#L139

https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/vm/_actions.py#L110

src/azure-cli/azure/cli/command_modules/acr/check_health.py
hkuadithya: From what I've been able to understand, `StrictVersion` raises exception if the software versioning is not appropriate or doesn't follow standard format. Strings such as a993.#@1.332 would raise exceptions.

If the versioning system follows industry standards, i.e.  majorVersion.minorVersion.Patch, `StrictVersion` and `LooseVersion` will have similar behavior.

In the CLI codebase, `LooseVersion` has been used throughout for comparing software versions. I didn't find any uses of `StrictVersion`. Therefore, I felt that `LooseVersion` would be more appropriate for our use case.

ni-reis: I see, makes sense. Good to know. Thanks!

src/azure-cli/azure/cli/command_modules/acr/check_health.py
ni-reis: Good to know it's been used in other places in CLI code itself. Should be safe then. Thanks!

ni-reis:
LGTM. @sajayantony can you please follow up with this PR? Thanks!

PR #9821, opened 2019-06-27
Autogrow

QingqingYuan2:
---


src/azure-cli/azure/cli/command_modules/rdbms/custom.py
```diff
@@ -18,7 +18,7 @@
 
 def _server_create(cmd, client, resource_group_name, server_name, sku_name, no_wait=False,
                    location=None, administrator_login=None, administrator_login_password=None, backup_retention=None,
-                   geo_redundant_backup=None, ssl_enforcement=None, storage_mb=None, tags=None, version=None):
+                   geo_redundant_backup=None, ssl_enforcement=None, storage_mb=None, tags=None, version=None, auto_grow='Enabled'):
```
yugangw-msft: I assume the service end has the capacity to handle "Enabled" and "Disable". Is `None` a valid value?

QingqingYuan2: None is also a valid. However, the requirement from PM is default to be "Enabled".

yugangw-msft: But with the default option of "Enabled", I don't think users will have a way to set to None though

PR #9822, opened 2019-06-27
Onboard Azure AppConfiguration

shenmuxiaosen:
Onboard Azure AppConfiguration CLI to Azure CLI core. Previously it is shipped as CLI extension. 
This pr contains both management and data plane commands. Documents are listed [here](https://docs.microsoft.com/en-us/cli/azure/ext/appconfig/appconfig?view=azure-cli-latest). 
---


src/azure-cli/azure/cli/command_modules/appconfig/_client_factory.py
```diff
@@ -0,0 +1,32 @@
...
+
+def cf_configstore(cli_ctx, *_):
+    return get_appconfig_service_client(cli_ctx).configuration_stores
+
+
+def cf_configstore_operations(cli_ctx, *_):
+    try:
+        return get_appconfig_service_client(cli_ctx).operations
+    except CLIError:
+        params = cli_ctx.data['safe_params']
```
tjprescott: You should not be using "safe_params" for any kind of logic. It is used for telemetry only.

src/azure-cli/azure/cli/command_modules/appconfig/_help.py
```diff
@@ -0,0 +1,209 @@
...
+            az appconfig credential list -g MyResourceGroup -n MyAppConfiguration
+    """
+
+helps['appconfig credential regenerate'] = """
+    type: command
+    short-summary: Regenerate an access key for an App Configuration.
+    examples:
+        - name: Regenerate a read only access key for an App Configuration
+          text:
+            az appconfig credential regenerate -g MyResourceGroup -n MyAppConfiguration --id ReadOnlyKeyId
```
tjprescott: What is the actual format for `--id`?  Is it a guid?  If so, some kind of dummy guid would be better here than "ReadOnlyKeyId"

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -0,0 +1,148 @@
...
+from ._validators import (validate_appservice_name_or_id,
+                          validate_connection_string, validate_datetime,
+                          validate_export, validate_import,
+                          validate_import_depth, validate_query_fields,
+                          validate_separator)
+
+
+def load_arguments(self, _):
+    with self.argument_context('appconfig create') as c:
+        c.argument('resource_group_name', options_list=['--resource-group', '-g'], required=True)
```
tjprescott: This should use the global alias for resource group already in the CLI.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -0,0 +1,148 @@
...
+        c.argument('name', options_list=['--name', '-n'], required=True, help='Name of the App Configuration to be created. You can configure the default name using `az configure --defaults app_configuration_store=<name>`', configured_default='app_configuration_store')
```
tjprescott: You should not need to provide `required=True`. The CLI can infer this from the command signature.  

Additionally, this should expose `id_part='name'` to enable `--ids`.
https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#supporting-the-ids-parameter

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -0,0 +1,148 @@
...
+
+    with self.argument_context('appconfig revision list') as c:
+        c.argument('key', help='If no key specified, return all keys by default. Support star sign as filters, for instance abc* means keys with abc as prefix. Similarly, *abc and *abc* are also supported.')
+        c.argument('label', help="If no label specified, list all labels. Support star sign as filters, for instance abc* means labels with abc as prefix. Similarly, *abc and *abc* are also supported.")
+        c.argument('datetime', validator=validate_datetime, help='Datetime to show history. Format: "YYYY-MM-DDThh:mm:ssZ". If no time zone specified, use UTC by default.')
+        c.argument('fields', nargs='+', help='Customize output fields.', validator=validate_query_fields, arg_type=get_enum_type(['key', 'value', 'label', 'content_type', 'etag', 'locked', 'last_modified']))
+        c.argument('top', options_list=['--top', '-t'], type=int, help='Maximum number of items to return. Default to 100.')
+        c.argument('all_', options_list=['--all'], action='store_true', help="List all revisions.")
```
tjprescott: There is a **lot** of unnecessary duplication here. You should use appropriate scopes for your argument contexts to minimize this and/or group metadata into CLICommandArgument classes to reuse definitions.

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -0,0 +1,100 @@
...
+        if not prompt_y_n(message):
+            raise CLIError('Operation cancelled.')
+    except NoTTYException:
+        raise CLIError(
+            'Unable to prompt for confirmation as no tty available. Use --yes.')
+
+
+def resolve_connection_string(cmd, config_store_name=None, connection_string=None):
+    string = ''
+    error_message = '''You may have specified both store name and connection string, which is a confliction.
```
tjprescott: Confliction isn't a word.

src/azure-cli/azure/cli/command_modules/appconfig/_validators.py
```diff
@@ -0,0 +1,124 @@
...
+
+from ._utils import is_valid_connection_string, resolve_resource_group
+from .azconfig.models import QueryFields
+
+
+def validate_tags(namespace):
+    if namespace.tags is not None:
+        if not namespace.tags:
+            return
+        tag_content = namespace.tags.split(',')
```
tjprescott: This is incorrect. The CLI does not allow comma-separated lists. You should be using the built-in tags_type and not need to duplicate or recreate any of this logic.

src/azure-cli/azure/cli/command_modules/appconfig/_validators.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def validate_separator(namespace):
+    if namespace.separator is not None:
+        if namespace.format_ == "properties":
+            raise CLIError("Separator is not needed for properties file.")
+        valid_separators = ['.', ',', ';', '-', '_', '__', '/', ':']
+        if namespace.separator not in valid_separators:
+            raise CLIError(
+                "Unsupported separator, allowed values: '.', ',', ';', '-', '_', '__', '/', ':'.")
```
tjprescott: This should just use `get_enum_type(['.', ',', ...])`, in which case all of the help text and validation will be handled automatically and consistent with the rest of the CLI.

src/azure-cli/azure/cli/command_modules/appconfig/_validators.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def validate_import(namespace):
+    source = namespace.source
+    if source == 'file':
+        if namespace.path is None:
+            raise CLIError("Please specify the file path.")
+        if namespace.format_ is None:
+            raise CLIError("Please specify the file format.")
```
tjprescott: Errors should favor usage-string format. Also, it is better to figure out ALL of the missing parameters in a parameter set and present one message. With this approach the first call can fail on the file path and then the next call on the file format. A useful error message will ensure the second call succeeds because the user will know to supply both.
Example: `usage error: --path PATH --format FORMAT`

src/azure-cli/azure/cli/command_modules/appconfig/_validators.py
```diff
@@ -0,0 +1,124 @@
...
+    elif source == 'appconfig':
+        if (namespace.src_name is None) and (namespace.src_connection_string is None):
+            raise CLIError(
+                "Please specify either the AppConfig name or its connection string.")
```
tjprescott: `usage error: --config-name NAME | --connection-string STR`

src/azure-cli/azure/cli/command_modules/appconfig/_validators.py
```diff
@@ -0,0 +1,124 @@
...
+    elif source == 'appservice':
+        if namespace.appservice_account is None:
+            raise CLIError(
+                "Please specify the name of the AppService account.")
```
tjprescott: `usage error: --appservice-account NAME_OR_ID`

src/azure-cli/azure/cli/command_modules/appconfig/_validators.py
```diff
@@ -0,0 +1,124 @@
...
+    elif destination == 'appconfig':
+        if (namespace.dest_name is None) and (namespace.dest_connection_string is None):
+            raise CLIError(
+                "Please specify either the AppConfig name or its connection string.")
+    elif destination == 'appservice':
+        if namespace.appservice_name is None:
+            raise CLIError("Please specify the name of the AppService.")
+        if namespace.appservice_resource_group is None:
+            raise CLIError(
+                "Please specify the name of the AppService's resource group.")
```
tjprescott: appservice_resource_group directly violates the CLI command guidelines:
https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#supporting-name-or-id-parameters

src/azure-cli/azure/cli/command_modules/appconfig/_validators.py
```diff
@@ -0,0 +1,124 @@
...
+            raise CLIError("Please specify the file path.")
+        if namespace.format_ is None:
+            raise CLIError("Please specify the file format.")
+    elif destination == 'appconfig':
+        if (namespace.dest_name is None) and (namespace.dest_connection_string is None):
+            raise CLIError(
+                "Please specify either the AppConfig name or its connection string.")
+    elif destination == 'appservice':
+        if namespace.appservice_name is None:
+            raise CLIError("Please specify the name of the AppService.")
```
tjprescott: Same comment on error messages.

src/azure-cli/azure/cli/command_modules/appconfig/azconfig/models.py
```diff
@@ -0,0 +1,164 @@
...
+        self.query_datetime = query_datetime
+        self.fields = fields
+        self.client_request_id = str(
+            uuid.uuid4()) if client_request_id is None else client_request_id
+        self.correlation_request_id = str(
+            uuid.uuid4()) if correlation_request_id is None else correlation_request_id
+
+
+class QueryKeyValueCollectionOptions(object):
+    '''
```
tjprescott: Is this your SDK?  It should not be embedded in the CLI...

src/azure-cli/azure/cli/command_modules/appconfig/commands.py
```diff
@@ -0,0 +1,61 @@
...
+    # Data Plane Commands
+    with self.command_group('appconfig kv', configstore_keyvalue_util) as g:
+        g.command('set', 'set_key')
+        g.command('delete', 'delete_key')
+        g.command('show', 'show_key')
+        g.command('list', 'list_key')
+        g.command('lock', 'lock_key')
+        g.command('unlock', 'unlock_key')
+        g.command('import', 'import_config')
+        g.command('export', 'export_config')
```
tjprescott: This module needs to be created in preview. Please add the following:

```
with self.command_group('appconfig', is_preview=True) as g:
  pass
```

src/azure-cli/azure/cli/command_modules/appconfig/_help.py
shenmuxiaosen: It is not a guid, but a string generated by our RP.  One example "0-l1-s0:B74joKwEhkXsOvEDA385".

src/azure-cli/azure/cli/command_modules/appconfig/commands.py
shenmuxiaosen: I think we have added in line 34 as 
```python
with self.command_group('appconfig', configstore_custom_util, is_preview=True) as g:
```

src/azure-cli/azure/cli/command_modules/appconfig/_validators.py
shenmuxiaosen: Actually this method is not used. Will remove it.

src/azure-cli/azure/cli/command_modules/appconfig/_validators.py
shenmuxiaosen: We used that before but the problem is these values are mostly punctuation. If we use get_enum_type, the help message will be very unclear like
```
Allowed values: ,, -, ., /, ;, _, __.
```
What we want is
```
Allowed values: '.', ',', ';', '-', '_', '__', '/'
```

src/azure-cli/azure/cli/command_modules/appconfig/azconfig/models.py
shenmuxiaosen: It is our internally used SDK. We don't plan to publish it and instead we will leverage Azure SDK team to ship our data plane Python SDK. And they are still working on it.

src/azure-cli/azure/cli/command_modules/appconfig/_validators.py
```diff
@@ -0,0 +1,124 @@
...
+def validate_tags(namespace):
+    if namespace.tags is not None:
+        if not namespace.tags:
+            return
+        tag_content = namespace.tags.split(',')
+        for tag_kv in tag_content:
+            tag = tag_kv.split("=")
+            if len(tag) != 2:
+                raise CLIError(
+                    "Tags are not valid. It should be this format 'key1=value1,key2=value2'")
```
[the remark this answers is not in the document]
shenmuxiaosen: This method is not used. Will delete it.

src/azure-cli/azure/cli/command_modules/appconfig/_client_factory.py
shenmuxiaosen: We are trying to bypass "az login" for some of our data plane commands by customer requests. Do you an example for how to check input arguments in _client_factory?

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
shenmuxiaosen: Done.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
shenmuxiaosen: Done.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
shenmuxiaosen: Done.

src/azure-cli/azure/cli/command_modules/appconfig/_validators.py
shenmuxiaosen: Fixed

src/azure-cli/azure/cli/command_modules/appconfig/_validators.py
shenmuxiaosen: Remove the check for resource group. Instead user can pass in name or just ID.

src/azure-cli/azure/cli/command_modules/appconfig/commands.py
tjprescott: You are right! Thanks!

src/azure-cli/azure/cli/command_modules/appconfig/_validators.py
tjprescott: That's reasonable.

src/azure-cli/azure/cli/command_modules/appconfig/_client_factory.py
tjprescott: You don't. Your data plane commands (which don't require login) should use a different client than your management plane commands. Storage would be the best example of this that I'm aware of.

src/azure-cli/azure/cli/command_modules/appconfig/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class AppconfigCommandsLoader(AzCommandsLoader):
+
+    def __init__(self, cli_ctx=None):
+        from azure.cli.core.commands import CliCommandType
+        configstore_custom = CliCommandType(
+            operations_tmpl='azure.cli.command_modules.appconfig.custom#{}')
+        super(AppconfigCommandsLoader, self).__init__(
+            cli_ctx=cli_ctx, custom_command_type=configstore_custom)
```
tjprescott: Please update this to suppress your extension:
https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#extension-suppression

shenmuxiaosen: Added

src/azure-cli/azure/cli/command_modules/appconfig/_client_factory.py
tjprescott: So, per our offline meeting, the simplest approach would be a validator on the connection string argument for your data plane commands that retrieves the connection string if it was not supplied. Then you shouldn't need two clients or any special logic in the client.

shenmuxiaosen: Another way to do this is for all data plane command, we don't require "az login" if connection string is provided. Otherwise, we will try to fetch connection string based on given resource name or ID, then throw exception if it fails.

PR #9839, opened 2019-07-01
CI to use Azdev

tjprescott:
Modifies CI scripts to utilize azdev commands (v0.1.0) in lieu of local scripts.

### DIFF
--- a/scripts/ci/precheck_pep8.sh
+++ b/scripts/ci/precheck_pep8.sh
@@ -2,12 +2,7 @@
 
 set -ev
 
-# install flake 8
-pip install -qqq flake8
+export AZDEV_CLI_REPO_PATH=$(pwd)
+export AZDEV_EXT_REPO_PATHS='_NONE_'
 
-echo "Run flake8."
-flake8 --statistics --config=./.flake8 
-
-# build packages
-. $(cd $(dirname $0); pwd)/artifacts.sh
-
+azdev style --pep8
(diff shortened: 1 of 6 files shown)


scripts/ci/precheck_pep8.sh
```diff
@@ -2,12 +2,7 @@
...
 
-# install flake 8
-pip install -qqq flake8
-
-echo "Run flake8."
-flake8 --statistics --config=./.flake8 
-
-# build packages
-. $(cd $(dirname $0); pwd)/artifacts.sh
+export AZDEV_CLI_REPO_PATH=$(pwd)
```
marstr: This line assumes that the script will always be run from the root of the repository. This is likely going to always be true, but I'd still prefer to be defensive and use pushd/popd to truly assert this.

For instance:


``` bash
#!/usr/bin/env bash
set -ev
pushd "$(dirname ${BASH_SOURCE[0]})/../.." > /dev/null 
azdev style --pep7
popd > /dev/null
```

marstr: Note, this comment is applicable a few places.

marstr: You know, given that there's so little in these scripts, you may as well skip this. When we move this over to ADO, we may consider just invoking azdev directly and deleting these scripts all-together.

tjprescott:
I'm going to close this and reopen with an upstream branch.

PR #9847, opened 2019-07-02
appservice: restore the idempotency of 'appservice plan create' and 'webapp create'

### LINKED ISSUE #9833: az appservice plan create causing error: "The server farm with name ... already exists in resource group ..." since new version
**Describe the bug**
We have been using the Azure DevOps Azure CLI task to deploy our infrastructure to Azure. In the most recent version the Azure CLI version it uses went from 2.0.61 to 2.0.67 which broke our pipelines on the 'az appservice plan create' step. It's now issueing the following error:

The server farm with name \<appservice plan name\> already exists in resource group \<resource group name\>.

The app service plan does already exist but that wasn't a problem before. It looks like it doesn't work idempotently anymore since version 2.0.67.

**To Reproduce**
Running below commands in that order cause the described error on version 2.0.67 but not on 2.0.66.

az group create --name testappservicerg
az appservice plan create --name testappservice --resource-group testappservicerg
az appservice plan create --name testappservice --resource-group testappservicerg

**Expected behavior**
No error and the app service plan gets updated with the given settings.

**Environment summary**
Installed with the [Azure DevOps Azure CLI task](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/deploy/azure-cli?view=azure-devops).

Tested locally with 2.0.66 because I still had that version installed. After upgrade to 2.0.67 it gave the described error.

**Additional context**

yugangw-msft:
Fix #9833

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/_validators.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/_validators.py
@@ -40,7 +40,7 @@ def validate_site_create(cmd, namespace):
             }
         }
         validation = client.validate(resource_group_name, validation_payload)
-        if validation.status.lower() == "failure":
+        if validation.status.lower() == "failure" and validation.error.code != 'SiteAlreadyExists':
             raise CLIError(validation.error.message)
 
 
@@ -69,5 +69,5 @@ def validate_asp_create(cmd, namespace):
             }
         }
         validation = client.validate(resource_group_name, validation_payload)
-        if validation.status.lower() == "failure":
+        if validation.status.lower() == "failure" and validation.error.code != 'ServerFarmAlreadyExists':
             raise CLIError(validation.error.message)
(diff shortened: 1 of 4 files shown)


src/azure-cli/azure/cli/command_modules/appservice/_validators.py
```diff
@@ -40,7 +40,7 @@ def validate_site_create(cmd, namespace):
             }
         }
         validation = client.validate(resource_group_name, validation_payload)
-        if validation.status.lower() == "failure":
+        if validation.status.lower() == "failure" and validation.error.code != 'SiteAlreadyExists':
```
panchagnula: Do we need case insensitive check here just to be sure?

panchagnula: How does this help - when someone actually trying to create a new ASP or APP with the same name accidentally? If it doesn't then we should just remove the validation check in these commands.

yugangw-msft: The code will just proceed to create the ASP/WebApp on name conflicts, which is the old behavior. I am keeping the validation logic here as I assume it does other kinds of validations which are still beneficial? However if the naming conflicts is the only thing the validate API checks, then I agree we should remove the validation check completely.

panchagnula: The API does do more than just name validation - keep it for now, & i will update az webapp up as needed.

yugangw-msft: I am merging. For error code change, case sensitive check is fine, as the code is supposed to be treated more prudently and less randomness on case change

PR #9875, opened 2019-07-05
Sql vm minor changes + new property in sql vm create

yareyes:
This PR contains small changes as: 

Adding a new property --sql-mgmt-type, not required with a default value. 
Modifying tests accordingly.
Removing a deprecated parameter.
Small update in sql vm group that did not allow customers to set the new storage account key in update.

---


src/azure-cli/azure/cli/command_modules/sqlvm/custom.py
```diff
@@ -171,14 +171,14 @@ def aglistener_update(instance, sql_virtual_machine_instances=None):
 
 # pylint: disable=too-many-arguments, too-many-locals, line-too-long, too-many-boolean-expressions
 def sqlvm_create(client, cmd, sql_virtual_machine_name, resource_group_name, sql_server_license_type,
-                 location=None, sql_image_sku=None, enable_auto_patching=None,
+                 location=None, sql_image_sku=None, enable_auto_patching=None, sql_management_mode="LightWeight",
```
SeliuMSFT: > sql_management_mode="LightWeight" [](start = 78, length = 34)

this is the by default value for create and update?

src/azure-cli/azure/cli/command_modules/sqlvm/custom.py
```diff
@@ -270,7 +272,7 @@ def sqlvm_create(client, cmd, sql_virtual_machine_name, resource_group_name, sql
 def sqlvm_update(instance, sql_server_license_type=None, sql_image_sku=None, enable_auto_patching=None,
                  day_of_week=None, maintenance_window_starting_hour=None, maintenance_window_duration=None,
                  enable_auto_backup=None, enable_encryption=False, retention_period=None, storage_account_url=None,
-                 storage_access_key=None, backup_password=None, backup_system_dbs=False, backup_schedule_type=None,
+                 storage_access_key=None, backup_password=None, backup_system_dbs=False, backup_schedule_type=None, sql_management_mode=None,
```
SeliuMSFT: >sql_management_mode=None [](start = 116, length = 24)

this means null for update?

src/azure-cli/azure/cli/command_modules/sqlvm/_help.py
```diff
@@ -67,19 +67,22 @@
         - name: Create a SQL virtual machine with specific sku type and license type.
           text: >
             az sql vm create -n sqlvm -g myresourcegroup -l eastus --image-sku Enterprise --license-type AHUB
+        - name: Create a SQL virtual machine with NoAgent type, only valid for EOS SQL 2008 and SQL 2008 R2.
+          text: >
+            az sql vm create -n sqlvm -g myresourcegroup -l eastus --license-type AHUB --sql-mgmt-type NoAgent --image-sku Enterprise --image-offer SQL2008-WS2008R2
         - name: Enable R services in SQL2016 onwards.
           text: >
-            az sql vm create -n sqlvm -g myresourcegroup -l eastus --license-type PAYG --enable-r-services true
+            az sql vm create -n sqlvm -g myresourcegroup -l eastus --license-type PAYG --sql-mgmt-type Full --enable-r-services true
```
SeliuMSFT: >--sql-mgmt-type Full - [](start = 87, length = 22)

I am thinking to change those text as LightWeight as well, since by default for create we are taking LightWeight

src/azure-cli/azure/cli/command_modules/sqlvm/_validators.py
```diff
@@ -97,3 +97,14 @@ def validate_subnet(cmd, namespace):
...
+
+# pylint: disable=too-many-statements,line-too-long
+def validate_sqlmanagement(namespace):
+    '''
+    Validates if sql management mode provided, the offer type and sku type has to be provided.
+    '''
+    sql_mgmt_mode = namespace.sql_management_mode
+
+    if (sql_mgmt_mode == "NoAgent" and (namespace.sql_image_sku is None or namespace.sql_image_offer is None)):
+        raise CLIError("incorrect usage: --sql-mgmt-type NoAgent --image-sku NAME --image-offer NAME")
```
SeliuMSFT: Can we explicitly mention that for NoAgent, the only available Image is SQL2008-WS2008 and SQL2008R2-WS2008?

yareyes: We do it, in the examples itself (_help.py). This validation is just for usage and should not have any examples.

src/azure-cli/azure/cli/command_modules/sqlvm/custom.py
yareyes: No, this means for update customer has the option to only pass Full (meaning they want to update from LightWeight or NoAgent -> Full). We will not be changing the current value that we get from the get call. If customer provides no value, we pass the same value the entity returns.

src/azure-cli/azure/cli/command_modules/sqlvm/custom.py
yareyes: Yes, we're setting LightWeight as default value for create. Are we changing that?

src/azure-cli/azure/cli/command_modules/sqlvm/_help.py
yareyes: This are just specific create options if customer wants to create + enable sql manageability. Do you think for the basic commands customer needs the LightWeight default set?

yareyes:
@tjprescott could you please help me review this PR? 
Also, the CLI Linter test is failing: 

-  FAIL: faulty_help_type_rule |  
-- | --
  | Help-Entry: `batch pool supported-images list` - Command should be of help-type `command`

But this is not related to my changes. 

Appreciate the help, we want to merge soon :)

src/azure-cli/azure/cli/command_modules/sqlvm/_params.py
```diff
@@ -168,6 +175,9 @@ def load_arguments(self, _):
                    options_list=['--image-sku'],
                    help='SQL image sku.',
                    arg_type=get_enum_type(SqlImageSku))
+        c.argument('sql_image_offer',
+                   options_list=['--image-offer'],
+                   help='SQL image offer. Examples include SQL2008R2-WS2008R2, SQL2008R2-WS2008, SQL2008-WS2008, SQL2008-WS2008R2.')
```
tjprescott: Where do these values come from?  It seems like this might need to use the `populator_commands` kwarg in `help.py`.

src/azure-cli/azure/cli/command_modules/sqlvm/_validators.py
tjprescott: Please change to "usage error: ..." to be consistent with how argparse shows usage errors.

src/azure-cli/HISTORY.rst
```diff
@@ -88,6 +88,12 @@ Release History
 
 * Document allowed values for sql db create --sample-name
 
+**SQL VM**
+
+* New optional paramenter to setup SQL management type on creation time of a SQL vm. --sql-mgmt-type
+* New optional parameter to upgrade SQL manageability on update of a SQL vm.
+* Minor fixes on SQL vm group that did not allow to update the key for storage accounts.
```
tjprescott: Ideally make this more specific:
```
* `vm sql create/update`: Added optional parameter `--sql-mgmt-type` to setup SQL management.
```

src/azure-cli/azure/cli/command_modules/sqlvm/_params.py
yareyes: These were just examples of possible combinations. I reduced the amount of examples just to give an idea of what the pattern should look like.

src/azure-cli/azure/cli/command_modules/sqlvm/_validators.py
yareyes: Done

src/azure-cli/HISTORY.rst
yareyes: Fixed

src/azure-cli/azure/cli/command_modules/sqlvm/_params.py
tjprescott: Is it purely pattern based? Is there an endpoint that provides actual supported values?

yareyes: No there is no endpoint that provides the supported values unfortunately. This parameter (image-offer) is only required if customer provides sql-mgmt-type as NoAgent, which is a minimal portion of customers (what we call End of Support Customers that have WS 2008).

PR #9876, opened 2019-07-06
Updated managedservices module to support api-version 2019-06-01

rajivnandivada:
---


rajivnandivada:
@tjprescott @limingu @yugangw-msft could you please review this?

src/azure-cli/azure/cli/command_modules/managedservices/custom.py
```diff
@@ -50,11 +44,11 @@ def cli_definition_create(cmd, client,
         Plan=plan,
     )
 
-    return client.create_or_update(
+    poller = client.create_or_update(
```
yugangw-msft: any reason you don't return `poller` here? CLI core has the common logic to detect the case that if it is a poller and will wait to get the result

rajivnandivada: I didnt know that. I have updated all such references to just return the poller.

panchagnula:
BTW update history as well - this file has moved here now ~src/azure-cli/HISTORY.rst

PR #9900, opened 2019-07-10
Relay: Fixed issue #8775 : 

v-Ajnava:
Fixed issue #8775 : Cannot create hybrid connection with disabled client authroization

---


src/azure-cli/HISTORY.rst
```diff
@@ -3,6 +3,11 @@
 Release History
 ===============
 
+**Relay**
```
tjprescott: This needs to be alphabetical in the list.

src/azure-cli/azure/cli/command_modules/relay/_params.py
```diff
@@ -71,13 +71,15 @@ def load_arguments_sb(self, _):
...
         c.argument('requires_client_authorization', arg_type=get_three_state_flag(), help='Indicates whether client authorization is required')
+        c.argument('requires_transport_security', arg_type=get_three_state_flag(), help='Indicates whether transport security is required')
         c.argument('user_metadata', help='Endpoint metadata')
 
     with self.argument_context('relay wcfrelay update') as c:
         c.argument('relay_name', arg_type=name_type, id_part='child_name_1', help='Name of WCF Relay')
         c.argument('status', arg_type=get_enum_type(['Active', 'Disabled', 'SendDisabled', 'ReceiveDisabled']), help='Enumerates the possible values for the status of a messaging entity.')
         c.argument('relay_type', arg_type=get_enum_type(['Http', 'NetTcp']), default='NetTcp', help='Relay type')
         c.argument('requires_client_authorization', arg_type=get_three_state_flag(), help='Indicates whether client authorization is required')
+        c.argument('requires_transport_security', arg_type=get_three_state_flag(), help='Indicates whether transport security is required')
```
tjprescott: You don't need this duplication. This should just be registered under a scope of 'relay wcfrelay'.

v-Ajnava: Let me check with PM, as 'requires_client_authorization' and 'requires_transport_security' can be set only at create wcfrelay and can not be changed. I think its should be removed from update command.

src/azure-cli/HISTORY.rst
v-Ajnava: sur, will make change

PR #9928, opened 2019-07-15
[storage] Add 'storage remove' command

limingu:
---


src/azure-cli/azure/cli/command_modules/storage/_params.py
```diff
@@ -742,6 +743,24 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals, too-many-statem
         c.argument('message_id', options_list='--id')
         c.argument('content', type=unicode_string, help='Message content, up to 64KB in size.')
 
+    with self.argument_context('storage remove') as c:
+        from .completers import file_path_completer
+
+        c.extra('container_name', container_name_type)
+        c.extra('blob_name', options_list=('--name', '-n'), arg_type=blob_name_type,
+                validator=validate_azcopy_remove_url)
```
Juliehzl: Why do you only validate azcopy remove url for blob_name? What about share name?

limingu: Updated with validate_azcopy_remove_arguments. It will validate all the arguments.

src/azure-cli/azure/cli/command_modules/storage/operations/azcopy.py
```diff
@@ -14,21 +14,18 @@ def storage_blob_copy(azcopy, source, destination, recursive=None):
...
-def storage_blob_remove(cmd, client, target, recursive=None):
-    azcopy = _azcopy_blob_client(cmd, client)
+def storage_remove(cmd, client, service, target, exclude=None, include=None, recursive=None):
+    if service == 'file':
+        azcopy = _azcopy_file_client(cmd, client)
+    else:
+        azcopy = _azcopy_blob_client(cmd, client)
     flags = []
     if recursive is not None:
         flags.append('--recursive')
```
Juliehzl: If you just append --recursive in flags, can it work? In azcopy, it should be --recursive=true.

src/azure-cli/azure/cli/command_modules/storage/tests/latest/test_storage_azcopy_scenarios.py
```diff
@@ -10,6 +10,8 @@
 
 
 class StorageAzcopyTests(StorageScenarioMixin, LiveScenarioTest):
```
Juliehzl: Do we need to write test_scenario for hybrid_2018_03_01 or hybrid_2019_03_01?

src/azure-cli/azure/cli/command_modules/storage/tests/latest/test_storage_azcopy_scenarios.py
```diff
@@ -72,3 +74,119 @@ def test_storage_blob_azcopy_sync(self, resource_group, storage_account_info, te
             test_dir, container, storage_account))
         self.cmd('storage blob list -c {} --account-name {}'.format(
             container, storage_account), checks=JMESPathCheck('length(@)', 0))
+
+    @ResourceGroupPreparer()
+    @StorageAccountPreparer()
+    @StorageTestFilesPreparer()
+    def test_storage_blob_azcopy_remove(self, resource_group, storage_account_info, test_dir):
```
Juliehzl: I think test_storage_remove() may be better because not only blob can be removed.

src/azure-cli/azure/cli/command_modules/storage/tests/latest/test_storage_azcopy_scenarios.py
```diff
@@ -72,3 +74,119 @@ def test_storage_blob_azcopy_sync(self, resource_group, storage_account_info, te
...
+        storage_account, _ = storage_account_info
+        container = self.create_container(storage_account_info)
+
+        # sync directory
+        self.cmd('storage blob sync -s "{}" -c {} --account-name {}'.format(
```
Juliehzl: Why do you run blob sync first?

src/azure-cli/azure/cli/command_modules/storage/operations/azcopy.py
limingu: '--recursive' equals '--recursive=true'.

src/azure-cli/azure/cli/command_modules/storage/tests/latest/test_storage_azcopy_scenarios.py
limingu: This command only supports the latest profile. No plan to support other profiles.

src/azure-cli/azure/cli/command_modules/storage/tests/latest/test_storage_azcopy_scenarios.py
limingu: test_storage_blob_azcopy_remove is focusing on blob remove. There is another test test_storage_file_azcopy_remove to cover file remove.

src/azure-cli/azure/cli/command_modules/storage/tests/latest/test_storage_azcopy_scenarios.py
limingu: To upload the blobs first. So they can be removed.

src/azure-cli/azure/cli/command_modules/storage/tests/latest/test_storage_azcopy_scenarios.py
```diff
@@ -72,3 +74,119 @@ def test_storage_blob_azcopy_sync(self, resource_group, storage_account_info, te
...
+        account_info = self.get_account_info(resource_group, storage_account)
+        s1 = self.create_share(account_info)
+        s2 = self.create_share(account_info)
+        d1 = 'dir1'
+        d2 = 'dir2'
+
+        self.storage_cmd('storage directory create --share-name {} -n {}', account_info, s1, d1)
+        self.storage_cmd('storage directory create --share-name {} -n {}', account_info, s2, d2)
+
+        local_file = self.create_temp_file(512, full_random=False)
```
Juliehzl: Why not use StorageTestFilesPreparer() here?

src/azure-cli/azure/cli/command_modules/storage/tests/latest/test_storage_azcopy_scenarios.py
```diff
@@ -72,3 +74,119 @@ def test_storage_blob_azcopy_sync(self, resource_group, storage_account_info, te
...
+        container = self.create_container(storage_account_info)
+
+        # sync directory
+        self.cmd('storage blob sync -s "{}" -c {} --account-name {}'.format(
+            test_dir, container, storage_account))
+        self.cmd('storage blob list -c {} --account-name {}'.format(
+            container, storage_account), checks=JMESPathCheck('length(@)', 41))
+
+        self.cmd('storage remove -c {} -n readme --account-name {}'.format(
+            container, storage_account))
```
Juliehzl: It looks that you do not provide a check for this command. In this way, how do you know test results?

limingu: The test checked the list of files after remove command. The total files became 40 from 41 after removing one file. See the next line for the check.

PR #9930, opened 2019-07-15
[storage] Fix 'storage blob update' content settings bug

### LINKED ISSUE #9888: `az storage blob update` incorrectly autofills parameters

**Describe the bug**
When calling `az storage blob update`, the cli makes 2 calls - one call to get the blob's information (if exists), and then makes the call to update the blob's system properties. However, in this process, the cli opts to intelligently autofill any content setting parameters that weren't explicitly set by the user. This means that users will never be able to `reset` content setting parameters through azure cli, and old values will incorrectly persist. 
I have a blob with content settings like:
```
    "contentSettings": {
      "cacheControl": "testcachecontrol",
      "contentDisposition": "testdisposition",
      "contentEncoding": "testencoding",
      "contentLanguage": "testlanguage",
      "contentMd5": "testing7",
      "contentType": "testtype"
    },
```
and I make an `az storage blob update` call with  `--content-type newtype`
the resulting contentSettings is incorrectly as follows:
```
    "contentSettings": {
      "cacheControl": "testcachecontrol",
      "contentDisposition": "testdisposition",
      "contentEncoding": "testencoding",
      "contentLanguage": "testlanguage",
      "contentMd5": "testing7",
      "contentType": "newtype"
    },
```
It's expected that all other content settings should be wiped and set to null/undefined/hidden.

**To Reproduce**
- set all the content settings parameters on a blob, or make note of the pre-existing values
- using `az storage blob update`, leave at least one property unspecified. Any unspecified property should be expected to have `null` value or to not show up in `contentSettings`.
- `az storage blob show` the blob, then compare `contentSettings` to the old values.

**Expected behavior**
https://docs.microsoft.com/en-us/rest/api/storageservices/set-blob-properties
Specifically:
```
If any one or more of the following properties is set in the request, then all of these properties are set together. If a value is not provided for a given property when at least one of the properties listed below is set, then that property will be cleared for the blob.
    x-ms-blob-cache-control
    x-ms-blob-content-type
    x-ms-blob-content-md5
    x-ms-blob-content-encoding
    x-ms-blob-content-language
    x-ms-blob-content-disposition
```
Example:
My pre-existing blob shows the content settings:
```
    "contentSettings": {
      "cacheControl": "testcachecontrol",
      "contentDisposition": "testdisposition",
      "contentEncoding": "testencoding",
      "contentLanguage": "testlanguage",
      "contentMd5": "testing7",
      "contentType": "testtype"
    },
```
I make the call `az storage blob update` with only `--contentLanguage newlanguage`
I get the blob again, and the contentSettings should be:
```
    "contentSettings": {
      "cacheControl": null,
      "contentDisposition": null,
      "contentEncoding": null,
      "contentLanguage": "newlanguage",
      "contentMd5": null,
      "contentType": null
    },
```

**Environment summary**
Install Method: `brew update && brew install azure-cli` / OS version: macOS Mojave version 10.14.3 / cli version: 2.0.66 / iTerm

**Additional context**
node client behaves as expected. if this is expected azure cli behavior please direct me to the appropriate documentation.

limingu:
Fix #9888
---


yugangw-msft:
@Juliehzl, please review. If all fine, please approve and merge.

src/azure-cli/azure/cli/command_modules/storage/_validators.py
```diff
@@ -337,11 +337,7 @@ def _class_name(class_type):
         return class_type.__module__ + "." + class_type.__class__.__name__
 
     def validator(cmd, namespace):
-        t_base_blob_service, t_file_service, t_blob_content_settings, t_file_content_settings = cmd.get_models(
-            'blob.baseblobservice#BaseBlobService',
-            'file#FileService',
-            'blob.models#ContentSettings',
-            'file.models#ContentSettings')
+        t_file_content_settings = cmd.get_models('file.models#ContentSettings')
```
Juliehzl: It looks that t_file_content_setting is not used in this file.

limingu: It is used in line 345. You may need to expand the lines.

PR #9939, opened 2019-07-16
[ACR] az acr check-health fixes - cli version, ignore-error description and faq message

hkuadithya:
1.  CLI version is correctly displayed now. ("not found" was shown before)

2.  Enhancement - searching component name in `working_set` `dictionary` instead of `list` for improved performance.

      * For reviewer's reference: https://github.com/pypa/setuptools/blob/master/pkg_resources/__init__.py#L560

3.  Updated URL in FAQ message to https://aka.ms/acr/health-check

4.  `--ignore-errors` description message has been updated

NOTE: Issue on GitHub can be raised and linked to this PR if required.

@ni-reis @sajayantony 

---
### Logger info, error and warning messages shown
Capture

### Git bash - cli component not found

---


src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -123,14 +123,14 @@ def _get_docker_status_and_version(ignore_errors, yes):
...
-        if component.key == acr_component_name:
-            acr_cli_version = component.version
 
-    print('ACR CLI version: {}'.format(acr_cli_version))
+    # working_set.by_key is a dictionary with component names as key
+    cli_component_name = "azure-cli"
+    cli_version = working_set.by_key[cli_component_name].version \
+        if cli_component_name in working_set.by_key else "not found"
+
+    print('ACR CLI version: {}'.format(cli_version))
```
tjprescott: This needs to print to stderr. 
https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md#general-patterns

hkuadithya: @tjprescott Thanks for pointing that out. New changes have been pushed.

ni-reis: Hi @tjprescott , this command is used by customers to get general information on the environment being used and connectivity to a registry. The command output consists of general info of the environment being used, including this, that's why we are using print (here and in other pieces of the command as well)

hkuadithya: @tjprescott @ni-reis 

CI build scripts seem to be failing while directly using `print('.....', file=sys.stderr)` due to incompatibility with Py 2.7. 

Apparently, an import statement `from __future__ import print_function` needs to be added to make it compatible with both Py2.7 and Py3+.  For instance, this file [uses the same fix here](https://github.com/Azure/azure-cli/blob/dev/scripts/dev_setup.py#L8).

As Nicholas mentioned before, maybe we should reconsider printing to `stderr` in the first place since this is just a general information being shown to the users. Any suggestions?

tjprescott: The fact that it is printed for informational purposes only is exactly why it should be printed to stderr. Anything printed to stdout is expected to be something one might pipe to another command. Informational messages don't fall into that category.

ni-reis: I see, thanks for clarifying! 
From our side, we want to make sure that the default is that this info will show in the command line (similar to a command output), i.e. we don't want this to be printed to another file that the user would have to check to see the output. Will we get this behavior printing to stderr?

tjprescott: The info will display in the command prompt in white font. But if the user redirects stdout to file or pipes stdout to another command, it will not pass those info message along to that.

ni-reis: I see. Thanks!
@hkuadithya , can you please make those changes and test then?

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -123,14 +124,15 @@ def _get_docker_status_and_version(ignore_errors, yes):
...
-            acr_cli_version = component.version
 
-    print('ACR CLI version: {}'.format(acr_cli_version))
+    # working_set.by_key is a dictionary with component names as key
+    cli_component_name = "azure-cli"
+    if cli_component_name in working_set.by_key:
+        print('ACR CLI version: {}'.format(
+            working_set.by_key[cli_component_name].version))
+    else:
+        print('ACR CLI version: not found', file=sys.stderr)
```
ni-reis: nit: can you change this to "Azure CLI version" or something similar, since it's no longer acr cli?

hkuadithya: @ni-reis Yes, I have pushed those changes now.

PR #9940, opened 2019-07-16
Update `prompt-toolkit` to 2.0.9

elibarzilay:
We ran into a problem when trying to install `azure-cli` on top of a conda environment.  The problem is that `azure-cli` specifies an old version of `prompt-toolkit`, and installing it forces a downgrade and that breaks the jupyter kernel.

This PR updates the version to the one that is current in conda.

### DIFF
--- a/src/azure-cli/setup.py
+++ b/src/azure-cli/setup.py
@@ -132,7 +132,7 @@ DEPENDENCIES = [
     'knack~=0.6,>=0.6.3',
     'mock~=2.0',
     'paramiko>=2.0.8,<3.0.0',
-    'prompt_toolkit~=1.0,>=1.0.15',
+    'prompt_toolkit~=2.0',
     'pydocumentdb>=2.0.1,<3.0.0',
     'pygments~=2.4',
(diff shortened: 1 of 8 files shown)


marstr:
As we talked about offline, I'm happy with the principle! Looks like you're still at the beginning of the journey here.

src/azure-cli/setup.py
```diff
@@ -132,7 +132,7 @@
     'knack~=0.6,>=0.6.3',
     'mock~=2.0',
     'paramiko>=2.0.8,<3.0.0',
-    'prompt_toolkit~=1.0,>=1.0.15',
+    'prompt_toolkit>=1.0.15',
```
marstr: Unless we're limiting ourselves to surface area available in both major versions 1 & 2, we should use `prompt_toolkit~=2.0`.

elibarzilay: I'll change it now.  FWIW, my reasoning in keeping just the `>=` constraint is that it looks like the code doesn't use anything specific for v2, but I might be wrong.

PR #9946, opened 2019-07-16
Add docker to image

ujovlado:
Changes:

- Add `docker` to `azure-cli` image

---

It can be really useful to have `docker` command available in `azure-cli` image.

If `docker` command is available in image, you can simply mount docker socket (e.g. `-v /var/run/docker.sock:/var/run/docker.sock`) and the run docker commands from container.

For example, it will allow you to push images from `azure-cli` container and there is no need to have `azure-cli` installed locally, you just pull the image.

Real world usage, CI platform:

- `docker build -t my-image .`
- `docker pull mcr.microsoft.com/azure-cli`
- `docker run --rm -v /var/run/docker.sock:/var/run/docker.sock mcr.microsoft.com/azure-cli sh -c 'az login ... && az acr login ... && docker push ...'` (dots mean there are more arguments in mentioned commands)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,9 +34,11 @@ LABEL maintainer="Microsoft" \
 # pip wheel - required for CLI packaging
 # jmespath-terminal - we include jpterm as a useful tool
 # libintl and icu-libs - required by azure devops artifact (az extension add --name azure-devops)
+# docker - include as a useful tool (with mounted docker socket, docker commands can be executed from a container)
 RUN apk add --no-cache bash openssh ca-certificates jq curl openssl git zip \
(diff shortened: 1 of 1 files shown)


marstr:
In general, I think this image is a bit bloated, so I'm not excited to add more to it. I can see why having the tools to interact with docker would be useful though, especially when working with ACR. However, for now I think it's best that people add it as a layer to their own images, i.e.

``` Dockerfile
FROM mcr.microsoft.com/azure-cli
RUN apk add --no-cache docker 
```

marstr:
Thanks for this though! I love getting PRs instead of just suggestion issues, many kudos @ujovlado.

ujovlado:
Thanks for explanation. We already have custom build, so this was only next step.

I understand. It also affects image size (about 100MB increase).

Issue #9861, opened 2019-07-04
nic is not added to a lb backend pool if the nic is referenced by the id


tspascoal:
## Describe the bug

**Command Name**
`az network nic ip-config address-pool add`

**Errors:**
```
The command failed with an unexpected error. Here is the traceback:


'child_name_1'
Traceback (most recent call last):
  File "C:\Users\VSSADM~1\AppData\Local\Temp\pip-install-dsvrkf2x\knack\knack\cli.py", line 206, in invoke
  File "C:\Users\VSSADM~1\AppData\Local\Temp\pip-install-dsvrkf2x\azure-cli-core\azure\cli\core\commands\__init__.py", line 508, in execute
  File "C:\Users\VSSADM~1\AppData\Local\Temp\pip-install-dsvrkf2x\knack\knack\cli.py", line 162, in raise_event
  File "C:\Users\VSSADM~1\AppData\Local\Temp\pip-install-dsvrkf2x\azure-cli-core\azure\cli\core\commands\arm.py", line 310, in parse_ids_arguments
KeyError: 'child_name_1'
```

## To Reproduce:
Steps to reproduce the behavior. Note that argument values have been redacted, as they may contain sensitive information.

- _Put any pre-requisite steps here..._
- `az network nic ip-config address-pool add --ids {} --address-pool {}`

The problem seems to lie in the use of id, because if --nic-name  and --name are specified the command works as expect. 

The id is correct, the problem not a malformed id, it was extracted using 

## Expected Behavior

the referenced NIC on ids parameter is added to the LB pool

## Environment Summary
```
Windows-10-10.0.18362-SP0
Python 3.6.6
Shell: cmd.exe

azure-cli 2.0.65 *

Extensions:
azure-devops 0.10.0
resource-graph 0.1.8

```

tjprescott:
The real issue here is that `--ids` should not even be exposed. You are expected to provide the name components.

PR #9957, opened 2019-07-17
Network.NIC: Remove --ids from the parameters

MyronFanQiu:
---
Fixes https://github.com/Azure/azure-cli/issues/9861

### DIFF
--- a/src/azure-cli/HISTORY.rst
+++ b/src/azure-cli/HISTORY.rst
@@ -3,13 +3,13 @@
 Release History
 ===============
 
-**Network**
-
-* Fixed issue #9861. Remove `--ids` from the parameters.
-
 **ACR**
 
 * Fixed issue #9952 (a regression in the `acr pack build` command).
+
+**Network**
+
+* network nic ip-config add: Fixes #9861 where --ids was inadvertently exposed but did not work.
 
 **RBAC**
 
--- a/src/azure-cli/HISTORY.rst
+++ b/src/azure-cli/HISTORY.rst
@@ -3,6 +3,10 @@
 Release History
 ===============
 
+**Network**
+
+* Fixed issue #9861. Remove `--ids` from the parameters.
+
 **ACR**
 
 * Fixed issue #9952 (a regression in the `acr pack build` command).
--- a/src/azure-cli/azure/cli/command_modules/network/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/network/_params.py
@@ -765,7 +765,7 @@ def load_arguments(self, _):
 
     with self.argument_context('network nic ip-config address-pool') as c:
         c.argument('load_balancer_name', options_list='--lb-name', help='The name of the load balancer containing the address pool (Omit if suppying an address pool ID).', completer=get_resource_name_completion_list('Microsoft.Network/loadBalancers'))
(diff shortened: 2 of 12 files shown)


MyronFanQiu:
@tjprescott I'm not sure whether this is a right way to fix this issue. In my understanding, should we do some fix in the azure.cli.core so that the `--ids` will not show up?

MyronFanQiu:
Besides targeting at this issue, please help me check, for a common scenario that we need to do some fix for an existed command, what do I miss? In this PR, I change the logic, add test and new recording, update the history.rst. Is there something I also need to do?

src/azure-cli/HISTORY.rst
```diff
@@ -3,6 +3,10 @@
 Release History
 ===============
 
+**Network**
+
+* Fixed issue #9861. Remove `--ids` from the parameters.
```
tjprescott: The network section should be located alphabetically. Also, the typical format would be:

* `network nic ip-config add`: Fixed issue where `--ids` was inadvertently exposed but did not work. 

Also, were it not for the fact that `--ids` caused a crash, we could not simply remove this. We would need to deprecate it using: https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#deprecating-commands-and-arguments

tjprescott:
> By prefacing your commits with “Fixes”, “Fixed”, “Fix”, “Closes”, “Closed”, or “Close” when the commit is merged into master, it will also automatically close the issue.
> 
> References make it possible to deeply connect the work being done with the bug being tracked, and are a great way to add visibility into the history of your project.

From: https://guides.github.com/features/issues/#notifications

In short, use "Fixes #9861" instead of "Fixed issue #9861" to get the auto-close functionality.

src/azure-cli/HISTORY.rst
MyronFanQiu: Actually, I don't understand what is the `--ids`. Is it by design to show the `--ids` when you add `id_parts=xxx` into the `c.arguments`? What do we need as the value for this field?
I will fix other comments, thanks very much!~

tjprescott: It is discussed here:
https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#supporting-the-ids-parameter

MyronFanQiu: Yes. I checked that doc before.  Maybe I misunderstand something, please correct me. To be simple, what should we used as an input for the `--ids` in the example below? `/subscription/xxxx-xxxx-xxxx/resoureceGroup/myResourceGroup/Microsoft.NIC/myNIC` or `/subscription/xxxx-xxxx-xxxx/resoureceGroup/myResourceGroup/Microsoft.NIC/myNIC/publicIP/mypublicIP` or both of them?
```
Resource Id Arguments
    --ids              : One or more resource IDs. If provided, no other 'Resource Id' arguments
                         should be specified.
    --name -n          : The IP config name.
    --nic-name         : The NIC name.
    --resource-group -g: Name of resource group.
```

tjprescott: Neither, but that is why `--ids` was being removed from the command. `--ids` should always refer to whatever the argument group's central item is.  In this case it is questionable or confusing what `--ids` should refer to, which is why it shouldn't be exposed.

PR #9985, opened 2019-07-19
webapp: Adding validation to az webapp config ssl

panchagnula:
---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/_validators.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/_validators.py
@@ -73,7 +73,7 @@ def validate_asp_create(cmd, namespace):
             raise CLIError(validation.error.message)
 
 
-def validate_app_exists_in_rg(cmd, namespace):
+def validate_app_or_slot_exists_in_rg(cmd, namespace):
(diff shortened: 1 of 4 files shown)


src/azure-cli/azure/cli/command_modules/appservice/_validators.py
```diff
@@ -85,3 +85,13 @@ def validate_app_exists_in_rg(cmd, namespace):
             app = client.web_apps.get(resource_group_name, webapp, None, raw=True)
         if app.response.status_code != 200:
             raise CLIError(app.response.text)
+
+
+def validate_app_exists_in_rg(cmd, namespace):
+    client = web_client_factory(cmd.cli_ctx)
+    if isinstance(namespace.name, str) and isinstance(namespace.resource_group_name, str):
```
yugangw-msft: This is unnecessary, as the command line parser should have taken care of this already. I know you copied and pasted from the old code, but I don't think you need it here, and probably it is unnecessary in the old place either

panchagnula: makes sense - yes, i did this while emulating the older existing code. Will remove these - will probably need to update the tests as well.

PR #9987, opened 2019-07-20
graph: fix bug on 'app update' command

### LINKED ISSUE #9895: az ad app update --id OBJID --remove replyUrls "http://my-reply-url" Fails as does remove by Index
I'm trying to remove a replyUrl from an app Registration. I've successfully used the az cli to add a replyUrl on the creation of a new test environment. Now I would like to remove it when the test environment is torn down.

## Describe the bug

**Command Name**
`az ad app update`
`az ad app update --id OBJ-ID --remove replyUrls "http://existing-reply-url.com"`
`az ad app update --id OBJ-ID --remove replyUrls 0`

**Errors:**
```
'NoneType' object has no attribute 'pop'
python3.7/site-packages/knack/cli.py, ln 206, in invoke
cli/core/commands/__init__.py, ln 560, in execute
    raise ex
cli/core/commands/__init__.py, ln 618, in _run_jobs_serially
cli/core/commands/__init__.py, ln 609, in _run_job
    cmd_copy.exception_handler(ex)
cli/command_modules/role/commands.py, ln 69, in graph_err_handler
    raise ex
cli/core/commands/__init__.py, ln 588, in _run_job
    result = cmd_copy(params)
cli/core/commands/__init__.py, ln 297, in __call__
    return self.handler(*args, **kwargs)
cli/core/commands/arm.py, ln 533, in handler
    remove_properties(instance, arg_values)
cli/core/commands/arm.py, ln 905, in remove_properties
    list_to_remove_from.pop(int(list_index))
AttributeError: 'NoneType' object has no attribute 'pop'
```

## To Reproduce:

- `az ad app update --id OBJ-ID --remove replyUrls "http://existing-reply-url.com"`
- `az ad app update --id OBJ-ID --remove replyUrls 0`

## Expected Behavior

The item to be removed from the replyUrls array

## Environment Summary
```
Darwin-18.6.0-x86_64-i386-64bit
Python 3.7.3
Shell: bash

azure-cli 2.0.64 *

```

yugangw-msft:
Fix #9895 

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/role/tests/latest/test_graph.py
+++ b/src/azure-cli/azure/cli/command_modules/role/tests/latest/test_graph.py
@@ -132,43 +132,57 @@ class ApplicationSetScenarioTest(ScenarioTest):
                 }
             ])
         })
-
+        app_id = None
         # create app through general option
         self.cmd('ad app create --display-name {name} --homepage {app} --identifier-uris {app}',
                  checks=self.check('identifierUris[0]', '{app}'))
 
         # set app password
         result = self.cmd('ad app credential reset --id {app} --append --password "test" --years 2').get_output_in_json()
-        self.assertTrue(result['appId'])
+        app_id = result['appId']
+        self.assertTrue(app_id)
 
-        # show/list app
-        self.cmd('ad app show --id {app}',
-                 checks=self.check('identifierUris[0]', '{app}'))
-        self.cmd('ad app list --display-name {name}', checks=[
-            self.check('[0].identifierUris[0]', '{app}'),
-            self.check('length([*])', 1)
-        ])
+        try:
+            # show/list app
+            self.cmd('ad app show --id {app}',
+                     checks=self.check('identifierUris[0]', '{app}'))
+            self.cmd('ad app list --display-name {name}', checks=[
+                self.check('[0].identifierUris[0]', '{app}'),
+                self.check('length([*])', 1)
+            ])
 
-        # update app
-        self.kwargs['reply_uri'] = "http://azureclitest-replyuri"
-        self.cmd('ad app update --id {app} --reply-urls {reply_uri}')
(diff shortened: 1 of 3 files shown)


yugangw-msft:
@zikalino and @Juliehzl, can one of you please take a quick look?

src/azure-cli/azure/cli/command_modules/role/tests/latest/test_graph.py
```diff
@@ -132,43 +132,57 @@ def test_application_set_scenario(self):
...
+        self.assertTrue(app_id)
 
-        # show/list app
-        self.cmd('ad app show --id {app}',
-                 checks=self.check('identifierUris[0]', '{app}'))
-        self.cmd('ad app list --display-name {name}', checks=[
-            self.check('[0].identifierUris[0]', '{app}'),
-            self.check('length([*])', 1)
-        ])
+        try:
```
MyronFanQiu: I am just curious about why we need this `try/finally`. It seems that if we meet any error during the try, this code will not fail and just delete the app.

MyronFanQiu: Will the `self.check` still work? For some commands which don't have `self.check`, does this `try/finally` matter?

yugangw-msft: This is a good question.
The `try/finally` is to ensure the graph application created will always get removed even the test exits before the step of `az ad app delete`; otherwise the leftover will accumulate in the test tenant like garbages 
The test code here is not in a good shape. The right way is to create a test preparer like `ResourceGroupPreparer` which guarantees the chance to clean up anything. I will let @leol7 handle that once he ramps up
`self.check` will only be picked up if the command succeeds. If the command fails, CLI's test FX will throw and control flow will get the the `finally` block

PR #9996, opened 2019-07-22
Add get-access-token CLI commands for RDBMS

rich617:
The intention of the PR is to easily allow a user to retrieve and print an access token suitable for the currently logged-in user to connect to an instance of an OSS RDBMS database in Azure using AAD. 

This is required because OSS RDBMS clients such as psql are community-supported, and so direct integration of AAD authentication into such clients is not possible. As a convenient workaround, these CLI commands allow a customer to retrieve an access token to pass to such clients in order to connect to a database instance using AAD. 

Although the existing command `az account get-access-token` with the `--resource` parameter set to the ossrdbmsEndpointId from `az cloud show` would work, it requires the user to type multiple commands and copy/paste the correct resourceId value.

The solution here is to add a new parameter `--resource-type` to `az account get-access-token`. The new parameter takes one of an enumerated set of well-known resource types. These types are then mapped to a resourceId field in cloud_endpoints.

For example, instead of typing:

`az account get-access-token --resource https://ossrdbms-aad.database.windows.net` 

while logged into the public cloud, the user would instead type:

`az account get-access-token --resource-type oss-rdbms`

---


src/azure-cli-core/azure/cli/core/adal_authentication.py
```diff
@@ -19,8 +19,7 @@ def __init__(self, token_retriever, external_tenant_token_retriever=None):
         self._token_retriever = token_retriever
         self._external_tenant_token_retriever = external_tenant_token_retriever
 
-    def signed_session(self, session=None):  # pylint: disable=arguments-differ
-        session = session or super(AdalAuthentication, self).signed_session()
+    def fetch_tokens(self):
```
yugangw-msft: There is [get_raw_token](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/azure/cli/core/_profile.py#L587) which might be enough, so that you don't need to refactor here.

src/azure-cli/azure/cli/command_modules/rdbms/_help.py
```diff
@@ -878,3 +878,27 @@
   - name: List log files for 'testsvr' less than 30Kb in size.
     text: az postgres server-logs list -g testgroup -s testsvr --max-file-size 30
 """
+
+helps['postgres get-access-token'] = """
```
yugangw-msft: I suggest don't do this; rather just redirect users to use the generic `az account -get-access-token`, and you can even add an example over there.

src/azure-cli-core/azure/cli/core/adal_authentication.py
rich617: I wasn't aware of that one. I will use that instead of changing the AdalAuthentication object.

src/azure-cli/azure/cli/command_modules/rdbms/_help.py
rich617: I see that az account get-access-token will work if --resource is specified. My question is how do I redirect users over there, and what is the appropriate way to show an example of looking up the RDBMS endpoint in "show az cloud", and using that in the command? Is it really that bad to just add a separate command to make it easier for the customer?

rich617: I kept the rdbms get-access-token commands, but implemented the same as az account get-access-token, plus updated the help documentation to indicate that these commands are equivalent to az account get-access-token with --resource set to the ossrdbmsEndpointId from az cloud show. Let me know if this is enough.

src/azure-cli-core/azure/cli/core/adal_authentication.py
rich617: I reverts the changes to AdalAuthentication and used get_raw_token as you requested. Let me know if you need anything else here.

src/azure-cli/azure/cli/command_modules/rdbms/_help.py
yugangw-msft: looking up the RDBMS endpoint in "show az cloud"
You had a good point here.

2 reasons that I suggest we don't introduce new commands on this
1. The rule of thumb is we should only have one command to do one specific thing
2. If we expose these 3 commands, for consistency, we need to do for any commands with unique resource such as `az storage`, `az ad`, `az dls`, etc

For the best of both world, maybe we can update the `--resource` argument `az account get-access-token` to accept an enum values such as `arm`, `rdbms`, `aad_graph`, `ms_graph`, `data_lake`, `storage`, etc?

rich617: I like this idea (allow enums for --resource). Let me work on it. If that's implemented, then I'm happy to remove the rdbms-specific commands.

rich617: I have made the suggested changes. Please review and advise whether or not the implementation is acceptable. Thanks!

rich617: My attempt at an upstream merge worked, but for some reason, github wants me to create a new pull request. I am going to abandon this one. The new PR is https://github.com/Azure/azure-cli/pull/10091 .

rich617:
My attempt at an upstream merge worked, but for some reason, github wants me to create a new pull request. I am going to abandon this one. The new PR is https://github.com/Azure/azure-cli/pull/10091 .

PR #10007, opened 2019-07-23
Appconfig kv restore

bariscaglar:
Added appconfig kv restore command, tests and help. Please add Azure appconfig relevant users to this PR.
---


src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -377,6 +378,66 @@ def list_key(cmd,
...
+
+    query_option_then = QueryKeyValueCollectionOptions(key_filter=key,
+                                                       label_filter=QueryKeyValueCollectionOptions.empty_label if label is not None and not label else label,
+                                                       query_datetime=datetime)
+    query_option_now = QueryKeyValueCollectionOptions(key_filter=key,
+                                                      label_filter=QueryKeyValueCollectionOptions.empty_label if label is not None and not label else label)
+
+    try:
+        restore_keyvalues = azconfig_client.get_keyvalues(query_option_then)
+        current_keyvalues = azconfig_client.get_keyvalues(query_option_now)
```
shenmuxiaosen: This can be simplified by calling list(). No need to create QueryKeyValueCollectionOptions
key_values = list_key(cmd,
                                  key=key,
                                  label=label,
                                  name=name,
                                  datetime=datetime,
                                  connection_string=connection_string,
                                  all_=True)

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -377,6 +378,66 @@ def list_key(cmd,
...
+        if not yes:
+            need_change = __print_restore_preview(kvs_to_restore, kvs_to_modify, kvs_to_delete)
+            if need_change is False:
+                return
+
+        for kv in chain(kvs_to_restore, kvs_to_modify):
+            success = None
+            while success is None and current_retry <= max_retry:
+                try:
+                    success = azconfig_client.set_keyvalue(kv, ModifyKeyValueOptions())
```
shenmuxiaosen: This API won't use etag to check precondition. In azconfig_client, there are three methods related to add/update kvs. 
1. add_keyvalue      use if-none-match:* to make sure adding a new kv
2. update_keyvalue       use if-match:etag to guarantee the kv we will change are based on what we get. 
3. set_keyvalue      this is a force set operation and doesn't check any precondition. 

In this case, we may want to use update_keyvalue

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -377,6 +378,66 @@ def list_key(cmd,
...
+
+        if not yes:
+            need_change = __print_restore_preview(kvs_to_restore, kvs_to_modify, kvs_to_delete)
+            if need_change is False:
+                return
+
+        for kv in chain(kvs_to_restore, kvs_to_modify):
+            success = None
+            while success is None and current_retry <= max_retry:
+                try:
```
shenmuxiaosen: what should we do if it still fails after retry?

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -125,6 +125,12 @@ def load_arguments(self, _):
         c.argument('key', help='If no key specified, return all keys by default. Support star sign as filters, for instance abc* means keys with abc as prefix. Similarly, *abc and *abc* are also supported.')
         c.argument('label', help="If no label specified, list all labels. Support star sign as filters, for instance abc* means labels with abc as prefix. Similarly, *abc and *abc* are also supported.")
 
+    with self.argument_context('appconfig kv restore') as c:
+        c.argument('name', id_part=None)
```
shenmuxiaosen: add argument for yes.   
c.argument('yes', help="Do not prompt for preview.")

since for restore command arguments, "yes" has a different meaning with argument context ('appconfig').

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -377,6 +378,66 @@ def list_key(cmd,
...
+
+    try:
+        restore_keyvalues = azconfig_client.get_keyvalues(query_option_then)
+        current_keyvalues = azconfig_client.get_keyvalues(query_option_now)
+        kvs_to_restore, kvs_to_modify, kvs_to_delete = __compare_kvs_for_restore(restore_keyvalues, current_keyvalues, no_deletes)
+
+        if not yes:
+            need_change = __print_restore_preview(kvs_to_restore, kvs_to_modify, kvs_to_delete)
+            if need_change is False:
+                return
```
shenmuxiaosen: add log if no-op

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
bariscaglar: After retrieving kvs, list_key loops through the keyvalue_iterable, putting them into another list. All I need is the iterable here. If the below table is correct, we can use that as a comment.

If label is            |       label_filter
None                        None
'' (empty string)        empty_label
anything else            label

Or simply:
if (not label)
     label = None

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
bariscaglar: I believe we discussed that we would do a force set operation. Let's say at time t1 the value is v1 and etag e1. At time t2, the value is updated to v2, with etag e2. What we are trying to do is set_keyvalue(k1, v1). If we use update_keyvalue(k1, v1, e1), this will fail, since the latest etag is e2. We could use e2 here but in any case, we had agreed that this should be a forced set since one value being changed in the meantime should not matter, since we are actually going to a time in the past.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
bariscaglar: We currently use c.argument('yes', options_list=['--yes', '-y'], help='Do not prompt for confirmation.'). Does it really matter if it is preview vs confirmation? Because preview==confirmation. If you preview, you must confirm, too. You can only confirm, if you previewed.

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
bariscaglar: added logger.debug

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
bariscaglar: If it fails, we give up after 3. We don't keep trying to restore the others, though. Best to leave it there, log the relevant information, and let the user know.

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
shenmuxiaosen: Yes, we should use force set kv. In this case, why we still catch precondition failed exception and keep retrying based on it?

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
shenmuxiaosen: Sounds good.

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
shenmuxiaosen: In your case, I think if label == None, it means all labels. If label is empty string (--label ""), then it means null label. As in general, null label == empty label. While, for different commands, if client doesn't specify --label, it may have different meanings.

bariscaglar: Oh, you're right. My comment with if (not label) is wrong. Maybe this would be clearer:

if label==''
    label = QueryKeyValueCollectionOptions.empty_label

then in the QueryKeyValueCollectionOptions:

label_filter = label

Because, really the confusing part is the ternary in the label_filter. The implementation for restore is identical to that of list. The help reflects that, too.

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
bariscaglar: I borrowed that part form set_key. (I didn't use that because that does an additional retrieve which we don't need). On the other hand if you look at the other mass operation, import, there's no retry logic. I'm OK with either way, or a third option is retrying for all failures (not just precondition). Is there other precondition failures, other than etag that set_key could run into? Or is that check also erroneous?

shenmuxiaosen: As for set_kv in line 182, it called either add_keyvalue or update_keyvalue which are both conditional. So we can retry based on precondition failed. However, for set_kv, it doesn't check precondition. For restore, it is fine that we don't retry. BTW, the data plane SDK will retry if it is retriable. On the other hand, since we restore sequentially, we need to consider the behavior for partial success. Either we fail as long as one kv failed or we just try best and error out kvs not restored.

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
shenmuxiaosen: That looks good.

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
bariscaglar: Agreed; removed the retry logic entirely. For partial success, I really believe we should fail where it fails, rather than going on with best effort. We are applying the changes in the same order they have confirmed and the user might not want to go on after a given failure. Also, there might be some bigger problems (server unavailable, connection, etc) which you might not want to try and fail for each and every key,

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
bariscaglar: I made the change here to simplify, but left the other locations alone where that ternary logic is used. If we feel like we should simplify those, too, we should file a different issue.

src/azure-cli/azure/cli/command_modules/appconfig/_azconfig/models.py
```diff
@@ -43,6 +43,14 @@ def __init__(self, key, value=None, label=None, tags=None, content_type=None):
         self.last_modified = None
         self.locked = None
 
+    def __str__(self):
+        return "\nKey: " + self.key + \
+               "\nValue: " + self.value + \
+               "\nLabel: " + (self.label if self.label else '') + \
+               "\netag: " + self.etag + \
+               "\nLast Modified: " + self.last_modified + \
```
shenmuxiaosen: Do we want to show etag? Also we should add content-type as well.

bariscaglar: I removed etag, added content type to the JSON output. This is now just being printed to the log file, in case something fails and not shown to the user. I'm adding content type here, and keeping etag, since it might help diagnose things in case of failure.

shenmuxiaosen:
@tjprescott can you take a look at this pr. Thanks!

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -644,6 +741,31 @@ def __print_preview(old_json, new_json):
...
+
+    if kvs_to_modify:
+        print('\nUpdating:')
+        print(json.dumps(__serialize_kv_list_to_comparable_json_list(kvs_to_modify), indent=2))
+
+    if kvs_to_delete:
+        print('\nDeleting:')
+        print(json.dumps(__serialize_kv_list_to_comparable_json_list(kvs_to_delete), indent=2))
+
+    print("")  # printing an empty line for formatting purpose
```
tjprescott: Printing should all occur to STDERR. 
https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md#general-patterns

shenmuxiaosen: It make sense to print to stderr for error messages or logs. But in this case, we are trying to print out preview of a command. For example, to import some key-values into App Configuration, we would like to prompt a preview for confirmation. While, this preview can be a little complex, including key-values objects and some verbs marking "Updated, Deleted or Added". Should we still print to stderr for this scenario?

tjprescott: Yes. The only thing that should go to STDOUT is JSON object structure that will obey CLI output format rules and can be programmatically ingested by other command line tools. This doesn't meet that standard. It's informational.

bariscaglar: @tjprescott is there a recommended way to write to STDERR? I've seen a few different ways people use but I'd want to be consistent with the rest of them.

bariscaglar: Changed the prints to error_print which now prints to sys,stderr

bariscaglar:
@tjprescott could you help complete this PR?

Juliehzl:
@bariscaglar  please resolve conflicts in your files.

Juliehzl:
@bariscaglar have you signed for license/cla?

PR #10021, opened 2019-07-23
[ACR] Enhance connectivity issue related diagnostic message

avtakkar:


src/azure-cli/azure/cli/command_modules/acr/_errors.py
```diff
@@ -76,7 +76,7 @@ def format_error_message(self, *args):
 
 CONNECTIVITY_FORBIDDEN_ERROR = ErrorClass(
     "CONNECTIVITY_FORBIDDEN_ERROR",
-    "Looks like you don't have access to registry '{}'. Are firewalls and virtual networks enabled?"
+    "Looks like you don't have access to registry '{}'. To see configured firewall rules, a) verify that you are logged in to az cli with a role that has the 'Azure Resource Manager' permission: https://aka.ms/acr/authentication/roles; b) run 'az acr network-rule list --name': https://docs.microsoft.com/en-us/cli/azure/acr/network-rule?view=azure-cli-latest#az-acr-network-rule-list"
```
djyou: IMO just mention running `az acr show -n <registry> --query networkRuleSet` is enough, which also shows the switch if firewall is enabled at all. If the user doesn't have permission to do so (i.e., read a registry), ARM has a proper error message already.

Could you also consolidate this message with this one https://github.com/Azure/azure-cli/blob/c48fbd6a58a4f0c3c6c585aaa3f7b9edb5837c69/src/azure-cli/azure/cli/command_modules/acr/_docker_utils.py#L217

/cc @ni-reis

avtakkar: Updated.

PR #10035, opened 2019-07-24
storage: fixed examples

zikalino:
This is mostly about fixing **az storage share create** example, where **--storage-name** argument was missing, and i also renamed some other arguments as they have to be lowercase.


yugangw-msft:
@Juliehzl, please review

Juliehzl:
Acctually, missing account name is not an issue. We do not provide account name argument because people have many ways to do such authentication. When you miss authentication argument, CLI will give you the following message:
```
Missing credentials to access storage service. The following variations are accepted:
    (1) account name and key (--account-name and --account-key options or
        set AZURE_STORAGE_ACCOUNT and AZURE_STORAGE_KEY environment variables)
    (2) account name and SAS token (--sas-token option used with either the --account-name
        option or AZURE_STORAGE_ACCOUNT environment variable)
    (3) account name (--account-name option or AZURE_STORAGE_ACCOUNT environment variable;
        this will make calls to query for a storage account key using login credentials)
    (4) connection string (--connection-string option or
        set AZURE_STORAGE_CONNECTION_STRING environment variable); some shells will require
        quoting to preserve literal character interpretation.
```
In this way, users can do the authentication through above four ways.  And you can see, it is the same for ```az storage container create``` command. If we change the examples for ```az storage share create```, it will be better do the same change to ``az storage container create```.

@yugangw-msft Do you think it is necessary?

zikalino:
@Juliehzl then I think we could have a few examples here, one for each combination.
I just tried the one we had and I had so user could just copy & paste.

yugangw-msft:
With comments from @juliehzl I suggest we don't make the change right now till we receive users feedback specifically on this.
Regarding @zikalino's comment, @sptramer do we have a general doc for storage authentication? If not, we should add one

zikalino:
@Juliehzl @yugangw-msft 
just one more thing. i called renamed "MyStorageAccount" to "storageaccont123". "123" part here is to indicate that name must be uniqe.
can you think of any better approach, i was also thinking of "mystorageaccountuniq" or something like that. i think we should have some kind of consistent approach.

another idea woudl be sth like "storageaccount$UNIQE"

yugangw-msft:
Let us discuss and reopen it with appropriate change

Issue #9845, opened 2019-07-02
az network watcher test-ip-flow fails without resource group and displays bad error


adammodlin:
When running the below command, it should work according to the docs [here](https://docs.microsoft.com/en-us/cli/azure/network/watcher?view=azure-cli-latest#az-network-watcher-test-ip-flow). The docs state that the `--resource-group` parameter is optional.

The second issue is that when this command fails, the incorrect parameter name is used in the failure message. The output is:

```
usage: az network watcher test-ip-flow [-h] [--verbose] [--debug]
                                       [--output {json,jsonc,table,tsv,yaml,none}]
                                       [--query JMESPATH] --vm VM --direction
                                       {Inbound,Outbound} --protocol
                                       {Tcp,Http,Https,Icmp} --local LOCAL
                                       --remote REMOTE
                                       **[--resource-group RESOURCE_GROUP_NAME]**
                                       [--nic NIC]
                                       [--subscription _SUBSCRIPTION]
az network watcher test-ip-flow: error: Parameter '**resource_group_name**' can not be None.
```

**Command Name**
`az network watcher test-ip-flow`

**Errors:**

## To Reproduce:
Steps to reproduce the behavior. Note that argument values have been redacted, as they may contain sensitive information.

- `az network watcher test-ip-flow --direction {} --protocol {} --local {} --remote {} --vm {}`

## Expected Behavior
The command should work according to the documentation. Either the docs need to be updated or the command should not require resource group. Also, when this fails, it should show the correct parameter name.

## Environment Summary
```
Windows-10-10.0.18362-SP0
Python 3.6.6
Shell: powershell.exe

azure-cli 2.0.67

```
## Additional Context
None

MyronFanQiu:
@adammodlin Hello, I checked the code. Now, If you provide the resource id of the vm, it doesn't need the `resource_group_name`. If you provide the name of the vm, it need the `resource_group_name`. The help information is a little bit unclear for now. Besides, I want to have a discussion about how to deal with such question? @tjprescott Is there any way like `required_if` to solve the dependency between several parameters?

tjprescott:
@MyronFanQiu there is a helper method called `validate_parameter_set` in `azure.cli.core.commands.validators` that can assist here.

MyronFanQiu:
@tjprescott Thanks. I will have a try.

MyronFanQiu:
@adammodlin Hello. I have updated the help information and added check for these two arguments. Could we close this issue?

[...]

Issue #9844, opened 2019-07-02
az network watcher test-ip-flow has unexpected error


adammodlin:
I am trying to use `az network watcher test-ip-flow` to diagnose RDP failures to an Azure VM. When I ran the command, I got an error.

**Command Name**
`az network watcher test-ip-flow`

**Errors:**
```
not enough values to unpack (expected 2, got 1)
Traceback (most recent call last):
Temp\pip-install-wl9gly17\knack\knack\cli.py, ln 206, in invoke
azure\cli\core\commands\__init__.py, ln 566, in execute
azure\cli\core\commands\__init__.py, ln 624, in _run_jobs_serially
azure\cli\core\commands\__init__.py, ln 617, in _run_job
Local\Temp\pip-install-wl9gly17\six\six.py, ln 693, in reraise
azure\cli\core\commands\__init__.py, ln 594, in _run_job
azure\cli\core\commands\__init__.py, ln 303, in __call__
azure-cli-core\azure\cli\core\__init__.py, ln 470, in default_command_handler
azure\cli\command_modules\network\custom.py, ln 3047, in check_nw_ip_flow
ValueError: not enough values to unpack (expected 2, got 1)
```

## To Reproduce:
Steps to reproduce the behavior. Note that argument values have been redacted, as they may contain sensitive information.

- _Put any pre-requisite steps here..._
- `az network watcher test-ip-flow --direction {} --protocol {} --local {} --remote {} --vm {} --resource-group {}`

## Expected Behavior
I expected to get the output of the command.

## Environment Summary
```
Windows-10-10.0.18362-SP0
Python 3.6.6
Shell: powershell.exe

azure-cli 2.0.67

```
## Additional Context
You can contact me via my Microsoft alias `admodlin` and I'll share the specifics of the command.

MyronFanQiu:
@adammodlin Hello, could you share the content of the `local` field you are setting? It should be something like `x.x.x.x:port`. If you don't have `:` in the `local`, it will fail. If this is the situation you met, I will make the code more robust and throw out useful error message. If not, let us figure out what's the root cause together.

MyronFanQiu:
@adammodlin Hello. Did your solve your issue? I plan to create a PR to make the code more robust. If it's not the root cause, please let me know.

adammodlin:
@MyronFanQiu I believe I had already included the port in that field. I'll make a note to verify it when I get a chance.

adammodlin:
[...]

PR #10047, opened 2019-07-25
Network: Correct the error messages and help message for the `--vm` and `--nic` arguments for watcher test-ip-flow command

MyronFanQiu:
https://github.com/Azure/azure-cli/issues/9845
https://github.com/Azure/azure-cli/issues/9844
Update the doc, add check and update the error messages.
---

### DIFF
--- a/src/azure-cli/HISTORY.rst
+++ b/src/azure-cli/HISTORY.rst
@@ -16,9 +16,9 @@ Release History
 **Network**
 
 * network nic ip-config add: Fixes #9861 where --ids was inadvertently exposed but did not work.
-* network application-gateway http-settings create/update: Fixes #9604. Add `--root-certs` to support user associate trusted root certificates with the HTTP settings.*
-* network dns record-set ns create: Fixes #9965. Support --subscription again by moving the supression into lower scope.*
-* network watcher test-ip-flow: Fixes #9845. Fixes #9844. Update doc and error messages.*
+* network application-gateway http-settings create/update: Fixes #9604. Add `--root-certs` to support user associate trusted root certificates with the HTTP settings.
+* network dns record-set ns create: Fixes #9965. Support --subscription again by moving the supression into lower scope.
+* network watcher test-ip-flow: Fixes #9845. Fixes #9844. Correct the error messages and help message for the `--vm` and `--nic` arguments. Only providing the name of the vm or nic without `--resource-group` will raise CLI usage error.
 
 **RBAC**
 
--- a/src/azure-cli/HISTORY.rst
+++ b/src/azure-cli/HISTORY.rst
@@ -18,6 +18,7 @@ Release History
 * network nic ip-config add: Fixes #9861 where --ids was inadvertently exposed but did not work.
 * network application-gateway http-settings create/update: Fixes #9604. Add `--root-certs` to support user associate trusted root certificates with the HTTP settings.*
 * network dns record-set ns create: Fixes #9965. Support --subscription again by moving the supression into lower scope.*
+* network watcher test-ip-flow: Fixes #9845. Fixes #9844. Update doc and error messages.*
 
 **RBAC**
 
--- a/src/azure-cli/azure/cli/command_modules/network/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/network/_params.py
@@ -873,8 +873,8 @@ def load_arguments(self, _):
             c.argument('watcher_name', ignore_type, validator=get_network_watcher_from_vm)
             c.ignore('location')
             c.ignore('watcher_rg')
-            c.argument('vm', help='Name or ID of the VM to target.')
-            c.argument('resource_group_name', help='Name of the resource group the target VM is in. Do not use when supplying VM ID.')
+            c.argument('vm', help='Name or ID of the VM to target. If the name of the VM is provided, the --resource_group_name is required.')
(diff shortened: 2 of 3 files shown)


src/azure-cli/azure/cli/command_modules/network/_params.py
```diff
@@ -873,8 +873,8 @@ def load_arguments(self, _):
             c.argument('watcher_name', ignore_type, validator=get_network_watcher_from_vm)
             c.ignore('location')
             c.ignore('watcher_rg')
-            c.argument('vm', help='Name or ID of the VM to target.')
-            c.argument('resource_group_name', help='Name of the resource group the target VM is in. Do not use when supplying VM ID.')
+            c.argument('vm', help='Name or ID of the VM to target. If the name of the VM is provided, the --resource_group_name is required.')
```
Juliehzl: It looks that the resource group name argument in ```az network watcher test-ip-flow``` command should be ```--resource-group```. You can check it again, and if I am right it will be better to update --resource_group_name to --resource-group in help message. In this way, it will be consistent.

src/azure-cli/azure/cli/command_modules/network/custom.py
```diff
@@ -3023,14 +3023,24 @@ def check_nw_ip_flow(cmd, client, vm, watcher_rg, watcher_name, direction, proto
                      resource_group_name=None, nic=None, location=None):
     VerificationIPFlowParameters = cmd.get_models('VerificationIPFlowParameters')
 
-    local_ip_address, local_port = local.split(':')
-    remote_ip_address, remote_port = remote.split(':')
+    try:
+        local_ip_address, local_port = local.split(':')
+        remote_ip_address, remote_port = remote.split(':')
```
Juliehzl: If you want to judge if ```:``` is provided in local or remove variable, split(':') may not work. Because it do not cause error when there is no ```:``` in string. I think contains() or other functions can help you, or you can do a more detailed check for the format.

MyronFanQiu: I think it will cause error when there is no `:` in the string. #9844

src/azure-cli/azure/cli/command_modules/network/_params.py
MyronFanQiu: Thanks. I will fix this

src/azure-cli/HISTORY.rst
```diff
@@ -18,6 +18,7 @@ Release History
 * network nic ip-config add: Fixes #9861 where --ids was inadvertently exposed but did not work.
 * network application-gateway http-settings create/update: Fixes #9604. Add `--root-certs` to support user associate trusted root certificates with the HTTP settings.*
 * network dns record-set ns create: Fixes #9965. Support --subscription again by moving the supression into lower scope.*
+* network watcher test-ip-flow: Fixes #9845. Fixes #9844. Update doc and error messages.*
```
tjprescott: This should be specific about what is actually fixed. No one is going to go back into Github history to find out what this means. Updating docs and error messages doesn't require a history entry.

src/azure-cli/azure/cli/command_modules/network/_params.py
```diff
@@ -873,8 +873,8 @@ def load_arguments(self, _):
             c.argument('watcher_name', ignore_type, validator=get_network_watcher_from_vm)
             c.ignore('location')
             c.ignore('watcher_rg')
-            c.argument('vm', help='Name or ID of the VM to target.')
-            c.argument('resource_group_name', help='Name of the resource group the target VM is in. Do not use when supplying VM ID.')
+            c.argument('vm', help='Name or ID of the VM to target. If the name of the VM is provided, the --resource-name is required.')
```
tjprescott: Do you mean `--resource-group`? I don't see `--resource-name` on the test-ip-flow command.

src/azure-cli/azure/cli/command_modules/network/_params.py
```diff
@@ -873,8 +873,8 @@ def load_arguments(self, _):
...
+            c.argument('resource_group_name', help='Name of the resource group the target VM is in. Required when you provide the name of vm or the name of the nic.')
```
tjprescott: This should not need to be stated. It is a universal CLI pattern.

src/azure-cli/azure/cli/command_modules/network/custom.py
```diff
@@ -3023,14 +3023,24 @@ def check_nw_ip_flow(cmd, client, vm, watcher_rg, watcher_name, direction, proto
...
+        if not resource_group_name:
+            raise CLIError("usage error: '--resource-group' required when use the name of the vm.")
+
         vm = resource_id(
             subscription=get_subscription_id(cmd.cli_ctx), resource_group=resource_group_name,
             namespace='Microsoft.Compute', type='virtualMachines', name=vm)
 
     if nic and not is_valid_resource_id(nic):
+        if not resource_group_name:
+            raise CLIError("usage error: '--resource-group' required when use the name of the nic.")
```
tjprescott: This should look like:
`usage error: --nic NAME --resource-group NAME | --nic ID`

Same with VM.

src/azure-cli/HISTORY.rst
MyronFanQiu: How about this one?

src/azure-cli/azure/cli/command_modules/network/_params.py
MyronFanQiu: fix

PR #10057, opened 2019-07-25
Provide missing resource id support for vnet and subnet

### LINKED ISSUE #9976: Missing support for subnet-id when creating a volume with netappfiles (reopening)

**Describe the bug**
Previously a subnet id could be used when creating volumes when it was an extension.  Since being moved to the main cli this is not available.  The impact here is that I cannot create a volume in a separate resource group to where the vnet is located.

**To Reproduce**
az netappfiles volume create --resource-group <volume-resource-group> --vnet <vname-name> --subnet <subnet-name> ...

This will fail unless the vnet is in the same resource group as the netappfiles.  

**Expected behavior**
An option for --subnet-id is specified for other commands when the vnet is in a different resource group.

**Environment summary**
Linux latest azure-cli.

**Additional context**
It works in the portal.

This is where the subnet-id is generated in the azure cli.  Here is it wrong to assume the resource_group also applies to the vnet:

https://github.com/Azure/azure-cli/blob/20a25fff6775e5dbc8d7ecae1c037426a25a927a/src/azure-cli/azure/cli/command_modules/netappfiles/custom.py#L93

leonardbf:
Fixes #9976

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/netappfiles/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/netappfiles/_help.py
@@ -251,9 +251,9 @@ helps['netappfiles volume create'] = """
         - name: --creation-token
           short-summary: A 1-80 character long alphanumeric string value that identifies a unique file share or mount point in the target subnet
         - name: --vnet
-          short-summary: The vnet for the volume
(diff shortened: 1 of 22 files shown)


tjprescott:
Closing in favor of more recent PR.

PR #10064, opened 2019-07-25
Correct Storage Unit of Measure

apeschel:
The help and usage information in this repository uses GB, when GiB is intended. This PR updates the help and usage information to accurately reflect the unit of measure actually being used by the underlying disks.

This PR is intended to only change help and usage information to correct the unit of measure for storage. Any changes to the logic are unintentional.

---


src/azure-cli/azure/cli/command_modules/container/_params.py
```diff
@@ -65,7 +65,7 @@ def load_arguments(self, _):
         c.argument('location', arg_type=get_location_type(self.cli_ctx), validator=get_default_location_from_resource_group)
         c.argument('image', validator=validate_image, help='The container image name')
         c.argument('cpu', type=float, help='The required number of CPU cores of the containers, accurate to one decimal place')
-        c.argument('memory', type=float, help='The required memory of the containers in GB, accurate to one decimal place')
+        c.argument('memory', type=float, help='The required memory of the containers in GiB, accurate to one decimal place')
```
yugangw-msft: This might be more accurate, but the Azure's doc has used the `GB` across, like [here](https://docs.microsoft.com/en-us/rest/api/container-instances/containergroups/createorupdate#container) and [there](https://docs.microsoft.com/en-us/rest/api/compute/disks/createorupdate). Unless there are users get confused, we should hold on this till doc gets updated first

yugangw-msft: Also, if you buy a physical machine, it is also the [GB](https://www.microsoft.com/en-us/store/config/Surface-Book-2/8MCPZJJCC98C?crosssellid=surfacebook2interstitial&selectedColor=&preview=&previewModes=) which gets mentioned

yonzhan:
Since GB is used across doc and swagger, we will change it together later.
Thanks for the good suggestion.

src/azure-cli/azure/cli/command_modules/container/_params.py
apeschel: @yugangw-msft 

> Also, if you buy a physical machine, it is also the GB which gets mentioned

When you by a physical disk, it's sized in GB, but when you allocate an Azure Disk, it's sized in GiB. Showing GB in these places is extremely misleading, because it does not accurately reflect the underlying size.

PR #10067, opened 2019-07-25
Add network-rule support for Cognitive Services

yangyuan:
---


tjprescott:
@MyronFanQiu evaluate this PR against the guidance here that we discussed yesterday:
https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md#network-rule-commands

src/azure-cli/azure/cli/command_modules/cognitiveservices/custom.py
```diff
@@ -79,24 +81,104 @@ def create(
...
+        rules.ip_rules.append(IpRule(value=ip_address))
+
+    return client.update(resource_group_name, account_name, properties={"networkAcls": rules})
+
+
+def remove_network_rule(cmd, client, resource_group_name, account_name, ip_address=None, subnet=None,
+                        vnet_name=None):  # pylint: disable=unused-argument
+    sa = client.get_properties(resource_group_name, account_name)
+    rules = sa.network_acls
+    if rules is None:
```
MyronFanQiu: could we just return?

yangyuan: Yes sure.

yangyuan:
@MyronFanQiu conflict solved.

For the `default_action` filed.

We want to provide similar experiences with Azure Storage:
https://docs.microsoft.com/en-us/cli/azure/storage/account/network-rule?view=azure-cli-latest
So in CLI we don't support modifying default_action (which is deny by default).

And we don't have `action` filed in rules. (you can consider it as "Allow" by default).

MyronFanQiu:
@yangyuan Got it. Please fix the Lint and Style errors. You can use `azdev style <service_name` to check it locally.

yangyuan:
@MyronFanQiu all checks passed :)

PR #10069, opened 2019-07-26
fix eventhub namespace example

zikalino:
Sometimes it's faster to create PR instead of issue.

Just fixing two things:
**--maximum-throughput-units** must be <=20
when **--maximum-throughput-units** is specified, **--enable-auto-inflate** must be set to true

Error messages:

"validation error: Parameter 'EHNamespace.maximum_throughput_units' must be less than 20"

 "Cannot set MaximumThroughputUnits property if AutoInflate is not enabled"

Also displayed validation error is wrong as **20** actually works for me.


zikalino:
thank you @Juliehzl !

PR #10091, opened 2019-07-30
Add --resource-type parameter to az account get-access-token

rich617:
Because OSS RDBMS clients such as psql are community-supported, direct integration of AAD authentication into such clients is not possible. As a convenient workaround, customers can generate access tokens using the `az account get-access-token` command for use with these clients. 

Although the existing command `az account get-access-token` with the `--resource` parameter set to the ossrdbmsResourceId from `az cloud show` would work, it requires the user to type multiple commands and copy/paste the correct resourceId value.

The solution here is to add a new parameter `--resource-type` to `az account get-access-token`. The new parameter takes one of an enumerated set of well-known resource types. These types are then mapped to a resourceId field in cloud_endpoints.

For example, instead of typing:

`az account get-access-token --resource https://ossrdbms-aad.database.windows.net` 

while logged into the public cloud, the user would instead type:

`az account get-access-token --resource-type oss-rdbms`

This new parameter provides a convenient way to specify well-known resources, including for the ossrdbmsResourceId, without having to lookup and copy the entire resource URL.

---


rich617:
This thing is now so hopelessly out of date and has got so many commits and reversions that I am just going to start over with a new PR again.

PR #10097, opened 2019-07-30
Support for Policy new API version 2019-01-01

RamyasreeChakka:
Support for Policy new API version 2019-01-01

1. Updated History.rst file
2. Added recordings
3. Ran tests in live and non live mode
4. Updated help file


src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -799,6 +799,19 @@
                 "effect": "deny"
             }
         }'
+  - name: Create a policy definition with mode. The mode 'Indexed' indicates the policy should be evaluated only for resource types that support tags and location.
```
mentat9: Nice addition here. It would be cool if you could include this explanation of 'Indexed' directly in the parameter help (_params.py), along with explanations for the other values.

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -102,15 +102,14 @@ def load_arguments(self, _):
...
-        from azure.mgmt.resource.policy.models import PolicyMode
         c.argument('policy_definition_name', arg_type=existing_policy_definition_name_type)
         c.argument('rules', help='JSON formatted string or a path to a file with such content', type=file_type, completer=FilesCompleter())
         c.argument('display_name', help='Display name of policy definition.')
         c.argument('description', help='Description of policy definition.')
         c.argument('params', help='JSON formatted string or a path to a file or uri with parameter definitions.', type=file_type, completer=FilesCompleter(), min_api='2016-12-01')
         c.argument('metadata', min_api='2017-06-01-preview', nargs='+', validator=validate_metadata, help='Metadata in space-separated key=value pairs.')
         c.argument('management_group', arg_type=management_group_name_type)
-        c.argument('mode', arg_type=get_enum_type(PolicyMode), options_list=['--mode', '-m'], help='Mode of the policy definition.', min_api='2016-12-01')
+        c.argument('mode', options_list=['--mode', '-m'], help='Mode of the policy definition, e.g. All, Indexed.', min_api='2016-12-01')
```
mentat9: This would be more helpful if it listed the allowed values (including new data plane values) and what they mean.

RamyasreeChakka: Till GA (Jan next year), only built-in policies are allowed for data plane. Adding extra values with which definitions can't be created will be confusing. I have already included All, Indexed in the help text.

src/azure-cli/azure/cli/command_modules/resource/_help.py
RamyasreeChakka: Adding explanation for all possible modes will result in long help text. Added aka.ms link https://aka.ms/azure-policy-mode for more information.

src/azure-cli/azure/cli/command_modules/resource/_params.py
RamyasreeChakka: Also mentioned link https://aka.ms/azure-policy-mode in the help text.

pilor:
LGTM

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -102,15 +102,14 @@ def load_arguments(self, _):
...
         c.argument('policy_definition_name', arg_type=existing_policy_definition_name_type)
         c.argument('rules', help='JSON formatted string or a path to a file with such content', type=file_type, completer=FilesCompleter())
         c.argument('display_name', help='Display name of policy definition.')
         c.argument('description', help='Description of policy definition.')
         c.argument('params', help='JSON formatted string or a path to a file or uri with parameter definitions.', type=file_type, completer=FilesCompleter(), min_api='2016-12-01')
         c.argument('metadata', min_api='2017-06-01-preview', nargs='+', validator=validate_metadata, help='Metadata in space-separated key=value pairs.')
         c.argument('management_group', arg_type=management_group_name_type)
-        c.argument('mode', arg_type=get_enum_type(PolicyMode), options_list=['--mode', '-m'], help='Mode of the policy definition.', min_api='2016-12-01')
+        c.argument('mode', options_list=['--mode', '-m'], help='Mode of the policy definition, e.g. All, Indexed. Please visit https://aka.ms/azure-policy-mode for more information.', min_api='2016-12-01')
```
mentat9: I think something like this would be more helpful: 'Mode of the policy definition. Values: one of All, Indexed, or <resource-provider>. See https://aka.ms/azure-policy-mode for more information.' The link already mentions resource provider modes, and will be updated as new ones are added. If you explicitly want to leave out <resource-provider> from that list, then I'd suggest wording like this: 'Mode of the policy definition, usually All or Indexed. See https://aka.ms/azure-policy-mode for more information.'

PR #10107, opened 2019-07-31
Updated Readme for non-root docker usage

kitingChris:
When mounting the home directory the container should not be running without the current user id and group id.
Otherwise all files that azure-cli writes will be written as root - in the container and on the host.
See also [https://medium.com/redbubble/running-a-docker-container-as-a-non-root-user-7d2e00f8ee15](https://medium.com/redbubble/running-a-docker-container-as-a-non-root-user-7d2e00f8ee15)

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ $ docker run -u $(id -u):$(id -g) -v ${HOME}:/home/az -e HOME=/home/az --rm -it 
 For automated builds triggered by pushes to this repo, see [azuresdk/azure-cli-python](https://hub.docker.com/r/azuresdk/azure-cli-python/tags).
 For example:
 ```bash
-docker run -v ${HOME}:/root -it --rm azuresdk/azure-cli-python:dev
+$ docker run -u $(id -u):$(id -g) -v ${HOME}:/home/az -e HOME=/home/az --rm -it azuresdk/azure-cli-python:dev
 ```
 
 ### Edge Builds
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ We maintain a Docker image preconfigured with the Azure CLI.
 See our [Docker tags](https://hub.docker.com/r/microsoft/azure-cli/tags/) for available versions.
 
 ```bash
-$ docker run -v ${HOME}:/root -it --rm mcr.microsoft.com/azure-cli:<version>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -106,7 +106,7 @@ We maintain a Docker image preconfigured with the Azure CLI.
 See our [Docker tags](https://hub.docker.com/r/microsoft/azure-cli/tags/) for available versions.
 
 ```bash
-$ docker run -v ${HOME}:/root -it --rm mcr.microsoft.com/azure-cli:<version>
+$ docker run -u $(id -u):$(id -g) -v ${HOME}:/home/az -e HOME=/home/az --rm -it mcr.microsoft.com/azure-cli:<version>
```
yugangw-msft: I suggest we keep it as it is unless we find the option which would work cross OSs including Windows. 
Thanks for sharing the doc though where I learnt some good things.

kitingChris: Hi @yugangw-msft no problem. But since the initial command works only in bash (${HOME} variable) this suggestion will work in a bash on windows as well ;-)

fengzhou-msft: @kitingChris shall we change [line 115](https://github.com/Azure/azure-cli/pull/10107/files/eacf4399c2d706b46a98ff1a7cd7f24f889757f6#diff-04c6e90faac2675aa89e2176d2eec7d8R115) as well, to keep consistency?

yonzhan:
@kitingChris  please take a look at the comments. Do we still need to merge this PR ?

kitingChris:
@yonzhan I would still recommend it. The current way is ok for everyone using root as main user running in the container. On windows you usually do not have any problem with that cause windows just ignores the linux user ids. 

If you run the commands in a linux environment having your ${HOME} mounted in root and therefore having files created as root inside the container but stored outside in your home results in files created by user id 0 (root) in your hosts file system. 

Which then can only removed by root or via sudo. In your home directory. 

That is why I would recommend running as non-root user the container services.

README.md
kitingChris: @fengzhou-msft yes! Applies there as well. Everywhere where you mount your home directory and you can expect files to be created.

kitingChris:
PR updated due to @fengzhou-msft suggestion

yonzhan:
@fengzhou-msft will help with this PR.

Issue #9657, opened 2019-06-13
aks create failed via cli while creating aks well by manual on the portal


dragonsappire:
> ### `az feedback` auto-generates most of the information requested below, as of CLI version 
azure-cli (2.0.46)
**Describe the bug**

**To Reproduce**
Steps to reproduce the behavior.
running cmd as below output:

**Expected behavior**

**Environment summary**
Install Method (e.g. pip, interactive script, apt-get, Docker, MSI, edge build) / CLI version (`az --version`) / OS version / Shell Type (e.g. bash, cmd.exe, Bash on Windows)  

**Additional context**
[work@monitor-vm ~]$ az account list-locations --output=table | grep chinaeast2
China East 2   31.3209     121.5891     chinaeast2
[work@monitor-vm ~]$ az aks create --resource-group xxxx --name xxxxx --node-count 1 --enable-addons monitoring --generate-ssh-keys --node-vm-size Standard_D2s_v3 --kubernetes-version 1.13.5 --subscription xxxxx --location chinaeast2
u'chinaeast2'
Traceback (most recent call last):
  File "/home/work/local/azure-cli/lib/lib/python2.7/site-packages/knack/cli.py", line 197, in invoke
    cmd_result = self.invocation.execute(args)
  File "/home/work/local/azure-cli/lib/lib/python2.7/site-packages/azure/cli/core/commands/__init__.py", line 369, in execute
    six.reraise(*sys.exc_info())
  File "/home/work/local/azure-cli/lib/lib/python2.7/site-packages/azure/cli/core/commands/__init__.py", line 343, in execute
    result = cmd(params)
  File "/home/work/local/azure-cli/lib/lib/python2.7/site-packages/azure/cli/core/commands/__init__.py", line 182, in __call__
    return self.handler(*args, **kwargs)
  File "/home/work/local/azure-cli/lib/lib/python2.7/site-packages/azure/cli/core/__init__.py", line 436, in default_command_handler
    result = op(**command_args)
  File "/home/work/local/azure-cli/lib/lib/python2.7/site-packages/azure/cli/command_modules/acs/custom.py", line 1459, in aks_create
    workspace_resource_id
  File "/home/work/local/azure-cli/lib/lib/python2.7/site-packages/azure/cli/command_modules/acs/custom.py", line 1726, in _handle_addons_args
    cmd, subscription_id, resource_group_name)
[...]

PR #10114, opened 2019-08-01
Add some better error messages for an edge case.

brendandburns:
Fixes https://github.com/Azure/azure-cli/issues/9657

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acs/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/custom.py
@@ -2071,6 +2071,12 @@ def _ensure_default_log_analytics_workspace_for_monitoring(cmd, subscription_id,
     workspace_region = "eastus"
     workspace_region_code = "EUS"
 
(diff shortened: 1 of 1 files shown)


MyronFanQiu:
@brendandburns Hello. The title should be meaningful which summarize the content of this PR. Besides, History.rst should be updated.

src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -2071,6 +2071,12 @@ def _ensure_default_log_analytics_workspace_for_monitoring(cmd, subscription_id,
     workspace_region = "eastus"
     workspace_region_code = "EUS"
 
+    # sanity check that locations and clouds match.
+    if ((cloud_name.lower() == 'azurecloud' and AzureChinaRegionToOmsRegionMap.get(rg_location, False)) or
```
yugangw-msft: nit, `rg_location in AzureChinaRegionToOmsRegionMap` might save a few characters here.

PR #10115, opened 2019-08-01
Update document for change log

Juliehzl:
---

### DIFF
--- a/doc/authoring_command_modules/README.md
+++ b/doc/authoring_command_modules/README.md
@@ -212,7 +212,6 @@ e.g.:
 Release History
 ===============
 
-0.0.3
 +++++
 * This is my customer-facing change.
 
--- a/doc/authoring_command_modules/README.md
+++ b/doc/authoring_command_modules/README.md
@@ -201,7 +201,7 @@ Submitting Pull Requests
 
 ### Modify Change Log
 
-Modify the `HISTORY.rst` for any customer-facing changes. If a module has changed at all since a previous release so that a version bump is required, it is fine to add a generic entry that says "* Minor fixes.".
+Modify the `HISTORY.rst` for any customer-facing changes. 
 
 This will be the release notes for the next release.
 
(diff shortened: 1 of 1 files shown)


doc/authoring_command_modules/README.md
```diff
@@ -212,7 +212,6 @@ e.g.:
 Release History
 ===============
 
-0.0.3
```
MyronFanQiu: any story?

Juliehzl: Customers do not need to add version number for next release. We will add it during release.

doc/authoring_command_modules/README.md
```diff
@@ -212,7 +212,6 @@ e.g.:
...
 +++++
```
MyronFanQiu: who does need to add this line?

doc/authoring_command_modules/README.md
MyronFanQiu: Sound reasonable. I prefer adding the next release info into history when we finish the current one.

doc/authoring_command_modules/README.md
MyronFanQiu: +++++

MyronFanQiu:
@yugangw-msft I prefer we have something like
```
History notes for next release
+++++++++++++++++++
**<command module>**

* This is my customer-facing change.

```
Then during release, we can change the release version to the right one. After release, we should provide the same content in the history.rst.

Juliehzl:
Because in document we have already declare how to add release notes for next release, the part I changed is just an sample to specify where users can add their release notes. I think we can just remove  ```version number``` and ```+++++++++++++++++++```,  and users can know how to add their release notes in the right position with document and sample shown.

yugangw-msft:
The goals is to make sure the sample here is as closer to the [real one](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/HISTORY.rst) as possible, so users can easily find the spot to input the change log correctly.

yugangw-msft:
Please do rebase 
```
git checkout dev
git pull upstream dev
git checkout changelog
git rebase dev
```

yonzhan:
Since this PR was very old. Let's discuss offline.

PR #10147, opened 2019-08-05
Fix#10024: Change default installation location without root privileges

Juliehzl:
---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/storage/azcopy/util.py
+++ b/src/azure-cli/azure/cli/command_modules/storage/azcopy/util.py
@@ -164,7 +164,7 @@ def _get_default_install_location():
             return None
         install_location = os.path.join(home_dir, r'.azcopy\azcopy.exe')
     elif system in ('Linux', 'Darwin'):
-        install_location = '/usr/local/bin/azcopy'
+        install_location = '~/bin/azcopy'
     else:
         install_location = None
     return install_location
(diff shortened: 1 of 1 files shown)


Juliehzl:
@yugangw-msft Can you review this PR to fix privilege requirement for azcopy related commands?

src/azure-cli/azure/cli/command_modules/storage/azcopy/util.py
```diff
@@ -164,7 +164,7 @@ def _get_default_install_location():
             return None
         install_location = os.path.join(home_dir, r'.azcopy\azcopy.exe')
     elif system in ('Linux', 'Darwin'):
-        install_location = '/usr/local/bin/azcopy'
+        install_location = '/home/bin/azcopy'
```
yugangw-msft: This should work. Consider OSX's user folder is `/users/<name>`,  to get 100% worry free, consider use `~` like [here](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/azure/cli/core/_environment.py#L9).

src/azure-cli/azure/cli/command_modules/storage/azcopy/util.py
```diff
@@ -164,7 +164,7 @@ def _get_default_install_location():
             return None
         install_location = os.path.join(home_dir, r'.azcopy\azcopy.exe')
     elif system in ('Linux', 'Darwin'):
-        install_location = '/usr/local/bin/azcopy'
+        install_location = '~/bin/azcopy'
```
yugangw-msft: Please expand the "~" through `expanduser`, as the underneath file system doesn't recognize "~"

PR #10148, opened 2019-08-05
[storage] emit error when windows is 32bit for azcopy

Juliehzl:
---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/storage/azcopy/util.py
+++ b/src/azure-cli/azure/cli/command_modules/storage/azcopy/util.py
@@ -36,8 +36,6 @@ class AzCopy(object):
                 os.makedirs(install_dir)
             base_url = 'https://azcopyvnext.azureedge.net/release20190423/azcopy_{}_amd64_10.1.0.{}'
             if self.system == 'Windows':
-                if platform.machine() != 'AMD64':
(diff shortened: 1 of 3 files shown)


src/azure-cli/azure/cli/command_modules/storage/azcopy/util.py
```diff
@@ -36,6 +36,8 @@ def __init__(self, creds=None):
                 os.makedirs(install_dir)
             base_url = 'https://azcopyvnext.azureedge.net/release20190423/azcopy_{}_amd64_10.1.0.{}'
             if self.system == 'Windows':
+                if platform.machine() != 'AMD64':
```
yugangw-msft: This would not work for OSX & Linux

Juliehzl: Right! I will deprecate this change and declare in help message. But when 32-bit is ready, we have to consider how to detect OS architecture again. From my side, platform.machine() can return values related to architecture. It seems that it will return AMD64 for 64-bit Windows and x86-64 for 64-bit Linux OS. For MacOS, I have no idea and cannot try anywhere.

Juliehzl: One more consideration is that it is a little strange to warning in help message. I think it will be a better way to detect the OS architecture and warning when not matched.

PR #10172, opened 2019-08-06
webapp: az webapp up detects the 'env' folder for python & removes this from the zip folder used for deployment

### LINKED ISSUE #9997: [web] - Az webapp up logic for removing dirs for python detected app, before performing zip deployment
look for site-packages, and backtrack from there,

`<virtualenv>/lib/<pythonversion>/<site-packages>`
or
`root/<virtual environment>/lib/<site-packages>`

So site packages, backtrack to the folder containing lib, assume it's the virtual environment and then ignore that entire folder

panchagnula:
Fixes #9997
---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/_create_util.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/_create_util.py
@@ -32,14 +32,13 @@ def zip_contents_from_dir(dirPath, lang):
     with zipfile.ZipFile("{}".format(zip_file_path), "w", zipfile.ZIP_DEFLATED) as zf:
         for dirname, subdirs, files in os.walk(dirPath):
             # skip node_modules folder for Node apps,
-            # since zip_deployment will perfom the build operation
-            if lang.lower() == NODE_RUNTIME_NAME and 'node_modules' in subdirs:
-                subdirs.remove('node_modules')
+            # since zip_deployment will perform the build operation
+            if lang.lower() == NODE_RUNTIME_NAME:
+                subdirs[:] = [d for d in subdirs if 'node_modules' not in d]
             elif lang.lower() == NETCORE_RUNTIME_NAME:
-                if 'bin' in subdirs:
-                    subdirs.remove('bin')
-                elif 'obj' in subdirs:
-                    subdirs.remove('obj')
+                subdirs[:] = [d for d in subdirs if d not in ['obj', 'bin']]
+            elif lang.lower() == PYTHON_RUNTIME_NAME:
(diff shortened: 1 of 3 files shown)


src/azure-cli/azure/cli/command_modules/appservice/_create_util.py
```diff
@@ -40,6 +40,17 @@ def zip_contents_from_dir(dirPath, lang):
...
                 elif 'obj' in subdirs:
                     subdirs.remove('obj')
+            elif lang.lower() == PYTHON_RUNTIME_NAME:
+                _folder_to_search = os.path.join('lib', 'site-packages')
+                path_val = dirname.lower()
+                if _folder_to_search in path_val:
+                    _parent_path = path_val[:path_val.index(_folder_to_search)]
+                    import shutil
+                    shutil.rmtree(_parent_path, ignore_errors=True)  # ignore errors so that we don't fail the operation
+            if '.github' in subdirs:
```
yugangw-msft: Are you sure to remove this as this will detach the source from git source control. To me it is a data loss. Can we copy to a temp folder, prune and deploy from there?

panchagnula: you are talking just about the .github check right, I will just remove that for now.

src/azure-cli/azure/cli/command_modules/appservice/_create_util.py
```diff
@@ -40,6 +40,15 @@ def zip_contents_from_dir(dirPath, lang):
...
                     subdirs.remove('obj')
+            elif lang.lower() == PYTHON_RUNTIME_NAME:
+                _folder_to_search = os.path.join('lib', 'site-packages')
+                path_val = dirname.lower()
+                if _folder_to_search in path_val:
+                    _parent_path = path_val[:path_val.index(_folder_to_search)]
+                    import shutil
+                    shutil.rmtree(_parent_path, ignore_errors=True)  # ignore errors so that we don't fail the operation
+            if '.azure' in subdirs:
+                subdirs.remove('.azure')
```
yugangw-msft: please also remove this, as otherwise you would wipe out the local context

yugangw-msft:
@panchagnula, I can merge this, but it could be a pretty bad idea to some users, that after the deployment, users would have to rebuild the virtual env which could take a long while. If `az webapp up` will be used repeatedly, then we should not delete, rather should use skip like mechanism
Let me know how to proceed

panchagnula:
@tjprescott  with Yugang out can you do quick review & merge please?

PR #10306, opened 2019-08-23
Update vm create -h. --size. Use populator-commands to add "Values from" message.

### LINKED ISSUE #10305: parameters dont have comment about valid values in several properties or have link that has valid values
parameters dont have comment about valid values in several properties or have link that has valid values
don't provide syntax for all values
az vm create does not have valid values for:
size, 

---
#### Document Details

* ID: 152b549b-d4b3-13e5-081c-c604b6e4d89e
* Version Independent ID: 3f79be78-ce75-ad31-304b-bad7d475ec9e
* Content: [az vm](https://docs.microsoft.com/en-us/cli/azure/vm?view=azure-cli-latest#feedback)
* Content Source: [src/azure-cli/azure/cli/command_modules/vm/_help.py](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/vm/_help.py)
* Service: **virtual-machines**
* GitHub Login: @rloutlaw

qwordy:
Fix #10305
---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/vm/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_help.py
@@ -731,6 +731,9 @@ parameters:
     populator-commands:
       - az vm image list
       - az vm image show
+  - name: --size
+    populator-commands:
+      - az vm list-sizes
   - name: --ssh-key-values
     short-summary: Space-separated list of SSH public keys or public key file paths.
   - name: --computer-name


qwordy:
Does anyone know populator-commands?

MyronFanQiu:
@qwordy I cannot find the corresponding filed. https://github.com/microsoft/knack/blob/44ee1799c891213c90916f186ab9aee7b9ff8344/knack/arguments.py#L299

src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -580,7 +580,7 @@ def load_arguments(self, _):
             c.argument('tags', tags_type)
             c.argument('no_wait', help='Do not wait for the long-running operation to finish.')
             c.argument('validate', options_list=['--validate'], help='Generate and validate the ARM template without creating any resources.', action='store_true')
-            c.argument('size', help='The VM size to be created. See https://azure.microsoft.com/pricing/details/virtual-machines/ for size info.')
+            c.argument('size', help='The VM size to be created. See https://azure.microsoft.com/pricing/details/virtual-machines/ for size info. Values from: az vm list-sizes.')
```
Juliehzl: Do you have default value for vm size to be created? If yes, you should declare here. And it will be better to say "Allowed values can be get through `az vm list-sizes` command."

MyronFanQiu: Looks like we should also change the description for populator-commands

> 
> Do you have default value for vm size to be created? If yes, you should declare here. And it will be better to say "Allowed values can be get through `az vm list-sizes` command."

Default value has been configured, but not here. "help" parameter's content is not the final output message. The format is copied from other parameters like --image.

qwordy:
> 
> 
> You can write more details into the PR's title.

OK

PR #10320, opened 2019-08-26
[Network] update help info for list-service-tags

### LINKED ISSUE #10244: Add more details to reduce customer confusion for "az network list-service-tags"

**Describe the Problem**
https://docs.microsoft.com/en-us/cli/azure/network?view=azure-cli-latest#az-network-list-service-tags
The location parameter is causing lots of customer confusion. 
Wee need to add more details:

"The location that will be used as a reference for version (not as a filter based on location, you will get the list of service tags with prefix details across all regions but limited to the cloud that your subscription belongs to)."

MyronFanQiu:
Fixes #10244 
---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/network/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/network/_help.py
@@ -12,6 +12,13 @@ type: group
 short-summary: Manage Azure Network resources.
 """
 
+helps['network list-service-tags'] = """
+type: command
+long-summary: >
+    A service tag represents a group of IP address prefixes to help minimize complexity for security rule creation.
+    To learn more about list-service-tags, visit https://docs.microsoft.com/en-us/azure/virtual-network/security-overview#service-tags
+"""
+
 helps['network application-gateway'] = """
 type: group
 short-summary: Manage application-level routing and load balancing services.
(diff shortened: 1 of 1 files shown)


src/azure-cli/azure/cli/command_modules/network/_help.py
```diff
@@ -12,6 +12,13 @@
 short-summary: Manage Azure Network resources.
 """
 
+helps['network list-service-tags'] = """
+type: command
+long-summary: >
```
Juliehzl: What does `>` mean here?

MyronFanQiu: you can split the long summary into multiple lines

src/azure-cli/azure/cli/command_modules/network/_help.py
```diff
@@ -12,6 +12,13 @@
 short-summary: Manage Azure Network resources.
 """
 
+helps['network list-service-tags'] = """
```
Juliehzl: One suggestion is that it will better to add a short summary for this command. Because if not, it will find short summary from SDK or other places and the fetched message use `Gets`, which is not consistent with [CLI command guideline](https://github.com/Azure/azure-cli/blob/dev/doc/authoring_help.md#tips-to-write-effective-help-for-your-command ) and other commands.

MyronFanQiu: now it will expose the SDK's help information to the user. Do you suggest us overwriting this help information?

Juliehzl: Right. Just my suggestion. It's not consistent with `list-usages` command when you run `az network -h` and don't follow CLI command guideline as you know.

PR #10358, opened 2019-08-28
Ensure transfer is complete before execution

zined:
This script is being published via https://aka.ms/InstallAzureCLIDeb and being used in a `curl | bash` way. In order to avoid partial execution in case of incomplete transfers, wrap the whole setup in a bash function which is called when the transfer is completed.

---


yugangw-msft:
@zined , thanks for the contribution.

Issue #1149, opened 2016-10-25
AKS: expose generic update


yugangw-msft:
Do it when one of the following conditions are met:
a. ACS has properties we can tweak
b. The 'Generic update' command can be customized, say to hide `--remove`,  `--add`.

//cc: @tjprescott

johanste:
Every resource should support the generic update. And every generic update should have --remove and --add (at least if the resource have one or more collection type properties).

tjprescott:
In the case of ACS, I believe the argument is that it in fact doesn't have collection properties and thus the motivation to suppress add and remove. We tried accomplishing this through aliasing and it didn't work, so after discussion we merged the PR without the generic update. There's a possibility that when #1059 is merged that we can accomplish this with just aliasing. If not it will be a little more work to make generic update support this scenario.

mboersma:
@tjprescott I'm not entirely clear on the generic update mechanism that was discussed here. I do know we don't plan to implement changes other than bug fixes and maintenance to ACS itself, so this can probably be closed or retargeted to AKS.

tjprescott:
Updated to target AKS, which apparently doesn't have an `update` method at all.

mboersma:
@yugangw-msft the Python SDK for AKS has [`create_or_update`](https://github.com/Azure/azure-sdk-for-python/blob/master/azure-mgmt-containerservice/azure/mgmt/containerservice/v2018_03_31/operations/managed_clusters_operations.py#L565) and [`update_tags`](https://github.com/Azure/azure-sdk-for-python/blob/master/azure-mgmt-containerservice/azure/mgmt/containerservice/v2018_03_31/operations/managed_clusters_operations.py#L672) methods, but I don't foresee other significant changes. I assume that's not equivalent to a "generic update" method.

Is there more we could do here or should this be closed?

tjprescott:
This shouldn't be closed. The generic update would make use of your `create_or_update` method to work. Please see:
https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#generic-update-commands

jnoller:
As this is an enhancement, we will look at sizing and possible improvements along with other CLI issues.

seanmck:
I've created a PR for updating ARM tags in our preview extension:

https://github.com/Azure/azure-cli-extensions/pull/851

[...]

PR #10379, opened 2019-08-30
[AKS] Add support of ACR integration

feiskyer:
Add support of ACR integration, which includes

* Add `--attach-acr <acr-name-or-resource-id>` to `az aks create` command, which allows for attach the ACR to AKS cluster.
* Add `--attach-acr <acr-name-or-resource-id>` and `--detach-acr <acr-name-or-resource-id>` to `az aks update` command, which allows to attach or detach the ACR from AKS cluster.

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acs/commands.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/commands.py
@@ -16,6 +16,7 @@ from ._format import aks_upgrades_table_format
 from ._format import aks_versions_table_format
 
 
+# pylint: disable=too-many-statements
 def load_command_table(self, _):
 
     container_services_sdk = CliCommandType(
@@ -85,6 +86,7 @@ def load_command_table(self, _):
                          'Are you sure you want to perform this operation?')
         g.custom_command('upgrade-connector', 'k8s_upgrade_connector', is_preview=True)
         g.custom_command('use-dev-spaces', 'aks_use_dev_spaces')
+        g.custom_command('update', 'aks_update', supports_no_wait=True)
         g.wait_command('wait')
 
     with self.command_group('aks', container_services_sdk, client_factory=cf_container_services) as g:
--- a/src/azure-cli/azure/cli/command_modules/acs/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/custom.py
@@ -73,6 +73,8 @@ from ._client_factory import cf_resource_groups
 from ._client_factory import get_auth_management_client
 from ._client_factory import get_graph_rbac_management_client
 from ._client_factory import cf_resources
+from ._client_factory import get_resource_by_name
+from ._client_factory import cf_container_registry_service
 
 logger = get_logger(__name__)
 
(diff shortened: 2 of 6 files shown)


feiskyer:
@zikalino Could you help to take a look?

src/azure-cli/azure/cli/command_modules/acs/commands.py
```diff
@@ -85,6 +86,7 @@ def load_command_table(self, _):
                          'Are you sure you want to perform this operation?')
         g.custom_command('upgrade-connector', 'k8s_upgrade_connector', is_preview=True)
         g.custom_command('use-dev-spaces', 'aks_use_dev_spaces')
+        g.custom_command('update', 'aks_update', supports_no_wait=True)
```
yugangw-msft: this should be a generic updater, check out how [VM does](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/vm/commands.py#L237)

src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -2264,6 +2400,58 @@ def _ensure_container_insights_for_monitoring(cmd, addon):
...
 
 
+def _ensure_aks_acr(cli_ctx,
+                    client_id,
+                    acr_name_or_id,
+                    subscription_id,
+                    detach=False):
+    from msrestazure.tools import is_valid_resource_id
+    # Check if the ACR exists by resource ID.
+    if is_valid_resource_id(acr_name_or_id):
```
yugangw-msft: You should go through ACR management client. ACR SDK is multu-api client, with potential to work in other cloud env, while your hard coded `2019-05-01` might not be available outside of public cloud. Let us follow the right pattern.

src/azure-cli/azure/cli/command_modules/acs/commands.py
feiskyer: I think it's a little different here, not all the options we updated are from AKS models. Should we still swith to generic updater?

yugangw-msft: Yes, I understand the difference.
The guideline calls out you should always start with generic updater, which is one liner simple change and gives you lots of other things for free such as `tagging`, and then you can expose strong arguments like the "--attach/detach acr".
For context missing `update` command from AKS is a long over-due, and since you are at it, I suggest very much you just get it done and resolve https://github.com/Azure/azure-cli/issues/1149.

feiskyer: That's actually one of the issues. When the cluster enables multi agent pools, the create_or_update call would also fail because agent pools updates have their own subresource APIs.

Could we keep it as custom command until we figured out how to fix #1149?

src/azure-cli/azure/cli/command_modules/acs/custom.py
feiskyer: Good catch. Fixed in the new commit.

feiskyer:
@yugangw-msft thanks a lot

PR #10383, opened 2019-08-30
[acr] import --image warn users if target_tag/image is prefixed with '/'

adewaleo:
---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acr/import.py
+++ b/src/azure-cli/azure/cli/command_modules/acr/import.py
@@ -4,12 +4,15 @@
 # --------------------------------------------------------------------------------------------
 
 from knack.util import CLIError
+from knack.log import get_logger
 from msrestazure.tools import is_valid_resource_id
 
 from ._utils import (
     validate_managed_registry,
     get_registry_from_name_or_login_server
 )
+
+logger = get_logger(__name__)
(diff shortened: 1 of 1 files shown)


adewaleo:
Simple validation to deal with the user-input issue @SteveLasker found. Does this look good or is this overkill @sajayantony

target tags warning

SteveLasker:
Is this an error or warning? I'm still not convinced we shouldn't just fix this up for the user. 
what happens when I `push demo42.azurecr.io/repo//image:tag` Does the // generate an error, or just get fixed up?

adewaleo:
@SteveLasker 

The warning is coming from the CLI (the yellow), whilst the error is coming from the service. This PR is only for the CLI

I have made a PR to the service's RP to use a shortened link instead of the longer url. The spacing error was actually a field reference not the period between two sentences.

I tried tagging an image locally so I could attempt to push the image to a private ACR instance. I got the following error:

```
> docker tag hello-world tosinreg5.azurecr.io//hello-world:v1
Error parsing reference: "tosinreg5.azurecr.io//hello-world:v1" is not a valid repository/tag: invalid reference format
```

adewaleo:
I will be closing this pr. Seems like we won't be moving ahead with client-side validation this time.

Issue #10124, opened 2019-08-02
Add terminate scheduled event configurability


szimmerMicrosoft:
**Resource Provider**
What is the Azure resource provider your feature is part of?

CRP, Microsoft.Compute

**Description of Feature or Work Requested**
Provide a brief description of the feature or work requested. A link to conceptual documentation may be helpful too.

Additional of a couple Virtual Machine Scale Set properties in ScheduledEventsProfile.TerminateNotificationProfile.

This is a configurable property (on create and update)
Which will enable a notification to be sent to the VM on the instance metadata server that a VM is being deleted (terminated).

**Minimum API Version Required**
What is the minimum API version of your service required to implement your feature?

2019-03-01

**Swagger Link**
Provide a link to the location of your feature(s) in the REST API specs repo. This should be on the master branch
of the REST API specs repo.

https://github.com/Azure/azure-rest-api-specs/blob/master/specification/compute/resource-manager/Microsoft.Compute/stable/2019-03-01/compute.json

All references from scheduledEventsProfile (line 8519)

**Target Date**
If you have a target date for release of this feature/work, please provide it. While we can't guarantee these dates,
it will help us prioritize your request against other requests.

8/8/2019

qwordy:
What is default value of enable? "false"?

szimmerMicrosoft:
This is by default disabled.

By default scheduledEventsProfile is null, which disables this feature.

This is equivalent to setting TerminateNotificationProfile.enable = false

qwordy:
> 
> 
> This is by default disabled.
> 
> By default scheduledEventsProfile is null, which disables this feature.
> 
> This is equivalent to setting TerminateNotificationProfile.enable = false

OK. Thank you!

qwordy:
Next time, provide a pull request link so that we can see the swagger changes.
Thank you!

qwordy:
Merged.

PR #10391, opened 2019-09-02
[Compute] vmss create/update: add --terminate-notification and --terminate-notification-time parameters to support terminate scheduled event configurability. Issue #10124

qwordy:
[VM] vmss create: --terminate-notification, add terminate scheduled event configurability #10124

---


src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -585,6 +585,7 @@ def load_arguments(self, _):
             c.argument('custom_data', help='Custom init script file or text (cloud-init, cloud-config, etc..)', completer=FilesCompleter(), type=file_type)
             c.argument('secrets', multi_ids_type, help='One or many Key Vault secrets as JSON strings or files via `@{path}` containing `[{ "sourceVault": { "id": "value" }, "vaultCertificates": [{ "certificateUrl": "value", "certificateStore": "cert store name (only on windows)"}] }]`', type=file_type, completer=FilesCompleter())
             c.argument('assign_identity', nargs='*', arg_group='Managed Service Identity', help="accept system or user assigned identities separated by spaces. Use '[system]' to refer system assigned identity, or a resource id to refer user assigned identity. Check out help for more examples")
+            c.argument('terminate_notification', help='Terminate Scheduled Event. Configurable length of time a Virtual Machine being deleted will have to potentially approve the Terminate Scheduled Event before the event is auto approved (timed out). The configuration must be specified in ISO 8601 format. e.g. PT5M')
```
yugangw-msft: I suggest we reword the whole comment a bit. I know it is from the swagger, but it makes me feel a bit out of breath to read it :). Maybe `length of time a notification to be sent to the VM on the instance metadata server till the VM gets deleted. `? Also please cross check the `PT5M` is 8061 format?

yugangw-msft: Please add api-version range

yugangw-msft:
> I cannot understand, how can the customer disable this field after creating it.

Good point. I believe "az vmss update" should handle it through the generic updater, but let us cross check.

src/azure-cli/HISTORY.rst
```diff
@@ -10,6 +10,10 @@ Release History
 * [BREAKING] When not specified, the default value for `--start-task-wait-for-success` on `az batch pool create` is now true (was false).
 * [BREAKING] The default value for Scope on AutoUserSpecification is now always Pool (was Task on Windows nodes, Pool on Linux nodes). This argument is not exposed via the commandline, but can be set in the `--json-file` arguments.
 
+**VM**
```
yugangw-msft: With one command module model, I suggest have the section called `Compute`, instead of `VM`, which is a bit confusing  to end users

src/azure-cli/azure/cli/command_modules/vm/_params.py
> 
> I suggest we reword the whole comment a bit. I know it is from the swagger, but it makes me feel a bit out of breath to read it :). Maybe `length of time a notification to be sent to the VM on the instance metadata server till the VM gets deleted. `? Also please cross check the `PT5M` is 8061 format?

Yes. It's hard to read. PT5M is an example provided by the swagger file.

> 
> Please add api-version range

OK

qwordy:
> 
> 
> Hi. Just some comments
> 
>     * Now, only the create command support this argument. However, the related issue also ask for update command. Please add this argument into update command.
> 
>     * I cannot understand, how can the customer disable this field after creating it.
> 
>     * At least, please add a test command for this argument.
> 
>     * You can include "Fix #10124", "Fixes #10124" or "Fixed #10124" in your first comment. When this PR is merged, the related issue will be closed automatically.

1. Good idea.
2. update --set command. But it's better to provide a dedicated parameter.
3. OK
4. OK
Thank you for your review.

src/azure-cli/HISTORY.rst
qwordy: > 
> 
> With one command module model, I suggest have the section called `Compute`, instead of `VM`, which is a bit confusing to end users

OK

src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -501,6 +501,13 @@ def load_arguments(self, _):
         c.argument('protect_from_scale_in', arg_type=protection_policy_type, help="Protect the VM instance from scale-in operations.")
         c.argument('protect_from_scale_set_actions', arg_type=protection_policy_type, help="Protect the VM instance from scale set actions (including scale-in).")
 
+    for scope in ['vmss create', 'vmss update']:
+        with self.argument_context(scope) as c:
+            c.argument('terminate_notification', min_api='2019-03-01', arg_type=get_three_state_flag(),
```
MyronFanQiu: Usually we use `enable_terminate_notification`

qwordy: https://docs.microsoft.com/en-us/cli/azure/vm?view=azure-cli-latest#az-vm-create

`vm create` has 3 boolean parameters now. They don't have a enable_ prefix.

`enable_` means you can only enable it. You can not pass true or false.

qwordy: Abbreviation is not urgent. If this parameter is used heavily, we can add it.

MyronFanQiu: `three_flags_arguments` supports user just inputs `--terminate-notification` without any value followed. So it's quite confusing to use this argument like that.

qwordy: I followed the pattern here.
https://github.com/Azure/azure-cli/blob/844005a08d19ab7a5544aa9147869e101426ea67/src/azure-cli/azure/cli/command_modules/vm/_params.py#L495
https://docs.microsoft.com/en-us/cli/azure/vmss?view=azure-cli-latest#az-vmss-create

qwordy: https://docs.microsoft.com/en-us/cli/azure/vmss?view=azure-cli-latest#az-vmss-create
Observation: parameters in same logic group share same prefix.

MyronFanQiu: Well, I think `enable_accelerated_networking` will be better. I discussed with @zikalino about it before. Taking this chance, we can discuss it again and follow the rule in the future.

MyronFanQiu: @zikalino What's your thoughts about `enable_` and `get_three_state_flag()`? We can make a conclusion and  I will update the [command authoring doc](https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md).

qwordy: It's time to escalate.
@yugangw-msft What's the best practice?

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -2048,6 +2050,8 @@ def _get_public_ip_address_allocation(value, sku):
     if secrets:
         secrets = _merge_secrets([validate_file_or_dict(secret) for secret in secrets])
 
+    validate_terminate_notification(terminate_notification, terminate_notification_time)
```
Juliehzl: This validate function is used only for arguments, right? If yes, it should be executed before command handler and you should specify your argument validator in argument context like https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/vm/_params.py#L238.

src/azure-cli/requirements.py2.Darwin.txt
```diff
@@ -26,7 +26,7 @@ azure-mgmt-billing==0.2.0
 azure-mgmt-botservice==0.2.0
 azure-mgmt-cdn==3.1.0
 azure-mgmt-cognitiveservices==5.0.0
-azure-mgmt-compute==6.0.0
+azure-mgmt-compute==7.0.0
```
Juliehzl: Can you declare package upgrade in changelog?

src/azure-cli/azure/cli/command_modules/vm/_params.py
Juliehzl: Just in suggestion, for storage module, we also use `enable_xxx` for such argument to enable a flag or argument.

yonzhan: There is one parameter called "--ultra-ssd-enabled" in below doc:
https://docs.microsoft.com/en-us/cli/azure/vmss?view=azure-cli-latest#az-vmss-create
Maybe you can also follow that pattern and use "terminate-notification-enabled" so that these two parameters can share the same prefix "terminate-notification".
@yugangw-msft can make the call for name convention.

yugangw-msft: let us go with `enable_terminate_notification`. `accelerated_networking` is a questionable naming which I probably should not have named that way.

src/azure-cli/HISTORY.rst
```diff
@@ -21,6 +21,10 @@ Release History
 * [BREAKING] When not specified, the default value for `--start-task-wait-for-success` on `az batch pool create` is now true (was false).
 * [BREAKING] The default value for Scope on AutoUserSpecification is now always Pool (was Task on Windows nodes, Pool on Linux nodes). This argument is not exposed via the commandline, but can be set in the `--json-file` arguments.
 
+**Compute**
+
+* vmss create/update: Add --terminate-notification and --terminate-notification-time parameters to support terminate scheduled event configurability.
```
yugangw-msft: For `create` command, we probably only need `--terminate-notification-time`. as long as the time argument value is provided, CLI should enable the notification. I assume by default the notification is off, right?

src/azure-cli/azure/cli/command_modules/vm/_params.py
qwordy: Thank you for your reviews.

src/azure-cli/azure/cli/command_modules/vm/custom.py
qwordy: It seems reasonable.

src/azure-cli/HISTORY.rst
qwordy: Yes. It is turned off by default. OK.

src/azure-cli/requirements.py2.Darwin.txt
qwordy: > 
> 
> Can you declare package upgrade in changelog?

OK

PR #10398, opened 2019-09-03
Multi-API: add resource type for appservice for consistency

yugangw-msft:

### DIFF
--- a/src/azure-cli-core/HISTORY.rst
+++ b/src/azure-cli-core/HISTORY.rst
@@ -2,6 +2,8 @@
 
 Release History
 ===============
+* Minor changes
+
 2.0.72
 ++++++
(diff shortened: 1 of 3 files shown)


src/azure-cli-core/HISTORY.rst
```diff
@@ -2,6 +2,8 @@
 
 Release History
 ===============
+* Minor changes
```
Juliehzl: Why do you use the minor changes here?

yugangw-msft: for changes related with implementation details which end users are not interested, we use the wording of "Minor changes"

Juliehzl:
I think it is good whenCI check pass.

PR #10443, opened 2019-09-06
Update issue templates - Service Team Support Request, request a link to the pull request

qwordy:


qwordy:
I changed "location" to "pull request". Other modifications are auto-generated.

Juliehzl:
I think "pull request" may be not very suitable here. Because the PR to swagger maybe not created for this feature. The "location" is used to let service team specify where the API for feature is in their service swagger file.

qwordy:
> 
> 
> I think "pull request" may be not very suitable here. Because the PR to swagger maybe not created for this feature. The "location" is used to let service team specify where the API for feature is in their service swagger file.

Yes. How about:

Provide a link to the location of your feature(s) in the REST API specs repo. If your feature(s) has corresponding commit or pull request in the REST API specs repo, provide them. This should be on the master branch of the REST API specs repo.

PR #10449, opened 2019-09-06
Aladdin generated examples for ams, appservice

AladdinBI:


src/azure-cli/azure/cli/command_modules/apim/_help.py
```diff
@@ -5,38 +5,38 @@
...
 helps['apim'] = """
-    type: group
-    short-summary: Manage Azure API Management services.
+type: group
+short-summary: Manage Azure API Management services.
+"""
+
+helps['apim backup'] = """
+type: command
+short-summary: Creates a backup of the API Management service to the given Azure Storage Account. This is long running operation and could take several minutes to complete.
```
qwordy: Creates -> Create

PR #10461, opened 2019-09-06
Aladdin generated examples for sql, storage

AladdinBI:


src/azure-cli/azure/cli/command_modules/storage/_help.py
```diff
@@ -349,6 +349,9 @@
 examples:
   - name: Upload all files that end with .py unless blob exists and has been modified since given date.
     text: az storage blob incremental-copy start --source-container MySourceContainer --source-blob MyBlob --source-account-name MySourceAccount --source-account-key MySourceKey --source-snapshot MySnapshot --destination-container MyDestinationContainer --destination-blob MyDestinationBlob
+  - name: Copies an incremental copy of a blob asynchronously. (autogenerated)
```
Juliehzl: CLI prefer to use active voice in the description. It will be better to use "Copy" not "Copies".

src/azure-cli/azure/cli/command_modules/storage/_help.py
```diff
@@ -1127,6 +1130,9 @@
         end=`date -u -d "30 minutes" '+%Y-%m-%dT%H:%MZ'`
         sas=`az storage share generate-sas -n MyShare --account-name MyStorageAccount --https-only --permissions dlrw --expiry $end -o tsv`
         az storage file upload -s MyShare --account-name MyStorageAccount --source file.txt  --sas-token $sas
+  - name: Generates a shared access signature for the share. (autogenerated)
```
Juliehzl: Generates -> Generate

PR #10462, opened 2019-09-06
Aladdin generated examples for vm

AladdinBI:


src/azure-cli/azure/cli/command_modules/vm/_help.py
```diff
@@ -293,6 +296,10 @@
 helps['image template show'] = """
 type: command
 short-summary: Show an image builder template.
+examples:
+  - name: Show an image builder template (autogenerated)
+    text: az image template show --name mytemplate --resource-group my-group
```
qwordy: To be consistent with other names, I prefer MyTemplate and MyResourceGroup.

PR #10465, opened 2019-09-06
az webapp create ACR support

Nking92:
az webapp create ACR support
---
Adds ACR / private registry support to the webapp create command.


panchagnula:
Thanks @Nking92 for getting this started.

src/azure-cli/azure/cli/command_modules/appservice/_help.py
```diff
@@ -1044,6 +1044,12 @@
   - name: Create a web app with a NodeJS 6.2 runtime and deployed from a local git repository.
     text: >
         az webapp create -g MyResourceGroup -p MyPlan -n MyUniqueAppName --runtime "node|6.2" --deployment-local-git
+  - name: Create a web app with an image from DockerHub.
+    text: >
+        az webapp create -g MyResourceGroup -p MyPlan -n MyUniqueAppName -i nginx
+  - name: Create a web app with an image from a private Azure Container Registry.
```
panchagnula: thanks for adding this - can you also add an example for Private Docker Registry with username & pwd parameters to use?

GetafixIT:
ACR integration still not working well... I am trying using a docker-compose file via the following params:

```bash
az webapp create \
-g $RG \
-p $ASP \
-n $NAME \
-i $TAG \
-s $REG_USER \
-w $REG_PASS \
--multicontainer-config-type "compose" \
--multicontainer-config-file ./docker-compose-deploy.yml
```

With or without -i -s -w flags there are failures regarding ACR credentials.

The only way to get this working is to run the above without -i -s -w flags then update the container config via 'az webapp config container set' and add the repo credentials that way.

Seems clunky. Would be better to have this in one command.

Also there doesn't appear to be a way to make slot deployments with a docker-compose file... we're having to make a new web app deployment to work around. Feels like docker-compose could be better implemented to level up Web Apps IMO.

PR #10473, opened 2019-09-09
POC: redact sensitive storage account info from debug logger

yugangw-msft:
@lmazuel, please let me know whether the patch here is sustainable, otherwise, we need to expose a hook from the pipeline for client applications to redact in anyway they prefer; or any other options you would suggest.
 
CC @Juliehzl

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/storage/_client_factory.py
+++ b/src/azure-cli/azure/cli/command_modules/storage/_client_factory.py
@@ -139,6 +139,9 @@ def cf_sa(cli_ctx, _):
 
 
 def cf_sa_for_keys(cli_ctx, _):
+    from knack.log import get_logger
+    logger = get_logger(__name__)
+    logger.debug('Redact HTTP response body to avoid including storage keys in body')
     client = storage_client_factory(cli_ctx)
     client.config.enable_http_logger = False
     return client.storage_accounts
--- a/src/azure-cli/azure/cli/command_modules/storage/_client_factory.py
+++ b/src/azure-cli/azure/cli/command_modules/storage/_client_factory.py
@@ -3,7 +3,6 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 
-import msrest
 from azure.cli.core.commands.client_factory import get_mgmt_service_client, get_data_service_client
 from azure.cli.core.profiles import ResourceType, get_sdk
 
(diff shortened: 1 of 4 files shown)


src/azure-cli/azure/cli/command_modules/storage/_validators.py
```diff
@@ -40,7 +40,7 @@ def _query_account_key(cli_ctx, account_name):
 
 def _query_account_rg(cli_ctx, account_name):
     """Query the storage account's resource group, which the mgmt sdk requires."""
-    scf = get_mgmt_service_client(cli_ctx, ResourceType.MGMT_STORAGE)
```
Juliehzl: Why not apply `storage_client_factory` for [_query_account_key](https://github.com/Azure/azure-cli/blob/2bc61d23b447585415b6d10611d36bc4c8c4f3a6/src/azure-cli/azure/cli/command_modules/storage/_validators.py#L29)? It looks `_query_account_rg` is used for listing storage accounts and `_query_account_key` is for listing keys.

yugangw-msft: Good question. Sorry this change is not very relevant,  as I wanted to have one less way to create the storage client. That is it. I can undo this if you like

src/azure-cli/azure/cli/command_modules/storage/_client_factory.py
```diff
@@ -138,6 +138,15 @@ def cf_sa(cli_ctx, _):
     return storage_client_factory(cli_ctx).storage_accounts
 
 
+def cf_sa_for_keys(cli_ctx, _):
+    from knack.log import get_logger
+    logger = get_logger(__name__)
+    logger.debug('Disable HTTP logging to avoid having storage keys in debug logs')
+    client = storage_client_factory(cli_ctx)
+    client.config.enable_http_logger = False
```
jiasli: ✔Tested with `storage account keys list -n {} --debug`.

src/azure-cli/azure/cli/command_modules/storage/_validators.py
```diff
@@ -32,6 +32,8 @@ def _query_account_key(cli_ctx, account_name):
     t_storage_account_keys = get_sdk(
         cli_ctx, ResourceType.MGMT_STORAGE, 'models.storage_account_keys#StorageAccountKeys')
 
+    scf.config.enable_http_logger = False
```
jiasli: ✔Tested with `storage blob copy start --source-uri {} --account-name {} -b {} -c {} --debug`

PR #10499, opened 2019-09-10
[ACR] Update some policy params

ankurkhemani:
- Add required `--type` parameter to retention command
 - use `-r` for registry name instead of `-n` for `config` command group.
- separate out `status` params
---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acr/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/acr/_params.py
@@ -79,6 +79,7 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('force', help='Overwrite the existing tag of the image to be imported.', action='store_true')
 
     with self.argument_context('acr config content-trust') as c:
+        c.argument('registry_name', options_list=['--registry', '-r', c.deprecate(target='-n', redirect='-r', hide=True), c.deprecate(target='--name', redirect='--registry', hide=True)])
         c.argument('status', help="Indicates whether content-trust is enabled or disabled.", arg_type=get_enum_type(PolicyStatus))
 
     with self.argument_context('acr config retention') as c:
--- a/src/azure-cli/azure/cli/command_modules/acr/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/acr/_params.py
@@ -85,7 +85,7 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('status', help="Indicates whether retention policy is enabled or disabled.", arg_type=get_enum_type(PolicyStatus))
         c.argument('registry_name', options_list=['--registry', '-r'])
         c.argument('days', type=int, help='The number of days to retain an untagged manifest after which it gets purged (Range: 0 to 365). Value "0" will delete untagged manifests immediately.', validator=validate_retention_days, default=7)
-        c.argument('type', help='The type of retention policy.', default=RetentionType.UntaggedManifests.name, arg_type=get_enum_type(RetentionType))
+        c.argument('policy_type', options_list=['--type'], help='The type of retention policy.', default=RetentionType.UntaggedManifests.name, arg_type=get_enum_type(RetentionType))
 
     with self.argument_context('acr login') as c:
         c.argument('resource_group_name', deprecate_info=c.deprecate(hide=True))
--- a/src/azure-cli/azure/cli/command_modules/acr/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/acr/_params.py
@@ -80,10 +80,10 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
 
     with self.argument_context('acr config content-trust') as c:
         c.argument('registry_name', options_list=['--registry', '-r', c.deprecate(target='-n', redirect='-r', hide=True), c.deprecate(target='--name', redirect='--registry', hide=True)])
-        c.argument('status', help="Indicates whether content-trust is enabled or disabled.", arg_type=get_enum_type(PolicyStatus))
+        c.argument('status', help="Indicates whether content-trust is enabled.", arg_type=get_enum_type(PolicyStatus))
 
     with self.argument_context('acr config retention') as c:
-        c.argument('status', help="Indicates whether retention policy is enabled or disabled.", arg_type=get_enum_type(PolicyStatus))
+        c.argument('status', help="Indicates whether retention policy is enabled.", arg_type=get_enum_type(PolicyStatus))
         c.argument('registry_name', options_list=['--registry', '-r'])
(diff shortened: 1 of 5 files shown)


ankurkhemani:
@djyou @sajayantony Can you please ack?

ankurkhemani:
@yugangw-msft Can you please help with failing CI?

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -77,11 +78,14 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('repository', help='The repository name for a manifest-only copy of images. Multiple copies supported by passing --repository multiple times.', action='append')
         c.argument('force', help='Overwrite the existing tag of the image to be imported.', action='store_true')
 
-    with self.argument_context('acr config') as c:
+    with self.argument_context('acr config content-trust') as c:
         c.argument('status', help="Indicates whether content-trust is enabled or disabled.", arg_type=get_enum_type(PolicyStatus))
 
     with self.argument_context('acr config retention') as c:
+        c.argument('status', help="Indicates whether retention policy is enabled or disabled.", arg_type=get_enum_type(PolicyStatus))
+        c.argument('registry_name', options_list=['--registry', '-r'])
```
djyou: Wouldn't this be inconsistent with other `acr config` commands?

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -77,11 +78,14 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('repository', help='The repository name for a manifest-only copy of images. Multiple copies supported by passing --repository multiple times.', action='append')
         c.argument('force', help='Overwrite the existing tag of the image to be imported.', action='store_true')
 
-    with self.argument_context('acr config') as c:
+    with self.argument_context('acr config content-trust') as c:
         c.argument('status', help="Indicates whether content-trust is enabled or disabled.", arg_type=get_enum_type(PolicyStatus))
 
     with self.argument_context('acr config retention') as c:
+        c.argument('status', help="Indicates whether retention policy is enabled or disabled.", arg_type=get_enum_type(PolicyStatus))
```
djyou: nit: probably can just say whether the policy is enabled to avoid defining it twice.

src/azure-cli/azure/cli/command_modules/acr/_params.py
ankurkhemani: Is that a problem? We need to reserve `--name` to uniquely identify a retention policy names in the future. Not applicable to `content-trust` though. We can make it `-r` for `content-trust` but that would be a breaking change.

src/azure-cli/azure/cli/command_modules/acr/_params.py
adewaleo: You can also use a for loop like

```suggestion
        for cmd_name in ["content-trust", "retention"]:
           with self.argument_context('acr config {}'.format(cmd_name)) as c:
              c.argument('status', help="Indicates whether content-trust is enabled or disabled.", arg_type=get_enum_type(PolicyStatus)) 
```

This would be its own separate block of code from other argument registration such as registry_name

src/azure-cli/azure/cli/command_modules/acr/_params.py
adewaleo: It would be good to keep parameter names consistent. I will suggest using a [deprecation](https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#deprecating-commands-and-arguments).

ankurkhemani: Added deprecation for `-r` and using `-n` instead

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -77,11 +78,15 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
...
-        c.argument('status', help="Indicates whether content-trust is enabled or disabled.", arg_type=get_enum_type(PolicyStatus))
+    with self.argument_context('acr config content-trust') as c:
+        c.argument('registry_name', options_list=['--registry', '-r', c.deprecate(target='-n', redirect='-r', hide=True), c.deprecate(target='--name', redirect='--registry', hide=True)])
+        c.argument('status', help="Indicates whether content-trust is enabled.", arg_type=get_enum_type(PolicyStatus))
 
     with self.argument_context('acr config retention') as c:
+        c.argument('status', help="Indicates whether retention policy is enabled.", arg_type=get_enum_type(PolicyStatus))
+        c.argument('registry_name', options_list=['--registry', '-r'])
         c.argument('days', type=int, help='The number of days to retain an untagged manifest after which it gets purged (Range: 0 to 365). Value "0" will delete untagged manifests immediately.', validator=validate_retention_days, default=7)
+        c.argument('policy_type', options_list=['--type'], help='The type of retention policy.', default=RetentionType.UntaggedManifests.name, arg_type=get_enum_type(RetentionType))
```
djyou: Why do we need this place holder if it's an optional parameter with default value? Can we add it when it's real?

src/azure-cli/azure/cli/command_modules/acr/_params.py
ankurkhemani: @djyou modified string.
@adewaleo nice suggestion, but here we can keep it separate as the help message is different.

src/azure-cli/azure/cli/command_modules/acr/_params.py
ankurkhemani: It's required. We need it since retention will have multiple types in future and don't want to break existing users in the future.

ankurkhemani: Actually, default value needs to be removed. Done.

ankurkhemani:
@yugangw-msft Can you please add this PR to [Sprint 74](https://github.com/Azure/azure-cli/milestone/73)?

src/azure-cli/azure/cli/command_modules/acr/policy.py
```diff
@@ -58,6 +63,7 @@ def acr_config_retention_show(cmd,
 def acr_config_retention_update(cmd,
                                 client,
                                 registry_name,
+                                policy_type,  # pylint: disable=unused-argument
```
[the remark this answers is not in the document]
ankurkhemani: It's a temporary placeholder. We are reserving it for future implementation, currently only type UntaggedManifest is supported and defaulted on the backend.

ankurkhemani:
@yugangw-msft Updated history file. Ready to merge.

Issue #10477, opened 2019-09-09
[DiskRP] Microsoft.Azure.Management.Compute CLI


pollock4595:
**Resource Provider**
DiskRP

**Description of Feature or Work Requested**
We need a Azure CLI support for the new API version for managed disks. This API version change allows direct upload (specifying the exact size of the upload for customers who upload their disks directly) , incremental snapshots for managed disks, and 'encryptionSettingsVersion'. It also returns some previously privileged fields to the user: 'diskSizeBytes', 'uniqueId', and 'sourceUniqueId'.

**Minimum API Version Required**
2019-03-01

**Swagger Link**
https://github.com/Azure/azure-rest-api-specs/blob/master/specification/compute/resource-manager/Microsoft.Compute/stable/2019-03-01/disk.json

Changes from pervious version can be seen here (although the description for 'uploadSizeBytes' has the wrong lower bound under description): https://github.com/Azure/azure-rest-api-specs/compare/1a3b855...4f5dbad

**Target Date**
September 13th

yonzhan:
@qwordy , please take a look and response whether we can finish this by September 13th.

qwordy:
> 
> 
> @qwordy , please take a look and response whether we can finish this by September 13th.

Yes, we can. I think I can provide a initial version on time. However, it may need review and revision.

qwordy:
Hi Joshua, I can't find the definition in Python SDK of the new features.
The Python SDK owner said
> they modified readme, but didn't add the tag to python release

I don't know the specific rules. Just remind if you forget doing something. Thanks.

qwordy:
`CreationData` is also used by snapshot. Does snapshot support `uploadSizeBytes` too?
Disk and snapshot share some common functions in our project.

qwordy:
After experiment, I know uploadSizeBytes must be a multiple of MiB + 512.
But still fail:
```
msrestazure.azure_exceptions.CloudError: Azure Error: InvalidParameter
Message: The value of parameter creationData.uploadSizeBytes is invalid.
Target: creationData.uploadSizeBytes
```
Can you help me with it?

qwordy:
> It also returns some previously privileged fields to the user: 'diskSizeBytes', 'uniqueId', and 'sourceUniqueId'.

This doesn't need modification of CLI.  CLI users can see them as long as the server provides them.

I can set `incremental` property in creation without error. But how can I know whether it is successful? I can't see this property in the returned JSON string.

[...]

PR #10503, opened 2019-09-11
[Compute] support --upload-size in disk creation, support incremental snapshots

qwordy:
Implement #10477 

---


qwordy:
I haven't modify the history file. Just review the changes I have made.

qwordy:
Any ideas of parameter name? --upload-size should be used together with --for-upload.

MyronFanQiu:
> Any ideas of parameter name? --upload-size should be used together with --for-upload.

However about `--blob-size` since you mentioned the `for uploading blobs` in `--for-upload` argument. I think `--upload-size` is a little bit confusing....

ramankumarlive:
@qwordy -upload-size must be called as -upload-size-bytes otherwise customers might enter a value in Gib. Also, the system will accept this parameter only when disk is create --for-upload

PR #10513, opened 2019-09-12
AKS support for SLB+VMSS

xuto2:
---


zikalino:
Looks good in general, but a few comments:
- could we also support public ip name if the same resource group for instance? I think passing is too complex
- I think it would be good to improve documentation and define what's the relationship between new parameters

zikalino:
I think it's ok to merge as is, some improvements could be made in the future

xuto2:
Thanks a lot for the comments, @zikalino . Will take your suggestion into the consideration in the next release(should not be too far).

xuto2:
@qwordy , can you take a look ?

Issue #3989, opened 2017-07-07
No module named azure in pyvenv


jaraco:
### Description

Installing azure-cli to a Python 3 venv produces invalid `az` command.

```
$ python3.6 -m venv ~/azure
$ ~/azure/bin/pip install -q azure-cli
$ ~/azure/bin/python --version        
Python 3.6.1
$ ~/azure/bin/az
/usr/local/opt/python/bin/python2.7: No module named azure
```

Whatever magic the az script is attempting to do is apparently not detecting the virtual environment and is instead invoking an old Python from a different environment where azure is not installed.

It's possible this installation method isn't supported, but as a Python engineer, I would have expected it to work.

---

### Environment summary

**Install Method:** How did you install the CLI? (e.g. pip, interactive script, apt-get, Docker, MSI, nightly)  
Answer here: pip

**CLI Version:** What version of the CLI and modules are installed? (Use `az --version`)  
Answer here: 2.0.10

**OS Version:** What OS and version are you using?  
Answer here: macOS 10.12.5

**Shell Type:** What shell are you using? (e.g. bash, cmd.exe, Bash on Windows)  
Answer here: zsh

troydai:
@derekbekoe

derekbekoe:
`python -m azure.cli --version`

What Python path does that refer to?

Also, does `/usr/local/opt/python/bin/python2.7` contain a different installation of the CLI?

This may be to do with the code in https://github.com/Azure/azure-cli/blob/master/src/azure-cli/az.

troydai:
The `az` script executes the `python` on the PATH, which I assume is `2.7`. To make the az aware of the virtual environment some logics are required to find the correct python executable.

jaraco:
> What Python path does that refer to?

I have `python` aliased to `python3` because I use Python 3 by default even though the system Python is `python2.7`.

```
$ which python
python: aliased to python3
$ which python3
/Library/Frameworks/Python.framework/Versions/3.6/bin/python3
```

Python 3.6 is the stock python.org Python 3.6.1.

The only installation of the Azure CLI is in that virtualenv, so it's not installed to `/usr/local/opt/python/bin/python2.7`.

> This may be to do with the code in az

Yes, almost certainly. I wonder if the invocation should be `/usr/bin/env python -m`. I'll test that.

jaraco:
Using `/usr/bin/env python` doesn't work either. Replacing it with `$DIR/python` does work, however:

```
[...]

Issue #1043, opened 2016-10-06
Support python3 executable when running az


derekbekoe:
The script files, 'az' and 'az.bat', that are installed through the `pip install` do not support machine configurations that only have `python3`.

An error is shown that `python` was not found as it has been named `python 3`.

The script files should potentially have a way to determine the correct python executable to use.

Related to these issues that were closed previously: https://github.com/Azure/azure-cli/issues/1016, https://github.com/Azure/azure-cli/issues/1008

derekbekoe:
cc @tjprescott

derekbekoe:
Related to PEP 394 -- The "python" Command on Unix-Like Systems - https://www.python.org/dev/peps/pep-0394/

tjprescott:
The only team that clamored for this was Datalake, and they were more than happy with the virtual environment solution.  Given that we have packaged released that include Python3... @derekbekoe can we close this issue?

kostrse:
Folks, this issue still persists and annoying...

pmarques:
@derekbekoe this bug annoying specially when dealing with custom stuff like brew for OS X. What is the propose of that bash script wrapper for az?
I'm not a python expert, but if this was a python script (with shebang pointing to python), pip would update the script to use the right python version. For instance:

pip2 install azure-cli would use python2(...)
pip3 install azure-cli would use python3(...)

kwakayama:
> Folks, this issue still persists and annoying...

Should this issue be re-opened?

PR #10516, opened 2019-09-12
Rewrite az in python

suominen:
Use python instead of bash, so that we get benefits from using `#!/usr/bin/env python`:

* `distutils` will replace it with an absolute path to the specific python interpreter being used
  * helps with `pip install` (#1043)
  * helps with tools like `pipx` (pipxproject/pipx#162)
* if running inside a `virtualenv` as is, it will run the python interpreter for that `virtualenv` (#3989)

There doesn't seem to be a measurable startup delay impact as has been reported with using `entry_points/console_script` (#7469).

---


suominen:
I've tested it with
* Python 2.7.16 and 3.7.3 on Debian 10.1 (buster).
* Python 2.7.10 and 3.7.4 on macOS 10.14.6 (Mojave).

suominen:
@fengzhou-msft - Didn't mean to dismiss any reviews, but just keep this branch without merge conflicts. Maybe it is better I don't resolve conflicts going forward -- it is just about `HISTORY.rst` changing in `dev`.

yonzhan:
@fengzhou-msft  please help deal with this PR.

src/azure-cli/az
```diff
@@ -1,14 +1,16 @@
-#!/usr/bin/env bash
+#!/usr/bin/env python
```
haroldrandom: Any earnings changing from bash to python?

suominen: I listed the benefits when opening this pull request: https://github.com/Azure/azure-cli/pull/10516#issue-316767937

Does that answer your question, or did you mean something else by "earnings"?

fengzhou-msft:
@yonzhan I have tested the code, it works as stated and solved several issues. This change impacts the entry point for running all commands when installed with pip. We need @zikalino to review and see if there's any drawbacks.

yonzhan:
@zikalino  please review this PR.

PR #10517, opened 2019-09-12
Adding buster

zikalino:
modifying pipelines to add buster


yonzhan:
@zikalino looks good.

yugangw-msft:
@zikalino, i believe release pipeline will need t be adjusted, particularly in the `Publish Debian Packages` task, there is a dictionary with all debian distros we need to add a new one.
I know you are already testing this, but in case it helps

jiasli:
Per https://wiki.debian.org/DebianReleases, Debian Buster reached EOL on 2024-06-30.

Issue #10478, opened 2019-09-09
Update Azure Compute VM/VMSS cmdlets with the changes in the PR "Added priority, eviction policy and billing profile to VirtualMachineProperties. Also, adding optional billing profile to VMSS Create and Update APIs"


rakeshakkera:
**Resource Provider**
What is the Azure resource provider your feature is part of?
Azure Compute

**Description of Feature or Work Requested**
Ability to create low-priority standalone VMs with variable pricing. Also adding variable pricing to low priority VMSS.

**Minimum API Version Required**
What is the minimum API version of your service required to implement your feature?
2019-03-01

**Swagger Link**
Provide a link to the location of your feature(s) in the REST API specs repo. This should be on the master branch
of the REST API specs repo.
PR: https://github.com/Azure/azure-rest-api-specs/commit/fb8f078deac2367f3dbcfbd8a542c77083c28a5b

**Target Date**
If you have a target date for release of this feature/work, please provide it. While we can't guarantee these dates,
it will help us prioritize your request against other requests.
09/24/2019

yonzhan:
@qwordy , please take a look and response whether we can finish this by this sprint.

qwordy:
> 
> 
> @qwordy , please take a look and response whether we can finish this by this sprint.

My answer is yes by my estimation

qwordy:
PR has been merged. Close.

PR #10536, opened 2019-09-16
[Compute] Update azure-mgmt-compute version to 8.0.0. vm/vmss create: Support --priority, --eviction-policy, --max-billing parameters. disk create: Allow specifying the exact size of the upload for customers who upload their disks directly. snapshot create: Support incremental snapshots for managed disks. (Issue #10478 & #10477)

qwordy:
[Compute] Issue #10478 & #10477
* Update azure-mgmt-compute version to 8.0.0.
* Update disk API version to 2019-03-01
* Update snapshot API version to 2019-03-01
* vm/vmss create: Support --priority, --eviction-policy, --max-billing parameters.
* disk create: Allow specifying the exact size of the upload for customers who upload their disks directly.
* snapshot create: Support incremental snapshots for managed disks.

---


src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -115,13 +115,17 @@ def load_arguments(self, _):
...
+        c.argument('upload_size_byte', type=int, min_api='2019-03-01',
+                   help='The size (in bytes) of the contents of the upload including the VHD footer. Min value: 20972032. Max value: 35183298347520')
     # endregion
 
     # region Snapshots
     with self.argument_context('snapshot', resource_type=ResourceType.MGMT_COMPUTE, operation_group='snapshots') as c:
         c.argument('snapshot_name', existing_snapshot_name, id_part='name', completer=get_resource_name_completion_list('Microsoft.Compute/snapshots'))
         c.argument('name', arg_type=name_arg_type)
         c.argument('sku', arg_type=snapshot_sku)
+        c.argument('incremental', arg_type=get_three_state_flag(),
```
yugangw-msft: need `api-version` guard?

src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -268,6 +272,9 @@ def load_arguments(self, _):
         c.argument('disk_caching', nargs='*', help="Use singular value to apply across, or specify individual disks, e.g. 'os=ReadWrite 0=None 1=ReadOnly' should enable update os disk and 2 data disks")
 
     with self.argument_context('vm create') as c:
+        VMPriorityTypes = self.get_models('VirtualMachinePriorityTypes', resource_type=ResourceType.MGMT_COMPUTE)
+        VirtualMachineEvictionPolicyTypes = self.get_models('VirtualMachineEvictionPolicyTypes',
```
yugangw-msft: make sure this won't throw under older api-version that the `VirtualMachinePriorityTypes` is unavailable

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -262,13 +262,17 @@ def create_managed_disk(cmd, resource_group_name, disk_name, location=None,  # p
     else:
         option = DiskCreateOption.empty
 
+    if upload_size_byte is not None and for_upload is not True:
```
yugangw-msft: When `for_upload` is None, should we default to `True`?

src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -115,13 +115,17 @@ def load_arguments(self, _):
         c.argument('os_type', arg_type=get_enum_type(OperatingSystemTypes), help='The Operating System type of the Disk.')
         c.argument('disk_iops_read_write', type=int, min_api='2018-06-01', help='The number of IOPS allowed for this disk. Only settable for UltraSSD disks. One operation can transfer between 4k and 256k bytes')
         c.argument('disk_mbps_read_write', type=int, min_api='2018-06-01', help="The bandwidth allowed for this disk. Only settable for UltraSSD disks. MBps means millions of bytes per second with ISO notation of powers of 10")
+        c.argument('upload_size_byte', type=int, min_api='2019-03-01',
```
zikalino: i think better name should be "upload_size_bytes"

src/azure-cli/azure/cli/command_modules/vm/_params.py
zikalino: I think we are running some old tests, so that should be covered

src/azure-cli/azure/cli/command_modules/vm/_params.py
qwordy: I should add it.

src/azure-cli/azure/cli/command_modules/vm/_params.py
qwordy: Agree with Zim.
It's better to know the exact behavior when can't load model from a specific API version.

src/azure-cli/azure/cli/command_modules/vm/_params.py
qwordy: OK

src/azure-cli/azure/cli/command_modules/vm/custom.py
qwordy: When `for_upload` is None, it means this disk is not created for upload.

src/azure-cli/azure/cli/command_modules/vm/_params.py
qwordy: Add a if statement, then get models
```
if self.supported_api_version(min_api='2019-03-01', resource_type=ResourceType.MGMT_COMPUTE):
```

PR #10418, opened 2019-09-04
Fix #8840: Tenant domain name is not accepted by Key Vault

jiasli:
Fix #8840: Convert tenant domain name to GUID if it is not.

PR #10547, opened 2019-09-17
Only resolve tenant domain name when not None

jiasli:
Patch #10418: Fix broken `az login`. Only resolve tenant domain name when `not None`.


yugangw-msft:
I merged it. This reveals a gap in our unit test which we should fill soon.

PR #10581, opened 2019-09-19
additional fix for sql

zikalino:


scripts/release/homebrew/docker/run.sh
```diff
@@ -7,7 +7,7 @@ pip install -U pip
 pip install -r $root/requirements.txt
 pip install azure-cli==$CLI_VERSION -f /mnt/pypi
 
-pip install msrestazure==0.4.34
+pip install msrestazure==0.6.2
```
qwordy: Try to delete this line in next sprint if possible. We should define package versions in requirements.txt.

PR #10583, opened 2019-09-19
webapp: Updating webapp up to not use F1 as the default

panchagnula:
---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -2766,7 +2766,7 @@ def get_history_triggered_webjob(cmd, resource_group_name, name, webjob_name, sl
     return client.web_apps.list_triggered_web_job_history(resource_group_name, name, webjob_name)
 
 
-def webapp_up(cmd, name, resource_group_name=None, plan=None, location=None, sku='F1', dryrun=False, logs=False,  # pylint: disable=too-many-statements,
+def webapp_up(cmd, name, resource_group_name=None, plan=None, location=None, sku=None, dryrun=False, logs=False,  # pylint: disable=too-many-statements,
               launch_browser=False):
     import os
     src_dir = os.getcwd()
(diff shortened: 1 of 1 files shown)


panchagnula:
@qwordy can i get  a PR on this please?

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2820,8 +2820,7 @@ def webapp_up(cmd, name, resource_group_name=None, plan=None, location=None, sku
         site_config = client.web_apps.get_configuration(rg_name, name)
     else:  # need to create new app, check if we need to use default RG or use user entered values
         logger.warning("webapp %s doesn't exist", name)
-        sku_value = get_sku_to_use(src_dir, sku)
-        sku = get_sku_name(sku_value)
+        sku = get_sku_to_use(src_dir, sku)
```
qwordy: Why do you switch from sku name to sku value

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2838,7 +2837,8 @@ def webapp_up(cmd, name, resource_group_name=None, plan=None, location=None, sku
                 "version_detected": "%s",
                 "runtime_version": "%s"
                 }
-                """ % (name, plan, rg_name, sku, os_name, loc, _src_path_escaped, detected_version, runtime_version)
+                """ % (name, plan, rg_name, get_sku_name(sku), os_name, loc, _src_path_escaped, detected_version,
+                       runtime_version)
```
qwordy: Why do you switch from sku value to sku name

src/azure-cli/azure/cli/command_modules/appservice/custom.py
panchagnula: if you see line 2824 - we initially did this on a separate line but we need this only when we display the SKU - to use SKU name otherwise in the commands we use the SKU value - so made that change

src/azure-cli/azure/cli/command_modules/appservice/custom.py
panchagnula: Same comment as above.

PR #10604, opened 2019-09-20
Aladdin generated examples for vm

AladdinBI:


src/azure-cli/azure/cli/command_modules/vm/_help.py
```diff
@@ -54,6 +54,9 @@
   - name: Delete a managed disk. (autogenerated)
     text: az disk delete --name MyManagedDisk --resource-group MyResourceGroup
     crafted: true
+  - name: Delete a managed disk. (autogenerated)
+    text: az disk delete --name MyManagedDisk --resource-group MyResourceGroup
+    crafted: true
```
qwordy: Duplicate examples

PR #10619, opened 2019-09-24
Update CI. Increase timeout for Run Automation, Python 2/3

qwordy:
Update CI. Increase timeout for Run Automation, Python 2/3

I have seen this timeout error more than once. The normal lasting time for this task is 50+ minutes. Sometimes, it’s over 60 minutes. We use a default timeout, which is 60 minutes. It’s not enough. We are adding more and more test cases, so we need a larger timeout.

```
Run Automation, Python 3
2 error(s), 0 warning(s)
The job running on agent Azure Pipelines 47 has exceeded the maximum execution time of 60. For more information, see https://go.microsoft.com/fwlink/?linkid=2077134
The operation was canceled. 
```

---

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -332,6 +332,7 @@ jobs:
 - job: RunAutomationFullPython2
   displayName: Run Automation, Python 2
   dependsOn: BuildPythonWheel
+  timeoutInMinutes: 90
 
   pool:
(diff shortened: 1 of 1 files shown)


qwordy:
> 
> 
> I suggest in future we pick up newer azdev and leverage the diff functionality so to only run commands from the module being updated

Strongly agree. Can we add machines to reduce running time?

PR #10622, opened 2019-09-24
[Monitor]Support "--workspace" for "vm create"

MyronFanQiu:
---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/vm/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_params.py
@@ -261,7 +261,7 @@ def load_arguments(self, _):
         c.argument('nsg', help='The name to use when creating a new Network Security Group (default) or referencing an existing one. Can also reference an existing NSG by ID or specify "" for none.', arg_group='Network')
         c.argument('nsg_rule', help='NSG rule to create when creating a new NSG. Defaults to open ports for allowing RDP on Windows and allowing SSH on Linux.', arg_group='Network', arg_type=get_enum_type(['RDP', 'SSH']))
         c.argument('application_security_groups', min_api='2017-09-01', nargs='+', options_list=['--asgs'], help='Space-separated list of existing application security groups to associate with the VM.', arg_group='Network')
-
+        c.argument('workspace', arg_group='Monitor', help='Name or ID of Log Analytics Workspace. If name is given, the workspace should be in the same resource group with the vm, otherwise a new workspace will be created.')
     with self.argument_context('vm capture') as c:
         c.argument('overwrite', action='store_true')
 
--- a/src/azure-cli/azure/cli/command_modules/vm/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_params.py
@@ -261,7 +261,7 @@ def load_arguments(self, _):
         c.argument('nsg', help='The name to use when creating a new Network Security Group (default) or referencing an existing one. Can also reference an existing NSG by ID or specify "" for none.', arg_group='Network')
         c.argument('nsg_rule', help='NSG rule to create when creating a new NSG. Defaults to open ports for allowing RDP on Windows and allowing SSH on Linux.', arg_group='Network', arg_type=get_enum_type(['RDP', 'SSH']))
         c.argument('application_security_groups', min_api='2017-09-01', nargs='+', options_list=['--asgs'], help='Space-separated list of existing application security groups to associate with the VM.', arg_group='Network')
-        c.argument('workspace', arg_group='Monitor', help='Name or ID of Log Analytics Workspace. If name is given, the workspace should be in the same resource group with the vm, otherwise a new workspace will be created.')
+        c.argument('workspace', arg_group='Monitor', help='Name or ID of Log Analytics Workspace. If you specify the workspace through its name, the workspace should be in the same resource group with the vm, otherwise a new workspace will be created.')
     with self.argument_context('vm capture') as c:
         c.argument('overwrite', action='store_true')
 
--- a/src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
@@ -884,18 +884,16 @@ def build_vm_mmaExtension_resource(_, vm_name, location, workspace_id):
             "typeHandlerVersion": "1.4",
             "autoUpgradeMinorVersion": 'true',
             "settings": {
-                "workspaceId": "",
+                "workspaceId": "[reference(parameters('workspaceId'), '2015-11-01-preview').customerId]",
                 "stopOnMultipleConnections": 'true'
             },
             "protectedSettings": {
-                "workspaceKey": ""
+                "workspaceKey": "[listKeys(parameters('workspaceId'), '2015-11-01-preview').primarySharedKey]"
             }
         }
     }
     mmaExtension_resource['name']= vm_name + '/OMSExtension'
     mmaExtension_resource['location'] = location
-    mmaExtension_resource['properties']['settings']['workspaceId'] = "[reference({0}, '2015-11-01-preview').customerId]".format(workspace_id)
-    mmaExtension_resource['properties']['protectedSettings']['workspaceKey'] = "[listKeys({0}, '2015-11-01-preview').primarySharedKey]".format(workspace_id)
     mmaExtension_resource['dependsOn'] = ['Microsoft.Compute/virtualMachines/' + vm_name]
     return mmaExtension_resource
 
--- a/src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
@@ -873,3 +873,44 @@ def build_av_set_resource(cmd, name, location, tags, platform_update_domain_coun
         av_set['properties']['proximityPlacementGroup'] = {'id': proximity_placement_group}
 
     return av_set
+
+def build_vm_mmaExtension_resource(_, vm_name, location, workspace_id):
+    mmaExtension_resource = {
+        "type": "Microsoft.Compute/virtualMachines/extensions",
+        "apiVersion": "2018-10-01",
+        "properties": {
+            "publisher": "Microsoft.EnterpriseCloud.Monitoring",
+            "type": "OmsAgentForLinux",
+            "typeHandlerVersion": "1.4",
+            "autoUpgradeMinorVersion": 'true',
+            "settings": {
+                "workspaceId": "",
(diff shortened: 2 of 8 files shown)


src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
```diff
@@ -862,3 +862,42 @@ def build_av_set_resource(cmd, name, location, tags, platform_update_domain_coun
...
+        "properties": {
+                "publisher": "Microsoft.Azure.Monitoring.DependencyAgent",
+                "type": "DependencyAgentLinux",
+                "typeHandlerVersion": "9.5",
+                "autoUpgradeMinorVersion": "true"
+        }
+    }
+    daExtensionName_resource['name']= vm_name + '/DependencyAgentLinux'
+    daExtensionName_resource['location'] = location
+    daExtensionName_resource['dependsOn'] = ['Microsoft.Compute/virtualMachines/' + vm_name + '/extensions/OMSExtension']
```
qwordy: can combine these 3 lines to variable definition

src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
```diff
@@ -862,3 +862,42 @@ def build_av_set_resource(cmd, name, location, tags, platform_update_domain_coun
...
+    return daExtensionName_resource
```
qwordy: Unnecessary variable

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -966,6 +976,33 @@ def update_vm(cmd, resource_group_name, vm_name, os_disk=None, disk_caching=None
     return sdk_no_wait(no_wait, _compute_client_factory(cmd.cli_ctx).virtual_machines.create_or_update,
                        resource_group_name, vm_name, **kwargs)
 
+
+def _prepare_workspace(cmd, resource_group_name, workspace):
+    from msrestazure.tools import parse_resource_id, is_valid_resource_id
+    from ._client_factory import cf_log_analytics
+    from azure.cli.core.commands.client_factory import get_subscription_id
+
+    workspace_id = None
```
qwordy: no effect

src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
```diff
@@ -862,3 +862,42 @@ def build_av_set_resource(cmd, name, location, tags, platform_update_domain_coun
...
+                "stopOnMultipleConnections": 'true'
+            },
+            "protectedSettings": {
+                "workspaceKey": "[listKeys(parameters('workspaceId'), '2015-11-01-preview').primarySharedKey]"
+            }
+        }
+    }
+    mmaExtension_resource['name']= vm_name + '/OMSExtension'
+    mmaExtension_resource['location'] = location
+    mmaExtension_resource['dependsOn'] = ['Microsoft.Compute/virtualMachines/' + vm_name]
```
qwordy: can combine these 3 lines to variable definition

src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
```diff
@@ -862,3 +862,42 @@ def build_av_set_resource(cmd, name, location, tags, platform_update_domain_coun
...
+    return mmaExtension_resource
```
qwordy: Unnecessary variable

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -967,6 +977,35 @@ def update_vm(cmd, resource_group_name, vm_name, os_disk=None, disk_caching=None
...
+    if not is_valid_resource_id(workspace):
+        workspace_name = workspace
+        subscription_id = get_subscription_id(cmd.cli_ctx)
+        log_client = cf_log_analytics(cmd.cli_ctx, subscription_id)
+        workspace_result = None
+        try:
+            workspace_result = log_client.get(resource_group_name, workspace_name)
+        except:  # pylint: disable=bare-except
```
Juliehzl: In this way, you will catch all kinds of exceptions. Why do you want to do this? If you just want to do a different thing, why not use condition to do what you want?

src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -261,7 +261,7 @@ def load_arguments(self, _):
         c.argument('nsg', help='The name to use when creating a new Network Security Group (default) or referencing an existing one. Can also reference an existing NSG by ID or specify "" for none.', arg_group='Network')
         c.argument('nsg_rule', help='NSG rule to create when creating a new NSG. Defaults to open ports for allowing RDP on Windows and allowing SSH on Linux.', arg_group='Network', arg_type=get_enum_type(['RDP', 'SSH']))
         c.argument('application_security_groups', min_api='2017-09-01', nargs='+', options_list=['--asgs'], help='Space-separated list of existing application security groups to associate with the VM.', arg_group='Network')
-
+        c.argument('workspace', arg_group='Monitor', help='Name or ID of Log Analytics Workspace. If name is given, the workspace should be in the same resource group with the vm, otherwise a new workspace will be created.')
```
Juliehzl: Do you mean "name or ID is given"? If yes, "If the parameter/argument is given" may be better??

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -967,6 +977,35 @@ def update_vm(cmd, resource_group_name, vm_name, os_disk=None, disk_caching=None
...
+            from azure.mgmt.loganalytics.models import Workspace, Sku
+            sku = Sku(name="pergb2018")
```
qwordy:
```
class SkuNameEnum(str, Enum):

    free = "Free"
    standard = "Standard"
    premium = "Premium"
    unlimited = "Unlimited"
    per_node = "PerNode"
    per_gb2018 = "PerGB2018"
    standalone = "Standalone"
```
I find a class in azure/mgmt/loganalytics/models/log_analytics_management_client_enums.py.
Avoid using magic numbers and string literals in your code.

src/azure-cli/azure/cli/command_modules/vm/tests/latest/test_vm_commands.py
```diff
@@ -1176,6 +1176,19 @@ def test_vm_create_none_options(self, resource_group):
...
+    @ResourceGroupPreparer(name_prefix='cli_test_vm_create_with_monitor', location='centralus')
+    def test_vm_create_with_monitor(self, resource_group):
+
+        self.kwargs.update({
+            'vm': 'monitorvm',
+            'workspace': 'vmlogworkspace20191009',
+            'rg': resource_group
+        })
+
+        self.cmd('vm create -n {vm} -g {rg} --image UbuntuLTS --workspace {workspace}')
```
qwordy: No assertion.

src/azure-cli/azure/cli/command_modules/vm/tests/latest/test_vm_commands.py
```diff
@@ -1176,6 +1176,19 @@ def test_vm_create_none_options(self, resource_group):
         ])
         self.cmd('network public-ip show -n {vm}PublicIP -g {rg}', expect_failure=True)
 
+class VMCreateMonitorTest(ScenarioTest):  # pylint: disable=too-many-instance-attributes
+
+    @ResourceGroupPreparer(name_prefix='cli_test_vm_create_with_monitor', location='centralus')
```
qwordy: too-many-instance-attributes (R0902):
Too many instance attributes (%s/%s) Used when class has too many instance attributes, try to reduce this to get a simpler (and so easier to use) class.
Do we need it?

src/azure-cli/azure/cli/command_modules/vm/_params.py
MyronFanQiu: no. If user give ID, it will try to connect the workspace. If user give name, it will check existence under same resource group. otherwise, create a new one.

src/azure-cli/azure/cli/command_modules/vm/custom.py
MyronFanQiu: I will delete the magic string. Actually, server can only support pergb2018. Other SKU are disabled by server.

src/azure-cli/azure/cli/command_modules/vm/tests/latest/test_vm_commands.py
MyronFanQiu: Good suggestion. I think we don't need this line.

src/azure-cli/azure/cli/command_modules/vm/_params.py
Juliehzl: In this way, it will be more clear to use "If you specify the workspace through its name, your workspace should be in the same resource group with vm...." Just for suggestion, you can change it as you like or keep your existing one.

Juliehzl: Another question for workspace for monitoring is when you need to create workspace for users, why not let user specify location, sku and other parameters like you do in CRUD command?

MyronFanQiu: the original design is to let user input an existed workspace instead of creating a new one. It's a vm command, so we only want to support day 0 scenario.

Juliehzl: To create a workspace, users need to use monitor command, right? Is it known for users?

MyronFanQiu: I don't think we should provide such information to user. For any commands which need vnet, we don't tell use to use "az network vnet create" to create a vnet. Log Analytics Workspace is a hint for user. If they don't know anything, we will create a default workspace for them.

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -689,8 +699,8 @@ def create_vm(cmd, vm_name, resource_group_name, image=None, size='Standard_DS1_
     if no_wait:
         return sdk_no_wait(no_wait, client.create_or_update,
                            resource_group_name, deployment_name, properties)
-    LongRunningOperation(cmd.cli_ctx)(sdk_no_wait(no_wait, client.create_or_update,
-                                                  resource_group_name, deployment_name, properties))
+    LongRunningOperation(cmd.cli_ctx)(client.create_or_update(resource_group_name, deployment_name, properties))
```
qwordy: sdk_no_wait has a no_wait parameter. no_wait is false here because of the above lines `if no_wait: return`. So it's a redundant code.

PR #10632, opened 2019-09-25
[REST] Fix #10268: Update rest --body help message

### LINKED ISSUE #10268: "az rest": allow the HTTP request body to be taken from a file (similar to curl's "--data @filename")
When using `az rest` under Windows, if you need to pass a large body you quickly hit the length limit for a single command.

I'd like to be able to put the body in a file and pass a reference to that file instead of the the actual body to the `az rest` command, e.g. 
`az rest -m PUT -u https://management.azure.com/bla --body @filename`

The alternative is to use curl instead but it requires obtaining the subscription ID and the authentication token beforehand:
```
az account show --query id
az account get-access-token --query accessToken
curl -X PUT "https://management.azure.com/subscriptions/%SUBSCRIPTION_ID%/..." -H "Authorization: Bearer %TOKEN%" -H "Content-Type: application/json" --data @filename

```
**Additional context**
According to the doc, headers and query params can already be loaded from a file using the @ notation, so it would seem natural to extend this to the body.

jiasli:
Fix #10268: Update `az rest --body` help message.


Juliehzl:
Hi @jiasli  can you write a test for your new argument?

PR #10634, opened 2019-09-25
set tabulate version to be exactly 0.8.3

fengzhou-msft:
Latest 0.8.4 errors when installing. https://bitbucket.org/astanin/python-tabulate/issues/182/tabulate-fails-to-install-on-windows

Locking to 0.8.3.

### DIFF
--- a/src/azure-cli/setup.py
+++ b/src/azure-cli/setup.py
@@ -149,7 +149,8 @@ DEPENDENCIES = [
     'websocket-client~=0.56.0',
     'xmltodict~=0.12',
     'javaproperties==0.5.1',
-    'jsondiff==1.2.0'
+    'jsondiff==1.2.0',
+    'tabulate==0.8.3'
 ]
 
 with open('README.rst', 'r', encoding='utf-8') as f:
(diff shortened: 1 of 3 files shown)


src/azure-cli/setup.py
```diff
@@ -149,7 +149,8 @@
     'websocket-client~=0.56.0',
     'xmltodict~=0.12',
     'javaproperties==0.5.1',
-    'jsondiff==1.2.0'
+    'jsondiff==1.2.0',
+    'tabulate==0.8.3'
```
yugangw-msft: I thought we will fix it in knack, or am I missing anything?

fengzhou-msft: Fix in knack is a better solution, so we don't need to put a tabulate version wherever knack appears. The fix here is just for a quick solution within our control. Since tabulate released 0.8.5 and fixed their issue, we could abandon the quick fix here and do it in knack.

fengzhou-msft:
tabulate is imported by knack, set its version in knack is a better solution. will do it in knack.

src/azure-cli/setup.py
yonzhan: Agree.

Issue #10156, opened 2019-08-05
VM/VMSS: Referencing VMSS during VM creation


shpa-microsoft:
**Resource Provider**
What is the Azure resource provider your feature is part of?
- Compute

**Description of Feature or Work Requested**
Provide a brief description of the feature or work requested. A link to conceptual documentation may be helpful too.
- Converged API VMSS now allows usage of VMSS in a way similar to availability Sets. Therefore we now need to reference VMSS instead of av-set during VM creation for clients who want to use this. Find internal API docs here: https://microsoft-my.sharepoint.com/%3Aw%3A/g/personal/shpa_microsoft_com1/Ef68QZMoy9BIpvnmq8RDeB8BoSvADf3afeUadvvs49MlBQ?e=gmWhky 
- https://microsoft-my.sharepoint.com/%3Aw%3A/g/personal/shpa_microsoft_com1/Ea1j9UXmUA5OnMul_N4fFrABjyLHBNjayNArtj3vRgSvCA?e=Xwrx0d

**Minimum API Version Required**
What is the minimum API version of your service required to implement your feature?
- 2019-03-01

**Swagger Link**
Provide a link to the location of your feature(s) in the REST API specs repo. This should be on the master branch of the REST API specs repo.
- https://github.com/Azure/azure-rest-api-specs/pull/6836 (needs to be merged yet)

**Target Date**
If you have a target date for release of this feature/work, please provide it. While we can't guarantee these dates,
it will help us prioritize your request against other requests.

qwordy:
We have received you feature request.

shpa-microsoft:
Need traction on this. Tried reaching out to whoever I could find. Perhaps posting here will do the job. @qwordy , @Juliehzl , @yonzhan Please help. We need this for Ignite too and it's been open for a very long time. Thank you! :)

qwordy:
> 
> 
> Need traction on this. Tried reaching out to whoever I could find. Perhaps posting here will do the job. @qwordy , @Juliehzl , @yonzhan Please help. We need this for Ignite too and it's been open for a very long time. Thank you! :)

OK. We will prioritize this issue. I have opened a PR, but it has problems. Do you know how to create a valid vmss in Azure CLI to be referenced in `vm create`? vmss of new type?

shpa-microsoft:
Yes, when you create a VMSS, please do not provide any VM profile in the template and that should be acceptable for this one. More details in the email thread I started.

yonzhan:
[...]

PR #10646, opened 2019-09-26
[Compute] (1) vm create: vm create: Add --vmss to specify an existing virtual machine scale set that the virtual machine should be assigned to. (2) vmss create: Add --orchestrator to specify how virtual machines are managed by the scale set.

qwordy:
* vm create: Add --vmss to specify an existing virtual machine scale set that the virtual machine should be assigned to.
* vmss create: Add --orchestrator to specify how virtual machines are managed by the scale set.

Implement #10156 

---


qwordy:
Testing has some problems now.

src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
```diff
@@ -373,6 +373,9 @@ def _build_storage_profile():
     if availability_set_id:
         vm_properties['availabilitySet'] = {'id': availability_set_id}
 
+    if vmss is not None:
+        vm_properties['virtualMachineScaleSet'] = {'id': vmss}
```
MyronFanQiu: Looks like you just support `ID`

src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -276,6 +276,7 @@ def load_arguments(self, _):
         c.argument('vm_name', name_arg_type, id_part=None, help='Name of the virtual machine.', completer=None)
         c.argument('os_disk_size_gb', type=int, help='the size of the os disk in GB', arg_group='Storage')
         c.argument('availability_set', help='Name or ID of an existing availability set to add the VM to. None by default.')
+        c.argument('vmss', help='Virtual machine scale set that the virtual machine should be assigned to.')
```
MyronFanQiu: Name or ID of Virtual machine scale set that the virtual machine should be assigned to.

src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
qwordy: > 
> 
> Looks like you just support `ID`

I copy the above line. I have ask this in email.

src/azure-cli/azure/cli/command_modules/vm/_params.py
> 
> Name or ID of Virtual machine scale set that the virtual machine should be assigned to.

Good

yonzhan:
@qwordy  what is the status of this PR right now ?

qwordy:
> 
> 
> @qwordy what is the status of this PR right now ?

Use case problem. 
And after many PR were merged, conflicts sprang up.
I need to fix both of them.

src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
qwordy: I have supported name now.

qwordy:
@shpa-microsoft, please help review. Thanks.

qwordy:
> 
> 
> > @qwordy what is the status of this PR right now ?
> 
> Use case problem.
> And after many PR were merged, conflicts sprang up.
> I need to fix both of them.

Solved.

src/azure-cli/HISTORY.rst
```diff
@@ -25,6 +25,8 @@ Release History
 **Compute**
 
 * vm create: Add warning when specifying accelerated networking and an existing NIC together.
+* vm create: Add --vmss to specify virtual machine scale set that the virtual machine should be assigned to.
+* vmss create: Add --empty to create a VMSS without VM profile, which can be then referenced in `vm create`.
```
shpa-microsoft: Empty keyword will give a wrong indication IMO cause one can create an empty VMSS even when VM profile exists. We have started to use orchestrationMode to specify this. The 2 modes are: VMSS and VM. I'd also seek feedback from Jun and Varun.

shpa-microsoft: Also vmss create should happen before vm create in this case. Not sure ordering matters in the doc.

src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -276,6 +276,7 @@ def load_arguments(self, _):
         c.argument('vm_name', name_arg_type, id_part=None, help='Name of the virtual machine.', completer=None)
         c.argument('os_disk_size_gb', type=int, help='the size of the os disk in GB', arg_group='Storage')
         c.argument('availability_set', help='Name or ID of an existing availability set to add the VM to. None by default.')
+        c.argument('vmss', help='Name of ID of a virtual machine scale set that the virtual machine should be assigned to. None by default.')
```
shpa-microsoft: Just like that mentioned in avset. We need to specify "existing"

src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -496,6 +497,7 @@ def load_arguments(self, _):
                    help="The eviction policy for virtual machines in a low priority scale set.", is_preview=True)
         c.argument('application_security_groups', resource_type=ResourceType.MGMT_COMPUTE, min_api='2018-06-01', nargs='+', options_list=['--asgs'], help='Space-separated list of existing application security groups to associate with the VM.', arg_group='Network', validator=validate_asg_names_or_ids)
         c.argument('computer_name_prefix', help='Computer name prefix for all of the virtual machines in the scale set. Computer name prefixes must be 1 to 15 characters long')
+        c.argument('empty', help='Create a VMSS without VM profile, which can be then referenced in `vm create`.', arg_type=get_three_state_flag())
```
shpa-microsoft: Nit: Creates 
Additional: If we change the term to orchestration mode, then all this description will change.

src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
```diff
@@ -387,6 +387,10 @@ def _build_storage_profile():
     if availability_set_id:
         vm_properties['availabilitySet'] = {'id': availability_set_id}
 
+    # vmss is ID
+    if vmss is not None:
+        vm_properties['virtualMachineScaleSet'] = {'id': vmss}
```
shpa-microsoft: How are we supporting just the name? We need full ARM ID of VMSS when it hits the rest layer. So if you guys are converting that to ARM ID, this is fine. Or else follow what avset does.

src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
```diff
@@ -847,6 +851,22 @@ def build_vmss_resource(cmd, name, naming_prefix, location, tags, overprovision,
     }
     if zones:
         vmss['zones'] = zones
+    # vmss without vm profile
+    if empty:
+        if platform_fault_domain_count is None:
+            platform_fault_domain_count = 2
```
shpa-microsoft: Why is it defaulted to 2? Can this not be passed down as null?

src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
qwordy: I handle this in _validate_vm_create_vmss(). This function can transform name to ID. I wrote examples in test cases.

src/azure-cli/HISTORY.rst
qwordy: > 
> 
> Also vmss create should happen before vm create in this case. Not sure ordering matters in the doc.

This doesn't matter. It's just change log.

qwordy: > 
> 
> Empty keyword will give a wrong indication IMO cause one can create an empty VMSS even when VM profile exists. We have started to use orchestrationMode to specify this. The 2 modes are: VMSS and VM. I'd also seek feedback from Jun and Varun.

I think "empty" is ambiguous, too. Do you have a suggested name?
Is there a change of vm and vmss creation ways in the future? You mentioned orchestrationMode.

src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
qwordy: When I use null,
Deployment failed. Correlation ID: e7f42bc9-2d4b-4310-90cd-8291da117c5d. {
  "error": {
    "code": "InvalidParameter",
    "message": "Required parameter 'platformFaultDomainCount' is missing (null).",
    "target": "platformFaultDomainCount"
  }
}

src/azure-cli/azure/cli/command_modules/vm/_params.py
qwordy: When will you change?

src/azure-cli/azure/cli/command_modules/vm/_params.py
qwordy: Added

qwordy:
Another question, after I create a vmss without vm profile, I can't run commands like list-instances, get-instance-view on it,
```
Operation 'VirtualMachineScaleSets.virtualMachines.GET' is not allowed on Virtual Machine Scale Set 'vmss1'
```
```
Operation 'VirtualMachineScaleSets.instanceView.GET' is not allowed on Virtual Machine Scale Set 'vmss1'.
```
Is it normal?

src/azure-cli/azure/cli/command_modules/vm/_params.py
shpa-microsoft: We have that term on portal already. Let's have a call today and clarify all the gaps?

src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
shpa-microsoft: Oh My bad. We do need this one; but not sure defaulting is the right thing to do v/s making clients always enter that value.

qwordy:
Shalaka and I have discussed in a phone call.

src/azure-cli/azure/cli/command_modules/vm/_params.py
shpa-microsoft: Decided on enum "orchestration" for this purpose. Keeping it in sync with Portal

src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -276,6 +276,7 @@ def load_arguments(self, _):
         c.argument('vm_name', name_arg_type, id_part=None, help='Name of the virtual machine.', completer=None)
         c.argument('os_disk_size_gb', type=int, help='the size of the os disk in GB', arg_group='Storage')
         c.argument('availability_set', help='Name or ID of an existing availability set to add the VM to. None by default.')
+        c.argument('vmss', help='Name of ID of an existing virtual machine scale set that the virtual machine should be assigned to. None by default.')
```
MyronFanQiu: Name or ID of

shpa-microsoft:
Can we name the param: **OrchestrationMode** instead? With options "VM" and "ScaleSetVM"

qwordy:
> 
> 
> Can we name the param: **OrchestrationMode** instead? With options "VM" and "ScaleSetVM"

I've merged it to enable related people to try it. I will open another PR for this problem.

PR #10658, opened 2019-09-26
Continue parsing image argument if aliases doc unreachable

jseely:
Currently the Azure CLI does not work in environments without public internet access when trying to use the managed disk image in `#5` of the _parse_image_argument function. 

This fix ignores the ConnectionError in `#4` when trying to lookup the aliases doc from `githubusercontent.com` and proceeds to `#5`. If `#5` does not match then it will return the error message:
```
Failed to connect to remote source of image aliases. Invalid image "{}". Use a valid image URN, customer image name, customer image id, or VHD blob URI.
See vm create -h for more information on specifying an image.
```

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/vm/_validators.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_validators.py
@@ -254,7 +254,7 @@ def _parse_image_argument(cmd, namespace):
 
     # 4 - attempt to match an URN alias (most likely)
     from azure.cli.command_modules.vm._actions import load_images_from_aliases_doc
-    from requests.exceptions import ConnectionError
+    import requests
     try:
         images = load_images_from_aliases_doc(cmd.cli_ctx)
         matched = next((x for x in images if x['urnAlias'].lower() == namespace.image.lower()), None)
@@ -264,7 +264,7 @@ def _parse_image_argument(cmd, namespace):
             namespace.os_sku = matched['sku']
             namespace.os_version = matched['version']
             return 'urn'
-    except ConnectionError:
+    except requests.exceptions.ConnectionError:
         pass
 
     # 5 - check if an existing managed disk image resource
--- a/src/azure-cli/azure/cli/command_modules/vm/_validators.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_validators.py
@@ -254,14 +254,17 @@ def _parse_image_argument(cmd, namespace):
 
     # 4 - attempt to match an URN alias (most likely)
     from azure.cli.command_modules.vm._actions import load_images_from_aliases_doc
-    images = load_images_from_aliases_doc(cmd.cli_ctx)
-    matched = next((x for x in images if x['urnAlias'].lower() == namespace.image.lower()), None)
-    if matched:
-        namespace.os_publisher = matched['publisher']
-        namespace.os_offer = matched['offer']
-        namespace.os_sku = matched['sku']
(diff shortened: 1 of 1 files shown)


src/azure-cli/azure/cli/command_modules/vm/_validators.py
```diff
@@ -254,14 +254,18 @@ def _parse_image_argument(cmd, namespace):
...
+        images = load_images_from_aliases_doc(cmd.cli_ctx)
+        matched = next((x for x in images if x['urnAlias'].lower() == namespace.image.lower()), None)
+        if matched:
+            namespace.os_publisher = matched['publisher']
+            namespace.os_offer = matched['offer']
+            namespace.os_sku = matched['sku']
+            namespace.os_version = matched['version']
+            return 'urn'
+    except requests.exceptions.ConnectionError:
+        pass
```
yugangw-msft: we can throw the exception here with the error string you put in the 5th step, since that step won't work anyway 
CC: @qwordy

qwordy: > 
> 
> we can throw the exception here with the error string you put in the 5th step, since that step won't work anyway
> CC: @qwordy

It can be an existing managed disk image resource.

src/azure-cli/azure/cli/command_modules/vm/_validators.py
```diff
@@ -271,9 +275,15 @@ def _parse_image_argument(cmd, namespace):
...
-              'pick an image from {}.\nSee vm create -h for more information on specifying an image.'
-        raise CLIError(err.format(namespace.image, [x['urnAlias'] for x in images]))
+        if images is not None:
+            err = 'Invalid image "{}". Use a valid image URN, custom image name, custom image id, ' \
+                  'VHD blob URI, or pick an image from {}.\nSee vm create -h for more information ' \
+                  'on specifying an image.'.format(namespace.image, [x['urnAlias'] for x in images])
+        else:
+            err = 'Failed to connect to remote source of image aliases. Invalid image "{}". Use a ' \
+                  'valid image URN, custom image name, custom image id, or VHD blob URI.\nSee vm ' \
+                  'create -h for more information on specifying an image.'.format(namespace.image)
```
qwordy: If it's not a valid alias nor an existing image name, what will happen?

src/azure-cli/azure/cli/command_modules/vm/_validators.py
jseely: @yugangw-msft no, that's what we are trying to avoid here. Even on failure in `#4`, `#5` should still be attempted.

src/azure-cli/azure/cli/command_modules/vm/_validators.py
jseely: If it's not a valid alias then it goes to `#5` as before, if it's not the name of an existing managed image it raises the same error message as before.

jseely:
@qwordy any update here? Is this good to merge?

qwordy:
> 
> 
> @qwordy any update here? Is this good to merge?

Merged.

jseely:
@qwordy Thanks! I know our mutual customers will be happy to hear!

qwordy:
> 
> 
> @qwordy Thanks! I know our mutual customers will be happy to hear!

You are welcome. I plan to add/update the alias file to/in project as a resource file, in each time we build and release. If can't access github, then use the local file.

PR #10686, opened 2019-09-29
Fix#10479: add transform function fo `az resource invoke-action`

### LINKED ISSUE #10479: HTTP post request wrong Json output
Hi,

When calling for my API from CLI, the Post request returns a Json string with all the escape character inside. it feels like the string has been json marshalled twice.
Example:  {\n  \"properties\": {\n   \"disabled\": true\n  }\n }
Expected output: 
{
  "properties": {
"disabled": true
   }
 }

in comparison I also have the default command which returns the correct value.
The code I used for those 2 API are identical. same methods So I don't understand the disparity in the returns.

This only happens when I call it through CLI, If I set up a local service, I can directly use CURL commands to call those APIS and I will get the expected returns.

Rrunning in cloudshell: 
az resource invoke-action --action enableConsole --ids "/subscriptions/$Subscription/providers/Microsoft.SerialConsole/consoleServices/default"
"{\n  \"properties\": {\n   \"disabled\": false\n  }\n }"

az resource show --ids "/subscriptions/$Subscription/providers/Microsoft.SerialConsole/consoleServices/default" | jq .properties
{
  "disabled": false
}

Both the Above API uses same method to create the json string and add the string to Header.

Juliehzl:
Fix #10479
---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/commands/__init__.py
+++ b/src/azure-cli-core/azure/cli/core/commands/__init__.py
@@ -634,15 +634,14 @@ class AzCliCommandInvoker(CommandInvoker):
             elif cmd_copy.no_wait_param and getattr(expanded_arg, cmd_copy.no_wait_param, False):
                 result = None
 
+            transform_op = cmd_copy.command_kwargs.get('transform', None)
+            if transform_op:
+                result = transform_op(result)
+
             if _is_poller(result):
                 result = LongRunningOperation(cmd_copy.cli_ctx, 'Starting {}'.format(cmd_copy.name))(result)
             elif _is_paged(result):
                 result = list(result)
-
-            # Re-arrange transform_op to make it work for LRO result
-            transform_op = cmd_copy.command_kwargs.get('transform', None)
-            if transform_op:
-                result = transform_op(result)
 
             result = todict(result, AzCliCommandInvoker.remove_additional_prop_layer)
             event_data = {'result': result}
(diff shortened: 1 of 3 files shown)


src/azure-cli/azure/cli/command_modules/resource/commands.py
```diff
@@ -171,15 +173,15 @@ def load_command_table(self, _):
         g.custom_command('list', 'list_locks')
         g.custom_show_command('show', 'get_lock')
         g.custom_command('update', 'update_lock')
-
+    from azure.cli.core.commands import DeploymentOutputLongRunningOperation
```
MyronFanQiu: Could we move `from azure.cli.core.commands import DeploymentOutputLongRunningOperation` under  `with self.command_group('resource', resource_custom, resource_type=ResourceType.MGMT_RESOURCE_RESOURCES) as g:`

src/azure-cli-core/azure/cli/core/commands/__init__.py
```diff
@@ -920,14 +920,20 @@ def __call__(self, result):
             result = super(DeploymentOutputLongRunningOperation, self).__call__(result)
             outputs = None
             try:
-                outputs = result.properties.outputs
+                if isinstance(result, str) and result:
+                    try:
+                        obj = json.loads(result)
+                        return obj
+                    except json.decoder.JSONDecodeError:
```
MyronFanQiu: What if the error is not json.decoder.JSONDecodeError?

Juliehzl: It will be throw out.

src/azure-cli/azure/cli/command_modules/resource/commands.py
Juliehzl: Maybe it will be better to combine it with CliCommandType at the top.

src/azure-cli-core/azure/cli/core/commands/__init__.py
```diff
@@ -920,14 +920,20 @@ def __call__(self, result):
...
+                        logger.info("Fail to transform result \"%s\" to dictionary", result)
```
qwordy: https://docs.python.org/3/library/string.html#string-formatting
This is new style. More fashionable.

PR #10690, opened 2019-09-29
Fix #10674: Add new parameter for az deployment create

### LINKED ISSUE #10674: --handle-extended-json-format missing from subscription level az deployment create command
**Describe the bug**
As of .74, the group deployment create command has the flag for --handle-extended-json-format, but it is missing from az deployment create. We will need it there as well.

As a side-note, can we have a short version of the flag? Maybe -j?

Juliehzl:
Fix #10674
---


src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -194,6 +194,8 @@ def load_arguments(self, _):
     with self.argument_context('deployment create') as c:
         c.argument('deployment_name', options_list=['--name', '-n'], required=False,
                    help='The deployment name. Default to template file base name')
+        c.argument('handle_extended_json_format', options_list=['--handle-extended-json-format', '-j'], action='store_true', is_preview=True,
+                   help='Support to handle extended template content including multiline and comments in deployment')
```
qwordy: Avoid identical code. Use CLIArgumentType or 
```
    for scope in ['group deployment create', 'deployment create']:
```

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -405,6 +405,109 @@ def on_request(self, request, **kwargs):
...
+
+    smc = get_mgmt_service_client(cli_ctx, ResourceType.MGMT_RESOURCE_RESOURCES)
+
+    class JsonCTemplate(object):
+        def __init__(self, template_as_bytes):
+            self.template_as_bytes = template_as_bytes
+
+    from msrest.serialization import Serializer
+
+    class MySerializer(Serializer):
```
qwordy: MySerializer. Naming could be improved.

qwordy:
Add documentation at a function level if you have time.

src/azure-cli/azure/cli/command_modules/resource/custom.py
Juliehzl: > MySerializer. Naming could be improved.

This is a temporary hack from Laurent, which will be removed and replaced with an official released library. Anyway, I can rename it with a more reasonable name.

src/azure-cli/azure/cli/command_modules/resource/_params.py
Juliehzl: Updated

PR #10749, opened 2019-10-03
update _help change the parameter name of blob name

moataseem:
update _help change the parameter name of blob name in cli command from --name to --blob-name

---


yonzhan:

Juliehzl:
Hi @moataseem, --name is by design for `az storage blob exists`.  For `az storage blob exists` command, the target of this command is just blob and we think it is better to use --name. Do you have any concern for changing it to `--blob-name`?

moataseem:
Hi @Juliehzl actually I've tried to execute the command and I've got the error: unrecognized arguments: --name 
I was able to execute it successfully through changing the parameter --name to --blob-name .
so it's not about naming. It's more about being able to execute the command to get the result if there is a blob under a specific container

Juliehzl:
> Hi @Juliehzl actually I've tried to execute the command and I've got the error: unrecognized arguments: --name
> I was able to execute it successfully through changing the parameter --name to --blob-name .
> so it's not about naming. It's more about being able to execute the command to get the result if there is a blob under a specific container

Can you verify it again? I tried it locally, but --name works well not --blob-name as follows:
```
C:\Users\zuh>az storage blob exists --name a.txt -c test --account-name zuhstoragetest1
{
  "exists": true
}
```

moataseem:
[image: azureBlob.png]
I have tried it by me you may say how it looks like in the image

yonzhan:
@moataseem have you figured out the problem? Do we still need this PR ?

moataseem:
@yonzhan As I've mentioned before. It worked for by changing the parameter
from --name to --blob-name so you can test against that if it works like
this by you too then we need it

yonzhan <notifications@github.com> schrieb am Sa., 2. Nov. 2019, 09:11:

yonzhan:
@Juliehzl  please help verify.

Juliehzl:
Hi @moataseem , it really strange and I think there should be something wrong. I have tried in az interactive and it shows as follows:
image
Could you please create an issue for your problem and share with me your debug information? One more thing to know is which cli version you are using.

PR #10751, opened 2019-10-04
[ARM] Refactor deployment commands

Tiano2017:
**History Notes:**  
[ARM] Add `az deployment group` and `az deployment operation group` for template deployment at resource groups. This is a duplicate of `az group deployment` and `az group deployment operation`. 
[ARM] Add `az deployment sub` and `az deployment operation sub` for template deployment at subscription scope. This is a duplicate of `az deployment` and `az deployment operation`.
[ARM] Add `az deployment mg` and `az deployment operation mg` for template deployment at management groups. 
[ARM] Add `az deployment tenant` and `az deployment operation tenant` for template deployment at tenant scope.

---


Tiano2017:
@Juliehzl can you help review this PR? I have a few more changes to make but most of them are in place.

yonzhan:

Juliehzl:
@Tiano2017 Can you resolve the conflict for your PR?

Tiano2017:
Please don't merge this PR as we'll make some design changes.

Juliehzl:
Any update for this PR?

Tiano2017:
> Any update for this PR?

Not yet. we'll need to redesign the commands. will update the PR once the new design is locked down.

Tiano2017:
@Juliehzl the PR is ready for review. can you help take a look?

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -191,21 +191,80 @@ def load_arguments(self, _):
...
+
+    with self.argument_context('deployment mg create') as c:
+        c.argument('deployment_name', options_list=['--name', '-n'], required=False,
+                   help='The deployment name. Default to template file base name')
+        c.argument('handle_extended_json_format', arg_type=extended_json_format_type)
+
+    with self.argument_context('deployment mg validate') as c:
+        c.argument('deployment_name', options_list=['--name', '-n'], required=False,
+                   help='The deployment name. Default to template file base name')
+        c.argument('handle_extended_json_format', arg_type=extended_json_format_type)
```
qwordy: Maybe you can use CLIArgumentType to define your argument, so that you can avoid duplicates.
E.g.
```
ultra_ssd_enabled_type = CLIArgumentType(
        arg_type=get_three_state_flag(), min_api='2018-06-01',
        help='Enables or disables the capability to have 1 or more managed data disks with UltraSSD_LRS storage account')
```

Tiano2017: sure I can update that.

Tiano2017:
@qwordy, @Juliehzl could you review this PR please?

qwordy:
Oh, some tests failed.
 e.g.
No match for the request (<Request (POST) https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourcegroups/clitest.rg000001/providers/Microsoft.Resources/deployments/domaintemplate/validate?api-version=2019-07-01>) was found.
Found 1 similar requests with 1 different matcher(s) 
You need to run test with --live to record it.

Tiano2017:
> Oh, some tests failed.
> e.g.
> No match for the request (<Request (POST) [https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourcegroups/clitest.rg000001/providers/Microsoft.Resources/deployments/domaintemplate/validate?api-version=2019-07-01>](https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourcegroups/clitest.rg000001/providers/Microsoft.Resources/deployments/domaintemplate/validate?api-version=2019-07-01%3E)) was found.
> Found 1 similar requests with 1 different matcher(s)
> You need to run test with --live to record it.

Thanks. There are two failures. I was able to record "test_sqlvm_add_and_remove", but for "test_deploymentmanager_scenario" I'm getting a weird error: 'test_deploymentmanager_scenario' exists twice in the 'deploymentmanager' module. Please rename one or both and re-run --discover.

I don't have any change in the "deploymentmanager" module. I also tried with the upstream/dev branch, and got the same error. Any suggestions around that?

qwordy:
> 
> 
> > Oh, some tests failed.
> > e.g.
> > No match for the request (<Request (POST) [https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourcegroups/clitest.rg000001/providers/Microsoft.Resources/deployments/domaintemplate/validate?api-version=2019-07-01>](https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourcegroups/clitest.rg000001/providers/Microsoft.Resources/deployments/domaintemplate/validate?api-version=2019-07-01%3E)) was found.
> > Found 1 similar requests with 1 different matcher(s)
> > You need to run test with --live to record it.
> 
> Thanks. I saw two test failures. I was able to record "test_sqlvm_add_and_remove", but for "test_deploymentmanager_scenario" I'm getting a weird error: 'test_deploymentmanager_scenario' exists twice in the 'deploymentmanager' module. Please rename one or both and re-run --discover.
> 
> I don't have any change in the "deploymentmanager" module. I also tried with the upstream/dev branch, and got the same error. Any suggestions around that?

Maybe there is a test case with same name in other modules?
Delete the recording file and try again.
Or You can checkout to a new branch from latest dev, create a new python virtual env. Add a test case "test_deploymentmanager_scenario". Then run it. If no error, then you can paste your changes to this new branch.

Juliehzl:
@Tiano2017 if it is ready to merge, pls remove the DoNotMerge label.

src/azure-cli/azure/cli/command_modules/resource/commands.py
```diff
@@ -233,19 +234,86 @@ def load_command_table(self, _):
         g.command('list', 'list')
         g.custom_show_command('show', 'get_deployment_operations', client_factory=cf_deployment_operations)
 
+    # az deployment
     with self.command_group('deployment', resource_deployment_sdk, min_api='2018-05-01', resource_type=ResourceType.MGMT_RESOURCE_RESOURCES) as g:
```
Juliehzl: Please make sure the min_api constraint is right for your new commands.

src/azure-cli/azure/cli/command_modules/resource/commands.py
```diff
@@ -233,19 +234,86 @@ def load_command_table(self, _):
...
+        g.custom_command('delete', 'delete_deployment_at_subscription_scope', supports_no_wait=True)
+        g.custom_command('validate', 'validate_arm_template_at_subscription_scope', validator=process_deployment_create_namespace,
+                         table_transformer=deployment_validate_table_format, exception_handler=handle_template_based_exception)
+        g.custom_command('create', 'deploy_arm_template_at_subscription_scope', supports_no_wait=True, validator=process_deployment_create_namespace,
+                         table_transformer=transform_deployment, exception_handler=handle_template_based_exception)
+        g.custom_command('export', 'export_template_at_subscription_scope')
+        g.custom_wait_command('wait', 'get_deployment_at_subscription_scope')
 
-    with self.command_group('deployment operation', resource_deployment_operation_sdk, min_api='2018-05-01', resource_type=ResourceType.MGMT_RESOURCE_RESOURCES) as g:
+    with self.command_group('deployment operation sub', resource_deployment_operation_sdk, min_api='2019-07-01', resource_type=ResourceType.MGMT_RESOURCE_RESOURCES) as g:
```
Juliehzl: It looks the commands in `az deployment operation sub` commands group are the same with commands in `az deployment operation`. https://github.com/Azure/azure-cli/pull/10751/files?file-filters%5B%5D=.json&file-filters%5B%5D=.py#diff-2057bbb8441166e4910b34b09d22b58cR249

src/azure-cli/azure/cli/command_modules/resource/commands.py
```diff
@@ -233,19 +234,86 @@ def load_command_table(self, _):
...
+                         exception_handler=handle_template_based_exception)
+        g.custom_command('export', 'export_template_at_subscription_scope')
+        g.custom_wait_command('wait', 'get_deployment_at_subscription_scope')
+
+    with self.command_group('deployment operation', resource_deployment_operation_sdk, min_api='2019-07-01', resource_type=ResourceType.MGMT_RESOURCE_RESOURCES) as g:
+        g.custom_command('list', 'list_deployment_operations_at_subscription_scope')
+        g.custom_show_command('show', 'get_deployment_operations_at_subscription_scope', client_factory=cf_deployment_operations)
+
+    # az deployment sub
+    with self.command_group('deployment sub', resource_deployment_sdk, min_api='2019-07-01', resource_type=ResourceType.MGMT_RESOURCE_RESOURCES) as g:
```
Juliehzl: I will prefer `deployment subscription` for your commands

src/azure-cli/azure/cli/command_modules/resource/commands.py
```diff
@@ -233,19 +234,86 @@ def load_command_table(self, _):
...
-        g.custom_wait_command('wait', 'wait_deployment_at_subscription_scope')
+        g.custom_command('validate', 'validate_arm_template_at_subscription_scope', validator=process_deployment_create_namespace,
+                         table_transformer=deployment_validate_table_format, exception_handler=handle_template_based_exception)
+        g.custom_command('create', 'deploy_arm_template_at_subscription_scope', supports_no_wait=True, validator=process_deployment_create_namespace,
+                         exception_handler=handle_template_based_exception)
+        g.custom_command('export', 'export_template_at_subscription_scope')
+        g.custom_wait_command('wait', 'get_deployment_at_subscription_scope')
+
+    with self.command_group('deployment operation', resource_deployment_operation_sdk, min_api='2019-07-01', resource_type=ResourceType.MGMT_RESOURCE_RESOURCES) as g:
+        g.custom_command('list', 'list_deployment_operations_at_subscription_scope')
```
Juliehzl: `deployment operation` commands are existing commands for other profiles. With your changes here, it will only show up in latest profile, which is really a breaking change to customers

Juliehzl:
@Tiano2017 There are several important thing to consider for your PR:
1. Make sure the original commands not affected by your changes, especially api version changes.
2. Can we add a parameter --scope in existing commands not adding new commands group?
2. Please remember to update history.rst for your changes.

src/azure-cli/azure/cli/command_modules/resource/commands.py
Tiano2017: yes, this is still the right min_api.

src/azure-cli/azure/cli/command_modules/resource/commands.py
Tiano2017: right. that's expected. The command group `az deployment operation sub` is the same as `az deployment operation`, and `az deployment sub` is the same as `az deployment`. We want to retire `az deployment` and `az deployment operation` but guess we'll do that in the future.

src/azure-cli/azure/cli/command_modules/resource/commands.py
Tiano2017: Is there a guideline on the command group name. we discussed in our team and decided to use the short format. Let me know if that's against the guidelines. @alex-frankel FYI.

src/azure-cli/azure/cli/command_modules/resource/commands.py
Tiano2017: makes sense. will fix it.

Tiano2017:
> @Tiano2017 There are several important things to consider for your PR:
> 
> 1. Make sure the original commands not affected by your changes, especially api version changes.
> 2. Can we add a parameter --scope in existing commands not adding new commands group?
> 3. Please update _help.py to introduce your new groups/commands and provide command usages if possible.
> 4. Please remember to update history.rst for your changes.

Thanks for the review. 
1. This is what we intend to do. I've fixed the min_api issue. 
2. We evaluated that option too, and decided the current implementation is cleaner and easier to use. 
3. Already included. 
4. Updated.

src/azure-cli/azure/cli/command_modules/resource/commands.py
Tiano2017: this was fixed.

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -378,9 +384,9 @@ def body(self, data, data_type, **kwargs):
             # Be sure to pass a DeploymentProperties
             template = data.properties.template
             if template:
-                data.properties.template = None
```
Tiano2017: @Juliehzl I made this change to make --handle-extended-json-format parameter work. Let me know if you see an issue.

Tiano2017:
@Juliehzl is there's anything else I need to update on this PR, except the conflicts which I'll resolve on Monday?

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
```diff
@@ -620,51 +622,54 @@ def restore_key(client, vault_base_url, file_path):
...
+
+    jwk.crv = supported_curves.get(curve, None)
+    if not jwk.crv:
+        raise CLIError("Import failed: Unsupported curve, {}.".format(curve))
+
+    jwk.x = _int_to_bytes(ec_key.private_numbers().public_numbers.x)
+    jwk.y = _int_to_bytes(ec_key.private_numbers().public_numbers.y)
+    jwk.d = _int_to_bytes(ec_key.private_numbers().private_value)
+
+
```
bim-msft: @Tiano2017 Why these irrelevant changes get involved? Actually, this part of change has already been merged in another PR. Can you clean your local git branch and force push it to remote again?

Tiano2017: thanks for catching that. I did a merge with dev and now it should be clean.

Tiano2017:
@bim-msft The CI run is passing now. let me know if there's any other issue.

src/azure-cli/HISTORY.rst
```diff
@@ -180,6 +180,11 @@ Release History
 **ARM**
 
 * Fix `az resource tag`: Recovery Services Vault tags cannot be updated
+* Added new command groups
```
Juliehzl: Please sync your PR with dev branch and make your release note in the right position.

src/azure-cli/azure/cli/command_modules/resource/commands.py
Juliehzl: If you will deprecate some commands group, please add deprecate info to them as shown in https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#deprecating-commands-and-arguments

src/azure-cli/azure/cli/command_modules/resource/commands.py
Juliehzl: For different profiles, is there any difference for command behavior?

src/azure-cli/azure/cli/command_modules/resource/commands.py
Juliehzl: Because there is existing subgroup named "operation", subscription looks more descriptive. Just my suggestion not required.

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -284,11 +284,14 @@ def _deploy_arm_template_core(cli_ctx, resource_group_name,
...
+    validation_result = smc.deployments.validate(resource_group_name=resource_group_name, deployment_name=deployment_name, properties=properties)
+
+    if validation_result and validation_result.error:
+        raise CLIError(validation_result.error)
     if validate_only:
-        response = sdk_no_wait(no_wait, smc.deployments.validate, resource_group_name, deployment_name, properties)
-        if response and response.error:
-            raise CLIError(response.error)
-        return response
+        return validation_result
```
Juliehzl: you don't want to support no wait for validate command, right? may I know the reason?

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -365,8 +368,13 @@ def _deploy_arm_template_unmodified(cli_ctx, resource_group_name, template_file=
         sender=PipelineRequestsHTTPSender(RequestsHTTPSender(smc.config))
     )
 
+    validation_result = deployments_operation_group.validate(resource_group_name=resource_group_name, deployment_name=deployment_name, properties=properties)
+
+    if validation_result and validation_result.error:
+        raise CLIError(validation_result.error)
     if validate_only:
-        return sdk_no_wait(no_wait, deployments_operation_group.validate, resource_group_name, deployment_name, properties)
+        return validation_result
```
Juliehzl: The same as above.

src/azure-cli/azure/cli/command_modules/resource/custom.py
Tiano2017: this has not changed.. it's the same as we do for `az deployment group validate`. I think the reason might be that the /validate API is not a long running operation, so no-wait is not needed. Let me know if my understanding is not correct.

src/azure-cli/HISTORY.rst
Tiano2017: fixed.

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -198,23 +201,76 @@ def load_arguments(self, _):
...
+        c.argument('deployment_name', arg_type=deployment_create_name_type)
+        c.argument('handle_extended_json_format', arg_type=extended_json_format_type)
+        c.argument('aux_subscriptions', nargs='*', options_list=['--aux-subs'],
+                   help='Auxiliary subscriptions which will be used during deployment across tenants.')
+
+    with self.argument_context('deployment group validate') as c:
+        c.argument('deployment_name', arg_type=deployment_create_name_type)
+        c.argument('handle_extended_json_format', arg_type=extended_json_format_type)
+
+    with self.argument_context('deployment mg') as c:
```
zhoxing-ms: `deployment mg` -> `deployment management-group`  Maybe the naming looks clearer and more elegant?

alex-frankel: our only concern is the number of characters to type out `management-group` especially when there is no autocomplete for cli (at least on windows from what I can find). MG is a fairly well-known acronym. Thoughts?

zhoxing-ms: @alex-frankel Personally, I mainly think `mg` doesn't look intuitive enough. However if `mg` is used in other commands, and the user already knows this, I think `mg` is fine.

Tiano2017: "mg" is actually something new.. I don't think other commands use it today. @alex-frankel what do you think?

alex-frankel: I didn't mean to suggest it is used in other cli commands, but it is a commonly used acronym in customer conversations, so I'm not worried about it

zhoxing-ms: Ok, I think `mg` is fine.

Tiano2017:
@Juliehzl @zhoxing-ms @bim-msft can you please merge this PR if there's no open issue?

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -41,6 +41,13 @@ def load_arguments(self, _):
     identity_role_type = CLIArgumentType(options_list=['--role'], help="Role name or id that will be assigned to the managed identity")
     extended_json_format_type = CLIArgumentType(options_list=['--handle-extended-json-format', '-j'], action='store_true', is_preview=True,
                                                 help='Support to handle extended template content including multiline and comments in deployment')
+    deployment_name_type = CLIArgumentType(options_list=['--name', '-n'], required=True, help='The deployment name.')
+    deployment_create_name_type = CLIArgumentType(options_list=['--name', '-n'], required=False, help='The deployment name. Default to template file base name')
+    management_group_id_type = CLIArgumentType(options_list=['--management-group-id', '-m'], required=True, help='The management group id.')
```
Juliehzl: Which kind of id is used here? When there is id as parameter suffix, we will expect GUID as parameter value.

Tiano2017: This is not necessarily a GUID, but it has been the official term for management group: https://docs.microsoft.com/en-us/azure/governance/management-groups/create.

Tiano2017:
> Please use resource_group_preparer() for your test. Refer to https://github.com/Azure/azure-cli/blob/dev/doc/authoring_tests.md#sample-4-prepare-a-resource-group-for-a-test

thanks for catching that. I've updated for the deployment test at resource group scope. The other scopes don't require a resource group to start, so I left those unchanged.

fengzhou-msft:
I added history notes in PR description. From now on, you do not need to modify HISTORY.rst. Instead please use PR title or PR description for history notes. See instructions [here](https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules#submitting-pull-requests).

Tiano2017:
> I added history notes in PR description. From now on, you do not need to modify HISTORY.rst. Instead please use PR title or PR description for history notes. See instructions [here](https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules#submitting-pull-requests).

I have updated the PR title. Is there anything else missing?

fengzhou-msft:
> > I added history notes in PR description. From now on, you do not need to modify HISTORY.rst. Instead please use PR title or PR description for history notes. See instructions [here](https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules#submitting-pull-requests).
> 
> Thanks! I have updated the PR title. Is there anything else missing? Do I need to revert the changes in HISTORY.rst?

Since the history notes is under `ARM`, you probably also want to put `[ARM]` instead of `[Resource]` in title?

Tiano2017:
> > > I added history notes in PR description. From now on, you do not need to modify HISTORY.rst. Instead please use PR title or PR description for history notes. See instructions [here](https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules#submitting-pull-requests).
> > 
> > 
> > Thanks! I have updated the PR title. Is there anything else missing? Do I need to revert the changes in HISTORY.rst?
> 
> Since the history notes is under `ARM`, you probably also want to put `[ARM]` instead of `[Resource]` in title?
> 
> You can delete your changes in HISTORY.rst or leave them there. The script to generate history notes will clean up manually written notes and then generate notes based on PR title or description. I suggest you to delete them so we can avoid conflicts when merging changes in HISTORY.rst for multiple PRs.

I've updated the PR title and reverted changes in HISTORY.rst.

src/azure-cli/azure/cli/command_modules/resource/_params.py
Juliehzl: It is a kind of management group name, right? In addition, in bot cli and powershell commands, we don't have parameter named xx-id,  but we all use --name to identify the specific management group. 
For management group parameter here, can we just use --management-group as parameter name and in help message, we can use something like "The name of management group. For root management group, it should be the Azure Active Directory ID"?

src/azure-cli/azure/cli/command_modules/resource/custom.py
Juliehzl: Yes. Thanks for explanation. I have checked the sdk and it is not long running operation.

src/azure-cli/azure/cli/command_modules/resource/_params.py
Tiano2017: I'd rather keep the "management-group-id". If you look at the link I put above, the description said "The Management Group ID is the directory unique identifier that is used to submit commands on this management group." I think there's a discussion on bringing everything consistent to "management group ID".

alex-frankel: We should do name *or* ID, but not management-group because then it is ambiguous until I go to help.

I would prefer ID because of how we present this field to user's in the portal, but @Juliehzl if you feel strongly about using name because that's how we refer to it in PS/CLI, I'm ok with that.

cc @rthorn17 as FYI

Tiano2017: Trying to close on this one. @Juliehzl let us know what you think.

Tiano2017: Also, please note the equivalent PowerShell cmdlets `*-AzManagementGroupDeployment` have the parameter as 'ManagementGroupId'.

alex-frankel: @Juliehzl - can you provide guidance on this one?

yonzhan:
@zhoxing-ms  please help with the PR.

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -198,23 +201,76 @@ def load_arguments(self, _):
...
+        c.argument('deployment_name', arg_type=deployment_create_name_type)
         c.argument('handle_extended_json_format', arg_type=extended_json_format_type)
 
     with self.argument_context('deployment validate') as c:
+        c.argument('deployment_name', arg_type=deployment_create_name_type)
         c.argument('handle_extended_json_format', arg_type=extended_json_format_type)
 
-    with self.argument_context('deployment operation show') as c:
+    with self.argument_context('deployment operation') as c:
         c.argument('operation_ids', nargs='+', help='A list of operation ids to show')
```
zhoxing-ms: @Tiano2017 I found a new question: Is this `operation_ids` is a parameter of the entire scope for `deployment operation` now? But the description on help looks like it's just a parameter only for `deployment operation show`.

Tiano2017: The `operation_ids` is only used for the `show` commands. It's put here so that it's accessible to all the `show` commands: we have `deployment operation show`, `deployment operation sub show`, `deployment operation group show`, `deployment operation mg show`, `deployment operation tenant show`.

zhoxing-ms: OK, it's fine

src/azure-cli/azure/cli/command_modules/resource/commands.py
Tiano2017: the behavior on this part should be same for different API versions. I've changed the minimum api version back to 2018-05-01.

src/azure-cli/azure/cli/command_modules/resource/_params.py
Juliehzl: This is what we follows in https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md#argument-naming-conventions.
And the following part is what I get for a management group using existing cli command. May I know which one is the value I should pass to your parameter if you are using `--management-group-id`? 
```
  {
    "displayName": "zunli",
    "id": "/providers/Microsoft.Management/managementGroups/zuhtestmg",
    "name": "zuhtestmg",
    "tenantId": "54826b22-38d6-4fb2-bad9-b7b93a3e9c5a",
    "type": "/providers/Microsoft.Management/managementGroups"
  }
```
From my understanding, one of the "name" and "id" part should work to specify a valid management group. If you can support both, you also should use --management-group with help message "Id or name of management group.“ If you only support name, it will be confused for me to use your commands.

alex-frankel: unfortunately, the property you need there is "name". The confusion is that in the Portal, that same property is called "ID". The question is what do we optimize for. We will take @zhoxing-ms approval as an acknowledgement that we can move forward with ID

src/azure-cli/azure/cli/command_modules/resource/commands.py
qianwens: vote for subscription. sub is not descriptive.

---
In reply to: [375117700](https://github.com/Azure/azure-cli/pull/10751#discussion_r375117700) [](ancestors = 375117700)

src/azure-cli/azure/cli/command_modules/resource/commands.py
qianwens: may I understand why az deployment operation will be retired? is there more subresource like sub in operation command group, if so, can you give an example? I ask this because if sub is already the top resource in operation, we would suggest to have it as a default resource in operation command group.

---
In reply to: [367718352](https://github.com/Azure/azure-cli/pull/10751#discussion_r367718352) [](ancestors = 367718352)

Tiano2017: so `az deployment operation` works at subscription scope today. It lists or shows the deployment operations for a deployment at subscription scope. With the changes in this PR, we are supporting deployment at tenant and management group scope. To make all commands consistent, we are introducing `az deployment operation group | sub | mg | tenant` for each scope. The `az deployment operation sub` will replace `az deployment operation`.

src/azure-cli/azure/cli/command_modules/resource/commands.py
```diff
@@ -218,6 +218,7 @@ def load_command_table(self, _):
         g.command('add-value', 'create_or_update_value')
         g.command('remove-value', 'delete_value')
 
+    # az group deployment
     with self.command_group('group deployment', resource_deployment_sdk) as g:
```
qianwens: >command_group [](start = 14, length = 13)

will group deployment, group deployment operation be deprecated, if so, you may need to add redirect here

src/azure-cli/azure/cli/command_modules/resource/commands.py
```diff
@@ -233,19 +234,86 @@ def load_command_table(self, _):
...
+                         table_transformer=transform_deployment, exception_handler=handle_template_based_exception)
+        g.custom_command('export', 'export_template_at_subscription_scope')
+        g.custom_wait_command('wait', 'get_deployment_at_subscription_scope')
+
+    with self.command_group('deployment operation sub', resource_deployment_operation_sdk, min_api='2019-07-01', resource_type=ResourceType.MGMT_RESOURCE_RESOURCES) as g:
+        g.custom_command('list', 'list_deployment_operations_at_subscription_scope')
+        g.custom_show_command('show', 'get_deployment_operations_at_subscription_scope', client_factory=cf_deployment_operations)
+
+    # az deployment group
+    with self.command_group('deployment group', resource_deployment_sdk, min_api='2019-07-01', resource_type=ResourceType.MGMT_RESOURCE_RESOURCES) as g:
```
qianwens: >'2019-07-01' [](start = 81, length = 12)

There is no min_api in group deployment -> with self.command_group('group deployment', resource_deployment_sdk).
Why do you need a min_api here?

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -288,11 +288,14 @@ def _deploy_arm_template_core(cli_ctx, resource_group_name,
                                       parameters=parameters, mode=mode, on_error_deployment=on_error_deployment)
 
     smc = get_mgmt_service_client(cli_ctx, ResourceType.MGMT_RESOURCE_RESOURCES, aux_subscriptions=aux_subscriptions)
+
+    validation_result = smc.deployments.validate(resource_group_name=resource_group_name, deployment_name=deployment_name, properties=properties)
+
+    if validation_result and validation_result.error:
+        raise CLIError(validation_result.error)
```
qianwens: It seems you do not check validate_only in this method now, is there any reason for this? And why do you not use sdk_no_wait?

src/azure-cli/azure/cli/command_modules/resource/commands.py
Tiano2017: they will be deprecated. we'll add the deprecation info later in a separate PR.

src/azure-cli/azure/cli/command_modules/resource/commands.py
Tiano2017: it's not required. does it cause any issue?

src/azure-cli/azure/cli/command_modules/resource/custom.py
Tiano2017: we want to do validate for both validate and create, so we are not checking validate_only. we don't need sdk_no_wait because the validate API is not a long running operation.

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -410,16 +418,229 @@ def on_request(self, request, **kwargs):
...
+                                         no_wait=False, handle_extended_json_format=False):
+    deployment_properties = None
+    if handle_extended_json_format:
+        deployment_properties = _prepare_deployment_properties_unmodified(cli_ctx=cli_ctx, template_file=template_file,
+                                                                          template_uri=template_uri,
+                                                                          parameters=parameters, mode='Incremental')
+    else:
+        deployment_properties = _prepare_deployment_properties(cli_ctx=cli_ctx, template_file=template_file,
+                                                               template_uri=template_uri,
+                                                               parameters=parameters, mode='Incremental')
```
qianwens: I see lot's of if handle_extended_json_format check in this file, may be you can add another method to wrap this logic

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -428,70 +649,45 @@ def _deploy_arm_template_subscription_scope_unmodified(cli_ctx,
         template_content = read_file_content(template_file)
         template_obj = _remove_comments_from_json(template_content)
 
+    if rollback_on_error == '':
```
qianwens: >if rollback_on_error == '' [](start = 4, length = 26)

this logic seems a little tricky, what about if rollback_on_error is None, what not just check if not rollback_on_error which means both Non and empty string

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -410,16 +418,229 @@ def on_request(self, request, **kwargs):
...
+    if validation_result and validation_result.error:
+        raise CLIError(validation_result.error)
+    if validate_only:
+        return validation_result
+
+    return sdk_no_wait(no_wait, mgmt_client.create_or_update_at_tenant_scope,
+                       deployment_name, deployment_properties, deployment_location)
+
+
+def _prepare_deployment_properties_unmodified(cli_ctx, template_file=None, template_uri=None, parameters=None, mode=None, rollback_on_error=None):
```
qianwens: >_prepare_deployment_properties_unmodified [](start = 4, length = 41)

there are lots of duplicated code between _prepare_deployment_properties_unmodified and _prepare_deployment_properties, can you refactor these two methods? Besides, I notice unmodified method does not remove operation from the template while modified does not, which is confusing

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -501,14 +697,42 @@ def _deploy_arm_template_subscription_scope(cli_ctx,
...
-        return sdk_no_wait(no_wait, smc.deployments.validate_at_subscription_scope,
-                           deployment_name, properties, deployment_location)
-    return sdk_no_wait(no_wait, smc.deployments.create_or_update_at_subscription_scope,
-                       deployment_name, properties, deployment_location)
+    return properties
+
+
+def _get_deployment_management_client(cli_ctx, handle_extended_json_format=False, aux_subscriptions=None):
+    smc = get_mgmt_service_client(cli_ctx, ResourceType.MGMT_RESOURCE_RESOURCES, aux_subscriptions)
+    deployments_operation_group = smc.deployments  # This solves the multi-api for you
```
qianwens: >deployments_operation_group [](start = 4, length = 27)

suggest to rename it as deployment_client, deployments_operation_group looks like a resource

src/azure-cli/azure/cli/command_modules/resource/_validators.py
```diff
@@ -31,12 +37,6 @@ def _validate_deployment_name(namespace):
             namespace.deployment_name = 'deployment1'
 
 
-def process_deployment_create_namespace(namespace):
```
qianwens: seems no change this file, can you revert this change?

src/azure-cli/azure/cli/command_modules/resource/custom.py
Tiano2017: I tried.. but didn't work out a cleaner solution.

src/azure-cli/azure/cli/command_modules/resource/custom.py
Tiano2017: I'm not sure about this one. This part was not changed. it's existing code.

src/azure-cli/azure/cli/command_modules/resource/custom.py
Tiano2017: I tried to refactor _prepare_deployment_properties_unmodified and _prepare_deployment_properties, but the current version is the best I could do. If you look at their implementations, there are slight differences. Let me know if you have suggestions on further refactoring. 

As for naming, again it's taken from existing code. I'm not sure if we want to rename the methods. Let me know.

src/azure-cli/azure/cli/command_modules/resource/_validators.py
Tiano2017: sure. updated.

src/azure-cli/azure/cli/command_modules/resource/custom.py
Tiano2017: updated.

src/azure-cli/azure/cli/command_modules/resource/commands.py
```diff
@@ -229,23 +230,91 @@ def load_command_table(self, _):
         g.custom_command('export', 'export_deployment_as_template')
         g.wait_command('wait')
 
-    with self.command_group('group deployment operation', resource_deployment_operation_sdk) as g:
+    with self.command_group('group deployment operation', resource_deployment_operation_sdk, deprecate_info=self.deprecate(redirect='deployment operation group', hide=True)) as g:
         g.command('list', 'list')
         g.custom_show_command('show', 'get_deployment_operations', client_factory=cf_deployment_operations)
 
+    # az deployment
     with self.command_group('deployment', resource_deployment_sdk, min_api='2018-05-01', resource_type=ResourceType.MGMT_RESOURCE_RESOURCES) as g:
```
Juliehzl: You also need to deprecate this command group because it is changed to `deployment sub`, right?

PR #10775, opened 2019-10-08
Openshift/azure monitoring cli

troy0820:
- Import `OpenShiftManagedClusterMonitorProfile` 
- Create help example for `openshift create` to be used with `--workspace_resource_id`
- Add `monitor_profile` to `OpenShiftManagedCluster`
---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acs/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/_help.py
@@ -896,6 +896,9 @@ parameters:
   - name: --customer-admin-group-id
     type: string
     short-summary: The Object ID of an Azure Active Directory Group that memberships will get synced into the OpenShift group "osa-customer-admins". If not specified, no cluster admin access will be granted.
+  - name: --loganalytics-workspace-resource-id
+    type: string
+    short-summary: The resource ID of an existing Log Analytics Workspace to use for storing monitoring data.
 
 
 examples:
@@ -907,6 +910,8 @@ examples:
     text: az openshift create -g MyResourceGroup -n MyManagedCluster --aad-client-app-id {APP_ID} --aad-client-app-secret {APP_SECRET} --aad-tenant-id {TENANT_ID} --compute-count 5
   - name: Create an Openshift cluster using a custom vnet
     text: az openshift create -g MyResourceGroup -n MyManagedCluster --vnet-peer "/subscriptions/0000000-0000-0000-0000-000000000000/resourceGroups/openshift-vnet/providers/Microsoft.Network/virtualNetworks/test"
+  - name: Create an Openshift cluster with Log Analytics monitoring enabled
+    text: az openshift create -g MyResourceGroup -n MyManagedCluster --loganalytics-workspace-resource-id {LOGANALYTICS_WORKSPACE_RESOURCE_ID}
 """
 
 helps['openshift delete'] = """
--- a/src/azure-cli/azure/cli/command_modules/acs/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/_help.py
@@ -896,7 +896,7 @@ parameters:
   - name: --customer-admin-group-id
     type: string
     short-summary: The Object ID of an Azure Active Directory Group that memberships will get synced into the OpenShift group "osa-customer-admins". If not specified, no cluster admin access will be granted.
-  - name: --loganalytics-workspace-resource-id
+  - name: --workspace-resource-id
     type: string
     short-summary: The resource ID of an existing Log Analytics Workspace to use for storing monitoring data.
 
@@ -911,7 +911,7 @@ examples:
   - name: Create an Openshift cluster using a custom vnet
     text: az openshift create -g MyResourceGroup -n MyManagedCluster --vnet-peer "/subscriptions/0000000-0000-0000-0000-000000000000/resourceGroups/openshift-vnet/providers/Microsoft.Network/virtualNetworks/test"
   - name: Create an Openshift cluster with Log Analytics monitoring enabled
-    text: az openshift create -g MyResourceGroup -n MyManagedCluster --loganalytics-workspace-resource-id {LOGANALYTICS_WORKSPACE_RESOURCE_ID}
+    text: az openshift create -g MyResourceGroup -n MyManagedCluster --workspace-resource-id {WORKSPACE_RESOURCE_ID}
 """
 
 helps['openshift delete'] = """
--- a/src/azure-cli/azure/cli/command_modules/acs/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/custom.py
@@ -3126,7 +3126,7 @@ def openshift_create(cmd, client, resource_group_name, name,  # pylint: disable=
                      vnet_peer=None,
                      tags=None,
                      no_wait=False,
-                     loganalytics_workspace_resource_id=None,
+                     workspace_resource_id=None,
                      customer_admin_group_id=None):
 
     if location is None:
@@ -3197,13 +3197,13 @@ def openshift_create(cmd, client, resource_group_name, name,  # pylint: disable=
                 namespace='Microsoft.Network', type='virtualNetwork',
                 name=vnet_peer
             )
-    if loganalytics_workspace_resource_id is not None:
-        loganalytics_workspace_resource_id = loganalytics_workspace_resource_id.strip()
-        if not loganalytics_workspace_resource_id.startswith('/'):
-            loganalytics_workspace_resource_id = '/' + loganalytics_workspace_resource_id
-        if loganalytics_workspace_resource_id.endswith('/'):
-            loganalytics_workspace_resource_id = loganalytics_workspace_resource_id.rstrip('/')
-        monitor_profile = OpenShiftManagedClusterMonitorProfile(enabled=True, loganalytics_workspace_resource_id=loganalytics_workspace_resource_id)  # pylint: disable=line-too-long
+    if workspace_resource_id is not None:
+        workspace_resource_id = workspace_resource_id.strip()
+        if not workspace_resource_id.startswith('/'):
+            workspace_resource_id = '/' + workspace_resource_id
+        if workspace_resource_id.endswith('/'):
+            workspace_resource_id = workspace_resource_id.rstrip('/')
+        monitor_profile = OpenShiftManagedClusterMonitorProfile(enabled=True, workspace_resource_id=workspace_resource_id)  # pylint: disable=line-too-long
     else:
         monitor_profile = None
 
--- a/src/azure-cli/azure/cli/command_modules/acs/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/custom.py
@@ -81,6 +81,8 @@ from azure.mgmt.containerservice.v2019_04_30.models import OpenShiftManagedClust
 from azure.mgmt.containerservice.v2019_04_30.models import OpenShiftRouterProfile
 from azure.mgmt.containerservice.v2019_04_30.models import OpenShiftManagedClusterAuthProfile
 from azure.mgmt.containerservice.v2019_04_30.models import NetworkProfile
+from azure.mgmt.containerservice.v2019_09_30_preview.models import OpenShiftManagedCluster as OpenShiftManagedClusterMonitor  # pylint: disable=line-too-long
+from azure.mgmt.containerservice.v2019_09_30_preview.models import OpenShiftManagedClusterMonitorProfile
 
 from ._client_factory import cf_container_services
 from ._client_factory import cf_resource_groups
@@ -3124,6 +3126,7 @@ def openshift_create(cmd, client, resource_group_name, name,  # pylint: disable=
                      vnet_peer=None,
                      tags=None,
                      no_wait=False,
+                     loganalytics_workspace_resource_id=None,
                      customer_admin_group_id=None):
 
     if location is None:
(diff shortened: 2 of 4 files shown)


src/azure-cli/azure/cli/command_modules/acs/_help.py
```diff
@@ -748,6 +751,8 @@
     text: az openshift create -g MyResourceGroup -n MyManagedCluster --aad-client-app-id {APP_ID} --aad-client-app-secret {APP_SECRET} --aad-tenant-id {TENANT_ID} --compute-count 5
   - name: Create an Openshift cluster using a custom vnet
     text: az openshift create -g MyResourceGroup -n MyManagedCluster --vnet-peer "/subscriptions/0000000-0000-0000-0000-000000000000/resourceGroups/openshift-vnet/providers/Microsoft.Network/virtualNetworks/test"
+  - name: Create an Openshift cluster with Log Analytics monitoring enabled
+    text: az openshift create -g MyResourceGroup -n MyManagedCluster --workspace-resource-id {WORKSPACE_ID}
 """
```
ganga1980: Azure Log Analytics Workspace has both Id (i.e. GUID) and Resource Id, I think, we should rename to "WORKSPACE_ID" to "WORKSPACE_RESOURCE_ID" . Since we have seen customers passing WorkspaceID (i.e. GUID) instead of Azure Resource Id of the Log analytics workspace.

src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -2950,6 +2952,16 @@ def openshift_create(cmd, client, resource_group_name, name,  # pylint: disable=
                 namespace='Microsoft.Network', type='virtualNetwork',
                 name=vnet_peer
             )
+    if workspace_resource_id is not None:
+        workspace_resource_id = workspace_resource_id.strip()
+        if not workspace_resource_id.startswith('/'):
+            workspace_resource_id = '/' + workspace_resource_id
+        if workspace_resource_id.endswith('/'):
+            workspace_resource_id = workspace_resource_id.rstrip('/')
+        monitor_profile = OpenShiftManagedClusterMonitorProfile(enabled=True, workspace_resource_id=workspace_resource_id)
```
ganga1980: We should also check the provided workspace_resource_id is valid and exists.  I have plan to do the same for AKS as well.

src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -2950,6 +2952,16 @@ def openshift_create(cmd, client, resource_group_name, name,  # pylint: disable=
                 namespace='Microsoft.Network', type='virtualNetwork',
                 name=vnet_peer
             )
+    if workspace_resource_id is not None:
```
ganga1980: In AKS, we create default Azure Log Analytics workspace if the customer passed on the "monitoring" addon and with no workspace-resource-id.  Looks like in ARO case, there is no monitor addon concept and  enabling monitoring governed by --workspace-resource-id parameter.  is that, right? or we will be going with the same behavior as AKS i.e. creating default log analytics workspace when customer provided no workspace-resource-id parameter.

src/azure-cli/azure/cli/command_modules/acs/_help.py
ganga1980: From reviewing the code, I understand that, customer can only add the monitoring as part of cluster creation.  what about the scenario to add monitoring for existing ARO cluster?

troy0820: I agree that changing this to `{WORKSPACE_RESOURCE_ID}` makes sense.  Change submitted in commit c5fe9dcc

src/azure-cli/azure/cli/command_modules/acs/custom.py
troy0820: ARO doesn't have a concept of add-ons at the moment.  I do not believe we want to create a default log analytics workspace if they do not pass in the `workspace-resource-id`.  We will not enable it if it's not passed in.

src/azure-cli/azure/cli/command_modules/acs/_help.py
```diff
@@ -737,6 +737,9 @@
   - name: --customer-admin-group-id
     type: string
     short-summary: The Object ID of an Azure Active Directory Group that memberships will get synced into the OpenShift group "osa-customer-admins". If not specified, no cluster admin access will be granted.
+  - name: --workspace-resource-id
```
jim-minter: would something like `--loganalytics-workspace-resource-id` be better here?  I know it's a lot to type.  Don't mind whether it's loganalytics or monitoring or some other name.

src/azure-cli/azure/cli/command_modules/acs/_help.py
jim-minter: I understand the RP will support changing the value - @troy0820 the `az` client will need to as well.

src/azure-cli/azure/cli/command_modules/acs/custom.py
jim-minter: Will the RP be doing this server side @JackQuincy ?

src/azure-cli/azure/cli/command_modules/acs/_help.py
troy0820: That is correct.  I think that scenario to add monitoring to an existing cluster is valid and I am open up to feedback on how AKS does this.

src/azure-cli/azure/cli/command_modules/acs/custom.py
jim-minter: @troy0820 I don't think it's quite as simple as this.  You can't just graft a struct from the preview API onto the released API - Azure won't recognise it because the request API version will be old.  If loganalytics is requested `az` will need to send a preview API request to the Azure API.  It will also have to handle receiving a preview API response from the Azure API.

If loganalytics is not requested, in some ways it doesn't matter which API version is sent to the Azure API.  It is possibly safer (although more work) to send a request using the released API, however.  Check with MSFT and check to see what the codebase does in this regard for other preview APIs.

I think I'd expect to see:
* if `az` parses server response, it can handle responses in either API version
* `az` creates request using released API, then, if loganalytics is requested, it converts that to the preview API and adds the additional monitorProfile.

src/azure-cli/azure/cli/command_modules/acs/_help.py
troy0820: To reduce complexity, I'm okay with the long `--loganalytics-workspace-resource-id` name.  I can change it to that to provide a concise naming convention.

src/azure-cli/azure/cli/command_modules/acs/_help.py
JackQuincy: I have tested on update if you add the monitor profile, or have one and enable it, it will add the agent to the cluster and logs will flow. Simarily you can change the workspace and the agents will be updated and logs will flow to the new namespace

src/azure-cli/azure/cli/command_modules/acs/custom.py
JackQuincy: Yes we are already checking the workspace is valid

JackQuincy: The way the cli is designed all az openshift calls have to have the same api version which should move to the preview version. The customer can override this value. We are running all tests on the preview version and it is passing.

src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -2950,6 +2952,16 @@ def openshift_create(cmd, client, resource_group_name, name,  # pylint: disable=
                 namespace='Microsoft.Network', type='virtualNetwork',
                 name=vnet_peer
             )
+    if loganalytics_workspace_resource_id is not None:
+        loganalytics_workspace_resource_id = loganalytics_workspace_resource_id.strip()
+        if not loganalytics_workspace_resource_id.startswith('/'):
+            loganalytics_workspace_resource_id = '/' + loganalytics_workspace_resource_id
+        if loganalytics_workspace_resource_id.endswith('/'):
+            loganalytics_workspace_resource_id = loganalytics_workspace_resource_id.rstrip('/')
```
JackQuincy: @jim-minter currently the RP is not doing this logic. We will fail the call if it has a trailing '/'

yonzhan:
@MyronFanQiu  please take a look at this PR.

troy0820:
@asalkeld can you take a look?

JackQuincy:
@troy0820 these changes are in west central us. We have been blocked due to dependency bugs. but you should be able to test there now. Can you provision a cluster there using a local build of your repo?

src/azure-cli/azure/cli/command_modules/acs/custom.py
ganga1980: @jim-minter , doing at client side would help user not running into cluster creation failure because of this. This is we learned from AKS experience.

src/azure-cli/azure/cli/command_modules/acs/custom.py
ganga1980: @troy0820, ok, thanks.

MyronFanQiu:
Hello. I'm working on the CLI monitor. How about supporting both workspace_name and workspace_id by exposing `--workspace` to user?

troy0820:
@MyronFanQiu I was trying to mirror what AKS does.  I'm not sure if they use the `workspace_name` for `workspace_resource_id` in AKS.  Also I was passing this in as I'm understanding the RP will handle the parsing logic for the `loganalytics_workspace_resource_id`

@ehashman can you TAL at this? 🙏

src/azure-cli/azure/cli/command_modules/acs/custom.py
jim-minter: @JackQuincy I don't quite get it - does that mean the PR can't merge until the new API is GA?

JackQuincy: No we have used preview versions in az in the past. We more have preview features.  But the rest of the api should be unaffected

src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -2944,17 +2946,37 @@ def openshift_create(cmd, client, resource_group_name, name,  # pylint: disable=
...
-    osamc = OpenShiftManagedCluster(
-        location=location, tags=tags,
-        open_shift_version="v3.11",
-        network_profile=network_profile,
-        auth_profile=auth_profile,
-        agent_pool_profiles=agent_pool_profiles,
-        master_pool_profile=agent_master_pool_profile,
-        router_profiles=[default_router_profile])
+    if monitor_profile is not None:
+        osamc = OpenShiftManagedCluster(
```
ehashman: Looking above...

```python
from azure.mgmt.containerservice.v2019_04_30.models import OpenShiftManagedCluster
```

These are both referring to the `v2019_04_30` model, we need to import the `v2019_09_30_preview` for the Azure Monitor case.

troy0820: 😬 How did I miss that 🤦‍♂ ?  Updated with an alias.

JackQuincy:
LGTM

JackQuincy:
Also update is the changes are out globally in the service.

MyronFanQiu:
@troy0820 Hello. Sorry for the delay due the Ignite. I have three questions about this PR.
- Could you create a support a workspace name in the same RG or creating a default workspace for user in the same RG? I think it will be easy for user to use. We can discuss about this one since I started to support workspace for several other service.
https://github.com/Azure/azure-cli/blob/7d25cdeea34b35f563ad8ef1b1b9086816a7fa01/src/azure-cli/azure/cli/command_modules/vm/custom.py#L992-L1020
- `--loganalytics-workspace-resource-id` is quite long.... I know that there is another workspace id. Do we really need the word `loganalytics`
- Update command doesn't support `--loganalytics-workspace-resource-id`. Based on your discussion, it should support. Did I miss something?

src/azure-cli/azure/cli/command_modules/acs/_help.py
```diff
@@ -737,6 +737,9 @@
   - name: --customer-admin-group-id
     type: string
     short-summary: The Object ID of an Azure Active Directory Group that memberships will get synced into the OpenShift group "osa-customer-admins". If not specified, no cluster admin access will be granted.
+  - name: --loganalytics-workspace-resource-id
```
olga-mir: @troy0820 @jim-minter is there a particular reason for choosing this name?
it seems in az cli context `--workspace-resource-id` already refers to Log Analytics workspace. If we use this name instead it will be consistent with `az aks create` and `az aks enable-addons`, also will make it shorter which is in my opinion is nicer experience (even if  with autocompletion available)

ganga1980: agree with @olga-mir about having consistent and precise naming

src/azure-cli/azure/cli/command_modules/acs/custom.py
ganga1980: @JackQuincy , we are good on this since the code in this PR takes care of the scenario which you have mentioned. so can you please resolve/close this comment.

src/azure-cli/azure/cli/command_modules/acs/_help.py
troy0820: > Azure Log Analytics Workspace has both Id (i.e. GUID) and Resource Id, I think, we should rename to "WORKSPACE_ID" to "WORKSPACE_RESOURCE_ID" . Since we have seen customers passing WorkspaceID (i.e. GUID) instead of Azure Resource Id of the Log analytics workspace.

I think the decision came from the comment @ganga1980 made earlier in this posting which @jim-minter suggested we go with the longer name.  

> would something like --loganalytics-workspace-resource-id be better here? I know it's a lot to type. Don't mind whether it's loganalytics or monitoring or some other name.

I can change this to `monitoring` or `loganalytics` which will reference the same expected value for it to work.  The concept of add-ons when drafting this PR was not a thing and still doesn't exist yet within the ARO cluster.  We didn't want to build that complexity into it as of right now.  

I agree that the `--workspace-resource-id` would work with naming consistency but we run up against the problem @ganga1980 mentioned earlier in this thread.  If we believe that customers will not pass in the WorkspaceID _(GUID)_ with it being defined as `--workspace-resource-id` I can change it back to this or change it to one of the suggestions @jim-minter mentioned above.

MyronFanQiu: @troy0820 Thanks for the explanation. Currently, it's already a little messy about the naming for the workspace. I'm fine that we have such name `--workspace-resource-id`. I just consider that we don't need `loganalytics`. In my opinion, the ideal situation should be that the command exposes `--workspace` and supports name/workspace-resource-id/workspace-id. However, I cannot find the conversion from workspace-id to workspace-resource-id.

amanohar:
@troy0820 we don't need to mirror AKS for update command. Let's discuss it before finalizing any changes.

`--workspace-resource-id` or `-workspace-id` sounds good to me.

src/azure-cli/azure/cli/command_modules/acs/_help.py
troy0820: Changed to `--workspace-resource-id` in commit 2f77e2dd

troy0820:
@MyronFanQiu I can update the `azure-cli/History.rst` as well.

troy0820:
Updated `azure-cli/History.rst` @Juliehzl @MyronFanQiu

PR #10784, opened 2019-10-09
Fix#10638: Add LFS for storage account

Juliehzl:
Fix#10638
---


src/azure-cli/azure/cli/command_modules/storage/_params.py
```diff
@@ -99,7 +106,7 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals, too-many-statem
     with self.argument_context('storage account check-name') as c:
         c.argument('name', options_list=['--name', '-n'])
 
-    with self.argument_context('storage account create') as c:
+    with self.argument_context('storage account create', resource_type=ResourceType.MGMT_STORAGE) as c:
```
qwordy: What will happen if resource_type is not specified

src/azure-cli/azure/cli/command_modules/storage/_params.py
```diff
@@ -66,6 +66,13 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals, too-many-statem
         default=5000, help='Specifies the maximum number of results to return. Provide "*" to return all.',
         validator=validate_storage_data_plane_list)
 
+    large_file_share_type = CLIArgumentType(
+        action='store_true', min_api='2019-04-01',
+        help='Enable the capability to support large file shares with more than 5 TiB capacity for storage account.'
+             'Once the property is enabled, the feature cannot be disabled. Currently only supported for LRS and ZRS '
+             'replication types, hence account conversions to geo-redundant accounts would not be possible. For more '
+             'information, please refer to https://go.microsoft.com/fwlink/?linkid=2086047.')
```
qwordy: Maybe a little long.

src/azure-cli/azure/cli/command_modules/storage/operations/account.py
```diff
@@ -145,7 +149,9 @@ def update_storage_account(cmd, instance, sku=None, tags=None, custom_domain=Non
             directory_service_options='AADDS' if enable_files_aadds else 'None')
     if assign_identity:
         params.identity = Identity()
-
+    if enable_large_file_share:
+        LargeFileSharesState = cmd.get_models('LargeFileSharesState')
+        params.large_file_shares_state = LargeFileSharesState("Enabled")
```
qwordy: I find `create` and `update` have same code pieces.

src/azure-cli/azure/cli/command_modules/storage/_params.py
Juliehzl: It will not use the defined API version.

src/azure-cli/azure/cli/command_modules/storage/_params.py
Juliehzl: This is the PM defined help message.

src/azure-cli/azure/cli/command_modules/storage/operations/account.py
Juliehzl: Right. For both create and update command, we just need to add a new property in the same way.

qwordy: Maybe we can abstract a function.
It's a general problem. `create` and `update` has lots of similar code. If we write them twice, it's error-prone and hard to maintain.

Juliehzl: But it is just a condition. How can we abstract a function for this? And for other properties, we also implement like this.

src/azure-cli/azure/cli/command_modules/storage/tests/latest/test_storage_account_scenarios.py
```diff
@@ -137,6 +137,15 @@ def test_create_storage_account(self, resource_group, location):
         self.cmd('storage account check-name --name {}'.format(name),
                  checks=JMESPathCheck('nameAvailable', True))
 
+        large_file_name = self.create_random_name(prefix='cli', length=24)
+        self.cmd('storage account create -g {} -n {} --sku {} --enable-large-file-share'.format(
+            resource_group, large_file_name, 'Standard_LRS'))
+        self.cmd('az storage account show -n {} -g {}'.format(large_file_name, resource_group), checks=[
+            JMESPathCheck('name', large_file_name),
```
qwordy: `self.check('name', large_file_name),`
is also OK.

src/azure-cli/azure/cli/command_modules/storage/tests/latest/test_storage_account_scenarios.py
```diff
@@ -137,6 +137,15 @@ def test_create_storage_account(self, resource_group, location):
         self.cmd('storage account check-name --name {}'.format(name),
                  checks=JMESPathCheck('nameAvailable', True))
 
+        large_file_name = self.create_random_name(prefix='cli', length=24)
+        self.cmd('storage account create -g {} -n {} --sku {} --enable-large-file-share'.format(
+            resource_group, large_file_name, 'Standard_LRS'))
```
qwordy: "rg" is predefined in ResourceGroupPreparer.
You don't need to type resource_group
e.g. `self.cmd('vm create -g {rg} -n {vm} --image UbuntuLTS')`

src/azure-cli/azure/cli/command_modules/storage/_params.py
```diff
@@ -67,11 +67,11 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals, too-many-statem
         validator=validate_storage_data_plane_list)
 
     large_file_share_type = CLIArgumentType(
-        action='store_true', min_api='2019-04-01',
+        action='store_true', min_api='2019-04-01', is_preview=True,
         help='Enable the capability to support large file shares with more than 5 TiB capacity for storage account.'
-             'Once the property is enabled, the feature cannot be disabled. Currently only supported for LRS and ZRS '
-             'replication types, hence account conversions to geo-redundant accounts would not be possible. For more '
-             'information, please refer to https://go.microsoft.com/fwlink/?linkid=2086047.')
+             'Once the property is enabled, the feature cannot be disabled. Currently onis_preview=True,ly supported '
```
qwordy: Typo?

src/azure-cli/azure/cli/command_modules/storage/operations/account.py
```diff
@@ -145,7 +149,9 @@ def update_storage_account(cmd, instance, sku=None, tags=None, custom_domain=Non
             directory_service_options='AADDS' if enable_files_aadds else 'None')
     if assign_identity:
         params.identity = Identity()
-
+    if enable_large_file_share:
+        LargeFileSharesState = cmd.get_models('LargeFileSharesState')
```
MyronFanQiu: why don't combine this into line 116?

src/azure-cli/azure/cli/command_modules/storage/_params.py
Juliehzl: Yes...

src/azure-cli/azure/cli/command_modules/storage/tests/latest/test_storage_account_scenarios.py
Juliehzl: check is also for JMESPathCheck.

src/azure-cli/azure/cli/command_modules/storage/tests/latest/test_storage_account_scenarios.py
Juliehzl: I know. But in this function, it is already introduced as resource_group. I don't want to change the convention now.

Issue #10702, opened 2019-09-30
Enable setting "auto pause" value when creating databases


mrwizard82d1:
**Is your feature request related to a problem? Please describe.**
I want to create a SQL database with a non-default auto-pause delay. I see no option to accomplish via azure cli (although it appears available via Powershell).

**Describe the solution you'd like**
I'd like to run the command
`az sql db create ... --auto-pause-delay-hours 2` to set the auto-pause delay to 2 hours

**Describe alternatives you've considered**
I can use Powershell; however, all my current scripts use Python with the azure-cli package. Using Powershell requires users to install yet another tool / module to create the database.

**Additional context**
I've found the azure CLI and the underlying Python code very useful in scripting deployments. Having to move to Powershell for this single piece is disappointing.

jaredmoo:
Thank you for the feedback! Let me chat with the team and find out what our plans are.

jaredmoo:
Looks like we plan to implement this in the next 1-2 months :)

jeroenterheerdt:
since this was implemented, let's go ahead and close this @jaredmoo

PR #10788, opened 2019-10-09
Serverless cli

weizeng-msft:
---


src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -149,6 +149,10 @@ def get_location_type_with_default_from_resource_group(cli_ctx):
     arg_group=sku_arg_group,
     options_list=['--elastic-pool'])
 
+compute_model_param_type = CLIArgumentType(
+    arg_group=sku_arg_group,
+    options_list=['--compute-model'])
```
jaredmoo: Can you add enum value list here?

jaredmoo: Actually, I think you don't want to use `get_enum_type` here (see comment at https://github.com/Azure/azure-cli/blob/29527352eafdebf6b39a66f7239cf56507347667/src/azure-cli/azure/cli/command_modules/sql/_params.py#L468 ). CLI core team, is there any way for a parameter to have autocomplete for suggested possible values, but still allow user to enter values that are not suggested?

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -277,6 +284,10 @@ def _configure_db_create_params(
     arg_ctx.argument('elastic_pool_id',
                      arg_type=elastic_pool_id_param_type,
                      help='The name or resource id of the elastic pool to create the database in.')
+    
+    arg_ctx.argument('compute_model',
+                     arg_type=compute_model_param_type,
+                     help='The compute model of this database. Currently we support Provisioned and Serverless.')
```
jaredmoo: `The compute model of the database. Allowed values include: Provisioned, Serverless.`

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -314,6 +325,15 @@ def _configure_db_create_params(
         # Provisioning with capacity is not applicable to DataWarehouse
         arg_ctx.ignore('capacity')
 
+        # Provisioning with autoPauseDelay is not applicable to DataWarehouse
```
jaredmoo: Consider grouping these 3 under 1 comment that just says DW doesn't support serverless

src/azure-cli/azure/cli/command_modules/sql/_params.py
jaredmoo: I think you can put this help text into `compute_model_param_type` so that it's not repeated later.

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -478,6 +498,16 @@ def load_arguments(self, _):
...
+        c.argument('compute_model',
+                    arg_type=compute_model_param_type,
+                    help='The compute model of this database. Currently we support Provisioned and Serverless.')
+    
+        c.argument('auto_pause_delay', 
+                    help='Time in minutes after which database is automatically paused. A value of -1 means that automatic pause is disabled.')
+
+        c.argument('min_capacity', 
+                    help='Minimal capacity that database will always have allocated, if not paused')
+
```
jaredmoo: What arg group are these all in? Should auto_pause_delay & min_capacity be in a new arg group in order to demonstrate that they are both related to serverless?

weizeng-msft:
@johnpaulkee

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -243,7 +243,14 @@ def _find_performance_level_capability(sku, supported_service_level_objectives,
         try:
             # Find requested service objective based on capacity & family.
             # Note that for non-vcore editions, family is None.
-            return next(slo for slo in supported_service_level_objectives
+            if requesting_serverless :
+                return next(slo for slo in supported_service_level_objectives
+                        if ((slo.sku.family == sku.family) or
+                            (slo.sku.family is None and allow_reset_family)) and
+                        int(slo.sku.capacity) == int(sku.capacity) and
+                        "_S_" in slo.sku.name)
```
jaredmoo: Instead of copy-pasting, just change existing code to 

```
                return next(slo for slo in supported_service_level_objectives
                        if ((slo.sku.family == sku.family) or
                            (slo.sku.family is None and allow_reset_family)) and
                        int(slo.sku.capacity) == int(sku.capacity) and
                        (not requesting_serverless or "_S_" in slo.sku.name)
```

jaredmoo: Is "_S_" really the only way we have of distinguishing serverless? This seems fragile. Doesn't capability have compute model or min capacity specified?

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -627,6 +635,10 @@ def _db_dw_create(
     Handles common concerns such as setting location and sku properties.
     '''
 
+    if kwargs['compute_model'] == "Serverless":
+        if sku == None or sku.tier == None or sku.family == None or sku.capacity == None:
+            raise CLIError('When creating a severless database, please pass in edition, family, and capacity parameters through -e -f -c')
```
jaredmoo: This check should not be needed. Tier, family, and capacity can all be specified in the sku name, in which case they don't need to be explicitly provided. This is why we have all the find_capability logic.

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -226,7 +226,7 @@ def _find_family_capability(sku, supported_families):
         return _get_default_capability(supported_families)
 
 
-def _find_performance_level_capability(sku, supported_service_level_objectives, allow_reset_family):
+def _find_performance_level_capability(sku, supported_service_level_objectives, allow_reset_family, requesting_serverless=False):
```
jaredmoo: This is not extensible to other compute models. Please just pass in `compute_model` as parameter.

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -639,7 +651,10 @@ def _db_dw_create(
 
     # If sku.name is not specified, resolve the requested sku name
     # using capabilities.
-    kwargs['sku'] = _find_db_sku_from_capabilities(cli_ctx, kwargs['location'], sku)
+    kwargs['sku'] = _find_db_sku_from_capabilities(cli_ctx, 
+        kwargs['location'], 
+        sku, 
+        requesting_serverless=(kwargs['compute_model'] == "Serverless"))
```
jaredmoo: Strings should not be hardcoded. If needed (which I hope not, because you should just compare capability's compute model with requested compute model), then use enum from SDK

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -1166,6 +1184,30 @@ def db_update(
...
+            server_name=server_name,
+            resource_group_name=resource_group_name)
+
+        # reset sku name for update
+        instance.sku.name = ""
+
+        instance.sku = _find_db_sku_from_capabilities(cmd.cli_ctx, 
+            server_location, 
+            instance.sku, 
+            requesting_serverless= compute_model == "Serverless")
```
jaredmoo: Why is this a separate step from `_db_elastic_pool_update_sku` ?

jaredmoo:
Fixes #10702

src/azure-cli/azure/cli/command_modules/sql/custom.py
weizeng-msft: I do not think we have a compute model in the capabilities, we have a different SLO name for serverless, but we did not introduce a new compute model just for it.

src/azure-cli/azure/cli/command_modules/sql/custom.py
weizeng-msft: yes, S is the only differentiation

src/azure-cli/azure/cli/command_modules/sql/custom.py
weizeng-msft: make sense, I think that is a good idea. I will update it

src/azure-cli/azure/cli/command_modules/sql/custom.py
jaredmoo: This is a problem. CLI (or any other client) really should discover capabilities with capabilities API. Parsing SKU name is not maintainable.

weizeng-msft: Agreed, but I think it was decided in some other design. Powershell is doing the same for this as well. Maybe we should add an item for repair?

jaredmoo: Sorry, I will not knowingly accept new technical debt into CLI.

weizeng-msft: Understood, and somewhat agree. We might need to discuss with the stakeholders on this issue.

src/azure-cli/azure/cli/command_modules/sql/_params.py
weizeng-msft: agreed!

src/azure-cli/azure/cli/command_modules/sql/custom.py
weizeng-msft: Yes, this should be part of the function. I have added it there.

src/azure-cli/azure/cli/command_modules/sql/_params.py
weizeng-msft: great idea, I grouped them, and added help text in _help.py as well

src/azure-cli/azure/cli/command_modules/sql/_params.py
weizeng-msft: I would say get_enum_type is not bad. If we were to introduce new compute type, we just need to update the enum to take in that from CLI side. it is trivial change to me.

weizeng-msft: I actually think it might be a good idea to add enum type here. Later if we want to introduce new types of offering, we will add it to the enum. Should be trivial change for CLI.

yonzhan:

src/azure-cli/azure/cli/command_modules/sql/custom.py
jaredmoo: Avoiding this will introduce further bugs, so this is the best option for now until capabilities api is improved. I would like to prioritise fixing this debt asap.

Juliehzl:
Hi @weizeng-msft , can you try to make CI pass?

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -150,6 +153,23 @@ def get_location_type_with_default_from_resource_group(cli_ctx):
...
+compute_model_param_type = CLIArgumentType(
+    arg_group=serverless_arg_group,
+    options_list=['--compute-model'],
+    help='The compute model of the database.',
+    arg_type=get_enum_type(ComputeModelType))
+
+auto_pause_delay_param_type = CLIArgumentType(
+    arg_group=serverless_arg_group,
+    options_list=['--auto-pause-delay'],
+    help='Time in minutes after which database is automatically paused. '
```
Juliehzl: If I don't specify this parameter, which value will be used? That is, what is the default value?

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -637,6 +666,13 @@ def _db_dw_create(
     Handles common concerns such as setting location and sku properties.
     '''
 
+    # This check needs to be here, because server side logic of
+    # finding a default sku for Serverless is not yet implemented.
+    if kwargs['compute_model'] == ComputeModelType.serverless:
+        if sku is None or sku.tier is None or sku.family is None or sku.capacity is None:
+            raise CLIError('When creating a severless database, please pass in edition, '
```
Juliehzl: I have a small question here.  When  losting sku value, why do you let user pass -e -f and -c? How is it related to sku?

src/azure-cli/azure/cli/command_modules/sql/_params.py
weizeng-msft: like the other commands above, it will be null, and it is the server code's responsibility to handle the case of null. Currently it is working as expected.

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -150,6 +153,23 @@ def get_location_type_with_default_from_resource_group(cli_ctx):
...
+    'A value of -1 means that automatic pause is disabled.')
+
+min_capacity_param_type = CLIArgumentType(
+    arg_group=serverless_arg_group,
+    options_list=['--min-capacity'],
+    help='Minimal capacity that database will always have allocated, if not paused')
```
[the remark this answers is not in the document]
weizeng-msft: No we do not have a default value for this parameter

src/azure-cli/azure/cli/command_modules/sql/custom.py
weizeng-msft: This is a design issue from the server side code. If we want a Serverless compute model database, we actually do not have a default edition (-e), family (-f), and capacity (-c) configured for the customer. We are making a server side improvement for that, but at the mean time, letting customer provide the edition, family, and capacity as input params should be the most desired experience.

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -55,7 +55,8 @@
     ClientType,
     DatabaseCapabilitiesAdditionalDetails,
     ElasticPoolCapabilitiesAdditionalDetails,
-    FailoverPolicyType
+    FailoverPolicyType,
+    ComputeModelType
```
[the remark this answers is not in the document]
weizeng-msft: will do!

weizeng-msft:
> Hi @weizeng-msft , can you try to make CI pass?
> In addition, you need to update HISTORY.rst file for your new command.

I have just updated HISTORY.rst

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -319,12 +341,18 @@ def _db_elastic_pool_update_sku(
     if capacity:
         instance.sku.capacity = capacity
 
+    # Wipe out sku name if serverless vs provisioned db offerings changed,
+    # only if sku name has not be wiped by earlier logic, and new compute model has been requested.
+    if instance.sku.name is not None and compute_model is not None:
```
jaredmoo: It's more pythonic to implicitly convert null check to boolean, i.e. `if instance.sku.name and compute_model:`

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -1152,6 +1201,12 @@ def db_update(
         server_name,
         resource_group_name)
 
+    # Finding out requesting compute_model
+    if compute_model is None:
```
jaredmoo: Similar command : `if not compute_model:`

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -637,6 +672,13 @@ def _db_dw_create(
     Handles common concerns such as setting location and sku properties.
     '''
 
+    # This check needs to be here, because server side logic of
+    # finding a default sku for Serverless is not yet implemented.
+    if kwargs['compute_model'] == ComputeModelType.serverless:
+        if sku is None or sku.tier is None or sku.family is None or sku.capacity is None:
```
jaredmoo: Similar comment about null checks

Juliehzl:
LGTM

PR #10808, opened 2019-10-10
Adds mimetype detection for WebAssembly files

### LINKED ISSUE #8708: When uploading blob .wasm files content type set to application/wasm rather then application/octet-stream
I do publish blazor apps on azure storage accounts from visual studio code.

Uploaded .wasm file's content type is set to application/octet-stream.

Browser expects for .wasm files content type application/wasm, so we are forced to correct later content types using different ways, which is not convenient and requires extra time.

If it is possible please address this issue.

kmfk:
This would resolve #8708.  We've run into the same issue with WebAssembly files being uploaded through the CLI in an Ansible playbook.

---

### DIFF
--- a/src/azure-cli/HISTORY.rst
+++ b/src/azure-cli/HISTORY.rst
@@ -25,7 +25,7 @@ Release History
 
 **Storage**
 
-* Adds support for WebAssembly (.wasm) mimetype detection
+* Add support for WebAssembly (.wasm) mimetype detection
 
 2.0.74
 ++++++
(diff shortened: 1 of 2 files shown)


yonzhan:
@Juliehzl  please take a look at this PR.

src/azure-cli/HISTORY.rst
```diff
@@ -23,6 +23,10 @@ Release History
 
 * az network private-dns link vnet create/update: Fixes #9851. Support cross-tenant virtual network linking.
 
+**Storage**
+
+* Adds support for WebAssembly (.wasm) mimetype detection
```
Juliehzl: LGTM. Can you change 'Adds' to 'Add' here?

kmfk: Done!

PR #10824, opened 2019-10-12
[Storage]: Change help message for --https-only parameter in az storage account create

Juliehzl:
---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/storage/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/storage/_params.py
@@ -116,6 +116,10 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals, too-many-statem
         c.argument('account_name', acct_name_type, options_list=['--name', '-n'], completer=None)
         c.argument('kind', help='Indicates the type of storage account.',
                    arg_type=get_enum_type(t_kind, default='storage'))
+        c.argument('https_only', arg_type=get_three_state_flag(), min_api='2019-04-01',
+                   help='Allow https traffic only to storage service if set to true. The default value is true.')
+        c.argument('https_only', arg_type=get_three_state_flag(), max_api='2018-11-01',
+                   help='Allow https traffic only to storage service if set to true. The default value is false.')
         c.argument('tags', tags_type)
         c.argument('custom_domain', help='User domain assigned to the storage account. Name is the CNAME source.')
         c.argument('sku', help='The storage account SKU.', arg_type=get_enum_type(t_sku_name, default='standard_ragrs'))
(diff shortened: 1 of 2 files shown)


qwordy:
You change help message. Why no deleted lines?

src/azure-cli/azure/cli/command_modules/storage/_params.py
```diff
@@ -116,6 +116,10 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals, too-many-statem
         c.argument('account_name', acct_name_type, options_list=['--name', '-n'], completer=None)
         c.argument('kind', help='Indicates the type of storage account.',
                    arg_type=get_enum_type(t_kind, default='storage'))
+        c.argument('https_only', arg_type=get_three_state_flag(), min_api='2019-04-01',
+                   help='Allow https traffic only to storage service if set to true. The default value is true.')
+        c.argument('https_only', arg_type=get_three_state_flag(), max_api='2018-11-01',
```
MyronFanQiu: why do we need two lines?

Juliehzl: A long story... Service team change the default values for different API version. So I have to split them and give the corresponding help message..

Juliehzl:
> You change help message. Why no deleted lines?

For different AP versionI, it will show up different help message.

src/azure-cli/azure/cli/command_modules/storage/_params.py
qwordy: Same help twice. You can assign it to a variable and reference it.
Same arg_type, help, use CLIArgumentType?

yonzhan:
@Juliehzl please take a look at this PR.

src/azure-cli/azure/cli/command_modules/storage/_params.py
Juliehzl: It is not the same. They are different.

qwordy: 'Allow https traffic only to storage service if set to true. The default value is true.' twice.
Minor issue🙂

MyronFanQiu: the second sentence is "'Allow https traffic only to storage service if set to true. The default value is **false**"

MyronFanQiu: @qwordy

qwordy: Oh! I'll go to ophthalmology department tomorrow.

qwordy:
Why you add `is not None`, but finally delete them?

src/azure-cli/azure/cli/command_modules/storage/_params.py
MyronFanQiu: > ophthalmology

well. You don't need. It's really hard to find.....

Issue #9495, opened 2019-05-29
Behavior of accelerated networking and NICs


dcbrown16:
Question about az vm create.
It appears that when using an existing NIC, the accelerated networking settings of that NIC override the accelerated networking setting specified when creating the VM.
i.e. if you create a NIC with accelerated networking disabled and then use that NIC to create a VM that would otherwise have the setting enabled, the VM is created with the setting disabled.
This specification probably makes the most sense, but if you can confirm that is intentional, can we add language to the --accelerated-networking option to clarify that the setting of any preexisting NIC takes precedence?

--accelerated-networking 
Enable accelerated networking. Unless specified, CLI will enable it based on machine image and size.
accepted values: false, true

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 152b549b-d4b3-13e5-081c-c604b6e4d89e
* Version Independent ID: 3f79be78-ce75-ad31-304b-bad7d475ec9e
* Content: [az vm](https://docs.microsoft.com/en-us/cli/azure/vm?view=azure-cli-latest#az-vm-create)
* Content Source: [src/command_modules/azure-cli-vm/azure/cli/command_modules/vm/_help.py](https://github.com/Azure/azure-cli/blob/dev/src/command_modules/azure-cli-vm/azure/cli/command_modules/vm/_help.py)
* Service: **virtual-machines**
* GitHub Login: @rloutlaw
* Microsoft Alias: **routlaw**

achandmsft:
@adewaleo could you please respond to Dillon's question?

adewaleo:
@dcbrown16 my apologies for the late response. According to this [doc](https://docs.microsoft.com/en-us/azure/virtual-network/create-vm-accelerated-networking-cli#enable-accelerated-networking-on-existing-vms), a vm (and its nic) can be updated to support accelerated networking.

Given this is possible, `vm create` should be able to enable accelerated networking on the existing nic, when `--accelerated-networking` is set to True.

I should be able to update this in an upcoming sprint.

dcbrown16:
@adewaleo I have a customer report that setting --accelerated-networking to true is not honored if an existing NIC with accelerated networking disabled is specified. If the intended specification is to have the NIC switch to enabled in that case, please verify on your end or contact me at v-dibr internally if we need IcM or something.

dcbrown16:
[...]

PR #10832, opened 2019-10-14
[Compute] vm create: Add warning when specifying accelerated networking and an existing NIC together

qwordy:
Issue #9495 

Accelerated networking is an option of NIC. When using an existing NIC, should not specifying accelerated networking. Add warning first, change to error 6 months later - April 2020.

---


src/azure-cli/HISTORY.rst
```diff
@@ -38,6 +38,7 @@ Release History
 * vm create: Use standard public IP SKU automatically when using zones.
 * vm create: Compose a valid computer name from VM name if computer name is not provided.
 * vm create: Add --workspace to enable log analytics workspace automatically.
+* vm create: Do not allow specifying accelerated networking when using an existing NIC.
```
MyronFanQiu: Well. It must be a breaking change. Right? Should we deprecated it first?

qwordy: Kind of. Can't use --accelerated-networking and --nics togother now. But no one is deprecated. What's the workflow of this condition?

qwordy: I think I can use warning first. Then after several releases, change it to error.

qwordy: Resolved. Added a reminder to onenote to change it to error later.

PR #10864, opened 2019-10-16
acr: support "--repositories" on token creation command

yugangw-msft:
This is to support fast ramp up on repository level of access control. One command will set up all for users. 
Notes:
1. For consistency, I also renamed the `--add` to `--repositories` to be consistent with `az acr token create`, but do let me know if you feel the `--add` is better. 
2. When the existing scope-map is detected, if the actions are the same, CLI will accept it; otherwise it will throw. This is to support command level idempotency that you can run the same command again and again.

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acr/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/acr/_params.py
@@ -254,21 +254,32 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
     with self.argument_context('acr scope-map') as c:
         c.argument('registry_name', options_list=['--registry', '-r'])
         c.argument('description', options_list=['--description'], help='Description for the scope map. Maximum 256 characters are allowed.', required=False)
-        c.argument('add_repository', options_list=['--add'], nargs='+', help='Actions to be added. Use the format "--add REPO [ACTION1 ACTION2 ...]" per flag. Valid actions are metadata/read, metadata/write, content/read, content/write and content/delete', action='append', required=False)
-        c.argument('remove_repository', options_list=['--remove'], nargs='+', help='Actions to be removed. Use the format "--remove REPO [ACTION1 ACTION2 ...]" per flag. Valid actions are metadata/read, metadata/write, content/read, content/write and content/delete', action='append', required=False)
         c.argument('scope_map_name', options_list=['--name', '-n'], help='The name of the scope map.', required=True)
 
+    valid_actions = "Valid actions are metadata/read, metadata/write, content/read, content/write and content/delete"
+    with self.argument_context('acr scope-map update') as c:
+        c.argument('add_repository', options_list=['--add'], nargs='+', action='append', required=False,
+                   help='repository permissions to be added. Use the format "--add REPO [ACTION1 ACTION2 ...]" per flag. ' + valid_actions)
+        c.argument('remove_repository', options_list=['--remove'], nargs='+', action='append', required=False,
+                   help='respsitory permissions to be removed. Use the format "--remove REPO [ACTION1 ACTION2 ...]" per flag. ' + valid_actions)
+
     with self.argument_context('acr scope-map create') as c:
-        c.argument('add_repository', options_list=['--add'], nargs='+', help='Actions to be added. Use the format "--add REPO [ACTION1 ACTION2 ...]" per flag. Valid actions are metadata/read, metadata/write, content/read, content/write and content/delete', action='append', required=True)
+        c.argument('repository_actions_list', options_list=['--repository'], nargs='+', action='append', required=True,
+                   help='repository permissions. Use the format "--repository REPO [ACTION1 ACTION2 ...]" per flag. ' + valid_actions)
 
     with self.argument_context('acr token') as c:
         c.argument('registry_name', options_list=['--registry', '-r'])
         c.argument('token_name', options_list=['--name', '-n'], help='The name of the token.', required=True)
         c.argument('scope_map_name', options_list=['--scope-map'], help='The name of the scope map associated with the token', required=False)
-        c.argument('status', options_list=['--status'], help='The status of the token. Allowed values are "enabled" or "disabled".', required=False, default="enabled")
+        c.argument('status', options_list=['--status'], arg_type=get_enum_type(['enabled', 'disabled']),
+                   help='The status of the token', required=False, default="enabled")
 
     with self.argument_context('acr token create') as c:
-        c.argument('scope_map_name', options_list=['--scope-map'], help='The name of the scope map associated with the token', required=True)
+        c.argument('scope_map_name', options_list=['--scope-map'],
+                   help='The name of the scope map with pre-configured repository permissions. Use "--repository" if you would like CLI to configure one for you')
+        c.argument('repository_actions_list', options_list=['--repository'], nargs='+', action='append',
+                   help='repository permissions. Use the format "--repository REPO [ACTION1 ACTION2 ...]" per flag. ' + valid_actions)
+        c.argument('no_passwords', arg_type=get_three_state_flag(), help='Do not generate passwords, instead use "az acr token credentials regenerate"')
 
     with self.argument_context('acr token update') as c:
         c.argument('scope_map_name', options_list=['--scope-map'], help='The name of the scope map associated with the token. If not specified, running this command will disassociate the current scope map related to the token.', required=False)
--- a/src/azure-cli/azure/cli/command_modules/acr/tests/latest/test_acr_token.py
+++ b/src/azure-cli/azure/cli/command_modules/acr/tests/latest/test_acr_token.py
@@ -1,0 +1,45 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+from azure_devtools.scenario_tests import AllowLargeResponse
+from azure.cli.testsdk import ScenarioTest, ResourceGroupPreparer
+
+
+class AcrTokenCommandsTests(ScenarioTest):
+
+    @AllowLargeResponse()
+    @ResourceGroupPreparer()
+    def test_repository_token_create(self):
+        self.kwargs.update({
+            'registry': self.create_random_name('clireg', 20),
+            'scope_map': 'scope-map',
+            'token': 'acr-token',
+            'token2': 'token-2'
+        })
+        self.cmd('acr create -g {rg} -n {registry} --sku premium')
+
+        # quick create
+        # ACR ever verifies the existence of the repository, hence we will feed a fake
+        self.cmd('acr token create -r {registry} -n {token} --repository foo content/read', checks=[
+            self.check('status', 'enabled'),
+            self.check('credentials.passwords[0].name', 'password1'),
+            self.check('credentials.passwords[1].name', 'password2'),
+            self.check('credentials.username', self.kwargs['token'])
+        ])
+        self.cmd('acr scope-map show -r {registry} -n {token}-scope-map', checks=[
+            self.check('actions', ['repositories/foo/content/read'])
+        ])
+
+        # create from an existing scope map
+        self.cmd('acr scope-map create -r {registry} -n {scope_map} --repository foo content/read metadata/read', checks=[
+            self.check('actions', ['repositories/foo/content/read', 'repositories/foo/metadata/read'])
+        ])
+        self.cmd('acr token create -r {registry} --scope-map {scope_map} -n {token2} --no-passwords', checks=[
+            self.check('credentials.passwords', [])
+        ])
+        self.cmd('acr token credential generate -r {registry} -n {token2} --password1 --password2', checks=[
+            self.check('passwords[0].name', 'password1'),
+            self.check('passwords[1].name', 'password2')
+        ])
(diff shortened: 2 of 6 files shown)


src/azure-cli/azure/cli/command_modules/acr/token.py
```diff
@@ -39,6 +50,53 @@ def acr_token_create(cmd,
...
+    token = LongRunningOperation(cmd.cli_ctx)(poller)
+    _ensure_default_passwords(cmd, resource_group_name, registry_name, token, logger)
+    return token
+
+
+def _ensure_default_scope_map(cmd, resource_group_name, registry_name, token_name, repositories, logger):
+    import re
+    from ._client_factory import cf_acr_scope_maps
+    scope_map_name = DEF_SCOPE_MAP_NAME_TEMPLATE.format(token_name)
+    scope_map_name = re.sub(r'[^a-zA-Z0-9-]', '-', scope_map_name)
```
hkuadithya: `scope_map_name` is derived from `<tokenName>-scope-map`.
If there are non permissible characters found in the `<tokenName>`, the token creation itself will fail.

Therefore, it might lead to a case where a scope map gets created but the token creation fails.

@ankurkhemani any thoughts?

yugangw-msft: Do token and scope-name share the same character set?

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -253,21 +253,32 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
...
+        c.argument('repositories', nargs='+', action='append', required=True,
+                   help='repository permissions. Use the format "--repositories REPO [ACTION1 ACTION2 ...]" per flag. ' + valid_actions)
 
     with self.argument_context('acr token') as c:
         c.argument('registry_name', options_list=['--registry', '-r'])
         c.argument('token_name', options_list=['--name', '-n'], help='The name of the token.', required=True)
         c.argument('scope_map_name', options_list=['--scope-map'], help='The name of the scope map associated with the token', required=False)
-        c.argument('status', options_list=['--status'], help='The status of the token. Allowed values are "enabled" or "disabled".', required=False, default="enabled")
+        c.argument('status', options_list=['--status'], arg_type=get_enum_type(['enabled', 'disabled']),
+                   help='The status of the token', required=False, default="enabled")
```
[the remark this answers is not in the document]
yugangw-msft: I am using the enum mainly to leverage CLI's built-in support to show the possible values to user.
Please let me know if I misunderstood.

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -253,21 +253,32 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
...
 
     with self.argument_context('acr token create') as c:
-        c.argument('scope_map_name', options_list=['--scope-map'], help='The name of the scope map associated with the token', required=True)
+        c.argument('scope_map_name', options_list=['--scope-map'],
+                   help='The name of the scope map with pre-configured repository permissions. Use "--repositories" if you would like CLI to configure one for you')
+        c.argument('repositories', nargs='+', action='append',
+                   help='repository permissions. Use the format "--repositories REPO [ACTION1 ACTION2 ...]" per flag. ' + valid_actions)
```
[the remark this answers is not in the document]
yugangw-msft: I am changing it to be `--repository` like you recommended which will be consistent with others

src/azure-cli/azure/cli/command_modules/acr/token.py
```diff
@@ -2,34 +2,45 @@
...
+                     repositories=None,
                      status=None,
-                     resource_group_name=None):
+                     resource_group_name=None,
+                     no_passwords=None):
+    from knack.log import get_logger
+    from ._utils import get_resource_id_by_registry_name
 
-    validate_premium_registry(cmd, registry_name, resource_group_name)
+    if bool(repositories) == bool(scope_map_name):
```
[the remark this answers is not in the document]
yugangw-msft: it is possible that both `repositories` and `scope_map_name` could be none, which the if statement also catches. The error of `usage error` is a conventional format, hence let us keep as it is.

src/azure-cli/azure/cli/command_modules/acr/_params.py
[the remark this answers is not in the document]
hkuadithya: @yugangw-msft 
My comment was about the argument variable `c.argument('repositories'` not the command line argument.  I suggested changing it to a more descriptive variable such as `c.argument('repository_actions_list'` since it's a list of Repositories and their corresponding actions.

I should've been clearer in the review. It's a nit suggestion. Feel free to change it to a more descriptive variable if possible.

src/azure-cli/azure/cli/command_modules/acr/token.py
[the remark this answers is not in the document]
hkuadithya: I see. It's like an XOR operation.  At least, repositories  or scope_map_name should be provided, but not both.  Looks good. Thanks for the clarification.

src/azure-cli/azure/cli/command_modules/acr/token.py
hkuadithya: Both tokens and scope-,map are ARM resources. Hence, the character set constraints for Scope map and Token is same as the constraints for any other ARM resource.

src/azure-cli/azure/cli/command_modules/acr/_params.py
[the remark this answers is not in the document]
hkuadithya: I see. Thanks for the clarification.

src/azure-cli/azure/cli/command_modules/acr/token.py
hkuadithya: Python SDK already has a validation for Token and Scope map resources (both are ARM resources and thus, there's also and existing ARM resource validation in RP).
In my opinion, we shouldn't perform any client side modifications such as    `scope_map_name = re.sub(r'[^a-zA-Z0-9-]', '-', scope_map_name)`.

If creation of `<tokenName>-scope-map` fails, then the operation should be stopped at that point with a suitable error message. 

@ankurkhemani please suggest changes here if needed.

src/azure-cli/azure/cli/command_modules/acr/token.py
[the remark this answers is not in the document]
hkuadithya: I see. Thanks for clarifying.

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -253,21 +253,32 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
...
+    valid_actions = "Valid actions are metadata/read, metadata/write, content/read, content/write and content/delete"
+    with self.argument_context('acr scope-map update') as c:
+        c.argument('add_repository', options_list=['--add'], nargs='+', action='append', required=False,
+                   help='repository permissions to be added. Use the format "--add REPO [ACTION1 ACTION2 ...]" per flag. ' + valid_actions)
+        c.argument('remove_repository', options_list=['--remove'], nargs='+', action='append', required=False,
+                   help='respsitory permissions to be removed. Use the format "--remove REPO [ACTION1 ACTION2 ...]" per flag. ' + valid_actions)
+
     with self.argument_context('acr scope-map create') as c:
-        c.argument('add_repository', options_list=['--add'], nargs='+', help='Actions to be added. Use the format "--add REPO [ACTION1 ACTION2 ...]" per flag. Valid actions are metadata/read, metadata/write, content/read, content/write and content/delete', action='append', required=True)
+        c.argument('repositories', options_list=['--repository'], nargs='+', action='append', required=True,
```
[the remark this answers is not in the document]
yugangw-msft: Agreed. I will make the change.

src/azure-cli/azure/cli/command_modules/acr/token.py
yugangw-msft: Agreed. I will pull out of the client side modification.

yonzhan:
@qwordy please help with this PR.

yugangw-msft:
@hkuadithya, @ankurkhemani, I have addressed all feedback, to give doc team enough time to update doc, I am going ahead to merge it, feel free to leave comment if you see any other places I should improve

src/azure-cli/azure/cli/command_modules/acr/tests/latest/test_acr_token.py
```diff
@@ -0,0 +1,47 @@
...
+import unittest
+from azure_devtools.scenario_tests import AllowLargeResponse
+from azure.cli.testsdk import ScenarioTest, ResourceGroupPreparer
+
+
+class AcrTokenCommandsTests(ScenarioTest):
+
+    @AllowLargeResponse()
+    @ResourceGroupPreparer()
+    @unittest.skip('blocked till the feature gets deployed to production')
```
yugangw-msft: @hkuadithya, this is the test we can re-record &enable once the deployment to prod is done

src/azure-cli/azure/cli/command_modules/acr/_params.py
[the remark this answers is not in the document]
ankurkhemani: @yugangw-msft You can use the enum from the model and the cli will still show the possible values.
`arg_type=get_enum_type(TokenStatus)`

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -254,21 +254,32 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
...
         c.argument('description', options_list=['--description'], help='Description for the scope map. Maximum 256 characters are allowed.', required=False)
-        c.argument('add_repository', options_list=['--add'], nargs='+', help='Actions to be added. Use the format "--add REPO [ACTION1 ACTION2 ...]" per flag. Valid actions are metadata/read, metadata/write, content/read, content/write and content/delete', action='append', required=False)
-        c.argument('remove_repository', options_list=['--remove'], nargs='+', help='Actions to be removed. Use the format "--remove REPO [ACTION1 ACTION2 ...]" per flag. Valid actions are metadata/read, metadata/write, content/read, content/write and content/delete', action='append', required=False)
         c.argument('scope_map_name', options_list=['--name', '-n'], help='The name of the scope map.', required=True)
 
+    valid_actions = "Valid actions are metadata/read, metadata/write, content/read, content/write and content/delete"
+    with self.argument_context('acr scope-map update') as c:
+        c.argument('add_repository', options_list=['--add'], nargs='+', action='append', required=False,
```
ankurkhemani: So we are still using `--add` `--remove` in update command, but `--repository` in create?

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -254,21 +254,32 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
     with self.argument_context('acr scope-map') as c:
         c.argument('registry_name', options_list=['--registry', '-r'])
         c.argument('description', options_list=['--description'], help='Description for the scope map. Maximum 256 characters are allowed.', required=False)
-        c.argument('add_repository', options_list=['--add'], nargs='+', help='Actions to be added. Use the format "--add REPO [ACTION1 ACTION2 ...]" per flag. Valid actions are metadata/read, metadata/write, content/read, content/write and content/delete', action='append', required=False)
-        c.argument('remove_repository', options_list=['--remove'], nargs='+', help='Actions to be removed. Use the format "--remove REPO [ACTION1 ACTION2 ...]" per flag. Valid actions are metadata/read, metadata/write, content/read, content/write and content/delete', action='append', required=False)
         c.argument('scope_map_name', options_list=['--name', '-n'], help='The name of the scope map.', required=True)
 
+    valid_actions = "Valid actions are metadata/read, metadata/write, content/read, content/write and content/delete"
```
ankurkhemani: Would prefer using a defined constant or enum for the actions.

src/azure-cli/azure/cli/command_modules/acr/token.py
```diff
@@ -39,6 +51,51 @@ def acr_token_create(cmd,
...
+        existing_scope_map = scope_map_client.get(resource_group_name, registry_name, scope_map_name)
+        # for command idempotency, if the actions are the same, we accept it
+        if sorted(existing_scope_map.actions) == sorted(actions):
+            return existing_scope_map.id
+        raise CLIError('The default scope map was already configured with different respository permissions. Please'
+                       ' use "az acr scope-map update -r {} -n {} --add <REPO> --remove <REPO>" to update'.format(
+                           registry_name, scope_map_name))
+    except CloudError:
+        pass
+    logger.warning('Creating a scope map of "%s" for provided respository permissions', scope_map_name)
```
ankurkhemani: typo: `repository`

src/azure-cli/azure/cli/command_modules/acr/token.py
```diff
@@ -39,6 +51,51 @@ def acr_token_create(cmd,
...
+    poller = scope_map_client.create(resource_group_name, registry_name, scope_map_name,
+                                     actions, "Token {}'s scope map".format(token_name))
```
ankurkhemani: Not sure if the scope-map `description` is generic enough since one scope-map can be tied to multiple tokens.
How about: `"Created by token: {}".format(token_name)`

ankurkhemani:
@yugangw-msft Added some minor comments. LGTM mostly. 

Thanks for this feedback and changes! 
It specially makes sense to also provide one command for token creation and do the magic underneath.

src/azure-cli/azure/cli/command_modules/acr/_params.py
hkuadithya: Done.

src/azure-cli/azure/cli/command_modules/acr/_params.py
hkuadithya: I see your point. I haven't changed this for now in the new PR.

PR #10865, opened 2019-10-16
enable ip whitelisting for create and update

gtracer:
---


yonzhan:
@qwordy please help with this PR.

src/azure-cli/azure/cli/command_modules/acs/_help.py
```diff
@@ -4,7 +4,7 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 
-from knack.help_files import helps  # pylint: disable=unused-import
+from knack.help_files import helps  # pylint: disable=import-error
```
qwordy: Why disable import-error

src/azure-cli/azure/cli/command_modules/acs/_validators.py
```diff
@@ -73,6 +75,31 @@ def validate_create_parameters(namespace):
...
+        raise CLIError('--api-server-authorized-ip-ranges can only be used with standard load balancer')
+
+    return validate_ip_ranges(namespace)
+
+
+def validate_ip_ranges(namespace):
+    if not namespace.api_server_authorized_ip_ranges:
+        return
+
+    allowAllTraffic = "0.0.0.0/32"
```
qwordy: Python style: allow_all_traffic

src/azure-cli/azure/cli/command_modules/acs/_validators.py
```diff
@@ -73,6 +75,31 @@ def validate_create_parameters(namespace):
         raise CLIError('--dns-prefix has no value')
 
 
+def validate_ip_ranges_on_create(namespace):
+    if (namespace.api_server_authorized_ip_ranges and
+            namespace.load_balancer_sku and
+            namespace.load_balancer_sku.lower() == "basic"):
```
qwordy: Does python SDK has a enum to lb sku? If so, use it instead of a string literal.

qwordy:
I prefer to use None in if statement explicitly. Like `if var is None` or `if var is not None`. Be careful.

src/azure-cli/HISTORY.rst
```diff
@@ -9,6 +9,7 @@ Release History
 * Set `--vm-set-type` default value to virtualmachinescalesets if supported by the kubernetes version
 * Add `az aks nodepool add`,`az aks nodepool show`, `az aks nodepool list`, `az aks nodepool scale`, `az aks nodepool upgrade`, `az aks nodepool update` and `az aks nodepool delete` commmands to support multiple nodepools in aks
 * Add `--zones` to `az aks create` and `az aks nodepool add` commands to support availability zones for aks
+* Enable GA support of apiserver authorized IP ranges
```
qwordy: Add "--api-server-authorized-ip-ranges" into this history entry.

src/azure-cli/azure/cli/command_modules/acs/_help.py
```diff
@@ -357,6 +365,12 @@
     text: az aks update -g MyResourceGroup -n MyManagedCluster --load-balancer-outbound-ip-prefixes <ip-prefix-resource-id-1,ip-prefix-resource-id-2>
   - name: Attach AKS cluster to ACR by name "acrName"
     text: az aks update -g MyResourceGroup -n MyManagedCluster --attach-acr acrName
+  - name: Update a kubernetes cluster with authorized apiserver ip ranges.
+    text: az aks update -g MyResourceGroup -n MyManagedCluster --api-server-authorized-ip-ranges 193.168.1.0/24,194.168.1.0/24
+  - name: Disable authorized apiserver ip ranges feature for a kubernetes cluster.
+    text: az aks update -g MyResourceGroup -n MyManagedCluster --api-server-authorized-ip-ranges ""
+  - name: Allow all traffic to an apiserver in a kubernetes cluster but don't disable authorized ip ranges feature to allow for future use.
+    text: az aks update -g MyResourceGroup -n MyManagedCluster --api-server-authorized-ip-ranges 0.0.0.0/32
```
qwordy: Great examples!

qwordy: Add testcases if possible.

src/azure-cli/azure/cli/command_modules/acs/_validators.py
gtracer: Good point, thanks - will update

src/azure-cli/azure/cli/command_modules/acs/_validators.py
gtracer: Good thinking, will check

src/azure-cli/HISTORY.rst
gtracer: Ok, will do

src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -1961,18 +1971,20 @@ def aks_update(cmd, client, resource_group_name, name,
                load_balancer_outbound_ip_prefixes=None,
                attach_acr=None,
                detach_acr=None,
+               api_server_authorized_ip_ranges=None,
```
xuto2: should we also validate no IPWL if cluster is Basic LB SKU? Or we rely on server side to validate?

I know we still want to support existing Basic LB cluster opted in the IPWL before GA, so not sure if there is a way to distinguish such case from trying to update IPWL for BLB cluster created after GA

src/azure-cli/azure/cli/command_modules/acs/_params.py
```diff
@@ -190,16 +190,18 @@ def load_arguments(self, _):
         c.argument('vnet_subnet_id')
         c.argument('workspace_resource_id')
         c.argument('skip_subnet_role_assignment', action='store_true')
+        c.argument('api_server_authorized_ip_ranges', type=str, validator=validate_ip_ranges_on_create)
```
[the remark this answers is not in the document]
gtracer: Interesting, I was told we would always be defaulting to standard now. I will update the validator to only check when k8 version >= 1.13.0. 

I want to keep it in the validator though instead of the core logic - I think the separate will lead to a cleaner, more re-useable code base

src/azure-cli/azure/cli/command_modules/acs/custom.py
gtracer: We validate it on the server side in sync pathway currently.

Don't quite follow the latter part of your comment - my understanding is you must have standard lb for IPW. @ruchikaguptaa is that right?

src/azure-cli/azure/cli/command_modules/acs/_params.py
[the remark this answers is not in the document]
gtracer: Actually I think this is still right. The validation only disallows basic LB with IPWL which is an illegal operation. If IPWL isn't provided then basic will be rightfully allowed

src/azure-cli/azure/cli/command_modules/acs/_help.py
gtracer: > I prefer to use None in if statement explicitly. Like `if var is None` or `if var is not None`. Be careful.

Ok - I will double check my code. I attempted to be careful to only use `if X, or if not X` in scenarios where I was comfortable checking for all truthy / falsy and where it made my code cleaner (e.g. avoiding having to do if not none and if not ""). But I will double check to make sure I didn't slip up on any of them.

src/azure-cli/azure/cli/command_modules/acs/_validators.py
gtracer: Ok, it looks to me like there is an enum for the standard sku but not basic. https://github.com/Azure/azure-sdk-for-python/blob/1dbf2d60c0de746a84465399354e743b3ac575c7/sdk/aks/azure-mgmt-devspaces/azure/mgmt/devspaces/models/dev_spaces_management_client_enums.py#L26

We could check against the standard sku and assume that if the customer specified standard then they explicitly didn't specify basic but I am hesitant for two reasons:
1. Capitalization, now I need to set the case to be upper case first letter and lower all which requires unneeded code
2. If we introduce a third sku in the future (e.g. premium) then this work around will fail for the new sku 

I prefer to leave it but open to your thoughts if you feel strongly about using the existing enum

src/azure-cli/azure/cli/command_modules/acs/custom.py
ruchikaguptaa: If you have a pre-GA cluster with BLB then you should be able to update the IPWL. The way we do that is check if the AuthorizedIPRange in MC has len > 0 and if BLB then we allow. Any new cluster that doesn't have IPWL enabled but has BLB will be blocked that way. We check that in RP sync, maybe we can skip it here.

src/azure-cli/azure/cli/command_modules/acs/_validators.py
qwordy: No problem.

src/azure-cli/azure/cli/command_modules/acs/_help.py
```diff
@@ -305,6 +305,9 @@
   - name: --attach-acr
     type: string
     short-summary: Grant the 'acrpull' role assignment to the ACR specified by name or resource ID.
+  - name: --api-server-authorized-ip-ranges
+    type: str
```
robbiezhang: string

src/azure-cli/azure/cli/command_modules/acs/_help.py
```diff
@@ -305,6 +305,9 @@
...
+    short-summary: Comma seperated list of authorized apiserver IP ranges. Set to "" to disable. If left unset, all external traffic will be allowed.
```
robbiezhang: "Set to "" to disable. If left unset, all external traffic will be allowed." feels like 2 options, but they are the same thing, right?

src/azure-cli/azure/cli/command_modules/acs/_help.py
```diff
@@ -348,6 +353,9 @@
   - name: --detach-acr
     type: string
     short-summary: Disable the 'acrpull' role assignment to the ACR specified by name or resource ID.
+  - name: --api-server-authorized-ip-ranges
+    type: str
+    short-summary: Comma seperated list of authorized apiserver IP ranges. Set to "" to disable. Set to 0.0.0.0/32 to allow all traffic in a previously restricted cluster.
```
robbiezhang: Set to 0.0.0.0/32 to disallow all traffic.

src/azure-cli/azure/cli/command_modules/acs/_help.py
```diff
@@ -357,6 +365,12 @@
     text: az aks update -g MyResourceGroup -n MyManagedCluster --load-balancer-outbound-ip-prefixes <ip-prefix-resource-id-1,ip-prefix-resource-id-2>
   - name: Attach AKS cluster to ACR by name "acrName"
     text: az aks update -g MyResourceGroup -n MyManagedCluster --attach-acr acrName
+  - name: Update a kubernetes cluster with authorized apiserver ip ranges.
+    text: az aks update -g MyResourceGroup -n MyManagedCluster --api-server-authorized-ip-ranges 193.168.1.0/24,194.168.1.0/24
+  - name: Disable authorized apiserver ip ranges feature for a kubernetes cluster.
+    text: az aks update -g MyResourceGroup -n MyManagedCluster --api-server-authorized-ip-ranges ""
+  - name: Allow all traffic to an apiserver in a kubernetes cluster but don't disable authorized ip ranges feature to allow for future use.
```
robbiezhang: this message feels obscure. I'd prefer "disallow all external traffic to the apiserver", thoughts?

src/azure-cli/azure/cli/command_modules/acs/custom.py
gtracer: Sounds good - thanks for the detailed response Ruchika

src/azure-cli/azure/cli/command_modules/acs/_help.py
gtracer: Your right - chatted with Saurya, I misunderstood the difference between 0.0.0.0/32 and "". Here is how it works:
1. If you have IPWL enabled and you want to disable it then use "". When disabled then all traffic can access your api server (still protected by kubeconfig and RBAC of course)
2. If you want to restrict api server traffic to only your node pools then set IPWL to 0.0.0.0/32

Updating the help section now

src/azure-cli/azure/cli/command_modules/acs/_help.py
gtracer: The empty string is to disable IPWL for a previously enabled cluster. Since IPWL isn't a default behavior we should only let users set "" for IPWL on update

src/azure-cli/azure/cli/command_modules/acs/_help.py
gtracer: Yup, your right - this wording is off. Updating the wording

qwordy:
@gtracer Add test cases because new parameters are added. Thanks.

gtracer-msft:
> @gtracer Add test cases because new parameters are added. Thanks.

@qwordy can you be more specific about your requirements for unit tests? As you can see from our CLI code there hasn't been a single test added in the last year so I want to avoid building out our entire testing infra in this PR. If there is something in particular you require to be unit tested due to the new param?

To me, the highest value add for testing would be to ensure the helpers/validators that were created (the main purpose of this PR) were tested.

src/azure-cli/HISTORY.rst
```diff
@@ -9,6 +9,7 @@ Release History
 * Set `--vm-set-type` default value to virtualmachinescalesets if supported by the kubernetes version
 * Add `az aks nodepool add`,`az aks nodepool show`, `az aks nodepool list`, `az aks nodepool scale`, `az aks nodepool upgrade`, `az aks nodepool update` and `az aks nodepool delete` commmands to support multiple nodepools in aks
 * Add `--zones` to `az aks create` and `az aks nodepool add` commands to support availability zones for aks
+* Enable GA support of apiserver authorized IP ranges via paramater `--api-server-authorized-ip-ranges` in `az aks create` and `az aks update
```
ruchikaguptaa: nit: missing backtick(`) after "az aks update"

src/azure-cli/azure/cli/command_modules/acs/_help.py
```diff
@@ -324,6 +327,8 @@
     text: az aks create -g MyResourceGroup -n MyManagedCluster --load-balancer-outbound-ip-prefixes <ip-prefix-resource-id-1,ip-prefix-resource-id-2>
   - name: Create a kubernetes cluster with basic SKU load balancer and AvailabilitySet vm set type.
     text: az aks create -g MyResourceGroup -n MyManagedCluster --load-balancer-sku basic --vm-set-type AvailabilitySet
+  - name: Create a kubernetes cluster with authorized apiserver ip ranges.
```
ruchikaguptaa: nit: upper case "IP" instead of lower case "ip"

src/azure-cli/azure/cli/command_modules/acs/_help.py
```diff
@@ -305,6 +305,9 @@
   - name: --attach-acr
     type: string
     short-summary: Grant the 'acrpull' role assignment to the ACR specified by name or resource ID.
+  - name: --api-server-authorized-ip-ranges
+    type: string
+    short-summary: Comma seperated list of authorized apiserver IP ranges. If left unset, all external traffic will be allowed.
```
ruchikaguptaa: Lets update this to be samilar to L358, except the "" flag. ie.
Comma seperated list of authorized apiserver IP ranges. Set to 0.0.0.0/32 to restrict apiserver traffic to node pools.

src/azure-cli/azure/cli/command_modules/acs/_params.py
```diff
@@ -20,7 +20,7 @@
     validate_create_parameters, validate_k8s_client_version, validate_k8s_version, validate_linux_host_name,
     validate_list_of_integers, validate_ssh_key, validate_connector_name, validate_max_pods, validate_nodepool_name,
     validate_vm_set_type, validate_load_balancer_sku, validate_load_balancer_outbound_ips,
-    validate_load_balancer_outbound_ip_prefixes, validate_taints)
+    validate_load_balancer_outbound_ip_prefixes, validate_taints, validate_ip_ranges, validate_acr, validate_ip_ranges_on_create)
```
ruchikaguptaa: Remove the validator "validate_ip_ranges_on_create" as its not defined.

src/azure-cli/azure/cli/command_modules/acs/_validators.py
```diff
@@ -73,6 +75,22 @@ def validate_create_parameters(namespace):
         raise CLIError('--dns-prefix has no value')
 
 
+def validate_ip_ranges(namespace):
+    if not namespace.api_server_authorized_ip_ranges:
+        return
+
+    allow_all_traffic = "0.0.0.0/32"
```
ruchikaguptaa: "dis"allow_all_traffic

src/azure-cli/azure/cli/command_modules/acs/_validators.py
```diff
@@ -73,6 +75,22 @@ def validate_create_parameters(namespace):
...
+    for ip in namespace.api_server_authorized_ip_ranges.split(','):
+        if ip == allow_all_traffic:
```
ruchikaguptaa: if the user inputs "0.0.0.0/32" then this should be the only IP address, otherwise we should fail. Can we add a check that the len of result of split(',') should be only 1 or else error. We should not continue on L85.

src/azure-cli/azure/cli/command_modules/acs/_validators.py
```diff
@@ -73,6 +75,22 @@ def validate_create_parameters(namespace):
...
+            continue
+        try:
+            ip = ip_network(ip)
+            if not ip.is_global:
```
ruchikaguptaa: Is it easy to add a check for IPv6 address? We can error from CLI itself. If not, then we can rely on the check in RP sync.

src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -1642,31 +1647,11 @@ def aks_create(cmd, client, resource_group_name, name, ssh_key_value,  # pylint:
...
-
-    if not vm_set_type:
-        vm_set_type = "VirtualMachineScaleSets"
-
-    # normalize as server validation is case-sensitive
-    if vm_set_type.lower() == "AvailabilitySet".lower():
-        vm_set_type = "AvailabilitySet"
-
-    if vm_set_type.lower() == "VirtualMachineScaleSets".lower():
-        vm_set_type = "VirtualMachineScaleSets"
```
ruchikaguptaa: Thanks for this!

src/azure-cli/azure/cli/command_modules/acs/_params.py
ruchikaguptaa: Or maybe it makes sense to have another method on_create which does not allow empty string - ""?

src/azure-cli/azure/cli/command_modules/acs/_validators.py
```diff
@@ -73,6 +75,22 @@ def validate_create_parameters(namespace):
...
+                raise CLIError("--api-server-authorized-ip-ranges must be global non-reserved addresses or CIDRs")
+        except ValueError:
```
ruchikaguptaa: Just confirming since not so good with python. Empty string "" will not throw this error right?

qwordy:
> 
> 
> > @gtracer Add test cases because new parameters are added. Thanks.
> 
> @qwordy can you be more specific about your requirements for unit tests? As you can see from our CLI code there hasn't been a single test added in the last year so I want to avoid building out our entire testing infra in this PR. If there is something in particular you require to be unit tested due to the new param?
> 
> To me, the highest value add for testing would be to ensure the helpers/validators that were created (the main purpose of this PR) were tested.

It's OK if aks doesn't have tests all the time. But be careful about quality. Sometimes a small change will break existing use case unexpectedly.
In some modules, there are lots of tests. E.g.
https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/vm/tests/latest/test_vm_commands.py

src/azure-cli/HISTORY.rst
gtracer: Nice eye, impressed - updated

src/azure-cli/azure/cli/command_modules/acs/_help.py
gtracer: Sure - done

src/azure-cli/azure/cli/command_modules/acs/_help.py
gtracer: Done

src/azure-cli/azure/cli/command_modules/acs/_params.py
gtracer: Nice catch, removed this dangling reference. I think we can avoid having another method to catch the "" on create. Technically, if a customer sets that it is ok right? It is essentially enabling IPWL to a state where it doesn't restrict any traffic. 

If you feel strongly about adding it, I am open to it :)

src/azure-cli/azure/cli/command_modules/acs/_validators.py
gtracer: Good catch - set to restrict_traffic_to_agentnodes

src/azure-cli/azure/cli/command_modules/acs/_validators.py
gtracer: Got it - didn't realize that but it makes sense. Added a check / error message

src/azure-cli/azure/cli/command_modules/acs/_validators.py
gtracer: It would but I had originally thought the for loop that splits the string into an array would not split "" into an array of any values -- thereby not entering the for loop. I re-tested that and I was wrong. 

I have set a few new safety checks to guard against a customer setting "" and some other IPs (which is illogical) and also to skip the IP global check for empty string. 

Thanks for being so observant :)

src/azure-cli/azure/cli/command_modules/acs/custom.py
gtracer: Happy to :)

src/azure-cli/azure/cli/command_modules/acs/_validators.py
gtracer: Added :)

src/azure-cli/azure/cli/command_modules/acs/_params.py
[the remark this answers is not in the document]
gtracer: Talked offline. I have changed the validation code to not kick out basic LB until after the LB has been set in custom.py. I have also moved the code to determine the LB type into a helper function for easy of reading / re-use

src/azure-cli/azure/cli/command_modules/acs/_help.py
gtracer: Removed as requested

src/azure-cli/azure/cli/command_modules/acs/_help.py
gtracer: Updated

src/azure-cli/HISTORY.rst
gtracer: Fixed

gtracer:
> > > @gtracer Add test cases because new parameters are added. Thanks.
> > 
> > 
> > @qwordy can you be more specific about your requirements for unit tests? As you can see from our CLI code there hasn't been a single test added in the last year so I want to avoid building out our entire testing infra in this PR. If there is something in particular you require to be unit tested due to the new param?
> > To me, the highest value add for testing would be to ensure the helpers/validators that were created (the main purpose of this PR) were tested.
> 
> It's OK if aks doesn't have tests all the time. But be careful about quality. Sometimes a small change will break existing use case unexpectedly.
> In some modules, there are lots of tests. E.g.
> https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/vm/tests/latest/test_vm_commands.py

That is a fair warning - certainly appreciate how valuable tests are. I have added in tests on the new helper function and validation function that I created.

src/azure-cli/azure/cli/command_modules/acs/_params.py
ruchikaguptaa: I think its fine. Can you just test this scenario that nothing unexpected is happening.

src/azure-cli/azure/cli/command_modules/acs/_validators.py
```diff
@@ -73,6 +75,34 @@ def validate_create_parameters(namespace):
...
+def validate_ip_ranges(namespace):
+    if not namespace.api_server_authorized_ip_ranges:
+        return
+
+    restrict_traffic_to_agentnodes = "0.0.0.0/32"
+    allow_all_traffic = ""
+    ip_ranges = [ip.strip() for ip in namespace.api_server_authorized_ip_ranges.split(",")]
+
+    if restrict_traffic_to_agentnodes in ip_ranges and len(ip_ranges) > 1:
+        raise CLIError("--api-server-authorized-ip-ranges cannot restrict traffic via 0.0.0.0/32 and \
```
ruchikaguptaa: Can we update the error message to be "Setting--api-server-authorized-ip-ranges to 0.0.0.0/32 is not allowed with other IP ranges. Refer to https://aka.ms/aks/whitelist for more details"? I think this is more direct and it matches with the error code in RP sync.

gtracer: Updated

src/azure-cli/azure/cli/command_modules/acs/tests/latest/test_validators.py
```diff
@@ -0,0 +1,52 @@
...
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+import unittest
+import mock
+from azure.cli.core.util import CLIError
+from azure.cli.command_modules.acs import _validators as validators
+
+
+class TestValidateIPRanges(unittest.TestCase):
```
ruchikaguptaa: Nice!

ruchikaguptaa:
One comment regarding updating error string. 
Otherwise /lgtm

src/azure-cli/azure/cli/command_modules/acs/_params.py
gtracer: Will run manually and let you know on teams

PR #10866, opened 2019-10-16
 Add reference to az find from help

mirdaki:
This update informs users about the `az find` command when they use `-h` on a command with examples. The goal is to increase adoption of the command by making more people aware of it. It will print in a format like this:

```
Examples
    Redeploy a VM.
        az vm redeploy -g MyResourceGroup -n MyVm

    Redeploy all VMs in a resource group.
        az vm redeploy --ids $(az vm list -g MyResourceGroup --query "[].id" -o tsv)

For more specific examples, use: az find 'az vm redeploy'
```

As this solely modifies help, I haven't updated the HISTORY.rst file.

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/_help.py
+++ b/src/azure-cli-core/azure/cli/core/_help.py
@@ -84,6 +84,8 @@ class CLIPrintMixin(CLIHelp):
                 _print_indent(u'{0}'.format(e.long_summary), indent)
             _print_indent(u'{0}'.format(e.command), indent)
             print('')
+        indent = 0
+        _print_indent('For more specific examples, use: az find \'az {}\' \n'.format(help_file.command), indent)
 
     @staticmethod
     def _process_value_sources(p):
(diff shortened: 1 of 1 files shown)


mirdaki:
I've added a highlight to the suggestion as requested by @JasonRShaver

src/azure-cli-core/azure/cli/core/_help.py
```diff
@@ -84,6 +85,9 @@ def _print_examples(help_file):
                 _print_indent(u'{0}'.format(e.long_summary), indent)
             _print_indent(u'{0}'.format(e.command), indent)
             print('')
+        indent = 0
+        message = 'For more specific examples, use: az find \'az {}\''.format(help_file.command)
```
yugangw-msft: please use double quotes like `az find "az vm redeploy"`, because single quote won't work on Windows

jiasli:
I would prefer to remove the color so that it is consistent with the previous output.

src/azure-cli-core/azure/cli/core/_help.py
mirdaki: OK I've updated it!

JasonRShaver:
This change (to add highlighting) was made after a conversation with @achandmsft, I'll let you guys figure out what the best option is =)

mirdaki:
Has there been any resolution on the text coloring?

PR #10900, opened 2019-10-17
Use entry_points instead of custom launcher az script

gferon:
When installing `azure-cli` with python3, it was not possible to use `az` (`python` command not found) because the custom `az` and `az.bat` scripts were tied to `python` even when installed with `python3`.

The `entry_points` section of `setup.py` is exactly designed to take care of that, and works out of the box.

As a sidenote, I think the `az.completion.sh` should also be installed in a different path, as it is currently installed under `/usr/local/bin` which won't be picked up by `bash` AFAIK.

---


yonzhan:
@fengzhou-msft please help with the PR.

fengzhou-msft:
`entry_points/console_script` is reported with startup delay and has been discussed [before](https://github.com/Azure/azure-cli/pull/7469#issuecomment-427119382). We are considering #10516 to fix python3.

fengzhou-msft:
I'm closing this PR since we used shebang(#10516) to fix the issue. @gferon thanks for submitting this PR and we benefited from this solution. We welcome you for any further contribution to Azure CLI.

PR #10928, opened 2019-10-21
Read scale

bradrich-msft:
Adding read scale support to SQL Database CRUD operations.
---


src/azure-cli/HISTORY.rst
```diff
@@ -5,6 +5,10 @@ Release History
 
 **SQL**
 
+* Added "--read-scale" and "--read-replicas", parameters to support read scale management with SQL Database CRUD operations."
+
+**SQL**
```
jaredmoo: Merge issue - 2 **SQL** sections

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -224,6 +225,18 @@ def get_location_type_with_default_from_resource_group(cli_ctx):
...
 
+read_scale_param_type = CLIArgumentType(
+    options_list=['--read-scale'],
+    help='If enabled, connections that have application intent set to readonly in their connection string may be routed to a readonly secondary replica. '
+    'This property is only settable for Premium and Business Critical databases.',
+    arg_type=get_enum_type(DatabaseReadScale))
+
+read_replicas_param_type = CLIArgumentType(
+    options_list=['--read-replicas'],
+    type=SizeWithUnitConverter('B', result_type=int),
```
jaredmoo: This `type` shouldn't be here. It's for converting storage units, e.g. `"1TB"` -> `1099511627776`

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -224,6 +225,18 @@ def get_location_type_with_default_from_resource_group(cli_ctx):
...
+    help='The number of readonly replicas to provision for the database. '
+    'Only settable for Hyerpscale edition.')
```
jaredmoo: `Hyerpscale ` typo

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -524,6 +537,15 @@ def load_arguments(self, _):
 
         c.argument('max_size_bytes', help='The new maximum size of the database expressed in bytes.')
 
+        c.argument('read_scale',
+                   arg_type=read_scale_param_type,
+                   help='If enabled, connections that have application intent set to readonly in their connection string may be routed to a readonly secondary replica. '
```
jaredmoo: I think you don't need to override this in `sql db update` layer if it's already set in `sql db` layer

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -400,12 +414,11 @@ def load_arguments(self, _):
...
                    arg_type=get_enum_type(DatabaseLicenseType))
 
-        # Needs testing
-        c.ignore('read_scale')
-        # c.argument('read_scale',
-        #            arg_type=get_three_state_flag(DatabaseReadScale.enabled.value,
-        #                                         DatabaseReadScale.disabled.value,
-        #                                         return_label=True))
+        c.argument('read_scale',
+                   arg_type=get_enum_type(DatabaseReadScale))
```
jaredmoo: Change this to `arg_type=read_scale_param_type`, then you shouldn't need to redefine it later in `sql db update` context

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -323,7 +338,6 @@ def _configure_db_create_params(
     if create_mode != CreateMode.default or engine != Engine.db:
         arg_ctx.ignore('sample_name')
         arg_ctx.ignore('catalog_collation')
-        arg_ctx.ignore('read_scale')
```
jaredmoo: You should still ignore read_scale and read_replica count if engine == dw (section below)

bradrich-msft: Done

src/azure-cli/azure/cli/command_modules/sql/_params.py
bradrich-msft: Done

src/azure-cli/HISTORY.rst
bradrich-msft: Fixed.

src/azure-cli/azure/cli/command_modules/sql/_params.py
bradrich-msft: Fixed.

src/azure-cli/azure/cli/command_modules/sql/_params.py
bradrich-msft: Fixed.

src/azure-cli/azure/cli/command_modules/sql/_params.py
bradrich-msft: Fixed.

src/azure-cli/azure/cli/command_modules/sql/_params.py
bradrich-msft: Fixed.

src/azure-cli/azure/cli/command_modules/sql/_params.py
bradrich-msft: Fixed

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -319,11 +333,16 @@ def _configure_db_create_params(
     arg_ctx.argument('min_capacity',
                      arg_type=min_capacity_param_type)
 
+    arg_ctx.argument('read_scale',
```
jaredmoo: I thought we decided that these should go into `sql db` context so that they apply to both `sql db create` and `sql db update` ?

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -523,7 +545,7 @@ def load_arguments(self, _):
                    help='The name or resource id of the elastic pool to move the database into.')
 
         c.argument('max_size_bytes', help='The new maximum size of the database expressed in bytes.')
-
+
```
jaredmoo: Nit: whitespace

src/azure-cli/azure/cli/command_modules/sql/_params.py
bradrich-msft: Isn't that where they are?  I removed them from lower down sql db update.  They are working for both create and update.  Maybe the name _configure_db_create_params is misleading?  It seems to be intended and working for both create and update: "Configures params for db/dw create/update commands."

yonzhan:
@Juliehzl  please help with this PR.

bradrich-msft:
> @Juliehzl please help with this PR.

Please have a look.  Thank you.

Juliehzl:
OK. I will look it soon.

bradrich-msft:
> OK. I will look it soon.

Hi, any update on this?  Thanks.

src/azure-cli/HISTORY.rst
```diff
@@ -3,13 +3,16 @@
 Release History
 ===============
 
+**SQL**
+
+* Added "--read-scale" and "--read-replicas", parameters to support read scale management with SQL Database CRUD operations."
```
Juliehzl: Why do you end with "? typo?

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -224,6 +225,18 @@ def get_location_type_with_default_from_resource_group(cli_ctx):
     options_list=['--object-id', '-i'],
     help='The unique ID of the Azure AD administrator.')
 
+read_scale_param_type = CLIArgumentType(
+    options_list=['--read-scale'],
+    help='If enabled, connections that have application intent set to readonly '
+    'in their connection string may be routed to a readonly secondary replica. '
+    'This property is only settable for Premium and Business Critical databases.',
+    arg_type=get_enum_type(DatabaseReadScale))
```
Juliehzl: What are DatabaseReadScale values or type? Boolean? 
In addition, can user disable this property?

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -224,6 +225,18 @@ def get_location_type_with_default_from_resource_group(cli_ctx):
...
+
+read_replicas_param_type = CLIArgumentType(
```
Juliehzl: What is the type for this parameter? int or str?

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -224,6 +225,18 @@ def get_location_type_with_default_from_resource_group(cli_ctx):
...
+    options_list=['--read-replicas'],
+    help='The number of readonly replicas to provision for the database. '
```
Juliehzl: According to my understanding, if I don't set this parameter but enable read scale, readReplicaCount will be set to 1 and if I disable read scale, readReplicaCount will be set to 0, right? If yes, Can you show up such behavior for this parameter and default value?

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -224,6 +225,18 @@ def get_location_type_with_default_from_resource_group(cli_ctx):
     options_list=['--object-id', '-i'],
     help='The unique ID of the Azure AD administrator.')
 
+read_scale_param_type = CLIArgumentType(
```
Juliehzl: Would you mind mark your new parameters as Preview using https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#preview-commands-and-arguments?

src/azure-cli/azure/cli/command_modules/sql/tests/latest/test_sql_commands.py
```diff
@@ -459,6 +466,40 @@ def test_sql_db_vcore_mgmt(self, resource_group, resource_group_location, server
                      JMESPathCheck('sku.capacity', vcore_capacity_updated),
                      JMESPathCheck('sku.family', vcore_family_updated)])
 
+    @ResourceGroupPreparer(name_prefix='clitest', location='eastus2')
```
Juliehzl: Can you change your name_prefix to something related to your service? e.g. clitest_sql_

Juliehzl:
@bradrich-msft  It will be better if you can provide examples for your new parameters usage.

src/azure-cli/HISTORY.rst
bradrich-msft: Yes, a typo.  I am also updating this comment to specify sql db create and update commands.

src/azure-cli/azure/cli/command_modules/sql/_params.py
bradrich-msft: I took a look at this and am concerned this will introduce some confusion and inconsistency.  The REST API and SDK expects "Enabled" or "Disabled" and Powershell client also expects "Enabled" and "Disabled" parameter values.  If CLI only starts using "True" and "False" it will not be a consistent experience.  I agree this should've been done as boolean from the beginning, but this API has been around now for 3 years and I don't want to make CLI only to be different.  If we want to make this change, I think API and all clients should be changed together.

src/azure-cli/HISTORY.rst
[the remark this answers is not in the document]
bradrich-msft: Read scale allows scale out of read workloads to readonly secondary database replicas.  The parameter/option to manage this depends on the edition of the database.  For Premium and Business Critical edition databases. the option available is read-scale enabled or disabled only.  It is enabled by default.  For Hyperscale edition databases, read scale can be scaled out to multiple readonly replicas.  To managed the number of replicas, read-replicas can be 0 or more.  It is always enabled when 1+.

src/azure-cli/azure/cli/command_modules/sql/_params.py
bradrich-msft: The parameters are independent and depend on the database edition.  Only one of the parameters matters for each edition.  The other is ignored.  Please see my other reply on prior comment and also the help text for the parameters:

read_scale_param_type = CLIArgumentType(
    options_list=['--read-scale'],
    help='If enabled, connections that have application intent set to readonly '
    'in their connection string may be routed to a readonly secondary replica. '
    'This property is only settable for Premium and Business Critical databases.',
    arg_type=get_enum_type(DatabaseReadScale))

read_replicas_param_type = CLIArgumentType(
    options_list=['--read-replicas'],
    help='The number of readonly replicas to provision for the database. '
    'Only settable for Hyperscale edition.')

src/azure-cli/azure/cli/command_modules/sql/_params.py
bradrich-msft: This functionality is in full release, not preview.  Can you let me know why we should mark this as preview?

src/azure-cli/azure/cli/command_modules/sql/_params.py
bradrich-msft: In the SDK it is an int.  I can add that int type here, but it looks like we have a lot of other int parameters being treated as string.  Not sure what value it is to have only this one parameter defined as int.

src/azure-cli/HISTORY.rst
bradrich-msft: Fixed

src/azure-cli/azure/cli/command_modules/sql/_params.py
bradrich-msft: Fixed.

src/azure-cli/azure/cli/command_modules/sql/tests/latest/test_sql_commands.py
bradrich-msft: Fixed.

src/azure-cli/azure/cli/command_modules/sql/_params.py
Juliehzl: In case you want to change your commands. But if you are sure that you will not change your commands, I am fine to it.

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -224,6 +225,19 @@ def get_location_type_with_default_from_resource_group(cli_ctx):
...
+    options_list=['--read-scale'],
+    help='If enabled, connections that have application intent set to readonly '
+    'in their connection string may be routed to a readonly secondary replica. '
+    'This property is only settable for Premium and Business Critical databases.',
+    arg_type=get_three_state_flag('Enabled','Disabled', False, True))
```
Juliehzl: The right way to use get_three_state_flag is  arg_type=get_three_state_flag()

src/azure-cli/azure/cli/command_modules/sql/_params.py
Juliehzl: You can initialize your SDK model  to "Enabled" when the parameter is true and "Disabled" when false.  It will be user friendly.

Juliehzl:
Can you make CI pass and resolve left comments?

src/azure-cli/azure/cli/command_modules/sql/_params.py
bradrich-msft: We cannot change these commands as they are in full release.

src/azure-cli/azure/cli/command_modules/sql/_params.py
bradrich-msft: We are not changing the SDK as it has already been released for 3 years and we have not had any issue with usage.  It's well documented and easy to use "Enabled" and "Disabled" values.

src/azure-cli/azure/cli/command_modules/sql/_params.py
bradrich-msft: I'm not following this comment.  I looked up the definition of get_three_state_flag()--pasted below--and this is exactly how it is defined.  In fact, the example given is 'enabled' and 'disabled.'  I have tested it as well and it is working as expected.  Can you please clarify what your request is?

def get_three_state_flag(positive_label='true', negative_label='false', invert=False, return_label=False):
    """ Creates a flag-like argument that can also accept positive/negative values. This allows
    consistency between create commands that typically use flags and update commands that require
    positive/negative values without introducing breaking changes. Flag-like behavior always
    implies the affirmative unless invert=True then invert the logic.
    - positive_label: label for the positive value (ex: 'enabled')
    - negative_label: label for the negative value (ex: 'disabled')
    - invert: invert the boolean logic for the flag
    - return_label: if true, return the corresponding label. Otherwise, return a boolean value

qwordy: You can use Enabled and Disabled, but I suggest the default true and false, which is consistent with other boolean parameters.
Maybe you should use Enum type if you expect a value of ['Enabled', 'Disabled'].

bradrich-msft: It's not a boolean parameter, it's an enum all the way from API through SDK and all clients.  I originally had it as an enum and was asked to change to use 3 value type.  Are you suggesting it needs to go back to enum?

qwordy: Oh, I am a rookie. I haven't read the previous conversation. You can listen to experienced reviewers like juliehzl

bradrich-msft: I'm actually in agreement with you.  I don't know that it was necessary to change from enum to three_state_flag, but I wasn't against doing it.  @Juliehzl can you please reply or approve?  It's been a week since the last comment.

Juliehzl: Hi @bradrich-msft , I know it is not boolean. But the functionality of this parameter is the same as other boolean parameter and as what @qwordy said above, using true and false will be consistent with others. You can refer to the parameter `--enable-llarg-file-share` in  `az storage account create`  and the implementation [here](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/storage/operations/account.py#L78).
In addition, you should use get_enum_type(['Enabled', 'Disabled']) if you still want to use enum values.

bradrich-msft: I've changed to enum type for consistency across our API, SDK, and other clients.

PR #10936, opened 2019-10-22
acr: support to enable audits on create

yugangw-msft:

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acr/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/acr/_params.py
@@ -112,6 +112,8 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('registry_name', completer=None)
         c.argument('deployment_name', validator=None)
         c.argument('location', validator=get_default_location_from_resource_group)
+        c.argument('workspace', is_preview=True,
+                   help='Name or ID of the Log Analytics workspace to send registry diagnostic logs to. All events will be enabled. You can use "az monitor log-analytics workspace create" to create one. Extra billing may apply.')
 
     with self.argument_context('acr check-name') as c:
         c.argument('registry_name', completer=None)
--- a/src/azure-cli/azure/cli/command_modules/acr/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/acr/custom.py
@@ -17,6 +17,7 @@ from ._docker_utils import get_login_credentials
 from .network_rule import NETWORK_RULE_NOT_SUPPORTED
 
 logger = get_logger(__name__)
+DEF_DIAG_SETTINGS_NAME_TEMPLATE = '{}-diagnostic-settings'
 
 
 def acr_check_name(client, registry_name):
@@ -37,7 +38,8 @@ def acr_create(cmd,
                location=None,
                admin_enabled=False,
                default_action=None,
-               tags=None):
+               tags=None,
+               workspace=None):
     if default_action and sku not in get_premium_sku(cmd):
         raise CLIError(NETWORK_RULE_NOT_SUPPORTED)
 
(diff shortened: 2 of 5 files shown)


sajayantony:
This has a cost impact to customer so expecting this to be optional and enable this only after confirming with the PMs.

yugangw-msft:
Closing and opening it to re-trigger CI

yugangw-msft:
@qwordy, could you please take a look?

src/azure-cli/azure/cli/command_modules/acr/custom.py
```diff
@@ -48,7 +49,21 @@ def acr_create(cmd,
...
+        from azure.cli.core.commands import LongRunningOperation
+        from msrestazure.tools import is_valid_resource_id, resource_id
+        from azure.cli.core.commands.client_factory import get_subscription_id
+        acr = LongRunningOperation(cmd.cli_ctx)(lro_poller)
+        if not is_valid_resource_id(workspace):
+            workspace = resource_id(subscription=get_subscription_id(cmd.cli_ctx),
+                                    resource_group=resource_group_name,
+                                    namespace='microsoft.OperationalInsights',
+                                    type='workspaces',
+                                    name=workspace)
```
qwordy: I wrote similar code recently, which transforms name to ID.  Maybe we can abstract a layer in core to handle this so that we don't need write same transformation code for every parameter everywhere.

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -110,6 +110,8 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('registry_name', completer=None)
         c.argument('deployment_name', validator=None)
         c.argument('location', validator=get_default_location_from_resource_group)
+        c.argument('workspace', is_preview=True,
```
qwordy: Add a history entry?

src/azure-cli/azure/cli/command_modules/acr/custom.py
yugangw-msft: I gave it a thought, but felt it doesn't save lots of code except the subscription part. If that is what you meant, then yes I can add a core.util

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -110,6 +110,8 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
...
+                   help='Name or ID of the Log Analytics workspace to send registry diagnostic logs to. All events will be enabled. You can use "az monitor log-analytics workspace create" to create one. Extra billing might apply')
```
hkuadithya: nit: add a period at the end of the sentence. _Extra billing may apply._

yugangw-msft:
>  recieved an error message that "The subscription '' is not registered to use microsoft.insights."

@hkuadithya, you can run  `az provider register -n microsoft.insights --wait` and then try again.

src/azure-cli/azure/cli/command_modules/acr/custom.py
```diff
@@ -245,3 +260,45 @@ def _check_wincred(login_server):
...
+        {
+            "category": "AllMetrics",
+            "enabled": True,
+            "retentionPolicy": {
+                "days": 0,
+                "enabled": False
+            },
+            "timeGrain": None
+        }
+    ]
```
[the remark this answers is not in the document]
yugangw-msft: Actually, using json/dict is a more favored style from Python sdk's aspect because of the simplicity and good readabilities; however since the surrounding code uses model classes, be consistent with them make more sense. Let me update it.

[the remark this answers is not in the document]
qwordy: By design, Azure CLI should call Python SDK. But in fact, you can use JSON directly. Being consistent with the existing code in your module is always a good choice.

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -112,6 +112,8 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('registry_name', completer=None)
         c.argument('deployment_name', validator=None)
         c.argument('location', validator=get_default_location_from_resource_group)
```
msyihchen: One dumb question: Is this PR adding log analytics workspace creation upon registry creation?

hkuadithya: @msyihchen 
It looks like you have to create a Log analytics workspace first. Then hook up a container registry with the workspace while creating.

`az acr create -n <containerRegName> --workspace <workspaceName> --sku basic`

```
--workspace [Preview] : Name or ID of the Log Analytics workspace to send registry
                            diagnostic logs to. All events will be enabled. You can use "az monitor
                            log-analytics workspace create" to create one. Extra billing may apply.
```

PR #10937, opened 2019-10-22
[ACR] Token list table formatting, scope map update and help text fixes

hkuadithya:
1. `az acr token list -o table` fix to display `PASS1 EXPIRY` and `PASS2 EXPIRY` in order

**Before fix**

**After fix**

 
2. `scope-map update` command fix to prevent inconsistencies with `--add` and `--remove` actions

3. Adding period to few sentences. Few fixes to help texts.
---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acr/scope_map.py
+++ b/src/azure-cli/azure/cli/command_modules/acr/scope_map.py
@@ -3,6 +3,7 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 
+from azure.cli.core.util import CLIError
 from ._utils import get_resource_group_name_by_registry_name, validate_premium_registry, parse_actions_from_repositories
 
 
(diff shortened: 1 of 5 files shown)


src/azure-cli/azure/cli/command_modules/acr/scope_map.py
```diff
@@ -14,12 +15,11 @@ def acr_scope_map_create(cmd,
                          resource_group_name=None,
                          description=None):
 
+    resource_group_name = get_resource_group_name_by_registry_name(cmd.cli_ctx, registry_name, resource_group_name)
     validate_premium_registry(cmd, registry_name, resource_group_name)
```
hkuadithya: `validate_premium_registry` internally fetches the `resource_group_name` if it's none. 

Since the method `get_resource_group_name_by_registry_name` already fetches the resource group, this can be used with `validate_premium_registry` to avoid an extra call to retrieve the resource group.

PR #10955, opened 2019-10-23
Fix #10578: `az login` hangs

### LINKED ISSUE #10578: az login hangs after selecting account
## Describe the bug

**Command Name**
`az login`

**Errors:**
The command does launches a browser and allows me to select an account. After selecting an account, the browser just spins and the shell just waits. 

I've left it running for 5 minutes before closing the browser or hitting `Ctrl-C`.

## To Reproduce:

- `az login`

## Expected Behavior
I expect the command to launch a browser and allow me to select an account to log in with, then login me in. 

## Environment Summary
```
Windows-10-10.0.18362-SP0
Python 3.6.6
Shell: powershell.exe

azure-cli 2.0.73

```
## Additional Context
What's strange is that if I use the `--debug` parameter, it works with no problems.

jiasli:
Fix #10578: `az login` hangs when more than one instances of `az login` are launched at the same time on Windows or WSL

The root cause is that `HTTPServer` can't detect port is in-use on Windows. 

To repro, launch `az login` but do not continue in the browser. Then run `az login` from another terminal and continue with the login. One of the `az` instance will get stuck. 

This PR disables `HTTPServer.allow_reuse_address` so that the correct routine is resumed. 

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/util.py
+++ b/src/azure-cli-core/azure/cli/core/util.py
@@ -367,9 +367,9 @@ def sdk_no_wait(no_wait, func, *args, **kwargs):
 def open_page_in_browser(url):
     import subprocess
     import webbrowser
-    platform_name, release = _get_platform_info()
+    platform_name, _ = _get_platform_info()
 
-    if _is_wsl(platform_name, release):   # windows 10 linux subsystem
+    if is_wsl():   # windows 10 linux subsystem
         try:
             return subprocess.call(['cmd.exe', '/c', "start {}".format(url.replace('&', '^&'))])
         except OSError:  # WSL might be too old  # FileNotFoundError introduced in Python 3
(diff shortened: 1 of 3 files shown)


yugangw-msft:
Try to test CI, will close and re-open

qwordy:
I can't cancel my approval now, which can be done before the permission change. Please don't merge before a real approval.

qwordy:
I think we don't need to support multiple `az login` at same time since we don't support multiple accounts. We can let all other `az login` fail if there is an `az login` running. `az login` should not interleave.
And if we support multiple `az login`, add write locks to token files. I've seen broken accessTokens.json because of concurrent write. The content is interleaving.

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -395,16 +395,21 @@ def _get_platform_info():
     return platform_name.lower(), release.lower()
 
 
-def _is_wsl(platform_name, release):
+def is_wsl():
```
fengzhou-msft: I'm wondering why the original  _is_wsl() is ignoring the passed in perameter `platform_name, release` and calls _get_platform_info() to overwrite them?

jiasli: Seems to be some forgotten logic.

welersonlisboa:
https://github.com/Azure/azure-cli/issues/26180

jiasli:
@welersonlisboa, this issue is not relevant to https://github.com/Azure/azure-cli/issues/26180.

PR #10968, opened 2019-10-24
Fix #7048: `az keyvault recover/purge` not working

### LINKED ISSUE #7048: az keyvault recover does not work
**Describe the bug**
The `az keyvault recover` command does not work. Even though a keyvault is listed in `az keyvault list-deleted`, the `recover` command fails with `The Resource 'Microsoft.KeyVault/vaults/${keyvaultname}' not found within subscription.`

This appears to be a misapplication of the `validate_resource_group_name` function in `command_modules/keyvault/_validators.py`.

**To Reproduce**
- Create a keyvault, then turn on soft deletes.
- Delete the keyvault.
- Try to recover the keyvault using `az keyvault recover`,

**Expected behavior**
`az keyvault recover` should validate against deleted keyvault names, not current keyvault names.

**Environment summary**
Azure CLI on Ubuntu 16.04:
```
ii  azure-cli                                  2.0.44-1~xenial                              all          Azure CLI
```

**Additional context**
This is almost certainly not an environment-specific problem and should be easy to reproduce.

jiasli:
Fix #7048: `az keyvault recover/purge` not working

Original PR #7950

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/keyvault/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/keyvault/_params.py
@@ -24,7 +24,8 @@ from ._validators import (
     validate_principal, validate_resource_group_name,
     validate_x509_certificate_chain,
     secret_text_encoding_values, secret_binary_encoding_values, validate_subnet,
-    validate_vault_id, validate_sas_definition_id, validate_storage_account_id, validate_storage_disabled_attribute)
+    validate_vault_id, validate_sas_definition_id, validate_storage_account_id, validate_storage_disabled_attribute,
+    validate_deleted_vault_name)
 
 # CUSTOM CHOICE LISTS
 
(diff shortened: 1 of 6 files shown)


src/azure-cli/azure/cli/command_modules/keyvault/tests/latest/test_keyvault_commands.py
```diff
@@ -755,7 +755,16 @@ def test_keyvault_softdelete(self, resource_group):
...
+        # recover and purge
+        self.cmd('keyvault delete -n {kv}')
+        self.cmd('keyvault recover -n {kv}')
+        self.cmd('keyvault delete -n {kv}')
+        self.cmd('keyvault purge -n {kv}')
+
+        # recover and purge with location
+        _create_keyvault(self, self.kwargs, additional_args=' --enable-soft-delete true').get_output_in_json()
+        self.cmd('keyvault delete -n {kv}')
+        self.cmd('keyvault recover -n {kv} -l {loc}')
```
yugangw-msft: check return values

src/azure-cli/azure/cli/command_modules/keyvault/_params.py
```diff
@@ -75,6 +76,18 @@ def load_arguments(self, _):
         c.argument('no_self_perms', arg_type=get_three_state_flag(), help="Don't add permissions for the current user/service principal in the new vault.")
         c.argument('location', validator=get_default_location_from_resource_group)
 
+    with self.argument_context('keyvault recover') as c:
+        c.argument('vault_name', help='Name of the deleted vault', required=True, completer=None,
+                   validator=validate_deleted_vault_name)
+        c.argument('resource_group_name', resource_group_name_type, id_part=None, required=False,
```
yugangw-msft: this is redundant, but maybe for the case that current principal doesn't have the permission to list soft deleted vaults?
either way, would be great to cross check, learn and make the CLI simpler and better

jiasli: We should keep `location` just in case "the current principal doesn't have the permission to list soft deleted vaults". For `resource_group_name`, we can keep it to align with `keyvault show`.

PR #10976, opened 2019-10-24
fix create and update bugs

gtracer:
Fixes two bugs:
1. On `az aks update  --api-server-authorized-ip-ranges ""` which is used to disable the resulting input to RP should be [] not ['']

2. On `az aks create` there is a bug on vm sku type naming that prevents clusters from being created 
Both bugs were introduced in a previous (non-released) PR: https://github.com/Azure/azure-cli/pull/10865
---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acs/_helpers.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/_helpers.py
@@ -8,15 +8,17 @@ from distutils.version import StrictVersion  # pylint: disable=no-name-in-module
 from azure.mgmt.containerservice.v2019_08_01.models import ManagedClusterAPIServerAccessProfile
 
 
-def _populate_api_server_access_profile(api_server_authorized_ip_ranges):
+def _populate_api_server_access_profile(api_server_authorized_ip_ranges, instance=None):
+    if instance is None or instance.api_server_access_profile:
+        profile = ManagedClusterAPIServerAccessProfile()
+
     if api_server_authorized_ip_ranges == "":
         authorized_ip_ranges = []
     else:
         authorized_ip_ranges = [ip.strip() for ip in api_server_authorized_ip_ranges.split(",")]
 
-    return ManagedClusterAPIServerAccessProfile(
-        authorized_ip_ranges=authorized_ip_ranges
-    )
+    profile.authorized_ip_ranges = authorized_ip_ranges
+    return profile
 
 
 def _set_load_balancer_sku(load_balancer_sku, kubernetes_version):
--- a/src/azure-cli/azure/cli/command_modules/acs/_helpers.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/_helpers.py
@@ -9,8 +9,13 @@ from azure.mgmt.containerservice.v2019_08_01.models import ManagedClusterAPIServ
 
 
 def _populate_api_server_access_profile(api_server_authorized_ip_ranges):
+    if api_server_authorized_ip_ranges == "":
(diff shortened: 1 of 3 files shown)


src/azure-cli/azure/cli/command_modules/acs/_helpers.py
```diff
@@ -9,8 +9,13 @@
 
 
 def _populate_api_server_access_profile(api_server_authorized_ip_ranges):
+    if api_server_authorized_ip_ranges == "":
+        authorized_ip_ranges = []
+    else:
+        authorized_ip_ranges = [ip.strip() for ip in api_server_authorized_ip_ranges.split(",")]
+
     return ManagedClusterAPIServerAccessProfile(
```
ruchikaguptaa: This would overwrite other properties in ManagedClusterAPIServerAccessProfile for eg. enablePrivateCluster? Won't this be a problem?

src/azure-cli/azure/cli/command_modules/acs/_helpers.py
```diff
@@ -31,6 +36,14 @@ def _set_vm_set_type(vm_set_type, kubernetes_version):
             not specified and kubernetes version(%s) less than 1.12.9 only supports \
             availabilityset\n' % (kubernetes_version))
             vm_set_type = "AvailabilitySet"
-        else:
-            vm_set_type = "VirtualMachineScaleSets"
-    return vm_set_type.capitalize()
+
+    if not vm_set_type:
```
ruchikaguptaa: Why add this again? Its already checked on L28. 
You can just keep L34, L35 as is and update L36 with the logic below from L43-L49

ruchikaguptaa: Something like 
` else:
            vm_set_type = "VirtualMachineScaleSets"

 # normalize as server validation is case-sensitive
    if vm_set_type.lower() == "AvailabilitySet".lower():
        vm_set_type = "AvailabilitySet"

    if vm_set_type.lower() == "VirtualMachineScaleSets".lower():
        vm_set_type = "VirtualMachineScaleSets"
    return vm_set_type`

src/azure-cli/azure/cli/command_modules/acs/_helpers.py
gtracer: Yup, you are right - will update

src/azure-cli/azure/cli/command_modules/acs/_helpers.py
gtracer: I just reverted the logic to the original logic that Tony had for SLB. I shouldn't have refactored it in my previous PR (unrelated to IPWL). 

Going to leave it as Tony's logic - we can consider updating it in a separate PR if you want

src/azure-cli/azure/cli/command_modules/acs/_helpers.py
gtracer: Updated

robbiezhang:
LGTM

src/azure-cli/azure/cli/command_modules/acs/_helpers.py
ruchikaguptaa: I see. Sounds good.

marwanad:
@yonzhan @qwordy i think this should also be for the S160 milestone.

yonzhan:
@qwordy please merge this PR before Ignite release.

qwordy:
> 
> 
> @qwordy please merge this PR before Ignite release.

OK

PR #10985, opened 2019-10-25
default task to v1.1.0 on run cmd

shahzzzam:
This will enable task to be run on v1.1.0 (which makes use of Aliases)

So instead of, 
```
az acr run -r MyRegistry --cmd "{{.Run.Registry}}/MyImage" /dev/null
```
You can do
```
az acr run -r MyRegistry --cmd "\$Registry/MyImage" /dev/null
```

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acr/_utils.py
+++ b/src/azure-cli/azure/cli/command_modules/acr/_utils.py
@@ -245,13 +245,13 @@ def get_validate_platform(cmd, platform):
 
 def get_yaml_template(cmd_value, timeout, file):
     """Generates yaml template
-    :param str cmd_value: The command to execute in each step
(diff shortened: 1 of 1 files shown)


src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -468,7 +468,7 @@
 examples:
   - name: Queue a run to execute a container command.
     text: >
-        az acr run -r MyRegistry --cmd MyImage /dev/null
+        az acr run -r MyRegistry --cmd '\\$Registry/MyImage' /dev/null
```
northtyphoon: what does `\\` do, skip `$`? It looks very odd. Then I start to think should we recommend users to use the pattern in the examples?

PR #10987, opened 2019-10-25
[Core] Expose example_provider hook to include AI examples

mirdaki:
This updates the Cloud Shell `-h` pages to use the generated examples from the Aladdin service if available (it falls back if none were found or if it could not connect). Because Cloud Shell is hosted in Azure, the latency for connecting to the Aladdin service should be reduced while displaying up-to-date examples to users.

I also updated the unit tests in the find module to use a mocked connection to the service and added a couple more tests to flesh it out.

This update did require dependency injection to let the find module interact with the core service. I tried to cause the fewest changes as possible to the project, but I recognize their might be a better way to do it. I'll point it out those changes below. 

This PR does not update HISTORY.rst, because it only affects help content.

---


src/azure-cli/azure/cli/utils.py
```diff
@@ -0,0 +1,26 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+def get_default_cli():
```
mirdaki: I was unable to find a better way to connect the find module to the core help page that didn't involve making more changes or leaving a circular dependency. This is mostly copied from the `azure.cli.core.__init__` file, with the goal of replacing the `get_generated_examples` in the help file with the one from find. 

I realize this is not a trivial change to the flow of the app, so please leave suggestions if you have them for how this could be better accomplished.

src/azure-cli-core/azure/cli/core/_help.py
```diff
@@ -77,14 +78,31 @@ def _print_examples(help_file):
...
-            if e.long_summary:
-                _print_indent(u'{0}'.format(e.long_summary), indent)
-            _print_indent(u'{0}'.format(e.command), indent)
-            print('')
+        use_default_examples = True
+        # TODO: Once Python 3.8 support is added, switch to an Assignment 
+        # Expressions to check if the the generated example list is not empty.
+        # That way it can fall back to the default one if there was a server
+        # error more elegantly.
+        if in_cloud_console():
```
mirdaki: I was unable to find information on how to debug/test from Cloud Shell. If there are resources, please let me know so I can verify this method works.

yonzhan:
@jiasli please help with this PR.

jiasli:
I think it'll be good to **append** the Aladdin generated examples instead of override the original ones. 

When I run `az vm create -h` with env var `ACC_CLOUD=1` to mock Cloud Shell, the output only has 3 examples, which are not as helpful as the original ones.

```
Examples
    Create an Azure Virtual Machine. (autogenerated)
        az vm create --admin-username deploy --generate-ssh-keys  --image Centos --name MyVm
        --resource-group MyResourceGroup

    Create a default Ubuntu VM with automatic SSH authentication.
        az vm create -n MyVm -g MyResourceGroup --image UbuntuLTS

    Create a default RedHat VM with automatic SSH authentication using an image URN.
        az vm create -n MyVm -g MyResourceGroup --image RedHat:RHEL:7-RAW:7.4.2018010506
```

src/azure-cli/azure/cli/utils.py
jiasli: This file isn't necessary and introduces code duplication. In `src/azure-cli/azure/cli/__main__.py`, we can call `get_default_cli` from `azure.cli.core` as before and do dependency injection with 

``` py
from azure.cli.core import get_default_cli
from azure.cli.core._help import AzCliHelp
from azure.cli.command_modules.find.custom import get_generated_examples

az_cli = get_default_cli()
AzCliHelp.get_generated_examples = get_generated_examples
```
We may also rename the attribute to `example_provider` to make it easier to understand.

jiasli: This file isn't necessary and introduces code duplication. In `src/azure-cli/azure/cli/__main__.py`, we can call `get_default_cli` from `azure.cli.core` as before and do dependency injection with 

``` py
from azure.cli.core import get_default_cli
from azure.cli.core._help import AzCliHelp
from azure.cli.command_modules.find.custom import get_generated_examples

az_cli = get_default_cli()
AzCliHelp.get_generated_examples = get_generated_examples
```
We may also rename the attribute to `example_provider` to make it easier to understand.

src/azure-cli-core/azure/cli/core/_help.py
mirdaki: Gotcha, set `ACC_CLOUD=1`.

src/azure-cli/azure/cli/utils.py
mirdaki: Good suggestions! I've improved the dependency injection and updated the name. Thank you for the help.

mirdaki:
Sorry for the long delay. I've made the suggested technical improvements. I've been told (at least for now) to continue replacing all the help examples when possible in Cloud Shell. Let me know what you think!

qwordy:
The examples written by human are better than generated ones in some perspectives. Maybe it is a little radical to remove all the high-quality examples.

yonzhan:
@mirdaki  I got request from Aladdin team that it will generate a separate single file which contains all commands from different module for us to refer to. Please verify if it is the case from your side.

mirdaki:
@yonzhan Yes, we are in the process of changing our example PRs to create one file with all the command examples in it as opposed to multiple PRs each with only some of the examples.

mirdaki:
The team is switching strategies to adding "hooks" into the core CLI, that will then be leveraged by an extension. That extension will replace the hooks and provide the rest of the logic. 

This strategy will impact a smaller audience, allowing for more experimentation.

The new goal of this PR is to add any hooks we think will be used in the near future and make a few improvements to the `az find` module (since that was already done as part of this process). The extension is being developed [here](https://github.com/mirdaki/azure-cli-extensions/tree/aladdin).

mirdaki:
@jiasli could you please re-review the PR? I don't have any more planned additions to it.

It adds a hook in the help file that the new [`ai-examples` extension](https://github.com/mirdaki/azure-cli-extensions/tree/ai-examples) will inject code into. This way the core CLI doesn't have any functionality changes, but still lets us experiment with the new functionality. It also updates the `az find` code to be a little cleaner and more tested.

src/azure-cli-core/azure/cli/core/_help.py
```diff
@@ -188,6 +188,11 @@ def update_loaders_with_help_file_contents(self, nouns):
                 file_contents[name] = self._name_to_content[name]
             self.versioned_loaders[ldr_cls_name].update_file_contents(file_contents)
 
+    # This method is meant to be a hook that can be overridden by an extension or module.
+    @staticmethod
+    def example_provider(help_file):
+        return help_file.examples
```
jiasli: Does your example provider rely on `help_file`? If not, I would prefer removing `help_file` from the param list to reduce coupling.

```py
    def _print_examples(help_file):
        ...
        # help_file.example can be overridden with `AzCliHelp.example_provider`
        generated_example = AzCliHelp.example_provider()
        if generated_example:
            help_file.example = generated_example
        for e in help_file.examples:
            indent = 1
```

```py
    def example_provider():
        return []
```

If so, the code is good to me. I can merge it after getting your confirmation.

jiasli:
Another problem is will the result of `az find` have conflicts with Aladdin generated examples which is added to CLI's `_help.py` files? Looping in @haroldrandom who handles Aladdin example integration.

src/azure-cli-core/azure/cli/core/_help.py
mirdaki: The `ai-examples` extension needs both the command that's being looked up and the examples that command has so it can prevent any duplicate examples from being added by the Aladdin service. `help_file` has both of those values, so I'd like to keep the current usage of it. 

It also gives the extension the possibility of changing the order of all the examples. For instance, we could do something like organize them by usage (to be clear: it's not something we're planning right now, but could be useful in the future).

mirdaki:
The extension looks at the built in examples and removes any duplicates it finds, whether it be the handcrafted or Aladdin generated ones, from what the Aladdin service provides, so there shouldn't be any conflicts.

The advantage this extension provides over the Aladdin `_help` integration is that new or edited examples can be added or experimented with more frequently than what can be shipped inthe standard Azure CLI.

haroldrandom:
@mirdaki Could you share some knowledge on how to remove duplicated exmaples? We may need it.

mirdaki:
@haroldrandom Sure! I do a simple string comparison with cleaned versions of the summary and snippet between the new and existing examples. If either summary or snippet match, I consider it a duplicate.

It's fairly rudimentary, but because of how example summaries are currently applied (which from what I was told, tries to use existing summaries when the same parameters are used) it works well. In my (somewhat brief) testing I didn't find any issues.

Code is pulled from the [`ai examples` extension PR](https://github.com/Azure/azure-cli-extensions/pull/1196/files#diff-f0770e2953bdc6e982b890fcbad3ffee):

```python
def concat_unique_examples(first_list, second_list):
    for first_item in first_list:
        for second_item in second_list:
            if are_examples_equal(first_item, second_item):
                second_list.remove(second_item)
    return first_list + second_list


def are_examples_equal(first, second):
    return clean_string(first.short_summary) == clean_string(second.short_summary) \
        or clean_string(first.command) == clean_string(second.command)


def clean_string(text):
    return text.strip()
```

haroldrandom:
> cleaned versions of the summary and snippet

What is "cleaned versions of the summary and snippet" you mentioned?

mirdaki:
Each individual example has a snippet of code and short summary associated with it. I clean those two strings of any white space (tabs, spaces, `/n`, or `/r`) characters on either side of the actual text. 

Their seems to be an inconsistency in how those characters are applied to the example snippets and summaries. so doing this cleaning makes them more consistent with each other.

PR #10999, opened 2019-10-28
Fix ACR linuxFxVersion for function apps

ankitkumarr:
Quick fix for Ignite release

@ahmedelnably @Hazhzeng

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -2292,7 +2292,8 @@ def create_function(cmd, resource_group_name, name, storage_account, plan=None,
                                                               value='true'))
                 if runtime not in RUNTIME_TO_IMAGE_FUNCTIONAPP.keys():
                     raise CLIError("An appropriate linux image for runtime:'{}' was not found".format(runtime))
-        site_config.linux_fx_version = _get_linux_fx_functionapp(is_consumption, runtime, runtime_version)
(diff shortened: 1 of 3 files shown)


src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2292,7 +2292,8 @@ def create_function(cmd, resource_group_name, name, storage_account, plan=None,
                                                               value='true'))
                 if runtime not in RUNTIME_TO_IMAGE_FUNCTIONAPP.keys():
                     raise CLIError("An appropriate linux image for runtime:'{}' was not found".format(runtime))
-        site_config.linux_fx_version = _get_linux_fx_functionapp(is_consumption, runtime, runtime_version)
+        if deployment_container_image_name is not None:
```
Hazhzeng: I think this line should be `if site_config.linux_fx_version is None`.
Because, hey think about this, when the **deployment_container_image_name** is provided and **Linux Dedicated** is chosen.

The first **site_config.linux_fx_version** assignment, `_format_fx_version(deployment_container_image_name)` in line 2289 will still be overridden by `_get_linux_fx_functionapp(is_consumption, runtime, runtime_version)` in line 2295.

ankitkumarr: This should be `is None`. My bad.

Hazhzeng:
This looks good to me

Issue #9853, opened 2019-07-03
scripts/ci/dependency_check.bat doesn't fail CI, even when it detects errors.


marstr:
If you check #9750, you'll see that CI fails dependency checks for Darwin and Linux, however the Windows check passes. Digging in, it's clear that the Windows script even detects the error here: [Build 60891, Verify src/azure-cli/requirements.*.Windows.txt, Line 342](https://dev.azure.com/azure-sdk/public/_build/results?buildId=60891&view=logs&j=796343fd-f04d-59ce-7e73-c4eab21e4249&t=3e5b4737-4f47-5027-dc25-be482d8c1eaf&l=342). However, the appropriate exit code isn't getting set.

yonzhan:
@haroldrandom add to next Sprint.

haroldrandom:
Fix at #11094. Feel free to reopen if there are futher problems.

PR #11094, opened 2019-11-04
Bug Fix: wrong invokation to call ./scripts/install_all.bat which let dependency check always pass on Windows platform

haroldrandom:
* Bug fix: issue #9853 to make dependency check valid on Windows platform

---

### DIFF
--- a/scripts/ci/dependency_check.bat
+++ b/scripts/ci/dependency_check.bat
@@ -9,8 +9,7 @@ if %errorlevel% neq 0 exit /b %errorlevel%
 del baseline_deps.txt
 
 Rem Install everything from our repository first.
-.\scripts\install_full.bat
-if %errorlevel% neq 0 exit /b %errorlevel%
+call .\scripts\install_full.bat
 
 pip check
 if %errorlevel% neq 0 exit /b %errorlevel%
--- a/scripts/ci/dependency_check.bat
+++ b/scripts/ci/dependency_check.bat
@@ -10,6 +10,7 @@ del baseline_deps.txt
 
 Rem Install everything from our repository first.
 call .\scripts\install_full.bat
+if %errorlevel% neq 0 exit /b %errorlevel%
 
 pip check
 if %errorlevel% neq 0 exit /b %errorlevel%
--- a/scripts/install_full.bat
+++ b/scripts/install_full.bat
@@ -1,7 +1,7 @@
 Rem This script is the windows counter-part to ./install_full.sh
 
 pushd %~dp0..
- 
+
 set cmd="dir .\src\ /b | findstr /v azure-cli-testsdk"
 for /f "tokens=*" %%e in (' %cmd% ') do (
     pip install --no-deps .\src\%%e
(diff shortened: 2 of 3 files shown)


haroldrandom:
FYI.

As you can see, in our previous dependency_check job **Verify src/azure-cli/requirements.*.Windows.txt**,

build [20190702.02](https://dev.azure.com/azure-sdk/public/_build/results?buildId=60891&view=logs&j=796343fd-f04d-59ce-7e73-c4eab21e4249&t=3e5b4737-4f47-5027-dc25-be482d8c1eaf&l=342) exited and return code 0 even though script detectes error. Which is not reasonable. 

build [20191031.26](https://dev.azure.com/azure-sdk/public/_build/results?buildId=176944&view=logs&j=796343fd-f04d-59ce-7e73-c4eab21e4249&t=3e5b4737-4f47-5027-dc25-be482d8c1eaf&l=390) exited and return code 0 even though script interupted.

So, whatever situation, dependency check of CI will always green even though it detects error on Windows platform.

My changes of [my first commit](https://github.com/Azure/azure-cli/pull/11094/commits/5faa98f1dd3e217cbe855792963672d35d50bc4f) can reproduce those build's situation.

scripts/install_full.bat
```diff
@@ -11,5 +11,4 @@ set cmd="python .\scripts\get-python-version.py"
 for /f "tokens=*" %%v in (' %cmd% ') do set SHORT_PY_VER=%%v
 pip install -r .\src\azure-cli\requirements.%SHORT_PY_VER%.Windows.txt
 
-echo %OUTPUT%
```
fengzhou-msft: what is the value of %OUTPUT%? Why remove it?

haroldrandom: %OUTPUT% is not a variable in those *.bat scripts context,  Azure DevOps and Windows Batch sctipt. The functionality is going to ECHO a non-existing variable. Can tell from [line 360 of build 20190702.7](https://dev.azure.com/azure-sdk/public/_build/results?buildId=60891&view=logs&j=796343fd-f04d-59ce-7e73-c4eab21e4249&t=3e5b4737-4f47-5027-dc25-be482d8c1eaf&l=360) . 

So, I think it's trivial.

scripts/ci/dependency_check.bat
```diff
@@ -9,8 +9,7 @@ if %errorlevel% neq 0 exit /b %errorlevel%
 del baseline_deps.txt
 
 Rem Install everything from our repository first.
-.\scripts\install_full.bat
-if %errorlevel% neq 0 exit /b %errorlevel%
+call .\scripts\install_full.bat
```
fengzhou-msft: what happens with `call` if install_full.bat returns non-zero exit code?

haroldrandom: the script won't fail as all it is going to do, is `pip install`

fengzhou-msft: should we still keep `if %errorlevel% neq 0 exit /b %errorlevel%`, just in case?

haroldrandom: I could do so, but it's not necessary I think

haroldrandom: Adding it back for further modification of ./scripts/install_all.bat

MyronFanQiu:
talked with @haroldrandom offline. LGTM.

PR #11125, opened 2019-11-06
Lock six version to ~=1.12 temporarily to let Travis CI pass

haroldrandom:
**Temporarily** lock six version to ~=1.12 until next Sprint due to [upgrade of package six from ~1.12.0 to 1.13.0](https://pypi.org/project/six/#history) in Nov 6, 2019 breaks dependency, which failed Travis CI

Related PR:  #11051

---

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ addons:
       - libssl-dev
       - libffi-dev
       - python-dev
-install: pip install --upgrade pip tox six
+install: pip install --upgrade pip tox six~=1.12.0
 cache: pip
 jobs:
   include:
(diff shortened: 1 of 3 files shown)


jiasli:
Astroid issue: https://github.com/PyCQA/astroid/issues/718

.travis.yml
```diff
@@ -11,7 +11,7 @@ addons:
       - libssl-dev
       - libffi-dev
       - python-dev
-install: pip install --upgrade pip tox six
+install: pip install --upgrade pip tox six~=1.12.0
```
fengzhou-msft: I think if you lock six version also in azure-cli-core/setup.py besides azure-cli/setup.py. You may not need to put six here anymore.

haroldrandom: would it be better to keep a note and sync with astroid?

haroldrandom: resolved offline

fengzhou-msft: ok, you can keep it here since six is also a dependency by other packages.

PR #11129, opened 2019-11-06
[keyvault] fix: --include-pending requires true or false

### LINKED ISSUE #11121: --include-pending requires a true or false parameter
**Describe the bug**
When using `az keyvault certificate list` context, passing `--include-pending` requires a value of `true` or `false`, unlike other boolean switches like `az keyvault create --enable-soft-delete`. Not passing a value results in an error that a value is required.

**To Reproduce**
`az keyvault certificate list --vault-name <vault name> --include-pending`

**Expected behavior**
I should see all certificates - pending or otherwise - since the default for a boolean switch is `true`.

**Environment summary**
Installed via pip. Happens on both azure-cli 2.0.55 and 2.0.74 on both Windows and Ubuntu (wsl, though using the linux distribution of azure-cli).

**Additional context**
Fixing this shouldn't be a breaking change, since you can still optionally pass `true` or `false` to a boolean switch.

jiasli:
Fix #11121: When using `az keyvault certificate list`, passing `--include-pending` now doesn't require a value of `true` or `false`, similar to `--enable-soft-delete`.

To test:
```
az keyvault certificate list --vault-name {} --include-pending
```
---


jiasli:
Good idea. Tests added.

PR #11136, opened 2019-11-07
Acr task create/update: branch parameter deprecated

rosanch:
---


northtyphoon:
@jaysterp @shahzzzam

src/azure-cli/azure/cli/command_modules/acr/task.py
```diff
@@ -982,3 +989,11 @@ def _get_trigger_event_list_patch(cmd,
             if SourceTriggerEvent.pullrequest.value in source_trigger_events:
                 source_trigger_events.remove(SourceTriggerEvent.pullrequest.value)
     return source_trigger_events
+
+
+def _get_branch_name(context_path):
+    # Context Path formats https://docs.docker.com/engine/reference/commandline/build/#git-repositories
+    branch = re.search(r'(?<=.git#)([^:/\n\s]*)', context_path)
```
northtyphoon: DevOps doesn't have `.git` in the path.  You might just look for the first occurrence of `#`.

src/azure-cli/azure/cli/command_modules/acr/task.py
```diff
@@ -383,6 +388,8 @@ def acr_task_update(cmd,  # pylint: disable=too-many-locals
         source_triggers = task.trigger.source_triggers
         base_image_trigger = task.trigger.base_image_trigger
         if (commit_trigger_enabled or pull_request_trigger_enabled) or source_triggers:
+            if not branch:
+                branch = _get_branch_name(context_path)
```
northtyphoon: >_get_branch_name [](start = 25, length = 16)

I notice it's for update scenario, but since the change is migrating to context_path for branch, if `context_path` is `[https://github.com/user/repo.git`](https://github.com/user/repo.git`), should the code set branch to `master`?

src/azure-cli/azure/cli/command_modules/acr/task.py
```diff
@@ -982,3 +989,11 @@ def _get_trigger_event_list_patch(cmd,
             if SourceTriggerEvent.pullrequest.value in source_trigger_events:
                 source_trigger_events.remove(SourceTriggerEvent.pullrequest.value)
     return source_trigger_events
+
+
+def _get_branch_name(context_path):
```
shahzzzam: We should default to `master` branch here if not found.

rosanch: I'm not doing this because in update the current logic is to send none by default. Should we switch to master when the branch is not specified in a context when updating a task?

src/azure-cli/azure/cli/command_modules/acr/task.py
rosanch: I do have the same question. Right now the default for update is none unless specified so I kept the same logic while deprecating the change, so we don't introduce a breaking change to the user.

src/azure-cli/azure/cli/command_modules/acr/task.py
jaysterp: No, you can leave it as None.

haroldrandom:
Please merge/rebase the latest branch to let CI gree. Some changes have be made.

jaysterp:
@rosanch please archive this change in `History.rst`.

src/azure-cli/HISTORY.rst
```diff
@@ -11,6 +11,10 @@ Release History
 
 * Fix #10996: Polish error for `--force-change-password-next-login` in `az ad user update` when `--password` is not specified
 
+**ACR**
```
jaysterp: @rosanch the release history is alphabetized by team. Please bring ACR to the top.

haroldrandom:
Small section of conflicts need to be resovled first.

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -187,7 +187,7 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('pull_request_trigger_enabled', help="Indicates whether the source control pull request trigger is enabled. The trigger is disabled by default.", arg_type=get_three_state_flag())
         c.argument('schedule', help="Schedule for a timer trigger represented as a cron expression. An optional trigger name can be specified using `--schedule name:schedule` format. Multiples supported by passing --schedule multiple times.", action='append')
         c.argument('git_access_token', help="The access token used to access the source control provider.")
-        c.argument('branch', help="The source control branch name.")
+        c.argument('branch', deprecate_info=c.deprecate(hide=True), help="The source control branch name. Please specify your source branch in the context parameter e.g. https://github.com/Azure-Samples/acr-build-helloworld-node.git#mybranch")
```
MyronFanQiu: why do you set `hide=True`? It will completely hide this argument from users.

MyronFanQiu: User cannot see the help information anymore.

rosanch: because we're going to deprecate this parameter. The user should pass the branch and directory in the context.

MyronFanQiu: but you changed the help information. User won't see the new help information if you set `hide=True`

qwordy:
What will happen if a user specify branch in both --context and --branch?

rosanch:
@qwordy If the user specifies a branch in both context and branch it will still work as before. Only when the branch is specified in the context and there is no input in the branch parameter this change will take effect by copying the value to the branch parameter.

src/azure-cli/azure/cli/command_modules/acr/_params.py
rosanch: We decided to remove it completely after I already modified the help info. I saw no need to revert the change, but I can revert if it's an issue.

src/azure-cli/azure/cli/command_modules/acr/task.py
```diff
@@ -982,3 +986,13 @@ def _get_trigger_event_list_patch(cmd,
...
```
adewaleo: This might be overkill, but do we want to add a unit test or tests for this method? Especially, given it relies on regexes? It might work fine now, but what if someone modifies its behavior and breaks it in the future?

rosanch: we're going to update the backend side to handle this case and remove this code completely. So I don't think is worth the trouble for now.

PR #11163, opened 2019-11-08
Updating help for task create/update commands

rosanch:
---


haroldrandom:
conflict can not let CI continue. Please solve it first.

northtyphoon:
@rosanch can you take a look at the conflict?

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -578,69 +578,36 @@
   - name: Create a task without the source location.
     text: >
         az acr task create -n hello-world -r MyRegistry --cmd '$Registry/myimage' -c /dev/null
-  - name: Create a task without the source location and with a timer trigger that runs the image `MyImage` at the top of every hour using the default trigger name.
-    text: >
-        az acr task create -n hello-world -r MyRegistry --cmd '$Registry/myimage' -c /dev/null --schedule "0 */1 * * *"
-  - name: Create a task with the definition from the standard input. Either 'Ctrl + Z'(Windows) or 'Ctrl + D'(Linux) terminates the input stream.
-    text: >
-        az acr task create -n hello-world -r MyRegistry -f - -c /dev/null
-  - name: Create a task with the definition from the pipe.
```
northtyphoon: keep the pipe example?

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -578,69 +578,36 @@
...
-        az acr task create -t hello-world:{{.Run.ID}} -n hello-world -r MyRegistry \\
-            -c https://github.com/Azure-Samples/acr-tasks.git#:multipleRegistries -f testtask.yaml \\
-            --commit-trigger-enabled false --schedule "0 12 * * Mon-Fri"\\
-            --assign-identity
-  - name: Create a Linux task from a public GitHub repository which builds the hello-world image with a git commit trigger and a timer trigger that runs that task at noon on Mondays through Fridays with the trigger name provided.
-    text: |
-        az acr task create -t hello-world:{{.Run.ID}} -n hello-world -r MyRegistry \\
-            -c https://github.com/Azure-Samples/acr-build-helloworld-node.git -f Dockerfile \\
+            -c https://github.com/Azure-Samples/acr-tasks.git#:multipleRegistries \\
+            -f testtask.yaml --assign-identity [system] "<myIdentityResourceId>" \\
```
northtyphoon: can you keep the example of resource id?

northtyphoon:
@jaysterp  @shahzzzam

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -578,69 +578,36 @@
...
-            -c https://github.com/Azure-Samples/acr-build-helloworld-node.git -f Dockerfile --commit-trigger-enabled false
+        az acr task create -n hello-world -r MyRegistry -f - -c /dev/null --schedule "0 */1 * * *"
   - name: Create a Linux task from a public GitHub repository which builds the hello-world image without triggers and uses a build argument.
     text: |
         az acr task create -t hello-world:{{.Run.ID}} -n hello-world -r MyRegistry \\
-            -c https://github.com/Azure/acr-builder.git -f Dockerfile --commit-trigger-enabled false \\
-            --arg DOCKER_CLI_BASE_IMAGE=docker:18.03.0-ce-git
-  - name: Create a Linux task using a private GitHub repository which builds the hello-world image without triggers on Arm architecture (V7 variant)
+            -c https://github.com/Azure/acr-builder.git -f Dockerfile \\
+            --base-image-trigger-enabled false --pull-request-trigger-enabled false \\
```
jaysterp: @rosanch the pr trigger is disabled by default. You can remove that parameter.

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -578,69 +578,36 @@
...
         az acr task create -t acb:{{.Run.ID}} -n acb-win -r MyRegistry \\
             -c https://github.com/Azure/acr-builder.git -f Windows.Dockerfile \\
-            --commit-trigger-enabled false --platform Windows/amd64
-  - name: Create a Linux multi-step task from a public GitHub repository with a git commit trigger and with the system-assigned managed identity
-    text: |
-        az acr task create -t hello-world:{{.Run.ID}} -n hello-world -r MyRegistry \\
-            -c https://github.com/Azure-Samples/acr-tasks.git#:multipleRegistries -f testtask.yaml \\
-            --assign-identity
-  - name: Create a Linux multi-step task from a public GitHub repository with a git commit trigger and with a user-assigned managed identity
+            --commit-trigger-enabled false --pull-request-trigger-enabled false \\
```
jaysterp: @rosanch ditto here on removing the pr trigger. Also let's include disabling the base image update trigger here as well as we specify "no triggers".

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -578,69 +578,36 @@
...
+            --platform Windows/amd64
+  - name: Create a Linux multi-step task from a public GitHub repository with with both system-assigned and user-assigned managed identities and with a git commit, pull request, base image, and timer trigger that runs the task at noon on Mondays through Fridays with the trigger name provided.
```
jaysterp: @rosanch nit: "...with git commit, pull request, base image, and timer triggers that run..."

jaysterp: nit: "...with the timer trigger name provided."

src/azure-cli/azure/cli/command_modules/acr/_help.py
jaysterp: @rosanch you will need to explicitly enable the pr trigger here as it is disabled by default.

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -902,8 +869,15 @@
   - name: Update platform for the Build step of your Task to Windows (prev Linux).
     text: >
         az acr task update -n MyTask -r MyRegistry --platform Windows
+  - name: Update task's triggers and context for an Azure Container Registry.
+    text: |
+        az acr task update -n hello-world -r MyRegistry -f Dockerfile \\
+            --commit-trigger-enabled true --pull-request-trigger-enabled false \\
```
jaysterp: @rosanch as the pr trigger is disabled by default, I would explicitly say in the description that we are disabling a pull-request-trigger because not all users will have this enabled on the tasks, like the commit trigger.

src/azure-cli/azure/cli/command_modules/acr/_help.py
rosanch: From my experience with Linux users as long as you say it takes the info from standard input they assume you can pipe it. But I was also conflicted as if to remove it or not. I do think more people will use the pipe, but the input stream seemed more educational. 
It's up to you. However, I do think it would be better to have only one Standard input example. In case you want both, I could also mention in the description that it can be piped, put the pipe example and mention in can be taken from the input stream, or just add both examples within the same text section.

src/azure-cli/azure/cli/command_modules/acr/_help.py
rosanch: sure, I removed it originally because it's so long that it breaks the help format when printed.

PR #11180, opened 2019-11-11
Trim memory configuration for Basic SKU redis cache

### LINKED ISSUE #2902: [Redis] Issues with Redis commands
Reviewing the updated Redis commands, I found the following issues:

- [ ] If I simply run `az redis update -g myRG -n myRedisCache` with no additional parameters (essentially PUTting the object back unchanged), I get the following error (created with Basic SKU):
```
The setting 'maxmemory-reserved' is not allowed on 'Basic' cache instances.
RequestID=f76b1c36-7b1b-4620-8afa-42549d39e3a4
```
- [ ] If I try to change the SKU from Basic (C2) to Premium, I get the following error:
```
Exception of type 'AzCacheRP.Exceptions.SkuFamilyNotSupportedException' was thrown.
RequestID=d7a662ad-5d0b-4222-b3c4-f9deb2a933e4
```
- [x] The help text for SKU on `redis create` shows the allowed values twice:
```
    --sku               [Required]: What type of redis cache to deploy. Valid values: (Basic,
                                    Standard, Premium).  Allowed values: Basic, Premium, Standard.
```

MECHANDR:
Avoid setting memory configs for Basic SKU cache. Fixes #2902
---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/redis/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/redis/custom.py
@@ -34,6 +34,16 @@ def cli_redis_update(cmd, instance, sku=None, vm_size=None):
     if vm_size is not None:
         instance.sku.family = vm_size[0]
         instance.sku.capacity = vm_size[1:]
+    
+    # avoid setting memory configs for basic sku
+    if instance.sku.name == 'Basic':
+        if 'maxmemory-reserved' in instance.redis_configuration:
+            instance.redis_configuration.pop('maxmemory-reserved')
+        if 'maxfragmentationmemory-reserved' in instance.redis_configuration:
+            instance.redis_configuration.pop('maxfragmentationmemory-reserved')
+        if 'maxmemory-delta' in instance.redis_configuration:
+            instance.redis_configuration.pop('maxmemory-delta')
+
 
     # pylint: disable=too-many-function-args
     update_params = RedisUpdateParameters(
--- a/src/azure-cli/azure/cli/command_modules/redis/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/redis/custom.py
@@ -34,7 +34,7 @@ def cli_redis_update(cmd, instance, sku=None, vm_size=None):
     if vm_size is not None:
         instance.sku.family = vm_size[0]
         instance.sku.capacity = vm_size[1:]
-    
+
     # avoid setting memory configs for basic sku
     if instance.sku.name == 'Basic':
         if 'maxmemory-reserved' in instance.redis_configuration:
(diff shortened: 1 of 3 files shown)


MECHANDR:
@haroldrandom Can you please review this PR

src/azure-cli/azure/cli/command_modules/redis/custom.py
```diff
@@ -35,6 +35,15 @@ def cli_redis_update(cmd, instance, sku=None, vm_size=None):
...
 
+    # avoid setting memory configs for basic sku
+    if instance.sku.name == 'Basic':
+        if 'maxmemory-reserved' in instance.redis_configuration:
+            instance.redis_configuration.pop('maxmemory-reserved')
+        if 'maxfragmentationmemory-reserved' in instance.redis_configuration:
+            instance.redis_configuration.pop('maxfragmentationmemory-reserved')
+        if 'maxmemory-delta' in instance.redis_configuration:
+            instance.redis_configuration.pop('maxmemory-delta')
+
```
haroldrandom: Why those three property values of `maxmemory-reserved`, `maxfragmentationmemory-reserved`, `maxmemory-delta` should be returned?
I mean those property values seem like shouldn't belong to the instance created with Basic SKU, should be trimmed.

MECHANDR: These property values are read-only for Basic SKU cache, so they need to be returned but they cannot be updated.

haroldrandom: These three values are allowed for Standard and Premium SKU?

MECHANDR: Yes. they are available for Standard and Premium SKUs. Documentation link: <https://docs.microsoft.com/en-us/azure/azure-cache-for-redis/cache-configure#memory-policies>

haroldrandom: I know. But I insist that these property values should be invisble for user and service-side should handle those exception Instead of cover that in the frontend side.

Imagine that, if user directly call API to manipulate Redis and he/she shoule get the same warning from service side because he/she saw those properties before and than he/shee finally aware of thoses properties are read-only for current SKU. What if those properties got trimmed at the first place? What if service-side can deal with those unnecessary properties for Basic SKU?

Those values are not present in azure-sdk-for-python too.  azure-cli can cover those details for service-side easily but I think too much.

MECHANDR: Service side cannot hide these properties on GET call, as these properties needs to be displayed to user and are currently displayed as uneditable fields in azure portal.
If user directly call API with these properties, service side will throw appropriate error. We cannot swallow this error, as user might think that they are able to set these properties.

haroldrandom: So there is no better way to walk it around. It would be better to do it in azure-cli

PR #11194, opened 2019-11-12
Adding context path in task run

ElianaLiberman:
---


src/azure-cli/azure/cli/command_modules/acr/task.py
```diff
@@ -34,6 +37,7 @@
 ALLOWED_TASK_FILE_TYPES = ('.yaml', '.yml', '.toml', '.json', '.sh', '.bash', '.zsh', '.ps1',
                            '.ps', '.cmd', '.bat', '.ts', '.js', '.php', '.py', '.rb', '.lua')
 
+NULL_SOURCE_LOCATION = "/dev/null"
```
ElianaLiberman: Please ignore the redundant imports. I was testing some code and forgot to remove it. I will send a new iteration removing it thanks!

.vscode/launch.json
```diff
@@ -9,7 +9,17 @@
             "program": "${workspaceRoot}/src/azure-cli/azure/cli/__main__.py",
             "cwd": "${workspaceRoot}",
             "args": [
-                "--help"
+                "acr",
```
northtyphoon: it's a global vscode setting on the project. can you revert it?

src/azure-cli/azure/cli/command_modules/acr/task.py
```diff
@@ -982,3 +990,34 @@ def _get_trigger_event_list_patch(cmd,
             if SourceTriggerEvent.pullrequest.value in source_trigger_events:
                 source_trigger_events.remove(SourceTriggerEvent.pullrequest.value)
     return source_trigger_events
+
+def prepare_context_path(source_location, client_registries, registry_name, resource_group_name):
```
northtyphoon: do we already have similar function?

northtyphoon:
@rosanch @shahzzzam @jaysterp

src/azure-cli/azure/cli/command_modules/acr/task.py
rosanch: yes this is pretty much the same as prepare_source_location in run.py

rosanch: @ElianaLiberman  would you mind moving this function to the utils so we can have the same code running for both task and run?

jaysterp:
@ElianaLiberman please update `_help.py` with an example on how to use this context in `az acr task run`. Also please archive this change info in `History.rst`.

.vscode/launch.json
ElianaLiberman: Sorry, this will be reverted

src/azure-cli/azure/cli/command_modules/acr/task.py
```diff
@@ -982,3 +990,34 @@ def _get_trigger_event_list_patch(cmd,
...
+    if source_location.lower() == NULL_SOURCE_LOCATION:
```
rosanch: Add null check. Otherwise, it will fail when there's no context. This function in run doesn't need the check because the context is mandatory.

ElianaLiberman:
> @ElianaLiberman please update `_help.py` with an example on how to use this context in `az acr task run`. Also please archive this change info in `History.rst`.

Do I manually change the file or is there any command to run?

jaysterp:
@ElianaLiberman you manually change the file. See this example from Rodrigo's PR https://github.com/Azure/azure-cli/pull/11136/commits/e5e8d15fbd53c0f4fd50707ed1381a79854badf7

src/azure-cli/azure/cli/command_modules/acr/task.py
rosanch: could you move this value to the _constants.py please?

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -821,6 +821,9 @@
   - name: Trigger a task run by overriding the context and file passed during Task create.
     text: >
         az acr task run -n MyTask -r MyRegistry -c https://github.com/Azure-Samples/acr-build-helloworld-node.git -f Dockerfile
+  - name: Trigger a task run by overriding the context and file passed during Task create with a local context.
+    text: >
+        az acr task run -n MyTask -r MyRegistry -c . -f Dockerfile
```
rosanch: could you add something more visual than a dot?
e.g. ./workspace or ./folder

src/azure-cli/azure/cli/command_modules/acr/_utils.py
```diff
@@ -21,8 +24,12 @@
 )
 from ._client_factory import cf_acr_registries
 
+from ._archive_utils import upload_source_code, check_remote_source_code
+
 logger = get_logger(__name__)
 
+NULL_CONTEXT = '/dev/null'
```
rosanch: would you mind moving this value to the _constants.py

src/azure-cli/azure/cli/command_modules/acr/_utils.py
```diff
@@ -431,7 +438,36 @@ def parse_actions_from_repositories(allow_or_remove_repository):
 
     return actions
 
+def prepare_source_location(source_location, client_registries, registry_name, resource_group_name):
+    if source_location.lower() == NULL_CONTEXT:
```
rosanch: null check for source_location

src/azure-cli/azure/cli/command_modules/acr/_utils.py
```diff
@@ -431,7 +438,36 @@ def parse_actions_from_repositories(allow_or_remove_repository):
...
+        source_location = None
+    elif os.path.exists(source_location):
+        if not os.path.isdir(source_location):
+            raise CLIError(
+                "Source location should be a local directory path or remote URL.")
+
+        tar_file_path = os.path.join(tempfile.gettempdir(
+        ), 'task_run_archive_{}.tar.gz'.format(uuid.uuid4().hex))
```
rosanch: can we have a dynamic name here? 
I would make easier to track the source based on the command (e.g. build_archive, run_archive, task_run_archive) when troubleshooting.

src/azure-cli/azure/cli/command_modules/acr/task.py
```diff
@@ -778,6 +779,10 @@ def acr_task_run(cmd,  # pylint: disable=too-many-locals
         update_trigger_token = base64.b64encode(update_trigger_token.encode()).decode()
 
     task_id = get_task_id_from_task_name(cmd.cli_ctx, resource_group_name, registry_name, task_name)
+
+    if context_path:
```
rosanch: If you add a null check in the function remove the if here

src/azure-cli/azure/cli/command_modules/acr/_utils.py
jaysterp: She has a null check around the task run call. As you stated earlier, we don't need a null check for all uses of this method. In that case, it's best to let the caller handle the null check.

rosanch: I'm good with that. That comment is missing a question mark. I was just wondering if it is better to have it on the function now that is gonna be in the utils folder.

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -789,6 +789,9 @@
   - name: Trigger a task run by overriding the context and file passed during Task create.
```
adewaleo: Should this add something like the following?
```suggestion
  - name: Trigger a task run by overriding the context and file passed during Task create with a remote repository
```

Not sure the correct wording / task context lingo. But suggesting this change, given the example you are adding is showing how to override the context with a local path and this existing is overriding with a remote repo / path...

Please choose a better wording, than the suggestion I made, the point is we should explain what kind of context we are overrideing the original one with.

src/azure-cli/azure/cli/command_modules/acr/_utils.py
```diff
@@ -431,7 +437,36 @@ def parse_actions_from_repositories(allow_or_remove_repository):
 
     return actions
 
+def prepare_source_location(source_location, client_registries, registry_name, resource_group_name):
+    if source_location is None or source_location.lower() == ACR_NULL_CONTEXT:
```
adewaleo: I know you simply moved this function around, but here is a comment:

What if the user passes an empty string? What behavior do we want? Should it also be handled as the null context or crash below when we check for remote repo? If you want `""` to resolve to `ACR_NULL_CONTEXT`, then consider:

```suggestion
    if not source_location or source_location.lower() == ACR_NULL_CONTEXT:
```

src/azure-cli/azure/cli/command_modules/acr/_utils.py
```diff
@@ -431,7 +437,36 @@ def parse_actions_from_repositories(allow_or_remove_repository):
...
+            raise CLIError(err)
+        finally:
+            try:
+                logger.debug(
+                    "Deleting the archived source code from '%s'...", tar_file_path)
+                os.remove(tar_file_path)
+            except OSError:
+                pass
+    else:
+        source_location = check_remote_source_code(source_location)
```
adewaleo: I am assuming this is the logic that currently handles the empty string `source_location`

src/azure-cli/azure/cli/command_modules/acr/task.py
```diff
@@ -778,6 +781,9 @@ def acr_task_run(cmd,  # pylint: disable=too-many-locals
         update_trigger_token = base64.b64encode(update_trigger_token.encode()).decode()
 
     task_id = get_task_id_from_task_name(cmd.cli_ctx, resource_group_name, registry_name, task_name)
+
+    context_path = prepare_source_location(context_path, client_registries, registry_name, resource_group_name)
```
adewaleo: Might be overkill to do this but it won't hurt to add a test for running acr task run with local context. This is if the operation returns/creates an arm resource with the context field. If so, the test should check that the created run resource contains the local context path...

src/azure-cli/azure/cli/command_modules/acr/task.py
```diff
@@ -16,16 +16,19 @@
     get_yaml_template,
     build_timers_info,
     remove_timer_trigger,
-    get_task_id_from_task_name
+    get_task_id_from_task_name,
+    prepare_source_location
 )
 from ._stream_utils import stream_logs
+from ._constants import (
```
jaysterp: @ElianaLiberman since we now have a const for `ACR_NULL_CONTEXT`, please also update the reference in run.py

ElianaLiberman:
@adewaleo Hi! I think this PR is ready, could you please merge it? thanks!

adewaleo:
@yonzhan @qwordy ? Could someone from the CLI team approve and merge?

yonzhan:
@qwordy please help with the PR.

PR #11229, opened 2019-11-14
Install work in python3.8

SabatierBoris:
[`platform.linux_distribution`](https://docs.python.org/2/library/platform.html#platform.linux_distribution) is removed in python 3.8 a we should use [distro](https://pypi.org/project/distro/) package instead.

---

### DIFF
--- a/scripts/curl_install_pypi/install.py
+++ b/scripts/curl_install_pypi/install.py
@@ -43,10 +43,10 @@ AZ_DISPATCH_TEMPLATE = """#!/usr/bin/env bash
 {install_dir}/bin/python -m azure.cli "$@"
 """
 
-VIRTUALENV_VERSION = '15.0.0'
+VIRTUALENV_VERSION = '16.7.7'
 VIRTUALENV_ARCHIVE = 'virtualenv-'+VIRTUALENV_VERSION+'.tar.gz'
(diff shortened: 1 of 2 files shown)


scripts/curl_install_pypi/install.py
```diff
@@ -24,6 +24,7 @@
 import shutil
 import subprocess
 import hashlib
+import distro
```
haroldrandom: 1. Don't need to add dependency requirement and update *.py3.*.txt and setup.py?
2. According to https://pypi.org/project/distro/ , "Distro currently supports Linux and BSD based systems but Windows and OS X support is also planned."  What about Windows platform?

haroldrandom:
Sorry, adding this line won't fix the problem.

We are considering using other way to keey this script working.

haroldrandom:
Hey, @SabatierBoris Thanks for you contribution!

We have received a lot issues about that installation problem. And sincerely considered your suggestion. [distro ](https://github.com/nir0s/distro)is a really good alternative of `platform.dist()` and `platform.linux_distribution()`. And it's recommended by **https://docs.python.org/3.7/library/platform.html**.

However, the installation script is a bit irrational, do much things for user and its code is not so clean. We may refine that script in the future. 
Also, considering we only need a small part of the functionality. we descided not to introduce **distro** into azure-cli. We fix that by a small simple alternative function. 

It's possible that we will bring in **distro** in some days.

We did get inspired by your idea. Thanks again.

Issue #6949, opened 2018-08-02
workspace-resource-id parameter format would be useful to document


mrdfuse:
I actually had to contact support and they also had some trouble figuring it out, the workspace-resource-id parameter for enabling the monitor addon needs to be in the format: The resourceid should be like this :
/subscriptions/[your sub]/resourcegroups/[your oms rg]/providers/Microsoft.OperationalInsights/workspaces/[your oms workspace name] 
and not just the actual id, as the parameter name implies

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: d75c3a8d-1187-bea1-a7d4-7ebba71c1790
* Version Independent ID: 75b0c352-762e-16ae-5cc6-d807ffc1dc3f
* Content: [az aks](https://docs.microsoft.com/en-us/cli/azure/aks?view=azure-cli-latest#az-aks-enable-addons)
* Content Source: [latest/docs-ref-autogen/aks.yml](https://github.com/Azure/azure-docs-cli-python/blob/live/latest/docs-ref-autogen/aks.yml)
* Service: **container-service**
* GitHub Login: @rloutlaw
* Microsoft Alias: **routlaw**

tjprescott:
How is that not the actual ID?

jemag:
I think the confusion stems from the fact that inside the Azure Portal, the "Workspace Id" is given as a simple GUID. However, the az cli needs the "workspace-resource-id" which is more of a URL (like the one provided by mrdfuse).

I do not know if you can actually get it from the portal, I personally found it using the azure cli:
az resource list

tjprescott:
@mrdfuse @jemag thanks for the clarification.

@mboersma we've added some guidance for parameters that end with `-id` recently. https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md#argument-naming-conventions

Based on that, `-id` would imply a GUID is acceptable. When something is an ARM resource ID, we omit the `-id` in the name and call out in the help text.  The main reason is that something which accepts an ARM resource ID should typically be following the "name or ID" convention in the CLI.

https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#supporting-name-or-id-parameters

In this case, my recommendation would be to expose `--workspace` (name or ARM ID) and deprecate `--workspace-resource-id`.

gaziqbal:
[...]

PR #11241, opened 2019-11-14
Change flag workspace-resource-id to workspace-resource

troy0820:
Issue https://github.com/Azure/azure-cli/issues/6949#issuecomment-456492503 describes this problem across AKS and Azure Red Hat Openshift when trying to use monitoring when creating clusters.  To clearly distinguish the parameter that Azure Red Hat Openshift needs to create a cluster with monitoring, changing this parameter to `workspace-resource` and providing the context in the help menu when using `az openshift create -h` will allow the user to see what is expected.  The conversation was in PR #10775 and settled on the former but subsequent work will create issues of handling the id with future commands. 

- [x] Change help and parameters to match "workspace-resource"
---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acs/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/_help.py
@@ -896,7 +896,7 @@ parameters:
   - name: --customer-admin-group-id
     type: string
     short-summary: The Object ID of an Azure Active Directory Group that memberships will get synced into the OpenShift group "osa-customer-admins". If not specified, no cluster admin access will be granted.
-  - name: --workspace-resource
+  - name: --workspace-id
     type: string
     short-summary: The resource id of an existing Log Analytics Workspace to use for storing monitoring data.
 
@@ -911,7 +911,7 @@ examples:
   - name: Create an Openshift cluster using a custom vnet
     text: az openshift create -g MyResourceGroup -n MyManagedCluster --vnet-peer "/subscriptions/0000000-0000-0000-0000-000000000000/resourceGroups/openshift-vnet/providers/Microsoft.Network/virtualNetworks/test"
   - name: Create an Openshift cluster with Log Analytics monitoring enabled
-    text: az openshift create -g MyResourceGroup -n MyManagedCluster --workspace-resource "/subscriptions/0000000-0000-0000-0000-000000000000/resourceGroups/<resourcegroup>/providers/Microsoft.OperationalInsights/workspaces/<workspace-resource-id>"
+    text: az openshift create -g MyResourceGroup -n MyManagedCluster --workspace-id "/subscriptions/0000000-0000-0000-0000-000000000000/resourceGroups/<resourcegroup>/providers/Microsoft.OperationalInsights/workspaces/<workspace-resource-id>"
 """
 
 helps['openshift delete'] = """
--- a/src/azure-cli/azure/cli/command_modules/acs/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/_help.py
@@ -911,7 +911,7 @@ examples:
   - name: Create an Openshift cluster using a custom vnet
     text: az openshift create -g MyResourceGroup -n MyManagedCluster --vnet-peer "/subscriptions/0000000-0000-0000-0000-000000000000/resourceGroups/openshift-vnet/providers/Microsoft.Network/virtualNetworks/test"
   - name: Create an Openshift cluster with Log Analytics monitoring enabled
-    text: az openshift create -g MyResourceGroup -n MyManagedCluster --workspace-id "/subscriptions/0000000-0000-0000-0000-000000000000/resourceGroups/<resourcegroup>/providers/Microsoft.OperationalInsights/workspaces/<workspace-resource-id>"
+    text: az openshift create -g MyResourceGroup -n MyManagedCluster --workspace-id "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/MyResourceGroup/providers/Microsoft.OperationalInsights/workspaces/{workspace-id}"
 """
 
 helps['openshift delete'] = """
--- a/src/azure-cli/azure/cli/command_modules/acs/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/_help.py
@@ -896,9 +896,9 @@ parameters:
   - name: --customer-admin-group-id
     type: string
     short-summary: The Object ID of an Azure Active Directory Group that memberships will get synced into the OpenShift group "osa-customer-admins". If not specified, no cluster admin access will be granted.
-  - name: --workspace-resource-id
+  - name: --workspace-resource
     type: string
-    short-summary: The resource ID of an existing Log Analytics Workspace to use for storing monitoring data.
+    short-summary: The resource id of an existing Log Analytics Workspace to use for storing monitoring data.
 
 
 examples:
@@ -911,7 +911,7 @@ examples:
   - name: Create an Openshift cluster using a custom vnet
     text: az openshift create -g MyResourceGroup -n MyManagedCluster --vnet-peer "/subscriptions/0000000-0000-0000-0000-000000000000/resourceGroups/openshift-vnet/providers/Microsoft.Network/virtualNetworks/test"
   - name: Create an Openshift cluster with Log Analytics monitoring enabled
-    text: az openshift create -g MyResourceGroup -n MyManagedCluster --workspace-resource-id {WORKSPACE_RESOURCE_ID}
+    text: az openshift create -g MyResourceGroup -n MyManagedCluster --workspace-resource "/subscriptions/0000000-0000-0000-0000-000000000000/resourceGroups/<resourcegroup>/providers/Microsoft.OperationalInsights/workspaces/<workspace-resource-id>"
 """
 
 helps['openshift delete'] = """
--- a/src/azure-cli/azure/cli/command_modules/acs/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/custom.py
@@ -3125,7 +3125,7 @@ def openshift_create(cmd, client, resource_group_name, name,  # pylint: disable=
                      vnet_peer=None,
                      tags=None,
                      no_wait=False,
-                     workspace_resource=None,
+                     workspace_id=None,
                      customer_admin_group_id=None):
 
     if location is None:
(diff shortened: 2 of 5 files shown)


troy0820:
@MyronFanQiu Can you take a look at this?

MyronFanQiu:
@troy0820 Hello. If you want to do this, you should support "Name of workspace". Actually I'm glad to see this PR. 😄 BTW, `workspace_resource` is a strange word.

olga-mir:
@troy0820 @MyronFanQiu what do you think abut using only `--workspace`? 
According to guidelines: `Arguments that end in -id should be GUIDs.`. I believe the id part of this name is what causing the majority of confusion among users, it invites to put the "id"-like value. In the Azure Monitor UI what you will see is "Workspace Resource id" field with a guid value, which doesn't help to either. 
Have a look at this issue and particularly this comment: https://github.com/Azure/azure-cli/issues/6949#issuecomment-456492503

MyronFanQiu:
@olga-mir @troy0820 Hi. Like @tjprescott said in the https://github.com/Azure/azure-cli/issues/6949#issuecomment-456492503, the main reason behind this guideline is that " The main reason is that something which accepts an ARM resource ID should typically be following the "name or ID" convention in the CLI." 
`--workspace` is a better argument for user if it supports both ID or Name of the workspace. We don't need to create a default workspace for user. If user gives the name of the workpsace, we usually assume that this workspace is under the same resource group with openshift.

olga-mir:
thanks for clarifying this @MyronFanQiu,  I can see your point, you can read this comment in both ways. If most people agree on this name then it's fine. @troy0820 did provide help text, so hopefully that helps to resolve confusion.

src/azure-cli/azure/cli/command_modules/acs/_help.py
```diff
@@ -911,7 +911,7 @@
   - name: Create an Openshift cluster using a custom vnet
     text: az openshift create -g MyResourceGroup -n MyManagedCluster --vnet-peer "/subscriptions/0000000-0000-0000-0000-000000000000/resourceGroups/openshift-vnet/providers/Microsoft.Network/virtualNetworks/test"
   - name: Create an Openshift cluster with Log Analytics monitoring enabled
-    text: az openshift create -g MyResourceGroup -n MyManagedCluster --workspace-resource-id {WORKSPACE_RESOURCE_ID}
+    text: az openshift create -g MyResourceGroup -n MyManagedCluster --workspace-id "/subscriptions/0000000-0000-0000-0000-000000000000/resourceGroups/<resourcegroup>/providers/Microsoft.OperationalInsights/workspaces/<workspace-resource-id>"
```
olga-mir: nit: would it be better to use `MyResourceGroup` or `MyRG` style instead of `<resourcegroup>? Other help text strings that include ARM ID don't use `<>` (`git grep "00000000-0000-0000-0000-000000000000" | grep -v tests`)

MyronFanQiu:
@troy0820 Hello. Any progress on this PR? I remembered the GA date is around today.

troy0820:
@MyronFanQui I can fix the nit that @olga-mir pointed out.  I didn't update the PR to support both the name of the workspace.  I can do that and have it hopefully land tomorrow.

troy0820:
@MyronFanQiu I've updated the PR to accept the string that is just the workspace-id.  If the string is just that, it will create the "/subscription/...." necessary to provide the monitor profile for `az openshift create`.  If the user inputs the "/subscription/..." it will default to that and create the monitor profile.  Can you take a look?

src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -3197,13 +3197,16 @@ def openshift_create(cmd, client, resource_group_name, name,  # pylint: disable=
...
+    if workspace_id is not None:
+        workspace_id = workspace_id.strip()
+        if not workspace_id.startswith('/'):
+            workspace_id = '/' + workspace_id
+        if workspace_id.endswith('/'):
+            workspace_id = workspace_id.rstrip('/')
+        if len(workspace_id) == 36 and bool(re.match(r'^(\d|\w{8})-(\d|\w{4})-(\d|\w{4})-(\d|\w{4})-(\d|\w{12})$', workspace_id)):  # pylint: disable=line-too-long
+            subscription = get_subscription_id(cmd.cli_ctx)
+            workspace_id = '/subscriptions/{}/resourceGroups/{}/providers/Microsoft.OperationalInsights/workspaces/{}'.format(subscription, resource_group_name, workspace_id)  # pylint: disable=line-too-long
+        monitor_profile = OpenShiftManagedClusterMonitorProfile(enabled=True, workspace_id=workspace_id)  # pylint: disable=line-too-long
```
MyronFanQiu: `workspace_id=workspace_id` instead of `workspace_resource_id=workspace_id`?

src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -3197,13 +3197,16 @@ def openshift_create(cmd, client, resource_group_name, name,  # pylint: disable=
...
-        monitor_profile = OpenShiftManagedClusterMonitorProfile(enabled=True, workspace_resource_id=workspace_resource_id)  # pylint: disable=line-too-long
+    if workspace_id is not None:
+        workspace_id = workspace_id.strip()
+        if not workspace_id.startswith('/'):
+            workspace_id = '/' + workspace_id
+        if workspace_id.endswith('/'):
+            workspace_id = workspace_id.rstrip('/')
+        if len(workspace_id) == 36 and bool(re.match(r'^(\d|\w{8})-(\d|\w{4})-(\d|\w{4})-(\d|\w{4})-(\d|\w{12})$', workspace_id)):  # pylint: disable=line-too-long
+            subscription = get_subscription_id(cmd.cli_ctx)
+            workspace_id = '/subscriptions/{}/resourceGroups/{}/providers/Microsoft.OperationalInsights/workspaces/{}'.format(subscription, resource_group_name, workspace_id)  # pylint: disable=line-too-long
```
MyronFanQiu: Is it possible for server end?

troy0820: @JackQuincy mentioned in the other PR that this has to be done client side.  AKS does the same validation for their workspace-resource-id on line 2204.  

The regex validation was checking the subscription but that is wrong.  Changing it to append the name of the workspace resource id to the string to build the monitor profile.

src/azure-cli/azure/cli/command_modules/acs/custom.py
troy0820: If the string passed in to `--workspace-id` is, in fact, jus the name of the workspace-id, the variable will still be workspace-id and reference the string to build the workspace-resource-id _(/subscriptions/...)_.

src/azure-cli/azure/cli/command_modules/acs/custom.py
MyronFanQiu: Can you add a test for this command? It's a little bit strange to have a id like `'/subscriptions/{}/resourceGroups/{}/providers/Microsoft.OperationalInsights/workspaces/{}'.format(subscription, resource_group_name, workspace_id)` The workspace_id is something like xxxx-xxxx-xxxx-xxxx-xxxx-xxxx which is strange.

troy0820: @MyronFanQiu  I'm confused.  When you do `az resource list` the **id** of the loganalytics workspace is this string:
 ```
/subscriptions/{subscription}/resourceGroups/{resourceGroup}/providers/Microsoft.OperationalInsights/workspaces/{workspace-name}
```
The `workspace name` is the just 4-63 characters after `/workspaces/`.  The logic allows for either the string to be passed in with the `workspace name` as the `id` or just the workspace name which builds the id because the OpenshiftMonitorProfile needs the whole id to create the managed cluster. 

I will indeed create a test for the functionality of the above logic.  

The id I believe you are describing is the subscription id that is passed with the id of the loganalytics workspace id.  Unless I'm mistaken.

Comment (MyronFanQiu) [reply src/azure-cli/azure/cli/command_modules/acs/custom.py]: @troy0820 Oh. Since the name of the argument is `workspace-id`, I thought this command supported both `workspace_resource_id` and `workspace_id`. If you are intend to support Name or Resource ID. Usually, the name of the argument should be just `workspace` and specify the detail information in the help message. We still need a test for this command. Thanks~

Comment (troy0820) [reply src/azure-cli/azure/cli/command_modules/acs/custom.py]: I will add a test for this but will also wait until this PR https://github.com/Azure/azure-cli/pull/11460 is merged and I can rebase.  It changes the imports to not obscure the API version from the caller and only use the preview 0930 imports.

File: src/azure-cli/azure/cli/command_modules/acs/tests/latest/test_osa_commands.py [hunk 4]
```diff
@@ -153,3 +153,39 @@ def test_openshift_create_default_service_no_aad(self, resource_group, resource_
...
+    def test_openshift_create_with_monitoring(self, resource_group, resource_group_location, aad_client_app_id, aad_client_app_secret):
+        # kwargs for string formatting
+        osa_name = self.create_random_name('clitestosa', 15)
+        self.kwargs.update({
+            'resource_group': resource_group,
+            'name': osa_name,
+            'location': resource_group_location,
+            'aad_client_app_id': aad_client_app_id,
+            'aad_client_app_secret': aad_client_app_secret,
+            'workspace_id': "test-workspace"
```
MyronFanQiu: Is it a workspace name instead of workspace id?

JackQuincy: It is a workspace resource id.  The workspace name and id are 2 other fields that exist and won't work here.  We are using the example inputs to try to make it clear what kind of id it is.

troy0820: This is the workspace resource id as Jack said.  We are using the same mechanism that we use to force users to grab this id as we do for VNET peering listed in the example inputs.  Trying to mock this means I need a subscription.  I can test that the monitor profile is indeed `None` when creating a cluster.  The cluster will provision as before but with a blank monitoring profile. 

We also do not provide a default log analytics workspace if none is provided like AKS does with its add-on capabilities.

MyronFanQiu: Looks like you delete the logic and it's just a renaming PR. Right?

src/azure-cli/azure/cli/command_modules/acs/tests/latest/test_osa_commands.py
```diff
@@ -153,3 +153,37 @@ def test_openshift_create_default_service_no_aad(self, resource_group, resource_
...
+        self.kwargs.update({
+            'resource_group': resource_group,
+            'name': osa_name,
+            'location': resource_group_location,
+            'aad_client_app_id': aad_client_app_id,
+            'aad_client_app_secret': aad_client_app_secret
+        })
+
+        # create
+        create_cmd = 'openshift create --resource-group={resource_group} --name={name} --location={location} ' \
```
MyronFanQiu: is it able to link a workspace since this PR is related with `workspace`? We have `az monitor log-analytics workspace create`

src/azure-cli/azure/cli/command_modules/acs/tests/latest/test_osa_commands.py
troy0820: I deleted the logic because as of right now we do not want to support workspace name for the reasons @JackQuincy explained.  I am just using this PR to rename the parameters around the flag for documentation purposes. The subsequent work in the future should address these issues.  The user of this flag will be demonstrated within the examples and forced to used the existing workspace in the format that is similar to VNET peering.

MyronFanQiu:
@troy0820 Actually, this should be a breaking change for users. However, I just checked usage data and it's quite low. So I think it's ok.

MyronFanQiu:
@troy0820 Could you add [Breaking Change] in History.rst?

troy0820:
@MyronFanQiu I don't mind if you add it.  I think that'll work if that is easier.

MyronFanQiu:
@troy0820 to resolve conflicts in history, I added history in another PR https://github.com/Azure/azure-cli/pull/11522

PR #11245, opened 2019-11-15
[IoTHub] Add missing parameter to create storage endpoint

anusapan:
---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/iot/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/iot/_params.py
@@ -203,6 +203,17 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
                    help='Encoding format for the container. The default is AVRO. '
                         'Note that this field is applicable only for blob container endpoints.')
 
+    with self.argument_context('iot hub routing-endpoint create') as c:
+        c.argument('batch_frequency', options_list=['--batch-frequency', '-b'], type=int,
+                   help='The maximum amount of time that can elapse before data is'
+                        ' written to a blob, between 60 and 720 seconds.')
+        c.argument('chunk_size_window', options_list=['--chunk-size', '-w'], type=int,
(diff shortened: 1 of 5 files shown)


fengzhou-msft:
@Juliehzl please take a look.

src/azure-cli/azure/cli/command_modules/iot/_params.py
```diff
@@ -203,6 +203,17 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
                    help='Encoding format for the container. The default is AVRO. '
                         'Note that this field is applicable only for blob container endpoints.')
 
+    with self.argument_context('iot hub routing-endpoint create') as c:
+        c.argument('batch_frequency', options_list=['--batch-frequency', '-b'], type=int,
```
Juliehzl: Can you specify unit for your parameter in help message? And from my understanding, the value must be in [60, 720], right?

src/azure-cli/azure/cli/command_modules/iot/_params.py
```diff
@@ -203,6 +203,17 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
...
+                   help='The maximum amount of time that can elapse before data is'
+                        ' written to a blob, between 60 and 720 seconds.')
+        c.argument('chunk_size_window', options_list=['--chunk-size', '-w'], type=int,
```
Juliehzl: The same as above. Please specify unit here.

Juliehzl:
LGTM in general.

src/azure-cli/azure/cli/command_modules/iot/_params.py
anusapan: Updated

src/azure-cli/azure/cli/command_modules/iot/_params.py
anusapan: Updated

Issue #10705, opened 2019-09-30
TestSDK utilities.py fails to convert byte to string for test recording


Hazhzeng:
> ### `az feedback` auto-generates most of the information requested below, as of CLI version 2.0.73

**Describe the bug**
When implementing an E2E test scenario to upload a zip file to a storage account, the _py3_byte_to_str() function inside testsdk/utilities.py fails to convert byte to string.

This affects the recording in `env\lib\site-packages\azure_devtools\scenario_tests\base.py:157` if the test is decorated by StorageAccountPreparer

**To Reproduce**
1. Paste the following e2e test scenario in `azure-cli\src\azure-cli\azure\cli\command_modules\appservice\tests\latest\test_webapp_commands.py`
```python
class FunctionappRemoteBuildScenarioTest(ScenarioTest):
    @ResourceGroupPreparer(name_prefix='azurecli-functionapp-c-e2e-remotebuild')
    @StorageAccountPreparer(sku='Standard_LRS')
    def test_functionapp_remote_build(self, resource_group, storage_account):
        functionapp_name = self.create_random_name('functionapp-remotebuild-test', 40)
        plan_name = self.create_random_name('functionapp-remotebuild-plan', 40)
        zip_file = os.path.join(TEST_DIR, 'test.zip')
        self.cmd('functionapp plan create -g {} -n {} --sku S1 --is-linux true'.format(resource_group, plan_name))
        self.cmd('functionapp create -g {} -n {} --plan {} -s {} --os-type Linux --runtime python'.format(resource_group, functionapp_name, plan_name, storage_account))
        self.cmd('functionapp deployment source config-zip -g {} -n {} --src "{}"'.format(resource_group, functionapp_name, zip_file)).assert_with_checks([
            JMESPathCheck('status', 4),
            JMESPathCheck('deployer', 'Push-Deployer'),
            JMESPathCheck('message', 'Created via a push deployment'),
            JMESPathCheck('complete', True)
        ])
```
2. Activate your virtual environment, setup the development environment.
3. Run `azdev test test_functionapp_remote_build`
4. The e2e test fails with the following traceback.
```
(env) C:\Users\hazeng\Projects\GitHub\azure-cli [hazeng-pr4-remotebuild +3 ~4 -1 !]>  azdev test test_functionapp_remote_build

=============
| Run Tests |
=============

The tests are set to run against current profile latest.

[...]

PR #11251, opened 2019-11-15
Ignore errors in UTF-8 decoding process

bim-msft:
---

This PR is for fixing issue #10705


yonzhan:
@qwordy please help review this PR.

src/azure-cli-testsdk/azure/cli/testsdk/utilities.py
```diff
@@ -54,7 +54,7 @@ def _py3_byte_to_str(byte_or_str):
     logger = logging.getLogger()
     logger.warning(type(byte_or_str))
     try:
-        return str(byte_or_str, 'utf-8') if isinstance(byte_or_str, bytes) else byte_or_str
+        return str(byte_or_str, 'utf-8', errors='ignore') if isinstance(byte_or_str, bytes) else byte_or_str
```
haroldrandom: Why ignore encoding error here?

bim-msft: @haroldrandom This modification intended to bypass the VCR recoding error because in this case the request body is a zip file content, not a valid UTF-8. But this would cause replay process fail, so I decided to revert this change, and just make the test as LiveScenarioTest.

bim-msft: LiveScenarioTest is always time-consuming, for furthor development, we should make it possible to record/replay binary requests.

bim-msft:
This PR is not necessary.

PR #11275, opened 2019-11-18
Add Python 3.8 Test

fengzhou-msft:


azure-pipelines.yml
```diff
@@ -408,7 +413,7 @@ jobs:
       REDUCE_SDK: 'True'
       AZURE_CLI_TEST_TARGET_PROFILE: '2019-03-01'
 
-- job: RunAutomationReduced20180301
+- job: RunAutomation20180301
```
Juliehzl: Why do you run reduced tests for profile 2019-03-01 but 2018-03-01 not? What is the difference?

fengzhou-msft: This is the current test settings for python 3.6. I modified the name to reflect the settings.

azure-pipelines.yml
```diff
@@ -434,6 +439,86 @@ jobs:
...
+    inputs:
+      TargetPath: '$(Build.ArtifactStagingDirectory)/pypi'
+      artifactName: pypi
+  - bash: |
+      pip install --force-reinstall wheel
+  - task: Bash@3
+    displayName: 'Run Test Automation'
+    inputs:
+      targetType: 'filePath'
+      filePath: ./scripts/ci/test_automation.sh
```
Juliehzl: Actually I also have a question for test_automation.sh file. Typically we run test based on `azdev test`, but in ADO CI system, tests are run based on python https://github.com/Azure/azure-cli/blob/dev/scripts/ci/test_automation.sh#L57.

Juliehzl:
LGTM in general and leave some questions.

azure-pipelines.yml
Juliehzl: I see. Just out of curiosity, ReducedAutomation is run when REDUCE_SDK is set to true.  But through the [scripts](https://github.com/Azure/azure-cli/blob/dev/scripts/ci/test_automation.sh#L35), I cannot see anything special besides reduced modules.

fengzhou-msft: Not familiar with testing details, add @haroldrandom for digging in.

azure-pipelines.yml
fengzhou-msft: Another good question to look into, this could be a discrepancy in developing and ci testing, @haroldrandom

haroldrandom: Yeah. It's a TODO for CI

PR #11303, opened 2019-11-19
[Storage] Fix issues with wrong error message

Juliehzl:
---


src/azure-cli/azure/cli/command_modules/storage/__init__.py
```diff
@@ -185,6 +185,13 @@ def handler(ex):
                 """
                 logger.error(message)
                 return
+            if isinstance(ex, t_error) and ex.status_code == 403 and ex.error_code == 'AuthorizationFailure':
+                message = """
+You maybe not allowed by the network rules of storage account. Please check network rule set using 'az storage account show -n accountname --query networkRuleSet'.
```
MyronFanQiu: allowed to access storage account by ... ?

src/azure-cli/HISTORY.rst
```diff
@@ -59,6 +59,7 @@ Release History
 **Storage**
 
 * storage account create: Add --enable-hierarchical-namespace to support filesystem semantics in blob service.
+* Fix issues with wrong error message "You do not have the required permissions needed to perform this operation."
```
jsntcy: Is there an issue related to this PR?

src/azure-cli/azure/cli/command_modules/storage/__init__.py
```diff
@@ -172,7 +172,7 @@ def handler(ex):
             t_error = get_sdk(self.command_loader.cli_ctx,
                               ResourceType.DATA_STORAGE,
                               'common._error#AzureHttpError')
-            if isinstance(ex, t_error) and ex.status_code == 403:
+            if isinstance(ex, t_error) and ex.status_code == 403 and ex.error_code == 'AuthorizationPermissionMismatch':
```
jsntcy: >AuthorizationPermissionMismatch [](start = 87, length = 31)

How do you know which error_code should use? Is there a document about the error codes and their meaning?

src/azure-cli/HISTORY.rst
jiasli: This sounds better to me:
> Fix incorrect error message "You do not have the required permissions needed to perform this operation." when blocked by network rules.

src/azure-cli/azure/cli/command_modules/storage/__init__.py
jiasli: Upon narrowing `ex.error_code` to `AuthorizationPermissionMismatch` and `AuthorizationFailure`, if the key or connection string is incorrect, `ex.error_code` will be `AuthenticationFailed` which is not handled.

src/azure-cli/azure/cli/command_modules/storage/__init__.py
jiasli: How about
> The request may be blocked by network rules of the storage account.

PR #11310, opened 2019-11-19
[Profile] Return subscriptions even with --allow-no-subscriptions

### LINKED ISSUE #11059: Can't sign in with the -allow-no-subscriptions parameter
**Resource Provider**
AZ CLI
az login
https://docs.microsoft.com/en-us/cli/azure/reference-index?view=azure-cli-latest#az-login

Cannot sign in with the following command:
az login --service-principal -u <principal-id>-password <password>-tenant <tenant-id>--allow-no-subscriptions

If don't have the --allow-no-subscriptions parameter,  can sign in.

AZ CLI version 2.0.75

jiasli:
Fix #11059: `az login --allow-no-subscriptions` fails if there are subscriptions in the tenant

In `az login --allow-no-subscriptions`, only tenants with no subscriptions are listed and available subscriptions are missing from the subscription list. If the only tenant has some subscriptions, this command fails. 

This PR fixes this and honor the word "allow" so that available subscriptions are preserved and other tenants with no subscriptions are appended to the subscription list. 

To test:

```
az login --allow-no-subscriptions
```

---


src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -223,7 +223,8 @@ def find_subscriptions_on_login(self,
             t_list = [s.tenant_id for s in subscriptions]
```
jsntcy: Is it better to distinct t_list as there may be many dupplicated tenant_id?

jiasli: This it not necessary as `t_list` is for tenants which are not used to construct `bare_tenants`.

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -342,7 +343,8 @@ def find_subscriptions_in_vm_with_msi(self, identity_id=None, allow_no_subscript
             if allow_no_subscriptions:
                 subscriptions = self._build_tenant_level_accounts([tenant])
             else:
-                raise CLIError('No access was configured for the VM, hence no subscriptions were found')
+                raise CLIError('No access was configured for the VM, hence no subscriptions were found. '
+                               "If this is expected, use '--allow-no-subscriptions' to have tenant level accesses.")
```
qwordy: access, uncountable noun

jiasli: Nice catch. Fixed.

PR #11325, opened 2019-11-20
Revert el8 package build

fengzhou-msft:
We have merged #11235 to build el7 package with Python 3, I have tested it on RHEL 8/ Centos 8/Fedora 31, it works fine. 

Building and releasing an extra el8 package has some overheads such as extra package management work, going through Linux repo onboarding process and users need to change the base url for the package repo. At this moment, an el8 package doesn't seem necessary as long as el7 package works.


aghassemlouei:
It's worth noting that newer version of rpm-build will fix the issues with sha256 signing and use of package on FIPS enabled systems and actually resolve #20719.

archoversight:
Is there a possibility that this reversion could get reverted?

https://github.com/Azure/azure-cli/issues/20719#issuecomment-1000614066

says that FIPS is not on the roadmap, but using a newer version of rpm-build (which is shipped with CentOS 8 or RHEL 8) would alleviate the issues as the package would be appropriately signed and thus installable on systems that no longer support MD5 hashing.

jiasli:
CentOS 8 will be deprecated in 2 days from now: 

https://www.centos.org/centos-linux-eol/

> CentOS Linux 8 will reach End of Life on December 31, 2021.

I don't think it is a good idea to use CentOS 8 anymore.

Issue #11117, opened 2019-11-05
Specialized Image Support + Snapshot as creation source in Az CLI for Shared Image Gallery


agrawalkshitij:
**Resource Provider**
What is the Azure resource provider your feature is part of? Microsoft.Compute

**Description of Feature or Work Requested**
Provide a brief description of the feature or work requested. A link to conceptual documentation may be helpful too.
Support OSState = Specialized in Gallery image-definition

And Snapshots ids in storageProfile.osdiskImage.source.id, make managed-image as optional.

**Minimum API Version Required**
What is the minimum API version of your service required to implement your feature?2019-07-01

**Swagger Link**
Provide a link to the location of your feature(s) in the REST API specs repo. If your feature(s) has corresponding commit or pull request in the REST API specs repo, provide them. This should be on the master branch of the REST API specs repo.[gallery.json]

**Target Date**
If you have a target date for release of this feature/work, please provide it. While we can't guarantee these dates,
it will help us prioritize your request against other requests. 11/06/2019

fengzhou-msft:
@qwordy please take a look and evaluate the work.

yonzhan:
@qwordy  add to S161.

qwordy:
I have modified code to use storage profile in #11092, commit - Fix VMGalleryImage.test_gallery_e2e
Full features will be covered in another PR.

qwordy:
I've added --os-state and --hyper-v-generation.

qwordy:
If you have other questions, comment in this issue or open a new issue. Thanks.

PR #11336, opened 2019-11-21
[Compute] sig image-version create: Support --os-snapshot and --data-snapshot

qwordy:
Issue #11117 

---


src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -2768,19 +2768,31 @@ def create_gallery_image(cmd, resource_group_name, gallery_name, gallery_image_n
...
-    if not is_valid_resource_id(managed_image):
+    if managed_image and not is_valid_resource_id(managed_image):
         managed_image = resource_id(subscription=client.config.subscription_id, resource_group=resource_group_name,
                                     namespace='Microsoft.Compute', type='images', name=managed_image)
+    if os_snapshot and not is_valid_resource_id(os_snapshot):
+        os_snapshot = resource_id(subscription=client.config.subscription_id, resource_group=resource_group_name,
+                                  namespace='Microsoft.Compute', type='snapshots', name=os_snapshot)
+    data_snapshots = []
+    if data_snapshot is not None:
+        data_snapshots = data_snapshot.split()
```
MyronFanQiu: you should use nargs='+'

src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -793,6 +793,8 @@ def load_arguments(self, _):
                    help='Gallery image version in semantic version pattern. The allowed characters are digit and period. Digits must be within the range of a 32-bit integer, e.g. <MajorVersion>.<MinorVersion>.<Patch>')
         c.argument('description', help='the description of the gallery image version')
         c.argument('managed_image', help='image name(if in the same resource group) or resource id')
+        c.argument('os_snapshot', help='Name or ID of OS disk snapshot')
+        c.argument('data_snapshot', help='Names or IDs (space-delimited) of data disk snapshots')
```
MyronFanQiu: please use nargs=+. you can find ref here https://docs.python.org/3/library/argparse.html#nargs

MyronFanQiu: It cannot be updated. Right?

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -2768,19 +2768,31 @@ def create_gallery_image(cmd, resource_group_name, gallery_name, gallery_image_n
...
     end_of_life_date = fix_gallery_image_date_info(end_of_life_date)
-    if not is_valid_resource_id(managed_image):
+    if managed_image and not is_valid_resource_id(managed_image):
         managed_image = resource_id(subscription=client.config.subscription_id, resource_group=resource_group_name,
                                     namespace='Microsoft.Compute', type='images', name=managed_image)
+    if os_snapshot and not is_valid_resource_id(os_snapshot):
+        os_snapshot = resource_id(subscription=client.config.subscription_id, resource_group=resource_group_name,
+                                  namespace='Microsoft.Compute', type='snapshots', name=os_snapshot)
+    data_snapshots = []
+    if data_snapshot is not None:
```
MyronFanQiu: This line always return true.

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -2789,13 +2801,26 @@ def create_image_version(cmd, resource_group_name, gallery_name, gallery_image_n
...
+            for i, s in enumerate(data_snapshots):
+                data_disk_images.append(GalleryDataDiskImage(source=GalleryArtifactVersionSource(id=s), lun=i))
+        if managed_image is None and os_snapshot is None:
+            raise CLIError('usage error: Please provide --managed-image or --os-snapshot')
+        storage_profile = GalleryImageVersionStorageProfile(source=source, os_disk_image=os_disk_image,
+                                                            data_disk_images=data_disk_images)
         image_version = ImageVersion(publishing_profile=profile, location=location, tags=(tags or {}),
                                      storage_profile=storage_profile)
     else:
+        if managed_image is None:
```
MyronFanQiu: I cannot find the usage of `managed_image`. Why do you need to check it?

src/azure-cli/azure/cli/command_modules/vm/custom.py
> 
> This line always return true.

data_snapshots and data_snapshot  are different

src/azure-cli/azure/cli/command_modules/vm/custom.py
> 
> you should use nargs='+'

OK

src/azure-cli/azure/cli/command_modules/vm/_params.py
> 
> please use nargs=+. you can find ref here https://docs.python.org/3/library/argparse.html#nargs

OK

src/azure-cli/azure/cli/command_modules/vm/custom.py
> 
> I cannot find the usage of `managed_image`. Why do you need to check it?

It's used in profile.

src/azure-cli/azure/cli/command_modules/vm/_params.py
> 
> It cannot be updated. Right?

image-version cannot update the source (image or snapshot) now

PR #11337, opened 2019-11-21
[Storage] Remove unrelated exception from error message

jiasli:
Remove unrelated exception from error message:

```
The command failed with an unexpected error. Here is the traceback:

'CommandResultItem' object is not iterable
  File "C:\Users\VSSADM~1\AppData\Local\Temp\pip-install-umfcm_n2\knack\knack\cli.py", line 212, in invoke
  File "C:\Users\VSSADM~1\AppData\Local\Temp\pip-install-umfcm_n2\knack\knack\output.py", line 132, in out
  File "C:\Users\VSSADM~1\AppData\Local\Temp\pip-install-umfcm_n2\knack\knack\output.py", line 38, in format_json
TypeError: 'CommandResultItem' object is not iterable
```

The reason for this exception is because `azure.cli.command_modules.storage.StorageCommandGroup.get_handler_suppress_some_400` silences the exception and makes Knack think the command is successful. 

This PR changes the behavior of `get_handler_suppress_some_400` so that it only rewrites (not prints) the error message and leaves the exception to CLI Core to handle. CLI core now handles `AzureException` and returns error code 1 as expected. 

To test: If the container doesn't exist, CLI should exit with exit code 3:
```
az storage blob show --account-name {} -c foo -n bar.txt --auth-mode key
```

If not authenticated, CLI should give `AuthenticationFailed`:
```
az storage blob show --account-name {} --account-key="YQ==" -c foo -n bar.txt 
```

---


Juliehzl:
Please try to make CI pass.

src/azure-cli/azure/cli/command_modules/storage/__init__.py
```diff
@@ -244,12 +239,10 @@ def _merge_new_exception_handler(kwargs, handler):
     first = kwargs.get('exception_handler')
 
     def new_handler(ex):
-        try:
-            handler(ex)
-        except Exception:  # pylint: disable=broad-except
```
MyronFanQiu: Is any reason to remove try/except?

src/azure-cli/azure/cli/command_modules/storage/__init__.py
```diff
@@ -181,14 +179,11 @@ def handler(ex):
...
+If you want to use the old authentication method and allow querying for the right account key, please use the "--auth-mode" parameter and "key" value. ErrorCode: AuthenticationFailed
                 """
-                logger.error(message)
-                return
+                ex.args = (message,)
             if isinstance(ex, t_error) and ex.status_code == 409 and ex.error_code == 'NoPendingCopyOperation':
-                logger.error(ex.args[0])
-                return
-            raise ex
+                pass
```
Juliehzl: If pass this exception scenario, how can users know the exception details?

src/azure-cli/azure/cli/command_modules/storage/__init__.py
```diff
@@ -181,14 +179,11 @@ def handler(ex):
...
-If you want to use the old authentication method and allow querying for the right account key, please use the "--auth-mode" parameter and "key" value.
+If you want to use the old authentication method and allow querying for the right account key, please use the "--auth-mode" parameter and "key" value. ErrorCode: AuthenticationFailed
                 """
-                logger.error(message)
-                return
+                ex.args = (message,)
             if isinstance(ex, t_error) and ex.status_code == 409 and ex.error_code == 'NoPendingCopyOperation':
-                logger.error(ex.args[0])
-                return
-            raise ex
```
Juliehzl: Can you tell me why do you remove `raise ex` here? In this way, how do you handle other exceptions?

src/azure-cli/azure/cli/command_modules/storage/__init__.py
```diff
@@ -166,9 +166,7 @@ def storage_custom_command_oauth(self, *args, **kwargs):
     def get_handler_suppress_some_400(self):
         def handler(ex):
```
MyronFanQiu: Looks like it's not a handler anymore?

src/azure-cli/azure/cli/command_modules/storage/__init__.py
jiasli: The details are in `ex.args[0]`.

src/azure-cli/azure/cli/command_modules/storage/__init__.py
jiasli: Other exceptions are raised in the outer layer of the call stack.

src/azure-cli/azure/cli/command_modules/storage/__init__.py
jiasli: `handler` previously raises exception. Now it doesn't and exceptions are raised later in `new_handler.

src/azure-cli/azure/cli/command_modules/storage/__init__.py
jiasli: It can't handle exceptions. As the design of Knack, no other handler should mute the exception or Knack will have error. See the description of the PR.

src/azure-cli/azure/cli/command_modules/storage/__init__.py
jiasli: `handler` previously raises exception. Now it doesn't and the exception is raised later in `new_handler`. See ⬇ code.

src/azure-cli/azure/cli/command_modules/storage/__init__.py
jiasli: During run-time, `first` is `show_exception_handler`, which raises exception anyway. 

https://github.com/Azure/azure-cli/blob/e47ee99ea2ff1476d3943d81f0403b4dcc638221/src/azure-cli-core/azure/cli/core/commands/arm.py#L754-L761

Issue #10447, opened 2019-09-06
az storage container create fails


matt-FFFFFF:
## Describe the bug
Cannot create storage container using az cli.

**Command Name**
`az storage container create`

**Errors:**
```
The command failed with an unexpected error. Here is the traceback:


'CommandResultItem' object is not iterable
Traceback (most recent call last):
  File "/opt/az/lib/python3.6/site-packages/knack/cli.py", line 212, in invoke
    self.output.out(cmd_result, formatter=formatter, out_file=out_file)
  File "/opt/az/lib/python3.6/site-packages/knack/output.py", line 132, in out
    output = formatter(obj)
  File "/opt/az/lib/python3.6/site-packages/knack/output.py", line 38, in format_json
    input_dict = dict(result) if hasattr(result, '__dict__') else result
TypeError: 'CommandResultItem' object is not iterable
```

## To Reproduce:
Steps to reproduce the behavior. Note that argument values have been redacted, as they may contain sensitive information.

- Have existing v2 storage account
- az login with account that has blob owner and contributor access
- `az storage container create -n {containername} --account-name {existingstorageaccountname}`

## Expected Behavior
Container is created

## Environment Summary
```
Linux-4.19.59-microsoft-standard-x86_64-with-debian-buster-sid
Python 3.6.5
Shell: bash

azure-cli 2.0.72

Extensions:
azure-devops 0.12.0
front-door 1.0.0
virtual-wan 0.1.0

```

Juliehzl:
@matt-FFFFFF Can you check the following two things for your storage account? 1. Contributor role to the storage account, not Storage Blob Data Contributor role. 2. Is there an firewall for your storage account? If everything is right but still failed, can you show me the debug information with `--debug` appending your command? Please note to hide your personal storage key and any credential information.

cherimckeown:
I am having this issues as well.  My user has both the Contributor and Storage Blob Data Contributor roles.  There is no firewall for my storage account.  Attached is the debug output (minus my keys and identifiers): 
[debug_for_container_access_issue.txt](https://github.com/Azure/azure-cli/files/3611819/debug_for_container_access_issue.txt)

matt-FFFFFF:
Hi,

Confirming no firewall and that the account has contributor access to the RG. Will test again with —debut and report back.

tamirkamara:
I have the exact same issue with brand new storage account, no firewall rules and I'm the owner of the subscription. Very simple flow:
```
[...]

Issue #10826, opened 2019-10-12
Can't access blob storage from AZ CLI


andrePKI:
## Describe the bug
Account is signed in correctly, has the appropriate roles. However commands fail with "You do not have the required permissions" and listing roles the account is member of

**Command Name**
`az storage blob upload`

**Errors:**
```

C:\WINDOWS\system32>az logout

C:\WINDOWS\system32>az login -u dataman@adhoekstra.ddns.net -p <erased>
[
  {
    "cloudName": "AzureCloud",
    "id": "b069085a-22df-4c6f-ab7e-f4f6f1d52d48",
    "isDefault": true,
    "name": "Pay-As-You-Go",
    "state": "Enabled",
    "tenantId": "cb356271-ecee-4f09-8c38-3a98bb49b3dd",
    "user": {
      "name": "DataMan@adhoekstra.ddns.net",
      "type": "user"
    }
  }
]

C:\WINDOWS\system32>az role assignment list --all
[
  {
    "canDelegate": null,
    "id": "/subscriptions/b069085a-22df-4c6f-ab7e-f4f6f1d52d48/resourceGroups/throw-away/providers/Microsoft.Storage/storageAccounts/koffiestorage/providers/Microsoft.Authorization/roleAssignments/a7ef4aa0-fa19-40ab-ae8e-d397de65a166",
    "name": "a7ef4aa0-fa19-40ab-ae8e-d397de65a166",
    "principalId": "caa984b3-ddc2-4359-b1e3-6b94a91d6acb",
    "principalName": "DataMan@adhoekstra.ddns.net",
    "principalType": "User",
    "resourceGroup": "throw-away",
    "roleDefinitionId": "/subscriptions/b069085a-22df-4c6f-ab7e-f4f6f1d52d48/providers/Microsoft.Authorization/roleDefinitions/17d1049b-9a84-46fb-8f53-869881c3d3ab",
    "roleDefinitionName": "Storage Account Contributor",
    "scope": "/subscriptions/b069085a-22df-4c6f-ab7e-f4f6f1d52d48/resourceGroups/throw-away/providers/Microsoft.Storage/storageAccounts/koffiestorage",
    "type": "Microsoft.Authorization/roleAssignments"
  }
]

[...]

Issue #10815, opened 2019-10-11
Firewalls and virtual network blob on a storage account show authentication error rather than firewall error


lufussel:
## Describe the bug

**Command Name**
`az storage container create`

**Errors:**
```
The command failed with an unexpected error. Here is the traceback:


'CommandResultItem' object is not iterable
Traceback (most recent call last):
  File "/opt/az/lib/python3.6/site-packages/knack/cli.py", line 212, in invoke
    self.output.out(cmd_result, formatter=formatter, out_file=out_file)
  File "/opt/az/lib/python3.6/site-packages/knack/output.py", line 132, in out
    output = formatter(obj)
  File "/opt/az/lib/python3.6/site-packages/knack/output.py", line 38, in format_json
    input_dict = dict(result) if hasattr(result, '__dict__') else result
TypeError: 'CommandResultItem' object is not iterable
```

## To Reproduce:
Steps to reproduce the behavior. Note that argument values have been redacted, as they may contain sensitive information.

- _Put any pre-requisite steps here..._
- `az storage container create --name {} --account-name {}`

## Environment Summary
```
Linux-4.15.0-1059-azure-x86_64-with-debian-stretch-sid
Python 3.6.5
Shell: bash

azure-cli 2.0.74

Extensions:
aks-preview 0.4.5
storage-preview 0.2.8

```

lufussel:
"Storage Blob Data Contributor" role is assigned to the user. Containers create without firewall rules, but are blocked when locked by firewall. Error message is not accurate.

Juliehzl:
Hi @lufussel, thanks for your feedback. What you said is really right. I think you are unblocked now with root cause found, right? We will improve error message for firewall scenario.

Juliehzl:
Hi @lufussel, given there is no response for a long time, I will close the issue and error message should already be update. If there is any issue, feel free to create a new one.

Issue #10436, opened 2019-09-05
az storage blob list not working properly within Azure environment


kloeffler:
## Describe the bug

**Command Name**
`az storage blob list`

**Errors:**
```
The command failed with an unexpected error. Here is the traceback:


'CommandResultItem' object is not iterable
Traceback (most recent call last):
  File "/usr/lib64/az/lib/python2.7/site-packages/knack/cli.py", line 212, in invoke
    self.output.out(cmd_result, formatter=formatter, out_file=out_file)
  File "/usr/lib64/az/lib/python2.7/site-packages/knack/output.py", line 132, in out
    output = formatter(obj)
  File "/usr/lib64/az/lib/python2.7/site-packages/knack/output.py", line 38, in format_json
    input_dict = dict(result) if hasattr(result, '__dict__') else result
TypeError: 'CommandResultItem' object is not iterable
```

This command works fine from VMs outside of the Azure environment, but not within. 
Roles are set properly, the same command will produce valid output.

## To Reproduce:
Steps to reproduce the behavior. Note that argument values have been redacted, as they may contain sensitive information.

- `az login --service-principal -u{} -p{}`
- `az storage blob list --container-name {} --auth-mode {} --account-name {}`

## Expected Behavior
Output of blob list, e.g.:
[]

## Environment Summary
```
Linux-3.10.0-957.27.2.el7.x86_64-x86_64-with-centos-7.6.1810-Core
Python 2.7.5
Shell: bash

azure-cli 2.0.71 *

```

yonzhan:
@Juliehzl  please take a look

Juliehzl:
@kloeffler can you show me more information using `az storage blob list --debug` ? P.S. Remember not post your account key information in the Internet.

kloeffler:
``
Command arguments: ['storage', 'blob', 'list', '--debug']
Event: Cli.PreExecute []
Event: CommandParser.OnGlobalArgumentsCreate [<function on_global_arguments at 0x7f86f0ed6f50>, <function on_global_arguments at 0x7f86f0a58aa0>, <function on_global_arguments at 0x7f86f0a5cd70>]
Event: CommandInvoker.OnPreCommandTableCreate []
[...]

PR #11345, opened 2019-11-21
[Stoarge] Fix issues with incorrect error message "You do not have the required permissions needed to perform this operation."

Juliehzl:
Fix issue #10447, #10126, #9995, #10815, #10436, #10826,
---
Resolve comments in PR #11303 


src/azure-cli/azure/cli/command_modules/storage/__init__.py
```diff
@@ -171,17 +171,28 @@ def handler(ex):
                               ResourceType.DATA_STORAGE,
                               'common._error#AzureHttpError')
             if isinstance(ex, t_error) and ex.status_code == 403:
-                message = """
+                if ex.error_code == 'AuthorizationPermissionMismatch':
```
jsntcy: What's the message if ex.status_code == 403 and ex.error_code is not in [AuthorizationPermissionMismatch, AuthorizationFailure, AuthenticationFailed]?

src/azure-cli/azure/cli/command_modules/storage/__init__.py
```diff
@@ -171,17 +171,28 @@ def handler(ex):
...
+                    message = """
+The request may be blocked by network rules of storage account. Please check network rule set using 'az storage account show -n accountname --query networkRuleSet'.
+If you want to change default action to apply when no rule matches, please use 'az storage account update'.
+                    """
+                    ex.args = (message,)
+                elif ex.error_code == 'AuthenticationFailed':
+                    message = """
+Failed to authenticate the request. Make sure account key, connection string or sas token value is valid for your storage account.
+                    """
+                    ex.args = (message,)
```
jsntcy: Only need write "ex.args = (message,)" once here instead of write it in each "if-elif-elif" as they are the same logic with different messages.

src/azure-cli/azure/cli/command_modules/storage/__init__.py
Juliehzl: Throw exception with raw message

src/azure-cli/azure/cli/command_modules/storage/__init__.py
```diff
@@ -171,17 +171,28 @@ def handler(ex):
...
+"Storage Blob Data Reader"
+"Storage Queue Data Contributor"
+"Storage Queue Data Reader"
+If you want to use the old authentication method and allow querying for the right account key, please use the "--auth-mode" parameter and "key" value.
+                    """
+                    ex.args = (message,)
+                elif ex.error_code == 'AuthorizationFailure':
+                    message = """
+The request may be blocked by network rules of storage account. Please check network rule set using 'az storage account show -n accountname --query networkRuleSet'.
+If you want to change default action to apply when no rule matches, please use 'az storage account update'.
```
MyronFanQiu: default action which is applied? I think it should be ` change default action to something`

src/azure-cli/azure/cli/command_modules/storage/__init__.py
```diff
@@ -171,17 +171,28 @@ def handler(ex):
...
+                    """
+                    ex.args = (message,)
+                elif ex.error_code == 'AuthenticationFailed':
+                    message = """
+Failed to authenticate the request. Make sure account key, connection string or sas token value is valid for your storage account.
```
MyronFanQiu: `either account key, connection string or sas token` may be better

src/azure-cli/azure/cli/command_modules/storage/__init__.py
```diff
@@ -171,17 +171,28 @@ def handler(ex):
...
-    "Storage Queue Data Contributor"
-    "Storage Queue Data Reader"
-
-If you want to use the old authentication method and allow querying for the right account key, please use the "--auth-mode" parameter and "key" value. ErrorCode: AuthenticationFailed
-                """
-                ex.args = (message,)
+"Storage Blob Data Contributor"
+"Storage Blob Data Reader"
+"Storage Queue Data Contributor"
+"Storage Queue Data Reader"
```
jiasli: Indents are removed by purpose?

src/azure-cli/azure/cli/command_modules/storage/__init__.py
jiasli: **the** default action

src/azure-cli/azure/cli/command_modules/storage/__init__.py
```diff
@@ -171,17 +171,28 @@ def handler(ex):
...
+If you want to use the old authentication method and allow querying for the right account key, please use the "--auth-mode" parameter and "key" value.
+                    """
+                    ex.args = (message,)
+                elif ex.error_code == 'AuthorizationFailure':
+                    message = """
+The request may be blocked by network rules of storage account. Please check network rule set using 'az storage account show -n accountname --query networkRuleSet'.
```
jiasli: Are network rules the only cause for `AuthorizationFailure`? If not, how about use 

> Authorization failure. This may be caused by ...

Something similar to https://github.com/Azure/azure-cli/blob/ffebdf45c957483745431ce0e2ed6d85242266af/src/azure-cli-core/azure/cli/core/util.py#L26-L30

src/azure-cli/azure/cli/command_modules/storage/__init__.py
Juliehzl: Fixed.

src/azure-cli/azure/cli/command_modules/storage/__init__.py
Juliehzl: Fixed

src/azure-cli/azure/cli/command_modules/storage/__init__.py
jiasli: Indents are removed on purpose?

src/azure-cli/azure/cli/command_modules/storage/__init__.py
Juliehzl: Fixed

src/azure-cli/azure/cli/command_modules/storage/__init__.py
Juliehzl: Fixed

Issue #11127, opened 2019-11-06
Do we need to relax/fasten version of package six?


haroldrandom:
**Is your feature request related to a problem? Please describe.**
PR #11125 temporarily lcok the version of package six for [upgrade of six from 1.12.0 to 1.13.0](https://pypi.org/project/six/#history) breaks dependency chain which failed Travis CI.

_Should we relax/fasten version of package six to enjon the latest benefits of upgrading?_

**Describe the solution you'd like**
Change the syntax format of package six `~1.12.0`. Need team discussion and many tests.

**Describe alternatives you've considered**
None

**Additional context**
None

haroldrandom:
Astroid issue: [PyCQA/astroid #718](https://github.com/PyCQA/astroid/issues/718)

haroldrandom:
astroid has upgraded six to 1.13.0

Issue #9781, opened 2019-06-24
Use requirements.txt to lock dependencies


marstr:
*More details to be editted in*

marstr:
#9785 takes care of the first part of this effort, providing the concrete requirements. The second part will be adjusting our installers/test infrastructure to use the appropriate requirements.*.txt file.

yonzhan:
add to S164.

fengzhou-msft:
RPM, Deb and Docker already use requirements.*.txt file now. MSI and Homebrew still use setup.py for dependencies. For pip, we can use `pip install -r requirements.txt`. 

I will evaluate the work to change MSI and homebrew.

fengzhou-msft:
add @haroldrandom to check CI and tests.

fengzhou-msft:
azdev setup, CI and packaging except pypi are using requirements.txt to lock dependencies now. This issue can be closed. 

Documentation on pip install with requirements.txt or modify the linux install script will be another discussion.

PR #11376, opened 2019-11-25
Remove six~=1.12.0 restriction in Travis CI since astroid's bug has been fixed

haroldrandom:
- Fix https://github.com/Azure/azure-cli/issues/11127
- One of our dependencies astroid's bug that block CI has been fixed [Update six version](https://github.com/PyCQA/astroid/issues/718), remove unnecessary restriction

---

### DIFF
--- a/src/azure-cli-core/setup.py
+++ b/src/azure-cli-core/setup.py
@@ -68,7 +68,7 @@ DEPENDENCIES = [
     'pyopenssl>=17.1.0',  # https://github.com/pyca/pyopenssl/pull/612
     'pyyaml',
     'requests~=2.20',
-    'six~=1.12',
+    'six',
     'wheel==0.30.0',
     'azure-mgmt-resource~=4.0',
 ]
--- a/src/azure-cli-core/setup.py
+++ b/src/azure-cli-core/setup.py
@@ -68,7 +68,7 @@ DEPENDENCIES = [
     'pyopenssl>=17.1.0',  # https://github.com/pyca/pyopenssl/pull/612
     'pyyaml',
     'requests~=2.20',
-    'six',
+    'six~=1.12',
     'wheel==0.30.0',
     'azure-mgmt-resource~=4.0',
 ]
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ addons:
       - libssl-dev
       - libffi-dev
       - python-dev
-install: pip install --upgrade pip tox six~=1.12.0
+install: pip install --upgrade pip tox six
 cache: pip
 jobs:
   include:


src/azure-cli-core/setup.py
```diff
@@ -68,7 +68,7 @@
     'pyopenssl>=17.1.0',  # https://github.com/pyca/pyopenssl/pull/612
     'pyyaml',
     'requests~=2.20',
-    'six~=1.12',
```
MyronFanQiu: `six` instead of `six~=1.13`?

haroldrandom: No need, the default behavior for `python setup.py` is to look up to the latest version

fengzhou-msft: I also recommand `six~=1.13` for compatibility. What if six upgrades to 2.0 and has breaking changes?

haroldrandom: I am not sure about it.

In \*.requirements.\*.txt, six has been already locked to ==1.12.0, but in other places, there is no such restriction.

haroldrandom: How about not upgrading to 1.13.0 ?
Since we have decided to lock down 1.12.0 in \*.requirements.\*.txt.

fengzhou-msft: requirements is for stable developing, it should be as strict as possible. setup is for install, it should be flexible as long as it does not break.  `six~=1.12` and `six~=1.13` has the same effect, it will install the latest compatible `1.x` version.

fengzhou-msft: Theoretically, `six~=1.12` and `six~=1.13` will both install the latest compatible `1.x` verison. https://stackoverflow.com/questions/39590187/in-requirements-txt-what-does-tilde-equals-mean. Need to figure oput why six version is locked to 1.12 if we put `six~=1.12` there.

jiasli: @fengzhou-msft, is the pipeline installing via requirements.txt? If so, setup.py should be as wide as possible. https://packaging.python.org/discussions/install-requires-vs-requirements/

haroldrandom: Also according to the [acticle ](https://caremad.io/posts/2013/07/setup-vs-requirement/)that yugang shared:
`setup.py` is used to distribution, `requirements.txt` is used to development, test and deployment.

`setup.py` is used where `pip install` by default, but if user wants to specify certain version on dev/test/prod env to get stable or fetch packages from private repo or dependency differentiate from different OS/different test/dev/prod env, it's what `requirements.txt` targets to.

In summary, the package version in `requirements.txt` should be greater or equal to that in `setup.py`. 
**In this PR, I think we should make `six~=1.12.0` in src/azure-cli-core/setup.py and `pip install --upgrade pip tox six` in .travis.yml.** 

Back to our case:
1. The first issue mentioned the purpose of differentiating envs is https://github.com/Azure/azure-cli/issues/9781.  he only addressed a part ot it.
2. The first PR I found about initiazling this requirements.txt ( https://github.com/Azure/azure-cli/pull/9785 ) just to _log_ all these dependencies into files such that he could run `pip check`. But I don't know why he can not check against what installed by `pip install` by using setup.py His purpose I don't fully understand.
3. The second PR about it is "[Add dep diagnostics](https://github.com/Azure/azure-cli/pull/9796)" which checks dependecies incompatibility among different requirements.txt as we saw today.
4. Back to now, those requiements.txt are used nowhere else ( azdev uses `pip install -e` instead )

So the next problem is about whether let those requirements.txt exists or not.

haroldrandom: ~=1.12.0 means >=1.12.0 and ==1.12.0, so only 1.12.0 will be installed.
~=1.12 means >=1.12 and ==1.* so 1.13 could an alternative.

PR #11395, opened 2019-11-26
improvement: Add label automatically when creating an issue based on issue template

jsntcy:
improvement: Add first level label automatically when creating an issue based on issue template

Following optimization:
Using a bot to extract issue title/description to add second level (service level) labels (such as network, storage, compute and etc.) automatically.

---

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug_report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_report.md
@@ -2,7 +2,7 @@
 name: "\U0001F41BBug report"
 about: Create a report to help us improve
 title: ''
-labels: ''
+labels: Bug
 assignees: ''
 
 ---
--- a/.github/ISSUE_TEMPLATE/Bug_report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_report.md
@@ -2,7 +2,7 @@
 name: "\U0001F41BBug report"
 about: Create a report to help us improve
 title: ''
-labels: Bug
+labels: ''
 assignees: ''
 
 ---
--- a/.github/ISSUE_TEMPLATE/Service_team_request.md
+++ b/.github/ISSUE_TEMPLATE/Service_team_request.md
@@ -3,7 +3,7 @@ name: "\U0001F6A7Service Team Support Request"
 about: For select service teams to request CLI-team support for work such as new features,
   commands, etc.
 title: ''
-labels: ''
+labels: Service Team Support Request
 assignees: ''
 
 ---
--- a/.github/ISSUE_TEMPLATE/Feature_request.md
+++ b/.github/ISSUE_TEMPLATE/Feature_request.md
@@ -2,7 +2,7 @@
 name: "\U0001F680Feature request"
 about: Suggest an idea for this project
 title: ''
-labels: ''
+labels: Feature Request
 assignees: ''
 
 ---


.github/ISSUE_TEMPLATE/Bug_report.md
```diff
@@ -2,7 +2,7 @@
 name: "\U0001F41BBug report"
 about: Create a report to help us improve
 title: ''
-labels: ''
+labels: Bug
```
fengzhou-msft: I suggest we do not label Bug automatically. Bugs have different SLA in terms of how soon it should be solved. We could have some numbers of false reports. We’d better manually label a bug after human investigation.

jsntcy: @fengzhou-msft, do we have statistics about false report? If the percentage of false report is too high, we should find a way to resolve false report issue. Anyway, automation labeling is our goal.

fengzhou-msft: I don't think we have such statistics. @Juliehzl @MyronFanQiu @jiasli may have better sense about how often that happens since they've been in the team a little bit longer. 

One reason for false reports is that in the new issue [page](https://github.com/Azure/azure-cli/issues/new/choose), the option to `Open a blank issue` is so small comapred to the above three options. Users may not see it and would mostly select `Bug report` for cases like a question. 

Another frequent false report case is that it's just a usage issue such as passing wrong parameter, not a real bug.

B.T.W. we have issues coming from documentation, we may want to label those automatically as well.

jsntcy:
request review for @jsntcy

.github/ISSUE_TEMPLATE/Bug_report.md
yonzhan: Bug label has two months resolve SLA which is one month shorter. I think we can create another label or just do not add Bug label.

jsntcy: @fengzhou-msft , @yonzhan , I removed auto-labeling for bug template, please help review again.

.github/ISSUE_TEMPLATE/Service_team_request.md
```diff
@@ -3,7 +3,7 @@ name: "\U0001F6A7Service Team Support Request"
 about: For select service teams to request CLI-team support for work such as new features,
   commands, etc.
 title: ''
-labels: ''
+labels: Service Team Support Request
```
MyronFanQiu: Personally, I will just use `Feature Request` for two issue templates. But I'm ok with this one too.

jsntcy:
Yes, I'll send notification to team about triage process update after PR merge.
The labels added automatically in this PR is only first level labels ("Bug" is not supported in this PR), and we still need to triage to add other labels such as service levels' labels (storage, network, and etc), "service attention".


---
In reply to: [324987958](https://github.com/Azure/azure-cli/pull/11395#issuecomment-324987958) [](ancestors = 324987958)

.github/ISSUE_TEMPLATE/Service_team_request.md
yungezz: one step further, what's the action for "service team support request" issues? should we az cli team take action, or service team take action? If you really want to add this label, I would suggest to separate feature request to az cli team, and to service team (for their owned cli/extension).

jsntcy: @yungezz currrently 

- This PR only adds first level labels (Feature Request for CLI self and Feature Request to support new module/command) automatically when creating an issue based on issue template. And the following optimization is using a bot to extract issue title/description to add second level (service level) labels (such as network, storage, compute and etc.) automatically.
- Service Team Support Request means service teams to request CLI-team support for work such as new features, commands, etc. Usually CLI team need take action: add new features/commands on behalf of service team if CLI team owns the service, or provide some guidance for service team if service team need adds features/commands themselves if CLI team doesn't own the service.

qwordy:
"Service Team Support Request" is also a "Feature Request". Maybe the naming of labels need to be discussed further.

PR #11409, opened 2019-11-27
[ARM] Support Cross Tenants in deployment

### LINKED ISSUE #11151: Support Cross-Tenant for az group deployment (arm template deployments) by taking auxilliary-tenant-id in input
**Resource Provider** 
Microsoft.Compute, ARM

Today, with SIG we support cross tenant vm deployment, i.e. the share image gallery (sig) is in different tenant than the vm being deployed to. This is only supported in az vm create. The way it works is cli takes a token from sig's tenant and also from vm's tenant (target) and passes both of them to ARM.
But this confuses the customers today, as they expect it to work from az group deployment (template deployments) also, but it doesnt work. It might be difficult to parse template, so that's why I request cli team supports an auxilliary tenant id in input to the az group deployment, and passes along the token to arm along with template deployment create request.

Not related

**Swagger Link**
N/A

**Target Date**
It is causing customer pain, and confusion. It will also unblock a major scneario in azure, and the work might turn out to be small. If it is a low hanging fruit, we will really appreciate your help to ease the customer pain.

Juliehzl:
Fix #11151
---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/_profile.py
+++ b/src/azure-cli-core/azure/cli/core/_profile.py
@@ -521,7 +521,7 @@ class Profile(object):
                 return parts[0], (None if len(parts) <= 1 else parts[1])
         return None, None
 
-    def get_login_credentials(self, resource=None, subscription_id=None, aux_subscriptions=None):
+    def get_login_credentials(self, resource=None, subscription_id=None, aux_subscriptions=None, aux_tenants=None):
         account = self.get_subscription(subscription_id)
         user_type = account[_USER_ENTITY][_USER_TYPE]
         username_or_sp_id = account[_USER_ENTITY][_USER_NAME]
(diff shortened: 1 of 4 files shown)


Juliehzl:
Please do not merge before it is ready

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -521,15 +521,15 @@ def _try_parse_msi_account_name(account):
                 return parts[0], (None if len(parts) <= 1 else parts[1])
         return None, None
 
-    def get_login_credentials(self, resource=None, subscription_id=None, aux_subscriptions=None):
+    def get_login_credentials(self, resource=None, subscription_id=None, aux_subscriptions=None, aux_tenants=None):
```
yungezz: just to confirm, this change will not only all cli modules right?  need change log

Juliehzl:
close it and send a new one in https://github.com/Azure/azure-cli/pull/11523

Issue #10869, opened 2019-10-16
Ability to associate application-gateway rewrite-rule sets to request routing rules


nesteffe:
I am unable to find a way to associate a rewrite rule set to a request routing rule using the CLI. I am not sure if I am just missing this or if it was an oversight when the rewrite-rule commands were added.

I believe the rule create & update commands should have a rewrite-rule-set-name switch or something similar. For example:

`az network application-gateway rule update -g MyResourceGroup --gateway-name MyAppGateway -n MyRule --rewrite-rule-set-name MyRewriteRuleSet`

I read in another issue about the --add and --set switches for the update commands that appear to allow setting arbitrary properties on the underlying objects but I can't seem to figure out how to make them work, or if this is even an appropriate use case for these methods.

In powershell the New/Set-AzApplicationGatewayRequestRoutingRule commands have a -RewriteRuleSet switch to make this association, but it doesn't appear this is surfaced in the CLI.

PaulCharlton:
+1

MyronFanQiu:
@nesteffe Hello. Is this issue similar with this one? https://github.com/Azure/azure-cli/issues/10854

nesteffe:
> @nesteffe Hello. Is this issue similar with this one? #10854

Yes, this is the same issue. Sorry, that issue didn't show up when I was searching for this earlier today.

MyronFanQiu:
@nesteffe oh. No problem. I just want to double check. We can leave this issue open. I'm quite busy right now for the Ignite. So I may need some time to take a look at this issue later. I'm sorry for it.

HeShaobo1:
hi all,

you can use Azure PowerShell by solving this probem. 

$requestHeaderConfiguration = New-AzApplicationGatewayRewriteRuleHeaderConfiguration -HeaderName "Strict-Transport-Security" -HeaderValue "max-age=3600"
$actionset = New-AzApplicationGatewayRewriteRuleActionSet -RequestHeaderConfiguration $requestHeaderConfiguration
$rewriterule = New-AzApplicationGatewayRewriteRule -Name hsts -ActionSet $actionset             
$rewriteruleset = New-AzApplicationGatewayRewriteRuleSet -Name hsts1rewriteruleset -RewriteRule $rewriterule 

#获取当前的APPGW
$appgw = Get-AzApplicationGateway -Name appgwv2 -ResourceGroupName shawnbian 

#获取当前需要添加重写的规则rule
[...]

PR #11421, opened 2019-11-28
[Network]Fix 10854 associate rule set to request routing rules

### LINKED ISSUE #10854: [AppGw] Rewrite rule set can't be associated with Request routing rules

Support request for reference: https://azuresupportcenter.msftcloudes.com/caseoverview?srId=119100425000173

**Describe the bug**
Application Gateway RewriteRuleSet can not be referenced in requestRoutingRule.
Similar to the backendPool, Listener, and HttpSettings, users should be able to *update* a requestRoutingRule with a RewriteRuleSet. Same for the UrlPathMap and PathRules in UrlPathMap.

**To Reproduce**
For the command, 
az network application-gateway rule update,
there is no option called "--rewrite-rule-set"

For this command,
az network application-gateway url-path-map update,
there is no option to add rewrite rule set.

There is no command to update an existing Path Rules to add a rewrite rule set.

**Expected behavior**
Explained above

**Environment summary**
Azure Cloud Shell

MyronFanQiu:
Fix #10854 #10869

---


src/azure-cli/HISTORY.rst
```diff
@@ -24,6 +24,11 @@ Release History
 * GA Release Large File Shares property for storage account create and update command
 * GA Release User Delegation SAS token Support
 
+**Network**
+
+* az network application-gateway url-path-map create/update: bug fix for `--waf-policy`
+* az network application-gateway: support `--redirect-rule-set`
```
jsntcy: --rewrite-rule-set

src/azure-cli/azure/cli/command_modules/network/_params.py
```diff
@@ -292,22 +293,25 @@ def load_arguments(self, _):
...
-        c.argument('firewall_policy', min_api='2019-09-01', help='Name or ID of a Firewall Policy resource.')
+        c.argument('firewall_policy', min_api='2019-09-01', help='Name or ID of a Firewall Policy resource.', arg_group='First Rule')
+        c.argument('rewrite_rule_set', min_api='2019-04-01', help='The name or ID of the rewrite rule set. If not specified, the default for the map will be used.', arg_group='First Rule')
 
     with self.argument_context('network application-gateway url-path-map create') as c:
         c.argument('default_address_pool', help='The name or ID of the default backend address pool, if different from --address-pool.', completer=get_ag_subresource_completion_list('backend_address_pools'))
         c.argument('default_http_settings', help='The name or ID of the default HTTP settings, if different from --http-settings.', completer=get_ag_subresource_completion_list('backend_http_settings_collection'))
+        c.argument('default_rewrite_rule_set', min_api='2019-04-01', help='The name or ID of the default rewrite rule set, if different from --http-settings.')
```
jsntcy: rewrite_rule_set

src/azure-cli/azure/cli/command_modules/network/_params.py
```diff
@@ -292,22 +293,25 @@ def load_arguments(self, _):
         c.argument('paths', nargs='+', help='Space-separated list of paths to associate with the rule. Valid paths start and end with "/" (ex: "/bar/")', arg_group='First Rule')
         c.argument('address_pool', help='The name or ID of the backend address pool to use with the created rule.', completer=get_ag_subresource_completion_list('backend_address_pools'), arg_group='First Rule')
         c.argument('http_settings', help='The name or ID of the HTTP settings to use with the created rule.', completer=get_ag_subresource_completion_list('backend_http_settings_collection'), arg_group='First Rule')
-        c.argument('firewall_policy', min_api='2019-09-01', help='Name or ID of a Firewall Policy resource.')
+        c.argument('firewall_policy', min_api='2019-09-01', help='Name or ID of a Firewall Policy resource.', arg_group='First Rule')
+        c.argument('rewrite_rule_set', min_api='2019-04-01', help='The name or ID of the rewrite rule set. If not specified, the default for the map will be used.', arg_group='First Rule')
```
jsntcy: Do we need add completer=get_ag_subresource_completion_list() for "rewrite_rule_set"?

src/azure-cli/azure/cli/command_modules/network/_validators.py
```diff
@@ -770,6 +770,10 @@ def process_ag_rule_create_namespace(cmd, namespace):  # pylint: disable=unused-
         namespace.redirect_config = _generate_ag_subproperty_id(
             cmd.cli_ctx, namespace, 'redirectConfigurations', namespace.redirect_config)
 
+    if namespace.rewrite_rule_set and not is_valid_resource_id(namespace.rewrite_rule_set):
```
yungezz: this one don't need to setup default_rewrite_rule_set?

src/azure-cli/azure/cli/command_modules/network/custom.py
```diff
@@ -740,14 +740,16 @@ def create_ag_request_routing_rule(cmd, resource_group_name, application_gateway
         url_path_map=SubResource(id=url_path_map) if url_path_map else None)
     if cmd.supported_api_version(min_api='2017-06-01'):
         new_rule.redirect_configuration = SubResource(id=redirect_config) if redirect_config else None
+    if cmd.supported_api_version(min_api='2019-04-01'):
```
yungezz: what's practice to set min_api version? I thought it's for azure stack. is it possible to specify version at global level, instead of in business logic code? it's hard for maintenance.

src/azure-cli/azure/cli/command_modules/network/custom.py
```diff
@@ -868,6 +872,14 @@ def create_ag_url_path_map(cmd, resource_group_name, application_gateway_name, i
         new_map.default_redirect_configuration = \
             SubResource(id=default_redirect_config) if default_redirect_config else None
 
+    if cmd.supported_api_version(min_api='2019-04-01'):
```
yungezz: again, it looks really bad for adding hardcode versions in different piece of codes but same purpose. at least define something like global var rewrite_rule_set_min_version

MyronFanQiu: A possible reason behind hardcode might be if we need to define a global var, the name might be `ag_url_path_map_rewrite_rule_set_min_version`. This api-version is limited in these two commands `az network application-gateway url-path-map create/update`

MyronFanQiu: @yungezz Currently, we set `min_api=2019-04-01` in _param.py too. I just checked the structure in `cmd` and we can also get this information. How about we support something like `cmd.support_parameter(rewrite_rule_set)` so that we can remove this hardcode version and keep the version information limited in _param.py

MyronFanQiu: It can also work for multi api if I write the check like `if rewrite_rule_set is None`

yungezz: thanks @MyronFanQiu . sounds good to add in cmd. Image one day, you'll need update versoin for a feature like rewrite_rule_set, how it's possible to find out all instance of hardcode version. meanwhile, bad readability.
Eventually, I would like we have this practice tested/verified, and make it into cli command guideline.

MyronFanQiu:
@yungezz Hello. I included two kinds of solution in this PR. After re-thinking the structure, I think `_param.py` is a quite good place to store the min_api version for each parameter. We can discuss the solution again.

src/azure-cli-core/azure/cli/core/commands/__init__.py
```diff
@@ -313,6 +313,13 @@ def get_api_version(self, resource_type=None, operation_group=None):
         resource_type = resource_type or self.command_kwargs.get('resource_type', None)
         return self.loader.get_api_version(resource_type=resource_type, operation_group=operation_group)
 
+    def get_param_min_api_version(self, parameter, parent_locals):
```
yungezz: just a wild thinking, do you think it make sense that cli might be asked to have multiple min_api_version, such as one for azure stack, one for JEDI, one for moon_cake?

src/azure-cli/azure/cli/command_modules/network/_params.py
```diff
@@ -40,6 +40,7 @@
     ag_url_map_rule_completion_list, tm_endpoint_completion_list, service_endpoint_completer,
     get_sdk_completer)
 from azure.cli.core.util import get_json_object
+from ._settings import application_gateway_url_path_map_rewrite_rule_set
```
yungezz: zan!

src/azure-cli-core/azure/cli/core/commands/__init__.py
MyronFanQiu: Currently, min_api is based on the first time this param appears in python SDK/swagger spec. Unless we need to choose different package version(not api-version) of sdk for different cloud, I cannot image we need to support multiple min_api_version.

jiasli: Since min API version is constant (will not change in the future), I prefer hard-code it to eliminate the over-complexity.

src/azure-cli-core/azure/cli/core/commands/__init__.py
```diff
@@ -313,7 +313,12 @@ def get_api_version(self, resource_type=None, operation_group=None):
         resource_type = resource_type or self.command_kwargs.get('resource_type', None)
         return self.loader.get_api_version(resource_type=resource_type, operation_group=operation_group)
 
-    def supported_api_version(self, resource_type=None, min_api=None, max_api=None, operation_group=None):
+    def supported_api_version(self, resource_type=None, min_api=None, max_api=None,
+                              operation_group=None, parameter_name=None):
```
jiasli: I don't think it is intuitive to insert `parameter_name` to `supported_api_version`. How about creating another function `supported_parameter` and call `supported_api_version` internally?

src/azure-cli-core/azure/cli/core/commands/__init__.py
yungezz: it's not about one api version, pls look at the whole file, or cli codes in general, lots of api version, even in one rp, one resource, min_api_version is on property level. This introduces bad readability and heavy maintenance effort. purpose of the comments is general improving code readability, maintenance. while for how to generalize the way, we can discuss. that's the context why @MyronFanQiu  provided 2 options open for discussion. If more concern, let's discuss offline.

PR #11451, opened 2019-12-02
[TestSDK] Add case_sensitive option for checks

qwordy:
Azure doesn't distinguish between upper case and lower case names. Current "equals" checks in testing framework is case_sensitive, which blocks the testing in some circumstances. In this PR, I add a case_sensitive option.

---

### DIFF
--- a/src/azure-cli-testsdk/azure/cli/testsdk/base.py
+++ b/src/azure-cli-testsdk/azure/cli/testsdk/base.py
@@ -158,7 +158,6 @@ class ScenarioTest(ReplayableTest, CheckerMixin, unittest.TestCase):
 
     def cmd(self, command, checks=None, expect_failure=False):
         command = self._apply_kwargs(command)
-        print(command)
         return execute(self.cli_ctx, command, expect_failure=expect_failure).assert_with_checks(checks)
 
     def get_subscription_id(self):
--- a/src/azure-cli-testsdk/azure/cli/testsdk/base.py
+++ b/src/azure-cli-testsdk/azure/cli/testsdk/base.py
@@ -43,11 +43,11 @@ class CheckerMixin(object):
             raise KeyError("Key '{}' not found in kwargs. Check spelling and ensure it has been registered."
                            .format(ex.args[0]))
 
-    def check(self, query, expected_results):
+    def check(self, query, expected_results, case_sensitive=True):
         from azure.cli.testsdk.checkers import JMESPathCheck
         query = self._apply_kwargs(query)
         expected_results = self._apply_kwargs(expected_results)
-        return JMESPathCheck(query, expected_results)
+        return JMESPathCheck(query, expected_results, case_sensitive)
 
     def exists(self, query):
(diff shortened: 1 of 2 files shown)


src/azure-cli-testsdk/azure/cli/testsdk/base.py
```diff
@@ -158,6 +158,7 @@ def create_guid(self):
 
     def cmd(self, command, checks=None, expect_failure=False):
         command = self._apply_kwargs(command)
+        print(command)
```
MyronFanQiu: should we support 2.7 currently?

src/azure-cli-testsdk/azure/cli/testsdk/checkers.py
```diff
@@ -23,7 +24,12 @@ def __call__(self, execution_result):
         except jmespath.exceptions.JMESPathTypeError:
             raise JMESPathCheckAssertionError(self._query, self._expected_result, actual_result,
                                               execution_result.output)
-        if actual_result != self._expected_result and str(actual_result) != str(self._expected_result):
+        if self._case_sensitive:
+            equals = actual_result == self._expected_result or str(actual_result) == str(self._expected_result)
+        else:
+            equals = isinstance(actual_result, str) and isinstance(self._expected_result, str)\
```
MyronFanQiu: what's the difference between `str(actual_result).lower() == str(self._expected_result).lower()` and your logic? if `self._case_sensitive==false`, `actual_result` and `self._expected_result` must be the type of string. Do I misunderstand?

src/azure-cli-testsdk/azure/cli/testsdk/base.py
qwordy: Oh. Sorry. It's just for debugging. I forgot resetting it.

src/azure-cli-testsdk/azure/cli/testsdk/checkers.py
qwordy: Good question. I have updated it.

src/azure-cli-testsdk/azure/cli/testsdk/checkers.py
```diff
@@ -23,7 +24,12 @@ def __call__(self, execution_result):
         except jmespath.exceptions.JMESPathTypeError:
             raise JMESPathCheckAssertionError(self._query, self._expected_result, actual_result,
                                               execution_result.output)
-        if actual_result != self._expected_result and str(actual_result) != str(self._expected_result):
+        if self._case_sensitive:
+            equals = actual_result == self._expected_result or str(actual_result) == str(self._expected_result)
+        else:
+            equals = actual_result == self._expected_result \
```
jiasli: I `lower()` is also needed here.

qwordy: The type of actual_result and self._expected_result may not be string. If actual_result == self._expected_result, the check should always pass no matter case_sensitive is true or false.

PR #11466, opened 2019-12-03
Fix #11458: az ad sp credential reset doesn't reset service principal credential, resets app credential

### LINKED ISSUE #11458: az ad sp credential reset doesn't reset service principal credential, resets app credential

**Describe the bug**
We have a multi-tenant app registration defined in Azure. We create a service principal in a *separate tenant* for that app, but when we try to provide a credential using the `az ad sp credential reset` command it tries to look up the app resource and can't find it in the tenant (because it isn't there). Why doesn't it look up the service principal resource like the corresponding [powershell command?](https://docs.microsoft.com/en-us/powershell/module/az.resources/new-azadspcredential?view=azps-3.1.0)

**To Reproduce**
```
> az ad sp create --id <app_id>
> az ad sp credential reset -n <app_id> --append
Resource '<app_id>' does not exist or one of its queried reference-property objects are not present.
```

**Expected behavior**
Similar behavior to the powershell command provided, the service principal should receive a new credential, which will be returned by the command, or provided by the user using the --password parameter.

**Environment summary**
cmd.exe on windows 
```
> az --version
azure-cli                         2.0.75 *

command-modules-nspkg              2.0.3
core                              2.0.75 *
nspkg                              3.0.4
telemetry                          1.0.4
```

**Additional context**
The documentation clearly states that this command operates the same as az ad app credential, but the reset command says it resets the service principal's credential. Is there actually a way to reset the service principal's credentials using azure cli?

mikeliddle:
Fixes for issue #11458

Changes the behavior of `az ad sp credential` to act on the service principal resource instead of the application resource. 

---

### DIFF
--- a/CredScanSuppressions.json
+++ b/CredScanSuppressions.json
@@ -1,0 +1,306 @@
+{
+  "tool": "Credential Scanner",
+  "suppressions": [
+    {
+      "placeholder": "r7rpDQLwRhJQF/WoSPp6sMKJB9GQJCsZ9/o8xkqP/xM=",
+      "_justification": "secret filed used among recoding yaml file of aks"
+    },
+    {
+      "placeholder": "1w4F14PX8TKt7rMaG1bymhJlfjab4cocxoL7edC9zJI=",
+      "_justification": "secret filed used among recoding yaml file of acr"
+    },
+    {
+      "file": [
+        "src\\azure-cli\\azure\\cli\\command_modules\\appconfig\\tests\\latest\\recordings\\test_azconfig_credential.yaml",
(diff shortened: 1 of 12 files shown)


jiasli:
If we decide to seperate sp and app credential commands. This line in `reset_application_credential` doesn't seem to be valid anymore:

https://github.com/Azure/azure-cli/blob/5baced969bb96d4138d8886661f302f142942255/src/azure-cli/azure/cli/command_modules/role/custom.py#L1683

We can simply replace it with `_resolve_application`.

yonzhan:
ad sp

jiasli:
Tracked by https://github.com/Azure/azure-cli/issues/11458

PR #11467, opened 2019-12-03
[AppConfig] Support Import/Export of feature flags

avanigupta:
---


avanigupta:
Please add @recao for review

src/azure-cli/azure/cli/command_modules/appconfig/_validators.py
```diff
@@ -76,7 +76,7 @@ def validate_export(namespace):
         if (namespace.dest_name is None) and (namespace.dest_connection_string is None):
             raise CLIError("usage error: --config-name NAME | --connection-string STR")
     elif destination == 'appservice':
-        if namespace.appservice_name is None:
+        if namespace.appservice_account is None:
```
Juliehzl: What is the difference between appservice_name and appservice_account?

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -48,14 +49,35 @@ def import_config(cmd,
...
             file_path=path, format_=format_, separator=separator, prefix_to_add=prefix, depth=depth)
+
+        if not skip_features:
+            # src_features is a list of KeyValue objects
+            src_features = __read_features_from_file(file_path=path, format_=format_)
+
     elif source == 'appconfig':
-        src_kvs = __read_kv_from_config_store(
-            cmd, name=src_name, connection_string=src_connection_string, key=src_key, label=src_label, prefix_to_add=prefix)
+        src_kvs = __read_kv_from_config_store(cmd, name=src_name, connection_string=src_connection_string,
```
Juliehzl: What will happened when the label is null?

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -74,48 +75,50 @@ def load_arguments(self, _):
         c.argument('id_', options_list=['--id'], help='Id of the key to be regenerated. Can be found using az appconfig credential list command.')
 
     with self.argument_context('appconfig kv import') as c:
-        c.argument('label', help="Imported KVs will be assigned with this label. If no label specified, will assign null label.")
-        c.argument('prefix', help="This prefix will be appended to the front of imported keys.")
-        c.argument('source', options_list=['--source', '-s'], choices=['file', 'appconfig', 'appservice'], validator=validate_import, help="The source of importing.")
+        c.argument('label', help="Imported KVs and feature flags will be assigned with this label. If no label specified, will assign null label.")
```
Juliehzl: How can I retrieve all the feature flags with null label?

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -74,48 +75,50 @@ def load_arguments(self, _):
...
+        c.argument('prefix', help="This prefix will be appended to the front of imported keys. Not applicable for feature flags.")
+        c.argument('source', options_list=['--source', '-s'], choices=['file', 'appconfig', 'appservice'], validator=validate_import, help="The source of importing. Note that importing feature flags from appservice is not supported.")
```
Juliehzl: Here `choices` is not recommended that you use this because the choice lists will be case sensitive. Instead see ["Registering Enums"](https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#registering-enums).

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -74,48 +75,50 @@ def load_arguments(self, _):
...
+        c.argument('separator', validator=validate_separator, help="Delimiter for flattening the json or yaml file to key-value pairs. Required for importing hierarchical structure. Not applicable for property files or feature flags. Supported values: '.', ',', ';', '-', '_', '__', '/', ':' ")
 
     with self.argument_context('appconfig kv import', arg_group='AppConfig') as c:
         c.argument('src_name', help='The name of the source App Configuration.')
-        c.argument('src_connection_string', validator=validate_connection_string, help="Connections of access key and endpoint of the source store. ")
-        c.argument('src_key', help='If no key specified, import all keys by default. Support star sign as filters, for instance abc* means keys with abc as prefix. Similarly, *abc and *abc* are also supported.')
+        c.argument('src_connection_string', validator=validate_connection_string, help="Combination of access key and endpoint of the source store. ")
```
Juliehzl: Is there a way to generate connection string quickly?

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -74,48 +75,50 @@ def load_arguments(self, _):
...
 
     with self.argument_context('appconfig kv export') as c:
-        c.argument('label', help="Only keys with this label will be exported. If no label specified, export keys with null label by default.")
-        c.argument('prefix', help="Prefix to be trimed from keys.")
-        c.argument('key', help='If no key specified, return all keys by default. Support star sign as filters, for instance abc* means keys with abc as prefix. Similarly, *abc and *abc* are also supported.')
-        c.argument('destination', options_list=['--destination', '-d'], choices=['file', 'appconfig', 'appservice'], validator=validate_export, help="The destination of exporting.")
+        c.argument('label', help="Only keys and feature flags with this label will be exported. If no label specified, export keys and feature flags with null label by default.")
+        c.argument('prefix', help="Prefix to be trimmed from keys. Not applicable for feature flags.")
```
Juliehzl: There are many parameters which are not applicable for feature flags. Can you explain for me in brief? I am a little confused about the behavior when not applicable for feature flags.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -0,0 +1,720 @@
...
+        if kv.tags:
+            for tag_k, tag_v in kv.tags.items():
+                tag_json[tag_k] = tag_v
+        kv_json['tags'] = tag_json
+        res.append(kv_json)
+    return res
+
+
+def __print_features_preview(old_json, new_json):
+    error_print('\n---------------- Preview Feature Flags (Beta) ----------------')
```
Juliehzl: print is not recommended please refer to https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md#general-patterns.
Try to use [logger](https://github.com/microsoft/knack/blob/master/docs/logging.md#logging) in knack.

src/azure-cli/azure/cli/command_modules/appconfig/_validators.py
avanigupta: This was a bug. The argument we want to check for is --appservice-account. We don't have anything like appservice-name.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
avanigupta: For this arg_group, we need 2 connection strings to connect to two different AppConfig stores. While the first (default) connection string can be easily configured, the second is not that straight forward as the second AppConfig store may or may not be in the same subscription that the user is logged in to. Moreover, since we only use this connection string for import/export commands when  arg_group='AppConfig', implementing a solution to set this quickly will be an overkill. 

---
In reply to: [355225819](https://github.com/Azure/azure-cli/pull/11467#discussion_r355225819) [](ancestors = 355225819)

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
avanigupta: The idea is that specifying --prefix argument will not have any effect of feature flags. It will be ignored completely. Same for other parameters which are not applicable for feature flags.

---
In reply to: [355226236](https://github.com/Azure/azure-cli/pull/11467#discussion_r355226236) [](ancestors = 355226236)

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
avanigupta: While importing, we assign a user specified label to KV or feature flags. If no label specified, we will assign null label.

---
In reply to: [355225468](https://github.com/Azure/azure-cli/pull/11467#discussion_r355225468) [](ancestors = 355225468)

avanigupta:
Please add @shenmuxiaosen  for review.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
avanigupta: done

---
In reply to: [355225688](https://github.com/Azure/azure-cli/pull/11467#discussion_r355225688) [](ancestors = 355225688)

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: done

---
In reply to: [355226690](https://github.com/Azure/azure-cli/pull/11467#discussion_r355226690) [](ancestors = 355226690)

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -158,7 +161,7 @@ def load_arguments(self, _):
         c.argument('fields', arg_type=feature_fields_arg_type)
 
     with self.argument_context('appconfig feature set') as c:
-        c.argument('feature', help='Name of the feature flag to be set.')
+        c.argument('feature', help="Name of the feature flag to be set. Only alphanumeric characters, '.', '-' and '_' are allowed.")
         c.argument('label', help="If no label specified, set the feature flag with null label by default")
```
shenmuxiaosen: Should we add a validator for the feature name?

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -74,48 +75,50 @@ def load_arguments(self, _):
...
+        c.argument('depth', validator=validate_import_depth, help="Depth for flattening the json or yaml file to key-value pairs. Flatten to the deepest level by default. Not appicable for property files or feature flags.")
         # bypass cli allowed values limition
-        c.argument('separator', validator=validate_separator, help="Delimiter for flatterning the json or yaml file to key-value pairs. Required for importing hierarchical structure. Not applicable for property files. Supported values: '.', ',', ';', '-', '_', '__', '/', ':' ")
+        c.argument('separator', validator=validate_separator, help="Delimiter for flattening the json or yaml file to key-value pairs. Required for importing hierarchical structure. Not applicable for property files or feature flags. Supported values: '.', ',', ';', '-', '_', '__', '/', ':' ")
 
     with self.argument_context('appconfig kv export', arg_group='AppConfig') as c:
         c.argument('dest_name', help='The name of the destination App Configuration.')
-        c.argument('dest_connection_string', validator=validate_connection_string, help="Connections of access key and endpoint of the destination store. ")
```
shenmuxiaosen: >  [](start = 153, length = 1)

nit: remove the white space

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -48,14 +49,35 @@ def import_config(cmd,
...
+                                              key=src_key, label=src_label, prefix_to_add=prefix)
```
shenmuxiaosen: It looks a little inefficient to fetch all items, remove feature flags and then fetch features if asked again. Instead, we can return two items like src_kvs, src_features = __read_kv_from_config_store.

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -48,14 +49,35 @@ def import_config(cmd,
                   src_key=None,
                   src_label=None,
                   # from-appservice parameters
-                  appservice_account=None):
+                  appservice_account=None,
+                  skip_features=False):
+    src_features = []
+    dest_features = []
+    dest_kvs = []
+    need_feature_change = False
```
shenmuxiaosen: nit: no need to create them here.  You can put it after user confirmed preview

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -96,41 +136,95 @@ def export_config(cmd,
                   dest_connection_string=None,
                   dest_label=None,
                   # to-app-service parameters
-                  appservice_account=None):
+                  appservice_account=None,
+                  skip_features=False):
     # fetch key values from user's configstore
     src_kvs = __read_kv_from_config_store(
```
shenmuxiaosen: >__read_kv_from_config_store [](start = 14, length = 27)

same with comment above, let __read_kv_from_config_store to fetch all items and return key-values/features separately

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -96,41 +136,95 @@ def export_config(cmd,
...
+    src_features = []
+    dest_features = []
+    dest_kvs = []
+    need_feature_change = False
+
+    if skip_features:
+        if destination == 'appconfig':
+            # dest_kvs contains features and KV that match the label. We discard only features.
+            dest_kvs = __read_kv_from_config_store(
+                cmd, name=dest_name, connection_string=dest_connection_string, key=None, label=dest_label)
```
shenmuxiaosen: fetch items from destination for comparison under if user requested the preview. If we should check "yes" if it is false then do the rest.

shenmuxiaosen:
>                   source,

since we are not case sensitive for source any more (get_enum_type in _params.py) we should convert source to lower case to be used later on. Same with export destination. 

---
Refers to: src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py:35 in bccd646. [](commit_id = bccd64621144c3c99b04bf7b8055d454cb0a7167, deletion_comment = False)

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
avanigupta: As discussed offline, this approach would not work if user specifies key filter like abc*. We need to make two separate calls to get all the keys and features.

---
In reply to: [355747275](https://github.com/Azure/azure-cli/pull/11467#discussion_r355747275) [](ancestors = 355747275)

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
avanigupta: As discussed offline, this approach would not work if user specifies key filter like abc*. We need to make two separate calls to get all the keys and features.

---
In reply to: [355743324](https://github.com/Azure/azure-cli/pull/11467#discussion_r355743324) [](ancestors = 355743324)

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
avanigupta: As discussed offline, will have a separate PR to implement validation of keys and feature names.

---
In reply to: [355687066](https://github.com/Azure/azure-cli/pull/11467#discussion_r355687066) [](ancestors = 355687066)

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -96,41 +138,97 @@ def export_config(cmd,
...
+        if destination == 'appconfig':
+            # dest_kvs contains features and KV that match the label. We discard only features.
             dest_kvs = __read_kv_from_config_store(
                 cmd, name=dest_name, connection_string=dest_connection_string, key=None, label=dest_label)
+
+            __discard_features_from_retrieved_kv(dest_kvs)
+
         elif destination == 'appservice':
             dest_kvs = __read_kv_from_app_service(
                 cmd, appservice_account=appservice_account, prefix_to_add="")
```
shenmuxiaosen: >prefix_to_add="" [](start = 60, length = 16)

nit: __read_kv_from_app_service has a default value for prefix_to_add

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -96,41 +138,97 @@ def export_config(cmd,
...
-            dest_kvs = []  # treat as empty file and overwrite exported key values
-        elif destination == 'appconfig':
+                  appservice_account=None,
+                  skip_features=False):
+    src_features = []
+    dest_features = []
+    dest_kvs = []
+    destination = destination.lower()
+
+    if skip_features and not yes:
```
shenmuxiaosen: >if skip_features and not yes: [](start = 4, length = 29)

the logic here can be optimized to avoid double negative check making codes hard to read. 
it can be something like below

```
# first get everything we need from source App Configuration.

__read_kv_from_config_store   3# fetch all key values 
__discard_features_from_retrieved_kv

if not skip_features:
    # fetch all features

# second check if we need preview
if not yes:
    # fetch all key values excluding features from destinations, "AppConfig" or "AppService"
    if not skip_feature:
         # fetch all features from destination, "AppConfig"
    #print preview and ask for confirmation 
# third export to destination

```

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -0,0 +1,720 @@
...
+        with io.open(file_path, 'r', encoding=__check_file_encoding(file_path)) as config_file:
+            if format_ == 'json':
+                config_data = json.load(config_file)
+                if 'FeatureManagement' in config_data:
+                    del config_data['FeatureManagement']
+            elif format_ == 'yaml':
+                for yaml_data in list(yaml.safe_load_all(config_file)):
+                    config_data.update(yaml_data)
+                if 'feature-management' in config_data:
+                    del config_data['feature-management']
```
shenmuxiaosen: I think we only support json. For other types like yaml and properties, we will treat everything as normal key values.

shenmuxiaosen:
>                   format_=None,

convert to lower case for later comparison.

---
Refers to: src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py:43 in bf93ec8. [](commit_id = bf93ec86fc1808276284c79e65eeb9905ae4b2f7, deletion_comment = False)

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -0,0 +1,720 @@
...
+    # pylint: disable=too-many-nested-blocks
+    key_values = []
+    default_conditions = {'client_filters': []}
+
+    default_value = {
+        "id": "",
+        "description": "",
+        "enabled": False,
+        "conditions": default_conditions
+    }
```
shenmuxiaosen: why not use FeatureFlagValue object

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -0,0 +1,720 @@
...
+        if format_ == 'json':
+            for k, v in features_dict.items():
+                key = FEATURE_FLAG_PREFIX + str(k)
+                default_value["id"] = str(k)
+
+                if isinstance(v, dict):
+                    # This is a conditional feature
+                    default_value["enabled"] = True
+                    default_value["conditions"] = {'client_filters': v.get('EnabledFor', [])}
+
```
shenmuxiaosen: log when "EnabledFor" does not exist.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -0,0 +1,720 @@
...
+                        # "FeatureU": {"EnabledFor": []}
+                        default_value["enabled"] = False
+                        default_value["conditions"] = default_conditions
+
+                    else:
+                        # Convert Name and Parameters to lowercase for backend compatibility
+                        for idx, val in enumerate(default_value["conditions"]["client_filters"]):
+                            # each val is a dict with two keys - Name, Parameters
+                            val = {filter_key.lower(): filter_val for filter_key, filter_val in val.items()}
+                            if val["name"] and val["name"].lower() == "alwayson":
```
shenmuxiaosen: As the schema is fixed, no need to loop val, but explicitly get "Name" and "Parameters". Also we can decide should we fail or log warrning if the schema is not valid(such as name or parameters is missing)

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -0,0 +1,720 @@
...
+                        # We support alternate format for specifying always OFF features
+                        # "FeatureU": {"EnabledFor": []}
+                        default_value["enabled"] = False
+                        default_value["conditions"] = default_conditions
+
+                    else:
+                        # Convert Name and Parameters to lowercase for backend compatibility
+                        for idx, val in enumerate(default_value["conditions"]["client_filters"]):
+                            # each val is a dict with two keys - Name, Parameters
+                            val = {filter_key.lower(): filter_val for filter_key, filter_val in val.items()}
```
shenmuxiaosen: >lower [](start = 46, length = 5)

why alwasy change to lower case?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -0,0 +1,720 @@
...
+                            if val["name"] and val["name"].lower() == "alwayson":
+                                # We support alternate format for specifying always ON features
+                                # "FeatureT": {"EnabledFor": [{ "Name": "AlwaysOn"}]}
```
shenmuxiaosen: >AlwaysOn [](start = 73, length = 8)

Is it case sensitve here?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -0,0 +1,720 @@
...
+                                default_value["conditions"] = default_conditions
+                                break
```
shenmuxiaosen: So if there are multiple filters, one of them is "alwayson" then we ignore other filters?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -0,0 +1,720 @@
...
+                            default_value["conditions"]["client_filters"][idx] = val
```
shenmuxiaosen: >idx [](start = 74, length = 3)

We don't need to remember index here, as the for loop preserve the order.  since default_value["conditions"]["client_filters"] is a list, so we can just keep appending items

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -0,0 +1,720 @@
...
+
+
+def __export_features(retrieved_features, format_):
+    exported_dict = {}
+    client_filters = []
+    if format_ == 'json':
+        exported_dict["FeatureManagement"] = {}
+
+    elif format_ in ('yaml', 'properties'):
+        # Currently, this condition will never be met
```
shenmuxiaosen: add log warning for formats not supported

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -0,0 +1,720 @@
...
+        featureSetValues["features"] = {}
+        exported_dict["feature-management"] = {"featureSet": featureSetValues}
+
+    try:
+        # retrieved_features is a list of FeatureFlag objects
+        for feature in retrieved_features:
+
+            # if feature state is on or off, it means there are no filters
+            if feature.state == "on":
+                feature_state = True
```
shenmuxiaosen: feature.state is always lower case? either we convert to lower or feature.state can be an enum.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -0,0 +1,720 @@
...
+
+            elif feature.state == "off":
+                feature_state = False
+
+            elif feature.state == "conditional":
+                feature_state = {}
+                feature_state["EnabledFor"] = []
```
shenmuxiaosen: nit: feature_state = {"EnabledFor":[]}

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -0,0 +1,720 @@
...
+                client_filters = feature.conditions["client_filters"]
+                # client_filters is a list of dictionaries, where all dictionaries have 2 keys - Name and Parameters
+                for filter_ in client_filters:
+                    feature_filter = {}
+                    feature_filter["Name"] = filter_.name
```
shenmuxiaosen: >feature_filter [](start = 20, length = 14)

shouldn't this be FeatureFilter type?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -0,0 +1,720 @@
...
+                    feature_filter["Parameters"] = filter_.parameters
+                    feature_state["EnabledFor"].append(feature_filter)
+
+            feature_entry = {}
+            feature_entry[feature.key] = feature_state
```
shenmuxiaosen: nit: feature_entry = {feature.key:feature_state}

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -0,0 +1,720 @@
...
+
+            if format_ == 'json':
+                exported_dict["FeatureManagement"].update(feature_entry)
+
+            elif format_ in ('yaml', 'properties'):
+                featureSetValues["features"].update(feature_entry)
+                exported_dict["feature-management"].update({"featureSet": featureSetValues})
+
+        if format_ == 'properties':
+            exported_dict = __flatten_dict_to_properties(exported_dict)
```
shenmuxiaosen: kind of confusing, for feature flags we only support json right? For properties or yaml, what is expected behavior?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: FeatureFlag object has state attribute whose value is always converted to lowercase.

---
In reply to: [356324912](https://github.com/Azure/azure-cli/pull/11467#discussion_r356324912) [](ancestors = 356324912)

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: For consistency with backend schema as defined here: [https://github.com/Azure/AppConfiguration/blob/master/docs/FeatureManagement/FeatureFlag.v1.0.0.schema.json](https://github.com/Azure/AppConfiguration/blob/master/docs/FeatureManagement/FeatureFlag.v1.0.0.schema.json) 

---
In reply to: [356317261](https://github.com/Azure/azure-cli/pull/11467#discussion_r356317261) [](ancestors = 356317261)

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: I can use FeatureFlagValue object, but I don't think its necessary to create a new object every time since we are dumping the value to  a new KeyValue object and assigning new values in each iteration.

---
In reply to: [356310927](https://github.com/Azure/azure-cli/pull/11467#discussion_r356310927) [](ancestors = 356310927)

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: We want to keep it consistent with the .NET schema for JSON files, so the attribute has to be "Name" instead of "name" from FeatureFilter object. Same for parameters.

---
In reply to: [356333207](https://github.com/Azure/azure-cli/pull/11467#discussion_r356333207) [](ancestors = 356333207)

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: We support 'alwayson' filter without being case sensitive.

---
In reply to: [356317592](https://github.com/Azure/azure-cli/pull/11467#discussion_r356317592) [](ancestors = 356317592)

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: Yes, if even one filter is true, feature is enabled. More discussion here: [https://github.com/microsoft/FeatureManagement-Dotnet/issues/28](https://github.com/microsoft/FeatureManagement-Dotnet/issues/28) 

---
In reply to: [356318399](https://github.com/Azure/azure-cli/pull/11467#discussion_r356318399) [](ancestors = 356318399)

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: Ok, I can log a warning and treat them as normal KV if someone tries to import features from yaml/properties file.

---
In reply to: [356304006](https://github.com/Azure/azure-cli/pull/11467#discussion_r356304006) [](ancestors = 356304006)

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: sounds good.

---
In reply to: [356322906](https://github.com/Azure/azure-cli/pull/11467#discussion_r356322906) [](ancestors = 356322906)

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: Logging warnings and ignoring feature flags while exporting to yaml/properties file.

---
In reply to: [356334456](https://github.com/Azure/azure-cli/pull/11467#discussion_r356334456) [](ancestors = 356334456)

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -167,6 +171,44 @@ def custom_serialize_conditions(conditions_dict):
...
+                                           default=lambda o: o.__dict__,
+                                           ensure_ascii=False),
+                          content_type=FEATURE_FLAG_CONTENT_TYPE,
+                          tags={})
+
+        set_kv.locked = featureflag.locked
+        set_kv.last_modified = featureflag.last_modified
+
+    except:
+        logger.debug("Exception while converting feature flag to key value:\n%s\n", featureflag)
```
Juliehzl: One question here is if the exception raised will throw clear error message.
If not, what about make logger.debug message here be part of exception message?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -0,0 +1,709 @@
...
+
+def __read_kv_from_file(file_path, format_, separator=None, prefix_to_add="", depth=None):
+    config_data = {}
+    try:
+        with io.open(file_path, 'r', encoding=__check_file_encoding(file_path)) as config_file:
+            if format_ == 'json':
+                config_data = json.load(config_file)
+                if 'FeatureManagement' in config_data:
+                    del config_data['FeatureManagement']
+            elif format_ == 'yaml':
```
Juliehzl: Should you log warning here as you said in https://github.com/Azure/azure-cli/pull/11467/files#r356356673?

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
Juliehzl: Thanks for explanation! But it will be better if you can make it more clear and easy understanding.

Juliehzl:
No other comment from CLI side. Please resolve left comments and make CI pass. We will freeze code tomorrow for the next release. If you want to make it merge before code freeze, please make your PR ready ASAP.

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
avanigupta: Done

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: done

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
avanigupta: done

---
In reply to: [356655044](https://github.com/Azure/azure-cli/pull/11467#discussion_r356655044) [](ancestors = 356655044)

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -167,6 +171,48 @@ def custom_serialize_conditions(conditions_dict):
...
+        set_kv = KeyValue(key=FEATURE_FLAG_PREFIX + featureflag.key,
+                          label=featureflag.label,
+                          value=json.dumps(feature_flag_value,
+                                           default=lambda o: o.__dict__,
+                                           ensure_ascii=False),
+                          content_type=FEATURE_FLAG_CONTENT_TYPE,
+                          tags={})
+
+        set_kv.locked = featureflag.locked
+        set_kv.last_modified = featureflag.last_modified
```
shenmuxiaosen: These two are server generated. No need to map to KeyValue. As in KeyValue, etag/last_modified/locked are not in ctor parameters. Although in Python basically there is no private property, it won't be included in request body and will aslo be ignored by backend.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -0,0 +1,715 @@
...
+                continue
+            kv.key = kv.key[len(prefix_to_remove):]
+            # add prefix if specified
+            kv.key = prefix_to_add + kv.key
+    except Exception as exception:
+        raise CLIError(str(exception))
+    return key_values
+
+
+def __read_features_from_config_store(cmd, name=None, connection_string=None, key=None, label=None):
```
shenmuxiaosen: >__read_features_from_config_store [](start = 4, length = 33)

this helper method looks unnecessary. caller can call list_feature directly. Also it is odd to let this internal helper method to depend on feature.py.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
shenmuxiaosen: The retrieved_features was originally got from list_feature() which will return a list of FeatureFlag, then featureFlag.condition["client_filters"] should be a list of FeatureFilter objects instead of dictionary. 

---
In reply to: [356347693](https://github.com/Azure/azure-cli/pull/11467#discussion_r356347693) [](ancestors = 356347693,356333207)

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -0,0 +1,715 @@
...
+                for filter_ in client_filters:
+                    feature_filter = {}
+                    feature_filter["Name"] = filter_.name
+                    if filter_.parameters:
+                        feature_filter["Parameters"] = filter_.parameters
+                    feature_state["EnabledFor"].append(feature_filter)
+
+            feature_entry = {feature.key: feature_state}
+
+            if format_ == 'json':
```
shenmuxiaosen: no need to check format here. Already done before

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -0,0 +1,715 @@
...
+    except Exception as exception:
+        raise CLIError("Failed to export feature flags. " + str(exception))
+
+
+def __convert_feature_dict_to_keyvalue_list(features_dict, format_):
+    # pylint: disable=too-many-nested-blocks
+    key_values = []
+    default_conditions = {'client_filters': []}
+    default_value = FeatureFlagValue(id_="")
+
```
shenmuxiaosen: can you create a new FeatureFlagValue instance inside of the for loop in line 613? Otherwise, we have to be careful that every iteration, we will clean the last round default. The default_conditions does make sense as it is an actual default value.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -0,0 +1,715 @@
...
+                    try:
+                        default_value.conditions = {'client_filters': v["EnabledFor"]}
+                    except KeyError:
+                        raise CLIError("Feature '{0}' must contain 'EnabledFor' definition or have a true/false value. \n".format(str(k)))
+
+                    if not default_value.conditions["client_filters"]:
+                        # We support alternate format for specifying always OFF features
+                        # "FeatureU": {"EnabledFor": []}
+                        default_value.enabled = False
+                        default_value.conditions = default_conditions
```
shenmuxiaosen: when every iteration you create a FeatureFlagValue, you can assign the default condition to it. Then in here, you don't need the if check again.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -0,0 +1,715 @@
...
+
+
+def __serialize_feature_list_to_comparable_json_object(features):
+    res = {}
+    for feature in features:
+        # state
+        feature_json = {'state': feature.state}
+        # description
+        feature_json['description'] = feature.description
+        # conditions
```
shenmuxiaosen: why we compare description in here but not in __serialize_features_from_kv_list_to_comparable_json_object

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -0,0 +1,715 @@
...
+def __serialize_features_from_kv_list_to_comparable_json_object(keyvalues):
+    res = {}
+    for kv in keyvalues:
+        feature = map_keyvalue_to_featureflag(kv)
+        # state
+        feature_json = {'state': feature.state}
+        # conditions
+        feature_json['conditions'] = feature.conditions
+        # key
+        res[feature.key] = feature_json
```
shenmuxiaosen: can we reuse __serialize_feature_list_to_comparable_json_object?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -0,0 +1,715 @@
...
+    logger.warning('\n---------------- Feature Flags Preview (Beta) -------------')
+    if not new_json:
+        logger.warning('\nSource configuration is empty. No changes will be made.')
+        return False
+
+    # perform diff operation
+    # to simplify output, add one shared key in src and dest configuration
+    new_json['@base'] = ''
+    old_json['@base'] = ''
+    differ = JsonDiffer(syntax='symmetric')
```
shenmuxiaosen: >symmetric [](start = 32, length = 9)

why we use symmetric for feature flag but explicit for key values?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: done

---
In reply to: [356883121](https://github.com/Azure/azure-cli/pull/11467#discussion_r356883121) [](ancestors = 356883121)

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: done

---
In reply to: [356885978](https://github.com/Azure/azure-cli/pull/11467#discussion_r356885978) [](ancestors = 356885978)

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: Oh I didn't realize __serialize_features_from_kv_list_to_comparable_json_object is dead code now. Removing it. As for the description, we should still compare it in case features are being exported/imported between two config stores.

---
In reply to: [356898632](https://github.com/Azure/azure-cli/pull/11467#discussion_r356898632) [](ancestors = 356898632)

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: done

---
In reply to: [356898740](https://github.com/Azure/azure-cli/pull/11467#discussion_r356898740) [](ancestors = 356898740)

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: explicit syntax doesn't work when comparing feature filters, probably due to whitespacing issue with the json formatting of filters. 

---
In reply to: [356899926](https://github.com/Azure/azure-cli/pull/11467#discussion_r356899926) [](ancestors = 356899926)

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -0,0 +1,689 @@
...
+def __convert_feature_dict_to_keyvalue_list(features_dict, format_):
+    # pylint: disable=too-many-nested-blocks
+    key_values = []
+    default_conditions = {'client_filters': []}
+
+    try:
+        if format_ == 'json':
+            for k, v in features_dict.items():
+                key = FEATURE_FLAG_PREFIX + str(k)
+                default_value = FeatureFlagValue(id_=str(k))
```
shenmuxiaosen: >default_value [](start = 16, length = 13)

nit: rename it to something meaningful, like feature = FeatureFlagValue(id_=str(k))

avanigupta: done

src/azure-cli/HISTORY.rst
```diff
@@ -3,6 +3,12 @@
 Release History
 ===============
 
+**AppConfig**
+
+* Add support for importing/exporting feature flags
+
+2.0.78
+++++++
```
haroldrandom: Don't do 
```
2.0.78
++++++
```

Please rebase/merge the lates the dev branch

avanigupta: done

PR #11508, opened 2019-12-08
Fix property name for Kubernetes version to show in az aks nodepool list

seanmck:
This PR fixes the `az aks nodepool list` custom formatter to reference the correct property name for the Kubernetes version, which is `orchestratorVersion`. This ensures that the Kubernetes version actually shows up when nodepools are listed.

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acs/_format.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/_format.py
@@ -19,7 +19,7 @@ def _aks_agentpool_table_format(result):
     parsed = compile_jmes("""{
         name: name,
         osType: osType,
-        kubernetesVersion: kubernetesVersion,
+        kubernetesVersion: orchestratorVersion,
         vmSize: vmSize,
         osDiskSizeGB: osDiskSizeGB,
         count: count,


haroldrandom:
@yugangw-msft Could you please review this PR?

src/azure-cli/azure/cli/command_modules/acs/_format.py
```diff
@@ -19,7 +19,7 @@ def _aks_agentpool_table_format(result):
     parsed = compile_jmes("""{
         name: name,
         osType: osType,
-        kubernetesVersion: kubernetesVersion,
+        kubernetesVersion: orchestratorVersion,
```
yugangw-msft: I assume this worked before, but there was a breaking change later on which invalidated it.

seanmck: I don't think it did. We have inconsistently named properties for Kubernetes version between the control plane (`kubernetesVersion`) and agent pools (`orchestratorVersion`), so I'm guessing they got mixed up when the formatter was added. You can see in the original PR (#10809) that the add node pool operation sets is correctly:

```
agent_pool = AgentPool(
        name=nodepool_name,
        count=int(node_count),
        vm_size=node_vm_size,
        os_type=os_type,
        storage_profile=ContainerServiceStorageProfileTypes.managed_disks,
        vnet_subnet_id=vnet_subnet_id,
        agent_pool_type="VirtualMachineScaleSets",
        max_pods=int(max_pods) if max_pods else None,
        orchestrator_version=kubernetes_version,
        availability_zones=zones,
        node_taints=taints_array
```

PR #11522, opened 2019-12-10
[AppConfig]Fix pagination and globalization bugs. 

shenmuxiaosen:
- Fix bug when appending api-version to request url. The existing solution doesn't work with pagination. 
- For globalization, we should support showing languages besides English as our backend service support unicode. 

---


shenmuxiaosen:
Please add @avanigupta as reviewer. Thanks.

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -855,7 +856,7 @@ def __export_keyvalues(fetched_items, format_, separator, prefix=None):
 
         if exported_dict and exported_list:
             logger.error("Can not export to a valid file! Some keys have been dropped. %s", json.dumps(
-                exported_dict, indent=2))
+                exported_dict, indent=2, ensure_ascii=False))
```
avanigupta: Can you please remove this one since I'm changing it in my PR?

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -233,7 +234,7 @@ def delete_key(cmd,
     if not_deleted_entries:
         if deleted_entries:
             logger.error('Deletion operation partially succeed. Some keys are not successfully deleted. \n %s',
```
avanigupta: do you know if %s formatting will support non-ascii characters?

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
shenmuxiaosen: sure thing

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
shenmuxiaosen: It does. It just the json.dump() has a default argument that converts all characters to ascii that we need change.

avanigupta:
>         url = 'https://{}{}'.format(endpoint, query_url)

Do we not need to append api version here? Same for unlock and delete urls.

---
Refers to: src/azure-cli/azure/cli/command_modules/appconfig/_azconfig/azconfig_client.py:319 in f75c1d3. [](commit_id = f75c1d304aa45540c8411e2eec581efa12f20067, deletion_comment = False)

avanigupta:
>     def delete_keyvalue_by_key_label(self, key, label=None, modify_options=None):

Looks like dead code. Can you remove it please?

---
Refers to: src/azure-cli/azure/cli/command_modules/appconfig/_azconfig/azconfig_client.py:153 in f75c1d3. [](commit_id = f75c1d304aa45540c8411e2eec581efa12f20067, deletion_comment = False)

shenmuxiaosen:
>     def delete_keyvalue_by_key_label(self, key, label=None, modify_options=None):

This is part of our python SDK. We have two methods to delete key values.  delete_keyvalue(keyvalue) will check etag and fail if mismatch. For delete_keyvalue_by_key_label(key, label) it is a force delete that won't check pre condition. This may not be used in CLI, but we should keep it in internal python sdk. We may consider to use the official python SDK once it is stabilized. 

---
In reply to: [564198076](https://github.com/Azure/azure-cli/pull/11522#issuecomment-564198076) [](ancestors = 564198076)

---
Refers to: src/azure-cli/azure/cli/command_modules/appconfig/_azconfig/azconfig_client.py:153 in f75c1d3. [](commit_id = f75c1d304aa45540c8411e2eec581efa12f20067, deletion_comment = False)

shenmuxiaosen:
>         url = 'https://{}{}'.format(endpoint, query_url)

yes, good catch. I will add it. Same with unlock and delete. 

---
In reply to: [564196480](https://github.com/Azure/azure-cli/pull/11522#issuecomment-564196480) [](ancestors = 564196480)

---
Refers to: src/azure-cli/azure/cli/command_modules/appconfig/_azconfig/azconfig_client.py:319 in f75c1d3. [](commit_id = f75c1d304aa45540c8411e2eec581efa12f20067, deletion_comment = False)

Juliehzl:
Hi @shenmuxiaosen, please update history.rst for your change.

Issue #11444, opened 2019-12-01
Plugging in HaTs survey to Azure CLI


yonzhan:
this survey would be adequate for getting a HaTs baseline score for the Azure CLI. Based on what has worked in the past, I wanted something that has no more than 3 mandatory questions and no more than 10 questions all up. 

https://aka.ms/clihats

Once this is scrubbed, we plan to rollout the same kind of survey for our other command line tools (Azure PowerShell, Azure Terraform provider,  Ansible, etc). 

Can plug in HasTs survey during Bootstrap/reminder me later (never or 3 months).

yonzhan:
Planned in S162.

fengzhou-msft:
We need to figure out the following to implement this feature:
1. Ways to identify user.
2.Where to save users' selected options (reminder me later, never or 3 months), whether the user has already done the survey.
3.Are we targeting all users or selected users?
4. When/How often do we check whether to show the survey?

PR #11526, opened 2019-12-10
Plugging in HaTs survey

fengzhou-msft:
Implement #11444


yungezz:
let's merge it after PM's confirmation

src/azure-cli-core/azure/cli/core/commands/constants.py
```diff
@@ -29,3 +30,6 @@
 DEFAULT_QUERY_TIME_RANGE = 3600000
 
 BLACKLISTED_MODS = ['context', 'shell', 'documentdb', 'component']
+
+SURVEY_PROMPT = Fore.YELLOW + Style.BRIGHT + 'Please let us know how we are doing: ' + Fore.BLUE \
```
qwordy: Starting with "Please" looks like a command. Maybe the prompt can be more friendly.

src/azure-cli-core/azure/cli/core/_help.py
```diff
@@ -150,6 +151,7 @@ def new_normalize_text(s):
     def show_help(self, cli_name, nouns, parser, is_group):
         self.update_loaders_with_help_file_contents(nouns)
         super(AzCliHelp, self).show_help(cli_name, nouns, parser, is_group)
+        print(SURVEY_PROMPT)
```
qwordy: each time when typing "-h", print this prompt?

fengzhou-msft: yes, that's by design.

qwordy: Will users be bothered? Reduce frequency?

src/azure-cli-core/azure/cli/core/commands/constants.py
fengzhou-msft: IMHO, using "Please" is  kind of friendly.

src/azure-cli-core/azure/cli/core/_help.py
fengzhou-msft: Originally, we thought of providing an option to turn the survey prompt off. But since users can just ignore it, let's just leave it there and see how it goes. We can improve this when we have more feedback.

mook-as: Hi there!

Please consider this as feedback that this is kind of annoying :) Especially since that's a coloured prompt (whereas the actual help text is not).

I've filled in the survey… but I could only select "neither good nor bad" for all entries, as I've _just started using the CLI_ (which is why I'm looking at the help).

Sorry for intruding on your PR, I'll go back to being quiet now, thanks!

yungezz: Thanks for the feedback. We'll look at then.

Issue #10768, opened 2019-10-08
az policy set-definition create is missing metadata argument


bgelens:
**Describe the bug**
az policy set-definition create is missing the metadata argument. This is necessary to provide, among else, the category information.

This is similar to #5798

**Environment summary**
Install Method (e.g. pip, interactive script, apt-get, Docker, MSI, edge build) / CLI version (`az --version`) / OS version / Shell Type (e.g. bash, cmd.exe, Bash on Windows)  

```
azure-cli                         2.0.74

command-modules-nspkg              2.0.3
core                              2.0.74
nspkg                              3.0.4
telemetry                          1.0.3

Extensions:
aks-preview                        0.4.5
azure-devops                      0.13.0

Python location '/usr/local/Cellar/azure-cli/2.0.74/libexec/bin/python'
Extensions directory '/Users/bengelens/.azure/cliextensions'

Python (Darwin) 3.7.4 (default, Sep  7 2019, 18:27:02) 
[Clang 10.0.1 (clang-1001.0.46.4)
```

yonzhan:
How is the status of this issue?

PR #11530, opened 2019-12-10
[ARM] policy set-definition support --metadata

zhoxing-ms:
Feature: #10768
`policy set-definition create`: Add support for `--metadata`


src/azure-cli/HISTORY.rst
```diff
@@ -16,6 +16,7 @@ Release History
 **ARM**
 
 * Update azure-mgmt-resource package to use 6.0.0
+* az policy set-definition create/update: Add support for `--metadata`.
```
Juliehzl: `Add new parameter `--metadata` to support adding metadata information for policy set definitions` may be more clear.

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -1448,6 +1449,9 @@ def create_policy_setdefinition(cmd, name, definitions, params=None, display_nam
     PolicySetDefinition = cmd.get_models('PolicySetDefinition')
     parameters = PolicySetDefinition(policy_definitions=definitions, parameters=params, description=description,
                                      display_name=display_name, policy_definition_groups=definition_groups)
+
+    if cmd.supported_api_version(min_api='2017-06-01-preview'):
```
Juliehzl: Can we add a condition here to check if metadata value is None?

src/azure-cli/HISTORY.rst
zhoxing-ms: Resolved

src/azure-cli/azure/cli/command_modules/resource/custom.py
zhoxing-ms: My writing refers to the similar logic in the "create policy definition" method. I want to keep consistent with the original writing style, so that the code is more neat.

zhoxing-ms: When metadata is none, it can work normally now. If the logic here becomes more complex in the future, I will add the verification of metadata

yonzhan:
Good first blood:)

PR #11551, opened 2019-12-11
Update Cosmos DB python package

shurd:
This change updates the azure-mgmt.-cosmosdb package to use the most recent version (0.11.0). This version uses "patch" to update accounts instead of "put", which helps prevent unwanted updates.

---


yonzhan:
add to S163.

src/azure-cli/azure/cli/command_modules/cosmosdb/custom.py
```diff
@@ -869,30 +836,10 @@ def cli_cosmosdb_network_rule_remove(cmd,
     if not rule_removed:
         raise CLIError("This rule does not exist for the Cosmos DB account")
 
-    locations = []
```
Juliehzl: Why do you remove location related code here?

src/azure-cli/azure/cli/command_modules/cosmosdb/custom.py
```diff
@@ -543,20 +510,22 @@ def cli_cosmosdb_mongodb_collection_update(client,
...
-    mongodb_collection_resource.indexes = mongodb_collection.indexes
+    mongodb_collection_resource.shard_key = mongodb_collection.resource.shard_key
+    mongodb_collection_resource.indexes = mongodb_collection.resource.indexes
 
     if _populate_mongodb_collection_definition(mongodb_collection_resource, None, indexes):
         logger.debug('replacing MongoDB collection')
 
+    mongodb_collection_create_update_resource = MongoDBCollectionCreateUpdateParameters(
+        resource=mongodb_collection_resource,
+        options={})
```
Juliehzl: Why do you want to pass options with {} here?

src/azure-cli/azure/cli/command_modules/cosmosdb/custom.py
```diff
@@ -465,12 +425,15 @@ def cli_cosmosdb_gremlin_graph_update(client,
                                           None):
         logger.debug('replacing Gremlin graph')
 
+    gremlin_graph_create_update_resource = GremlinGraphCreateUpdateParameters(
+        resource=gremlin_graph_resource,
+        options={})
```
Juliehzl: Why do you want to pass options with {} here?

src/azure-cli/azure/cli/command_modules/cosmosdb/custom.py
```diff
@@ -356,12 +309,15 @@ def cli_cosmosdb_sql_container_update(client,
                                           None):
         logger.debug('replacing SQL container')
 
+    sql_container_create_update_resource = SqlContainerCreateUpdateParameters(
+        resource=sql_container_resource,
+        options={})
```
Juliehzl: Why do you want to pass options with {} here?

src/azure-cli/azure/cli/command_modules/cosmosdb/custom.py
```diff
@@ -178,52 +158,21 @@ def cli_cosmosdb_update(client,
...
-
-    if virtual_network_rules is None:
-        virtual_network_rules = existing.virtual_network_rules
-
-    if tags is None:
-        tags = existing.tags
-
-    if enable_multiple_write_locations is None:
-        enable_multiple_write_locations = existing.enable_multiple_write_locations
-
```
Juliehzl: Why do you delete this part?

shurd: Originally, we were using "put" to update the account, and now we are using "patch". With "patch", null values are ignored in our RP so we do not need to copy over the existing value of each property. All this code was used to ensure that an account update (via "put") did not incorrectly reset a property to the default value.

src/azure-cli/azure/cli/command_modules/cosmosdb/custom.py
shurd: options is a required keyword argument for this object, but we do not allow option updates via this CLI command. Currently, they can update the options by performing a throughput update.

src/azure-cli/azure/cli/command_modules/cosmosdb/custom.py
shurd: This is the same reasoning as with MongoDBCollectionCrateUpdateParameters. It is a require parameter in the SDK, but we do not allow option updates via this command.

src/azure-cli/azure/cli/command_modules/cosmosdb/custom.py
shurd: This is the same reasoning as with MongoDBCollectionCrateUpdateParameters. It is a required parameter in the SDK, but we do not allow option updates via this command.

src/azure-cli/azure/cli/command_modules/cosmosdb/custom.py
shurd: This command is specifically to update virtual_network_rules on the account. All the other code I removed was just copying the existing account state so that other properties are not incorrectly update or reset to defaults. Now that we are using "patch", we only need to include the relevant properties since empty/null values are ignored by the RP.

PR #11571, opened 2019-12-12
[AppConfig] Adding new command 'set-keyvault' to kv subgroup

avanigupta:
---


avanigupta:
Please add @shenmuxiaosen as reviewer.

yonzhan:
@jiasli please help with the PR.

src/azure-cli/azure/cli/command_modules/appconfig/_validators.py
```diff
@@ -147,3 +147,17 @@ def validate_filter_parameter(string):
...
+    from azure.keyvault.key_vault_id import KeyVaultIdentifier
+
+    identifier = getattr(namespace, 'secret_identifier', None)
+    if identifier:
+        try:
+            KeyVaultIdentifier(uri=identifier, collection="secrets")
+        except Exception as e:
+            raise CLIError("--secret-identifier is invalid.\n{0}".format(str(e)))
+    else:
+        raise CLIError("usage error: --secret-identifier is required.")
```
shenmuxiaosen: No need to validate the existence for secret identifier. In keyvalue.py, set_keyvault method already has a required input "secret_identifier".

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -193,6 +194,65 @@ def set_key(cmd,
...
+                 name=None,
+                 label=None,
+                 tags=None,
+                 yes=False,
+                 connection_string=None):
+    connection_string = resolve_connection_string(cmd, name, connection_string)
+    azconfig_client = AzconfigClient(connection_string)
+
+    keyvault_ref_value = json.dumps({"uri": secret_identifier}, ensure_ascii=False)
+    retry_times = 3
```
shenmuxiaosen: why we need json dumps here to create a dictionary?

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -193,6 +194,65 @@ def set_key(cmd,
...
+
+        verification_kv = {
+            "key": set_kv.key,
+            "label": set_kv.label,
+            "content_type": set_kv.content_type,
+            "value": set_kv.value,
+            "tags": set_kv.tags
+        }
+        entry = json.dumps(verification_kv, indent=2, sort_keys=True)
+        confirmation_message = "Are you sure you want to set the keyvault reference: \n" + entry + "\n"
```
shenmuxiaosen: add ensure_ascii=False in json.dumps

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
avanigupta: 'value' attribute of KeyValue object is supposed to be a string. In order to preserve double quotes while converting this dict to str, we use json dumps.

---
In reply to: [358477565](https://github.com/Azure/azure-cli/pull/11571#discussion_r358477565) [](ancestors = 358477565)

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
avanigupta: done

---
In reply to: [358478003](https://github.com/Azure/azure-cli/pull/11571#discussion_r358478003) [](ancestors = 358478003)

src/azure-cli/azure/cli/command_modules/appconfig/_validators.py
avanigupta: done

---
In reply to: [358474008](https://github.com/Azure/azure-cli/pull/11571#discussion_r358474008) [](ancestors = 358474008)

src/azure-cli/azure/cli/command_modules/appconfig/_validators.py
```diff
@@ -147,3 +147,15 @@ def validate_filter_parameter(string):
...
+
+def validate_secret_identifier(namespace):
+    """ Validate the format of keyvault reference secret identifier """
+    from azure.keyvault.key_vault_id import KeyVaultIdentifier
+
+    identifier = getattr(namespace, 'secret_identifier', None)
+    try:
+        # this throws an exception for invalid format of secret identifier
+        KeyVaultIdentifier(uri=identifier)
+    except Exception as e:
```
shenmuxiaosen: >Exception [](start = 11, length = 9)

Any specific type of exception this can throw? Since our error message is very specific to secret identifier value.

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
shenmuxiaosen: Okay, so the key value will look like
{
"Key":"K1"
"Value":"{"uri":"identitfier"}"
} ?

---
In reply to: [358518179](https://github.com/Azure/azure-cli/pull/11571#discussion_r358518179) [](ancestors = 358518179,358477565)

src/azure-cli/azure/cli/command_modules/appconfig/_validators.py
avanigupta: Going through the keyvault SDK, it looks like they only throw ValueError and TypeError at this time. But I didn't find any official documentation which says they can't throw anything else in future. I could change our error message to clarify that the exception was thrown by keyvault. 

---
In reply to: [358527193](https://github.com/Azure/azure-cli/pull/11571#discussion_r358527193) [](ancestors = 358527193)

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
avanigupta: Actually, like this:

{
"key": "K1",
"value": "{\"uri\": \"[https://contoso.vault.azure.net/Secrets/DummySecret/Dummyversion\](https://contoso.vault.azure.net/Secrets/DummySecret/Dummyversion\)"}"
}

---
In reply to: [358528290](https://github.com/Azure/azure-cli/pull/11571#discussion_r358528290) [](ancestors = 358528290,358518179,358477565)

src/azure-cli/azure/cli/command_modules/appconfig/_help.py
```diff
@@ -138,6 +138,16 @@
     text: az appconfig kv set --connection-string Endpoint=https://contoso.azconfig.io;Id=xxx;Secret=xxx --key color --value red --tags key1=value1 key2=value2
 """
 
+helps['appconfig kv set-keyvault'] = """
```
Juliehzl: From my understanding, it is a little similar with `az appconfig kv set` command. Why not use `az appconfig kv set` to set keyvault value?

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -273,6 +274,65 @@ def set_key(cmd,
...
+def set_keyvault(cmd,
+                 key,
+                 secret_identifier,
+                 name=None,
+                 label=None,
+                 tags=None,
+                 yes=False,
+                 connection_string=None):
+    connection_string = resolve_connection_string(cmd, name, connection_string)
+    azconfig_client = AzconfigClient(connection_string)
```
Juliehzl: For your azconfig_client, you should configure it in your configstore_keyvalue_util sdk with client_factory. After configuring in configstore_keyvalue_util, you can just import the client in your set_keyvalue() function.

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/test_appconfig_commands.py
```diff
@@ -206,6 +206,42 @@ def test_azconfig_kv(self, resource_group, location):
                          self.check('[0].value', updated_entry_value),
                          self.check('[0].label', updated_label)])
 
+        # KeyVault reference tests
```
Juliehzl: You can consider separate the test_azconfig_kv to several parts. In this way, you don't need to re-recording the unrelated tests and only focus on the tests related to your new features.

src/azure-cli/azure/cli/command_modules/appconfig/_help.py
avanigupta: To set key vault reference, we need to provide users with different arguments than the set kv command. This also aligns with the portal experience, where set kv and set key vault are different.

---
In reply to: [361830243](https://github.com/Azure/azure-cli/pull/11571#discussion_r361830243) [](ancestors = 361830243)

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/test_appconfig_commands.py
avanigupta: In this case, it would be better to add the new feature tests in the same scenario test since I have to reuse/update some of the pre-existing resources created in the commands above. 

---
In reply to: [361830867](https://github.com/Azure/azure-cli/pull/11571#discussion_r361830867) [](ancestors = 361830867)

PR #11588, opened 2019-12-16
Fix: Recovery Services Vault tags cannot be updated

zhoxing-ms:
Fix Recovery Services Vault tags cannot be updated
Link: [https://icm.ad.msft.net/imp/v3/incidents/details/162448402/home](url)


src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -2131,22 +2131,13 @@ def update(self, parameters):
...
-            identity=resource.identity)
+        parameters = GenericResource(tags=tags)
 
         if self.resource_id:
-            return self.rcf.resources.create_or_update_by_id(self.resource_id, self.api_version,
-                                                             parameters)
-        return self.rcf.resources.create_or_update(self.resource_group_name,
+            return self.rcf.resources.update_by_id(self.resource_id, self.api_version,parameters)
+
+        return self.rcf.resources.update(self.resource_group_name,
```
Juliehzl: Is there any scenario which will call create using this class?

yonzhan:
add to S163.

src/azure-cli/HISTORY.rst
```diff
@@ -7,6 +7,10 @@ Release History
 
 * Add support for importing/exporting feature flags
 
+**ARM**
+
+* Fix Recovery Services Vault tags cannot be updated
```
Juliehzl: Please specify the command you changed here.

src/azure-cli/HISTORY.rst
```diff
@@ -7,6 +7,10 @@ Release History
...
+
 2.0.78
```
Juliehzl: Why two 2.0.78 here? Please help fixed it.

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -2132,6 +2132,24 @@ def update(self, parameters):
 
     def tag(self, tags):
         resource = self.get_resource()
+
+        # please add the service type that needs to be requested with PATCH type here
+        # for example: the properties of RecoveryServices/vaults must be filled, and a PUT request that passes back
+        # to properties will fail due to the lack of properties, so the PATCH type should be used
+        need_patch_service = ['Microsoft.RecoveryServices/vaults']
+
+        if resource is not None and resource.type in need_patch_service:
```
Juliehzl: Can you add a test for Recovery Service tag scenario?

zhoxing-ms: Done

src/azure-cli/HISTORY.rst
zhoxing-ms: Done

src/azure-cli/HISTORY.rst
zhoxing-ms: Done

PR #11664, opened 2019-12-25
update for docker doc

fengzhou-msft:


yonzhan:
@fengzhou-msft  nice work, thanks.

Issue #9399, opened 2019-05-15
`pip install --upgrade azure-cli` fails on Cygwin due to psutil


ElvenSpellmaker:
**Describe the bug**
I tried to upgrade `azure-cli` from 2.0.31 to the latest (2.0.62 as of writing)

**To Reproduce**
`pip install --upgrade azure-cli`

**Expected behavior**
az should upgrade.

**Environment summary**
Windows 10
Cygwin (latest)

**Additional context**
```
Collecting psutil>=5.6.1 (from azure-cli-core==2.0.64->azure-cli)
  Using cached https://files.pythonhosted.org/packages/c6/c1/beed5e4eaa1345901b595048fab1c85aee647ea0fc02d9e8bf9aceb81078/psutil-5.6.2.tar.gz
    ERROR: Complete output from command python setup.py egg_info:
    ERROR: platform cygwin is not supported
    ----------------------------------------
ERROR: Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-install-NwRvbi/psutil/
```

yugangw-msft:
@adewaleo, the `psutil` was pulled in for the new `az feedback`, anyway we can find an alternative?

ElvenSpellmaker:
`2.0.65` seems to install under Cygwin now, but has problems running everything:
![image](https://user-images.githubusercontent.com/2286713/58321354-d5545e80-7e15-11e9-9487-997bbb29fabb.png)

Should I open a new issue for this? Thanks.

ElvenSpellmaker:
Weirdly enough I uninstalled all traces of the `azure-cli` (`pip list | awk '{print $1}' | grep '^azure' | xargs pip uninstall`) and tried re-installing and `2.0.65` complains about `psutil` again... 🤔

ElvenSpellmaker:
Also it seems I can't even install an older version any more as the `ps-util` breaks that.
```
Collecting psutil>=5.6.1 (from azure-cli-core->azure-cli==2.0.31)
  Using cached https://files.pythonhosted.org/packages/c6/c1/beed5e4eaa1345901b595048fab1c85aee647ea0fc02d9e8bf9aceb81078/psutil-5.6.2.tar.gz
```

adewaleo:
@ElvenSpellmaker, Yes, unfortunately it turns out that [psutil does not support cygwin](https://github.com/giampaolo/psutil/issues/82) yet. 

[...]

PR #11665, opened 2019-12-25
{Packaging} Get rid of psutil dependency

avoidik:


avoidik:
An attempt to fix https://github.com/Azure/azure-cli/issues/9399

My apologies for such rude approach, but I'd better provide shell as textual information rather than relying on contradictive dependencies.

jsntcy:
@fengzhou-msft, please help review it, thanks.

src/azure-cli/azure/cli/command_modules/feedback/custom.py
```diff
@@ -526,7 +513,6 @@ def _get_parent_proc_name():
     format_dict["cli_version"] = _get_az_version_summary()
     format_dict["python_info"] = "Python {}".format(platform.python_version())
     format_dict["platform"] = "{}".format(platform.platform())
-    format_dict["shell"] = "Shell: {}".format(_get_parent_proc_name())
```
fengzhou-msft: What's the consideration for removing psutil as we would lose shell information here?

avoidik: To bring back Cygwin support on Windows, psutil has compatibility issues. What's the purpose in collecting shell information? Why it cannot be provide by any other means?

fengzhou-msft: I agree with this trade. If future versions of psutil add support for Cygwin, we can then add it back.

avoidik: Thank you, I hope this change wouldn't cause to much impact for a feedback process flow

jiasli:
With the current limitation of psutil https://github.com/giampaolo/psutil/issues/82, I agree we should remove psutil dependency to ease of the life of Cygwin users. 

There seems to be [another way](https://stackoverflow.com/a/55598796/2199657) to detect if `az` is run in a PowerShell terminal. We can do that later.

qwordy:
Can we detect environment and then decide what code to run to avoid this bug? 
E.g. https://docs.python.org/3/library/platform.html
Deleting it is kind of rude as you mentioned.

fengzhou-msft:
> Can we detect environment and then decide what code to run to avoid this bug?
> E.g. https://docs.python.org/3/library/platform.html
> Deleting it is kind of rude as you mentioned.

Let's unblock Cygwin users first in this release. Shell info in feedback is a good-to-have thing. We can check the alternatives later.

ElvenSpellmaker:
Amazing guys thanks!

src/azure-cli/azure/cli/command_modules/feedback/custom.py
adewaleo: @avoidik @fengzhou-msft, the reason we including shell info was because some issues are platform and / or  shell-based

cmd and powershell behave differently for example.....

PR #11670, opened 2019-12-26
[Aladdin Parser] Parse aladdin.py into commands _help.py

haroldrandom:


fengzhou-msft:
Can you write some description about the whole process and how the file will be used?

PR #11704, opened 2019-12-30
[Compute][Feature]az vm create: Support default data source settings for vm workspace

MyronFanQiu:
support default data source settings during vm creation if `--workspace` is provided.

---


src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -1071,6 +1081,40 @@ def _prepare_workspace(cmd, resource_group_name, workspace):
...
+        from ._workspace_data_source_settings import default_windows_data_sources
+        default_data_sources = default_windows_data_sources
+
+    if default_data_sources is not None:
+        for data_source_kind, data_source_settings in default_data_sources.items():
+            for data_source_setting in data_source_settings:
+                data_source = DataSource(kind=data_source_kind,
+                                         properties=data_source_setting)
+                data_source_name = data_source_name_template.format(data_source_kind, _gen_guid())
+                data_sources_client.create_or_update(resource_group_name,
```
qwordy: What will happen if this operation fails? Will user know it?

MyronFanQiu: It will fail. Do you mean I should add try/except here?

src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
```diff
@@ -916,7 +916,7 @@ def build_av_set_resource(cmd, name, location, tags, platform_update_domain_coun
 
 
 # used for log analytics workspace
-def build_vm_mmaExtension_resource(_, vm_name, location):
+def build_vm_linux_log_analytics_workspace_agent(_, vm_name, location):
```
qwordy: Put comments below function name

src/azure-cli/azure/cli/command_modules/vm/custom.py
qwordy: I think both ways are OK. We need to define when the command execution should be considered successful.

src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
```diff
@@ -915,8 +915,10 @@ def build_av_set_resource(cmd, name, location, tags, platform_update_domain_coun
     return av_set
 
 
-# used for log analytics workspace
-def build_vm_mmaExtension_resource(_, vm_name, location):
+def build_vm_linux_log_analytics_workspace_agent(_, vm_name, location):
+    '''
```
qwordy: Triple double-quoted strings should be used for docstrings.

PR #11717, opened 2019-12-30
[RecoveryServices Backup]Added disk exclusion Feature for IAASVM

sambitratha:
---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/backup/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/backup/_help.py
@@ -323,7 +323,7 @@ type: command
 short-summary: Update disk exclusion settings associated with a backed up VM item.
 examples:
   - name: Update disk exclusion settings associated with a backed up VM item.
-    text: az backup protection update-for-vm --vault-name MyVault --resource-group MyResourceGroup --container-name MyContainer --item-name MyItem  --disk-list-setting exclude --diskslist 1
+    text: az backup protection update-for-vm --vault-name MyVault --resource-group MyResourceGroup --container-name MyContainer --item-name MyItem --disk-list-setting exclude --diskslist 1
 """
 
 helps['backup recoveryconfig'] = """
--- a/src/azure-cli/azure/cli/command_modules/backup/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/backup/_help.py
@@ -318,6 +318,14 @@ examples:
     text: az backup protection resume --vault-name MyVault --resource-group MyResourceGroup --container-name MyContainer --item-name MyItem --policy-name MyPolicy
 """
 
+helps['backup protection update-for-vm'] = """
+type: command
+short-summary: Update disk exclusion settings associated with a backed up VM item.
+examples:
+  - name: Update disk exclusion settings associated with a backed up VM item.
+    text: az backup protection update-for-vm --vault-name MyVault --resource-group MyResourceGroup --container-name MyContainer --item-name MyItem  --disk-list-setting exclude --diskslist 1
+"""
+
 helps['backup recoveryconfig'] = """
 type: group
 short-summary: Manage recovery configuration of an Azure workload backed up item.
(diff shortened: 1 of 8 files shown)


yonzhan:
@qwordy  please help with the PR.

src/azure-cli/azure/cli/command_modules/backup/_help.py
```diff
@@ -318,6 +318,14 @@
     text: az backup protection resume --vault-name MyVault --resource-group MyResourceGroup --container-name MyContainer --item-name MyItem --policy-name MyPolicy
 """
 
+helps['backup protection update-for-vm'] = """
+type: command
+short-summary: Update disk exclusion settings associated with a backed up VM item.
+examples:
+  - name: Update disk exclusion settings associated with a backed up VM item.
+    text: az backup protection update-for-vm --vault-name MyVault --resource-group MyResourceGroup --container-name MyContainer --item-name MyItem  --disk-list-setting exclude --diskslist 1
```
qwordy: two spaces in "--item-name MyItem  --disk-list-setting exclude".

src/azure-cli/azure/cli/command_modules/backup/_params.py
```diff
@@ -227,6 +238,9 @@ def load_arguments(self, _):
         c.argument('storage_account', help='Name or ID of the staging storage account. The VM configuration will be restored to this storage account. See the help for --restore-to-staging-storage-account parameter for more info.')
         c.argument('restore_to_staging_storage_account', arg_type=get_three_state_flag(), help='Use this flag when you want disks to be restored to the staging storage account using the --storage-account parameter. When not specified, disks will be restored to their original storage accounts. Default: false.')
         c.argument('target_resource_group', options_list=['--target-resource-group', '-t'], help='Use this to specify the target resource group in which the restored disks will be saved')
+        c.argument('diskslist', diskslist_type)
+        c.argument('restore_only_osdisk', arg_type=get_three_state_flag(), help='Use this flag to restore only OS disks of a backed up VM.')
+        c.argument('restoredisks', arg_type=get_three_state_flag(), help='Use this flag to restore specified disks of a backed up VM.')
```
qwordy: Maybe we can add a "_" between "restore" and "disks"

src/azure-cli/azure/cli/command_modules/backup/_params.py
```diff
@@ -227,6 +238,9 @@ def load_arguments(self, _):
         c.argument('storage_account', help='Name or ID of the staging storage account. The VM configuration will be restored to this storage account. See the help for --restore-to-staging-storage-account parameter for more info.')
         c.argument('restore_to_staging_storage_account', arg_type=get_three_state_flag(), help='Use this flag when you want disks to be restored to the staging storage account using the --storage-account parameter. When not specified, disks will be restored to their original storage accounts. Default: false.')
         c.argument('target_resource_group', options_list=['--target-resource-group', '-t'], help='Use this to specify the target resource group in which the restored disks will be saved')
+        c.argument('diskslist', diskslist_type)
```
qwordy: Maybe we can omit "list" in "diskslist"

src/azure-cli/azure/cli/command_modules/backup/_params.py
```diff
@@ -36,6 +36,8 @@
 protectable_item_name_type_help = """Specify the resource name to be protected by Azure Backup service."""
 backup_type_help = """'Full, Differential, Log, Copy-only-full' for backup Item type 'MSSQL'. 'Full, Differential' for backup item type 'SAPHANA'."""
 retain_until_help = """The date until which this backed up copy will be available for retrieval, in UTC (d-m-Y). For SAPHANA and SQL workload, retain-until parameter value will be overridden by the underlying policy."""
+diskslist_help = """List of disks to be excluded or included."""
```
qwordy: Usually Azure CLI supports both name or ID of resource.

pvrk: This is not a azure managed disk ID. This is the LUN of that disk in the context of that backed up VM. Hence this is list of LUNs.

src/azure-cli/azure/cli/command_modules/backup/_params.py
pvrk: As per my comment below, this is a list of disk LUNs. Hence I prefer diskslist

src/azure-cli/azure/cli/command_modules/backup/_params.py
pvrk: On second thoughts, this may not be required at all. The command name is already 'az backup restore restore-disks'. Here the expected arguments are either 'restore_only_osdisk' or "diskslist". Both are optional and only one of them is expected. If both of them are not given, then the service will go with default behavior. @sambitratha : Please remove the restoredisks parameter.

sambitratha: Done

PR #11719, opened 2019-12-30
[App Service] Add az webapp up flag for HTML static sites

### LINKED ISSUE #11592: az webapp up flag for html static sites
There is no reliable detection logic for plain old html apps.

we should add a -html flag that ignores detection logic and deploys this as a html app.

See:
- https://github.com/MicrosoftDocs/azure-docs/issues/44124

ammanifold:
Fixes #11592 
---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/_help.py
@@ -1722,6 +1722,7 @@ short-summary: >
     Create a webapp and deploy code from a local workspace to the app. The command is required to run from the folder
     where the code is present. Current support includes Node, Python, .NET Core and ASP.NET. Node,
     Python apps are created as Linux apps. .Net Core, ASP.NET apps are created as Windows apps.
+    Append the html flag to deploy as an HTML app.
 examples:
   - name: View the details of the app that will be created, without actually running the operation
     text: >
--- a/src/azure-cli/azure/cli/command_modules/appservice/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/_help.py
@@ -1741,7 +1741,7 @@ examples:
         az webapp up -n MyUniqueAppName --logs
   - name: Create a web app and deploy as a static HTML app.
     text: >
-        az webapp up -n MyUniqueAppName --html      
+        az webapp up -n MyUniqueAppName --html
 """
 
 helps['webapp update'] = """
--- a/src/azure-cli/azure/cli/command_modules/appservice/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/_help.py
@@ -1721,8 +1721,8 @@ type: command
 short-summary: >
     Create a webapp and deploy code from a local workspace to the app. The command is required to run from the folder
     where the code is present. Current support includes Node, Python, .NET Core and ASP.NET. Node,
-    Python apps are created as Linux apps. .Net Core, ASP.NET apps are created as Windows apps.
-    Append the html flag to deploy as an HTML app.
+    Python apps are created as Linux apps. .Net Core, ASP.NET, and static HTML apps are created as Windows apps.
+    Append the html flag to deploy as a static HTML app.
 examples:
   - name: View the details of the app that will be created, without actually running the operation
     text: >
@@ -1739,6 +1739,9 @@ examples:
   - name: Create a web app and enable log streaming after the deployment operation is complete. This will enable the default configuration required to enable log streaming.
     text: >
         az webapp up -n MyUniqueAppName --logs
+  - name: Create a web app and deploy as a static HTML app.
+    text: >
+        az webapp up -n MyUniqueAppName --html      
 """
 
 helps['webapp update'] = """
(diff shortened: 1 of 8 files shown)


ammanifold:
@panchagnula FYI :)

src/azure-cli/azure/cli/command_modules/appservice/_help.py
```diff
@@ -1722,6 +1722,7 @@
     Create a webapp and deploy code from a local workspace to the app. The command is required to run from the folder
     where the code is present. Current support includes Node, Python, .NET Core and ASP.NET. Node,
     Python apps are created as Linux apps. .Net Core, ASP.NET apps are created as Windows apps.
+    Append the html flag to deploy as an HTML app.
```
panchagnula: Can you add an example here on how to use this for static HTML

src/azure-cli/azure/cli/command_modules/appservice/_help.py
```diff
@@ -1722,6 +1722,7 @@
     Create a webapp and deploy code from a local workspace to the app. The command is required to run from the folder
     where the code is present. Current support includes Node, Python, .NET Core and ASP.NET. Node,
```
panchagnula: Update the help text "Current support includes Node, Python, .NET Core and ASP.NET. Node" to say and static html.

panchagnula:
@ammanifold just makes sure that you have tested the re-deploy scenario for HTML content with & without the flag & see how we handle this.

yonzhan:
@Juliehzl  please help with the PR.

src/azure-cli/azure/cli/command_modules/appservice/_params.py
```diff
@@ -420,6 +420,7 @@ def load_arguments(self, _):
         c.argument('location', arg_type=get_location_type(self.cli_ctx))
         c.argument('launch_browser', help="Launch the created app using the default browser", default=False, action='store_true', options_list=['--launch-browser', '-b'])
         c.argument('logs', help="Configure default logging required to enable viewing log stream immediately after launching the webapp", default=False, action='store_true')
+        c.argument('html', help="Ignore app detection and deploy as an html app", default=False, action='store_true')
```
qwordy: Is it a bool type? Do we need a "arg_type=" to restrict the value?

ammanifold: Not needed, we are just detecting if the flag is used or not (see 'logs' and 'launch_browser' argument above)

panchagnula:
> @ammanifold just makes sure that you have tested the re-deploy scenario for HTML content with & without the flag & see how we handle this.

Thanks for testing this out @ammanifold

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -3023,7 +3023,7 @@ def webapp_up(cmd, name, resource_group_name=None, plan=None, location=None, sku
         site_config = client.web_apps.get_configuration(rg_name, name)
     else:  # need to create new app, check if we need to use default RG or use user entered values
         logger.warning("webapp %s doesn't exist", name)
-        sku = get_sku_to_use(src_dir, sku)
+        sku = get_sku_to_use(src_dir, html, sku)
```
Juliehzl: can we avoid using positional arguments for each function?

Juliehzl:
> > @ammanifold just makes sure that you have tested the re-deploy scenario for HTML content with & without the flag & see how we handle this.
> 
> Thanks for testing this out @ammanifold

@ammanifold @panchagnula can we add this test scenario in our test file, which will help other team test for your command?

src/azure-cli/azure/cli/command_modules/appservice/custom.py
ammanifold: we need to pass these in case of the flag being set

ammanifold:
> > > @ammanifold just makes sure that you have tested the re-deploy scenario for HTML content with & without the flag & see how we handle this.
> > 
> > 
> > Thanks for testing this out @ammanifold
> 
> @ammanifold @panchagnula can we add this test scenario in our test file, which will help other team test for your command?

We are working on a suite of tests for az webapp up. Those will go in a separate PR

panchagnula:
> > > > @ammanifold just makes sure that you have tested the re-deploy scenario for HTML content with & without the flag & see how we handle this.
> > > 
> > > 
> > > Thanks for testing this out @ammanifold
> > 
> > 
> > @ammanifold @panchagnula can we add this test scenario in our test file, which will help other team test for your command?
> 
> We are working on a suite of tests for az webapp up. Those will go in a separate PR

Correct that work is on me to add tests & will work on it, for now we focus on manual testing for this scenario. thanks. This shouldn't block the change from being merged. Thanks!

src/azure-cli/azure/cli/command_modules/appservice/custom.py
Juliehzl: I know. I will prefer something like get_sku_to_use(src_dir=src_dir, html=html, sku=sku) to avoid positional arguments

yonzhan:
@panchagnula We can merge this PR if there are several requests pending on this one. While please take a look at @Juliehzl 's comments about get_sku_to_use(src_dir=src_dir, html=html, sku=sku) to avoid positional arguments. in this way it will not cause error when argument positions change.

yonzhan:
After this PR is merged. It is better to address the improvement in another PR.

PR #11722, opened 2019-12-31
Fix typo in `KeyVaultMgmtScenarioTest`

bim-msft:
---

This is a minor change, change `--enable-for-deploymen` to `--enable-for-deployment`.

### DIFF
--- a/src/azure-cli/HISTORY.rst
+++ b/src/azure-cli/HISTORY.rst
@@ -13,7 +13,7 @@ Release History
 
 **AppService**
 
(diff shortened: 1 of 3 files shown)


jiasli:
Typo fix doesn't need to go to HISTORY. Kindly remove it from HISTORY so we don't spam it. 

Also this won't actually cause problem because **argparse** is using [prefix matching](https://docs.python.org/3/library/argparse.html#argument-abbreviations-prefix-matching). But it is still good to fix it.

bim-msft:
Also, fix a typo in `HISTORY.rst` (`comand` -> `command`)

Issue #5110, opened 2017-12-15
[Resource] ManagedAppScenarioTest fails run live.


tjprescott:
Test is changed to live only temporarily to avoid breaking CI.  Cannot be re-recorded currently.

```
azure.cli.core.util.CLIError: Deployment failed. Correlation ID: 78dcc99e-2ef0-4ef9-a58f-8778b9b606d4. Principal 5e91139ac94b462ea6ff1ee95e8aac07 does not exist in the directory 54826b22-38d6-4fb2-bad9-b7b93a3e9c5a
```

yonzhan:
@Juliehzl  add to S162.

PR #11727, opened 2019-12-31
[ARM] Fix the test of test_managedapp

zhoxing-ms:
Fix the test of test_managedapp
issue: #5110 

- Because the group of Azure Active Directory which used in the case was deleted, I changed it to a new group-id
- The matching parameters of the template were not added to the previous command，So I changed the template which case used and give the matching parameters of the template.

---


yonzhan:
add to S164.

yonzhan:
Please merge this PR before this build.

PR #11750, opened 2020-01-02
{SQL} Add example for sql db create hyperscale

### LINKED ISSUE #11662: Include Azure SQL Hyperscale create sample
Hyperscale is a undocumented tier in documentation:

--edition --tier -e

The edition component of the sku. Allowed values include: Basic, Standard, Premium, GeneralPurpose, BusinessCritical.

Running list-editions command:

```
gabriel@Azure:~$ az sql db list-editions -l westus --query "[].name"
[
  "System",
  "Free",
  "Basic",
  "Standard",
  "Premium",
  "DataWarehouse",
  "Stretch",
  "GeneralPurpose",
  "BusinessCritical",
  "Hyperscale"
]
```
A example could also be included with other commom parameters:

Create a Hyperscale database with five read replicas
```
az sql db create -g mygroup -s myserver -n mydb -e Hyperscale --read-replicas 5
```

---
#### Document Details

* ID: c1bdd674-b454-e863-c943-9c979b466a92
* Version Independent ID: b15341f9-3f1b-00e0-a1c2-970f25a6ab03
* Content: [az sql db](https://docs.microsoft.com/en-us/cli/azure/sql/db?view=azure-cli-latest#az-sql-db-create)
* Content Source: [src/azure-cli/azure/cli/command_modules/sql/_help.py](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/sql/_help.py)
* Service: **sql-database**
* GitHub Login: @rloutlaw

jaredmoo:
Fixes #11662

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/sql/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/sql/_params.py
@@ -237,7 +237,7 @@ read_replicas_param_type = CLIArgumentType(
     help='The number of readonly replicas to provision for the database. '
     'Only settable for Hyperscale edition.')
 
-db_service_objective_examples = 'Basic, S0, P1, GP_Gen4_1, BC_Gen5_2, GP_Gen5_S_8.'
+db_service_objective_examples = 'Basic, S0, P1, GP_Gen4_1, GP_Gen5_S_8, BC_Gen5_2, HS_Gen5_32.'
 dw_service_objective_examples = 'DW100, DW1000c'
 
 
@@ -436,7 +436,7 @@ def load_arguments(self, _):
         c.argument('tier',
                    arg_type=tier_param_type,
                    help='The edition component of the sku. Allowed values include: Basic, Standard, '
-                   'Premium, GeneralPurpose, BusinessCritical.')
+                   'Premium, GeneralPurpose, BusinessCritical, Hyperscale.')
 
         c.argument('capacity',
                    arg_type=capacity_param_type,
--- a/src/azure-cli/azure/cli/command_modules/sql/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/sql/_help.py
@@ -73,6 +73,9 @@ examples:
     text: az sql db create -g mygroup -s myserver -n mydb -z false
   - name: Create a GeneralPurpose Gen5 2 vcore serverless database with auto pause delay of 120 minutes
     text: az sql db create -g mygroup -s myserver -n mydb -e GeneralPurpose -f Gen5 -c 2 --compute-model Serverless --auto-pause-delay 120
+  - name: Create a Hyperscale Gen5 2 vcore database with 2 read replicas
+    text: az sql db create -g mygroup -s myserver -n mydb -e Hyperscale -f Gen5 -c 2 --read-replicas 2
+
 """
 
 helps['sql db delete'] = """


jaredmoo:
Note: HISTORY.rst is not modified because the change only affects help text.

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -436,7 +436,7 @@ def load_arguments(self, _):
         c.argument('tier',
                    arg_type=tier_param_type,
                    help='The edition component of the sku. Allowed values include: Basic, Standard, '
-                   'Premium, GeneralPurpose, BusinessCritical.')
+                   'Premium, GeneralPurpose, BusinessCritical, Hyperscale.')
```
Juliehzl: here the allowed values change, right? There two small questions here:
1. can you try to use get_enum_type() for allowed values?
2. given more supported values here, it will be better to declare it in release note.

Juliehzl:
@jaredmoo  Please resolve my comments

src/azure-cli/azure/cli/command_modules/sql/_params.py
jaredmoo: I am deliberately not using get_enum_type() because there are some values of in `DatabaseEdition` that we don't want to allow for this specific command. I also am not declaring in release note because it's only a documentation change. Hyperscale was already supported.

jaredmoo:
Replied - thank you for the guidance but in this specific circumstance I think it does not apply :)

jaredmoo:
@Juliehzl can you review when available? Thank you :)

src/azure-cli/azure/cli/command_modules/sql/_params.py
Juliehzl: you can customize the allowed values using get_enum_type(["Basic", "Standard", "Premium", "GeneralPurpose", "BusinessCritical", "Hyperscale"]). We will recommend to use get_enum_type() because it will follow the same string format and case insensitive for parameter values.

yonzhan:
@jaredmoo  please help address the comments raised by @Juliehzl , thx.

src/azure-cli/azure/cli/command_modules/sql/_params.py
jaredmoo: Customizing the list is good, because that will allow us to hide the editions that aren't supported. However we still have the problem of forward compatibility: we are likely to release new editions in the future, so we would like the ability for the user to type in any value that they like so that today's version of CLI has support for future editions.

jaredmoo: Just to clarify, my understanding is that get_enum_list() will force the value to be validated in client-side. We want to avoid the validation so that "unknown values" are still passed to the server and let the server decide if it's valid.

Juliehzl: Although I can understand your concern, users should only input values allowed as shown in help message. 
get_enum_type() will check allowed values but it also can help transform enum values to allowed ones, which is very convenient to users.

Juliehzl: @jaredmoo any thought here?

jaredmoo: For us, the forward compatibility is important. We don't want to force customers to upgrade their CLI just to access new editions. This is why get_enum_type() doesn't work for us in this scenario.

Juliehzl: Got it.

PR #11753, opened 2020-01-02
[AppConfig] Validate key and feature names before setting and importing

avanigupta:
---


src/azure-cli/azure/cli/command_modules/appconfig/_validators.py
```diff
@@ -159,3 +162,32 @@ def validate_secret_identifier(namespace):
         KeyVaultIdentifier(uri=identifier)
     except Exception as e:
         raise CLIError("Received an exception while validating the format of secret identifier.\n{0}".format(str(e)))
+
+
+def validate_key(namespace):
+    if namespace.key:
+        input_key = str(namespace.key)
+        if input_key == '.' or input_key == '..' or '%' in input_key:
+            raise CLIError("Key is invalid. Key cannot be a '.' or '..', or contain the '%' character.")
```
shenmuxiaosen: why we need to check for these three cases? It makes sense to make sure key shouldn't start with feature flag prefix. But I am not sure why we have a shallow check for some characters here.

src/azure-cli/azure/cli/command_modules/appconfig/_validators.py
```diff
@@ -159,3 +162,32 @@ def validate_secret_identifier(namespace):
...
+        if input_key.startswith(FEATURE_FLAG_PREFIX):
+            raise CLIError("Key is invalid. Key cannot start with the reserved prefix for feature flags.")
+    else:
+        raise CLIError("Key cannot be empty.")
+
+
+def validate_content_type(namespace):
+    if namespace.content_type is not None:
+        if namespace.content_type == FEATURE_FLAG_CONTENT_TYPE:
+            raise CLIError("Content type is invalid. It's a reserved content type for feature flags.")
```
shenmuxiaosen: Are we case sensitive here? Or when we retrieve feature flags from config store, does the content has to be identical(case sensitive) with FEATURE_FLAG_CONTENT_TYPE?

src/azure-cli/azure/cli/command_modules/appconfig/_validators.py
```diff
@@ -159,3 +162,32 @@ def validate_secret_identifier(namespace):
...
+            raise CLIError("Content type is invalid. It's a reserved content type for KeyVault references.")
+
+
+def validate_feature(namespace):
+    if namespace.feature:
+        invalid_pattern = re.compile(r'[^a-zA-Z0-9._-]')
+        invalid = re.search(invalid_pattern, namespace.feature)
+        if invalid:
+            raise CLIError("Feature name is invalid. Only alphanumeric characters, '.', '-' and '_' are allowed.")
+    else:
```
shenmuxiaosen: We don't allow some characters even if they are escaped? Like for normal key value, if "*" or "," is part of the key, it will be accepted as long as they are escaped.

src/azure-cli/azure/cli/command_modules/appconfig/_validators.py
avanigupta: To be consistent with all the checks in portal.

---
In reply to: [362633661](https://github.com/Azure/azure-cli/pull/11753#discussion_r362633661) [](ancestors = 362633661)

src/azure-cli/azure/cli/command_modules/appconfig/_validators.py
avanigupta: You're right. I'll make it case insensitive to align with the portal.

---
In reply to: [362634100](https://github.com/Azure/azure-cli/pull/11753#discussion_r362634100) [](ancestors = 362634100)

src/azure-cli/azure/cli/command_modules/appconfig/_validators.py
avanigupta: For kv, we accept any special char except % in the name. For features, we accept only these 3 special char. Even \ is not supported for escaping.

---
In reply to: [362634642](https://github.com/Azure/azure-cli/pull/11753#discussion_r362634642) [](ancestors = 362634642)

Juliehzl:
@avanigupta please resolve conflicts first.

CredScanSuppressions.json
```diff
@@ -16,6 +16,7 @@
         "src\\azure-cli\\azure\\cli\\command_modules\\appconfig\\tests\\latest\\recordings\\test_azconfig_feature_filter.yaml",
         "src\\azure-cli\\azure\\cli\\command_modules\\appconfig\\tests\\latest\\recordings\\test_azconfig_import_export_naming_conventions.yaml",
         "src\\azure-cli\\azure\\cli\\command_modules\\appconfig\\tests\\latest\\recordings\\test_azconfig_import_export.yaml",
+        "src\\azure-cli\\azure\\cli\\command_modules\\appconfig\\tests\\latest\\recordings\\test_azconfig_key_validation.yaml",
```
Juliehzl: @haroldrandom please look at the credential scan issue.

haroldrandom: image
Test PR: https://github.com/Azure/azure-cli/pull/11911/checks?check_run_id=400111935
Build Raw Log: https://dev.azure.com/azure-sdk/29ec6040-b234-4e31-b139-33dc4287b756/_apis/build/builds/239104/logs/193

Credential Suppression Confirmed.

PR #11773, opened 2020-01-04
[AppConfig] Support import/export of keyvault ref from appservice

avanigupta:
---


yonzhan:
add to S164.

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -300,7 +298,7 @@ def set_keyvault(cmd,
     connection_string = resolve_connection_string(cmd, name, connection_string)
     azconfig_client = AzconfigClient(connection_string)
 
-    keyvault_ref_value = json.dumps({"uri": secret_identifier}, ensure_ascii=False)
+    keyvault_ref_value = json.dumps({"uri": secret_identifier}, ensure_ascii=False, separators=(',', ':'))
```
shenmuxiaosen: >separators [](start = 84, length = 10)

Why we care about the separator.  Are we strict about the indent and whitespace?

avanigupta: This is just to make CLI consistent with portal. json dump adds extra space after : , while portal creates compact json without any whitespaces.

---
In reply to: [363912684](https://github.com/Azure/azure-cli/pull/11773#discussion_r363912684) [](ancestors = 363912684)

shenmuxiaosen:
Add release note in history.rst

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/test_appconfig_commands.py
```diff
@@ -393,6 +393,99 @@ def test_azconfig_import_export(self, resource_group, location):
         assert imported_kvs == exported_kvs
 
 
+class AppConfigAppServiceImportExportLiveScenarioTest(LiveScenarioTest):
```
Juliehzl: Because you mark this test class as LiveScenarioTest, is it expected to you? May I know the reason that you want to make it live?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -269,14 +268,44 @@ def __read_kv_from_app_service(cmd, appservice_account, prefix_to_add=""):
             cmd, resource_group_name=appservice_account["resource_group"], name=appservice_account["name"], slot=None)
         for item in settings:
             key = prefix_to_add + item['name']
-            value = item['value']
             tags = {'AppService:SlotSetting': str(item['slotSetting']).lower()} if item['slotSetting'] else {}
+            value = item['value']
+
+            # Value will look like one of the following if it is a KeyVault reference:
+            # @Microsoft.KeyVault(SecretUri=https://myvault.vault.azure.net/secrets/mysecret/ec96f02080254f109c51a1f14cdb1931)
+            # @Microsoft.KeyVault(VaultName=myvault;SecretName=mysecret;SecretVersion=ec96f02080254f109c51a1f14cdb1931)
```
shenmuxiaosen: Is the schema fixed? It is possible to have whitespace, something like "@Microsoft.KeyVault(SecretUri = [https://myvault.vault.azure.net/secrets/mysecret/ec96f02080254f109c51a1f14cdb1931](https://myvault.vault.azure.net/secrets/mysecret/ec96f02080254f109c51a1f14cdb1931))"

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -269,14 +268,44 @@ def __read_kv_from_app_service(cmd, appservice_account, prefix_to_add=""):
...
+                        secret_name = appsvc_value_dict_lower.get('secretname')
+                        secret_version = appsvc_value_dict_lower.get('secretversion')
+                        secret_identifier = "https://{0}.vault.azure.net/secrets/{1}/{2}".format(vault_name, secret_name, secret_version)
+                    try:
+                        from azure.keyvault.key_vault_id import KeyVaultIdentifier
+                        # this throws an exception for invalid format of secret identifier
+                        KeyVaultIdentifier(uri=secret_identifier)
+                        appconfig_value = json.dumps({"uri": secret_identifier}, ensure_ascii=False, separators=(',', ':'))
+                        content_type = KeyVaultConstants.KEYVAULT_CONTENT_TYPE
+                        kv = KeyValue(key=key, value=appconfig_value, tags=tags, content_type=content_type)
```
shenmuxiaosen: nit:
```
kv = KeyValue(key=key,
                        value=json.dumps({"uri": secret_identifier}, ensure_ascii=False, separators=(',', ':')),
                        tags=tags,
                        content_type=KeyVaultConstants.KEYVAULT_CONTENT_TYPE)
```

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -286,6 +315,18 @@ def __write_kv_to_app_service(cmd, key_values, appservice_account):
...
             value = kv.value
+            # If its a KeyVault ref, convert the format to AppService KeyVault ref format
+            if kv.content_type and kv.content_type.lower() == KeyVaultConstants.KEYVAULT_CONTENT_TYPE:
+                try:
+                    secret_uri = json.loads(value).get("uri")
+                    if secret_uri:
+                        value = KeyVaultConstants.APPSVC_KEYVAULT_PREFIX + '(SecretUri={0})'.format(secret_uri)
+                    else:
+                        raise ValueError("Not a valid KeyVault reference.")
+                except (AttributeError, TypeError, ValueError) as e:
```
shenmuxiaosen: nit: no need to add error message as it will be caught later and swallowed.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -269,14 +268,44 @@ def __read_kv_from_app_service(cmd, appservice_account, prefix_to_add=""):
...
             tags = {'AppService:SlotSetting': str(item['slotSetting']).lower()} if item['slotSetting'] else {}
+            value = item['value']
+
+            # Value will look like one of the following if it is a KeyVault reference:
+            # @Microsoft.KeyVault(SecretUri=https://myvault.vault.azure.net/secrets/mysecret/ec96f02080254f109c51a1f14cdb1931)
+            # @Microsoft.KeyVault(VaultName=myvault;SecretName=mysecret;SecretVersion=ec96f02080254f109c51a1f14cdb1931)
+            if value and value.lower().startswith(KeyVaultConstants.APPSVC_KEYVAULT_PREFIX.lower()):
+                try:
+                    appsvc_value_dict = dict(x.split('=') for x in value[len(KeyVaultConstants.APPSVC_KEYVAULT_PREFIX) + 1: -1].split(';'))
+                    appsvc_value_dict_lower = {k.lower(): v for k, v in appsvc_value_dict.items()}
```
shenmuxiaosen: >appsvc_value_dict_lower [](start = 20, length = 23)

nit: we can reuse appsvc_value_dict

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: In appservice documentation, they haven't mentioned anything about whitespaces. To be on the safer side, I can strip all whitespaces before parsing. 
[https://docs.microsoft.com/en-us/azure/app-service/app-service-key-vault-references#reference-syntax](https://docs.microsoft.com/en-us/azure/app-service/app-service-key-vault-references#reference-syntax)  

---
In reply to: [364445905](https://github.com/Azure/azure-cli/pull/11773#discussion_r364445905) [](ancestors = 364445905)

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: Just like whitespacing, its not clear from appservice docs whether the keys "secreturi", "vaultname", etc, are case sensitive or not. We need to convert all keys to lowercase to be case insensitive.

---
In reply to: [364481792](https://github.com/Azure/azure-cli/pull/11773#discussion_r364481792) [](ancestors = 364481792)

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: If value doesn't contain a "uri" field, or the uri value is empty string, we want to treat it as a regular key value and print this message in debugger. Since this case will not throw an exceptoin, to catch this error, we need an else block. So I can either raise a ValueError from else block or duplicate the same logger.debug statement as below.

---
In reply to: [364480722](https://github.com/Azure/azure-cli/pull/11773#discussion_r364480722) [](ancestors = 364480722)

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -36,7 +36,7 @@ def resolve_resource_group(cmd, config_store_name):
     config_store_client = cf_configstore(cmd.cli_ctx)
     all_stores = config_store_client.list()
     for store in all_stores:
-        if store.name == config_store_name:
+        if store.name.lower() == config_store_name.lower():
             # Id has a fixed structure /subscriptions/subscriptionName/resourceGroups/groupName/providers/providerName/configurationStores/storeName"
```
shenmuxiaosen: Store name is case sensitive. Although we use it as part of the DNS record, so we don't allow store name "ABC" and "abc" both to exist.

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -99,3 +99,10 @@ def is_valid_connection_string(connection_string):
             return False
         return True
     return False
+
+
+def get_store_name_from_connection_string(connection_string):
+    if is_valid_connection_string(connection_string):
+        segments = connection_string.split(';')
+        return segments[0].split('//')[1].split('.')[0]
```
shenmuxiaosen: There is work item about the connection string validation in CLI. As of now, we expect the connection string to exactly like "Endpoint=abv.azconfig.io;Id=xxx;Secret=xxx". The order of endpoint, id and secret is fixed, which doesn't need to. So for this helper method, we can ignore the order, split by ";" and convert the three segments to dictionary. 

```
segments= dict(seg.split("=", 1) for seg in connection.split(";"))
endpoint = segments.get("Endpoint")
if not endpoint:
  return None
store = endpoint.split("//")[1].split('.')[0]
```

avanigupta: done

---
In reply to: [364503307](https://github.com/Azure/azure-cli/pull/11773#discussion_r364503307) [](ancestors = 364503307)

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: done

---
In reply to: [364446718](https://github.com/Azure/azure-cli/pull/11773#discussion_r364446718) [](ancestors = 364446718)

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
shenmuxiaosen: I see. Didn't notice you also log the exception in line 333. Then I am fine with throwing Value Error here. 

---
In reply to: [364498010](https://github.com/Azure/azure-cli/pull/11773#discussion_r364498010) [](ancestors = 364498010,364480722)

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
shenmuxiaosen: Yes, just we don't need to assign it to another variable called appsvc_value_dict_lower. 
```
appsvc_value_dicts = {k.lower(): v for k, v in appsvc_value_dict.items()}
```
Really minor comment. Up to you. 

---
In reply to: [364496405](https://github.com/Azure/azure-cli/pull/11773#discussion_r364496405) [](ancestors = 364496405,364481792)

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: On second thoughts, logging the error in else block will be more efficient than throwing exception and then logging same error through except block.

---
In reply to: [364925012](https://github.com/Azure/azure-cli/pull/11773#discussion_r364925012) [](ancestors = 364925012,364498010,364480722)

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: Makes sense. Changed variable name.

---
In reply to: [364925760](https://github.com/Azure/azure-cli/pull/11773#discussion_r364925760) [](ancestors = 364925760,364496405,364481792)

CredScanSuppressions.json
```diff
@@ -11,6 +11,7 @@
     },
     {
       "file": [
+        "src\\azure-cli\\azure\\cli\\command_modules\\appconfig\\tests\\latest\\recordings\\test_appconfig_to_appservice_import_export.yaml",
```
haroldrandom: Could you tell which secret/password is against the CredScan?

avanigupta: my bad, cred scan suppression is not needed for this test because its a livescenario test. I'll remove it.

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/test_appconfig_commands.py
avanigupta: These tests only pass when we run them live. We are fine with making this LiveScenarioTest because we don't expect major code changes in this area.

Issue #9392, opened 2019-05-14
Cannot re-record AKS tests


tjprescott:
Whilst re-recording tests for network API version 2019-02-01, several AKS tests failed to re-record. Since these are also failing in the daily live test run, they have been marked @live_only to permit the network changes to be merged.

**CONVERTED TO LIVE ONLY**
- [ ] test_aks_create_default_service
- [ ] test_aks_create_scale_with_custom_nodepool_name
- [ ] test_aks_create_default_service_without_skip_role_assignment
- [ ] test_aks_create_default_service_with_skip_role_assignment
- [ ] test_aks_create_default_service_without_SP_and_with_role_assignment
- [ ] test_aks_create_default_service_with_monitoring_addon
- [ ] test_aks_create_service_no_wait

mboersma:
@zqingqing1 @GaneshaThirumurthi @samkreter could someone on the AKS team please take a look?

tjprescott:
Updated to add test_aks_create_scale_with_custom_nodepool_name

Juliehzl:
Remember [here](https://github.com/Azure/azure-cli/blob/2361360c79de792e3f68a0598cb66ca14cc3616d/src/azure-cli/azure/cli/command_modules/acs/tests/latest/test_aks_commands.py#L281) and [here.](https://github.com/Azure/azure-cli/blob/2361360c79de792e3f68a0598cb66ca14cc3616d/src/azure-cli/azure/cli/command_modules/acs/tests/latest/test_aks_commands.py#L326). Also need to re-record because there are many conflicts in [test_aks_create_default_service_with_skip_role_assignment.yaml ](https://github.com/Azure/azure-cli/blob/2361360c79de792e3f68a0598cb66ca14cc3616d/src/azure-cli/azure/cli/command_modules/acs/tests/latest/recordings/test_aks_create_default_service_with_skip_role_assignment.yaml#L4) and [test_aks_create_default_service_without_skip_role_assignment.yaml](https://github.com/Azure/azure-cli/blob/2361360c79de792e3f68a0598cb66ca14cc3616d/src/azure-cli/azure/cli/command_modules/acs/tests/latest/recordings/test_aks_create_default_service_without_skip_role_assignment.yaml#L4)

Juliehzl:
@jluk Can you have a look at this?

jluk:
[...]

PR #11778, opened 2020-01-06
fix:  breaking change in enabling monitoring for OpenShift managed cluster

ganga1980:
This PR has the fix to revert the name change to workspace_resource_id for the OpenShiftManagedClusterMonitorProfile.
---


src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -3202,7 +3202,7 @@ def openshift_create(cmd, client, resource_group_name, name,  # pylint: disable=
             workspace_id = '/' + workspace_id
         if workspace_id.endswith('/'):
             workspace_id = workspace_id.rstrip('/')
-        monitor_profile = OpenShiftManagedClusterMonitorProfile(enabled=True, workspace_id=workspace_id)  # pylint: disable=line-too-long
+        monitor_profile = OpenShiftManagedClusterMonitorProfile(enabled=True, workspace_resource_id=workspace_id)  # pylint: disable=line-too-long
```
MyronFanQiu: Is this a breaking change? I don't think it's a breaking change.

troy0820:
I previously wrote a test for this command to ensure the flag `workspace-id` works with the OpenShiftMonitorProfile:  `src/azure-cli/azure/cli/command_modules/acs/tests/latest/test_osa_commands.py`.  

If we are changing it back to `workspace_resource_id` the test should reflect that.

MyronFanQiu:
@troy0820 @ganga1980 Hello. Is there any reason to mark the test as `live_only()`? 
https://github.com/Azure/azure-cli/blob/fd1c96077c8dd57b168439b6280009d84d40c7f5/src/azure-cli/azure/cli/command_modules/acs/tests/latest/test_osa_commands.py#L157-L160
- I don't think it's a breaking change for users. You can check and modify the title.
- You changed the invoking logic of python sdk and didn't change the python sdk. However, this PR passed the CI system. That's the reason I think before we don't have test. This test is decorated as `live_only()` which will not be run in CI system. If you don't have special reason to mark it as `live_only()`, could we recording the test and include the recording files in this PR?

src/azure-cli/azure/cli/command_modules/acs/custom.py
ganga1980: > Is this a breaking change? I don't think it's a breaking change.
> yes, this caused enabling of monitoring failed. Monitor profile should have the enabled=true and workspace_resource_id=<workspace_resource_id> since workspace_id set instead of workspace_resource_id so the Monitor Profile in the request has only enabled flag set not the workspace_resource_id. because of this RP throwing 400 bad request.

ganga1980:
> I previously wrote a test for this command to ensure the flag `workspace-id` works with the OpenShiftMonitorProfile: `src/azure-cli/azure/cli/command_modules/acs/tests/latest/test_osa_commands.py`.
> 
> If we are changing it back to `workspace_resource_id` the test should reflect that.
> @troy0820 , I am not changing back to workspace_resource_id and just changing 
from
 monitor_profile = OpenShiftManagedClusterMonitorProfile(enabled=True, workspace_resource_id=workspace_id)

 monitor_profile = OpenShiftManagedClusterMonitorProfile(enabled=True, workspace_resource_id=workspace_id)

> I previously wrote a test for this command to ensure the flag `workspace-id` works with the OpenShiftMonitorProfile: `src/azure-cli/azure/cli/command_modules/acs/tests/latest/test_osa_commands.py`.
> 
> If we are changing it back to `workspace_resource_id` the test should reflect that.

> @troy0820, just to clear,  the change is

from
 monitor_profile = OpenShiftManagedClusterMonitorProfile(enabled=True, workspace_id=workspace_id)  

 monitor_profile = OpenShiftManagedClusterMonitorProfile(enabled=True, workspace__resource_id=workspace_id)

ganga1980:
> @troy0820 @ganga1980 Hello. Is there any reason to mark the test as `live_only()`?
> https://github.com/Azure/azure-cli/blob/fd1c96077c8dd57b168439b6280009d84d40c7f5/src/azure-cli/azure/cli/command_modules/acs/tests/latest/test_osa_commands.py#L157-L160
> 
> * I don't think it's a breaking change for users. You can check and modify the title.
> @MyronFanQiu,  This fails below error with released version of the CLI 2.0.78 
----------------------------------------------------------
Operation failed with status: 'Bad Request'. Details: An error has occurred in subscription 99f960ec-3d95-46c2-a55c-3c11c37fd45b, resourceGroup: aro-cluster-mon-rg request: [properties.monitorProfile.workspaceResourceID cannot be empty if properties.monitorProfile.enabled = true].
----------------------------------------------------------
> * You changed the invoking logic of python sdk and didn't change the python sdk. However, this PR passed the CI system. That's the reason I think before we don't have test. This test is decorated as `live_only()` which will not be run in CI system. If you don't have special reason to mark it as `live_only()`, could we recording the test and include the recording files in this PR?
> @troy0820 , I see you have added below comment on this test 
# It works in --live mode but fails in replay mode.get rid off @live_only attribute once this resolved
is this a still issue in replay mode?

troy0820:
If that is the only change, _(workspace_id=workspace_id => workspace_resource_id=workspace_id)_, I'm okay with that to align with the python sdk.  We should take out the decorator `live_only()` to allow testing for this command in CI and validate the test works.  @ganga1980 if you could change that test in this PR, that would help out a lot.

ganga1980:
> > @troy0820 @ganga1980 Hello. Is there any reason to mark the test as `live_only()`?
> > https://github.com/Azure/azure-cli/blob/fd1c96077c8dd57b168439b6280009d84d40c7f5/src/azure-cli/azure/cli/command_modules/acs/tests/latest/test_osa_commands.py#L157-L160
> > 
> > 
> > * I don't think it's a breaking change for users. You can check and modify the title.
> >   @MyronFanQiu,  This fails below error with released version of the CLI 2.0.78
> 
> ## Operation failed with status: 'Bad Request'. Details: An error has occurred in subscription 99f960ec-3d95-46c2-a55c-3c11c37fd45b, resourceGroup: aro-cluster-mon-rg request: [properties.monitorProfile.workspaceResourceID cannot be empty if properties.monitorProfile.enabled = true].
> > * You changed the invoking logic of python sdk and didn't change the python sdk. However, this PR passed the CI system. That's the reason I think before we don't have test. This test is decorated as `live_only()` which will not be run in CI system. If you don't have special reason to mark it as `live_only()`, could we recording the test and include the recording files in this PR?
> >   @troy0820 , I see you have added below comment on this test
> 
> # It works in --live mode but fails in replay mode.get rid off @live_only attribute once this resolved
> is this a still issue in replay mode?
>


> @troy0820 @ganga1980 Hello. Is there any reason to mark the test as `live_only()`?
> https://github.com/Azure/azure-cli/blob/fd1c96077c8dd57b168439b6280009d84d40c7f5/src/azure-cli/azure/cli/command_modules/acs/tests/latest/test_osa_commands.py#L157-L160
> 
> * I don't think it's a breaking change for users. You can check and modify the title.
> * You changed the invoking logic of python sdk and didn't change the python sdk. However, this PR passed the CI system. That's the reason I think before we don't have test. This test is decorated as `live_only()` which will not be run in CI system. If you don't have special reason to mark it as `live_only()`, could we recording the test and include the recording files in this PR?

> @MyronFanQiu , @troy0820,  Based on the investigation and found that, these tests are marked because of this open issue https://github.com/Azure/azure-cli/issues/9392.  This issue applicable for ARO also since tests have dependency on the AAD App. So looks like until the issue #9392 resolved, these tests have in  `live_only()`. Once the #9392 issue resolved then we can remove the `live_only()` attribute and add the recordings.

MyronFanQiu:
@ganga1980 @troy0820 Hello. I cannot run the live test locally due to the error messages
```
"You cannot purchase reservation because required AAD tenant information is missing. Please ask your tenant admin to fill this form: https://aka.ms/orgprofile
```
If you confirm that the command is correct now and you can run the live test. I will merge this PR.

ganga1980:
> @ganga1980 @troy0820 Hello. I cannot run the live test locally due to the error messages
> 
> ```
> "You cannot purchase reservation because required AAD tenant information is missing. Please ask your tenant admin to fill this form: https://aka.ms/orgprofile
> ```
> 
> If you confirm that the command is correct now and you can run the live test. I will merge this PR.

> @MyronFanQiu , I have added tenantId to the command and verified test passes. And also as you have requested, I have added recorded file. can you please approve this PR?

Hi, @troy0820 , can you please review and sign off on this PR?  I think, we need to fix remaining tests also, can you please have a task to do that?

troy0820:
I can't sign off on the PR but it looks good to me.  @MyronFanQiu can you sign off on this Pull Request? All tests are now passing.

MyronFanQiu:
@troy0820 @ganga1980 Thanks for the contribution. I merged this PR.

PR #11782, opened 2020-01-06
[Storage] Update help for pattern

### LINKED ISSUE #3982: Bad checking of the `--pattern` argument
First of all, the `--pattern` argument description is giving very little
information: "used for files globbing"??  I had to try it out to see
how it's used.

Second, the way that it's implemented is probably going to be surprising
for most people.  I'm guessing that almost everyone would expect a
pattern like `x*` to match all files that begin with `x`, but it looks
like the pattern is actually matched against the full path that is
copied, and therefore a pattern like `x*` will either copy a whole
toplevel directory that starts with `x` or nothing at all if there is
none.

Related to this, it should be clear exactly *what* gets matched.  In an
attempt to avoid bogus matches in a list that iterates over a list of
globs, I prepended `*/` to each one (yes, that was a mistake) which
resulted in skipping a toplevel `index.html` file, and this is
regardless of specifying a full path as my `--source` or `cd`ing into
the directory and using a `--source .`.  So I'm guessing that this
pattern is matched against the full path *without* the prefix `--source`
directory (which is the same as the full container + `/` + blob name
*without* the part that is specified in `--destination`).  Summarizing
this in a short CLI description is going to be difficult, but it is
*very* needed.

Yet another bit that must be documented is the fact that `*` matches
*any* character, *including* `/`.  It would have been *much* better to
have `*` not match `/`s, and instead add a `**` pattern for that.

Given that there are probably scripts that rely on this behavior of
`--pattern` (at least the script that I wrote does so), a way to resolve
this would be to add some new `--file-pattern` that gets matched only
against the basename of files, and not against any directories.  Maybe
also another `--path-pattern` flag with `*` not matching `/`s, and `**`
that does.

### LINKED ISSUE #10382: Improve `--pattern` documentation with additional examples
**Resource Provider**
Azure Storage Blobs

**Description of Work Requested**
I'd like to suggest expanding the documentation around the `--pattern` parameter. Specifically, there is a lack of examples for common globbing scenarios. The tools for powerful patterns seem to be available, but the exact syntax required to achieve certain outcomes is not obvious.

I opened [this SO question](https://stackoverflow.com/questions/57713726/azure-cli-storage-blob-delete-batch-to-delete-all-blobs-excluding-two-directori) about making an exclusion pattern for the `az storage blob delete-batch` command. You can find [other folks](https://stackoverflow.com/questions/52070362/azure-cli-storage-delete-batch-with-exclude-pattern) having trouble with [different scenarios](https://stackoverflow.com/questions/56046447/azure-storage-copy-list-of-blobs).

I'm sure you can think of many example patterns that one might want to use, but here's a quick brain dump:

- Delete everything except for one specific directory
- Delete everything except for multiple specific directories
- Delete everything in a specific nested directory
- Delete all but one particular blob in a specific nested directory
- Delete everything from a specific directory that matches an extension (\*.txt)
- Delete everything from a specific directory that matches multiple extensions (\*.txt or \*.html)
- Delete everything from a specific directory that does NOT match an extension (\*.txt)
- Delete everything from a specific directory that does NOT match multiple extensions (\*.txt or \*.html)

Providing such examples in the documentation will help others to form the correct patterns for their needs.

Juliehzl:
Fix #3982, Fix #10382 
---


src/azure-cli/azure/cli/command_modules/storage/_help.py
```diff
@@ -960,6 +996,7 @@
   - name: Download files from an Azure Storage File Share to a local directory in a batch operation. (autogenerated)
     text: az storage file download-batch --account-key 00000000 --account-name MyAccount --destination . --no-progress --source /path/to/file
     crafted: true
+
```
haroldrandom: Additional new line

src/azure-cli/azure/cli/command_modules/storage/_help.py
```diff
@@ -564,7 +587,17 @@
     short-summary: The active lease id for the blob
 examples:
   - name: Upload all files that end with .py unless blob exists and has been modified since given date.
-    text: az storage blob upload-batch -d MyContainer --account-name MyStorageAccount -s directory_path --pattern *.py --if-unmodified-since 2018-08-27T20:51Z
+    text: |
+        az storage blob upload-batch -d mycontainer --account-name mystorageaccount -s <path-to-directory> --pattern *.py --if-unmodified-since 2018-08-27T20:51Z
+  - name: Upload all files in local path directory to a container named "mycontainer".
```
haroldrandom: Would `from local path` better?

src/azure-cli/azure/cli/command_modules/storage/_help.py
Juliehzl: done

src/azure-cli/azure/cli/command_modules/storage/_help.py
Juliehzl: done

src/azure-cli/azure/cli/command_modules/storage/_help.py
```diff
@@ -282,7 +282,8 @@
     short-summary: The blob container where the selected source files or blobs will be copied to.
   - name: --pattern
     type: string
-    short-summary: The pattern used for globbing files or blobs in the source. The supported patterns are '*', '?', '[seq', and '[!seq]'.
+    short-summary: The pattern used for globbing files or blobs in the source. The supported patterns are '*', '?', '[seq]', and '[!seq]'. For more information, please refer to https://docs.python.org/3.7/library/fnmatch.html.
+    long-summary: When you use '*' in --pattern, it will match any character including the the directory separator '/'.
```
haroldrandom: Directory separator on Windows is not `/`. Do you think we should mention it or don't end with `/`.

Issue #9916, opened 2019-07-11
"az resource list --tag" gives strange error


mick62:
## Describe the bug

**Command Name**
`az resource list`

**Errors:**
```
you cannot use the tag filter with other filters
```

## To Reproduce:
Steps to reproduce the behavior. Note that argument values have been redacted, as they may contain sensitive information.

- I assigned tag (key) 'customer' for 2 of my storage accounts.
- `az resource list --tag 'customer' `

## Expected Behavior
JSON text with description of the 2 resources which have tag 'customer' assigned. 

## Environment Summary
```
Windows-10-10.0.16299-SP0
Python 3.6.6
Shell: powershell.exe

azure-cli 2.0.65 *

```

tjprescott:
If you run the command with --debug, what is the first line (starts with command_arguments:)?

mick62:
Command entered:
az resource list --tag 'customer' --debug
Result:
Command arguments: ['resource', 'list', '--tag', 'customer', '--debug']...az_command_data_logger : command args: resource list --tag {} --debug
...cli.azure.cli.core.util : you cannot use the tag filter with other filters
you cannot use the tag filter with other filters...


Content of log file in ~\.azure\commands:
CMD-LOG-LINE-BEGIN 21488 | 2019-07-12 18:56:44,648 | INFO | az_command_data_logger | command args: resource list --tag {} --debug
CMD-LOG-LINE-BEGIN 21488 | 2019-07-12 18:56:45,161 | ERROR | cli.azure.cli.core.util | you cannot use the tag filter with other filters
CMD-LOG-LINE-BEGIN 21488 | 2019-07-12 18:56:45,164 | INFO | az_command_data_logger | exit code: 1

Why is it changed to "--tag {}" in the log file?


    Am Donnerstag, 11. Juli 2019, 20:42:23 MESZ hat Travis Prescott <notifications@github.com> Folgendes geschrieben:  
 
 
If you run the command with --debug, what is the first line (starts with command_arguments:)?

tjprescott:
Probably to avoid logging potentially sensitive info. If you run `az configure` do you have any defaults set like resource group?

mick62:
az configure
entered in VS Code terminal displays:
...
Your current configuration is as follows:
[cloud]
name = AzureCloud

[defaults]
output = table
location = eastus
--
entered in Web Shell displays:...
Your current configuration is as follows:

[cloud]
name = AzureCloud

[core]
first_run = yes

[...]

PR #11787, opened 2020-01-07
[ARM] Fix az resource list cannot use tag when there is a default location

zhoxing-ms:
Fix `az resource list` cannot use tag when there is a default location
Issue: #9916 
Improve the error message of the conflict between tag and other filter conditions for `az resource list` command.

---


src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -504,9 +504,24 @@ def _deploy_arm_template_subscription_scope(cli_ctx,
                        deployment_name, properties, deployment_location)
 
 
-def _list_resources_odata_filter_builder(resource_group_name=None, resource_provider_namespace=None,
+def _list_resources_odata_filter_builder(cli=None, resource_group_name=None, resource_provider_namespace=None,
                                          resource_type=None, name=None, tag=None, location=None):
     """Build up OData filter string from parameters """
+    if tag:
```
Juliehzl: try to use `if tag is not None` to specify that there is parameter value for tag

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -504,9 +504,24 @@ def _deploy_arm_template_subscription_scope(cli_ctx,
                        deployment_name, properties, deployment_location)
 
 
-def _list_resources_odata_filter_builder(resource_group_name=None, resource_provider_namespace=None,
+def _list_resources_odata_filter_builder(cli=None, resource_group_name=None, resource_provider_namespace=None,
```
Juliehzl: Pls keep parameter consistent with original function. cli is not consistent.

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -504,9 +504,24 @@ def _deploy_arm_template_subscription_scope(cli_ctx,
...
+        if cli and cli.data and isinstance(cli.data['safe_params'], list):
+            for arg_name in cli.data['safe_params']:
+                if arg_name in ['-l', '--l', '--location']:
+                    manual_input_location = True
+                    break
+        if manual_input_location is False:
+            location = None
+
+        if name or location:
+            raise IncorrectUsageError('you cannot use the tag filter with other filters')
```
Juliehzl: pls list `other filters` here

Juliehzl:
Through reconsidering this scenario, we should keep consistent for the location argument. If they don't want to utilize the build-in type, we can define location by ourselves. But if we use, users should know the influence of such location type.

src/azure-cli/azure/cli/command_modules/resource/custom.py
zhoxing-ms: Done

src/azure-cli/azure/cli/command_modules/resource/custom.py
zhoxing-ms: Done

src/azure-cli/azure/cli/command_modules/resource/custom.py
zhoxing-ms: Done

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -507,6 +507,20 @@ def _deploy_arm_template_subscription_scope(cli_ctx,
 def _list_resources_odata_filter_builder(resource_group_name=None, resource_provider_namespace=None,
                                          resource_type=None, name=None, tag=None, location=None):
     """Build up OData filter string from parameters """
+    if tag is not None:
+        if resource_group_name:
+            raise IncorrectUsageError('you cannot use the tag filter with resource group filter'
+                                      '(If the default value for resource group is set, please clear it with \'az configure\' command first)')
```
Juliehzl: use "az configure" command

zhoxing-ms: Done

Juliehzl:
@zhoxing-ms Why do you close this PR?

Issue #9342, opened 2019-05-08
What about managedBy param ?


alexbenedek:
How can i define the managedBy param when i create the Resource Group in the cli?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 58c81c4d-10d2-fc04-e707-fcf914f4d540
* Version Independent ID: 8598f788-bd33-a4c8-ed66-593e0fde75d5
* Content: [az group](https://docs.microsoft.com/en-us/cli/azure/group?view=azure-cli-latest#feedback)
* Content Source: [src/command_modules/azure-cli-resource/azure/cli/command_modules/resource/_help.py](https://github.com/Azure/azure-cli/blob/dev/src/command_modules/azure-cli-resource/azure/cli/command_modules/resource/_help.py)
* Service: **azure-resource-manager**
* GitHub Login: @rloutlaw
* Microsoft Alias: **routlaw**

yugangw-msft:
@kristapratico, this is a simple fix towards `az group create` by exposing a new argument.

tjprescott:
This parameter may be read-only. It would be worth contacting ARM to be sure (i.e. I would *not* look to Swagger for truth unless it says it is)

yonzhan:
add to S164.

zhoxing-ms:
@alexbenedek Why would you want to add the managedBy param when creating a Resource Group?

zhoxing-ms:
Could you tell me the specific meaning of the `managedBy` parameter in the `Resource Groups - Create Or Update` service? @Tiano2017 
doc link: [https://docs.microsoft.com/zh-cn/rest/api/resources/resourcegroups/createorupdate](url)

Tiano2017:
@dandanwang0320 to help.

zhoxing-ms:
@dandanwang0320 Please take a look. I don't think the description of the `--managedBy` parameter in the document is clear what type the ID it is. Could you give a specific example for it?

haymansfield:
I'm not sure why this is closed.

Managed by doesn't appear to be an available parameter on the cli yet is still in the documentation.

zhoxing-ms:
@haymansfield Please use `pip install` to download the latest version of azure cli (2.0.81) and have a try.

dcbrown16:
[...]

PR #11796, opened 2020-01-08
[ARM] Add new parameter to support adding managedBy information for az group create

zhoxing-ms:
Add new parameter `--managed-by` to support adding managedBy information for command `az group create`
Issue: #9342 

---


yonzhan:
add to S164.

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -221,6 +221,7 @@ def load_arguments(self, _):
 
     with self.argument_context('group create') as c:
         c.argument('rg_name', options_list=['--name', '--resource-group', '-n', '-g'], help='name of the new resource group', completer=None)
+        c.argument('managed_by', min_api='2016-09-01', help='The ID of the resource that manages this resource group.', resource_type=ResourceType.MGMT_RESOURCE_RESOURCES)
```
Juliehzl: You don't need to specify the resource_type here. It already registered when module initialization.

Juliehzl: What does the ID mean? resource id , guid or something else?

zhoxing-ms: OK, I'll make it registered with the initialized __init__.py

zhoxing-ms: This comment is from the documentation of the Rest API, because it's exact meaning has not been found yet.

Issue #11780, opened 2020-01-06
get_raw_token should take tenant name/id as input


mit2nil:
Current interface as shown below takes only subscription and resource as inputs. However, auth token are specific to a tenant and not a subscription. Lack of tenant input also complicates design for scenarios requiring auth token for tenants with no subscription. 

`def get_raw_token(self, resource=None, subscription=None):`

bim-msft:
@jiasli Please take a look.

yonzhan:
add to S165.

mit2nil:
> add to S165.
@jiasli @yonzhan Is it possible to pull this into S164?

yonzhan:
@mit2nil  add to S164.

Issue #11351, opened 2019-11-22
Add --tenant option to the az account get-access-token command 


jiasli:
Add `--tenant` option to `az account get-access-token`

PR #11798, opened 2020-01-08
[Core][Profile] Add `tenant` parameter to `get_raw_token`

jiasli:
Inspired by #11351, #11780

**In cli core**: `get_raw_token`: Add `tenant` parameter to acquire token for the tenant directly, needless to specify a subscription. The structure of the return value is not changed - a `(cred, subscription, tenant)` tuple, but the second element of the tuple - `subscription` is set to `None` for tenant token for backward compatibility. 

**In command module**: `az account get-access-token`: Add `--tenant` parameter to acquire token for the tenant directly, needless to specify a subscription. In that case, the returned JSON doesn't have `subscription` field.

```jsonc
{
  "accessToken": "eyJ0e...",
  "expiresOn": "2020-01-08 17:09:52.554667",
  "tenant": "72f988bf-86f1-41af-91ab-2d7cd011db47",
  "tokenType": "Bearer"  // no subscription if --tenant is specified**
}
```

The returned token can be decoded at https://jwt.ms/ for verification. This is a Microsoft website so there is no security concern to paste the token there. Check the `tid` field for the tenant ID. 

`tenant` is only available for user and service principal account, not for MSI or Cloud Shell account, because MSI or Cloud Shell account is a single-tenant account and it is meaningless to explicitly specify a tenant. We error it out for simplicity, otherwise we need to check whether the specified tenant is the home tenant of the account.

To test:

* Test help: `az account get-access-token -h`
* Test with user identity: 
    ```sh
    az login
    account get-access-token --tenant {}
    ```
* Test with service principal:
    ```sh
    az ad sp create-for-rbac
    az login --service-principal -u {} -p {} -t {}
    az account get-access-token --tenant {}
    ```
* Test with MSI (this will fail!): create a VM with MSI and setup the dev environment, then run 
    ```sh
    az login --identity
    az account get-access-token --tenant {}
    ```
* Cloud Shell can't be tested currently, so unit test `test_get_raw_token_in_cloud_console` is added to mock Cloud Shell

---


jiasli:
@mit2nil please kindly help review and test this PR.

jiasli:
@fengzhou-msft @haroldrandom, there seems to be some issue with the CI. The title is for Python 3.8 but it is still using 2.7 internally: https://dev.azure.com/azure-sdk/public/_build/results?buildId=227317&view=logs&jobId=4ac732a2-b9e4-5fde-3ff2-aac0b888c429&j=4ac732a2-b9e4-5fde-3ff2-aac0b888c429&t=7defb86d-ca26-5ebc-39fe-803ea5c317dd

```
============================= test session starts ==============================
platform linux2 -- Python 2.7.12, pytest-4.6.9, py-1.8.1, pluggy-0.13.1
cachedir: .tox/py27/.pytest_cache
rootdir: /home/vsts/work/1/s/src/azure-cli-core
collected 269 items
```

haroldrandom:
> @fengzhou-msft @haroldrandom, there seems to be some issue with the CI. The title is for Python 3.8 but it is still using 2.7 internally: https://dev.azure.com/azure-sdk/public/_build/results?buildId=227317&view=logs&jobId=4ac732a2-b9e4-5fde-3ff2-aac0b888c429&j=4ac732a2-b9e4-5fde-3ff2-aac0b888c429&t=7defb86d-ca26-5ebc-39fe-803ea5c317dd
> 
> ```
> ============================= test session starts ==============================
> platform linux2 -- Python 2.7.12, pytest-4.6.9, py-1.8.1, pluggy-0.13.1
> cachedir: .tox/py27/.pytest_cache
> rootdir: /home/vsts/work/1/s/src/azure-cli-core
> collected 269 items
> ```
> 
> `assertRaisesRegex` is the modern name. I need to use the deprecated version `assertRaisesRegexp` to pass the CI. Maybe we can deprecate Python 2.7 support as well? Given Python 2.7 is already giving the warning:
> 
> > py27 installed: DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date.

Fixed here: https://github.com/Azure/azure-cli/pull/11802

Reason: 
>tox can find system-wide Python which is Python 2.7 in ubuntu 16.04.
Even though CI job "Unit Test for Core and Telemetry" can run pre-set Python version to test but also run an extra test on system-wide Python which is Python 2.7

So I add environment variable to let tox use only the specific Python version I want.

But still need to check the schedule of retiring Python 2.7 with @fengzhou-msft to verify whether need to add Python 2.7 support or not.

src/azure-cli/azure/cli/command_modules/profile/__init__.py
```diff
@@ -75,6 +75,7 @@ def load_arguments(self, command):
 
         with self.argument_context('account get-access-token') as c:
             c.argument('resource_type', get_enum_type(cloud_resource_types), options_list=['--resource-type'], arg_group='', help='Type of well-known resource.')
+            c.argument('tenant', options_list=['--tenant', '-t'], is_preview=True, help='Tenant ID for which the token is acquired. Only available for user and service principal account, not for MSI or Cloud Shell account')
```
jsntcy: why do not we support MSI?

src/azure-cli-core/azure/cli/core/tests/test_profile.py
```diff
@@ -716,6 +716,17 @@ def test_get_raw_token(self, mock_get_token, mock_read_cred_file):
         self.assertEqual(sub, '1')
         self.assertEqual(tenant, self.tenant_id)
 
+        # Test get_raw_token with tenant
+        creds, sub, tenant = profile.get_raw_token(resource='https://foo', tenant=self.tenant_id)
+
+        self.assertEqual(creds[0], self.token_entry1['tokenType'])
```
jsntcy: Can we also add tests for MSI and cloud shell? (check exception)

jiasli: Yes. It is already in this file.

src/azure-cli/azure/cli/command_modules/profile/__init__.py
jiasli: Please see the PR description.

jiasli: The explanation is in the PR description.

mit2nil:
> @mit2nil please kindly help review and test this PR.

@jiasli I was able to verify the change for our AOBO and CSP use cases.

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -587,31 +587,42 @@ def get_refresh_token(self, resource=None,
         sp_secret = self._creds_cache.retrieve_secret_of_service_principal(username_or_sp_id)
         return username_or_sp_id, sp_secret, None, str(account[_TENANT_ID])
 
-    def get_raw_token(self, resource=None, subscription=None):
+    def get_raw_token(self, resource=None, subscription=None, tenant=None):
+        if subscription and tenant:
+            raise CLIError("Please specify only one of subscription and tenant, not both")
```
qianwens: this error message pattern seems not align with existing pattern like "subscription shouldn't e specified when tenant is specified"

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -587,31 +587,42 @@ def get_refresh_token(self, resource=None,
...
             creds = self._get_token_from_cloud_shell(resource)
-
-        elif user_type == _USER:
-            creds = self._creds_cache.retrieve_token_for_user(username_or_sp_id,
-                                                              account[_TENANT_ID], resource)
         else:
-            creds = self._creds_cache.retrieve_token_for_service_principal(username_or_sp_id,
-                                                                           resource,
-                                                                           account[_TENANT_ID])
+            tenant_dest = tenant if tenant else account[_TENANT_ID]
```
qianwens: when subscription is specified, it is checked if exist in cached file in function "get_subscription", is there anyway we can have a pre-check before send retrieve token request?

src/azure-cli-core/azure/cli/core/_profile.py
jiasli: This expression is used by many other modules. You may Find All with "not both" to see other usage.

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -587,31 +587,42 @@ def get_refresh_token(self, resource=None,
         sp_secret = self._creds_cache.retrieve_secret_of_service_principal(username_or_sp_id)
         return username_or_sp_id, sp_secret, None, str(account[_TENANT_ID])
 
-    def get_raw_token(self, resource=None, subscription=None):
+    def get_raw_token(self, resource=None, subscription=None, tenant=None):
+        if subscription and tenant:
+            raise CLIError("Please specify only one of subscription and tenant, not both")
         account = self.get_subscription(subscription)
```
qianwens: if subscription is not specified, account is empty? then can we get the user_type below?

jiasli: If subscription is not specified, it retrieves the `default` subscription which is the one that is currently selected by `az account set`.

src/azure-cli/azure/cli/command_modules/profile/custom.py
```diff
@@ -59,20 +59,20 @@ def show_subscription(cmd, subscription=None, show_auth_for_sdk=None):
     print(json.dumps(profile.get_sp_auth_info(subscription), indent=2))
 
 
-def get_access_token(cmd, subscription=None, resource=None, resource_type=None):
-    '''
+def get_access_token(cmd, subscription=None, resource=None, resource_type=None, tenant=None):
+    """
     get AAD token to access to a specified resource
     :param resource: Azure resource endpoints. Default to Azure Resource Manager
     :param resource-type: Name of Azure resource endpoints. Can be used instead of resource.
```
qianwens: please also add tenent in comment

src/azure-cli/azure/cli/command_modules/profile/__init__.py
arrownj: do we need to provide an example for this parameter ?

src/azure-cli/azure/cli/command_modules/profile/custom.py
jiasli: The doc string is overridden by https://github.com/Azure/azure-cli/blob/fbfe69172600936798387e2b5c9c1c9887bb38f4/src/azure-cli/azure/cli/command_modules/profile/__init__.py#L78
I'd prefer to omit here to eliminate duplication.

src/azure-cli-core/azure/cli/core/_profile.py
jiasli: For a **user** account we don't need to check if the tenant is in `az account list`, as the `refreshToken` in `accessTokens.json` is the same for all tenants. **user : refresh token** is a 1:1 mapping. The refresh token can be used to retrieve a token in ANY tenant, even if the user is not a member of that tenant. The token doesn't has any permission in that tenant. 

A **service principal** is a binding between an **app** and a **tenant**, in other words, a projection of an **app** in a **tenant**. 

In `accessTokens.json`, the entry for a service principal account is:

```
{
    "servicePrincipalId": "6827...",
    "servicePrincipalTenant": "5482...",
    "accessToken": "43b2..."
}
```

`servicePrincipalId` is usually the `appId` returned by `az ad sp create-for-rbac`.

`accessToken` is actually the credential(password or certificate) of the service principal or app. If the credential is at app level, it is the same across tenants; if the credential is at service principal level, it is different across tenants. So **service principal : access token(credential)** is a 1:+ mapping. 

Even if for app credential, **service principal : access token(credential)** is a 1:1 mapping, we can't tell whether the credential is valid in other tenants until the AAD server returns error. I prefer not to do the bet. So to retrieve the credential, we need to use both `servicePrincipalId` and `servicePrincipalTenant`, which is searched at 

https://github.com/Azure/azure-cli/blob/fbfe69172600936798387e2b5c9c1c9887bb38f4/src/azure-cli-core/azure/cli/core/_profile.py#L940-L941

Therefore, for a Service Principal to work across tenants, we need to run `az login` twice for each tenant.

jiasli:
I have changed the behavior of `get_raw_token` for service principal account. 

The current behavior of CLI is to search for the credential from local cache (`~/.azure/accessTokens.json`) using a `(servicePrincipalId, servicePrincipalTenant)` combination. But **if the corresponding app of the service principal is a multi-tenant app and the credential is at [app level](https://docs.microsoft.com/en-us/previous-versions/azure/ad/graph/api/entity-and-complex-type-reference#application-entity), we can safely use the credential from the logged-in tenant to retrieve token for another tenant**. 

So what the changed logic does is it first searches for the credential from local cache using `servicePrincipalId`. If that is not found, error out. If found, filter again with tenant. If found, use that credential. If not, display a warning and fallback to the credential for the logged-in tenant. 

For example, say app1 is a multi-tenant app registered in tenant t1 and t2. The user runs `az login -u {app1Id} -p {} -t {t1}`. Then when the user runs `az account get-access-token -t {t2}`, CLI will search the local cache and found `(app1, t2)` is not present, so it warns that 

> Could not retrieve credential from local cache for service principal {app1Id} under tenant {t2}. Trying credential under tenant {t1}, assuming that is an app credential.

The token can be successfully retrieved and you can verify the tenant ID `tid` with https://jwt.ms/

The only situation where this won't work is if the credential is at [service principal level](https://docs.microsoft.com/en-us/previous-versions/azure/ad/graph/api/entity-and-complex-type-reference#serviceprincipal-entity), which is bound to a specified tenant and only works in that tenant. CLI currently doesn't support that. Related PR is https://github.com/Azure/azure-cli/pull/11466, but we haven't reached an agreement about this breaking change. 


ℹ Remark: Multi-tenant app can't be created using CLI and can only be done in Azure Portal. After creating the app in Azure Portal, use `az account set -s {}` to switch to another tenant where you want to create a service principal and run `az ad sp create --id {appId}` to register the app under this tenant as a service principal.

src/azure-cli/azure/cli/command_modules/profile/__init__.py
jiasli: @arrownj, good point. Added as requested.

jiasli:
Reply to https://github.com/Azure/azure-cli/pull/11798#pullrequestreview-343719541. @Juliehzl, good point. Added as requested.

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -587,31 +587,42 @@ def get_refresh_token(self, resource=None,
...
+                                                                  tenant_dest, resource)
+            else:
+                # Service Principal
+                creds = self._creds_cache.retrieve_token_for_service_principal(username_or_sp_id,
+                                                                               resource,
+                                                                               tenant_dest)
         return (creds,
-                str(account[_SUBSCRIPTION_ID]),
-                str(account[_TENANT_ID]))
+                str('N/A' if tenant else account[_SUBSCRIPTION_ID]),
```
jsntcy: is it better not to show subscription instead of 'N/A' in result? As it's confused for users to understand what's 'N/A' meaning, does it mean no sub or just don't show sub? 'N/A' makes no sense to users.

jiasli: You may search all in the project to find the usage of `get_raw_token`. It is currently used all over the place. I prefer not to change the data structure of the returned value. Instead, I will update `az account get-access-token`'s output to incorporate this proposal. Thanks for it.

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -587,31 +587,42 @@ def get_refresh_token(self, resource=None,
...
+                                                                  tenant_dest, resource)
+            else:
+                # Service Principal
+                creds = self._creds_cache.retrieve_token_for_service_principal(username_or_sp_id,
+                                                                               resource,
+                                                                               tenant_dest)
         return (creds,
-                str(account[_SUBSCRIPTION_ID]),
-                str(account[_TENANT_ID]))
+                None if tenant else str(account[_SUBSCRIPTION_ID]),
```
arrownj: why change previous 'N/A' to None here ?

src/azure-cli/azure/cli/command_modules/profile/custom.py
```diff
@@ -59,27 +59,30 @@ def show_subscription(cmd, subscription=None, show_auth_for_sdk=None):
...
+
+    result = {
         'tokenType': creds[0],
         'accessToken': creds[1],
         'expiresOn': creds[2].get('expiresOn', 'N/A'),
-        'subscription': subscription,
         'tenant': tenant
     }
+    if subscription:
+        result['subscription'] = subscription
```
arrownj: Seems we changed previous behavior here. `subscription` property value is N/A before, after this change, the property will be removed. Is this a break change to some customers ?

jiasli: Nope. Previously there is no way for the user to get a token for a tenant, so this is a new feature rather than a breaking change.

src/azure-cli-core/azure/cli/core/_profile.py
jiasli: This is implemented according to https://github.com/Azure/azure-cli/pull/11798#discussion_r368358503. For the upper level `az account get-access-token --tenant` command, `subscription` property isn't included.

src/azure-cli/azure/cli/command_modules/profile/custom.py
jiasli: Nope. Previously there is no way for the user to get a token for a tenant, so this is a new feature rather than a breaking change. 

The previous behavior for `az account get-access-token --subscription` is not affected. This can be verified by the test cases.

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -926,12 +937,21 @@ def retrieve_token_for_user(self, username, tenant, resource):
...
+                           "Please run 'az login' for this service principal."
+                           .format(sp_id))
+        matched_with_tenant = [x for x in matched if tenant == x[_SERVICE_PRINCIPAL_TENANT]]
+        if matched_with_tenant:
+            cred = matched_with_tenant[0]
+        else:
+            logger.warning("Could not retrieve credential from local cache for service principal %s under tenant %s. "
+                           "Trying credential under tenant %s, assuming that is an app credential.",
+                           sp_id, tenant, matched[0][_SERVICE_PRINCIPAL_TENANT])
+            cred = matched[0]
```
arrownj: So if I only login with tenant A, and run `az account get-access-token -t B`, it will return an access token of A, right ?

jiasli: Nope. It will return the token for tenant B automatically using the credential for tenant A. That's why it is called single-sign-on. +@qianwens for awareness

mit2nil:
@jiasli Is this going to be merged and be part of next release? We need to change our cli extension PR based on the timelines of release of this feature.

jiasli:
> @jiasli Is this going to be merged and be part of next release? We need to change our cli extension PR based on the timelines of release of this feature.

Yes. We will deliver it in this release.

mattduguid:
thanks this helped solve following error for us,
ValidationError: Could not retrieve credential from local cache for service principal ***REDACTED***. Please run 'az login' for this service principal.

we received that error when logging in from azure devops pipeline using,
az login --service-principal -u ***REDACTED*** -p ***REDACTED*** --tenant ***REDACTED*** --allow-no-subscriptions

jiasli:
Hi @mattduguid, glad to know it is working for your, but I didn't really get your point. `--tenant` has been there for some time and we plan to GA it soon. Could you describe why the message "argument --tenant is in preview, it may be removed/changed in a future release" should be kept?

mattduguid:
> Hi @mattduguid, glad to know it is working for your, but I didn't really get your point. `--tenant` has been there for some time and we plan to GA it soon. Could you describe why the message "argument --tenant is in preview, it may be removed/changed in a future release" should be kept?

Sorry meant keep the —tenant parameter the message can go, the way it was worded made me think it may not be kept 😉

jiasli:
Haha, I got it. Sure we will definitely keep `--tenant`.

PR #11800, opened 2020-01-08
[SQL] az sql db audit-policy: Fix for empty audit actions and groups

ranisha2:
In case the policy is enabled and there are no audit actions and groups (Was overridden when disabling the policy, or removed while enabling), set the default value

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/sql/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/sql/custom.py
@@ -1434,6 +1434,14 @@ def db_audit_policy_update(
     if audit_actions_and_groups:
         instance.audit_actions_and_groups = audit_actions_and_groups
 
+    # If auditing is enabled, make sure that the actions and groups are set to default
+    # value in case they were removed previously (When disabling auditing)
+    if enabled and (not instance.audit_actions_and_groups or instance.audit_actions_and_groups == []):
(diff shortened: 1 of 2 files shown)


jsntcy:
@Juliehzl, please help review it, thanks.

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -1432,6 +1432,14 @@ def db_audit_policy_update(
     if audit_actions_and_groups:
         instance.audit_actions_and_groups = audit_actions_and_groups
 
+    # If auditing is enabled, make sure that the actions and groups are set to default
+    # value in case they were removed previously (When disabling auditing)
+    if enabled and (not instance.audit_actions_and_groups or instance.audit_actions_and_groups == []):
```
Juliehzl: Can we add test for this scenario?

Juliehzl:
Add @jaredmoo to take a look.

jaredmoo:
Why add new logic on client side, instead of updating backend behaviour?

Juliehzl:
> Why add new logic on client side, instead of updating backend behaviour?

There are two concerns here:
1. I am not sure about what is your default value

> Why add new logic on client side, instead of updating backend behaviour?

make sense.

jaredmoo:
I still don't approve of this yet.

PR #11843, opened 2020-01-14
[Rest] Make --method -m optional

jiasli:
`--method -m` in `az rest`  is self-contradictory. It has default value but is `[Required]`.
```
--method -m      [Required] : HTTP request method.  Allowed values: delete, get, head, options,
                              patch, post, put.  Default: get.
```

This is fixed by making `method` optional in `rest_call`. 

To test, run 

```
> az rest -h
--method -m                 : HTTP request method.  Allowed values: delete, get, head, options,
                              patch, post, put.  Default: get.

> az rest --uri https://management.azure.com/
```

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/resource/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/resource/custom.py
@@ -2011,7 +2011,7 @@ def list_resource_links(cmd, scope=None, filter_string=None):
 # endregion
 
 
-def rest_call(cmd, method, uri, headers=None, uri_parameters=None,
+def rest_call(cmd, uri, method=None, headers=None, uri_parameters=None,
               body=None, skip_authorization_header=False, resource=None, output_file=None):
     from azure.cli.core.util import send_raw_request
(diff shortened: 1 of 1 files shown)


src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -2011,7 +2011,7 @@ def list_resource_links(cmd, scope=None, filter_string=None):
 # endregion
 
 
-def rest_call(cmd, method, uri, headers=None, uri_parameters=None,
+def rest_call(cmd, uri, method=None, headers=None, uri_parameters=None,
```
qwordy: 1. Why you swap the order?
2. Why you use method='get'?

jiasli: 1. Python requires non-default params to appear first
2. This 'get' will be overridden by https://github.com/Azure/azure-cli/blob/f93602026bb1af862eb571ecbd2e469153bba8e4/src/azure-cli/azure/cli/command_modules/resource/_params.py#L304

qwordy: reasonable

PR #11846, opened 2020-01-14
[KeyVault] Add example for container sas-definition

bim-msft:
Fix: https://github.com/Azure/azure-cli-extensions/issues/1063

---


jiasli:
`--account-key 00000000` is only used to generate a SAS definition which actually doesn't work. It is the permission part `sv=2019-02-02&ss=bfqt&srt=sco&sp=rwdlacup&se=2020-01-14T16:57:57Z&st=2020-01-14T08:57:57Z&spr=https` that matters, not the `sig=oD7M...` part. `sig` is recalculated by keyvault later. 

This reduces the permission to query storage account key via [Storage Accounts - List Keys](https://docs.microsoft.com/en-us/rest/api/storagerp/storageaccounts/listkeys) which requires RBAC write permission.

PR #11852, opened 2020-01-14
[SQL] 'az sql mi update`: Add `--tier` and `--family` parameters to change hw generation

v-nestan:
This change adds support for changing existing managed instance hardware generation.

TFS item [link](http://sqlbuvsts01:8080/Main/SQL%20Server/_workitems#_a=edit&id=13293296)


jsntcy:
@Juliehzl,  please help review it, thanks.

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -2134,6 +2134,12 @@ def _find_managed_instance_sku_from_capabilities(cli_ctx, location, sku):
     # Find family level capability, based on requested sku properties
     family_capability = _find_family_capability(sku, edition_capability.supported_families)
 
+    # Check whether hardware family is being changed to a deprecated hardware family
+    if not is_instance_create:
+        if current_sku_family == 'Gen5' and 'Gen4' in family_capability.sku:
```
jaredmoo: Why is this being done client-side instead of server-side (MI provision operation authorizer)? These sorts of policy decisions should not be coded into CLI

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -1261,6 +1261,16 @@ def _configure_security_policy_storage_params(arg_ctx):
                    'for use with key management services like Azure KeyVault. '
                    'If identity is already assigned - do nothing.')
 
+        # Create args that will be used to build up the Managed Instance's Sku object
+        create_args_for_complex_type(
+            c, 'sku', Sku, [
+                'family',
+                'name',
+                'tier',
+            ])
```
jaredmoo: Can you show the output of `az sql mi create -h` and make sure that sku-related params have same UX as `az sql db create -h`?

src/azure-cli/azure/cli/command_modules/sql/custom.py
v-nestan: The idea was to fast fail if a customer tries to change hardware family to a deprecated hardware family. Synced with PM, and we agreed to remove this check as you suggested. Thanks

src/azure-cli/azure/cli/command_modules/sql/_params.py
v-nestan: I've aligned them 

az sql mi create -h
image

az sql mi update -h
image

az sql db create -h
image

src/azure-cli/HISTORY.rst
```diff
@@ -74,6 +74,10 @@ Release History
 * Add new parameters `--enable-delete-retention` and `--delete-retention-days` to support managing delete retention policy for storage account blob-service-properties.
 
 
+**SQL**
```
Juliehzl: Please rearrange the release note in the right position.

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -1156,7 +1156,8 @@ def _configure_security_policy_storage_params(arg_ctx):
 
         c.argument('tier',
                    arg_type=tier_param_type,
-                   help='The edition component of the sku. Allowed values: GeneralPurpose, BusinessCritical.')
+                   help='The edition component of the sku. Allowed values include:'
```
Juliehzl: please use get_enum_types() for allowed values.

Juliehzl:
@v-nestan please fix CI errors and if @jaredmoo agree, I am fine to merge it.

src/azure-cli/azure/cli/command_modules/sql/_help.py
```diff
@@ -559,6 +559,8 @@
 examples:
   - name: Updates a mi with specified parameters and with identity
     text: az sql mi update -g mygroup -n myinstance -i -p mypassword --license-type mylicensetype --capacity vcorecapacity --storage storagesize
+  - name: Updates mi edition and hardware family
```
Juliehzl: Updates -> Update

src/azure-cli/azure/cli/command_modules/sql/_help.py
```diff
@@ -559,6 +559,8 @@
...
+    text: az sql mi update -g mygroup -n myinstance -tier GeneralPurpose -family Gen5
```
Juliehzl: It should be --tier and --family.

Juliehzl:
Please follow above template for your PR title in future.

src/azure-cli/HISTORY.rst
v-nestan: Fixed

src/azure-cli/azure/cli/command_modules/sql/_help.py
v-nestan: Fixed

src/azure-cli/azure/cli/command_modules/sql/_help.py
v-nestan: Corrected. Thanks

src/azure-cli/azure/cli/command_modules/sql/_params.py
v-nestan: Done. I've done the same thing for family too

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -56,7 +56,9 @@
     ComputeModelType,
     DatabaseCapabilitiesAdditionalDetails,
     ElasticPoolCapabilitiesAdditionalDetails,
-    FailoverPolicyType
+    FailoverPolicyType,
+    ManagedInstanceTiers,
```
jaredmoo: style nitpick: It should be singular, i.g. `ManagedInstanceTier` & `ManagedInstanceFamily`

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -1159,13 +1171,12 @@ def _configure_security_policy_storage_params(arg_ctx):
...
-                   help='The edition component of the sku. Allowed values: GeneralPurpose, BusinessCritical.')
+                   arg_type=mi_tier_param_type,
+                   help='The edition component of the sku.')
 
         c.argument('family',
-                   arg_type=family_param_type,
-                   help='The compute generation component of the sku. '
-                   'Allowed values include: Gen4, Gen5.')
+                   arg_type=mi_family_param_type,
+                   help='The compute generation component of the sku.')
```
jaredmoo: Does `get_enum_type()` throw error if you don't specify one of the allowed values? This will prevent forward compatibility (i.e. when we release Gen1234, customers will not be able to manually type it in until they download newer CLI version) - this has previously been a frustration for SQL DB customers. We should allow customer to type in anything, then if we validate it must be against a dynamic list (e.g. capabilities API) and NOT a hardcoded list.

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -2098,22 +2098,30 @@ def encryption_protector_update(
...
 
     logger.debug('_find_managed_instance_sku_from_capabilities input: %s', sku)
 
-    if sku.name:
-        # User specified sku.name, so nothing else needs to be resolved.
-        logger.debug('_find_managed_instance_sku_from_capabilities return sku as is')
-        return sku
-
-    if not _any_sku_values_specified(sku):
+    if is_instance_create and not _any_sku_values_specified(sku):
```
jaredmoo: Why does behaviour need to be different for create vs. update?

jaredmoo: Why is `is_instance_create` logic needed?

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -2183,14 +2191,16 @@ def managed_instance_list(
...
         instance,
         administrator_login_password=None,
         license_type=None,
         vcores=None,
         storage_size_in_gb=None,
         assign_identity=False,
         proxy_override=None,
-        public_data_endpoint_enabled=None):
+        public_data_endpoint_enabled=None,
+        sku=None):
```
jaredmoo: In the interest of alignment & reducing overall complexity, would it make sense to use same syntax as db_update function (i.e. split it up into components):
```
        tier=None,
        family=None,
        capacity=None,
```
?

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -2211,6 +2221,16 @@ def managed_instance_update(
     instance.proxy_override = (
         proxy_override or instance.proxy_override)
```
jaredmoo: Should you set `sku.name = None` here?

jaredmoo: I think it makes sense to null out the sku name here, because there is no requested sku name, and you don't yet know what the sku name will be (it will be found based on tier & family).

src/azure-cli/azure/cli/command_modules/sql/custom.py
jaredmoo: Just remembered that MI has vcores property, not capacity. Blergh

src/azure-cli/azure/cli/command_modules/sql/_params.py
v-nestan: I've checked now and it does throws an error. For example

az sql mi update --subscription "DS-CloudLifter_jovanc_R&D_60843" --resource-group 201tmpl --name nestan-testing109 --family Gen52
az sql mi update: 'Gen52' is not a valid value for '--family'. See 'az sql mi update --help'.

The most similar choice to 'Gen52' is:
        Gen5

I'll remove the get_enum_type :)

v-nestan: Removed

src/azure-cli/azure/cli/command_modules/sql/_params.py
v-nestan: I've removed those because they are not needed any more since I've removed get_enum_type()

src/azure-cli/azure/cli/command_modules/sql/custom.py
v-nestan: I'll replace sku with tier and family. I'll omit the third component of sku because it is not needed

src/azure-cli/azure/cli/command_modules/sql/custom.py
v-nestan: Yes, it makes sense to set it to `None` here

src/azure-cli/azure/cli/command_modules/sql/custom.py
v-nestan: You are right it is redundant.
The idea was to protect from getting a default sku name  when sku components are not entered during update SLO

However, since SKU for existing instance is populated before this method is called, default sku will not be taken for update SLO.

I'll remove it

src/azure-cli/azure/cli/command_modules/sql/_params.py
v-nestan: Thanks for pointing this up

yonzhan:
add to S165.

src/azure-cli/HISTORY.rst
```diff
@@ -51,6 +51,10 @@ Release History
 * `az storage remove`: Change `--inlcude` and `--exclude` parameters to `--include-path`, `--include-pattern`, `--exclude-path` and`--exclude-pattern` parameters
 * `az storage sync`: Add `--include-pattern`, `--exclude-path` and`--exclude-pattern` parameters
 
+**SQL**
+
+* Update az sql mi update with two new properties tier and family.
```
petrajkogit: nit: please format this string little better. It sound weird: "Update az sql mi update". Also "properties: tier and family"

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -2098,21 +2098,18 @@ def encryption_protector_update(
 #                sql managed instance         #
 ###############################################
 
-
```
petrajkogit: removing this line is causing validation to fail, please return it

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -2211,6 +2211,16 @@ def managed_instance_update(
     instance.proxy_override = (
         proxy_override or instance.proxy_override)
 
+    instance.sku.name = None
```
jaredmoo: tiny nit: consider commenting the reason why you do this, since it would not be obvious when first reading

src/azure-cli/HISTORY.rst
v-nestan: Done. Thanks

PR #11874, opened 2020-01-16
[Storage] Upgrade Azcopy for storage copy/remove and blob sync

### LINKED ISSUE #11735: [Storage] Integrate Azcopy 10.3 and support for win 32
Issue #11509, #10924，#10302，#9576， #6466

**Additional context**

Juliehzl:
Fix #11735
---
Feature description：
1. Upgrade Azcopy 10.3.3 to utilize some new features
2. If no azcopy in target folder or the version is not the same as what we expect, we will install for users.
3. --include-pattern/--include-path/--exclude-pattern/--exclude-path are involved to improve the performance for target operation.
The new parameters have behaviour that is better defined in complex situations (such as
recursion). The * wildcard is supported in the pattern parameters, but not in the path ones.

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/storage/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/storage/_params.py
@@ -488,6 +488,17 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals, too-many-statem
                    'to ensure destination storage account support setting access tier. In the cases that setting '
                    'access tier is not supported, please use `--preserve-s2s-access-tier false` to bypass copying '
                    'access tier. (Default true)')
+        c.argument('exclude_pattern', arg_group='Additional Flags',
+                   help='Exclude these files when copying. This option supports wildcard characters (*)')
+        c.argument('include_pattern', arg_group='Additional Flags',
+                   help='Include only these files when copying. This option supports wildcard characters (*). '
+                   'Separate files by using a ";"')
+        c.argument('exclude_path', arg_group='Additional Flags',
+                   help='Exclude these paths when copying. This option does not support wildcard characters (*). '
+                   'Checks relative path prefix. For example: myFolder;myFolder/subDirName/file.pdf.')
+        c.argument('include_path', arg_group='Additional Flags',
+                   help='Include only these paths when copying. This option does not support wildcard characters (*). '
+                   'Checks relative path prefix. For example: myFolder;myFolder/subDirName/file.pdf')
 
     with self.argument_context('storage blob copy') as c:
         for item in ['destination', 'source']:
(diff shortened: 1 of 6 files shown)


src/azure-cli/azure/cli/command_modules/storage/_params.py
```diff
@@ -97,6 +97,20 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals, too-many-statem
                                                   "Required when --enable-files-adds is set to True")
     sas_help = 'The permissions the SAS grants. Allowed values: {}. Do not use if a stored access policy is ' \
```
qianwens: This line was above line "with self.argument_context('storage') as c:", it is embeded in the argumenttype definition now

src/azure-cli/azure/cli/command_modules/storage/azcopy/util.py
```diff
@@ -23,35 +23,52 @@
...
+        else:
+            raise CLIError('Azcopy ({}) does not exist.'.format(self.system))
+        try:
+            _urlretrieve(file_url, install_location)
+            os.chmod(install_location,
+                     os.stat(install_location).st_mode | stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH)
+        except IOError as err:
+            raise CLIError('Connection error while attempting to download azcopy ({})'.format(err))
+
+    def check_version(self):
```
qianwens: is there any way to get az version from memory instead of start a new process? the output message may change which has potential to break the result

src/azure-cli/azure/cli/command_modules/storage/azcopy/util.py
```diff
@@ -23,35 +23,52 @@
...
+        try:
+            import re
+            args = [self.executable] + ["--version"]
```
qianwens: although self.executable file not exist exception can be caught, it is better practise to check if the file exists here

src/azure-cli/azure/cli/command_modules/storage/operations/azcopy.py
```diff
@@ -94,27 +95,49 @@ def get_url_with_sas(source, account_name, container, blob, share, file_path, lo
...
+def storage_remove(cmd, client, service, target, recursive=None, exclude_pattern=None, include_pattern=None,
+                   exclude_path=None, include_path=None):
     if service == 'file':
         azcopy = _azcopy_file_client(cmd, client)
     else:
         azcopy = _azcopy_blob_client(cmd, client)
     flags = []
     if recursive is not None:
         flags.append('--recursive')
-    if include is not None:
```
qianwens: This is a question, does this means --include won't be supported since this change?

src/azure-cli/azure/cli/command_modules/storage/operations/azcopy.py
```diff
@@ -94,27 +95,49 @@ def get_url_with_sas(source, account_name, container, blob, share, file_path, lo
...
+    flags = []
+    if recursive is not None:
+        flags.append('--recursive')
+    if include_pattern is not None:
+        flags.append('--include-pattern=' + include_pattern)
+    if exclude_pattern is not None:
+        flags.append('--exclude-pattern=' + exclude_pattern)
+    if exclude_path is not None:
+        flags.append('--exclude-path=' + exclude_path)
     azcopy.sync(source, _add_url_sas(destination, azcopy.creds.sas_token), flags=['--delete-destination', 'true'])
```
qianwens: flags that you defined above is not used

src/azure-cli/azure/cli/command_modules/storage/azcopy/util.py
Juliehzl: checked here https://github.com/Azure/azure-cli/pull/11874/files#diff-218741868927572140fea7e37a1eb4f3R35

src/azure-cli/azure/cli/command_modules/storage/operations/azcopy.py
Juliehzl: yes

src/azure-cli/azure/cli/command_modules/storage/operations/azcopy.py
Juliehzl: thanks for reminder!

yonzhan:
add to S164.

PR #11886, opened 2020-01-17
[Core][Profile] Support lighthouse multi-tenant subscription

jiasli:
**History Notes:**  

[Profile] Preview: Add new attributes `homeTenantId` and `managedByTenants` to subscription accounts. Please re-run `az login` for the changes to take effect
[Profile] `az login`: Show a warning when a subscription is listed from more than one tenants and default to the first one. To select a specific tenant when accessing this subscription, please include `--tenant` in `az login`

---

**Update on 2020-02-04**

Rename:
```
subscriptionTenantId    > homeTenantId
subscription_tenant_id  > home_tenant_id
_SUBSCRIPTION_TENANT_ID > _HOME_TENANT_ID
```

---

This PR does these tasks:

1. Bump resource subscription client to API version 2019-06-01
2. Display `homeTenantId` and `managedByTenants` for subscription accounts while not changing the fields of bare tenant accounts generated by specifying `--allow-no-subscriptions`
3. Show a warning when a subscription is listed from more than one tenants
4. Add corresponding tests for both 2016-06-01 API and 2019-06-01 API

To test:
 
Run `az login` which will show a warning when a subscription is listed from more than one tenants:
 
> Subscription 9535 'Visual Studio Enterprise' can be accessed from tenants 72f9(default) and 2b8e. To select a specific tenant when accessing this subscription, please include --tenant in 'az login'.
 
In the response `homeTenantId` and `managedByTenants` are added. `homeTenantId` maps to `tenantId` of [Subscriptions - List REST API](https://docs.microsoft.com/en-us/rest/api/resources/subscriptions/list) (GUID is trimmed for simplicity):

```jsonc
{
    "cloudName": "AzureCloud",
    "id": "9535",
    "isDefault": false,
    "managedByTenants": [
        {
        "tenantId": "2b8e"
        },
        {
        "tenantId": "d6ad"
        }
    ],
    "name": "Visual Studio Enterprise",
    "state": "Disabled",
    "homeTenantId": "72f9",          // maps to tenantId from REST API
    "tenantId": "72f9",              // this is the tenant ID of the token
    "user": {
        "name": "@microsoft.com",
        "type": "user"
}
```
 
Then login with `az login -t 2b8e`, this subscription is now as below. Notice that `tenantId` has changed to reflect the token tenant ID.
 
```jsonc
{
    "cloudName": "AzureCloud",
    "id": "9535",
    "isDefault": false,
    "managedByTenants": [
        {
        "tenantId": "2b8e"
        }
    ],
    "name": "Visual Studio Enterprise",
    "state": "Disabled",
    "homeTenantId": "72f9",
    "tenantId": "2b8e",              // tenant ID is switched to the one specified by -t
    "user": {
        "name": "@microsoft.com",
        "type": "user"
    }
}
```

This allows the user to operate on subscription `9535` from tenant `2b8e`.
 
For backward compatibility, switch profile with `az cloud set -n AzureCloud --profile 2019-03-01-hybrid` and login again `az login -t 2b8e`

```jsonc
{
    "cloudName": "AzureCloud",
    "id": "9535",
    "isDefault": false,
    "name": "Visual Studio Enterprise",
    "state": "Disabled",
    "tenantId": "2b8e",
    "user": {
        "name": "@microsoft.com",
        "type": "user"
    }
}
```

This time, everything is the same as before. To switch cloud back, run `az cloud set -n AzureCloud --profile latest`


src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -259,7 +264,15 @@ def _normalize_properties(self, user, subscriptions, is_service_principal, cert_
                 _IS_DEFAULT_SUBSCRIPTION: False,
                 _TENANT_ID: s.tenant_id,
                 _ENVIRONMENT_NAME: self.cli_ctx.cloud.name
-            })
+            }
+            # for Subscriptions - List REST API 2019-06-01's subscription account
+            if subscription_dict[_SUBSCRIPTION_NAME] != _TENANT_LEVEL_ACCOUNT_NAME:
+                if hasattr(s, 'subscription_tenant_id'):
```
qianwens: what is the difference between subscription_tenant_id and tenant_id?

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -259,7 +264,15 @@ def _normalize_properties(self, user, subscriptions, is_service_principal, cert_
...
+                    subscription_dict[_SUBSCRIPTION_TENANT_ID] = s.subscription_tenant_id
+                if hasattr(s, 'managed_by_tenants'):
```
qianwens: Can we move code like subscription_dict[_SUBSCRIPTION_TENANT_ID] = s.subscription_tenant_id to the initialization code of subscription_dict = {...} to make code more clean?

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -259,7 +264,15 @@ def _normalize_properties(self, user, subscriptions, is_service_principal, cert_
                 _IS_DEFAULT_SUBSCRIPTION: False,
                 _TENANT_ID: s.tenant_id,
                 _ENVIRONMENT_NAME: self.cli_ctx.cloud.name
-            })
+            }
+            # for Subscriptions - List REST API 2019-06-01's subscription account
+            if subscription_dict[_SUBSCRIPTION_NAME] != _TENANT_LEVEL_ACCOUNT_NAME:
```
Juliehzl: What about using supported_api_version() to check here?

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -855,7 +868,21 @@ def _find_using_common_tenant(self, access_token, resource):
             subscriptions = self._find_using_specific_tenant(
                 tenant_id,
                 temp_credentials[_ACCESS_TOKEN])
-            all_subscriptions.extend(subscriptions)
+
+            # When a subscription can be listed by multiple tenants, only the first appearance is retained
+            for sub_to_add in subscriptions:
+                add_sub = True
+                for sub_to_compare in all_subscriptions:
```
Juliehzl: It looks that `all_subscriptions` in your current code is [].

src/azure-cli-core/azure/cli/core/tests/test_profile.py
```diff
@@ -15,7 +15,7 @@
 
 from adal import AdalError
 from azure.mgmt.resource.subscriptions.models import \
-    (SubscriptionState, Subscription, SubscriptionPolicies, SpendingLimit)
+    (SubscriptionState, Subscription, SubscriptionPolicies, SpendingLimit, ManagedByTenant)
```
Juliehzl: Because subscription models have api contraint, please use get_models() here.

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -855,7 +868,21 @@ def _find_using_common_tenant(self, access_token, resource):
...
+                    if sub_to_add.subscription_id == sub_to_compare.subscription_id:
+                        logger.warning("Subscription %s '%s' can be accessed from tenants %s(default) and %s. "
```
qianwens: For these new output formats, do we have some spec or context for this new multi-tenant behavior? I ask this because I want to learn if we should only output one subscription while it exists multi tenants or if we should output all of them and mark one of them as default one.

Juliehzl:
In addition, changing api version will influence many other commands. Please make sure all the tests pass and re-record related test files.

src/azure-cli-core/azure/cli/core/_profile.py
jiasli: Explained at https://github.com/Azure/azure-cli/pull/11886/files#diff-e35be24f5c5a220ff6948bcd6f18150cR37-R41

src/azure-cli-core/azure/cli/core/_profile.py
jiasli: `managed_by_tenants` is only present since API version `2019-06-01`. It is not possible to do conditional initialization according to python syntax.

jiasli: `managed_by_tenants` is only present since API version `2019-06-01`. It is not possible to do conditional initialization according to python syntax. https://stackoverflow.com/a/14263905

src/azure-cli-core/azure/cli/core/_profile.py
jiasli: We SHOULD and WILL display them all. But CLI is currently keying on `subscriptionId` in the account list which makes showing all of them impossible. This requires an overhaul of the account persistence mechanism.

src/azure-cli-core/azure/cli/core/_profile.py
jiasli: `all_subscriptions` will be filled at L885 in the first iteration.

src/azure-cli-core/azure/cli/core/tests/test_profile.py
jiasli: This test is only for the latest. For specific test, it is at `test_profile_v2016_06_01.py`.

jiasli: This test is only for `latest`. For specific test, it is in `test_profile_v2016_06_01.py`.

src/azure-cli-core/azure/cli/core/_profile.py
jiasli: This is still under debate. For me, this approach looks faster and safer and is not built on any assumption about API version and the actual response body.

jiasli:
@Juliehzl, thanks for the notification. I was planning to do that but don't have time yet.

jiasli:
This [test failure](https://dev.azure.com/azure-sdk/public/_build/results?buildId=239113&view=logs&j=53e6a249-cf91-5484-c166-713a9b7e2083&t=1bcc0c43-795c-57a1-0db2-a0a67fcc4ffe&l=7937)

```
2020-01-21T05:00:05.1483290Z FAIL: test_dls_file_mgmt (azure.cli.command_modules.dls.tests.latest.test_dls_commands.DataLakeStoreFileScenarioTest)
2020-01-21T05:00:05.1484511Z ----------------------------------------------------------------------
2020-01-21T05:00:05.1491248Z Traceback (most recent call last):
2020-01-21T05:00:05.1492359Z   File "/opt/hostedtoolcache/Python/3.8.0/x64/lib/python3.8/site-packages/azure_devtools/scenario_tests/preparers.py", line 50, in _preparer_wrapper
2020-01-21T05:00:05.1493019Z     fn(test_class_instance, **kwargs)
2020-01-21T05:00:05.1494298Z   File "/opt/hostedtoolcache/Python/3.8.0/x64/lib/python3.8/site-packages/azure/cli/command_modules/dls/tests/latest/test_dls_commands.py", line 285, in test_dls_file_mgmt
2020-01-21T05:00:05.1495634Z     self.cmd('dls fs list -n {dls} --path "{folder1}"', checks=[
2020-01-21T05:00:05.1496572Z   File "/opt/hostedtoolcache/Python/3.8.0/x64/lib/python3.8/site-packages/azure/cli/testsdk/base.py", line 161, in cmd
2020-01-21T05:00:05.1497635Z     return execute(self.cli_ctx, command, expect_failure=expect_failure).assert_with_checks(checks)
2020-01-21T05:00:05.1499168Z   File "/opt/hostedtoolcache/Python/3.8.0/x64/lib/python3.8/site-packages/azure/cli/testsdk/base.py", line 201, in __init__
2020-01-21T05:00:05.1499518Z     self._in_process_execute(cli_ctx, command, expect_failure=expect_failure)
2020-01-21T05:00:05.1499997Z   File "/opt/hostedtoolcache/Python/3.8.0/x64/lib/python3.8/site-packages/azure/cli/testsdk/base.py", line 257, in _in_process_execute
2020-01-21T05:00:05.1500189Z     raise AssertionError(ex)
2020-01-21T05:00:05.1500731Z AssertionError: Can't overwrite existing cassette ('/opt/hostedtoolcache/Python/3.8.0/x64/lib/python3.8/site-packages/azure/cli/command_modules/dls/tests/latest/recordings/test_dls_file_mgmt.yaml') in your current record mode ('once').
2020-01-21T05:00:05.1501286Z No match for the request (<Request (GET) https://cliadls000002.azuredatalakestore.net/webhdfs/v1/?OP=LISTSTATUS&api-version=2018-09-01&listSize=4000>) was found.
2020-01-21T05:00:05.1501550Z Found 5 similar requests with 1 different matcher(s) :
...
2020-01-21T05:00:05.1509663Z 5 - (<Request (GET) https://cliadls000002.azuredatalakestore.net/webhdfs/v1/.?OP=LISTSTATUS&api-version=2018-09-01&listSize=4000>).
2020-01-21T05:00:05.1510350Z Matchers succeeded : ['method', 'scheme', 'host', 'port', '_custom_request_query_matcher']
2020-01-21T05:00:05.1510574Z Matchers failed :
2020-01-21T05:00:05.1510938Z path - assertion failure :
2020-01-21T05:00:05.1511149Z /webhdfs/v1/ != /webhdfs/v1/.
```

is caused by **urllib3**'s breaking change https://github.com/urllib3/urllib3/pull/1647 at [util/url.py#L253-L255](https://github.com/urllib3/urllib3/blob/d2a5a5920fffcc48e6141f4126c491343800d16a/src/urllib3/util/url.py#L253-L255).

**urllib3** 1.24 keeps the trailing dot `.`, so the url `/webhdfs/v1/.` generated by `dls` is not modified. However the CI is using the latest [`urllib3==1.25.7`](https://dev.azure.com/azure-sdk/public/_build/results?buildId=239113&view=logs&j=201a69fb-575c-5f90-d252-d1ed1e360cc3&t=fb98e434-311f-5204-eaf4-065ec0de8a99&l=511) which trims the trailing `.` and fails due to the mismatch. 

We need to record with at least `urllib3==1.25.7`. To update urllib3, run 

```
pip install -U urllib3
```

jiasli:
For botservice, the recorded YAML files are manually edited to replace `2016-06-01` with `2019-06-01` due to the issue discussed at https://github.com/Azure/azure-cli/pull/8384#discussion_r368392620.

src/azure-cli-core/azure/cli/core/_profile.py
jiasli: Good point. I also think these concepts are very confusing. I am following up with Lighthouse and our PMs trying to persuade them to change `subscriptionTenantId` to `homeTenantId` which is a read-only field indicating where the subscription is hosted. `managedByTenants` is a mapping from REST API's [`managedByTenants`](https://docs.microsoft.com/en-us/rest/api/resources/subscriptions/list#subscription).

jiasli:
Per discussion with Lighthouse team, closing as `subscriptionTenantId` will be changed to `homeTenantId`.

src/azure-cli-core/azure/cli/core/_profile.py
qianwens: >_TENANT_LEVEL_ACCOUNT_NAME [](start = 56, length = 26)

If a subscription is a tenant level account, does it mean it only belong to one tenant? In this case, the rest api response does not has homeTenentId? why do we need this check here?

jiasli: > If a subscription is a tenant level account, does it mean it only belong to one tenant? 

In that case, the account **is** a tenant with no subscriptions, not "belong to". The naming of `subscription_dict` is inaccurate. It contains both subscription and tenant accounts. 

> In this case, the rest api response does not has homeTenentId? 

This tenant account is not generated from REST, but manually built up:

https://github.com/Azure/azure-cli/blob/944c0c3b1f30530a280bfaa6b7c7d799e3248f28/src/azure-cli-core/azure/cli/core/_profile.py#L270-L279

> Why do we need this check here?

It doesn't make sense to include `home_tenant_id` and `managed_by_tenants` (whichs are subscription account's attributes) for a tenant account.

In fact, this is all due to the faulty original design that mixes subscription and tenant in the same list. This will and must be changed in the future to reflect the hierarchy. Tenant is a one-level-higher concept than subscription.

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -866,7 +879,21 @@ def _find_using_common_tenant(self, access_token, resource):
...
+                                       "To select a specific tenant when accessing this subscription, "
+                                       "please include --tenant in 'az login'.",
+                                       sub_to_add.subscription_id, sub_to_add.display_name,
+                                       sub_to_compare.tenant_id, sub_to_add.tenant_id)
+                        add_sub = False
+                        break
```
arrownj: if there are n subscriptions below to multi tenants, we will write n lines of such similar warning here ? Can we just use a common message to combine these similar messages ?

arrownj:
Hi @jiasli , all the logic implement here is to set homeTenantTd and managedByTenants value. I'm wondering, where do we use these properties? Just used to show to end user ?

qwordy:
There are conflicting files

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -866,7 +879,21 @@ def _find_using_common_tenant(self, access_token, resource):
             subscriptions = self._find_using_specific_tenant(
                 tenant_id,
                 temp_credentials[_ACCESS_TOKEN])
-            all_subscriptions.extend(subscriptions)
```
jsntcy: When are all_subscriptions different between old behavior and your new logic? when a subscription exists in multiple tenants? for example, it exists in two tenants, will it only show once in all_subscriptions?

jiasli: Yes. A subscription will only appear once since we are keying by `subscriptionId`. The old behavior is **newly listed ones overwriting previously listed ones**, while the new behavior is **newly listed ones being discarded**. Thus, we can make sure subscriptions from **user's** home tenant is preferred (home tenant is listed earlier).

PR #11900, opened 2020-01-19
[Network]az  network dns record-set add/remove: show warning when a new record-set is created automatically

### LINKED ISSUE #2092: [DNS] DNS record management commands should support --if-match
The commands under 'az network dns record ...' are used to add/remove records from DNS record sets. They can also create a new record set if one does not already exist.

These operations need to support optimistic concurrency to protect against concurrent changes.

To address this, the '--if-match <value>' parameter should be supported on both the 'add' and 'remove' commands (and 'update-soa').

When the record set doesn't already exist and thus a new record set is being created, the command should *always* include '--if-none-match *' to protect against concurrent operations. This this should always be applied in this case, there is no need to expose it in the command parameters.

MyronFanQiu:
Fix #2092 

---

### DIFF
--- a/src/azure-cli/HISTORY.rst
+++ b/src/azure-cli/HISTORY.rst
@@ -24,6 +24,10 @@ Release History
 
 * Fix #6371: Support filename and environment variable completion in Bash
 
+**Network**
+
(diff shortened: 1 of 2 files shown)


yonzhan:
add to S164.

PR #11903, opened 2020-01-20
{Storage} Refine help for storage data plane commands and log warning

Juliehzl:
---
**Background** :
Existing data plane commands will query keys when only storage account name provided. To give more convenience to users, CLI will query keys for storage account, but it is manage plane operation, which is not expected by users.


yonzhan:
add to S164.

PR #11912, opened 2020-01-21
[Key Vault] Add a new command `az keyvault key download` for downloading keys

bim-msft:
---

This is for BYOK feature. We need to add CLI support for downloading keys such as downloading certificates. Currently, we have the command `az keyvault certificate download` for downloading certificates, this time we will add a command `az keyvault key download`.

For a specific Azure Key Vault key, whether you generate it or import it, it's always a private key (If you try to import a public key, service will throw an exception). So according to Key Vault's security policy, we cannot export a private key to outside world, we can only use it internally. So for `az keyvault key show` command, service team just return the public part of key to us using `JWK` (`Json Web Key`) format.

Here is a sample JWK data:
```
{
    "crv": null,
    "d": null,
    "dp": null,
    "dq": null,
    "e": "AAEAAQ==",
    "k": null,
    "keyOps": [
      "sign",
      "verify",
      "wrapKey",
      "unwrapKey",
      "encrypt",
      "decrypt"
    ],
    "kid": "https://bim-kv6.vault.azure.net/keys/rsa-hsm/xxxxxxxx",
    "kty": "RSA-HSM",
    "n": "4vo+nidE6MIvEqUxkOa+phqQryh8fj0/i2pvRSZaqISz1RcT/tDWE1FHPotPPFflCsxhRKWpb3FpIZXgVil6F+TsvF8wbOEi1rZY7NVCedpZOdxuoSoFotI7iYGBct7X5EzQx3K9B/u3CAQ9mgAgCPY8
hkfq8teK6VAbSX2k/RfRXZDap/Ex4I3b5TFMpChrPAJKM/WfxJ6fdd/iNPvOzUliYOJc+6thNpVk88YKQxq4WQwy1ETg9kN4G5QTTpsh69cBoVQLlkiNjTc31GGx+VmGNzvTLwDhZmjhqCZAfFLIcAFLz9X4HJmr/W
vaUc51QgedxqEOmAZQRp8U9pp3Jw==",
    "p": null,
    "q": null,
    "qi": null,
    "t": null,
    "x": null,
    "y": null
  },
```

`kty` means key type (RSA/RSA-HSM/EC/EC-HSM), `e` and `n` are part of RSA public key (See RFC-7517 for JWK detail: https://tools.ietf.org/html/rfc7517)

For the `az keyvault key download` command, all we need to do is convert JWK format to PEM (base64 encoded string) or DER (binary) format, which is similar to certificate downloading.

This is a sample PEM file:
```
-----BEGIN PUBLIC KEY-----
MFYwEAYHKoZIzj0CAQYFK4EEAAoDQgAE4D+Wa+0jgEL2nt4oveOUTtcPQanwTr6d
fD5JqNVya8lc3pCWut2Z3CHsbQOAi3Hz0eaJGA5b9vnOYJYiGLW6LA==
-----END PUBLIC KEY-----
```

In this PR, I use the library `cryptography` to do the tedious work such as construct a RSA/EC public key. Here I have learned some library usages from JOSE (https://github.com/mpdavis/python-jose/blob/master/jose/backends/cryptography_backend.py).

For the scenario test part, I copied some codes from `test_certificate_download`, which use `OpenSSL` library to verify public keys. I have prepared some keys in directory `test/latest/keys`, the command for generating these keys are listed as follows:

```
RSA Keys
* Generate: openssl genrsa -out rsa-4096.pem 4096
* Extract public key: openssl rsa -in rsa-4096.pem -pubout > rsa-4096.pub.pem`
* Public key PEM to DER: openssl rsa -pubin -inform PEM -in rsa-4096.pub.pem -outform DER -out rsa-4096.pub.der

EC Keys
* Generate: openssl ecparam -genkey -name secp521r1 -out ec521.key
* Extract public key (PEM): openssl ec -in ec-p521.pem -pubout -out ec-p521.pub.pem
* Extract public key (DER): `openssl ec -in ec-p521.pem -pubout -outform DER -out ec-p521.pub.der
```

Test methods:
1. Generate a RSA or EC key using Azure CLI or Portal, then use `az keyvault key download --vault-name {kv} -n {key_name} -f {target_file} -e {encoding}` to download it.
2. Generate a RSA or EC key using openssl command and import it into Key Vault, then download it. The openssl usage are listed in previous paragraph.


src/azure-cli/HISTORY.rst
```diff
@@ -25,6 +25,10 @@ Release History
 
 * Support app creation/update with the new sku name ST0, ST1, ST2.
 
+**Key Vault**
+
+* Add a new command `az keyvault key download` for downloading keys.
```
yungezz: in format of .... and I think `History.rst` file is auto-generated now, right? pls confirm with @fengzhou-msft

src/azure-cli/azure/cli/command_modules/keyvault/_params.py
```diff
@@ -30,7 +30,7 @@
 # CUSTOM CHOICE LISTS
 
 secret_encoding_values = secret_text_encoding_values + secret_binary_encoding_values
-certificate_format_values = ['PEM', 'DER']
+key_format_values = certificate_format_values = ['PEM', 'DER']
```
yungezz: why delete certificate_format_values?

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
```diff
@@ -694,6 +696,121 @@ def _private_ec_key_to_jwk(ec_key, jwk):
...
     return client.import_key(vault_base_url, key_name, key_obj, protection == 'hsm', key_attrs, tags)
+
+
+def download_key(client, file_path, vault_base_url=None, key_name=None, key_version='',
+                 encoding=None, identifier=None):  # pylint: disable=unused-argument
+    """ Download a key from a KeyVault. """
+    if os.path.isfile(file_path) or os.path.isdir(file_path):
+        raise CLIError("File or directory named '{}' already exists.".format(file_path))
+
+    def _bytes_to_int(b):
```
yungezz: it will be great to add comments on where the code comes from for future maintainance

src/azure-cli/azure/cli/command_modules/keyvault/_params.py
bim-msft: @yungezz It is not deleted, I just change `certificate_format_values = ['PEM', 'DER']` to `key_format_values = certificate_format_values = ['PEM', 'DER']`.

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
bim-msft: @yungezz Good idea, I will add some comments for the complicated code parts.

src/azure-cli/HISTORY.rst
bim-msft: @yungezz Just confirmed with him offline, we still need to modify `HISTORY.rst` manually in this sprint (S164). After the early Feb release, this file will be generated automatically.

src/azure-cli/azure/cli/command_modules/keyvault/tests/latest/test_keyvault_commands.py
```diff
@@ -237,6 +238,82 @@ def test_keyvault_key(self, resource_group):
...
+
+        # EC Keys
+        # * Generate: `openssl ecparam -genkey -name secp521r1 -out ec521.key`
+        # * Extract public key (PEM): `openssl ec -in ec-p521.pem -pubout -out ec-p521.pub.pem`
+        # * Extract public key (DER): `openssl ec -in ec-p521.pem -pubout -outform DER -out ec-p521.pub.der`
+
+        for key_name in key_names:
+            var_name = key_name.split('.')[0] + '-file'
+            if 'hsm' in key_name:  # Should be generated
+                continue
```
bim-msft: Oh, I missed something here, will send a patch.

bim-msft:
> pls let keyvault team test the download feature to make sure file downloaded met their need.

Sure, I will provide them a private version for testing after this PR is merged.

jiasli:
I think keyvault service team can test first before merging the PR. Otherwise reverting and fixing may be more complicated. 

Also, please refine the PR's description, including a brief summary, test methods and where complicated code like `_extract_ec_public_key_from_jwk` comes from.

bim-msft:
> I think keyvault service team can test first before merging the PR. Otherwise reverting and fixing may be more complicated.
> 
> Also, please refine the PR's description, including a brief summary, test methods and where complicated code like `_extract_ec_public_key_from_jwk` comes from.

@jiasli Thanks for your advice. I've updated the summary and background of this PR. Hope this will be helpful.

yungezz:
> I think keyvault service team can test first before merging the PR. Otherwise reverting and fixing may be more complicated.
> 
> Also, please refine the PR's description, including a brief summary, test methods and where complicated code like `_extract_ec_public_key_from_jwk` comes from.

to understand context, is it general guideline to let service team testing before PR merged? For Keyvault and this specific feature, I'd encourage more testing. But to be clear on it, it's a nice thing we  do, but not a must.

bim-msft:
Just send a private package to service team for testing, waiting for their feedback.

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
```diff
@@ -694,6 +699,144 @@ def _private_ec_key_to_jwk(ec_key, jwk):
...
+        d['y'] = _bytes_to_int(jwk.y)
+
+    return d
+
+
+def _extract_rsa_public_key_from_jwk(jwk_dict):
+    """
+    Please refer to:
+    https://github.com/mpdavis/python-jose/blob/eed086d7650ccbd4ea8b555157aff3b1b99f14b9/jose/backends
+    /cryptography_backend.py#L249-L254
```
jiasli: We usually don't wrap links but use `# pylint: disable=line-too-long` to disable the warnings instead.

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
```diff
@@ -694,6 +699,144 @@ def _private_ec_key_to_jwk(ec_key, jwk):
...
+            format=PublicFormat.SubjectPublicKeyInfo
+        )
+
+    methods = {
+        'DER': _to_der,
+        'PEM': _to_pem
+    }
+
+    if encoding not in methods.keys():
+        raise CLIError('Unsupported encoding: {}'.format(encoding))
```
jiasli: Let's prompt the supported encoding as well.

src/azure-cli/azure/cli/command_modules/keyvault/tests/latest/test_keyvault_commands.py
```diff
@@ -237,6 +238,98 @@ def test_keyvault_key(self, resource_group):
...
+                             ' -p hsm --kty EC-HSM --curve ' + curve_names[ec_curve])
+            else:  # Should be imported (Have already been generated offline)
+                self.kwargs[var_name] = os.path.join(KEYS_DIR, key_name)
+                self.cmd('keyvault key import --vault-name {kv} -n ' + var_name + ' --pem-file "{' + var_name + '}"')
+
+            der_downloaded_filename = var_name + '.der'
+            pem_downloaded_filename = var_name + '.pem'
+
+            try:
+                self.cmd('keyvault key download --vault-name {kv} -n ' + var_name + ' -f "' + der_downloaded_filename + '" -e DER')
```
jiasli: Let's use [string format](https://docs.python.org/3/library/string.html#format-string-syntax) for readability.

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
bim-msft: @jiasli I have tried not to wrap the links and use `# pylint: disable=line-too-long` to disable the check: https://github.com/Azure/azure-cli/pull/11912/commits/cf25aebc0f6e0da66e03e56882430a1be98f29e3 , but it didn't work.
Do you have any idea on this?

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
bim-msft: @jiasli Good idea, I will add this.

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
bim-msft: Problem solved: https://stackoverflow.com/questions/5466451/how-can-i-print-literal-curly-brace-characters-in-python-string-and-also-use-fo

src/azure-cli/azure/cli/command_modules/keyvault/tests/latest/test_keyvault_commands.py
bim-msft: @jiasli Refined this.

jiasli:
Pylint by default will ignore `line-too-long` on a line with only a single URL: http://pylint.pycqa.org/en/latest/technical_reference/features.html#format-checker-options

> ignore-long-lines:
> Regexp for a line that is allowed to be longer than the limit.
> Default: ^\s*(# )?<?https?://\S+>?$

jiasli:
Thanks @bim-msft for the below explanation. 

Note: we manually implement `_bytes_to_int` and `_int_to_bytes` instead of using the built-in [`int.from_bytes`](https://docs.python.org/3/library/stdtypes.html#int.from_bytes) and [`int.to_bytes`](https://docs.python.org/3/library/stdtypes.html#int.to_bytes) for Python 2 compatibility. 

We will drop them and use the built-in functions after we totally drop Python 2 support. I wrote a sample to demo the usage of the built-in functions:

```py
import codecs, math, struct


def _int_to_bytes(i):
    h = hex(i)
    if len(h) > 1 and h[0:2] == '0x':
        h = h[2:]
    # need to strip L in python 2.x
    h = h.strip('L')
    if len(h) % 2:
        h = '0' + h
    return codecs.decode(h, 'hex')


def _bytes_to_int(b):
    """Convert bytes to hex integer"""
    len_diff = 4 - len(b) % 4 if len(b) % 4 > 0 else 0
    b = len_diff * b'\x00' + b  # We have to patch leading zeros for using struct.unpack
    bytes_num = int(math.floor(len(b) / 4))
    ans = 0
    items = struct.unpack('>' + 'I' * bytes_num, b)
    for sub_int in items:  # Accumulate all items into a big integer
        ans *= 2 ** 32
        ans += sub_int
    return ans


x = 31415926535897932384626433832795028841971693993751058209749445923078164062862
b1 = _int_to_bytes(x)
b2 = x.to_bytes(length=(x.bit_length() + 7) // 8, byteorder='big', signed=False)
print(b1)
print(b2)

i1 = _bytes_to_int(b1)
i2 = int.from_bytes(b2, byteorder='big', signed=False)
print(i1)
print(i2)
```

Output:
```
b'Et\xc8\xc7]n\x88\xac\xd2\x8f~F}\xac\x97\xb5\xc6\x0c88\xd9\xda\xd9\x93\x90\x0b\xdf@!R"\x8e'
b'Et\xc8\xc7]n\x88\xac\xd2\x8f~F}\xac\x97\xb5\xc6\x0c88\xd9\xda\xd9\x93\x90\x0b\xdf@!R"\x8e'
31415926535897932384626433832795028841971693993751058209749445923078164062862
31415926535897932384626433832795028841971693993751058209749445923078164062862
```

Issue #11785, opened 2020-01-07
Cannot add secrets to virtual machine


ctolkien:
## Describe the bug

**Command Name**
`az vm secret list`

**Errors:**
```
'NoneType' object has no attribute 'secrets'
Traceback (most recent call last):
cliextensions\interactive\azext_interactive\azclishell\app.py, ln 662, in cli_execute
    result = invocation.execute(args)
azure\cli\core\commands\__init__.py, ln 603, in execute
azure\cli\core\commands\__init__.py, ln 661, in _run_jobs_serially
azure\cli\core\commands\__init__.py, ln 654, in _run_job
Local\Temp\pip-install-umfcm_n2\six\six.py, ln 693, in reraise
azure\cli\core\commands\__init__.py, ln 631, in _run_job
azure\cli\core\commands\__init__.py, ln 306, in __call__
azure-cli-core\azure\cli\core\__init__.py, ln 485, in default_command_handler
azure\cli\command_modules\vm\custom.py, ln 1730, in list_vm_secrets
AttributeError: 'NoneType' object has no attribute 'secrets'
```

- _Put any pre-requisite steps here..._
- `az vm secret list --name {} -g {}`

## Expected Behavior
Should work?

## Environment Summary
```
Windows-10-10.0.18362-SP0
Python 3.6.6
Shell: cmd.exe

azure-cli 2.0.76 *

Extensions:
azure-devops 0.12.0
interactive 0.4.3

```

bim-msft:
@qwordy Please take a look.

qwordy:
@ctolkien, How do you create the vm?
We have written a patch #11914 . The problem is vm's os_profile is None.

ctolkien:
This was made via the portal, I Bieber from an existing HDD?

qwordy:
The patch has released in latest Azure CLI. Hope it can solve your issue. If any problem, you can open a new issue.

Issue #11533, opened 2019-12-10
Vmaas Extension List not working


kotharij:
ubuntu@ubuntu$ az vmss  extension list --resource-group coke-b375c-tms-loc_gateway --vmss-name vmscaleset --subscription &lt;...&gt;
The command failed with an unexpected error. Here is the traceback:

'NoneType' object has no attribute 'extension_profile'
Traceback (most recent call last):
  File "/opt/az/lib/python3.6/site-packages/knack/cli.py", line 206, in invoke
    cmd_result = self.invocation.execute(args)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 603, in execute
    raise ex
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 661, in _run_jobs_serially
    results.append(self._run_job(expanded_arg, cmd_copy))
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 654, in _run_job
    six.reraise(*sys.exc_info())
  File "/opt/az/lib/python3.6/site-packages/six.py", line 693, in reraise
    raise value
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 631, in _run_job
    result = cmd_copy(params)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 306, in __call__
    return self.handler(*args, **kwargs)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/__init__.py", line 487, in default_command_handler
    return op(**command_args)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/command_modules/vm/custom.py", line 2645, in list_vmss_extensions
    return None if not vmss.virtual_machine_profile.extension_profile \
AttributeError: 'NoneType' object has no attribute 'extension_profile'

To open an issue, please run: 'az feedback'

ubuntu@ubuntu$ az vmss list
[
  {
    "additionalCapabilities": null,
    "automaticRepairsPolicy": null,
    "doNotRunExtensionsOnOverprovisionedVms": null,
    "id": "/subscriptions/&lt;...&gt;/resourceGroups/COKE-B375C-TMS-LOC_GATEWAY/providers/Microsoft.Compute/virtualMachineScaleSets/vmscaleset",
    "identity": null,
    "location": "eastus",
    "name": "vmscaleset",
    "overprovision": null,
[...]

PR #11914, opened 2020-01-21
{Compute} fix none check when list vmss extension.

yungezz:
Fix error in issues

#11533
#11785
Add None check to vmss.virtual_machine_profile

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/vm/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/custom.py
@@ -1879,7 +1879,9 @@ def add_vm_secret(cmd, resource_group_name, vm_name, keyvault, certificate, cert
 
 def list_vm_secrets(cmd, resource_group_name, vm_name):
     vm = get_vm(cmd, resource_group_name, vm_name)
-    return vm.os_profile.secrets
+    if vm.os_profile:
+        return vm.os_profile.secrets
+    return None
 
 
 def remove_vm_secret(cmd, resource_group_name, vm_name, keyvault, certificate=None):
@@ -2815,8 +2817,9 @@ def list_vmss_extensions(cmd, resource_group_name, vmss_name):
     client = _compute_client_factory(cmd.cli_ctx)
     vmss = client.virtual_machine_scale_sets.get(resource_group_name, vmss_name)
     # pylint: disable=no-member
-    return None if not vmss.virtual_machine_profile.extension_profile \
-        else vmss.virtual_machine_profile.extension_profile.extensions
+    if vmss.virtual_machine_profile and vmss.virtual_machine_profile.extension_profile:
+        return vmss.virtual_machine_profile.extension_profile.extensions
+    return None
 
 
 def set_vmss_extension(cmd, resource_group_name, vmss_name, extension_name, publisher, version=None,
--- a/src/azure-cli/azure/cli/command_modules/vm/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/custom.py
@@ -1881,7 +1881,7 @@ def list_vm_secrets(cmd, resource_group_name, vm_name):
     vm = get_vm(cmd, resource_group_name, vm_name)
     if vm.os_profile:
         return vm.os_profile.secrets
-    return None
+    return []
 
 
 def remove_vm_secret(cmd, resource_group_name, vm_name, keyvault, certificate=None):


jiasli:
@fengzhou-msft, we are seeing [this error](https://dev.azure.com/azure-sdk/public/_build/results?buildId=239301&view=logs&j=b21ded15-2ef6-5c2b-875f-d00fdf573f12&t=ae271903-525b-5a25-197b-c8c2e710d08d&l=908) more and more often

```
git checkout --progress --force bf1d02f8058230e6bfc720668f78d5f926d37de4
fatal: reference is not a tree: bf1d02f8058230e6bfc720668f78d5f926d37de4
##[error]Git checkout failed with exit code: 128
```

jiasli:
I tried `az vm show` and `az vmss show` and saw `osProfile` and `virtualMachineProfile` appearing in the response json respectively. Under what circumstances will them be `null`?

yungezz:
> I tried `az vm show` and `az vmss show` and saw `osProfile` and `virtualMachineProfile` appearing in the response json respectively. Under what circumstances will they be `null`?

ideally those properties are in results, while customers do see NoneType exception as showed in below issues:
#11533
#11785

jiasli:
Thanks. I saw the exceptions in those two issues. While this PR solves them, I think it will still be great for us to find out why that happened in the first place. @qwordy

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -1879,7 +1879,9 @@ def add_vm_secret(cmd, resource_group_name, vm_name, keyvault, certificate, cert
 
 def list_vm_secrets(cmd, resource_group_name, vm_name):
     vm = get_vm(cmd, resource_group_name, vm_name)
-    return vm.os_profile.secrets
+    if vm.os_profile:
+        return vm.os_profile.secrets
+    return None
```
qwordy: `return None` will show nothing in the output.
how about `return []`

yungezz: Nice. actually I have same confusion on return value when result is None. sometimes az cli returns  [], sometimes None. Is there any general guideline?

yungezz:
> Thanks. I saw the exceptions in those two issues. While this PR solves them, I think it will still be great for us to find out why that happened in the first place. If the `null` property is generated by CLI, maybe we can fix that in the create or update command. @qwordy

It could happen in parallel. Eventually it's not right to throw NoneType exception. While we can keep understanding users specific scenario, it takes time and to the end, specific to environments. because both commands are quite straightforward, CLI does nothing except wrapping rest call and output directly.

src/azure-cli/azure/cli/command_modules/vm/custom.py
qwordy: In this command, the output is [] if there is no secret.
Let me search the doc.

qwordy:
> 
> 
> Thanks. I saw the exceptions in those two issues. While this PR solves them, I think it will still be great for us to find out why that happened in the first place. If the `null` property is generated by CLI, maybe we can fix that in the create or update command. @qwordy

Adding null pointer checker is a general way to mitigate this kind of bug. I agree. We need to find out why it's null.

fengzhou-msft:
I haven't identified the root cause for the git checkout error, seems random.  But I noticed we checkout source code for all tasks even those package/image testing tasks that don't need the source code. 

@haroldrandom do we have a way to skip checking out code in tasks?

src/azure-cli/azure/cli/command_modules/vm/custom.py
jiasli: Tested with a VM and the results confirms @qwordy's comment:

```
az vm secret list -n {} -g {}
[]
```

yonzhan:
@haroldrandom  please take a look at the question raised by @fengzhou-msft

haroldrandom:
> I haven't identified the root cause for the git checkout error, seems random. But I noticed we checkout source code for all tasks even those package/image testing tasks that don't need the source code.
> 
> @haroldrandom do we have a way to skip checking out code in tasks?

As for randomly fail on `git checkout`, I haven't dig it out. It occurs often suddenly. Intuitively, ADO could make some changes.

As for skip-able git checkout, I searched Document of ADO, found some options like [Don't sync sources](https://docs.microsoft.com/zh-cn/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#dont-sync-sources) and [Shallow Fetch](https://docs.microsoft.com/zh-cn/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#shallow-fetch) might help. Nice suggestion to speed up cost time of CI, will dig it later.

yungezz:
CI passed in Azure DevOps https://dev.azure.com/azure-sdk/public/_build/results?buildId=276518&view=results, seems status not sycing into github. Looking at it

PR #11915, opened 2020-01-21
{AppService} Refine test_linux_webapp_ssh on Windows

jiasli:
Skip `test_linux_webapp_ssh` on Windows as it will fail at
https://github.com/Azure/azure-cli/blob/c5b5772ab958e755a9649e455a8e6bb17edc6400/src/azure-cli/azure/cli/command_modules/appservice/custom.py#L3282-L3283

and break the test on `pytest` or `azdev test`.

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -3274,13 +3274,13 @@ def _start_ssh_session(hostname, port, username, password):
 
 
 def ssh_webapp(cmd, resource_group_name, name, port=None, slot=None, timeout=None):  # pylint: disable=too-many-statements
-    config = get_site_configs(cmd, resource_group_name, name, slot)
-    if config.remote_debugging_enabled:
-        raise CLIError('remote debugging is enabled, please disable')
-
     import platform
     if platform.system() == "Windows":
         raise CLIError('webapp ssh is only supported on linux and mac')
+
+    config = get_site_configs(cmd, resource_group_name, name, slot)
+    if config.remote_debugging_enabled:
+        raise CLIError('remote debugging is enabled, please disable')
     create_tunnel_and_session(cmd, resource_group_name, name, port=port, slot=slot, timeout=timeout)
 
 
--- a/src/azure-cli/azure/cli/command_modules/appservice/tests/latest/test_webapp_commands.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/tests/latest/test_webapp_commands.py
@@ -590,10 +590,14 @@ class LinuxWebappScenarioTest(ScenarioTest):
 class LinuxWebappSSHScenarioTest(ScenarioTest):
     @ResourceGroupPreparer(location='japanwest')
     def test_linux_webapp_ssh(self, resource_group):
-        # Skip this test on Windows as it will fail at 'webapp ssh'
+        # On Windows, test 'webapp ssh' throws error
         import platform
         if platform.system() == "Windows":
+            from azure.cli.core.util import CLIError
+            with self.assertRaises(CLIError):
+                self.cmd('webapp ssh -g {} -n {} --timeout 5'.format("foo", "bar"))
             return
+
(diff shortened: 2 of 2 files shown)


src/azure-cli/azure/cli/command_modules/appservice/tests/latest/test_webapp_commands.py
```diff
@@ -590,6 +590,10 @@ def test_linux_webapp(self, resource_group):
 class LinuxWebappSSHScenarioTest(ScenarioTest):
     @ResourceGroupPreparer(location='japanwest')
     def test_linux_webapp_ssh(self, resource_group):
+        # Skip this test on Windows as it will fail at 'webapp ssh'
+        import platform
+        if platform.system() == "Windows":
```
Juliehzl: As there are many people using Windows system, if we skip this test, the test may be missed in some scenarios. But if we can make sure our CI system always have tests on Linux system. I am fine to skip this test because our CI system is running on Linux now. +@haroldrandom for awareness.
In addition, I will prefer that when system detected is Windows, the test can assert exception as expected.

jiasli: Nice idea, but `webapp ssh` does more than failing on Windows directly - it generates some network traffic before the platform detection: 
https://github.com/Azure/azure-cli/blob/c5b5772ab958e755a9649e455a8e6bb17edc6400/src/azure-cli/azure/cli/command_modules/appservice/custom.py#L3276-L3284

I have changed the logic of `ssh_webapp` so that it fails as early as possible.

yonzhan:
add to S165.

yonzhan:
add to S165.

src/azure-cli/azure/cli/command_modules/appservice/tests/latest/test_webapp_commands.py
haroldrandom: I would prefer to use decorator `unittest.skipIf()` leverage this.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -3274,13 +3274,13 @@ def _start_ssh_session(hostname, port, username, password):
 
 
 def ssh_webapp(cmd, resource_group_name, name, port=None, slot=None, timeout=None):  # pylint: disable=too-many-statements
-    config = get_site_configs(cmd, resource_group_name, name, slot)
-    if config.remote_debugging_enabled:
-        raise CLIError('remote debugging is enabled, please disable')
-
     import platform
     if platform.system() == "Windows":
         raise CLIError('webapp ssh is only supported on linux and mac')
```
qwordy: Can we make Windows also support this feature? Windows also has ssh command. Do you know the reason?

src/azure-cli/azure/cli/command_modules/appservice/tests/latest/test_webapp_commands.py
jiasli: It is tested on Windows as well for raising the exception.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
jiasli: I don't know. I am only trying to unblock azdev. @panchagnula @Juliehzl , kindly comment on this.

src/azure-cli/azure/cli/command_modules/appservice/tests/latest/test_webapp_commands.py
haroldrandom: I am not sure what's the purpose exactly? Is going to raise on Windows or Skip if on Windows.
If you want to skip on Windows, I would recommend to use `unitest.skipIf()` decorator.

jiasli: The command itself will raise an exception on Windows and the exception is also tested. So we are not skipping anything.

PR #11955, opened 2020-01-27
[AppService] Add az webapp|functionapp config ssl create

### LINKED ISSUE #11872: Create App Service Managed Certificates via CLI?
I would like to secure the webapp I am currently deploying via CLI by adding a custom domain with a proper TLS certificate. The page linked below shows the option to create a certificate through the portal labeled "Create App Service Managed Certificates". This option does not appear to be exposed through the CLI. Are there plans to add support for it in the CLI as well?

https://docs.microsoft.com/en-us/azure/app-service/configure-ssl-certificate#create-a-free-certificate-preview

---
#### Document Details

* ID: 2a3f984d-0d48-e075-7e2f-2da5667b15cf
* Version Independent ID: 98e06d47-0ed4-4d8e-ccf5-d3a78e510539
* Content: [az webapp config ssl](https://docs.microsoft.com/en-us/cli/azure/webapp/config/ssl?view=azure-cli-latest&source=docs#feedback)
* Content Source: [latest/docs-ref-autogen/webapp/config/ssl.yml](https://github.com/MicrosoftDocs/azure-docs-cli/blob/live/latest/docs-ref-autogen/webapp/config/ssl.yml)
* Service: **app-service-web**
* GitHub Login: @rloutlaw

madsd:
Fixes #11872 
---

### DIFF
--- a/src/azure-cli/HISTORY.rst
+++ b/src/azure-cli/HISTORY.rst
@@ -41,6 +41,7 @@ Release History
 
 * Azure Stack: surface commands under the profile of 2019-03-01-hybrid
 * functionapp: Add ability to create Java function apps in Linux
+* Add support for Managed Certificate: az webapp|functionapp config ssl create-managed-certificate
 
 **ARM**
 
--- a/src/azure-cli/HISTORY.rst
+++ b/src/azure-cli/HISTORY.rst
@@ -41,7 +41,6 @@ Release History
 
 * Azure Stack: surface commands under the profile of 2019-03-01-hybrid
 * functionapp: Add ability to create Java function apps in Linux
-* Add az webapp|functionapp config ssl create-managed-certificate
 
 **ARM**
 
--- a/src/azure-cli/HISTORY.rst
+++ b/src/azure-cli/HISTORY.rst
@@ -41,7 +41,7 @@ Release History
 
 * Azure Stack: surface commands under the profile of 2019-03-01-hybrid
 * functionapp: Add ability to create Java function apps in Linux
-* Add support for Managed Certificate: az webapp|functionapp config ssl create-managed-certificate
+* Add az webapp|functionapp config ssl create-managed-certificate
 
 **ARM**
 
--- a/src/azure-cli/azure/cli/command_modules/appservice/commands.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/commands.py
@@ -144,7 +144,7 @@ def load_command_table(self, _):
         g.custom_command('unbind', 'unbind_ssl_cert', validator=validate_app_or_slot_exists_in_rg)
         g.custom_command('delete', 'delete_ssl_cert', exception_handler=ex_handler_factory())
         g.custom_command('import', 'import_ssl_cert', exception_handler=ex_handler_factory(), is_preview=True)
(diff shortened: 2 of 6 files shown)


yonzhan:
add to S165.

qwordy:
Please add at least one test case

madsd:
Unit test added - although it had to be mocked. It is dependent on a CNAME lookup.

src/azure-cli/HISTORY.rst
```diff
@@ -37,6 +37,7 @@ Release History
 
 * Azure Stack: surface commands under the profile of 2019-03-01-hybrid
 * functionapp: Add ability to create Java function apps in Linux
+* Add az webapp|functionapp config ssl create-managed-certificate
```
Juliehzl: Please make your release note in next release part.

madsd: removed the History.rst release note as this is extracted from the title now

Juliehzl:
> Unit test added - although it had to be mocked. It is dependent on a CNAME lookup.

Why not use scenario test here?

madsd:
> > Unit test added - although it had to be mocked. It is dependent on a CNAME lookup.
> 
> Why not use scenario test here?

As mentioned in our chat - because scenario test will not be able to handle the CNAME registration part that is necessary for a successful run. You will see the same pattern for e.g. setting domain name.

panchagnula:
@madsd this feature is not supported only on SKUs Basic & above - so free & shared don't support this can you add a validation for this & throw an error - the API doesn't handle this currently but the FrontEnds block - so we need to handle this on CLI end, we do the same in portal. Thanks!

madsd:
> @madsd this feature is not supported only on SKUs Basic & above - so free & shared don't support this can you add a validation for this & throw an error - the API doesn't handle this currently but the FrontEnds block - so we need to handle this on CLI end, we do the same in portal. Thanks!

I added a check for Free and Shared, and will throw a CLI error if it is. I updated the unit tests to handle this.

src/azure-cli/azure/cli/command_modules/appservice/commands.py
```diff
@@ -144,6 +144,7 @@ def load_command_table(self, _):
         g.custom_command('unbind', 'unbind_ssl_cert', validator=validate_app_or_slot_exists_in_rg)
         g.custom_command('delete', 'delete_ssl_cert', exception_handler=ex_handler_factory())
         g.custom_command('import', 'import_ssl_cert', exception_handler=ex_handler_factory(), is_preview=True)
+        g.custom_command('create-managed-certificate', 'create_managed_ssl_cert', exception_handler=ex_handler_factory(), is_preview=True)
```
panchagnula: @yonzhan / @Juliehzl can you comment if this naming follows the recommended naming convention
@madsd  i was thinking if we should call this create simply & have another property that defaults to Managed certs as default for now so that we can re-purpose the same command to create AppService certificates when we are ready to add that support as well? What do you think?  If you agree , we can run this by Yutang as well as the PM for this feature.

panchagnula: @yutanglin16 - FYI

PR #11960, opened 2020-01-28
[ACS] aks create/update: add --load-balancer-outbound-ports and --load-balancer-idle-timeout

gtracer:
---


yonzhan:
add to S165.

gtracer:
@yonzhan noticed you added a tag for S164. We are looking to get this PR merged before the 29th deadline for S164. 

CC: @jluk

jluk:
@yonzhan can you please move this PR into [S164](https://github.com/Azure/azure-cli/milestone/81)? It looks like we are still 2 days prior to cutoff. 

As Ganesh mentions above, there are open AKS CRIs and a few large customers waiting on this release in an official cli build so he has put in extra hours to land this on time.

yonzhan:
@qwordy @MyronFanQiu  please help with this PR before this release.

src/azure-cli/azure/cli/command_modules/acs/_help.py
```diff
@@ -372,6 +382,14 @@
     type: string
     short-summary: Load balancer outbound IP prefix resource IDs.
     long-summary: Comma-separated public IP prefix resource IDs for load balancer outbound connection. Valid for Standard SKU load balancer cluster only.
+  - name: --load-balancer-outbound-ports
+    type: string
+    short-summary: Load balancer outbound allocated ports.
```
gtracer: Will update to `Desired static number of outbound ports per VM in the load balancer backend pool. By default, set to 0 which will use the default allocation table.`

gtracer: Desired static number of outbound ports per VM in the load balancer backend pool. By default, set to 0 which uses the default allocation based on the number of VMs

src/azure-cli/HISTORY.rst
```diff
@@ -3,6 +3,10 @@
 Release History
 ===============
 
+**ACS**
+
+* Added support to set outbound allocated ports and idle timeouts on standard load balancer
```
weinong: also changed api-version, right?

gtracer: Yes, good catch. Will update

src/azure-cli/azure/cli/command_modules/acs/_validators.py
```diff
@@ -197,6 +197,22 @@ def validate_load_balancer_outbound_ip_prefixes(namespace):
             raise CLIError("--load-balancer-outbound-ip-prefixes cannot contain whitespace")
 
 
+def validate_load_balancer_outbound_ports(namespace):
```
weinong: total ports allocated to VMs cannot exceed 64k * number_of_front_end_ip. I wonder if we should add this check

gtracer: Let's do it - we have it in RP but probably makes sense to have it here in the client too. Will add

gtracer: added

src/azure-cli/azure/cli/command_modules/acs/_validators.py
```diff
@@ -202,15 +202,15 @@ def validate_load_balancer_outbound_ports(namespace):
     if namespace.load_balancer_outbound_ports is not None:
         if namespace.load_balancer_outbound_ports % 8 != 0:
             raise CLIError("--load-balancer-allocated-ports must be a multiple of 8")
-        if namespace.load_balancer_outbound_ports < 0:
-            raise CLIError("--load-balancer-allocated-ports cannot be negative")
+        if namespace.load_balancer_outbound_ports < 0 or namespace.load_balancer_outbound_ports > 64000:
```
weinong: you know this number is based on number of VMs and number of FE IP, right? is it what we have implemented in RP? I guess it's better than before...

gtracer: Yup, that is what we implemented in RP. 

We do a check that: IPs * 64K > # of vms * desired allocated ports 

where # of vms accounts for things like buffer for upgrading and over provisioning, auto scaler etc.

src/azure-cli/azure/cli/command_modules/acs/_help.py
```diff
@@ -252,6 +252,14 @@
     type: string
     short-summary: Load balancer outbound IP prefix resource IDs.
     long-summary: Comma-separated public IP prefix resource IDs for load balancer outbound connection. Valid for Standard SKU load balancer cluster only.
+  - name: --load-balancer-outbound-ports
+    type: string
```
robbiezhang: why the type is not int?

src/azure-cli/azure/cli/command_modules/acs/_help.py
```diff
@@ -252,6 +252,14 @@
...
+    short-summary: Load balancer outbound allocated ports.
+    long-summary: Desired static number of outbound ports per VM in the load balancer backend pool. By default, set to 0 which uses the default allocation based on the number of VMs.
+  - name: --load-balancer-idle-timeout
+    type: string
```
robbiezhang: same question, int?

src/azure-cli/azure/cli/command_modules/acs/_loadbalancer.py
```diff
@@ -0,0 +1,101 @@
...
+            public_ip_prefixes=outbound_ip_prefix_resources
+        )
+    if outbound_ports:
+        profile.allocated_outbound_ports = outbound_ports
+    if idle_timeout:
+        profile.idle_timeout_in_minutes = idle_timeout
+    return profile
+
+
+def is_load_balancer__profile_provided(managed_outbound_ip_count, outbound_ips, ip_prefixes,
```
robbiezhang: double "_" between balancer and profile. in purpose?

src/azure-cli/azure/cli/command_modules/acs/_help.py
gtracer: Should be - good eye. Updating

src/azure-cli/azure/cli/command_modules/acs/_help.py
gtracer: Should be - good eye. Updating them all

src/azure-cli/azure/cli/command_modules/acs/_loadbalancer.py
gtracer: Accidental, thanks for catching. Fixing

robbiezhang:
LGTM

jsntcy:
@gtracer, please help fix CI check errors above first.

gtracer:
> @gtracer, please help fix CI check errors above first.

Loo

gtracer:
@jsntcy we good to merge this?

jluk:
@yonzhan it looks like checks have all passed, but want to check if you can merge quickly before your cut? PR looks clear?

yonzhan:
@qwordy  will help you with this PR.

jluk:
Thank you @yonzhan - @qwordy could you ack if we can get this in for the CLI release on Tuesday? Looking to ship this to some blocked customers impacted by SNAT exhaustion which would be fixed by the changes in this PR.

src/azure-cli/azure/cli/command_modules/acs/_loadbalancer.py
```diff
@@ -0,0 +1,101 @@
...
+from azure.mgmt.containerservice.v2019_11_01.models import ManagedClusterLoadBalancerProfileOutboundIPPrefixes
+from azure.mgmt.containerservice.v2019_11_01.models import ManagedClusterLoadBalancerProfileOutboundIPs
+from azure.mgmt.containerservice.v2019_11_01.models import ResourceReference
+
+
+def set_load_balancer_sku(sku, kubernetes_version):
+    if sku:
+        return sku
+    if kubernetes_version and StrictVersion(kubernetes_version) < StrictVersion("1.13.0"):
+        print('Setting load_balancer_sku to basic as it is not specified and kubernetes \
```
yungezz: pls not use print, but logger.warning or error, eg. https://github.com/Azure/azure-cli/blob/8adf97355a605d147afc26f6c1b6f6afb6651c2f/src/azure-cli/azure/cli/command_modules/cognitiveservices/custom.py#L45

src/azure-cli/azure/cli/command_modules/acs/_validators.py
```diff
@@ -197,6 +197,22 @@ def validate_load_balancer_outbound_ip_prefixes(namespace):
             raise CLIError("--load-balancer-outbound-ip-prefixes cannot contain whitespace")
 
 
+def validate_load_balancer_outbound_ports(namespace):
+    """validate load balancer profile outbound allocated ports"""
+    if namespace.load_balancer_outbound_ports is not None:
+        if namespace.load_balancer_outbound_ports % 8 != 0:
+            raise CLIError("--load-balancer-allocated-ports must be a multiple of 8")
+        if namespace.load_balancer_outbound_ports < 0 or namespace.load_balancer_outbound_ports > 64000:
```
yungezz: suggest to add the constraint in doc above also

yungezz:
hi @gtracer , could you pls address comments? Thanks. could you pls give @qwordy  push permission to your fork repo so that we can help on minor change? Thanks.

src/azure-cli/azure/cli/command_modules/acs/_loadbalancer.py
```diff
@@ -0,0 +1,101 @@
...
+    profile = ManagedClusterLoadBalancerProfile()
+    return configure_load_balancer_profile(managed_outbound_ip_count, outbound_ips, outbound_ip_prefixes,
+                                           outbound_ports, idle_timeout, profile)
+
+
+def configure_load_balancer_profile(managed_outbound_ip_count, outbound_ips, outbound_ip_prefixes, outbound_ports,
+                                    idle_timeout, profile):
+    """configure a load balancer with customer supplied values"""
+    if not profile:
+        return profile
```
qwordy: If profile is null, skip?

src/azure-cli/azure/cli/command_modules/acs/_help.py
```diff
@@ -252,6 +252,14 @@
     type: string
     short-summary: Load balancer outbound IP prefix resource IDs.
     long-summary: Comma-separated public IP prefix resource IDs for load balancer outbound connection. Valid for Standard SKU load balancer cluster only.
+  - name: --load-balancer-outbound-ports
```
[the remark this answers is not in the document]
gtracer: @yungezz your right that in most cases it is a list of ports. But in this case we are taking the name from SLB which has a property called allocated-outbound-ports (interchangeable referred to  as outbound ports) which is actually an int of the # of ports you want per VM. We want to keep the name consistent with SLB since we are using this CLI param to set the SLB in the customers AKS cluster

You can see more details here: https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-outbound-rules-overview#snatports

src/azure-cli/azure/cli/command_modules/acs/_loadbalancer.py
gtracer: @yungezz happy to update this - certainly good feedback, especially for a team like ours that doesn't use python on a daily basis. 

However, I have only moved this code - didn't actually write it - so to scope my PR I would prefer to leave this change for a separate PR.

Please let me know what you think, if you feel strongly about it I can modify it but please advise on the best way to share an "info" (e.g. is there a logger.info) -- in this case it isn't a error or warning it is just a fact that they should be aware of given the rest of their configuration

src/azure-cli/azure/cli/command_modules/acs/_validators.py
gtracer: In the _loadbalancer.py file? If so, can you explain your recommendation further? Are you worried that another operation will bypass the validation and then erroneously set it? 

I am not particularly worried if it get's by passed as this validation is just "best effort / fail fast" we do the exact same validation in the RP itself

src/azure-cli/azure/cli/command_modules/acs/_loadbalancer.py
gtracer: Thanks for the reminder, appreciate that. 

I don't expect the profile to be null at this point - because the callers of this function only call it when there is a profile. But since I am editing the profile in this func, to avoid a panic/exception I am just double checking that it isn't null.

jluk:
Thank you all for your help in merging this, it is very appreciated!

PR #11964, opened 2020-01-28
{Packaging} Generate Homebrew formula by updating existing one by default

### LINKED ISSUE #11839: Generate Homebrew Formula based on existing one
Currently we generate Homebrew formula based on a [template](https://github.com/Azure/azure-cli/blob/dev/scripts/release/homebrew/docker/formula_template.txt).

From time to time Homewbrew team may help fix some issues such as https://github.com/Homebrew/homebrew-core/pull/48846, [fix for Xcode 11](https://github.com/Homebrew/homebrew-core/commit/644b9f78e7e13560f752a2e1137da754f11b8124#diff-279b5540b1246df8d64827e6e2ef62be). We may not get notified well with such changes, and the template is not updated with the changes. A regeneration of formula based on the template will lose the changes.

A proposed solution is to generate the new formula based on existing one and only update the version of azure-cli, resource url and resource sha256.

fengzhou-msft:
To resolve #11839.

Add the method to generate a new homebrew formula based on the latest existing formula in `formula_generate.py`. And it's the default way now.

Invoke `formula_generate.py` with `-b use_template` option to generate the formula using a template which is the old way.

### DIFF
--- a/scripts/release/homebrew/docker/formula_generate.py
+++ b/scripts/release/homebrew/docker/formula_generate.py
@@ -11,6 +11,9 @@ from typing import List, Tuple
 import requests
 import jinja2
 from poet.poet import make_graph, RESOURCE_TEMPLATE
+from collections import OrderedDict
+import bisect
+import argparse
 
 TEMPLATE_FILE_NAME='formula_template.txt'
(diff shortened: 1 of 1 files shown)


fengzhou-msft:
Plan in S164.

PR #11976, opened 2020-01-29
[SQL] BREAKING CHANGE: az sql db create: Remove WideWorldImportersStd and WideWorldImportersFull as allowed values for --sample-name values

jaredmoo:
The WideWorldImportersStd and WideWorldImportersFull sql db samples were never supported in production and were only documented here by mistake. Removing them from allowed values so that users do not attempt to create them. Note that this is not a breaking change because these options already caused db creation to fail.

---


yonzhan:
add to S164.

jaredmoo:
Thank you again Julie!

Issue #11883, opened 2020-01-17
az role assignment create scope default to subscription if empty


yungezz:
> ### `az feedback` auto-generates most of the information requested below, as of CLI version 2.0.62

**Describe the bug**
# Summary

Using `azure-cli` (in actual version of `2.0.79`) and `az role assignment create` command.

The problem I see is the use of `--scope` parameter.
Empty (I mean empty = specified but undefined parameter) scope equals to default (therefore subscription) scope.

Why is that? How about the error message for this action? User is not aware of the issue and scope automatically applies on the subscription.
That can be harmful in some cases (see examples below) - user or service principal can be assigned more powerfull scope and you can imagine the consequences.

# Description

Following examples can cause elevation of privileges. In case of resource name typo, non-existing resource, etc...

## Steps to Reproduce:

```bash
# login to the portal
az login
# obtain assignee identifier (user in this case)
ASSIGNEE=$(az account show --query 'user.name' -o tsv)

# acceptable version: undefined scope => default scope => applied to subscription, OK
az role assignment create \
--assignee "${ASSIGNEE}" \
--role Contributor

# Why empty (empty, not undefined!) scope equals to default scope? Following examples can cause elevation of privileges.

# version 1: EMPTY scope => applied to subscription (nonintended elevation of privileges)
az role assignment create \
--assignee "${ASSIGNEE}" \
--scope "" \
--role Contributor

# version 2: non existing resource => EMPTY scope => applied to subscription (nonintended elevation of privileges)
az role assignment create \
--assignee "${ASSIGNEE}" \
--scope "$(az storage account show --name iMakeATypoInName --subscription ${AZURE_SUBSCRIPTION} --query 'id' -o tsv)" \
--role Contributor

# version 3: non existing resource in different way => EMPTY scope => applied to subscription (nonintended elevation of privileges)
OBJECT=$(az storage account show --name iMakeATypoInName --subscription ${AZURE_SUBSCRIPTION} --query 'id' -o tsv)
az role assignment create \
--assignee "${ASSIGNEE}" \
--scope "${OBJECT}" \
--role Contributor

# check for all the three versions:
az role assignment list --assignee "${ASSIGNEE}"
```

```bash
# output:
[
{
"canDelegate": null,
"id": "/subscriptions/<subscription_guid>/providers/Microsoft.Authorization/roleAssignments/<assignment_guid>",
"name": "<assignment_guid>",
"principalId": "<principal_id>",
[...]

PR #11977, opened 2020-01-29
[RBAC] BREAKING CHANGE: fix issue #11883 az role assignment create scope default to subscription if empty

arrownj:
---

### DIFF
--- a/src/azure-cli/HISTORY.rst
+++ b/src/azure-cli/HISTORY.rst
@@ -3,6 +3,10 @@
 Release History
 ===============
 
+**RBAC**
+
+* [BREAKING CHANGE] Fix issue #11883 az role assignment create scope default to subscription if empty
+
 **ACR**
 
 * [BREAKING CHANGE] `az acr delete` will prompt
--- a/src/azure-cli/azure/cli/command_modules/role/_validators.py
+++ b/src/azure-cli/azure/cli/command_modules/role/_validators.py
@@ -91,6 +91,18 @@ def process_assignment_namespace(cmd, namespace):  # pylint: disable=unused-argu
     if namespace.scope and resource_group and getattr(resource_group, 'is_default', None):
         namespace.resource_group_name = None  # drop configured defaults
 
+    if not namespace.scope and not namespace.resource_group_name:
+        raise CLIError('usage error: please specify at least one of "--scope" and "--resource-group".')
+
+
+def process_list_assignment_namespace(cmd, namespace):  # pylint: disable=unused-argument
+    resource_group = namespace.resource_group_name
+    if namespace.scope and resource_group and getattr(resource_group, 'is_default', None):
+        namespace.resource_group_name = None  # drop configured defaults
+
+    if not namespace.show_all and not namespace.scope and not namespace.resource_group_name:
+        raise CLIError('usage error: please specify at least one of "--all", "--scope" and "--resource-group".')
+
 
 def process_msi_namespace(cmd, namespace):
     get_default_location_from_resource_group(cmd, namespace)
--- a/src/azure-cli/azure/cli/command_modules/role/commands.py
+++ b/src/azure-cli/azure/cli/command_modules/role/commands.py
@@ -12,7 +12,8 @@ from azure.cli.core.commands import CliCommandType
 from ._client_factory import (_auth_client_factory, _graph_client_factory,
                               _msi_user_identities_operations, _msi_operations_operations)
 
-from ._validators import process_msi_namespace, process_assignment_namespace, validate_change_password
+from ._validators import (process_msi_namespace, process_assignment_namespace, process_list_assignment_namespace,
+                          validate_change_password)
 
 
 def transform_definition_list(result):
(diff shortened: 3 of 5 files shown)


src/azure-cli/HISTORY.rst
```diff
@@ -3,6 +3,10 @@
 Release History
 ===============
 
+**RBAC**
+
+* [BREAKING CHANGE] Fix issue #11883 az role assignment create scope default to subscription if empty
```
jiasli: Please follow the history convention like:

> Fix #11883: `az role assignment create`: scope defaults to subscription if not specified

jiasli:
Why the behavior of `az role assignment list` is also changed? There doesn't seem to be any security concern in this command.

Also, it is also a breaking change to force one of `--all`, `--scope` and `-resource-group`, which needs to go to the history.

src/azure-cli/azure/cli/command_modules/role/_validators.py
```diff
@@ -91,6 +91,18 @@ def process_assignment_namespace(cmd, namespace):  # pylint: disable=unused-argu
...
+        raise CLIError('usage error: please specify at least one of "--scope" and "--resource-group".')
+
+
+def process_list_assignment_namespace(cmd, namespace):  # pylint: disable=unused-argument
+    resource_group = namespace.resource_group_name
+    if namespace.scope and resource_group and getattr(resource_group, 'is_default', None):
+        namespace.resource_group_name = None  # drop configured defaults
+
+    if not namespace.show_all and not namespace.scope and not namespace.resource_group_name:
+        raise CLIError('usage error: please specify at least one of "--all", "--scope" and "--resource-group".')
```
jiasli: We usually use single quotes in error messages.

jiasli: Why the behavior of `az role assignment list` is also changed? There doesn't seem to be any security concern in this command.

Also, it is also a breaking change to force one of `--all`, `--scope` and `-resource-group`, which needs to go to the history.

src/azure-cli/azure/cli/command_modules/role/_validators.py
```diff
@@ -91,6 +91,18 @@ def process_assignment_namespace(cmd, namespace):  # pylint: disable=unused-argu
...
         namespace.resource_group_name = None  # drop configured defaults
 
+    if not namespace.scope and not namespace.resource_group_name:
+        raise CLIError('usage error: please specify at least one of "--scope" and "--resource-group".')
+
+
+def process_list_assignment_namespace(cmd, namespace):  # pylint: disable=unused-argument
+    resource_group = namespace.resource_group_name
+    if namespace.scope and resource_group and getattr(resource_group, 'is_default', None):
+        namespace.resource_group_name = None  # drop configured defaults
```
jiasli: This is a duplicate of L90-L92. I'd suggested to extract it to a function like `drop_default_group`.

src/azure-cli/azure/cli/command_modules/role/_validators.py
arrownj: I searched similar error message in our code base, seems we usually use nothing in such case. The error message will be `at least one of --scope and --resource-group` in the new PR.

jiasli:
I think the security concern in the original issue #11883 is when `--scope` is **specified but with an empty argument**: `--scope ""`. Therefore, we don't need to change the behavior (fallback to subscription scope) when `--scope` is **not specified**.

We can error out here: https://github.com/Azure/azure-cli/blob/85f04e71dc4babd1bad7cbd4e0da37080ca4fb29/src/azure-cli/azure/cli/command_modules/role/custom.py#L493

The error can be something like:

```py
    if scope == "":
        raise CLIError('Invalid scope. Please use --help to view the valid format.')
```

One interesting thing I noticed is that the [role assignment REST API](https://docs.microsoft.com/en-us/azure/role-based-access-control/role-assignments-rest) does support empty scope. In that case, the create call is

```http
PUT https://management.azure.com/providers/Microsoft.Authorization/roleAssignments/a5b85ff6-2fbe-454d-bbc3-6a90f9c54d62?api-version=2018-09-01-preview
```

And the response is 
```json
  {
    "canDelegate": null,
    "id": "/providers/Microsoft.Authorization/roleAssignments/a5b85ff6-2fbe-454d-bbc3-6a90f9c54d62",
    "name": "a5b85ff6-2fbe-454d-bbc3-6a90f9c54d62",
    "principalId": "",
    "principalName": "",
    "principalType": "User",
    "roleDefinitionId": "/subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/providers/Microsoft.Authorization/roleDefinitions/b24988ac-6180-42a0-ab88-20f7382dd24c",
    "roleDefinitionName": "Contributor",
    "scope": "/",
    "type": "Microsoft.Authorization/roleAssignments"
  }
```
Notice `"scope": "/"`. But this behavior is not documented. 

Anyway, let's error out for empty scope `""` for now and discuss it furtherly later.

PR #11983, opened 2020-01-29
[RBAC] BREAKING CHANGE: Fix #11883: `az role assignment create`: empty scope will prompt error

### LINKED ISSUE #11883: az role assignment create scope default to subscription if empty

arrownj:
This is a PR to fix issue #11883 

The previous design can cause elevation of privileges. 

In the original PR #11977 , we plan to fix this problem by adding a validation for `--scope`, the validator will promt error when `--scope` is not specified or specified an empty string. And the change will impact `assignment create`, `assignment delete` and `assignment list`. Of cause, this is a breaking change to end users.

As discussed in https://github.com/Azure/azure-cli/pull/11977#issuecomment-579774320, we decided not to change the behavior when `--scope` is not specified.

Just as user said in origin issue(paste below), when `--scope` is not specified, it is OK to use default subscription value. What user concerns is that when `--scope` is specified an empty value, it shouldn't use subscription anymore.

```bash
# login to the portal
az login
# obtain assignee identifier (user in this case)
ASSIGNEE=$(az account show --query 'user.name' -o tsv)

# acceptable version: undefined scope => default scope => applied to subscription, OK
az role assignment create \
--assignee "${ASSIGNEE}" \
--role Contributor

# Why empty (empty, not undefined!) scope equals to default scope? Following examples can cause elevation of privileges.

# version 1: EMPTY scope => applied to subscription (nonintended elevation of privileges)
az role assignment create \
--assignee "${ASSIGNEE}" \
--scope "" \
--role Contributor
```

So we finally change the solution to promt error only when `--scope` is specified an empty value. This is still a breaking chang(to user who specify an empty value to `--scope` and expect subscription  value), however, the impacts are much reduced compared to previous solution.

---


yonzhan:
add to S165.

src/azure-cli/azure/cli/command_modules/role/tests/latest/test_role.py
```diff
@@ -295,6 +296,11 @@ def test_role_assignment_e2e(self, resource_group):
                 self.cmd('role assignment list --assignee {upn}',
                          checks=self.check("length([])", 1))
                 self.cmd('role assignment delete --assignee {upn} --role reader')
+
+                # test role assignment on empty scope
+                with self.assertRaisesRegexp(CLIError, 'Invalid scope. Please use --help to view the valid format.'):
```
jiasli: Usually we just use some phrases in the error message, like "Invalid scope", to have more flexibility of the error message in the future. But this is totally fine to be more strict on the error message. 😉

src/azure-cli/HISTORY.rst
```diff
@@ -3,6 +3,10 @@
 Release History
 ===============
 
+**RBAC**
+
+* [BREAKING CHANGE] Fix #11883: `az role assignment create`: empty scope will prompt error
```
jiasli: Please place this entry in the right place in the history file following alphabetic order.

arrownj: Thanks, updated.

jiasli:
It seems we missed another case where `--resource-group ""`, this will again fallback to subscription level assignment, with even incorrect `principalType`:

```
az role assignment create --assignee-object-id 0d504196-1423-4569-9a6e-15149656f0ee --role "Contributor" --resource-group ""
{
  "canDelegate": null,
  "id": "/subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/providers/Microsoft.Authorization/roleAssignments/4290ea7b-9cca-414f-9f0d-eeb931cc9e40",
  "name": "4290ea7b-9cca-414f-9f0d-eeb931cc9e40",
  "principalId": "00215dd3-8494-4737-8f44-47c950eee2a8",
  "principalName": "",
  "principalType": "ServicePrincipal",
  "roleDefinitionId": "/subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/providers/Microsoft.Authorization/roleDefinitions/b24988ac-6180-42a0-ab88-20f7382dd24c",
  "roleDefinitionName": "Contributor",
  "scope": "/subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590",
  "type": "Microsoft.Authorization/roleAssignments"
}
```

yungezz:
let's create a new issue for the resourcegroup empty string. Worth to check policy assignment/definition also.

PR #12029, opened 2020-02-04
{Compute} Fix typo in error message when attaching unmanaged disk

gaoruifeng:
replace `Pleae` with `Please`, adding missing `s`


jiasli:
Nice catch. Changing title to use curly brackets so that this PR doesn't appear in history.

qwordy:
Thanks for you contribution!

Issue #12003, opened 2020-01-31
Keyvault: Create key via importing - BYOK preview


yungezz:
**Resource Provider**
What is the Azure resource provider your feature is part of?
Keyvault

**Description of Feature or Work Requested**
Provide a brief description of the feature or work requested. A link to conceptual documentation may be helpful too.
To support create key via importing. such as
`az keyvault key create --vault-name bim-kv6 --name mykek --kty RSA-HSM --size 2048 --ops import`

**Minimum API Version Required**
What is the minimum API version of your service required to implement your feature?

**Swagger Link**
Provide a link to the location of your feature(s) in the REST API specs repo. If your feature(s) has corresponding commit or pull request in the REST API specs repo, provide them. This should be on the master branch of the REST API specs repo.
https://github.com/Azure/azure-rest-api-specs/pull/8175/files

**Target Date**
If you have a target date for release of this feature/work, please provide it. While we can't guarantee these dates,
it will help us prioritize your request against other requests.

yonzhan:
Keyvault BYOK

yungezz:
> Keyvault BYOK

? :)

bim-msft:
Supported in PR: #12030 
This feature is only supported on canary regions (region with suffix `euap`, like `eastus2euap`)

PR #12030, opened 2020-02-04
[KeyVault] az keyvault key create: add a new value `import` for parameter `--ops`

bim-msft:
Support for #12003

This is a part of BYOK feature. Currently we use a monkey patch to support new value for `--ops` as the latest data-plane SDK is not released. Once the SDK is released, we should deprecate this monkey patch and use SDK struct directly.

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/keyvault/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/keyvault/_params.py
@@ -2,6 +2,7 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
+from enum import Enum
 
 from argcomplete.completers import FilesCompleter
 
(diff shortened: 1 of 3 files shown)


src/azure-cli/azure/cli/command_modules/keyvault/_params.py
```diff
@@ -41,6 +42,17 @@ def load_arguments(self, _):
...
          resource_type=ResourceType.DATA_KEYVAULT)
 
+    class CLIJsonWebKeyOperation(str, Enum):
+        encrypt = "encrypt"
+        decrypt = "decrypt"
+        sign = "sign"
+        verify = "verify"
+        wrap_key = "wrapKey"
+        unwrap_key = "unwrapKey"
+        import_ = "import"
```
jiasli: Is this a typo?

bim-msft: @jiasli Do you mean `import_`? No, it's not a typo, since `import` is a Python keyword so we can't use it as a variable name.

jiasli: 🚢

jiasli: BTW, this is mentioned in [PEP 8](https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles).

fengzhou-msft:
Will this PR be merged before new data-plane SDK gets released? Do we have estimation of SDK release time?

bim-msft:
> Will this PR be merged before new data-plane SDK gets released? Do we have estimation of SDK release time?

Yes, this need to be merged urgently in order to catch S165 release. We don't know SDK release time yet, service team recommended us to do a monkey patch like this instead of waiting SDK to be released.

PR #12033, opened 2020-02-04
[SQL] Sql midb Add: list-deleted, show-deleted, update-retention, show-retention

djnisic:
Adding new cmdlets that will support short term retention and restore of deleted databases on MI.

---


yonzhan:
add to S165.

jaredmoo:
For `az sql db` I intentionally implemented only `list-deleted` and not `show-deleted` because the resource name of deleted db is a bit weird and getting single deleted db seems like a relatively corner case scenario. What are your thoughts?

src/azure-cli/azure/cli/command_modules/sql/tests/latest/test_sql_commands.py
```diff
@@ -2954,6 +2954,214 @@ def test_sql_mi_tdebyok(self, resource_group, resource_group_location):
...
+
+        # test get short term retention on live database
+        self.cmd('sql midb show-retention -g {rg} --mi {managed_instance_name} -n {database_name}',
+                 checks=[
+                     self.check('resourceGroup', '{rg}'),
+                     self.check('retentionDays', '{retention_days_inc}')])
+
+        # Wait for first backup before dropping
+        if self.in_recording:
+            time.sleep(300)
```
adeal: we should change this to polling to reduce chances of flakiness...you can query the GetDatabase API until EasrliestRestoreDate becomes available. Thoughts?

jaredmoo: Agreed, the existing test for sql db pitr will wait for earliest restore date.

djnisic: Sure, I will add that.

src/azure-cli/azure/cli/command_modules/sql/_help.py
```diff
@@ -596,12 +596,52 @@
     text: az sql midb list -g mygroup --mi myinstance
 """
 
+helps['sql midb list-deleted'] = """
+type: command
+short-summary: Lists restorable deleted managed databases.
```
Juliehzl: Lists -> List

src/azure-cli/azure/cli/command_modules/sql/_help.py
```diff
@@ -596,12 +596,52 @@
...
+examples:
+  - name: List all restorable deleted managed databases on Managed Instance .
+    text: az sql midb list-deleted -g mygroup --mi myinstance
+"""
+
+helps['sql midb show-deleted'] = """
+type: command
+short-summary: Shows restorable deleted managed database.
```
Juliehzl: Shows -> Show

src/azure-cli/azure/cli/command_modules/sql/_help.py
```diff
@@ -596,12 +596,52 @@
...
     text: az sql midb restore -g mygroup --mi myinstance -n mymanageddb --dest-name targetmidb --time "2018-05-20T05:34:22"
+  - name: Restore a dropped managed database using Point in time restore
+    text: az sql midb restore -g mygroup --mi myinstance -n mymanageddb --dest-name targetmidb --time "2018-05-20T05:34:22" --deletion-date "2018-05-20T05:34:22"
+  - name: Restore a live managed database using Point in time restore from another instance
+    text: az sql midb restore -g mygroup --mi myinstance -n mymanageddb --dest-name targetmidb --time "2018-05-20T05:34:22" --dest-mi targetmi --dest-resource-group targetrg
+"""
+
+helps['sql midb update-retention'] = """
+type: command
+short-summary: Updates short term retention for automated backups on a single database.
```
Juliehzl: Updates -> Update

src/azure-cli/azure/cli/command_modules/sql/_help.py
```diff
@@ -596,12 +596,52 @@
...
+examples:
+  - name: Update backup short term retention for live managed database.
+    text: az sql midb update-retention -g mygroup --mi myinstance -n mymanageddb --retention-days retentionindays
+  - name: Update backup short term retention for dropped managed database.
+    text: az sql midb update-retention -g mygroup --mi myinstance -n mymanageddb --deletion-date "2018-05-20T05:34:22" --retention-days retentionindays
+"""
+
+helps['sql midb show-retention'] = """
+type: command
+short-summary: Shows short term retention for automated backups on a single database.
```
Juliehzl: Shows -> Show

src/azure-cli/azure/cli/command_modules/sql/commands.py
```diff
@@ -220,6 +222,17 @@ def load_command_table(self, _):
...
+    restorable_dropped_managed_databases_operations = CliCommandType(
+        operations_tmpl='azure.mgmt.sql.operations#RestorableDroppedManagedDatabasesOperations.{}',
+        client_factory=get_sql_restorable_dropped_managed_databases_operations)
+
+    with self.command_group('sql midb',
+                            restorable_dropped_managed_databases_operations,
+                            client_factory=get_sql_restorable_dropped_managed_databases_operations) as g:
+
+        g.command('list-deleted', 'list_by_instance')
+        g.custom_command('show-deleted', 'show_deleted_database_mi')
```
Juliehzl: The two commands are a little werid. What about add a new parameter `--include-deleted` in existing show and list commands like storage blob command?

src/azure-cli/azure/cli/command_modules/sql/commands.py
```diff
@@ -547,6 +560,17 @@ def load_command_table(self, _):
...
+    managed_backup_short_term_retention_policies_operations = CliCommandType(
+        operations_tmpl='azure.mgmt.sql.operations#ManagedBackupShortTermRetentionPoliciesOperations.{}',
+        client_factory=get_sql_managed_backup_short_term_retention_policies_operations)
+
+    with self.command_group('sql midb',
+                            managed_backup_short_term_retention_policies_operations,
+                            client_factory=get_sql_managed_backup_short_term_retention_policies_operations) as g:
+
+        g.custom_command('update-retention', 'update_short_term_retention_mi', supports_no_wait=True)
+        g.custom_command('show-retention', 'get_short_term_retention_mi')
```
Juliehzl: What about `az sql midb delete-retention update/show` or `az sql midb delete-policy update/show` ?

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -457,6 +474,45 @@ def _get_managed_db_resource_id(cli_ctx, resource_group_name, managed_instance_n
...
+        deletion_date):
+    '''
+    Gets the Managed db resource id in this Azure environment.
+    '''
+
+    # url parse package has different names in Python 2 and 3. 'six' package works cross-version.
+    from six.moves.urllib.parse import quote  # pylint: disable=import-error
+    from azure.cli.core.commands.client_factory import get_subscription_id
+
+    return ('/subscriptions/{}/resourceGroups/{}/providers/Microsoft.Sql/managedInstances/'
```
Juliehzl: You can consider using resource_id in msrestazure. `from msrestazure.tools import resource_id`

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -2498,23 +2555,129 @@ def managed_db_restore(
...
+
+    Custom function makes create mode more convenient.
+    '''
+
+    if deletion_date:
+        database_name = '{},{}'.format(
+            database_name,
+            _to_filetimeutc(deletion_date))
+
+        client = \
```
Juliehzl: why not use client in parameter list here?

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -2498,23 +2555,129 @@ def managed_db_restore(
...
```
Juliehzl: The same as above.

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -2498,23 +2555,129 @@ def managed_db_restore(
...
+            get_sql_restorable_dropped_database_managed_backup_short_term_retention_policies_operations(
+                cmd.cli_ctx,
+                None)
+
+        policy = client.get(
+            restorable_dropped_database_id=database_name,
+            managed_instance_name=managed_instance_name,
+            resource_group_name=resource_group_name)
+    else:
+        policy = client.get(
```
Juliehzl: If deletion date is given, you will return dropped database related information and if not, you will return database information directly, right?
Can live database and dropped database be the same name? If yes, in your current scenario, as soon as deletion date is given, you will only show up the dropped database information, right?

src/azure-cli/azure/cli/command_modules/sql/_help.py
```diff
@@ -596,12 +596,52 @@
...
+examples:
+  - name: List all restorable deleted managed databases on Managed Instance .
+    text: az sql midb list-deleted -g mygroup --mi myinstance
+"""
+
+helps['sql midb show-deleted'] = """
+type: command
+short-summary: Shows restorable deleted managed database.
+examples:
+  - name: Shows specific restorable deleted managed database.
```
Juliehzl: Shows -> Show

src/azure-cli/azure/cli/command_modules/sql/_help.py
```diff
@@ -596,12 +596,52 @@
...
+    text: az sql midb update-retention -g mygroup --mi myinstance -n mymanageddb --deletion-date "2018-05-20T05:34:22" --retention-days retentionindays
+"""
+
+helps['sql midb show-retention'] = """
+type: command
+short-summary: Shows short term retention for automated backups on a single database.
+examples:
+  - name: Shows backup short term retention for live managed database.
+    text: az sql midb show-retention -g mygroup --mi myinstance -n mymanageddb
+  - name: Shows backup short term retention for dropped managed database.
```
Juliehzl: Shows -> Show

Juliehzl:
Another suggestion here is that can we make the new commands in Preview flag?

src/azure-cli/azure/cli/command_modules/sql/commands.py
djnisic: That could work, however I wanted to keep cmdlets for MI similar to the SQLDB ones. As you can see couple of lines above SQLDB is using list-deleted. 

Other reason is that live and dropped databases are not same resource type. In PowerShell that breaks piping but I am not sure if this is an issue in CLI.

src/azure-cli/azure/cli/command_modules/sql/custom.py
djnisic: Correct, there can exist live and multiple dropped databases with the same name. If deletion date is provided I will use client for dropped databases and fetch them, otherwise I will use client for live databases.

djnisic:
> For `az sql db` I intentionally implemented only `list-deleted` and not `show-deleted` because the resource name of deleted db is a bit weird and getting single deleted db seems like a relatively corner case scenario. What are your thoughts?

I think you are right. In order to use show for dropped databases, you would have to use list first and get it's dropped time, which defeats it's purpose.

src/azure-cli/azure/cli/command_modules/sql/commands.py
djnisic: You mean as in renaming current cmdlets or as in adding a new ones for deleted databases. 

If we are talking about altering existing, we are open for suggestions. az sql midb retention-policy update/show could also be an option. However I would avoid delete-retention as this retention is for automated backups on database not database itself.

I wouldn't separate cmdlets for dropped and live databases since policy doesn't change once database is dropped. We might separate them in the future though.

src/azure-cli/azure/cli/command_modules/sql/commands.py
Juliehzl: I see. The reason I recommend such design is that we have similar implementation in storage commands and what we do is add an new parameters as what I have mentioned. And as far as I know, Powershell also has similar design. 
But as you said, in your scenario, it will break piping in Powershell. So there maybe some difference.
In this way, I am fine with the command if you already have similar design.

src/azure-cli/azure/cli/command_modules/sql/commands.py
Juliehzl: I mean rename currrent cmdlets. You can find the option you like. 

Actually, the reason I have such suggestion is that cli do not recommend commands with hyphen for the last part. The only scenario to use hyphen in command is that the abstracted subgroup only has single command, e.g.  `az stroage account show-connection-string`.

Juliehzl:
No other concern to me. @djnisic please resolve my comments ASAP if you want to catch the next realase tomorrow and make sure @jaredmoo approve.

djnisic:
> No other concern to me. @djnisic please resolve my comments ASAP if you want to catch the next realase tomorrow and make sure @jaredmoo approve.

Will do. I am currently running tests with Austin's suggestion. Each tests takes a lot for MI(~4 hours). Regarding new release, I don't think this is that critical for MI, but I will try to get it done today.

djnisic:
@jaredmoo everything should be OK now, please review.

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -1419,11 +1419,17 @@ def _configure_security_policy_storage_params(arg_ctx):
...
         create_args_for_complex_type(
             c, 'parameters', ManagedDatabase, [
+                'deletion_date',
                 'target_managed_database_name',
                 'target_managed_instance_name',
                 'restore_point_in_time'
             ])
 
+        c.argument('deletion_date',
+                   options_list=['--deletion-date'],
```
jaredmoo: Can you align this with `az sql db restore`, which has `--deleted-time`? 

https://docs.microsoft.com/en-us/cli/azure/sql/db?view=azure-cli-latest#az-sql-db-restore

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -67,6 +72,18 @@ def _get_server_location(cli_ctx, server_name, resource_group_name):
         resource_group_name=resource_group_name).location
 
 
+def _get_managed_restorable_dropped_database_backup_short_term_retention_client(cli_ctx):
+    '''
+    Returns the location (i.e. Azure region) that the specified server is in.
```
jaredmoo: Is this comment correct?

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -446,15 +463,59 @@ def _get_managed_db_resource_id(cli_ctx, resource_group_name, managed_instance_n
...
+        subscription=get_subscription_id(cli_ctx),
+        resource_group=resource_group_name,
+        namespace='Microsoft.Sql', type='managedInstances',
+        name=managed_instance_name,
+        child_type_1='databases',
+        child_name_1=database_name)
+
+
+def _to_filetimeutc(dateTime):
+    NET_epoch = datetime(1601, 1, 1)
```
jaredmoo: Add function comment

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -2509,23 +2571,129 @@ def managed_db_restore(
...
+
+
+def show_deleted_database_mi(
+        client,
+        database_name,
+        managed_instance_name,
+        resource_group_name,
+        deletion_date):
+    '''
+    Updates short term retention for database
```
jaredmoo: This comment is also not correct :)

jaredmoo:
Please also make sure that @adeal  approves. Cheers

src/azure-cli/azure/cli/command_modules/sql/_help.py
```diff
@@ -601,12 +601,24 @@
...
+
 helps['sql midb restore'] = """
 type: command
 short-summary: Restore a managed database.
 examples:
-  - name: Restore a managed database using Point in time restore
+  - name: Restore a live managed database using Point in time restore
     text: az sql midb restore -g mygroup --mi myinstance -n mymanageddb --dest-name targetmidb --time "2018-05-20T05:34:22"
+  - name: Restore a dropped managed database using Point in time restore
+    text: az sql midb restore -g mygroup --mi myinstance -n mymanageddb --dest-name targetmidb --time "2018-05-20T05:34:22" --deleted-time "2018-05-20T05:34:22"
```
Juliehzl: What is the difference for --time and --deleted-time?

src/azure-cli/azure/cli/command_modules/sql/_help.py
```diff
@@ -617,6 +629,31 @@
...
 """
 
+helps['sql midb retention-policy'] = """
+type: group
+short-summary: Manage SQL managed instance database backup retention policy.
+"""
+
+helps['sql midb retention-policy update'] = """
+type: command
+short-summary: Update short term retention for automated backups on a single database.
```
Juliehzl: Hi, I am trying to understand your service.  And I am a little confused about the retention policy here. Can you help me better understand your service? I am not sure if your service also has delete retention policy and restore policy. And for your retention policy days, is it defined to specify the days how long the deleted sql database will be retained or restored?

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -1450,8 +1456,28 @@ def _configure_security_policy_storage_params(arg_ctx):
...
+
+        c.argument('deleted_time',
+                   options_list=['--deleted-time'],
+                   help='If specified, updates retention days for a deleted database, instead of an existing database.'
+                   'Must match the deleted time of a deleted database on the source Managed Instance.')
+
+        c.argument('retention_days',
+                   options_list=['--retention-days'],
+                   required=True,
+                   help='New retention policy in days.')
```
Juliehzl: Can you give a more descriptive help here? and is there a valid range for the days?

src/azure-cli/azure/cli/command_modules/sql/_help.py
djnisic: --deleted-time is required to identify dropped database, because primary key to identify dropped database is database name + drop time. For live databases name is good enough, because you cannot have multiple databases with the same name. --time on other had represents the point we are restoring to.

src/azure-cli/azure/cli/command_modules/sql/_help.py
djnisic: We are taking automated backups for every database on MI. Automated backups can be used to restore databases to any point withing the retention period. This policy will determine for how long are we going to keep those backups before deleting them(retention period). 

There is only one policy per database, and it ranges for 7 to 35 for live databases, and 0 to 35 for dropped databases. Note we are going to reduce minimum retention days for live databases to 1 in the future.

src/azure-cli/azure/cli/command_modules/sql/_params.py
djnisic: Sure I can add additional information here.

djnisic:
@adeal ping.

src/azure-cli/azure/cli/command_modules/sql/_help.py
```diff
@@ -601,12 +601,24 @@
...
 helps['sql midb restore'] = """
 type: command
 short-summary: Restore a managed database.
 examples:
-  - name: Restore a managed database using Point in time restore
+  - name: Restore a live managed database using Point in time restore
     text: az sql midb restore -g mygroup --mi myinstance -n mymanageddb --dest-name targetmidb --time "2018-05-20T05:34:22"
+  - name: Restore a dropped managed database using Point in time restore
+    text: az sql midb restore -g mygroup --mi myinstance -n mymanageddb --dest-name targetmidb --time "2018-05-20T05:34:22" --deleted-time "2018-05-20T05:34:22"
+  - name: Restore a live managed database using Point in time restore from another instance
```
adeal: nit: I think it might be more clear to say "Restore a live managed database from another instance using Point in time restore"

src/azure-cli/azure/cli/command_modules/sql/_help.py
adeal: Other clients (PS) have the following as the documentation for this api, do we want consistency?

> Sets a backup short term retention policy.

src/azure-cli/azure/cli/command_modules/sql/_help.py
```diff
@@ -617,6 +629,31 @@
...
+type: command
+short-summary: Update short term retention for automated backups on a single database.
+examples:
+  - name: Update backup short term retention for live managed database.
+    text: az sql midb retention-policy update -g mygroup --mi myinstance -n mymanageddb --retention-days retentionindays
+  - name: Update backup short term retention for dropped managed database.
+    text: az sql midb retention-policy update -g mygroup --mi myinstance -n mymanageddb --deleted-time "2018-05-20T05:34:22" --retention-days retentionindays
+"""
+
+helps['sql midb retention-policy show'] = """
```
adeal: I think we need to be explicit in what retention-policy is being updated. Our naming convention is unfortunate, but it's best to make sure the apis and clients all align. From that, we should use short-term-retention-policy

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -1419,11 +1419,17 @@ def _configure_security_policy_storage_params(arg_ctx):
     with self.argument_context('sql midb restore') as c:
         create_args_for_complex_type(
             c, 'parameters', ManagedDatabase, [
+                'deleted_time',
```
adeal: `deleted_time` does not align with other apis/clients as well. I think it should be `dropped_time`?

jaredmoo: Deleted-time is correct: https://docs.microsoft.com/en-us/cli/azure/sql/db?view=azure-cli-latest#az-sql-db-restore

Consistency with an experience (e.g.: AzCLI, PowerShell, or Portal) is higher priority than consistency across experiences.

jaredmoo: Note that I deliberately chose to use `--deleted-time` when implementing `az sql db restore` because the command is `az sql db delete`, not `az sql db drop`.

adeal: Makes sense and I agree on consistency throughout any one given client. Thanks for the extra context!

src/azure-cli/azure/cli/command_modules/sql/_help.py
djnisic: Sure, will change.

src/azure-cli/azure/cli/command_modules/sql/_help.py
djnisic: You are right, will change update to set.

src/azure-cli/azure/cli/command_modules/sql/_help.py
djnisic: Sure.

djnisic:
@adeal, @Juliehzl all comments addressed. Can we merge pull request now?

djnisic:
Ping!

djnisic:
@jaredmoo, @Juliehzl ping.

PR #12045, opened 2020-02-05
{Compute} Add missing parameter to attach disk example command

gaoruifeng:
Attaching unmanaged disk without --name will cause error.
Fix the example by adding `--name MyDataDisk`

When following the example online, I am getting this error
```
$ az vm unmanaged-disk attach -g my-rg --vm-name my-vm \
>     --vhd-uri https://dummy-storage-account.blob.core.windows.net/container/vm-datadisk.vhd
Pleae provide the name of the existing disk to attach
```

error goes away after adding `--name MyDataDisk`


arrownj:
Hi @qwordy , could you please help review this PR ?

gaoruifeng:
Code also shows disk_name is required when attaching existing disk:

https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/vm/custom.py#L1921,L1922
```
    if not new and not disk_name:
        raise CLIError('Please provide the name of the existing disk to attach')
```

qwordy:
Thanks for your contribution. You are right. It's a mistake in examples.

Issue #11646, opened 2019-12-23
stat: path too long for Windows - error when using large parameters.json file


ajklotz:
## Describe the bug

**Command Name**
`az group deployment create`

**Errors:**
```
stat: path too long for Windows
```

## To Reproduce:
Create a parameters.json file that has 490+ lines with name value pairs. The total character count of said JSON file is 37343. Total file size is 37KB. The structure of the json file looks similar this:

```
{
    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#",
    "contentVersion": "1.0.0.0",
    "parameters": {
        "applicationSettings": {
            "value": {
                "CI": {
                    "AzureWebJobsStorage": "areallylongvaluethatyoumustputhereinordertoreproducetheproblemsojustaddabunchofletterstoproducetheproblemandthenaddthistoeachfieldletmeknowhowitworks",
                    "AzureWebJobsDashboard": "",
                    "AzureWebJobsServiceBus": "",
                    "AzureWebJobsServiceBusSource": ""
                },
                "DEV": {
                    "AzureWebJobsStorage": "",
                    "AzureWebJobsDashboard": "",
                    "AzureWebJobsServiceBus": "",
                    "AzureWebJobsServiceBusSource": ""
                },
                "QA": {
                    "AzureWebJobsStorage": "",
                    "AzureWebJobsDashboard": "",
                    "AzureWebJobsServiceBus": "",
                    "AzureWebJobsServiceBusSource": ""
                },
                "STG": {
                    "AzureWebJobsStorage": "",
                    "AzureWebJobsDashboard": "",
                    "AzureWebJobsServiceBus": "",
                    "AzureWebJobsServiceBusSource": ""
                },
                "TRN": {
                    "AzureWebJobsStorage": "",
                    "AzureWebJobsDashboard": "",
                    "AzureWebJobsServiceBus": "",
                    "AzureWebJobsServiceBusSource": ""
                },
                "PROD": {
                    "AzureWebJobsStorage": "",
                    "AzureWebJobsDashboard": "",
                    "AzureWebJobsServiceBus": "",
                    "AzureWebJobsServiceBusSource": ""
                }
            }
        }
    }
}
```

For each object in applicationSettings, there are 78 name value pairs with a value. When I remove chunks of JSON, the command works. The JSON file is formatted correctly and there are no errors. There appears to be a problem parsing a json file with a lot of content.

- _Put any pre-requisite steps here..._
[...]

PR #12047, opened 2020-02-05
[ARM] az group deployment create: Fix "stat: path too long for Windows" error when using large parameters.json file (#11646)

zhoxing-ms:
Fix issue: #11646

When the prefix of the `--paramters` parameter passed in is `@`, the JSON data will be loaded ahead of time by `_pre_command_table_create()`.
Later in the process, `os.path.isfile()` is called to determine whether the JSON data is a file name . 
When the length of the JSON data is too long, `os.stat()` will have an error of "stat: path too long for Windows" (belonging to ValueError). 
In Python 3.8, the ValueError would be caught, whereas in versions before 3.6, it will make a direct error.
So add the logic to catch ValueError in order to compatible with the older versions of Python.

**History Notes:**  

---


Juliehzl:
A small question here is that "path too long" is caused by file name or file content size?

zhoxing-ms:
> A small question here is that "path too long" is caused by file name or file content size?

@Juliehzl The prefix of the filename contains `@` and the contents of the file are too large will cause this problem.

Juliehzl:
> > A small question here is that "path too long" is caused by file name or file content size?
> 
> @Juliehzl The prefix of the filename contains `@` and the contents of the file are too large will cause this problem.

What about no `@` in parameter value? From my previous understanding, `@` is just used to help find file path in some scenario. And eventually they will call the same load file function. So I think it should also have the same error although there is no `@` in parameter value. Is it right?

zhoxing-ms:
> What about no @ in parameter value? From my previous understanding, @ is just used to help find file path in some scenario. And eventually they will call the same load file function. So I think it should also have the same error although there is no @ in parameter value. Is it right?

@Juliehzl There is no problem if the prefix of the parameter does not contain `@`, because when `os.path.isfile(value)` is called in `_process_parameters()`, the value is a filename.
But when the parameter prefix is `@`, the contents of the file are loaded ahead of time by `_pre_command_table_create()`. Then in the `_process_parameters()` process, `os.path.isfile(value)` is used to determine if the value is a filename, at which point the value has become the contents of the file.

PR #12050, opened 2020-02-05
{Resource} Bump to azure-mgmt-resource~=8.0.1

jiasli:
Bump to `azure-mgmt-resource~=8.0.1` to isolate test hang issue in #12021

### DIFF
--- a/src/azure-cli-core/setup.py
+++ b/src/azure-cli-core/setup.py
@@ -69,7 +69,7 @@ DEPENDENCIES = [
     'requests~=2.20',
     'six~=1.12',
     'wheel==0.30.0',
-    'azure-mgmt-resource~=6.0',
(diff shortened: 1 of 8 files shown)


src/azure-cli-core/setup.py
```diff
@@ -69,7 +69,7 @@
     'requests~=2.20',
     'six~=1.12',
     'wheel==0.30.0',
-    'azure-mgmt-resource~=6.0',
+    'azure-mgmt-resource~=8.0.1',
```
fengzhou-msft: This will lock the version to be 8.0.*. Is this wanted behavior?

jiasli: I understand. We'd better lock it in case we hit issues like https://github.com/Azure/azure-cli/pull/10407 again.

PR #12071, opened 2020-02-06
[CDN] Add CDN WAF commands

lsmith130:
**History Notes:**  

[CDN] Add CDN WAF commands.

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/cdn/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/cdn/_help.py
@@ -95,24 +95,6 @@ examples:
             --origin www.example.com --no-http --enable-compression
 """
 
-helps['cdn endpoint set'] = """
-type: command
-short-summary: Create or update a named endpoint to connect to a CDN.
-examples:
-  - name: Set an endpoint to service content for hostname over HTTP or HTTPS.
-    text: >
-        az cdn endpoint set -g group -n endpoint --profile-name profile \\
-            --origin www.example.com
-  - name: Set an endpoint with a custom domain origin with HTTP and HTTPS ports.
-    text: >
-        az cdn endpoint set -g group -n endpoint --profile-name profile \\
-            --origin www.example.com 88 4444
-  - name: Set an endpoint with a custom domain with compression and only HTTPS.
-    text: >
-        az cdn endpoint set -g group -n endpoint --profile-name profile \\
-            --origin www.example.com --no-http --enable-compression
-"""
-
 helps['cdn endpoint delete'] = """
 type: command
 short-summary: Delete a CDN endpoint.
@@ -290,6 +272,64 @@ examples:
   - name: show delivery rules asscociate with the endpoint.
     text: >
         az cdn endpoint rule action show -g group --profile-name profile-name -n endpoint
+"""
+
+helps['cdn endpoint waf'] = """
+type: group
+short-summary: Manage WAF properties of a CDN endpoint.
+"""
+
+helps['cdn endpoint waf policy'] = """
+type: group
+short-summary: Apply a CDN WAF policy to a CDN endpoint.
+"""
+
+helps['cdn endpoint waf policy set'] = """
+type: command
+short-summary: Set the CDN WAF policy applied to a CDN endpoint
+parameters:
+  - name: --waf-policy-id
+    type: string
+    short-summary: >
+        The Resource ID of the CDN WAF policy to apply to this endpoint.
+  - name: --waf-policy-subscription-id
+    type: string
+    short-summary: >
+        The Resource ID of the CDN WAF policy to apply to this endpoint. ignored
+        if --waf-policy-id is set.
+  - name: --waf-policy-resource-group-name
+    type: string
+    short-summary: >
+        The resource group of the CDN WAF policy to apply to this endpoint.
+        Ignored if --waf-policy-id is set.
+  - name: --waf-policy-name
+    type: string
+    short-summary: >
+        The name of the CDN WAF policy to apply to this endpoint. Ignored if
+        --waf-policy-id is set.
+examples:
+  - name: Set the CDN WAF policy applied to a CDN endpoint by WAF Policy name.
+    text: >
+        az cdn endpoint waf policy set -g group --endpoint-name endpoint \\
+            --profile-name profile --waf-policy-subscription-id xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx \\
+"""
+
+helps['cdn endpoint waf policy remove'] = """
+type: command
+short-summary: Remove a CDN WAF policy from a CDN endpoint.
+examples:
+  - name: Remove a CDN WAF policy from a CDN endpoint.
+    text: >
+        az cdn endpoint waf policy remove -g group --endpoint-name endpoint --profile-name profile
+"""
+
+helps['cdn endpoint waf policy show'] = """
+type: command
+short-summary: Show which CDN WAF policy is applied to a CDN endpoint.
+examples:
+  - name: Show which CDN WAF policy is applied to a CDN endpoint.
+    text: >
+        az cdn endpoint waf policy show -g group --endpoint-name endpoint --profile-name profile
 """
 
 helps['cdn origin'] = """
@@ -573,7 +613,7 @@ parameters:
   - name: --action
     type: string
     short-summary: >
-        The action to take when the rule is matched. Valid values are 'Block', 'Redirect', 'Allow', and 'Log'.
+        The action to take when the rule is matched.
   - name: --match-condition -m
     type: string
     short-summary: Conditions used to determine if the rule is matched for a request.
@@ -642,7 +682,7 @@ parameters:
   - name: --action
     type: string
     short-summary: >
-        The action to take when the rule is matched. Valid values are 'Block', 'Redirect', 'Allow', and 'Log'.
+        The action to take when the rule is matched.
   - name: --match-condition -m
     type: string
     short-summary: Conditions used to determine if the rule is matched for a request.
--- a/src/azure-cli/azure/cli/command_modules/cdn/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/cdn/_help.py
@@ -392,52 +392,6 @@ type: group
 short-summary: Manage CDN WAF policies.
 """
 
-helps['cdn waf managed-rule-set'] = """
-type: group
-short-summary: View available CDN WAF managed rule sets.
-"""
-
-helps['cdn waf managed-rule-set list'] = """
-type: command
-short-summary: List available CDN WAF managed rule sets.
-examples:
-  - name: List all available CDN WAF managed rule sets.
-    text: az cdn waf managed-rule-set list
-"""
-
-helps['cdn waf managed-rule-set show'] = """
-type: command
-short-summary: List available CDN WAF managed rule sets.
-examples:
-  - name: Show the CDN WAF managed rule set DefaultRuleSet_1.0.
-    text: az cdn waf managed-rule-set show --rule-set-type DefaultRuleSet --rule-set-version 1.0
-"""
-
-helps['cdn waf managed-rule-set rule-group'] = """
-type: group
-short-summary: View available rule groups of a CDN WAF managed rule set.
-"""
-
-helps['cdn waf managed-rule-set rule-group list'] = """
-type: command
-short-summary: List available CDN WAF managed rule sets.
-examples:
-  - name: List available rule groups for DefaultRuleSet_1.0.
-    text: |
-      az cdn waf managed-rule-set rule-group list \\
-        --rule-set-type DefaultRuleSet --rule-set-version 1.0
-"""
-
-helps['cdn waf managed-rule-set rule-group show'] = """
-type: command
-short-summary: List available CDN WAF managed rule sets.
-examples:
-  - name: Show the SQLI rule group of DefaultRuleSet_1.0.
-    text: |
-      az cdn waf managed-rule-set rule-group show \\
-        --rule-set-type DefaultRuleSet --rule-set-version 1.0 -n SQLI
-"""
-
 helps['cdn waf policy set'] = """
 type: command
 short-summary: Create a new CDN WAF policy.
@@ -540,6 +494,23 @@ examples:
           --rule-set-type DefaultRuleSet --rule-set-version 1.0
 """
 
+helps['cdn waf policy managed-rule-set list-available'] = """
+type: command
+short-summary: List available CDN WAF managed rule sets.
+examples:
+  - name: List all available CDN WAF managed rule sets.
+    text: az cdn waf policy managed-rule-set list-available
+"""
+
+helps['cdn waf policy managed-rule-set show-available'] = """
+type: command
+short-summary: List available CDN WAF managed rule sets.
+examples:
+  - name: Show the CDN WAF managed rule set DefaultRuleSet_1.0.
+    text: az cdn waf policy managed-rule-set show-available \\
+            --rule-set-type DefaultRuleSet --rule-set-version 1.0
+"""
+
 helps['cdn waf policy managed-rule-set rule-group-override'] = """
 type: group
 short-summary: Manage rule group overrides of a managed rule on a CDN WAF policy.
@@ -547,10 +518,10 @@ short-summary: Manage rule group overrides of a managed rule on a CDN WAF policy
 
 helps['cdn waf policy managed-rule-set rule-group-override set'] = """
 type: command
-short-summary: Add a rule group override to a managed rule set on a CDN WAF policy.
+short-summary: Add or update a rule group override to a managed rule set on a CDN WAF policy.
 parameters:
   - name: --rule-override -r
-    short-summary: Override a rule in the rule group
+    short-summary: Override a rule in the rule group.
     long-summary: |
         rule overrides are specified as key value pairs in the form "KEY=VALUE [KEY=VALUE ...]".
         Available keys are 'id', 'action', and 'enabled'. 'id' is required. Valid values for
@@ -600,6 +571,25 @@ examples:
           --rule-set-type DefaultRuleSet --rule-set-version 1.0 -n SQLI
 """
 
+helps['cdn waf policy managed-rule-set rule-group-override list-available'] = """
+type: command
+short-summary: List available CDN WAF managed rule groups of a managed rule set.
+examples:
+  - name: List available rule groups for DefaultRuleSet_1.0.
+    text: |
+      az cdn waf policy managed-rule-set rule-group-override list-available \\
+        --rule-set-type DefaultRuleSet --rule-set-version 1.0
+"""
+
+helps['cdn waf policy managed-rule-set rule-group-override show-available'] = """
+type: command
+short-summary: Show an available CDN WAF managed rule group.
+examples:
+  - name: Show the SQLI rule group of DefaultRuleSet_1.0.
+    text: |
+      az cdn waf policy managed-rule-set rule-group-override show-available \\
+        --rule-set-type DefaultRuleSet --rule-set-version 1.0 -n SQLI
+"""
 
 helps['cdn waf policy custom-rule'] = """
 type: group
--- a/src/azure-cli/azure/cli/command_modules/cdn/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/cdn/_help.py
@@ -40,6 +40,10 @@ examples:
     text: >
         az cdn custom-domain create -g group --endpoint-name endpoint --profile-name profile \\
             -n domain-name --hostname www.example.com
+  - name: Enable custom https with a minimum
+    text: >
+        az cdn custom-domain create -g group --endpoint-name endpoint --profile-name profile \\
+            -n domain-name --hostname www.example.com
 """
 
 helps['cdn custom-domain delete'] = """
@@ -75,6 +79,7 @@ short-summary: Manage CDN endpoints.
 helps['cdn endpoint create'] = """
 type: command
 short-summary: Create a named endpoint to connect to a CDN.
+
 examples:
   - name: Create an endpoint to service content for hostname over HTTP or HTTPS.
     text: >
@@ -87,6 +92,24 @@ examples:
   - name: Create an endpoint with a custom domain with compression and only HTTPS.
     text: >
         az cdn endpoint create -g group -n endpoint --profile-name profile \\
+            --origin www.example.com --no-http --enable-compression
+"""
+
+helps['cdn endpoint set'] = """
+type: command
+short-summary: Create or update a named endpoint to connect to a CDN.
+examples:
+  - name: Set an endpoint to service content for hostname over HTTP or HTTPS.
+    text: >
+        az cdn endpoint set -g group -n endpoint --profile-name profile \\
+            --origin www.example.com
+  - name: Set an endpoint with a custom domain origin with HTTP and HTTPS ports.
+    text: >
+        az cdn endpoint set -g group -n endpoint --profile-name profile \\
+            --origin www.example.com 88 4444
+  - name: Set an endpoint with a custom domain with compression and only HTTPS.
+    text: >
+        az cdn endpoint set -g group -n endpoint --profile-name profile \\
             --origin www.example.com --no-http --enable-compression
 """
 
@@ -170,12 +193,12 @@ short-summary: Add a delivery rule to a CDN endpoint.
 examples:
   - name: Create a global rule to disable caching.
     text: >
-        az cdn endpoint rule add -g group -n endpoint --profile-name profile --order 0\\
+        az cdn endpoint rule add -g group -n endpoint --profile-name profile --order 0 \\
             --rule-name global --action-name CacheExpiration --cache-behavior BypassCache
   - name: Create a rule for http to https redirect
     text: >
-        az cdn endpoint rule add -g group -n endpoint --profile-name profile --order 1\\
-            --rule-name "redirect" --match-variable RequestScheme --operator Equal --match-values HTTPS\\
+        az cdn endpoint rule add -g group -n endpoint --profile-name profile --order 1 \\
+            --rule-name "redirect" --match-variable RequestScheme --operator Equal --match-values HTTPS \\
             --action-name "UrlRedirect" --redirect-protocol Https --redirect-type Moved
 """
 
@@ -185,7 +208,7 @@ short-summary: Remove a delivery rule from an endpoint.
 examples:
   - name: Remove the global rule.
     text: >
-        az cdn endpoint rule remove -g group -n endpoint --profile-name profile --rule-name Global\\
+        az cdn endpoint rule remove -g group -n endpoint --profile-name profile --rule-name Global
 """
 
 helps['cdn endpoint rule show'] = """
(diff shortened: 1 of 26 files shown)


yonzhan:
add to S165.

lsmith130:
@Juliehzl I've fixed merge conflicts and CI errors on this PR and it is ready for review.

src/azure-cli/azure/cli/command_modules/cdn/_params.py
```diff
@@ -154,3 +140,55 @@ def load_arguments(self, _):
...
+        c.argument('policy_name', name_arg_type, id_part='name', help='The name of the CDN WAF policy.')
+
+    with self.argument_context('cdn waf policy managed-rule-set') as c:
+        c.argument('policy_name', help='Name of the CDN WAF policy.')
+        c.argument('rule_set_type', help='Type name of the managed rule set.')
+        c.argument('rule_set_version', help='Version the managed rule set.')
+    with self.argument_context('cdn waf policy managed-rule-set add') as c:
+        c.argument('enabled', arg_type=get_three_state_flag())
+
+    with self.argument_context('cdn waf policy managed-rule-set rule-group-override') as c:
```
Juliehzl: What do you mean for rule-group-override?

src/azure-cli/azure/cli/command_modules/cdn/_params.py
```diff
@@ -154,3 +140,55 @@ def load_arguments(self, _):
...
+        c.argument('name', name_arg_type, id_part='name', help='The name of the rule group override.')
+        c.argument('policy_name', help='Name of the CDN WAF policy.')
+        c.argument('rule_set_type', help='Type name of the managed rule set.')
```
Juliehzl: Which types are allowed here?

src/azure-cli/azure/cli/command_modules/cdn/_params.py
```diff
@@ -154,3 +140,55 @@ def load_arguments(self, _):
...
+        c.argument('rule_set_version', help='Version of the managed rule set.')
+    with self.argument_context('cdn waf policy managed-rule-set rule-group-override set') as c:
+        c.argument('rule_overrides',
+                   options_list=['-r', '--rule-override'],
+                   action=ManagedRuleOverrideAction,
+                   nargs='+')
+
+    with self.argument_context('cdn waf policy custom-rule') as c:
```
Juliehzl: What is the difference between custom-rule and rate-limit-rule? Why not make the one subgroup named "rule"?

Juliehzl:
@lsmith130 It looks you forget to change version in some requirements.txt files, e.g. requirements.py2.Darwin.txt, requirements.py2.Linux.txt, requirements.py2.Windows.txt

src/azure-cli/azure/cli/command_modules/cdn/_params.py
lsmith130: In our service, you can use a rule group override to override the default rules applied by adding a managed rule set to a policy. I considered shortening this to `cdn waf policy managed-rule-set rule-group`, but I wanted to make it clear that overriding the defaults of one policy doesn't modify the managed rule set which may be applied to other policies. This also follows the naming convention used by our service's swagger API.

src/azure-cli/azure/cli/command_modules/cdn/_params.py
lsmith130: rate-limit-rule and custom-rule have the same format except that rate-limit-rule has two extra required fields and have different behavior for determining when a rule's action is applied. There is also a third type of rule, managed rules, which are grouped as a managed-rule-set. The reasons why I didn't move these three types of rules under one subgroup are:

1. I thought it would be confusing if managed rules didn't fall under the "rule" subgroup as well, but because they are defined under rule sets it seems odd to have the rule above the rule set. If you're ok with managed rules not falling under the "rule" subgroup then this concern doesn't apply.

2. Our service and documentation refer to the titles of these properties as "Managed Rules" "Custom Rules" and "Rate Limit Rules". We don't typically refer to all of them together as just "rules".

If you still think it would be better to move two of the three types of rules under a "rule" subgroup so the commands are `az cdn waf policy rule custom <action>`, `az cdn waf policy rule rate-limit <action>`, and `az cdn waf policy managed-rule-set rule-group set --rule <rule definition>`, I'm happy to make the change.

src/azure-cli/azure/cli/command_modules/cdn/_params.py
lsmith130: Available rule sets and rule groups may be added dynamically by our service without updating the client. I've updated the help text for --rule-set-type, --rule-set-version, and the --name of rule-groups to include the command to list the currently available options.

lsmith130:
> @lsmith130
> It looks you forget to change version in some requirements.txt files, e.g. requirements.py2.Darwin.txt, requirements.py2.Linux.txt, requirements.py2.Windows.txt

@Juliehzl, I originally updated the py2 requirements files, but earlier today I resolved a merge conflict where I had updated them but Azure/azure-cli:dev had deleted them, so I kept the deleted version. I also noticed that python2 checks have been removed from the CI pipeline. I assumed they have been removed because python2 reached end-of-life on January 1, 2020. Is my assumption correct?

lsmith130:
@Juliehzl, could you please give review my responses to your comments when you can? If I need to make changes I'd like to know early enough that I have time to make needed changes before the code complete deadline of S166. Thanks!

src/azure-cli/azure/cli/command_modules/cdn/tests/latest/test_waf_scenarios.py
```diff
@@ -0,0 +1,447 @@
...
+        with self.assertRaises(CLIError):
+            self.cmd('cdn waf policy rate-limit-rule show -g {resource_group} --policy-name {policy} -n {rate_limit_rule_1}')
+
+        # Delete policy
+        self.cmd('cdn waf policy delete -g {resource_group} -n {policy}')
+
+
+class CdnWafEndpointLinkScenarioTest(CdnScenarioMixin, ScenarioTest):
+    @ResourceGroupPreparer(name_prefix='cli_test_cdn_domain')
+    def test_cdn_waf_endpoint_link(self, resource_group):
```
MyronFanQiu: This is not related with this PR. right?

src/azure-cli/azure/cli/command_modules/cdn/_help.py
```diff
@@ -323,3 +346,360 @@
...
+        az cdn waf policy managed-rule-set show -g group --policy-name policy \\
+          --rule-set-type DefaultRuleSet --rule-set-version 1.0
+"""
+
+helps['cdn waf policy managed-rule-set rule-group-override'] = """
+type: group
+short-summary: Manage rule group overrides of a managed rule on a CDN WAF policy.
+"""
+
+helps['cdn waf policy managed-rule-set rule-group-override set'] = """
```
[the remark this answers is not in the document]
lsmith130: Should I use `add` even if it is modifying an existing rule-group-override?

src/azure-cli/azure/cli/command_modules/cdn/_help.py
```diff
@@ -323,3 +346,360 @@
...
+          --rule-set-type DefaultRuleSet --rule-set-version 1.0 -n SQLI
+"""
+
+
+helps['cdn waf policy custom-rule'] = """
+type: group
+short-summary: Manage custom rules of a CDN WAF policy.
+"""
+
+helps['cdn waf policy custom-rule set'] = """
```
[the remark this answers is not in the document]
lsmith130: Again, this has create-or-update semantics because we are required to use PUT requests. I saw in the docs that create-or-update semantics should use set, but if this doesn't apply to sub-collections then I'll change it to add.

src/azure-cli/azure/cli/command_modules/cdn/commands.py
```diff
@@ -136,3 +156,43 @@ def _inner_not_found(ex):
 
     with self.command_group('cdn edge-node', cdn_edge_sdk) as g:
         g.command('list', 'list')
+
+    with self.command_group('cdn waf managed-rule-set', cdn_waf_ruleset_sdk, is_preview=True) as g:
```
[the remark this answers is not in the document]
lsmith130: `cdn waf managed-rule-set` provides read-only commands to view the currently available managed rule sets. `cdn waf policy managed-rule-set` provides commands to add and remove the available rulesets to a particular policy.

src/azure-cli/azure/cli/command_modules/cdn/tests/latest/test_waf_scenarios.py
lsmith130: This is related - this tests adding a CDN WAF policy to a CDN endpoint, which is new functionality I've added in this PR.

src/azure-cli/azure/cli/command_modules/cdn/_help.py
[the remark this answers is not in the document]
MyronFanQiu: yes. I got your point at last. It's ok to use set. 😄

src/azure-cli/azure/cli/command_modules/cdn/commands.py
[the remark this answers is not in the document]
MyronFanQiu: Sure. I got your point. What I want to know is that these two `managed-rule-set`s are the same things or not? I'm a little bit confused to see two `managed-rule-set` appear under two different sub command group if they are same thing.

src/azure-cli/azure/cli/command_modules/cdn/commands.py
```diff
@@ -136,3 +156,43 @@ def _inner_not_found(ex):
...
+        g.custom_show_command('show', 'show_waf_managed_rule_group', client_factory=cf_waf_rule_set)
+
+    with self.command_group('cdn waf policy', cdn_waf_policy_sdk, is_preview=True) as g:
+        g.show_command('show', 'get')
+        g.command('list', 'list')
+        g.custom_command('set', 'set_waf_policy', client_factory=cf_waf_policy)
+        g.command('delete', 'delete')
+
+    with self.command_group('cdn waf policy managed-rule-set', cdn_waf_policy_sdk, is_preview=True) as g:
+        g.custom_command('add', 'add_waf_policy_managed_rule_set', client_factory=cf_waf_policy)
```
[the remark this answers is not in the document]
lsmith130: I thought so as well, but when I wrote it that way originally I got an error that there was no `client` parameter to call the custom functions with. After deeper inspection, all existing `custom_command` calls in CDN are supplying the `client_factory` explicitly for each command.

src/azure-cli/azure/cli/command_modules/cdn/commands.py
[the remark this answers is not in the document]
lsmith130: They are the same. Policies contain "links" to certain managed rule sets. Would it be less confusing if I changed the name of the `cdn waf policy managed-rule-set` group to `cdn waf policy applied-managed-rule-set` or `cdn waf policy managed-rule-set-link`?

yonzhan:
add to S166.

src/azure-cli/azure/cli/command_modules/cdn/_params.py
```diff
@@ -76,6 +53,13 @@ def load_arguments(self, _):
                         'user requests for a compressed version. Content won\'t be compressed '
                         'on CDN when requested content is smaller than 1 byte or larger than 1 '
                         'MB.')
+        c.argument('waf_policy_link', help='The ResourceId of the CDN WAF Policy this endpoint is linked to.')
```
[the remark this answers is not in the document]
lsmith130: Based on your feedback, I've moved managing this field to a new command subgroup `az cdn endpoint waf policy`. `az cdn endpoint waf policy set` allows to specify the CDN WAF policy by either id or name, and `az cdn endpoint waf policy remove` allows the user to remove a WAF policy without dealing with the Azure python sdk limitation that fields cannot be unset using PATCH. The field will still be present on retrieved endpoints, but I've added an `az cdn ednpoint waf policy show` command for consistency.

src/azure-cli/azure/cli/command_modules/cdn/_help.py
```diff
@@ -323,3 +346,360 @@
...
+"""
+
+helps['cdn waf policy rate-limit-rule set'] = """
+type: command
+short-summary: Add a rate limit rule to a CDN WAF policy.
+parameters:
+  - name: --action
+    type: string
+    short-summary: >
+        The action to take when the rule is matched. Valid values are 'Block', 'Redirect', 'Allow', and 'Log'.
```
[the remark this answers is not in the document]
lsmith130: I've made the change.

src/azure-cli/azure/cli/command_modules/cdn/custom.py
```diff
@@ -444,3 +459,299 @@ def create_profile(client, resource_group_name, name,
...
+        raise CLIError("managed rule set type '{}' version '{}' is already added to WAF policy '{}'"
+                       .format(rule_set_type, rule_set_version, policy_name))
+
+    # Add the managed rule set to the policy.
+    policy.managed_rules.managed_rule_sets.append(ManagedRuleSet(rule_set_type=rule_set_type,
+                                                                 rule_set_version=rule_set_version))
+    result = client.create_or_update(resource_group_name, policy_name, policy).result()
+
+    # Return the new managed rule set from the updated policy.
+    for r in result.managed_rules.managed_rule_sets:
```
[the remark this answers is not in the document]
lsmith130: Good catch, I've changed this to use `_find_policy_managed_rule_set`.

lsmith130:
@MyronFanQiu, I've made the requested changes and tried to answer your questions - please review again.

src/azure-cli/azure/cli/command_modules/cdn/commands.py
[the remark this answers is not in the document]
MyronFanQiu: right. I find the reason because you're using custom_command. You can set `custom_command_type=cdn_waf_policy_sdk` in `self.command_group`

src/azure-cli/azure/cli/command_modules/cdn/_help.py
```diff
@@ -323,3 +386,350 @@
...
+  - name: --action
+    type: string
+    short-summary: >
+        The action to take when the rule is matched.
+  - name: --match-condition -m
+    type: string
+    short-summary: Conditions used to determine if the rule is matched for a request.
+    long-summary: >
+        Match conditions are specified as key value pairs in the form "KEY=VALUE [KEY=VALUE ...]".
+        Available keys are 'match-variable', 'operator', 'match-value', 'selector', 'negate', and
```
MyronFanQiu: Hi. How can user know the valid values for each key?

src/azure-cli/azure/cli/command_modules/cdn/_help.py
```diff
@@ -323,3 +386,350 @@
...
```
MyronFanQiu: same here.

src/azure-cli/azure/cli/command_modules/cdn/_params.py
```diff
@@ -154,3 +138,60 @@ def load_arguments(self, _):
...
+                   options_list=['-r', '--rule-override'],
+                   action=ManagedRuleOverrideAction,
+                   nargs='+')
+
+    with self.argument_context('cdn waf policy custom-rule') as c:
+        c.argument('name', name_arg_type, id_part='', help='The name of the custom rule.')
+        c.argument('policy_name', id_part='name', help='Name of the CDN WAF policy.')
+    with self.argument_context('cdn waf policy custom-rule list') as c:
+        # List commands cannot use --ids flag
+        c.argument('policy_name', id_part='')
```
MyronFanQiu: Usually we set `id_part=None`

src/azure-cli/azure/cli/command_modules/cdn/_help.py
```diff
@@ -40,6 +40,10 @@
     text: >
         az cdn custom-domain create -g group --endpoint-name endpoint --profile-name profile \\
             -n domain-name --hostname www.example.com
+  - name: Enable custom https with a minimum
+    text: >
+        az cdn custom-domain create -g group --endpoint-name endpoint --profile-name profile \\
+            -n domain-name --hostname www.example.com
```
haroldrandom: a single `\` will appear in the line. It's YAML restriced.
If you want to make the example stay in a line, pelease remove the `\\`.
If you want to the examples wrap like you want, please use `text: |` instead of `>`

src/azure-cli/azure/cli/command_modules/cdn/_help.py
```diff
@@ -75,6 +79,7 @@
 helps['cdn endpoint create'] = """
 type: command
 short-summary: Create a named endpoint to connect to a CDN.
+
```
haroldrandom: trivial empty line

src/azure-cli/azure/cli/command_modules/cdn/_help.py
haroldrandom: mamy case below.

lsmith130: I decided to remove the back slashes altogether since not all shells use backslash as a line continuation character. Thanks for the input.

src/azure-cli/azure/cli/command_modules/cdn/_help.py
lsmith130: Fixed

src/azure-cli/azure/cli/command_modules/cdn/_help.py
lsmith130: I've updated the long-summary to include this information.

src/azure-cli/azure/cli/command_modules/cdn/_params.py
lsmith130: Fixed

src/azure-cli/azure/cli/command_modules/cdn/_params.py
```diff
@@ -154,3 +138,60 @@ def load_arguments(self, _):
...
+                   options_list=['-r', '--rule-override'],
+                   action=ManagedRuleOverrideAction,
+                   nargs='+')
+
+    with self.argument_context('cdn waf policy custom-rule') as c:
+        c.argument('name', name_arg_type, id_part=None, help='The name of the custom rule.')
+        c.argument('policy_name', id_part='name', help='Name of the CDN WAF policy.')
+    with self.argument_context('cdn waf policy custom-rule list') as c:
+        # List commands cannot use --ids flag
+        c.argument('policy_name', id_part=None)
```
Juliehzl: It looks the id_part here doesn't work. You don't need to add `id_part` kwargs here.

src/azure-cli/azure/cli/command_modules/cdn/_validators.py
```diff
@@ -17,3 +17,7 @@ def check_port_range(port, msg):
             check_port_range(origin.http_port, msg.format('HTTP', origin.name))
             check_port_range(origin.https_port, msg.format('HTTPS', origin.name))
     return True
+
+
+def validate_priority(namespace):
+    return 0 <= namespace.priority <= 1000
```
Juliehzl: with the validator, when priority is not in [0,1000], no error throws, right? This function only return False for the scenario that priority is not in [0,1000].

src/azure-cli/azure/cli/command_modules/cdn/commands.py
```diff
@@ -136,3 +151,39 @@ def _inner_not_found(ex):
...
+        g.custom_command('remove', 'remove_waf_policy_managed_rule_set', client_factory=cf_waf_policy)
+        g.custom_command('list', 'list_waf_policy_managed_rule_sets', client_factory=cf_waf_policy)
+        g.custom_show_command('show', 'show_waf_policy_managed_rule_set', client_factory=cf_waf_policy)
+        g.custom_command('list-available', 'list_waf_managed_rule_set', client_factory=cf_waf_rule_set)
+        g.custom_command('show-available', 'show_waf_managed_rule_set', client_factory=cf_waf_rule_set)
```
Juliehzl: `list-available` can be used for the scenario to list all available resources. What do you mean for `show-available`? Typically we only use `show` command to show the specified resource. May I know why there is a `show-available` command?

src/azure-cli/azure/cli/command_modules/cdn/commands.py
```diff
@@ -136,3 +151,39 @@ def _inner_not_found(ex):
...
+
+    with self.command_group('cdn waf policy managed-rule-set rule-group-override',
+                            cdn_waf_policy_sdk,
+                            is_preview=True) as g:
+        g.custom_command('set', 'set_waf_managed_rule_group_override', client_factory=cf_waf_policy)
+        g.custom_command('delete', 'delete_waf_managed_rule_group_override', client_factory=cf_waf_policy)
```
Juliehzl: it will be better to have `confirmation` for delete command.

src/azure-cli/azure/cli/command_modules/cdn/commands.py
```diff
@@ -136,3 +151,39 @@ def _inner_not_found(ex):
...
+        g.custom_command('list', 'list_waf_policy_managed_rule_group_overrides', client_factory=cf_waf_policy)
+        g.custom_show_command('show', 'show_waf_managed_rule_group_override', client_factory=cf_waf_policy)
+        g.custom_command('list-available', 'list_waf_managed_rule_groups', client_factory=cf_waf_rule_set)
+        g.custom_command('show-available', 'show_waf_managed_rule_group', client_factory=cf_waf_rule_set)
```
Juliehzl: The same as above

src/azure-cli/azure/cli/command_modules/cdn/commands.py
```diff
@@ -136,3 +151,39 @@ def _inner_not_found(ex):
...
+
+    with self.command_group('cdn waf policy custom-rule', cdn_waf_policy_sdk, is_preview=True) as g:
+        g.custom_command('set', 'set_waf_custom_rule', client_factory=cf_waf_policy)
+        g.custom_command('delete', 'delete_waf_custom_rule', client_factory=cf_waf_policy)
```
Juliehzl: confirmation

src/azure-cli/azure/cli/command_modules/cdn/commands.py
```diff
@@ -136,3 +151,39 @@ def _inner_not_found(ex):
...
+        g.custom_command('list', 'list_waf_custom_rules', client_factory=cf_waf_policy)
+        g.custom_show_command('show', 'show_waf_custom_rule', client_factory=cf_waf_policy)
+
+    with self.command_group('cdn waf policy rate-limit-rule', cdn_waf_policy_sdk, is_preview=True) as g:
+        g.custom_command('set', 'set_waf_rate_limit_rule', client_factory=cf_waf_policy)
+        g.custom_command('delete', 'delete_waf_rate_limit_rule', client_factory=cf_waf_policy)
```
Juliehzl: confirmation

src/azure-cli/azure/cli/command_modules/cdn/custom.py
```diff
@@ -94,6 +100,60 @@ def update_endpoint(instance,
...
+                                  resource_group_name: str,
+                                  profile_name: str,
+                                  endpoint_name: str):
+
+    from azure.mgmt.cdn.models import (EndpointPropertiesUpdateParametersWebApplicationFirewallPolicyLink)
+
+    link = client.get(resource_group_name, profile_name, endpoint_name).web_application_firewall_policy_link
+    if link is not None:
+        return link
+    return EndpointPropertiesUpdateParametersWebApplicationFirewallPolicyLink(id=None)
```
Juliehzl: Can you help understand this part?
When link resource cannot be got, we will get Error, right? And in this way, we should just throw this error and with `custom_show_command` build in exception handler, cli will exit with exit code 3 and error message.

Why do you want to return such object?

src/azure-cli/azure/cli/command_modules/cdn/commands.py
lsmith130: The show available command shows one of the available managed rule sets, retrieved by name. Managed rule sets are deeply nested resources, and it is helpful to retrieve one individually for inspecting available rule groups and rules within the managed rule set rather than sifting through the list of all of them. Please let me know if there is some issue with `show-available` and I can just remove it for now in order to expedite this release.

src/azure-cli/azure/cli/command_modules/cdn/commands.py
lsmith130: Same as above, please let me know if there is any issue here and I'll just remove it for now to unblock this PR.

src/azure-cli/azure/cli/command_modules/cdn/_validators.py
lsmith130: Fixed

src/azure-cli/azure/cli/command_modules/cdn/commands.py
lsmith130: Fixed

src/azure-cli/azure/cli/command_modules/cdn/commands.py
lsmith130: Fixed

src/azure-cli/azure/cli/command_modules/cdn/commands.py
lsmith130: Fixed

src/azure-cli/azure/cli/command_modules/cdn/custom.py
lsmith130: If the endpoint exists but there is no link, `client.get(resource_group_name, profile_name, endpoint_name).web_application_firewall_policy_link` will simply return `None`, which `custom_show_command` does not handle. I've changed it to raise a CLIError instead of an empty WAF policy link model object in this case, which is the practice used by the cli core util function `find_child_item`.

src/azure-cli/azure/cli/command_modules/cdn/commands.py
Juliehzl: I think you want to show all available rule groups in a specific rule set, right? 
In this way, you can make your command design like `az cdn waf policy managed-rule-set group list` to show all rule groups information in a rule set.

src/azure-cli/azure/cli/command_modules/cdn/commands.py
Juliehzl: Is there any special design for rule-group-override? What is the difference between override and standard rule group?

lsmith130: Managed rules (including managed rule sets and rule groups) are defined by our service, and cannot be modified. The default settings of a rule group can be overridden for a particular policy, but this does not change the same managed rules which may be applied to a different policy. For this reason, the service refers to these as Rule Group Overrides, and I kept this same convention for our CLI module.

src/azure-cli/azure/cli/command_modules/cdn/commands.py
lsmith130: Unfortunately, our service is a little confusing because available managed rule sets, available managed rule groups, and their available managed rules are separate from the managed rule sets with their rule group overrides and rule overrides that are applied to a particular policy. For this reason it's difficult to differentiate between whether an action is related to an available managed rule set/group vs an applied managed rule set/group. I've removed the `show-available` commands for the `cdn waf policy managed-rule-set` and `cdn waf policy managed-rule-set rule-group` command groups since it isn't essential for our service and is causing confusion.

Issue #11038, opened 2019-10-30
Add CLI Support for Private Link for Key Vault 


ShaneBala-keyvault:
**Resource Provider**
What is the Azure resource provider your feature is part of?
Azure Key Vault

**Description of Feature or Work Requested**
Provide a brief description of the feature or work requested. A link to conceptual documentation may be helpful too.
Add CLI Support for Private Link for Azure Key Vault 

**Minimum API Version Required**
What is the minimum API version of your service required to implement your feature?
See Swagger Doc Below

**Swagger Link**
Provide a link to the location of your feature(s) in the REST API specs repo. If your feature(s) has corresponding commit or pull request in the REST API specs repo, provide them. This should be on the master branch of the REST API specs repo.
https://github.com/Azure/azure-rest-api-specs/pull/7631

**Target Date**
If you have a target date for release of this feature/work, please provide it. While we can't guarantee these dates,
it will help us prioritize your request against other requests.
End of November 2019

yonzhan:
@jiasli  planned in S162.

bim-msft:
The creation and the deletion of Private Endpoint should be restrained in `network` command, which already supports command groups `private-endpoint` and `private-link-service`. @MyronFanQiu needs additional work for full support (see #11667)

For `keyvault` command, should we do the same thing as network command? Or just adding functionality like `show/get/list` for private endpoints is enough?

How do you think? @ShaneBala-keyvault @yungezz

bim-msft:
PR merged, will release on Feb 18.

PR #12073, opened 2020-02-06
[KeyVault] Support private endpoint connections

bim-msft:
Support for #11038 
Bump the Key Vault mgmt-plane SDK version to 2.1.1 (This is the fixed version, 2.1.0 is buggy)

Add two command groups, all of them are marked as `preview`:
1. `az keyvault private-endpoint`: manage vault private endpoint connections.
2. `az keyvault private-link-resource`: manage vault private link resources.

Whole scenario and test instructions:
1. Prepare the resources: a `Key Vault` and a `Virtual Network` within the same region (preferred region: `Central US EUAP`).
2. Disable private endpoint network policy for the new created vnet: https://docs.microsoft.com/en-us/azure/private-link/disable-private-endpoint-network-policy#using-azure-cli
3. Create a private endpoint connection request via network command: `network private-endpoint create -g {rg} -n {pe} --vnet-name {vnet} --subnet {subnet} -l {loc} --connection-name {pe_connection} --private-connection-resource-id {kv_id} --group-ids vault` (The `group-ids` can be obtained from command `az keyvault private-link-resource show`, for Key Vault service, the group id is `vault`, I remember that @MyronFanQiu planned to implement a new command in `network` group, which can be used to obtain the group id of all service types).
4. Show the private endpoint connection on Key Vault side: `keyvault private-endpoint show --connection-id {kv_pe_id}`.
5. Approve or reject this connection request: `keyvault private-endpoint approve -n {kv} --connection-name {kv_pe_name} --approval-description "{approval_desc}"`, `keyvault private-endpoint reject --connection-id {kv_pe_id} --rejection-description "{rejection_desc}"`

---


src/azure-cli/azure/cli/command_modules/keyvault/_validators.py
```diff
@@ -166,6 +166,27 @@ def validate_policy_permissions(ns):
...
 
+def validate_private_endpoint_connection_id(cmd, ns):
+    connection_id = ns.connection_id
+    connection_name = ns.connection_name
+    vault_name = ns.vault_name
+
+    if not connection_id:
+        if not all([connection_name, vault_name]):
+            raise argparse.ArgumentError(
+                None, 'specify both: --connection-name and --vault-name')
```
jiasli: In `az keyvault private-endpoint show -h`, it should be `--name` instead of `--vault-name`.
```suggestion
                None, 'specify both: --connection-name and --name')
```

src/azure-cli/azure/cli/command_modules/keyvault/_help.py
```diff
@@ -144,6 +144,31 @@
...
+"""
+
+helps['keyvault private-link-resource'] = """
+type: group
+short-summary: Manage vault private link resources.
+"""
+
+helps['keyvault private-link-resource show'] = """
+type: command
+short-summary: Show the private link resources supported for a Key Vault.
```
jiasli: https://github.com/Azure/azure-cli/blob/c0e345c7719cf20898ad4257171557b8e7e4d955/src/azure-cli/azure/cli/command_modules/network/_params.py#L745

The `az network private-endpoint create -h`'s help message doesn't conform to this command name. @MyronFanQiu

src/azure-cli/azure/cli/command_modules/keyvault/tests/latest/test_keyvault_commands.py
```diff
@@ -49,6 +49,85 @@ def test_parse_asn1_date(self):
...
+        # Prepare vault and network
+        keyvault = _create_keyvault(self, self.kwargs).get_output_in_json()
+        self.kwargs['kv_id'] = keyvault['id']
+        self.cmd('network vnet create -n {vnet} -g {rg} -l {loc} --subnet-name {subnet}',
+                 checks=self.check('length(newVNet.subnets)', 1))
+        self.cmd('network vnet subnet update -n {subnet} --vnet-name {vnet} -g {rg} '
+                 '--disable-private-endpoint-network-policies true',
+                 checks=self.check('privateEndpointNetworkPolicies', 'Disabled'))
+
+        # Create a private point connection
```
jiasli:
```suggestion
        # Create a private endpoint connection
```

src/azure-cli/azure/cli/command_modules/keyvault/_validators.py
bim-msft: @jiasli I prefer to replace all `--name` to `--vault-name` as this is consistent with other subgroup like `az keyvault key`, for the subgroup, `--name/-n` should represent the name of the sub resource itself (`key`, `secret`, `private-endpoint`), the name of vault should be represented by `--vault-name`.
How do you think?

jiasli: Makes sense. In that case, it should be --vault-name here: https://github.com/Azure/azure-cli/blob/cc53d1f95e3956444b885db0f3f48c2076cd3388/src/azure-cli/azure/cli/command_modules/keyvault/_params.py#L133

bim-msft: @jiasli Yea, I will change it.

src/azure-cli/azure/cli/command_modules/keyvault/tests/latest/test_keyvault_commands.py
```diff
@@ -49,6 +49,85 @@ def test_parse_asn1_date(self):
...
+                 '--rejection-description "{rejection_desc}"', checks=[
+                     self.check('privateLinkServiceConnectionState.status', 'Rejected'),
+                     self.check('privateLinkServiceConnectionState.description', '{rejection_desc}')
+                 ])
+
+        max_retries = 20
+        retries = 0
+        while self.cmd('keyvault private-endpoint show --connection-id {kv_pe_id}').\
+                get_output_in_json()['provisioningState'] != 'Succeeded' or retries > max_retries:
+            time.sleep(5)
```
jiasli: Better to sleep only during live run: https://github.com/Azure/azure-cli/blob/eb910a9cb389da8c1a5ebb37001a4068dbfd5bd6/src/azure-cli/azure/cli/command_modules/keyvault/tests/latest/test_keyvault_commands.py#L438-L439

src/azure-cli/azure/cli/command_modules/keyvault/tests/latest/test_keyvault_commands.py
```diff
@@ -49,6 +49,85 @@ def test_parse_asn1_date(self):
...
+            retries += 1
+
+        self.cmd('keyvault private-endpoint approve -n {kv} --connection-name {kv_pe_name} '
+                 '--approval-description "{approval_desc}"', checks=[
+                     self.check('privateLinkServiceConnectionState.status', 'Approved'),
+                     self.check('privateLinkServiceConnectionState.description', '{approval_desc}')
+                 ])
```
jiasli: Let's check `provisioningState` in the return value as well.

src/azure-cli/azure/cli/command_modules/keyvault/commands.py
```diff
@@ -64,6 +77,27 @@ def load_command_table(self, _):
...
 
+    with self.command_group('keyvault private-endpoint',
+                            kv_private_endpoint_connections_sdk,
+                            min_api='2018-02-14',
+                            client_factory=keyvault_client_private_endpoint_connections_factory,
+                            is_preview=True) as g:
+        g.custom_command('approve', 'approve_private_endpoint_connection',
+                         validator=validate_private_endpoint_connection_id)
+        g.custom_command('reject', 'reject_private_endpoint_connection',
+                         validator=validate_private_endpoint_connection_id)
```
jiasli: `approve` and `reject` are not common CLI commands. An alternative ways may be putting them in `keyvault private-endpoint update --approve true`. Has this been confirmed with PowerShell team @isra-fel or Key Vault PM?

src/azure-cli/azure/cli/command_modules/keyvault/tests/latest/test_keyvault_commands.py
bim-msft: Good idea

src/azure-cli/azure/cli/command_modules/keyvault/commands.py
bim-msft: @jiasli Yes, this is confirmed with Key Vault team, will forward you the email later.

src/azure-cli/azure/cli/command_modules/keyvault/tests/latest/test_keyvault_commands.py
bim-msft: Sure, will add this.

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
```diff
@@ -1154,3 +1154,71 @@ def restore_storage_account(client, vault_base_url, file_path):
...
         return client.restore_storage_account(vault_base_url, data)
 # endregion
+
+
+# region private_endpoint
+def show_private_endpoint_connection(client, resource_group_name, vault_name, connection_name,
+                                     connection_id=None):  # pylint: disable=unused-argument
+    """Show details of a private endpoint connection associated with a Key Vault."""
+    return client.get(resource_group_name=resource_group_name, vault_name=vault_name,
+                      private_endpoint_connection_name=connection_name)
```
jiasli: If this is a 1:1 mapping, maybe we can eliminate the custom function?

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
```diff
@@ -1154,3 +1154,71 @@ def restore_storage_account(client, vault_base_url, file_path):
...
+
+    return client.put(resource_group_name=resource_group_name,
+                      vault_name=vault_name,
+                      private_endpoint_connection_name=connection_name,
+                      private_endpoint=private_endpoint_connection.private_endpoint,
+                      properties=PrivateEndpointConnection(
+                          tags=private_endpoint_connection.tags,
+                          location=private_endpoint_connection.location,
+                          private_link_service_connection_state=PrivateLinkServiceConnectionState(
+                              status='Approved',
```
jiasli: Let's not hard-code it but use the value in `PrivateEndpointServiceConnectionStatus` instead.

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
bim-msft: @jiasli The only thing I have to do here is using name `connection_name` instead of `private_endpoint_connection_name`, because of this minor mismatch, I can not directly map it to SDK method, do you have any suggestion?

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
bim-msft: Good idea, will refine this.

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
```diff
@@ -1154,3 +1154,71 @@ def restore_storage_account(client, vault_base_url, file_path):
...
+    private_endpoint_connection = client.get(resource_group_name=resource_group_name, vault_name=vault_name,
+                                             private_endpoint_connection_name=connection_name)
+
+    return client.put(resource_group_name=resource_group_name,
+                      vault_name=vault_name,
+                      private_endpoint_connection_name=connection_name,
+                      private_endpoint=private_endpoint_connection.private_endpoint,
+                      properties=PrivateEndpointConnection(
+                          tags=private_endpoint_connection.tags,
+                          location=private_endpoint_connection.location,
```
jiasli: According to the code of `PrivateEndpointConnection`, it seems `tags` and `location` will be discarded. Also, can we just modify the returned value of `client.get`?

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
```diff
@@ -1154,3 +1154,71 @@ def restore_storage_account(client, vault_base_url, file_path):
...
+                          private_link_service_connection_state=PrivateLinkServiceConnectionState(
+                              status='Approved',
+                              description=approval_description
+                          )
+                      ))
+
+
+def reject_private_endpoint_connection(cmd, client, resource_group_name, vault_name, connection_name,
```
jiasli: We'd better to extract the common code and call it with a switch.

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
bim-msft: Yes, makes sense, will refine this.

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
bim-msft: Good idea.

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
```diff
@@ -1154,3 +1154,71 @@ def restore_storage_account(client, vault_base_url, file_path):
...
+    PrivateLinkServiceConnectionState = cmd.get_models('PrivateLinkServiceConnectionState',
+                                                       resource_type=ResourceType.MGMT_KEYVAULT)
+
+    private_endpoint_connection = client.get(resource_group_name=resource_group_name, vault_name=vault_name,
+                                             private_endpoint_connection_name=connection_name)
+
+    return client.put(resource_group_name=resource_group_name,
+                      vault_name=vault_name,
+                      private_endpoint_connection_name=connection_name,
+                      private_endpoint=private_endpoint_connection.private_endpoint,
```
jiasli: `private_endpoint` has been moved down to `properties` due to the parameter count change.  This is caused by autorest's `payload-flattening-threshold=2` config which will flatten the parameters when the parameter count <= 2. 

Anyway, let's remove it.

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
jiasli: Let's use `extra` to attach parameters. https://github.com/Azure/azure-cli/blob/cc53d1f95e3956444b885db0f3f48c2076cd3388/src/azure-cli/azure/cli/command_modules/keyvault/_params.py#L148

src/azure-cli/azure/cli/command_modules/keyvault/_help.py
```diff
@@ -144,6 +144,31 @@
 short-summary: Manage vault network ACLs.
 """
 
+helps['keyvault private-endpoint'] = """
+type: group
+short-summary: Manage vault private endpoint connections.
+"""
```
jiasli: On second thought, we'd better name it as `private-endpoint-connection` as it corresponds to  `privateEndpointConnections` from `az keyvault show`.

```jsonc
    "privateEndpointConnections": [
      {
        "etag": "9fae19d5c0704012a6b0544574fe24d2",
        "id": "/subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/resourceGroups/rg1/providers/Microsoft.KeyVault/vaults/kv0130/privateEndpointConnections/c9e56b4f661f4cf382aeb5d832bf6f9f",
        "privateEndpoint": {
          "id": "/subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/resourceGroups/rg1/providers/Microsoft.Network/privateEndpoints/pe1",
          "resourceGroup": "rg1"
        },
        "privateLinkServiceConnectionState": {
          "actionRequired": null,
          "actionsRequired": "None",
          "description": "",
          "status": "Approved"
        },
        "provisioningState": "Succeeded",
        "resourceGroup": "rg1"
      }
```

`privateEndpoint` stands for another resource type `Microsoft.Network/privateEndpoints`. I think we shouldn't mix these two concepts.

bim-msft: @jiasli Thanks for this, for the convenience, I shortened the group name as `private-endpoint` (in fact, I copied the group name from `network private-endpoint`), but as you mentioned, this name stands for another resource type and may cause confusion, I will modify this later.

src/azure-cli/azure/cli/command_modules/keyvault/_params.py
```diff
@@ -108,15 +108,37 @@ class CLIJsonWebKeyOperation(str, Enum):
...
         c.argument('ip_address', help='IPv4 address or CIDR range.')
         c.argument('subnet', help='Name or ID of subnet. If name is supplied, `--vnet-name` must be supplied.')
         c.argument('vnet_name', help='Name of a virtual network.', validator=validate_subnet)
+
+    with self.argument_context('keyvault private-endpoint', min_api='2018-02-14') as c:
+        c.argument('approval_description', help='Comments for the approval.')
+        c.argument('private_endpoint_connection_name', options_list=['--connection-name', '-n'], required=False,
+                   help='The name of the private endpoint connection associated with the Key Vault. '
+                        'Required if --connection-id is not specified')
+        c.argument('vault_name', vault_name_type, required=False,
```
MyronFanQiu: should it be `service_name` or keep different in different sub command groups.

src/azure-cli/azure/cli/command_modules/keyvault/_validators.py
```diff
@@ -166,6 +166,27 @@ def validate_policy_permissions(ns):
...
 
 
+def validate_private_endpoint_connection_id(cmd, ns):
+    connection_id = ns.connection_id
+    connection_name = ns.private_endpoint_connection_name
+    vault_name = ns.vault_name
+
+    if not connection_id:
+        if not all([connection_name, vault_name]):
+            raise argparse.ArgumentError(
```
MyronFanQiu: Based on my knowledge, usually we would `raise CLIError('usage error:')`. We can discuss.

src/azure-cli/azure/cli/command_modules/keyvault/commands.py
MyronFanQiu: here. I think this design is good.

src/azure-cli/azure/cli/command_modules/keyvault/commands.py
```diff
@@ -64,6 +77,25 @@ def load_command_table(self, _):
...
+        g.command('delete', 'delete', validator=validate_private_endpoint_connection_id)
+        g.show_command('show', 'get', validator=validate_private_endpoint_connection_id)
+
+    with self.command_group('keyvault private-link-resource',
+                            kv_private_link_resources_sdk,
+                            min_api='2018-02-14',
+                            client_factory=keyvault_client_private_link_resources_factory,
+                            is_preview=True) as g:
+        g.show_command('show', 'list_by_vault')
```
MyronFanQiu: should it support id?

src/azure-cli/azure/cli/command_modules/keyvault/tests/latest/test_keyvault_commands.py
```diff
@@ -49,6 +49,104 @@ def test_parse_asn1_date(self):
...
+            'rejection_desc': 'You are rejected!'
+        })
+        self.cmd('keyvault private-endpoint reject --connection-id {kv_pe_id} '
+                 '--rejection-description "{rejection_desc}"', checks=[
+                     self.check('privateLinkServiceConnectionState.status', 'Rejected'),
+                     self.check('privateLinkServiceConnectionState.description', '{rejection_desc}'),
+                     self.check('provisioningState', 'Updating')
+                 ])
+
+        max_retries = 20
```
MyronFanQiu: I think this is because of the errors of the swagger spec. Could we add some comments here?

Issue #6343, opened 2018-05-13
Authorization issue when calling az policy assignment create with policy id.


cmendible:
**Describe the bug**
If I run: 

`az policy assignment create --name 'Allowed locations' --policy "/providers/Microsoft.Authorization/policyDefinitions/e56962a6-4747-49cd-b67b-bf8b01975c4c" -p '{\"listOfAllowedLocations\":{\"value\":[\"westeurope\",\"northeurope\",\"eastus\"]}}' -g 'Resource Group Name' `

I get the following error:

`The client 'live.com#[my account] with object id '[GUID]' does not have authorization to perform action 'Microsoft.Authorization/policyDefinitions/Microsoft.Authorization/e56962a6-4747-49cd-b67b-bf8b01975c4c/read' over scope '/providers/Microsoft.Authorization/policyDefinitions/providers/Microsoft.Authorization/policyDefinitions'.`

**To Reproduce**
Run:
`az policy assignment create --name 'Allowed locations' --policy "/providers/Microsoft.Authorization/policyDefinitions/e56962a6-4747-49cd-b67b-bf8b01975c4c" -p '{\"listOfAllowedLocations\":{\"value\":[\"westeurope\",\"northeurope\",\"eastus\"]}}' -g 'Resource Group Name' `

**Expected behavior**
The cli help (**az policy assignment create -h**) states:  **--policy: Name or id of the policy definition**. Therefore I expect the same result as when running the command with just the name of the policy:

`az policy assignment create --name 'Allowed locations' --policy "e56962a6-4747-49cd-b67b-bf8b01975c4c" -p '{\"listOfAllowedLocations\":{\"value\":[\"westeurope\",\"northeurope\",\"eastus\"]}}' -g 'Resource Group Name' `

which results in:

`{                                                                                                                                                                                                       
  "description": null,                                                                                                                                                                                  
  "displayName": null,                                                                                                                                                                                  
  "id": "/subscriptions/[subscription id]/resourceGroups/[resource group]/providers/Microsoft.Authorization/policyAssignments/Allowed locations",                                              
  "metadata": null,                                                                                                                                                                                     
  "name": "Allowed locations",                                                                                                                                                                          
  "notScopes": null,                                                                                                                                                                                    
  "parameters": {                                                                                                                                                                                       
    "listOfAllowedLocations": {                                                                                                                                                                         
      "value": [                                                                                                                                                                                        
        "westeurope",                                                                                                                                                                                   
[...]

PR #12075, opened 2020-02-06
[ARM] az policy assignment create: Fix the error message when the `--policy` parameter is invalid (#6343)

zhoxing-ms:
Issue: #6343

If an exception occurs when judging whether `--policy` parameter is a built-in policy, the Authorization error message will be thrown. This error message will mislead users,so it's necessary to change the error information to help the user troubleshoot problems.

**History Notes:**  

---


yonzhan:
add to S165.

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -646,7 +646,10 @@ def _get_custom_or_builtin_policy(cmd, client, name, subscription=None, manageme
     except (CloudError, HttpOperationError) as ex:
         status_code = ex.status_code if isinstance(ex, CloudError) else ex.response.status_code
         if status_code == 404:
-            return policy_operations.get_built_in(name)
+            try:
+                return policy_operations.get_built_in(name)
+            except CloudError:
```
qianwens: >CloudError [](start = 19, length = 10)

Can CloudError caused by other reason like internalServerError, if so, you may need to raise IncorrectUsageError for a more specific error code

zhoxing-ms: Thanks for your suggestion, I have improved the processing logic of this part.

PR #12080, opened 2020-02-06
{AppService} az webapp config ssl import: Fix to allow key vault in other subscription

madsd:
**History Notes:**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -2039,10 +2039,11 @@ def import_ssl_cert(cmd, resource_group_name, name, key_vault, key_vault_certifi
     kv_id_parts = parse_resource_id(kv_id)
     kv_name = kv_id_parts['name']
     kv_resource_group_name = kv_id_parts['resource_group']
+    kv_subscription = kv_id_parts['subscription']
     cert_name = '{}-{}-{}'.format(resource_group_name, kv_name, key_vault_certificate_name)
     lnk = 'https://azure.github.io/AppService/2016/05/24/Deploying-Azure-Web-App-Certificate-through-Key-Vault.html'
     lnk_msg = 'Find more details here: {}'.format(lnk)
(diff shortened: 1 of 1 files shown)


arrownj:
Hi @qwordy @Juliehzl , could you please help review this PR ?

yonzhan:
add to S165.

qwordy:
What's the disadvantage of allowing key vault in other subscription? Was it supported before?

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2054,10 +2055,15 @@ def import_ssl_cert(cmd, resource_group_name, name, key_vault, key_vault_certifi
...
-def _check_service_principal_permissions(cmd, resource_group_name, key_vault_name):
+def _check_service_principal_permissions(cmd, resource_group_name, key_vault_name, key_vault_subscription):
     from azure.cli.command_modules.keyvault._client_factory import keyvault_client_vaults_factory
     from azure.cli.command_modules.role._client_factory import _graph_client_factory
     from azure.graphrbac.models import GraphErrorException
+    from azure.cli.core.commands.client_factory import get_subscription_id
+    subscription = get_subscription_id(cmd.cli_ctx)
+    # Cannot check if key vault is in another subscription
+    if subscription != key_vault_subscription:
+        return False
```
qwordy: Your code looks good

madsd:
> What's the disadvantage of allowing key vault in other subscription? Was it supported before?

It is allowed and is still supported - all this change does is remove the access check, as the key vault sdk does not allow cross subscription lookup.

madsd:
@zhoxing-ms , @qianwens , @Juliehzl can you have a look at this. I just pushed with a fresh rebase. It is a minor change to allow use of Key Vaults from other subscriptions. The functionality was working, but the check for permissions did not, so essentially I just removed the check, if the subscriptions are different (and the call will then just fail in the backend as expected)

madsd:
@yonzhan is there a reason that this is just pushed?

Issue #10714, opened 2019-10-01
az rest response code


jurgyy:
**Is your feature request related to a problem? Please describe.**
Currently when making `put` or `patch` requests using `az rest` it can be unclear whether the request is executed successfully. Unsuccessful execution returns an error, but there are certain commands that have multiple successful responses such as [api create/update](https://docs.microsoft.com/en-us/rest/api/apimanagement/2019-01-01/apis/createorupdate#response): codes 200, 201 and 202. In this case getting no returned value is confusing as to what has happened.

**Describe the solution you'd like**
It would be nice if `az rest` was able to return those codes to ease the flow of scripts.

**Describe alternatives you've considered**
In the current scenario I have to send a get request to check if everything is set as expected while this information is (sometimes) already present in the response header so perhaps return the header entirely.

maggiepint:
Hi,
I've routed this to the appropriate team for follow-up.

mattgwic:
It should at least show up when I use --debug

jurgyy:
Any update on this?

I've just spend the past hour debugging a problem with trying to create an API and the source of the problem was that in the Swagger specification the property `properties.value.info.title` was identical to an already existing API. Sending the request manually I got a 202 Accepted response with a location header which when I send a GET request I received the following message:

```json
{
	"error": {
		"code": "ValidationError",
		"message": "One or more fields contain incorrect values:",
		"details": [{
				"code": "ValidationError",
				"target": "name",
				"message": "API with specified name 'myApiName' already exists"
			}
		]
	}
}
```

Would've been nice if `az rest` returns something along those lines...

zhoxing-ms:
This issue is related to asynchronous operation, asynchronous execution will be adopted for some time-consuming operations.
Usage documents of asynchronous API: [link](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/async-operations)

zhoxing-ms:
@jurgyy What do you expect to return? Is the location URL returned when you manually request to create the API or the ValidationError returned when you send a GET request for the second time?

jurgyy:
@zhoxing-ms Due to the asynchronous nature of the problem I'd suppose the location URL would suffice and let the end user decide what to do with it. But not returning anything is undesired behavior in my opinion.

zhoxing-ms:
OK, I will submit a PR to fix it. 
After that, you can get information about the header by using `--debug`, such as the location URL.

jiasli:
[...]

PR #12096, opened 2020-02-07
[ARM] az rest: Print the header of response (#10714)

zhoxing-ms:
Issue: #10714

Some services executed asynchronously will put useful information into the header, like: location URL, Retry-After, etc. So I print the header of response for `az rest`.

**History Notes:**  

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/util.py
+++ b/src/azure-cli-core/azure/cli/core/util.py
@@ -608,6 +608,7 @@ def send_raw_request(cli_ctx, method, uri, headers=None, uri_parameters=None,  #
     try:
         r = requests.request(method, uri, params=uri_parameters, data=body, headers=headers,
                              verify=not should_disable_connection_verify())
+        logger.debug("Response Header : %s", r.headers if r else None)
     except Exception as ex:  # pylint: disable=broad-except
         raise CLIError(ex)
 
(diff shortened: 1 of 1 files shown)


yonzhan:
add to S165.

fengzhou-msft:
Please change PR title to begin with `{ARM}` since the change do not need to be put in history notes.

zhoxing-ms:
OK, but I think this PR can be written into history notes because it is associated with a customer's issue

zhoxing-ms:
> Please change PR title to begin with {ARM} since the change do not need to be put in history notes.

OK, but I think this PR can be written into history notes because it is associated with a customer's issue

fengzhou-msft:
> > Please change PR title to begin with {ARM} since the change do not need to be put in history notes.
> 
> @fengzhou-msft OK, but I think this PR title can be written into history notes because it is associated with a customer's issue

Sure. It's up to you.

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -608,6 +608,7 @@ def send_raw_request(cli_ctx, method, uri, headers=None, uri_parameters=None,  #
     try:
         r = requests.request(method, uri, params=uri_parameters, data=body, headers=headers,
                              verify=not should_disable_connection_verify())
+        logger.debug("Response Header : %s", r.headers if r else '')
```
jiasli: We can dump the headers regardless of `r.__bool__()` so that headers for non-successful response such as 404, 403 are also printed. I will create a new PR to address this.

PR #12100, opened 2020-02-07
[ACS] (BREAKING CHANGE:) (az aks:) support msi changes for GF and BF for omsagent (Container monitoring)(#1)

rashmichandrashekar:
msi role assignment changes for greenfield and brownfield onboarding for omsagent


src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -1802,10 +1813,28 @@ def aks_create(cmd, client, resource_group_name, name, ssh_key_value,  # pylint:
...
+                # Else, provide permissions to MSI
+                if hasattr(result, 'service_principal_profile') and hasattr(result.service_principal_profile, 'client_id'):
+                    logger.warning('service principal exists, using it')
+                    service_principal_msi_id = result.service_principal_profile.client_id
+                    isServicePrincipal = True
+                elif ((hasattr(result, 'addon_profiles')) and
+                     ('omsagent' in result.addon_profiles) and
+                     (hasattr(result.addon_profiles['omsagent'], 'identity')) and
+                     (hasattr(result.addon_profiles['omsagent'].identity, 'object_id'))):
+                    logger.warning('omsagent MSI exists, using it')
```
ganga1980: similarly here, info?

src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -1802,10 +1813,28 @@ def aks_create(cmd, client, resource_group_name, name, ssh_key_value,  # pylint:
...
+                    service_principal_msi_id = result.addon_profiles['omsagent'].identity.object_id
+                    isServicePrincipal = False
+
+                if service_principal_msi_id is not None:
```
ganga1980: throwing warning if service_principal_msi_id is None will be useful to troubleshoot and catch issues.

src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -1854,13 +1889,28 @@ def aks_enable_addons(cmd, client, resource_group_name, name, addons, workspace_
                 namespace='Microsoft.ContainerService', type='managedClusters',
                 name=name
             )
+        service_principal_msi_id = None
+        # Check if service principal exists, if it does, assign permissions to service principal
+        # Else, provide permissions to MSI
+        if hasattr(result, 'service_principal_profile') and hasattr(result.service_principal_profile, 'client_id'):
```
ganga1980: I think, wrapping this logic into function  would be cleaner.

src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -1833,17 +1862,23 @@ def aks_disable_addons(cmd, client, resource_group_name, name, addons, no_wait=F
...
     instance = _update_addons(cmd, instance, subscription_id, resource_group_name, addons, enable=True,
                               workspace_resource_id=workspace_resource_id, subnet_name=subnet_name, no_wait=no_wait)
 
     if 'omsagent' in instance.addon_profiles:
         _ensure_container_insights_for_monitoring(cmd, instance.addon_profiles['omsagent'])
+        
+    # send the managed cluster representation to update the addon profiles
+    result = sdk_no_wait(no_wait, client.create_or_update,
+                          resource_group_name, name, instance)
+    result = LongRunningOperation(cmd.cli_ctx)(result)
```
ganga1980: similarly here, I think, we should execute LongRunningOperation function only if the monitoring enabled. This way we dont have impact on other scenarios.

yonzhan:
add to S165.

src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -1802,10 +1815,33 @@ def aks_create(cmd, client, resource_group_name, name, ssh_key_value,  # pylint:
...
+                ):
+                    logger.warning('omsagent MSI exists, using it')
+                    service_principal_msi_id = result.addon_profiles['omsagent'].identity.object_id
+                    isServicePrincipal = False
+
+                if service_principal_msi_id is not None:
+                    if not _add_role_assignment(cmd.cli_ctx, 'Monitoring Metrics Publisher',
+                                                service_principal_msi_id, isServicePrincipal,
+                                                scope=cluster_resource_id):
+                        logger.warning('Could not create a role assignment for monitoring addon. '
```
MyronFanQiu: I think try/except is better than `if not`

src/azure-cli/azure/cli/command_modules/acs/tests/latest/test_custom.py
```diff
@@ -56,7 +56,7 @@ def test_add_role_assignment_basic(self):
         with mock.patch(
                 'azure.cli.command_modules.acs.custom.create_role_assignment') as create_role_assignment:
             ok = _add_role_assignment(cli_ctx, role, sp, delay=0)
-            create_role_assignment.assert_called_with(cli_ctx, role, sp, scope=None)
+            create_role_assignment.assert_called_with(cli_ctx, role, sp, True, scope=None)
```
MyronFanQiu: is it able to add test for `False` scenario?

MyronFanQiu:
@jiasli Could you take a look at this PR? I'm not familiar with service pricinpal and lack the background.

src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -1791,6 +1798,10 @@ def aks_create(cmd, client, resource_group_name, name, ssh_key_value,  # pylint:
                                  client.create_or_update,
                                  resource_group_name=resource_group_name,
                                  resource_name=name, parameters=mc)
+
+            # adding a wait here since we rely on the result for role assignment
```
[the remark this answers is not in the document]
rashmichandrashekar: @MyronFanQiu - I tested out that if --no-wait is set to False, the command doesnt wait for return value and leads to false evaluation of the following conditions. We need this in order for it to use the right values from the result.

src/azure-cli/azure/cli/command_modules/acs/custom.py
rashmichandrashekar: @MyronFanQiu - I have retained the same behavior as before and with other places where the role assignments are done within this file. Is it a new guideline to use try/catch instead of if not?

src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -551,7 +551,7 @@ def _build_service_principal(rbac_client, cli_ctx, name, url, client_secret):
     return service_principal
 
 
-def _add_role_assignment(cli_ctx, role, service_principal, delay=2, scope=None):
+def _add_role_assignment(cli_ctx, role, service_principal_msi_id, isServicePrincipal=True, delay=2, scope=None):
```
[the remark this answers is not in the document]
rashmichandrashekar: @MyronFanQiu - As mentioned in the previous reply to the comment, the command doesnt seem to wait if LongRunningOperation is not used, irrespective of the --no-wait flag. I will move it to do this only if monitoring is enabled.

src/azure-cli/azure/cli/command_modules/acs/custom.py
rashmichandrashekar: The method _add_role_assignment returns a boolean value. I think adding try except would changes the behavior of this method and all the depending methods. I will retain the same behavior as before. Unless, there is an absolute need to change it

src/azure-cli/azure/cli/command_modules/acs/tests/latest/test_custom.py
rashmichandrashekar: Sure i can do that.

src/azure-cli/azure/cli/command_modules/acs/custom.py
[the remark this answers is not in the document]
MyronFanQiu: currently, if user set --no-wait, it just won't work. This should be a breaking change. If you want to do this, you could just remove the sdk_no_wait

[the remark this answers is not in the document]
rashmichandrashekar: addressed this as discussed offline

src/azure-cli/azure/cli/command_modules/acs/tests/latest/test_custom.py
MyronFanQiu: Hello. Could we have a test case for `False` scenario?

src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -1395,11 +1395,12 @@ def create_service_principal(cli_ctx, identifier, resolve_app=True, rbac_client=
...
 
 
-def create_role_assignment(cli_ctx, role, assignee, resource_group_name=None, scope=None):
-    return _create_role_assignment(cli_ctx, role, assignee, resource_group_name, scope)
+def create_role_assignment(cli_ctx, role, assignee, is_service_principal, resource_group_name=None, scope=None):
+    return _create_role_assignment(cli_ctx, role, assignee, is_service_principal, resource_group_name, scope)
 
 
-def _create_role_assignment(cli_ctx, role, assignee, resource_group_name=None, scope=None, resolve_assignee=True):
+def _create_role_assignment(cli_ctx, role, assignee, is_service_principal,
```
jiasli: Maybe `is_service_principal` is a little bit duplicated? To me, it does very similar thing as `resolve_assignee`. `_resolve_object_id` can be skipped by setting `resolve_assignee` to `False`.

jiasli:
I ran the command twice and got error:

```
> aks enable-addons -a monitoring -g {} -n {}

The monitoring addon is already enabled for this managed cluster.
To change monitoring configuration, run "az aks disable-addons -a monitoring"before enabling it again.
```

I'd suggest to make the command idempotent.

src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -1793,21 +1833,23 @@ def aks_create(cmd, client, resource_group_name, name, ssh_key_value,  # pylint:
                                  client.create_or_update,
                                  resource_group_name=resource_group_name,
                                  resource_name=name, parameters=mc)
+
             # add cluster spn with Monitoring Metrics Publisher role assignment to the cluster resource
             # mdm metrics supported only in azure public cloud so add the  role assignment only in this cloud
             cloud_name = cmd.cli_ctx.cloud.name
             if cloud_name.lower() == 'azurecloud' and monitoring:
+                # adding a wait here since we rely on the result for role assignment
+                result = LongRunningOperation(cmd.cli_ctx)(result)
```
jiasli: Per my understanding, waiting is necessary here. 

However, if `no_wait` is set to `True`, I don't think `LongRunningOperation` works anymore. In such case, `result` is a [`LROPoller`](https://github.com/Azure/msrest-for-python/blob/master/msrest/polling/poller.py#L103) whose `polling_method` is [`NoPolling`](https://github.com/Azure/msrest-for-python/blob/master/msrest/polling/poller.py#L66) whose [`finished()`](https://github.com/Azure/msrest-for-python/blob/master/msrest/polling/poller.py#L91) method always returns `True`. 

If you want to force `LongRunningOperation` to wait here, you need to remove `sdk_no_wait`. For example, during VM creation: 

https://github.com/Azure/azure-cli/blob/d3beab81c9245a3e815a08fe3034a89ba1b52f88/src/azure-cli/azure/cli/command_modules/vm/custom.py#L826-L830

src/azure-cli/azure/cli/command_modules/acs/custom.py
rashmichandrashekar: Yes. That makes sense. I will get rid of this. I can just use the resolve_assignee.

rashmichandrashekar:
> I ran the command twice and got error:
> 
> ```
> > aks enable-addons -a monitoring -g {} -n {}
> 
> The monitoring addon is already enabled for this managed cluster.
> To change monitoring configuration, run "az aks disable-addons -a monitoring"before enabling it again.
> ```
> 
> I'd suggest to make the command idempotent.

rashmichandrashekar:
> > I ran the command twice and got error:
> > ```
> > > aks enable-addons -a monitoring -g {} -n {}
> > 
> > The monitoring addon is already enabled for this managed cluster.
> > To change monitoring configuration, run "az aks disable-addons -a monitoring"before enabling it again.
> > ```
> > 
> > 
> > I'd suggest to make the command idempotent.

Another PUT request on an existing addon might not trigger the addon recreation. If the new PUT request has a different workspace id this might not go into effect(which will only happen when the pods are restarted) This is not in scope of the PR. I would prefer handling this separately since this requires further testing. I would leave the existing behavior as is.

src/azure-cli/azure/cli/command_modules/acs/custom.py
rashmichandrashekar: As discussed offline, made chnages as per our conversation

PR #12112, opened 2020-02-10
[Storage] BREAKING CHANGE: az storage account create: Change default storage account kind to StorageV2

Juliehzl:
**History Notes:**  

---
**Description**:

The feature is required for a long time. Given there is more and more feature only available in StorageV2 Kind, we need to switch the defualt kind to StoargeV2, which will be align with Powershell, two.

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/storage/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/storage/_params.py
@@ -154,8 +154,10 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals, too-many-statem
         c.argument('location', get_location_type(self.cli_ctx), validator=get_default_location_from_resource_group)
         c.argument('account_type', help='The storage account type', arg_type=get_enum_type(t_account_type))
         c.argument('account_name', acct_name_type, options_list=['--name', '-n'], completer=None)
-        c.argument('kind', help='Indicates the type of storage account.',
-                   arg_type=get_enum_type(t_kind, default='storage'))
+        c.argument('kind', help='Indicates the type of storage account.', min_api="2017-10-01",
+                   arg_type=get_enum_type(t_kind), default='StorageV2')
+        c.argument('kind', help='Indicates the type of storage account.', max_api="2017-06-01",
+                   arg_type=get_enum_type(t_kind), default='Storage')
         c.argument('https_only', arg_type=get_three_state_flag(), min_api='2019-04-01',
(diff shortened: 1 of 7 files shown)


src/azure-cli/azure/cli/command_modules/storage/_params.py
```diff
@@ -154,8 +154,10 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals, too-many-statem
         c.argument('location', get_location_type(self.cli_ctx), validator=get_default_location_from_resource_group)
         c.argument('account_type', help='The storage account type', arg_type=get_enum_type(t_account_type))
         c.argument('account_name', acct_name_type, options_list=['--name', '-n'], completer=None)
-        c.argument('kind', help='Indicates the type of storage account.',
-                   arg_type=get_enum_type(t_kind, default='storage'))
+        c.argument('kind', help='Indicates the type of storage account.', min_api="2017-10-01",
```
MyronFanQiu: Can you share the reason about changing the default to V2?

Juliehzl: Most features are only available for storageV2 storage account. In this way, StorageV2 is a general purpose storage account.
In addition, there are requirements from other team to have such default value and it is the same with PSH behavior.

src/azure-cli/azure/cli/command_modules/storage/_params.py
```diff
@@ -154,8 +154,10 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals, too-many-statem
...
+                   arg_type=get_enum_type(t_kind), default='StorageV2')
+        c.argument('kind', help='Indicates the type of storage account.', max_api="2017-06-01",
+                   arg_type=get_enum_type(t_kind), default='Storage')
```
qianwens: >Storage [](start = 60, length = 7)

the default value was "storage" while it is "Storage" now, is it by design?

Juliehzl: In cli commands, if you use get_enum_types(), the input value will be case insensitive. In this way, it is doesn't matter for "storage: or "Storage" here.

qianwens: so this default value is only used by our code, it is not visible to user?

---
In reply to: [376912505](https://github.com/Azure/azure-cli/pull/12112#discussion_r376912505) [](ancestors = 376912505)

yonzhan:
add to S165.

PR #12115, opened 2020-02-10
[Compute] BREAKING CHANGE: Fix #10728: `az vm create`: create subnet automatically if vnet is specified and subnet not exists

### LINKED ISSUE #10728: az vm create - option "--subnet" having incorrect behavior
## Describe the bug

**Command Name**
`az vm create`

**Errors:**
Unable to delete the subnet 'Default' because the subnet is in used

## To Reproduce:
- Multiple VM instances are created with the same vnet "MyRG-vnet" and subnet "Default" within it
- Only a single vnet and a single subnet exist in the resource group
- `az vm create -n MyRG --image Win2019Datacenter --size Standard_D2s_v3 --nsg sql-server --vnet-name MyRG-vnet`

## Expected Behavior
The VM should be created since the option "--subnet" is not required to be provided. Mentions in the documents, `--subnet The name of the subnet when creating a new VNet or referencing an existing one. Can also reference an existing subnet by ID. If omitted, an appropriate VNet and subnet will be selected automatically, or a new one will be created.` Which means, if the option "--subnet" is not provided, it will select the one and only subnet "Default" in my vnet. It did not select the subnet, instead seems like it tried to delete the subnet "Default" and attempted to create a new one (probably with the same name)

After I manually added the option "--subnet Default", it works, but the behavior contradicts with the description in the command reference.

## Environment Summary
```
Linux-4.15.0-1059-azure-x86_64-with-debian-stretch-sid
Python 3.6.5
Shell: bash

azure-cli 2.0.74

Extensions:
aks-preview 0.4.8
```

arrownj:
This PR is to fix #10728 

Previously, when user use below command:
`
az vm create -n MyRG --image Win2019Datacenter --size Standard_D2s_v3 --nsg sql-server --vnet-name MyRG-vnet
`
As subnet is not specified here, Azure CLI will generate a default value for it (the format is {VMName}Subnet, in this case, subnet value will be MyRGSubnet).

Azure CLI will try to create/update the vnet with this subnet. So if the vnet exists and already has subnets, the behavior here will be that remove all existing subnets first and then add this new one. This is wrong.

After a discussion, we decide to update the logic to add this new subnet to the list instead of clearing the list. To reduce the impacts to the least, the code change only impact such case: vnet is specified with a existing value and subnet is not specified. All other cases will not be impacted and follow original logic.

---

**History Notes:**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/vm/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_params.py
@@ -677,7 +677,7 @@ def load_arguments(self, _):
         with self.argument_context(scope, arg_group='Network') as c:
             c.argument('vnet_name', help='Name of the virtual network when creating a new one or referencing an existing one.')
             c.argument('vnet_address_prefix', help='The IP address prefix to use when creating a new VNet in CIDR format.')
-            c.argument('subnet', help='The name of the subnet when creating a new VNet or referencing an existing one. Can also reference an existing subnet by ID. If omitted, an appropriate VNet and subnet will be selected automatically only if VNet is also omitted, or a new one will be created if not exists.')
+            c.argument('subnet', help='The name of the subnet when creating a new VNet or referencing an existing one. Can also reference an existing subnet by ID. If both vnet-name and subnet are omitted, an appropriate VNet and subnet will be selected automatically, or a new one will be created.')
             c.argument('subnet_address_prefix', help='The subnet IP address prefix to use when creating a new VNet in CIDR format.')
             c.argument('nics', nargs='+', help='Names or IDs of existing NICs to attach to the VM. The first NIC will be designated as primary. If omitted, a new NIC will be created. If an existing NIC is specified, do not specify subnet, VNet, public IP or NSG.')
             c.argument('private_ip_address', help='Static private IP address (e.g. 10.0.0.5).')
--- a/src/azure-cli/azure/cli/command_modules/vm/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_params.py
@@ -677,7 +677,7 @@ def load_arguments(self, _):
         with self.argument_context(scope, arg_group='Network') as c:
             c.argument('vnet_name', help='Name of the virtual network when creating a new one or referencing an existing one.')
             c.argument('vnet_address_prefix', help='The IP address prefix to use when creating a new VNet in CIDR format.')
-            c.argument('subnet', help='The name of the subnet when creating a new VNet or referencing an existing one. Can also reference an existing subnet by ID. If omitted, an appropriate VNet and subnet will be selected automatically, or a new one will be created.')
+            c.argument('subnet', help='The name of the subnet when creating a new VNet or referencing an existing one. Can also reference an existing subnet by ID. If omitted, an appropriate VNet and subnet will be selected automatically only if VNet is also omitted, or a new one will be created if not exists.')
             c.argument('subnet_address_prefix', help='The subnet IP address prefix to use when creating a new VNet in CIDR format.')
             c.argument('nics', nargs='+', help='Names or IDs of existing NICs to attach to the VM. The first NIC will be designated as primary. If omitted, a new NIC will be created. If an existing NIC is specified, do not specify subnet, VNet, public IP or NSG.')
             c.argument('private_ip_address', help='Static private IP address (e.g. 10.0.0.5).')
--- a/src/azure-cli/azure/cli/command_modules/vm/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/custom.py
@@ -679,10 +679,13 @@ def create_vm(cmd, vm_name, resource_group_name, image=None, size='Standard_DS1_
                         address_prefix=subnet_address_prefix
                     )
                     upsert_to_collection(vnet, 'subnets', subnet_obj, 'name')
-                    vnet = cached_put(cmd, client.create_or_update, vnet, resource_group_name, vnet_name).result()
-                    subnet_created = next((x for x in vnet.subnets if x.name.lower() == subnet.lower()), None)
-                    if not subnet_created:
-                        raise CLIError('Create subnet failed')
+                    try:
+                        cached_put(cmd, client.create_or_update, vnet, resource_group_name, vnet_name)
+                    except Exception:
+                        raise CLIError('Subnet({}) does not exist, fail to create new subnet with address prefix {}. '
+                                       'It may be caused by name or address prefix conflict. Please specify an '
+                                       'appropriate subnet name with --subnet or an valid address prefix value with '
+                                       '--subnet-address-prefix.'.format(subnet, subnet_address_prefix))
             if not vnet_exists:
                 vnet_name = vnet_name or '{}VNET'.format(vm_name)
                 nic_dependencies.append('Microsoft.Network/virtualNetworks/{}'.format(vnet_name))
--- a/src/azure-cli/azure/cli/command_modules/vm/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/custom.py
@@ -680,7 +680,7 @@ def create_vm(cmd, vm_name, resource_group_name, image=None, size='Standard_DS1_
                     )
                     upsert_to_collection(vnet, 'subnets', subnet_obj, 'name')
                     try:
-                        cached_put(cmd, client.create_or_update, vnet, resource_group_name, vnet_name)
+                        cached_put(cmd, client.create_or_update, vnet, resource_group_name, vnet_name).result()
                     except Exception:
                         raise CLIError('Subnet({}) does not exist, but failed to create a new subnet with address '
                                        'prefix {}. It may be caused by name or address prefix conflict. Please specify '
--- a/src/azure-cli/azure/cli/command_modules/vm/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/custom.py
@@ -660,11 +660,34 @@ def create_vm(cmd, vm_name, resource_group_name, image=None, size='Standard_DS1_
 
         nic_dependencies = []
         if vnet_type == 'new':
-            vnet_name = vnet_name or '{}VNET'.format(vm_name)
             subnet = subnet or '{}Subnet'.format(vm_name)
-            nic_dependencies.append('Microsoft.Network/virtualNetworks/{}'.format(vnet_name))
-            master_template.add_resource(build_vnet_resource(
-                cmd, vnet_name, location, tags, vnet_address_prefix, subnet, subnet_address_prefix))
+            vnet_exists = False
+            if vnet_name:
+                from azure.cli.command_modules.vm._vm_utils import check_existence
+                vnet_exists = \
+                    check_existence(cmd.cli_ctx, vnet_name, resource_group_name, 'Microsoft.Network', 'virtualNetworks')
+                if vnet_exists:
+                    from azure.cli.core.commands import cached_get, cached_put, upsert_to_collection
+                    from azure.cli.command_modules.vm._validators import get_network_client
+                    client = get_network_client(cmd.cli_ctx).virtual_networks
+                    vnet = cached_get(cmd, client.get, resource_group_name, vnet_name)
+
+                    Subnet = cmd.get_models('Subnet', resource_type=ResourceType.MGMT_NETWORK)
+                    subnet_obj = Subnet(
+                        name=subnet,
+                        address_prefixes=[subnet_address_prefix],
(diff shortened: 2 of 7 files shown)


yonzhan:
add to S165.

src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -677,7 +677,7 @@ def load_arguments(self, _):
         with self.argument_context(scope, arg_group='Network') as c:
             c.argument('vnet_name', help='Name of the virtual network when creating a new one or referencing an existing one.')
             c.argument('vnet_address_prefix', help='The IP address prefix to use when creating a new VNet in CIDR format.')
-            c.argument('subnet', help='The name of the subnet when creating a new VNet or referencing an existing one. Can also reference an existing subnet by ID. If omitted, an appropriate VNet and subnet will be selected automatically, or a new one will be created.')
+            c.argument('subnet', help='The name of the subnet when creating a new VNet or referencing an existing one. Can also reference an existing subnet by ID. If omitted, an appropriate VNet and subnet will be selected automatically only if VNet is also omitted, or a new one will be created if not exists.')
```
fengzhou-msft: This doesn't seem very clear to me: If omitted, an appropriate VNet and subnet will be selected automatically only if VNet is also omitted, or a new one will be created if not exists.

My understanding is if subnet is ommited, we will use a default name. Then the process is the same as subnet not ommited?

jiasli: The word "appropriate" is ambiguous in my mind.

arrownj: The logic of selecting automatically only happens when both vnet and subnet are omitted. And the auto create subnet logic happens when the subnet not exists, regardless where the value come from (user specified or generate from default format).

arrownj: Hi @jiasli , could you please help give some suggestions to make this message clearer ?

fengzhou-msft: Then `A new one will be created if not exists.` should be in a separate sentence regardless of whether subnet is omitted.

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -660,11 +660,34 @@ def create_vm(cmd, vm_name, resource_group_name, image=None, size='Standard_DS1_
...
+                    from azure.cli.command_modules.vm._validators import get_network_client
+                    client = get_network_client(cmd.cli_ctx).virtual_networks
+                    vnet = cached_get(cmd, client.get, resource_group_name, vnet_name)
+
+                    Subnet = cmd.get_models('Subnet', resource_type=ResourceType.MGMT_NETWORK)
+                    subnet_obj = Subnet(
+                        name=subnet,
+                        address_prefixes=[subnet_address_prefix],
+                        address_prefix=subnet_address_prefix
+                    )
```
jiasli: Adding a subnet is not as simple as adding a new resource. We need to take the **vnet's address space** and **subnet's prefix** into consideration. 

It is very hard to deduce the **new subnet's prefix** based on the **existing vnet's address space** and **existing subnet's prefix**. Even Azure Portal doesn't do the decision for the user. When creating a subnet, it just leave the field `Address range (CIDR block)` empty.

Some example edge cases:

1. If the existing vnet has address space `10.3.0.0/16`, this command will try to create a subnet with default prefix `10.0.0.0/24` and will fail, because the prefix is not in the allowed address space:
    > Azure Error: NetcfgInvalidSubnet
    > Message: Subnet 'vm3Subnet' is not valid in virtual network 'vnet3'.

2. If the existing vnet's address space is the same as the existing subnet's prefix - both `10.0.0.0/16`, the subnet creation will fail in any case, because there is no space left

3. If the existing vnet's address space is `10.0.0.0/24`, how do we decide the mask of the subnet? 24, 25? It is not practical to do the decision for the user

I prefer to error out if vnet is specified while subnet is not. 

cc @MyronFanQiu

src/azure-cli/azure/cli/command_modules/vm/_params.py
jiasli: I don't know the behavior of this param. That's why I raised this question. 😆 
We need to state the actual behavior based on the code logic.

src/azure-cli/azure/cli/command_modules/vm/custom.py
arrownj: Yes, we discussed the cases you mentioned above before. Previously, there are some problems with the vnet/subnet prefix. However, for this PR, we don't want to change these behaviors, just keep them as before. 

What we changed in this PR this that, when vnet is specified and exists, no matter subnet is specified or not, if the subnet not exists, create it without removing other existing subnets. 

As the cases above, if the user not specify subnet prefix, it will use a default value (10.0.0.0/24) which may be wrong, however this is not the problems we want to solve in the PR.

yungezz: to share some context, not related with this PR. There's default vnet/subnet azure CLI created when creating VM. This is convenient scenario CLI providing: minimum effort to get started on a VM.

src/azure-cli/azure/cli/command_modules/vm/tests/latest/test_vm_commands.py
```diff
@@ -4058,5 +4058,38 @@ def test_vmss_create_disk_iops_mbps(self, resource_group):
...
+        })
+
+        # Expecting no results
+        self.cmd('vm list --resource-group {rg}',
+                 checks=self.is_empty())
+        self.cmd('network vnet list --resource-group {rg}',
+                 checks=self.is_empty())
+
+        self.cmd('network vnet create --resource-group {rg} --name {vnet} --location {loc}')
+        self.cmd('vm create --resource-group {rg} --location {loc} --name {vm} --admin-username ubuntu --image UbuntuLTS --admin-password testPassword0 --authentication-type password --vnet-name {vnet}')
```
jiasli: The test didn't fail because the vnet doesn't have subnets. Assume we create a vnet with this command:

```
az network vnet create -g {} -n vnet5 --location eastasia --address-prefix 10.0.0.0/16 --subnet-name sub1 --subnet-prefix 10.0.0.0/24 
```

Then the `vm create` will fail:

```
vm create --resource-group {} --location southeastasia --name {} --admin-username ubuntu --image UbuntuLTS --admin-password testPassword0 --authentication-type password --vnet-name vnet5
Azure Error: NetcfgInvalidSubnet
Message: Subnet '{}Subnet' is not valid in virtual network 'vnet5'.
```
Because `10.0.0.0/24` is already occupied by `sub1`.

src/azure-cli/azure/cli/command_modules/vm/_params.py
fengzhou-msft: Actually a `new one will be created` is already mentioned in the first sentence. So don't bother to change.

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -679,10 +679,13 @@ def create_vm(cmd, vm_name, resource_group_name, image=None, size='Standard_DS1_
...
                     )
                     upsert_to_collection(vnet, 'subnets', subnet_obj, 'name')
-                    vnet = cached_put(cmd, client.create_or_update, vnet, resource_group_name, vnet_name).result()
-                    subnet_created = next((x for x in vnet.subnets if x.name.lower() == subnet.lower()), None)
-                    if not subnet_created:
-                        raise CLIError('Create subnet failed')
+                    try:
+                        cached_put(cmd, client.create_or_update, vnet, resource_group_name, vnet_name)
+                    except Exception:
+                        raise CLIError('Subnet({}) does not exist, fail to create new subnet with address prefix {}. '
```
jiasli:
```suggestion
                        raise CLIError('Subnet({}) does not exist, but failed to create new subnet with address prefix {}. '
```

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -679,10 +679,13 @@ def create_vm(cmd, vm_name, resource_group_name, image=None, size='Standard_DS1_
...
+                                       'It may be caused by name or address prefix conflict. Please specify an '
+                                       'appropriate subnet name with --subnet or an valid address prefix value with '
```
jiasli:
```suggestion
                                       'appropriate subnet name with --subnet or a valid address prefix value with '
```

src/azure-cli/azure/cli/command_modules/vm/custom.py
arrownj: Thanks, updated.

src/azure-cli/azure/cli/command_modules/vm/custom.py
arrownj: Thanks, updated.

qwordy:
So the problem is how can we create a valid subnet in a vnet. If there is no subnet in this vnet, I think we can create one for user. But if there are some existing subnets in this vnet, then it becomes harder.

src/azure-cli/azure/cli/command_modules/vm/custom.py
> 
> Adding a subnet is not as simple as adding a new resource. We need to take the **vnet's address space** and **subnet's prefix** into consideration.
> 
> It is very hard to deduce the **new subnet's prefix** based on the **existing vnet's address space** and **existing subnet's prefix**.
> 
> This PR will always try to add subnet `10.0.0.0/24` to the vnet. It'll succeed only if `10.0.0.0/24` is valid within the vnet, which is very unlikely in most cases. In other cases:
> 
>     1. If the existing vnet `10.0.0.0/16` already has subnet `10.0.0.0/24`, the creation will fail because the space is already occupied
> 
>     2. If the existing vnet is `10.3.0.0/16`, the creation will fail because the prefix is not in the allowed address space:
>        > Azure Error: NetcfgInvalidSubnet
>        > Message: Subnet 'vm3Subnet' is not valid in virtual network 'vnet3'.
> 
>     3. If the existing vnet's address space is the same as the existing subnet's prefix - both `10.0.0.0/16`, the subnet creation will fail in any case, because there is no space left
> 
>     4. If the existing vnet's address space is `10.0.0.0/24`, how do we decide the mask of the subnet? 24, 25? It is not practical to do the decision for the user
> 
> 
> **My suggestion**: If `--vnet-name` and `--subnet` are neither specified, `az create` will detect if there is already a vnet and subnet in the resource group and put the vm there. We can apply the same when only `--vnet-name` is specified with an existing vnet - if there is a subnet, put the vm there; if not, error out, instead of calculating the available prefix and creating a subnet voluntarily.
> 
> cc @MyronFanQiu

1, retry 25, 26..., or 23, 22...
2, we can create subnet with same address of vnet
3. retry 25, 26..., or 23, 22...
4, Azure CLI always makes decision for users so that users can use a very simple command to create resources.

I agree it's complex and tricky. Users might be astonished when a strange subnet is created.

qwordy:
How do we know the user wants to use an existing subnet in the vnet or create a new subnet?
Users may think "I have specified an existing vnet. I want to reuse these network resources"

arrownj:
> How do we know the user wants to use an existing subnet in the vnet or create a new subnet?
> Users may think "I have specified an existing vnet. I want to reuse these network resources"

try to find the specified one first, if not exists, then will try to create a new one. In this case, user should also specify the subnet value, we update the help message to let user this.

qwordy:
The commit history has problem. So many unrelated commit. And involve so many reviewers. You need to git reset and make the commit history correct. I guess the reason is that you checkout a new branch in your new machine and then you pull from this PR.

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -660,11 +660,37 @@ def create_vm(cmd, vm_name, resource_group_name, image=None, size='Standard_DS1_
...
+                if vnet_exists:
+                    from azure.cli.core.commands import cached_get, cached_put, upsert_to_collection
+                    from azure.cli.command_modules.vm._validators import get_network_client
+                    client = get_network_client(cmd.cli_ctx).virtual_networks
+                    vnet = cached_get(cmd, client.get, resource_group_name, vnet_name)
+
+                    Subnet = cmd.get_models('Subnet', resource_type=ResourceType.MGMT_NETWORK)
+                    subnet_obj = Subnet(
+                        name=subnet,
+                        address_prefixes=[subnet_address_prefix],
```
qwordy: default value: subnet_address_prefix='10.0.0.0/24'
It may be incompatible with vnet. You can deduce a valid value from vnet.
I request a change here.

Issue #11908, opened 2020-01-20
az rest should show request/response details


jongio:
When using `az rest` I would like to know the exact request and response that is sent and received by the command.  For example, please refer to the httpie --print option: https://httpie.org/doc#what-parts-of-the-http-exchange-should-be-printed

This is very helpful when debugging issues.

yonzhan:
add to S165.

jiasli:
The HTTP traces appeared in `stderr` with `--debug` enabled for specific commands are populated by `msrest.http_logger`. `az rest` directly uses [requests](https://github.com/psf/requests) library, so traces are not included in `stderr` with `--debug` enabled:

https://github.com/Azure/azure-cli/blob/26c8ead7536cb3fe56490fd86db6b0a83b3c7066/src/azure-cli-core/azure/cli/core/util.py#L609-L610

Currently there is no good way to log requests for **requests** library: https://stackoverflow.com/questions/16337511/log-all-requests-from-the-python-requests-module

Given the limitation, we prefer not to implement this feature by now. If you would like to capture HTTP trace for debugging purpose, we recommend following [Working behind a proxy](https://github.com/Azure/azure-cli/blob/dev/doc/use_cli_effectively.md#working-behind-a-proxy) and using Fiddler.

jongio:
@jiasli Can you please have a look at this: 

https://stackoverflow.com/questions/20658572/python-requests-print-entire-http-request-raw 

https://requests.readthedocs.io/en/master/user/advanced/#prepared-requests


It looks like it is possible to get the headers, body, via .prepare() method.

jiasli:
Thanks for the link. However, dumping the HTTP trace is not as simple as it seems, since we also need to remove the authorization header. It also involves `merge_environment_settings`-related logic that we need to play around and fully test.

Actually it is a pretty nice idea to have HTTP trace with `az rest`. I have marked this PR as **feature request**. But currently we have many other priorities to work on and since there is workaround with Fiddler, I may not have time to develop this feature. It will also be great if you can draft a PR. We will review and merge it.

jiasli:
As this is not blocking issue, we will close it and plan it in future releases.

jiasli:
It appears we are over-complicate things. Instead of letting **requests** dump headers by itself, why don't we dump for it? I have submitted PR #12117 to show the headers. @jongio, kindly let me know if the change suits your need.

yonzhan:
@jongio any feedback?

[...]

PR #12117, opened 2020-02-10
[REST] az rest: Dump request and response

### LINKED ISSUE #10714: az rest response code

jiasli:
Fix #10714, #11908
Inspired by #12096

With `--verbose` flag, `az rest` can now dump the raw request and response to `stderr` stream, using the same format as msrest. 

For example:

```sh
az rest --uri /subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/resourcegroups/myrg?api-version=2019-07-01 --verbose
```

`stderr`:

```
Request URL: 'https://management.azure.com/subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/resourcegroups/myrg?api-version=2019-07-01'
Request method: 'GET'
Request headers:
    'User-Agent': 'AZURECLI/2.2.0'
    'Accept-Encoding': 'gzip, deflate'
    'Accept': '*/*'
    'Connection': 'keep-alive'
    'x-ms-client-request-id': '16b7f3c5-153f-43af-b30f-0bc7a12569df'
    'CommandName': 'rest'
    'ParameterSetName': '--uri --verbose'
    'Authorization': 'Bearer eyJ0eXAi*****'
Request body:
None
Response status: 200
Response headers:
    'Cache-Control': 'no-cache'
    'Pragma': 'no-cache'
    'Content-Type': 'application/json; charset=utf-8'
    'Content-Encoding': 'gzip'
    'Expires': '-1'
    'Vary': 'Accept-Encoding'
    'x-ms-ratelimit-remaining-subscription-reads': '11992'
    'x-ms-request-id': '5d8d43d8-eab3-4d5b-bc88-71fc547b7747'
    'x-ms-correlation-request-id': '5d8d43d8-eab3-4d5b-bc88-71fc547b7747'
    'x-ms-routing-request-id': 'SOUTHEASTASIA:20200319T113507Z:5d8d43d8-eab3-4d5b-bc88-71fc547b7747'
    'Strict-Transport-Security': 'max-age=31536000; includeSubDomains'
    'X-Content-Type-Options': 'nosniff'
    'Date': 'Thu, 19 Mar 2020 11:35:06 GMT'
    'Content-Length': '308'
Response content:
{"id":"/subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/resourceGroups/myrg","name":"myrg","type":"Microsoft.Resources/resourceGroups","location":"westus","tags":{"StorageType":"Standard_LRS","type":"test"},"properties":{"provisioningState":"Succeeded"}}
```

`stdout`:

```jsonc
{
  "id": "/subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/resourceGroups/myrg",
  "location": "westus",
  "name": "myrg",
  "properties": {
    "provisioningState": "Succeeded"
  },
  "type": "Microsoft.Resources/resourceGroups"
}
```

Authentication header is trimmed to only keep the first 15 characters. 

```
'Authorization': 'Bearer eyJ0eXAi*****'
```


yonzhan:
add to S165.

jiasli:
Rather than JSON, another way is to dump RAW HTTP response headers like:

```
>curl https://microsoft.com -I
HTTP/1.1 301 Moved Permanently
Date: Tue, 11 Feb 2020 06:23:06 GMT
Server: Kestrel
Location: https://www.microsoft.com/
```

@yonzhan @jongio @achandmsft, any suggestions?

jongio:
If you output JSON the output has to be valid JSON as in no text in between the JSON snippets.
Also consider if we are outputting the authorization header.  I think it is helpful, but consider redacting some of it so you don't print the whole token.

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -584,6 +584,10 @@ def send_raw_request(cli_ctx, method, uri, headers=None, uri_parameters=None,  #
     if '{subscriptionId}' in uri:
         uri = uri.replace('{subscriptionId}', profile.get_subscription_id())
 
+    # Dump custom request headers before adding Authorization header
+    logger.info("Request headers:")
```
jongio: please have output be valid JSON, so I can easily programmatically parse the entire output

jiasli: I agree. The problem is there are two parts - request and response headers. We want to display request headers before sending the request, so they have to be torn apart. 

Also, `stderr` may contain other trash like `command ran in 4.608 seconds.`, so it is not a good practice to extract info from `stderr`. 

For the `Authorization` header, the current behavior of msrest with `--debug` is to totally remove it. We previously also removed similar information with https://github.com/Azure/azure-cli/pull/10473 due to security concerns. So currently let's just hide it until we get approval from PM @achandmsft .

What I am thinking is to add a param `--include-headers`, so that result in `stdout` will become something like:

```jsonc
{
    "request": {
        "headers": {
            "User-Agent": "AZURECLI/2.0.81",
            "x-ms-client-request-id": "b43346df-c51f-475a-bf7d-eec64d394f15",
            "CommandName": "rest",
            "ParameterSetName": "-u --verbose"
        }
    },
    "response": {
        "headers": {
            "Cache-Control": "no-cache",
            "Pragma": "no-cache",
            "Content-Type": "application/json; charset=utf-8",
            "Content-Encoding": "gzip",
            "Expires": "-1",
            "Vary": "Accept-Encoding",
            "x-ms-ratelimit-remaining-subscription-reads": "11999",
            "x-ms-request-id": "05761bc6-1b26-44be-a676-a6999e950a8a",
            "x-ms-correlation-request-id": "05761bc6-1b26-44be-a676-a6999e950a8a",
            "x-ms-routing-request-id": "EASTASIA:20200210T113239Z:05761bc6-1b26-44be-a676-a6999e950a8a",
            "Strict-Transport-Security": "max-age=31536000; includeSubDomains",
            "X-Content-Type-Options": "nosniff",
            "Date": "Mon, 10 Feb 2020 11:32:38 GMT",
            "Content-Length": "279"
        },
        "status": 200,
        "body": {
            "id": "/subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/resourceGroups/xxx",
            "location": "southeastasia",
            "name": "xxx",
            "properties": {
                "provisioningState": "Succeeded"
            },
            "type": "Microsoft.Resources/resourceGroups"
        }
    }
}
```

In thus case, the output can be easily scripted from `stdout` instead of `stderr`. Any suggestion is welcome.

jongio: I like stdout and full json, just make sure it is escaped properly, i.e. if body contains quotes

--include-headers should be something like the print flag in the httpie library that I referenced in the bug.  Sometimes you only want certain headers.

For auth, I would still include the header, so the user knows the header was there, but redact the actual token.

`Authorization: Bearer ey=oo97er...0789er`

Not the full token.

jiasli: Let's address the full json output (including headers) in another PR. I have updated this one to follow the format from msrest (see the updated description).

For `Authentication` header, I have included it in the log and trimmed it to only keep the first 15 characters. Does this suit the needs?

```
'Authorization': 'Bearer eyJ0eXAi*****'
```

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -623,9 +624,16 @@ def send_raw_request(cli_ctx, method, uri, headers=None, uri_parameters=None,  #
             logger.warning("Can't derive appropriate Azure AD resource from --url to acquire an access token. "
                            "If access token is required, use --resource to specify the resource")
     try:
-        r = requests.request(method, uri, params=uri_parameters, data=body, headers=headers,
-                             verify=not should_disable_connection_verify())
-        logger.debug("Response Header : %s", r.headers if r else '')
```
zhoxing-ms: This is a slight breaking change. Do we need to add `BREAKING CHANGE` to the title to let users notice the change?

jiasli: Not really. Things in `stderr` is never meant to be stable.

zhoxing-ms: OK

zhoxing-ms: If it is released, please  just remember to inform the customer of this issue.  #12096

jiasli: Sure.

src/azure-cli-core/azure/cli/core/util.py
julienstroheker: @jiasli any update on been able to print `--verbose` logs as JSON ? That will be useful to parse it :)

nascarsayan: @jiasli any update on the `--include-headers` proposal?

Issue #4911, opened 2017-11-18
Assigning a role to a service principal by name yields a HTTP 400


carlosonunez:
---

### Environment summary

**Install Method**: Docker
**CLI version**: 2.0.21
**Shell Type**: Bash

### Bug Summary

Running `az role assignment create` where `--assignee` is a service principal name yields a HTTP 400. (Using the service principal ID works.)

### Actual behavior

```shell
bash-4.3# az ad sp list | jq '.[] | select(.displayName == "Packer") | { id: .objectId, name: .displayName }'
{
  "id": "6924e191-262b-49a2-91ac-78e63c5cfa3f",
  "name": "Packer"
}
bash-4.3# az role assignment create --assignee Packer --role Owner                                                 Operation failed with status: 'Bad Request'. Details: 400 Client Error: Bad Request for url: https://graph.windows.net/83072677-0859-4f53-b309-0ec405db441d/getObjectsByObjectIds?api-version=1.6
```

### Expected behavior

```shell
bash-4.3# az ad sp list | jq '.[] | select(.displayName == "Packer") | { id: .objectId, name: .displayName }'
{
  "id": "6924e191-262b-49a2-91ac-78e63c5cfa3f",
  "name": "Packer"
}
bash-4.3# az role assignment create --assignee "6924e191-262b-49a2-91ac-78e63c5cfa3f" --role Owner
{
  "id": "/subscriptions/SUB_ID/providers/Microsoft.Authorization/roleAssignments/7c2b7644-8002-4f9b-9c19-d37f6dc36c4f",
  "name": "7c2b7644-8002-4f9b-9c19-d37f6dc36c4f",
  "properties": {
    "principalId": "6924e191-262b-49a2-91ac-78e63c5cfa3f",
    "roleDefinitionId": "/subscriptions/8bbbc92b-6d16-4eb2-8f95-7a0769748c8d/providers/Microsoft.Authorization/roleDefinitions/8e3af657-a8ff-443c-a75c-2fe8c4bcb635",
    "scope": "/subscriptions/SUB_ID"
  },
  "type": "Microsoft.Authorization/roleAssignments"
}
```

yugangw-msft:
@carlosonunez  I agree there is a bit confusion, but please do not use displayName, instead using `app_id` or any one from the array from `servicePrincipalNames` should work

carlosonunez:
Will do, but can we get a more friendly error message instead of a 400?

JamesDawson:
[...]

PR #12133, opened 2020-02-11
[Role] az role assignment create: Fix the error that assigning a role to a service principal by display name yields a HTTP 400

qianwens:
Fix issue [#4911](https://github.com/Azure/azure-cli/issues/4911)

Root cause: CLI code tries to use assignee as objectId to get objects. When the assignee is a display name, service throws bad request because the display name is not a GUID.
How to fix: Try to use the assignee as objectId to get objects when it is a GUID.

**History Notes:**  

---


yonzhan:
add to S165.

Issue #11767, opened 2020-01-03
az vm image list can't find publisher


SamaraSoucy-MSFT:
Describe the bug

**Command Name**
`az vm image list`

**Errors:**
```
Publisher: Symantec.test.ru4mp1.latest was not found.
```

## To Reproduce:
Steps to reproduce the behavior. Note that argument values have been redacted, as they may contain sensitive information.

- _Put any pre-requisite steps here..._
- `az vm image list --offer {} --all --location {} --output {} --debug`
(may only reproduce with the exact 'az vm image list --offer UbuntuServer --all --location eastus2' command

## Environment Summary
```
Windows-10-10.0.18362-SP0
Python 3.6.6
Shell: powershell.exe

azure-cli 2.0.74

Extensions:
azure-devops 0.11.0
dev-spaces 1.0.3
webapp 0.2.19

```
## Additional Context

Original customer thread https://github.com/MicrosoftDocs/azure-docs-cli/issues/1732
I was able to reproduce it and I've sent an email with the debug log to the DL.

bim-msft:
@qwordy Please take a look.

qwordy:
I have reproduced the bug. Using '--debug', you can see this command can get lots of images, but in the end of the output, there is an error when retrieving image info from publisher Symantec.test.ru4mp1.latest. I think I may need to add some fault tolerance mechanism.

farialima:
Having a similar issue :

```
bash-3.2$ az vm image list --offer Debian --all --location francecentral
Publisher: Microsoft.Azure.Extensions.Testb2e612e0-1c8c-40ed-93fe-a169ed7619cb-20190219154622 was not found.
bash-3.2$ 
```

This is problematic because it prevent discovering images. The only way to workaround it is to find (somehow) a publisher, and re-rerun with it, eg. 
```
az vm image list --offer Debian --all --location francecentral --publisher credativ
```

Running with `debug` shows the following potential causes :

```
(...)
urllib3.connectionpool : https://management.azure.com:443 "GET /subscriptions/c0c11638-e31a-4ea3-815e-d33c8ef7e55b/providers/Microsoft.Compute/locations/francecentral/publishers/Microsoft.Az\
ure.Extensions.Testb2e612e0-1c8c-40ed-93fe-a169ed7619cb-20190219154622/artifacttypes/vmimage/offers?api-version=2019-07-01 HTTP/1.1" 404 175
```
(...)
```
[...]

PR #12134, opened 2020-02-11
[Compute] Increase robustness of vm image list

qwordy:
Issue #11767 

In this command's implementation,
Step 1. use `client.virtual_machine_images.list_publishers(location)` to get publishers
Step 2. use `client.virtual_machine_images.list_offers(location, publisher)` to get offers

Step 2 fails for some publishers retrieved from step 1, then the whole command fails. It's a data inconsistency from service.
Anyway, add try-catch. Use `logger.warning` if exception occurs.

**History Notes:**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/vm/_actions.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_actions.py
@@ -40,15 +40,27 @@ def load_images_thru_services(cli_ctx, publisher, offer, sku, location):
         location = get_one_of_subscription_locations(cli_ctx)
 
     def _load_images_from_publisher(publisher):
-        offers = client.virtual_machine_images.list_offers(location, publisher)
+        try:
+            offers = client.virtual_machine_images.list_offers(location, publisher)
+        except Exception as e:
+            logger.warning(str(e))
+            return
         if offer:
             offers = [o for o in offers if _matched(offer, o.name)]
         for o in offers:
-            skus = client.virtual_machine_images.list_skus(location, publisher, o.name)
+            try:
+                skus = client.virtual_machine_images.list_skus(location, publisher, o.name)
+            except Exception as e:
+                logger.warning(str(e))
+                continue
             if sku:
                 skus = [s for s in skus if _matched(sku, s.name)]
             for s in skus:
-                images = client.virtual_machine_images.list(location, publisher, o.name, s.name)
+                try:
+                    images = client.virtual_machine_images.list(location, publisher, o.name, s.name)
+                except Exception as e:
+                    logger.warning(str(e))
+                    continue
                 for i in images:
                     all_images.append({
                         'publisher': publisher,
--- a/src/azure-cli/azure/cli/command_modules/vm/_actions.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_actions.py
@@ -40,9 +40,10 @@ def load_images_thru_services(cli_ctx, publisher, offer, sku, location):
         location = get_one_of_subscription_locations(cli_ctx)
 
     def _load_images_from_publisher(publisher):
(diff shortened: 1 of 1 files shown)


yonzhan:
add to S165.

src/azure-cli/azure/cli/command_modules/vm/_actions.py
```diff
@@ -40,15 +40,28 @@ def load_images_thru_services(cli_ctx, publisher, offer, sku, location):
...
+                logger.warning(str(e))
+                continue
             if sku:
                 skus = [s for s in skus if _matched(sku, s.name)]
             for s in skus:
-                images = client.virtual_machine_images.list(location, publisher, o.name, s.name)
+                try:
+                    images = client.virtual_machine_images.list(location, publisher, o.name, s.name)
+                except CloudError as e:
+                    logger.warning(str(e))
```
MyronFanQiu: e.message? I'm not familiar with it. What's the difference between `e.message` and `str(e)`? @jiasli

arrownj: As I know, str(object) will call the object.__str__() at last. and e.message is just a property of e. Whether these two values are the same depends on how to implement the function __str__(). If it return message directly, they will be the same. However, for most cases, __str__ will serialize all the properties of the object. @MyronFanQiu

src/azure-cli/azure/cli/command_modules/vm/_actions.py
```diff
@@ -40,15 +40,28 @@ def load_images_thru_services(cli_ctx, publisher, offer, sku, location):
         location = get_one_of_subscription_locations(cli_ctx)
 
     def _load_images_from_publisher(publisher):
-        offers = client.virtual_machine_images.list_offers(location, publisher)
+        from msrestazure.azure_exceptions import CloudError
+        try:
+            offers = client.virtual_machine_images.list_offers(location, publisher)
+        except CloudError as e:
+            logger.warning(str(e))
```
arrownj: Previously, if there is an exception, it will be treated as an error. After updating, it will be treated as a warning, right ?

src/azure-cli/azure/cli/command_modules/vm/_actions.py
```diff
@@ -40,15 +40,28 @@ def load_images_thru_services(cli_ctx, publisher, offer, sku, location):
...
+            return
         if offer:
             offers = [o for o in offers if _matched(offer, o.name)]
         for o in offers:
-            skus = client.virtual_machine_images.list_skus(location, publisher, o.name)
+            try:
+                skus = client.virtual_machine_images.list_skus(location, publisher, o.name)
+            except CloudError as e:
+                logger.warning(str(e))
```
arrownj: if meet exception for some `offer`, the skus of it will be skipped ?

src/azure-cli/azure/cli/command_modules/vm/_actions.py
qwordy: str(e) can tell more details including which publisher cannot be found. It's ok to use e.message here. The difference depends on specific implementation.
code:
msrestazure\azure_exceptions.py, class CloudError(ClientException):

src/azure-cli/azure/cli/command_modules/vm/_actions.py
qwordy: Yes. There are many publishers. This function will be invoked many times. An error will fail the command. But there are other valid publishers. So I catch it.

src/azure-cli/azure/cli/command_modules/vm/_actions.py
qwordy: Yes.

PR #12139, opened 2020-02-11
[AKS] Add cluster-autoscaler profile

marwanad:
**History Notes:**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acs/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/_help.py
@@ -270,6 +270,9 @@ parameters:
   - name: --max-count
     type: int
     short-summary: Maximum nodes count used for autoscaler, when "--enable-cluster-autoscaler" specified. Please specify the value in the range of [1, 100].
+  - name: --cluster-autoscaler-profile
+    type: list
+    short-summary: Space-separated list of key=value pairs for configuring cluster autoscaler.
   - name: --vm-set-type
     type: string
     short-summary: Agent pool vm set type. VirtualMachineScaleSets or AvailabilitySet.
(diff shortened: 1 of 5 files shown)


yungezz:
For review only, service not ready for merge.

src/azure-cli/azure/cli/command_modules/acs/_validators.py
```diff
@@ -249,3 +250,28 @@ def validate_vnet_subnet_id(namespace):
...
+    """
+    _extract_cluster_autoscaler_params(namespace)
+    if namespace.cluster_autoscaler_profile is not None:
+        for key in namespace.cluster_autoscaler_profile.keys():
+            _validate_cluster_autoscaler_key(key)
+
+
+def _validate_cluster_autoscaler_key(key):
+    if not key:
+        raise CLIError('Empty key specified for cluster-autoscaler parameter')
```
fengzhou-msft: paramter name is `cluster-autoscaler-profile`, please change it in the error message to keep consistent.

src/azure-cli/azure/cli/command_modules/acs/_help.py
```diff
@@ -370,6 +373,12 @@
   - name: --max-count
     type: int
     short-summary: Maximum nodes count used for autoscaler, when "--enable-cluster-autoscaler" specified. Please specify the value in the range of [1, 100]
+  - name: --clear-cluster-autoscaler-profile
+    type: bool
+    short-summary: Clears the cluster-autoscaler profile.
+  - name: --cluster-autoscaler-profile
+    type: list
+    short-summary: Space-separated list of key=value pairs for configuring cluster autoscaler.
```
qwordy: Maybe you use empty string to represent clear operation. Then we can delete --clear-cluster-autoscaler-profile and only reserve --cluster-autoscaler-profile. For example, `--cluster-autoscaler-profile ""`

marwanad: If we go with this, with that example a user does:
```
az aks update {{some stuff}}
az aks update --cluster-autoscaler-profile ""
```
I could be missing something here but how do we differentiate between both intentions in the validator? To the validator in both cases, `namespace.cluster_autoscaler_profile` is falsy but the intention in the first command is to not clear the profile while the intention of the second one is to clear it.

marwanad:
made some changes, pending the new API vendor - CI won't pass.

src/azure-cli/azure/cli/command_modules/acs/_help.py
qwordy: This is the only request of mine. After this is revolved, I'll approve.

marwanad:
will reopen after rebase.

PR #12151, opened 2020-02-12
{KeyVault} Modify command group name `private-endpoint` to `private-endpoint-connection`

bim-msft:
Create this PR for eliminating the ambiguous meaning mentioned in https://github.com/Azure/azure-cli/pull/12073#discussion_r378153569

The previous group name `private-endpoint` stands for resource `Microsoft.Network/privateEndpoints`, however, what we need is `privateEndpointConnections`.

---


src/azure-cli/azure/cli/command_modules/keyvault/_params.py
```diff
@@ -122,20 +122,20 @@ class CLIJsonWebKeyOperation(str, Enum):
...
         c.argument('rejection_description', help='Comments for the rejection.')
 
     for item in ['approve', 'reject', 'delete', 'show']:
-        with self.argument_context('keyvault private-endpoint {}'.format(item), min_api='2018-02-14') as c:
-            c.extra('connection_id', required=False,
+        with self.argument_context('keyvault private-endpoint-connection {}'.format(item), min_api='2018-02-14') as c:
+            c.extra('connection_id', options_list=['--id'], required=False,
                     help='The ID of the private endpoint connection associated with the Key Vault. '
-                         'If specified --vault-name and --connection-name/-n, this should be omitted.')
+                         'If specified --vault-name and --name/-n, this should be omitted.')
```
fengzhou-msft: I think `connection-id` and `connection-name` could be consolidated into one parameter `connection`. It can support both a name or id. You already has a validator to parse id.

Parameter names need to be changed in the error message in validator as well. https://github.com/Azure/azure-cli/blob/54ff67cb6e8fa9c2bb02e1e5bd32b99c368f0ee0/src/azure-cli/azure/cli/command_modules/keyvault/_validators.py#L182

bim-msft: @fengzhou-msft Thanks for pointing it out, for the consistency concern, I prefer to maintain the status quo as we already have commands like `az keyvault key show`, they use both `--id` and `--name/-n`, the change for these commands would be a breaking change, we can revisit this later:
```
(env3) C:\repos\azure-cli>az keyvault key show -h

Command
    az keyvault key show : Gets the public part of a stored key.
        The get key operation is applicable to all key types. If the requested key is symmetric,
        then no key material is released in the response. This operation requires the keys/get
        permission.

Arguments

Id Arguments
    --id           : Id of the key.  If specified all other 'Id' arguments should be omitted.
    --name -n      : Name of the key. Required if --id is not specified.
    --vault-name   : Name of the key vault. Required if --id is not specified.
    --version -v   : The key version. If omitted, uses the latest version.

```

However, you pointed out a typo I ignored, many thanks!
`--connection-id` -> `--id`
`--connection-name` -> `--name`
https://github.com/Azure/azure-cli/blob/54ff67cb6e8fa9c2bb02e1e5bd32b99c368f0ee0/src/azure-cli/azure/cli/command_modules/keyvault/_validators.py#L177
https://github.com/Azure/azure-cli/blob/54ff67cb6e8fa9c2bb02e1e5bd32b99c368f0ee0/src/azure-cli/azure/cli/command_modules/keyvault/_validators.py#L182

PR #12156, opened 2020-02-12
[ACR] `az acr login`: Throw a CLIError if there are errors returned by docker command

### LINKED ISSUE #11355: AZ Cli incorrectly returns exit code 0 even on failure
## Describe the bug

**Command Name**
`az acr login`

**Errors:**
The command `az acr login` silently fails on a network error and the process returns exit code 0.  This makes breaks Unix convention and makes it more difficult in automation to detect failure.

## To Reproduce:
We are seeing the error intermittently based on network failures.  The network failures are not the issue here, the exit code of 0 is.  Whenever we do see a network failure, the exit code is consistently 0.  Here is the relevant section of output from running the command with the `--debug` flag.

`
DEBUG: urllib3.connectionpool : Starting new HTTPS connection (1): REDACTED.azurecr.io:443
DEBUG: urllib3.connectionpool : https://REDACTED.azurecr.io:443 "POST /oauth2/exchange HTTP/1.1" 200 None
Error response from daemon: Get https://REDACTED.azurecr.io/v2/: dial tcp: lookup REDACTED.azurecr.io on [::1]:53: read udp [::1]:55595->[::1]:53: read: connection refused
DEBUG: Event: CommandInvoker.OnTransformResult [<function _resource_group_transform at 0x7f87591552a8>, <function _x509_from_base64_to_hex_transform at 0x7f8759155320>]
DEBUG: Event: CommandInvoker.OnFilterResult []
DEBUG: Event: Cli.PostExecute []
INFO: az_command_data_logger : exit code: 0
`
## Expected Behavior

On any failure to successfully complete the login, I would expect the process to return an exit code other than 0 to indicate failure.

## Environment Summary
```
Linux-3.10.0-1062.4.3.el7.x86_64-x86_64-with-centos-7.7.1908-Core
Python 2.7.5
Shell: bash

azure-cli 2.0.76

```

Wwwsylvia:
Fix #11355 

If there are errors returned by docker command, the CLI exit code will be non-zero 

image

**History Notes:**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acr/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/acr/custom.py
@@ -193,12 +193,14 @@ def acr_login(cmd,
                        login_server])
             p.wait()
         else:
+            stderr_messages = stderr.decode()
             if b'--password-stdin' in stderr:
-                errors = [err for err in stderr.decode().split('\n') if '--password-stdin' not in err and err != '']
-
-                if errors:
-                    stderr = '\n'.join(errors)
-                    raise CLIError('docker: ' + stderr)
+                errors = [err for err in stderr_messages.split('\n') if err and '--password-stdin' not in err]
+                # Will not raise CLIError if there is no error other than '--password-stdin'
+                if not errors:
+                    return None
+                stderr_messages = '\n'.join(errors)
+            raise CLIError('docker: ' + stderr_messages)
 
     return None
 
--- a/src/azure-cli/azure/cli/command_modules/acr/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/acr/custom.py
@@ -194,12 +194,11 @@ def acr_login(cmd,
             p.wait()
         else:
             if b'--password-stdin' in stderr:
-                errors = [err for err in stderr.decode().split('\n') if '--password-stdin' not in err]
-                stderr = '\n'.join(errors).encode()
+                errors = [err for err in stderr.decode().split('\n') if '--password-stdin' not in err and err != '']
 
-            import sys
-            output = getattr(sys.stderr, 'buffer', sys.stderr)
-            output.write(stderr)
+                if errors:
+                    stderr = '\n'.join(errors)
+                    raise CLIError('docker: ' + stderr)
 
     return None
 


src/azure-cli/azure/cli/command_modules/acr/custom.py
```diff
@@ -194,12 +194,11 @@ def acr_login(cmd,
...
             if b'--password-stdin' in stderr:
-                errors = [err for err in stderr.decode().split('\n') if '--password-stdin' not in err]
-                stderr = '\n'.join(errors).encode()
+                errors = [err for err in stderr.decode().split('\n') if '--password-stdin' not in err and err != '']
 
-            import sys
-            output = getattr(sys.stderr, 'buffer', sys.stderr)
-            output.write(stderr)
+                if errors:
+                    stderr = '\n'.join(errors)
```
yugangw-msft: @djyou, before I merge this change, want to confirm that it was not on purpose we void the exit code of 1

src/azure-cli/azure/cli/command_modules/acr/custom.py
```diff
@@ -194,12 +194,11 @@ def acr_login(cmd,
             p.wait()
         else:
             if b'--password-stdin' in stderr:
-                errors = [err for err in stderr.decode().split('\n') if '--password-stdin' not in err]
-                stderr = '\n'.join(errors).encode()
+                errors = [err for err in stderr.decode().split('\n') if '--password-stdin' not in err and err != '']
```
djyou: if err and '--password-stdin' not in err

src/azure-cli/azure/cli/command_modules/acr/custom.py
djyou: We should raise regardless if ever reaching here, try the following
```
            if b'--password-stdin' in stderr:
                errors = ...
                stderr = '\n'.join(errors) if errors else 'some generic message'
            raise CLIError('docker: ' + stderr)
```
nit: not sure about the docker prefix in message but will let you decide.

Wwwsylvia: > We should raise regardless if ever reaching here, try the following
> 
> ```
>             if b'--password-stdin' in stderr:
>                 errors = ...
>                 stderr = '\n'.join(errors) if errors else 'some generic message'
>             raise CLIError('docker: ' + stderr)
> ```


I think the idea here is that, if the error is just "--password-stdin..." (meaning the login has succeed), the function should return normally and we should not show any error messages to user. if there are errors other than "--password-stdin", then we should raise

djyou: What do we print if --password-stdin is not in stderr?

Wwwsylvia: Should raise error. Updated in latest commit.

djyou: LGTM

Issue #11981, opened 2020-01-29
The az resource create example doesn't work


jacekmackiewicz:
Example: az resource create -g myRG -n myApiApp --resource-type Microsoft.web/sites --is-full-object --properties '{
            "kind": "api",
            "location": "West US",
            "properties": {
                "serverFarmId": "/subscriptions/{SubID}/resourcegroups/{ResourceGroup}/providers/Microsoft.Web/serverfarms/{ServicePlan}"
            }
        }'
doesn't work, giving a "Error parsing JSON.
{
Expecting property name enclosed in double quotes: line 1 column 2 (char 1)".

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: e76a8ad2-efd8-9e8a-9b2b-eebeb1e9d1d4
* Version Independent ID: 280b7ea9-eb6c-9601-464b-09bf8e0cff5d
* Content: [az resource](https://docs.microsoft.com/en-us/cli/azure/resource?view=azure-cli-latest&source=docs#feedback)
* Content Source: [src/azure-cli/azure/cli/command_modules/resource/_help.py](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/resource/_help.py)
* Service: **azure-resource-manager**
* GitHub Login: @rloutlaw
* Microsoft Alias: **routlaw**

yonzhan:
add to S165.

happyiphone:
Any news on this? I need to be able to use this command or an alternative for it. Any help on this will be highly appreciated. thanks!

zhoxing-ms:
@happyiphone Thanks for your feedback, It seems that example is not friendly to users, I will submit a PR to improve it

zhoxing-ms:
@happyiphone For the `az resource create` command, please use like this:
`az resource create -g myRG -n myApiApp --resource-type Microsoft.web/sites --is-full-object --properties "{ \"kind\": \"api\", \"location\": \"West US\", \"properties\": { \"serverFarmId\": \"/subscriptions/{SubID}/resourcegroups/{ResourceGroup}/providers/Microsoft.Web/serverfarms/{ServicePlan}\" } }"`

surexxx:
Examples above doesn't work for me in Azure Portal Cloud Shell.
This is a working example that will successfully run: 
[...]

PR #12167, opened 2020-02-13
[ARM] az resource: Improve the examples of the resource module (#11981)

zhoxing-ms:
Issue: #11981

In some of our original examples, there are the following problems:
1. The JSON data in the parameter is newlined
2. `"` in the JSON data is not escaped
3. The parameters are referenced with `'`

These problems prevent users from using our example intuitively.
So I'm going to modify the example of resource module first. If there are similar problems in other modules in the future, they can be solved in this way.

**History Notes:**  

---


yonzhan:
add to S165.

PR #12173, opened 2020-02-13
[netappfiles] az volume replication: Modified volume create to allow data protection volumes and added replication operations

audunn:
**History Notes:**  

[NetAppFiles] az volume create: Allow data protection volumes adding replication operations: approve, suspend, resume, status, remove

---


yonzhan:
add to S166.

haroldrandom:
Hi, @audunn, I found two things may help:
1. You have modified your test code but without updating YAML accordingly. You have to re-recording those recording YAML files as well.
2. I found you are using Python 2.7? Recently, Python 2.7 has not been supported any more, could you use Python 3.5+ instead?

audunn:
Thank you @haroldrandom  I am updating with new test recordings. Had some issues getting JIT access etc.  to run the tests those should be sorted now.  
I can run them again f.ex. on Python 3.7.5 if that is required. Can you advice if that is needed?

audunn:
Please advise if we need to do more on this.

audunn:
Can someone look at this and advise if more work is needed on this.

audunn:
Hi, This has been sitting for a while but we found out we need to update this PR changing the command pause to suspend.

Explanation is that this is also implemented in Azure Powershell and we found that that pause is not on the list for approved verbs for Cmdlets in Powershell according to [cmdlet best practices](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/design-guidelines/cmdlet-best-practices.md) document. 

To avoid confusion we would like to use the same verbs so I will update this with the change shortly.

src/azure-cli/azure/cli/command_modules/netappfiles/_help.py
```diff
@@ -338,6 +338,14 @@
...
   - name: --protocol-types
     short-summary: Space seperated list of protocols that the volume can use
+  - name: --volume-type
+    short-summary: Whether the volume should be a data protection volume ("DataProtection")
+  - name: --endpoint-type
+    short-summary: Whether the volume is source ("src") or destination ("dst")
+  - name: --remote-volume-resource-id
+    short-summary: The volume id of the remote volume of the replication (the destination for "src" volume endpoints and the source for "dst" endpoints)
+  - name: --replication-schedule
+    short-summary: The replication schedule, e.g. "_10minutely"
```
Juliehzl: I have some questions for the paramwter:
1. So the available value will be something like "_10minutely"? and how can I represent seconds? hours? 
2. Is it something like interval?

src/azure-cli/azure/cli/command_modules/netappfiles/_help.py
```diff
@@ -362,6 +370,93 @@
...
+short-summary: Authorize a volume as a replication destination for a specified source.
+parameters:
+  - name: --account-name -a
+    short-summary: The name of the ANF account
+  - name: --pool-name -p
+    short-summary: The name of the ANF pool
+  - name: --name --volume-name -n -v
+    short-summary: The name of the replication source volume
+  - name: --remote-volume-resource-id -d
+    short-summary: The resource id of the destination replication volume
```
Juliehzl: Can we both support destination volume name  and resource id here?

src/azure-cli/azure/cli/command_modules/netappfiles/_help.py
```diff
@@ -338,6 +338,14 @@
     short-summary: The ARM Id or name of the subnet for the vnet. If omitted 'default' will be used
   - name: --protocol-types
     short-summary: Space seperated list of protocols that the volume can use
+  - name: --volume-type
+    short-summary: Whether the volume should be a data protection volume ("DataProtection")
```
Juliehzl: Which are available volume types? Can you show up them in help message?

src/azure-cli/azure/cli/command_modules/netappfiles/_help.py
```diff
@@ -338,6 +338,14 @@
...
+  - name: --endpoint-type
+    short-summary: Whether the volume is source ("src") or destination ("dst")
```
Juliehzl: There are two questions here:
1. Can you use get_enum_type() for allowed values?
2. With this parameter, I can create a source  or destination volume, how can I associate the source and destination?

src/azure-cli/azure/cli/command_modules/netappfiles/_help.py
```diff
@@ -362,6 +370,93 @@
...
+    short-summary: The name of the ANF pool
+  - name: --name --volume-name -n -v
+    short-summary: The name of the replication destination volume
+examples:
+  - name: Delete the replication objects of the paired volumes
+    text: >
+        az netappfiles volume replication remove -g mygroup --account-name myaccname --pool-name mypoolname --name mydestinationvolname
+"""
+
+helps['netappfiles volume replication status'] = """
```
Juliehzl: What about changing to show command?

src/azure-cli/azure/cli/command_modules/netappfiles/commands.py
```diff
@@ -126,6 +126,16 @@ def load_command_table(self, _):
...
                                  exception_handler=netappfiles_exception_handler)
 
+    with self.command_group('netappfiles volume replication', netappfiles_volumes_sdk) as g:
+        g.custom_command('approve', 'authorize_replication',
+                         client_factory=volumes_mgmt_client_factory,
+                         doc_string_source='azure.mgmt.netapp.models#Volume',
+                         exception_handler=netappfiles_exception_handler)
+        g.command('suspend', 'break_replication')
+        g.command('resume', 'resync_replication')
+        g.command('remove', 'delete_replication')
```
Juliehzl: what about changing to `show` command and provide confirmation kwargs?

src/azure-cli/azure/cli/command_modules/netappfiles/tests/latest/test_volume_commands.py
```diff
@@ -84,21 +96,75 @@ def test_create_volume_with_subnet_in_different_rg(self):
...
         subs_id = self.current_subscription()
-        self.cmd("az group create -n %s --subscription %s -l %s" % (subnet_rg, subs_id, LOCATION)).get_output_in_json()
+        self.cmd("az group create -n %s --subscription %s -l %s" % (subnet_rg, subs_id, RG_LOCATION)).get_output_in_json()
 
         rg = '{rg}'
-        self.setup_vnet(subnet_rg, vnet_name, subnet_name, '10.0.0.0')
-        self.cmd("az netappfiles account create -g %s -a %s -l %s" % (rg, account_name, LOCATION)).get_output_in_json()
-        self.cmd("az netappfiles pool create -g %s -a %s -p %s -l %s %s" % (rg, account_name, pool_name, LOCATION, POOL_DEFAULT)).get_output_in_json()
+        self.setup_vnet(subnet_rg, vnet_name, subnet_name, '10.0.0.0', RG_LOCATION)
+        self.cmd("az netappfiles account create -g %s -a %s -l %s" % (rg, account_name, ANF_LOCATION)).get_output_in_json()
```
Juliehzl: For all your tests, can we add some checks or assert for command results?

src/azure-cli/azure/cli/command_modules/netappfiles/tests/latest/test_volume_commands.py
```diff
@@ -84,21 +96,75 @@ def test_create_volume_with_subnet_in_different_rg(self):
...
+        self.cmd("az netappfiles volume replication resume -g %s -a %s -p %s -v %s" % (rg_r, account_name_r, pool_name_r, volume_name_r))
+        self.wait_for_replication_status("Mirrored", rg_r, account_name_r, pool_name_r, volume_name_r)
+
+        # break
+        self.cmd("az netappfiles volume replication suspend -g %s -a %s -p %s -v %s" % (rg_r, account_name_r, pool_name_r, volume_name_r))
+        self.wait_for_replication_status("Broken", rg_r, account_name_r, pool_name_r, volume_name_r)
+
+        # delete
+        self.cmd("az netappfiles volume replication remove -g %s -a %s -p %s -v %s" % (rg_r, account_name_r, pool_name_r, volume_name_r))
+        time.sleep(2)
```
Juliehzl: why sleep here?

src/azure-cli/azure/cli/command_modules/netappfiles/_help.py
audunn: Yes this in interval  hourly, daily, weekly, monthly.

src/azure-cli/azure/cli/command_modules/netappfiles/_help.py
audunn: The remote-volume-resource-id parameter expects the resource id.

src/azure-cli/azure/cli/command_modules/netappfiles/_help.py
audunn: At this time this can be left empy or set to "DataProtection".  There are plans to change this to an enum in a later version.

audunn:
> 
> 
> In summary, I have some concern for your PR:
> 
>     1. I will prefer you move parameter definitions to _params.py and provide available values for your some types related parameter.
>        You can abstract common parameters using CLIArgumentType and reuse them in commands.
> 
>     2. For existing volume create command, it is a long running operation, right? If yes, I will recommend you support no wait for such commands and add wait command.
> 
>     3. For all your tests, it will be better to add checks.

Your points are valid and appreciated. Note that this is in preview and marked as such.  
This has been waiting for a while now, and we will need to get the next version out very soon, therefor I suggest we implement those recommendations in the next version if acceptable.

src/azure-cli/azure/cli/command_modules/netappfiles/commands.py
audunn: Unless I am misunderstanding I am not sure show is appropriate here this is a custom command that performs an action on approves/authorize the replication connection on the source volume.

src/azure-cli/azure/cli/command_modules/netappfiles/_help.py
audunn: I suggest we do this change for next version/update that will come very soon.

src/azure-cli/azure/cli/command_modules/netappfiles/_help.py
audunn: I suggest we do this change for next version/update that will come very soon.

src/azure-cli/azure/cli/command_modules/netappfiles/tests/latest/test_volume_commands.py
audunn: I suggest we do this change for next version/update that will come very soon.

src/azure-cli/azure/cli/command_modules/netappfiles/tests/latest/test_volume_commands.py
audunn: I think this is needed to ensure this completes before any cleanup after else it mayl try and delete the resources before they are ready.

fengzhou-msft:
@audunn I flattened the history notes in PR description as the current script does not support multiple lines for one history note.

audunn:
Hi this has been sitting for a while, what is needed to get this merged.

yonzhan:
@audunn @Juliehzl  There are several conflicting files in this PR.

audunn:
@yonzhan  Apparently Network was updated in the meantime causing this conflict. Has been resolved.

Issue #6687, opened 2018-06-29
Table Storage Policy List Doesn't return an array


ninijay:
**Describe the bug**
When listing all the policies for a given table storage, the az tool returns an object with sub- objects rather than an array of objects

**To Reproduce**
1. Create Table Storage
2. Add 2 or more Access Policies
![image](https://user-images.githubusercontent.com/6756364/42080272-63a38dce-7b82-11e8-9c7d-d64cd56e7f23.png)
3. Query for a list of policies for that table:
```bash
az storage table policy list --table-name <tablename> --account-name <storagename>
```
4. Get the "malformed" result:
```
{
  "policy1": {
    "expiry": null,
    "permission": "rau",
    "start": null
  },
  "policy2": {
    "expiry": null,
    "permission": "d",
    "start": null
  }
}
```

**Expected behavior**
Since it is a list, the result should be returned as an array/list and should look like this:
```
[
  "policy1": {
    "expiry": null,
    "permission": "rau",
    "start": null
  },
  "policy2": {
    "expiry": null,
    "permission": "d",
    "start": null
  }
]
```

**Environment summary**
I'm working in the cloud bash shell

**Additional context**
none

ninijay:
Same in other storage types like queues and shares

tjprescott:
@williexu, since we are using the list verb, we should consider using a transform to force the returned object to be a list. We would also need to move the policy name into the dictionary or wrap each policy in a dictionary.

limingu:
@ninijay, we can change the the result as a list. It will be like this:
[
  {
    "policy1": {
      "expiry": null,
      "permission": "rau",
      "start": null
    },
  },
  {
    "policy2": {
      "expiry": null,
      "permission": "d",
      "start": null
    }
  }
]

That actually makes it hard to find a policy name. As a list, you may have to go through the whole list to find the policy. The result as a dictionary seems make more sense here.

tjprescott:
@limingu the issue is that the command is called a list command does not return a list. We can force it to a more normal format like:

```
[
  {
      "name": "policy1",
      "expiry": null,
      "permission": "rau",
      "start": null
  }
]
```

This would be consistent with the rest of Azure and thus the user could employ the same patterns for searching for the desired object.

zhoxing-ms:
@ninijay Hi, after discussions with our PM, we think this will trigger a breaking change, and we are cautious about it. 
[...]

PR #12183, opened 2020-02-13
[Storage] BREAKING CHANGE: az storage table policy list: Change the returned data to a list (#6687)

zhoxing-ms:
Feature request: #6687

**History Notes:**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/storage/_transformers.py
+++ b/src/azure-cli/azure/cli/command_modules/storage/_transformers.py
@@ -130,4 +130,17 @@ def transform_url(result):
     result = re.sub('/', '//', result, count=1)
     return encode_url_path(result)
 
+
+def transform_storage_table_policy_list_output(result):
+    new_result = []
+    if result is None:
+        return new_result
+    for policy_name in result.keys():
+        new_entry = {'name': policy_name}
+        if result[policy_name] is not None:
+            for attr in result[policy_name].__dict__:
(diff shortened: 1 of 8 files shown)


yonzhan:
add to S166.

src/azure-cli/azure/cli/command_modules/storage/_transformers.py
```diff
@@ -130,4 +130,17 @@ def transform_url(result):
     result = re.sub('/', '//', result, count=1)
     return encode_url_path(result)
 
+
+def transform_storage_table_policy_list_output(result):
+    new_result = []
+    if result is None:
+        return new_result
+    for policy_name in result.keys():
+        new_entry = {'name': policy_name}
```
Juliehzl: It looks that you changed output structure compared to previous format, which may cause breaking change to users and is not expected. May I know the reason why you have such design? and can we consider not changing output format only with [] change?

zhoxing-ms: Yes, this is a breaking change.
If only change the outermost `{}` to `[]`, the returned one will not be in the correct JSON format
If only `[]` is added to the outermost layer, it is also breaking change, and this format is not really returned as a list
My return format is the one recommended by tjprescott in the issue, he means
> This would be consistent with the rest of Azure and thus the user could employ the same patterns for searching for the desired object.

PR #12204, opened 2020-02-14
[KeyVault] keyvault create: enable soft-delete by default

### LINKED ISSUE #11039: Soft-delete on by default via the Azure CLI
For Azure Key Vault, the new version of the API enables by default that soft-delete is true.  We are asking the CLI team to support "on" or true by default.

The existing behavior will now change.  By default, the CLI should set this to true by default.  If they leave it blank, this will now default to true.  They will need to specific false specifically to ensure this is off.

Today's behavior is as follows:
https://docs.microsoft.com/en-us/azure/key-vault/key-vault-soft-delete-cli

Specifying blank should enable this to be true (not false) as it is today.

jiasli:
Resolve #11039: Soft-delete on by default via the Azure CLI

Previously in `keyvault create`, if the user doesn't specify `--enable-soft-delete`, soft-delete will not be turned on - the `properties.enableSoftDelete` attribute in the returned JSON is `null`.

This PR bumps the API version so that now in `keyvault create`, if `--enable-soft-delete` is not specified, the keyvault will be created with soft-delete turned on by default -  the `properties.enableSoftDelete` attribute is `true`.

User will need to explicitly specify `--enable-soft-delete false` to turn off soft-delete.

To test, run 
```
az keyvault create -g {} -n {}
``` 
and check the `properties.enableSoftDelete` attribute to be `true`.

---


src/azure-cli-core/azure/cli/core/profiles/_shared.py
```diff
@@ -138,7 +138,7 @@ def default_api_version(self):
         ResourceType.MGMT_RESOURCE_RESOURCES: '2019-07-01',
         ResourceType.MGMT_RESOURCE_SUBSCRIPTIONS: '2019-06-01',
         ResourceType.MGMT_NETWORK_DNS: '2018-05-01',
-        ResourceType.MGMT_KEYVAULT: '2018-02-14',
+        ResourceType.MGMT_KEYVAULT: '2019-09-01',
```
bim-msft: Many of KV features are using `2018-02-14` folder, including newly added Private Link, would this change causes multi-api problems?

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/recordings/test_azconfig_mgmt.yaml
```diff
@@ -513,7 +513,7 @@ interactions:
       - python/3.5.2 (Windows-10-10.0.18362-SP0) msrest/0.6.10 msrest_azure/0.6.2
         azure-mgmt-keyvault/1.1.0 Azure-SDK-For-Python AZURECLI/2.0.81
```
bim-msft: It seems that you are still using `azure-mgmt-keyvault/1.1.0`, actually the new version (2.1.1) of this SDK had already been released, and it was bumped in the Private Link PR.

jiasli: I don't want to re-run those tests other than keyvault. There are so many pitfalls and manual intervention needed in the live run. Also the request header isn't considered in the vcr matcher.

src/azure-cli/azure/cli/command_modules/keyvault/_params.py
```diff
@@ -87,6 +87,7 @@ class CLIJsonWebKeyOperation(str, Enum):
         c.argument('sku', arg_type=get_enum_type(SkuName, default=SkuName.standard.value))
         c.argument('no_self_perms', arg_type=get_three_state_flag(), help="Don't add permissions for the current user/service principal in the new vault.")
         c.argument('location', validator=get_default_location_from_resource_group)
+        c.argument('enable_soft_delete', arg_type=get_three_state_flag(), help='Enable vault deletion recovery for the vault, and all contained entities. If omitted, assume true as default value.')
```
Juliehzl: I have one concern here. As what we have mentioned yesterday, if there is no misunderstanding, we cannot disable the feature after enabling, right?

src/azure-cli-core/azure/cli/core/profiles/_shared.py
jiasli: The `latest` profile should always use the latest API. As long as the live tests pass, we can assume it won't cause problems for new features like private link.

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/recordings/test_azconfig_mgmt.yaml
Juliehzl: I have the same concern.  And it looks in azure-mgmt-keyvault 1.1.0 there is no v2019-09-01 SDKs.

src/azure-cli/azure/cli/command_modules/keyvault/_params.py
jiasli: > We cannot disable the feature after enabling, right?

No, we can't. This sentence will be included in the next version of REST spec: 

https://github.com/Azure/azure-rest-api-specs/pull/8301/files#diff-ba200dfe0b4825793d1b20fb873061d2R1125

"description": "Property to specify whether the 'soft delete' functionality is enabled for this key vault. If it's not set to any value(true or false) when creating new key vault, it will be set to true by default. Once it's been set to true value, it can NOT be reverted to false."

I am still working with keyvault team to change the phrasing so that we can directly map from the REST spec. 

> And if no --enable-soft-delete parameter disable is provided, it will enable soft delete for keyvault resource.

Yes. Soft-delete applies to data-plane objects too.

src/azure-cli/azure/cli/command_modules/sql/tests/latest/recordings/test_sql_mi_tdebyok.yaml
```diff
@@ -13956,7 +13956,7 @@ interactions:
       accept-language:
       - en-US
     method: PUT
-    uri: https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/clitest000001/providers/Microsoft.KeyVault/vaults/sqltdebyok000005?api-version=2018-02-14
+    uri: https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/clitest000001/providers/Microsoft.KeyVault/vaults/sqltdebyok000005?api-version=2019-09-01
```
Juliehzl: You just replace the api version string and no actual live test running for these scenarios, right? Is it any problem when running test in live mode?

jiasli: I didn't. I am blocked by `azdev` indefinitely hang on Linux. Still working on it.

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/recordings/test_azconfig_mgmt.yaml
bim-msft: > I don't want to re-run those tests other than keyvault. There are so many pitfalls and manual intervention needed in the live run. Also the request header isn't considered in the vcr matcher.

Yes, I think it's not necessary to re-run all tests, but my concern is: are you using `1.1.0` for recording? If yes, you should better upgrade it to `2.1.1` because this version was already written in pip requirements, all newly modified yaml should use the latest packages for the consistency concern.

jiasli: I didn't re-run the test live. The recording YAML is manually edited. I couldn't run `azdev` on all tests as it hangs forever. I couldn't even find out which test hangs.

bim-msft: Thanks, got it.

jiasli: @Juliehzl, I have live run the storage test `test_customer_managed_key` and it works well. The recording now shows `azure-mgmt-keyvault/2.1.1`:

https://github.com/Azure/azure-cli/blob/d09618e9669a097fde3c50fa3ac0ea0785b4053c/src/azure-cli/azure/cli/command_modules/storage/tests/latest/recordings/test_customer_managed_key.yaml#L128-L130

Kindly review it.

yonzhan:
add to S166.

jiasli:
As indicated by https://github.com/Azure/azure-rest-api-specs/pull/8301, `softDeleteRetentionInDays` will still be in [2019-09-01/keyvault.json](https://github.com/Azure/azure-rest-api-specs/blob/c3329e8fd5b2745c9fa1572d9c59ebe2554151d7/specification/keyvault/resource-manager/Microsoft.KeyVault/stable/2019-09-01/keyvault.json#L1128), so we don't need to bump API version again later. Let's merge this PR.

bim-msft:
Core functionalities look good to me.

bim-msft:
Another concern: should the title prefix be `[KeyVault]` instead of `[Keyvault]`? Is the release notes  script case sensitive?

jiasli:
Nice catch!

src/azure-cli/azure/cli/command_modules/keyvault/_params.py
qwordy: Where is this variable used? I can't find usage in this PR.

jiasli: It is not used. I added it to separate the help message of `--enable-soft-delete` in `keyvault create` from that in `keyvault update`.

Issue #11765, opened 2020-01-03
[ARM] Expose cross tenant for group deployment command


Juliehzl:
**Is your feature request related to a problem? Please describe.**
`az group deployment create --aux-tenants`

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

Juliehzl:
asked by customer

yonzhan:
add to S164.

PR #12221, opened 2020-02-15
[ARM] az group deployment create: Add parameter `--aux-tenants` to support cross tenants (#11765)

zhoxing-ms:
Feature request: #11765

**History Notes:**  

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/_profile.py
+++ b/src/azure-cli-core/azure/cli/core/_profile.py
@@ -534,7 +534,10 @@ class Profile(object):
                 return parts[0], (None if len(parts) <= 1 else parts[1])
         return None, None
 
-    def get_login_credentials(self, resource=None, subscription_id=None, aux_subscriptions=None):
+    def get_login_credentials(self, resource=None, subscription_id=None, aux_subscriptions=None, aux_tenants=None):
+        if aux_tenants and aux_subscriptions:
+            raise CLIError("Please specify only one of aux_subscriptions and aux_tenants, not both")
+
         account = self.get_subscription(subscription_id)
         user_type = account[_USER_ENTITY][_USER_TYPE]
         username_or_sp_id = account[_USER_ENTITY][_USER_NAME]
@@ -543,12 +546,14 @@ class Profile(object):
         identity_type, identity_id = Profile._try_parse_msi_account_name(account)
 
         external_tenants_info = []
-        ext_subs = [aux_sub for aux_sub in (aux_subscriptions or []) if aux_sub != subscription_id]
-        for ext_sub in ext_subs:
-            sub = self.get_subscription(ext_sub)
-            if sub[_TENANT_ID] != account[_TENANT_ID]:
-                # external_tenants_info.append((sub[_USER_ENTITY][_USER_NAME], sub[_TENANT_ID]))
-                external_tenants_info.append(sub)
+        if aux_tenants:
+            external_tenants_info = [tenant for tenant in aux_tenants if tenant != account[_TENANT_ID]]
+        if aux_subscriptions:
+            ext_subs = [aux_sub for aux_sub in aux_subscriptions if aux_sub != subscription_id]
+            for ext_sub in ext_subs:
+                sub = self.get_subscription(ext_sub)
+                if sub[_TENANT_ID] != account[_TENANT_ID]:
+                    external_tenants_info.append(sub[_TENANT_ID])
 
         if identity_type is None:
             def _retrieve_token():
@@ -564,13 +569,13 @@ class Profile(object):
 
             def _retrieve_tokens_from_external_tenants():
                 external_tokens = []
-                for s in external_tenants_info:
+                for sub_tenant_id in external_tenants_info:
                     if user_type == _USER:
                         external_tokens.append(self._creds_cache.retrieve_token_for_user(
-                            username_or_sp_id, s[_TENANT_ID], resource))
+                            username_or_sp_id, sub_tenant_id, resource))
                     else:
                         external_tokens.append(self._creds_cache.retrieve_token_for_service_principal(
-                            username_or_sp_id, resource, s[_TENANT_ID], resource))
+                            username_or_sp_id, resource, sub_tenant_id, resource))
                 return external_tokens
 
             from azure.cli.core.adal_authentication import AdalAuthentication
--- a/src/azure-cli/azure/cli/command_modules/resource/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/resource/_params.py
@@ -241,8 +241,11 @@ def load_arguments(self, _):
     with self.argument_context('deployment group create') as c:
         c.argument('deployment_name', arg_type=deployment_create_name_type)
         c.argument('handle_extended_json_format', arg_type=extended_json_format_type)
-        c.argument('aux_subscriptions', nargs='*', options_list=['--aux-subs'],
-                   help='Auxiliary subscriptions which will be used during deployment across tenants.')
+        c.argument('aux_subscriptions', nargs='+', options_list=['--aux-subs'],
+                   help='Auxiliary subscriptions which will be used during deployment across tenants.',
+                   deprecate_info=c.deprecate(target='--aux-subs', redirect='--aux-tenants'))
+        c.argument('aux_tenants', nargs='+', options_list=['--aux-tenants'],
+                   help='Auxiliary tenants which will be used during deployment across tenants.')
 
     with self.argument_context('deployment group validate') as c:
         c.argument('deployment_name', arg_type=deployment_create_name_type)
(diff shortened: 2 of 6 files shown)


yonzhan:
add to S166.

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -543,12 +543,17 @@ def get_login_credentials(self, resource=None, subscription_id=None, aux_subscri
...
-        for ext_sub in ext_subs:
-            sub = self.get_subscription(ext_sub)
-            if sub[_TENANT_ID] != account[_TENANT_ID]:
-                # external_tenants_info.append((sub[_USER_ENTITY][_USER_NAME], sub[_TENANT_ID]))
-                external_tenants_info.append(sub)
+        if isinstance(aux_tenants, list):
+            external_tenants_info = [tenant for tenant in aux_tenants if tenant != account[_TENANT_ID]]
+            if aux_subscriptions is not None:
+                logger.warning("When aux-subscriptions and aux-tenants pass in at the same time, "
+                               "only aux-tenants will work")
```
jiasli: I'd prefer to error out when both are specified, as giving warning instead encourages incorrect usage. 

https://github.com/Azure/azure-cli/blob/5edbd822ee860d7ba5d5182811b68233e0846ba8/src/azure-cli-core/azure/cli/core/_profile.py#L605

zhoxing-ms: It has been verified in the validator and an error has been returned for this scenario.
This method is a underlying method, which is also called in many other places. Just logging warning will have a smaller impact on different businesses, unless we can confirm that all scenarios calling it in the future can directly report an error.

jiasli: In other places, no invocation involves `aux_tenants`. Making them mutual-exclusive won't cause any breaking change.

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -543,12 +543,17 @@ def get_login_credentials(self, resource=None, subscription_id=None, aux_subscri
         identity_type, identity_id = Profile._try_parse_msi_account_name(account)
 
         external_tenants_info = []
-        ext_subs = [aux_sub for aux_sub in (aux_subscriptions or []) if aux_sub != subscription_id]
-        for ext_sub in ext_subs:
-            sub = self.get_subscription(ext_sub)
-            if sub[_TENANT_ID] != account[_TENANT_ID]:
-                # external_tenants_info.append((sub[_USER_ENTITY][_USER_NAME], sub[_TENANT_ID]))
-                external_tenants_info.append(sub)
+        if isinstance(aux_tenants, list):
```
jiasli:
```suggestion
        if aux_tenants:
```

src/azure-cli-core/azure/cli/core/_profile.py
zhoxing-ms: Yes, I know it has no effect on the existing calling methods. I mean, I didn't know if future calls should be waring or error.
I think you're right. I'll change the logic.

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -534,7 +534,10 @@ def _try_parse_msi_account_name(account):
                 return parts[0], (None if len(parts) <= 1 else parts[1])
         return None, None
 
-    def get_login_credentials(self, resource=None, subscription_id=None, aux_subscriptions=None):
+    def get_login_credentials(self, resource=None, subscription_id=None, aux_subscriptions=None, aux_tenants=None):
+        if aux_tenants and aux_subscriptions:
+            raise CLIError("Please specify only one of subscription and tenant, not both")
```
jiasli:
```suggestion
            raise CLIError("Please specify only one of aux_subscriptions and aux_tenants, not both")
```

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -543,12 +546,14 @@ def get_login_credentials(self, resource=None, subscription_id=None, aux_subscri
         identity_type, identity_id = Profile._try_parse_msi_account_name(account)
 
         external_tenants_info = []
-        ext_subs = [aux_sub for aux_sub in (aux_subscriptions or []) if aux_sub != subscription_id]
```
qianwens: > or []) [](start = 61, length = 7)

if both ext_sub and ext_tenants are empty, the original code append sub = self.get_subscription([]), your code removed this logic, please check if it will cause regression

zhoxing-ms: In the original code, when aux_subscriptions is empty, ext_subs is [], sub = self.get_subscription (ext_sub) will not be executed.
I understand that the previous code was written this way to prevent aux_subscriptions from having null pointer problems.

jiasli:
Let's add some tests for the change at 

https://github.com/Azure/azure-cli/blob/34549d2db993cfeab6dad21aa38ba1a4b7fc33e7/src/azure-cli-core/azure/cli/core/tests/test_profile.py#L548

zhoxing-ms:
> Let's add some tests for the change at
> 
>   
>     
>       azure-cli/src/azure-cli-core/azure/cli/core/tests/test_profile.py
>     
>     
>          Line 548
>       in
>       34549d2
>     
>     
>     
>     
> 
>         
>           
>            def test_get_login_credentials_aux_subscriptions(self, mock_get_token, mock_read_cred_file):

done

zhoxing-ms:
> Let's add some tests for the change at

done

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -189,7 +189,10 @@ def load_arguments(self, _):
                    help='The deployment name. Default to template file base name')
         c.argument('handle_extended_json_format', arg_type=extended_json_format_type)
         c.argument('aux_subscriptions', nargs='*', options_list=['--aux-subs'],
-                   help='Auxiliary subscriptions which will be used during deployment across tenants.')
+                   help='Auxiliary subscriptions which will be used during deployment across tenants.',
+                   deprecate_info=c.deprecate(target='--aux-subs', redirect='--aux-tenants'))
+        c.argument('aux_tenants', nargs='*', options_list=['--aux-tenants'],
```
Juliehzl: Please use + not * to avoid [] for aux_tenants and aux_subs

src/azure-cli/azure/cli/command_modules/resource/tests/latest/test_resource.py
```diff
@@ -1514,6 +1519,48 @@ def test_group_deployment_crosstenant(self, resource_group):
             self.check('resourceGroup', '{vm_rg}')
         ])
 
+        self.cmd('group deployment create -g {vm_rg} -n {dn1} --template-file "{tf}" --parameters SIG_ImageVersion_id={sig_id} NIC_id={nic_id} --aux-tenants "{aux_tenant}"', checks=[
```
Juliehzl: Can you consider abstracting this part to a new test function?

Juliehzl:
Another concern here is that when PR #10751 merged, you may need to re-apply your commands change to all new commands in #10751.

src/azure-cli/azure/cli/command_modules/resource/tests/latest/test_resource.py
zhoxing-ms: This question has been considered before, my thoughts are: 
The name of this method is  `test_group_deployment_crosstenant()`, but the original test was only for cross aux_sub. 
Now I've added the test logic for cross aux_tenant, which is consistent with the method name.
My original plan was to directly delete the logic of testing aux_sub in `test_group_deployment_crosstenant()` after deprecating `--aux-sub`. So that some common code can be reused and the test with two parameters bound together can be implemented properly.
What do you think about it?

zhoxing-ms:
> Another concern here is that when PR #10751 merged, you may need to re-apply your commands change to all new commands in #10751.

Good suggestions! I am considering submitting a new PR to solve this problem, so that everyone can distinguish the new code I added later when reviewing.

src/azure-cli/azure/cli/command_modules/resource/tests/latest/test_resource.py
zhoxing-ms: And I will add a new test function for the refactored code, so I think it might be better to use only a function for the old code to distinguish the tests of the old code from the new code.

src/azure-cli/azure/cli/command_modules/resource/tests/latest/test_resource.py
```diff
@@ -1675,6 +1680,171 @@ def test_group_deployment_crosstenant(self, resource_group):
...
```
Tiano2017: >group deployment create [](start = 18, length = 23)

I don't see test recordings for this. Is it not needed for this change?

zhoxing-ms: Because this case is not easy to write as rerecord, so now it is a live_only mode.

PR #12233, opened 2020-02-17
[Monitor] az monitor clone: support clone metric rules from one resource to another.

MyronFanQiu:
**Background.**

Clone the metric alerts of one resource to another resource.
image

The alert rules and the resource must be in the same subscription. Thus, I add a requirement for this command right now that the source_resource and the target_resource must be in the same sub first.

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/monitor/_client_factory.py
+++ b/src/azure-cli/azure/cli/command_modules/monitor/_client_factory.py
@@ -6,9 +6,9 @@
 
 # MANAGEMENT CLIENT FACTORIES
 def cf_monitor(cli_ctx, **kwargs):
-    from azure.mgmt.monitor import MonitorManagementClient
+    from azure.cli.core.profiles import ResourceType
     from azure.cli.core.commands.client_factory import get_mgmt_service_client
-    return get_mgmt_service_client(cli_ctx, MonitorManagementClient, **kwargs)
+    return get_mgmt_service_client(cli_ctx, ResourceType.MGMT_MONITOR, **kwargs)
 
 
 def cf_alert_rules(cli_ctx, _):
--- a/src/azure-cli/azure/cli/command_modules/monitor/_client_factory.py
+++ b/src/azure-cli/azure/cli/command_modules/monitor/_client_factory.py
@@ -8,7 +8,7 @@
 def cf_monitor(cli_ctx, **kwargs):
     from azure.mgmt.monitor import MonitorManagementClient
     from azure.cli.core.commands.client_factory import get_mgmt_service_client
-    return get_mgmt_service_client(cli_ctx, MonitorManagementClient)
+    return get_mgmt_service_client(cli_ctx, MonitorManagementClient, **kwargs)
 
 
 def cf_alert_rules(cli_ctx, _):
@@ -28,7 +28,7 @@ def cf_diagnostics(cli_ctx, _):
 
 
 def cf_diagnostics_category(cli_ctx, _):
-    return cf_monitor(cli_ctx, _).diagnostic_settings_category
+    return cf_monitor(cli_ctx).diagnostic_settings_category
 
 
 def cf_log_profiles(cli_ctx, _):
@@ -66,7 +66,7 @@ def cf_metric_alerts(cli_ctx, _):
 def _log_analytics_client_factory(cli_ctx, **kwargs):
     from azure.mgmt.loganalytics import LogAnalyticsManagementClient
     from azure.cli.core.commands.client_factory import get_mgmt_service_client
-    return get_mgmt_service_client(cli_ctx, LogAnalyticsManagementClient)
+    return get_mgmt_service_client(cli_ctx, LogAnalyticsManagementClient, **kwargs)
 
 
 def cf_log_analytics_workspace(cli_ctx, _):
--- a/src/azure-cli/azure/cli/command_modules/monitor/_client_factory.py
+++ b/src/azure-cli/azure/cli/command_modules/monitor/_client_factory.py
@@ -5,26 +5,26 @@
 
 
 # MANAGEMENT CLIENT FACTORIES
-def cf_monitor(cli_ctx, _):
+def cf_monitor(cli_ctx, **kwargs):
     from azure.mgmt.monitor import MonitorManagementClient
     from azure.cli.core.commands.client_factory import get_mgmt_service_client
     return get_mgmt_service_client(cli_ctx, MonitorManagementClient)
 
 
 def cf_alert_rules(cli_ctx, _):
-    return cf_monitor(cli_ctx, _).alert_rules
+    return cf_monitor(cli_ctx).alert_rules
 
 
 def cf_alert_rule_incidents(cli_ctx, _):
-    return cf_monitor(cli_ctx, _).alert_rule_incidents
+    return cf_monitor(cli_ctx).alert_rule_incidents
 
 
 def cf_autoscale(cli_ctx, _):
-    return cf_monitor(cli_ctx, _).autoscale_settings
+    return cf_monitor(cli_ctx).autoscale_settings
 
 
 def cf_diagnostics(cli_ctx, _):
-    return cf_monitor(cli_ctx, _).diagnostic_settings
+    return cf_monitor(cli_ctx).diagnostic_settings
 
 
 def cf_diagnostics_category(cli_ctx, _):
@@ -32,42 +32,42 @@ def cf_diagnostics_category(cli_ctx, _):
 
 
 def cf_log_profiles(cli_ctx, _):
-    return cf_monitor(cli_ctx, _).log_profiles
+    return cf_monitor(cli_ctx).log_profiles
 
 
 def cf_action_groups(cli_ctx, _):
-    return cf_monitor(cli_ctx, _).action_groups
+    return cf_monitor(cli_ctx).action_groups
 
 
 def cf_activity_log_alerts(cli_ctx, _):
-    return cf_monitor(cli_ctx, _).activity_log_alerts
+    return cf_monitor(cli_ctx).activity_log_alerts
 
 
 def cf_metrics(cli_ctx, _):
-    return cf_monitor(cli_ctx, _).metrics
+    return cf_monitor(cli_ctx).metrics
 
 
 def cf_metric_def(cli_ctx, _):
-    return cf_monitor(cli_ctx, _).metric_definitions
+    return cf_monitor(cli_ctx).metric_definitions
 
 
 def cf_activity_log(cli_ctx, _):
-    return cf_monitor(cli_ctx, _).activity_logs
+    return cf_monitor(cli_ctx).activity_logs
 
 
 def cf_event_categories(cli_ctx, _):
-    return cf_monitor(cli_ctx, _).event_categories
+    return cf_monitor(cli_ctx).event_categories
 
 
 def cf_metric_alerts(cli_ctx, _):
-    return cf_monitor(cli_ctx, _).metric_alerts
+    return cf_monitor(cli_ctx).metric_alerts
 
 
-def _log_analytics_client_factory(cli_ctx, _):
+def _log_analytics_client_factory(cli_ctx, **kwargs):
     from azure.mgmt.loganalytics import LogAnalyticsManagementClient
     from azure.cli.core.commands.client_factory import get_mgmt_service_client
     return get_mgmt_service_client(cli_ctx, LogAnalyticsManagementClient)
 
 
 def cf_log_analytics_workspace(cli_ctx, _):
-    return _log_analytics_client_factory(cli_ctx, _).workspaces
+    return _log_analytics_client_factory(cli_ctx).workspaces
--- a/src/azure-cli/azure/cli/command_modules/monitor/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/monitor/_help.py
@@ -1201,3 +1201,9 @@ examples:
         az monitor metrics list-definitions --resource /subscriptions/{subscriptionID}/resourceGroups/Space1999/providers/Microsoft.Network/networkSecurityGroups/ADDS-NSG
     crafted: true
 """
+
+helps['monitor clone'] = """
+type: command
+short-summary: Clone alert rules from one resource to another.
+long-summary: Due to the service limitation, this command would only clone the alert rules in the current subscription.
+"""
--- a/src/azure-cli/azure/cli/command_modules/monitor/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/monitor/_help.py
@@ -1204,6 +1204,5 @@ examples:
 
 helps['monitor clone'] = """
 type: command
-short-summary: Clone alert rules from one resource to another.
-long-summary: Due to the service limitation, this command would only clone the alert rules in the current subscription.
+short-summary: Clone metrics alert rules from one resource to another resource.
 """
--- a/src/azure-cli/azure/cli/command_modules/monitor/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/monitor/_help.py
@@ -1205,4 +1205,8 @@ examples:
 helps['monitor clone'] = """
 type: command
 short-summary: Clone metrics alert rules from one resource to another resource.
+examples:
+  - name: Clone the metric alert settings from one VM to another
+    text: |
+        az monitor clone --source-resource /subscriptions/{subscriptionID}/resourceGroups/Space1999/providers/Microsoft.Compute/virtualMachines/vm1 --target-resource /subscriptions/{subscriptionID}/resourceGroups/Space1999/providers/Microsoft.Compute/virtualMachines/vm2
 """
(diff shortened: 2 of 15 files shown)


yonzhan:
add to S166.

src/azure-cli/azure/cli/command_modules/monitor/operations/general_operations.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def _clone_monitor_metrics_alerts(cmd, source_resource, target_resource):
+    if not _is_resource_type_same(source_resource, target_resource):
+        raise CLIError('The target resource should be the same type with the source resource')
+    monitor_client = cf_monitor(cmd.cli_ctx)
+    # we can only clone the alert rules belonging to the current subscription.
+    alert_rules = list(monitor_client.metric_alerts.list_by_subscription())
+    updated_alert_rules = []
+    ErrorResponseException = cmd.get_models('ErrorResponseException', operation_group='metric_alerts')
```
Juliehzl: A question here: you get the client through directly import without resource type message, but here you use get_modes() here, which means you will import the model according to the api version defined in resource type.
How can you make sure the client (operations) and model are in the same api version?

src/azure-cli/azure/cli/command_modules/monitor/_params.py
```diff
@@ -343,3 +343,9 @@ def load_arguments(self, _):
         c.argument('intelligence_pack_name', options_list=['--name', '-n'])
         c.argument('workspace_name', options_list='--workspace-name')
     # endregion
+
+    # region monitor clone
+    with self.argument_context('monitor clone') as c:
+        c.argument('source_resource', help="Resource ID of the source resource.")
+        c.argument('target_resource', help="Resource ID of the target resource.")
```
Juliehzl: I have another concern here. It looks that you only have two parameters here with resource id as values. What about  the subscription in resource id are from two different tenants?

src/azure-cli/azure/cli/command_modules/monitor/operations/general_operations.py
```diff
@@ -0,0 +1,61 @@
...
+    metrics_alert_rules = _clone_monitor_metrics_alerts(cmd, source_resource, target_resource)
+    return metrics_alert_rules
+
+
+def _clone_monitor_metrics_alerts(cmd, source_resource, target_resource):
+    if not _is_resource_type_same(source_resource, target_resource):
+        raise CLIError('The target resource should be the same type with the source resource')
+    monitor_client = cf_monitor(cmd.cli_ctx)
+    # we can only clone the alert rules belonging to the current subscription.
+    alert_rules = list(monitor_client.metric_alerts.list_by_subscription())
```
Juliehzl: According to here, you only list alert rules in one subscription, right? And the concern is the same as above.

src/azure-cli/azure/cli/command_modules/monitor/operations/general_operations.py
```diff
@@ -0,0 +1,61 @@
...
+        return None
+    return updated_alert_rules if len(updated_alert_rules) > 1 else updated_alert_rules[0]
+
+
+def _is_resource_type_same(source_resource, target_resource):
+    source_dict = parse_resource_id(source_resource)
+    source_type = "{}/{}".format(source_dict['namespace'], source_dict['type'])
+    target_dict = parse_resource_id(target_resource)
+    target_type = "{}/{}".format(target_dict['namespace'], target_dict['type'])
+    return source_type == target_type
```
Juliehzl: It should the same as comparing both namespace part and type part, right?

src/azure-cli/azure/cli/command_modules/monitor/operations/general_operations.py
MyronFanQiu: yes.

src/azure-cli/azure/cli/command_modules/monitor/_params.py
MyronFanQiu: list alert in the sub of source resource

src/azure-cli/azure/cli/command_modules/monitor/operations/general_operations.py
MyronFanQiu: fix. init client from resource type

src/azure-cli/azure/cli/command_modules/monitor/_params.py
```diff
@@ -343,3 +343,9 @@ def load_arguments(self, _):
         c.argument('intelligence_pack_name', options_list=['--name', '-n'])
         c.argument('workspace_name', options_list='--workspace-name')
     # endregion
+
+    # region monitor clone
+    with self.argument_context('monitor clone') as c:
+        c.argument('source_resource', help="Resource ID of the source resource.")
```
qianwens: what is the options_list?

src/azure-cli/azure/cli/command_modules/monitor/operations/general_operations.py
```diff
@@ -0,0 +1,62 @@
...
+from azure.cli.core.commands.transform import _parse_id
+from knack.log import get_logger
+from knack.util import CLIError
+from msrestazure.tools import parse_resource_id
+
+
+logger = get_logger(__name__)
+
+
+def clone_existed_settings(cmd, source_resource=None, target_resource=None):
```
qianwens: >source_resource [](start = 32, length = 15)

do you need to check none for these two parameters?

src/azure-cli/azure/cli/command_modules/monitor/operations/general_operations.py
```diff
@@ -0,0 +1,62 @@
...
+    for alert_rule in alert_rules:
+        if source_resource in alert_rule.scopes:
+            if target_resource not in alert_rule.scopes:
+                try:
+                    alert_rule.scopes.append(target_resource)
+                    resource_group_name, name = _parse_id(alert_rule.id).values()
+                    alert_rule = monitor_client.metric_alerts.create_or_update(resource_group_name=resource_group_name,
+                                                                               rule_name=name, parameters=alert_rule)
+                except ErrorResponseException as ex:  # Create new alert rule
+                    if ex.response.status_code == 400:
```
qianwens: >ex.response.status_code == 400 [](start = 23, length = 30)

May I understand in what kind of situation, service return 400?

src/azure-cli/azure/cli/command_modules/monitor/operations/general_operations.py
```diff
@@ -0,0 +1,62 @@
...
+                                                                                   parameters=alert_rule)
+                    else:
+                        raise ex
+                updated_alert_rules.append(alert_rule)
+            else:
+                logger.warning('The target resource already has alert rule %s. '
+                               'Skip cloning this one.', alert_rule.name)
+    if not updated_alert_rules:
+        return None
+    return updated_alert_rules if len(updated_alert_rules) > 1 else updated_alert_rules[0]
```
qianwens: >updated_alert_rules [](start = 11, length = 19)

Why do you return different type : list, object, None. Is this CLI's recommendated output?

src/azure-cli/azure/cli/command_modules/monitor/commands.py
```diff
@@ -244,3 +250,6 @@ def load_command_table(self, _):
         g.command('list', 'list_intelligence_packs')
         g.command('enable', 'enable_intelligence_pack')
         g.command('disable', 'disable_intelligence_pack')
+
+    with self.command_group('monitor', metric_alert_sdk, custom_command_type=monitor_general_custom) as g:
+        g.custom_command('clone', 'clone_existed_settings')
```
qianwens: You only clone metric alerts in _clone_monitor_metrics_alerts, will you clone all settings in this command or you only clone alerts?

src/azure-cli/azure/cli/command_modules/monitor/_params.py
MyronFanQiu: `options_list` is used to rename the arguments. I don't want to rename them. Do you think we need a shorter name?

src/azure-cli/azure/cli/command_modules/monitor/operations/general_operations.py
MyronFanQiu: good catch. I will remove the default value to make it as a required parameter.

src/azure-cli/azure/cli/command_modules/monitor/operations/general_operations.py
MyronFanQiu: metric alert has different types. Each service support one or some of the types. For instance, metric alert for vm support multiple vms using one metric alert. However, for storage account, it's just a one-one mapping. So if I want to directly add a vm into the scope, it would succeed. If I want to add a storage account into the scope, service would return 400. There might be other situation that server returns 400. So I just have another try to create a new metric alert rule for the target_resource.

src/azure-cli/azure/cli/command_modules/monitor/commands.py
MyronFanQiu: yeah. good question. The original plan is to clone all settings. After talking with Arun, he said that currently the feature request is to clone metric alerts only. But I want to leave enough space for future change. So I don't change the command name.

src/azure-cli/azure/cli/command_modules/monitor/operations/general_operations.py
MyronFanQiu: I don't find any existed guideline for such scenario. I think it make senses that we only return list.(list, list with only one object, empty list). It's just a little bit strange for me to have an empty list. I will change the return type to list to keep consistence.

src/azure-cli/azure/cli/command_modules/monitor/commands.py
qianwens: what if later we need to support specific other settings to clone? should we add a parameter like --settings metric-alert

src/azure-cli/azure/cli/command_modules/monitor/operations/general_operations.py
qianwens: Can we just create new rules instead of adding resources into scopes? It is easier for user to track the copy result and revert the copy operation.

src/azure-cli/azure/cli/command_modules/monitor/commands.py
MyronFanQiu: good suggestion. I think we can do that. Currently, we can make `--settings` have default value `metric-alert`

src/azure-cli/azure/cli/command_modules/monitor/operations/general_operations.py
MyronFanQiu: sounds good. The cost of metric alert is not high. let me change it.

src/azure-cli/azure/cli/command_modules/monitor/util.py
```diff
@@ -54,4 +54,9 @@ def get_autoscale_scale_direction_map():
     from azure.mgmt.monitor.models import ScaleDirection
     return {'to': ScaleDirection.none, 'out': ScaleDirection.increase,
             'in': ScaleDirection.decrease}
+
+
+def _gen_guid():
+    import uuid
```
haroldrandom: no prefix `_` since this function is already used outside of this file.

src/azure-cli/azure/cli/command_modules/monitor/_params.py
```diff
@@ -343,3 +343,14 @@ def load_arguments(self, _):
...
+        c.argument('target_resource', help="Resource ID of the target resource.")
+        c.argument('always_clone', action='store_true',
+                   help="If this argument is applied, "
+                        "all monitor settings would be cloned instead of expanding its scope.")
+        c.argument('monitor_types', options_list=['--types', '-t'], arg_type=get_enum_type(['metricsAlert']),
+                   nargs='+', help='List of types of monitor settings which would be cloned.', default=['metricsAlert'])
```
haroldrandom: No enum class?

MyronFanQiu: no. It's decided by me. we can define an enum class for it. But I think list of string is enough.

haroldrandom: what's the purpose of this value?

MyronFanQiu: currently, we just support clone the metric alerts. We may support clone more monitor settings in the future. This is used to control the kinds of the monitor settings which would be cloned by this command.

haroldrandom: Usually, the option enum value is capitalized. Is it better to follow the convention?

src/azure-cli/azure/cli/command_modules/monitor/_help.py
```diff
@@ -1201,3 +1201,8 @@
         az monitor metrics list-definitions --resource /subscriptions/{subscriptionID}/resourceGroups/Space1999/providers/Microsoft.Network/networkSecurityGroups/ADDS-NSG
     crafted: true
 """
+
+helps['monitor clone'] = """
+type: command
+short-summary: Clone metrics alert rules from one resource to another resource.
+"""
```
jiasli: Maybe we can add some examples?

jiasli:
I don't think `test_keyvault_private_endpoint_connection.yaml` is affected. Maybe we can revert it?

src/azure-cli/azure/cli/command_modules/monitor/commands.py
jiasli: Shouldn't we mark this command as previous?

jiasli: Shouldn't we mark this command as preview?

src/azure-cli/azure/cli/command_modules/monitor/_help.py
MyronFanQiu: Sure

src/azure-cli/azure/cli/command_modules/monitor/_params.py
MyronFanQiu: Sure

src/azure-cli/azure/cli/command_modules/monitor/commands.py
MyronFanQiu: Sure

src/azure-cli/azure/cli/command_modules/monitor/_help.py
jsntcy: missing examples?

---
In reply to: [391417892](https://github.com/Azure/azure-cli/pull/12233#discussion_r391417892) [](ancestors = 391417892)

src/azure-cli/azure/cli/command_modules/monitor/operations/monitor_clone_util.py
```diff
@@ -0,0 +1,107 @@
...
+                    raise ex
+        updated_metrics_alert_rules.append(alert_rule)
+
+    return updated_metrics_alert_rules
+
+
+def _is_resource_type_same_and_sub_same(source_resource, target_resource):
+    source_dict = parse_resource_id(source_resource)
+    target_dict = parse_resource_id(target_resource)
+    same_rp = source_dict['namespace'] == target_dict['namespace'] and source_dict['type'] == target_dict['type']
```
jsntcy: should be case-insensitive here?

src/azure-cli/azure/cli/command_modules/monitor/_client_factory.py
```diff
@@ -5,69 +5,69 @@
...
-    return cf_monitor(cli_ctx, _).metric_alerts
+    return cf_monitor(cli_ctx).metric_alerts
 
 
-def _log_analytics_client_factory(cli_ctx, _):
+def _log_analytics_client_factory(cli_ctx, **kwargs):
     from azure.mgmt.loganalytics import LogAnalyticsManagementClient
     from azure.cli.core.commands.client_factory import get_mgmt_service_client
-    return get_mgmt_service_client(cli_ctx, LogAnalyticsManagementClient)
+    return get_mgmt_service_client(cli_ctx, LogAnalyticsManagementClient, **kwargs)
```
jsntcy: >kwargs [](start = 76, length = 6)

Why do we need add "**kwargs" here?

MyronFanQiu: left the room for future in case we need to input sub or aux_sub in the future.

src/azure-cli/azure/cli/command_modules/monitor/_help.py
MyronFanQiu: added

src/azure-cli/azure/cli/command_modules/monitor/operations/monitor_clone_util.py
jsntcy: I mean if target_dict['namespace'] should be case-insensitive?

---
In reply to: [394248241](https://github.com/Azure/azure-cli/pull/12233#discussion_r394248241) [](ancestors = 394248241)

MyronFanQiu: yes. I have make source_resource and target_resource to be lower case. So everything in the dict should be lower case

PR #12234, opened 2020-02-17
[Core] Remove color when stdout is not a tty

### LINKED ISSUE #2924: JSONC is not parsed successfully by json.loads in Python2.
### Environment summary

**Install Method:** How did you install the CLI? (e.g. pip, interactive script, apt-get, Docker, MSI, nightly)  
Answer here: dev_setup.py

**CLI Version:** What version of the CLI and modules are installed? (Use `az --version`)  
Answer here: 2.0.2

**OS Version:** What OS and version are you using?  
Answer here: macOS Sierra

**Shell Type:** What shell are you using? (e.g. bash, cmd.exe, Bash on Windows)  
Answer here: bash, cmd.exe

**Python Version:** 2.7.11

---

### Description

Running the example in the example:
```
az keyvault certificate create --vault-name myvault -n cert1 -p "$(az keyvault certificate get-default-policy)"
```

results in the following error on bash (Python 2.7.11):
```
invalid syntax: "attributes": {
```

Alternate ways of accomplishing this (for example piping) generate the same error.

If I run this on Windows with cmd.exe it works in Python 3.  In Python 2 I get an error:
```
Property policy.x509_props has invalid value. Invalid X.500 distinguished name
Property policy.key_props has invalid value. Invalid Kty UlNB
Property policy.secret_props has invalid value. ContentType is invalid
```

jiasli:
Fix #2924: JSONC is not parsed successfully by json.loads in Python2

On Linux, when the output format is set to `jsonc` or `yamlc` whether by changing the `~/.azure/config` file or by using `--output`, there will be ANSI control sequences for coloring like `^[[94m` in `stdout`. If `stdout` is redirected, downstream commands like `jq` will fail:

```sh
$ az vm list --output jsonc | jq '.[0]'
parse error: Invalid numeric literal at line 3, column 6

$ az vm list --output jsonc > out.txt
$ cat out.txt --show-nonprinting
[
  {
    ^[[94m"additionalCapabilities"^[[39;49;00m: ^[[34mnull^[[39;49;00m,
    ^[[94m"availabilitySet"^[[39;49;00m: ^[[34mnull^[[39;49;00m,
```

This PR detects whether `stdout` is a tty. If not, use `json` or `yaml` instead, thus allowing the output to be piped to downstream commands:

```sh
$ az vm list --output jsonc | jq '.[].name'
"azureclitestlinux"
"azureclitestwin"

$ az vm list --output json > out.txt
$ cat out.txt --show-nonprinting
[
  {
    "additionalCapabilities": null,
    "availabilitySet": null,
```

ℹ This issue doesn't happen on Windows because ANSI control sequences are removed by colorama.

### DIFF
--- a/src/azure-cli-core/azure/cli/core/_output.py
+++ b/src/azure-cli-core/azure/cli/core/_output.py
@@ -6,6 +6,7 @@
 import knack.output
 import sys
 
+
 class AzOutputProducer(knack.output.OutputProducer):
     def __init__(self, cli_ctx=None):
         super(AzOutputProducer, self).__init__(cli_ctx)
--- a/src/azure-cli-core/azure/cli/core/_output.py
+++ b/src/azure-cli-core/azure/cli/core/_output.py
@@ -4,7 +4,7 @@
 # --------------------------------------------------------------------------------------------
 
 import knack.output
-
+import sys
 
 class AzOutputProducer(knack.output.OutputProducer):
     def __init__(self, cli_ctx=None):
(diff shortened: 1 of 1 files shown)


jiasli:
Closing as this should go to https://github.com/microsoft/knack

Issue #7692, opened 2018-10-29
Download all secrets in a key vault in on go


andreacassioli:
**Is your feature request related to a problem? Please describe.**
I would like to download all the secrets contained in a key vault in one go. Instead it seems the only way is to get them one at the time and only dumped in a file.

**Describe the solution you'd like**
a command to get all secrets in one go

andreacassioli:
Any thoughts about this?

tjprescott:
@schaabs for comment

andreacassioli:
Bump...

yonzhan:
add to S164.

yungezz:
need a design discussion.

yungezz:
PR ready, need confirm with PM. @bim-msft could you pls follow up on this? thanks

bim-msft:
> PR ready, need confirm with PM. @bim-msft could you pls follow up on this? thanks

Sure, will send mail to PM to further discuss this.

nihar:
Is there any update to this? Would love to have [12243](https://github.com/Azure/azure-cli/pull/12243) merged

evelyn-ys:
Sorry we had a discussion with keyvault service PM and they don't think it's safe and necessary. Please consider `az keyvault secret list` and then download one by one with simple script

PR #12243, opened 2020-02-18
[KeyVault] Add new command `az keyvault secret download-all` for downloading all secrets at once

bim-msft:
This is for supporting #7692 
Command usage: `az keyvault secret download-all --vault-name {kv} --dir/-d {local_dir} --encoding/-e {e} --maxresults {mr}`

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/keyvault/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/keyvault/_params.py
@@ -211,29 +211,51 @@ def load_arguments(self, _):
     # endregion
 
     # region KeyVault Secret
-    with self.argument_context('keyvault secret set') as c:
-        c.argument('content_type', options_list=['--description'], help='Description of the secret contents (e.g. password, connection string, etc)')
-        c.attributes_argument('secret', SecretAttributes, create=True)
+    for item in ['set']:
+        with self.argument_context('keyvault secret {}'.format(item)) as c:
+            c.argument('content_type', options_list=['--description'],
+                       help='Description of the secret contents (e.g. password, connection string, etc)')
+            c.attributes_argument('secret', SecretAttributes, create=True)
 
-    with self.argument_context('keyvault secret set', arg_group='Content Source') as c:
-        c.argument('value', options_list=['--value'], help="Plain text secret value. Cannot be used with '--file' or '--encoding'", required=False)
-        c.extra('file_path', options_list=['--file', '-f'], type=file_type, help="Source file for secret. Use in conjunction with '--encoding'", completer=FilesCompleter())
-        c.extra('encoding', arg_type=get_enum_type(secret_encoding_values, default='utf-8'), options_list=['--encoding', '-e'], help='Source file encoding. The value is saved as a tag (`file-encoding=<val>`) and used during download to automatically encode the resulting file.')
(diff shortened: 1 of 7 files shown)


src/azure-cli/azure/cli/command_modules/keyvault/_params.py
```diff
@@ -211,29 +211,51 @@ class CLIJsonWebKeyOperation(str, Enum):
...
-        c.argument('encoding', arg_type=get_enum_type(secret_encoding_values), options_list=['--encoding', '-e'], help="Encoding of the secret. By default, will look for the 'file-encoding' tag on the secret. Otherwise will assume 'utf-8'.", default=None)
-
-    for scope in ['backup', 'restore']:
-        with self.argument_context('keyvault secret {}'.format(scope)) as c:
-            c.argument('file_path', options_list=['--file', '-f'], type=file_type, completer=FilesCompleter(), help='File to receive the secret contents.')
-
-    for scope in ['list', 'list-deleted', 'list-versions']:
-        with self.argument_context('keyvault secret {}'.format(scope)) as c:
-            c.argument('maxresults', options_list=['--maxresults'], type=int)
+    for item in ['set']:
```
jiasli: Why is `set` extracted to an one-item array?

bim-msft: I

> Why is `set` extracted to an one-item array?

I intend to make the code pieces more consistent like the following template:
```
with item in {command_list}:
    with self.argument_context('keyvault secret {}'.format(item)) as c:
         ...
```

But if `len(command_list) == 1`, I'm not sure what is the best practice, do we have any guiding on this?

bim-msft:
Also, there is another way to implement this functionality as @yungezz mentioned:
```
az keyvault secret download --all
```
The user interface is more consistent with single key download.

But in this PR, I’m following the design principle of `show` and `list`, for almost all resources, `list` is equivalent to the functionality of `show --all`, but they are different commands as their parameters are so different, pack them into one command would increase the complexity of argument parsing process and the test. So just separate them into two different commands is fine.
For example:

```
az keyvault secret show
az keyvault secret list (we do not use "az keyvault secret show --all")
```

If you guys have any thought on this, we can discuss it in this thread.

jiasli:
For `--all`, I have no preference. Better to confirm with PM.

src/azure-cli/azure/cli/command_modules/keyvault/_params.py
jiasli: We don't have guideline on this. Using a loop is simply for eliminating duplication. For single item array, I prefer to unwrap it to improve the readability. But it's up to you. 😉

bim-msft: Thanks, I'd like to keep this until we have further strict guideline on this.

NithinPrasad-GEP:
Any update on this ? Is this in the roadmap to be released?

davildaran:
not trying to add scope-creep. I created [this Issue](https://github.com/Azure/azure-cli/issues/26321) prior to seeing this PR. Thoughts on including another optional param to select a subset of secrets?

evelyn-ys:
@jiasli Any context about why we have this PR and why we leave it unmerged?

Issue #11420, opened 2019-11-28
Update CLI specifications for Resources to be created/updated as part of Microsoft.Compute/ProximityPlacementGroup


rosingla:
**Resource Provider**
Microsoft.Compute

**Description of Feature or Work Requested**
This issue propose changes in cmdlet w.r.t. Proximity Placement Group to support following features.

- To update the Proximity Placement Group reference of a resource (VM/AvSet/VMSS)
- To show colocation status of a Proximity Placement Group.

More details alongwith the proposed changes in the corresponding powershell commands can be found at : https://github.com/Azure/azure-powershell-cmdlet-review-pr/issues/431#event-2826967563

**Minimum API Version Required**

Min versions for the below two items:
- To update the Proximity Placement Group reference of a resource (VM/AvSet/VMSS) **[2018-04-01]**
- To show colocation status of a Proximity Placement Group. **[2019-07-01]**

**Swagger Link**
Swagger changes are checked in through : 
Azure/azure-rest-api-specs#7745
https://github.com/Azure/azure-rest-api-specs/pull/7745/files

**Target Date**
Proximity Placement Group feature is already in GA

haroldrandom:
@qwordy Please take a look

qwordy:
> 
> 
> @qwordy Please take a look

OK

qwordy:
> To update the Proximity Placement Group reference of a resource (VM/AvSet/VMSS)

PR is merged.

> To show colocation status of a Proximity Placement Group.

Will implement in a separate PR

qwordy:
Azure CLI already has a parameter --include-colocation-status. However, Azure Python SDK is not up-to-date. I have requested an update. Once the new version of SDK is published, we can see colocation status in output. This feature will be available in next release of Azure CLI. Release date: 03/10/2020

PR #12248, opened 2020-02-18
[Compute] ppg show: add --colocation-status to Enable fetching the colocation status of all the resources in the proximity placement group

qwordy:
#11420 

**History Notes:**  

---


src/azure-cli/azure/cli/command_modules/vm/tests/latest/test_vm_commands.py
```diff
@@ -3468,6 +3468,8 @@ def test_proximity_placement_group(self, resource_group, resource_group_location
             self.check('location', '{loc}'),
         ])
 
+        self.cmd('ppg show -g {rg} -n {ppg1} --colocation-status')
```
jiasli: How about checking the returned JSON?

jiasli:
Verified the usage of SDK is correct: `Lib/site-packages/azure/mgmt/compute/v2019_07_01/operations/_proximity_placement_groups_operations.py:238`

```py
    def get(
            self, resource_group_name, proximity_placement_group_name, include_colocation_status=None, custom_headers=None, raw=False, **operation_config):
```

src/azure-cli/azure/cli/command_modules/vm/commands.py
```diff
@@ -464,7 +464,7 @@ def load_command_table(self, _):
         g.wait_command('wait')
 
     with self.command_group('ppg', compute_proximity_placement_groups_sdk, min_api='2018-04-01', client_factory=cf_proximity_placement_groups) as g:
-        g.command('show', 'get')
+        g.custom_command('show', 'show_proximity_placement_group')
```
MyronFanQiu: just one question. Should we use g.custom_command? Based on your implementation, I think we can still use command.

MyronFanQiu: well. I got your point. It's a required filed. Right? I think you can also set `required=False` to use it. But this one is fine.

MyronFanQiu: looks like if this filed is not required in SDK. I don't think you need change this one. You can just rewrite the c.argument.

jiasli: I agree. We don't need to use `custom_command`. Also we should use `show_command` and `custom_show_command` so that the [exit code](https://github.com/Azure/azure-cli#exit-codes) is correctly set to 3 on a 404.

arrownj:
In 2.1.0, there is already a --include-colocation-status parameter, why we change the name to --colocation-status ?

```
(Version 2.1.0)
PS C:\Users\xiaojxu.FAREAST> az ppg show -h

Command
    az ppg show : Get a proximity placement group.

Arguments
    --name -n           [Required] : The name of the proximity placement group.
    --resource-group -g [Required] : Name of resource group. You can configure the default group
                                     using `az configure --defaults group=<name>`.
    --include-colocation-status    : IncludeColocationStatus=true enables fetching the colocation
                                     status of all the resources in the proximity placement group.
```

Actually current problem is that colocation status is in the REST api response, however we didn't return the value to user.

```
(Version 2.1.0)
PS C:\Users\xiaojxu.FAREAST> az ppg show -g xiaojianxu -n xxj-ppg-1 --include-colocation-status true --debug
......................
... unrelated logs ...
......................
msrest.http_logger : Request URL: 'https://management.azure.com/subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/resourceGroups/xiaojianxu/providers/Microsoft.Compute/proximityPlacementGroups/xxj-ppg-1?includeColocationStatus=true&api-version=2019-07-01'
......................
... unrelated logs ...
......................
msrest.http_logger : {
  "name": "xxj-ppg-1",
  "id": "/subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/resourceGroups/xiaojianxu/providers/Microsoft.Compute/proximityPlacementGroups/xxj-ppg-1",
  "type": "Microsoft.Compute/proximityPlacementGroups",
  "location": "eastasia",
  "tags": {},
  "properties": {
    "proximityPlacementGroupType": "Standard",
    "virtualMachines": [],
    "virtualMachineScaleSets": [],
    "availabilitySets": [],
    "colocationStatus": {
      "code": "ColocationStatus/Aligned",
      "level": "Info",
      "displayStatus": "Aligned",
      "message": "All resources in the proximity placement group are aligned."
    }
  }
}
......................
... unrelated logs ...
......................
```

We can see that the current REST request already has a **includeColocationStatus=true** parameter and **colocationStatus** is already in the response body(properties sections).

After these code changes, I find that the problem still exists. ColocationStatus is still not returned to user. All the behaviors are same with before except we changed the parameter name.

@qwordy , so I think the main work here is how to transfer the value from properties to user.

src/azure-cli/azure/cli/command_modules/vm/commands.py
qwordy: @MyronFanQiu Agree. command() is enough. I didn't realize that there is such an argument. Then I added a new one.

qwordy: @jiasli Good advice. An interesting fact is that there is no show_command() is vm module's params. I'll try it.

qwordy:
@arrownj Your comment has deep insight. Thanks for finding the problem.

qwordy:
@arrownj This colocation field was add to swagger 2 months ago. The root cause is Python SDK is not up-to-date. CLI will not output response directly, instead, it creates an object using SDK class then output JSON. I need to find someone to publish a new version of SDK.

qwordy:
I have requested a Python SDK update.

src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -841,6 +841,9 @@ def load_arguments(self, _):
         c.argument('ppg_type', options_list=['--type', '-t'], help="The type of the proximity placement group. Allowed values: Standard.")
         c.argument('tags', tags_type)
 
+    with self.argument_context('ppg show', min_api='2019-07-01') as c:
+        c.argument('include_colocation_status', options_list=['--colocation-status'], arg_type=get_three_state_flag(), help='Enable fetching the colocation status of all the resources in the proximity placement group. Default is false.')
```
fengzhou-msft: Remove options_list to avoid breaking change? In addition, the name `include_colocation_status` seems more clear. I suggest you to use `action='store_true'` since include_colocation_status is not a property of the resource, it's more like a command switch.

jiasli: I agree. As `include_colocation_status` doesn't go to a property of the resource.

qwordy: Good idea. Keep the original name.
Right. It's not a property.
Fixed.

qwordy:
> 
> 
> We need to wait azure-mgmt-compute upgraded and update the version number in our side. Also it's better to add the check for colocation status property in test.

Done. Upgraded compute SDK to 11.0.0. And added checks in test.

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -2876,7 +2876,7 @@ def set_vmss_extension(cmd, resource_group_name, vmss_name, extension_name, publ
         extensions = extension_profile.extensions
         if extensions:
             extension_profile.extensions = [x for x in extensions if
-                                            x.type.lower() != extension_name.lower() or x.publisher.lower() != publisher.lower()]  # pylint: disable=line-too-long
+                                            x.type1.lower() != extension_name.lower() or x.publisher.lower() != publisher.lower()]  # pylint: disable=line-too-long
```
arrownj: due to some update of sdk ?

> 
> due to some update of sdk ?

Yes. I don't like this new name.

PR #12277, opened 2020-02-20
[Network] az network application-gateway rewrite-rule create: support url configuration

### LINKED ISSUE #12088: CLI support for URL Rewrite in Application Gateway
**Resource Provider**
What is the Azure resource provider your feature is part of?

NRP

URL Rewrite refers to that action by the Application Gateway that replaces the original URL path with a new URL path while it routes the incoming request to the backend. Using URL Rewrite, user can use the App Gateway to rewrite request URL path before it is processed by the backend pools. User can define URL path rewriting logic that includes regular expressions, capture groups and rules can be applied based on the request URL, HTTP headers, and server. User has an option to rewrite the URL path and use that URL path either to go to the route to same backend pool or reroute to the new path based rule based on rewritten URL path inside same listener. User can also change the host header but the updated host header.

This parameter will be valid only for Application Gateway Standard_v2 and WAF_v2 SKUs. The validations in NRP are in place for the SKU verification.

2019-09-01

**Swagger Link**
PR: https://github.com/Azure/azure-rest-api-specs/pull/7905/files
https://github.com/Azure/azure-rest-api-specs/blob/master/specification/network/resource-manager/Microsoft.Network/stable/2019-11-01/applicationGateway.json

**Target Date**
If you have a target date for release of this feature/work, please provide it. While we can't guarantee these dates,
it will help us prioritize your request against other requests.

End of Feb, 2020. The API is already out and we are working on PS and Portal, once done, we'll be targeting Public Preview.

MyronFanQiu:
Fix #12088

**History Notes:**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/network/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/network/_params.py
@@ -238,6 +238,11 @@ def load_arguments(self, _):
         c.argument('request_headers', nargs='+', help='Space-separated list of HEADER=VALUE pairs.', validator=get_header_configuration_validator('request_headers'), completer=get_sdk_completer('application_gateways', 'list_available_request_headers'))
         c.argument('sequence', type=int, help='Determines the execution order of the rule in the rule set.')
 
+    with self.argument_context('network application-gateway rewrite-rule', arg_group='URL Configuration') as c:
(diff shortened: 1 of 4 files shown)


haroldrandom:
Context?

MyronFanQiu:
> Context?

Fix #12088

Issue #12031, opened 2020-02-04
az deployment validate returns exit code 0 when an error is encountered in json


chaos-panda:
## Describe the bug
When you run az deployment validate command against your ARM templates locally or on a CI/CD returns exit code 0 even though you encountered an error and your json error response is not null.

**Command Name**
`az deployment validate`

## To Reproduce:
Steps to reproduce the behavior. Note that argument values have been redacted, as they may contain sensitive information.

- _Put any pre-requisite steps here..._
- `az deployment validate --location {} --subscription {} --template-file {} --parameters {} --parameters {} --parameters {}`

## Expected Behavior
Expected to have other than 0 exit code for non-valid templates so those can be captured locally or in CI/CD pipelines.

## Environment Summary
```
Linux-4.15.0-76-generic-x86_64-with-debian-buster-sid
Python 3.6.5
Shell: bash

azure-cli 2.0.81

Extensions:
aks-preview 0.4.11

```

arrownj:
Hi @zhoxing-ms , could you please help take a look at this ?

yonzhan:
add to S166.

PR #12286, opened 2020-02-20
[ARM] az deployment validate: Fix the exit code when the verification fails (#12031)

zhoxing-ms:
Issue: #12031

Because the SDK did not throw an exception when the status code of the service returned result is 400, the exit code is 0 when the verification fails.
I wrapped the error result returned by the verification as CLIError, so the exit code returned when the verification fails is 1.
The same error reason as #11842 occurs on several other commands.

**History Notes:**  

---


yonzhan:
add to S166.

zhoxing-ms:
This logic has been refactored and is now fixed in the new code.
PR: #10751

PR #12295, opened 2020-02-20
{doc} Import Directive from docutils.parsers.rst

bluca:
sphinx.util.compat is deprecated since 1.6 and removed in 1.7.

http://www.sphinx-doc.org/en/master/extdev/deprecated.html


yonzhan:
add to S166.

Issue #12299, opened 2020-02-21
Ability to specify lun while creating a SIG Image Version with snapshots


axayjo:
In the sample below, 
'sig image-version create -g {rg} --gallery-name {gallery} --gallery-image-definition {image} --gallery-image-version 1.0.0 --os-snapshot s1 --data-snapshots s2 s’
 
How do we specify LUN for data disks?

Code:
if data_snapshots:
    data_disk_images = []
    for i, s in enumerate(data_snapshots):
        data_disk_images.append(GalleryDataDiskImage(source=GalleryArtifactVersionSource(id=s), lun=i))
lun is set automatically from 0, 1, 2 ...

I can add a parameter to specify lun, like --data-snapshot-luns. It receives a list of luns, separated by space.

yungezz:
we're working on in this release.

qwordy:
This feature will be available on 03/10/2020. If any problem, you can comment here or open a new issue.

PR #12303, opened 2020-02-21
[Compute] sig image-version: add --data-snapshot-luns

qwordy:
#12299 

Add parameter
--data-snapshot-luns                     : LUNs (space-delimited) of data disk snapshots.

**History Notes:**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/vm/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_params.py
@@ -816,6 +816,7 @@ def load_arguments(self, _):
         c.argument('managed_image', help='image name(if in the same resource group) or resource id')
         c.argument('os_snapshot', help='Name or ID of OS disk snapshot')
         c.argument('data_snapshots', nargs='+', help='Names or IDs (space-delimited) of data disk snapshots')
+        c.argument('data_snapshot_luns', nargs='+', help='LUNs (space-delimited) of data disk snapshots')
         c.argument('exclude_from_latest', arg_type=get_three_state_flag(), help='The flag means that if it is set to true, people deploying VMs with version omitted will not use this version.')
         c.argument('version', help='image version')
         c.argument('end_of_life_date', help="the end of life date, e.g. '2020-12-31'")
--- a/src/azure-cli/azure/cli/command_modules/vm/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_params.py
@@ -816,7 +816,7 @@ def load_arguments(self, _):
         c.argument('managed_image', help='image name(if in the same resource group) or resource id')
         c.argument('os_snapshot', help='Name or ID of OS disk snapshot')
         c.argument('data_snapshots', nargs='+', help='Names or IDs (space-delimited) of data disk snapshots')
-        c.argument('data_snapshot_luns', nargs='+', help='LUNs (space-delimited) of data disk snapshots')
+        c.argument('data_snapshot_luns', nargs='+', help='Logical unit numbers (space-delimited) of data disk snapshots')
         c.argument('exclude_from_latest', arg_type=get_three_state_flag(), help='The flag means that if it is set to true, people deploying VMs with version omitted will not use this version.')
         c.argument('version', help='image version')
         c.argument('end_of_life_date', help="the end of life date, e.g. '2020-12-31'")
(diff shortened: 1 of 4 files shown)


src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -816,6 +816,7 @@ def load_arguments(self, _):
         c.argument('managed_image', help='image name(if in the same resource group) or resource id')
         c.argument('os_snapshot', help='Name or ID of OS disk snapshot')
         c.argument('data_snapshots', nargs='+', help='Names or IDs (space-delimited) of data disk snapshots')
+        c.argument('data_snapshot_luns', nargs='+', help='LUNs (space-delimited) of data disk snapshots')
```
jiasli: Maybe we can use the full name of "LUN" for readability, add more description and add the constraints that it should contain the same number of arguments as `--data-snapshots`.

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -3020,10 +3020,17 @@ def create_image_version(cmd, resource_group_name, gallery_name, gallery_image_n
             source = GalleryArtifactVersionSource(id=managed_image)
         if os_snapshot is not None:
             os_disk_image = GalleryOSDiskImage(source=GalleryArtifactVersionSource(id=os_snapshot))
+        if data_snapshot_luns and not data_snapshots:
```
yungezz: just curious, is there general way in CLI to force parameters combination?

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -3020,10 +3020,17 @@ def create_image_version(cmd, resource_group_name, gallery_name, gallery_image_n
...
+            raise CLIError('usage error: --data-snapshot-luns must be used together with --data-snapshots')
         if data_snapshots:
+            if data_snapshot_luns and len(data_snapshots) != len(data_snapshot_luns):
+                raise CLIError('usage error: Length of --data-snapshots and --data-snapshot-luns should be equal.')
+            if not data_snapshot_luns:
+                data_snapshot_luns = [i for i in range(len(data_snapshots))]
             data_disk_images = []
             for i, s in enumerate(data_snapshots):
-                data_disk_images.append(GalleryDataDiskImage(source=GalleryArtifactVersionSource(id=s), lun=i))
+                data_disk_images.append(GalleryDataDiskImage(source=GalleryArtifactVersionSource(id=s),
```
yungezz: what if data_snapshot_luns None?

src/azure-cli/azure/cli/command_modules/vm/_params.py
yungezz: to align with `data_snapshots` on meaning, should it be `data_snapshots_luns`?

> 
> Maybe we can use the full name of "LUN" for readability, add more description and add the constraints that it should contain the same number of arguments as `--data-snapshots`.

I will add full name of LUN in help.
There is a validation in custom function.

> 
> to align with `data_snapshots` on meaning, should it be `data_snapshots_luns`?

I treat "data-snapshot-lun" as an entity, then I add a "s" to represent plural form. I think both names are reasonable.

src/azure-cli/azure/cli/command_modules/vm/custom.py
qwordy: I don't think we have, otherwise, we don't need so many validation code. This a common code pattern in Azure CLI.

src/azure-cli/azure/cli/command_modules/vm/custom.py
qwordy: Assign luns 0, 1, 2, ..., n - 1
```
if not data_snapshot_luns:
  data_snapshot_luns = [i for i in range(len(data_snapshots))]
```

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -3020,10 +3020,17 @@ def create_image_version(cmd, resource_group_name, gallery_name, gallery_image_n
             source = GalleryArtifactVersionSource(id=managed_image)
         if os_snapshot is not None:
             os_disk_image = GalleryOSDiskImage(source=GalleryArtifactVersionSource(id=os_snapshot))
+        if data_snapshot_luns and not data_snapshots:
+            raise CLIError('usage error: --data-snapshot-luns must be used together with --data-snapshots')
         if data_snapshots:
+            if data_snapshot_luns and len(data_snapshots) != len(data_snapshot_luns):
+                raise CLIError('usage error: Length of --data-snapshots and --data-snapshot-luns should be equal.')
```
arrownj: we validate the length of luns here, do we need add validation for the value of each lun ? For example, do we need to check whether there are duplicate luns(1, 1, 1) or invliad character(1, a, b) ? Or we just rely on the service to do the validation of these cases ?

qwordy: You are right. They should be non-negative integers.

PR #12317, opened 2020-02-21
[AppService] functionapp: Updated container image configuration for Linux apps

### LINKED ISSUE #11257: [Azure Functions]Update Linux Dedicated and consumption create flow to set linuxFxVersion to runtime|runtimeVersion
cc @ankitkumarr 

After ANT87 deploys, please update the create flow for linux dedicated apps.

gzuber:
Resolves #11257 

**History Notes:**  
[AppService] az functionapp create: Updated the way linuxFxVersion and container image name were being set for linux function apps.

---

### DIFF
--- a/src/azure-cli/HISTORY.rst
+++ b/src/azure-cli/HISTORY.rst
@@ -4,9 +4,7 @@ Release History
 ===============
 
 **AppService**
-
 az webapp create : Fixing issue when running the command with --runtime 
-* functionapp: Updated container image configuration for Linux apps
 
 2.1.0
 ++++++
--- a/src/azure-cli/HISTORY.rst
+++ b/src/azure-cli/HISTORY.rst
@@ -4,7 +4,9 @@ Release History
 ===============
 
 **AppService**
+
 az webapp create : Fixing issue when running the command with --runtime 
+* functionapp: Updated container image configuration for Linux apps
 
 2.1.0
 ++++++
--- a/src/azure-cli/azure/cli/command_modules/appservice/_constants.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/_constants.py
@@ -39,39 +39,18 @@ RUNTIME_TO_DEFAULT_VERSION_FUNCTIONAPP = {
         'java': '8'
     }
 }
-# functions version -> runtime -> runtime version : container image
-RUNTIME_TO_IMAGE_FUNCTIONAPP = {
+# functions version -> runtime : runtime versions
+RUNTIME_TO_SUPPORTED_VERSIONS_FUNCTIONAPP = {
     '2': {
-        'node': {
-            '8': 'mcr.microsoft.com/azure-functions/node:2.0-node8-appservice',
-            '10': 'mcr.microsoft.com/azure-functions/node:2.0-node10-appservice'
-        },
-        'python': {
-            '3.6': 'mcr.microsoft.com/azure-functions/python:2.0-python3.6-appservice',
-            '3.7': 'mcr.microsoft.com/azure-functions/python:2.0-python3.7-appservice'
-        },
-        'dotnet': {
-            '2': 'mcr.microsoft.com/azure-functions/dotnet:2.0-appservice'
-        },
-        'java': {
-            '8': 'mcr.microsoft.com/azure-functions/java:2.0-java8-appservice'
-        }
+        'node': ['8', '10'],
+        'python': ['3.6', '3.7'],
+        'dotnet': ['2'],
+        'java': ['8']
     },
     '3': {
-        'node': {
-            '10': 'mcr.microsoft.com/azure-functions/node:3.0-node10-appservice',
-            '12': 'mcr.microsoft.com/azure-functions/node:3.0-node12-appservice'
-        },
-        'python': {
-            '3.6': 'mcr.microsoft.com/azure-functions/python:3.0-python3.6-appservice',
-            '3.7': 'mcr.microsoft.com/azure-functions/python:3.0-python3.7-appservice',
-            '3.8': 'mcr.microsoft.com/azure-functions/python:3.0-python3.8-appservice'
-        },
-        'dotnet': {
-            '3': 'mcr.microsoft.com/azure-functions/dotnet:3.0-appservice'
-        },
-        'java': {
-            '8': 'mcr.microsoft.com/azure-functions/java:3.0-java8-appservice'
-        }
+        'node': ['10', '12'],
+        'python': ['3.6', '3.7', '3.8'],
+        'dotnet': ['3'],
+        'java': ['8']
     }
 }
--- a/src/azure-cli/azure/cli/command_modules/appservice/_constants.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/_constants.py
@@ -18,14 +18,14 @@ PYTHON_VERSIONS = ['3.7', '3.6', '2.7']
 NETCORE_VERSIONS = ['1.0', '1.1', '2.1', '2.2']
 DOTNET_VERSIONS = ['3.5', '4.7']
 LINUX_SKU_DEFAULT = "P1V2"
-FUNCTIONS_VERSIONS_FUNCTIONAPP = ['2', '3']
+FUNCTIONS_VERSIONS = ['2', '3']
 # functions version : default node version
-NODE_VERSION_DEFAULT_FUNCTIONAPP = {
+FUNCTIONS_VERSION_TO_DEFAULT_NODE_VERSION = {
     '2': '~10',
     '3': '~12'
 }
 # functions version -> runtime : default runtime version
-RUNTIME_TO_DEFAULT_VERSION_FUNCTIONAPP = {
+FUNCTIONS_VERSION_TO_DEFAULT_RUNTIME_VERSION = {
     '2': {
         'node': '8',
         'dotnet': '2',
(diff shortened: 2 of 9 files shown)


gzuber:
@Hazhzeng @ankitkumarr Can you take a look when you have time?

@pragnagopa FYI

src/azure-cli/HISTORY.rst
```diff
@@ -4,7 +4,9 @@ Release History
 ===============
 
 **AppService**
+
 az webapp create : Fixing issue when running the command with --runtime 
+* functionapp: Updated container image configuration for Linux apps
```
yungezz: same comments as in another PR, history.rst is automatically generated from PR titles. Pls refine PR titles and no need to change this file.

src/azure-cli/azure/cli/command_modules/appservice/_constants.py
```diff
@@ -39,39 +39,18 @@
         'java': '8'
     }
 }
-# functions version -> runtime -> runtime version : container image
-RUNTIME_TO_IMAGE_FUNCTIONAPP = {
+# functions version -> runtime : runtime versions
+RUNTIME_TO_SUPPORTED_VERSIONS_FUNCTIONAPP = {
```
yungezz: just to confirm, container image supporting is replaced with supported version of different runtime?

yungezz:
add to S166. Hi @qwordy  could you pls review this PR? thanks.

src/azure-cli/azure/cli/command_modules/appservice/_constants.py
```diff
@@ -39,39 +39,18 @@
...
     '2': {
-        'node': {
-            '8': 'mcr.microsoft.com/azure-functions/node:2.0-node8-appservice',
```
qwordy: Will the values like 'mcr.microsoft.com/azure-functions/node:2.0-node8-appservice' not be used any more?

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2420,13 +2420,10 @@ def create_function(cmd, resource_group_name, name, storage_account, plan=None,
             else:
                 site_config.app_settings.append(NameValuePair(name='WEBSITES_ENABLE_APP_SERVICE_STORAGE',
                                                               value='true'))
-                if runtime not in RUNTIME_TO_IMAGE_FUNCTIONAPP[functions_version].keys():
+                if runtime not in RUNTIME_TO_SUPPORTED_VERSIONS_FUNCTIONAPP[functions_version]:
                     raise CLIError("An appropriate linux image for runtime:'{}' was not found".format(runtime))
```
pragnagopa: Include runtime version in the error message

src/azure-cli/azure/cli/command_modules/appservice/_constants.py
```diff
@@ -39,39 +39,18 @@
...
-        },
-        'dotnet': {
-            '3': 'mcr.microsoft.com/azure-functions/dotnet:3.0-appservice'
-        },
-        'java': {
-            '8': 'mcr.microsoft.com/azure-functions/java:3.0-java8-appservice'
-        }
+        'node': ['10', '12'],
+        'python': ['3.6', '3.7', '3.8'],
+        'dotnet': ['3'],
```
pragnagopa: I do not think dotnet needs to be included here as Functions_Extension_Version App setting determines the which functions runtime version is used by the app. @ankitkumarr - can you confirm?

gzuber: That's a good point. My reasoning for leaving it here was that if a customer attempted to explicitly make a v2 function app with dotnet 3 (or something like that), they would get a clear error message as to why that fails.

gzuber: Per our discussion offline I've created this tracking issue: https://github.com/Azure/Azure-Functions/issues/1496 for these changes

src/azure-cli/azure/cli/command_modules/appservice/_constants.py
gzuber: Correct -- the back end has been updated to know which container image to use, so it's not something we have to set here anymore

src/azure-cli/azure/cli/command_modules/appservice/_constants.py
gzuber: Yeah the only difference is we don't need to specify the container image anymore so I renamed this because "Image" no longer applies. We still need to track which `runtime_version`'s we support for which runtime and functions version.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
gzuber: So if the user specifies a `runtime_version`, an error message will be emitted here: https://github.com/Azure/azure-cli/blob/dd691a3bff8ffa5230871ecb3ca7bbc216cdbf82/src/azure-cli/azure/cli/command_modules/appservice/custom.py#L2399-L2401

I added `functions_version` to this error message to make it more clear

qwordy:
@gzuber PR ready to merge？

src/azure-cli/azure/cli/command_modules/appservice/_constants.py
Francisco-Gamino: Hi @gzuber, to keep the language consistent, would it make sense to rename this to something like FUNCTIONSVERSION_TO_SUPPORTED_RUNTIMEVERSION given that you take into consideration the `Host version (function version)` ?

src/azure-cli/azure/cli/command_modules/appservice/_constants.py
Francisco-Gamino: @gzuber, you are missing `PowerShell` version `6`. If this lookup table is only use for `Linux`, do you have a separate table for `Windows`?

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2394,7 +2394,7 @@ def create_function(cmd, resource_group_name, name, storage_account, plan=None,
     if runtime_version is not None:
         if runtime is None:
             raise CLIError('Must specify --runtime to use --runtime-version')
-        allowed_versions = RUNTIME_TO_IMAGE_FUNCTIONAPP[functions_version][runtime].keys()
+        allowed_versions = RUNTIME_TO_SUPPORTED_VERSIONS_FUNCTIONAPP[functions_version][runtime]
         if runtime_version not in allowed_versions:
```
Francisco-Gamino: Do you do any validate for --runtime-version for `Windows` function apps?

gzuber: Apologies if I misunderstood the question -- the `if` block before this one will check that the `runtime` they've selected is valid for the OS, and this block will check that the `runtime_version` is valid for the `runtime` regardless of the OS. As far as I understand (at least to this point) supported `runtime_version`'s don't change across OS's. They're only different across `functions_version`'s. So as long as the `runtime` is supported for the OS, the list of supported `runtime_version`'s should be the same.

src/azure-cli/azure/cli/command_modules/appservice/_constants.py
ankitkumarr: @pragnagopa, yes that's true that dotnet doesn't need to be here. @gzuber, yes throwing an error like that makes sense. I wonder if we should just throw an error if `--runtime-version` is set to any value for `--runtime dotnet`. Although that would be a breaking change and is not necessary.
So, I think this looks good, except maybe when setting linuxFxVersion for dotnet, I would just set `dotnet` to keep it consistent with other UX tools.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2494,13 +2492,10 @@ def _get_extension_version_functionapp(functions_version):
...
 
-def _get_linux_fx_functionapp(is_consumption, functions_version, runtime, runtime_version):
+def _get_linux_fx_functionapp(functions_version, runtime, runtime_version):
     if runtime_version is None:
         runtime_version = RUNTIME_TO_DEFAULT_VERSION_FUNCTIONAPP[functions_version][runtime]
-    if is_consumption:
-        return '{}|{}'.format(runtime.upper(), runtime_version)
-    # App service or Elastic Premium
-    return _format_fx_version(RUNTIME_TO_IMAGE_FUNCTIONAPP[functions_version][runtime][runtime_version])
+    return '{}|{}'.format(runtime.upper(), runtime_version)
```
ankitkumarr: For `dotnet`, it would be nice to set it to `DOTNET` and not `DOTNET|2` or `DOTNET|3`.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
Francisco-Gamino: Thanks for your reply. There are some runtime versions that are only supported on FunctionsVersion 2, and not in 3 and vice versa. For the cmdlets, I will validate against a static json file that @pragnagopa is working on. It would probably make sense that all clients (`az cli`, `Portal`, and `Az.Functions cmdlets`) do the same. However, that does not apply in this PR. This is more like a future work item.

gzuber: Right so this will do the validation to make sure the `runtime_version` is valid for the `functions_version`. And yeah using the json files pragna is working on is definitely planned for the cli

src/azure-cli/azure/cli/command_modules/appservice/_constants.py
gzuber: Yeah that's a good point, I didn't need to hold on to the old names. I've changed a lot of the constants to better reflect what they are.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
gzuber: Ok that makes sense. I've made that change because it's directly related to how we're handling linuxFxVersion. There's also https://github.com/Azure/Azure-Functions/issues/1496 for further dotnet related changes.

gzuber:
@qwordy This should be good to merge, thanks!

PR #12323, opened 2020-02-23
{Packaging/windows pip} Use local python for az.bat

### LINKED ISSUE #7372: Incorrect az.bat when installed through pip on Windows
**Describe the bug**
The az.bat (https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/az.bat) that you get when you "pip install azure-cli" is incorrect,. The correct one is: https://github.com/Azure/azure-cli/blob/dev/build_scripts/windows/scripts/az.cmd

**To Reproduce**
Install Azure CLI on Windows through "pip install azure-cli"

**Expected behavior**
Remove az.bat and include the az.cmd in the **Scripts** directory when installing through pip. 

**Environment summary**
Windows. My test was on a Server 2016 on a command prompt. 

**Additional context**
The impact of this issue is that the AZ CLI is not running in the correct Python.exe where the azure cli libraries are installed. Our servers have multiple Python instances and the the one activated / default Python environment may not be the one where we pip installed the azure-cli but we still want to run the "az" command to manage Azure resource from the command line . Our workaround is to manually remove az.bat and copy the az.cmd in the "Scripts" directory. That seems to work fine for us though I want to understand if there is something that may still break. Also looks like the Azure CLI MSI installer correctly included the az.cmd.

fengzhou-msft:
Fix #7372

---

### DIFF
--- a/src/azure-cli/az.bat
+++ b/src/azure-cli/az.bat
@@ -1,9 +1,10 @@
 @echo off
 setlocal
 
+SET PYTHONPATH=%~dp0\src;%PYTHONPATH%
+
 IF EXIST "%~dp0\python.exe" (
   "%~dp0\python.exe" -m azure.cli %*
 ) ELSE (
-  SET PYTHONPATH=%~dp0/src;%PYTHONPATH%
   python -m azure.cli %*
 )
--- a/src/azure-cli/az.bat
+++ b/src/azure-cli/az.bat
@@ -1,5 +1,9 @@
 @echo off
 setlocal
 
-SET PYTHONPATH=%~dp0/src;%PYTHONPATH%
-python -m azure.cli %*
+IF EXIST "%~dp0\python.exe" (
+  "%~dp0\python.exe" -m azure.cli %*
+) ELSE (
+  SET PYTHONPATH=%~dp0/src;%PYTHONPATH%
+  python -m azure.cli %*
+)


fengzhou-msft:
Planned in S166.

src/azure-cli/az.bat
```diff
@@ -1,5 +1,9 @@
 @echo off
 setlocal
 
-SET PYTHONPATH=%~dp0/src;%PYTHONPATH%
-python -m azure.cli %*
+IF EXIST "%~dp0\python.exe" (
+  "%~dp0\python.exe" -m azure.cli %*
+) ELSE (
+  SET PYTHONPATH=%~dp0/src;%PYTHONPATH%
```
jiasli: I think `PYTHONPATH` is not necessary anymore. By showing it with `echo %PYTHONPATH%`:
```
D:\cli\env38\Scripts\/src;
```
This is not a valid path.

jiasli:
The script itself works fine. We can verify by running

```
D:\cli\env38\Scripts\az.bat -v
```

It shows:

```
Python location 'D:\cli\env38\Scripts\python.exe'
``` 

By renaming `D:\cli\env38\Scripts\python.exe` to something else, it show:

```
Python location 'C:\Users\jiasli\AppData\Local\Programs\Python\Python38\python.exe'
```

src/azure-cli/az.bat
jiasli: Also path separator `/` is not valid on Windows.

fengzhou-msft: I will fix the path separator. 

PYTHONPATH is not needed for normal cli run, but let's leave it there for the convenience of using additional custom Python libraries that users do not want to install in the global default location (i.e., the site-packages directory).

fengzhou-msft: Linux counterpart script also has this setting: https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/az#L6..L14

jiasli: Not sure if there is a historical reason. As far as I can find, the only possible operation for `src` to exist is installing via [`pip install -e <VCS>`](https://pip.pypa.io/en/stable/reference/pip_install/#vcs-support): `pip install -e "git+https://github.com/Azure/azure-sdk-for-python/#egg=azure-mgmt-eventgrid&subdirectory=sdk\eventgrid\azure-mgmt-eventgrid"`

Even in that case `src` should be under `%~dp0\..`. The path to `azure-mgmt-eventgrid` is added to `sys.path` via `env\Lib\site-packages\azure-mgmt-eventgrid.egg-link`.

fengzhou-msft: I agree `%~dp0\..` is a better location to put `src` as `%~dp0` is the `Scripts` directory (or `bin` directory on Linux). I would consider changing it together with the Linux script when we have more knowledge about the real uses cases and history for this `PYTHONPATH` setting.

PR #12333, opened 2020-02-24
[Storage] az storage account file-service-properties update/show: Add File Properties Support for Storage Account

Juliehzl:
**History Notes:**  

---


yonzhan:
add to S166.

src/azure-cli/azure/cli/command_modules/storage/_validators.py
```diff
@@ -1146,16 +1146,3 @@ def validator_delete_retention_days(namespace):
     if namespace.enable_delete_retention is False and namespace.delete_retention_days is not None:
         raise ValueError(
             "incorrect usage: '--delete-retention-days' is invalid when '--enable-delete-retention' is set to false")
-
-    if namespace.enable_delete_retention is None and namespace.delete_retention_days is not None:
-        raise ValueError(
-            "incorrect usage: please specify '--enable-delete-retention true' if you want to set the value for "
-            "'--delete-retention-days'")
```
zhoxing-ms: Removing the validation of the common call here affects the logic of `update_blob_service_properties()` in this parameter scenario.

src/azure-cli/azure/cli/command_modules/storage/_validators.py
```diff
@@ -1146,16 +1146,3 @@ def validator_delete_retention_days(namespace):
...
-
-    if namespace.delete_retention_days or namespace.delete_retention_days == 0:
-        if namespace.delete_retention_days < 1:
-            raise ValueError(
-                "incorrect usage: '--delete-retention-days' must be greater than or equal to 1")
-        if namespace.delete_retention_days > 365:
-            raise ValueError(
-                "incorrect usage: '--delete-retention-days' must be less than or equal to 365")
```
zhoxing-ms: The reason why this verification was added here is that the verification of SDK is not correct, and the range of verification prompt is [2, 364].            
If you want to remove this verification logic, please make sure that they have fixed the error range prompt.

The previous errors of SDK were as follows:
> validation error: Parameter 'DeleteRetentionPolicy.days' must be less than 365.  
   validation error: Parameter 'DeleteRetentionPolicy.days' must be greater than 1.

src/azure-cli/azure/cli/command_modules/storage/operations/account.py
```diff
@@ -363,3 +360,27 @@ def update_blob_service_properties(cmd, instance, enable_change_feed=None, enabl
...
+    # If already enabled, only update days
+    if enable_delete_retention is None and delete_retention_days is not None:
+        delete_retention_policy = client.get_service_properties(
+            resource_group_name=resource_group_name,
+            account_name=account_name).share_delete_retention_policy
+        if delete_retention_policy is not None and delete_retention_policy.enabled:
+            delete_retention_policy.days = delete_retention_days
+        else:
+            raise CLIError("Delete Retention Policy hasn't been enabled, and you cannot set delete retention days. "
+                           "Please set --enabled-delete-retention as true to enable Delete Retention Policy.")
```
zhoxing-ms: In such a parameter scenario, the `storage account blob-service-properties update` command will report an error directly.
Should we consider keeping the user experience consistent?

src/azure-cli/azure/cli/command_modules/storage/_validators.py
Juliehzl: I think it will be better to make server side give days range check. Otherwise, if server changes the days range, we will block users and currently service error message is good enough to me.
But if you have any concern, feel free to let me know.

src/azure-cli/azure/cli/command_modules/storage/_validators.py
Juliehzl: Yes, I know.  I removed this part because I think we should allow users to update days when delete retention policy is already enabled.
But I think I also need some logic to check inside update_blob_service_properties(). What do you think of such design?

src/azure-cli/azure/cli/command_modules/storage/operations/account.py
qianwens: This is a question, can we enable delete retention as long as delete_retention_days is set? 

---
In reply to: [383229549](https://github.com/Azure/azure-cli/pull/12333#discussion_r383229549) [](ancestors = 383229549)

src/azure-cli/azure/cli/command_modules/storage/_validators.py
zhoxing-ms: Because the error message of the SDK validation is not correct, so we may need to verify it in advance to provide the correct error message.
It is a problem about msrest in essence, and I suspect they are unlikely to fix it in a short time:
https://github.com/Azure/msrest-for-python/issues/191
If they can fix this problem, I think it's ok to remove this check.

src/azure-cli/azure/cli/command_modules/storage/_validators.py
zhoxing-ms: I think it's ok, or maybe we can change the update_blob_service_properties logic to be consistent with the file-service-properties.

src/azure-cli/azure/cli/command_modules/storage/operations/account.py
Juliehzl: sure

PR #12336, opened 2020-02-24
[ACR] Add new command `az acr helm install-cli`

### LINKED ISSUE #8974: az acr helm install-cli
It bad to relay on external installations of helm. There is no official docker image with helm pre installed

An easy way to install helm cli e.g. `az acr helm install-cli` which is similar to `az aks install-cli`

Wwwsylvia:
`az acr helm install-cli`:  To download and install Helm command-line tool
Close #8974

**History Notes:**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acr/helm.py
+++ b/src/azure-cli/azure/cli/command_modules/acr/helm.py
@@ -51,7 +51,7 @@ def acr_helm_list(cmd,
 def acr_helm_show(cmd,
                   registry_name,
                   chart,
-                  client_version=None,
+                  version=None,
                   repository='repo',
                   resource_group_name=None,  # pylint: disable=unused-argument
                   tenant_suffix=None,
@@ -69,7 +69,7 @@ def acr_helm_show(cmd,
     return request_data_from_registry(
         http_method='get',
         login_server=login_server,
-        path=_get_charts_path(repository, chart, client_version),
+        path=_get_charts_path(repository, chart, version),
         username=username,
         password=password)[0]
 
@@ -77,7 +77,7 @@ def acr_helm_show(cmd,
 def acr_helm_delete(cmd,
                     registry_name,
                     chart,
-                    client_version=None,
+                    version=None,
                     repository='repo',
                     resource_group_name=None,  # pylint: disable=unused-argument
                     tenant_suffix=None,
@@ -85,9 +85,9 @@ def acr_helm_delete(cmd,
                     password=None,
                     prov=False,
                     yes=False):
-    if client_version:
+    if version:
         message = "This operation will delete the chart package '{}'".format(
-            _get_chart_package_name(chart, client_version, prov))
+            _get_chart_package_name(chart, version, prov))
     else:
         message = "This operation will delete all versions of the chart '{}'".format(chart)
     user_confirmation("{}.\nAre you sure you want to continue?".format(message), yes)
@@ -104,7 +104,7 @@ def acr_helm_delete(cmd,
     return request_data_from_registry(
         http_method='delete',
         login_server=login_server,
-        path=_get_blobs_path(repository, chart, client_version, prov) if client_version else _get_charts_path(repository, chart),
+        path=_get_blobs_path(repository, chart, version, prov) if version else _get_charts_path(repository, chart),
         username=username,
         password=password)[0]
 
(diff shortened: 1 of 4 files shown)


yungezz:
hi @fengzhou-msft @qwordy could you pls help to look at this PR? thanks.

qwordy:
What's the advantage of this command compared to https://github.com/helm/helm#install?

yugangw-msft:
> What's the advantage of this command compared to https://github.com/helm/helm#install?
@qwordy, the goal is to install the right version (2.x instead of 3.x) from the right place, which has not been easy for users to figure out

qwordy:
> 
> 
> ```
>  > What's the advantage of this command compared to https://github.com/helm/helm#install?
> ```
> 
> @qwordy, the goal is to install the right version (2.x instead of 3.x) from the right place, which has not been easy for users to figure out

Thanks for your answer. Helm 2's page says "You are viewing Helm 2 (legacy).   Helm 3 is here.". Helm 2 is not recommended. Maybe it's time to support Helm 3. Just like Azure CLI has abandoned Python 2.

Wwwsylvia:
> > ```
> >  > What's the advantage of this command compared to https://github.com/helm/helm#install?
> > ```
> > 
> > 
> > @qwordy, the goal is to install the right version (2.x instead of 3.x) from the right place, which has not been easy for users to figure out
> 
> Thanks for your answer. Helm 2's page says "You are viewing Helm 2 (legacy). Helm 3 is here.". Helm 2 is not recommended. Maybe it's time to support Helm 3. Just like Azure CLI has abandoned Python 2.

I think the idea of command is to install the current supported Helm CLI for users. It's v2 at current stage, and will be v3 after other commands support v3. @yugangw-msft  Am I correct?

src/azure-cli/azure/cli/command_modules/acr/helm.py
```diff
@@ -175,6 +182,81 @@ def acr_helm_repo_add(cmd,
     p.wait()
 
 
+def acr_helm_install_cli(client_version='2.16.3', install_location=None, yes=False):
+    """Install Helm command-line tool."""
+
+    if client_version >= '3':
+        raise CLIError('Helm v3 is not supported yet.')
```
yugangw-msft: Please warn people that it is not we don't support helm3, but just can't get along with "az acr helm" commands. 
Maybe output a warning: 
```
Please note helm3 doesn't work with "az acr helm" commands, and but you can still push helm chat using a different command flow to ACR. For more information, please check out  https://docs.microsoft.com/en-us/azure/container-registry/container-registry-helm-repos

File: src/azure-cli/azure/cli/command_modules/acr/helm.py [hunk 2]
```diff
@@ -175,6 +182,81 @@ def acr_helm_repo_add(cmd,
...
+
+    if not install_location or install_location.isspace():
+        raise CLIError('Invalid install location.')
+
+    install_dir, cli = os.path.dirname(install_location), os.path.basename(install_location)
+    if not cli:
+        raise CLIError("{} is a directory. Please specify a complete file path.".format(install_location))
```
yugangw-msft: can we default to "helm" if it is an folder?

src/azure-cli/azure/cli/command_modules/acr/helm.py
```diff
@@ -175,6 +182,81 @@ def acr_helm_repo_add(cmd,
...
+                # Rename helm to specified name
+                target_path = install_location if os.path.splitext(f.name)[0] == 'helm' \
+                    else os.path.join(install_dir, f.name)
+                logger.debug('Moving %s to %s', f.path, target_path)
+                shutil.move(f.path, target_path)
+
+                if os.path.splitext(f.name)[0] in ('helm', 'tiller'):
+                    os.chmod(target_path, os.stat(target_path).st_mode | stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH)
+    except IOError as e:
+        raise CLIError('Error while installing {} to {}: {}'.format(cli, install_dir, e))
```
yugangw-msft: will this cause the original stack trace to lost? Otherwise I suggest we catch a particular line which might throw this exception, or just start with uncaught.

Wwwsylvia: We can log the stack trace so that it won't lose and users won't get disturbed

Wwwsylvia:
> LGTM. Just one comment, in the case of helm already installed when running this command again, the helm package downloaded in the second run will silently overwrite the first one. You may want to tell users about this behavior in help.

Thanks for the advice! Updated.

PR #12341, opened 2020-02-24
{Container monitoring} - Add case insensitive string compare for msi string

rashmichandrashekar:
Adding case insensitive string comparison for msi string


yonzhan:
add to S166.

Issue #9177, opened 2019-04-23
Default name for policy assignment exceeds '24' characters.


murarisumit:
**Describe the bug**
Default name for policy assignment  exceed '24' characters, which results in error.

**To Reproduce**
* Try creating any policy.
* Creating policy assignment without `name` flag 
  ```
  az policy assignment create \
                --scope "/providers/Microsoft.Management/managementGroups/test-group" \
                --policy "/providers/Microsoft.Management/managementgroups/test-group/providers/Microsoft.Authorization/policyDefinitions/<policy-name>"

  ```

Steps to reproduce the behavior.

**Expected behavior**
I should be able to create assignment of policy at given management group with some system generated name.

**Environment summary**
Running it in bash with az cli <2.0.60> in mac mojave.

yonzhan:
@Juliehzl, add to S162

PR #12352, opened 2020-02-26
[Policy] az policy assignment create: Fix the bug that automatically generated name of policy assignment exceeds the limit (#9177)

zhoxing-ms:
Issue: #9177

When the user does not pass in the policy assignment name, we will automatically generate a random name that does not repeat. 
The previous solution was to use UUID, but the string generated by using Python's `uuid.uuid4()` is 36 characters, which exceeds the limit of 24 characters required by server. So I used Base64 to compress the UUID to generate a unique random name with a length of 22.

**History Notes:**  

---


yonzhan:
add to S166.

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -1356,7 +1357,10 @@ def create_policy_assignment(cmd, policy=None, policy_set_definition=None,
             identity = _build_identities_info(cmd, assign_identity)
         assignment.identity = identity
 
-    createdAssignment = policy_client.policy_assignments.create(scope, name or uuid.uuid4(), assignment)
+    if name is None:
+        name = (base64.urlsafe_b64encode(uuid.uuid4().bytes).decode())[:-2]
```
qianwens: >uuid4 [](start = 46, length = 5)

why not just cut the uuid?

zhoxing-ms: Because the name generated by using Python's uuid.uuid4() is 36 characters, which exceeds the limit of 24 characters required by server. So I used Base64 to compress the UUID to generate a unique random name with a length of 22.

yonzhan: [:-22]?

PR #12367, opened 2020-02-26
[Extensions] Allow to install extensions in the system path via packages

bluca:
On Linux, check if there are extension in the system Python
library path (eg: /usr/lib/python3/dist-packages/azure-cli-extensions).

This allows extensions to be installed via system packages in
dist-packages, and registered by simply creating an extension directory
and symlinking the egg-info and the module.

EG:

$ ls -l /usr/lib/python3/dist-packages/azure-cli-extensions/azure-devops/
total 0
lrwxrwxrwx 1 root root 18 Feb 23 12:31 azext_devops -> ../../azext_devops
lrwxrwxrwx 1 root root 34 Feb 23 12:31 azure_devops-0.17.0.egg-info -> ../../azure_devops-0.17.0.egg-info
$ az extension list
[
  {
    "extensionType": "whl",
    "name": "azure-devops",
    "version": "0.17.0"
  }
]

Locally installed extensions still have higher priority if available.

### DIFF
--- a/src/azure-cli-core/azure/cli/core/extension/__init__.py
+++ b/src/azure-cli-core/azure/cli/core/extension/__init__.py
@@ -7,9 +7,11 @@ import os
 import traceback
 import json
 import re
+import sys
+import pkginfo
 
 from azure.cli.core._config import GLOBAL_CONFIG_DIR, ENV_VAR_PREFIX
-
+from distutils.sysconfig import get_python_lib
 from knack.config import CLIConfig
 from knack.log import get_logger
 
@@ -19,6 +21,7 @@ _DEV_EXTENSION_SOURCES = az_config.get('extension', 'dev_sources', None)
 EXTENSIONS_DIR = os.path.expanduser(_CUSTOM_EXT_DIR) if _CUSTOM_EXT_DIR else os.path.join(GLOBAL_CONFIG_DIR,
                                                                                           'cliextensions')
 DEV_EXTENSION_SOURCES = _DEV_EXTENSION_SOURCES.split(',') if _DEV_EXTENSION_SOURCES else []
+EXTENSIONS_SYS_DIR = os.path.join(get_python_lib(), 'azure-cli-extensions') if sys.platform.startswith('linux') else ""
 
 EXTENSIONS_MOD_PREFIX = 'azext_'
 
@@ -33,7 +36,7 @@ EXT_METADATA_ISPREVIEW = 'azext.isPreview'
 WHEEL_INFO_RE = re.compile(
     r"""^(?P<namever>(?P<name>.+?)(-(?P<ver>\d.+?))?)
     ((-(?P<build>\d.*?))?-(?P<pyver>.+?)-(?P<abi>.+?)-(?P<plat>.+?)
-    \.whl|\.dist-info)$""",
+    \.whl|\.dist-info|\.egg-info)$""",
     re.VERBOSE).match
 
 logger = get_logger(__name__)
@@ -136,8 +139,13 @@ class WheelExtension(Extension):
                 if os.path.isfile(whl_metadata_filepath):
                     with open(whl_metadata_filepath) as f:
                         metadata.update(json.loads(f.read()))
+                elif os.path.isfile(os.path.join(dist_info_dirname, 'PKG-INFO')):
+                    metadata.update(pkginfo.Develop(dist_info_dirname).__dict__)
 
         return metadata
+
+    def __eq__(self, other):
+        return other.name == self.name
 
     @staticmethod
     def get_azext_metadata(ext_dir):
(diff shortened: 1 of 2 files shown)


yungezz:
HI @fengzhou-msft could you pls help to have a look? thanks.

haroldrandom:
Why linux only?

src/azure-cli-core/azure/cli/core/extension/__init__.py
```diff
@@ -7,9 +7,11 @@
 import traceback
 import json
 import re
+import sys
+import pkginfo
 
 from azure.cli.core._config import GLOBAL_CONFIG_DIR, ENV_VAR_PREFIX
-
+from distutils.sysconfig import get_python_lib
```
haroldrandom: What if there is extension containing C code (non-pure Python), will the get_python_lib() work?

bluca: On my Debian box cpython libraries are in dist-packages too, so I'm not expecting issues:

```
>>> from distutils.sysconfig import get_python_lib
>>> get_python_lib()
'/usr/lib/python3/dist-packages'
```
```
$ find /usr/lib/python3/dist-packages/ -name *.so -type f | wc -l
241
```

Also, what we are after is the azure-cli-extensions directory where we store the symlinks - the symlinks themselves can point to anywhere

bluca:
> Why Linux only? As far as I know, windows should also work for this.

Because I can only test it on Linux :-) If someone can confirm it works on Windows too, and we can establish a canonical installation path, then we can add it in the future. OSX too.

fengzhou-msft:
@bluca While this PR adds a useful feature to use extensions in the system path, I plan to make the system directory configurable just like the user extension directory. And it's also useful to add an option in `az extension add` to install extensions in the system directory. I will work on it in another PR.

bluca:
> @bluca While this PR adds a useful feature to use extensions in the system path, I plan to make the system directory configurable just like the user extension directory. And it's also useful to add an option in `az extension add` to install extensions in the system directory. I will work on it in another PR.

That's fine, as long as the default works as-is without any additional configuration - otherwise it's not useful

bluca:
> > @bluca While this PR adds a useful feature to use extensions in the system path, I plan to make the system directory configurable just like the user extension directory. And it's also useful to add an option in `az extension add` to install extensions in the system directory. I will work on it in another PR.
> 
> That's fine, as long as the default works as-is without any additional configuration - otherwise it's not useful

@fengzhou-msft can we start by merging this and then doing additional work on top?

fengzhou-msft:
> > > @bluca While this PR adds a useful feature to use extensions in the system path, I plan to make the system directory configurable just like the user extension directory. And it's also useful to add an option in `az extension add` to install extensions in the system directory. I will work on it in another PR.
> > 
> > 
> > That's fine, as long as the default works as-is without any additional configuration - otherwise it's not useful
> 
> @fengzhou-msft can we start by merging this and then doing additional work on top?

Yes. That's the plan.

bluca:
Great, thank you!

fengzhou-msft:
@bluca on Ubuntu, our package uses a wrapped python instead of the system python, and the python_lib path is `/opt/az/lib/python3.6/site-packages`, since you mentioned using symlinks so I think it would also work for you, but not as convenient? On CentOS/RHEL, we use the system python.

bluca:
I don't use your packages because of that. I've uploaded azure-cli to Debian and Ubuntu using system's python and dependencies.

https://packages.debian.org/sid/azure-cli
https://packages.ubuntu.com/focal/azure-cli

bluca:
Thanks!

jiasli:
@bluca, after we supported Python 3.10 (https://github.com/Azure/azure-cli/pull/20195), `distutils.sysconfig.get_python_lib` has now been deprecated in Python 3.10:

```py
>>> from distutils.sysconfig import get_python_lib
<stdin>:1: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
<stdin>:1: DeprecationWarning: The distutils.sysconfig module is deprecated, use sysconfig instead
```

According to https://bugs.python.org/issue41282#msg393021, it is not recommended/possible to get with `/usr/lib/python3/dist-packages` with the new `sysconfig.get_path`:

```py
>>> import distutils.sysconfig
>>> distutils.sysconfig.get_python_lib()
'/usr/lib/python3/dist-packages'
>>> import sysconfig
>>> sysconfig.get_paths()
{'stdlib': '/usr/lib/python3.8', 'platstdlib': '/usr/lib/python3.8', 'purelib': '/usr/lib/python3.8/site-packages', 'platlib': '/usr/lib/python3.8/site-packages', 'include': '/usr/include/python3.8', 'platinclude': '/usr/include/python3.8', 'scripts': '/usr/bin', 'data': '/usr'}
```

Do you have any idea for how we should fix this?

bluca:
I think it's still being discussed. As a temporary workaround you could do something like `sysconfig.get_paths()['platlib'].replace("site", "dist")` but it's just a warning for now, so I'd suggest to wait

jiasli:
It doesn't seem `distutils.sysconfig.get_python_lib()` and `sysconfig.get_paths()['platlib'].replace("site", "dist")` are the same:

```
>>> import distutils.sysconfig
>>> import sysconfig
>>> distutils.sysconfig.get_python_lib()
'/usr/lib/python3/dist-packages'
>>> sysconfig.get_paths()['platlib'].replace("site", "dist")
'/usr/lib/python3.8/dist-packages'
```

> but it's just a warning for now, so I'd suggest to wait

Now it is flooding our CI with `DeprecationWarning`. I would prefer to fix it, even with a hardcode/workaround.

bluca:
Right, it doesn't seem like Python added a way to get that path then - you can just hard-code it until they fix it, if you want to get rid of the warnings I guess

jiasli:
Thanks @bluca, this is helpful information. Could you help create a PR? It is difficult for me to test it. Appreciate it!

bluca:
Sure, this works for me: https://github.com/Azure/azure-cli/pull/20391

Issue #12319, opened 2020-02-22
VMSS automatic instance repair


yungezz:
**Resource Provider**
What is the Azure resource provider your feature is part of?
VMSS

**Description of Feature or Work Requested**
Support VMSS automatic instance repair. speclet
https://microsoft.sharepoint.com/:w:/t/AzureManagementExperience/Eat9Q3n2AX1LiWKRhkSABgMBYwUnpeo_KpA356xScHdzqg?e=kPpZzs

**Minimum API Version Required**
What is the minimum API version of your service required to implement your feature?

**Swagger Link**
will add later.

**Target Date**
3rd week of March

yungezz:
HI @arrownj, pls take this as high priority, thanks.

yungezz:
released in 2.2.0, close the issue now.

PR #12374, opened 2020-02-27
[Compute] az vmss create/update: support automatic repairs

arrownj:
This PR is adding support for automatic repairs, which was mentioned in this issue #12319 .

Please note that automatic repairs is still in preview, subscription must be added to the white list for testing.

There are 2 different scenarios if user want to enable automatic repairs for a vmss. A health probe or a health extension must provide. This PR add support for both these 2 scenarios.

**Health probe**

- Enable automatic repairs in az vmss create

`az vmss create -g xiaojianxu -n xxj-vmss-1 --load-balancer xxj-lb--1 --health-probe hp --automatic-repairs-grace-period 30`

> --load-balancer and --health-probe are also required if create vmss with --automatic-repairs-grace-period. The specified --load-balancer and --health-probe must exist, otherwise will report usage error.

- Enable automatic repairs in az vmss update

```
# create vmss
az vmss create -g xiaojianxu -n xxj-vmss-1 --image UbuntuLTS

# enable health probe for vmss (you must have a valid health probe first)
az vmss update -g xiaojianxu -n xxj-vmss-1 --set virtualMachineProfile.networkProfile.healthProbe.id=<Valid_Health_Probe_Id>

# list vmss instances to get the instance ids, these ids will be used in following command
az vmss list-instances -g xiaojianxu -n xxj-vmss-1 --query "[].instanceId"

# update instances, if not do this, will report error in following command (Error msg: Automatic repairs not supported for this Virtual Machine Scale Set because a health probe is not applied to all instances in the virtual machine scale set prior to switching to instance health repair mode.)
az vmss update-instances -g xiaojianxu -n xxj-vmss-1 --instance-ids 0 1

# enable automatic repairs 
az vmss update -g xiaojianxu -n xxj-vmss-1 --enable-automatic-repairs true --automatic-repairs-grace-period 30
```

**Health extension**

- Enable automatic repairs in az vmss create
Not support. We can only add health extension with a separate `az vmss extension` command.

- Enable automatic repairs in az vmss update

```
# create vmss
az vmss create -g xiaojianxu -n xxj-vmss-1 --image UbuntuLTS

# add health extension for vmss
az vmss extension set -g xiaojianxu --vmss-name xxj-vmss-1 --name ApplicationHealthLinux --version 1.0 --publisher Microsoft.ManagedServices --settings "{'port':80, 'protocol':'http', 'requestPath':'/'}"

# list vmss instances to get the instance ids, these ids will be used in following command
az vmss list-instances -g xiaojianxu -n xxj-vmss-1 --query "[].instanceId"

# update instances, if not do this, will report error in following command (Error msg: Automatic repairs not supported for this Virtual Machine Scale Set because a health extension is not applied to all instances in the virtual machine scale set prior to switching to instance health repair mode.)
az vmss update-instances -g xiaojianxu -n xxj-vmss-1 --instance-ids 0 1

# enable automatic repairs 
az vmss update -g xiaojianxu -n xxj-vmss-1 --enable-automatic-repairs true --automatic-repairs-grace-period 30
```

**History Notes:**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/vm/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_params.py
@@ -513,6 +513,9 @@ def load_arguments(self, _):
         c.argument('orchestration_mode', help='Choose how virtual machines are managed by the scale set. In VM mode, you manually create and add a virtual machine of any configuration to the scale set. In ScaleSetVM mode, you define a virtual machine model and Azure will generate identical instances based on that model.',
                    arg_type=get_enum_type(['VM', 'ScaleSetVM']), is_preview=True)
         c.argument('scale_in_policy', scale_in_policy_type)
+        c.argument('automatic_repairs_grace_period', min_api='2018-10-01', is_preview=True,
+            help='The amount of time (in minutes, between 30 and 90) for which automatic repairs are suspended due to a state change on VM.'
+        )
 
     with self.argument_context('vmss create', arg_group='Network Balancer') as c:
         LoadBalancerSkuName = self.get_models('LoadBalancerSkuName', resource_type=ResourceType.MGMT_NETWORK)
@@ -545,12 +548,18 @@ def load_arguments(self, _):
         c.argument('enable_automatic_repairs', min_api='2018-10-01', arg_type=get_three_state_flag(),
                    help='Enable automatic repairs', is_preview=True)
 
+    with self.argument_context('vmss update', min_api='2018-10-01', arg_group='Automatic Repairs', is_preview=True) as c:
+        c.argument('enable_automatic_repairs', arg_type=get_three_state_flag(), help='Enable automatic repairs')
+        c.argument(
+            'automatic_repairs_grace_period',
+            help='The amount of time (in minutes, between 30 and 90) for which automatic repairs are suspended due to a state change on VM.'
+        )
+
     for scope in ['vmss create', 'vmss update']:
         with self.argument_context(scope) as c:
             c.argument('terminate_notification_time', min_api='2019-03-01',
                        help='Length of time (in minutes, between 5 and 15) a notification to be sent to the VM on the instance metadata server till the VM gets deleted')
-            c.argument('automatic_repairs_grace_period', min_api='2018-10-01', is_preview=True,
-                       help='The amount of time (in minutes, between 30 and 90) for which automatic repairs are suspended due to a state change on VM.')
+
 
     for scope, help_prefix in [('vmss update', 'Update the'), ('vmss wait', 'Wait on the')]:
         with self.argument_context(scope) as c:
--- a/src/azure-cli/azure/cli/command_modules/vm/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_params.py
@@ -514,8 +514,7 @@ def load_arguments(self, _):
                    arg_type=get_enum_type(['VM', 'ScaleSetVM']), is_preview=True)
         c.argument('scale_in_policy', scale_in_policy_type)
         c.argument('automatic_repairs_grace_period', min_api='2018-10-01', is_preview=True,
-            help='The amount of time (in minutes, between 30 and 90) for which automatic repairs are suspended due to a state change on VM.'
-        )
+                   help='The amount of time (in minutes, between 30 and 90) for which automatic repairs are suspended due to a state change on VM.')
 
     with self.argument_context('vmss create', arg_group='Network Balancer') as c:
         LoadBalancerSkuName = self.get_models('LoadBalancerSkuName', resource_type=ResourceType.MGMT_NETWORK)
@@ -559,7 +558,6 @@ def load_arguments(self, _):
         with self.argument_context(scope) as c:
             c.argument('terminate_notification_time', min_api='2019-03-01',
                        help='Length of time (in minutes, between 5 and 15) a notification to be sent to the VM on the instance metadata server till the VM gets deleted')
-
 
     for scope, help_prefix in [('vmss update', 'Update the'), ('vmss wait', 'Wait on the')]:
         with self.argument_context(scope) as c:
--- a/src/azure-cli/azure/cli/command_modules/vm/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_params.py
@@ -550,7 +550,7 @@ def load_arguments(self, _):
             c.argument('terminate_notification_time', min_api='2019-03-01',
                        help='Length of time (in minutes, between 5 and 15) a notification to be sent to the VM on the instance metadata server till the VM gets deleted')
             c.argument('automatic_repairs_grace_period', min_api='2018-10-01', is_preview=True,
-                       help='The amount of time (in minutes, not less than 30) for which automatic repairs are suspended due to a state change on VM.')
+                       help='The amount of time (in minutes, between 30 and 90) for which automatic repairs are suspended due to a state change on VM.')
 
     for scope, help_prefix in [('vmss update', 'Update the'), ('vmss wait', 'Wait on the')]:
         with self.argument_context(scope) as c:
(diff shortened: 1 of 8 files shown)


arrownj:
I have a question here, there is a limitation for --automatic-repairs-grace-period, the value should between 30 to 90. For this case, should we validate it in CLI side ? or just rely on service side to validate this value ?
I see some similar cases which didn't do such check in CLI side, so I raise this problem for discussion. 

@qwordy @MyronFanQiu @fengzhou-msft @jiasli

src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
```diff
@@ -877,6 +877,13 @@ def build_vmss_resource(cmd, name, naming_prefix, location, tags, overprovision,
         }
         vmss_properties['virtualMachineProfile']['scheduledEventsProfile'] = scheduled_events_profile
 
+    if automatic_repairs_grace_period is not None:
+        automatic_repairs_policy = {
+            'enabled': 'true',
+            'gracePeriod': automatic_repairs_grace_period
```
yungezz: is grace period required property in rest api? if yes, it must be paired applied by user right? then need validation. is there default value? the 30 mins limitation is done at service side? If CLI hints user that constraint, would it be good to add constraint at CLI side? not sure what's the common practice there..

arrownj: From REST api perspective, it is not required. If it is omitted, service will give it a default value 30 to make sure the property has value.
Currently we only add --automatic-repairs-grace-period for `az vmss create` (no --enable-automatic-repairs). If user specify --automatic-repairs-grace-period, there means he/she want to enable automatic repairs. However this will cause that there will be no chance to trigger the service side default value logic.
I prefer our current design(not add --enable-automatic-repairs for `az vmss create`), because we have showed the value range in help message and it's better to make user know what value they are using explicitly. 
I'm not sure about this, so want to collect more suggestions here. 

For the question whether we should validate in CLI side, I prefer to not add. I think it's a duplicate work and service may change this logic later, if we add this validation, we also need to change according to service's changes later. Also want to listen your opinions.

yungezz: got it, thanks @arrownj .  Firstly, about parameters, `enable-automatic-repairs` and `automatic-repairs-grace-period` are different meaning, it's better to separate them to corresponding service api property instead of use implication that configuring grace period means enabling,
Secondly, for default value, yes it's more reliable to reply on service validation. If there's any description about value constraints at restapi spec, it will good if link that in help message to direct user to source of truth.

MyronFanQiu: I prefer not adding `--enable-automatic-repairs` in `vmss create`. `vmss update` has two arguments which is suitable. However, user cannot disable it since it's a create command. So it make senses to me that we just have one argument `--automatic-repairs-grace-period`.

yungezz: thanks @MyronFanQiu . we've offline discussed last Friday, will keep current behavior: use one parameter for setting grace period at creating, no enable parameter.

src/azure-cli/azure/cli/command_modules/vm/_validators.py
```diff
@@ -1633,3 +1635,27 @@ def _validate_vmss_terminate_notification(cmd, namespace):  # pylint: disable=un
...
+        if namespace.load_balancer is None or namespace.health_probe is None:
+            raise CLIError("usage error: --load-balancer and --health-probe are required "
+                           "when create vmss with automatic repairs")
+    _validate_vmss_automatic_repairs(cmd, namespace)
+
+
+def _validate_vmss_update_automatic_repairs(cmd, namespace):  # pylint: disable=unused-argument
+    if namespace.enable_automatic_repairs is False and namespace.automatic_repairs_grace_period is not None:
+        raise CLIError("usage error: please enable --enable-automatic-repairs")
+    if namespace.enable_automatic_repairs is True and namespace.automatic_repairs_grace_period is None:
```
MyronFanQiu: Is it necessary? Based on your description, I think we can drop this check here.

src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -542,11 +542,15 @@ def load_arguments(self, _):
                    help='Enable terminate notification')
         c.argument('ultra_ssd_enabled', ultra_ssd_enabled_type)
         c.argument('scale_in_policy', scale_in_policy_type)
+        c.argument('enable_automatic_repairs', min_api='2018-10-01', arg_type=get_three_state_flag(),
```
fengzhou-msft: Shall we use `action='store_true'` instead of `get_three_state_flag`?

src/azure-cli/azure/cli/command_modules/vm/_validators.py
```diff
@@ -1633,3 +1635,27 @@ def _validate_vmss_terminate_notification(cmd, namespace):  # pylint: disable=un
     """
     if namespace.terminate_notification_time is not None:
         namespace.terminate_notification_time = 'PT' + namespace.terminate_notification_time + 'M'
+
+
+def _validate_vmss_create_automatic_repairs(cmd, namespace):  # pylint: disable=unused-argument
+    if namespace.automatic_repairs_grace_period is not None:
+        if namespace.load_balancer is None or namespace.health_probe is None:
+            raise CLIError("usage error: --load-balancer and --health-probe are required "
+                           "when create vmss with automatic repairs")
```
fengzhou-msft: when creating

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -2686,6 +2688,11 @@ def update_vmss(cmd, resource_group_name, name, license_type=None, no_wait=False
             TerminateNotificationProfile(not_before_timeout=terminate_notification_time,
                                          enable=enable_terminate_notification)
 
+    if enable_automatic_repairs is not None or automatic_repairs_grace_period is not None:
+        AutomaticRepairsPolicy = cmd.get_models('AutomaticRepairsPolicy')
+        vmss.automatic_repairs_policy = \
+            AutomaticRepairsPolicy(enabled="true", grace_period=automatic_repairs_grace_period)
```
fengzhou-msft: With `enable_automatic_repairs` as `get_three_state_flag`, the value could be False when it's not None.

src/azure-cli/azure/cli/command_modules/vm/_params.py
fengzhou-msft: Maybe `get_three_state_flag` is also fine. `automaticRepairsPolicy` is a property of the resource, and `enable_automatic_repairs` could be considered as the `enabled` property of `automaticRepairsPolicy`.

fengzhou-msft: I suggest to put `enable_automatic_repairs` and `automatic_repairs_grace_period` in an argument group so customers know that these two arguments need to be used together.

arrownj: yes, they are in the same argument group. `enable_automatic_repairs` and `automatic_repairs_grace_period` are in `az vmss update`. While only `automatic_repairs_grace_period` in `az vmss create`. You can reference above discussion with Catherine, I explain why we design like this. And want to hear your options. Thanks

src/azure-cli/azure/cli/command_modules/vm/_validators.py
arrownj: you mean we leave this validation to service side ?

src/azure-cli/azure/cli/command_modules/vm/_params.py
fengzhou-msft: I mean you can add something like `arg_group='Automatic Repair Policy'`, otherwise the two options are apart from each other in alpabetic order in help message. Usually we name the second level property with a prefix of first level property, they will be showed together in help message, for instance, the names could be `automatic_repairs_policy_enable`, `automatic_reapirs_policy_grace_period`, but the first name looks weird with `enable` as suffix, so I think arg_group can work well in this case.

src/azure-cli/azure/cli/command_modules/vm/_validators.py
qwordy: default value 30? maybe you can omit this validation

src/azure-cli/azure/cli/command_modules/vm/_params.py
arrownj: arg_group added

src/azure-cli/azure/cli/command_modules/vm/_validators.py
arrownj: per discussed offline, let's just keep this validation.

PR #12414, opened 2020-03-02
[SQL DB, SQL MI] Add minimal_tls_version property for MI and SQL DB

cicovica:


yonzhan:
add to S166.

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -1039,6 +1041,10 @@ def _configure_security_policy_storage_params(arg_ctx):
                    help='Generate and assign an Azure Active Directory Identity for this server'
                    'for use with key management services like Azure KeyVault.')
 
+        c.argument('minimal_tls_version',
+                   arg_type=get_enum_type(SqlServerMinimalTlsVersionType),
```
Juliehzl: You can abstract minimal_tls_version using CLIArgumentType to reuse the type for all `minimal_tls_version` parameters.
e.g. https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/storage/_params.py#L41

cicovica:
> 2\. From my understanding, with SqlServerMinimalTlsVersion and SqlManagedInstanceMinimalTlsVersion, you give different allowed values for `create` and `update` command.  The actual difference is `None` value, right?

That is correct. We do not want to enable updating this property to `None` for SqlServer, only for SqlManagedInstance.

> So what if user doesn't specify min_tls_version when creating sql server? I mean default value.

If a user does not specify min_tls_version during creation, the behavior is the same as 'None'. We just do not want to allow updates to value `None` for SqlServer.

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -1246,6 +1252,10 @@ def _configure_security_policy_storage_params(arg_ctx):
                    arg_type=get_enum_type(ServerConnectionType),
                    help='The connection type used for connecting to the instance.')
 
+        c.argument('minimal_tls_version',
+                   arg_type=get_enum_type(SqlManagedInstanceMinimalTlsVersionType),
+                   help='The minimal TLS version enforced by the managed instance for inbound connections.')
```
jaredmoo: You already defined this in `sql server` scope, why do you need to repeat this in `sql server create` scope?

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -425,6 +425,19 @@ class FailoverPolicyType(Enum):
     manual = 'Manual'
 
 
+class SqlServerMinimalTlsVersionType(Enum):
```
jaredmoo: You should have defined this enum in swagger so that it showed up in generated sdk. Then you wouldn't need to define it here. This is what I was asking about on your swagger review. :)

src/azure-cli/azure/cli/command_modules/sql/_params.py
cicovica: I think this is the `sql mi` scope, please correct me if I'm wrong

src/azure-cli/azure/cli/command_modules/sql/custom.py
cicovica: Yes, sorry about that, we have plans to fix it and do it the right way :)

jaredmoo: Thank you, please do :)

PR #12425, opened 2020-03-03
[Storage] az storage account encryption-scope: Add support for customizing encryption scope

### LINKED ISSUE #12188: Request to support Encryption Scopes for Microsoft.Storage.
**Resource Provider**
Microsoft.Storage

Encryption scopes provide the ability to create isolated encryption enclaves within a storage account. Currently, Storage provides server-side encryption through AES-256 symmetric encryption in 2 flavors – using platform-managed keys(i.e. Storage creates and manages the life-cycle of these keys internally) & BYOK (Bring Your Own Key) using Azure Key Vaults. However, customers using a single storage account for multi-tenancy scenarios were limited to using a single account-scoped encryption key for all the data in the account. 
With encryption scopes, a customer creates and manages the life-time of this key material and is able to apply it at a blob-level or default it at a container level. These scopes are created via the control plane through the API’s described in the swagger PR and are referenced in data-plane operations via headers. Particularly, when an encryption scope is created, a customer can: 
-	Create a blob using the referenced encryption scope by passing the name in the HTTP request headers.
-	Create a container setting the default encryption scope for all blobs created in it via HTTP request headers. 

This allows for customer blobs within a container/storage account to be isolated between encryption scope boundaries, since the key material used by each encryption scope is different. At the same time, customers don’t explicitly need to see/store the key itself – while the customer has control over it’s lifecycle, Storage manages the key itself and customers simply reference them by encryption scope name in requests. 

Re-iterating a few customer scenarios:
For eg. 
-	A customer can create an encryption scope and by passing a reference to this created scope while creating a blob, this blob will be encrypted on server-side using key material from this encryption scope. Note that once a scope is assigned to a blob, this cannot be changed. 
-	A customer can assign an encryption scope to a container during container creation time and this will default all blob creates within this container to use the key material from the specified scope. 
-	A customer can optionally explicitly override the default container encryption scope by passing a custom header during blob create. 
-	A customer can choose to opt-in to use platform-managed or BYOK with Azure Key Vault for the encryption scope. To opt-in to platform-managed, he would specify source=”Microsoft.Storage”. For BYOK, he would specify source=”Microsoft.KeyVault” with a reference to the keyUri in the “keyVaultProperties” attribute. Note that customers can freely move between these 2 flavors just as is the case with server-side encryption on storage accounts. 
-	Encryption scopes are created as state=”Enabled” by default. But, a customer can move the encryption scope state to “Disabled” which will cause all blob read/write operations which use this encryption scope to fail on the data-plane. Additionally, the customer will no longer be billed for this encryption scope once the state is set to “Disabled”. 

Below is the API specification doc for the REST changes we effected to support this in case this is useful. 
[SRP Encryption Scopes](https://microsoft.sharepoint.com/:w:/t/AzureStorage/EdGMgSU6-rRGiOnD6Bl3OFIBIoMkS3Q66cFUI4wK7r2AZw?e=7vVRUv)

2019-06-01

**Swagger Link**
https://github.com/Azure/azure-rest-api-specs/pull/8381

**Target Date**
As soon as possible.

Juliehzl:
Fix #12188 

`New feature`:
1. Add support to customize encryption scope for storage account:
`az storage account encryption-scope create/update/show/list`
2. Specify default encryption scope in container level
`az storage container create` with `--default-encryption-scope` and `deny-encryption-scope-override`

**History Notes:**  

[Storage] az storage account encryption-scope create/update/list/show: Add support to customize encryption scope.
[Storage] az storage container create: Add --default-encryption-scope and --deny-encryption-scope-override to set encryption scope for container level.

---


yonzhan:
add to S166.

src/azure-cli/azure/cli/command_modules/storage/operations/account.py
```diff
@@ -443,3 +443,36 @@ def update_blob_service_properties(cmd, instance, enable_change_feed=None, enabl
...
+def create_encryption_scope(cmd, client, resource_group_name, account_name, encryption_scope_name,
+                            encryption_key_source=None, encryption_key_uri=None):
+    EncryptionScope = cmd.get_models('EncryptionScope')
+
+    if encryption_key_source:
+        encryption_scope = EncryptionScope(source=encryption_key_source)
+
+    if encryption_key_uri:
+        EncryptionScopeKeyVaultProperties = cmd.get_models('EncryptionScopeKeyVaultProperties')
+        encryption_scope.key_vault_properties = EncryptionScopeKeyVaultProperties(key_uri=encryption_key_uri)
```
zhoxing-ms: When the `encryption_key_uri` has a value but the `encryption_key_source` has no value, the `encryption_scope` will not be assigned, will it cause the null pointer exception of the `encryption_scope.key_vault_properties`?

src/azure-cli/azure/cli/command_modules/storage/operations/account.py
```diff
@@ -443,3 +443,36 @@ def update_blob_service_properties(cmd, instance, enable_change_feed=None, enabl
...
+
+    if state is not None:
+        encryption_scope.state = EncryptionScopeState(state)
```
zhoxing-ms: Same as above

src/azure-cli/azure/cli/command_modules/storage/operations/blob.py
```diff
@@ -21,6 +22,26 @@
...
+                                       deny_encryption_scope_override=deny_encryption_scope_override)
+        container = client.create(resource_group_name=resource_group_name, account_name=account_name,
+                                  container_name=container_name, blob_container=blob_container)
+        return container is not None
+
+    from .._client_factory import blob_data_service_factory
+    kwargs['account_name'] = account_name
+    client = blob_data_service_factory(cmd.cli_ctx, kwargs)
+    return client.create_container(container_name, metadata=metadata, public_access=public_access,
+                                   fail_on_exist=fail_on_exist, timeout=timeout)
```
zhoxing-ms: Are the data type of `container is not None` and `client.create_container` consistent?

src/azure-cli/azure/cli/command_modules/storage/README.md
```diff
@@ -0,0 +1,63 @@
...
+    -g MyResourceGoup
+```
+
+#### Disable specified encryption scope in a storage account.
+```
+az storage account encryption-scope update \
+    -n myencryption \
+    --account-name mystorageaccount \
+    -g MyResourceGoup \
+    --disabled
```
zhoxing-ms: Why the usage of the example in help is `--state Disabled`? Are both `--state Disabled` and `--disabled` supported?

src/azure-cli/azure/cli/command_modules/storage/operations/blob.py
Juliehzl: yes

src/azure-cli/azure/cli/command_modules/storage/operations/account.py
Juliehzl: Actually it will key uri will be ignored by server side. But it is a good point to check key uri for key source.

src/azure-cli/azure/cli/command_modules/storage/README.md
Juliehzl: It is already updated.

src/azure-cli/azure/cli/command_modules/storage/operations/account.py
Juliehzl: key source will always has default value and they have validator to check usage.

src/azure-cli/azure/cli/command_modules/storage/_help.py
```diff
@@ -83,6 +83,59 @@
...
+helps['storage account encryption-scope'] = """
+type: group
+short-summary: Manage encryption scope for a storage account.
+"""
+
+helps['storage account encryption-scope create'] = """
+type: command
+short-summary: Create an encryption scope within storage account.
+examples:
+  - name: Create an encryption scope within storage account based on Micosoft.Storage key source.
```
jiasli, line 95: "Micosoft" is misspelled.

PR #12426, opened 2020-03-03
[RBAC] az ad group show: fix --group value treated as regex problem

### LINKED ISSUE #11599: single group subcommand values treated as regex
azure-cli                         2.0.75 

```

az ad group list -o tsv --query "[? displayName == 'PAC.G.ODTA' || mailNickname == 'PAC.G.ODTA' ].[objectId,displayName]"
1__REDACTED__4	PAC.G.ODTA

az ad group show --group 'PAC.G.ODTA' -o tsv --query "[objectId,displayName]"
More than one groups match the name of 'PAC.G.ODTA'
```

arrownj:
This is a PR to fix #11599 

Previously, if there are two groups, one is the prefix of anther, take below cases for example:

- show_regex_1
- show_regex_11

If we use `az ad gourp show --group show-regex_1`, it will raise "More than one groups match the name of show_regex_1" error, which is not correctly. The right behavior should be match the exact name first, if not found, then match with prefix. Previously, use would never be able to show show_regex_1.

And please be noted that, this update also impact previous `az ad group delete` and `az ad group get-member-groups`, they share the same logic with --group parameter. To cover this case in tests, only one command of these three is enough.

**History Notes:**  

---


yonzhan:
az ad group

src/azure-cli/azure/cli/command_modules/role/_validators.py
```diff
@@ -13,25 +13,31 @@
...
+    count = len(result)
+    ret = False
+    if count == 1:
+        setattr(namespace, attr, result[0].object_id)
+        ret = True
+    elif count == 0:
+        if not has_next_filter:
+            raise CLIError("No group matches the name of '{}'".format(value))
+    else:
+        raise CLIError("More than one groups match the name of '{}'".format(value))
```
yungezz: group

src/azure-cli/azure/cli/command_modules/role/tests/latest/test_graph.py
```diff
@@ -262,17 +262,14 @@ def test_graph_group_scenario(self):
         account_info = self.cmd('account show').get_output_in_json()
         if account_info['user']['type'] == 'servicePrincipal':
             return  # this test delete users which are beyond a SP's capacity, so quit...
-        upn = account_info['user']['name']
-        domain = upn.split('@', 1)[1]
         self.kwargs = {
             'user1': 'deleteme1',
             'user2': 'deleteme2',
-            'domain': domain,
+            'domain': 'azuresdkteam.onmicrosoft.com',
```
jiasli: Let's not hard code it. But create a new user in the `azuresdkteam.onmicrosoft.com` tenant with **global admin** permission to run this test again.

jiasli:
Our tenant contains many trash groups and all of them are included in the recording. We can use a PowerShell script to clean the groups in the tenant: https://github.com/jiasli/azure-cli/blob/dev/doc/use_cli_effectively.md#windows-powershell-scripts-to-loop-through-a-list

src/azure-cli/azure/cli/command_modules/role/_validators.py
```diff
@@ -13,25 +13,31 @@
 VARIANT_GROUP_ID_ARGS = ['object_id', 'group_id', 'group_object_id']
 
 
+def _validate_group(namespace, attr, value, group_filter, has_next_filter):
```
jiasli: Personally, I think the usage of `has_next_filter` lowers the cohesion of this function. It makes `_validate_group` rely on the external logic and dependent between executions. I prefer not to exact the logic to `_validate_group` but put it in the original `validate_group` instead.

arrownj: I'm also hesitated at it. @qianwens , want to hear your opinions.

src/azure-cli/azure/cli/command_modules/role/tests/latest/test_graph.py
```diff
@@ -262,17 +262,14 @@ def test_graph_group_scenario(self):
         account_info = self.cmd('account show').get_output_in_json()
```
jiasli: `user.name` from `az account show` may be different from the real **user principal name** if the user is a guest account in this tenant. To get the real user principal name, look for `userPrincipalName` from `az ad signed-in-user show`. 

For example my `user.name` from `az account show` is `myid@microsoft.com`, but `userPrincipalName` from `az ad signed-in-user show` is `myid_microsoft.com#EXT#@AzureSDKTeam.onmicrosoft.com`

src/azure-cli/azure/cli/command_modules/role/_validators.py
qianwens: I think it is correct to have _validate_group function or the handle exception logic in validate_group looks complicated. But for _validate_group, I agree that we should not raise different kind of error message based on the has_next_filter paramete. Currently _validate_group 's return logic is not predictable, sometime it return false, sometime it raise error, could you please refine the logic here, or you can replace this functio with _get_group_count to just return the count and let the validate_group to raise the exception

---
In reply to: [387564846](https://github.com/Azure/azure-cli/pull/12426#discussion_r387564846) [](ancestors = 387564846)

src/azure-cli/azure/cli/command_modules/role/tests/latest/test_graph.py
arrownj: updated

src/azure-cli/azure/cli/command_modules/role/tests/latest/test_graph.py
```diff
@@ -272,7 +274,6 @@ def test_graph_group_scenario(self):
             'group': 'deleteme_g',
             'pass': 'Test1234!!'
         }
-        self.recording_processors.append(AADGraphUserReplacer('@' + domain, '@example.com'))
```
jiasli: We'd better still have this line to remove the tenant info from recordings.

arrownj: Add the replay_processor. Have no idea why previous test can pass.

arrownj: updated

src/azure-cli/azure/cli/command_modules/role/_validators.py
arrownj: updated

src/azure-cli/azure/cli/command_modules/role/_validators.py
```diff
@@ -13,25 +13,34 @@
...
+            )
+            if prefix_match_count == 1:
+                setattr(namespace, attr, object_id)
+            elif prefix_match_count == 0:
+                raise CLIError("No group matches the name of '{}'".format(value))
+            else:
+                raise CLIError("More than one group match the name of '{}'".format(value))
         else:
-            raise CLIError("More than one groups match the name of '{}'".format(value))
+            raise CLIError("More than one group match the name of '{}'".format(value))
```
jiasli: This will never be reached. But yes, let's keep it in case the service doesn't return 0 or 1 item as expected.

PR #12431, opened 2020-03-03
{Core} Change help example hook

mirdaki:
This PR updates the "hook" that was previously added to the core help file. The change moves the hook from the printing method to the help file loading method. This allows examples to be added to commands that previously had none (such as `az aks list`) and to groups. It will be utilized by the `ai-examples` extension (with [this PR](https://github.com/Azure/azure-cli-extensions/pull/1345)),

---


yonzhan:
add to S166.

src/azure-cli-core/azure/cli/core/_help.py
```diff
@@ -192,8 +195,8 @@ def update_loaders_with_help_file_contents(self, nouns):
 
     # This method is meant to be a hook that can be overridden by an extension or module.
     @staticmethod
-    def example_provider(help_file):
-        return help_file.examples
+    def check_for_new_examples(command):  # pylint: disable=unused-argument
+        return []
```
jiasli: We can use the name `check_for_ai_examples` to make the purpose clearer.

jiasli: My concern is by taking `command` as argument and use `self.examples = new_examples` later, this will totally remove the built-in examples. This is really intended?

src/azure-cli-core/azure/cli/core/_help.py
```diff
@@ -248,6 +251,12 @@ def load(self, options):
         for loader in ordered_loaders:
             loader.versioned_load(self, options)
 
+        # Used as a hook for changing the default examples
+        if self.help_ctx.original_nouns == self.command:
```
jiasli: To me, it seems the purpose of `original_nouns` is to detect whether the help is loaded for the original command like `vm` or its children, like `vm availability-set`, so that we only load AI examples for the command itself, not its children.

```
> az vm -h

Group
    az vm              << Load AI examples for it

Subgroups:
    availability-set   << Don't load AI examples for these
    boot-diagnostics
    diagnostics
...
```
If so, an easier approach is to copy the logic from `knack.help.CLIHelp.show_help` to `azure.cli.core._help.AzCliHelp.show_help` and add the replacement/addition logic here, instead of going very deep and evaluate for every `CliHelpFile.load` whether the help is retrieved for the original command - `vm`. In other words, we can actively add the AI examples after the help is rendered, instead of passively evaluating the given command.

```py
    # azure.cli.core._help.AzCliHelp.show_help
    def show_help(self, cli_name, nouns, parser, is_group):
        self.update_loaders_with_help_file_contents(nouns)
        self.original_nouns = ' '.join(nouns)
        import colorama
        colorama.init(autoreset=True)
        delimiters = ' '.join(nouns)
        help_file = self.command_help_cls(self, delimiters, parser) if not is_group \
            else self.group_help_cls(self, delimiters, parser)
        help_file.load(parser)
        if not nouns:
            help_file.command = ''
        else:
            new_examples = self.check_for_new_examples(help_file.command)  # ai examples are added here
            if new_examples:
                help_file.examples = new_examples
        self._print_detailed_help(cli_name, help_file)

        print(SURVEY_PROMPT)
```

Also, I would suggest to write some tests similar to 

https://github.com/Azure/azure-cli/blob/39ca86e9e9e98981024ed95db5167f440de1d4f6/src/azure-cli-core/azure/cli/core/tests/test_help.py#L345-L353

so that we can verify the hook is working as expected.

src/azure-cli-core/azure/cli/core/_help.py
mirdaki: So the current plan is to entirely replace the default examples, but I have debated how to best build this hook. 

I like the idea of leaving it extensible, so in the future the extension could do some form of combination without needing to update the core CLI again. I'll make a change that will pass in the default examples, which should also simplify the logic here too.

src/azure-cli-core/azure/cli/core/_help.py
mirdaki: Correct, that is the purpose! If that's a cleaner approach I can switch to it.

I will add test in the next PR.

src/azure-cli-core/azure/cli/core/_help.py
```diff
@@ -147,10 +147,21 @@ def new_normalize_text(s):
...
+        import colorama
+        colorama.init(autoreset=True)
+        delimiters = ' '.join(nouns)
+        help_file = self.command_help_cls(self, delimiters, parser) if not is_group \
+            else self.group_help_cls(self, delimiters, parser)
+        help_file.load(parser)
+        if not nouns:
+            help_file.command = ''
+        else:
+            help_file.examples = AzCliHelp.check_for_new_examples(help_file.command, help_file.examples)
```
jiasli: Let's simplify the function signature.
```suggestion
            AzCliHelp.check_for_new_examples(help_file)
```

src/azure-cli-core/azure/cli/core/_help.py
```diff
@@ -192,8 +203,8 @@ def update_loaders_with_help_file_contents(self, nouns):
 
     # This method is meant to be a hook that can be overridden by an extension or module.
     @staticmethod
-    def example_provider(help_file):
-        return help_file.examples
+    def check_for_new_examples(command, examples):  # pylint: disable=unused-argument
+        return examples
```
jiasli:
```suggestion
    def check_for_new_examples(help_file):  # pylint: disable=unused-argument
        pass
```

src/azure-cli-core/azure/cli/core/_help.py
jiasli: That makes sense. How about `update_examples`?  `check_for_new_examples` is also fine to me. It's up to you. 😊

jiasli: The change is great!

mirdaki: `update_examples` sounds good to me!

src/azure-cli-core/azure/cli/core/_help.py
mirdaki: Updated!

src/azure-cli-core/azure/cli/core/_help.py
mirdaki: Updated!

Issue #12185, opened 2020-02-13
Add application insight endpoint to azure cli


kicha0:
**Is your feature request related to a problem? Please describe.**
Currently azure cli does not provide some of the application insight endpoint for different cloud regions. For example, ingestion endpoint for application insight in us government region is
https://dc.applicationinsights.us/v2/track

**Describe the solution you'd like**
Azure cli team maintain application insight endpoint mapping and provide the information in az cloud show.
The current endpoint list for application insight can be found here:
https://docs.microsoft.com/en-us/azure/azure-monitor/app/custom-endpoints

**Describe alternatives you've considered**
We currently maintain a mapping from cloud region to application insight ingestion endpoint in our application. 

**Additional context**
As part of effort to automate azureml service deployment in new region, we would like to be able to query the correct application insight endpoint given cloud region so that correct endpoint can be consumed at build time.

arrownj:
@kicha0 , thanks for reporting, we are looking at it.

yungezz:
the feature is released in Azure CLI 2.2.0. close the issue

PR #12442, opened 2020-03-04
[Core] az cloud show: add application insights endpoint for China/US cloud

arrownj:
This is a PR to achieve the request of #12185 .

Currently, only application insight channel endpoint is needed. As the offical document only have China/US cloud's endpoints, so we only add these 2 regions for user as they required.

**History Notes:**  

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/cloud.py
+++ b/src/azure-cli-core/azure/cli/core/cloud.py
@@ -68,7 +68,8 @@ class CloudEndpoints(object):  # pylint: disable=too-few-public-methods,too-many
                  media_resource_id=None,
                  ossrdbms_resource_id=None,
                  log_analytics_resource_id=None,
-                 app_insights_resource_id=None):
+                 app_insights_resource_id=None,
+                 app_insights_telemetry_channel_resource_id=None):
         # Attribute names are significant. They are used when storing/retrieving clouds from config
         self.management = management
         self.resource_manager = resource_manager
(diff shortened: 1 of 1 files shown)


jiasli:
1. Without corresponding code usage of `app_insights_telemetry_channel_resource_id`, no command except for `az cloud show` will be affected. Is this intended?
2. Don't we need to add `app_insights_telemetry_channel_resource_id` to `AzureCloud` as well? I think it should be `https://dc.services.visualstudio.com/v2/track` according to https://docs.microsoft.com/en-us/azure/azure-monitor/faq#example-applicationinsightsconfig-with-default-endpoints. Better to confirm with Application Insight service team

arrownj:
> Without corresponding code usage of app_insights_telemetry_channel_resource_id, no command except for az cloud show will be affected. Is this intended?
> Don't we need to add app_insights_telemetry_channel_resource_id to AzureCloud as well? I think it should be https://dc.services.visualstudio.com/v2/track according to https://docs.microsoft.com/en-us/azure/azure-monitor/faq#example-applicationinsightsconfig-with-default-endpoints. Better to confirm with Application Insight service team

1. yes, user just want to get the output of `az cloud show`. I think besides `az cloud show`, it also impact the output of `az cloud list`.

2. yes, Kin give me this link before, as it is not an official way, I didn't add it to AzureCloud. Your suggestion is very useful. I will send an email to app insights team to get their confirmation.

jiasli:
In that case, let's merge it for now and add `app_insights_telemetry_channel_resource_id` to `AzureCloud` in the future.

fengzhou-msft:
Do we need to change code in application insight extension as well?

MyronFanQiu:
Can you share the context with me offline in case I need to do something for application insight extension. Based on my understanding, this PR just have impact on az cloud show/list command.

Issue #12066, opened 2020-02-06
Make "az deployment create" commands use --handle-extended-json-format by default


alex-frankel:
**Resource Provider**
Microsoft.Resources/deployments

**Description of Feature or Work Requested**
Make the flag `--handle-extended-json-format` the default behavior for all variants of `az deployment create` (`deployment group create`, `deployment sub create`, `deployment tenant create`, `deployment mg create`)

**Minimum API Version Required**
all

**Swagger Link**
Provide a link to the location of your feature(s) in the REST API specs repo. If your feature(s) has corresponding commit or pull request in the REST API specs repo, provide them. This should be on the master branch of the REST API specs repo.
N/A

**Target Date**
If you have a target date for release of this feature/work, please provide it. While we can't guarantee these dates,
it will help us prioritize your request against other requests.
2/14

yonzhan:
add to S166.

PR #12455, opened 2020-03-05
[ARM] az deployment create/validate: Use --handle-extended-json-format by default (#12066)

zhoxing-ms:
Feature request: #12066

**History Notes:**  

---


yonzhan:
add to S167.

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -190,15 +190,17 @@ def load_arguments(self, _):
 
     with self.argument_context('group deployment create') as c:
         c.argument('deployment_name', arg_type=deployment_create_name_type)
-        c.argument('handle_extended_json_format', arg_type=extended_json_format_type)
+        c.argument('handle_extended_json_format', arg_type=extended_json_format_type,
```
Juliehzl: you can only add deprecate_info in extended_json_format_type.

src/azure-cli/azure/cli/command_modules/resource/tests/latest/management_group_level_template.json
```diff
@@ -43,73 +43,13 @@
...
       "apiVersion": "2016-12-01",
-      "dependsOn": [ "policy2" ],
+      "dependsOn": [
+        "policy2"
+      ],
       "properties": {
         "scope": "[variables('mgScope')]",
         "policyDefinitionId": "[concat(variables('mgScope'), '/providers/', 'Microsoft.Authorization/policyDefinitions/', 'policy2')]"
       }
-    },
```
Juliehzl: May I know why we remove deployment resource below?

src/azure-cli/azure/cli/command_modules/resource/tests/latest/management_group_level_parameters.json
```diff
@@ -2,14 +2,8 @@
...
   "contentVersion": "1.0.0.0",
   "parameters": {
-    "targetMG": {
-      "value": "cli_test_management_group_deployment"
-    },
     "nestedsubId": {
-      "value": "fb3a3d6b-44c8-44f5-88c9-b20917c9b96b"
-    },
-    "nestedRG": {
-      "value": "cli_test_management_group_deployment"
```
Juliehzl: May I know why we remove such parameter values?

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -671,72 +603,36 @@ def _prepare_deployment_properties_unmodified(cli_ctx, template_file=None, templ
...
+    smc.config.pipeline = Pipeline(
+        policies=[
+            JsonCTemplatePolicy(),
+            smc.config.user_agent_policy,
+            RequestsPatchSession(),
+            smc.config.http_logger_policy,
+            RequestsCredentialsPolicy(smc.config.credentials)
+        ],
+        sender=PipelineRequestsHTTPSender(RequestsHTTPSender(smc.config))
+    )
```
Juliehzl: I will prefer to abstract this part as a new "config client" function.

src/azure-cli/azure/cli/command_modules/resource/_params.py
zhoxing-ms: Yes, that's what I wanted to write at the beginning.  But I found that this would cause an error if the specific command's argument_context was not bound. 
If you have a good way to solve this problem, please share with me, thank you~

src/azure-cli/azure/cli/command_modules/resource/tests/latest/management_group_level_template.json
zhoxing-ms: Because it will cause an internal error of the Service, but there is no specific error message (only service internal error), the Service seems to have an unknown problem. I don't think these resources affect the main logic of the test deployment, so I removed them first.

src/azure-cli/azure/cli/command_modules/resource/custom.py
zhoxing-ms: Actually, I just deleted the logic of `if handle_extended_json_format`.  
So I didn't understand your suggestion, Could you please be more specific

src/azure-cli/azure/cli/command_modules/resource/tests/latest/management_group_level_parameters.json
zhoxing-ms: As above comment, some data is no longer needed. Some other data is not written in the configuration parameters, changed to get in the code.

src/azure-cli/azure/cli/command_modules/resource/tests/latest/management_group_level_template.json
zhoxing-ms: Resolved

PR #12475, opened 2020-03-06
[RDBMS] Updating RDBMS Private Endpoint Tests

ramnov:
**History Notes:**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/rdbms/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/rdbms/custom.py
@@ -440,7 +440,7 @@ def _update_private_endpoint_connection_status(cmd, client, resource_group_name,
     return client.create_or_update(resource_group_name=resource_group_name,
                                    server_name=server_name,
                                    private_endpoint_connection_name=private_endpoint_connection_name,
-                                   properties=private_endpoint_connection)
+                                   private_link_service_connection_state=private_endpoint_connection.private_link_service_connection_state)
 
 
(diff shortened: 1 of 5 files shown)


src/azure-cli/azure/cli/command_modules/rdbms/custom.py
```diff
@@ -440,7 +440,7 @@ def _update_private_endpoint_connection_status(cmd, client, resource_group_name,
     return client.create_or_update(resource_group_name=resource_group_name,
                                    server_name=server_name,
                                    private_endpoint_connection_name=private_endpoint_connection_name,
-                                   properties=private_endpoint_connection)
+                                   private_link_service_connection_state=private_endpoint_connection.private_link_service_connection_state)
```
MyronFanQiu: a little bit confused about this PR. should it be `parameter` or `private_endpoint_connection_parameter`? Why do we just send the state instead of the whole `private_endpoint_connection`?

ramnov: Private Endpoint Connection State can be updated from Pending to Approved (or) Pending to Rejected. In order to update the state of the Private Endpoint Connection, only the required state, description is sufficient to pass to client.create_or_update()

MyronFanQiu: I used to think private_link_resource_state should be part of private_endpoint. I checked the python SDK. it's a little bit weird for me. But it's fine and it would work. 😄

PR #12477, opened 2020-03-06
{Core} Relax humanfriendly dependency

### LINKED ISSUE #12283: Please support newer versions of humanfriendly
The ```azure-cli-core``` package currently depends on ```humanfriendly~=4.7``` meaning the minimum version is 4.7 and the maximum version supported is 5.0.

The current upstream version is 7.1.1 (https://pypi.org/project/humanfriendly) and in openSUSE, ```humanfriendly``` got just bumped to 5.0 making ```azure-cli-core``` uninstallable.

I will try to relax the dependency of the openSUSE package a bit and see if it works with 5.0 as well, but in any case, it would be desirable if newer versions of ```humanfriendly``` could be supported.

jiasli:
Fix #12283: Please support newer versions of humanfriendly

Tested with humanfriendly 8.0, and `az deployment group create -g {} --template-file template.json --parameters "@parameters.json"` works as expected:

```
 - Running ..  
```

### DIFF
--- a/src/azure-cli-core/azure/cli/core/commands/progress.py
+++ b/src/azure-cli-core/azure/cli/core/commands/progress.py
@@ -114,7 +114,7 @@ class IndeterminateStandardOut(ProgressViewBase):
         :param args: dictionary containing key 'message'
         """
         if self.spinner is None:
-            self.spinner = humanfriendly.Spinner(
+            self.spinner = humanfriendly.Spinner(  # pylint: disable=no-member
                 label='In Progress', stream=self.out, hide_cursor=False)
         msg = args.get('message', 'In Progress')
         try:
--- a/src/azure-cli-core/azure/cli/core/tests/test_progress.py
+++ b/src/azure-cli-core/azure/cli/core/tests/test_progress.py
@@ -60,7 +60,7 @@ class TestProgress(unittest.TestCase):
         before = view.spinner.total
         view.write({})
         after = view.spinner.total
-        self.assertTrue(after >= before)
+        self.assertTrue(after == before)
         view.write({'message': 'TESTING'})
 
     def test_progress_indicator_det_stdview(self):
--- a/src/azure-cli-core/setup.py
+++ b/src/azure-cli-core/setup.py
@@ -55,7 +55,7 @@ DEPENDENCIES = [
     'argcomplete~=1.8',
     'azure-cli-telemetry',
     'colorama>=0.3.9',
-    'humanfriendly~=4.7',
+    'humanfriendly>=4.7,<9.0',
     'jmespath',
     'knack~=0.6.2',
     'msrest>=0.4.4',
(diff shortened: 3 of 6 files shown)


jiasli:
@Juliehzl , storage module heavily relies on this lib. 

https://github.com/Azure/azure-cli/blob/d08a69c33e438978baf6232ce2ee315c9d45aae2/src/azure-cli/azure/cli/command_modules/storage/_validators.py#L737

Please verify progress is displayed correctly after bumping the version.

src/azure-cli-core/azure/cli/core/tests/test_progress.py
```diff
@@ -60,7 +60,7 @@ def test_progress_indicator_indet_stdview(self):
         before = view.spinner.total
         view.write({})
         after = view.spinner.total
-        self.assertTrue(after >= before)
+        self.assertTrue(after == before)
```
jiasli: In humanfriendly 4.7 `spinner.total` is always the default value 0. In 8.0, the default value is `None` which makes `>=` invalid.

fengzhou-msft:
tested with 8.0 in extension add, LGTM

src/azure-cli-core/azure/cli/core/commands/progress.py
```diff
@@ -114,7 +114,7 @@ def write(self, args):
         :param args: dictionary containing key 'message'
         """
         if self.spinner is None:
-            self.spinner = humanfriendly.Spinner(
+            self.spinner = humanfriendly.Spinner(  # pylint: disable=no-member
```
jiasli: `humanfriendly.Spinner` is moved to `humanfriendly.terminal.spinners.Spinner` in later versions. Use the deprecated `humanfriendly.Spinner` for compatibility.

src/azure-cli-core/setup.py
```diff
@@ -55,7 +55,7 @@
     'argcomplete~=1.8',
     'azure-cli-telemetry',
     'colorama>=0.3.9',
-    'humanfriendly~=4.7',
+    'humanfriendly>=4.7,<9.0',
```
jiasli: @glaubitz, I have relaxed the dependency. Hope it will work on all Linux distros.

dbaio:
Hi. Could we support `humanfriendly` `9.1` as well?

Just `pluralize()` is backwards incompatible [1] and `azure-cli-core` is not using this function.

1 - https://humanfriendly.readthedocs.io/en/latest/changelog.html#release-9-1-2020-12-10

Thank you.

jiasli:
Hi @dbaio, just our of curiosity, are you using `azure-cli-core` directly, installing it in one of your virtual environment or using some unofficial installation source like OpenSUSE?

I will try to bump it and do some testing in one of the following sprints.

dbaio:
@jiasli Sorry for the late reply. I use it on `FreeBSD` (pkg system, not pypi/virtualenv).
And thank you!

jiasli:
Yeah, I see. It looks like **FreeBSD 13** only ships with [py37-humanfriendly-9.1.txz](https://freebsd.pkgs.org/13/freebsd-i386/py37-humanfriendly-9.1.txz.html).

https://pkgs.org/search/?q=humanfriendly

image

(These unofficial packing systems really cause us lots of issues as they may not support multi-version packages. 😕)

Issue #6962, opened 2018-08-03
`az login` fails due to MFA


davidobrien1985:
**Describe the bug**
`az login` doesn't work for me anymore with an MFA enabled user.

```
Failed to authenticate '{'additional_properties': {}, 'id': '/tenants/4253165e-ba77-4eaa-bd15-e7abb69a74ef', 'tenant_id': '4253165e-ba77-4eaa-bd15-e7abb69a74ef'}' due to error 'Get Token request returned http error: 400 and server response: {"error":"interaction_required","error_description":"AADSTS50076: Due to a configuration change made by your administrator, or because you moved to a new location, you must use multi-factor authentication to access '797f4846-ba00-4fd7-ba43-dac1f8f63013'.\r\nTrace ID: 384cee7c-cf4c-4dfe-8a11-a510c4e90c00\r\nCorrelation ID: 66394469-089e-42dc-aaf6-e5d6e77c7987\r\nTimestamp: 2018-08-03 12:27:46Z","error_codes":[50076],"timestamp":"2018-08-03 12:27:46Z","trace_id":"384cee7c-cf4c-4dfe-8a11-a510c4e90c00","correlation_id":"66394469-089e-42dc-aaf6-e5d6e77c7987","suberror":"basic_action"}'
```

**To Reproduce**
`az login`, also `read -sp "Azure password: " AZ_PASS && echo && az login -u auobrien.david@outlook.com -p $AZ_PASS` doesn't work.

**Expected behavior**
I receive an MFA prompt on my phone.

**Environment summary**
`az --version`
azure-cli (2.0.43)

Installed on WSL.

yugangw-msft:
For MFA, you have to use interactive login through `az login` w/o `-u`. This requirement comes from AAD token service, not CLI.

davidobrien1985:
Sorry if I wasn't clear.

Interactive AND non-interactive fail with MFA.

Sent from my iPhone

On 4 Aug 2018, at 2:05 am, Yugang Wang <notifications@github.com<mailto:notifications@github.com>> wrote:


For MFA, you have to use interactive login through az login w/o -u. This requirement comes from AAD token service, not CLI.

yugangw-msft:
Are you able to login in to portal? If yes, can you try `az login` again?

davidobrien1985:
Yes, all fine via the portal and cloud shell.

Only CLI on my local is broken.

Sent from my iPhone

[...]

PR #12516, opened 2020-03-10
[Profile] az login: Show warning for MFA error

jiasli:
Partially mitigate https://github.com/Azure/azure-cli/issues/6962. MFA cross-tenant trust is still under development by service team.

During `az login`, show warning for MFA error. This happens when the home tenant doesn't require MFA but the other tenants where the user is a guest of do. When getting access token with refresh token for those MFA tenants, CLI gives the raw error which isn't very instructive:

> AADSTS50076: Due to a configuration change made by your administrator, or because you moved to a new location, you must use multi-factor authentication to access 'xxx'.

This PR checks the exception and provides instructions on how to work with MFA tenants:

> Tenant da086fa3-023f-4500-bf56-0f19d3065440 requires Multi-Factor Authentication (MFA). To access this tenant, use 'az login --tenant' to explicitly login to this tenant.

For the test, since CI still uses nosetests while we locally use pytest, it is difficult to verify captured logs in both test frameworks: https://docs.pytest.org/en/latest/logging.html#caplog-fixture

We can manually run the test with `-o log_cli=True` to check that the warning is displayed correctly:

```
pytest -o log_cli=True src/azure-cli-core/azure/cli/core/tests/test_profile.py::TestProfile::test_find_using_common_tenant_mfa_warning
```

### DIFF
--- a/src/azure-cli-core/azure/cli/core/_profile.py
+++ b/src/azure-cli-core/azure/cli/core/_profile.py
@@ -879,7 +879,7 @@ class SubscriptionFinder(object):
                 # because user creds went through the 'common' tenant, the error here must be
                 # tenant specific, like the account was disabled. For such errors, we will continue
                 # with other tenants.
-                msg = getattr(ex, 'error_response', {}).get('error_description', '')
+                msg = (getattr(ex, 'error_response', None) or {}).get('error_description') or ''
                 if 'AADSTS50076' in msg:
                     logger.warning("Tenant %s requires Multi-Factor Authentication (MFA). "
                                    "To access this tenant, use 'az login --tenant' to explicitly "
--- a/src/azure-cli-core/azure/cli/core/_profile.py
+++ b/src/azure-cli-core/azure/cli/core/_profile.py
@@ -879,7 +879,13 @@ class SubscriptionFinder(object):
                 # because user creds went through the 'common' tenant, the error here must be
                 # tenant specific, like the account was disabled. For such errors, we will continue
                 # with other tenants.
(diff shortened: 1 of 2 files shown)


yonzhan:
add to S167

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -879,7 +879,13 @@ def _find_using_common_tenant(self, access_token, resource):
                 # because user creds went through the 'common' tenant, the error here must be
                 # tenant specific, like the account was disabled. For such errors, we will continue
                 # with other tenants.
-                logger.warning("Failed to authenticate '%s' due to error '%s'", t, ex)
+                msg = (getattr(ex, 'error_response', None) or {}).get('error_description') or ''
```
arrownj: why not use `getattr(ex, 'error_response', {})` and `get('error_description', '')` ?

jiasli: Good question. I previously have exactly the same question as you. Because `error_response` can be `None`, instead of not-exist. In that case, `getattr(ex, 'error_response', {})` returns `None` and cause following `.get` to fail.

jiasli: Same for `get('error_description', '')`, which may return `None` if `error_description` exists, but the value is `None`. `get('error_description') or ''` guarantees the returned value to be at least an empty string.

arrownj: got it. make sense.

Issue #12313, opened 2020-02-21
Command to wait for SQL Server state fails


valterc:
## Describe the bug

Using Azure CLI to wait for a SQL Server state fails with the following stack trace:

**Command Name**
`az sql server wait`

**Errors:**
```
'NoneType' object has no attribute 'casefold'
Traceback (most recent call last):
Temp\pip-install-3pnsb50r\knack\knack\cli.py, ln 206, in invoke
azure\cli\core\commands\__init__.py, ln 608, in execute
azure\cli\core\commands\__init__.py, ln 666, in _run_jobs_serially
azure\cli\core\commands\__init__.py, ln 659, in _run_job
Local\Temp\pip-install-3pnsb50r\six\six.py, ln 703, in reraise
azure\cli\core\commands\__init__.py, ln 636, in _run_job
azure\cli\core\commands\__init__.py, ln 306, in __call__
azure\cli\core\commands\arm.py, ln 684, in handler
AttributeError: 'NoneType' object has no attribute 'casefold'
```

## To Reproduce:

- `az login`
- `az account set --subscription {}`
- `az sql server wait --name {} --resource-group {} --created`

## Expected Behavior
Command waits for a few seconds or returns immediately with no errors.

## Environment Summary
```
Windows-10-10.0.18362-SP0
Python 3.6.6

azure-cli 2.1.0

```

yonzhan:
az sql

a-vishar:
Any update on this?

jaredmoo:
Error:

```
'NoneType' object has no attribute 'casefold'
Traceback (most recent call last):
Temp\pip-install-3pnsb50r\knack\knack\cli.py, ln 206, in invoke
azure\cli\core\commands\__init__.py, ln 608, in execute
azure\cli\core\commands\__init__.py, ln 666, in _run_jobs_serially
azure\cli\core\commands\__init__.py, ln 659, in _run_job
Local\Temp\pip-install-3pnsb50r\six\six.py, ln 703, in reraise
azure\cli\core\commands\__init__.py, ln 636, in _run_job
azure\cli\core\commands\__init__.py, ln 306, in __call__
azure\cli\core\commands\arm.py, ln 684, in handler
AttributeError: 'NoneType' object has no attribute 'casefold'  
```

Code:
              provisioning_state = get_provisioning_state(instance)
                # until we have any needs to wait for 'Failed', let us bail out on this
                if provisioning_state.casefold() == 'failed':
[...]

PR #12532, opened 2020-03-11
{Core} Only lower provisioning_state when it has value

jiasli:
Fix the regression in https://github.com/Azure/azure-cli/pull/12154: `provisioning_state.casefold()` raises exception `AttributeError: 'NoneType' object has no attribute 'casefold'` when `provisioning_state` is `None`.

This causes issue to `az sql server wait` since sql server doesn't use `"provisioningState": "Succeeded"`, but `"state": "Ready"`: https://github.com/Azure/azure-cli/issues/12313

Only lower `provisioning_state` when it has value (not `None` or empty string `''`).

Tested with

```
az vm wait --ids {} --created
```

Also, use `lower()` to be consistent with other code.

### DIFF
--- a/src/azure-cli-core/azure/cli/core/commands/arm.py
+++ b/src/azure-cli-core/azure/cli/core/commands/arm.py
@@ -681,10 +681,12 @@ def _cli_wait_command(context, name, getter_op, custom_command=False, **kwargs):
                     return None
                 provisioning_state = get_provisioning_state(instance)
                 # until we have any needs to wait for 'Failed', let us bail out on this
-                if provisioning_state.casefold() == 'failed':
+                if provisioning_state:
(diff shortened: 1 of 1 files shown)


Juliehzl:
My concern is that although your current change will not break users ay more, the sql wait command still doesn't work now.
I mean because the wait condition is from "state" and "provisionstate" will none, the wait condition will never match as expected for this scenario.

jiasli:
> My concern is that although your current change will not break users ay more, the sql wait command still doesn't work now.
> I mean because the wait condition is from "state" and "provisionstate" will none, the wait condition will never match as expected for this scenario.

I have put my thought in https://github.com/Azure/azure-cli/issues/12313#issuecomment-597423556. But anyway, let's fix the regression first.

Issue #11267, opened 2019-11-18
[KeyVault] Support softDeleteRetentionInDays


jiasli:
Copied from https://github.com/Azure/azure-powershell/issues/10556

Support making purge protection on/off as a one time only operation (already did), and soft delete retention configurable by number of days (from 7 to 90).

fengzhou-msft:
@jiasli please evaluate this request.

PR #12542, opened 2020-03-11
[KeyVault] Support --retention-days

jiasli:
Fix https://github.com/Azure/azure-cli/issues/11267: Support softDeleteRetentionInDays, configurable by number of days from 7 to 90

Support `--retention-days` to configure `softDeleteRetentionInDays` during `keyvault create/update`.

The description of `--enable-soft-delete` is also updated to be consistent with the [REST spec](https://github.com/Azure/azure-rest-api-specs/blob/e9428e8594204d2bb9cdbcbf3e35992b879df6b4/specification/keyvault/resource-manager/Microsoft.KeyVault/stable/2019-09-01/keyvault.json#L1126).

Though the CI passes because msrest can build objects even when keys are missing from SDK, the actually command won't work, as the SDK on PyPI doesn't have `softDeleteRetentionInDays`. 

ℹ msrest can handle additional properties with `msrest.serialization.Deserializer._build_additional_properties`.


src/azure-cli/azure/cli/command_modules/keyvault/_params.py
```diff
@@ -88,7 +87,12 @@ class CLIJsonWebKeyOperation(str, Enum):
         c.argument('sku', arg_type=get_enum_type(SkuName, default=SkuName.standard.value))
         c.argument('no_self_perms', arg_type=get_three_state_flag(), help="Don't add permissions for the current user/service principal in the new vault.")
         c.argument('location', validator=get_default_location_from_resource_group)
-        c.argument('enable_soft_delete', arg_type=get_three_state_flag(), help='Enable vault deletion recovery for the vault, and all contained entities. If omitted, assume true as default value.')
+        c.argument('enable_soft_delete', arg_type=get_three_state_flag(), help="Enable 'soft delete' functionality for this key vault and all contained entities. If omitted, it will be set to true by default. Once set to true, it cannot be reverted to false.")
```
jiasli: Instead of previously assigning `enable_soft_delete` to both `create` and `update` under `keyvault`, and then override the one in `create`, I separated `create` and `update` so that there is no overriding relationship for `enable_soft_delete` and `retention_days` .

src/azure-cli/azure/cli/command_modules/keyvault/tests/latest/test_keyvault_commands.py
```diff
@@ -216,27 +219,23 @@ def test_keyvault_mgmt(self, resource_group):
...
-        ])
-
-        # test enableSoftDelete is True if omitted
-        self.cmd('keyvault create -g {rg} -n {kv3} -l {loc} --enabled-for-deployment true '
-                 '--enabled-for-disk-encryption true --enabled-for-template-deployment true',
+        self.cmd('keyvault create -g {rg} -n {kv2} -l {loc} '  # enableSoftDelete is True if omitted
+                 '--retention-days 7 '
+                 '--sku premium '
+                 '--enabled-for-deployment true --enabled-for-disk-encryption true --enabled-for-template-deployment true '
+                 '--no-self-perms ',  # no self permission assigned
```
jiasli: The old 3 tests for **permission**, **sku** and **soft delete** are combined to one test to save running time.

PR #12543, opened 2020-03-11
[Core] Add experimental tag

jiasli:
This PR requires https://github.com/microsoft/knack/pull/180

`is_experimental=True` flag can be used (same as `is_preview`) with 

1. Command groups
2. Commands
3. Arguments


yonzhan:
add to S167

jiasli:
Changed the behavior to make `is_preview` and `is_experimental` mutual-exclusive. If the developer uses both, CLI will errror out:

> Error loading command module 'profile': Failed to register command 'account list-locations', is_preview and is_experimental can't be true at the same time

PR #12544, opened 2020-03-11
[Core] Support --only-show-errors

jiasli:
This PR requires https://github.com/microsoft/knack/pull/179

There are 3 ways to achieve this effect:

- Use `--only-show-errors` as command parameter
- Use .azure/config 
    ```
    [core]
    only_show_errors=True
    ```
- Use env var `AZURE_CORE_ONLY_SHOW_ERRORS=True`


yonzhan:
add to S167

PR #12547, opened 2020-03-11
{Core} Fix is_preview for commands

jiasli:
When a command group is marked as preview:

```py
    with self.command_group('vm', compute_vm_sdk, is_preview=True) as g:
```

during invocation, the warning message isn't very accurate - what is invoked is actually a command, not a command group.

```sh
> az vm list
This command group is in preview. It may be changed/removed in a future release.
```

This PR corrects the behavior by removing the inheritance of `preview_info` from command group to command, so that the implicit preview resolution is activated - it searches `preview_info` from the command to its parent group, then the parent group's parent group (if any). Otherwise, preview resolution stops at the command level (L711).

https://github.com/Azure/azure-cli/blob/d7d7a25236f360fbbb3b799b18f46d6cc206cd20/src/azure-cli-core/azure/cli/core/commands/__init__.py#L709-L725

Now it shows a more accurate warning following Knack's behavior:

```sh
> az vm list
Command group 'vm' is in preview. It may be changed/removed in a future release.
```

Also, this PR fixes `show_command`, `wait_command` and `generic_update_command` where `is_preview` doesn't work before.


yonzhan:
add to S167

PR #12554, opened 2020-03-11
[ACR] 'az acr agentpool': new feature

rosanch:
New ACR feature AgentPool

This allows the client to create their own agent pool and customize the resources. This also allows the use of Tasks inside a VNET


src/azure-cli/azure/cli/command_modules/acr/agentpool.py
```diff
@@ -0,0 +1,127 @@
...
+
+def acr_agentpool_show(cmd,
+                       client,
+                       agentpool_name,
+                       registry_name,
+                       resource_group_name=None,
+                       queue_count=False):
+    _, resource_group_name = validate_managed_registry(
+        cmd, registry_name, resource_group_name)
+    if queue_count:
```
rosanch: Couldn't decide if it should be a separate command or just a flag

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
```diff
@@ -0,0 +1,127 @@
...
+    agentpool_create_paramters = AgentPoolUpdateParameters(
+        tags=tags,
+        count=count
+    )
+
+    try:
+        return client.update(resource_group_name=resource_group_name,
+                             registry_name=registry_name,
+                             agent_pool_name=agentpool_name,
+                             tags=tags,
```
rosanch: should update take the AgentPoolParameters DataModel instead? 
I think we should follow the same format for tasks, plus it will make it easier to modify and maintain in the future.

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -316,6 +316,18 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('password1', options_list=['--password1'], help='Flag indicating if first password should be deleted', action='store_true', required=False)
         c.argument('password2', options_list=['--password2'], help='Flag indicating if second password should be deleted.', action='store_true', required=False)
 
+    with self.argument_context('acr agentpool') as c:
+        c.argument('registry_name', options_list=['--registry', '-r'])
+        c.argument('agentpool_name', options_list=['--name', '-n'], help='The name of the agentpool.')
+        c.argument('tag', help='The tag of the agentpool.')
+        c.argument('count', type=int, help='The count of the agentpool.')
+        c.argument('tier', help='The tier of the agentpool.')
```
rosanch: shouldn't we make a data model for the agent pool tiers just like ACR has for the SKU?

src/azure-cli/azure/cli/command_modules/acr/_format.py
```diff
@@ -216,6 +220,14 @@ def _taskrun_format_group(item):
                                    _get_value(item, 'runResult', 'finishTime')))
     ])
 
+def _agentpool_format_group(item):
+    return OrderedDict([
+        ('NAME', _get_value(item, 'name')),
+        ('STATE', _get_value(item, 'provisioningState')),
+        ('OS', _get_value(item, 'os')),
```
huanwu: os is in properties, not in root. So you might need write as ('OS', _get_value(item, 'properties', 'os')),

huanwu:
Need update schedulerun to add agentpoolname parameter so we can schedule a run on dedicated pool.

yonzhan:
add to S167

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -1037,6 +1037,74 @@
...
 """
 
+helps['acr agentpool'] = """
+type: group
+short-summary: Manage private Tasks agent pools with Azure Container Registries.
+"""
+
+helps['acr agentpool create'] = """
+type: command
+short-summary: Creates an agent pool for an Azure Container Registry.
```
qwordy: Creates -> Create

src/azure-cli/azure/cli/command_modules/acr/commands.py
```diff
@@ -294,3 +302,11 @@ def load_command_table(self, _):  # pylint: disable=too-many-statements
 
     with self.command_group('acr token credential', acr_token_credential_generate_util) as g:
         g.command('generate', 'acr_token_credential_generate')
+
+    with self.command_group('acr agentpool', acr_agentpool_util, is_preview=True) as g:
+        g.command('create', 'acr_agentpool_create')
+        g.command('update', 'acr_agentpool_update')
+        g.command('delete', 'acr_agentpool_delete')
+        g.command('list', 'acr_agentpool_list')
+        g.command('show', 'acr_agentpool_show')
```
qwordy: How about g.show_command()?

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -157,6 +157,7 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('values', help="The task values file path relative to the source context.")
         c.argument('set_value', options_list=['--set'], help="Value in 'name[=value]' format. Multiples supported by passing --set multiple times.", action='append', validator=validate_set)
         c.argument('set_secret', help="Secret value in '--set name[=value]' format. Multiples supported by passing --set multiple times.", action='append', validator=validate_set_secret)
+        c.argument('agent_pool_name', options_list=['--agent-pool'], help='The name of the agentpool.')
```
northtyphoon: later, can you add a preview flag for the agent_pool_name? Also mark the agent_pool command group as preview.

northtyphoon:
@djyou  @shahzzzam @juliusl

src/azure-cli/azure/cli/command_modules/acr/_params.py
rosanch: done

src/azure-cli/azure/cli/command_modules/acr/_format.py
rosanch: Right now is working, I do see that it shows as part of the properties but when I call the get method in sdk I get this response. 
{
  "count": 1,
  "id": "/subscriptions/3647315e-0c5b-4ce4-8739-b071e144b2c9/resourceGroups/roywusdf/providers/Microsoft.ContainerRegistry/registries/royagent/agentPools/rosanchagentpool",
  "location": "WESTUS",
  "name": "rosanchagentpool",
  "os": "Linux",
  "provisioningState": "Succeeded",
  "resourceGroup": "roywusdf",
  "tags": null,
  "tier": "S1",
  "type": "Microsoft.ContainerRegistry/registries/agentPools",
  "virtualNetworkSubnetResourceId": null
}

src/azure-cli/azure/cli/command_modules/acr/commands.py
rosanch: done. Do you know what is the difference? they both seem to work when I tested.

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -1037,6 +1037,74 @@
...
+"""
+
+helps['acr agentpool create'] = """
+type: command
+short-summary: Create an agent pool for an Azure Container Registry.
+examples:
+  - name: Create the agent pool 'MyAgentName' associated with the registry 'MyRegistry'.
+    text: >
+        az acr agentpool create -n MyAgentName -r MyRegistry
+  - name: Create the agent pool 'MyAgentName' with n agent count.
```
shahzzzam: lets give a concrete number because this is an example?

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -1037,6 +1037,74 @@
...
+    text: >
+        az acr agentpool create -n MyAgentName -r MyRegistry --count n
+  - name: Create the agent pool 'MyAgentName' associated with the registry 'MyRegistry' in VNET 'MyVnetId'.
+    text: >
+        az acr agentpool create -n MyAgentName -r MyRegistry --vnet-id MyVnetId
```
shahzzzam: Should this be SubnetID? @northtyphoon

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -1037,6 +1037,74 @@
...
+"""
+
+helps['acr agentpool update'] = """
+type: command
+short-summary: Update an agent pool for an Azure Container Registry.
+examples:
+  - name: Update the agent pool 'MyAgentName' count
+    text: >
+        az acr agentpool update -n MyAgentName -r MyRegistry --tag NewCount
```
shahzzzam: confused. What is a tag?

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -1037,6 +1037,74 @@
...
+        az acr agentpool delete -n MyAgentName -r MyRegistry
+"""
+
+helps['acr agentpool list'] = """
+type: command
+short-summary: List the agent pools for an Azure Container Registry.
+examples:
+  - name: List agent pools and show the result in a table.
+    text: >
+        az acr agentpool list -r MyRegistry
```
shahzzzam: Does this show as a table? I believe to show it as a table you need `-o table`

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
```diff
@@ -0,0 +1,127 @@
...
+                         resource_group_name=None,
+                         tags=None,
+                         count=None):
+
+    _, resource_group_name = validate_managed_registry(
+        cmd, registry_name, resource_group_name)
+
+    AgentPoolUpdateParameters = cmd.get_models('AgentPoolUpdateParameters')
+
+    agentpool_create_paramters = AgentPoolUpdateParameters(
```
shahzzzam: nit: agentpool_update_parameters*

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
```diff
@@ -0,0 +1,127 @@
...
+                         tier=DEFAULT_TIER,
+                         os_type=DEFAULT_OS,
+                         vnet_id=None):
+
+    registry, resource_group_name = get_registry_by_name(
+        cmd.cli_ctx, registry_name, resource_group_name)
+
+    AgentPool = cmd.get_models('AgentPool')
+
+    agentpool_create_paramters = AgentPool(
```
shahzzzam: nit:parameters*

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
shahzzzam: What do you do with this object?

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
```diff
@@ -0,0 +1,127 @@
...
+                         client,
+                         agent_pool_name,
+                         registry_name,
+                         resource_group_name=None):
+
+    _, resource_group_name = validate_managed_registry(
+        cmd, registry_name, resource_group_name)
+
+    try:
+        return client.update(resource_group_name=resource_group_name,
```
shahzzzam: delete?

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
shahzzzam: a flag is fine IMO

src/azure-cli/azure/cli/command_modules/acr/_params.py
shahzzzam: My assumption is that SKU is an Enum, and ACR does not model it as string. Downside of Enums are if they want to add any new SKU's they will have to go thru new API version. @northtyphoon can correct me if I am wrong

src/azure-cli/azure/cli/command_modules/acr/_help.py
shahzzzam: @rosanch Can you give a valid Subnet ResourceID in the example? This is too vague.

src/azure-cli/azure/cli/command_modules/acr/_help.py
shahzzzam:
```
examples:
  - name: Scale up my agentpool to 3 agents.
    text: >
        az acr agentpool update -n MyAgentName -r MyRegistry --count 3
  - name: Scale down my agentpool to 0 agents.
    text: >
        az acr agentpool update -n MyAgentName -r MyRegistry --count 0

```

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
shahzzzam: remove this

src/azure-cli/azure/cli/command_modules/acr/_format.py
```diff
@@ -216,6 +220,14 @@ def _taskrun_format_group(item):
                                    _get_value(item, 'runResult', 'finishTime')))
     ])
 
+def _agentpool_format_group(item):
```
northtyphoon: The instance count will be very useful. Can you add to the list?

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
northtyphoon: should it be delete?

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
```diff
@@ -0,0 +1,127 @@
...
+                         tags=None,
+                         count=None):
+
+    _, resource_group_name = validate_managed_registry(
+        cmd, registry_name, resource_group_name)
+
+    AgentPoolUpdateParameters = cmd.get_models('AgentPoolUpdateParameters')
+
+    agentpool_create_paramters = AgentPoolUpdateParameters(
+        tags=tags,
```
huanwu: The agentpoolupdateparameter only contains count no tags.

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
```diff
@@ -0,0 +1,127 @@
...
+        count=count
+    )
+
+    try:
+        return client.update(resource_group_name=resource_group_name,
+                             registry_name=registry_name,
+                             agent_pool_name=agent_pool_name,
+                             tags=tags,
```
huanwu: I think you need pass in AgentPoolUpdateParameters here by the RP definition.

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
huanwu: Yes, refer https://msazure.visualstudio.com/AzureContainerRegistry/_git/Build?path=%2Fsrc%2FACR.BuildRP%2Fsrc%2Fv18_09_GA%2FControllers%2FAgentPoolsController.cs&version=GBmaster&line=241&lineEnd=242&lineStartColumn=1&lineEndColumn=1&lineStyle=plain

src/azure-cli/azure/cli/command_modules/acr/_help.py
northtyphoon: @huanwu it reminds me since agent pool is tracked resource, do we need to do anything in rp to support tags? same question for task run.

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
northtyphoon: @huanwu  ^^

huanwu: I see. But we need remove the tag parameter from the update method.

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
huanwu: It seems the SDK generate that. We just need remove the tag from the input.

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
```diff
@@ -0,0 +1,127 @@
...
+def acr_agentpool_show(cmd,
+                       client,
+                       agent_pool_name,
+                       registry_name,
+                       resource_group_name=None,
+                       queue_count=False):
+    _, resource_group_name = validate_managed_registry(
+        cmd, registry_name, resource_group_name)
+    if queue_count:
+        return client.get_queue_status(resource_group_name, registry_name, agent_pool_name)
```
huanwu: Why this method support queue_count here? I think you already had another method 'acr_agentpool_show_queue' to show the queue count.

src/azure-cli/azure/cli/command_modules/acr/_help.py
huanwu: I double checked the swagger and seems the swagger will always generate the arm resource tags for the UpdateParameters including task and taskrun. What kind of scenario will we use the tags?

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -324,6 +330,18 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('password1', options_list=['--password1'], help='Flag indicating if first password should be deleted', action='store_true', required=False)
         c.argument('password2', options_list=['--password2'], help='Flag indicating if second password should be deleted.', action='store_true', required=False)
 
+    with self.argument_context('acr agentpool') as c:
+        c.argument('registry_name', options_list=['--registry', '-r'])
+        c.argument('agent_pool_name', options_list=['--name', '-n'], help='The name of the agentpool.')
+        c.argument('tag', help='The tag of the agentpool.')
+        c.argument('count', type=int, help='The count of the agentpool.')
+        c.argument('tier', help='The tier of the agentpool.')
```
huanwu: Should we add more description about the tier here e.g S1 maps to D2_V3 (2CPU, 4GB Memory)

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -1037,6 +1037,74 @@
...
+helps['acr agentpool update'] = """
+type: command
+short-summary: Update an agent pool for an Azure Container Registry.
+examples:
+  - name: Update the agent pool 'MyAgentName' count
+    text: >
+        az acr agentpool update -n MyAgentName -r MyRegistry --tag NewCount
+  - name: Update the agent pool 'MyAgentName' tag
+    text: >
+        az acr agentpool update -n MyAgentName -r MyRegistry --tag MyTag
```
huanwu: The update is to update the count of agentpool. e.g. az acr agentpool update -n MyAgentName -r MyrRegistry --count 3

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -1037,6 +1037,74 @@
...
+        az acr agentpool list -r MyRegistry
+"""
+
+helps['acr agentpool show'] = """
+type: command
+short-summary: Get the properties of a specified agent pool for an Azure Container Registry.
+examples:
+  - name: Get the properties of an agent pool, displaying the results in a table.
+    text: >
+        az acr agentpool show -n MyAgentName -r MyRegistry -o table
```
huanwu: I feel -o table is not necessary for acr agentpool show, which actually won't show everything.

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -1037,6 +1037,74 @@
...
+"""
+
+helps['acr agentpool show-queue'] = """
+type: command
+short-summary: Get the agent pool queue count.
+examples:
+  - name: Get the agent pool queue count, displaying only the value.
+    text: >
+        az acr agentpool show -n MyAgentName -r MyRegistry -o tsv
```
huanwu: Is this sample right? It doesn't get the agent pool queue count. And from what you defined in commands, you need call az acr gentpool show-queue to get the queue count.

src/azure-cli/azure/cli/command_modules/acr/_params.py
huanwu: Or we need a option_list here too.

huanwu:
And can we add agentpoolname filter in acr_task_list_runs?

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
rosanch: @shahzzzam @huanwu @northtyphoon @juliusl

src/azure-cli/azure/cli/command_modules/acr/_params.py
rosanch: I know that is the downside, but if they already have the SKU like this which is like our tier I think is better to be consistent. I doubt we will be changing the tiers very often anyway
it is an enum
`        c.argument('sku', help='The SKU of the container registry', arg_type=get_enum_type(SkuName))`

and a data model 
`SkuName, PasswordName, DefaultAction, PolicyStatus, WebhookAction, WebhookStatus, TaskStatus, \
        BaseImageTriggerType, RunStatus, SourceRegistryLoginMode, UpdateTriggerPayloadType, TokenStatus = self.get_models(
            'SkuName', 'PasswordName', 'DefaultAction', 'PolicyStatus', 'WebhookAction', 'WebhookStatus',
            'TaskStatus', 'BaseImageTriggerType', 'RunStatus', 'SourceRegistryLoginMode', 'UpdateTriggerPayloadType',
            'TokenStatus')
`

src/azure-cli/azure/cli/command_modules/acr/_help.py
rosanch: I was confused as well but I saw it in both the data model for update and create. So I set the variable I can hide for now with the deprecate functionally to block it

src/azure-cli/azure/cli/command_modules/acr/_help.py
rosanch: the two parameters that the SDK accepts are count and tag. But like I said I'll be hiding all the code related to tag for now

src/azure-cli/azure/cli/command_modules/acr/_help.py
rosanch: good catch thanks

src/azure-cli/azure/cli/command_modules/acr/_help.py
rosanch: I agree but the examples I saw from other shows are like this. I think is to show the -o functionality. But if you prefer I can just remove it

src/azure-cli/azure/cli/command_modules/acr/_help.py
rosanch: right is wrong the command should be show-queue. But I'm removing this command anyway and leaving only show with a flag for this

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
rosanch: yes, the model is not being used I just wanted to have the discussion. Because it was weird to me that we have an update parameters but we have to pass the values raw to the Update function.
I left a comment on the PR exactly for this line. Shouldn't we pass the updateParameters model to the update function like in tasks? So if the future the capability of update increases we only modify the data model and not the update function parameters? To be more consistant and it looks cleaner to me as well.

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
rosanch: my bad I did have the delete method I must have overwrite when cleaning the code.

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
rosanch: I wanted to see which one you all prefer I left comment to point that out in the PR. I'll be removing this one

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
rosanch: for example. Right now we want to remove tag from update. by doing so this might break the code because we're passing a tag value that is no longer needed but if we added a tag value to the datamodel it would just be ignored, and it will not break the code.

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
rosanch: this happened because of the optimization right?

src/azure-cli/HISTORY.rst
```diff
@@ -135,6 +135,7 @@ Release History
 * Add a new argument `--expose-token` for `az acr login`
 * Fix the incorrect output of `az acr task identity show -n Name -r Registry -o table`
 * az acr login: Throw a CLIError if there are errors returned by docker command
+* Add a new command group 'az acr task agentpool create/update/show/list/delete' for taskrun management
```
yungezz: history.rst file is automatically generated from PR title now, no need to manual update

yungezz:
LGTM, minor comments

src/azure-cli/HISTORY.rst
rosanch: thanks

rosanch: thanks

src/azure-cli/azure/cli/command_modules/acr/_help.py
shahzzzam: Lets remove it from show. 

```suggestion
        az acr agentpool show -n MyAgentName -r MyRegistry 
```

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
```diff
@@ -0,0 +1,112 @@
...
+                         resource_group_name=None):
+
+    _, resource_group_name = validate_managed_registry(
+        cmd, registry_name, resource_group_name)
+
+    try:
+        response = client.delete(resource_group_name=resource_group_name,
+                                 registry_name=registry_name,
+                                 agent_pool_name=agent_pool_name)
+        if response.status_code == 404:
```
shahzzzam: Borrowed from Saravanan's PR;
```
/* Since agent pool is a tracked resource in arm,
                                 * arm, as well, pings the async deletion api at the
                                 * same time to get the status. If arm gets the 200 status
                                 * first and knows that the resource is deleted, it marks the
                                 * resource as deleted and stop routing further requests
                                 * to the resource including the async deletion status
                                 * api. Hence arm will directly return 404. Consider this as
                                 * successful delete
                                 */
```

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -324,6 +330,17 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('password1', options_list=['--password1'], help='Flag indicating if first password should be deleted', action='store_true', required=False)
         c.argument('password2', options_list=['--password2'], help='Flag indicating if second password should be deleted.', action='store_true', required=False)
 
+    with self.argument_context('acr agentpool') as c:
+        c.argument('registry_name', options_list=['--registry', '-r'])
+        c.argument('agent_pool_name', options_list=['--name', '-n'], help='The name of the agent pool.')
+        c.argument('count', type=int, help='The count of the agent pool.')
+        c.argument('tier', help='Sets the VM your agent pool will run on. Valid values are: S1(Standard_D2_v3), S2(Standard_D4_v3) or S3(Standard_D8_v3)')
```
northtyphoon: S1 (2 cpus)
S2 (4 cpus)
S3 (8 cpus)

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -324,6 +330,17 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
...
+        c.argument('os_type', options_list=['--os'], help='The os of the agent pool.', deprecate_info=c.deprecate(hide=True))
+        c.argument('vnet_id', options_list=['--vnet-id'], help='The Virtual Network Subnet Resource Id of the agent machine.')
```
northtyphoon: might use `subnet` to be consist with existing acr commands

shahzzzam: Agreed! Less confusion.

src/azure-cli/azure/cli/command_modules/acr/_params.py
shahzzzam: Sets the VM tier*

@northtyphoon , we should probably also mention memory to be thorough
S1 (2 cpus, 3 MB RAM)
S2 (4 cpus, 8 MB RAM)
S3 (8 cpus, 16 MB RAM)

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
```diff
@@ -0,0 +1,112 @@
...
+        cmd.cli_ctx, registry_name, resource_group_name)
+
+    AgentPool = cmd.get_models('AgentPool')
+
+    agentpool_create_parameters = AgentPool(
+        location=registry.location,
+        count=count,
+        tier=tier.upper(),
+        os=os_type,
+        VirtualNetworkSubnetResourceId=vnet_id
```
northtyphoon: Is it supposed to be `virtual_network_subnet_resource_id`

https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/containerregistry/azure-mgmt-containerregistry/azure/mgmt/containerregistry/v2019_06_01_preview/models/_models_py3.py#L159

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
```diff
@@ -0,0 +1,112 @@
...
+
+    registry, resource_group_name = get_registry_by_name(
+        cmd.cli_ctx, registry_name, resource_group_name)
+
+    AgentPool = cmd.get_models('AgentPool')
+
+    agentpool_create_parameters = AgentPool(
+        location=registry.location,
+        count=count,
+        tier=tier.upper(),
```
yungezz: any validation on tier value?

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
```diff
@@ -0,0 +1,112 @@
...
+
+    _, resource_group_name = validate_managed_registry(
+        cmd, registry_name, resource_group_name)
+
+    try:
+        response = client.delete(resource_group_name=resource_group_name,
+                                 registry_name=registry_name,
+                                 agent_pool_name=agent_pool_name)
+        if response.status_code == 404:
+            response.status_code = 200
```
yungezz: could you pls share more context why 404 but map it to 200 for user?

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -1040,6 +1040,63 @@
...
+        az acr agentpool create -n MyAgentName -r MyRegistry --vnet-id /subscriptions/{SubscriptionId}/resourceGroups/{ResourceGroupName}/providers/Microsoft.ClassicNetwork/virtualNetworks/<myNetwork>/subnets/<subNet>
+"""
+
+helps['acr agentpool update'] = """
+type: command
+short-summary: Update an agent pool for an Azure Container Registry.
+examples:
+  - name: Update the agent pool 'MyAgentName' count to 5
+    text: >
+        az acr agentpool update -n MyAgentName -r MyRegistry --count 5
```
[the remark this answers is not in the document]
rosanch: I thought about it, but wasn't sure if I should because -c for context
@northtyphoon  @huanwu what do you tink?

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
rosanch: I'm letting the SDK side handle that. Output error is already descriptive

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
rosanch: Yes there seems to be a timing issue between arm and our rp, and it sometimes return 404 when deleted. Although I'm changing the workarround please ignore this.

rosanch: I'm adding this comment for context in next solution:
/* Since agent pool is a tracked resource in arm,
                                 * arm, as well, pings the async deletion api at the
                                 * same time to get the status. If arm gets the 200 status
                                 * first and knows that the resource is deleted, it marks the
                                 * resource as deleted and stop routing further requests
                                 * to the resource including the async deletion status
                                 * api. Hence arm will directly return 404. Consider this as
                                 * successful delete
                                 */

src/azure-cli/azure/cli/command_modules/acr/commands.py
qwordy: https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#register-commands

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
```diff
@@ -0,0 +1,129 @@
...
+    _, resource_group_name = validate_managed_registry(
+        cmd, registry_name, resource_group_name)
+
+    try:
+        response = client.delete(resource_group_name=resource_group_name,
+                                 registry_name=registry_name,
+                                 agent_pool_name=agent_pool_name)
+
+        if no_wait:
+            return client.get(resource_group_name, registry_name, agent_pool_name)
```
northtyphoon: in theory, the delete can return 200 which indicates the agent pool has been successfully removed. if you call client.get, it will get 404 error. Is it why you want to wrap response in the longrunningoperation initially?

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
```diff
@@ -0,0 +1,129 @@
...
+
+        #Since agent pool is a tracked resource in arm, arm also pings the async deletion api at the
+        #same time to get the status. If arm gets the 200 status first and knows that the resource is deleted,
+        #it marks the resource as deleted and stop routing further requests to the resource including the
+        #async deletion status api. Hence arm will directly return 404. Consider this as successful delete.
+        from ._agentpool_polling import delete_agentpool_with_polling
+        delete_agentpool_with_polling(cmd, client, agent_pool_name, registry_name, resource_group_name)
+        return response
```
northtyphoon: in this case, can you directly return the new agent_poller from `delete_agentpool_with_polling`

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
northtyphoon: Ok, I think you can just do something like 
```
        if no_wait:
            logger.warning("Started to delete the agent pool: %s", response.status())
            return
```

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
```diff
@@ -0,0 +1,129 @@
...
+                         agent_pool_name,
+                         registry_name,
+                         no_wait=False,
+                         resource_group_name=None):
+
+    _, resource_group_name = validate_managed_registry(
+        cmd, registry_name, resource_group_name)
+
+    try:
+        response = client.delete(resource_group_name=resource_group_name,
```
northtyphoon: Can you add a prompt to confirm the deletion? check registry deletion

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
northtyphoon: Yeah, I think you can just return `delete_agentpool_with_polling`

src/azure-cli/azure/cli/command_modules/acr/_agentpool_polling.py
```diff
@@ -0,0 +1,95 @@
...
+    def finished(self):
+        return self.operation_status in get_finished_agentpool_status(self._cmd)
+
+    def resource(self):
+        return self.operation_result
+
+    def _set_operation_status(self, response):
+        AgentPoolStatus = self._cmd.get_models('ProvisioningState')
+        if response.status_code == 200 or response.status_code == 404:
+            self.operation_result = self._deserialize(response)
```
northtyphoon: maybe just set to null, otherwise, user may see something like 
{
  "count": null,
  "error": {
    "code": "ResourceNotFound",
    "message": "The Resource 'Microsoft.ContainerRegistry/registries/binduwestus/agentPools/agent2' under resource group 'bindu' was not found."
  },
  "id": null,
  "location": null,
  "name": null,
  "os": null,
  "provisioningState": null,
  "tags": null,
  "tier": null,
  "type": null,
  "virtualNetworkSubnetResourceId": null
}

src/azure-cli/azure/cli/command_modules/acr/agentpool.py
rosanch: good catch. 
This was for the original version of the command. Behavior was a little bit different

src/azure-cli/azure/cli/command_modules/acr/_agentpool_polling.py
rosanch: yes that's exactly the case that's why I'm not passing the response of the polling

northtyphoon: actually you might not be able to set none, my intention is just hide the resourcenotfound code and return an alternative result as long as it indicates succeeded.

qwordy:
Is this PR ready to merge?

yungezz:
> Is this PR ready to merge?
hi @qwordy thanks for driving :) we're confirming the release plan, this one is not ready to merge now.

yungezz:
put the PR is next release per release plan. Will remove Do Not Merge label after tomorrow's CC.

rosanch:
@qwordy we're waiting for another chance to be released. We can merge after 3/25

rosanch:
@qwordy We're ready to merge

rosanch:
@yungezz We're ready to merge. Thanks

yungezz:
hi @rosanch 3/25 build will be cut at tomorrow, will merge it after cutting so that your change will not be in this release.

src/azure-cli/azure/cli/command_modules/acr/_format.py
```diff
@@ -217,6 +221,17 @@ def _taskrun_format_group(item):
     ])
 
 
+def _agentpool_format_group(item):
+    return OrderedDict([
+        ('NAME', _get_value(item, 'name')),
+        ('COUNT', _get_value(item, 'count')),
```
rosanch: @qwordy is there a way I can make the table format to write value zero instead of a leaving it blank? I didn't find anything in the documentation.

rosanch: cc: @yungezz

rosanch: never mind. Fixed

PR #12556, opened 2020-03-12
[Core] Add experimental tag for extensions

jiasli:
An extension can now be marked as **experimental** by adding `"azext.isExperimental": true` to `azext_metadata.json`:

```jsonc
{
    "azext.minCliCoreVersion": "2.0.46",
    "azext.isPreview": true,
    "azext.isExperimental": true   // Add here
}
```

## When it appears

Same as **preview**, **experimental** appears when 

### Installing the extension from index

```sh
> az extension add -n webapp
The installed extension 'webapp' is experimental and not covered by customer support. Please use with discretion.
```

### Installing the extension from wheel 

This is newly added, also applies to preview

```sh
> az extension add -s "webapp-0.2.24-py2.py3-none-any.whl"
The installed extension 'webapp' is experimental and not covered by customer support. Please use with discretion.
```

### Command help

```sh
> az webapp container up -h
This command is from the following extension: webapp
The extension is experimental and not covered by customer support. Please use with discretion.
```

ℹ If `true`, `azext.isExperimental` will override `azext.isPreview`
ℹ Experimental/Preview warning doesn't appear when invoking a command within an experimental/preview extension

## Change of existing commands

For `extension list-available`, `extension` property is added to each entry.

```jsonc
[
  {
    "experimental": false,   // New
    "installed": false,
    "name": "vm-repair",
    "preview": false,
    "summary": "Auto repair commands to fix VMs.",
    "version": "0.2.6"
  },
  {
    "experimental": false,   // New
    "installed": true,
    "name": "webapp",
    "preview": true,
    "summary": "Additional commands for Azure AppService.",
    "version": "0.2.24"
  },
  ...
]
```

For `extension list`, `preview` and `extension` properties are added to each entry. 

```jsonc
> az extension list
[
  {
    "experimental": false,   // New
    "extensionType": "whl",
    "name": "support",
    "preview": true,         // New
    "version": "0.1.1"
  },
  {
    "experimental": true,    // New
    "extensionType": "whl",
    "name": "webapp",
    "preview": true,         // New
    "version": "0.2.24"
  },
  ...
]
```

### DIFF
--- a/src/azure-cli-core/azure/cli/core/_help.py
+++ b/src/azure-cli-core/azure/cli/core/_help.py
@@ -115,7 +115,7 @@ class CLIPrintMixin(CLIHelp):
         if isinstance(help_file.command_source, ExtensionCommandSource):
             logger.warning(help_file.command_source.get_command_warn_msg())
             if help_file.command_source.preview:
-                logger.warning(help_file.command_source.get_preview_warn_msg())
+                logger.warning(help_file.command_source.get_preview_experimental_warn_msg())
 
 
 class AzCliHelp(CLIPrintMixin, CLIHelp):
--- a/src/azure-cli-core/azure/cli/core/_help.py
+++ b/src/azure-cli-core/azure/cli/core/_help.py
@@ -114,8 +114,11 @@ class CLIPrintMixin(CLIHelp):
             return
         if isinstance(help_file.command_source, ExtensionCommandSource):
             logger.warning(help_file.command_source.get_command_warn_msg())
-            if help_file.command_source.preview:
-                logger.warning(help_file.command_source.get_preview_experimental_warn_msg())
+            # If experimental is true, it overrides preview
+            if help_file.command_source.experimental:
+                logger.warning(help_file.command_source.get_experimental_warn_msg())
+            elif help_file.command_source.preview:
+                logger.warning(help_file.command_source.get_preview_warn_msg())
 
 
 class AzCliHelp(CLIPrintMixin, CLIHelp):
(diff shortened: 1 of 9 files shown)


src/azure-cli-core/azure/cli/core/_help.py
```diff
@@ -115,7 +115,7 @@ def _print_extensions_msg(help_file):
         if isinstance(help_file.command_source, ExtensionCommandSource):
             logger.warning(help_file.command_source.get_command_warn_msg())
             if help_file.command_source.preview:
-                logger.warning(help_file.command_source.get_preview_warn_msg())
+                logger.warning(help_file.command_source.get_preview_experimental_warn_msg())
```
MyronFanQiu: a little bit confused by this line.

jiasli: This function previously only gets the preview warning message. Now it gets both. Just renaming it to reflect the real functionality.

MyronFanQiu: but the condition is `if help_file.command_source.preview:`. Did I miss something?

jiasli: My fault. Will fix it.

yonzhan:
add to S167

jiasli:
> LGTM. The overiding feature seems reasonable. It's interesting to think that if more tags need to be added, how we change the design and existing code.

This is the tricky part because we need to keep `azext.isPreview` so that old extensions can still be installed while adding more tags. But luckily, it is not very likely to require more tags (hopefully).

PR #12563, opened 2020-03-12
[Compute] az vm update: Support --workspace

MyronFanQiu:
**History Notes:**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
@@ -943,7 +943,7 @@ def build_vm_linux_log_analytics_workspace_agent(_, vm_name, location):
         'properties': {
             'publisher': 'Microsoft.EnterpriseCloud.Monitoring',
             'type': 'OmsAgentForLinux',
-            'typeHandlerVersion': '1.4',
+            'typeHandlerVersion': '1.0',
             'autoUpgradeMinorVersion': 'true',
             'settings': {
                 'workspaceId': "[reference(parameters('workspaceId'), '2015-11-01-preview').customerId]",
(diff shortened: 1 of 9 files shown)


yonzhan:
add to S167

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -794,9 +803,7 @@ def create_vm(cmd, vm_name, resource_group_name, image=None, size='Standard_DS1_
         master_template.add_secure_parameter('workspaceId', workspace_id)
         if os_type.lower() == 'linux':
             vm_mmaExtension_resource = build_vm_linux_log_analytics_workspace_agent(cmd, vm_name, location)
-            vm_daExtensionName_resource = build_vm_daExtension_resource(cmd, vm_name, location)
             master_template.add_resource(vm_mmaExtension_resource)
-            master_template.add_resource(vm_daExtensionName_resource)
         elif os_type.lower() == 'windows':
             vm_mmaExtension_resource = build_vm_windows_log_analytics_workspace_agent(cmd, vm_name, location)
             master_template.add_resource(vm_mmaExtension_resource)
```
arrownj: seem `Linux` and `windows` have the same logic for mmaExtension now. The if-elif is not necessary now.

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -1127,78 +1135,19 @@ def update_vm(cmd, resource_group_name, vm_name, os_disk=None, disk_caching=None
...
 
+    if workspace is not None:
+        workspace_id = _prepare_workspace(cmd, resource_group_name, workspace)
+        workspace_name = parse_resource_id(workspace_id)['name']
+        _set_log_analytics_workspace_extension(cmd=cmd,
+                                               resource_group_name=resource_group_name,
+                                               vm=vm,
+                                               vm_name=vm_name,
+                                               workspace_name=workspace_name)
+        os_type = vm.storage_profile.os_disk.os_type.value if vm.storage_profile.os_disk.os_type else None
```
arrownj: what if set_extension failed here ?

src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
```diff
@@ -943,7 +943,7 @@ def build_vm_linux_log_analytics_workspace_agent(_, vm_name, location):
         'properties': {
             'publisher': 'Microsoft.EnterpriseCloud.Monitoring',
             'type': 'OmsAgentForLinux',
-            'typeHandlerVersion': '1.4',
+            'typeHandlerVersion': '1.0',
```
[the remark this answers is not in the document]
MyronFanQiu: I want to keep the consistence with portal and powershell. They are using 1.0.

src/azure-cli/azure/cli/command_modules/vm/custom.py
MyronFanQiu: good catch.

src/azure-cli/azure/cli/command_modules/vm/custom.py
MyronFanQiu: If it fail with an error, this command would stop. I think it's ok. If it fail with warning. It's still ok to set data source. I check the logic of `set_extension` and look like it will only fail with an error. Is there any concern?

src/azure-cli/azure/cli/command_modules/vm/custom.py
MyronFanQiu: I double check the logic. Looks like we still need this one since we need to build different templates for different os. Meanwhile, I have another `else` logic for other os type except for `linux` and `windows`.

src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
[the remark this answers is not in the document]
MyronFanQiu: I checked the telemetry. Only 6 subs(including 2 subs I used) are using it. So it should be ok to change this behavior.

Issue #12332, opened 2020-02-24
Certain passwords don't allow you to log-in


SimonTheLeg:
## Describe the bug

**Command Name**
`az login`

**Errors:**
```
az login: error: argument --password/-p: expected one argument
```

## To Reproduce:
- Create a SPP with the following password `-_-PZ4I11Aw&ao*!=%<adjB@Y<%_iFDKAMiRudu!m15mca=Z4A##EUm`
- Try to log-in `az login --service-principal -u <id> -p '-_-PZ4I11Aw&ao*!=%<adjB@Y<%_iFDKAMiRudu!m15mca=Z4A##EUm' --tenant <tenant>`

## Expected Behavior
Login succeeds

## Environment Summary
```
Darwin-19.2.0-x86_64-i386-64bit
Python 3.7.5
Shell: zsh (also tried with bash)

azure-cli 2.0.77 *

```

yonzhan:
add to S167.

qianwens:
@SimonTheLeg , this is a known issue in python that the leading character '-' will confusing the argument parser to make it as an option name. https://github.com/Azure/azure-cli/issues/7054 . You can fix this issue by adding '=' between the option name and value : `az login --username=$azureUserName --password=$azurePassword`

SimonTheLeg:
ah okay. Well that explains why it was such a weird error :D 
But yeah with the workaround not too much of a problem. :) Maybe it would be good idea to add a small message stating this to the cli's output?

qianwens:
Thanks for the advice. The help message is added in sprint 167.

Issue #11985, opened 2020-01-29
What is an OData filter?  


armtuk:
The term 'OData' filter under --filter option for az ad sp list should be a link here to a description of what that means.

This is REALLY important because az ad sp list shows pages and pages and pages of information, most of which isn't relevant, so filtering feels critical.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: e99d25ff-03e2-e015-37f4-a29b59ec50d4
* Version Independent ID: 1a8e88ed-72c7-fbe7-b0ad-68f23b3946cd
* Content: [az ad sp](https://docs.microsoft.com/en-us/cli/azure/ad/sp?view=azure-cli-latest#feedback)
* Content Source: [src/azure-cli/azure/cli/command_modules/role/_help.py](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/role/_help.py)
* Service: **active-directory**
* GitHub Login: @rloutlaw
* Microsoft Alias: **routlaw**

yonzhan:
add to S165.

yonzhan:
add to S168

richeney:
I was really hoping this issue would have an explanation and examples for OData filters!

richeney:
Some old documentation:
<https://docs.microsoft.com/en-us/previous-versions/azure/ad/graph/howto/azure-ad-graph-api-supported-queries-filters-and-paging-options#filter>

Examples: 
```
az ad sp list --filter "displayName eq 'governance'" --output jsonc
az ad sp list --filter "startswith(displayName, 'gov')" --query "[].displayName" --output yaml
```

qianwens:
This issue is fixed in the sprint 167

PR #12567, opened 2020-03-12
{Core} Polish the help messages for Az login -p and Az ad --filter

qianwens:
https://github.com/Azure/azure-cli/issues/12332
https://github.com/Azure/azure-cli/issues/11985

**History Notes:**  

---


yonzhan:
add to S167

src/azure-cli/azure/cli/command_modules/role/_params.py
```diff
@@ -120,7 +120,7 @@ def load_arguments(self, _):
         c.argument('identifier_uri', help='graph application identifier, must be in uri format')
         c.argument('spn', help='service principal name')
         c.argument('upn', help='user principal name, e.g. john.doe@contoso.com')
-        c.argument('query_filter', options_list=['--filter'], help='OData filter')
+        c.argument('query_filter', options_list=['--filter'], help='OData filter, e.g. displayname eq \'test\' and servicePrincipalType eq \'Application\'')
```
jiasli: It'll be great to mention that the query needs to be quotes, like `--filter "displayname eq 'test'"`.

jiasli: Also. We may put the link here for more information: https://docs.microsoft.com/en-us/previous-versions/azure/ad/graph/howto/azure-ad-graph-api-supported-queries-filters-and-paging-options

src/azure-cli/azure/cli/command_modules/profile/_help.py
```diff
@@ -14,10 +14,10 @@
         - name: Log in interactively.
           text: >
             az login
-        - name: Log in with user name and password. This doesn't work with Microsoft accounts or accounts that have two-factor authentication enabled.
+        - name: Log in with user name and password. This doesn't work with Microsoft accounts or accounts that have two-factor authentication enabled. Use -p=secret if first character of password is '-'.
```
jiasli:
```suggestion
        - name: Log in with user name and password. This doesn't work with Microsoft accounts or accounts that have two-factor authentication enabled. Use -p=secret if the first character of password is '-'.
```

src/azure-cli/azure/cli/command_modules/profile/_help.py
```diff
@@ -14,10 +14,10 @@
...
           text: >
             az login -u johndoe@contoso.com -p VerySecret
-        - name: Log in with a service principal using client secret.
+        - name: Log in with a service principal using client secret. Use -p=secret if first character of password is '-'.
```
jiasli:
```suggestion
        - name: Log in with a service principal using client secret. Use -p=secret if first the character of password is '-'.
```

src/azure-cli/azure/cli/command_modules/profile/_help.py
qianwens: ok

src/azure-cli/azure/cli/command_modules/role/_params.py
qianwens: This page is out-of-date so that I prefer not to mention this link here in case it is not accessible one day

src/azure-cli/azure/cli/command_modules/role/_params.py
```diff
@@ -120,7 +120,7 @@ def load_arguments(self, _):
         c.argument('identifier_uri', help='graph application identifier, must be in uri format')
         c.argument('spn', help='service principal name')
         c.argument('upn', help='user principal name, e.g. john.doe@contoso.com')
-        c.argument('query_filter', options_list=['--filter'], help='OData filter')
+        c.argument('query_filter', options_list=['--filter'], help='OData filter, e.g. --filter \"displayname eq \'test\' and servicePrincipalType eq \'Application\'\"')
```
jiasli: We don't need to escape double quote within single quoted string. 

```suggestion
        c.argument('query_filter', options_list=['--filter'], help='OData filter, e.g. --filter "displayname eq \'test\' and servicePrincipalType eq \'Application\'"')
```

src/azure-cli/azure/cli/command_modules/role/_params.py
jiasli: Makes sense! Maybe we can improve this during MS Graph integration.

jiasli:
Shit it!

jiasli:
Ship it!

PR #12572, opened 2020-03-12
[AKS] aks update: Override outbound IPs with input values when update.

JunSun17:
Twin: https://github.com/Azure/azure-cli-extensions/pull/1383

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acs/_loadbalancer.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/_loadbalancer.py
@@ -59,18 +59,22 @@ def configure_load_balancer_profile(managed_outbound_ip_count, outbound_ips, out
     outbound_ip_resources = _get_load_balancer_outbound_ips(outbound_ips)
     outbound_ip_prefix_resources = _get_load_balancer_outbound_ip_prefixes(outbound_ip_prefixes)
 
-    if managed_outbound_ip_count:
-        profile.managed_outbound_ips = ManagedClusterLoadBalancerProfileManagedOutboundIPs(
-            count=managed_outbound_ip_count
-        )
-    if outbound_ip_resources:
-        profile.outbound_ips = ManagedClusterLoadBalancerProfileOutboundIPs(
-            public_ips=outbound_ip_resources
-        )
-    if outbound_ip_prefix_resources:
-        profile.outbound_ip_prefixes = ManagedClusterLoadBalancerProfileOutboundIPPrefixes(
-            public_ip_prefixes=outbound_ip_prefix_resources
-        )
+    if managed_outbound_ip_count or outbound_ip_resources or outbound_ip_prefix_resources:
+        profile.managed_outbound_ips = None
+        profile.outbound_ips = None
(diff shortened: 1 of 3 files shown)


src/azure-cli/azure/cli/command_modules/acs/_loadbalancer.py
```diff
@@ -59,18 +59,22 @@ def configure_load_balancer_profile(managed_outbound_ip_count, outbound_ips, out
...
-    if outbound_ip_resources:
-        profile.outbound_ips = ManagedClusterLoadBalancerProfileOutboundIPs(
-            public_ips=outbound_ip_resources
-        )
-    if outbound_ip_prefix_resources:
-        profile.outbound_ip_prefixes = ManagedClusterLoadBalancerProfileOutboundIPPrefixes(
-            public_ip_prefixes=outbound_ip_prefix_resources
-        )
+    if managed_outbound_ip_count or outbound_ip_resources or outbound_ip_prefix_resources:
+        profile.managed_outbound_ips = None
```
yungezz: just to confirm behavior, when any of `managed_outbound_ip_count`, `outbound_ip_resources`, `outbound_ip_prefix_resources` from customer inputs, all others will be set to None?

qwordy:
Usually, in `update`, if a parameter is not set, it remains the original value.
1. Why make this change?
2. How about noticing users the change in help message?

src/azure-cli/azure/cli/command_modules/acs/_loadbalancer.py
gtracer: @JunSun17 

you can use multiple assignment in python and set them all to none in one go

x, y, z = None, None, None

FYI don't do x = y = z = None or I think it will assign the values to the same block of memory 

https://note.nkmk.me/en/python-multi-variables-values/

src/azure-cli/azure/cli/command_modules/acs/_loadbalancer.py
```diff
@@ -59,18 +59,22 @@ def configure_load_balancer_profile(managed_outbound_ip_count, outbound_ips, out
...
+        profile.outbound_ips = None
+        profile.outbound_ip_prefixes = None
+        if managed_outbound_ip_count:
+            profile.managed_outbound_ips = ManagedClusterLoadBalancerProfileManagedOutboundIPs(
+                count=managed_outbound_ip_count
+            )
```
gtracer: if x:
   //recreate

doesn't need to be nested and I think will read more easily when it isn't

JunSun17:
@qwordy 

Thanks for the comments!

> Usually, in `update`, if a parameter is not set, it remains the original value.
> 
> 1. Why make this change?
This has been discussed inside our team and this is the desired client side logic we want to implement.

> 2. How about noticing users the change in help message?
Yes, will add some logging as well as tests.

JunSun17:
> could you pls add some tests?

Unit test added.

JunSun17:
@qwordy help message is also updated, ready for review.

Issue #12582, opened 2020-03-13
remove azure-cli-core wheel==0.30 wheel pin


kapilt:
instead of hard coding a static wheel pin, azure cli should use a semver constraint
latest wheel is (0.34.2) vs current pin (0.30.0)

yonzhan:
@fengzhou-msft  please take a look

yonzhan:
add to S169

PR #12583, opened 2020-03-13
{Packaging} upgrade wheel to 0.34.2

kapilt:
addresses #12582 

**History Notes:**  

[azure-cli-core]: upgrade wheel package

### DIFF
--- a/src/azure-cli-core/setup.py
+++ b/src/azure-cli-core/setup.py
@@ -66,7 +66,7 @@ DEPENDENCIES = [
     'pyyaml~=5.2',
     'requests~=2.20',
     'six~=1.12',
-    'wheel==0.30.0',
+    'wheel==0.34.2',
     'azure-mgmt-resource~=8.0.1',
 ]
 


yonzhan:
add to S167

yungezz:
this is related to issue  #12582

haroldrandom:
Hi, @kapilt, appreciated the contribution!

Could you give us some context about the upgrade of wheel?
We found the wheel version is not consistent inside several Azure CLI repos, also want to make it consistency.

kapilt:
its causing spurious upgrades/downgrades for projects with azure-cli-core as a dependency.

the underlying wheel api here is public and stable so its unclear why its pinned at all.

probably the bigger issue, its also unclear why its being listed as an install dep since its not used by the library just the wheel packaging.

haroldrandom:
We should all the wheel version for all packages inside Azure/azure-cli.
Also, we shall align the wheel version of Azure/azure-cli-extensions and Azure/azure-cli-dev-tools with this.
At least the should be the same.

fengzhou-msft:
Original PR to pin wheel to 0.30.0: https://github.com/Azure/azure-cli/pull/8896. As long as the issue no longer exists, we should upgrade wheel probably with 'wheel~=0.34.2' or even unpin it.

haroldrandom:
Add context https://github.com/Azure/azure-cli-extensions/issues/128

haroldrandom:
Sorry, upgrade of wheel needs to involve our infrasturre of how to identify Python package. It needs a bit effort to refactor.
But we will plan to relex the restriction ASAP. Sorry for the inconvenience.

haroldrandom:
Relaxed the limit in #12782.

haroldrandom:
plan changed, since relax this need more tests, unpining wheel would be available in the next releases. Won't be late before Azure CLI 3.0

Issue #12597, opened 2020-03-16
{Core} The name of extension of dev is None if has incompatibility problem


haroldrandom:
**Describe the bug**
install an extension which has compatibility restriction, like connnection-monitor-preview (azure-cli-core <= 2.0.82 ). and some comamnds might give warning, but the name of the extension is 'None'.

**To Reproduce**
install connection-monitor-preivew exntension, and then `az -h`:
![image](https://user-images.githubusercontent.com/14357159/76735761-988c9600-67a0-11ea-8ba6-77d2a9587a3a.png)

**Expected behavior**
Show up the extension name.

**Environment summary**
No required

**Additional context**
1. [request for upgrade wheel](https://github.com/Azure/azure-cli/pull/12583)
2. [PEP 426, the metadata of Python Package](https://www.python.org/dev/peps/pep-0426/)
3. [Decided to pin wheel version in CLI](https://github.com/Azure/azure-cli/pull/8896)
4. [Change logs of wheel](https://github.com/pypa/wheel/blob/master/docs/news.rst)

yonzhan:
add to S167

PR #12598, opened 2020-03-16
{Core} hotfix the name of extension of dev type is None if it has incompatibility problem

haroldrandom:
Temporary hotfix the name not found problem when an extension has incompatibility problem.
Fix https://github.com/Azure/azure-cli/issues/12597

image

### Background:
CLI and extension currently relies on metadata.json to load/record meta info, which is generated from wheel <= 0.30.0 while packaging. However it's buggy and wheel starting from 0.32.0, has removed the metadata.json as wheel. 

Also, src/index.json use that that to update. and azdev and extension release pipeline

So, we cannot upgrade wheel for now until we decide to migrate metadata infrastructure. It takes a big effort

**That's why I named the PR title "hotfix"**. It involves how we extract metadata from Python package considering breaking changes of wheel. It's a small step of upgrade wheel.

**Additional context**
1. [request for upgrade wheel](https://github.com/Azure/azure-cli/pull/12583)
2. [PEP 426, the metadata of Python Package](https://www.python.org/dev/peps/pep-0426/)
3. [Decided to pin wheel version in CLI](https://github.com/Azure/azure-cli/pull/8896)
4. [Change logs of wheel](https://github.com/pypa/wheel/blob/master/docs/news.rst)

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/extension/__init__.py
+++ b/src/azure-cli-core/azure/cli/core/extension/__init__.py
@@ -193,6 +193,11 @@ class DevExtension(Extension):
                             key = key.lower()
                             if key == 'version':
                                 metadata[key] = '{}'.format(val.strip())
+                            elif key == 'name':
+                                # temporary fix extension name is None
+                                # until wheel upgrade and metadata structure in azure-cli-extensions has upgraded too.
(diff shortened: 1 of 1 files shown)


yonzhan:
add to S167

Issue #12615, opened 2020-03-17
Azure CLI for role assignment fails if the app registration is created from CLI also.


azabbasi:
> ### `az feedback` auto-generates most of the information requested below, as of CLI version 2.0.62

**Describe the bug**
If I create an app registration from the command line (azure cli) and try to do a role assignment using its application Id, I get the following error:
No matches in graph database for 'xxxxxx' which is the app Id.

This issue doesn't repro if I use azure portal to create the application registration.
I also waited for an hour in case this was an propagation issue. That does not seem to be the case.

**To Reproduce**
* Create an App registration using azure cli
* `$appId = az ad app create --display-name $AppRegistrationName --reply-urls https://api.loganalytics.io/ --available-to-other-tenants false --query 'appId' --output tsv`
* Perform the role assignment from the azure cli
* `az role assignment create --role Reader --assignee $appId --resource-group $rgName`

**Expected behavior**
I expect the role assignment to go through with no issues.

**Environment summary**
Install Method (e.g. pip, interactive script, apt-get, Docker, MSI, edge build) / CLI version (`az --version`) / OS version / Shell Type (e.g. bash, cmd.exe, Bash on Windows)  

Window 10
Powershell version: 5.1.18362.628
Azure CLI version :
azure-cli  : 2.2.0
command-modules-nspkg: 2.0.3
core: 2.2.0
nspkg: 3.0.4
telemetry : 1.0.4
Extensions:
azure-iot: 0.9.1

**Additional context**
We are blocked on this since it is part of our environment deployments.

jiasli:
Hi @azabbasi, Per the source code https://github.com/Azure/azure-cli/blob/c66411567b87eefabbc78b75d27faa91043d8772/src/azure-cli/azure/cli/command_modules/role/custom.py#L1658

`az role assignment create` works on a Service Principal, instead an App. `az ad app create` only creates the App, but not the Service Principal. You will need an extra step to create the corresponding Service Principal in the tenant:

```pwsh
$appId = az ad app create --display-name $AppRegistrationName --reply-urls https://api.loganalytics.io/ --available-to-other-tenants false --query 'appId' --output tsv
az ad sp create --id $appId
az role assignment create --role Reader --assignee $appId --resource-group $rgName
```

azabbasi:
> Hi @azabbasi, Per the source code
> 
[...]

PR #12617, opened 2020-03-17
{RBAC} Polich error message for service principal resolution

jiasli:
Resolve https://github.com/Azure/azure-cli/issues/12615

When the user creates an app but doesn't create the corresponding service principal, then does role assignment

```pwsh
# pwsh
$AppRegistrationName = 'myapp0317'
$appId = az ad app create --display-name $AppRegistrationName --reply-urls https://api.loganalytics.io/ --available-to-other-tenants false --query 'appId' --output tsv
# az ad sp create --id $appId

$rgName = 'myrg0317'
az group create -n $rgName -l westus
az role assignment create --role Reader --assignee $appId --resource-group $rgName
```

`az role assignment create` fails with

> No matches in graph database for '64ba0967-d3b7-4346-b7b6-803fdea011a4'

This PR polishes the error message to make it more instructive:

> Cannot find user or service principal in graph database for '64ba0967-d3b7-4346-b7b6-803fdea011a4'. If the assignee is an appId, make sure the corresponding service principal is created with 'az ad sp create --id 64ba0967-d3b7-4346-b7b6-803fdea011a4'.

ℹ Clean up:
```pwsh
az ad app delete --id $appId
```

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/role/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/role/custom.py
@@ -1661,7 +1661,9 @@ def _resolve_object_id(cli_ctx, assignee, fallback_to_object_id=False):
 
         # 2+ matches should never happen, so we only check 'no match' here
         if not result:
-            raise CLIError("No matches in graph database for '{}'".format(assignee))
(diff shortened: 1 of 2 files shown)


yonzhan:
add to S167

PR #12657, opened 2020-03-19
{CI} Enable incremental test on commnad modules only

haroldrandom:
Scenario of modified files:
1. Modified files in `src/azure-cli-core/azure/cli/core` (https://github.com/Azure/azure-cli/pull/12658) - 83mins
2. Modified files in `src/azure-cli-telemetry/` (https://github.com/Azure/azure-cli/pull/12662) - 53mins
3. Modified files in `src/azure-cli/azure/cli/command_modules/network` (https://github.com/Azure/azure-cli/pull/12661) take 11 mins.
4. Modified files in `srv/azure-cli/azure/cli/command_modules/vm` (https://github.com/Azure/azure-cli/pull/12664) take 11 mins.
5. Modified files in `src/azure-cli-testsdk/` (https://github.com/Azure/azure-cli/pull/12666) - 48mins
6. Modified files in vm and network (https://github.com/Azure/azure-cli/pull/12665) - 14mins.
7. Modified files in network and storage (https://github.com/Azure/azure-cli/pull/12676), - 14 mins.

---

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -424,6 +424,7 @@ jobs:
       filePath: ./scripts/ci/test_automation.sh
     env:
       REDUCE_SDK: 'True'
+      ADO_PULL_REQUEST_TARGET_BRANCH: $(System.PullRequest.TargetBranch)
 
 - job: RunAutomationReduced20190301
   displayName: Run Automation Reduced Python 3.6, Profile 2019-03-01
@@ -451,6 +452,7 @@ jobs:
     env:
       REDUCE_SDK: 'True'
       AZURE_CLI_TEST_TARGET_PROFILE: '2019-03-01'
+      ADO_PULL_REQUEST_TARGET_BRANCH: $(System.PullRequest.TargetBranch)
 
 - job: RunAutomation20180301
   displayName: Run Automation, Profile 2018-03-01
@@ -477,6 +479,7 @@ jobs:
       filePath: ./scripts/ci/test_automation.sh
     env:
       AZURE_CLI_TEST_TARGET_PROFILE: '2018-03-01'
+      ADO_PULL_REQUEST_TARGET_BRANCH: $(System.PullRequest.TargetBranch)
 
 - job: RunAutomationReducedPython3_8
   displayName: Run Automation Reduced Python 3.8
@@ -504,6 +507,7 @@ jobs:
       filePath: ./scripts/ci/test_automation.sh
     env:
       REDUCE_SDK: 'True'
+      ADO_PULL_REQUEST_TARGET_BRANCH: $(System.PullRequest.TargetBranch)
 
 - job: RunAutomationReduced20190301Python3_8
   displayName: Run Automation Reduced Python 3.8, Profile 2019-03-01
@@ -531,6 +535,7 @@ jobs:
     env:
       REDUCE_SDK: 'True'
       AZURE_CLI_TEST_TARGET_PROFILE: '2019-03-01'
+      ADO_PULL_REQUEST_TARGET_BRANCH: $(System.PullRequest.TargetBranch)
 
 - job: TestExtensionsLoading
   displayName: Test Extensions Loading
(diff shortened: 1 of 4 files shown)


yonzhan:
incremental test

azure-pipelines.yml
```diff
@@ -423,7 +423,7 @@ jobs:
       targetType: 'filePath'
       filePath: ./scripts/ci/test_automation.sh
     env:
-      REDUCE_SDK: 'True'
+      ADO_PULL_REQUEST_TARGET_BRANCH: $(System.PullRequest.TargetBranch)
```
qianwens: why do you add this environment ADO_PULL_REQUEST_TARGET_BRANCH

haroldrandom: Because in scripts (both in .sh and .py) can't reference ADO pre-defined variables. So I have to add it here.

Also, I enable incremental test on all target branch instead of dev only on PR stage. This is also another keypoint I set this ADO variable environment.

tools/automation/tests/__init__.py
```diff
@@ -42,6 +102,20 @@ def execute(args):
         output('Running in CI Mode')
         selected_modules = [('All modules', 'azure.cli', 'azure.cli'),
                             ('CLI Linter', 'automation.cli_linter', 'automation.cli_linter')]
+
+        modified_modules = _extract_top_level_modified_modules()
+        if any(base_mod in modified_modules for base_mod in ['core', 'testsdk', 'telemetry']):
+            pass
```
qianwens: >pass [](start = 12, length = 4)

can you add some comments here, like pass means run all test, it took me sometime to understand

tools/automation/tests/__init__.py
```diff
@@ -42,6 +102,20 @@ def execute(args):
...
+        else:
+            test_paths = []
+            for mod in modified_modules:
+                try:
+                    test_paths.append(os.path.normpath(test_index[mod]))
```
qianwens: >test_paths [](start = 20, length = 10)

what if there is duplicated test path? how about use set to have unique value and then transfer to list?

tools/automation/tests/__init__.py
haroldrandom: It's Python list comprehension and will yield a generator as return to any(), expand it to:
```Python
tmp = []
for base_mod in ['core', 'testsdk', 'telemetry']:
    tmp.append(base_mod in modified_modules)

if any(tmp):
    pass
```

tools/automation/tests/__init__.py
haroldrandom: test_path won't be duplicated becase it's like `c:\\users\\jianzen\\documents\\repo\\azure-cli\\src\\azure-cli\\azure\\cli\\command_modules\\storage\\tests\\latest\\test_storage_acl_scenarios.py:StorageAccessControlListTests.test_storage_container_acl_scenarios` .
It's like command_module_folder + tests + profile_name + test_file_name + test_class_name + test_case_name.

tools/automation/tests/__init__.py
haroldrandom: Comments added

tools/automation/tests/__init__.py
haroldrandom: Changed to set()

tools/automation/tests/__init__.py
```diff
@@ -42,6 +106,20 @@ def execute(args):
...
+        else:
+            test_paths = set()
+            for mod in modified_modules:
+                try:
+                    test_paths.add(os.path.normpath(test_index[mod]))
+                except KeyError:
+                    display("no tests found in module: {}".format(mod))
+            args.tests = test_paths
+
+            selected_modules = []
```
fengzhou-msft: In the beloew code, both selected_modules and args.tests are set, but here you cleaned selected_modules. What's the difference? 
https://github.com/Azure/azure-cli/blob/a590a3059cf10d4746f2540a4c5da1b87cc5cfda/tools/automation/tests/__init__.py#L133-L157

haroldrandom: The selected_modules is a list of tuple which consist of 3 elements, like:
https://github.com/Azure/azure-cli/blob/450f7511788f02f2a44eac252e58e4dfa1dfef6e/tools/automation/tests/__init__.py#L43-L44
The last element of the tuple is a Python Namesapce.

In the incremental test case, I identify module from modified files and then extract module name from it.
So the modified_modules is a list of Python module name like `core`, `vm`, `storage`.
Then I extract corresponding tests from testIndex.json which is generated from `get_test_index()`.

In the runner:
https://github.com/Azure/azure-cli/blob/450f7511788f02f2a44eac252e58e4dfa1dfef6e/tools/automation/tests/main.py#L15-L38

tests has higher priority than the module.
So, when I have the extract test need to run, no need to provide `selected_module`.

PR #12683, opened 2020-03-20
{Packaging} Remove universal=1

jiasli:
Remove `universal=1` so that PyPI doesn't mark the package as Python 2 supported: 

https://pypi.org/project/azure-cli/#files

azure_cli-2.2.0-py2.py3-none-any.whl (1.5 MB) | Wheel | py2.py3
-- | -- | --


yonzhan:
add to S167

PR #12580, opened 2020-03-13
[Rest] az rest: Use configured ARM's resource ID

jiasli:
Fix #12566: `az rest` fails to retrieve access token and fail with 

> Get Token request returned http error: 400 and server response: 
> AADSTS65001: The user or administrator has not consented to use the application with ID '04b07795-8ddb-461a-bbee-02f9e1bf7b46' named 'Microsoft Azure CLI'. Send an interactive authorization request for this user and resource

## Cause

To retrieve an access token for ARM API `--endpoint-resource-manager`, the client needs to provide a resource ID configured by `--endpoint-active-directory-resource-id` representing this API. The resource ID may or may not be the same as the API's URL. 

Say for public cloud `AzureCloud`

https://github.com/Azure/azure-cli/blob/a221f49b9af6cf889cde20bbe4b42956900acd37/src/azure-cli-core/azure/cli/core/cloud.py#L226

https://github.com/Azure/azure-cli/blob/a221f49b9af6cf889cde20bbe4b42956900acd37/src/azure-cli-core/azure/cli/core/cloud.py#L231

its API URL is `https://management.azure.com/`, and AAD accepts both `https://management.core.windows.net/` and `https://management.azure.com/` as resource ID, so it doesn't have issue with `az rest`. But still `https://management.core.windows.net/` is the official resource ID.

For dogfood environment 
```
az cloud register
--endpoint-resource-manager https://api-dogfood.resources.windows-int.net/
--endpoint-active-directory-resource-id https://management.core.windows.net/
```
only `https://management.core.windows.net/` is accepted as the resource ID of the ARM API `https://api-dogfood.resources.windows-int.net/`. Using `https://api-dogfood.resources.windows-int.net/` as resource ID will cause AAD to return the mentioned error.

## Change

This PR changes the behavior of `az rest` to follow the same behavior as `azure.cli.core.commands.client_factory._get_mgmt_service_client` 

https://github.com/Azure/azure-cli/blob/a221f49b9af6cf889cde20bbe4b42956900acd37/src/azure-cli-core/azure/cli/core/commands/client_factory.py#L126

so that it uses `active_directory_resource_id` as resource ID, instead of `resource_manager`.

PR #12690, opened 2020-03-20
{REST} az rest: Refinement

jiasli:
**History Notes:**  

[REST] az rest: do not retrieve token if `Authorization` header is already provided
[REST] az rest: support custom `User-Agent`

---

Refine #12117, #12580 

Also, when showing `Authorization` in the http dump, trim at least half of the token but keep at most 20 characters. For example, 
```
> az rest --uri /subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/resourceGroups/?api-version=2019-07-01 --headers "Authorization=abcdefg" --verbose
    'Authorization': 'abc...'
```
```
> az rest --uri /subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/resourceGroups/?api-version=2019-07-01 --verbose
    'Authorization': 'Bearer eyJ0eXAiOiJKV...'
```
Add more tests.

### DIFF
--- a/src/azure-cli-core/azure/cli/core/util.py
+++ b/src/azure-cli-core/azure/cli/core/util.py
@@ -671,7 +671,7 @@ def _log_request(request):
             if header.lower() == 'authorization':
                 # Trim at least half of the token but keep at most 20 characters
                 preserve_length = min(int(len(value) * 0.5), 20)
-                value = value[:preserve_length] + '*****'
+                value = value[:preserve_length] + '...'
             logger.info("    %r: %r", header, value)
         logger.info("Request body:")
 
--- a/src/azure-cli-core/azure/cli/core/util.py
+++ b/src/azure-cli-core/azure/cli/core/util.py
@@ -542,9 +542,10 @@ def send_raw_request(cli_ctx, method, uri, headers=None, uri_parameters=None,  #
                      generated_client_request_id_name='x-ms-client-request-id'):
     import uuid
     from requests import Session, Request
+    from requests.structures import CaseInsensitiveDict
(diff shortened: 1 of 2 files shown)


yonzhan:
add to S167

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -662,7 +669,9 @@ def _log_request(request):
         logger.info("Request headers:")
         for header, value in request.headers.items():
             if header.lower() == 'authorization':
-                value = value[:15] + '*****'
+                # Trim at least half of the token but keep at most 20 characters
+                preserve_length = min(int(len(value) * 0.5), 20)
+                value = value[:preserve_length] + '*****'
```
zhoxing-ms: Do we need to consider whether the number of characters of `value` should remain the same after the last few bits are encrypted by `*`? Is the length of `value` in the previous logic unchanged after encryption?

jiasli: I don't think so, otherwise a Bearer token is too long.

zhoxing-ms: Personally, if `*****` does not contain information about the number of characters, perhaps `...` is more appropriate to indicate omission.
For example: `'Authorization': 'Bearer eyJ0eXAiOiJKV...'`
It's just my personal preference, what do you think about it~

jiasli: Nice idea. 😃 Changed as requested.

jiasli:
```suggestion
                value = value[:preserve_length] + '...'
```

PR #12694, opened 2020-03-21
[Packaging] drop support for python 3.5

### LINKED ISSUE #12525: az 2.2.0 does not appear to be compatible with Python 3.5
**Describe the bug**
az 2.2.0 does not appear to be compatible with Python 3.5.

**To Reproduce**
Run `azdev linter` on a custom extension using Python 3.5.9.  The latest az published at https://packages.microsoft.com/repos/azure-cli started breaking this.  It looks like a Python 3.6-style format string at azure/cli/command_modules/cdn/custom.py line 113 is to blame.

**Expected behavior**
Either builds with 3.5 should work, or 3.5 should not be marked as supported.

**Environment summary**
VSTS pipeline, latest Ubuntu, python 3.5.9, az 2.2.0.

fengzhou-msft:
[knack 0.7.0rc1](https://github.com/microsoft/knack/blob/master/HISTORY.rst) has dropped support for python 3.5.

Resolve #12525

**History Notes:**  

---


yonzhan:
drop python 3.5

yungezz:
hi @fengzhou-msft  why knack drop 3.5 support? what's impact to users if stopping support 3.5?

fengzhou-msft:
knack relies on mock 4.0 which dropped python 3.5. This impacts users using pip install with python 3.5. Telemetry shows the rate for python 3.5 in the past week is less than 0.5%.

src/azure-cli/setup.py
```diff
@@ -154,9 +153,6 @@
 with open('HISTORY.rst', 'r', encoding='utf-8') as f:
     HISTORY = f.read()
 
-if sys.version_info < (3, 4):
```
bim-msft: Why just remove this statement instead of changing it to `if sys.version_info < (3, 6)`?

jiasli: Same question.

fengzhou-msft: This check was added in this [PR](https://github.com/Azure/azure-cli/pull/11363). Adding this sets the hard limit for python versions at installation and block lower verisons. When we remove a version from the classifier, we're droppiong support for that version. It doesn't mean the code must not be compatible with that version. We can remove python 3.6 simply becaused it reaches end-of-life in the future, but our CLI code could still be totally compatible with python 3.6 at that time until at one point it's not. We will not fix it as we stated it's not supported.

bim-msft: @fengzhou-msft I think I got your point, thanks for the clarification.

PR #12702, opened 2020-03-23
{Core} Add . in prefix for operation handler

Juliehzl:
**History Notes:**  

---
`Description:`
`ResourceType` is defined as enum with the following structure for each element:
`DATA_STORAGE = ('azure.multiapi.storage', None)`

If we have two Resource Types named `DATA_STORAGE` and `DATA_STORAGESYNC` with the following definition:
```
DATA_STORAGESYNC = ('azure.multiapi.storagesync.', None)
DATA_STORAGE = ('azure.multiapi.storage.', None)
```
and the following api versions in latest profile:
```
ResourceType.DATA_STORAGESYNC: '2015-12-01',
ResourceType.DATA_STORAGE: '2016-09-01',
```
For the command type with `operations_tmpl = 'azure.mgmt.storagesync.operations#ManagementPoliciesOperations.{}'`, the condition will be fit for resource type `DATA_STORAGE` in original code and CLI will go into the path `azure.multiapi.storage` 
 with api version defined in  `DATA_STORAGE`, not `azure.multiapi.storagesync` :
https://github.com/Azure/azure-cli/blob/ac20369eb7465e56f2b6043b131230310786ec02/src/azure-cli-core/azure/cli/core/__init__.py#L537-L540

But the right api version should be in `ResourceType.DATA_STORAGESYNC`.
With the change, we can prevent such scenario and go into  azure-multiapi-storagesync package as expected.

### DIFF
--- a/src/azure-cli-core/azure/cli/core/__init__.py
+++ b/src/azure-cli-core/azure/cli/core/__init__.py
@@ -534,7 +534,7 @@ class AzCommandsLoader(CLICommandsLoader):  # pylint: disable=too-many-instance-
         from azure.cli.core.profiles._shared import get_versioned_sdk_path
 
         for rt in AZURE_API_PROFILES[self.cli_ctx.cloud.profile]:
-            if operation.startswith(rt.import_prefix):
+            if operation.startswith(rt.import_prefix + '.'):
                 operation = operation.replace(rt.import_prefix,
                                               get_versioned_sdk_path(self.cli_ctx.cloud.profile, rt,
                                                                      operation_group=operation_group))


yonzhan:
add to S167

haroldrandom:
May I have some contexts?

src/azure-cli-core/azure/cli/core/__init__.py
```diff
@@ -534,7 +534,7 @@ def get_op_handler(self, operation, operation_group=None):
         from azure.cli.core.profiles._shared import get_versioned_sdk_path
 
         for rt in AZURE_API_PROFILES[self.cli_ctx.cloud.profile]:
-            if operation.startswith(rt.import_prefix):
+            if operation.startswith(rt.import_prefix + '.'):
```
jiasli: While this does handles some scenarios (please provide some real use cases in the PR description), this still can't handle the keyvault track 1 and 2 SDKs' compatibility:

Track 1 SDK:
- `azure.keyvault`

Track 2 SDK:
- `azure.keyvault.secrets`
- `azure.keyvault.keys`

`azure.keyvault.secrets.SecretClient.xxx` and `azure.keyvault.keys.KeyClient.xxx` will be matched as Track 1 SDK `azure.keyvault.`.

bim-msft:
What's the background of this modification?

src/azure-cli-core/azure/cli/core/__init__.py
yungezz: the change itself looks resolving the issue mentioned in description, same question on is there multiple namespaces containing dot after first dot? such as `azure.multipleapi.storage.blob`, `azure.multipleapi.storage.file`

qwordy:
This sentence "With the change, we can prevent such scenario and go into the" is incomplete

Juliehzl:
> What's the background of this modification?

I have updated the description and you can take a look.

Juliehzl:
> This sentence "With the change, we can prevent such scenario and go into the" is incomplete

Thanks! Updated!

src/azure-cli-core/azure/cli/core/__init__.py
Juliehzl: @jiasli If `azure.keyvault` is totally replaced with submodules such as `azure.keyvault.secrets` and `azure.keyvault.keys`, there will be no problem. 
But if the resource type for `azure.keyvault` still exists with new resource types with `azure.keyvault.secrets` and `azure.keyvault.keys`, it will have some problem.

@yungezz There are packages as you mentioned, e.g. `azure.mgmt.resource` in https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/azure/cli/core/profiles/_shared.py#L44-L49 .

jiasli: https://github.com/Azure/azure-cli/blob/31a9724478c67751ae0bee6cc0d9b75b763df17c/src/azure-cli-core/azure/cli/core/profiles/_shared.py#L44-L49

This doesn't have problem because there is no `azure.mgmt.resource` that can cause the same conflict as `azure.keyvault`. 

I am ok with this for now, but still we need to figure out a better solution.

Juliehzl: Yes. If you want to using submodules for `keyvault`, you cannot use `azure.keyvault` any more.

PR #12712, opened 2020-03-23
[SQL] Sql midb Add: update/show long term retention policy,  show/delete long term retention backups, restore long term retention backup 

xaliciayang:


yonzhan:
add to S167

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -2699,10 +2731,128 @@ def get_short_term_retention_mi(
...
+        yearly_retention=None,
+        week_of_year=None):
+    '''
+    Updates long term retention for managed database
+    '''
+    if not (weekly_retention or monthly_retention or yearly_retention):
+        raise CLIError('Please specify retention setting(s).')
+
+    if yearly_retention and not week_of_year:
+        raise CLIError('Please specify week of year for yearly retention.')
```
arrownj: We can also put these in a separate validator. Anyway, it is also ok to put it here.

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -1500,6 +1533,39 @@ def _configure_security_policy_storage_params(arg_ctx):
...
+        c.argument('week_of_year',
+                    help='The Week of Year, 1 to 52, to save for the Yearly Retention.')
+
+    with self.argument_context('sql midb long-term-retention-backup') as c:
+        c.argument('location_name',
+                    arg_type=get_location_type(self.cli_ctx))
+
+        c.argument('resource_group_name',
+                   required=False,
+                   help='If specified, the resource group the for which a managed instance/database resource belongs to.')
```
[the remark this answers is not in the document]
xaliciayang: dropped instances are not associated with a resource group; so user can specify RG or not depending on which database they want to get a backup from

src/azure-cli/azure/cli/command_modules/sql/tests/latest/test_sql_commands.py
```diff
@@ -3194,6 +3194,77 @@ def test_sql_managed_db_short_retention(self, resource_group, resource_group_loc
                      self.check('retentionDays', '{retention_days_dec}')])
 
 
+class SqlManagedInstanceDbLongTermRetentionScenarioTest(ScenarioTest):
```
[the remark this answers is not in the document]
xaliciayang: yes, that is coming

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -2628,6 +2630,36 @@ def managed_db_restore(
         parameters=kwargs)
 
 
+def managed_db_restore_ltr_backup(
+        cmd,
+        client,
+        database_name,
+        managed_instance_name,
+        resource_group_name,
+        backup_id,
```
pixia: why do you need database_name, managed_instance_name and resource_group_name?

It looks like backup_id is the ltr backup resource id ?

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -2628,6 +2630,36 @@ def managed_db_restore(
...
+        target_managed_instance_name=None,
+        target_resource_group_name=None,
+        **kwargs):
+    '''
+    Restores a managed db from an LTR backup (i.e. create with 'RestoreLongTermRetentionBackup' create mode.)
+    '''
+
+    kwargs['location'] = _get_managed_instance_location(
+        cmd.cli_ctx,
+        managed_instance_name=managed_instance_name,
```
pixia: shouldn't this be managed_instance_name=target_managed_instance_name?

src/azure-cli/azure/cli/command_modules/sql/custom.py
xaliciayang: Whatever is preferred; I'm okay with either.

src/azure-cli/azure/cli/command_modules/sql/commands.py
```diff
@@ -559,6 +561,11 @@ def load_command_table(self, _):
         g.command('list', 'list_by_instance')
         g.command('delete', 'delete', confirmation=True, supports_no_wait=True)
 
+    with self.command_group('sql midb ltr restore',
+                            managed_databases_operations,
+                            client_factory=get_sql_managed_databases_operations) as g:
+        g.custom_command('restore', 'managed_db_restore_ltr_backup', supports_no_wait=True)
```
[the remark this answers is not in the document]
xaliciayang: took out support_no_wait

src/azure-cli/azure/cli/command_modules/sql/custom.py
xaliciayang: yes

src/azure-cli/azure/cli/command_modules/sql/custom.py
xaliciayang: yes, backup_id is the ltr backup resource id 

the other *_name variables should all represent target resources.  fix on the way.

src/azure-cli/azure/cli/command_modules/sql/tests/latest/recordings/test_sql_managed_db_long_term_retention.yaml
```diff
@@ -0,0 +1,346 @@
...
+      User-Agent:
+      - python/3.8.2 (Windows-10-10.0.14393-SP0) msrest/0.6.11 msrest_azure/0.6.3
+        azure-mgmt-network/9.0.0 Azure-SDK-For-Python AZURECLI/2.2.0
+      accept-language:
+      - en-US
+    method: PUT
+    uri: https://api-dogfood.resources.windows-int.net/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/clitest000001/providers/Microsoft.Network/routeTables/vcCliTestRouteTable?api-version=2019-11-01
+  response:
+    body:
+      string: '{"error":{"code":"NoRegisteredProviderFound","message":"No registered
```
xaliciayang: @arrownj can you weigh in on how to address this?  We spoke briefly on Teams about this but I want to make sure I'm running my unit tests correctly

src/azure-cli/azure/cli/command_modules/sql/tests/latest/test_sql_commands.py
```diff
@@ -3194,6 +3194,199 @@ def test_sql_managed_db_short_retention(self, resource_group, resource_group_loc
...
+                self.check('resourceGroup', '{rg}'),
+                self.check('managedInstance', '{managed_instance_name}'),
+                self.check('database', '{database_name}')])
+
+        # test delete long term retention backup
+        self.cmd(
+            'sql midb long-term-retention-backup delete -l {loc} -g {rg} --mi {managed_database_name} -n {database_name} --backup-name {backup_name}')
+        self.cmd('sql midb long-term-retention-backup list-by-database -l {loc} --mi {managed_instance_name} -n {database_name}',
+                 checks=[
+                     self.check(self, None)])
```
xaliciayang: is this the best way to verify nothing returns?

azure-cli1.pyproj
```diff
@@ -0,0 +1,1535 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
pixia: do you need this file?

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -82,8 +82,18 @@ def __init__(
...
-                                         TB=1024 * 1024 * 1024 * 1024)
+        self.unit_map = unit_map or dict(
+            B=1,
+            kB=1024,
+            MB=1024 *
+            1024,
+            GB=1024 *
+            1024 *
+            1024,
+            TB=1024 *
```
pixia: revert this change

pixia: Looks like your env setting made a lot of unnecessary change, I would suggest you to revert those changes to avoid any surprises.

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -1483,22 +1647,62 @@ def _configure_security_policy_storage_params(arg_ctx):
...
-                   options_list=['--deleted-time'],
-                   help='If specified, shows retention days for a deleted database, instead of an existing database.'
-                   'Must match the deleted time of a deleted database on the source Managed Instance.')
+        c.argument(
+            'deleted_time',
+            options_list=['--deleted-time'],
+            help='If specified, shows retention days for a deleted database, instead of an existing database.'
+            'Must match the deleted time of a deleted database on the source Managed Instance.')
+
+    with self.argument_context('sql midb long-term-retention-policy set') as c:
```
pixia: do you also need sql midb long-term-retention-policy set show?

jaredmoo:
I would love if the length of `short-term-retention-policy` can be aggressively reduced in length. Try typing this 10 times in a row and I'm sure you would not want to be a customer using this command repeatedly :) Ergonomics is more important than clarify here - abbreviations can be easily explained with help text, as has been done for `sql mi`, `sql midb`, etc. My first idea is just `short-term`, but there are more possibilities.

src/azure-cli/azure/cli/command_modules/sql/commands.py
```diff
@@ -570,6 +577,30 @@ def load_command_table(self, _):
...
 
+    managed_database_long_term_retention_policies_operations = CliCommandType(
+        operations_tmpl='azure.mgmt.sql.operations#ManagedInstanceLongTermRetentionPoliciesOperations.{}',
+        client_factory=get_sql_managed_database_long_term_retention_policies_operations)
+
+    with self.command_group('sql midb long-term-retention-policy',
+                            managed_database_long_term_retention_policies_operations,
+                            client_factory=get_sql_managed_database_long_term_retention_policies_operations) as g:
+
+        g.custom_command('set', 'update_long_term_retention_mi', supports_no_wait=True)
```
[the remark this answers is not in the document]
xaliciayang: took out support_no_wait

azure-cli1.pyproj
xaliciayang: yes; removing

src/azure-cli/azure/cli/command_modules/sql/_params.py
xaliciayang: correct; for sql midb long-term-retention-policy show, we can use it "out of box" from the SDK

xaliciayang:
> I would love if the length of `short-term-retention-policy` can be aggressively reduced in length. Try typing this 10 times in a row and I'm sure you would not want to be a customer using this command repeatedly :) Ergonomics is more important than clarify here - abbreviations can be easily explained with help text, as has been done for `sql mi`, `sql midb`, etc. My first idea is just `short-term`, but there are more possibilities.

@jaredmoo I will talk to the code owners (I'm just doing long term retention) and see if we can get that cleaned up!  I think I'm going to be doing another PR with LTR for sterling DB's so I can fix up long-term-retention-policy then

src/azure-cli/azure/cli/command_modules/sql/_help.py
```diff
@@ -654,6 +667,76 @@
     text: az sql midb short-term-retention-policy show -g mygroup --mi myinstance -n mymanageddb --deleted-time "2018-05-20T05:34:22"
 """
 
+helps['sql midb long-term-retention-policy'] = """
```
Juliehzl: `long-term-retention-policy` may be too long for users to type in command line. Can we consider abbrevation for this?

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -1476,29 +1476,133 @@ def _configure_security_policy_storage_params(arg_ctx):
...
+        create_args_for_complex_type(
+            c, 'parameters', ManagedDatabase, [
+                'weekly_retention',
+                'monthly_retention',
+                'yearly_retention',
+                'week_of_year'
```
Juliehzl: can we just use one parameter to define retention days with unit defined?

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -1476,29 +1476,133 @@ def _configure_security_policy_storage_params(arg_ctx):
...
+        c.argument('yearly_retention',
+                   help='The Yearly Retention.'
+                   'If just a number is passed instead of an ISO 8601 string, days will be assumed as the units.'
+                   'There is a minimum of 7 days and a maximum of 10 years.')
+
+        c.argument('week_of_year',
+                   help='The Week of Year, 1 to 52, to save for the Yearly Retention.')
+
+    with self.argument_context('sql midb long-term-retention-backup') as c:
+        c.argument('location_name',
```
Juliehzl: location is required for backup show, right? may I know why location is required but no resource group?

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -1476,29 +1476,133 @@ def _configure_security_policy_storage_params(arg_ctx):
...
+                   required=True,
+                   arg_type=get_location_type(self.cli_ctx))
+
+        c.argument('resource_group_name',
+                   required=False,
+                   help='If specified, the resource group that the managed instance/database resource belongs to.')
+
+    with self.argument_context('sql midb long-term-retention-backup show') as c:
```
Juliehzl: What does the --ids mean in your command?
image
Typically if you want to use id part in your command, the --ids should be id of backup, and --name should be backup name.

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -1476,29 +1476,133 @@ def _configure_security_policy_storage_params(arg_ctx):
...
+        c.argument('backup_name',
+                   options_list=['--backup-name', '-b'],
+                   required=True,
+                   help='The long term retention backup name.')
+
+    with self.argument_context('sql midb long-term-retention-backup list-by-database') as c:
```
Juliehzl: May I know  why --location is required for this command?

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -1476,29 +1476,133 @@ def _configure_security_policy_storage_params(arg_ctx):
...
+        c.argument('database_state',
+                   required=False,
+                   help='\'All\', \'Live\', or \'Deleted\'')
+
+        c.argument('only_latest_per_database',
```
Juliehzl: `--only-latest-per-database` is too long as parameter. Can you give a short name for this parameter?

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -2628,6 +2628,32 @@ def managed_db_restore(
...
+    '''
+    kwargs['location'] = _get_managed_instance_location(
+        cmd.cli_ctx,
+        managed_instance_name=target_managed_instance_name,
+        resource_group_name=target_resource_group_name)
+
+    kwargs['create_mode'] = CreateMode.restore_long_term_retention_backup.value
+    kwargs['long_term_retention_backup_resource_id'] = long_term_retention_backup_resource_id
+
+    return client.create_or_update(
```
Juliehzl: Is this operation a long running operation?

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -2628,6 +2628,32 @@ def managed_db_restore(
...
+        **kwargs):
+    '''
+    Restores a managed database from a long term retention backup.
+    '''
+    kwargs['location'] = _get_managed_instance_location(
+        cmd.cli_ctx,
+        managed_instance_name=target_managed_instance_name,
+        resource_group_name=target_resource_group_name)
+
+    kwargs['create_mode'] = CreateMode.restore_long_term_retention_backup.value
```
Juliehzl: Should we expose the restore mode to users? Currently you are hard-code the mode with longtermretention, what if point in time restore?

xaliciayang: Sure; it looked like the command "sql midb restore" already hard-coded PITR, so I just did something similar with "sql midb ltr restore"

src/azure-cli/azure/cli/command_modules/sql/_help.py
xaliciayang: yep, planning to shorten

src/azure-cli/azure/cli/command_modules/sql/_params.py
xaliciayang: No matter what, a LTR backup resource id will include a location.  

For resource group - for example, if a server is dropped, then the resources on that server are also dropped, and therefore they are all disassociated with any resource group.  However, in this case we'd still like the LTR backup to be available.

src/azure-cli/azure/cli/command_modules/sql/_params.py
xaliciayang: location is required to find the right server/db to list LTR backups for

src/azure-cli/azure/cli/command_modules/sql/_params.py
xaliciayang: noted

src/azure-cli/azure/cli/command_modules/sql/_params.py
xaliciayang: It's a little more complex than that.  "Weekly retention" means how long each weekly backup should be kept; "monthly retention" means how long each monthly backup should be kept; "yearly backup" means how long to keep a backup that happens during "week_of_year". 

More info: https://docs.microsoft.com/en-us/azure/sql-database/sql-database-long-term-retention or just message me if this is confusing!

src/azure-cli/azure/cli/command_modules/sql/_params.py
xaliciayang: How do I remove --ids from resource Id arguments? For this function I'm just using the get operation from python sdk.

src/azure-cli/azure/cli/command_modules/sql/custom.py
xaliciayang: yes - where should the wait command go?

xaliciayang:
@Juliehzl 

> I have several suggestions for command design:
> 
> 1. Can we consider shorter name for commands? currently the name is too long.
> 2. It looks long-term-retention-policy/backup/ltr are related command groups. It will be better if we can group them together.
>    image
> 3. I am curious about `restore` command. It should be a general command for midb with different options, why do you only add it in long term retention part?

1. Yes - I'm changing all instances of 'long-term-retention-*' to 'ltr-*' (@jaredmoo)
2. They are indeed all related; however I think it makes most sense to keep ltr policy and ltr backup separate still (to make it clear what the show, set, list-by-* commands are doing) 
3. I will move restore into the midb restore.  I originally made a new function just for ltr restore because currently, the managed db restore command is hard-coded to be point in time restore.

src/azure-cli/azure/cli/command_modules/sql/_params.py
xaliciayang: Doesn't --ids refer to resource ids?

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -1442,13 +1442,27 @@ def _configure_security_policy_storage_params(arg_ctx):
...
             ])
 
-        c.argument('deleted_time',
-                   options_list=['--deleted-time'],
-                   help='If specified, restore from a deleted database instead of from an existing database.'
-                   ' Must match the deleted time of a deleted database on the source Managed Instance.')
+        c.argument(
+            'restore_mode',
+            options_list=['--restore-mode', '--mode'],
+            required=True,
```
jaredmoo, line 1453: Adding a new required param is a breaking change :( we shouldn't do this.

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -1459,46 +1473,120 @@ def _configure_security_policy_storage_params(arg_ctx):
...
-                   'When not specified it defaults to source managed instance.')
+                   'When not specified for PITR it defaults to source managed instance. '
+                   'Required for LTR.')
 
         c.argument('target_resource_group_name',
                    options_list=['--dest-resource-group'],
                    help='Name of the resource group of the managed instance to restore managed database to. '
-                   'When not specified it defaults to source resource group.')
+                   'When not specified for PITR it defaults to source resource group. '
+                   'Required for LTR.')
```
jaredmoo, line 1483: Having a different set of required/non-required parameters is a code smell :(

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -2699,10 +2731,171 @@ def get_short_term_retention_mi(
...
+    policy = client.create_or_update(
+        database_name=database_name,
+        managed_instance_name=managed_instance_name,
+        resource_group_name=resource_group_name,
+        parameters=kwargs)
+
+    return policy
+
+
+def list_by_database_long_term_retention_mi_backups(
```
jaredmoo, line 2770: please use `_` prefix to indicate that it's a private helper function

src/azure-cli/azure/cli/command_modules/sql/_help.py
xaliciayang: done

src/azure-cli/azure/cli/command_modules/sql/_params.py
xaliciayang: Would it be better to split into 2 different functions then?

jaredmoo: One command should not have 2 completely different set of parameters, instead it should just be 2 commands. (Please note the difference between this case, where you have 1 group of required parameters for mode X and another group of required parameters for mode Y, versus the list scenario where you have a bunch of optional parameters and then we are able to intelligently choose the correct SDK method based on the parameters you specified).

xaliciayang:
Previously, I was recommended to add `az sql midb ltr restore` to the existing `az sql midb restore` command (which currently uses point in time restore).  However, after review from @jaredmoo, we decided not to go this route since the two restore modes have different required parameters, and introducing LTR restore into the existing command would be a breaking change.

I wanted to use `az sql midb ltr-backup restore`, but the restore command belongs to a different operations group (managed_databases_operations) from the rest of `ltr-backup` (long_term_retention_managed_instance_operations), so I was not able to include restore in the `ltr-backup` group; please let me know if there is indeed a way to achieve this.  

Therefore, I am going with my original idea of simply doing `az sql midb ltr restore`.  This distinguishes from point in time restore, avoids breaking changes, and indicates that the managed database will be restored from a long term retention backup.

src/azure-cli/azure/cli/command_modules/sql/_params.py
xaliciayang: moved restore under ltr-backup

src/azure-cli/azure/cli/command_modules/sql/_params.py
xaliciayang: reverted this

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -1483,22 +1483,103 @@ def _configure_security_policy_storage_params(arg_ctx):
...
+        create_args_for_complex_type(
+            c, 'parameters', ManagedDatabase, [
+                'weekly_retention',
+                'monthly_retention',
+                'yearly_retention',
+                'week_of_year'
+            ])
+
+        c.argument('weekly_retention',
+                   help='The Weekly Retention.'
```
pixia, line 1516: Retention of weekly backup

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -1483,22 +1483,103 @@ def _configure_security_policy_storage_params(arg_ctx):
...
+                   'If just a number is passed instead of an ISO 8601 string, days will be assumed as the units.'
+                   'There is a minimum of 7 days and a maximum of 10 years.')
+
+        c.argument('monthly_retention',
+                   help='The Monthly Retention.'
```
pixia, line 1521: Retention of monthly backup

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -1483,22 +1483,103 @@ def _configure_security_policy_storage_params(arg_ctx):
...
+                   'If just a number is passed instead of an ISO 8601 string, days will be assumed as the units.'
+                   'There is a minimum of 7 days and a maximum of 10 years.')
+
+        c.argument('yearly_retention',
+                   help='The Yearly Retention.'
```
pixia, line 1526: Retention of yearly backup

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -1483,22 +1483,103 @@ def _configure_security_policy_storage_params(arg_ctx):
...
+                   'If just a number is passed instead of an ISO 8601 string, days will be assumed as the units.'
+                   'There is a minimum of 7 days and a maximum of 10 years.')
+
+        c.argument('week_of_year',
+                   help='The Week of Year, 1 to 52, to save for the Yearly Retention.')
```
pixia, line 1531: The Week of Year, 1 to 52, on which to take the yearly backup.

src/azure-cli/azure/cli/command_modules/sql/_help.py
```diff
@@ -654,6 +654,77 @@
...
+  - name: List long term retention backups for a location (with resource group argument).
+    text: az sql midb ltr-backup list -l southeastasia -g mygroup
+"""
+
+helps['sql midb ltr-backup delete'] = """
+type: command
+short-summary: Delete a long term retention backup.
+examples:
+  - name: Delete long term retention backup for a managed database.
+    text: az sql midb ltr-backup delete -l southeastasia --mi myinstance -n mymanageddb --backup-name "3214b3fb-fba9-43e7-96a3-09e35ffcb336;132292152080000000"
```
jaredmoo, line 712: Because the noun here is `ltr-backup`, `--backup-name` should just be name

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -1483,22 +1483,103 @@ def _configure_security_policy_storage_params(arg_ctx):
...
+        c.argument('target_resource_group_name',
+                   options_list=['--dest-resource-group'],
+                   required=True,
+                   help='Name of the resource group of the managed instance to restore managed database to.')
+
+        c.argument('long_term_retention_backup_resource_id',
+                   options_list=['--backup-id'],
+                   required=True,
+                   help='The resource id of the long term retention backup to be restored. '
+                   'Use \'az sql midb ltr-backup show\' or \'az sql midb ltr-backup list\' for backup id.')
```
jaredmoo, line 1582: I love the help here!

src/azure-cli/azure/cli/command_modules/sql/_help.py
jaredmoo: And database name should be something else

jaredmoo: `--database -d` (see `az sql db op cancel -h`

src/azure-cli/azure/cli/command_modules/sql/_params.py
xaliciayang: B)

src/azure-cli/azure/cli/command_modules/sql/_params.py
Juliehzl: With the link you provided, it becomes clear to me now. So can we consider refine the help message for these parameters and provide the link for more information?

src/azure-cli/azure/cli/command_modules/sql/_params.py
Juliehzl: I am fine with --location optional. Because you add this argument with `required=True`, it will be required for user. 

I am thinking if --location is not necessary to be `required`, as you can see for command `az sql ltr-backup delete` as shown here:
image

src/azure-cli/azure/cli/command_modules/sql/_params.py
Juliehzl: So for this argument, we will assume it to be true when provided, right? If yes, can we consider using `action='store_true'` for this argument to make sure it will be true when specified?

src/azure-cli/azure/cli/command_modules/sql/_help.py
```diff
@@ -654,6 +654,77 @@
...
+  - name: Delete long term retention backup for a managed database.
+    text: az sql midb ltr-backup delete -l southeastasia --mi myinstance -d mymanageddb --name "3214b3fb-fba9-43e7-96a3-09e35ffcb336;132292152080000000"
+"""
+
+helps['sql midb ltr-backup restore'] = """
+type: command
+short-summary: Restore a long term retention backup to a new database.
+examples:
+  - name: Restore a managed database using LTR backup.
+    text: az sql midb ltr-backup restore --dest-database targetmidb --dest-mi myinstance --dest-resource-group mygroup --backup-id "/subscriptions/6caa113c-794c-42f8-ab9d-878d8aa104dc/resourceGroups/mygroup/providers/Microsoft.Sql/locations/southeastasia/longTermRetentionManagedInstances/myinstance/longTermRetentionDatabases/sourcemidb/longTermRetentionManagedInstanceBackups/3214b3fb-fba9-43e7-96a3-09e35ffcb336;132292152080000000"
```
Juliehzl, line 720: Adding `\` in your command will be better format:
image

src/azure-cli/azure/cli/command_modules/sql/commands.py
```diff
@@ -567,9 +569,45 @@ def load_command_table(self, _):
...
+        client_factory=get_sql_managed_database_long_term_retention_backups_operations)
+
+    with self.command_group('sql midb ltr-backup',
+                            managed_database_long_term_retention_backups_operations,
+                            client_factory=get_sql_managed_database_long_term_retention_backups_operations) as g:
+        g.show_command('show', 'get')
+        g.custom_command(
+            'list',
+            'list_long_term_retention_mi_backups')
+        g.command('delete', 'delete')
```
Juliehzl, line 600: Please add `confirmation=True` for delete command.

src/azure-cli/azure/cli/command_modules/sql/tests/latest/test_sql_commands.py
```diff
@@ -3194,8 +3194,114 @@ def test_sql_managed_db_short_retention(self, resource_group, resource_group_loc
                      self.check('retentionDays', '{retention_days_dec}')])
 
 
-class SqlManagedInstanceRestoreDeletedDbScenarioTest(ScenarioTest):
+class SqlManagedInstanceDbLongTermRetentionScenarioTest(ScenarioTest):
+    @record_only()
```
Juliehzl, line 3198: Why do you make it record_only() here? Is there any existing resource reference?

src/azure-cli/azure/cli/command_modules/sql/tests/latest/test_sql_commands.py
```diff
@@ -3194,8 +3194,114 @@ def test_sql_managed_db_short_retention(self, resource_group, resource_group_loc
...
+    def test_sql_managed_db_long_term_retention(
+            self):
+
+        self.kwargs.update({
+            'rg': 'clitestxj6awmetud',
+            'loc': 'westus',
+            'managed_instance_name': 'ayang-ltr-mi',
+            'database_name': 'test-1',
+            'weekly_retention': 'P1W',
```
Juliehzl, line 3207: I am a little confusing about the allowed values for `weekly_retention`, ...?

src/azure-cli/azure/cli/command_modules/sql/tests/latest/test_sql_commands.py
```diff
@@ -3194,8 +3194,114 @@ def test_sql_managed_db_short_retention(self, resource_group, resource_group_loc
...
+        self.cmd(
+            'sql midb ltr-backup list -l {loc} --mi {managed_instance_name} -d {database_name} -g {rg}',
+            checks=[
+                self.greater_than('length(@)', 0)])
+
+        # without resource group
+        self.cmd(
+            'sql midb ltr-backup list -l {loc} --mi {managed_instance_name} -d {database_name}',
+            checks=[
+                self.greater_than('length(@)', 0)])
```
Juliehzl, line 3268: Please change the check values to a specific number using something like `self.check('length(@)', 1)`

src/azure-cli/azure/cli/command_modules/sql/tests/latest/test_sql_commands.py
```diff
@@ -3194,8 +3194,114 @@ def test_sql_managed_db_short_retention(self, resource_group, resource_group_loc
...
-class SqlManagedInstanceRestoreDeletedDbScenarioTest(ScenarioTest):
+class SqlManagedInstanceDbLongTermRetentionScenarioTest(ScenarioTest):
+    @record_only()
+    def test_sql_managed_db_long_term_retention(
+            self):
+
+        self.kwargs.update({
+            'rg': 'clitestxj6awmetud',
+            'loc': 'westus',
+            'managed_instance_name': 'ayang-ltr-mi',
```
Juliehzl, line 3205: Because you are using an existing managed instance, one day someone else takes over this part and it will be hard to re-run the test. 
And CLI live run test will skipped.
Is this acceptable for you?

src/azure-cli/azure/cli/command_modules/sql/tests/latest/test_sql_commands.py
xaliciayang: https://docs.microsoft.com/en-us/azure/sql-database/sql-database-long-term-retention

Here is a little more info.  When using the command line, we expect retention settings to be an ISO 8601 string.

src/azure-cli/azure/cli/command_modules/sql/tests/latest/test_sql_commands.py
xaliciayang: The exact number of backups in this situation can change quickly (since backups can be added/deleted - even this test does a delete at the very end), but in general should be nonzero.

src/azure-cli/azure/cli/command_modules/sql/tests/latest/test_sql_commands.py
xaliciayang: Yes; I've spoken to @arrownj about this scenario.  In fact I prefer not to create new resources every time.

src/azure-cli/azure/cli/command_modules/sql/tests/latest/test_sql_commands.py
xaliciayang: Can also explain further on Teams if needed

src/azure-cli/azure/cli/command_modules/sql/tests/latest/test_sql_commands.py
Juliehzl: the check is used to check the output for this command. it doesn't matter for changing later.

src/azure-cli/azure/cli/command_modules/sql/tests/latest/test_sql_commands.py
Juliehzl: yes. Please ping me in Teams for this.

src/azure-cli/azure/cli/command_modules/sql/_params.py
xaliciayang: We will assume false unless provided.

src/azure-cli/azure/cli/command_modules/sql/_params.py
xaliciayang: Sure.  I can add those

xaliciayang: discussed on Teams

src/azure-cli/azure/cli/command_modules/sql/tests/latest/test_sql_commands.py
xaliciayang: discussed on Teams - users know to use ISO format

src/azure-cli/azure/cli/command_modules/sql/tests/latest/test_sql_commands.py
xaliciayang: spoke on teams

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -2699,10 +2699,287 @@ def get_short_term_retention_mi(
...
+    Extract resource values from the backup id
+    '''
+
+    backup_id = backup_id.replace('\'', '')
+    backup_id = backup_id.replace('"', '')
+
+    if backup_id[0] == '/':
+        # remove leading /
+        backup_id = backup_id[1:]
+        print(backup_id)
```
Juliehzl, line 2749: print() is not recommended here. Is there any reason to print these information? If you really want to print these information, please use logger.

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -2699,10 +2699,287 @@ def get_short_term_retention_mi(
...
+
+    resources_list = backup_id.split('/')
+    print(resources_list)
```
Juliehzl, line 2752: The same

src/azure-cli/azure/cli/command_modules/sql/tests/latest/test_sql_commands.py
```diff
@@ -3194,8 +3194,114 @@ def test_sql_managed_db_short_retention(self, resource_group, resource_group_loc
...
+            'sql midb ltr-policy set -g {rg} --mi {managed_instance_name} -n {database_name} --weekly-retention {weekly_retention} --monthly-retention {monthly_retention} --yearly-retention {yearly_retention} --week-of-year {week_of_year}',
+            checks=[
+                self.check('resourceGroup', '{rg}'),
+                self.check('weeklyRetention', '{weekly_retention}'),
+                self.check('monthlyRetention', '{monthly_retention}'),
+                self.check('yearlyRetention', '{yearly_retention}')])
+
+        # test get long term retention policy on live database
+        self.cmd(
+            'sql midb ltr-policy show -g {rg} --mi {managed_instance_name} -n {database_name}',
```
Juliehzl, line 3224: Please also add test for show command with back up id

src/azure-cli/azure/cli/command_modules/sql/tests/latest/test_sql_commands.py
```diff
@@ -3194,8 +3194,114 @@ def test_sql_managed_db_short_retention(self, resource_group, resource_group_loc
...
+            'sql midb ltr-backup list -l {loc}',
+            checks=[
+                self.greater_than('length(@)', 0)])
+
+        # test list long term retention backups for instance
+        # with resource group
+        self.cmd(
+            'sql midb ltr-backup list -l {loc} --mi {managed_instance_name} -g {rg}',
+            checks=[
+                self.greater_than('length(@)', 0)])
```
Juliehzl, line 3249: test check still need to be a certain value.

src/azure-cli/azure/cli/command_modules/sql/_params.py
```diff
@@ -1483,22 +1483,124 @@ def _configure_security_policy_storage_params(arg_ctx):
...
+                   'Use \'az sql midb ltr-backup show\' or \'az sql midb ltr-backup list\' for backup name.',
+                   id_part=None)
+
+        c.argument('backup_id',
+                   options_list=['--backup-id', '--id'],
+                   help='The resource id of the backups. '
+                   'Use \'az sql midb ltr-backup show\' or \'az sql midb ltr-backup list\' for backup id. '
+                   'If provided, other arguments are not required. ')
+
+    with self.argument_context('sql midb ltr-backup list') as c:
```
Juliehzl, line 1559: Please also add is_preview=True to new commands, which will give us buffer time to test command behavior.

src/azure-cli/azure/cli/command_modules/sql/custom.py
xaliciayang: oops; never meant to push this

src/azure-cli/azure/cli/command_modules/sql/_params.py
xaliciayang: Sure;  added.  Is there a reason no other sql commands are in preview now?

src/azure-cli/azure/cli/command_modules/sql/tests/latest/test_sql_commands.py
xaliciayang: changed to specific lengths

src/azure-cli/azure/cli/command_modules/sql/_params.py
xaliciayang: added backup_id to show/delete commands (discussed on teams)

PR #12716, opened 2020-03-23
[KeyVault] az keyvault create: support `--network-acls`, `--network-acls-ips` and `--network-acls-vnets` for specifying network rules while creating vault

bim-msft:
Support for AB#1150

Usage:

Use JSON string directly, for the `vnet` field, `{vnet_name}/{subnet_name}` and `{subnet_id}` are both supported:
```
az keyvault create --network-acls "{
     \"ip\":  [\"1.2.3.4\", \"2.3.4.0/24\"], 
     \"vnet\": [
           \"vnet_name_1/subnet_name1\", 
           \"/subscriptions/000000-0000-0000/resourceGroups/MyResourceGroup/providers/Microsoft.Network/virtualNetworks/vnet2/subnets/subnet2\"
     ]
}"
```

Use JSON file:
```
az keyvault create --network-acls network-acls-example.json
```

Use flattened parameters (IP rule):
```
az keyvault create --network-acls-ips 1.2.3.4 2.3.4.0/24
```

Virtual network rule:
```
az keyvault create --network-acls-vnets vnet_name_1/subnet_name1 /subscriptions/000000-0000-0000/resourceGroups/MyResourceGroup/providers/Microsoft.Network/virtualNetworks/vnet2/subnets/subnet2
```

Also, the three parameters `--network-acls`, `--network-acls-ips` and `--network-acls-vnets` are all optional and they can be used together:
```
az keyvault create --network-acls ... --network-acls-ips ... --network-acls-vnets ...
```

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/keyvault/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/keyvault/_help.py
@@ -138,7 +138,7 @@ examples:
 
   - name: Create a key vault with network ACLs specified (use --network-acls, --network-acls-ips and --network-acls-vnets together, redundant rules will be removed, finally there will be 4 IP rules and 3 VNet rules).
     text: |
-        az keyvault create --location westus2 --name MyKeyVault --resource-group MyResourceGroup --network-acls "{\\"ip\\": [\\"1.2.3.4\\", \\"2.3.4.0/24\\"], \\"vnet\\": [\\"vnet_name_1/subnet_name1\\", \\"vnet_name_2/subnet_name2\\"]} --network-acls-ips 3.4.5.0/24 4.5.6.0/24 --network-acls-vnets vnet_name_2/subnet_name_2 vnet_name_3/subnet_name_3 /subscriptions/000000-0000-0000/resourceGroups/MyResourceGroup/providers/Microsoft.Network/virtualNetworks/vnet_name_4/subnets/subnet_name_4
+        az keyvault create --location westus2 --name MyKeyVault --resource-group MyResourceGroup --network-acls "{\\"ip\\": [\\"1.2.3.4\\", \\"2.3.4.0/24\\"], \\"vnet\\": [\\"vnet_name_1/subnet_name1\\", \\"vnet_name_2/subnet_name2\\"]}" --network-acls-ips 3.4.5.0/24 4.5.6.0/24 --network-acls-vnets vnet_name_2/subnet_name_2 vnet_name_3/subnet_name_3 /subscriptions/000000-0000-0000/resourceGroups/MyResourceGroup/providers/Microsoft.Network/virtualNetworks/vnet_name_4/subnets/subnet_name_4
 """
 
 helps['keyvault delete'] = """
--- a/src/azure-cli/azure/cli/command_modules/keyvault/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/keyvault/_help.py
@@ -119,6 +119,13 @@ examples:
     text: |
         az keyvault create --location westus2 --name MyKeyVault --resource-group MyResourceGroup
     crafted: true
+  - name: Create a key vault with network ACLs specified (use JSON string).
+    text: |
+        az keyvault create --location westus2 --name MyKeyVault --resource-group MyResourceGroup 
+        --network-acls "{\\"ip\\": [\\"1.2.3.4\\", \\"2.3.4.0/24\\"], \\"vnet\\": [\\"vnet_name_1/subnet_name1\\", \\"vnet_name_2/subnet_name2\\"]}"
+  - name: Create a key vault with network ACLs specified (use JSON file).
+    text: |
+        az keyvault create --location westus2 --name MyKeyVault --resource-group MyResourceGroup --network-acls network-acls-example.json
 """
 
 helps['keyvault delete'] = """
--- a/src/azure-cli/azure/cli/command_modules/keyvault/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/keyvault/_help.py
@@ -119,23 +119,23 @@ examples:
     text: |
         az keyvault create --location westus2 --name MyKeyVault --resource-group MyResourceGroup
     crafted: true
-  
+
   - name: Create a key vault with network ACLs specified (use --network-acls to specify IP and VNet rules by using a JSON string).
     text: |
         az keyvault create --location westus2 --name MyKeyVault --resource-group MyResourceGroup --network-acls "{\\"ip\\": [\\"1.2.3.4\\", \\"2.3.4.0/24\\"], \\"vnet\\": [\\"vnet_name_1/subnet_name1\\", \\"vnet_name_2/subnet_name2\\", \"/subscriptions/000000-0000-0000/resourceGroups/MyResourceGroup/providers/Microsoft.Network/virtualNetworks/MyVNet/subnets/MySubnet\"]}"
-  
+
   - name: Create a key vault with network ACLs specified (use --network-acls to specify IP and VNet rules by using a JSON file).
     text: |
         az keyvault create --location westus2 --name MyKeyVault --resource-group MyResourceGroup --network-acls network-acls-example.json
-  
+
   - name: Create a key vault with network ACLs specified (use --network-acls-ips to specify IP rules).
     text: |
         az keyvault create --location westus2 --name MyKeyVault --resource-group MyResourceGroup --network-acls-ips 3.4.5.0/24 4.5.6.0/24
-  
+
   - name: Create a key vault with network ACLs specified (use --network-acls-vnets to specify VNet rules).
     text: |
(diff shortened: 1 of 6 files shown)


src/azure-cli/azure/cli/command_modules/keyvault/custom.py
```diff
@@ -264,6 +264,49 @@ def recover_keyvault(cmd, client, vault_name, resource_group_name, location):
...
+
+    try:
+        network_acls_json = json.loads(network_acls)
+    except:  # pylint: disable=bare-except
+        with open(network_acls) as f:
+            logger.info('network_acls_filename: {}'.format(network_acls))
+            network_acls_json = json.load(f)
+
+    try:
+        network_acls_json = json.loads(json.dumps(network_acls_json))
```
yungezz: why load again?

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
```diff
@@ -264,6 +264,49 @@ def recover_keyvault(cmd, client, vault_name, resource_group_name, location):
...
+            raise CLIError('Invalid VNet rule: {}. Format: {{vnet}}/{{subnet}}'.format(vnet_rule))
+
+        vnet_name = items[0].lower()
+        subnet_name = items[1].lower()
+        vnet = _construct_vnet(cmd, resource_group_name, vnet_name, subnet_name)
+        network_acls.virtual_network_rules.append(VirtualNetworkRule(id=vnet))
+
+    network_acls.ip_rules = []
+    for ip_rule in network_acls_json.get('ip', []):
+        network_acls.ip_rules.append(IPRule(value=ip_rule))
```
yungezz: does rules only support IPRule? Is there any other rules such as network security group, application security group etc?

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
```diff
@@ -264,6 +264,49 @@ def recover_keyvault(cmd, client, vault_name, resource_group_name, location):
...
+    VirtualNetworkRule = cmd.get_models('VirtualNetworkRule', resource_type=ResourceType.MGMT_KEYVAULT)
+    IPRule = cmd.get_models('IPRule', resource_type=ResourceType.MGMT_KEYVAULT)
+
+    network_acls = _create_network_rule_set(cmd)
+
+    network_acls.virtual_network_rules = []
+    for vnet_rule in network_acls_json.get('vnet', []):
+        items = vnet_rule.split('/')
+        if len(items) != 2:
+            raise CLIError('Invalid VNet rule: {}. Format: {{vnet}}/{{subnet}}'.format(vnet_rule))
```
yungezz: can user input resource id directly?

src/azure-cli/azure/cli/command_modules/keyvault/_params.py
```diff
@@ -89,6 +89,9 @@ class CLIJsonWebKeyOperation(str, Enum):
         c.argument('location', validator=get_default_location_from_resource_group)
         c.argument('enable_soft_delete', arg_type=get_three_state_flag(), help="Enable 'soft delete' functionality for this key vault and all contained entities. If omitted, it will be set to true by default. Once set to true, it cannot be reverted to false.")
         c.argument('retention_days', help='Soft delete data retention days. It accepts >=7 and <=90.', default='90')
+        c.argument('network_acls', help='Network ACLs. It accepts a JSON filename or a JSON string. '
```
yungezz: besides ip and vnet, any other supported rules? how about flatten it as --ips and --vnets?

yonzhan:
keyvault

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
bim-msft: > why load again?

@yungezz Good point, actually I added this sentence in order to make sure the input is a valid JSON. But later I modified the code that makes this part useless... I will remove this part.

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
bim-msft: > does rules only support IPRule? Is there any other rules such as network security group, application security group etc?

@yungezz Rules only contain `IPRule` and `VirtualNetworkRule` currently. Consistent with `az keyvault network-rule add`. NSG and ASG are not supported.

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
bim-msft: > can user input resource id directly?

@yungezz It should be supported, I will update the code.

src/azure-cli/azure/cli/command_modules/keyvault/_params.py
bim-msft: > besides ip and vnet, any other supported rules? how about flatten it as --ips and --vnets?

@yungezz Only IP and VNet rules now, flatten is also a good idea. How about we support `--network-acls <json>`, `--network-acls-ips <ip list>`, `--network-acls-vnets <vnet list>` at the same time? They can work together and all three parameters are optional.

src/azure-cli/azure/cli/command_modules/keyvault/_help.py
```diff
@@ -119,6 +119,15 @@
...
+    text: |
+        az keyvault create --location westus2 --name MyKeyVault --resource-group MyResourceGroup --network-acls "{\\"ip\\": [\\"1.2.3.4\\", \\"2.3.4.0/24\\"], \\"vnet\\": [\\"vnet_name_1/subnet_name1\\", \\"vnet_name_2/subnet_name2\\", \"/subscriptions/000000-0000-0000/resourceGroups/MyResourceGroup/providers/Microsoft.Network/virtualNetworks/MyVNet/subnets/MySubnet\"]}"
+  - name: Create a key vault with network ACLs specified (use JSON file).
+    text: |
+        az keyvault create --location westus2 --name MyKeyVault --resource-group MyResourceGroup --network-acls network-acls-example.json
+  - name: Create a key vault with network ACLs specified (use --network-acls, --network-acls-ips and --network-acls-vnets at the same time).
+    text: |
+        az keyvault create --location westus2 --name MyKeyVault --resource-group MyResourceGroup --network-acls network-acls-example.json --network-acls-ips 3.4.5.0/24 4.5.6.0/24 --network-acls-vnets vnet1/subnet1 vnet2/subnet2 /subscriptions/000000-0000-0000/resourceGroups/MyResourceGroup/providers/Microsoft.Network/virtualNetworks/vnet3/subnets/subnet3
```
yungezz: thanks @bim-msft for quick change. Should we fail validation when json and flatten params from user at same time?

bim-msft: > thanks @bim-msft for quick change. Should we fail validation when json and flatten params from user at same time?

@yungezz I think it's safe to pass validation for this case which all three params are specified at the same time. As far as I can see, there is one scenario may benefits from this: user has one basic network rules json file, and this file is always applied to a new created vault, also there are different additional rules for each new vault.

yungezz: :)

src/azure-cli/azure/cli/command_modules/keyvault/_help.py
```diff
@@ -119,6 +119,26 @@
...
         az keyvault create --location westus2 --name MyKeyVault --resource-group MyResourceGroup
     crafted: true
+
+  - name: Create a key vault with network ACLs specified (use --network-acls to specify IP and VNet rules by using a JSON string).
+    text: |
+        az keyvault create --location westus2 --name MyKeyVault --resource-group MyResourceGroup --network-acls "{\\"ip\\": [\\"1.2.3.4\\", \\"2.3.4.0/24\\"], \\"vnet\\": [\\"vnet_name_1/subnet_name1\\", \\"vnet_name_2/subnet_name2\\", \"/subscriptions/000000-0000-0000/resourceGroups/MyResourceGroup/providers/Microsoft.Network/virtualNetworks/MyVNet/subnets/MySubnet\"]}"
+
+  - name: Create a key vault with network ACLs specified (use --network-acls to specify IP and VNet rules by using a JSON file).
+    text: |
+        az keyvault create --location westus2 --name MyKeyVault --resource-group MyResourceGroup --network-acls network-acls-example.json
```
fengzhou-msft: No `@` prefix for file?

bim-msft: > No `@` prefix for file?

@fengzhou-msft Is `@` a convention? I think in this case users don't need to specify the marker explicitly, the program can detect whether it's a JSON string or a JSON filename.

fengzhou-msft: It seems json str or @file.json is a convention, and `type=validate_file_or_dict` can be used, but it makes file completer not working. @Juliehzl @MyronFanQiu any guideline for this?

fengzhou-msft: Tested with validate_file_or_dict, it can support json string and file path with or without @ prefix.

bim-msft: @fengzhou-msft That's good, I will use `type=validate_file_or_dict` to refine the code.

bim-msft: @fengzhou-msft For the help messages and examples, I decided not to add `@` as a prefix since this is not mandatory and may cause confusion.

PR #12724, opened 2020-03-23
[RDBMS] Added support for minimal tls version and deny public access

akamath46297:
This PR add an ability to provide the --minimal-tls-version and --public-network-access properties for create and update server


src/azure-cli/azure/cli/command_modules/rdbms/custom.py
```diff
@@ -67,6 +72,8 @@ def _server_create(cmd, client, resource_group_name, server_name, sku_name, no_w
                 administrator_login_password=administrator_login_password,
                 version=version,
                 ssl_enforcement=ssl_enforcement,
+                minimal_tls_version=minimal_tls_version,
```
ramnov: MariaDB does not support minimal_tls_version. can remove this

akamath46297: I am removing it in the params. So this should be fine. If the user tried to provide this value it will show unrecognized arguments. Going ahead we will be supporting TLS for mariadb as well and then its one line change

akamath46297:
@yonzhan , @Juliehzl  and @haroldrandom could you guys help include this in S167 https://github.com/Azure/azure-cli/milestone/84

yonzhan:
add to S167

src/azure-cli/azure/cli/command_modules/rdbms/_params.py
```diff
@@ -80,6 +80,8 @@ def _complex_params(command_group):
             c.argument('administrator_login', options_list=['--admin-user', '-u'], help='Administrator username for the server. Once set, it cannot be changed.')
             c.argument('administrator_login_password', options_list=['--admin-password', '-p'], help='The password of the administrator. Minimum 8 characters and maximum 128 characters. Password must contain characters from three of the following categories: English uppercase letters, English lowercase letters, numbers, and non-alphanumeric characters.')
             c.argument('ssl_enforcement', arg_type=get_enum_type(['Enabled', 'Disabled']), options_list=['--ssl-enforcement'], help='Enable or disable ssl enforcement for connections to server. Default is Enabled.')
+            c.argument('minimal_tls_version', arg_type=get_enum_type(['TLS1_0', 'TLS1_1', 'TLS1_2', 'TLSEnforcementDisabled']), options_list=['--minimal-tls-version'], help='Set the minimal TLS version for connections to server when SSL is enabled. Default is TLS1_2.', validator=tls_validator)
```
Juliehzl: The default value is applied in your service server side, right?

src/azure-cli/azure/cli/command_modules/rdbms/_params.py
```diff
@@ -80,6 +80,8 @@ def _complex_params(command_group):
...
+            c.argument('public_network_access', arg_type=get_enum_type(['Enabled', 'Disabled']), options_list=['--public-network-access'], help='Set whether or not public network access is allowed for this server. Default is Enabled')
```
Juliehzl: Can you change to use get_three_state_flags() like https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/sql/_params.py#L1048?

src/azure-cli/azure/cli/command_modules/rdbms/tests/latest/test_rdbms_commands.py
```diff
@@ -101,29 +101,53 @@ def _test_server_mgmt(self, database_engine, resource_group_1, resource_group_2)
...
+                        checks=[
+                            JMESPathCheck('name', servers[0]),
+                            JMESPathCheck('resourceGroup', resource_group_1),
+                            JMESPathCheck('administratorLogin', admin_login),
+                            JMESPathCheck('sslEnforcement', 'Enabled'),
+                            JMESPathCheck('tags.key', '1'),
+                            JMESPathCheck('sku.capacity', old_cu),
+                            JMESPathCheck('sku.tier', edition),
+                            JMESPathCheck('storageProfile.backupRetentionDays', backupRetention),
+                            JMESPathCheck('storageProfile.geoRedundantBackup', geoRedundantBackup)])
```
Juliehzl: Please add test for default value check for tsl_version and public_network_access here.

src/azure-cli/azure/cli/command_modules/rdbms/_params.py
akamath46297: Thats correct. As discussed I didnt add Preview for now. I have also removed the TLS from the params as the api changes have not yet landed. Will get this in in next release

src/azure-cli/azure/cli/command_modules/rdbms/_params.py
akamath46297: All three services covered in the rdbms (mysql, postgres and mariadb uses different model though same) hence changing this to use get_three_state_flags will add additional conversions for each of these types separately. We also have ssl enforcement which has the same problem hence we have kept it as enabled and disabled for now. Will think about this to see how we can make a generic change.

src/azure-cli/azure/cli/command_modules/rdbms/tests/latest/test_rdbms_commands.py
akamath46297: Done

src/azure-cli/azure/cli/command_modules/rdbms/tests/latest/test_rdbms_commands.py
```diff
@@ -202,17 +213,6 @@ def _test_server_mgmt(self, database_engine, resource_group_1, resource_group_2)
                      JMESPathCheck('sku.tier', edition),
                      JMESPathCheck('administratorLogin', admin_login)])
 
-        self.cmd('{} server restore -g {} --name {} '
```
Juliehzl: Why do we delete restore part in test?

Juliehzl: I will help you merge it first. But please add it as we discussed offline.

PR #12730, opened 2020-03-24
[AppService] fix #11412: az webapp create is removing existing application settings

### LINKED ISSUE #11412: az webapp create is removing existing application settings
## Describe the bug
When `az webapp create` is called every time a new pipeline is running, all existing appsettings is removed and it's force the site to break down if ex. WEBSITE_RUN_FROM_PACKAGE=1

**Command Name**
`az webapp create`

**Errors:**
See above description.

## To Reproduce:

- `az webapp create --name {} --plan {}`

Trough the portal or anywhere you can add an application setting do that. Then run the same command again.

- `az webapp create --name {} --plan {}`

Now the application setting you created before is deleted.

We are using Azure CLI in our CI/CD pipeline so we are running the same command over and over again. Previous we used Azure ARM, which has not the same issue.

## Expected Behavior
Keep existing application settings.

## Environment Summary
```
Windows-10-10.0.18362-SP0
Python 3.6.6
Shell: powershell.exe

azure-cli 2.0.76 *

Extensions:
azure-devops 0.8.0
dev-spaces-preview 0.1.6
interactive 0.4.3

```
## Additional Context

It looks like `az webapp deployment slot create` is not removing existing application settings when called, so it would be cool if `az webapp create` can act the same way.

Kotasudhakarreddy:
**History Notes:**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -140,17 +140,6 @@ def create_webapp(cmd, resource_group_name, name, plan, runtime=None, startup_fi
     else:  # windows webapp without runtime specified
         site_config.app_settings.append(NameValuePair(name="WEBSITE_NODE_DEFAULT_VERSION",
                                                       value=node_default_version))
-
-    # This is to keep the existing appsettings for a newly created webapp on existing webapp name.
-    name_validation = client.check_name_availability(name, 'Site')
-    if not name_validation.name_available:
-        existing_app_settings = _generic_site_operation(cmd.cli_ctx, resource_group_name,
-                                                        name, 'list_application_settings')
-        site_config.app_settings = []
-        site_config.app_settings += [NameValuePair(name=k, value=v)
-                                     for k, v
-                                     in existing_app_settings.properties.items()]
-
     if site_config.app_settings:
         for setting in site_config.app_settings:
             logger.info('Will set appsetting %s', setting)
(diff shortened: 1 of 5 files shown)


Kotasudhakarreddy:
@panchagnula - FYA
@ThejaChoudary - FYI

Kotasudhakarreddy:
@panchagnula - need your help here. Build is failing and I am not sure the exact issue. Could you please suggest me here.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -141,6 +141,16 @@ def create_webapp(cmd, resource_group_name, name, plan, runtime=None, startup_fi
         site_config.app_settings.append(NameValuePair(name="WEBSITE_NODE_DEFAULT_VERSION",
                                                       value=node_default_version))
 
+    # This is to keep the existing appsettings for a newly created webapp on existing webapp name.
+    name_validation = client.check_name_availability(name, 'Site')
```
yungezz: why not directly call get? If get return with none, means no existing site, else appending to existing app settings. this saves a rest call.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -141,6 +141,16 @@ def create_webapp(cmd, resource_group_name, name, plan, runtime=None, startup_fi
...
+    if not name_validation.name_available:
+        existing_app_settings = _generic_site_operation(cmd.cli_ctx, resource_group_name,
+                                                        name, 'list_application_settings')
+        site_config.app_settings = []
+        site_config.app_settings += [NameValuePair(name=k, value=v)
```
yungezz: need to check updating existing app_settings: some config in app_settings already, and user want to update it.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
Kotasudhakarreddy: @yungezz - agreed  that making a `get` call will work but this is an expensive call where it returns an entire webapp object and we need to hold it through out. But in my scenario I will make second call only when the app already exists, and the second call is only to fetch the appsettings. And it will not occupy more memory even though we hold it. Both the rest calls are light weight here.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
Kotasudhakarreddy: thanks will work and send the updated code

Kotasudhakarreddy:
This is has some unwanted changes and unable to revert them. closing and will create a new request

PR #12737, opened 2020-03-25
[AppConfig] Add kv export-secret command

shenmuxiaosen:
**History Notes:**  

---


shenmuxiaosen:
@avanigupta Can you please take a look? Thanks.

src/azure-cli/azure/cli/command_modules/appconfig/_help.py
```diff
@@ -121,6 +121,16 @@
     text: az appconfig kv export -n MyAppConfiguration -d appconfig --dest-name AnotherAppConfiguration --key * --label * --dest-label ExportedKeys
 """
 
+helps['appconfig kv export-secret'] = """
+type: command
+short-summary: Export values of key vault reference from your App Configuration.
+examples:
+  - name: Export key-vault reference values.
+    text: az appconfig kv export -n MyAppConfiguration --label test
```
avanigupta: >export [](start = 26, length = 6)

export-secret

shenmuxiaosen: Done.

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -245,6 +248,51 @@ def export_config(cmd,
...
+                  label=None,
+                  key=None,
+                  prefix="",
+                  path=None,
+                  format_=None):
+    # fetch key values from user's configstore
+    src_kvs = __read_kv_from_config_store(
+        cmd, name=name, connection_string=connection_string, key=key, label=label, prefix_to_remove=prefix)
+
+    kv_references = [keyvaule for keyvaule in src_kvs if keyvaule.content_type == KeyVaultConstants.KEYVAULT_CONTENT_TYPE]
```
avanigupta: >keyvaule [](start = 21, length = 8)

nit: keyvalue

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -245,6 +248,51 @@ def export_config(cmd,
...
 
 
+def export_secret(cmd,
+                  name=None,
+                  connection_string=None,
+                  label=None,
+                  key=None,
+                  prefix="",
+                  path=None,
+                  format_=None):
```
avanigupta: >format_ [](start = 18, length = 7)

What is the default format?

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -245,6 +248,51 @@ def export_config(cmd,
...
+        try:
+            secret = keyvault_client.get_secret(vault_base_url=kv_identifier.vault,
+                                                secret_name=kv_identifier.name,
+                                                secret_version=kv_identifier.version)
+            secrets[kv.key] = secret.value
+        except Exception as exception:   # pylint: disable=broad-except
+            logger.error("Cannot fetch secret: %s. Exception: %s", secret_id, str(exception))
+
+    if path:
+        with open(path, 'w') as fp:
```
avanigupta: >with open(path, 'w') as [](start = 8, length = 23)

Do we want to add encoding='utf-8' here?

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
shenmuxiaosen: We don't have default format. If no --path specified, then this command will return secrets, as storing secrets in persist files may be a concern for customers. If --path specified but no --format, then we will just create an empty file. Same behavior like kv export.

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
shenmuxiaosen: Fixed.

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
shenmuxiaosen: Yeah, we should use utf8 as default encoding. However, from keyvault perspective, each secret can have its own encoding. For instance ```az keyvault secret download --encoding```. Since for our command, we kind of batch export secrets into one file, there is no way to customize encoding for each secret.

src/azure-cli/azure/cli/command_modules/appconfig/_help.py
```diff
@@ -121,6 +121,16 @@
     text: az appconfig kv export -n MyAppConfiguration -d appconfig --dest-name AnotherAppConfiguration --key * --label * --dest-label ExportedKeys
 """
 
+helps['appconfig kv export-secret'] = """
+type: command
+short-summary: Export values of key vault reference from your App Configuration.
```
avanigupta: >values [](start = 22, length = 6)

The "values"  being exported may be confused with values of app config keys. Can we make it something like "Export Key Vault secret values from Key Vault references in your App Configuration"?

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
avanigupta: In that case, it may be better to call it "get-secret" instead of "export-secret". I feel like the expectation with "export" commands is to save the output somewhere. If customers only want to see their secret values in the terminal, or optionally save it in a file, "get" pattern may be better. And in that case, we wouldn't even have to support file formats, as users can pipe the output from CLI in any format they like.

---
In reply to: [398245409](https://github.com/Azure/azure-cli/pull/12737#discussion_r398245409) [](ancestors = 398245409)

avanigupta: Or we could use "get-keyvault" since we already have a "set-keyvault" command.

---
In reply to: [398254227](https://github.com/Azure/azure-cli/pull/12737#discussion_r398254227) [](ancestors = 398254227,398245409)

Juliehzl:
@shenmuxiaosen please rebase dev to fix CI failure.

shenmuxiaosen:
Abandon this pr due to design change.

Issue #12668, opened 2020-03-19
Equivalent for az cosmosdb collection exists


bojingo:
**Is your feature request related to a problem? Please describe.**

I am writing a script that will update a property of a Cosmos DB container and perform a variety of other operations, **but only if the container already exists**. I'm not interested in getting an error if the container doesn't exist as in my case it doesn't really represent an error condition. 

I am using the command `az cosmosdb sql container update` for this, which works fine. However, I can't find a way to test if the container exists or not. The `show` command returns an error rather than no result if it doesn't exist. *(Side note: in my Google searches, many people claim that `show` would return nothing if the object doesn't exist and is a good method to test for existence. I have no reason to think these people are all incorrect which probably means the contract for the `show` command is inconsistent across the Azure CLI)* I don't see an `exists` command at all in the help, and even tried without success to see if perhaps it was undocumented.

However, I did find `exists` within the collection subgroup of cosmosdb. This does what I want it to, however I get the following message:
```
This command has been deprecated and will be removed in a future release. Use 'cosmosdb sql container, cosmosdb mongodb collection, cosmosdb cassandra table, cosmosdb gremlin graph or cosmosdb table' instead.
This command is implicitly deprecated because command group 'cosmosdb collection' is deprecated and will be removed in a future release. Use 'cosmosdb sql container, cosmosdb mongodb collection, cosmosdb cassandra table, cosmosdb gremlin graph or cosmosdb table' instead.
```
Doh - took me back full-circle. Not helpful.

**Describe the solution you'd like**

Command `az cosmosdb sql container exists` returns true if the container exists, and false if not. The same required parameters would need provided as with the other commands of the container subgroup.

**Describe alternatives you've considered**

1. Continuing to use the deprecated variant for now.
2. Determine existence of the container based on error detection (YUCKY!!!!)

**Additional context**

In my opinion, the `exists` command should be a standard part of the CLI contract of any resource type, and any subgroup where the concept of "existence" is relevant. If I should change my thinking around this, please explain in your replies.

yungezz:
thanks for reporting, routing to appropriate team

shurd:
Hi @bojingo, thank you for opening this issue. When adding the new commands, we did not create an "exists" command. As you indicate, currently you must use the "show" command and determine if there was an error. We will work on adding an "exists" option. We are tracking this internally as work item 672701 and have an ETA of the end of April.

yonzhan:
move to S169

[...]

PR #12774, opened 2020-03-27
[Cosmos DB] az cosmosdb: add exists command to database and container groups

shurd:
This change addresses https://github.com/Azure/azure-cli/issues/12668 by adding supports for "exists" in the new commands that replace the deprecated "cosmosdb database" and "cosmosdb collection" commands. It also updates parameter descriptions to address https://github.com/Azure/azure-cli/issues/13096

These commands are added:
az cosmosdb sql database exists
az cosmosdb sql container exists
az cosmosdb mongodb database exists
az cosmosdb mongodb collection exists
az cosmosdb gremlin database exists
az cosmosdb gremlin graph exists
az cosmosdb cassandra keyspace exists
az cosmosdb cassandra table exists
az cosmosdb table exists

**Testing Guide**  

**History Notes:**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/cosmosdb/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/cosmosdb/custom.py
@@ -8,6 +8,7 @@
 from enum import Enum
 from knack.log import get_logger
 from knack.util import CLIError
+from msrestazure.azure_exceptions import CloudError
 
 from azure.mgmt.cosmosdb.models import (
     ConsistencyPolicy,
@@ -216,6 +217,12 @@ def cli_cosmosdb_keys(client, resource_group_name, account_name, key_type=Cosmos
     raise CLIError("az cosmosdb keys list: '{0}' is not a valid value for '--type'. See 'az cosmosdb keys list --help'.".format(key_type))
 
 
+def _handle_exists_exception(cloud_error):
+    if cloud_error.status_code == 404:
+        return False
+    raise cloud_error
+
+
 def cli_cosmosdb_sql_database_create(client,
                                      resource_group_name,
                                      account_name,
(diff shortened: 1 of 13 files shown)


yonzhan:
add to S168

shurd:
@Juliehzl can you please review this PR?

Juliehzl:
@shurd  LGTM in general. But could you resolve all conflicts and I will merge it for you?

yonzhan:
add to S170

yonzhan:
@shurd please resolve the conflicts.

yonzhan:
add to S171

shurd:
@yonzhan @Juliehzl I believe the conflicts are now resolved

src/azure-cli/azure/cli/command_modules/cosmosdb/custom.py
```diff
@@ -216,6 +217,12 @@ def cli_cosmosdb_keys(client, resource_group_name, account_name, key_type=Cosmos
     raise CLIError("az cosmosdb keys list: '{0}' is not a valid value for '--type'. See 'az cosmosdb keys list --help'.".format(key_type))
 
 
+def _handle_exists_exception(cloud_error):
+    if cloud_error.status_code == 404:
+        return False
```
zhoxing-ms, line 222: Is there any other error conditions when `status_code` is 404 other than where the resource already exists? 
And the original meaning of `404` as http staus should be **not found**, `cloud_error.status_code` is on the contrary, it seems misleading. If the error status provided by the server can be modified, I think it might be better to use `409 Conflict` for this scenario.

Juliehzl: It should be fine to use `404` for not found scenario. When not found, it means the db doesn't exist and return false for the command.

src/azure-cli/azure/cli/command_modules/cosmosdb/custom.py
```diff
@@ -615,6 +676,19 @@ def cli_cosmosdb_mongodb_database_create(client,
...
 
+def cli_cosmosdb_mongodb_database_exists(client,
+                                         resource_group_name,
+                                         account_name,
+                                         database_name):
+    """Checks if an Azure Cosmos DB MongoDB database exists"""
+    try:
+        client.get_mongo_db_database(resource_group_name, account_name, database_name)
+    except CloudError as ex:
+        return _handle_exists_exception(ex)
```
Juliehzl, line 687: By the way, you can add exception_handler kwargs in command for exception handler as shown in https://github.com/Azure/azure-cli/blob/master/src/azure-cli/azure/cli/command_modules/storage/commands.py#L231.

shurd: I tried this, but I do not think I can return "True" in that case like I do in the exception handler here. If I do that, I get:

cli.azure.cli.core.util : 'CommandResultItem' object is not iterable

Is it possible to return something other than an exception from the exception handler kwags method?

PR #12778, opened 2020-03-29
{Extension} insert extension directory in head of sys.path

### LINKED ISSUE #12062: CLI calls the wrong version SDK when a different version SDK is installed in current pyenv

**Describe the bug**
Details are in https://github.com/microsoft/MLOpsPython/issues/136

**To Reproduce**
NOTE: to repro the error reported in https://github.com/microsoft/MLOpsPython/issues/136 you will need more setup, such as creating ML workspace, register a model etc. The following repro step is simplified. You won't see an error but you will see the symptom of different version SDK is loaded.

step 1: install latest azure-cli-ml: az extension add -n azure-cli-ml (or az extension update -n azure-cli-ml). 
   You should get version 1.0.85. SDK version is correct as you can see from 
      $HOME/.azure/cliextensions/azure-cli-ml/azureml_core*

step 2: create a pyenv, can be conda, venv, doesn't matter. I use venv
    python -m venv test-venv
    source test-venv/bin/activate

step 3: install a different version of azureml-core
    pip install azureml-core==1.0.79

step 4: run an ml CLI command:  az ml workspace list -g abc --debug (doesn't matter whether resource group abc exists or not)
    In the debug level log, you will see SDK 1.0.79 is loaded (you will see 'azureml-core 1.0.79' in debug output)

In the real error case, there is a change in azureml-core 1.0.81 which introduces a new required parameter.

**Expected behavior**

**Environment summary**

**Additional context**

fengzhou-msft:
Fix #12062.
The extension directory was appended to the sys.path, and the virtual env site-package directory was in front of it. So CLI will load the SDK installed in the virtual env instead of the one in extension directory. Now we put extension directory in front position of sys.path to be loaded priorly.

**Testing Guide**  

**History Notes:**  

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/__init__.py
+++ b/src/azure-cli-core/azure/cli/core/__init__.py
@@ -228,7 +228,7 @@ class MainCommandsLoader(CLICommandsLoader):
                         continue
                     ext_name = ext.name
                     ext_dir = ext.path or get_extension_path(ext_name)
-                    sys.path.append(ext_dir)
+                    sys.path.insert(0, ext_dir)
                     try:
                         ext_mod = get_extension_modname(ext_name, ext_dir=ext_dir)
                         # Add to the map. This needs to happen before we load commands as registering a command
--- a/src/azure-cli-core/azure/cli/core/extension/operations.py
+++ b/src/azure-cli-core/azure/cli/core/extension/operations.py
@@ -339,7 +339,7 @@ def reload_extension(extension_name, extension_module=None):
 
 def add_extension_to_path(extension_name, ext_dir=None):
     ext_dir = ext_dir or get_extension(extension_name).path
-    sys.path.append(ext_dir)
+    sys.path.insert(0, ext_dir)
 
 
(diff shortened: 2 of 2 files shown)


yonzhan:
add to S167

yungezz:
hi @fengzhou-msft when user is using venv, it's by design all packages should be loaded from venv, right? Will sync with you offline on this.

jiasli:
FYI, I debugged into 
https://github.com/Azure/azure-cli/blob/34f9033b96e6ea31f82626a9f2b152c1a061300e/src/azure-cli-core/azure/cli/core/__init__.py#L231
`ext_dir` is like `'C:\\Users\\jiasli\\.azure\\cliextensions\\support'`. 

In that case, if two extensions contain the same SDK but with different versions, it will still has conflict. So as an ultimate solution, the extension should vendor and rename the SDK or put it under `vendored_sdks` to avoid the conflict.

jiasli:
Another concern is that

PR #12813, opened 2020-03-31
[Batch] Update Batch API

bgklein:

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/batch/_parameter_format.py
+++ b/src/azure-cli/azure/cli/command_modules/batch/_parameter_format.py
@@ -23,7 +23,8 @@ BASIC_TYPES = {
 
 # Parameters that we want to be present in the namespace, but not exposed to the user as arguments
 SILENT_PARAMETERS = [
-    'pool.virtual_machine_configuration.image_reference'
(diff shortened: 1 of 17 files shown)


yonzhan:
add to S168

src/azure-cli/HISTORY.rst
```diff
@@ -60,6 +60,10 @@ Release History
 * Added support for Backing up OS Disks only
 * Added restore-as-unmanaged-disk parameter to specify unmanaged restore
 
+**Batch**
+* az batch pool create: Added support to specify --disk-encryption-targets
+* Update Python SDK dependency to 9.0
```
qwordy, line 65: It is no longer required to add history notes here.
Instead, put it in PR title or add them to description of PR. Azure CLI will generate history notes automatically when releasing a new version.

bgklein:
With the new azdev setup those tests will only record the old help text until the CLI is actually released. Don't think its really needed anymore as well.

bgklein:
@qwordy Should be good for merge unless there are any new CLI steps we missed

qwordy:
> 
> 
> @qwordy Should be good for merge unless there are any new CLI steps we missed

Done

Issue #11909, opened 2020-01-20
“Az Snapshot Create” fails with 404 when the Managed Disk and Resource Group locations are different.


RishiMaini:
“Az Snapshot Create” fails with 404 when the Managed Disk and Resource Group locations are different.

Steps to repro 
===============
Create a Resource Group in location 1

Create a managed disk under the same Resource Group but the Managed disk location should be different from Resource Group location.

Now if you try to create a Snapshot of the Disk using “Az Snapshot Create”, it will fail with 404 error.

Same operation works either through PowerShell or directly from portal.

abc@Azure:~$ az snapshot create -g "DiskRG" -n "Snap99" --source "testdisk2"
**Resource testdisk2 is not found.**

Looks like Az Snapshot is checking for Disk existence under the Resource Group location rather than actual disk location.

Expectation is that CLI command should also work fine independent of Resource Group and Disk location, just like PowerShell and Portal.

This is same issue as https://github.com/Azure/azure-cli/issues/9233 which was closed earlier without further research.

yonzhan:
issue labelled.

qwordy:
I reproduce the problem using the following code.
```
az group create -g fyrg1 -l westus
az disk create -g fyrg1 -n d1 --size-gb 10 -l eastus
az snapshot create -g fyrg1 -n s1 --source d1
```
Some logs of `snapshot create` command
```
Request URL: 'https://management.azure.com/subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/resourceGroups/fyrg1/providers/Microsoft.Compute/snapshots/s1?api-version=2019-07-01'
Request body:
{"location": "westus", "tags": {}, "sku": {"name": "Standard_LRS"}, "properties": {"hyperVGeneration": "V1", "creationData": {"createOption": "Copy", "sourceResourceId": "/subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/resourceGroups/fyrg1/providers/Microsoft.Compute/disks/d1"}}}
Response content:
{
  "error": {
    "code": "NotFound",
    "message": "Resource d1 is not found."
  }
}
```
Azure CLI doesn't have location limitation in this command. In this example, it sends request to service and gets an error from service. I am not very familiar with PowerShell. Can you share PowerShell commands? Or can you paste HTTP request of PowerShell snapshot create command from log if the log has this info? Thank you!

RishiMaini:
[...]

PR #12828, opened 2020-04-01
{Compute} Update help of disk/snapshot create

qwordy:
#11909 
Describe --location more precisely. Tell users it will inherit from location of resource group if not specified. There are many other commands do not state it explicitly. And we can't modify it in global parameter definition because inheritance from location of resource group is a customized behavior of each command.

**Testing Guide**  
`az disk create -h`
`az snapshot create -h`

**History Notes:**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/vm/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_params.py
@@ -118,6 +118,7 @@ def load_arguments(self, _):
             c.argument('encryption_type', min_api='2019-07-01', arg_type=get_enum_type(self.get_models('EncryptionType')),
                        help='Encryption type. EncryptionAtRestWithPlatformKey: Disk is encrypted with XStore managed key at rest. It is the default encryption type. EncryptionAtRestWithCustomerKey: Disk is encrypted with Customer managed key at rest.')
             c.argument('disk_encryption_set', min_api='2019-07-01', help='Name or ID of disk encryption set that is used to encrypt the disk.')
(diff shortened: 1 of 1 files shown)


src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -118,6 +118,7 @@ def load_arguments(self, _):
             c.argument('encryption_type', min_api='2019-07-01', arg_type=get_enum_type(self.get_models('EncryptionType')),
                        help='Encryption type. EncryptionAtRestWithPlatformKey: Disk is encrypted with XStore managed key at rest. It is the default encryption type. EncryptionAtRestWithCustomerKey: Disk is encrypted with Customer managed key at rest.')
             c.argument('disk_encryption_set', min_api='2019-07-01', help='Name or ID of disk encryption set that is used to encrypt the disk.')
+            c.argument('location', help='Location. Values from: `az account list-locations`. You can configure the default location using `az configure --defaults location=<location>`. If location is not specified and no default location specified, location will be automatically set as same as the resource group.')
```
arrownj, line 121: Should we start a session to align all the location definitions in different modules? I found that there are some modules which also have their own definition, and obviously here is another different definition.

qwordy: I agree with you. I opened another issue to track it.
This PR will focus on problem of this specific command reported by customer.

yonzhan:
Compute

PR #12856, opened 2020-04-02
[WIP][Extension] Support installing extensions in a system directory

### LINKED ISSUE #9831: `az extension add` should give an option to install extension for all users (i.e. at a machine level)
`az extension add` should give an option to install extension for all users (i.e. at a machine level)

Currently it is possible that azure cli is installed for all users in the machine but extension installation is bound only to a particular user..

it will be great if there can be an option to install extension at machine level instead of just user level

fengzhou-msft:
Resolve #9831.
Currently users can set `AZURE_EXTENSION_DIR` env variable to install extensions in a customized path. But this has a limit to install all extensions in one directory. 

This PR adds an option `--system` to `az extension add`. So users can choose to install an extension in either a user path or a system path, but not able to install the same extension in two paths. az will load the extensions from both paths. 

The system directory is also configurable by env var `AZURE_EXTENSION_SYS_DIR`. The default is the `azure-cli-extensions` folder under the lib path of the python that az is using.

**Testing Guide**  

**History Notes:**  

[Extension] az extension add: Add --system to enable installing extensions in a system path

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/extension/__init__.py
+++ b/src/azure-cli-core/azure/cli/core/extension/__init__.py
@@ -316,8 +316,6 @@ def build_extension_path(ext_name, system=None):
 
 def get_extensions(ext_type=None):
     logger.debug("Extensions directory: '%s'", EXTENSIONS_DIR)
-    if os.path.isdir(EXTENSIONS_SYS_DIR):
-        logger.debug("Extensions system directory: '%s'", EXTENSIONS_SYS_DIR)
     extensions = []
     if not ext_type:
         ext_type = EXTENSION_TYPES
--- a/src/azure-cli-core/azure/cli/core/extension/__init__.py
+++ b/src/azure-cli-core/azure/cli/core/extension/__init__.py
@@ -2,6 +2,7 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
+# pylint: disable=line-too-long
 
 import os
 import traceback
@@ -18,10 +19,11 @@ from knack.log import get_logger
 az_config = CLIConfig(config_dir=GLOBAL_CONFIG_DIR, config_env_var_prefix=ENV_VAR_PREFIX)
 _CUSTOM_EXT_DIR = az_config.get('extension', 'dir', None)
 _DEV_EXTENSION_SOURCES = az_config.get('extension', 'dev_sources', None)
+_CUSTOM_EXT_SYS_DIR = az_config.get('extension', 'sys_dir', None)
 EXTENSIONS_DIR = os.path.expanduser(_CUSTOM_EXT_DIR) if _CUSTOM_EXT_DIR else os.path.join(GLOBAL_CONFIG_DIR,
                                                                                           'cliextensions')
 DEV_EXTENSION_SOURCES = _DEV_EXTENSION_SOURCES.split(',') if _DEV_EXTENSION_SOURCES else []
-EXTENSIONS_SYS_DIR = os.path.join(get_python_lib(), 'azure-cli-extensions') if sys.platform.startswith('linux') else ""
+EXTENSIONS_SYS_DIR = os.path.expanduser(_CUSTOM_EXT_SYS_DIR) if _CUSTOM_EXT_SYS_DIR else os.path.join(get_python_lib(), 'azure-cli-extensions')
 
 EXTENSIONS_MOD_PREFIX = 'azext_'
 
@@ -134,10 +136,10 @@ class WheelExtension(Extension):
 
     def get_metadata(self):
         from glob import glob
-        if not extension_exists(self.name):
-            return None
         metadata = {}
         ext_dir = self.path or get_extension_path(self.name)
+        if not ext_dir or not os.path.isdir(ext_dir):
+            return None
         info_dirs = glob(os.path.join(ext_dir, '*.*-info'))
         azext_metadata = WheelExtension.get_azext_metadata(ext_dir)
         if azext_metadata:
@@ -204,11 +206,10 @@ class DevExtension(Extension):
         return self.metadata.get('version')
 
     def get_metadata(self):
-
-        if not extension_exists(self.name):
-            return None
         metadata = {}
         ext_dir = self.path
+        if not ext_dir or not os.path.isdir(ext_dir):
+            return None
         egg_info_dirs = [f for f in os.listdir(ext_dir) if f.endswith('.egg-info')]
         azext_metadata = DevExtension.get_azext_metadata(ext_dir)
         if azext_metadata:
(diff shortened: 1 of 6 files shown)


src/azure-cli-core/azure/cli/core/extension/__init__.py
```diff
@@ -18,10 +18,11 @@
 az_config = CLIConfig(config_dir=GLOBAL_CONFIG_DIR, config_env_var_prefix=ENV_VAR_PREFIX)
 _CUSTOM_EXT_DIR = az_config.get('extension', 'dir', None)
 _DEV_EXTENSION_SOURCES = az_config.get('extension', 'dev_sources', None)
+_CUSTOM_EXT_SYS_DIR = az_config.get('extension', 'sys_dir', None)
 EXTENSIONS_DIR = os.path.expanduser(_CUSTOM_EXT_DIR) if _CUSTOM_EXT_DIR else os.path.join(GLOBAL_CONFIG_DIR,
                                                                                           'cliextensions')
 DEV_EXTENSION_SOURCES = _DEV_EXTENSION_SOURCES.split(',') if _DEV_EXTENSION_SOURCES else []
-EXTENSIONS_SYS_DIR = os.path.join(get_python_lib(), 'azure-cli-extensions') if sys.platform.startswith('linux') else ""
+EXTENSIONS_SYS_DIR = os.path.expanduser(_CUSTOM_EXT_SYS_DIR) if _CUSTOM_EXT_SYS_DIR else os.path.join(get_python_lib(), 'azure-cli-extensions')
```
fengzhou-msft, line 25: Need to test if this also works on Windows and Mac.

src/azure-cli-core/azure/cli/core/extension/operations.py
```diff
@@ -220,13 +223,14 @@ def add_extension(cmd, source=None, extension_name=None, index_url=None, yes=Non
...
+    ext = get_extension(extension_name)
+    _augment_telemetry_with_ext_info(extension_name, ext)
     try:
-        if extension_name and get_extension(extension_name).experimental:
+        if extension_name and ext.experimental:
             logger.warning("The installed extension '%s' is experimental and not covered by customer support. "
                            "Please use with discretion.", extension_name)
-        elif extension_name and get_extension(extension_name).preview:
+        elif extension_name and ext.preview:
             logger.warning("The installed extension '%s' is in preview.", extension_name)
```
fengzhou-msft, line 234: get_extension calls get_extensions to load all extension, then fetch the extension with the name, we should avoid calling it mulitple times.

yonzhan:
add to S168

src/azure-cli-core/azure/cli/core/extension/__init__.py
fengzhou-msft: On windows, users need to open the shell as administrator. I have added [message](https://github.com/Azure/azure-cli/pull/12856/files#diff-c9aa0ae7daec922ec8ebbd390137a745R59) when users encounter the permission error.

src/azure-cli-core/azure/cli/core/extension/__init__.py
```diff
@@ -303,7 +303,15 @@ def get_extension_modname(ext_name=None, ext_dir=None):
 
 def get_extension_path(ext_name):
     # This will simply form the path for a WHEEL extension.
-    return os.path.join(EXTENSIONS_DIR, ext_name)
+    ext_sys_path = os.path.join(EXTENSIONS_SYS_DIR, ext_name)
+    ext_path = os.path.join(EXTENSIONS_DIR, ext_name)
+    return ext_path if os.path.isdir(ext_path) else (
```
arrownj, line 308: add `os.path.isdir(ext_path)` check here, is it because previous code may case potential bug ?

src/azure-cli-core/azure/cli/core/extension/operations.py
```diff
@@ -220,13 +225,14 @@ def add_extension(cmd, source=None, extension_name=None, index_url=None, yes=Non
             logger.debug(err)
             raise CLIError("No matching extensions for '{}'. Use --debug for more information.".format(extension_name))
     extension_name = _add_whl_ext(cmd=cmd, source=source, ext_sha256=ext_sha256,
-                                  pip_extra_index_urls=pip_extra_index_urls, pip_proxy=pip_proxy)
-    _augment_telemetry_with_ext_info(extension_name)
+                                  pip_extra_index_urls=pip_extra_index_urls, pip_proxy=pip_proxy, system=system)
     try:
-        if extension_name and get_extension(extension_name).experimental:
+        ext = get_extension(extension_name)
+        _augment_telemetry_with_ext_info(extension_name, ext)
```
arrownj, line 231: move telemetry part into the try/except, so if `get_extension` throw exception, it will not be executed, right ?

src/azure-cli-core/azure/cli/core/extension/__init__.py
fengzhou-msft: Previous code use `get_extension_path` both before install when this path does not exist and when loading extensions. I splitted it to use the below `build_extension_path` for building the path before install, and to use this `get_extension_path` only for loading extensions. So we can add the isdir check here now.

src/azure-cli-core/azure/cli/core/extension/__init__.py
```diff
@@ -303,7 +303,15 @@ def get_extension_modname(ext_name=None, ext_dir=None):
 
 def get_extension_path(ext_name):
     # This will simply form the path for a WHEEL extension.
```
fengzhou-msft, line 305:
```suggestion
    # This will return the path for a WHEEL extension if exists.
```

src/azure-cli-core/azure/cli/core/extension/operations.py
fengzhou-msft: Yes. get_extension was also called in _augment_telemetry_with_ext_info in previous code and it will be catched inside. And I think it's the same flow.

src/azure-cli-core/azure/cli/core/extension/__init__.py
yungezz: run as admin only required when user set customized install dir right? if yes, then less concern since no change to existing behavoir

src/azure-cli-core/azure/cli/core/extension/__init__.py
```diff
@@ -134,10 +135,10 @@ def get_version(self):
 
     def get_metadata(self):
         from glob import glob
-        if not extension_exists(self.name):
-            return None
         metadata = {}
         ext_dir = self.path or get_extension_path(self.name)
+        if not ext_dir or not os.path.isdir(ext_dir):
```
yungezz, line 140: create extension dir if not exists?

src/azure-cli-core/azure/cli/core/extension/operations.py
```diff
@@ -53,6 +56,8 @@ def _run_pip(pip_exec_args):
     except CalledProcessError as e:
         logger.debug(e.output)
         logger.debug(e)
+        if "PermissionError: [WinError 5]" in e.output:
```
yungezz, line 59: what's default install path after the change? suppose no change

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -191,6 +191,9 @@ def _get_version_string(name, version_dict):
         _print()
     _print("Python location '{}'".format(sys.executable))
     _print("Extensions directory '{}'".format(EXTENSIONS_DIR))
+    import os
+    if os.path.isdir(EXTENSIONS_SYS_DIR) and os.listdir(EXTENSIONS_SYS_DIR):
+        _print("Extensions system directory '{}'".format(EXTENSIONS_SYS_DIR))
```
yungezz, line 196: it's little confusing, how about direct show as `Extensions directory`

fengzhou-msft: The idea is to have no impact on users who don't use --system to install extensions. This will only be shown when users have installed at least one extension with --system.

src/azure-cli-core/azure/cli/core/extension/operations.py
fengzhou-msft: yes. no change for the default extension path.

src/azure-cli-core/azure/cli/core/extension/__init__.py
fengzhou-msft: yes, not affecting default extension installation.

src/azure-cli-core/azure/cli/core/extension/__init__.py
fengzhou-msft: The logic to create extension dir if not exists is only added when installing extensions. Here we may not want that.

haroldrandom:
Would it be better install inside `site-package/` directly?

fengzhou-msft:
> Would it be better install to inside `site-package/` directly?

I think it's more organized with the azure-cli-extensions folder. What's the benefit of installing inside `site-package/` directly?

haroldrandom:
> What's

a dedicated folder will do as this PR https://github.com/Azure/azure-cli/pull/12367

PR #12942, opened 2020-04-10
[ARM] az deployment sub/group what-if：Add new commands to support What-If for ARM template deployment

shenglol:
The PR adds new commands for [ARM template deployment What-If](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/template-deploy-what-if) (in preview):
- `az deployment sub what-if`
- `az deployment group what-if`

which can tell users what will happen when they deploy a template if it is valid.

A new flag `--confirm-with-what-if` was also added to the existing `deployment sub/group create` commands to allow users to review and acknowledge resource changes and execute deployment with one operation.

Here's a sample output:
image

**Testing Guide**  

**History Notes:**  

[ARM] Add template deployment What-If commands

---


yonzhan:
add to S168

zhoxing-ms:
@shenglol Why doesn't `az deployment mg`, `az deployment tenant` have `what-if`?

zhoxing-ms:
I suggest that it might be better to change the PR title, like: `[ARM] az deployment sub/group/... what-if：Add new commands to support What-If for ARM template deployment`
PR title writing guide: [link](https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules#submitting-pull-requests)

src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -275,6 +275,8 @@
     short-summary: The location to store the deployment metadata.
   - name: --name -n
     short-summary: The deployment name.
+  - name: --what-if-result-format
+    short-summary: The format of What-If results. Applicable when --confirm-with-what-if is set.
```
zhoxing-ms, line 279: `--what-if-result-format` the name of this parameter looks long, could you add a short alias?

shenglol:
> @shenglol Why doesn't `az deployment mg`, `az deployment tenant` have `what-if`?

Currently the What-If back-end doesn't support mg-level and tenant-level deployment.

src/azure-cli/azure/cli/command_modules/resource/_help.py
shenglol: Create an alias `-r`. Also added it to the `what-if` commands for the `--result-format` parameter to keep it consistent.

shenglol:
> I suggest that it might be better to change the PR title, like: `[ARM] az deployment sub/group/... what-if：Add new commands to support What-If for ARM template deployment`
> PR title writing guide: [link](https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules#submitting-pull-requests)

Thanks. I changed the title.

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -55,6 +55,12 @@ def load_arguments(self, _):
     no_prompt = CLIArgumentType(arg_type=get_three_state_flag(), help='The option to disable the prompt of missing parameters for ARM template. '
                                 'When the value is true, the prompt requiring users to provide missing parameter will be ignored. The default value is false.')
 
+    deployment_what_if_result_format_type = get_enum_type(WhatIfResultFormat, WhatIfResultFormat.full_resource_payloads)
+    deployment_what_if_no_pretty_print_type = CLIArgumentType(options_list=['--no-pretty-print'], action='store_true',
+                                                              help='Disables pretty-print for What-If results. When set, the output format type will be used.')
```
zhoxing-ms, line 60: Please use active voice, like `Disables pretty-print` -> `Disable pretty-print`

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -55,6 +55,12 @@ def load_arguments(self, _):
...
+    deployment_what_if_confirmation_type = CLIArgumentType(options_list=['--confirm-with-what-if'], action='store_true',
+                                                           help='Instructs the command to run deployment What-If before excuting the deployment. It then prompts you to acknowledge resource changes before it continues.')
```
zhoxing-ms, line 62: Please use active voice, like `Instructs the command` -> `Instruct the command`

zhoxing-ms, line 62: Parameters `--no-pretty-print` and `--confirm-with-what-if` also look a bit long and may be inconvenient to input. Could you give them a simple alias as well?

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -228,6 +234,9 @@ def load_arguments(self, _):
         c.argument('handle_extended_json_format', arg_type=extended_json_format_type,
                    deprecate_info=c.deprecate(target='--handle-extended-json-format/-j'))
         c.argument('no_prompt', arg_type=no_prompt)
+        c.argument('confirm_with_what_if', arg_type=deployment_what_if_confirmation_type, is_preview=True)
+        c.argument('what_if_result_format', options_list=['--what-if-result-format', '-r'],
+                   arg_type=deployment_what_if_result_format_type, is_preview=True)
```
zhoxing-ms, line 239: Are these two parameters only related to deployment? Why are they also under the command group of resource?

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -275,6 +293,21 @@ def load_arguments(self, _):
...
+
+    with self.argument_context('deployment group what-if') as c:
+        c.argument('deployment_name', arg_type=deployment_create_name_type)
+        c.argument('aux_subscriptions', nargs='+', options_list=['--aux-subs'],
+                   help='Auxiliary subscriptions which will be used during deployment across tenants.',
+                   deprecate_info=c.deprecate(target='--aux-subs', redirect='--aux-tenants'))
```
zhoxing-ms, line 304: The parameters to be discarded should not be added in the new command

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -249,6 +258,15 @@ def load_arguments(self, _):
         c.argument('handle_extended_json_format', arg_type=extended_json_format_type,
                    deprecate_info=c.deprecate(target='--handle-extended-json-format/-j'))
         c.argument('no_prompt', arg_type=no_prompt)
+        c.argument('confirm_with_what_if', arg_type=deployment_what_if_confirmation_type, is_preview=True)
+        c.argument('what_if_result_format', options_list=['--what-if-result-format', '-r'],
+                   arg_type=deployment_what_if_result_format_type, is_preview=True)
+
+    with self.argument_context('deployment sub what-if') as c:
```
zhoxing-ms, line 265: Does the `what if` command of sub need to support `aux_tenants`?

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -383,7 +386,16 @@ def deploy_arm_template_at_subscription_scope(cmd,
                                               template_file=None, template_uri=None, parameters=None,
                                               deployment_name=None, deployment_location=None,
                                               no_wait=False, handle_extended_json_format=None,
+                                              confirm_with_what_if=None, what_if_result_format=None,
                                               no_prompt=False):
```
zhoxing-ms, line 390: It is suggested that the newly added parameters should be placed at the end

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -431,8 +443,18 @@ def deploy_arm_template_at_resource_group(cmd,
                                           resource_group_name=None,
                                           template_file=None, template_uri=None, parameters=None,
                                           deployment_name=None, mode=None, rollback_on_error=None,
-                                          no_wait=False, handle_extended_json_format=None,
-                                          aux_subscriptions=None, aux_tenants=None, no_prompt=False):
+                                          no_wait=False, aux_subscriptions=None, aux_tenants=None,
+                                          confirm_with_what_if=None, what_if_result_format=None,
+                                          no_prompt=False):
```
zhoxing-ms, line 448: Same as above

zhoxing-ms:
Is the `az deployment what-if` mentioned in the PR description not done? If not, please delete it in the PR description.

zhoxing-ms:
Do we need to consider adding tests for the new parameters `confirm_with_what_if` and `what_if_result_format` ?

src/azure-cli/azure/cli/command_modules/resource/_params.py
shenglol: They are not. GitHub folds unchanged code so it looks like they are under the `resource` command group (notice that it's on line 66). If you expand you'll see that they are under `deployment create`.

src/azure-cli/azure/cli/command_modules/resource/_params.py
shenglol: No. If I understand it correctly, the parameter is only useful for cross tenant resource group vm deployment: https://github.com/Azure/azure-cli/issues/11151.

src/azure-cli/azure/cli/command_modules/resource/custom.py
shenglol: That's good to know. I moved them to the end.

src/azure-cli/azure/cli/command_modules/resource/_params.py
shenglol: Done.

src/azure-cli/azure/cli/command_modules/resource/_params.py
shenglol: Changed.

shenglol: I added an alias `-c` for `--confirm-with-what-if`. However, I couldn't think of a good one for `--no-pretty-print`. `-n` might make sense, but it's already taken by `--name`. Do you have any suggestions?

src/azure-cli/azure/cli/command_modules/resource/_params.py
shenglol: Makes sense. I removed it.

src/azure-cli/azure/cli/command_modules/resource/custom.py
shenglol: Good catch. I think it was deleted by accident after resolving some rebasing conflicts. I added it back.

shenglol:
> Is the `az deployment what-if` mentioned in the PR description not done? If not, please delete it in the PR description.

Removed it. `az deployment what-if` was implemented at the beginning, but later I realized that the `az deployment` group will be deprecated, so I removed it from the code. Not sure why I added it to the description though...

shenglol:
> Do we need to consider adding tests for the new parameters `confirm_with_what_if` and `what_if_result_format` ?

I thought about it, but the challenge is that when `--confirm-with-what-if` is set, the `deployment sub/group create` command will print out formatted What-If results and then prompt to ask users to confirm before it continues execution. I'm not sure if it's worthy verifying console output, and if there's a way to mock user behavior in a e2e test. We do have unit tests to very formatted What-If results. Please let me know if that's enough.

src/azure-cli/azure/cli/command_modules/resource/_params.py
zhoxing-ms: I couldn't find a better alias. How about keeping it `--no-pretty-print` or changing it to `--no-pretty`?

qianwens:
@jiasli to help review the colorama related code in custom.py

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -228,6 +234,9 @@ def load_arguments(self, _):
         c.argument('handle_extended_json_format', arg_type=extended_json_format_type,
                    deprecate_info=c.deprecate(target='--handle-extended-json-format/-j'))
         c.argument('no_prompt', arg_type=no_prompt)
+        c.argument('confirm_with_what_if', arg_type=deployment_what_if_confirmation_type, is_preview=True)
```
qianwens, line 237: you need specify `min_api='***'` if this feature is only supported in new API version

src/azure-cli/azure/cli/command_modules/resource/_params.py
shenglol: I think I'm leaning towards keeping `--no-pretty-print`. Although `--no-pretty` is a shorter name, the meaning is less accurate.

src/azure-cli/azure/cli/command_modules/resource/_params.py
shenglol: Added `min_api` constraints.

shenglol:
Hi @zhoxing-ms @qianwens @jiasli is there any other changes I need to make to merge this PR?

zhoxing-ms:
@shenglol Hi, we have other reviewers need to review your PR, please wait for their review first

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -580,6 +600,58 @@ def _deploy_arm_template_at_tenant_scope(cli_ctx,
...
+            # Enable virtual terminal mode for Windows console so it processes color codes.
+            if platform.system() == "Windows":
+                from ._win_vt import enable_vt_mode
+                enable_vt_mode()
+
+        print(format_what_if_operation_result(what_if_result, cli_ctx.enable_color))
+    finally:
+        if cli_ctx.enable_color:
+            from colorama import init
+            init()
```
jiasli, line 650: Colorama usage part looks good to me.

shenglol: Thanks for reviewing it, Jiashuo!

jiasli: BTW, here is the discussion on Colorama: https://github.com/tartley/colorama/issues/160#issuecomment-372687727

shenglol:
@zhoxing-ms @jiasli do you mind helping merge it once the build pipe line passes? It would be great if we can still catch the next release.

src/azure-cli/azure/cli/command_modules/resource/custom.py
shenglol: That would be awesome 😀! Btw I think I found out an small bug related to colorama while test it the pipe with CLI. I'll open an issue in the knack repo soon.

jiasli:
> @zhoxing-ms @jiasli do you mind helping merge it once the build pipe line passes? It would be great if we can still catch the next release.

Absolutely!

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -8,7 +8,7 @@
 def load_arguments(self, _):
     from argcomplete.completers import FilesCompleter
 
-    from azure.mgmt.resource.resources.models import DeploymentMode
+    from azure.mgmt.resource.resources.models import DeploymentMode, WhatIfResultFormat
```
Juliehzl, line 11: Please use get_models() to import model.

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -228,6 +234,9 @@ def load_arguments(self, _):
         c.argument('handle_extended_json_format', arg_type=extended_json_format_type,
                    deprecate_info=c.deprecate(target='--handle-extended-json-format/-j'))
         c.argument('no_prompt', arg_type=no_prompt)
+        c.argument('confirm_with_what_if', arg_type=deployment_what_if_confirmation_type, is_preview=True, min_api='2019-07-01')
```
Juliehzl, line 237: is_preview and min_api also can be moved to deployment_what_if_confirmation_type description.

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -249,6 +258,15 @@ def load_arguments(self, _):
...
+        c.argument('what_if_result_format', options_list=['--what-if-result-format', '-r'],
```
Juliehzl, line 262: I think what_if_result_format also can be abstracted using CLIArgumentType.

src/azure-cli/azure/cli/command_modules/resource/_utils.py
```diff
@@ -0,0 +1,51 @@
...
+# --------------------------------------------------------------------------------------------
+
+import re
+
+_subscription_pattern = r"^\/?subscriptions\/(?P<subscription_id>[a-f0-9-]+)"
+_resource_group_pattern = r"^\/resourceGroups\/(?P<resource_group_name>[-\w\._\(\)]+)"
+_relative_resource_id_pattern = r"^\/providers/(?P<relative_resource_id>.+$)"
+
+
+def parse_resource_id(resource_id):
```
Juliehzl, line 13: Is it possible to utilize  parse_resource_id here?
`from msrestazure.tools import parse_resource_id`

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -55,6 +55,12 @@ def load_arguments(self, _):
     no_prompt = CLIArgumentType(arg_type=get_three_state_flag(), help='The option to disable the prompt of missing parameters for ARM template. '
                                 'When the value is true, the prompt requiring users to provide missing parameter will be ignored. The default value is false.')
 
+    deployment_what_if_result_format_type = get_enum_type(WhatIfResultFormat, WhatIfResultFormat.full_resource_payloads)
```
Juliehzl, line 58: May I know what is the difference for these two formats?

shenglol: With `full_resource_payloads` the What-If API returns detailed resource changes with the resource ID, before state, after state and delta information calculated from the before and after states. With `resource_id_only` the API only return resource changes with resource ID.

src/azure-cli/azure/cli/command_modules/resource/_utils.py
shenglol: Just learnt there's already an existing one. However it may not be suitable to use it here, because it returns a dictionary with unpredictable number of keys, which is dependent on how many child resource namespaces a resource ID has. What I need is just to split a resource id into two parts, the resource scope, and the rest (relative resource ID). If using the existing function I have to process the result dictionary carefully to construct the relative resource ID with the key/value pairs in the result dictionary, which seems to be a harder task. I'm open to changing the utility function's name to `split_resource_id` if that you are okay with that.

src/azure-cli/azure/cli/command_modules/resource/_params.py
shenglol: Done. I didn't changed the two imports below though. I didn't find any ResourceType enum definition for `managedapplications`. Loading it using get_models also failed a number of tests. Since I'm not familiar with the code I'll keep them as is.

shenglol: BTW may I also ask a question about get_models? So when can the `resource_type` be omitted? I searched the code base - it appears that sometimes you need to explicitly pass a ResourceType enum and sometimes you don't.

src/azure-cli/azure/cli/command_modules/resource/_params.py
shenglol: Good to know. I moved them to the argument type definition.

src/azure-cli/azure/cli/command_modules/resource/_params.py
shenglol: Created a CLIArgumentType for it.

shenglol:
@Juliehzl could you take another look at the PR when you get a chance? Thanks!

src/azure-cli/azure/cli/command_modules/resource/_params.py
Juliehzl: when the resource type is the same as the one in __init__  as shown here https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/resource/__init__.py#L19

src/azure-cli/azure/cli/command_modules/resource/_utils.py
Juliehzl: I see

Juliehzl:
> @Juliehzl could you take another look at the PR when you get a chance? Thanks!

LGTM in general but curious about the fantastic function you add. 😊

src/azure-cli/azure/cli/command_modules/resource/tests/latest/test_resource_formatters.py
```diff
@@ -0,0 +1,424 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+import unittest
+
+from azure.mgmt.resource.resources.models import (
```
Juliehzl, line 8: We also need to use get_model() here.

src/azure-cli/azure/cli/command_modules/resource/_params.py
shenglol: Got it. Thanks!

src/azure-cli/azure/cli/command_modules/resource/tests/latest/test_resource_formatters.py
shenglol: Replaced `import` with `get_model()`. Regarding using `mock()`, I don't think it's needed in the formatters unit tests (since there's no dependencies to assert/verify) but I added two tests with mocks to verify prompt logic with `-confirm-with-what-if`, which seems to be worth testing to me.

PR #12949, opened 2020-04-10
{Core} Make CLI respect file logging configuration option: enable_log_file and AZURE_LOGGING_ENABLE_LOG_FILE env variable

haroldrandom:
I am migrating test framwork from nose to pytest and found CLI would log every command's invocation output to a filep even though I configured in ~/.azure/config
```
[logging]
enable_log_file = no
```
And consequently, pytest wouldn't clear the AzCLI instance cleanly such that the test process would crash because of holding too many file descriptor (over the limit by OS):
image
image

**So, here, I want to enable the ability to repect the logging configuratin even though I won't use it in CI. I will fire another PR to fix the bug that CLI holding too many opened files.**

**Testing Guide**  
It affects two instance variables' assignment `command_logger_handler` and `command_metadata_logger`.
In the places they are referenced, there is a fault protection, so, it won't be a problem if user disbale the output of a logger to a log file.

Verify by running command with prefix `AZURE_LOGGING_ENABLE_LOG_FILE=no` or config in ~/.azure/config like above, then there should be no files under `~/.azure/commands/`. Otherwise, there should be a file under that folder

**History Notes:**  

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/azlogging.py
+++ b/src/azure-cli-core/azure/cli/core/azlogging.py
@@ -77,8 +77,10 @@ class AzCliLogging(CLILogging):
 
             cmd_logger = logging.getLogger(AzCliLogging._COMMAND_METADATA_LOGGER)
 
-            self._init_command_logfile_handlers(cmd_logger, args)  # pylint: disable=protected-access
-            get_logger(__name__).debug("metadata file logging enabled - writing logs to '%s'.", self.command_log_dir)
+            if self.file_log_enabled:
+                self._init_command_logfile_handlers(cmd_logger, args)  # pylint: disable=protected-access
+                get_logger(__name__).debug("metadata file logging enabled - writing logs to '%s'.",
+                                           self.command_log_dir)
 
             _delete_old_logs(self.command_log_dir)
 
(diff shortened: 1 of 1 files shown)


yonzhan:
pytest

src/azure-cli-core/azure/cli/core/azlogging.py
```diff
@@ -77,10 +77,15 @@ def _delete_old_logs(log_dir):
 
             cmd_logger = logging.getLogger(AzCliLogging._COMMAND_METADATA_LOGGER)
 
-            self._init_command_logfile_handlers(cmd_logger, args)  # pylint: disable=protected-access
-            get_logger(__name__).debug("metadata file logging enabled - writing logs to '%s'.", self.command_log_dir)
+            # overwrite CLILogging._is_file_log_enabled() from knack
+            self.file_log_enabled = cli_ctx.config.getboolean('logging', 'enable_log_file', fallback=True)
```
fengzhou-msft, line 81: The default for logging is no in az configure.

fengzhou-msft: However, changing it to no could be a breaking change since the bug made the actual default to be logging enabled.

haroldrandom: I overwrite the implementation from knack to enable logging by default like currenty CLI does, so it's transparency to user. Not a breaking it is.

PR #12950, opened 2020-04-10
[Core] fix generic remove property NoneType exception

### LINKED ISSUE #12119: VMSS update failure AttributeError: 'NoneType' object has no attribute 'pop'
## Describe the bug

**Command Name**
`az vmss update`

**Errors:**
```
'NoneType' object has no attribute 'pop'
python3.8/site-packages/knack/cli.py, ln 206, in invoke
cli/core/commands/__init__.py, ln 608, in execute
    raise ex
cli/core/commands/__init__.py, ln 666, in _run_jobs_serially
cli/core/commands/__init__.py, ln 659, in _run_job
    six.reraise(*sys.exc_info())
...
cli/core/commands/__init__.py, ln 306, in __call__
    return self.handler(*args, **kwargs)
cli/core/commands/arm.py, ln 546, in handler
    remove_properties(instance, arg_values)
cli/core/commands/arm.py, ln 921, in remove_properties
    list_to_remove_from.pop(int(list_index))
AttributeError: 'NoneType' object has no attribute 'pop'
```

## To Reproduce:

- `az vmss update --resource-group {} --name {} --remove {} {}`

## Environment Summary
```
macOS-10.15.3-x86_64-i386-64bit
Python 3.8.1
Shell: bash

azure-cli 2.0.81

Extensions:
azure-firewall 0.1.5
aks-preview 0.4.23

```

arrownj:

This PR is trying to fix #12119 

It is also a common problem for generic update.

When we try to remove an element from a list, previously we only catch the `IndexError`. However the list itself could also be None. So we also need to catch `AttributeError` here.

With this change, user will see a clean error message instead of previous trace stack.

**Testing Guide**  
`az vmss create -g myRG -n myVMSS --image UbuntuLTS`

`az vmss update --resource-group myRG--name myVMSS --remove virtualMachineProfile.networkProfile.networkInterfaceConfigurations[0].ipConfigurations[0].loadBalancerInboundNatPools 0`

`az vmss update --resource-group myRG--name myVMSS --remove virtualMachineProfile.networkProfile.networkInterfaceConfigurations[0].ipConfigurations[0].loadBalancerInboundNatPools 0`

Previously, it will print the trace stack. After this change, it will promt the error `index 0 doesn't exist on loadBalancerInboundNatPools `

**History Notes:**  

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/commands/arm.py
+++ b/src/azure-cli-core/azure/cli/core/commands/arm.py
@@ -922,9 +922,11 @@ def remove_properties(instance, argument_values):
         list_to_remove_from = _find_property(instance, list_attribute_path)
         try:
             list_to_remove_from.pop(int(list_index))
-        except (IndexError, AttributeError):
+        except IndexError:
             raise CLIError('index {} doesn\'t exist on {}'
                            .format(list_index, list_attribute_path[-1]))
+        except AttributeError:
+            raise CLIError('{} doesn\'t exist'.format(list_attribute_path[-1]))
 
 
 def throw_and_show_options(instance, part, path):
--- a/src/azure-cli-core/azure/cli/core/commands/arm.py
+++ b/src/azure-cli-core/azure/cli/core/commands/arm.py
@@ -922,7 +922,7 @@ def remove_properties(instance, argument_values):
         list_to_remove_from = _find_property(instance, list_attribute_path)
         try:
             list_to_remove_from.pop(int(list_index))
-        except IndexError:
+        except (IndexError, AttributeError):
             raise CLIError('index {} doesn\'t exist on {}'
                            .format(list_index, list_attribute_path[-1]))
 


qwordy:
Move the description to **Description** section.

src/azure-cli-core/azure/cli/core/commands/arm.py
```diff
@@ -922,7 +922,7 @@ def remove_properties(instance, argument_values):
         list_to_remove_from = _find_property(instance, list_attribute_path)
         try:
             list_to_remove_from.pop(int(list_index))
-        except IndexError:
+        except (IndexError, AttributeError):
```
qwordy, line 925: If it is an AttributeError, do we need to compose a new error message?

arrownj: The error message is `index {} doesn't exist on {}`, I think it is also valid if the list is none.

jiasli:
> Move the description to **Description** section.

Also would like to see some **Testing Guide**.

yonzhan:
add to S168

src/azure-cli-core/azure/cli/core/commands/arm.py
fengzhou-msft: How about when update the property? There are multiple `IndexError` in this file, maybe double check all of them?

arrownj:
> Move the description to Description section.
> 
> Also would like to see some Testing Guide.

**Test Guide** added

src/azure-cli-core/azure/cli/core/commands/arm.py
arrownj: > How about when update the property? There are multiple IndexError in this file, maybe double check all of them?

Confirmed 2 other places, both already catch AttributeError.

jiasli: I agree with @qwordy. Something like "_{list_attribute_path} doesn't exist_" sounds more reasonable.

arrownj: > I agree with @qwordy. Something like "{list_attribute_path} doesn't exist" sounds more reasonable.

message updated as recommend.

src/azure-cli-core/azure/cli/core/commands/arm.py
```diff
@@ -925,6 +925,8 @@ def remove_properties(instance, argument_values):
         except IndexError:
             raise CLIError('index {} doesn\'t exist on {}'
                            .format(list_index, list_attribute_path[-1]))
+        except AttributeError:
+            raise CLIError('{} doesn\'t exist'.format(list_attribute_path[-1]))
```
jiasli, line 929: On a second thought, we can use `if not list_to_remove_from: raise ...`. But it's OK. No difference.

PR #12954, opened 2020-04-10
[RBAC] az ad app create/update: support --optional-claims as a parameter

### LINKED ISSUE #11534: optional claims through azure cli
I am trying to create app registration through cli. I need to have AD return the group information as roles. I am trying to get this done by setting the optional claims additional properties as 'emit_as_roles' in manifest.json. But this is not happening. Would have been nice if there is an option to 'az ad app create --display-name testapp **_--optional-claims_** manifest.json'

I was wondering if this optionalClaims to the app registration can be done via any other way. Please help.

arrownj:

This PR is to fix #11534 

There are three attributes of optional claims, `idToken`, `accessToken` and `samlToken`.

As a result of the graph api version problem, version 2.0 has deprecated `samlToken` and use a new `saml2Token` instead. Although we use version 1.6, service side still not accept `samlToken` any more. So Azure CLI will only support `idToken` and `accessToken`.

This may change if we migrate to the latest Microsoft Graph API in the future.

**Testing Guide**  
Please reference the case in test_graph.py.

**History Notes:**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/role/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/role/_help.py
@@ -47,6 +47,26 @@ examples:
             "isEnabled": "true",
             "value": "approver"
         }]
+  - name: Create an application with optional claims
+    text: |
+        az ad app create --display-name mytestapp --optional-claims @manifest.json
+        ("manifest.json" contains the following content)
+        {
+            "idToken": [
+                {
+                    "name": "auth_time",
+                    "source": null,
+                    "essential": false
+                }
+            ],
+            "accessToken": [
+                {
+                    "name": "email",
+                    "source": null,
+                    "essential": false
+                }
+            ]
+        }
 """
 
 helps['ad app credential'] = """
@@ -220,6 +240,26 @@ examples:
             "isEnabled": "true",
             "value": "approver"
         }]
+  - name: update optional claims
+    text: |
+        az ad app update --id e042ec79-34cd-498f-9d9f-123456781234 --optional-claims @manifest.json
+        ("manifest.json" contains the following content)
+        {
+            "idToken": [
+                {
+                    "name": "auth_time",
+                    "source": null,
+                    "essential": false
+                }
+            ],
+            "accessToken": [
+                {
+                    "name": "email",
+                    "source": null,
+                    "essential": false
+                }
+            ]
+        }
   - name: update an application's group membership claims to "All"
     text: >
         az ad app update --id e042ec79-34cd-498f-9d9f-123456781234 --set groupMembershipClaims=All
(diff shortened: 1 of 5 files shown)


yonzhan:
add to S168

src/azure-cli/azure/cli/command_modules/role/_help.py
```diff
@@ -47,6 +47,26 @@
             "isEnabled": "true",
             "value": "approver"
         }]
+  - name: Create an application with optional claims
+    text: |
+        az ad app create --display-name mytestapp --optional-claims @manifest.json
+        ("manifest.json" contains the following content)
```
yungezz, line 53: how about some description on fields of manifest.json

src/azure-cli/azure/cli/command_modules/role/custom.py
```diff
@@ -1025,6 +1032,29 @@ def _build_app_roles(app_roles):
...
+def _build_optional_claims(optional_claims):
+    if not optional_claims:
+        return None
+    available_keys = {
+        'idToken': 'id_token',
+        'accessToken': 'access_token'
+    }
+    result = OptionalClaims()
+    for key, object_key in available_keys.items():
+        if optional_claims.get(key, []):
```
yungezz, line 1044: since no schema check on json content, what if input optional_claims doesn't contain either idToken or accessToken?

arrownj: It will skip these properties if it is not in the json file. It's because none of these properties is required.

src/azure-cli/azure/cli/command_modules/role/_help.py
arrownj: I checked the swagger file, the descriptions of `idToken` and `accessToken` are just same with the name. Do you mean that we need to clarify why we only support these two properties rather than `smalToken` ?

yungezz: not why, but clear description/guide/sample(either way ok) of each fields, what's name/source/essentail mean to help user write a correct manifest.json

src/azure-cli/azure/cli/command_modules/role/tests/latest/test_graph.py
```diff
@@ -543,3 +543,47 @@ def test_graph_required_access_e2e(self):
         finally:
             if app_id:
                 self.cmd('ad app delete --id ' + app_id)
+
+
+class ApplicationOptionalClaimsScenarioTest(ScenarioTest):
+
+    def test_application_optional_claims(self):
```
jiasli, line 550: Maybe consider moving the test to `ApplicationSetScenarioTest.test_application_set_scenario`?

src/azure-cli/azure/cli/command_modules/role/_help.py
jiasli: FYI, the docs are at [optionalClaims attribute](https://docs.microsoft.com/en-us/azure/active-directory/develop/reference-app-manifest#optionalclaims-attribute) and [OptionalClaim type](https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-optional-claims#optionalclaim-type).

arrownj: > not why, but clear description/guide/sample(either way ok) of each fields, what's name/source/essentail mean to help user write a correct manifest.json

I add OptionalClaim properties doc link in the help message of `--optional-claims`. It's because I think it will be a little strange to put the description in the example part. Do you think it is OK?

src/azure-cli/azure/cli/command_modules/role/tests/latest/test_graph.py
arrownj: moved.

Issue #12882, opened 2020-04-06
Logging leaks file handles leading to FD exhaustion for long running scripts


elpollouk:
**Describe the bug**
Each call to `get_default_cli().invoke(...)` will leak a file handle due to logging.

**To Reproduce**
Run this script under Linux:
```
from azure.cli.core import get_default_cli
import os
import tempfile

for _ in range(10):
    with tempfile.TemporaryFile('w+t') as temp:
        get_default_cli().invoke(["account", "show"], out_file=temp)

open_fds = os.listdir(f"/proc/{os.getpid()}/fd")
print("Leaked file handles:")
for fd in open_fds:
    try:
        target = os.readlink(f"/proc/{os.getpid()}/fd/{fd}")
        print(f"  {fd} - {target}")
    except:
        pass
```
**Expected behavior**
When the `invoke()` command returns, all internal resources specific to the child process are released.

**Environment summary**
Untunu 18.04
Python 3.6.9
azure-cli 2.3.1
azure-cli-core 2.3.1

**Additional context**
The problem is due to [azlogging.py](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/azure/cli/core/azlogging.py) not cleaning up the `logging.FileHandler` resource allocation on [line 101](https://github.com/Azure/azure-cli/blob/a29b6a82566303b3c2f0e3489f52ec1ef1bffec3/src/azure-cli-core/azure/cli/core/azlogging.py#L101)

I did try updating `end_cmd_metadata_logging` to clean up the file handler, but this method is not called if the `az` process errors, e.g. a problem with the user's credentials. This is because exception handling bypasses most clean up code.

yonzhan:
add to S169

haroldrandom:
Hi, @elpollouk , we will fix the bug that CLI doesn't close fd of logger properly in `end_cmd_metadata_logging`  https://github.com/Azure/azure-cli/pull/12971

Issue #10435, opened 2019-09-05
Azure command logging is leaking file handles


rickardp:
## Describe the bug
When running the azure-cli from within Python and running commands in a loop, the azlogging.py seems to leak file handles.

## To Reproduce:
Run an 'az' command using the Python library in a loop.

## Expected Behavior
Files are deleted, file handles are closed. Disk space is reclaimed.

## Observed behavior

Files are deleted, but left open. `lsof` gives:

```
1       /usr/local/bin/python3.7        /root/.azure/commands/2019-09-05.08-45-54.resource_list.1.log (deleted)
1       /usr/local/bin/python3.7        /root/.azure/commands/2019-09-05.08-45-55.resource_list.1.log (deleted)
1       /usr/local/bin/python3.7        /root/.azure/commands/2019-09-05.08-45-57.resource_list.1.log (deleted)
1       /usr/local/bin/python3.7        /root/.azure/commands/2019-09-05.08-45-58.resource_list.1.log (deleted)
1       /usr/local/bin/python3.7        /root/.azure/commands/2019-09-05.08-45-59.monitor_metrics_list.1.log (deleted)
1       /usr/local/bin/python3.7        /root/.azure/commands/2019-09-05.08-46-00.monitor_metrics_list.1.log (deleted)
1       /usr/local/bin/python3.7        /root/.azure/commands/2019-09-05.08-46-01.monitor_metrics_list.1.log (deleted)
1       /usr/local/bin/python3.7        /root/.azure/commands/2019-09-05.08-46-02.monitor_metrics_list.1.log (deleted)
1       /usr/local/bin/python3.7        /root/.azure/commands/2019-09-05.08-46-08.monitor_metrics_list.1.log (deleted)
1       /usr/local/bin/python3.7        /root/.azure/commands/2019-09-05.08-46-09.monitor_metrics_list.1.log (deleted)
1       /usr/local/bin/python3.7        /root/.azure/commands/2019-09-05.08-46-10.monitor_metrics_list.1.log (deleted)
1       /usr/local/bin/python3.7        /root/.azure/commands/2019-09-05.08-46-11.monitor_metrics_list.1.log (deleted)
1       /usr/local/bin/python3.7        /root/.azure/commands/2019-09-05.08-46-13.monitor_metrics_list.1.log (deleted)
1       /usr/local/bin/python3.7        /root/.azure/commands/2019-09-05.08-46-14.monitor_metrics_list.1.log (deleted)
[...]

PR #12971, opened 2020-04-11
{Core} Reducing the risk of logging file fd leaking

haroldrandom:
**Description**  
As I discovered in https://github.com/Azure/azure-cli/pull/12949, CLI would crash after long time running due to running out of available fd because it didn't close the fd properly.

Running tests with pytest, testing is terminated after the number of avaliable fd is consumed.
image
image

I add a [hook `doCleanups` of unittest.Testcase](https://docs.python.org/3/library/unittest.html#unittest.TestCase.doCleanups) to mandatorily clean those unclosed fd hold by CLI's logger. 

**This could reduce the risk of fd leaking but still need developer to manually call `end_cmd_metadata_logging()` of AzLogging, like:**
https://github.com/Azure/azure-cli/blob/806e4318d058c3037566fa5a7a07cfc99d1ff51d/src/azure-cli/azure/cli/__main__.py#L49-L50

**Why I say reducing**
1.  there are still some test classes didn't inherit from `ScenarioTest/LiveScenarioTest` but `unittest.TestCase`, so, they couldn't benifit from this fix.
2.  A log files rotated logic is applied to delete older log files to keep at most 25 log files in order to save the disk. If there are too many commands/tests running at the same time and CLI instance can't recover by Python Garbage Collector, it will keep leak the fd until CLI instance is del or the process is terminated. Like these 2 PRs: https://github.com/Azure/azure-cli/issues/12882, https://github.com/Azure/azure-cli/issues/10435

In both scenarios, we will still have chance to see `(deleted)` files that are unclosed and leaking

**Testing Guide**  
1.  uncomment the new added lines and run a test with lots of test units like `test_network_commands.py`,
2.  what over the opened fd by your process, `lsof -p {TestProcessID} | grep "azure"`,
3.  After a while, you will see lots of items end up with `(deleted)` like the image above. That's the leaking fds. 
4.  If you run with pytest `pytest -x -v --capture=no` and with tons of tests, process will crash.
5.  Adding the new added line in this PR, there are still sometime that the **deleted** file will show in `lsof -p {TestProcessID}`, but they will disappear very soon.

**History Notes**  

---


yonzhan:
add to S168

jiasli:
Handlers are added by `azure.cli.core.azlogging.AzCliLogging._init_command_logfile_handlers` which is called when `EVENT_INVOKER_PRE_CMD_TBL_TRUNCATE` is [raised](https://github.com/Azure/azure-cli/blob/34f9033b96e6ea31f82626a9f2b152c1a061300e/src/azure-cli-core/azure/cli/core/commands/__init__.py#L503). 

I think we can use the same mechanism to call "`deinit_command_file_logging`" duing

The only place where `end_cmd_metadata_logging` is called is [azure-cli/azure/cli/__main__.py](https://github.com/Azure/azure-cli/blob/1bab7985899f267b208cfe6048fb253485357c4c/src/azure-cli/azure/cli/__main__.py#L50) which is never called during a test case, because it is generating its own dummy cli:

https://github.com/Azure/azure-cli/blob/74d9ee52a5606d741ea69762376f9e8c65e58e21/src/azure-cli-testsdk/azure/cli/testsdk/base.py#L181

jiasli:
Handlers are added by `azure.cli.core.azlogging.AzCliLogging._init_command_logfile_handlers`:

https://github.com/Azure/azure-cli/blob/9015bb49290547da4615a7238ce88b09c0060d54/src/azure-cli-core/azure/cli/core/azlogging.py#L106

which is called when `EVENT_INVOKER_PRE_CMD_TBL_TRUNCATE` is raised

https://github.com/Azure/azure-cli/blob/9015bb49290547da4615a7238ce88b09c0060d54/src/azure-cli-core/azure/cli/core/azlogging.py#L48

https://github.com/Azure/azure-cli/blob/34f9033b96e6ea31f82626a9f2b152c1a061300e/src/azure-cli-core/azure/cli/core/commands/__init__.py#L503-L504

I think we can use the same mechanism to call `end_cmd_metadata_logging` in https://github.com/Azure/azure-cli/blob/34f9033b96e6ea31f82626a9f2b152c1a061300e/src/azure-cli-core/azure/cli/core/commands/__init__.py#L490 

Currently, the only place where `end_cmd_metadata_logging` is called is

https://github.com/Azure/azure-cli/blob/1bab7985899f267b208cfe6048fb253485357c4c/src/azure-cli/azure/cli/__main__.py#L50

which is never called during in a `ScenarioTest`, because `ScenarioTest` is generating its own dummy cli:

https://github.com/Azure/azure-cli/blob/74d9ee52a5606d741ea69762376f9e8c65e58e21/src/azure-cli-testsdk/azure/cli/testsdk/base.py#L181

src/azure-cli-core/azure/cli/core/azlogging.py
```diff
@@ -177,6 +177,10 @@ def end_cmd_metadata_logging(self, exit_code):
         if self.command_metadata_logger:
             self.command_metadata_logger.info("exit code: %s", exit_code)
 
+            for handler in self.command_metadata_logger.handlers[:]:
+                handler.close()
+                self.command_metadata_logger.removeHandler(handler)
```
qianwens, line 182: >  self.command_metadata_logger.removeHandler(handler) [](start = 14, length = 53)

remove elements in the array loop may cause unexpected result

haroldrandom: it's a copy from slice `[:]`, won't have side affect.

haroldrandom:
@jiasli What evet name do you suggest?

zhoxing-ms:
I see that the same log handler(delete) appears repeatedly in the second picture. Could we design the log handler object which handle the same type of business as a singleton? This can also avoid repeated creation after multiple calls and reduce the fd.

haroldrandom:
> I see that the same log handler(delete) appears repeatedly in the second picture. Could we design the log handler object which handle the same type of business as a singleton? This can also avoid repeated creation after multiple calls and reduce the fd.

Good suggestion. Can singletion affect the logging order? If CLI were used concurrenly, will it affect anything? I am concered about it.

zhoxing-ms:
The idea comes from the fact that when I used a logging framework in Java before, I always create a LogFactory to generates singleton logger objects for different businesses:
1. The logger object performs the printed invocation method mainly according to the log settings of corresponding business and does not affect the log printing order.
2. Because different business threads use different logger objects, they do not affect each other. And the configuration of multi-threaded print logs for the same business should be the same, so singletons can reduce unnecessary re-creation of objects.

But I don't have a deep understanding for Python's log, just a guess based on my previous experience in Java, maybe you can use it as a reference

src/azure-cli-core/azure/cli/core/azlogging.py
elpollouk: Will this get cleaned up if the `az` process returns an error? I seem to remember the call to `end_cmd_metadata_logging` being skipped if an exception was raised when I was testing as it was only called on a successful path rather than from a `finally` block.

haroldrandom: @elpollouk If you were using `az`, that would go with code in `__main__.py` 

https://github.com/Azure/azure-cli/blob/1bab7985899f267b208cfe6048fb253485357c4c/src/azure-cli/azure/cli/__main__.py#L48-L73

whenever an exception is raised, the final exception on top of exception stack would be `SystemExit` because there is a hook:
https://github.com/Azure/azure-cli/blob/4770a437fb9fbef119b50248bc9853c9e3fb9855/src/azure-cli-core/azure/cli/core/parser.py#L148-L157

That's my understanding, if anything missing/wrong, could you please give a detailed report?

haroldrandom:
> The idea comes from the fact that when I used a logging framework in Java before, I always create a LogFactory to generates singleton logger objects for different businesses:
> 
> 1. The logger object performs the printed invocation method mainly according to the log settings of corresponding business and does not affect the log printing order.
> 2. Because different business threads use different logger objects, they do not affect each other. And the configuration of multi-threaded print logs for the same business should be the same, so singletons can reduce unnecessary re-creation of objects.
> 3. If the logger object of the same business is often created repeatedly, the singleton pattern is more recommended. Otherwise, it might be better to use up and release.
> 
> However, I don't have a deep understanding for Python's log, just a guess based on my previous experience in Java, maybe you can use it as a reference

Got it. 
The Python logging module does this similarly. the logging module will deal with duplicated logger once we call `logging.getLogger(AzCliLogging._COMMAND_METADATA_LOGGER)`.
But once logger can hold more than one file handlers, meanwhile different handler can hold different fd to the same logger file if the program executes one command more that 1 time and fast because the file name ends with minutes only:
https://github.com/Azure/azure-cli/blob/ec56c6681ba1a974d45af8e5d4c327bfd0f3372b/src/azure-cli-core/azure/cli/core/azlogging.py#L99-L111

If we are going to implement the  singleton on logger level, I think it's unncessary for now. Because it's more meaningful under the cocurrent scenaro. And if we decide to make it singleton, the feedback it's involved to change because it needs logs.

haroldrandom:
Closed due to mistaken merged.

PR #12999, opened 2020-04-14
{Core} support get msal accesstoken with adal refresh token for vm ssh feature

arrownj:
**Description**  
This PR is try to provide an workaround solution for vmssh on adal.

As vmssh feature can only be implemented based on a msal access token. After investigation, we find it's possible to get msal access token with adal refresh token. Thus we can unblock vmssh before we adapt to msal completely.

These codes may be deprecated after the msal adoption finished.

**Testing Guide**  
It should be tested through `ssh` extension which is in developing.

**History Notes**  

---


src/azure-cli-core/azure/cli/core/_msal.py
```diff
@@ -0,0 +1,30 @@
+from knack.util import CLIError
+from msal import ClientApplication
+
+
+class SSHCertificateClientApplication(ClientApplication):
+    """
+    This class is added only for vmssh feature.
+    This is a temporary solution and will deprecate after adoption to MSAL completely.
+    """
+    def _acquire_token_silent_by_finding_rt_belongs_to_me_or_my_family(
```
yungezz, line 10: seems account not used, how about remove this wrapper?

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -1008,6 +1029,29 @@ def retrieve_token_for_user(self, username, tenant, resource):
             self.persist_cached_creds()
         return (token_entry[_TOKEN_ENTRY_TOKEN_TYPE], token_entry[_ACCESS_TOKEN], token_entry)
 
+    def retrieve_ssh_certificate_for_user(self, tenant, modulus, exponent, refresh_token):
+        """
+        This is added only for vmssh feature.
+        It is a temporary solution and will deprecate after adoption to MSAL completely.
+        """
+        from azure.cli.core._msal import SSHCertificateClientApplication
+        scopes = ["https://pas.windows.net/CheckMyAccess/Linux/user_impersonation"]
```
yungezz, line 1038: will this be different under different cloud env?

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -1008,6 +1029,29 @@ def retrieve_token_for_user(self, username, tenant, resource):
...
+        tenant = tenant or 'organizations'
+        authority = self._ctx.cloud.endpoints.active_directory + '/' + tenant
+        app = SSHCertificateClientApplication(_CLIENT_ID, authority=authority)
+        jwk = {
+            "kty": "RSA",
+            "n": modulus,
+            "e": exponent,
+            "kid": str(hash((modulus, exponent)))
```
rlrossiter, line 1046: This shouldn't be using the built-in hash because the value changes between runs. Here's the code that I've modified it to locally to make it generate the same hash each run to use the same SSH cert out of the token cache.

```
        key_hash = hashlib.sha256()
        key_hash.update(modulus.encode('utf-8'))
        key_hash.update(exponent.encode('utf-8'))
        key_id = key_hash.hexdigest()

        jwk = {
            "kty": "RSA",
            "n": modulus,
            "e": exponent,
            "kid": key_id
        }
```

yonzhan:
vm ssh

src/azure-cli-core/azure/cli/core/_profile.py
arrownj: Thanks @rlrossiter . Updated.

src/azure-cli-core/azure/cli/core/_msal.py
arrownj: It's a method which defined in the parent class. We'd better not to change the method signature  in subclasses.

src/azure-cli-core/azure/cli/core/_profile.py
arrownj: Currently there is only one link. This was described in [this document](https://microsoft.sharepoint.com/:w:/r/teams/AutomationConfigAll/_layouts/15/guestaccess.aspx?rtime=MB8KjvQD10g&share=EXj-_T0zafdLpB63hyi82fUBLnUfp9t-isLgaNOh6KTzCQ). The document says it may changed to `sshservice.azure.net` in the future.

src/azure-cli-core/azure/cli/core/commands/ssh_credential_factory.py
```diff
@@ -0,0 +1,30 @@
...
+This is a temporary solution and will deprecate after adoption to MSAL completely.
+The function is try to retrieve MSAL access token with ADAL refresh token.
+"""
+from knack import log
+
+
+logger = log.get_logger(__name__)
+
+
+def get_ssh_credentials(cli_ctx, modulus, exponent):
```
qianwens, line 16: can you add test for this api

qianwens, line 16: modules?

yungezz: i have exact same question :) , then I did a search...

arrownj: @qianwens @yungezz , these codes were original added by Ryan, however I will take the ownership since then. I will add a test for this function.

yonzhan:
vm ssh

src/azure-cli-core/azure/cli/core/_profile.py
qianwens: This feature currently seems only supported in public cloud [https://docs.microsoft.com/en-us/azure/active-directory/devices/howto-vm-sign-in-azure-ad-windows#supported-azure-regions-and-windows-distributions](https://docs.microsoft.com/en-us/azure/active-directory/devices/howto-vm-sign-in-azure-ad-windows#supported-azure-regions-and-windows-distributions), but we should add this endpoint in the cloud.py and throw error for non-public cloud

---
In reply to: [412720221](https://github.com/Azure/azure-cli/pull/12999#discussion_r412720221) [](ancestors = 412720221)

src/azure-cli-core/azure/cli/core/commands/ssh_credential_factory.py
qianwens: modulus is correct:smile: it means 模數

---
In reply to: [413529228](https://github.com/Azure/azure-cli/pull/12999#discussion_r413529228) [](ancestors = 413529228)

src/azure-cli-core/azure/cli/core/_profile.py
qianwens: I think maybe you should let the client pass scopes and data to get token to make this method more generic, this method contain too much implementation detail for SSHcert

---
In reply to: [413537394](https://github.com/Azure/azure-cli/pull/12999#discussion_r413537394) [](ancestors = 413537394,412720221)

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -525,6 +525,16 @@ def get_access_token_for_resource(self, username, tenant, resource):
             username, tenant, resource)
         return access_token
 
+    def _get_ssh_certificate_for_resource(self, tenant, modulus, exponent):
```
qianwens, line 528: >ssh_certificate [](start = 13, length = 15)

please just implement a method like get_msal_for_resource and put the SSH related logic in vm module

src/azure-cli-core/azure/cli/core/_profile.py
arrownj: > I think you should let the client pass scopes and data to get token to make this method more generic, this method contain too much implementation detail for SSHcert
> 
> In reply to: [413537394](https://github.com/Azure/azure-cli/pull/12999#discussion_r413537394) [](ancestors = 413537394,412720221)

Agreed. I will change it.

Issue #12858, opened 2020-04-02
List of VM SKUs which support accelerated networking has some undocumented SKUs


arvindshmicrosoft:
The [code](https://github.com/Azure/azure-cli/blob/86d8163b9baaac785829c0b45731501979b466cd/src/azure-cli/azure/cli/command_modules/vm/_validators.py#L712-L744) which checks for VM SKUs supporting accelerated networking (AN) has a different list of SKUs compared to enumeration of SKUs via command like the below.
```
az vm list-skus --resource-type virtualMachines --query '[].{name:name, an_enabled:capabilities[?name==`AcceleratedNetworkingEnabled`].value | [0]} | [?an_enabled==`True`][].name | sort(@)'  | uniq
```
In the existing code there are also several SKUs which are not regularly consumable or documented, and some outright errors. For example Standard_PB6s does not currently support AN - based on practical testing. Some other SKUs like AZAP_Performance_ComputeV17C_DDA, Standard_D3_v2_ABC are not even documented anywhere.

What I'm looking for by filing this issue is a clarification on how the list in the current code was arrived at, and if you are open to a PR to align the CLI code with a more accurate list as returned by the capabilities method I mentioned.

yungezz:
hi @qwordy could you pls have a look

arvindshmicrosoft:
hello again, just pinging to check if you had a chance to review this item?

qwordy:
The list of SKUs in code is written in 2018. It may be out of date. I wonder why CLI makes decision for users. CLI will enable it based on image and size. I can update the list now. But it is hard to keep the list always up to date. We don't know when the list should be updated. @yugangw-msft This piece of code is written by you. Can you share the story of this design? Do you think we can return the right to users?
```
--accelerated-networking
Enable accelerated networking. Unless specified, CLI will enable it based on machine image and size.
```

arvindshmicrosoft:
Thank you @qwordy - I believe if we can switch to using the [ResourceSkus.list](https://docs.microsoft.com/en-us/rest/api/compute/resourceskus/list) API and then filtering based on the capabilities where the SKU is AccelNet capable, it would be a robust way.

qwordy:
Thank you @arvindshmicrosoft . It is a robust way. And it does not break the current behavior of this parameter.

arvindshmicrosoft:
If it helps, I have some Python code which already uses this API in some Ansible related automation. You could consider leveraging something similar in CLI.
```
[...]

PR #13045, opened 2020-04-16
[Compute] Update VM SKUs that support accelerated networking

qwordy:
**Description**  

https://github.com/Azure/azure-cli/issues/12858
VM SKUs that support accelerated networking are hard coded in code. Last time when they are updated is in 2018. They are stale now. I change it to fetching from service on the fly so that they are always up to date.

**Testing Guide**  

**History Notes**  

---


yonzhan:
add to S168

src/azure-cli/azure/cli/command_modules/vm/_validators.py
```diff
@@ -714,40 +714,10 @@ def _validate_vm_vmss_accelerated_networking(cli_ctx, namespace):
...
-                      'Standard_L80s_v2', 'Standard_M8ms', 'Standard_M8-4ms', 'Standard_M8-2ms', 'Standard_M16ms',
-                      'Standard_M16-8ms', 'Standard_M16-4ms', 'Standard_M32ms', 'Standard_M32-8ms',
-                      'Standard_M32-16ms', 'Standard_M32ls', 'Standard_M32ts', 'Standard_M64ls', 'Standard_E64i_v3',
-                      'Standard_E64is_v3', 'Standard_E4-2s_v3', 'Standard_E8-4s_v3', 'Standard_E8-2s_v3',
-                      'Standard_E16-4s_v3', 'Standard_E16-8s_v3', 'Standard_E20s_v3', 'Standard_E20_v3']
-        aval_sizes = [x.lower() for x in aval_sizes]
+
+        skus = list_sku_info(cli_ctx, namespace.location)
+        aval_sizes = [x.name.lower() for x in skus if x.resource_type == 'virtualMachines' and
+                      any(c.name == 'AcceleratedNetworkingEnabled' and c.value == 'True' for c in x.capabilities)]
```
arrownj, line 720: only list AcceleratedNetworkingEnabled skus ?

qwordy: Yes. Then enable accelerated networking for users if the size support this feature. I have some reservations for this design.

yungezz:
hi @qwordy is this PR ready to merge after CI pass?

qwordy:
> 
> 
> hi @qwordy is this PR ready to merge after CI pass?

No. The data is not convincing. I consulted VM people in the issue.

yungezz:
hi @qwordy any update on this PR? thanks

qwordy:
> 
> 
> hi @qwordy any update on this PR? thanks

According to the issue, a service engineer commented 16 days ago. No update yet.

qwordy:
This PR causes more than 100 test failures. It affects all `vm/vmss create` commands.

yungezz:
> This PR causes 100+ test failures in replay mode. It affects all `vm/vmss create` commands. When I rerun them in live mode, still 100+ errors. Most of them are large response errors because list_sku_info returns very large response.

hi @qwordy I remembered that you mentioned add a decorator for this. What's status of it?

qwordy:
> 
> 
> > This PR causes 100+ test failures in replay mode. It affects all `vm/vmss create` commands. When I rerun them in live mode, still 100+ errors. Most of them are large response errors because list_sku_info returns very large response.
> 
> hi @qwordy I remembered that you mentioned add a decorator for this. What's status of it?

The test case author needs to add a large response decorator if the response is large.

qwordy:
This PR fails all vm create tests in replay mode and all of them need to be rerun lively. 
And I don't agree with the existing behavior. Why we make decision for user to enable accelerated networking automatically?
I suggest abandon this PR. The motivation is wrong! We should correct it to the right way, but not improve it in the opposite way.

src/azure-cli/azure/cli/command_modules/vm/_validators.py
```diff
@@ -706,12 +706,68 @@ def _validate_vm_vmss_accelerated_networking(cli_ctx, namespace):
         size = getattr(namespace, 'size', None) or getattr(namespace, 'vm_sku', None)
         size = size.lower()
 
+        # Deprecated
```
yungezz, line 709: it would be nice to add detail why deprecate

yungezz:
> This PR fails all vm create tests in replay mode and all of them need to be rerun lively.
> And I don't agree with the existing behavior. Why we make decision for user to enable accelerated networking automatically?
> I suggest abandon this PR. The motivation is wrong! We should correct it to the right way, but not improve it in the opposite way.

thanks @qwordy . The original intention is to just by default `on` for user. I don't see problem of default `on`. Will it cause other impact? If the concern is that dynamic `list_sku_infos` too time consuming, can we manual maintain the hardcoded sku list to keep them up-to-date?

qwordy:
> 
> 
> > This PR fails all vm create tests in replay mode and all of them need to be rerun lively.
> > And I don't agree with the existing behavior. Why we make decision for user to enable accelerated networking automatically?
> > I suggest abandon this PR. The motivation is wrong! We should correct it to the right way, but not improve it in the opposite way.
> 
> thanks @qwordy . The original intention is to just by default `on` for user. I don't see problem of default `on`. Will it cause other impact? If the concern is that dynamic `list_sku_infos` too time consuming, can we manual maintain the hardcoded sku list to keep them up-to-date?

The current status of PR does not change default on. It just updates the list. And I comment that how to update the list in the code.

yungezz:
> > > This PR fails all vm create tests in replay mode and all of them need to be rerun lively.
> > > And I don't agree with the existing behavior. Why we make decision for user to enable accelerated networking automatically?
> > > I suggest abandon this PR. The motivation is wrong! We should correct it to the right way, but not improve it in the opposite way.
> > 
> > 
> > thanks @qwordy . The original intention is to just by default `on` for user. I don't see problem of default `on`. Will it cause other impact? If the concern is that dynamic `list_sku_infos` too time consuming, can we manual maintain the hardcoded sku list to keep them up-to-date?
> 
> The current status of PR does not change default on. It just updates the list. And I comment that how to update the list in the code.

yes, got it. above comments are for your previous comments that `should abandon this PR`, and cocnern on existing behavior

src/azure-cli/azure/cli/command_modules/vm/_validators.py
qwordy: Let me delete these 3 lines.

jkewley:
Point of clarification - the strategy here is to update the hard coded list on a regular basis? Is there some sort of automation or process which will drive a refresh when appropriate, like when VM capabilities are updated??

qwordy:
> 
> 
> Point of clarification - the strategy here is to update the hard coded list on a regular basis? Is there some sort of automation or process which will drive a refresh when appropriate, like when VM capabilities are updated??

I suggest not enable accelerated networking automatically for users. But I keep the original behavior. Currently, no. We are considering supporting it.

PR #13066, opened 2020-04-17
[ARO] az aro: Fix table output

m1kola:
**Description**

Fixes a bug in `az aro {command} --output table` (`az aro list --output table`, for example) when the comamnd fails due to not taking into account potential `None` values in data in some edge cases table output currently fails. For example:

* When there are no worker profiles (cluster only has masters or we failed to get data from API server)
* When a worker profile does not have count (machine set was scaled to 0)

**Testing Guide**  

Create an ARO cluster, scale all machine sets to 0 and try to run `az aro list --output table`.

**History Notes**  

[ARO] az aro: Fix table output

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/aro/_format.py
+++ b/src/azure-cli/azure/cli/command_modules/aro/_format.py
@@ -15,11 +15,13 @@ def aro_list_table_format(results):
 def aro_show_table_format(result):
     parts = parse_resource_id(result['id'])
 
+    worker_profiles = result['workerProfiles'] or []
+
     return collections.OrderedDict(
         Name=result['name'],
         ResourceGroup=parts['resource_group'],
         Location=result['location'],
         ProvisioningState=result['provisioningState'],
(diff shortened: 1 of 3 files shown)


yonzhan:
add to S169

haroldrandom:
Could you please add \_\_init\_\_.py under `tests/` and its child directory `latest/`?

m1kola:
Good catch! Added `__init__.py` into these dirs. But I think we also need to update request recordings. I'll look into how to do it now.

haroldrandom:
> Good catch! Added `__init__.py` into these dirs. But I think we also need to update request recordings. I'll look into how to do it now.

One reason why CI failed is azure-mgmt-network has bumped to a newer version, you could re-recordin test to resolve this.
But still need you fix the problem for your modifcation if any.

m1kola:
@jiasli @haroldrandom please take a look. I rebased on top of `dev` branch which now has `__init__.py` files for tests and new recordings and added some extra test for table output.

PR #13067, opened 2020-04-17
{REST} az rest: Refine help

jiasli:
**Description**  
Refine the help doc of `az rest`.

Also use the correct name `URL` instead of `URI` for the HTTP request. See [What is the difference between a URI, a URL and a URN?](https://stackoverflow.com/questions/176264/what-is-the-difference-between-a-uri-a-url-and-a-urn).

**Testing Guide**  
```
az rest --help
```


yonzhan:
S169

Issue #12090, opened 2020-02-07
'az webapp create' with different resource group to plan


danielbecroft:
## Describe the bug
It is not possible to create a Web App in a different resource group to the App Service Plan. This functionality is possible in the Azure Portal UI.

Running `az webapp create` to create a webapp in a different resource group to the service plan results in an error message. 

**Command Name**
`az webapp create`

**Errors:**
```
az webapp create: error: 'NoneType' object has no attribute 'location'
```

## To Reproduce:

```powershell
$APP_SERVICE_PLAN_NAME = "appserviceplan"
$PLAN_RESOURCE_GROUP_NAME = "appservice-plan-group"
$APP_RESOURCE_GROUP_NAME = "webapp-group"

az group create --name $PLAN_RESOURCE_GROUP_NAME --location australiaeast --subscription $SUBSCRIPTION_ID

az group create --name $APP_RESOURCE_GROUP_NAME --location australiaeast --subscription $SUBSCRIPTION_ID

az appservice plan create --name $APP_SERVICE_PLAN_NAME --resource-group $PLAN_RESOURCE_GROUP_NAME --sku B1 --is-linux

$WEBAPP_NAME = [System.Guid]::NewGuid().ToString()

az webapp create --name $WEBAPP_NAME --resource-group $APP_RESOURCE_GROUP_NAME --plan $APP_SERVICE_PLAN_NAME --runtime """python|3.6""" --query id
```

## Environment Summary
```
Windows-10-10.0.18362-SP0
Python 3.6.6
Shell: cmd.exe

azure-cli 2.0.80 *

Extensions:
db-up 0.1.13
dev-spaces 1.0.3
interactive 0.4.2

```
## Additional Context
Related discussion on the azure-xplat-cli repo: https://github.com/Azure/azure-xplat-cli/issues/3485

yungezz:
thanks for reporting the issue. we're looking at it.

panchagnula:
Can you please run the command with --debug and check the details. Thanks!

danielbecroft:
Hi @panchagnula 

Here is the output:

```
DEBUG: Command arguments: ['webapp', 'create', '--name', '486097f6-81bc-4702-a983-f64d5aacce22', '--resource-group', 'zz-webapp-group', '--plan', 'zz-appserviceplan', '--runtime', 'python|3.6', '--debug']
DEBUG: Event: Cli.PreExecute []
[...]

PR #13068, opened 2020-04-17
[Web App] Web app and plan different rgs

Guilhermeslucas:
**Description**  

This PR adds the possibility of creating a web app in a resource group that is different from the plan you are trying to create

**Testing Guide**  

In order to test this PR, you should follow the instructions on this [issue ](https://github.com/Azure/azure-cli/issues/12090 )that explain how to reproduce it. You should run:
```
$APP_SERVICE_PLAN_NAME = "appserviceplan"
$PLAN_RESOURCE_GROUP_NAME = "appservice-plan-group"
$APP_RESOURCE_GROUP_NAME = "webapp-group"

az group create --name $PLAN_RESOURCE_GROUP_NAME --location australiaeast --subscription $SUBSCRIPTION_ID

az group create --name $APP_RESOURCE_GROUP_NAME --location australiaeast --subscription $SUBSCRIPTION_ID

az appservice plan create --name $APP_SERVICE_PLAN_NAME --resource-group $PLAN_RESOURCE_GROUP_NAME --sku B1 --is-linux

$WEBAPP_NAME = [System.Guid]::NewGuid().ToString()

az webapp create --name $WEBAPP_NAME --resource-group $APP_RESOURCE_GROUP_NAME --plan $APP_SERVICE_PLAN_NAME --runtime """python|3.6""" --query id
```

**History Notes**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -80,7 +80,7 @@ def create_webapp(cmd, resource_group_name, name, plan, runtime=None, startup_fi
         raise CLIError('usage error: --deployment-source-url <url> | --deployment-local-git')
 
     docker_registry_server_url = parse_docker_image_name(deployment_container_image_name)
-    
+  
     client = web_client_factory(cmd.cli_ctx)
     if is_valid_resource_id(plan):
         parse_result = parse_resource_id(plan)
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -86,7 +86,7 @@ def create_webapp(cmd, resource_group_name, name, plan, runtime=None, startup_fi
         parse_result = parse_resource_id(plan)
         plan_info = client.app_service_plans.get(parse_result['resource_group'], parse_result['name'])
     else:
-        plan_info = client.app_service_plans.get(resource_group_name, plan)
+        plan_info = _get_plan_by_name(client, plan)
     if not plan_info:
         raise CLIError("The plan '{}' doesn't exist".format(plan))
     is_linux = plan_info.reserved
@@ -3529,3 +3529,13 @@ def _verify_hostname_binding(cmd, resource_group_name, name, hostname, slot=None
             verified_hostname_found = True
 
     return verified_hostname_found
+
+def _get_plan_by_name(client, plan):
+    plan_info = None
+    plan_list = client.app_service_plans.list()
+    if plan_list:
+        candidate_plans = list(filter(lambda x: x.name == plan, plan_list))
+        if candidate_plans:
+            plan_info = candidate_plans[0]
+
+    return plan_info
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -3532,14 +3532,3 @@ def _verify_hostname_binding(cmd, resource_group_name, name, hostname, slot=None
             verified_hostname_found = True
 
     return verified_hostname_found
-
-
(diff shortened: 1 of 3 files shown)


Guilhermeslucas:
Fixes #12090

yonzhan:
add to S169

panchagnula:
@Kotasudhakarreddy can you PR this change please?

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -3529,3 +3529,14 @@ def _verify_hostname_binding(cmd, resource_group_name, name, hostname, slot=None
...
     return verified_hostname_found
+
+
+def _get_plan_by_name(client, plan):
+    plan_info = None
+    plan_list = client.app_service_plans.list()
+    if plan_list:
+        candidate_plans = list(filter(lambda x: x.name == plan, plan_list))
+        if candidate_plans:
+            plan_info = candidate_plans[0]
```
qwordy, line 3540: Why choose the first one?

Guilhermeslucas: If we are choosing based on the name, it shouldn't really matter if this is the first, second, etc! If the name matches, should be enough.

qwordy: Is it possible that multiple plans have same names but different configurations?

Guilhermeslucas: Yes! And that can happen because we are not using the resource group that the web app is located anymore!

Kotasudhakarreddy: It is possible that multiple plans can have same names under different resource groups. So when we consider only the first plan, then there is a possibility of assigning a plan from different RG than the user anticipated RG. So as per me the current implementation will not solve the problem 100%.
**proposed implementation:**  In the portal plans are populated based on the selected `AppService OS` type. I think we should consider the same implementation for CLI command as well.

Guilhermeslucas: I see! So, whats the way you suggest implementing that, since we have no mandatory parameter that tells the OS type on the az webapp create command?

Kotasudhakarreddy: yes we don't have a parameter to tell the type of OS. we should either add a parameter as `--plan-resource-group`  to know the plan RG name or `--is_linux` to know the OS type.
@panchagnula - please let us know your suggestion on it.

Guilhermeslucas: I am happy with both of the implementations! I feel that in order to avoid any type of confusing, we should add the ```--plan-resource-group``` as a optional parameter, and if its not passed, we try to get from the same rg as web app is! What do you think?

Guilhermeslucas: Added the implementation for --plan-resource-group, let me know what you think!

panchagnula: @Guilhermeslucas first of all thanks for taking the initiative in starting this PR. Unfortunately at this point we cannot take your change. One big feedback we get regarding CLI commands is to keep them simple & not add new parameters since for a non Azure domain knowledge customer - the commands are overwhelming, hence in order to fix the original issue here, our team needed to come up with a better solution to fix this ,avoiding adding any additional parameters. I am closing this PR for now since we cannot take this  with the current approach. I see you are an MS employee please feel free to contact me via email if you need more details. Thank you.

Guilhermeslucas: huuum, thats sad :( No problem, I understand! Thanks for the answer!

Issue #11867, opened 2020-01-15
az webapp config ssl import assumes the key vault is in the resource group of the web app


collinstevens:
**Describe the bug**
`az webapp config ssl import` assumes the key vault is in the resource group of the web app.
An error is output describing that the key vault was unable to be found within the resource group of the web app.

**To Reproduce**
1. Create a key vault named kv-key-vault in resource group rg-key-vaults
2. Upload a certificate to the certificate store of the key vault kv-key-vault named cert-certificate
2. Create a web app named app-web-app in resource group rg-web-apps
3. Execute `az webapp config ssl import --key-vault kv-key-vault --key-vault-certificate-name cert-certificate -n app-web-app -g rg-web-apps`
4. Receive error `The Resource 'Microsoft.KeyVault/vaults/kv-key-vault' under resource group 'rg-web-app' was not found.`

**Expected behavior**
The certificate should be imported to the web app.

**Environment summary**
- Azure CLI version 2.0.80
- Powershell Core
- Windows 10 x64

**Additional context**
I would prefer a required argument of --key-vault-resource-group-name or something similar because I reference a common key vault from many web apps and function apps. This key vault is not in the same resource groups of these web apps and function apps.

tsgenge:
We have the same issue; when the key vault is in a different RG (very common, as you want your compute to be torn down easily without impacting config elements) from your web app, this command fails. Setting the RG with -g to the web app RG causes an error that it can't find the keyvault, and vice versa. Note setting "az configure --defaults group" doesn't help.

derekmckinnon:
Just hit the same problem as well, with LetsEncrypt certs in Key Vault in a different RG then the Web Apps.

It would be great to have `--key-vault-resource-group` option in order to get around this issue.

jvanderbiest:
Same here... I have a key vault that is used cross resource group because it contains the 1 wildcard certificate that I want to bind a number of apps in different resource groups :-(.

jsmed-renson:
Same problem here, looking for that `--key-vault-resource-group` parameter ;)

gavinbarron:
I think the correct fix is not to require the resource group that the keyvault is in as a parameter at all.

I guess I'll just keep on downloading and uploading the SSL cert in PowerShell scripts until this actually works

The docs on this should clearly callout the that the current implementation require the webapp and the keyvault to be in the same resource group.

panchagnula:
@madsd FYI.

madsd:
[...]

PR #13099, opened 2020-04-21
[AppService] az webapp|functionapp config ssl import: Lookup key vault across resources groups in subscription and improve help and examples.

madsd:
**Description**  
Prior to this change, issues were raised #11867 as the command were failing if the key vault was not in the same resource group as the web app. An option existed to provide a resource id, but this was not documented enough to be discoverable. The change allows key vault to be imported across all resource groups in the subscription in context without any change, and if it is not found, a suggestion to provide the full resource id is displayed. Further a help example of using resource id is added.
There are no changes to the parameter list or breaking changes.

**Testing Guide**  
The two examples in az webapp config ssl import --help

**History Notes**  

[AppService] az webapp|functionapp config ssl import: Lookup key vault across resources groups in subscription and improve help and examples.

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -2136,7 +2136,26 @@ def import_ssl_cert(cmd, resource_group_name, name, key_vault, key_vault_certifi
         raise CLIError("'{}' app doesn't exist in resource group {}".format(name, resource_group_name))
     server_farm_id = webapp.server_farm_id
     location = webapp.location
-    kv_id = _format_key_vault_id(cmd.cli_ctx, key_vault, resource_group_name)
+    kv_id = None
+    if not is_valid_resource_id(key_vault):
+        kv_client = get_mgmt_service_client(cmd.cli_ctx, ResourceType.MGMT_KEYVAULT)
+        key_vaults = kv_client.vaults.list_by_subscription()
+        for kv in key_vaults:
+            if key_vault == kv.name:
+                kv_id = kv.id
+                break
+    else:
+        kv_id = key_vault
+
+    if kv_id is None:
+        kv_msg = 'The Key Vault {0} was not found in the subscription in context. ' \
+                 'If your Key Vault is in a different subscription, please specify the full Resource ID: ' \
+                 '\naz .. ssl import -n {1} -g {2} --key-vault-certificate-name {3} ' \
+                 '--key-vault /subscriptions/[sub id]/resourceGroups/[rg]/providers/Microsoft.KeyVault/' \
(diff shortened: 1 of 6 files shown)


yonzhan:
add to S169

madsd:
@yonzhan @Juliehzl I can see it is in failed state due to the PR Title - I changed that, but it does not restart the build - is there a way to restart the build or do I need to push a small change?

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2136,7 +2136,26 @@ def import_ssl_cert(cmd, resource_group_name, name, key_vault, key_vault_certifi
         raise CLIError("'{}' app doesn't exist in resource group {}".format(name, resource_group_name))
     server_farm_id = webapp.server_farm_id
     location = webapp.location
-    kv_id = _format_key_vault_id(cmd.cli_ctx, key_vault, resource_group_name)
+    kv_id = None
+    if not is_valid_resource_id(key_vault):
+        kv_client = get_mgmt_service_client(cmd.cli_ctx, ResourceType.MGMT_KEYVAULT)
+        key_vaults = kv_client.vaults.list_by_subscription()
```
qwordy, line 2142: Maybe you can use a get method to avoid traversing all key vaults.

madsd: I looked at that, but get method requires a resource group to be specified, and the point is to search across resource groups.

madsd:
@panchagnula fyi

madsd:
@Juliehzl - are you waiting for something from me?

qwordy:
@Juliehzl Please review.

madsd:
> Do you have a test for key vault resource id?

@Juliehzl , the code does not change the functionality for import with resource id - but I am happy to add a test for that. I was looking to add a test with multiple resource groups. Do you know if ResourceGroupPreparer can handle that?

Juliehzl:
@madsd You can add multiple ResourceGroupPreparer() to create multiple resource groups. Here is a sample for StorageAccountPreparer(), which is the same. https://github.com/Azure/azure-cli/blob/dev/doc/authoring_tests.md#sample-8-prepare-multiple-storage-accounts-for-tests

madsd:
@Juliehzl thanks, I added two additional tests - one for kv by name in different resource group and one for kv by resource id.

Juliehzl:
@madsd Can you take a look at CI test? There are something wrong in your test.

madsd:
> @madsd Can you take a look at CI test? There are something wrong in your test.

It is strange, because that part I did not change - it is the import of the cert to Key Vault. I will do a bit of debugging to see if I can find the error.

madsd:
@Juliehzl Finally the tests passed. Still don't know what caused it, but when I combined the cross RG with Resource Id, it worked.

Issue #11950, opened 2020-01-26
az keyvault set-policy: passing a variable to --object-id fails.


ghstahl:
## Describe the bug

**Command Name**
`az keyvault set-policy`

**Errors:**
```
An invalid value was provided for 'accessPolicies'.
```

## To Reproduce:
Steps to reproduce the behavior. Note that argument values have been redacted, as they may contain sensitive information.

```
AZ_FUNC_NAME="azfun-oauth2"
APP_FRIENDLY_NAME="organics"
LOCATION="eastus2"
RESOURCE_GROUP_NAME="rg-$APP_FRIENDLY_NAME-openhack"
KEYVAULT_NAME="kv-$APP_FRIENDLY_NAME"



az webapp identity assign \
    --name $AZ_FUNC_NAME  \
    --resource-group $RESOURCE_GROUP_NAME
PRINCIPAL_ID=$(az webapp show -n $AZ_FUNC_NAME -g $RESOURCE_GROUP_NAME --query 'identity.principalId' -o json)
echo "KEYVAULT_NAME:$KEYVAULT_NAME"
echo "principalId:$PRINCIPAL_ID"



az keyvault set-policy \
    --name $KEYVAULT_NAME \
    --object-id $PRINCIPAL_ID \
    --secret-permissions get list \
    --key-permissions get list sign verify create \
    --certificate-permissions get list getissuers listissuers
```
## The Problem
```
--object-id $PRINCIPAL_ID \
```
## This Works
```
--object-id "02c27055-efb4-4b9d-bd5d-acd671ea5a10" \
```

I query for the **identity.principalId** and assign it to **PRINCIPAL_ID**.
Passing **$PRINCIPAL_ID** as the argument to **--object-id** FAILS.
Hardcoding it works.  

## Expected Behavior
Passing **$PRINCIPAL_ID** as the argument to **--object-id** SHOULD SUCCEED.

## Environment Summary
```
Linux-4.4.0-18362-Microsoft-x86_64-with-debian-buster-sid
Python 3.6.5
Shell: bash

azure-cli 2.0.78 *

Extensions:
azure-devops 0.16.0

```

fengzhou-msft:
@bim-msft please take a look.

bim-msft:
@ghstahl Hi, could you please upgrade your cli client to the latest (2.4.0) and try it again? Thanks.
I cannot reproduce your issue, this works for me:

[...]

PR #13146, opened 2020-04-24
{KeyVault} Support quoted `object id`, `spn`, `upn` for `set-policy` and `delete-policy`

bim-msft:
Support for #11950 

The case is when user try to obtain the object id using the following command (-o json):
```
PRINCIPAL_ID=$(az webapp show -n $AZ_FUNC_NAME -g $RESOURCE_GROUP_NAME --query 'identity.principalId' -o json)
echo "principalId: $PRINCIPAL_ID"
```

The ID will be quoted like this:
```
principalId: "02c27055-efb4-4b9d-bd5d-acd671ea5a10"
```

This quoted ID cannot be passed into the command:
```
az keyvault set-policy -n {kv} --object-id $PRINCIPAL_ID --key-permissions get list
An invalid value was provided for 'accessPolicies'.
```

---


src/azure-cli/azure/cli/command_modules/keyvault/_validators.py
```diff
@@ -189,6 +189,11 @@ def validate_principal(ns):
         raise argparse.ArgumentError(
             None, 'specify exactly one: --object-id, --spn, --upn')
 
+    attrs = ['object_id', 'spn', 'upn']
+    for attr in attrs:
+        if getattr(ns, attr):
+            setattr(ns, attr, getattr(ns, attr).strip('"'))
```
yungezz, line 195: for upn, is it safe to strip `"` ? what's the name convention? suggest only change object_id for now.

bim-msft: > for upn, is it safe to strip `"` ? what's the name convention? suggest only change object_id for now.

Most of UPN names are email-like: https://docs.microsoft.com/en-us/azure/active-directory/hybrid/plan-connect-userprincipalname
As far as I know there is also a type of UPN for external users like this: `bim_microsoft.com#EXT#@AzureSDKTeam.onmicrosoft.com`

Both formats don't have leading or tailing quotes.
But in case that there are maybe some unknown formats may have quotes, I will keep the original format for UPN and SPN. Will investigate them again when we have new user ask.

yonzhan:
Key Vault

fengzhou-msft:
While the change LGTM, this is a very common usage in other commands as well, we may need to further consider fixing it in the core.

bim-msft:
> While the change LGTM, this is a very common usage in other commands as well, we may need to further consider fixing it in the core. Probably remove quotes in `--query` result if it has no spaces.

Graceful solution. As this is not urgent, I will change the implementation in the same PR later.

fengzhou-msft:
Using `-o tsv` instead of `-o json` would strip the double quotes. In the case to use the output of a CLI command, we should use `-o tsv` and I don't think we need to add support for `-o json` in this case as well.

PR #13158, opened 2020-04-24
[ACR] `az acr check-health`: Fix "DOCKER_PULL_ERROR" on Windows

Wwwsylvia:
**Description**  

Fix [az acr check-health : Error image operating system "linux" cannot be used on this platform](https://github.com/Azure/acr/issues/328)
Fix the bug that `az acr check-health` always throws "DOCKER_PULL_ERROR" on Windows 

**Before:**
image

**After:**
image

**Testing Guide**  

`az acr check-health --name myregistry --ignore-errors --yes`

**History Notes**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acr/check_health.py
+++ b/src/azure-cli/azure/cli/command_modules/acr/check_health.py
@@ -31,9 +31,7 @@ DOCKER_PULL_WRONG_PLATFORM = 'cannot be used on this platform'
 
 # Utilities functions
 def print_pass(message):
-    from colorama import Fore, Style, init
-    init()
-    logger.warning(str(message) + " : " + Fore.GREEN + "OK" + Style.RESET_ALL)
+    logger.warning("%s : OK", str(message))
 
 
 def _handle_error(error, ignore_errors):
@@ -112,7 +110,7 @@ def _get_docker_status_and_version(ignore_errors, yes):
         if stderr:
             if DOCKER_PULL_WRONG_PLATFORM in stderr:
                 print_pass("Docker pull of '{}'".format(IMAGE))
-                logger.warning("Image %s can be pulled but cannot be used on this platform", IMAGE)
+                logger.warning("Image '%s' can be pulled but cannot be used on this platform", IMAGE)
             else:
                 _handle_error(DOCKER_PULL_ERROR.append_error_message(stderr), ignore_errors)
         else:
--- a/src/azure-cli/azure/cli/command_modules/acr/check_health.py
+++ b/src/azure-cli/azure/cli/command_modules/acr/check_health.py
@@ -6,7 +6,6 @@
 from __future__ import print_function
 
 import re
-import sys
 from knack.util import CLIError
 from knack.log import get_logger
 from .custom import get_docker_command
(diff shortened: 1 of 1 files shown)


Wwwsylvia:
@yugangw-msft Could you help to review?

yonzhan:
add to S169

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -108,7 +111,11 @@ def _get_docker_status_and_version(ignore_errors, yes):
         output, warning, stderr = _subprocess_communicate([docker_command, "pull", IMAGE])
 
         if stderr:
-            _handle_error(DOCKER_PULL_ERROR.append_error_message(stderr), ignore_errors)
+            if DOCKER_PULL_WRONG_PLATFORM in stderr:
+                print_pass("Docker pull of '{}'".format(IMAGE))
```
yungezz, line 115: in cli we don't use print but logger, just checked this print_pass is used several places in this file. what scenario it's used for besides normal logger methods, could you pls change to logger.xxx?

fengzhou-msft: For `print_pass`, it's redirected to std_err. I think it should be fine.
https://github.com/Azure/azure-cli/blob/f5cbbb28b388ad1dd114ffaedea26446fe5acc8c/src/azure-cli/azure/cli/command_modules/acr/check_health.py#L37

Other information like versions is also printed in stderr, how do users use the result of `acr check-health` command in a script?

qwordy:
How about output error without modification?
This line is not very robust. Usually, we should rely on error code rather than string matching.
```
if DOCKER_PULL_WRONG_PLATFORM in stderr:
```

src/azure-cli/azure/cli/command_modules/acr/check_health.py
Wwwsylvia: Thanks, I've replaced print() with logger.warning().

> Other information like versions is also printed in stderr, how do users use the result of `acr check-health` command in a script?

`az acr check-health` is for troubleshooting purpose, usually users would not need the output of this command in script. A exit code is enough to check the status.

Wwwsylvia:
> How about output error without modification?
> This line is not very robust. Usually, we should rely on error code rather than string matching.
> 
> ```
> if DOCKER_PULL_WRONG_PLATFORM in stderr:
> ```

We knew string matching is definitely not a good practice, but unfortunately it seems to be the only way to implement this, since currently Docker CLI does not return different exit codes to distinguish different errors types.

I found this [[Proposal] docker pull return code](https://github.com/docker/cli/issues/538) issue in their repo is still open...

Wwwsylvia:
This PR is ready for review

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -27,13 +26,14 @@
...
 NOTARY_VERSION_REGEX = re.compile(r'Version:\s+([.\d]+)')
+DOCKER_PULL_WRONG_PLATFORM = 'cannot be used on this platform'
 
 
 # Utilities functions
 def print_pass(message):
     from colorama import Fore, Style, init
     init()
-    print(str(message) + " : " + Fore.GREEN + "OK" + Style.RESET_ALL, file=sys.stderr)
+    logger.warning(str(message) + " : " + Fore.GREEN + "OK" + Style.RESET_ALL)
```
yungezz, line 36: pls be aware of that warning will display to stderr. Some customers who're doing automation don't want warning in stderr to break auto pipeline

yungezz: and we don't suggest color in output, except it's from CLI underlying framework.

Wwwsylvia: Removed

Wwwsylvia:
Can anyone help to merge if no more concerns?

PR #13230, opened 2020-04-28
{Packaging} Remove wheel as a runtime dependency

fengzhou-msft:
**Description**  

wheel should not be needed to run Azure CLI. It is only used in packaging.

To remove wheel from azure-cli-core dependencies, we need to install wheel in packaging scripts. ~~It seems they're already doing this, will check them one by one.~~ Verified that MSI, homebrew and pypi use setup.py to build wheels during packaging, and their script all include a step to install wheel. Other packages use requirements file to install dependency, wheel is not used.

For the packaging of extensions, as azdev has the dependency of wheel, it should be fine.

**Testing Guide**  

**History Notes**  

---

### DIFF
--- a/src/azure-cli-core/setup.py
+++ b/src/azure-cli-core/setup.py
@@ -62,7 +62,6 @@ DEPENDENCIES = [
     'msrest>=0.4.4',
     'msrestazure>=0.6.3',
     'paramiko>=2.0.8,<3.0.0',
-    'pkginfo',
     'PyJWT',
     'pyopenssl>=17.1.0',  # https://github.com/pyca/pyopenssl/pull/612
     'requests~=2.20',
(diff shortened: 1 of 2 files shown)


yonzhan:
remove wheel

haroldrandom:
Fix https://github.com/Azure/azure-cli/issues/12582

src/azure-cli-core/setup.py
```diff
@@ -62,12 +62,10 @@
     'msrest>=0.4.4',
     'msrestazure>=0.6.3',
     'paramiko>=2.0.8,<3.0.0',
-    'pkginfo',
```
fengzhou-msft, line 65: remove duplicates

PR #13253, opened 2020-04-29
{Packaging} Clean some azure-cli dependencies

fengzhou-msft:
**Description**  

1. Remove the dependencies in azure-cli that is already in azure-cli-core dependencies.
2. mock is only needed in testing, move it to the tests_require dependency of `azure-cli`. `azure-cli-core` has done the same:
https://github.com/Azure/azure-cli/blob/1d64349de51139471011c49febb1ac5a4de5d2c3/src/azure-cli-core/setup.py#L74-L76

Since azdev has the dependency of mock, nothing needs to be changed for setting up testing environment with azdev.
https://github.com/Azure/azure-cli-dev-tools/blob/a25a8c9e5dfb416215711f4240623bfd4db13049/setup.py#L68

**Testing Guide**  

**History Notes**  

---


yonzhan:
add to S169

src/azure-cli/setup.py
```diff
@@ -148,6 +139,10 @@
     'jsondiff==1.2.0'
 ]
 
+TESTS_REQUIRE = [
```
haroldrandom, line 142: I think, we don't need this. 

Because `mock` is [merged into Python standard library since Python 3.3](mock is now part of the Python standard library, available as unittest.mock in Python 3.3 onwards), and CLI support Python 3.6+ only.

So, I think we should also change code from `import mock` to `from unittest import mock`

fengzhou-msft: Just tried replacing `import mock` with `from unittest import mock`, some tests failed. I believe some code is using the latest feature of mock 4.0, so mock bundled in python 3.6 might not work. @jiasli to comment.

Issue #13137, opened 2020-04-23
 UnicodeEncodeError when passing JSON with Japanese Characters into "az deployment group create" 


hiyosh-msft:
**Describe the bug**
When you input a JSON with Japanese characters into the following command, you get the following error.

Command:
`az deployment group create -g <resource-group> --template-file <arm-template.json> `

Error:
` latin-1' codec can't encode characters in position 363-375: Body ('ここに日本語で説明します。') is not valid Latin-1.`

`az group create --location <location> --name <resource-group> --tags "test=日本語"`
seems to work successfully. Therefore, I'm pretty sure that Azure CLI itself is capable of posting JSON with Japanese, and that `az deployment group create` failing with Japanese characters is most likely unintentional.

**To Reproduce**
Pass a JSON with Japanese characters in it to `az deployment group create -g <resource-group> --template-file <arm-template.json>`.

A sample ARM-template (sample.txt), and the full trace from --debug option (result.txt) is attached.
[result.txt](https://github.com/Azure/azure-cli/files/4523079/result.txt)
[sample.txt](https://github.com/Azure/azure-cli/files/4523080/sample.txt)

**Expected behavior**
`az deployment group create` should be able to post JSON with Japanese characters.

**Environment summary**
`NAME="Ubuntu"
VERSION="18.04.4 LTS (Bionic Beaver)"
Python 3.6.9
 az version = 
  "azure-cli": "2.4.0",
  "azure-cli-command-modules-nspkg": "2.0.3",
  "azure-cli-core": "2.4.0",
  "azure-cli-nspkg": "3.0.4",
  "azure-cli-telemetry": "1.0.4",
  "extensions": {}`

But it repro with Windows as well.

yonzhan:
add to S170

bureado:
FYI, if this is related to another customer inquiry I saw on Twitter, it could be related to certain libraries or the version of Python, as it seems to work in Cloud Shell but not in DevOps Pipeline: https://twitter.com/jabex9/status/1253611479929991169

ns-abe-jun:
It worked fine with CloudShell until the other day, but now it doesn't.
Until the other day, it was Az CLI ver 2.2.0, but now it is ver 2.4.0.
Probably it is related Az CLI ver 2.4.0.

Incidentally , The same phenomenon occurs in DevOps Pipeline (Ubuntu 16.04 / 18.04), but this is ver 2.3.1.

zhoxing-ms:
[...]

PR #13269, opened 2020-04-30
[ARM] az deployment: Fix the problem that deployment/validate does not support specially encoded character (#13137)

zhoxing-ms:
Issue: #13137 #13316

**Description**  
The reason for this problem is that the pipeline configuration of the requesting client was modified during the processing of `handle_extended_json_format`, resulting in incorrect request encoding:

> 'latin-1' codec can't encode characters in position 376-388: Body ('ここに日本語で説明します。') is not valid Latin-1. Use body.encode('utf-8') if you want to send it encoded in UTF-8.

As prompted by the error message, change the encoding of the request to `utf-8`.

**Testing Guide**  
Use ARM templates with special encoding characters which not supported by `latin-1` for deployment or validation.

**History Notes**  

---


yonzhan:
add to S169

PR #13274, opened 2020-04-30
{Identity} az logout

jiasli:
## Description

Support `az login`. 

This PR changes the behavior of `az logout` and `az account clear` so that they only clear the profiles and legacy cache. 

To clear shared token from MSAL encrypted cache, user will need to specify `--clear-credential`. 

## MSAL cache handling

Regardless of whether the user is logged in, CLI will always check the MSAL cache for the specified username or all users, and show warnings accordingly. 

It first checks whether the cache exists (for the username):

- ✔ Yes, check if the command wants to clear the cache
  - ✔ Yes, clear the cache and show a warning that the user is also logged out from other SDKs
  - ❌ No, show a warning that the credential is still stored in MSAL encrypted cache (ℹ The word **encrypted** is very important and can't be omitted.)
- ❌ No, show a warning that the cache doesn't exist

In the future we need to create some extra commands to manage MSAL cache directly and bootstrap CLI from MSAL cache.

### DIFF
--- a/src/azure-cli-core/azure/cli/core/_profile.py
+++ b/src/azure-cli-core/azure/cli/core/_profile.py
@@ -417,46 +417,68 @@ class Profile(object):
         set_cloud_subscription(self.cli_ctx, active_cloud.name, result[0][_SUBSCRIPTION_ID])
         self._storage[_SUBSCRIPTIONS] = subscriptions
 
-    def logout(self, user_or_sp):
+    def logout(self, user_or_sp, clear_credential):
         subscriptions = self.load_cached_subscriptions(all_clouds=True)
         result = [x for x in subscriptions
                   if user_or_sp.lower() == x[_USER_ENTITY][_USER_NAME].lower()]
-        if not result:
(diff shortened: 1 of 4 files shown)


yonzhan:
add to S169

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -126,8 +127,8 @@ def login(self,
 
         credential=None
         auth_profile=None
-        authority = self.cli_ctx.cloud.endpoints.active_directory.replace('https://', '')
-        identity = Identity(authority, tenant, cred_cache=self._adal_cache)
+        adal_cache = ADALCredentialCache(cli_ctx=self.cli_ctx)
```
qianwens, line 130: >adal_cache = ADALCredentialCache(cli_ctx=self.cli_ctx) [](start = 8, length = 54)

not used?

jiasli: No idea how this line get inserted. Removed.

PR #13286, opened 2020-05-01
[Extension] support .egg-info to store wheel type extension metadata

haroldrandom:
**Description**  
As this PR https://github.com/Azure/azure-cli/pull/12367 pointing out, some user may manually install extension from source code and still want to load it.

Same scenario could happen in the developing of azure-cli-ml as discussed offline. The develop CLI extension without azdev, and use that way to work in editable mode in older CLI (<= 2.4.0)

**Testing Guide**  

**History Notes**  

---


yonzhan:
add to S170

src/azure-cli-core/azure/cli/core/extension/tests/latest/test_dev_type_extension.py
```diff
@@ -0,0 +1,44 @@
...
+
+        # assert Python metadata
+        self.assertEqual(metadata['name'], ext_name)
+        self.assertEqual(metadata['version'], ext_version)
+        self.assertEqual(metadata['author'], 'Microsoft Corporation')
+        self.assertNotIn('metadata.json', os.listdir(os.path.join(self.ext_dir, ext_name + '-' + ext_version)))
+
+        # assert Azure CLI extended metadata
+        self.assertTrue(metadata['azext.isPreview'])
+        self.assertTrue(metadata['azext.isExperimental'])
```
fengzhou-msft, line 43: I remember these two tags are not supposed to be both True in the future. @zhoxing-ms for confirmation.

src/azure-cli-core/azure/cli/core/extension/tests/latest/test_wheel_type_extension.py
```diff
@@ -0,0 +1,107 @@
...
+        """
+
+        source_code_packaged = get_test_data_file('hello-0.1.0.tar.gz')
+
+        with tarfile.open(source_code_packaged, 'r:gz') as tar:
+            tar.extractall(self.ext_dir)
+
+        ext_name, ext_version = 'hello', '0.1.0'
+
+        ext_extension = WheelExtension(ext_name, os.path.join(self.ext_dir, ext_name + '-' + ext_version))
```
fengzhou-msft, line 95: not use DevExtension?

src/azure-cli-core/azure/cli/core/extension/tests/latest/test_dev_type_extension.py
zhoxing-ms: Yes, according to the previous design of @jiasli for `isExperimental`, an error will be reported if both `isPreview` and `isExperimental` are True.

src/azure-cli-core/azure/cli/core/extension/tests/latest/test_wheel_type_extension.py
haroldrandom: This is to test WheelExtension class not DevExtension.
The user using this way to develop Azure CLI extension is to hack the machenism. I think it's doable but not recommended. And supporting .egg-info is just one month ago.

src/azure-cli-core/azure/cli/core/extension/tests/latest/test_dev_type_extension.py
haroldrandom: Thanks for sharing this, I just learnt that.
I think the class `WheelExtension` and `DevExtension` is to represent data model not validation. Validation should be a higher level business.
This PR is about it test reading the metadata correctly, not validation.

Please @jiasli make sure that the concern above has been made in CLI.

Issue #12137, opened 2020-02-11
az ad app grants not working or usable as expected


dmprantz:
**Describe the bug**
"az ad app permission grant" only seems to grant a single scope.  If I call it successive times, the existing scope is overwritten.  This behaviour is not clearly documented, nor is the way to grant. multiple scopes.

"az ad app permission admin-consent" appears to be non-functional.  How can admin consent be granted without three key pieces of information, AppID, APIID, and Scope?  What is this doing?  How can a single scope be granted at the application level?

I would also consider it a bug that "az ad app permission grant" cannot be used to grant application type scopes.

**To Reproduce**
Try to execute the above commands.  "az ad app permission admin-consent" is not documented in any way that makes sense.  "az ad app permission grant" seems to work, but successive calls over-write your work.

**Expected behavior**
There should be a single command that allows you to grant individual (or comma separated lists) of scopes for an API to an app.  That same command should have a flag for Delegated vs Application Type.  It should not overwrite itself.  Worst case, if there have to be two commands, the one used to grant admin-consent to the application type should allow appropriate parameters.

Documentation of these calls should be clear and concise.

**Environment summary**
Windows 10.  Installed via downloaded MSI for x64.  cmd.exe shell.

C:\Program Files\Microsoft SDKs\Azure\.NET SDK\v2.9\bin>az --version
azure-cli                         2.0.81

command-modules-nspkg              2.0.3
core                              2.0.81
nspkg                              3.0.4
telemetry                          1.0.4

Python location 'C:\Program Files (x86)\Microsoft SDKs\Azure\CLI2\python.exe'
Extensions directory 'C:\Users\dpomerantz\.azure\cliextensions'

Python (Windows) 3.6.6 (v3.6.6:4cf1f54eb7, Jun 27 2018, 02:47:15) [MSC v.1900 32 bit (Intel)]

arrownj:
Hi @jiasli , could you please help take a look at this ?

yonzhan:
add to S166.

dmprantz:
This was pushed to a future release.  How many times is it going to get pushed before being addressed and resolved.  Is any one able to comment on a commitment that this will be resolved with a definitive timeline?

jiasli:
[...]

PR #13325, opened 2020-05-04
[RBAC] Updates to az ad app permission list/add/delete

psignoret:
**Description**

Various updates to `az ad app permission` commands for managing required API permissions on app registrations:

* `az ad app permission add`
  * Remove message suggesting that running `az ad app permission grant --id {id} --api {api}` will grant the permissions added. (This was inaccurate and there's no command currently that can be used to grant both application and delegated permissions.)
  * Add support for identifying target API by identifier URI and service principal object ID.
* `az ad app permission delete`
  * Add support for removing an individual required permission.
  * Add support for identifying target API by identifier URI and service principal object ID.
* `az ad app permission list`
  * Set delegated permission grant expiry time to a date far in the future. This is slightly more accurate than the current approach, since the real "expiryTime" values are ignored by the service.
  * Add option to skip retrieving delegated permission grants (avoids several unnecessary queries and improves performance).
* Other:
  * Refactor `show_service_principal` to avoid an unnecessary query in several scenarios.
  * Expanded tests to cover more scenarios and to tear down more cleanly.

**Testing Guide**  

See examples added in _help.py.

**History Notes**  

[RBAC] az ad app permission list: Now includes a date far in the future as "expiryTime" when listing required permissions and there exists at least one tenant-wide delegated permission grant for the API in question. (The real "expiryTime" values are not enforced by the server, so it is more accurate to state the the permission grant expires far in the future.) Use `az ad app permission list-grants` to list all delegated permission grants.

[RBAC] az ad app permission add: Add support for identifying the API by identifier URI.

[RBAC] az ad app permission delete: Add support for identifying the API by identifier URI, and add support for removing individual required permissions.

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/role/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/role/_help.py
@@ -146,16 +146,23 @@ short-summary: remove an application owner.
 
 helps['ad app permission'] = """
 type: group
-short-summary: manage an application's OAuth2 permissions.
+short-summary: Manage an app registration's required API permissions (requiredResourceAccess).
 """
 
 helps['ad app permission add'] = """
 type: command
-short-summary: add an API permission
-long-summary: invoking "az ad app permission grant" is needed to activate it
+short-summary: Add required API permissions.
+long-summary: Adds required API permissions to the app registration. Does not grant permissions.
 examples:
-  - name: add a Graph API permission of "Sign in and read user profile"
-    text: az ad app permission add --id eeba0b46-78e5-4a1a-a1aa-cafe6c123456 --api 00000002-0000-0000-c000-000000000000 --api-permissions 311a71cc-e848-46a1-bdf8-97ff7156d8e6=Scope
+  - name: Add the delegated permission User.Read for Microsoft Graph to the required permissions.
+    text: az ad app permission add --id eeba0b46-78e5-4a1a-a1aa-cafe6c123456 --api "00000002-0000-0000-c000-000000000000" --api-permissions e1fe6dd8-ba31-4d61-89e7-88639da4683d=Scope
+  - name: Retrieve the permission IDs for the delegated permission User.Read and the app role Application.Read.All, both for Microsoft Graph, then add these as required permissions.
+    text: |
+      az ad sp show --id "https://graph.microsoft.com" --query "oauth2Permissions[?value == 'User.Read'].[id, value]" -o tsv
+      e1fe6dd8-ba31-4d61-89e7-88639da4683d    User.Read
+      az ad sp show --id "https://graph.microsoft.com" --query "appRoles[?value == 'Application.Read.All'].[id, value]" -o tsv
+      9a5d68dd-52b0-4cc2-bd40-abcf44ac3a30    Application.Read.All
+      az ad app permission add --id eeba0b46-78e5-4a1a-a1aa-cafe6c123456 --api "https://graph.microsoft.com" --api-permissions e1fe6dd8-ba31-4d61-89e7-88639da4683d=Scope 9a5d68dd-52b0-4cc2-bd40-abcf44ac3a30=Role
 """
 
 helps['ad app permission admin-consent'] = """
@@ -170,16 +177,18 @@ examples:
 
 helps['ad app permission delete'] = """
 type: command
-short-summary: remove an API permission
+short-summary: Remove required API permissions.
+long-summary: Removes required API permissions from the app registration. Does not revoke any granted permissions.
 examples:
-  - name: remove an AAD graph permission
-    text: az ad app permission delete --id eeba0b46-78e5-4a1a-a1aa-cafe6c123456 --api 00000002-0000-0000-c000-000000000000
+  - name: Remove all required permissions to Microsoft Graph.
+    text: az ad app permission delete --id eeba0b46-78e5-4a1a-a1aa-cafe6c123456 --api 00000003-0000-0000-c000-000000000000
+  - name: Remove a specific required delegated permission to Microsoft Graph (User.Read, in this example).
+    text: az ad app permission delete --id eeba0b46-78e5-4a1a-a1aa-cafe6c123456 --api "https://graph.microsoft.com" --api-permissions e1fe6dd8-ba31-4d61-89e7-88639da4683d=Scope
 """
 
 helps['ad app permission grant'] = """
 type: command
-short-summary: Grant the app an API Delegated permissions
-long-summary: for Application permissions, please use "ad app permission admin-consent"
+short-summary: Grant delegated permissions
 examples:
   - name: Grant a native application with permissions to access an existing API with TTL of 2 years
     text: az ad app permission grant --id e042ec79-34cd-498f-9d9f-1234234 --api a0322f79-57df-498f-9d9f-12678 --expires 2
@@ -187,9 +196,10 @@ examples:
 
 helps['ad app permission list'] = """
 type: command
-short-summary: List API permissions the application has requested
+short-summary: List required API permissions for an app registration.
+long-summary: Lists required API permissions for an app registration. Does not list permission grants.
 examples:
-  - name: List the OAuth2 permissions for an existing AAD app
+  - name: List the required API permissions for an app registration.
     text: az ad app permission list --id e042ec79-34cd-498f-9d9f-1234234
 """
 
--- a/src/azure-cli/azure/cli/command_modules/role/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/role/custom.py
@@ -812,12 +812,39 @@ def _get_grant_permissions(graph_client, client_sp_object_id=None, query_filter=
     return grant_info
 
 
-def list_required_permissions(cmd, identifier):
+def list_required_permissions(cmd, identifier, skip_grant_expiry_time=False):
     # Returns the required API permissions (requiredResourceAccess)
+
     graph_client = _graph_client_factory(cmd.cli_ctx)
+    sp_client = graph_client.service_principals
+
     application = show_application(graph_client.applications, identifier)
-    permissions = application.required_resource_access
-    return permissions
+    required_permissions = application.required_resource_access
+
+    if not skip_grant_expiry_time:
+        # Retrieve any tenant-wide delegated permission grants where this app is the client app
+        granted_delegated_permissions = []
+        client = list(sp_client.list(filter="appId eq '{}'".format(application.app_id)))
+        if client:
+            granted_delegated_permissions = graph_client.oauth2_permission_grant.list(
+                filter="consentType eq 'AllPrincipals' and clientId eq '{}'".format(client[0].object_id))
+
+        for p in required_permissions:
+            expiry_time = 'N/A'
+            resource = list(sp_client.list(filter="appId eq '{}'".format(p.resource_app_id)))
+            if resource:
+                if any([x.resource_id == resource[0].object_id for x in granted_delegated_permissions]):
+                    # We use a date far in the future to signal that there does exist at least one
+                    # tenant-wide delegated permission grant for this resource app. The actual
+                    # values returned are not enforced by the service (i.e. even if the expiryTime
+                    # value is in the past, the permission is still considered granted). This does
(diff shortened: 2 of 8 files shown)


yonzhan:
add to S169

jiasli:
Hi @psignoret, thanks a lot for the contribution. We are on public holiday. I will review this PR tomorrow. By the way, `az ad app permission grant` does work with delegated permissions. Could you take a look at https://github.com/Azure/azure-cli/issues/12137#issuecomment-596567479?

psignoret:
Yes, `az ad app permission grant` does work for delegated permissions, but it _only_ works for delegated permissions, not for application permissions, where `az ad app permission add|delete|list` apply for both delegated permissions ("Scope") and application permissions ("Role"). The message telling the user to call `az ad app permission grant --id {id} --api {api}` had two issues:

1. The message suggests that with `az ad app permission grant` any permission can be granted (when in fact it's only for delegated permissions).
2. If the user was to copy-paste the command proposed by that message (without any scopes specified), the scope value "user_impersonation" would be granted, which is not necessarily valid, nor is it necessarily the permission which the user had added to the required permissions. 

I'm working on a separate PR to add support for application permissions (app role assignments). Once that's implemented, we can update `admin-consent` to use this.

psignoret:
@jiasli Have you had a change to review this PR?

jiasli:
Thanks @psignoret, I still haven't got a chance to review it as AD Graph API is not our priority. I don't think we will use the term `admin-consent` anymore. In fact, the whole command group needs an overhaul to better distinguish between **delegated permissions** and **application permissions**. 

We actually plan to do this in MS Graph migration, as the current work may be discarded during the process.

src/azure-cli/azure/cli/command_modules/role/custom.py
```diff
@@ -854,37 +836,77 @@ def list_permission_grants(cmd, identifier=None, query_filter=None, show_resourc
...
+    update_parameter = ApplicationUpdateParameters(required_resource_access=updated_rras)
     return graph_client.applications.patch(application.object_id, update_parameter)
 
 
+def _resolve_api_identifier_to_app_id(graph_client, api_identifier):
+    """Resolves an API identifier to an app ID
+
+    The identifier can be an object ID or service principal name for an existing service principal
+    object. If no matching service principal object is found but the identifier is a Guid, it is
+    assumed to be the app ID of an API which is not instantiated in the tenant.
```
jiasli, line 899:
```suggestion
    The identifier can be an object ID or service principal name, like `https://graph.microsoft.com`
    for an existing service principal object.
    If no matching service principal object is found but the identifier is a Guid, it is
    assumed to be the app ID of an API which is not instantiated in the tenant.
```

src/azure-cli/azure/cli/command_modules/role/custom.py
```diff
@@ -1222,12 +1239,16 @@ def delete_service_principal_credential(cmd, identifier, key_id, cert=False):
 
 
 def _resolve_service_principal(client, identifier):
-    # todo: confirm with graph team that a service principal name must be unique
```
jiasli, line 1225: This question remains unanswered. If a service principal name is not unique, some extra complexity may occur.

src/azure-cli/azure/cli/command_modules/role/custom.py
```diff
@@ -812,29 +812,11 @@ def _get_grant_permissions(graph_client, client_sp_object_id=None, query_filter=
...
+    # Returns the required API permissions (requiredResourceAccess)
     graph_client = _graph_client_factory(cmd.cli_ctx)
     application = show_application(graph_client.applications, identifier)
     permissions = application.required_resource_access
-    for p in permissions:
-        result = list(graph_client.service_principals.list(
-            filter="servicePrincipalNames/any(c:c eq '{}')".format(p.resource_app_id)))
-        expiry_time = 'N/A'
-        if result:
-            expiry_time = ', '.join([x.expiry_time for x in grant_permissions if
```
jiasli, line 835: I agree with you. The expiry time we retrieved is only for delegated permissions. We shouldn't fix it with application permisions.

When the app only has application permissions, I got empty `expiryTime`:

```
> az ad app permission list --id 46eb4122-bd2b-4f54-af7b-6d79b46ee31a
[
  {
    "additionalProperties": null,
    "expiryTime": "",
    "resourceAccess": [
      {
        "additionalProperties": null,
        "id": "9a5d68dd-52b0-4cc2-bd40-abcf44ac3a30",
        "type": "Role"
      }
    ],
    "resourceAppId": "00000003-0000-0000-c000-000000000000"
  }
]
```

However, I am not sure if this BREAKING CHANGE is necessary or if it can cause customers' script to fail. @achandmsft, do you think removing the misleading `expiryTime` is a good idea?

src/azure-cli/azure/cli/command_modules/role/custom.py
jiasli: As a side note, this is the value of `service_principal_names` for `api_identifier='https://graph.microsoft.com'`:

```
<class 'list'>: ['https://dod-graph.microsoft.us', 'Microsoft.Azure.AgregatorService', 
'00000003-0000-0000-c000-000000000000', '00000003-0000-0000-c000-000000000000/ags.windows.net',
'https://ags.windows.net', 'https://graph.microsoft.com', 'https://canary.graph.microsoft.com', 
'https://graph.microsoft.com/', 'https://graph.microsoft.us']
```

src/azure-cli/azure/cli/command_modules/role/custom.py
```diff
@@ -854,37 +836,77 @@ def list_permission_grants(cmd, identifier=None, query_filter=None, show_resourc
...
+    if rra is None:
+        rra = RequiredResourceAccess(resource_app_id=api_app_id, resource_access=[])
+        rras.append(rra)
+
+    # Add each required permission unless it already exists in the RRA.
     for e in api_permissions:
-        access_id, access_type = e.split('=')
-        resource_accesses.append(ResourceAccess(id=access_id, type=access_type))
+        permission_id, permission_type = e.split('=')
+        if not any((ra.id == permission_id and ra.type == permission_type) for ra in rra.resource_access):
```
jiasli, line 856: This is great change! If we don't detect duplication, we end up with duplicated entries in the request body:

```
> az ad app permission add --id 46eb4122-bd2b-4f54-af7b-6d79b46ee31a  --api 00000002-0000-0000-c000-000000000000 --api-permissions 311a71cc-e848-46a1-bdf8-97ff7156d8e6=Scope --debug

{
    "requiredResourceAccess": [
        {
            "resourceAccess": [
                {
                    "id": "311a71cc-e848-46a1-bdf8-97ff7156d8e6",
                    "type": "Scope"
                },
                {
                    "id": "311a71cc-e848-46a1-bdf8-97ff7156d8e6",
                    "type": "Scope"
                },
                {
                    "id": "311a71cc-e848-46a1-bdf8-97ff7156d8e6",
                    "type": "Scope"
                },
                {
                    "id": "311a71cc-e848-46a1-bdf8-97ff7156d8e6",
                    "type": "Scope"
                }
            ],
            "resourceAppId": "00000002-0000-0000-c000-000000000000"
        },
        {
            "resourceAccess": [
                {
                    "id": "c79f8feb-a9db-4090-85f9-90d820caa0eb",
                    "type": "Scope"
                },
                {
                    "id": "9a5d68dd-52b0-4cc2-bd40-abcf44ac3a30",
                    "type": "Role"
                }
            ],
            "resourceAppId": "00000003-0000-0000-c000-000000000000"
        }
    ]
}
```

However, the server should also be responsible for detecting duplication, right? 😉

src/azure-cli/azure/cli/command_modules/role/_help.py
```diff
@@ -146,16 +146,23 @@
...
-short-summary: manage an application's OAuth2 permissions.
+short-summary: Manage an app registration's required API permissions (requiredResourceAccess).
 """
 
 helps['ad app permission add'] = """
 type: command
-short-summary: add an API permission
-long-summary: invoking "az ad app permission grant" is needed to activate it
+short-summary: Add required API permissions.
+long-summary: Adds required API permissions to the app registration. Does not grant permissions.
```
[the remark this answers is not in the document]
psignoret: Updated.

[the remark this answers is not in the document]
psignoret: I agree, but I'm not trying to address that issue (which already exists) in this PR.

src/azure-cli/azure/cli/command_modules/role/_help.py
```diff
@@ -170,26 +177,29 @@
...
+  - name: Remove all required permissions to Microsoft Graph.
+    text: az ad app permission delete --id eeba0b46-78e5-4a1a-a1aa-cafe6c123456 --api 00000003-0000-0000-c000-000000000000
+  - name: Remove a specific required delegated permission to Microsoft Graph (User.Read, in this example).
+    text: az ad app permission delete --id eeba0b46-78e5-4a1a-a1aa-cafe6c123456 --api "https://graph.microsoft.com" --api-permissions e1fe6dd8-ba31-4d61-89e7-88639da4683d=Scope
 """
 
 helps['ad app permission grant'] = """
 type: command
-short-summary: Grant the app an API Delegated permissions
-long-summary: for Application permissions, please use "ad app permission admin-consent"
```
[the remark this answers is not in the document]
psignoret: I agree that is needed. I'm working on a separate PR to address app role assignments.

src/azure-cli/azure/cli/command_modules/role/_help.py
```diff
@@ -170,26 +177,29 @@
...
 examples:
   - name: Grant a native application with permissions to access an existing API with TTL of 2 years
     text: az ad app permission grant --id e042ec79-34cd-498f-9d9f-1234234 --api a0322f79-57df-498f-9d9f-12678 --expires 2
 """
 
 helps['ad app permission list'] = """
 type: command
-short-summary: List API permissions the application has requested
+short-summary: List required API permissions for an app registration.
+long-summary: Lists required API permissions for an app registration. Does not list permission grants.
```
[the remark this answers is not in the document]
psignoret: I agree. I'm working on a proposal in a separate PR.

src/azure-cli/azure/cli/command_modules/role/custom.py
psignoret: Added example value.

src/azure-cli/azure/cli/command_modules/role/custom.py
psignoret: If any script is depending on this value, it is a good thing that it will fail, as the value itself is already incorrect.

src/azure-cli/azure/cli/command_modules/role/custom.py
psignoret: I can confirm: a `servicePrincipalName` value uniquely identifies a **servicePrincipal** object within a tenant. In other words `$filter=servicePrincipalNames/any(n:n eq '{something')` will always return exactly 0 or 1 values.

src/azure-cli/azure/cli/command_modules/role/custom.py
psignoret: As discussed offline, I will add this property back (for permissions which have been granted tenant-wide, but the value will be set to a time far in the future, to ensure it is not misinterpreted as being expired.

psignoret: Updated to leave `expiryTime` in the response (if and only if there exists at least one tenant-wide delegated permission grant for the API), but sets the value far in the future. This avoids the breaking change, so I also updated the PR description to remove the BREAKING CHANGE.

src/azure-cli/azure/cli/command_modules/role/custom.py
psignoret: Bug opened internally to detect and avoid server-side duplicates. (It's not harmful in itself, but we can certainly reduce the chances of surprising a client which doesn't handle this edge case correctly.)

yonzhan:
add to S171

Issue #13317, opened 2020-05-04
Null templateLink returned in deployment().properties in 2.5.1 azure cli


priyank-b:
> ### `az feedback` auto-generates most of the information requested below, as of CLI version 2.5.1

**Describe the bug**

Following az command is brokern:
az deployment group create -g <resource_group> --template-uri "https://devtestzipacc.blob.core.windows.net/test/mainTemplate.json" --name "test"

It used to return templateLink correctly in older version. With latest update, it is sending null.

**To Reproduce**

Use this command in 2.5.1 azure cli. It returns null templateLink property.
az deployment group create -g <resource_group> --template-uri "https://devtestzipacc.blob.core.windows.net/test/mainTemplate.json" --name "test"

**Expected behavior**

Above command should populate templateLink as explained in public documentation:
- https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/linked-templates#using-variables-to-link-templates
- https://docs.microsoft.com/bs-latn-ba/azure/azure-resource-manager/templates/template-functions-deployment#remarks

**Environment summary**

**Additional context**

yonzhan:
add to S169

Issue #12990, opened 2020-04-13
az deployment {sub} create --template-uri... templateLink no longer available. 


Svessir:
## Describe the bug

**Command Name**
`az deployment create`

**Errors:**
```
Deployment failed. Correlation ID: 21bcc642-ce46-4888-848b-300b1ee9634b. {
  "error": {
    "code": "InvalidTemplate",
    "message": "Unable to process template language expressions for resource '/subscriptions/c88c80a7-d849-4bce-8907-ac15d5ca178b/providers/Microsoft.Resources/deployments/innResourceGroup' at line '61' and column '5'. 'The language expression property 'templateLink' doesn't exist, available properties are 'template, templateHash, parameters, mode, provisioningState'.'",
    "additionalInfo": [
      {
        "type": "TemplateViolation",
        "info": {
          "lineNumber": 61,
          "linePosition": 5,
          "path": ""
        }
      }
    ]
  }
}
```

## To Reproduce:
Steps to reproduce the behavior. Note that argument values have been redacted, as they may contain sensitive information.

- _Put any pre-requisite steps here..._
- `az deployment create --name {} --location {} --template-uri {} --parameters {} --parameters {} --verbose`

## Expected Behavior
I used the --template-uri to supply the template so I expect deployment().properties.templateLink
to be available to use in the template like in earlier versions of az cli. I'm guessing the az cli is now downloading the template from the source and running it similar to a local template? az deployment sub create has the same behavior. I ran the templates with an earlier version of az cli successfully (version 2.0.69).

## Environment Summary
```
Linux-5.3.0-46-generic-x86_64-with-debian-buster-sid
Python 3.6.5
Installer: DEB

azure-cli 2.3.1

```

yonzhan:
add to S169

ckittel:
I can confirm this is happening as well, it looks like the template is downloaded and then provided to `/validate` as if it was a local file.

yonzhan:
add to S171

zhoxing-ms:
Hi @Svessir, this is the same problem as #13317, and I have submitted a PR #13344 to fix it~ You can track the progress of the problem through this PR.

Issue #12890, opened 2020-04-06
2.3.0: Breaking changes to template deployments when using --template-uri


noelbundick:
## Describe the bug

Starting with azure-cli 2.3.0, the default behavior has been updated to process extended JSON syntax.

Previously, a template deployed with `--template-uri` would be executed with the `templateLink` property set. At runtime, `[deployment().properties.templateLink.uri]` would be accessible

With the new behavior, the CLI will download the template from the remote host, preprocess it, then PUT a deployment with `template` set. At runtime, `[deployment().properties.templateLink.uri]` is not accessible. 

This appears to be an undocumented breaking change and causes my templates to fail

The following test recording was last run in Dec 2019. I believe if re-run today, this would demonstrate different results showing a PUT with the full template rather than the templateUri: https://github.com/Azure/azure-cli/blob/991cb7cc7c9e577d8bec3b78e0316dfb5d39fe10/src/azure-cli/azure/cli/command_modules/resource/tests/latest/recordings/test_group_deployment_thru_uri.yaml

**Command Name**
`az deployment group create`

**Errors:**
Template deploys successfully. `[deployment().properties.templateLink]` does not exist

## To Reproduce:

- `az group create -n test1 -l westus2`
- `az deployment group create -g test1 --template-uri https://gist.githubusercontent.com/noelbundick/2a599b3954c9c4e47754151f138d942a/raw/45e64f07f4c82462ebcb09c6bd083ae01c3e4d3c/azuredeploy.json`

## Expected Behavior

`[deployment().properties.templateLink]` is available per the previous behavior and documentation (https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-deployment#deployment)

## Environment Summary
```
Linux-4.19.84-microsoft-standard-x86_64-with-debian-buster-sid
Python 3.6.5
Installer: DEB

azure-cli 2.3.0 *
azure-cli-ml 1.0.85

Extensions:
azure-cli-ml 1.0.85
azure-devops 0.16.0

```
## Additional Context

I have a sample zero resource template here that demonstrates the issue: https://gist.github.com/noelbundick/2a599b3954c9c4e47754151f138d942a

Works (2.2.0):

```shell
docker run --rm -it -v /home/$USER/.azure:/root/.azure mcr.microsoft.com/azure-cli:2.2.0 bash
[...]

PR #13344, opened 2020-05-06
[ARM] az deployment: Fix the problem that the templateLink will not be returned when deploying or validating template-uri (#13317)

zhoxing-ms:
Issue: #13317 #12890 #12990

**Description**  
This problem is caused by the simultaneous values of the `template` and `template_link` parameters in the `DeploymentProperties` when handling `--handle-extended-json-format`:
The service API returns templateLink only if the request parameter is templateLink.

When both `template_link` and `template` have values in `DeploymentProperties`, the Python SDK will take the data of `template` as the main data, and the request body for service API are:
```
{"properties": {"parameters": {}, "mode": "Incremental", template:{
    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
    "contentVersion": "1.0.0.0",
    "resources": [],
    "outputs": {
        "deploymentOutput": {
            "value": "[deployment()]",
            "type" : "object"
        }
    }
}}}
```
Therefore, the service API does not return templateLink.

And When there is only `templatelink` in `DeploymentProperties`，the request body for service API are:
```
{"properties": {"templateLink": {"uri": "https://devtestzipacc.blob.core.windows.net/test/mainTemplate.json"}, "parameters": {}, "mode": "Incremental"}}
```
In this case, the service API will return templateLink

By the way, if the request parameter in `DeploymentProperties` is only `templateLink`, there is no need to change the configuration of the deployment_client that sent the parameter as JSON template, otherwise the service API will report an error.

**Testing Guide**  
When deploying or validating the ARM template, please use the `--template-uri` parameter to see if tempalteLink will be returned.

**History Notes**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/resource/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/resource/custom.py
@@ -291,8 +291,7 @@ def _deploy_arm_template_core_unmodified(cli_ctx, resource_group_name, template_
     template_content = None
     if template_uri:
         template_link = TemplateLink(uri=template_uri)
-        template_content = _urlretrieve(template_uri).decode('utf-8')
-        template_obj = _remove_comments_from_json(template_content)
+        template_obj = _remove_comments_from_json(_urlretrieve(template_uri).decode('utf-8'))
     else:
         template_content = read_file_content(template_file)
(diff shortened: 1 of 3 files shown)


yonzhan:
add to S169

PR #13353, opened 2020-05-06
[Packaging] Build MSI/Homebrew packages with dependecies resolved from requirements.txt

fengzhou-msft:
**Description**  

This PR finishes the Packaging deployment part for #9781 to build with requirements.\*.txt files for all packages. To resolve the issue, we need to change azdev and CI to use requirements.\*.txt to resolve depedencies as well. @haroldrandom for awareness.

I have also submitted the PR https://github.com/Azure/azure-cli-dev-tools/pull/194 for azdev setup to use requirements file.

**Testing Guide**  

---


yonzhan:
Packaging

haroldrandom:
What do azdev need to do?

fengzhou-msft:
> What do azdev need to do?

I have submitted the PR https://github.com/Azure/azure-cli-dev-tools/pull/194 in azdev. azdev setup will use requirements file to download dependencies by default. But we still provide the option with `--deps setup` to download dependencies based on setup.py. The azdev PR currently has failed CI because some dependencies in the requirements file is old and not compatible with dependencies of azdev. I updated them in this PR. After merging  this PR, the azdev PR should pass CI.

After merging the azdev PR, we need to modify pipeline script to test both setup.py and requirements file. I'm thinking about using setup.py for python 3.8 tests and requirements file for python 3.6 since we use requirements file for all packages except pip install with this PR and some packages have a bundled python 3.6.

What do you think?

src/azure-cli/requirements.py3.windows.txt
```diff
@@ -119,7 +119,7 @@ pywin32==225
 requests==2.22.0
 requests-oauthlib==1.2.0
 scp==0.13.2
-six==1.12.0
+six==1.14.0
```
arrownj, line 122: should we consider remove all `six` dependency as we only support python 3 now ?

fengzhou-msft: Good point. I will work on cleaning up dependencies in another PR. Feel free to suggest any other dependencies that could be removed.

haroldrandom:
@fengzhou-msft 

> What do you think?

I could be done.

I have one question. How do we keep up with the version uprgade of those dependencies? How often we update requirements.txt?

fengzhou-msft:
> @fengzhou-msft
> 
> > What do you think?
> 
> I could be done.
> 
> I have one question. How do we keep up with the version uprgade of those dependencies? How often we update requirements.txt?

I think the current practice is when we need to use new features or when customers requested to do so, we update the versions.

haroldrandom:
we can also make setup.py use requirements.txt too.
https://stackoverflow.com/questions/14399534/reference-requirements-txt-for-the-install-requires-kwarg-in-setuptools-setup-py

build_scripts/windows/scripts/build.cmd
```diff
@@ -86,26 +86,17 @@ set PYTHON_EXE=%PYTHON_DIR%\python.exe
...
 )
-echo Built CLI packages successfully.
+%BUILDING_DIR%\python.exe -m pip install -r %CLI_SRC%\azure-cli\requirements.py3.windows.txt
 
 if %errorlevel% neq 0 goto ERROR
 
-set ALL_MODULES=
-for %%i in (%TEMP_SCRATCH_FOLDER%\*.whl) do (
-    set ALL_MODULES=!ALL_MODULES! %%i
-)
```
jiasli, line 105: The removal of `TEMP_SCRATCH_FOLDER` breaks the `privates` private wheel injection mechanism.

PR #13356, opened 2020-05-06
{Storage} Use get_login_credentials from Core for --auth-mode login in track2

Juliehzl:
**Description**  

Change the way to get token credential when `--auth-mode` login used in track 2.

Storage data plane operations are using an addition token updater to retrieve token credential, but we can utilize the `get_login_credentials()` in core module to do the same thing, which will make data plane and mgmt plane align in track 2.

 The method to distinguish track1 and track2 is api version here, because track 2 SDK starts from v2019-02-02.

**Testing Guide**  

Try `az storage container create` with `--auth-mode login`.

**History Notes**  

---


yonzhan:
add to S169

src/azure-cli/azure/cli/command_modules/storage/_validators.py
```diff
@@ -117,7 +102,9 @@ def validate_client_parameters(cmd, namespace):
         if not n.account_name:
             n.account_name = get_config_value(cmd, 'storage', 'account', None)
         if auth_mode == 'login':
-            n.token_credential = _create_token_credential(cmd.cli_ctx)
+            from azure.cli.core._profile import Profile
+            profile = Profile(cli_ctx=cmd.cli_ctx)
+            n.token_credential, _, _ = profile.get_login_credentials(resource="https://storage.azure.com")
```
jiasli, line 107: Do we want to hard-code `resource`? Shouldn't it be added to the cloud? Just want to confirm.

jiasli:
Please title the PR with meaningful description, like "{Storage} Use get_login_credentials from Core for --auth-mode login".

Also, please fix the formatting of the PR description.

src/azure-cli/azure/cli/command_modules/storage/oauth_token_util.py
```diff
@@ -1,48 +0,0 @@
-# --------------------------------------------------------------------------------------------
-# Copyright (c) Microsoft Corporation. All rights reserved.
-# Licensed under the MIT License. See License.txt in the project root for license information.
-# --------------------------------------------------------------------------------------------
-
-import threading
-
-
-class TokenUpdater(object):
```
jiasli, line 9: I am actually curios about the design of this `TokenUpdater`. Why does it actively refresh the token instead of passively 
 refreshing it on demand? Be careful of pitfalls that may be introduced by this change.

src/azure-cli/azure/cli/command_modules/storage/oauth_token_util.py
```diff
@@ -1,48 +0,0 @@
...
-        self.timer_callback()
-
-    def timer_callback(self):
-        # call to get a new token and set a timer
-        from azure.cli.core._profile import Profile
-        from datetime import datetime
-        # should give back token that is valid for at least 5 mins
-        token = Profile(cli_ctx=self.cli_ctx).get_raw_token(resource="https://storage.azure.com")[0][2]
-        try:
-            self.token_credential.token = token['accessToken']
```
jiasli, line 30: It seems the token is accessed via `token_credential.token`. `AdalAuthentication` doesn't have this field, but only has `get_token`. Won't this break track 1 SDK?

src/azure-cli/azure/cli/command_modules/storage/_validators.py
```diff
@@ -117,7 +117,12 @@ def validate_client_parameters(cmd, namespace):
         if not n.account_name:
             n.account_name = get_config_value(cmd, 'storage', 'account', None)
         if auth_mode == 'login':
-            n.token_credential = _create_token_credential(cmd.cli_ctx)
+            if cmd.supported_api_version(min_api='2019-02-02'):
```
qianwens, line 120: >if cmd.supported_api_version(min_api='2019-02-02'): [](start = 12, length = 51)

add comment that this check is for track1/track2

src/azure-cli/azure/cli/command_modules/storage/_validators.py
Juliehzl: it is hard-coded resource, not related to cloud.

src/azure-cli/azure/cli/command_modules/storage/oauth_token_util.py
Juliehzl: actually what we need is signed session in track 1. `token` is used in signed session in the updater.

src/azure-cli/azure/cli/command_modules/storage/_validators.py
Juliehzl: done

src/azure-cli/azure/cli/command_modules/storage/oauth_token_util.py
Juliehzl: refreshing it when expired. Here the code is changed to keep previous token credential for track 1 SDK.

Juliehzl:
@qianwens and @jiasli Please help review the PR again because we want to make it merged asap to let adls utilize this part and test with service team.

src/azure-cli/azure/cli/command_modules/storage/oauth_token_util.py
jiasli: Aha, I see: https://github.com/Azure/azure-multiapi-storage-python/blob/726b1b0cf3c61aae52fc410451f968cbc4e675bc/azure/multiapi/storage/v2018_11_09/common/tokencredential.py#L25

```py
    def signed_session(self, session=None):
        """
        Sign requests session with the token. This method is called every time a request is going on the wire.
        The user is responsible for updating the token with the preferred tool/SDK.
        In general there are two options:
            - override this method to update the token in a preferred way and set Authorization header on session
            - not override this method, and have a timer that triggers periodically to update the token on this class
        The second option is recommended as it tends to be more performance-friendly.
        :param session: The session to configure for authentication
        :type session: requests.Session
        :rtype: requests.Session
        """
        session = session or requests.Session()
        session.headers['Authorization'] = "Bearer {}".format(self.token)

        return session

    def token(self, new_value):
        """
        :param new_value: new value to be set as the token.
        """
        self.token = new_value
```

PR #13315, opened 2020-05-04
[ACS] bug fix: remove fields cleanup for fields that no longer exist

olga-mir:
**Description**  
There was an error introduced in https://github.com/Azure/azure-cli/pull/12240 where AROv3   has been upgraded to latest API version. Fields `name` and `os_type` has been dropped from master pool profile in azure-sdk-for-python: https://github.com/Azure/azure-sdk-for-python/pull/10464/files

**Testing Guide**  
create an OpenShift cluster 
`az openshift create -g <group_name> -n <resource_name> --location eastus2`
try show or list command:
`az openshift list`
`az openshift show -g <group_name> -n <resource_name>`
They should execute without error and display the cluster info.

**History Notes**  

[Component Name 1] BREAKING CHANGE: az command a: Make some customer-facing breaking change.  
[Component Name 2] az command b: Add some customer-facing feature.

---

This checklist is used to make sure that common guidelines for a pull request are followed.

- [x] The PR title and description has followed the guideline in [Submitting Pull Requests](https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules#submitting-pull-requests).

- [x] I adhere to the [Command Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md).

PR #13358, opened 2020-05-06
{ACS} az openshift show/list: Refactoring to reduce the probability of AttributeError

m1kola:
**Description**  

The changes in the original PR were a duplicate of https://github.com/Azure/azure-cli/pull/13315 which got merged recently. I rebased the PR and I updated it so we hopefully reduce the probability of `AttributeError` in the similar cases.

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acs/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/custom.py
@@ -3234,7 +3234,7 @@ def _remove_osa_nulls(managed_clusters):
     attrs = ['tags', 'plan', 'type', 'id']
     for managed_cluster in managed_clusters:
         for attr in attrs:
-            if getattr(managed_cluster, attr, None) is None:
+            if hasattr(managed_cluster, attr) and getattr(managed_cluster, attr) is None:
                 delattr(managed_cluster, attr)
     return managed_clusters
 


src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -3232,14 +3232,10 @@ def _remove_osa_nulls(managed_clusters):
     by the server, but get recreated by the CLI's own "to_dict" serialization.
     """
     attrs = ['tags', 'plan', 'type', 'id']
-    ap_master_attrs = ['name', 'os_type']
```
m1kola, line 3235: It doesn't seem like we ever had `name` and `os_type` attributes in the public API.

m1kola:
@julienstroheker could you also please take a look?

src/azure-cli/azure/cli/command_modules/acs/custom.py
m1kola: But this code been here since the beginning of times too. Not sure what broke the command.

yonzhan:
add to S169

src/azure-cli/azure/cli/command_modules/acs/custom.py
arrownj: Hi @m1kola , is this because the different API version? Are you sure we can remove these code safely?

m1kola: @arrownj I'm currently not sure why it's happening. It was working fine on Azure CLI 2.3.1 (maybe on newer versions too) and `az openshift show` and `az openshift list` are broken on 2.5.1 and `dev` branch.

As far as I see, we did not change anything in the API and we never had `name` and `os_type` fields in the master pool profile API on the server (I checked all the supported versions in [our codebase](https://github.com/openshift/openshift-azure/tree/c70152ea551524eb7470c19519d01a1d7035ea8d/pkg/api)).

jim-minter: Here's the break:

https://github.com/Azure/azure-sdk-for-python/commit/3f6238b2a224df11fa88c6501e078aac96697277#diff-66c3074fc0745e8946eba5783a5f8f32R21

The client library we were using was stale having been generated on an ancient pre-GA API and ^ brings us up to date.

src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -3232,14 +3232,10 @@ def _remove_osa_nulls(managed_clusters):
...
     for managed_cluster in managed_clusters:
         for attr in attrs:
             if getattr(managed_cluster, attr, None) is None:
```
jim-minter, line 3237:
```suggestion
            if hasattr(managed_cluster, attr) and getattr(managed_cluster, attr) is None:
```

arguably ^ is better, but it doesn't really matter.

m1kola: If remember Python correctly `getattr(managed_cluster, attr)` still needs to be `getattr(managed_cluster, attr, None)`. Otherwise there will be an exception in case when there is no attribute.

m1kola: But it makes sense. I'll update.

m1kola: Ahh. I see... `hasattr` :) Tired brain.

m1kola:
It looks like it is a duplicated of #13315

m1kola:
The changes in the original PR were a duplicate of https://github.com/Azure/azure-cli/pull/13315 which got merged recently. I rebased the PR and I updated it so we hopefully reduce the probability of `AttributeError` in the similar cases.

Please take another look.

jim-minter:
LGTM.

PR #13366, opened 2020-05-06
[AppService] az webapp auth update: Add optional parameter to update runtime-version

mkarmark:
**Description**  
We want to offer support via the Azure CLI tool to update the runtime version of the App Service Authentication/Authorization module. This is done by exposing a runtime version parameter in the "webapp auth update" command, already offered through the Azure CLI tool.

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -740,7 +740,17 @@ def update_auth_settings(cmd, resource_group_name, name, enabled=None, action=No
         auth_settings.unauthenticated_client_action = UnauthenticatedClientAction.allow_anonymous
     elif action:
         auth_settings.unauthenticated_client_action = UnauthenticatedClientAction.redirect_to_login_page
-        auth_settings.default_provider = AUTH_TYPES[action]
+        auth_settings.default_provider = AUTH_TYPES[action]    
+    # validate runtime version
+    from version_parser.version import Version
+    try:
+        if runtime_version is not None:
+            if runtime_version.startswith("~") and len(runtime_version) > 1:
+                Version(runtime_version[1:])
+            else:
+                Version(runtime_version)
+    except(Exception, SystemExit) as ex:
+        raise CLIError('Usage Error: --runtime-version set to invalid value')
 
     import inspect
     frame = inspect.currentframe()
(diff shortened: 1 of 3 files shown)


yonzhan:
add to S169

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -724,7 +724,7 @@ def get_auth_settings(cmd, resource_group_name, name, slot=None):
 
 
 def update_auth_settings(cmd, resource_group_name, name, enabled=None, action=None,  # pylint: disable=unused-argument
-                         client_id=None, token_store_enabled=None,  # pylint: disable=unused-argument
+                         client_id=None, token_store_enabled=None, runtime_version=None,  # pylint: disable=unused-argument
```
yungezz, line 727: will be param be passed to sdk call? didn't see it

qwordy: `runtime_version` is not used.

qwordy:
The PR title does not conform to the requirement.
https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules#format-pr-title

src/azure-cli/azure/cli/command_modules/appservice/custom.py
mkarmark: runtime_version will be set in lines 752-4

qwordy: I am still confused. I search "runtime_version" and can't find usage.

panchagnula:
@mkarmark can you add a test case for this - I am still confused how this is working? You might also need to update params.py and help for this new property.

mkarmark:
@panchagnula Added the test in the latest PR, but a lot of the work (i.e. params.py) for runtime version was already done, I am just connecting the final pieces in order to get cli support for pinning.

yonzhan:
add to S171

mkarmark:
@panchagnula, here is the expected behavior demonstrated with the runtime version listed in an invalid format
image

panchagnula:
> @panchagnula, here is the expected behavior demonstrated with the runtime version listed in an invalid format
> image

this means if we pass even an invalid format the API just sets the value? is this API going to handle this? if not should CLI validate the string & then pass it to API?

mkarmark:
> > @panchagnula, here is the expected behavior demonstrated with the runtime version listed in an invalid format
> > image
> 
> this means if we pass even an invalid format the API just sets the value? is this API going to handle this? if not should CLI validate the string & then pass it to API?

I pushed some changes that result in the following error message upon an invalid version:
image

Issue #12775, opened 2020-03-27
az vm create: add flag for "requireGuestProvisionSignal" api field


colemickens:
**Is your feature request related to a problem? Please describe.**
Yes. I would like to boot VMs without the agent, but I do want to boot them through the CLI.

**Describe the solution you'd like**
Expose a `az vm create` flag that corresponds to the `requireGuestProvisionSignal` that is part of hte VM Create REST API.

Ex: 
```
az vm create --image /sub/image-without-agent --require-guest-provisioning-signal false
```

**Describe alternatives you've considered**
I can't come up with any.

yonzhan:
@qwordy please take a look.

qwordy:
I get an error when testing this field.
```
The property 'requireGuestProvisionSignal' is not valid because the 'Microsoft.Compute/Agentless' feature is not enabled for this subscription
```
When I try to register it, I get
```
The feature 'Agentless' could not be found.
```
Can you use this field successfully?

colemickens:
It seems like it's been part of the public Compute API for years, I just assumed it was available. I was waiting for CLI support to test it out.

adritrix:
@qwordy I’m having the same problem with the 'requireGuestProvisionSignal' error. Did you find a solution? If so I would appreciate the help you could provide me.

qwordy:
Hi service team, do you know how to use requireGuestProvisionSignal? I try to register the feature, I get "The feature 'Agentless' could not be found."

adritrix:
@qwordy In my case, I received that error while trying to deploy Rome resources through a template. I fixed it by removing the line. 

  **"requireGuestProvisionSignal": true**

Drewm3:
This property was inappropriately added to the public APIs, and it is not currently supported. This was added at one point in the past by someone doing experimentation, but it should not have shown up in the public APIs. 

I will work with the team to get this property removed from the public APIs because it is not supported for general usage.

colemickens:
Okay, I'll explore other options in the meantime. I don't know if y'all want to leave this open to track removal from API, but we can close this one whenever. Thanks for the update @Drewm3!

colemickens:
Also, since I didn't link them already... for anyone else here following breadcrumbs, the related `walinuxagent` issue: https://github.com/Azure/WALinuxAgent/issues/1888

qwordy:
Thank you @Drewm3

davejdyer:
[...]

PR #13369, opened 2020-05-07
[Compute] az vm create: Add --guest-provision-signal

qwordy:
**Description**  

https://github.com/Azure/azure-cli/issues/12775
[Compute] az vm create: Add --guest-provision-signal

I get an error when testing this field.
```
The property 'requireGuestProvisionSignal' is not valid because the 'Microsoft.Compute/Agentless' feature is not enabled for this subscription
```
When I try to register it, I get
```
The feature 'Agentless' could not be found.
```

**Testing Guide**  

**History Notes**  

---


src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -321,7 +321,7 @@ def load_arguments(self, _):
                        help="The eviction policy for the Spot priority virtual machine. Default eviction policy is Deallocate for a Spot priority virtual machine")
         c.argument('enable_agent', arg_type=get_three_state_flag(), min_api='2018-06-01',
                    help='Indicates whether virtual machine agent should be provisioned on the virtual machine. When this property is not specified, default behavior is to set it to true. This will ensure that VM Agent is installed on the VM so that extensions can be added to the VM later')
-
+        c.argument('guest_provision_signal', arg_type=get_three_state_flag(), help='Specifies whether the guest provision signal is required to infer provision success of the virtual machine.')
```
jiasli, line 324: I'd prefer `require_guest_provision_signal`. Without "require", the mean is vague and not intuitive.

yonzhan:
add to S169

qwordy:
> This property was inappropriately added to the public APIs, and it is not currently supported. This was added at one point in the past by someone doing experimentation, but it should not have shown up in the public APIs.
I will work with the team to get this property removed from the public APIs because it is not supported for general usage.

from service team

Issue #12938, opened 2020-04-09
Disconnected and Offline CLI Functionality Issues


rjaffee:
**Describe the bug**
CLI does not function correctly when run within a VNet that is isolated from having access to the Internet. CLI makes calls to the internet for commands such as AZ VM Create and AZ Copy and others. This presents an issue when the VNet is isolated from the internet or an organization is not able to download and install software on the fly.

**To Reproduce**
1)	Setup a VNet with a windows VM inside.
2)	Install CLI (2.2.0/2.3.1) on Windows VM (Server 2016)
3)	Isolate Subnet (NSG) by denying all Internet traffic inbound and outbound (ability to still reach internal Azure Endpoints)
Examples: 
•	For authorization, in the python adal library az cli uses, there is a hardcoded whitelist of known AAD endpoints and if it cannot find the endpoint, it makes a call to logon.windows.net.  
•	Attempt to create a VM (Az VM Create) – Issue #12902
•	Attempt to Copy files to VHD (Az Storage Copy) – Calls to internet to download AzCopy if AZCopy doesn’t exist already.

**Expected behavior**
CLI should have full functionality when the ability to make calls to the internet and/or download additional content is restricted or unavailable. CLI commands should not be restricted based on internet availability of the end point they are being run from. 

**Environment summary**
•	Windows Server 2016 VM on an Azure VNet Isolated from the internet
•	CLI 2.2.0/2.3.1 and PowerShell 7.0 

**Additional context**
Requesting a code review to enable full CLI functionality when internet access is not available or heavily restricted on what can be downloaded and installed.  
Suggestion - a standalone or offline installer package for CLI that includes additional files and packages that would normally be checked for and downloaded if not available locally. 

Example environment(s): 
•	Isolated VNet or resource from Internet
•	Disconnected Azure Stack 
For disconnected stack the following addition command was needed for Azure Copy functionality:
$env:AZCOPY_DEFAULT_SERVICE_API_VERSION="2017-11-09"

CLI should set these types of environmental variables based on the registered Azure cloud profile - (Az Cloud Register). In this case the profile was 2018-03-01-hybrid.

yonzhan:
add to S168

jiasli:
[...]

PR #13371, opened 2020-05-07
{Profile} Bump ADAL version to 1.2.3

jiasli:
**Description**  
Fix https://github.com/Azure/azure-cli/issues/12938: `az login` fails because instance discovery failes with `login.windows.net` in private cloud

This PR bumps ADAL version to 1.2.3 so that private cloud can use `login.microsoftonline.com` to make instance discovery. The old ADAL uses `login.windows.net` to make instance discovery, which fails in a private cloud.

Corresponding ADAL issue: https://github.com/AzureAD/azure-activedirectory-library-for-python/issues/224
Corresponding ADAL PR: https://github.com/AzureAD/azure-activedirectory-library-for-python/pull/227

**Testing Guide**  
`az login`

### DIFF
--- a/src/azure-cli-core/setup.py
+++ b/src/azure-cli-core/setup.py
@@ -50,7 +50,7 @@ CLASSIFIERS = [
 ]
 
 DEPENDENCIES = [
-    'adal~=1.2',
+    'adal~=1.2.3',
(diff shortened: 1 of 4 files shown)


yonzhan:
add to S169

PR #13377, opened 2020-05-07
[Network] az network dns zone: support - character

stegag:
**Description**  
See internal ICM 187136418
We are failing to import zones when domain name in $origin contains hypens.
This is because we don't parse it correctly in the regex (only alphanumeric char are evaluated with "\w" and that doesn't include hypen, like it is for outher resurces like CNAME.

**Testing Guide**  

**History Notes**  

---


yonzhan:
add to S169

PR #13392, opened 2020-05-08
[ACR]: bug fixes on token create and image import commands

yugangw-msft:
**Description**  
2 ACR command updates, see history notes below for details

**Testing Guide**  
Added e2e test for token creation
Conducted manual test on image bug fixes
**History Notes**  
[ACR] az acr token create: expose --days argument
[ACR] az acr import: accept --source argument values which contain login in server name through client end correction

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acr/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/acr/_params.py
@@ -320,17 +320,16 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('repository_actions_list', options_list=['--repository'], nargs='+', action='append',
                    help='repository permissions. Use the format "--repository REPO [ACTION1 ACTION2 ...]" per flag. ' + valid_actions)
         c.argument('no_passwords', arg_type=get_three_state_flag(), help='Do not generate passwords, instead use "az acr token credential generate"')
+        c.argument('expiration_in_days', help='Number of days for which the credentials will be valid. If not specified, the expiration will default to the max value "9999-12-31T23:59:59.999999+00:00"', type=int, required=False)
 
     with self.argument_context('acr token update') as c:
         c.argument('scope_map_name', options_list=['--scope-map'], help='The name of the scope map associated with the token. If not specified, running this command will disassociate the current scope map related to the token.', required=False)
 
-    for scope in ['acr token create', 'acr token credential generate']:
-        with self.argument_context(scope) as c:
-            c.argument('days', options_list=['--days'], help='Number of days for which the credentials will be valid. If not specified, the expiration will default to the max value "9999-12-31T23:59:59.999999+00:00"', type=int, required=False)
-
     with self.argument_context('acr token credential generate') as c:
         c.argument('password1', options_list=['--password1'], help='Flag indicating if password1 should be generated.', action='store_true', required=False)
         c.argument('password2', options_list=['--password2'], help='Flag indicating if password2 should be generated.', action='store_true', required=False)
+        c.argument('expiration_in_days', options_list=['--expiration-in-days', c.deprecate(target='--days', redirect='--expiration-in-days', hide=True)],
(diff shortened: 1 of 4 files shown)


yonzhan:
add to S169

src/azure-cli/azure/cli/command_modules/acr/token.py
```diff
@@ -21,12 +21,15 @@ def acr_token_create(cmd,
...
-                     no_passwords=None):
+                     no_passwords=None,
+                     days=None):
     from knack.log import get_logger
     from ._utils import get_resource_id_by_registry_name
 
     if bool(repository_actions_list) == bool(scope_map_name):
         raise CLIError("usage error: --repository | --scope-map-name")
+    if no_passwords and days is not None:
+        raise CLIError("usage error: --no-passwords and --days are mutually exclusive.")
```
qwordy, line 32: What if no_passwords is not None and days? Will CLI throw an error?

yugangw-msft: if `no_password` is False or just missing, then it is a valid case to have `days` for user specified expiry time, or `9999-12-31` otherwise

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -324,10 +324,13 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
     with self.argument_context('acr token update') as c:
         c.argument('scope_map_name', options_list=['--scope-map'], help='The name of the scope map associated with the token. If not specified, running this command will disassociate the current scope map related to the token.', required=False)
 
+    for scope in ['acr token create', 'acr token credential generate']:
+        with self.argument_context(scope) as c:
+            c.argument('days', options_list=['--days'], help='Number of days for which the credentials will be valid. If not specified, the expiration will default to the max value "9999-12-31T23:59:59.999999+00:00"', type=int, required=False)
```
yungezz, line 329: expiration_in_days? I think there's existing guideline of not putting unit in parameter name.

src/azure-cli/azure/cli/command_modules/acr/import.py
```diff
@@ -50,10 +50,15 @@ def acr_import(cmd,
     if source_registry:
         if is_valid_resource_id(source_registry):
             source = ImportSource(resource_id=source_registry, source_image=source_image)
-
         else:
             registry = get_registry_from_name_or_login_server(cmd.cli_ctx, source_registry, source_registry)
             if registry:
+                # trim away redundant login server name, a common error
+                prefix = registry.login_server + '/'
```
yungezz, line 57: how about add validation earlier at params?

src/azure-cli/azure/cli/command_modules/acr/_params.py
yugangw-msft: You are right. The existing `az acr token credential generate` already uses the same name, let me update that one as well . Will put in deprecate. Since the whole command is still under preview, so the eventual breaking is fine.

src/azure-cli/azure/cli/command_modules/acr/import.py
yugangw-msft: This change will be a bit involving, as to do it right, other 3 parameters should also be moved to the validator not just this one. Can I log an issue and do it later?

yugangw-msft:
@qwordy @yungezz, I have addressed the review feedback and also opened a tracking issue for improving the parameter validating. Could you merge the PR when ready; or let me know if there are other things I should address. Thanks.

src/azure-cli/azure/cli/command_modules/acr/import.py
yugangw-msft: Logged https://github.com/Azure/azure-cli/issues/13415

yungezz:
hi @haroldrandom could you pls have a look at the CI failure? why the timeout is 10 mins?

haroldrandom:
> hi @haroldrandom could you pls have a look at the CI failure? why the timeout is 10 mins?

It passed after rerun. 
10 mins is enough for this job based on previous experience, it failed maye be caused by network jetlet occationally.

PR #13412, opened 2020-05-09
{AppService} Use azure.multiapi.storage for storage blob

Juliehzl:
**Description**  

Because the latest `azure.storage.blob` is for track2 without supporting for `BlockBlobService`, `AppendBlobService` ..., which are in track1, using `azure.multiapi.storage` can prevent the breaking change in `azure.stroage.blob`.

**Testing Guide**  

The influenced command is `functionapp deployment source config-zip`

**History Notes**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -36,7 +36,6 @@ from azure.mgmt.storage import StorageManagementClient
 from azure.mgmt.applicationinsights import ApplicationInsightsManagementClient
 from azure.mgmt.relay.models import AccessRights
 from azure.cli.command_modules.relay._client_factory import hycos_mgmt_client_factory, namespaces_mgmt_client_factory
-from azure.storage.blob import BlockBlobService, BlobPermissions
 from azure.cli.command_modules.network._client_factory import network_client_factory
 
 from azure.cli.core.commands.client_factory import get_mgmt_service_client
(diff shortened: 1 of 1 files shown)


Juliehzl:
@ankitkumarr Can you also look at this change?

yonzhan:
add to S169

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -502,7 +501,7 @@ def upload_zip_to_storage(cmd, resource_group_name, name, src, slot=None):
 
     container_name = "function-releases"
     blob_name = "{}-{}.zip".format(datetime.datetime.today().strftime('%Y%m%d%H%M%S'), str(uuid.uuid4()))
-
+    BlockBlobService = get_sdk(cmd.cli_ctx, ResourceType.DATA_STORAGE, 'blob#BlockBlobService')
```
qwordy, line 504: Can we import it statically like `from azure.storage.blob import BlockBlobService`?
What is the advantage of this approach?

Juliehzl: Because the latest azure.storage.blob is for track2 without supporting for BlockBlobService, AppendBlobService ..., which are in track1, using azure.multiapi.storage can prevent the breaking change in azure.stroage.blob.

Juliehzl:
@ankitkumarr  @panchagnula  Can you also take a look?

ankitkumarr:
@Juliehzl, I think this looks fine, although I am not fully aware `get_sdk`. Assuming it returns the right Service back.

Were you able to run the command (using Linux Consumption function app) and make sure it still works as expected? Please let me know if you have questions executing an e2e scenario. If that works fine, it should be ok.

Juliehzl:
> @Juliehzl Zunli Hu FTE, I think this looks fine, although I am not fully aware of the `get_sdk`. Assuming it returns the right Service back.
> 
> Were you able to run the command (using Linux Consumption function app) and make sure it still works as expected? Please let me know if you have questions executing an e2e scenario. If that works fine, it should be ok.

tested and it works well.

Issue #13340, opened 2020-05-05
Azure REST API for SQL not working for CLI but works for Post Man


rrmistry:
## Describe the bug

**Command Name**
`az rest`

**Errors:**
```
Bad Request({"error":{"details":[{"code":"InvalidResourceIdSegment","message":"","target":"parameters.properties.weeklyRetention"}],"code":"InvalidResourceIdSegment","message":"The 'parameters.properties.weeklyRetention' segment in the url is invalid."}})
```

## To Reproduce:
Steps to reproduce the behavior. Note that argument values have been redacted, as they may contain sensitive information.

```powershell
# Language : PowerShell
$server = "YOUR_AZURE_SQL_SERVER"
$database = "YOUR_AZURE_SQL_DATABASE"
$weeklyRetention = "P2W"
$policyName = "default"
$apiVersion = "2019-06-01-preview"

$srvs = az sql server list --query "[?name=='$server']" | ConvertFrom-Json
$srv = $srvs[0]
$dbs = az sql db list --resource-group $srv.resourceGroup --server $srv.name --query "[?name=='$database']" | ConvertFrom-Json
$db = $dbs[0]

# This works 👍 
az rest --method "GET" --uri "$($srv.id)/databases/$($db.name)/backupLongTermRetentionPolicies/$($policyName)?api-version=$apiVersion" | ConvertFrom-Json

# This Fails 👎 
az rest --method "PUT" --uri "$($srv.id)/databases/$($db.name)/backupLongTermRetentionPolicies/$($policyName)?api-version=$apiVersion" --header "Content-Type=application/json" --body "{ ""properties"": { ""weeklyRetention"": ""$weeklyRetention"" } }" --debug
```

## Expected Behavior

Should update with `202` response:

```json
{
    "operation": "UpsertDatabaseBackupArchivalPolicyV2",
    "startTime": "2020-05-05T15:04:16.397Z"
}
```

## Environment Summary
```
Windows-10-10.0.14393-SP0
Python 3.6.6

azure-cli 2.2.0

```
## Additional Context

Also note that setting `backupShortTermRetentionPolicies` on the same resource works without a problem:

```
az rest --method "PUT" --uri "$($srv.id)/databases/$($db.name)/backupShortTermRetentionPolicies/$($policyName)?api-version=$apiVersion" --header "Content-Type=application/json" --body "{ ""properties"": { ""retentionDays"": 35 } }"
```

So I know it's not an issue with the headers or how I pass the JSON body.

[...]

PR #13249, opened 2020-04-29
{Docs} Increase visibility of doc/use_cli_effectively.md

jiasli:
## Description

- Increase visibility of `doc/use_cli_effectively.md`, thus avoiding issues like 
    - Output formatting: #8579
    - Command chaining: #6821
    - Quoting issues: #13208, #13152, #11008, #12970, #11641, #11668, #11003, #10997, #8827, #8070, #7054, #10370, #9228, #8630, #9742
    - Proxy/SSL issues: #13032, #12911, #12310, #12122, #11436, #11356, #11166, #11113, #11069, #11068, #10921, #10860, #10463, #10272, #10236, #10079, #11654, #12566, #11806, #9232, #9030 , #8734, #5841
    - Generic update arguments: #10287, #10297, #10568, #10073, #1582
    - Storage Commands: #12242, #9852
- Add docs for `az rest`
- Add quoting issues for `--query`
- Add quoting issue troubleshooting guide

In order to be consistent with public docs like [Create an Azure service principal with the Azure CLI](https://docs.microsoft.com/en-us/cli/azure/create-an-azure-service-principal-azure-cli?view=azure-cli-latest),

- Non-leading upper case letters in headings are converted to lower case letters
    ```diff
    - Generic Resource Commands
    + Generic resource commands
    ```
- Gerunds (verb + -ing) are converted to their base forms
    ```diff
    - Working behind a proxy
    + Work behind a proxy
    ```

## Testing Guide

View the rendered file by clicking **View file**.

image

# Remark

⚠ Renaming the titles may break anchors in previous issues, without any workarounds or redirection, but this is inevitable in order to be more friendly to future readers.

Issue #13208, opened 2020-04-27
Cloud Shell - Azure CLI: passing headers not working but does with version 2.4.0


Krumelur:
## Describe the bug

I'm trying to use `az rest` to access a CosmosDB's REST API. This works fine locally with Azure CLI 2.4.0 installed on macOS. The problem is caused by the `--headers` parameter.

According to the docs (https://docs.microsoft.com/en-us/cli/azure/reference-index?view=azure-cli-latest#az-rest), multiple header values can be specified as a blank-separated string. This does not work. Only the first header is sent. This is why I'm using the JSON format which is optionally supported. This works fine when running locally.

```
headers="{\"x-ms-date\": \"$now\", \"x-ms-version\": \"2018-12-31\", \"x-ms-documentdb-isquery\": \"true\", \"Content-Type\": \"application/query+json\", \"Authorization\": \"$authString\"}"
echo "Headers:" $headers

# Make the API call by combining base URL and resource type.
url="$baseUrl$resourceType"
echo "URL: $url"
az rest --verbose -m get -u $url --headers $headers
```

When running the same command in Cloud Shell in Azure Portal (v 2.2.0), the error below occurs.

**Command Name**
`az rest`

**Errors:**
```
not enough values to unpack (expected 2, got 1)
Traceback (most recent call last):
azure/cli/core/util.py, ln 280, in shell_safe_json_parse
    return json.loads(json_or_dict_string)
lib/python3.6/json/__init__.py, ln 354, in loads
    return _default_decoder.decode(s)
lib/python3.6/json/decoder.py, ln 339, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
lib/python3.6/json/decoder.py, ln 357, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 14 (char 13)

...

azure/cli/core/util.py, ln 552, in send_raw_request
    key, value = s.split('=', 1)
ValueError: not enough values to unpack (expected 2, got 1)
```

## To Reproduce:
Steps to reproduce the behavior. Note that argument values have been redacted, as they may contain sensitive information.

- _Put any pre-requisite steps here..._
- `az rest --verbose -m {} -u {} --headers {} {} {} {} {} {} {} {} {} {} {} {} {} {} {}`

## Expected Behavior
1. JSON format for headers should work
2. Space separated headers should work

## Environment Summary
```
[...]

Issue #13152, opened 2020-04-24
az version --query "azure-cli" should return version number and not throw error


silentbobbert:
## Describe the bug

When I run ```az version --query "azure-cli"``` it should return the version number: ```"2.4.0"```

**Command Name**
`az version`

**Errors:**
```
az version: error: argument --query: invalid jmespath_type value: "azure-cli"'
```

## To Reproduce:
Steps to reproduce the behavior. Note that argument values have been redacted, as they may contain sensitive information.

- `az version --query "azure-cli"`

## Expected Behavior

According to the tutorial pages you suggest people use to learn about [JMESPath](https://jmespath.org/tutorial.html), it should work like this:

![image](https://user-images.githubusercontent.com/3257988/80186315-b8b43e00-8605-11ea-8950-8e00aaf5aed5.png)

I copied the output from a plain `az version` execution, and selected the field I wanted manually in the online tool.

I would expect `az version --query "azure-cli"` to perform the same logical function and return `2.4.0`

## Environment Summary
```
Windows-10-10.0.17134-SP0
Python 3.6.6
Installer: MSI

azure-cli 2.4.0

```

yonzhan:
add to S170

jiasli:
This is the design of JMESPath that `-` is parsed as minus. See https://github.com/jmespath/jmespath.py/issues/109#issuecomment-240315651. 

So in the query `azure-cli` needs to be quoted as `"azure-cli"` as you did in the screenshot:

![image](https://user-images.githubusercontent.com/3257988/80186315-b8b43e00-8605-11ea-8950-8e00aaf5aed5.png)

**However, since the terminal interprets double quotes, the command needs to be escaped.**

In Windows CMD:

 ```cmd
>az version --query "\"azure-cli\""
"2.4.0"
```

In Windows PowerShell (Actually `\` shouldn't be necessary, but we need them here due to https://github.com/PowerShell/PowerShell/issues/1995): 

```pwsh
> az version --query '\"azure-cli\"'
"2.4.0"
> az version --query "\`"azure-cli\`""
"2.4.0"
```

In Bash:

```sh
$ az version --query "\"azure-cli\""
"2.4.0"
$ az version --query '"azure-cli"'
"2.4.0"
```

silentbobbert:
[...]

PR #13419, opened 2020-05-09
{Doc} Add quoting-issues-with-powershell.md

jiasli:
## Description
This PR serves as an extension to https://github.com/Azure/azure-cli/pull/13249. 

PowerShell quoting issues are prevailing in Azure CLI repo: #13340, #13208, #13152, #11008, #12970, #11641, #11668, #11003, #10997, #8827, #8070, #7054, #10370, #9228, #8630, #9742

Since PowerShell quoting issues are not related to Azure CLI but a known bug, separate it to a new doc file for easy management.

### DIFF
--- a/doc/quoting-issues-with-powershell.md
+++ b/doc/quoting-issues-with-powershell.md
@@ -63,6 +63,15 @@ az: 'a' is not in the 'az' command group.
 operable program or batch file.
 ```
 
+This is what `cmd.exe` or Windows system sees:
+
+```cmd
+>az a&b --debug
+az: 'a' is not in the 'az' command group. 
+'b' is not recognized as an internal or external command,
+operable program or batch file.
+```
+
 To solve it:
 
 ```powershell
@@ -88,7 +97,18 @@ This typically happens when passing a JSON to `az`. This is because double quote
 # Wrong! Note that the double quotes (") are lost
 > python.exe -c "import sys; print(sys.argv)" '{"key": "value"}'
 ['-c', '{key: value}']
+```
 
+This is what `cmd.exe` or Windows system sees:
+
+```cmd
+>python.exe -c "import sys; print(sys.argv)" "{"key": "value"}"
+['-c', '{key: value}']
+```
+
+To solve it:
+
+```powershell
 # Escape double quotes (") with backward-slashes (\) as required by Command Prompt,
 # then quote the string with single quotes (') as required by PowerShell
 > python.exe -c "import sys; print(sys.argv)" '{\"key\": \"value\"}'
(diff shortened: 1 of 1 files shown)


yonzhan:
add to S169

qwordy:
Use cli effectively.md has similar instructions. Will you move them here?

jiasli:
> Use cli effectively.md has similar instructions. Will you move them here?

Already removed them in #13249.

qwordy:
Can we put the doc at https://docs.microsoft.com/en-us/cli/azure/?view=azure-cli-latest?
Many users don't know what github is. They cannot find this doc and they do not trust it.

doc/quoting-issues-with-powershell.md
```diff
@@ -0,0 +1,165 @@
...
+> az "a&b" --debug
+az: 'a' is not in the 'az' command group.
+'b' is not recognized as an internal or external command,
+operable program or batch file.
+```
+
+This is what `cmd.exe` or Windows system sees:
+
+```cmd
+>az a&b --debug
```
qwordy, line 71:
```suggestion
> az a&b --debug
```

jiasli: This how CMD is. CMD doesn't have space in between.

jiasli:
> Can we put the doc at https://docs.microsoft.com/en-us/cli/azure/?view=azure-cli-latest?
> Many users don't know what github is. They cannot find this doc and they do not trust it.

This is exactly I have been talking about. We need to work with doc team to duplicate this to public doc as well. + @dbradish-microsoft

doc/quoting-issues-with-powershell.md
qwordy: Interesting

doc/quoting-issues-with-powershell.md
```diff
@@ -0,0 +1,165 @@
...
+operable program or batch file.
+```
+
+To solve it:
+
+```powershell
+# Double quotes are parsed by Command Prompt to treat ampersand (&) as a literal character
+> az '"a&b"' --debug
+Command arguments: ['a&b', '--debug']
+
```
fengzhou-msft, line 83: Based on what I learned from the following section, these also works:
>az "\`"a&b\`"" --debug
>az """a&b""" --debug

Basically, the quotes will first be parsed by Powershell, then be parsed by Command Prompt. If after 2 parses, the double quotes still exist, then it works. This is an alternative to the the `stop-parsing symbol` way. If you could summarize it and put under the Workaround section below `stop-parsing symbol`, it would be very clear to the readers.

jiasli: Thanks for the great suggestion. Add to [here](https://github.com/Azure/azure-cli/pull/13419/files#diff-9d47dca9ba5356e75bb84c36621bd267R21) and [here](https://github.com/Azure/azure-cli/pull/13419/files#diff-9d47dca9ba5356e75bb84c36621bd267R87-R93).

Issue #8941, opened 2019-04-01
Support az feature unregister


jluk:
**Is your feature request related to a problem? Please describe.**
There are feature previews that impact the ongoing behavior of CLI due to using new versions of APIs.

**Describe the solution you'd like**
I want one command to revert my API version to one that is supported and not on preview terms.

**Describe alternatives you've considered**
API calls can be made but that is not a good experience after enabling something.

tjprescott:
It seems like there's a PS cmdlet for this, but there doesn't seem to be an ARM endpoint for this functionality. @yugangw-msft do you know a good contact on the ARM team to confirm?

yugangw-msft:
I have not found the PS cmdlet doing that. Under the source [folder](https://github.com/Azure/azure-powershell/tree/master/src/Resources/ResourceManager/Implementation/Features) only has the `register` cmdlet.
@vladimirjoanovic, can you please chip in the api details?

jluk:
Talking to @yangl900 looks like a post can be made similar to: `armclient post /subscriptions/<somesubscription>/providers/Microsoft.Features/providers/Microsoft.Xrm/features/uxdevelopment/unregister?api-version=2015-12-01`

yugangw-msft:
Maybe improve the swagger api [here](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/resources/resource-manager/Microsoft.Features/stable/2015-12-01/features.json) to unblock SDK/POSH/CLI?

jluk:
@vladimirjoanovic can you add the proper owner to this discussion?

jluk:
Opened an issue on swagger as well https://github.com/Azure/azure-rest-api-specs/issues/5607

zparnold:
Has this gotten any love?

yonzhan:
ARM service attention

ifeoluwaokunoren:
We're actively working on providing this command - 
PR link: https://github.com/Azure/azure-cli/pull/13487
I'd add update this thread once the PR is merged.
In the meanwhile if you're blocked, you can directly unregister a subscription from a feature using the REST API: https://github.com/Azure/azure-rest-api-specs/blob/38388fadb2a31238262c96520a96f413ff7c7311/specification/resources/resource-manager/Microsoft.Features/stable/2015-12-01/features.json#L230

ifeoluwaokunoren:
[...]

PR #13487, opened 2020-05-13
[ARM] Add az feature unregister api

diwudd:
Add az feature unregister api
Issue Raised: https://github.com/Azure/azure-cli/issues/8941

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/resource/tests/latest/test_resource.py
+++ b/src/azure-cli/azure/cli/command_modules/resource/tests/latest/test_resource.py
@@ -1129,6 +1129,10 @@ class FeatureScenarioTest(ScenarioTest):
 
         # Once a feature goes GA , it will be removed from the feature list. Once that happens, use other ones to test
         self.cmd('feature show --namespace Microsoft.Network -n AllowLBPreview')
+    
+    @AllowLargeResponse(8192)
+    def test_feature_unregister(self):
+        self.cmd('feature unregister --namespace Microsoft.Network --name AllowLBPreview', checks=self.check('properties.state', 'Unregistered'))
 
 
 class PolicyScenarioTest(ScenarioTest):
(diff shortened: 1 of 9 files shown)


yonzhan:
add to S170

yonzhan:
add to S171

src/azure-cli/azure/cli/command_modules/resource/tests/latest/test_resource.py
```diff
@@ -1130,6 +1130,10 @@ def test_feature_list(self):
         # Once a feature goes GA , it will be removed from the feature list. Once that happens, use other ones to test
         self.cmd('feature show --namespace Microsoft.Network -n AllowLBPreview')
 
+    @AllowLargeResponse(8192)
+    def test_feature_unregister(self):
+        self.cmd('feature unregister --namespace Microsoft.Network --name AllowLBPreview', checks=self.check('properties.state', 'Unregistered'))
```
haroldrandom, line 1135: Should the test need pre-setup step like register Network namespace before running this?

diwudd: No, it does not need pre-setup

diwudd:
Thanks Jianhui.

@Juliehzl,   can you please take another look?

Thanks,

Di

ifeoluwaokunoren:
@Juliehzl can you please confirm your comments have been addressed, and approve so we can merge?
Thanks.

src/azure-cli/azure/cli/command_modules/resource/tests/latest/test_resource.py
Juliehzl, line 1135: Can we add some check for command response? Or it is possible to utilize `az feature show` to check state after unregistering?

diwudd: I might miss something. It does check the response with "checks=self.check('properties.state', 'Unregistered')". Anything else you think we should check here?

Juliehzl: yes. you can check the state in response.

PR #13538, opened 2020-05-17
{Telemetry} Log exception

fengzhou-msft:
**Description**  

Add error of command result to telemetry when it's not added inside the module.
Example data:
```
{"name": "azurecli/fault",
 "properties": {
                "Reserved.DataModel.EntityType": "Fault",
                "Reserved.DataModel.Fault.Description": "",
                "Reserved.DataModel.Correlation.1": "7a3155c0-7237-4ae2-8500-f547e99713d5,UserTask,",
                "Reserved.DataModel.Fault.TypeString": "CLIError",
                "Reserved.DataModel.Fault.Exception.Message": "Please provide subscription id or unique name.",
                "Reserved.DataModel.Fault.Exception.StackTrace": "[]",
                ...}
}
```

**Testing Guide**  

**History Notes**  

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/telemetry.py
+++ b/src/azure-cli-core/azure/cli/core/telemetry.py
@@ -203,6 +203,10 @@ class TelemetrySession(object):  # pylint: disable=too-many-instance-attributes
 _session = TelemetrySession()
 
 
+def has_exceptions():
+    return len(_session.exceptions) > 0
+
+
 def _user_agrees_to_telemetry(func):
     @wraps(func)
     def _wrapper(*args, **kwargs):
--- a/src/azure-cli/azure/cli/__main__.py
+++ b/src/azure-cli/azure/cli/__main__.py
@@ -40,6 +40,9 @@ try:
 
     exit_code = cli_main(az_cli, sys.argv[1:])
 
+    if az_cli.result.error is not None and not telemetry.has_exceptions():
+        telemetry.set_exception(az_cli.result.error, fault_type='')
+
     if exit_code and exit_code != 0:
         telemetry.set_failure()
     else:
--- a/src/azure-cli/azure/cli/__main__.py
+++ b/src/azure-cli/azure/cli/__main__.py
@@ -40,10 +40,9 @@ try:
 
     exit_code = cli_main(az_cli, sys.argv[1:])
 
-    if az_cli.result.error is not None and not telemetry.has_exceptions():
-        telemetry.set_exception(az_cli.result.error, fault_type='')
-
     if exit_code and exit_code != 0:
+        if az_cli.result.error is not None and not telemetry.has_exceptions():
+            telemetry.set_exception(az_cli.result.error, fault_type='')
         telemetry.set_failure()
     else:
         telemetry.set_success()
--- a/src/azure-cli-core/azure/cli/core/parser.py
+++ b/src/azure-cli-core/azure/cli/core/parser.py
@@ -143,7 +143,7 @@ class AzCliCommandParser(CLICommandParser):
                 _parser=command_parser)
 
     def validation_error(self, message):
-        telemetry.set_user_fault('validation error')
+        telemetry.set_user_fault('validation error: {}'.format(message))
         return super(AzCliCommandParser, self).error(message)
 
     def error(self, message):


src/azure-cli-core/azure/cli/core/telemetry.py
```diff
@@ -203,6 +203,10 @@ def product_version(self):
 _session = TelemetrySession()
 
 
+def has_exceptions():
+    return len(_session.exceptions) > 0
```
yungezz, line 207: this could be serivce error or client error, right?

src/azure-cli/azure/cli/__main__.py
```diff
@@ -40,6 +40,9 @@ def cli_main(cli, args):
 
     exit_code = cli_main(az_cli, sys.argv[1:])
 
+    if az_cli.result.error is not None and not telemetry.has_exceptions():
```
yungezz, line 43: when has_exceptions(), the exceptions will be logged also, but not to telemetry exception right?

yonzhan:
Telemetry

src/azure-cli-core/azure/cli/core/telemetry.py
fengzhou-msft: yes, could be any error.

src/azure-cli/azure/cli/__main__.py
fengzhou-msft: when has_exceptions(), it means some raised exception is manually added to telemetry session already, so here no need to add it again.

src/azure-cli-core/azure/cli/core/telemetry.py
haroldrandom: What type of this `exceptions`? What's the content of the `exceptions`?

src/azure-cli/azure/cli/__main__.py
```diff
@@ -40,6 +40,9 @@ def cli_main(cli, args):
...
+        telemetry.set_exception(az_cli.result.error, fault_type='')
+
     if exit_code and exit_code != 0:
         telemetry.set_failure()
```
haroldrandom, line 47: Can we resuse this part of code?

src/azure-cli-core/azure/cli/core/telemetry.py
arrownj: `has_exceptions` is defined outside `TelemetrySession` ? How can below `telemetry.has_exceptions` work ?

fengzhou-msft: `_session` is a shared instance of `TelemetrySession` in `telemetry.py`, so you can do `from azure.cli.core import telemetry` and just call `telemetry.has_exceptions`

fengzhou-msft: > What type of this `exceptions`? What's the content of the `exceptions`?

This contains all sorts of exceptions that are added to the result of a command. It could be CLIError, CloudError or others. It usually contains an error message, please see the PR description for an example.

src/azure-cli/azure/cli/__main__.py
fengzhou-msft: Yes. You can call `telemetry.set_exception()` with a self-defined fault_type in the module code. Here we check that if you have an error but have't added it to telemetry, then we add it when the command finishes executing, but we cannot decide the fault_type.

haroldrandom: I was wonder why `telemetry.has_exceptions` and `az_cli.result.exit_code != 0` occur in different timming.

As far as I understand, when `az_cli.result.error is not None` which also means `az_cli.result.exit_code != 0`. 
At this time, `telmetry.has_exceptions` and `telemetry.set_failure` should put together under this condition.

Which means, should put `telemetry.set_exception(az_cli.result.error, fault_type='')` under the if statement `if exit_code and exit_code != 0:` as `telemetry.set_failure()`.

fengzhou-msft: I moved the new code under `if exit_code and exit_code != 0:` in [knack](https://github.com/microsoft/knack/blob/master/knack/util.py#L16-L17), but since `exit_code` and `error` are set differently, I would keep checking the error to be robust.

Issue #13159, opened 2020-04-24
Validation failures give error location based on minified arm template


Tyriar:
Repro:

1. Create `arm-test.json` with this content
```json
{
    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
    "contentVersion": "1.0.0.0",
    "parameters": {
        "appservice_name": {
            "defaultValue": "testing",
            "type": "string"
        },
        "location": {
            "defaultValue": "West US",
            "type": "string"
        }
    },
    "resources": [
        {
            "type": "Microsoft.Web/serverfarms",
            "apiVersion": "2018-02-01",
            "name": "[parameters('appservice_name')]",
            "location": "[parameters('location')]",
            "sku": {
                "name": "P1v2",
                "tier": "PremiumV2",
                "size": "P1v2",
                "family": "Pv2",
                "capacity": 1
            },
            "kind": "linux",
            "properties": {
                "perSiteScaling": false,
                "maximumElasticWorkerCount": 1,
                "isSpot": false,
                "reserved": true,
                "isXenon": false,
                "hyperV": false,
                "targetWorkerCount": 0,
                "targetWorkerSizeId": 0
            }
        },
    ]
}
```
2. Set $vSubscription and $vResourceGroup
3. Run `az deployment group create --subscription $vSubscription -g $vResourceGroup--template-file arm-test.json`

Result:

![image](https://user-images.githubusercontent.com/2193314/80231640-7ccea800-8608-11ea-9105-6891165f1558.png)

So I think there is a bug here and I also have a suggestion. The error says line 1 column 679 clearly the arm template is not that long, it's complaining about some minified version of it being validated.

If that gets fixed, the suggestion is to print this with the path to the file, that way using VS Code's integrated terminal you can click the error and go directly to the problem line, for example:

![image](https://user-images.githubusercontent.com/2193314/80232032-15fdbe80-8609-11ea-994d-2ce19fe562d2.png)

Compilers/linters normally do this which makes it convenient to fix the problem.

yonzhan:
add to S170

zhoxing-ms:
@Tyriar Thank you for your feedback and suggestions~ As you said, this problem is caused by the validation after the ARM template is compressed and comments are deleted.

[...]

PR #13561, opened 2020-05-19
[ARM] az deployment: Fix the problem of incorrect message of JSON after removing comments and compressing (#13159)

zhoxing-ms:
Issue: #13159

**Description**  
When the JSON format of the template file or parameter file is not correct, the following error will be reported:
```pwsh
Expecting value: line 1 column 337 (char 336)
```
There are two problems with it:
1. Because the processing of removing comments and compression in the method `_remove_comments_from_json` will lead to misplacement of error location in error message, so the error message should be wrapped.
2. It does not prompt the specific error file. It may be a template file or a parameter file.

**Testing Guide**  
Deploy or verify the template file or parameter file with incorrect JSON format.

**History Notes**  

---


yonzhan:
add to S170

PR #13572, opened 2020-05-19
{Extensions} tests: do not import from absolute path

bluca:
Test files/modules are not installed. Import from the current directory, or they will
fail when running in detached mode (eg: autopkgtest in Debian/Ubuntu).

### DIFF
--- a/src/azure-cli-core/azure/cli/core/extension/tests/latest/test_wheel_type_extension.py
+++ b/src/azure-cli-core/azure/cli/core/extension/tests/latest/test_wheel_type_extension.py
@@ -11,7 +11,10 @@ import shutil
 import unittest
 
 from azure.cli.core.extension import WheelExtension
-from azure.cli.core.extension.tests.latest import ExtensionTypeTestMixin, get_test_data_file
+try:
+    from azure.cli.core.extension.tests.latest import ExtensionTypeTestMixin, get_test_data_file
+except ImportError:
+    from . import ExtensionTypeTestMixin, get_test_data_file
 
 
 class TestWheelTypeExtensionMetadata(ExtensionTypeTestMixin):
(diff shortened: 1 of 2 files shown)


yonzhan:
add to S170

src/azure-cli-core/azure/cli/core/extension/tests/latest/test_wheel_type_extension.py
```diff
@@ -11,7 +11,7 @@
 import unittest
 
 from azure.cli.core.extension import WheelExtension
-from azure.cli.core.extension.tests.latest import ExtensionTypeTestMixin, get_test_data_file
+from . import ExtensionTypeTestMixin, get_test_data_file
```
jiasli, line 14: I don't quite understanding. Isn't `azure.cli.core.extension.tests.latest` part of `azure.cli.core.extension`? If `azure.cli.core.extension.tests.latest` is not found, how can `azure.cli.core.extension` be found? 

As stated in [PEP 8](https://www.python.org/dev/peps/pep-0008/#imports):

> Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages) if the import system is incorrectly configured.

bluca: No, the tests are not part of the module and are not installed normally. And they shouldn't be - it's a waste of space.

bluca: autopkgtest installs the packages (which means the azure.cli modules will be in /usr/lib/python3/dist-packages/) and runs the unit tests from the sources in a working path (eg: /tmp/autopkgtest/somerandomchar)

bluca: How about this: I pushed an update that will try the absolute import first, and then fall back to import from current directory only if that fails. Does that sound better?

jiasli: Ah, I see. So you are installing Azure CLI under `/usr/lib/python3/dist-packages/`, but running tests from source code in a separate location.

PR #13616, opened 2020-05-21
{Find} az find: Change telemetry settings

mirdaki:
**Description**  

We wanted to update the find module to use a more relevant piece of telemetry and respect the user set telemetry flag.

No user facing changed. If the flag is false, the service is sent empty strings as telemetry. 

**Testing Guide**  

No new commands or tests. Local testing was done to ensure this feature works.

---


yonzhan:
add to S170

src/azure-cli/azure/cli/command_modules/find/custom.py
```diff
@@ -95,16 +95,20 @@ def should_enable_styling():
 
 
 def call_aladdin_service(query):
-    session_id = telemetry_core._session._get_base_properties()['Reserved.SessionId']  # pylint: disable=protected-access
+    correlation_id = telemetry_core._session.correlation_id  # pylint: disable=protected-access
```
yungezz, line 98: why put correlation_id in sessionId?

mirdaki: We're trying to switch to using the `correlation_id`, since it looks more appropriate for our use case (though the service isn't yet setup to expect it). 

Though it looks like the `correlation_id` isn't being used in recent telemetry from the CLI, so we are trying to figure that out now. The PR should stay pending until we have a resolution on that.

PR #13617, opened 2020-05-21
[ACR] add support for region endpoint disable / routing disable

adewaleo:
**Description**  

Test draft acr sdk. Test region endpoint enabled feature.

**Testing Guide**  

`az acr replication create -h`
`az acr replication update -h`

`az acr replication create --region-endpoint-enabled true/false`

**History Notes**  

[Component Name 2] az acr replication [create/update]: Add support for disabling routing to a registry replication via `--region-endpoint-enabled`. The replication's data will continue to be synced with other registries.

---


yonzhan:
ACR

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -156,6 +156,14 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
     with self.argument_context('acr replication create') as c:
         c.argument('replication_name', help='The name of the replication. Default to the location name.', completer=None)
 
+    for scope in ['acr replication create', 'acr replication update']:
+        help_str = "Allow routing to this replication. Requests will not be routed to a disabled replication." \
+                   " Data syncing will continue regardless of the region endpoint status."
+        help_str += ' Default: true.' if 'create' in scope else ''  # suffix help with default if command is for create
+
+        with self.argument_context(scope) as c:
+            c.argument('region_endpoint_enabled', arg_type=get_three_state_flag(), help=help_str, is_preview=True)
```
djyou, line 165: Can we actually preview a flag?

src/azure-cli/azure/cli/command_modules/acr/replication.py
```diff
@@ -35,11 +39,37 @@ def acr_replication_create(cmd,
         raise CLIError('Replication could not be created in the same location as the registry.')
 
     from msrest.exceptions import ValidationError
+    ReplicationType = cmd.get_models('Replication')
+
+    # version below adds region_endpoint_enabled and a breaking change where some create params are grouped into a model
+    grouped_params_create_api_version = "2019-12-01-preview"
```
djyou, line 45: What's preventing us from updating the api-version for all commands?

src/azure-cli/azure/cli/command_modules/acr/replication.py
```diff
@@ -26,6 +29,7 @@ def acr_replication_create(cmd,
                            registry_name,
                            resource_group_name=None,
                            replication_name=None,
+                           region_endpoint_enabled=None,
```
djyou, line 32: region_endpoint_enabled = True

src/azure-cli/azure/cli/command_modules/acr/_params.py
adewaleo: Yes we can, it looks like this: 

image

src/azure-cli/azure/cli/command_modules/acr/replication.py
adewaleo: So for some reason, their CI might use a different profile which uses an older api version. Versions older than this version have a different signature for create.

This is why I check for the api version.

src/azure-cli/azure/cli/command_modules/acr/replication.py
adewaleo: Would have done this but for some reason this causes the following help output. Not sure why the CLI renders the boolean True as a number.

image

djyou: Can you check how similar flags such as admin enabled is implemented?

src/azure-cli/azure/cli/command_modules/acr/_params.py
adewaleo: @djyou admin_enabled defaults to false, however, false defaults aren't surfaced in help text. public access enabled's default is left to the sdk if None. Sdk defaults to None.

Our sdk does default to True for --region-endpoint-enabled. I added the logic I had to ensure it is always consistent in behavior. But I guess that's overkill. Should I leave default to sdk but document that region endpoint will always be true (assuming it would be breaking change to change this)

See here for acr_create defaults: https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/acr/custom.py#L35-L47

src/azure-cli/azure/cli/command_modules/acr/replication.py
adewaleo: @djyou I thought I  responded, but just seeing I didn't Basically admin_enabled defaults to false in the signature (this isn't exposed automatically in CLI). public_network_enabled defaults to sdk default which is allowed/enabled. That being said we write the default in help text.

I suppose I can say default is true and leave to sdk to set default (sdk will set field to true if None is passed). I'll remove the defaulting logic I added, was paranoid.

adewaleo:
@qwordy @MyronFanQiu @fengzhou-msft, I've gotten ACR sign-off, please can one of you review as well? Thanks!

src/azure-cli/azure/cli/command_modules/acr/_params.py
fengzhou-msft: You can add `default='true'` or  `default='false'` in `c.argument()`.

src/azure-cli/setup.py
```diff
@@ -74,7 +74,7 @@
     'azure-mgmt-compute~=12.0',
     'azure-mgmt-consumption~=2.0',
     'azure-mgmt-containerinstance~=1.4',
-    'azure-mgmt-containerregistry==3.0.0rc12',
+    'azure-mgmt-containerregistry~=3.0.0rc13',
```
yungezz, line 77: why use `~=`, could  you pls pin exact version again? thanks.

adewaleo: @yungezz I did this because other dependencies aren't pinned to an exact version. Do you want me to do something different?

yungezz: I remembered we had the problem of breaking change in patch version from 9.0 -> 11.0.  that's the reason we pinned the exact version at that time.  What's your thinking on this.

yugangw-msft: @adewaleo let us just use `==` to pin it

adewaleo: @yungezz @yugangw-msft sounds good! Will pin.

Issue #13625, opened 2020-05-22
Issue with special characters in SP credentials in latest version of Az CLI


AdamSharif-MSFT:
**Describe the bug**

When created a new SP (or resetting credentials) n Azure CLI, special characters appear such as !, $, and ' which cause issues with variables/commands in Linux:

Example (obfuscated):

```
> az ad sp create-for-rbac
Creating a role assignment under the scope of "/subscriptions/**[redacted]**"
  Retrying role assignment creation: 1/36
{
  "appId": "dummy-app-id",
  "displayName": "**[redacted]**",
  "name": "**[redacted]**",
  "password": "**[redacted]**$9=:p;**[redacted]**(<5\\**[redacted]**@<a3r/N!**[redacted]**",
  "tenant": "**[redacted]**"
}

> pass="[redacted]" --> value from above
zsh: event not found: >I
```

**To Reproduce**

1. `az ad sp create-for-rbac --skip-assignment` (or `az ad sp credential reset`)
2. set variable as password value
3. Try to use variable

**Expected behavior**

Secret value should contain only character which are permitted and usable in scripts/variables

**Environment summary**

```
> az version
{
  "azure-cli": "2.6.0",
  "azure-cli-command-modules-nspkg": "2.0.3",
  "azure-cli-core": "2.6.0",
  "azure-cli-nspkg": "3.0.4",
  "azure-cli-telemetry": "1.0.4",
  "extensions": {
    "aks-preview": "0.4.44",
    "azure-firewall": "0.3.1"
  }
}
```

WSL & Bash/Zsh, Ubuntu VM with Bash

**Additional context**

Testing a previous version of CLI such as 2.4, we see that secrets are generated as GUID format with only - for special characters. Same case for Azure PowerShell module.

jiasli:
This was introduced in https://github.com/Azure/azure-cli/pull/13357. @qianwens, please help take a look.

jiasli:
Special characters like `'`, `"`, `!` and `$` will cause issues in Linux Bash even though the password is quoted.

```sh
$ echo "some"password"
> ^C
$ echo 'some'password'
> ^C
$ echo "some!password"
-bash: !password: event not found
$ echo "some$password"
some
```

`!` is treated as [History Expansion](https://www.gnu.org/software/bash/manual/html_node/History-Interaction.html). `$` is treated as [Parameter Expansion](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion).

[...]

PR #13357, opened 2020-05-06
[RBAC] `az ad sp credential reset`: fix weak credential generation

qianwens:
**Description**  

Password autogenerated by `az ad sp credential reset` is a Guid which is not secure.

**Testing Guide**  

az ad sp credential reset -n test
Expected result: password should be a string with at least one special character
**History Notes**  

[Component Name 1] BREAKING CHANGE: az command a: Make some customer-facing breaking change.  
[Component Name 2] az command b: Add some customer-facing feature.

---

This checklist is used to make sure that common guidelines for a pull request are followed.

- [ ] The PR title and description has followed the guideline in [Submitting Pull Requests](https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules#submitting-pull-requests).

- [ ] I adhere to the [Command Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md).

Issue #12946, opened 2020-04-10
Microsoft Graph API Support


jiasli:
[**Azure Active Directory Graph API**](https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-graph-api) has been deprecated:

> We strongly recommend that you use [Microsoft Graph](https://developer.microsoft.com/graph) instead of Azure AD Graph API to access Azure Active Directory (Azure AD) resources. Our development efforts are now concentrated on Microsoft Graph and no further enhancements are planned for Azure AD Graph API. There are a very limited number of scenarios for which Azure AD Graph API might still be appropriate; for more information, see the [Microsoft Graph or the Azure AD Graph](https://developer.microsoft.com/office/blogs/microsoft-graph-or-azure-ad-graph/) blog post and [Migrate Azure AD Graph apps to Microsoft Graph](https://docs.microsoft.com/graph/migrate-azure-ad-graph-overview).

We need to switch Azure CLI to use [**Microsoft Graph API**](https://docs.microsoft.com/en-us/graph/api/overview?toc=./ref/toc.json&view=graph-rest-1.0) instead.

yonzhan:
Microsoft Graph API Support

jiasli:
## Call Microsoft Graph API with `az rest`

As [**Microsoft Graph REST API v1.0**](https://docs.microsoft.com/en-us/graph/api/overview?view=graph-rest-1.0) is now GA, we can call it directly with `az rest` to achieve the same effect as `az ad` commands, including all latest features from Microsoft Graph. 

## Examples

### Update `redirectUris` for an [Application](https://docs.microsoft.com/en-us/graph/api/resources/application?view=graph-rest-1.0)

Originally posted at https://github.com/Azure/azure-cli/issues/9501#issuecomment-610979930. We call the [Update application](https://docs.microsoft.com/en-us/graph/api/application-update?view=graph-rest-1.0&tabs=http) API. The GUID part in the following URLs are the object ID of the application. 

```sh
# Get the application
az rest --method GET --uri 'https://graph.microsoft.com/v1.0/applications/b4e4d2ab-e2cb-45d5-a31a-98eb3f364001'

# Update `redirectUris` for `web` property
[...]

Issue #12561, opened 2020-03-12
Unable to append password credential to ServicePrincipal


x10an14:
## Describe the bug
Unable to append a new password credential to a Service Principal (read: `az ad app` Azure Application).

**Command Name**
`az ad sp credential reset`

**Errors:**
```
Update to existing credential with KeyId 'b88fbf58-8ff1-420d-b0b2-640bfd5262b9' is not allowed.
```

## To Reproduce:
Steps to reproduce the behavior. Note that argument values have been redacted, as they may contain sensitive information.

- _Put any pre-requisite steps here..._
- `az ad sp credential reset --append --name 1fabc626-4309-49e4-bb47-ec9e67251a4a --debug --query "password" -o tsv`

<details>
<summary>Hopefully scrubbed-for-sensitive-details debug output</summary>

```
[2020-03-12 09:49:09] 2 x10an14@x10-desktop:~
-> $ az ad sp credential reset --append --name 1fabc626-4309-49e4-bb47-ec9e67251a4a --debug
Command arguments: ['ad', 'sp', 'credential', 'reset', '--append', '--name', '1fabc626-4309-49e4-bb47-ec9e67251a4a', '--debug']
Event: Cli.PreExecute []
Event: CommandParser.OnGlobalArgumentsCreate [<function CLILogging.on_global_arguments at 0x7f1c16787d08>, <function OutputProducer.on_global_arguments at 0x7f1c166c8ae8>, <function CLIQuery.on_global_arguments at 0x7f1c166f2b70>]
Event: CommandInvoker.OnPreCommandTableCreate []
[...]

PR #13643, opened 2020-05-22
[RBAC] `az ad sp credential reset`: modify credential generation to avoid troublesome special characters

chef-davin:
Signed-off-by: Davin Taddeo <davin@chef.io>

**Description**  

This is an update to the change made in #13357 in the hopes of fixing issue #13625.  I updated the new _random_password() method to avoid special characters that can cause problems in a lot of scripting languages.

**Testing Guide**  

az ad sp credential reset -n test

Expected Result: password/secret should be a string with at least one special character, but whose first character is alpha-numeric and does not contain dangerous special characters like !$#;\/

**History Notes**  

[RBAC] `az ad sp credential reset`: modify credential generation to avoid troublesome special characters

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/role/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/role/custom.py
@@ -1762,7 +1762,7 @@ def _random_password(length):
     # ensure that the password starts with an alphanumeric to avoid issues in shell scripting
     while password_list[0] not in alphanumeric:
         random.SystemRandom().shuffle(password_list)
-    
+
     # turn the password back into a string befor returning it
     password = ''.join(password_list)
     return password
--- a/src/azure-cli/azure/cli/command_modules/role/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/role/custom.py
@@ -1238,7 +1238,7 @@ def _process_service_principal_creds(cli_ctx, years, app_start_date, app_end_dat
 
     if not any((cert, create_cert, password, keyvault)):
         # 1 - Simplest scenario. Use random password
-        return _random_password(32), None, None, None, None
+        return _random_password(34), None, None, None, None
 
     if password:
         # 2 - Password supplied -- no certs
@@ -1745,26 +1745,28 @@ def _gen_guid():
 def _random_password(length):
     import random
     import string
-    safe_punctuation = '@%_-+=:,.(){}[]<>'
+    safe_punctuation = '-_.~'
     random_source = string.ascii_letters + string.digits + safe_punctuation
     alphanumeric = string.ascii_letters + string.digits
 
-    password = random.choice(alphanumeric)
+    # make sure first character is not a punctuation like '--' which will make CLI command break
+    first_character = random.choice(alphanumeric)
+
+    # make sure we have special character in the password
+    password = random.choice(string.ascii_lowercase)
+    password += random.choice(string.ascii_uppercase)
+    password += random.choice(string.digits)
+    password += random.choice(safe_punctuation)
 
     # generate a password of the given length from the options in the random_source variable
-    for _ in range(length):
+    for _ in range(length - 5):
         password += random.choice(random_source)
 
     # turn it into a list for some extra shuffling
     password_list = list(password)
     random.SystemRandom().shuffle(password_list)
 
-    # ensure that the password starts with an alphanumeric to avoid issues in shell scripting
-    while password_list[0] not in alphanumeric:
-        random.SystemRandom().shuffle(password_list)
-
-    # turn the password back into a string befor returning it
-    password = ''.join(password_list)
+    password = first_character + ''.join(password_list)
     return password
 
 
--- a/src/azure-cli/azure/cli/command_modules/role/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/role/custom.py
@@ -1750,17 +1750,17 @@ def _random_password(length):
     alphanumeric = string.ascii_letters + string.digits
 
     # make sure first character is not a punctuation like '--' which will make CLI command break
-    first_character = random.choice(alphanumeric)
+    first_character = random.SystemRandom().choice(alphanumeric)
 
     # make sure we have special character in the password
-    password = random.choice(string.ascii_lowercase)
-    password += random.choice(string.ascii_uppercase)
-    password += random.choice(string.digits)
-    password += random.choice(safe_punctuation)
+    password = random.SystemRandom().choice(string.ascii_lowercase)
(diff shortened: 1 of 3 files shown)


yonzhan:
add to S170

src/azure-cli/azure/cli/command_modules/role/custom.py
```diff
@@ -1745,17 +1745,25 @@ def _gen_guid():
 def _random_password(length):
     import random
     import string
-    random_source = string.ascii_letters + string.digits + string.punctuation
-    password = random.choice(string.ascii_lowercase)
-    password += random.choice(string.ascii_uppercase)
-    password += random.choice(string.digits)
-    password += random.choice(string.punctuation)
+    safe_punctuation = '@%_-+=:,.(){}[]<>'
```
jiasli, line 1748: Thank you very much for submitting the PR. 

I don't think `<>()` are safe when they are not quoted. (Some users new to bash may forget to do that.) `<>` can be parsed as `stdin` `stdout` redirection, and `()` can be parsed as [subshell execution](https://www.gnu.org/software/bash/manual/html_node/Command-Grouping.html):

```
$ echo a>b
# Output is redirected

$ echo (ab)
-bash: syntax error near unexpected token `ab'
```

`%` is also not safe in a Windows Command Prompt: https://ss64.com/nt/syntax-percent.html

We will sync internally with the security team first and try to find out the best solution. 

Possible solutions are:

1. The first character must be a letter or number, then the rest characters can only use punctuation in `-_.~`
2. Show a warning when generating the password, something like "when using the password, it must be quoted to avoid shell interpretation"

jiasli: Microsoft Graph API generates the password by itself, instead of letting the user generate one https://github.com/Azure/azure-cli/issues/12561#issuecomment-605752672:

```
$ az rest -m "POST" -u https://graph.microsoft.com/v1.0/applications/b4e4d2ab-e2cb-45d5-a31a-98eb3f364001/addPassword --headers "Content-Type=application/json" -b '{"passwordCredential":{"displayName":"Password friendly name"}}'
{
  "@odata.context": "https://graph.microsoft.com/v1.0/$metadata#microsoft.graph.passwordCredential",
  "customKeyIdentifier": null,
  "displayName": "Password friendly name",
  "endDateTime": "2022-05-25T03:39:02.2389196Z",
  "hint": "4cc",
  "keyId": "b50bcc87-0f02-4aaf-b5c0-a871189871cb",
  "secretText": "4ccP5-~jc_fWK_4AJ-pzn.y2iNy~51tGnU",
  "startDateTime": "2020-05-25T03:39:02.2389196Z"
}
```

It only uses punctuation marks from `-_.~`. Some more examples:

```
4ccP5-~jc_fWK_4AJ-pzn.y2iNy~51tGnU
Pn~2983OPGQfi0.T.Jaai~v3664xRreKZi
-tw1b2Y46gfnzw0-2~WtfsbTpKy4q9H~8S
--zz5zs._yz-zSC5P~QcR23MNku3mQ2TVg
k-39_0y_83qmHgm.oMkypb._F7T7.5-Jbz
vaeV.4vDtu1HW~e-Fn-O.PM57leL1DqDa6
b2.vx-7Q9~Kn~Sc2Ii5dGnbEhPBj-1-PS7
~y4_UnMLgnnBu5.aXAQj-0j0X2p1-UEk_n
```

jiasli: If the password starts with dash `-`, Python's `argparse` will treat it as a parameter, like `--name`, instead of an argument. See https://github.com/Azure/azure-cli/issues/10692. 

For now, even if we only use letter or number as the first character, this issue will emerge after we migrate to Microsoft Graph (https://github.com/Azure/azure-cli/issues/12946).

src/azure-cli/azure/cli/command_modules/role/custom.py
```diff
@@ -1745,17 +1745,25 @@ def _gen_guid():
...
+    random_source = string.ascii_letters + string.digits + safe_punctuation
+    alphanumeric = string.ascii_letters + string.digits
 
-    for i in range(length - 4):    # pylint: disable=unused-variable
+    password = random.choice(alphanumeric)
```
jiasli, line 1752: A better choice is to use [`secrets.choice`](https://docs.python.org/3/library/secrets.html#secrets.choice) which is the the "most secure source of randomness" instead of `random.choice`.

One problem is that `secrets` is introduced in Python 3.6. I am not sure whether using it will cause problem in Python 3.5 for which we have already dropped support.  + @fengzhou-msft

qianwens:
@chef-davin thanks for your contribution:) After internal discussion, we decided to use punctuation from '-_.~' to align with portal behavior.

src/azure-cli/azure/cli/command_modules/role/custom.py
```diff
@@ -1745,18 +1745,28 @@ def _gen_guid():
...
+    first_character = random.SystemRandom().choice(alphanumeric)
 
+    # make sure we have special character in the password
+    password = random.SystemRandom().choice(string.ascii_lowercase)
+    password += random.SystemRandom().choice(string.ascii_uppercase)
+    password += random.SystemRandom().choice(string.digits)
+    password += random.SystemRandom().choice(safe_punctuation)
+
+    # generate a password of the given length from the options in the random_source variable
+    for _ in range(length - 5):
```
arrownj, line 1762: For `_random_password` function itself, do we need to consider the case when length is less than 5 ?

qianwens: the length is 34 which is not configurable for the customer

chef-davin:
Sorry for not responding to comments sooner. Thank you all for making my code better and pushing this change. I think it'll definitely make things a bit better in the future.

singularpoint:
Just FYI, this problem cases problems with Azure IoT Edge runtime, because this special symbols in password broke the recommended way to connect to Azure Container Registry with service principal credentials.

Azure Edge stores the credentials in the deployment manifest as a plaintext stings in JSON, so it may be connected with JSON deserialization when doing Docker authorization.

lastcoolnameleft:
IIRC, it used to generate a GUID for the password, which I thought was brilliant when I first encountered it.  Easy to copy/paste and automate with no special characters other than '-'.

jiasli:
> IIRC, it used to generate a GUID for the password, which I thought was brilliant when I first encountered it. Easy to copy/paste and automate with no special characters other than '-'.

Hi @lastcoolnameleft, the current implementation only contains `-_.~` as special characters, which shouldn't have any issue in popular shells.

Issue #13361, opened 2020-05-06
webapp: az appservice ase create command fails on azure cli version 2.4.0, 2.5.0 or 2.5.1 (last worked on v2.3.1)


tocorobo:
**DESCRIPTION**
The command *az appservice ase create* fails with azure cli version 2.4.0, 2.5.0 or 2.5.1 on CentOS 7.8 [last working version was 2.3.1].

**ERROR**
The request content was invalid and could not be deserialized: 'Could not find member 'template' on object of type 'DeploymentDefinition'. Path 'template', line 1, position 12.'.

**STEPS TO REPRODUCE**
- Procure CentOS 7.8 Linux OS instance
- Patch OS with latest updates; reboot.
*sudo yum update*
*sudo shutdown -r -time 0*
- Follow steps in https://docs.microsoft.com/en-us/cli/azure/install-azure-cli-yum?view=azure-cli-latest to setup azure cli repo.

*sudo rpm --import https://packages.microsoft.com/keys/microsoft.asc*

*sudo sh -c 'echo -e "[azure-cli]
name=Azure CLI
baseurl=https://packages.microsoft.com/yumrepos/azure-cli
enabled=1
gpgcheck=1
gpgkey=https://packages.microsoft.com/keys/microsoft.asc" > /etc/yum.repos.d/azure-cli.repo'*

- Install azure-cli version 2.4.0, 2.5.0 or 2.5.1 (pick one):
*sudo yum install azure-cli-2.4.0-1.el7.x86_64*
OR
*sudo yum install azure-cli-2.5.0-1.el7.x86_64*
OR
*sudo yum install azure-cli-2.5.1-1.el7.x86_64*

- Login to your azure subscription.
az login

- Create a bash script similar to the following which will try to create the azure appservice environment using the *az appservice ase create* command.

#!/bin/bash
aseRgName="ASE_RESOURCE_GROUP_NAME_HERE"
aseName="ASE_NAME_HERE"
aseVnetName="VNET_NAME_CONTAINING_SUBNET_THAT_ASE_WILL_BE_TIED_TO"
aseVnetRgName="VNET_RESOURCE_GROUP_NAME_HERE"
aseSubnetName="NAME_OF_SUBNET_TO_ASSOCIATE_WITH_ASE_HERE"
aseSubnetId=$(az network vnet subnet list --resource-group $aseVnetRgName --vnet-name $aseVnetName --query "[].id" -o tsv | grep $aseSubnetName)
[...]

Issue #13140, opened 2020-04-23
Error when running az appservice ase list-addresses -n ASEName


djr29578:
## Describe the bug
When running this command a list of VIP addresses should be the output.  This worked on az cli verison 2.0.78 but started failing when I upgraded to 2.4

**Command Name**
`az appservice ase list-addresses`

**Errors:**
```
'AppServiceEnvironmentsOperations' object has no attribute 'list_vips'
Traceback (most recent call last):
python3.6/site-packages/knack/cli.py, ln 215, in invoke
    cmd_result = self.invocation.execute(args)
cli/core/commands/__init__.py, ln 631, in execute
    raise ex
cli/core/commands/__init__.py, ln 695, in _run_jobs_serially
    results.append(self._run_job(expanded_arg, cmd_copy))
...
azure/cli/core/__init__.py, ln 574, in default_command_handler
    return op(**command_args)
cli/command_modules/appservice/appservice_environment.py, ln 100, in list_appserviceenvironment_addresses
    return ase_client.list_vips(resource_group_name, name)
AttributeError: 'AppServiceEnvironmentsOperations' object has no attribute 'list_vips'
```

## To Reproduce:
Steps to reproduce the behavior. Note that argument values have been redacted, as they may contain sensitive information.

- _Put any pre-requisite steps here..._
- `az appservice ase list-addresses -n {}`

## Expected Behavior
Expected outcome based on command ran in previous versions of AZ CLI (replaced IP addresses with ##):    

{
  "internalIpAddress": null,
  "outboundIpAddresses": [
    "##"
  ],
  "serviceIpAddress": "##",
  "vipMappings": [
    {
      "inUse": false,
      "internalHttpPort": 20003,
      "internalHttpsPort": 20001,
      "serviceName": null,
      "virtualIp": "##"
    },
    {
      "inUse": false,
      "internalHttpPort": 20004,
      "internalHttpsPort": 20002,
      "serviceName": null,
      "virtualIp": "##"
    }
  ]
}

## Environment Summary
```
Linux-4.4.0-18362-Microsoft-x86_64-with-debian-buster-sid
Python 3.6.5
Installer: DEB

azure-cli 2.4.0

Extensions:
application-insights 0.1.1

```

yonzhan:
appservice

yungezz:
bug

djr29578:
Is there any update to the timeline on when this fix will be complete?

panchagnula:
@madsd FYI,

[...]

PR #13737, opened 2020-05-29
[AppService] Defaulting AppServiceEnviromentOperations to 2018-02-01 version since…

panchagnula:
… swagger is currently broken for a fee operations on the latest version

**Description**  

**Testing Guide**  

**History Notes**  

[Component Name 1] BREAKING CHANGE: az command a: Make some customer-facing breaking change.  
[Component Name 2] az command b: Add some customer-facing feature.

---

This checklist is used to make sure that common guidelines for a pull request are followed.

- [ ] The PR title and description has followed the guideline in [Submitting Pull Requests](https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules#submitting-pull-requests).

- [ ] I adhere to the [Command Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md).

PR #13744, opened 2020-05-29
[AppService] Fixing failing az appservice ase create and list-addresses commands

madsd:
**Description**  

**Testing Guide**  

**History Notes**  

[AppService] az appservice ase create: Fix create issue #13361
[AppService] az appservice ase list-addresses: Fix change of SDK #13140.

---


yonzhan:
AppService

src/azure-cli/azure/cli/command_modules/appservice/appservice_environment.py
```diff
@@ -59,7 +58,8 @@ def create_appserviceenvironment_arm(cmd, resource_group_name, name, subnet, vne
 
     logger.info('Create App Service Environment...')
     deployment_name = _get_unique_deployment_name('cli_ase_deploy_')
-    ase_deployment_properties = _build_ase_deployment_properties(name, location, subnet_id, virtual_ip_type,
+    ase_deployment_properties = _build_ase_deployment_properties(cmd.cli_ctx, name, location,
```
panchagnula, line 61: @madsd lets make sure to not use positional arguments so that we don't hit these sort of issues anymore

src/azure-cli/azure/cli/command_modules/appservice/appservice_environment.py
```diff
@@ -350,8 +350,11 @@ def _get_unique_deployment_name(prefix):
     return prefix + random_string(16)
 
 
-def _build_ase_deployment_properties(name, location, subnet_id, virtual_ip_type,
+def _build_ase_deployment_properties(cli_ctx, name, location, subnet_id, virtual_ip_type,
                                      front_end_scale_factor=None, front_end_sku=None, tags=None):
+    from azure.cli.core.profiles import ResourceType, get_sdk
+    DeploymentProperties = get_sdk(cli_ctx, ResourceType.MGMT_RESOURCE_RESOURCES, 'DeploymentProperties', mod='models')
```
panchagnula, line 356: same here as well

src/azure-cli/azure/cli/command_modules/appservice/appservice_environment.py
madsd: After offline discussion with @panchagnula we decided to keep it this way.

src/azure-cli/azure/cli/command_modules/appservice/appservice_environment.py
madsd: After offline discussion with @panchagnula we decided to keep it this way.

madsd:
Agreed to keep these changes and prioritize keyword arguments going forward.

src/azure-cli/azure/cli/command_modules/appservice/appservice_environment.py
```diff
@@ -118,7 +122,7 @@ def _get_ase_client_factory(cli_ctx, api_version=None):
 
 def _get_deployment_client_factory(cli_ctx):
     from azure.mgmt.resource import ResourceManagementClient
-    client = get_mgmt_service_client(cli_ctx, ResourceManagementClient).deployments
+    client = get_mgmt_service_client(cli_ctx, ResourceManagementClient, api_version=VERSION_2019_10_01).deployments
```
qwordy, line 125: Is it feasible to use a central configuration center instead of specifying in each method? E.g. _shared.py

madsd: Great suggestion, I have centralized the api_version, but made it possible to override in a specific method if needed (e.g. to use new features). I also made a few other optimizations now that the file was open 👍

src/azure-cli/azure/cli/command_modules/appservice/appservice_environment.py
```diff
@@ -3,33 +3,40 @@
...
 def list_appserviceenvironments(cmd, resource_group_name=None):
-    ase_client = _get_ase_client_factory(cmd.cli_ctx, VERSION_2019_02_01)
+    ase_client = _get_ase_client_factory(cmd.cli_ctx)
     if resource_group_name is None:
         return ase_client.list()
     return ase_client.list_by_resource_group(resource_group_name)
 
 
 def show_appserviceenvironment(cmd, name, resource_group_name=None):
-    ase_client = _get_ase_client_factory(cmd.cli_ctx, VERSION_2019_02_01)
```
panchagnula, line 32: I don't think the version here makes any difference AFAIK the version gets set from the shared.py, currently we will always use the default api-version, whichever is the latest. Have you checked if changing this to 2018-02-01 here actually makes a difference. See this PR where I was going to force ASE commands use 2018-02-01 https://github.com/Azure/azure-cli/pull/13737

madsd: You are right - for the WebServiceManagementClient it seems to fix the version to 2019-08-01. The other ManagementClient adapt to the set api_version. I would be nice to get this fixed, but for now this change with work with both the changed ASE SDK and the changed ARM SDK.

panchagnula: Your change does fix the issue, all I am asking is to remove setting the version here since it does nothing.

madsd: You could just have said so :-) - done.

panchagnula: Thanks @madsd

PR #13762, opened 2020-06-01
{RBAC} Fix #11210: `az role assignment list` should include groups if `--include-groups` is specified

### LINKED ISSUE #11210: az role assignment list not seeing groups
## Describe the bug
Updated to the lasted cli 2.0.76 and no longer view groups when running az role assignment list
This worked on 2.0.5X (not sure the exact version)

**Command Name**
`az role assignment list`

**Errors:**

## To Reproduce:

- `az role assignment list --all --assignee {} --subscription {} --include-groups --include-inherited --include-classic-administrators -o {}`

## Expected Behavior
List Groups & Users role assignments.

## Environment Summary
```
Windows-10-10.0.15063-SP0
Python 3.6.6
Shell: powershell.exe

azure-cli 2.0.76

Extensions:
aks-preview 0.4.4

```

qianwens:
**Description**  

Fix https://github.com/Azure/azure-cli/issues/11210

`az role assignment list --all --assignee "***" --include-groups` does not include the group that the assignee user is a member of. This is a regression issue in CLI.
Root cause: The role assignment list filtered the result based on the assignee object id which will exclude the group. 
How to fix: do not filter the assignee object id if the "scope" is not specified since the list query already includes the filter for assignee object id.  

**Testing Guide**  

`az role assignment list --all --assignee "***" --include-groups` should list the group which the user is member of.
**History Notes**  

---


yonzhan:
add to S171

jiasli:
Please kindly link the PR to the related issue in the title (for it to appear in history) and the description (for the issue to be linked).

jiasli:
You may rebase on `dev` to rectify the history.

src/azure-cli/azure/cli/command_modules/role/custom.py
```diff
@@ -483,7 +483,8 @@ def _search_role_assignments(cli_ctx, assignments_client, definitions_client,
             role_id = _resolve_role_id(role, scope, definitions_client)
             assignments = [i for i in assignments if worker.get_role_property(i, 'role_definition_id') == role_id]
 
-        if assignee_object_id:
+        # filter the assignee if "scope" is provided
+        if assignee_object_id and scope:
```
jiasli, line 487: I don't think this solves the issue when `scope` is indeed provided - group will still be excluded.

Maybe we can take advantage of the `filter` parameter of `list_for_scope`, similar to `list`? I haven't tested yet. 

https://github.com/Azure/azure-sdk-for-python/blob/71baf7ea47489a73d5b84508f5e70f8705f41e1e/sdk/authorization/azure-mgmt-authorization/azure/mgmt/authorization/v2018_09_01_preview/operations/_role_assignments_operations.py#L660-L661

```py
    def list_for_scope(
            self, scope, filter=None, custom_headers=None, raw=False, **operation_config):
        """Gets role assignments for a scope.

        :param filter: The filter to apply on the operation. Use
         $filter=atScope() to return all role assignments at or above the
         scope. Use $filter=principalId eq {id} to return all role assignments
         at, above or below the scope for the specified principal.
        :type filter: str
        """
```

https://docs.microsoft.com/en-us/azure/role-based-access-control/role-assignments-list-rest#list-role-assignments

| Filter | Description |
| --- | --- |
| `$filter=atScope()` | Lists role assignments for only the specified scope, not including the role assignments at subscopes. |
| `$filter=assignedTo('{objectId}')` | Lists role assignments for a specified user or service principal.<br/>If the user is a member of a group that has a role assignment, that role assignment is also listed. This filter is transitive for groups which means that if the user is a member of a group and that group is a member of another group that has a role assignment, that role assignment is also listed.<br/>This filter only accepts an object ID for a user or a service principal. You cannot pass an object ID for a group. |
| `$filter=atScope()+and+assignedTo('{objectId}')` | Lists role assignments for the specified user or service principal and at the specified scope. |
| `$filter=principalId+eq+'{objectId}'` | Lists role assignments for a specified user, group, or service principal. |

jiasli: BTW, the doc for REST API [Role Assignments - List](https://docs.microsoft.com/en-us/rest/api/authorization/roleassignments/list) is missing description for `assignedTo`.


Name | In | Required | Type | Description
-- | -- | -- | -- | --
subscriptionId | path | True | string | The ID of the target subscription.
api-version | query | True | string | The API version to use for this operation.
$filter | query |   | string | The filter to apply on the operation. Use $filter=atScope() to return all role assignments at or above the scope. Use $filter=principalId eq {id} to return all role assignments at, above or below the scope for the specified principal.

src/azure-cli/azure/cli/command_modules/role/custom.py
```diff
@@ -483,7 +486,8 @@ def _search_role_assignments(cli_ctx, assignments_client, definitions_client,
             role_id = _resolve_role_id(role, scope, definitions_client)
             assignments = [i for i in assignments if worker.get_role_property(i, 'role_definition_id') == role_id]
 
-        if assignee_object_id:
+        # filter the assignee if "include_groups" is not provided
+        if assignee_object_id and not include_groups:
```
jiasli, line 490: Is this still necessary? I think `f = f + " and assignedTo('{}')".format(assignee_object_id)` already has the filter.

PR #13763, opened 2020-06-01
[Core] Add new get command loader mechanism when loading CommandLoader

Juliehzl:
**Description**  

 
According to spec[ design](https://microsoft.sharepoint.com/:w:/t/AzureManagementExperience/EZ9Zu0EaSq9OvLQgJDVcaj0BRs9ewFhKMAy5HIFjjKmBEw?e=M0suFg), two kinds of CommandLoader will be introduced in storage module.

Here in storage initialization, AzureStackStorageCommandLoader will be loaded when cloud profile is for Azure Stack and StorageCommandLoader will be used in public cloud.
For other modules, if there is requirement to load different CommandLoader, the `get_command_loader()` function can be utilized.

Here `command_v2.py` and `params_v2.py` are totally the same as existing `command.py` and `params.py`.

**Testing Guide**  

In public cloud, `az storage -h` will see all commands for public cloud.

Switch the profile with `az cloud set`, `az storage -h` will only show commands Azure Stack supports.

**History Notes**  

---


yonzhan:
azure stack command loader

src/azure-cli/azure/cli/command_modules/storage/_params_v2.py
```diff
@@ -0,0 +1,1236 @@
...
+
+    from azure.cli.core.commands.parameters import get_resource_name_completion_list
+
+    from .sdkutil import get_table_data_type
+    from .completers import get_storage_name_completion_list
+
+    t_base_blob_service = self.get_sdk('blob.baseblobservice#BaseBlobService')
+    t_file_service = self.get_sdk('file#FileService')
+    t_queue_service = self.get_sdk('queue#QueueService')
+    t_table_service = get_table_data_type(self.cli_ctx, 'table', 'TableService')
```
zhoxing-ms, line 38: The definition of these xxxServices comes from track 1, right? Is there any way to reduce the dependence of track 1 they introduce on track 2?

Juliehzl: chat offline already.

src/azure-cli/azure/cli/command_modules/storage/__init__.py
```diff
@@ -259,4 +282,8 @@ def new_handler(ex):
     kwargs['exception_handler'] = new_handler
 
 
-COMMAND_LOADER_CLS = StorageCommandsLoader
+def get_command_loader(cli_ctx):
+    if cli_ctx.cloud.profile != 'latest':
```
MyronFanQiu, line 286:
```suggestion
    if cli_ctx.cloud.profile.lower() != 'latest':
```

src/azure-cli/azure/cli/command_modules/storage/__init__.py
```diff
@@ -23,12 +23,35 @@ def __init__(self, cli_ctx=None):
...
         load_command_table(self, args)
         return self.command_table
 
     def load_arguments(self, command):
         super(StorageCommandsLoader, self).load_arguments(command)
+        from azure.cli.command_modules.storage._params_v2 import load_arguments
+        load_arguments(self, command)
+
+
+class AzureStackStorageCommandsLoader(AzCommandsLoader):
```
MyronFanQiu, line 36: Naming is little bit tricky for me. Here are two command modules `storage.commands` and `storage.commands_v2`. Beside we have two command loader names `StorageCommandsLoader` and `AzureStackStorageCommandsLoader`. Is it possible to align these three names into two set of names? Why do we need such name `AzureStackStorage`?

jiasli: The naming of `AzureStack` requires a background knowledge of the internal requirement and dependencies. I agree with @MyronFanQiu that we should keep the name in sync.

jsntcy:
> class StorageCommandsLoader(AzCommandsLoader):

Better to always keep this loader for latest so the naming is more consistent and easier to maintain.

---
Refers to: src/azure-cli/azure/cli/command_modules/storage/__init__.py:13 in f45a298. [](commit_id = f45a298149a53ca63b2687de146a9a751f786a98, deletion_comment = False)

src/azure-cli/azure/cli/command_modules/storage/__init__.py
jsntcy: Better to keep loader name and name of command file and params file consistent for easier understand.

---
In reply to: [439186279](https://github.com/Azure/azure-cli/pull/13763#discussion_r439186279) [](ancestors = 439186279)

Juliehzl: done

Juliehzl:
> > class StorageCommandsLoader(AzCommandsLoader):
> 
> Better to always keep this loader for latest so the naming is more consistent and easier to maintain.
> 
> Refers to: src/azure-cli/azure/cli/command_modules/storage/**init**.py:13 in [f45a298](https://github.com/Azure/azure-cli/commit/f45a298149a53ca63b2687de146a9a751f786a98). [](commit_id = [f45a298](https://github.com/Azure/azure-cli/commit/f45a298149a53ca63b2687de146a9a751f786a98), deletion_comment = False)

sure

PR #13774, opened 2020-06-01
{Doc} Add Git Bash doc

fengzhou-msft:
**Description**  

Create an intial doc for Azure CLI usage on Git Bash. Best practice or workaround for issues with Git Bash can be put in this doc.

**Testing Guide**  

**History Notes**  

---

### DIFF
--- a/doc/use_cli_with_git_bash.md
+++ b/doc/use_cli_with_git_bash.md
@@ -45,5 +45,5 @@ The [entry script](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/az)
 
 According to [_exec, _wexec Functions](https://docs.microsoft.com/en-us/cpp/c-runtime-library/exec-wexec-functions?view=vs-2019), `os.exec` internally uses [CreateProcess](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw) which doesn't wait for the sub-process.
 
-#### Exit Code is 0 even the command fails
+#### Exit Code is 0 even when the command fails
 Related to the above issue, Windows creates a new process and exits the current one with `os.exec`. Hence the calling program only sees that the script has terminated without an issue (See this [comment](https://bugs.python.org/issue9148#msg109179)). The exit code does not reflect the actual result of the command execution in the new process.
--- a/doc/use_cli_with_git_bash.md
+++ b/doc/use_cli_with_git_bash.md
@@ -13,7 +13,7 @@ $ az vm show --ids "/subscriptions/00000000-0000-0000-0000-000000000000/resource
 
 invalid resource ID: C:/Program Files/Git/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/my-rg/providers/Microsoft.Compute/virtualMachines/my-vm
 ```
-To disable the path conversion. You can set enviroment variable MSYS_NO_PATHCONV=1 or set it temporarily when a running command:
+To disable the path conversion. You can set enviroment variable `MSYS_NO_PATHCONV=1` or set it temporarily when a running command:
 ```
 $ MSYS_NO_PATHCONV=1 az vm show --ids "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/my-rg/providers/Microsoft.Compute/virtualMachines/my-vm"
 ```
--- a/doc/use_cli_with_git_bash.md
+++ b/doc/use_cli_with_git_bash.md
@@ -1,0 +1,38 @@
+# Use Azure CLI with Git Bash
+
+## Introduction
+The MSI package for Windows now contains an entry script for running az on Git Bash. You can directly call `az` on Git Bash now. While using Git Bash on Windows gives you a similar experience on a Linux shell, it has some unexpected issues that impact the user experience of Azure CLI.
+
+## Issues
+
+### Auto-translation of Resource IDs
+
+On Git Bash, if you specify command-line options starting with a slash, POSIX-to-Windows path conversion will kick in converting. This causes an issue for passing ARM Resource IDs e.g.:
+```
+$ az vm show --ids "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/my-rg/providers/Microsoft.Compute/virtualMachines/my-vm"
+
+invalid resource ID: C:/Program Files/Git/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/my-rg/providers/Microsoft.Compute/virtualMachines/my-vm
+```
+To disable the path conversion. You can set enviroment variable MSYS_NO_PATHCONV=1 or set it temporarily when a running command:
+```
+$ MSYS_NO_PATHCONV=1 az vm show --ids "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/my-rg/providers/Microsoft.Compute/virtualMachines/my-vm"
+```
+
+More discussions [here](https://stackoverflow.com/questions/7250130/how-to-stop-mingw-and-msys-from-mangling-path-names-given-at-the-command-line#34386471).
+
+### Quoting Issues
+If you install Azure CLI with `pip install`, the double quotes will be stripped in the value and cause trouble when there are spaces inside the quotes since Git Bash will treat each parts separated by spaces as an invividual command option.
+```
+$ az find "vm create"
+$ Command arguments: ['find', 'vm', 'create', '--debug']
+...
+az: error: unrecognized arguments: create
+...
+```
+You can wrap the value with additional single quotes to avoid the issue:
+```
+$ az find '"vm create"' --debug
+$ Command arguments: ['find', 'vm create', '--debug']
+...
+```
+We do recommend to use MSI to install Azure CLI on Windows when it's possible.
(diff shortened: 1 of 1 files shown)


yonzhan:
Git Bash doc

doc/use_cli_with_git_bash.md
```diff
@@ -0,0 +1,38 @@
+# Use Azure CLI with Git Bash
+
+## Introduction
+The MSI package for Windows now contains an entry script for running az on Git Bash. You can directly call `az` on Git Bash now. While using Git Bash on Windows gives you a similar experience on a Linux shell, it has some unexpected issues that impact the user experience of Azure CLI.
```
jiasli, line 4:
```suggestion
The MSI package for Windows now contains an entry script for running `az` on Git Bash. You can directly call `az` on Git Bash now. While using Git Bash on Windows gives you a similar experience on a Linux shell, it has some unexpected issues that impact the user experience of Azure CLI.
```

doc/use_cli_with_git_bash.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Issues
+
+### Auto-translation of Resource IDs
+
+On Git Bash, if you specify command-line options starting with a slash, POSIX-to-Windows path conversion will kick in converting. This causes an issue for passing ARM Resource IDs e.g.:
```
jiasli, line 10:
```suggestion
On Git Bash, if you specify command-line options starting with a slash, POSIX-to-Windows path conversion will kick in. This causes an issue for passing ARM Resource IDs, like:
```

doc/use_cli_with_git_bash.md
```diff
@@ -0,0 +1,38 @@
...
+```
+To disable the path conversion. You can set enviroment variable `MSYS_NO_PATHCONV=1` or set it temporarily when a running command:
+```
+$ MSYS_NO_PATHCONV=1 az vm show --ids "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/my-rg/providers/Microsoft.Compute/virtualMachines/my-vm"
+```
+
+More discussions [here](https://stackoverflow.com/questions/7250130/how-to-stop-mingw-and-msys-from-mangling-path-names-given-at-the-command-line#34386471).
+
+### Quoting Issues
+If you install Azure CLI with `pip install`, the double quotes will be stripped in the value and cause trouble when there are spaces inside the quotes since Git Bash will treat each parts separated by spaces as an invividual command option.
```
jiasli, line 24: Why does this happen? Does this have anything to do with https://github.com/Azure/azure-cli/blob/dev/doc/quoting-issues-with-powershell.md?

jiasli: I figured it out. It it because [`execl`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/exec-wexec-functions?view=vs-2019) internally calls [CreateProcessW ](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw):

https://docs.microsoft.com/en-us/cpp/c-runtime-library/exec-wexec-functions?view=vs-2019

> Spaces embedded in strings may cause unexpected behavior; for example, passing `_exec` the string `"hi there"` will result in the new process getting two arguments, `"hi"` and `"there"`. If the intent was to have the new process open a file named "hi there", the process would fail. You can avoid this by quoting the string: `"\"hi there\""`.

[`subprocess.Popen`](https://docs.python.org/3/library/subprocess.html#subprocess.Popen) can handle escaping correctly by converting `args` to a string in a manner described in [Converting an argument sequence to a string on Windows](https://docs.python.org/3/library/subprocess.html#converting-argument-sequence). The implementation is [`subprocess.list2cmdline`](https://github.com/python/cpython/blob/1def7754b7a41fe57efafaf5eff24cfa15353444/Lib/subprocess.py#L529).

jiasli: The way how Windows parses command-line arguments can be found at [Parsing C++ command-Line arguments](https://docs.microsoft.com/en-us/cpp/cpp/main-function-command-line-args?view=vs-2019#parsing-c-command-line-arguments):

> Microsoft C/C++ startup code uses the following rules when interpreting arguments given on the operating system command line:
> 
> - Arguments are delimited by white space, which is either a space or a tab.
> 
> - The caret character (^) is not recognized as an escape character or delimiter. The character is handled completely by the command-line parser in the operating system before being passed to the `argv` array in the program.
> 
> - A string surrounded by double quotation marks ("*string*") is interpreted as a single argument, regardless of white space contained within. A quoted string can be embedded in an argument.
> 
> - A double quotation mark preceded by a backslash (\\") is interpreted as a literal double quotation mark character (").
> 
> - Backslashes are interpreted literally, unless they immediately precede a double quotation mark.
> 
> - If an even number of backslashes is followed by a double quotation mark, one backslash is placed in the `argv` array for every pair of backslashes, and the double quotation mark is interpreted as a string delimiter.
> 
> - If an odd number of backslashes is followed by a double quotation mark, one backslash is placed in the `argv` array for every pair of backslashes, and the double quotation mark is "escaped" by the remaining backslash, causing a literal double quotation mark (") to be placed in `argv`.

doc/use_cli_with_git_bash.md
```diff
@@ -0,0 +1,49 @@
...
+$ Command arguments: ['find', 'vm create', '--debug']
+...
+```
+
+#### Entry script exits before command finishes
+The [entry script](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/az) exits once `os.execl()` is called. It does not wait for the actual command to finish.
+
+According to [_exec, _wexec Functions](https://docs.microsoft.com/en-us/cpp/c-runtime-library/exec-wexec-functions?view=vs-2019), `os.exec` internally uses [CreateProcess](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw) which doesn't wait for the sub-process.
+
+#### Exit Code is 0 even the command fails
```
fengzhou-msft, line 48:
```suggestion
#### Exit Code is 0 even when the command fails
```

doc/use_cli_with_git_bash.md
```diff
@@ -0,0 +1,49 @@
+# Use Azure CLI with Git Bash
+
+## Introduction
+The MSI package for Windows now contains an [az entry script](https://github.com/Azure/azure-cli/blob/dev/build_scripts/windows/scripts/az) for running `az` on Git Bash. You can directly call `az` on Git Bash now. While using Git Bash on Windows gives you a similar experience on a Linux shell, it has some unexpected issues that impact the user experience of Azure CLI. 
+
+We do not recommend to use `az` installed through `pip install azure-cli` on Git Bash as that [az entry script](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/az) was meant to be run on Linux and have even more issues when run on Git Bash on Windows.
```
jiasli, line 6:
```suggestion
We do not recommend to use `az` installed through `pip install azure-cli` on Git Bash as that [az entry script](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/az) was meant to be run on Linux and have even more issues when running on Git Bash on Windows.  
```

doc/use_cli_with_git_bash.md
```diff
@@ -0,0 +1,49 @@
...
+invalid resource ID: C:/Program Files/Git/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/my-rg/providers/Microsoft.Compute/virtualMachines/my-vm
+```
+To disable the path conversion. You can set enviroment variable `MSYS_NO_PATHCONV=1` or set it temporarily when a running command:
+```
+$ MSYS_NO_PATHCONV=1 az vm show --ids "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/my-rg/providers/Microsoft.Compute/virtualMachines/my-vm"
+```
+
+More discussions [here](https://stackoverflow.com/questions/7250130/how-to-stop-mingw-and-msys-from-mangling-path-names-given-at-the-command-line#34386471).
+
+### Issues with pip installed az
```
jiasli, line 25:
```suggestion
### Issues with pip-installed az
```

doc/use_cli_with_git_bash.md
```diff
@@ -0,0 +1,49 @@
...
+...
+```
+
+#### Entry script exits before command finishes
+The [entry script](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/az) exits once `os.execl()` is called. It does not wait for the actual command to finish.
+
+According to [_exec, _wexec Functions](https://docs.microsoft.com/en-us/cpp/c-runtime-library/exec-wexec-functions?view=vs-2019), `os.exec` internally uses [CreateProcess](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw) which doesn't wait for the sub-process.
+
+#### Exit Code is 0 even when the command fails
+Related to the above issue, Windows creates a new process and exits the current one with `os.exec`. Hence the calling program only sees that the script has terminated without an issue (See this [comment](https://bugs.python.org/issue9148#msg109179)). The exit code does not reflect the actual result of the command execution in the new process.
```
jiasli, line 49:
```suggestion
Related to the above issue, Windows creates a new process and exits the current one with `os.exec`. Hence the calling program only sees that the script has terminated without an issue (See [Issue9148: os.execve puts process to background on windows](https://bugs.python.org/issue9148#msg109179)). The exit code does not reflect the actual result of the command execution in the new process.
```

PR #13783, opened 2020-06-02
{Util} Add util module for version and rest

jiasli:
## Description

Peel off `az version` and `az rest` from `resource` module. A new module `util` is created. Thus they don't carry the burden of 

1. `azure-mgmt-resource` initialization when running these commands
2. scenario tests on `resource` command module when developing these commands

This makes running and developing `az version` and `az rest` significantly faster. 

## Testing Guide
```powershell
> az version
{
  "azure-cli": "2.6.0",
  "azure-cli-command-modules-nspkg": "2.0.3",
  "azure-cli-core": "2.6.0",
  "azure-cli-dev-tools": "0.1.1",
  "azure-cli-nspkg": "3.0.4",
  "azure-cli-telemetry": "1.0.4",
  "azure-cli-testsdk": "0.2.4",
  "extensions": {
    "application-insights": "0.1.1",
    "interactive": "0.4.4",
    "mesh": "0.10.6",
    "timeseriesinsights": "0.1.2",
    "webapp": "0.2.24"
  }
}

> az rest -m GET -u https://management.azure.com/subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/resourcegroups/rg1?api-version=2019-07-01
{
  "id": "/subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/resourceGroups/rg1",
  "location": "westus",
  "name": "rg1",
  "properties": {
    "provisioningState": "Succeeded"
  },
  "tags": {
    "StorageType": "Standard_LRS",
    "type": "test"
  },
  "type": "Microsoft.Resources/resourceGroups"
}
```


yonzhan:
add to S171

src/azure-cli/azure/cli/command_modules/util/custom.py
```diff
@@ -0,0 +1,31 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+from __future__ import print_function
```
zhoxing-ms, line 6: We don't support Python 2.0+ anymore, right?
Do we still need `from __future__ import print_function`? [print-as-a-function](https://docs.python.org/3/whatsnew/2.6.html?highlight=print_function#pep-3105-print-as-a-function)

jiasli: Nice catch! I simply migrated the code from `src/azure-cli/azure/cli/command_modules/resource/custom.py`. Removed.

src/azure-cli/azure/cli/command_modules/util/custom.py
```diff
@@ -0,0 +1,29 @@
...
+    from azure.cli.core.util import send_raw_request
+    r = send_raw_request(cmd.cli_ctx, method, url, headers, uri_parameters, body,
+                         skip_authorization_header, resource, output_file)
+    if not output_file and r.content:
+        try:
+            return r.json()
+        except ValueError:
+            logger.warning('Not a json response, outputting to stdout. For binary data '
+                           'suggest use "--output-file" to write to a file')
+            print(r.text)
```
Juliehzl, line 22: Prefer to not use print() according to command guideline.

PR #13785, opened 2020-06-02
[Identity] Rename auth_only to tenant_access

jiasli:
**Description**  

1. Rename `auth_only` to a better name `tenant_access`, indicating bare tenant-level access.
2. Deprecate `--allow-no-subscriptions`, because:
   1. Even with `--allow-no-subscriptions`, `az login` fails when ARM request is blocked.
   2. Depending on whether the tenant has subscriptions, the `id` property in `az account list/show` is ambiguous: it can mean either tenant ID or subscription ID. 

**Testing Guide**  
```
> az login -h

    --allow-no-subscriptions [Deprecated] : Support access tenants without subscriptions.
                                            It's uncommon but useful to run tenant level commands,
                                            such as `az ad`.
        Option '--allow-no-subscriptions' has been deprecated and will be removed in version
        '3.0.0'. Use '--tenant-access' instead.

    --tenant-access                       : Only log in to the home tenant or the tenant specified
                                            by --tenant. CLI will not perform ARM operations to list
                                            tenants and subscriptions. Then you may run tenant-level
                                            commands, such as `az ad`, `az account get-access-
                                            token`.
```

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/profile/__init__.py
+++ b/src/azure-cli/azure/cli/command_modules/profile/__init__.py
@@ -57,7 +57,12 @@ class ProfileCommandsLoader(AzCommandsLoader):
             c.argument('service_principal', action='store_true', help='The credential representing a service principal.')
             c.argument('username', options_list=['--username', '-u'], help='user name, service principal, or managed service identity ID')
             c.argument('tenant', options_list=['--tenant', '-t'], help='The AAD tenant, must provide when using service principals.', validator=validate_tenant)
-            c.argument('allow_no_subscriptions', action='store_true', help="Support access tenants without subscriptions. It's uncommon but useful to run tenant level commands, such as 'az ad'")
+            c.argument('tenant_access', action='store_true',
+                       help='Only log in to the home tenant or the tenant specified by --tenant. CLI will not perform '
+                            'ARM operations to list tenants and subscriptions. Then you may run tenant-level commands, '
+                            'such as `az ad`, `az account get-access-token`.')
+            c.argument('allow_no_subscriptions', action='store_true', deprecate_info=c.deprecate(target='--allow-no-subscriptions', expiration='3.0.0', redirect="--tenant-access", hide=False),
+                       help="Support access tenants without subscriptions. It's uncommon but useful to run tenant level commands, such as `az ad`")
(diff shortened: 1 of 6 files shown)


src/azure-cli/requirements.py3.windows.txt
```diff
@@ -115,7 +115,7 @@ pypiwin32==223
 pyreadline==2.1
 python-dateutil==2.8.0
 pytz==2019.1
-pywin32==225
+pywin32==227
```
jiasli, line 118: Update `pywin32` to 227, otherwise `portalocker` will fail (https://github.com/mhammond/pywin32/issues/1431).

src/azure-cli/azure/cli/command_modules/profile/__init__.py
```diff
@@ -57,7 +57,12 @@ def load_arguments(self, command):
             c.argument('service_principal', action='store_true', help='The credential representing a service principal.')
             c.argument('username', options_list=['--username', '-u'], help='user name, service principal, or managed service identity ID')
             c.argument('tenant', options_list=['--tenant', '-t'], help='The AAD tenant, must provide when using service principals.', validator=validate_tenant)
-            c.argument('allow_no_subscriptions', action='store_true', help="Support access tenants without subscriptions. It's uncommon but useful to run tenant level commands, such as 'az ad'")
+            c.argument('tenant_access', action='store_true',
+                       help='Only log in to the home tenant or the tenant specified by --tenant. CLI will not perform '
+                            'ARM operations to list tenants and subscriptions. Then you may run tenant-level commands, '
+                            'such as `az ad`, `az account get-access-token`.')
+            c.argument('allow_no_subscriptions', action='store_true', deprecate_info=c.deprecate(target='--allow-no-subscriptions', expiration='3.0.0', redirect="--tenant-access", hide=False),
```
qianwens, line 64: MSAL adoption won't bump version to 3.0, the expiration version could be 2.15.0?

yonzhan:
Identity

src/azure-cli/azure/cli/command_modules/profile/__init__.py
jiasli: I'd like to set the expiration date to a more explicit time, instead of deprecating it unintentionally.

PR #13567, opened 2020-05-19
[Identity] Support Cloud Shell login

jiasli:
## Description

~~Currently can't test in Cloud Shell due to the lack of Python 3.6+. See internal email: Python 3.6+ support in Cloud Shell~~

Use [`pyenv`](https://github.com/pyenv/pyenv) to install Python 3.6+ in Cloud Shell. See https://stackoverflow.com/a/61890236/2199657. Special thanks to @NullMDR for this fascinating solution. 

Log in to Cloud Shell.

```
$ az login --identity
Using Cloud Shell Managed Identity: {"tenant_id": "54826b22-38d6-4fb2-bad9-b7b93a3e9c5a", "unique_name": "xxx@microsoft.com"}
[
  {
    "environmentName": "AzureCloud",
    "homeTenantId": "54826b22-38d6-4fb2-bad9-b7b93a3e9c5a",
    "id": "0b1f6471-1bf0-4dda-aec3-cb9272f09590",
    "isDefault": true,
    "managedByTenants": [
      {
        "tenantId": "2f4a9838-26b7-47ee-be60-ccc1fdec5953"
      }
    ],
    "name": "AzureSDKTest",
    "state": "Enabled",
    "tenantId": "54826b22-38d6-4fb2-bad9-b7b93a3e9c5a",
    "user": {
      "cloudShellID": true,
      "name": "xxx@microsoft.com",
      "type": "user"
    }
  }
]
```

We can also test on a VM with managed identity and set env var `ACC_CLOUD=1` to mock Cloud Shell.

```
C:\Users\admin\env38\Scripts\python.exe C:/Users/admin/azure-cli/src/azure-cli/azure/cli/__main__.py login --identity
Using Cloud Shell Managed Identity: {"tenant_id": "54826b22-38d6-4fb2-bad9-b7b93a3e9c5a", "unique_name": "N/A"}
[
  {
    "environmentName": "AzureCloud",
    "homeTenantId": "54826b22-38d6-4fb2-bad9-b7b93a3e9c5a",
    "id": "0b1f6471-1bf0-4dda-aec3-cb9272f09590",
    "isDefault": true,
    "managedByTenants": [
      {
        "tenantId": "2f4a9838-26b7-47ee-be60-ccc1fdec5953"
      }
    ],
    "name": "AzureSDKTest",
    "state": "Enabled",
    "tenantId": "54826b22-38d6-4fb2-bad9-b7b93a3e9c5a",
    "user": {
      "cloudShellID": true,
[...]

PR #13822, opened 2020-06-04
{Identity} Fix logout

jiasli:
**Description**  
Fix regression introduced in https://github.com/Azure/azure-cli/pull/13567. Use MSAL-style `authority` to build the `PublicClientApplication`.

**Testing Guide**  
```
az login
az logout
```

### DIFF
--- a/src/azure-cli-core/azure/cli/core/_identity.py
+++ b/src/azure-cli-core/azure/cli/core/_identity.py
@@ -64,10 +64,10 @@ class Identity:
 
     CLOUD_SHELL_IDENTITY_UNIQUE_NAME = "unique_name"
 
-    def __init__(self, authority=None, tenant_id=None, client_id=_CLIENT_ID, **kwargs):
+    def __init__(self, authority=None, tenant_id=None, client_id=None, **kwargs):
         self.authority = authority
-        self.tenant_id = tenant_id
-        self.client_id = client_id
+        self.tenant_id = tenant_id or "organizations"
+        self.client_id = client_id or _CLIENT_ID
         self._cred_cache = kwargs.pop('cred_cache', None)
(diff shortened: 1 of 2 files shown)


src/azure-cli-core/azure/cli/core/_identity.py
```diff
@@ -256,16 +258,15 @@ def _decode_managed_identity_token(self, credential, resource):
         return decoded
 
     def get_user(self, user_or_sp=None):
-        try:
-            return self._msal_app.get_accounts(user_or_sp)
-        except ValueError:
-            pass
+        accounts = self._msal_app.get_accounts(user_or_sp)
```
qianwens, line 261: why remove the try-catch? it throws an exception if the user does not exist which is not handled in the logout function

src/azure-cli-core/azure/cli/core/_identity.py
```diff
@@ -256,16 +258,15 @@ def _decode_managed_identity_token(self, credential, resource):
         return decoded
 
     def get_user(self, user_or_sp=None):
-        try:
-            return self._msal_app.get_accounts(user_or_sp)
-        except ValueError:
```
jiasli, line 261: `ValueError` is thrown due to the incorrect `authority` given to `PublicClientApplication`. If the `username` is not found, `[]` will be return.

yonzhan:
add to S171

src/azure-cli-core/azure/cli/core/_identity.py
jiasli: `ValueError` is thrown due to the incorrect `authority` given to `PublicClientApplication`. If the `username` is not found, `[]` will be return.

PR #13844, opened 2020-06-05
{Misc} Fix timer for __main__.py

jiasli:
## Description

Fix the incorrect timing logic. Previous it only takes `invoke` duration into account, while forgetting about the `init` duration. 

This PR fixes the timing logic to include both `init` and `invoke` duration. 

## Testing Guide

Before:

```
> az version -h --verbose
...
command ran in 0.881 seconds.
```

After:
```
> az version -h --verbose
...
Command ran in 1.679 seconds (init: 0.824, invoke: 0.855)
```


yonzhan:
add to S171

jiasli:
Because `azdev` doesn't honor `azure-cli` repo's `.flake8` (https://github.com/Azure/azure-cli-dev-tools/issues/209), disabling `E402` doesn't work in the CI and causes the [failure](https://dev.azure.com/azure-sdk/public/_build/results?buildId=415569&view=logs&j=36dd4138-4d53-5e46-00d9-e5cd9744be05&t=1cf3879c-0a42-5ccd-7693-7a4781d739d8&l=107):

```
/home/vsts/work/1/s/src/azure-cli/azure/cli/__main__.py:13:1: E402 module level import not at top of file
/home/vsts/work/1/s/src/azure-cli/azure/cli/__main__.py:14:1: E402 module level import not at top of file
/home/vsts/work/1/s/src/azure-cli/azure/cli/__main__.py:16:1: E402 module level import not at top of file
/home/vsts/work/1/s/src/azure-cli/azure/cli/__main__.py:17:1: E402 module level import not at top of file
/home/vsts/work/1/s/src/azure-cli/azure/cli/__main__.py:18:1: E402 module level import not at top of file
/home/vsts/work/1/s/src/azure-cli/azure/cli/__main__.py:19:1: E402 module level import not at top of file
6     E402 module level import not at top of file
```

Thus, [`per-file-ignores`](https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-per-file-ignores) won't work wither.

Since there is no support for file-level `# flake8: noqa` to disable certain errors (https://gitlab.com/pycqa/flake8/-/issues/399), temporarily comply to flake8's `E402` rule to make the CI pass.

src/azure-cli/azure/cli/__main__.py
```diff
@@ -47,27 +50,24 @@ def cli_main(cli, args):
...
     raise ex
 
 finally:
     telemetry.conclude()
 
     try:
-        logger.info("command ran in %.3f seconds.", elapsed_time)
+        # Log the invoke finish time
+        invoke_finish_time = timeit.default_timer()
+        logger.info("Command ran in %.3f seconds (init: %.3f, invoke: %.3f)",
```
jiasli, line 68: I am not sure if this **running time info** should be moved to `logger.debug`. I am open to this.

jiasli:
> Would it better to calculate the execution time under certain situations such as debug mode, --verbose mode, etc.
> 
> But this modifcation doesn't bring any (significant) regression, So, LGTM.

The execution time is calculated regardless of the execution mode (verbose, debug). `--verbose` and `--debug` only decide if the execution time is printed. 

Anyway, the overhead of calling `timeit.default_timer()` is pretty small. So we don't need to worry about it.

PR #13854, opened 2020-06-05
{Find} az find: Change telemetry name

mirdaki:
**Description**  

The Correlation ID was previously set to be the Session ID. The service at the time had not been updated to accept it as a separate value. That has since been resolved.

No user facing changes.

**Testing Guide**  

No new commands or tests. Local testing was done to ensure this feature works.

---


yonzhan:
add to S171

yungezz:
hi @mirdaki could you pls share more about the change? `service` here means aladdin service? you sent telemetry from Azure CLI to aladdin telemetry store?

mirdaki:
Hi @yungezz! Yes the telemetry, when the user authorizes it, is shared with the Aladdin services. It helps us build the model that powers the example creation.

This change in particular updates the property name that the Aladdin service uses. Previously we did not have a Correlation ID parameter defined, so that data would always be thrown out if we sent it as such. Now we have it defined, we can move away from passing it in as the Session ID.

PR #13294, opened 2020-05-01
{Core} Build and use command index

jiasli:
## Description

A command index is built and saved in file `.azure/commandIndex.json`. It contains a mapping from **top-level commands / command group names** to their **command module namespaces**.

Both built-in modules and extensions are included in the index:

```jsonc
{
    "version": "2.7.0",
    "cloudProfile": "latest",
    "commandIndex": {
        "acr": [
            "azure.cli.command_modules.acr"
        ],
        ...
        "network": [
            "azure.cli.command_modules.natgateway",
            "azure.cli.command_modules.network",
            "azure.cli.command_modules.privatedns"
        ],
        ...
        "sql": [
            "azure.cli.command_modules.sql",
            "azure.cli.command_modules.sqlvm"
        ],
        ...
        "timeseriesinsights": [
            "azext_timeseriesinsights"
        ]
    }
}

```

CLI checks if a command can be found from the index (either from built-in modules, extensions, or both):

- ✔ Yes, then only corresponding modules from the index are loaded. Then check if the loaded modules do have the command, in case of an outdated cache:
    -  ✔ Yes, continue parsing with only the loaded modules
    -  ❌ No, go to (*)
- ❌ (*) No, then all modules and extensions will be loaded and a new index will be built from the loaded command table

This significantly decreases the run time of commands, since unnecessary modules are not loaded anymore.

Together with other PRs, Azure CLI's run time are improved from 2.3s to 0.6s (`az version`, without the loading time of `pkg_resources`):

- **Fix built-in timer** (https://github.com/Azure/azure-cli/pull/13844): Previous it only takes `invoke` duration into account, while forgetting about the `init` duration. The timing logic is fixed to include both `init` and `invoke` duration. 

- **Delay unnecessary imports** (https://github.com/Azure/azure-cli/pull/13843): Previously during CLI initialization, importing `azure.core` and `msrestazure.tools` are very time-consuming, but actually unnecessary. This PR delays the import until `azure.core` and `msrestazure.tools` are actually used. 

- **Remove unnecessary imports of msrest**: Follow [EAFP](https://docs.python.org/3/glossary.html#term-eafp) (Easier to Ask for Forgiveness than Permission) style to remove unnecessary imports of msrest. Also see https://stackoverflow.com/a/11360880/2199657

[...]

PR #13867, opened 2020-06-08
{AppService} Delay importing is_valid_resource_id, parse_resource_id

jiasli:
**Description**  

`from msrestazure.tools import is_valid_resource_id, parse_resource_id` is slow and can slow down the whole Azure CLI!

image

Delay importing them and import then only when absolutely necessary.

Also migrate to `azure.mgmt.core` at the same time.


jiasli:
Closing as this usage of **top-level import** appears in many modules. This won't be a problem when https://github.com/Azure/azure-cli/pull/13294 is merged.

PR #13879, opened 2020-06-08
{Misc} Remove unnecessary base class method invocations

jiasli:
### `knack.commands.CLICommandsLoader.load_command_table`

Calling `knack.commands.CLICommandsLoader.load_command_table` does no more than generating a useless event which results in a useless line in the debug log:

```
Event: CommandLoader.OnLoadCommandTable []
```

`[]` stands for **no handler**. 

`knack.commands.CLICommandsLoader.load_command_table`:

https://github.com/microsoft/knack/blob/fe3bf5d3a79a3dd2ce5ddb0c38d93843a3380f6f/knack/commands.py#L203-L212

```py
    def load_command_table(self, args):  # pylint: disable=unused-argument
        """ Load commands into the command table

        :param args: List of the arguments from the command line
        :type args: list
        :return: The ordered command table
        :rtype: collections.OrderedDict
        """
        self.cli_ctx.raise_event(EVENT_CMDLOADER_LOAD_COMMAND_TABLE, cmd_tbl=self.command_table)
        return OrderedDict(self.command_table)
```

### `knack.commands.CLICommandsLoader.load_arguments`

[`knack.commands.CLICommandsLoader.load_arguments`](https://github.com/microsoft/knack/blob/fe3bf5d3a79a3dd2ce5ddb0c38d93843a3380f6f/knack/commands.py#L214-L232)'s work is already taken care of by `azure.cli.core.MainCommandsLoader.load_arguments` at 

https://github.com/Azure/azure-cli/blob/168e50e956111d460c74d6cd596edd7430009240/src/azure-cli-core/azure/cli/core/__init__.py#L344

If you add a breakpoint here:

https://github.com/microsoft/knack/blob/fe3bf5d3a79a3dd2ce5ddb0c38d93843a3380f6f/knack/commands.py#L105

```py
    def load_arguments(self):
        if self.arguments_loader:
            cmd_args = self.arguments_loader()
            if self.confirmation:
                cmd_args.append(('yes',
                                 CLICommandArgument(dest='yes', options_list=['--yes', '-y'],
                                                    action='store_true', help='Do not prompt for confirmation.')))
            self.arguments.update(cmd_args)
```

you will see the breakpoint being hit twice when calling `storage account delete --debug`.

Other command modules like `network` don't have these 2 redundant invocations:

https://github.com/Azure/azure-cli/blob/38c2930469af4f00a4a352eb9bd9db4f6c0e70de/src/azure-cli/azure/cli/command_modules/network/__init__.py#L32-L39


yonzhan:
add to S171

src/azure-cli-core/azure/cli/core/tests/test_command_registration.py
```diff
@@ -170,7 +170,6 @@ def _mock_load_command_loader(loader, args, name, prefix):
         class TestCommandsLoader(AzCommandsLoader):
 
             def load_command_table(self, args):
-                super(TestCommandsLoader, self).load_command_table(args)
```
jiasli, line 173: @jsntcy Nice catch!

jsntcy: Looks like there are still other places using super.load_command_table(),could you search and remove them as well?

---
In reply to: [459828026](https://github.com/Azure/azure-cli/pull/13879#discussion_r459828026) [](ancestors = 459828026)

jiasli: This PR is only for removing the extra debug log entries. We can handle them when time permits.

PR #13899, opened 2020-06-09
[Cloud] Cloud endpoint discovery

fengzhou-msft:
**Description**  

This PR adds logic to get cloud endpoints in the order of a local file (`~/.azure/cloudEndpoints.json`), environment variable `ARM_CLOUD_METADATA_URL`, a fixed DNS address (commented as not ready), hard-coded list in public clouds or throw error in air-gapped clouds. 

`az cloud import --endpoint` is added to support manually import the endpoints from a URL and should be used to set up endpoints in an air-gapped cloud for the first time. It can also be used to refresh endpoints.

This PR also makes use of a local `versionCheck.json` file. When an update is detected by comparing the current core version and the core version stored in the file, endpoints will be refreshed by cleaning the endpoints data in the `cloudEndpoints.json` first. It will use the `metadata_url` stored in `cloudEndpoints.json` to fetch the endpoints.

**Testing Guide**  

**History Notes**  

[Cloud] Add `az cloud import` command.  

---


src/azure-cli/azure/cli/__main__.py
```diff
@@ -29,6 +29,19 @@
...
+        CLOUD_ENDPOINTS.load(CLOUD_ENDPOINTS_FILE)
+        CLOUD_ENDPOINTS['clouds'] = {}
+        if len(sys.argv) == 3:
+            CLOUD_ENDPOINTS['metadata_url'] = ''
+        elif sys.argv[3] == '--endpoint':
+            CLOUD_ENDPOINTS['metadata_url'] = sys.argv[4]
+except Exception:  # pylint: disable=broad-except
+    pass
 
 az_cli = get_default_cli()
```
fengzhou-msft, line 46: Cloud information is retrieved when init `AzCli` in `get_default_cli`. So we need to process the `metadata_url` before calling `get_default_cli`.

src/azure-cli/azure/cli/__main__.py
```diff
@@ -30,6 +30,19 @@ def cli_main(cli, args):
     return cli.invoke(args)
 
 
+try:
+    if sys.argv[1] == 'cloud' and sys.argv[2] == 'import':
```
qianwens, line 34: I don't understand why you have to load the CLOUD_ENDPOINTS here. why it is required when the command is `az cloud import`, why not other command

src/azure-cli/azure/cli/command_modules/cloud/custom.py
```diff
@@ -148,3 +148,8 @@ def list_profiles(cmd, cloud_name=None, show_all=False):
     if not cloud_name:
         cloud_name = get_active_cloud_name(cmd.cli_ctx.cloud)
     return list(API_PROFILES)
+
+
+def import_cloud_endpoints(cmd, endpoint=None):
```
qianwens, line 153: where is the parameter endpoint used?

src/azure-cli-core/azure/cli/core/cloud.py
```diff
@@ -163,13 +172,62 @@ def _get_microsoft_graph_resource_id(cloud_name):
     return graph_endpoint_mapper.get(cloud_name, None)
 
 
+def _get_database_server_endpoint(db_prefix, sql_server_hostname, cloud_name):
+    if cloud_name == 'AzureCloud':
+        return db_prefix + '.database.azure.com'
```
qianwens, line 177: why do you need different logic for AzureCloud? why not just add '.database.azure.com' as its hostname in AzureCloud

src/azure-cli-core/azure/cli/core/cloud.py
```diff
@@ -334,18 +402,58 @@ def __str__(self):
         postgresql_server_endpoint='.postgres.database.cloudapi.de',
         mariadb_server_endpoint='.mariadb.database.cloudapi.de'))
 
-KNOWN_CLOUDS = [AZURE_PUBLIC_CLOUD, AZURE_CHINA_CLOUD, AZURE_US_GOV_CLOUD, AZURE_GERMAN_CLOUD]
+KNOWN_CLOUDS = []
```
qianwens, line 405: I think it is better to use delegate function here like KNOWN_CLOUDS = get_known_clouds(), the logic looks complicated to update `global KNOWN_CLOUDS` in get_know_clouds while you still have a public variable KNOWN_CLOUDS, it is easy to make mistake

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -982,3 +976,18 @@ def get_linux_distro():
             release_info[k.lower()] = v.strip('"')
 
     return release_info.get('name', None), release_info.get('version_id', None)
+
+
+def handle_version_update():
```
qianwens, line 981: please add comments for this method, it is not easy to understand from the name

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -982,3 +976,18 @@ def get_linux_distro():
...
+    try:
+        from azure.cli.core._session import VERSIONS, CLOUD_ENDPOINTS
+        from distutils.version import LooseVersion  # pylint: disable=import-error,no-name-in-module
+        from azure.cli.core import __version__
+        if not VERSIONS['versions']:
+            get_cached_latest_versions()
```
qianwens, line 987: the logic is confusing here, what is the object of this function, does it want to refresh the version to latest version or reset the version to empty? If you want to handle some file refresh based on the version, it is better not make it work by some tricky logic

src/azure-cli-core/azure/cli/core/cloud.py
```diff
@@ -334,18 +402,58 @@ def __str__(self):
...
-        arm_cloud_dict = json.loads(urlretrieve(os.getenv('ARM_CLOUD_METADATA_URL')))
-        cli_cloud_dict = _convert_arm_to_cli(arm_cloud_dict)
-        if 'AzureCloud' in cli_cloud_dict:
-            cli_cloud_dict['AzureCloud'].endpoints.active_directory = 'https://login.microsoftonline.com'  # pylint: disable=line-too-long # change once active_directory is fixed in ARM for the public cloud
-        KNOWN_CLOUDS = list(cli_cloud_dict.values())
-    except Exception as ex:  # pylint: disable=broad-except
-        logger.warning('Failed to load cloud metadata from the url specified by ARM_CLOUD_METADATA_URL')
-        raise ex
+
+def get_known_clouds():
```
qianwens, line 408: please add some function tests for this method.

src/azure-cli-core/azure/cli/core/cloud.py
```diff
@@ -226,6 +285,15 @@ def __str__(self):
         }
         return pformat(o)
 
+    def toJSON(self):
+        return {'name': self.name, "endpoints": self.endpoints.__dict__, "suffixes": self.suffixes.__dict__}
+
+    @classmethod
+    def fromJSON(cls, json_str):
```
qianwens, line 292: what if the json file corrupt, please give a proper error message to help the user to recover.

src/azure-cli-core/azure/cli/core/cloud.py
```diff
@@ -71,7 +76,8 @@ def __init__(self,
                  ossrdbms_resource_id=None,
                  log_analytics_resource_id=None,
                  app_insights_resource_id=None,
-                 app_insights_telemetry_channel_resource_id=None):
+                 app_insights_telemetry_channel_resource_id=None,
+                 **entries):
```
qianwens, line 80: add comments for **entries

src/azure-cli/azure/cli/__main__.py
fengzhou-msft: This is mainly for use in air-gapped clouds. Users need to provide an URL and use `az cloud import --endpoint` to first fetch the endpoints before they can use azure cli. When they run other commands, an error will be thrown because no cloud endpoints are set unless they have set an environment variable for the url which is another way to provide the endpoint url.

src/azure-cli/azure/cli/command_modules/cloud/custom.py
fengzhou-msft: It's set in `__main__.py` and used in `cloud.py`, then the command execution reaches here, but the endpoints are actually already imported in AzCli init process.

src/azure-cli-core/azure/cli/core/cloud.py
fengzhou-msft: The data is based on the endpoints in https://management.azure.com/metadata/endpoints?api-version=2019-05-01 and `AzureCloud` uses "database.windows.net" for `sqlServerHostname`(may be outdated), need to change to `.database.azure.com`.

We'll need to update the data in the url to remove this patch.

src/azure-cli-core/azure/cli/core/util.py
fengzhou-msft: The purpose is to detect that azure cli has updated to a new version, then we can invalidate some cached files like version and cloudEndpoints by setting to empty. These files will be refreshed if they are found to be empty in a get operation. No need to refresh them here in advance.

yonzhan:
Cloud endpoint discovery

src/azure-cli-core/azure/cli/core/util.py
qianwens: should we update the cloudendpoints when CLI upgrades to a new version? It seems cloudendpoints refresh has no relationship with CLI version.

src/azure-cli/azure/cli/__main__.py
qianwens: Can we refactor the code to defer the error throw instead of adding the check here

src/azure-cli/azure/cli/command_modules/cloud/custom.py
qianwens: I don't see where this parameter used in this method

src/azure-cli-core/azure/cli/core/util.py
fengzhou-msft: cloudendpoints update is not necessarily related to CLI. It could add some endpoints but not used in CLI. But when a new endpoint is also added in CLI, the new version of CLI needs to use the latest endpoints.

PR #13900, opened 2020-06-09
{Core} Add storage sync endpoint suffix

houk-ms:
**Description**  

This PR adds the storage sync endpoint suffix for AzureCloud and AzureUSGovernment. 

It will use the property name `storageSyncEndpointSuffix` to find the suffix in ARM cloud discovery, and it will use the suffix of AzureCloud as default if couldn't find it.

**Testing Guide**  

Run the command below, the PR will make the suffix `storageSyncEndpoint` appear.
`az cloud show -n AzureCloud`


src/azure-cli-core/azure/cli/core/cloud.py
```diff
@@ -304,6 +308,7 @@ def __str__(self):
         app_insights_telemetry_channel_resource_id='https://dc.applicationinsights.us/v2/track'),
     suffixes=CloudSuffixes(
         storage_endpoint='core.usgovcloudapi.net',
+        storage_sync_endpoint='afs.azure.net',
```
qianwens, line 311: The endpoint is the same for AzureUsGovernment and AzureCloud?

houk-ms: Thanks @qianwens for pointing it out. My negligence, updated.

yonzhan:
storage sync endpoint suffix

src/azure-cli-core/azure/cli/core/cloud.py
```diff
@@ -191,6 +193,7 @@ def _arm_to_cli_mapper(arm_dict):
             log_analytics_resource_id=arm_dict['logAnalyticsResourceId'] if 'logAnalyticsResourceId' in arm_dict else None),  # pylint: disable=line-too-long
         suffixes=CloudSuffixes(
             storage_endpoint=arm_dict['suffixes']['storage'],
+            storage_sync_endpoint=arm_dict['suffix']['storageSyncEndpointSuffix'] if 'storageSyncEndpointSuffix' in arm_dict['suffixes'] else 'afs.azure.net',  # pylint: disable=line-too-long
```
fengzhou-msft, line 196:
```suggestion
            storage_sync_endpoint=arm_dict['suffix']['storageSyncEndpointSuffix'] if 'storageSyncEndpointSuffix' in arm_dict['suffixes'] else _get_storage_sync_endpoint(arm_dict['name']),  # pylint: disable=line-too-long
```
Can you add a method similar to `_get_microsoft_graph_resource_id` to return 'afs.azure.us' for US cloud and 'afs.azure.net' for other clouds?

houk-ms: Sure, updated.

src/azure-cli-core/azure/cli/core/cloud.py
```diff
@@ -163,6 +165,14 @@ def _get_microsoft_graph_resource_id(cloud_name):
     return graph_endpoint_mapper.get(cloud_name, None)
 
 
+def _get_storage_sync_endpoint(cloud_name):
+    storage_sync_endpoint_mapper = {
+        'AzureCloud': 'afs.azure.net',
+        'AzureUSGovernment': 'afs.azure.us',
```
jiasli, line 171: I think this is a duplication of https://github.com/Azure/azure-cli/blob/57596cdcf3463ae8c6fe5879f5f3ca95ba914711/src/azure-cli-core/azure/cli/core/cloud.py#L319

Maintaining 1 string in 2 places is not a good programming practice.

houk-ms: It's a good suggestion. While I find two other functions in the file are also using the same way right now, so I'd prefer we stay the same code sytle until we decide to change all of them.

PR #13939, opened 2020-06-11
[App Service] Fix: Surface error when invalid webapp name provided 

### LINKED ISSUE #11884: az webapp up provides misleading information when invalid name is provided
## Describe the bug

**Command Name**
`az webapp up`

**Errors:**
```
Unable to retrieve details of the existing app mpae_webapp. Please check that the app is a part of the current subscription
```

## To Reproduce:

- `az webapp up --resource-group {} --name {} --html`

## Expected Behavior
The error that is captured is incomplete, on the screen it gives

```
Webapp random_webapp already exists. The command will deploy contents to the existing app.
Unable to retrieve details of the existing app random_webapp. Please check that the app is a part of the current subscription
```

The first portion of that output is incorrect and misleading.  The error seems to be that underscore is not allowed in the name.  However, the error says that the instance exists instead.

## Environment Summary
```
Darwin-19.2.0-x86_64-i386-64bit
Python 3.7.4
Shell: bash

azure-cli 2.0.80
azure-cli-iot-ext 0.8.3

Extensions:
azure-cli-iot-ext 0.8.3

```

calvinsID:
**Description**  
Closes #11884 

Show descriptive error when providing an invalid name in `az webapp up -n appName`.
Currently, providing an invalid name (for example, a name with an underscore) shows the error 'Unable to retrieve details of the existing app {invalid_name}. Please check that the app is a part of the current subscription', even though there is no existing app. 

**Testing Guide**  
Run `az webapp up -n invalid_name` in the directory of a test app you want to deploy. The error should now be more descriptive of the problem, and show 'Hostname must be less than 64 chars and must follow allowed character guidelines at http://tools.ietf.org/html/rfc952'. 

All other functionality of the command should stay the same. 

**History Notes**  

---


panchagnula:
That was quick :) what is the error message we show in this case?

panchagnula:
@btardif FYI

calvinsID:
It returns the error message that the name validation API gives, which is  'Hostname must be less than 64 chars and must follow allowed character guidelines at http://tools.ietf.org/html/rfc952'

btardif:
Linking to the RFC is not going to be user friendly can we trim it for now to be:
"Hostname must be less than 64 chars and must follow allowed character guidelines"

I don't think we have a doc covering valid hostname characters and I don't think we support the RFC to the full extent.

yonzhan:
App Service

panchagnula:
@qwordy could you review this please?

Issue #11431, opened 2019-11-29
Could we remove this functional token?


mmyyrroonn:
**Is your feature request related to a problem? Please describe.**
Looks like this token is still valid and we should remove it.

haroldrandom:
add it to S163

MyronFanQiu:
@haroldrandom Please take a look at these two lines https://github.com/Azure/azure-cli/blob/807faccca1009e75e41c663d56b6d3171e630484/CredScanSuppressions.json#L285 and https://github.com/Azure/azure-cli/blob/807faccca1009e75e41c663d56b6d3171e630484/CredScanSuppressions.json#L293

haroldrandom:
I think it should own by service team. The accessToken is used in test_ams_live_event_scenarios.py is hard coded. Should be get online through command. 
Added "Service Attension" for help.

yonzhan:
@fredericky @akucer @yulei, could you please help with above hard coded accessToken issue from Media Services team?

haroldrandom:
Add an example https://github.com/Azure/azure-cli/pull/12589#issuecomment-599158870

hivyas:
The accessToken is used in test_ams_live_event_scenarios.py is actually named poorly because it is a GUID and not an actual security token. I have updated the code to use a generated GUID instead of a hardcoded one in order to avoid confusion: https://github.com/Azure/azure-cli/pull/13948.

The second hardcoded string is found in the container module and AMS does not own that.

haroldrandom:
Add one related issue https://github.com/Azure/azure-cli-dev-tools/issues/215

akucer:
@haroldrandom - I removed the Media Services label and assignment to AMS team members as AMS issue has been fixed. As per previous comment the other hardcoded string is not owned by AMS and needs assignment to appropriate team / owner. Thanks.

PR #13948, opened 2020-06-12
[AMS] Change test case to use a generated GUID

hivyas:
**Description**  

Changing a variable in the Live Events test case from a hardcoded GUID to a generated one due to this issue: https://github.com/Azure/azure-cli/issues/11431
**Testing Guide**  

Run `azdev test AmsLiveEventTests`
**History Notes**  

---


yonzhan:
AMS

qwordy (dismissed):

yungezz:
hi @hivyas could you pls look at CI failure then we can go ahead to merge the PR? @qwordy

hivyas:
Hi there, there is a bug in the code that we are currently unable to solve so we might have to revert these changes to the original. Will try to resolve this PR by the end of the week. Thanks @yungezz

haroldrandom:
Try rebase with the latest dev branch. You branch is way too behind. Maybe the result after merging is not correct.

And I think the key point is that `self.create_guid` is a dummy GUID generator. And it's actually an instance-scope counter increased by one every time it's invoked.

https://github.com/Azure/azure-cli/blob/bbed0711093b5d1ccabafc3449e628d84e5fa396/src/azure-cli-testsdk/azure/cli/testsdk/base.py#L150-L163

So the value would be `88888888000000000000000000000002`. 
However, from the YAML file I can see the token value from response is `3cc2274785b346549072ad1b9686e5b3` instead of the one you insert. 

I am not sure whether it's a mistake from servier that I can't get what I insert or the token value is designed to not be inserted.

haroldrandom:
The problem is in live-mode or non-live-mode. Please take a closer look of the `create_guid` function

Issue #11031, opened 2019-10-30
Clarification of --scopes parameter for `az ad sp create-for-rbac --skip-assignment ...`


BenjaminHerbert:
Could you explain the meaning of the `--scopes` parameter for the command 
`az ad sp create-for-rbac --skip-assignment ...`.

The help says:

> --scopes          : Space-separated list of scopes the service principal's role assignment applies to. Defaults to the root of the current subscription.

As I have no assignment due to `--skip-assignment`, there is no assignment yet...

What does `--scopes` do? Does it limit the scope where a role can be assigned to the service principal?
```
az role assignment create --assignee $ASSIGNEE --scope $SCOPE --role $ROLE
```

Or does it just do nothing in the `--skip-assignment` case? 

I'd be glad if you could clarify this and add some more information to the documentation regarding `--scopes`.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: e99d25ff-03e2-e015-37f4-a29b59ec50d4
* Version Independent ID: 1a8e88ed-72c7-fbe7-b0ad-68f23b3946cd
* Content: [az ad sp](https://docs.microsoft.com/en-us/cli/azure/ad/sp?view=azure-cli-latest#az-ad-sp-create-for-rbac)
* Content Source: [src/azure-cli/azure/cli/command_modules/role/_help.py](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/role/_help.py)
* Service: **active-directory**
* GitHub Login: @rloutlaw
* Microsoft Alias: **routlaw**

yonzhan:
@jiasli add to S162.

jiasli:
@BenjaminHerbert, thanks for this nice feedback. Please https://github.com/Azure/azure-cli/issues/10724#issuecomment-539345221 where I described what a scope is. 

I will improve the help message and make it clearer.

yonzhan:
add to S171

PR #13977, opened 2020-06-16
{RBAC} Improve help message for `az ad sp create-for-rbac`

jiasli:
**Description**  
Close https://github.com/Azure/azure-cli/issues/11031

Improve help message for `az ad sp create-for-rbac` and  `--scopes` parameter:

```
> az ad sp create-for-rbac -h

...

        When --skip-assignment is specified, --scopes will be ignored. You may use `az role
        assignment create` to create role assignments for this service principal later.

Arguments
    ...
    --scopes           : Space-separated list of scopes the service principal's role assignment
                         applies to. Defaults to the root of the current subscription. e.g.,
                         /subscriptions/0b1f6471-1bf0-4dda-aec3-111122223333,
                         /subscriptions/0b1f6471-1bf0-4dda-aec3-111122223333/resourceGroups/myGroup,
                         or /subscriptions/0b1f6471-1bf0-4dda-aec3-111122223333/resourceGroups/myGro
                         up/providers/Microsoft.Compute/virtualMachines/myVM.
    --skip-assignment  : Skip creating the default assignment, which allows the service principal to
                         access resources under the current subscription.  Allowed values: false,
                         true.
```

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/role/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/role/_help.py
@@ -353,6 +353,9 @@ examples:
 helps['ad sp create-for-rbac'] = """
 type: command
 short-summary: Create a service principal and configure its access to Azure resources.
+long-summary: >
+    When --skip-assignment is specified, --scopes will be ignored. You may use `az role assignment create` to create
+    role assignments for this service principal later.
 parameters:
   - name: --name -n
     short-summary: a URI to use as the logic name. It doesn't need to exist. If not present, CLI will generate one.
@@ -369,7 +372,9 @@ parameters:
   - name: --scopes
     short-summary: >
         Space-separated list of scopes the service principal's role assignment applies to.
-        Defaults to the root of the current subscription.
+        Defaults to the root of the current subscription. e.g., /subscriptions/0b1f6471-1bf0-4dda-aec3-111122223333,
+        /subscriptions/0b1f6471-1bf0-4dda-aec3-111122223333/resourceGroups/myGroup, or
+        /subscriptions/0b1f6471-1bf0-4dda-aec3-111122223333/resourceGroups/myGroup/providers/Microsoft.Compute/virtualMachines/myVM
   - name: --role
     short-summary: Role of the service principal.
 examples:
--- a/src/azure-cli/azure/cli/command_modules/role/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/role/_help.py
@@ -353,9 +353,6 @@ examples:
 helps['ad sp create-for-rbac'] = """
 type: command
 short-summary: Create a service principal and configure its access to Azure resources.
-long-summary: >
-    When --skip-assignment is specified, --scopes will be ignored. You may use `az role assignment create` to create
-    role assignments for this service principal later.
 parameters:
   - name: --name -n
     short-summary: a URI to use as the logic name. It doesn't need to exist. If not present, CLI will generate one.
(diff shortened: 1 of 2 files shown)


yonzhan:
add to S171

src/azure-cli/azure/cli/command_modules/role/_help.py
```diff
@@ -353,6 +353,9 @@
 helps['ad sp create-for-rbac'] = """
 type: command
 short-summary: Create a service principal and configure its access to Azure resources.
+long-summary: >
+    When --skip-assignment is specified, --scopes will be ignored. You may use `az role assignment create` to create
+    role assignments for this service principal later.
```
qianwens, line 358: why not add this message in --skip-assignment parameter

jiasli: Because I think this message involves 2 parameters `--skip-assignment` and `--scopes`. In my mind, it is better to keep the message in a common place.

qianwens: I think you can add this message in `--skip-assignment` because it only impact when user set this parameter

---
In reply to: [441289495](https://github.com/Azure/azure-cli/pull/13977#discussion_r441289495) [](ancestors = 441289495)

jiasli: In such case, the help for `--scopes` won't indicate it will be ignored by `--skip-assignment`. It will be clearer to put cross-parameter help message in one place. Besides, it appears in `long-summary` and saves the help from being crowded.

arrownj: My personal experience is that I prefer to put parameter related information in the parameter help message. Because sometimes there may be too many parameters and we can't display all them in one screen, the user may not notice this information when he try to check the parameter specific message.

If it is a common pattern which related lots of parameters and will display ugly in each parameter, I think it's OK to put it in long summary.

For this case, I prefer to put this message in `--skip-assignment`. Sorry for not realize this before.

jiasli: To summarize, the debate is **where to put a help message involving multiple parameters**:

> When --skip-assignment is specified, --scopes will be ignored. You may use `az role assignment create` to create role assignments for this service principal later.

jiasli: Option 1: Put the help message in command help's `long-summary` section

jiasli: Option 2: Put the help message in any of the parameter, like `--skip-assignment` in this case, or both `--skip-assignment` and `--scope`.

qwordy: Do not use long summary. It is not shown in CLI reference. And docs teams does not plan to fix it.

qwordy: Put it in both places.

jiasli: @qwordy, this is tricky, and makes the `long-summary` totally broken doc-wise. 

Besides, I already applied option 1 in `az rest`:

```
> az rest -h

Command
    az rest : Invoke a custom request.
        This command automatically authenticates using the logged-in credential: If Authorization
        header is not set, it attaches header `Authorization: Bearer <token>`, where `<token>` is
        retrieved from AAD. The target resource of the token is derived from --url if --url starts
        with an endpoint from `az cloud show --query endpoints`. You may also use --resource for a
        custom resource.
        If Content-Type header is not set and --body is a valid JSON string, Content-Type header
        will default to application/json.
```

jiasli: > Another option: Put it in both places.

This introduces duplication, which IMO is bad.

jiasli: Moved the message to `--skip-assignment`:

```
> az ad sp create-for-rbac -h

Arguments
    ...
    --skip-assignment  : Skip creating the default assignment, which allows the service principal to
                         access resources under the current subscription. When specified, --scopes
                         will be ignored. You may use `az role assignment create` to create role
                         assignments for this service principal later.  Allowed values: false, true.
```

PR #13996, opened 2020-06-17
[AKS] Change to support 0 node agent pool and block manual scale for CAS enabled pool

xhl873:
**Description**  

1. allow scale nodepool to zero count
2. If autoscaler is enabled for a nodepool, manual scale will be blocked.

**Testing Guide**  

Create a user nodepool without autoscaler enabled, scale it to 0 node using the following command should be supported.
az aks nodepool scale --cluster-name <cluster name>--name <nodepool name>--resource-group <rg> --node-count 0
Enable autoscaler for that nodepool, use the command above, and changed the node-count to a different number, the manual should be blocked by cli.

**History Notes**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acs/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/custom.py
@@ -2098,10 +2098,11 @@ def aks_scale(cmd, client, resource_group_name, name, node_count, nodepool_name=
         raise CLIError('There are more than one node pool in the cluster. '
                        'Please specify nodepool name or use az aks nodepool command to scale node pool')
 
-    if node_count == 0:
-        raise CLIError("Can't scale down to 0 nodes.")
     for agent_profile in instance.agent_pool_profiles:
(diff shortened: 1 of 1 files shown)


src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -2098,10 +2098,11 @@ def aks_scale(cmd, client, resource_group_name, name, node_count, nodepool_name=
         raise CLIError('There are more than one node pool in the cluster. '
                        'Please specify nodepool name or use az aks nodepool command to scale node pool')
 
-    if node_count == 0:
```
paulgmiller, line 2101: I wonder if we already validate node_count >= 0 somewhere so you can't pass negative numbers.

xhl873: AgentPool validation will limit the count between 0 to 100, negative number won't be allowed. :)

yonzhan:
AKS

PR #13997, opened 2020-06-17
[AppService] Improved error messages for SSH Tunnel Connection

sanchitmehta:
**Description**  

This PR improves the error message(s) and adds fail fast scenarios for the customers using Azure CLI to SSH to their App Service on Linux instances. We leverage the new version of the status API to determine if an app container exposes the port 2222 and and provide fail fast errors with links to the documentation for customers who have incorrectly configured the SSH setup on their custom docker images.  

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -3467,7 +3467,9 @@ def _wait_for_webapp(tunnel_server):
         if tries == 0:
             logger.warning('Connection is not ready yet, please wait')
         if tries == 60:
-            raise CLIError("SSH timeout, your app must be running before it can accept SSH connections. Use `az webapp log tail` to review the app startup logs.")
+            raise CLIError('SSH timeout, your app must be running before'
+                           ' it can accept SSH connections. '
+                           'Use `az webapp log tail` to review the app startup logs.')
         tries = tries + 1
         logger.warning('.')
         time.sleep(1)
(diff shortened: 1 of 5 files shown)


sanchitmehta:
cc: @panchagnula @btardif

src/azure-cli/azure/cli/command_modules/appservice/tunnel.py
```diff
@@ -91,23 +92,32 @@ def is_webapp_up(self):
...
-        logger.info('Status response message: %s', msg)
-        if 'FAIL' in msg.upper():
-            logger.info('WARNING - Remote debugging may not be setup properly. Reponse content: %s', msg)
-            return False
-        if 'SUCCESS' in msg.upper():
-            return True
+        resp_msg = r.read().decode('utf-8')
+        json_data = json.loads(resp_msg)
+
+        if 'STARTED' in json_data["state"].upper():
```
qwordy, line 111: Is it possible the key is not found

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -3467,7 +3467,7 @@ def _wait_for_webapp(tunnel_server):
         if tries == 0:
             logger.warning('Connection is not ready yet, please wait')
         if tries == 60:
-            raise CLIError("Timeout Error, Unable to establish a connection")
+            raise CLIError("SSH timeout, your app must be running before it can accept SSH connections. Use `az webapp log tail` to review the app startup logs.")
```
qwordy, line 3470: I don't think it's a timeout error. Maybe "Exceed maximum retry times"

src/azure-cli/azure/cli/command_modules/appservice/tunnel.py
sanchitmehta: Added a check if key is not found, returning false from this method would invoke the retry logic

src/azure-cli/azure/cli/command_modules/appservice/custom.py
sanchitmehta: This is just polling for the app status (at 1 second intervals), before making an ssh tunnel connection. This message would be useful for users to understand what went wrong

PR #14038, opened 2020-06-19
[ACR] Handle log artifact link from Registry to stream logs

savaradh:
Handle new parameter logArtifactLink when streaming logs.
This new parameter will point to a registry artifact which contains the task run logs

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acr/_stream_utils.py
+++ b/src/azure-cli/azure/cli/command_modules/acr/_stream_utils.py
@@ -202,7 +202,7 @@ def _stream_logs(no_format,  # pylint: disable=too-many-locals, too-many-stateme
             raise CLIError("Run was canceled")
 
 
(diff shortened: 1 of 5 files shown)


src/azure-cli/azure/cli/command_modules/acr/_stream_utils.py
```diff
@@ -191,6 +202,24 @@ def _stream_logs(no_format,  # pylint: disable=too-many-locals, too-many-stateme
             raise CLIError("Run was canceled")
 
 
+def _stream_artifact_logs(log_file_sas,  # pylint: disable=too-many-locals, too-many-statements, too-many-branches
```
northtyphoon, line 205: does the following code trigger the lint warnings?

savaradh: i dont think so. will disable it.

yonzhan:
ACR

PR #14042, opened 2020-06-22
[Compute] BREAKING CHANGE: `az vmss create`: Make parameter single-placement-group default to false when creating a vmss

houk-ms:
**Description**  

This PR makes the `single-placement-group` parameter default to **false** when creating a vmss, which is required by the service team to better facilitate the vmss placement.

**Testing Guide**  

1. Create a vmss with the `single-placement-group` parameter not been set, e.g, with the command below
`az vmss create -n houk-vmss-spg -g houk-test --image Centos`

2. After creating the vmss, show the status of `single-placement-group` parameter by
`az vmss show -g houk-test -n houk-vmss-spg --query 'singlePlacementGroup'`

**Before**
You will get **true**.

**Now**
You will get  **false**.


yonzhan:
Compute

qwordy:
The original default value is null. Does the service set it true when it is not set explicitly?

houk-ms:
> The original default value is null. Does the service set it true when it is not set explicitly?

Yes

src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -514,7 +514,7 @@ def load_arguments(self, _):
         c.argument('name', name_arg_type)
         c.argument('nat_backend_port', default=None, help='Backend port to open with NAT rules. Defaults to 22 on Linux and 3389 on Windows.')
         c.argument('single_placement_group', arg_type=get_three_state_flag(), help="Limit the scale set to a single placement group."
-                   " See https://docs.microsoft.com/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-placement-groups for details")
+                   " See https://docs.microsoft.com/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-placement-groups for details, default to false", default=False)
```
fengzhou-msft, line 517:
```suggestion
                   " See https://docs.microsoft.com/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-placement-groups for details. Default to false", default=False)
```

PR #14068, opened 2020-06-23
[AKS] az aks create: add --enable-aad argument

bingosummer:
**Description**  

This PR is to portal the support of creating AKS-managed Azure AD cluster from aks-preview extension. It adds a new argument `--enable-aad` to `az aks create` command. Customers can specify the `--aad-admin-group-object-ids` and `--aad-tenant-id` when creating/updating AKS cluster.

**Testing Guide**  

* Create an AKS-managed Azure AD cluster

```
az aks create -g MyResourceGroup -n MyManagedCluster --enable-aad [--aad-admin-group-object-ids <id>] [--aad-tenant-id <id>]
```

* Update an AKS-managed Azure AD cluster

```
az aks update -g MyResourceGroup -n MyManagedCluster [--aad-admin-group-object-ids <id>] [--aad-tenant-id <id>]
```

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acs/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/_help.py
@@ -385,6 +385,8 @@ examples:
     text: az aks create -g MyResourceGroup -n MyManagedCluster --outbound-type userDefinedRouting --load-balancer-sku standard --vnet-subnet-id customUserSubnetVnetID
   - name: Create a kubernetes cluster with supporting Windows agent pools.
     text: az aks create -g MyResourceGroup -n MyManagedCluster --load-balancer-sku Standard --vm-set-type VirtualMachineScaleSet --network-plugin azure --windows-admin-username azure --windows-admin-password 'replacePassword1234$'
+  - name: Create a kubernetes cluster with managed AAD enabled.
+    text: az aks create -g MyResourceGroup -n MyManagedCluster --enable-aad --aad-admin-group-object-ids <id-1,id-2> --aad-tenant-id <id>
 """
 
 helps['aks update'] = """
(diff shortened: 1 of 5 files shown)


yonzhan:
AKS

src/azure-cli/azure/cli/command_modules/acs/_help.py
```diff
@@ -222,6 +222,12 @@
   - name: --windows-admin-password
     type: string
     short-summary: Password to create on Windows node VMs.
+  - name: --enable-aad
+    type: bool
+    short-summary: Enable managed AAD feature for cluster.
+  - name: --aad-admin-group-object-ids
+    type: string
+    short-summary: Comma seperated list of aad group object IDs that will be set as cluster admin.
```
arrownj, line 230: Usually we use space separated for this case. And add `nargs='+'` for this argument in `_params.py`. There are so many examples in our code base, take [this ](https://github.com/Azure/azure-cli/blob/b5937aa49d572fe93e02bb2d30fa6b1cc575d224/src/azure-cli/azure/cli/command_modules/acr/_params.py#L144)for example.

If there is no special requirement for this parameter, I strongly suggest to follow above pattern. With that, we do not need to handle the parse logic ourselves.

bingosummer: Thanks for reviewing the PR. In azure-cli-extension, we use the comma separated list for this argument. Is it OK that we use space separated for azure-cli and comma separated for azure-cli-extension?

arrownj: If they are the same parameter, it's better to keep them the same.

PR #14077, opened 2020-06-23
{Telemetry} Fix for file move in cloud shell

fengzhou-msft:
**Description**  

When sending telemetry notes to application insight on Cloud Shell, the following error occurs in the step of moving `cache.x` file to a tmp directory:

> telemetry.records : Fail to move file from /home/feng/.azure/telemetry/cache.1 to /tmp/tmpdj9u1r70/cache.1. Reason: [Errno 18] Invalid cross-device link: '/home/feng/.azure/telemetry/cache.1' -> '/tmp/tmpdj9u1r70/cache.1'.

`os.rename` only works if source and destination are on the same file system which is not the case in Cloud Shell. [`shutil.move`](https://docs.python.org/3/library/shutil.html#shutil.move) can handle it well by copying src to dst using copy_function and then remove src.

In addition, I added Cloud Shell info to feedback and telemetry. 

**Testing Guide**  

**History Notes**  

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/telemetry.py
+++ b/src/azure-cli-core/azure/cli/core/telemetry.py
@@ -437,7 +437,7 @@ def _get_azure_subscription_id():
 
 
 def _get_shell_type():
-    ## This method is not accurate and needs improvement, for instance all shells on Windows return 'cmd'.
+    # This method is not accurate and needs improvement, for instance all shells on Windows return 'cmd'.
     if 'ZSH_VERSION' in os.environ:
         return 'zsh'
     if 'BASH_VERSION' in os.environ:
--- a/src/azure-cli-core/azure/cli/core/telemetry.py
+++ b/src/azure-cli-core/azure/cli/core/telemetry.py
@@ -437,6 +437,7 @@ def _get_azure_subscription_id():
 
 
 def _get_shell_type():
+    ## This method is not accurate and needs improvement, for instance all shells on Windows return 'cmd'.
     if 'ZSH_VERSION' in os.environ:
         return 'zsh'
     if 'BASH_VERSION' in os.environ:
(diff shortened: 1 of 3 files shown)


yonzhan:
Telemetry

src/azure-cli-telemetry/azure/cli/telemetry/components/records_collection.py
```diff
@@ -52,7 +52,7 @@ def snapshot_and_read(self):
             if stat.S_ISREG(each[1].st_mode):
                 try:
                     # Platform question: if this op is atom
-                    os.rename(os.path.join(folder, each[0]), os.path.join(tmp, each[0]))
+                    shutil.move(os.path.join(folder, each[0]), os.path.join(tmp, each[0]))
```
yungezz, line 55: this is an existing external dependency right, just want to double confirm

fengzhou-msft: It's from Python library and already imported in this file.

jiasli:
BTW, here is the doc from Python:

https://docs.python.org/3/library/os.html#os.rename

> `os.rename(src, dst, *, src_dir_fd=None, dst_dir_fd=None)`
>
> The operation may fail on some Unix flavors if src and dst are on different filesystems.

Also see https://stackoverflow.com/questions/11578443/trigger-io-errno-18-cross-device-link, https://stackoverflow.com/questions/42392600/oserror-errno-18-invalid-cross-device-link

src/azure-cli-core/azure/cli/core/telemetry.py
```diff
@@ -437,6 +437,7 @@ def _get_azure_subscription_id():
 
 
 def _get_shell_type():
+    ## This method is not accurate and needs improvement, for instance all shells on Windows return 'cmd'.
```
fengzhou-msft, line 440:
```suggestion
    # This method is not accurate and needs improvement, for instance all shells on Windows return 'cmd'.
```

PR #14097, opened 2020-06-24
[Rdbms] Add Infrastructure Encryption for Azure Postgres and MySQL

shinilm:
**Description**  

We are currently adding a new capability in our platform Azure Database for MySQL/Azure Database for PostgreSQL called infrastructure encryption. Currently,  Azure Database for MySQL/Azure Database for PostgreSQL are always encrypted at rest using Microsoft’s managed keys. This new optional feature adds a second layer of encryption using 2nd encryption algorithm which protects the data from compromise of the first algorithm and can have performance impact on the workloads

**Testing Guide**  
az postgres server update –name <server name>  -g <resoure_group> --infrastructure-encryption Enabled
az mysql server update --name  <server name>  -g <resoure_group> --infrastructure-encryption Enabled

**History Notes**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/rdbms/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/rdbms/_params.py
@@ -32,7 +32,7 @@ def load_arguments(self, _):    # pylint: disable=too-many-statements
             c.argument('geo_redundant_backup', arg_type=get_enum_type(['Enabled', 'Disabled']), options_list=['--geo-redundant-backup'], help='Enable or disable geo-redundant backups. Default value is Disabled. Not supported in Basic pricing tier.')
             c.argument('storage_mb', options_list=['--storage-size'], type=int, help='The storage capacity of the server (unit is megabytes). Minimum 5120 and increases in 1024 increments. Default is 51200.')
             c.argument('auto_grow', arg_type=get_enum_type(['Enabled', 'Disabled']), options_list=['--auto-grow'], help='Enable or disable autogrow of the storage. Default value is Enabled.')
-            c.argument('infrastructure_encryption', arg_type=get_enum_type(['Enabled', 'Disabled']), options_list=['--infrastructure-encryption', 'i'], help='Add an optional second layer of encryption for data using new encryption algorithm. Default value is Disabled.')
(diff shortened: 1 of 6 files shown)


yonzhan:
add to S172

src/azure-cli/azure/cli/command_modules/rdbms/_params.py
```diff
@@ -89,11 +90,13 @@ def _complex_params(command_group):
             c.argument('storage_mb', options_list=['--storage-size'], type=int, help='The storage capacity of the server (unit is megabytes). Minimum 5120 and increases in 1024 increments. Default is 51200.')
             c.argument('backup_retention', options_list=['--backup-retention'], type=int, help='The number of days a backup is retained. Range of 7 to 35 days. Default is 7 days.', validator=retention_validator)
             c.argument('auto_grow', arg_type=get_enum_type(['Enabled', 'Disabled']), options_list=['--auto-grow'], help='Enable or disable autogrow of the storage. Default value is Enabled.')
+            c.argument('infrastructure_encryption', arg_type=get_enum_type(['Enabled', 'Disabled']), options_list=['--infrastructure_encryption'], help='Add an optional second layer of encryption for data using new encryption algorithm. Default value is Disabled.')
```
Juliehzl, line 93:
```suggestion
            c.argument('infrastructure_encryption', arg_type=get_enum_type(['Enabled', 'Disabled']), options_list=['--infrastructure-encryption'], help='Add an optional second layer of encryption for data using new encryption algorithm. Default value is Disabled.')
```

Juliehzl:
Hi @shinilm, is it possible to use similar interface with storage part for infrastructure encryption? e.g.  --require-infrastructure-encryption as a boolean argument,  Allowed values: false, true..

src/azure-cli/azure/cli/command_modules/rdbms/_params.py
shinilm: Resolved

shinilm:
> Hi @shinilm, is it possible to use similar interface with storage part for infrastructure encryption? e.g. --require-infrastructure-encryption as a boolean argument, Allowed values: false, true..

Thanks for the suggestion.  But we have  Arm Templates and Swagger went live with the same property name and we wanted keep it consistent with the same

PR #14124, opened 2020-06-28
[Security] Add commands to manage IoT security solution

ofdadu:
**Description**  
addition of iot security solutionREST to Azure CLI under security module

**Testing Guide**  
as documented 

**History Notes**  

[1] addition of az security iot-solution/ iot-alerts/iot-recommendations/iot-analytics REST to Azure CLI 

---


yonzhan:
add to S172

src/azure-cli/azure/cli/command_modules/security/commands.py
```diff
@@ -261,5 +285,32 @@ def load_command_table(self, _):
         g.custom_command('list', 'list_security_sub_assessments')
         g.custom_command('show', 'get_security_sub_assessment')
 
+    with self.command_group('security iot-solution',
+                            security_iot_solution_sdk,
+                            client_factory=cf_security_iot_solution) as g:
+        g.custom_command('list', 'list_security_iot_solution')
+        g.custom_command('show', 'get_security_iot_solution')
```
Juliehzl, line 292: please use show_command for all show command.

src/azure-cli/azure/cli/command_modules/security/commands.py
```diff
@@ -261,5 +285,32 @@ def load_command_table(self, _):
...
+        g.custom_command('create', 'create_security_iot_solution')
+        g.custom_command('delete', 'delete_security_iot_solution')
```
Juliehzl, line 294: could we also add confirmation=True for all delete commands?

src/azure-cli/azure/cli/command_modules/security/commands.py
```diff
@@ -261,5 +285,32 @@ def load_command_table(self, _):
         g.custom_command('list', 'list_security_sub_assessments')
         g.custom_command('show', 'get_security_sub_assessment')
 
+    with self.command_group('security iot-solution',
```
Juliehzl, line 288: could we add update command not integrate with create command for update operation?

src/azure-cli/azure/cli/command_modules/security/tests/latest/test_iot_scenario.py
```diff
@@ -0,0 +1,27 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+from azure.cli.testsdk import ScenarioTest
```
Juliehzl, line 6: please make sure all new commands included in test.

ofdadu: the other commands are alerts and recommendations
alerts get deleted after 14 days 
recreating alerts on a regular basis on an unknown subscription is problematic  
I consulted with ASC and they did the same on their alerts

src/azure-cli/azure/cli/command_modules/security/commands.py
ofdadu: ASC asked to stay with the convention of create, delete, show, list 
and our REST have a createOrUpdate function meaning putting it on one category is problematic

src/azure-cli/azure/cli/command_modules/security/commands.py
ofdadu: I kept it consistent with all the rest of the show commands in the security module 
I don't think we should change only this commands

src/azure-cli/azure/cli/command_modules/security/commands.py
ofdadu: I can do that 
but it will not stay consistent with the rest of the delete commands
the SDK handles it and will return an error if needed

src/azure-cli/azure/cli/command_modules/security/commands.py
haroldrandom: why not from now on? Keep consistent with other CLI commands style is better.

Juliehzl: show_command will add exception_handler for status code 404, which will be more friendly to customers when error occurs and is consistent with CLI command guideline.

src/azure-cli/azure/cli/command_modules/security/commands.py
Juliehzl: You can consider utilize generic_update_command to add `update` command with `PATCH-like` operation, which is consistent with CLI command guideline https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md#standard-command-types.

src/azure-cli/azure/cli/command_modules/security/tests/latest/test_iot_scenario.py
Juliehzl: actually we have a test subscription. If there is special policy or service to config, we can add it in our test subscription.

ofdadu: in order to test the alerts and recommendation, we need to simulate an attack on a specific device on always live IoT hub with our solution on continuously
it is not an easy task and even if we do that the test can take up to an hour or creating a side service that we need to maintain. (ASC didn’t do that for that reason)

src/azure-cli/azure/cli/command_modules/security/commands.py
ofdadu: changed to show_command

src/azure-cli/azure/cli/command_modules/security/commands.py
ofdadu: separated create and update as requested

src/azure-cli/azure/cli/command_modules/security/tests/latest/test_iot_scenario.py
Juliehzl: I see.

src/azure-cli/azure/cli/command_modules/security/commands.py
```diff
@@ -261,5 +285,33 @@ def load_command_table(self, _):
...
+                            security_iot_analytics_sdk,
+                            client_factory=cf_security_iot_analytics) as g:
+        g.custom_command('list', 'list_security_iot_analytics')
+        g.custom_command('show', 'show_security_iot_analytics')
+
+    with self.command_group('security iot-alerts',
+                            security_iot_alerts_sdk,
+                            client_factory=cf_security_iot_alerts) as g:
+        g.custom_command('list', 'list_security_iot_alerts')
+        g.custom_command('show', 'show_security_iot_alerts')
```
Juliehzl, line 307:
```suggestion
        g.custom_show_command('show', 'show_security_iot_alerts')
```

src/azure-cli/azure/cli/command_modules/security/commands.py
```diff
@@ -261,5 +285,33 @@ def load_command_table(self, _):
...
+        g.custom_command('delete', 'dismiss_security_iot_alerts')
+
+    with self.command_group('security iot-recommendations',
+                            security_iot_recommendations_sdk,
+                            client_factory=cf_security_iot_recommendations) as g:
+        g.custom_command('list', 'list_security_iot_recommendations')
+        g.custom_command('show', 'show_security_iot_recommendations')
```
Juliehzl, line 314:
```suggestion
        g.custom_show_command('show', 'show_security_iot_recommendations')
```

src/azure-cli/azure/cli/command_modules/security/commands.py
Juliehzl: adding confirmation will align with current command guideline. adding `-y` in command will by pass the prompt. so I will recommend to add confirmation in all delete command.

src/azure-cli/azure/cli/command_modules/security/commands.py
```diff
@@ -261,5 +285,33 @@ def load_command_table(self, _):
         g.custom_command('list', 'list_security_sub_assessments')
         g.custom_command('show', 'get_security_sub_assessment')
 
+    with self.command_group('security iot-solution',
+                            security_iot_solution_sdk,
+                            client_factory=cf_security_iot_solution) as g:
+        g.custom_command('list', 'list_security_iot_solution')
+        g.custom_command('show', 'show_security_iot_solution')
```
Juliehzl, line 292:
```suggestion
        g.custom_show_command('show', 'show_security_iot_solution')
```

PR #14134, opened 2020-06-29
[Storage] Remove check in token credential

Juliehzl:
**Description**  

For managed identity, users will be blocked by exception. Because token should be valid and refreshed by server side and we honor it, removing the check should be no effect and add logger to find why users sometimes will get token to be expired in 3mins.

**Testing Guide**  

**History Notes**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/storage/oauth_token_util.py
+++ b/src/azure-cli/azure/cli/command_modules/storage/oauth_token_util.py
@@ -36,7 +36,7 @@ class TokenUpdater(object):
             seconds_left = (datetime.strptime(token['expiresOn'], "%Y-%m-%d %H:%M:%S.%f") - datetime.now()).seconds
         except KeyError:  # needed to deal with differing unserialized MSI token payload
             self.token_credential.token = token['access_token']
-            expire = datetime.datetime.fromtimestamp(int(token['expires_on']))
+            expire = datetime.fromtimestamp(int(token['expires_on']))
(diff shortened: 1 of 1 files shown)


yonzhan:
add to S172

src/azure-cli/azure/cli/command_modules/storage/oauth_token_util.py
```diff
@@ -29,13 +32,15 @@ def timer_callback(self):
...
             seconds_left = (datetime.strptime(token['expiresOn'], "%Y-%m-%d %H:%M:%S.%f") - datetime.now()).seconds
         except KeyError:  # needed to deal with differing unserialized MSI token payload
             self.token_credential.token = token['access_token']
+            expire = datetime.datetime.fromtimestamp(int(token['expires_on']))
             seconds_left = (datetime.fromtimestamp(int(token['expires_on'])) - datetime.now()).seconds
+
         if seconds_left < 180:
-            # acquired token expires in less than 3 mins
-            raise Exception("Acquired a token expiring in less than 3 minutes")
+            logger.warning("Acquired token will expire on %s.", expire)
```
zhoxing-ms, line 43: May I ask why does `seconds_left` compare with 180s, but the error is `will expire on %s.", expire`? Why not `seconds_left < expire`

Juliehzl: expire is actual time and seconds left is time range. The compare is used to make sure we there is enough buffer time for the acquired token to be used in the request.

PR #14140, opened 2020-06-30
{ARM} Fix the bugs for failed cases in live tests

zhoxing-ms:
**Description**  

***Fix the bugs for failed cases in live tests***
1. `test_delete_dependent_resources/test_invoke_action`
    Error: block by the policy `deny-nsg-internet-inbound`  
    Solution: The parameters `--nsg-rule None` need to be added to bypass policy restrictions
2. `test_tag_scenario`  
    Error: Predefined tag name 'test' could not be found.
    Solution: In the test, reverse the parameter positions of tagName and tagvalue, so correct the corresponding positions
3. `test_rest`
    Error: Unsupported Media Type({"error":{"code":"UnsupportedMediaType","message":"The content media type \'<null>\' is not supported. Only \'application/json\' is supported."}}) 
    Solution: The file content cannot be loaded into the context due to the direct input of the file name, the value `@rest_storage_account_create_body.json` is passed to the Service as the body content. 

**Testing Guide**  

**History Notes**  

---


yonzhan:
add to S172

src/azure-cli/azure/cli/command_modules/util/tests/latest/test_rest.py
```diff
@@ -13,14 +14,19 @@ class ResourceGroupScenarioTest(ScenarioTest):
 
     @ResourceGroupPreparer(name_prefix='cli_test_rest')
     def test_rest(self, resource_group):
+        curr_dir = os.path.dirname(os.path.realpath(__file__))
         self.kwargs.update({
-            'sa': self.create_random_name("tmpst", length=10)
+            'sa': self.create_random_name("tmpst", length=10),
+            'storage_account_create_body': os.path.join(curr_dir,
+                                                        'rest_storage_account_create_body.json').replace('\\', '\\\\'),
```
jiasli, line 21: Let's add a comment saying `\` will be removed by `shlex.split`.

jiasli, line 21: Maybe we can use the same mechanism of `TEST_DIR` from keyvault module?

https://github.com/Azure/azure-cli/blob/a8c1a56758dfa9c4d862e6ed46dd50080eff8391/src/azure-cli/azure/cli/command_modules/keyvault/tests/latest/test_keyvault_commands.py#L31

(Though let's use `os.path.dirname` instead of `..`)

zhoxing-ms: OK, this way is more readable. I've modified it

src/azure-cli/azure/cli/command_modules/util/tests/latest/test_rest.py
```diff
@@ -5,22 +5,28 @@
...
+            'sa': self.create_random_name("tmpst", length=10),
+            'storage_account_create_body': os.path.join(TEST_DIR, 'rest_storage_account_create_body.json'),
+            'storage_account_sas_body': os.path.join(TEST_DIR, 'rest_storage_account_sas_body.json')
         })
 
         # Create a storage account
         # https://docs.microsoft.com/en-us/rest/api/storagerp/storageaccounts/create
         self.cmd('az rest -m PUT -u /subscriptions/{{subscriptionId}}/resourceGroups/{rg}/providers/Microsoft.Storage/storageAccounts/{sa}?api-version=2019-06-01 '
-                 '-b @rest_storage_account_create_body.json')
+                 '-b @"{storage_account_create_body}"')
```
jiasli, line 29: Let's make a comment about why `\` -> `\\` escaping is not necessary.

zhoxing-ms: Done
```py
self.kwargs.update({
            'sa': self.create_random_name("tmpst", length=10)
            'sa': self.create_random_name("tmpst", length=10),
            # Please remember to add double quote to the variable reference of the file path in tests,
            # otherwise \\ will be parsed to \, which will lead to the wrong path
            'storage_account_create_body': os.path.join(TEST_DIR, 'rest_storage_account_create_body.json'),
            'storage_account_sas_body': os.path.join(TEST_DIR, 'rest_storage_account_sas_body.json')
        })
```

src/azure-cli/azure/cli/command_modules/util/tests/latest/test_rest.py
```diff
@@ -5,22 +5,30 @@
...
 class ResourceGroupScenarioTest(ScenarioTest):
 
     @ResourceGroupPreparer(name_prefix='cli_test_rest')
     def test_rest(self, resource_group):
+
         self.kwargs.update({
-            'sa': self.create_random_name("tmpst", length=10)
+            'sa': self.create_random_name("tmpst", length=10),
+            # Please remember to add double quote to the variable reference of the file path in tests,
+            # otherwise \\ will be parsed to \, which will lead to the wrong path
```
jiasli, line 23:
```suggestion
            # otherwise \ will be removed by shlex.split, which will lead to a wrong path.
```

PR #14153, opened 2020-06-30
Azure sphere authentication

tibortakacs:
This PR exposes scopes and app registration client id from the `Profile` class. This change would be essential for third-party applications which are built on top of the `azure-cli-core`. One example is Azure Sphere CLI.

The main concept of PR is the following:
* Authority is in the, i.e. it is constant in an object. This is just an extension of the original implementation which uses the `cli_ctx.cloud.endpoints.active_directory` everywhere. This change just make it configurable.
* The scopes and the app registration client id are in the constructor of the `Profile` class.
* The "resource" and "scopes" are similar terms: the former is for the ADAL world, the latter one is the new MSAL universe. It is not accepted to set up both, so where it is possible, there is validation step. However, this PR does not aim to resolve all possible collisions, so for example in the `AuthenticationWrapper` the "scopes" has just higher priority than the "resouce". I believe there is a tendency to move completely to MSAL, so all "resource" references will be removed at some point, and/or coherent logic is needed how to handle them together.
* The introduced "scopes" and "client id" are not published in the `login` command layer yet but of course nothing blocks to make them available in the CLI.

Note that all change are intented to be backward compatible, i.e. it should be totally transpartent in part of the Azure CLI stack.


yonzhan:
sphere

src/azure-cli-core/azure/cli/core/commands/client_factory.py
```diff
@@ -137,15 +137,24 @@ def _get_mgmt_service_client(cli_ctx,
                              api_version=None,
                              base_url_bound=True,
                              resource=None,
+                             scopes=None,
```
qianwens, line 140: >scopes=None, [](start = 29, length = 12)

please move `scopes` to the end of the parameter list for backward compatibility.

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -97,17 +97,15 @@ def _get_cloud_console_token_endpoint():
 # pylint: disable=too-many-lines,too-many-instance-attributes,unused-argument
 class Profile(object):
 
-    def __init__(self, storage=None, auth_ctx_factory=None, use_global_creds_cache=True,
+    def __init__(self, authority=None, storage=None, auth_ctx_factory=None, use_global_creds_cache=True,
```
qianwens, line 100: >authority=None, [](start = 23, length = 15)

it seems tricky here to pass your authority instead of use cli_ctx, could you please configure self.cli_ctx.cloud.endpoints.active_directory in your CLI code?

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -117,6 +115,8 @@ def login(self,
               password,
               is_service_principal,
               tenant,
+              scopes=None,
+              client_id=None,
```
qianwens, line 119: why need specify scopes when login? I assume you only need to use it when get_token for the service API call

src/azure-cli-core/azure/cli/core/_identity.py
```diff
@@ -413,19 +413,25 @@ def save_service_principal_cred(self, sp_entry):
 
     # noinspection PyBroadException
     # pylint: disable=protected-access
-    def add_credential(self, credential):
+    def add_credential(self, credential, **kwargs):
```
qianwens, line 416: >add_credential [](start = 8, length = 14)

ADALCredentialCache is used by Azure CLI to store ADAL format token file for backward compatibility. Azure Sphere CLI may not need this. You can add a parameter like `store_ADAL_cache` in Profile and only call add_credential when it is true.

src/azure-cli-core/azure/cli/core/authentication.py
```diff
@@ -25,7 +25,7 @@ def _create_scopes(resource):
         scope = resource + '/.default'
     else:
         scope = resource.rstrip('/') + '/.default'
-    return scope
+    return scope,
```
qianwens, line 28: >scope, [](start = 11, length = 6)

typo

src/azure-cli-core/azure/cli/core/authentication.py
```diff
@@ -35,16 +35,21 @@ def __init__(self, credential, **kwargs):
         # _external_credentials and _resource are only needed in Track1 SDK
         self._external_credentials = kwargs.pop("external_credentials", None)
         self._resource = kwargs.pop("resource", None)
+        self._scopes = kwargs.pop("scopes", None)
```
qianwens, line 38: >self._scopes = kwargs.pop("scopes", None) [](start = 8, length = 41)

seems not needed since `scopes` should be set in `get_token`

src/azure-cli-core/azure/cli/core/commands/client_factory.py
tibortakacs: Of course, I am happy to do that.

Let me ask a clarification question here. I put it next to `resource` because they have the same purpose. I thought that `_get_mgmt_service_client` is an internal private function for only internal purposes, and I checked all occurrence in the `azure-cli` project. Please let me know if considering this you still think it is good to move the parameter to the end of the function. Thank you.

src/azure-cli-core/azure/cli/core/authentication.py
tibortakacs: This is a tricky one.

The `scopes` input parameter of `_get_token(...)` function is a tuple:
https://github.com/Azure/azure-cli/blob/b6c8e436888f93bee8a1e4b45700671e754b8500/src/azure-cli-core/azure/cli/core/authentication.py#L40

If the `scopes` argument has multiple items, the original implementation had a bug here:
`token = self._credential.get_token(scopes)`
Because the tuple is not unpacked, the `_credential.get_token(...)` function received a tuple of tuples, for example, here:
https://github.com/Azure/azure-sdk-for-python/blob/aff33bd40c12b16ec3eeaf9ccb7925f884a8a913/sdk/identity/azure-identity/azure/identity/_internal/msal_credentials.py#L161
Therefore, during the token acquiring process, the "scope" was the appended string of the elements in the tuple which is of course wrong, the authentication failed.

So, I changed the line to:
`token = self._credential.get_token(*scopes)`

This solved the original problem but led to a different one. The `_create_scopes` function returned a simple string in which the unpack operation resulted a tuple of characters. This, of course, does not work either.

So, I changed the `_create_scopes` function to return a tuple with one element, the `,` after `scope` does this.

I believe this change is necessary to resolve the original bug (independently from the original purpose of this PR), however, I am happy to either add a short comment why it is done so, or moving the tuple creation out of the `_create_scopes` function to the `_get_token` function.

Do you have any preference here?

src/azure-cli-core/azure/cli/core/_profile.py
tibortakacs: This is due to the fact that a 3rd party application does not always have access to the common management scope/resource (e.g. to the https://management.core.windows.net/). 

For example, this is the error message I receive if I do not specify the proper scope in the login call:

Authentication failed: ["AADSTS650057: Invalid resource. The client has requested access to a resource which is not listed in the requested permissions in the client's application registration. Client app ID: 0b1c8f7e-28d2-4378-97e2-7d7d63f7c87f(Azure Sphere Utility). Resource value from request: https://management.core.windows.net/. Resource app ID: 797f4846-ba00-4fd7-ba43-dac1f8f63013. List of valid resources from app registration: f0277259-56be-4121-a579-5a57fcd1dbac, 00000002-0000-0000-c000-000000000000.\r\nTrace ID: 820a9c22-4c5b-44c5-82ca-1e5675584800\r\nCorrelation ID: fbfb8bfc-1421-4241-820a-e05940c18b01\r\nTimestamp: 2020-07-02 12:12:54Z"]

By adding the `scopes` the parameter to the `login` function, the "log in" call means, that the user logs in into an AAD `tenant` from an app specified by the `client_id` to access a resource specified by the `scopes`.

What do you think?

src/azure-cli-core/azure/cli/core/_identity.py
tibortakacs: Great point, going to do the modification.

src/azure-cli-core/azure/cli/core/authentication.py
tibortakacs: See my detailed response at the related discussion.

src/azure-cli-core/azure/cli/core/_profile.py
tibortakacs: Great idea, I have changed the code, thank you.

I might keep the `self._authority` though because there is the `.replace('https://', ''` part of the authority creation:
https://github.com/Azure/azure-cli/blob/b6c8e436888f93bee8a1e4b45700671e754b8500/src/azure-cli-core/azure/cli/core/_profile.py#L107

I think it is safer if it is done only once.

What do you think?

src/azure-cli-core/azure/cli/core/_profile.py
tibortakacs: Reading your other comment about `cli_ctx`, just had a thought about this. Would it be a feasible option to overwrite the `cli_ctx.cloud.endpoints.active_directory_resource_id` by the `login` command? If yet, would it be an option to introduce a `cli_ctx.cloud.endpoints.active_directory_scopes` member to be compatible with MSAL?

src/azure-cli-core/azure/cli/core/commands/client_factory.py
```diff
@@ -137,15 +137,23 @@ def _get_mgmt_service_client(cli_ctx,
...
                              sdk_profile=None,
                              aux_subscriptions=None,
                              aux_tenants=None,
                              **kwargs):
```
tibortakacs, line 144: @qianwens @jiasli I have an interesting finding here. It turned out that if I use the `--credential-scopes` parameter in the autorest generation, it automatically adds my scope into the generated client library:

```
class PublicApiClientConfiguration(Configuration):

    def __init__(
        self,
        credential,  # type: "TokenCredential"
        **kwargs  # type: Any
    ):
        # type: (...) -> None
        if credential is None:
            raise ValueError("Parameter 'credential' must not be None.")
        super(PublicApiClientConfiguration, self).__init__(**kwargs)

        self.credential = credential
        self.credential_scopes = ['https://ostechnology.onmicrosoft.com/azsphereapi/user_impersonation']
        self.credential_scopes.extend(kwargs.pop('credential_scopes', []))
        kwargs.setdefault('sdk_moniker', 'sphere-publicapi/{}'.format(VERSION))
        self._configure(**kwargs)
```

This means that extending this function is not really necessary if the generated SDK already has the scope internally as in my example above.

I still believe that it does make sense to extend this API with the `scopes` since the `resource` is already there. I just wanted to draw your attention on this. What is your view about this?

src/azure-cli-core/azure/cli/core/_profile.py
jiasli: @qianwens, according to [authorization code flow](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow#request-an-authorization-code), the initial request also requires a `scope` to which the `client_id` may not have access:

```
https://login.microsoftonline.com/{tenant}/oauth2/v2.0/authorize?
client_id=6731de76-14a6-49ae-97bc-6eba6914391e
&response_type=code
&redirect_uri=http%3A%2F%2Flocalhost%2Fmyapp%2F
&response_mode=query
&scope=openid%20offline_access%20https%3A%2F%2Fgraph.microsoft.com%2Fmail.read
&state=12345
```

That's why we need to make it customizable.

src/azure-cli-core/azure/cli/core/_identity.py
```diff
@@ -96,19 +96,20 @@ def _msal_app(self):
         msal_authority = "https://{}/{}".format(self.authority, self.tenant_id)
         return PublicClientApplication(authority=msal_authority, client_id=self.client_id, token_cache=cache)
 
-    def login_with_interactive_browser(self):
+    def login_with_interactive_browser(self, scopes=None):
```
jiasli, line 99: If multiple methods requires `scopes`, how about put it in `Identity.__init__`?

src/azure-cli-core/azure/cli/core/_profile.py
jiasli: Azure Identity has some built-in default scopes:

https://github.com/Azure/azure-sdk-for-python/blob/fdf11b302b20821e94eae61a2eabd8eb1d776162/sdk/identity/azure-identity/azure/identity/_internal/msal_credentials.py#L41-L46

```py
_DEFAULT_AUTHENTICATE_SCOPES = {
    "https://" + KnownAuthorities.AZURE_CHINA: ("https://management.core.chinacloudapi.cn//.default",),
    "https://" + KnownAuthorities.AZURE_GERMANY: ("https://management.core.cloudapi.de//.default",),
    "https://" + KnownAuthorities.AZURE_GOVERNMENT: ("https://management.core.usgovcloudapi.net//.default",),
    "https://" + KnownAuthorities.AZURE_PUBLIC_CLOUD: ("https://management.core.windows.net//.default",),
}
```

src/azure-cli-core/azure/cli/core/authentication.py
```diff
@@ -25,7 +25,7 @@ def _create_scopes(resource):
         scope = resource + '/.default'
     else:
         scope = resource.rstrip('/') + '/.default'
```
jiasli, line 27: @qianwens, let's not hard-code `datalake`, `batch` and `database`, but `rstrip` anyway.

src/azure-cli-core/azure/cli/core/commands/client_factory.py
jiasli: I think we should totally remove `resource` as it is an ADAL concept and has been replaced by `scopes`.

src/azure-cli-core/azure/cli/core/commands/client_factory.py
jiasli: This approach makes the SDK lose the capability to customize the scope from outside the SDK. For example, ARM in `AzureCloud` has https://github.com/Azure/azure-cli/blob/81ff7bfad67383a9d8aba2e2e0bd5a7457b0f8b9/src/azure-cli-core/azure/cli/core/cloud.py#L250 

while `AzureChinaCloud` has https://github.com/Azure/azure-cli/blob/81ff7bfad67383a9d8aba2e2e0bd5a7457b0f8b9/src/azure-cli-core/azure/cli/core/cloud.py#L281

src/azure-cli-core/azure/cli/core/_profile.py
jiasli: Azure Identity has some built-in default scopes:

https://github.com/Azure/azure-sdk-for-python/blob/fdf11b302b20821e94eae61a2eabd8eb1d776162/sdk/identity/azure-identity/azure/identity/_internal/msal_credentials.py#L41-L46

```py
_DEFAULT_AUTHENTICATE_SCOPES = {
    "https://" + KnownAuthorities.AZURE_CHINA: ("https://management.core.chinacloudapi.cn//.default",),
    "https://" + KnownAuthorities.AZURE_GERMANY: ("https://management.core.cloudapi.de//.default",),
    "https://" + KnownAuthorities.AZURE_GOVERNMENT: ("https://management.core.usgovcloudapi.net//.default",),
    "https://" + KnownAuthorities.AZURE_PUBLIC_CLOUD: ("https://management.core.windows.net//.default",),
}
```

src/azure-cli-core/azure/cli/core/authentication.py
jiasli: Yes. I prefer `token = self._credential.get_token(*scopes)`.

src/azure-cli-core/azure/cli/core/commands/client_factory.py
tibortakacs: This is correct. However, if the SDK is generate with `azure-arm` flag (which is should be), it adds `https://management.core.windows.net/` as the default `credential_scopes` item. For example, in storage, as a random example: https://github.com/Azure/azure-sdk-for-python/blob/ba66869562f94559b3f167bc8d3a733157b67573/sdk/storage/azure-mgmt-storage/azure/mgmt/storage/_configuration.py#L46

src/azure-cli-core/azure/cli/core/_identity.py
tibortakacs: I was thinking about the same. I did not want to change the "state" of the `Identity` object with the `scopes` member variable because I did not know whether it is expected or not. As you mentioned this here, I am super happy to go into this direction, going to do that.

src/azure-cli-core/azure/cli/core/commands/client_factory.py
tibortakacs: I am happy to do this in this PR, however, it might be that removing all resource reference in one go would cleaner. Up to you. What do you think?

src/azure-cli-core/azure/cli/core/_profile.py
tibortakacs: This is correct. It picks up the one which belongs to the passed authority.

src/azure-cli-core/azure/cli/core/authentication.py
tibortakacs: @jiasli Just to be sure: so, are you happy then to return tuple with line `return scope,`, or would you recommend a modification?

src/azure-cli-core/azure/cli/core/commands/client_factory.py
tibortakacs: @jiasli I was thinking a little bit about this, and my concern is that this function's parameter list is practically part of the public API since the `get_mgmt_service_client(...)` function just forwards the `kwargs` arguments. So, if anybody uses the `resource` named parameter by calling the `get_mgmt_service_client` method, it would break. And there are many occasions for this. Therefore, we might consider to deprecated this public function and create a new one which has only the `scopes` as accepted parameter. (This is of course just an idea, you are expert in this field.)

However, the `_get_mgmt_service_client` function internally forwards the `resource` to the `Profile::get_login_credentials` function. I think this is only for backward compatibility with the ADAL, so we can get easier rid of it. This would require, however, slightly bigger change because there are a few places in the `azure-cli` project which calls `get_login_credentials` with resource parameter. I might do this is in a separated PR.

src/azure-cli-core/azure/cli/core/_identity.py
tibortakacs: I have push a new commit with this change. Looking forward to your opinion.

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -204,9 +208,8 @@ def login_with_managed_identity(self, identity_id=None, allow_no_subscriptions=N
         # Managed identities for Azure resources is the new name for the service formerly known as
         # Managed Service Identity (MSI).
 
-        resource = self.cli_ctx.cloud.endpoints.active_directory_resource_id
```
tibortakacs, line 207: Put back resource -> scope here.

src/azure-cli-core/azure/cli/core/_profile.py
jiasli: Possibly we can move the `client_id` to `__init__`?

src/azure-cli-core/azure/cli/core/authentication.py
jiasli: `return scope,` is good. Let's just add more comments in case of confusion.

src/azure-cli-core/azure/cli/core/commands/client_factory.py
jiasli: Let's put this on hold. We will see how we can deprecate `resource` in the future.

src/azure-cli-core/azure/cli/core/commands/client_factory.py
jiasli: @jsntcy, let's make sure we pass the correct `credential_scopes` to Track 2 SDKs.

src/azure-cli-core/azure/cli/core/extension/__init__.py
```diff
@@ -251,20 +251,21 @@ def get_all():
...
                 ext_name = os.path.split(ext_path)[-1]
                 exts.append(DevExtension(ext_name, ext_path))
             else:
                 for item in os.listdir(path):
-                    _collect(os.path.join(path, item), depth + 1, max_depth)
+                    _collect(os.path.join(path, item), packaging_pattern, depth + 1, max_depth)
         for source in DEV_EXTENSION_SOURCES:
-            _collect(source)
+            _collect(source, '*.dist-info')
+            _collect(source, '*.egg-info')
```
jiasli, line 268: Let's create a separate PR for this change and add some detailed descriptions. @fengzhou-msft, @haroldrandom who are our packaging experts will help review it.

tibortakacs: Good point, I have reverted the commit.

src/azure-cli-core/azure/cli/core/authentication.py
tibortakacs: I have added a short comment to avoid confusion.

src/azure-cli-core/azure/cli/core/_profile.py
tibortakacs: I have moved both `client_id` and `scopes` into the constructor.

src/azure-cli-core/azure/cli/core/_identity.py
```diff
@@ -64,10 +65,13 @@ class Identity:
...
     CLOUD_SHELL_IDENTITY_UNIQUE_NAME = "unique_name"
 
-    def __init__(self, authority=None, tenant_id=None, client_id=None, **kwargs):
+    def __init__(self, authority=None, tenant_id=None, client_id=None, scopes=None, **kwargs):
         self.authority = authority
         self.tenant_id = tenant_id or "organizations"
         self.client_id = client_id or _CLIENT_ID
+        self.scopes = scopes or _DEFAULT_SCOPES
+        if self.scopes and not isinstance(self.scopes, (list, tuple)):
+            self.scopes = (self.scopes,)
```
jiasli, line 74: I will unify `scopes` to a `list` as MSAL's `scope` use `list[str]` as the type:

https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/ece3918c62a403b95a078dbccbaf093bc5c6415e/msal/application.py#L523

```py
        :param list[str] scopes: (Required)
            Scopes requested to access a protected API (a resource).
```

src/azure-cli-core/azure/cli/core/_identity.py
```diff
@@ -23,6 +23,7 @@
 )
 
 _CLIENT_ID = '04b07795-8ddb-461a-bbee-02f9e1bf7b46'
+_DEFAULT_SCOPES = ('https://management.core.windows.net/.default',)
```
jiasli, line 26: I will remove this as the default `scopes` is related to the current cloud.

jiasli:
As this PR has diverged so much from the current `identity-feature` branch, I will manually merge it line-by-line.

Issue #13752, opened 2020-05-29
az keyvault create not idempotent with access policies


mcalnd70:
> ### `az feedback` auto-generates most of the information requested below, as of CLI version 2.0.62

**Describe the bug**

The "**az keyvault create**" command is not idempotent with **Access Policies**. On re-running the command the Access Policies are deleted/removed.

**To Reproduce**

Create a KeyVault with **az keyvault create**  e.g. "**_az keyvault create --location westus2 --name MyKeyVault --resource-group MyResourceGroup_**"

Assign an access policy to the KeyVault, either manually in the portal, or via another Azure CLI or PowershellAz command.

Run the "**az keyvault create --location westus2 --name MyKeyVault --resource-group MyResourceGroup**" command again

The access policy has been removed/deleted

**Expected behavior**
Leave the Access Policy alone

**Environment summary**

All environments

**Additional context**

yonzhan:
@bim-msft please take a look

bim-msft:
Duplicate with #13212 
Hi @mcalnd70 , thanks for the feedback. Unfortunately, this is the expected behavior on service side. You can refer to the above issue to see more detail.
If you have any question, feel free to ping me again.

mcalnd70:
@bim-msft The other one has been closed? Is this being worked on?

bim-msft:
@mcalnd70 Hi, sorry for the late response. We are working on it.

yungezz:
This one is a feature request because previously az keyvault don't support idempotent at client side, but depends on service to provide idempotency. To support at client side, there's little debate at existing user experience impact. PR is ready, we're working on reviewing and fine design on it.

pascalnaber:
@yungezz Maybe I don't understand what you mean with client side idempotency. Because I would not know why somebody needs that. Idempotency is about server side behavior if you ask me.

If I read about the az cli Standard Command Types, The CREATE command MUST be idempotent according to the documentation.

https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md#standard-command-types

Hopefully, the behavior of the Keyvault can be changed soon according to the cli guidelines.

puagarwa:
@bim-msft Is there any ETA for this fix ? I am also facing this issue in my CMK code where i create key vault and then add ACL for CMK to work but once we deploy template again, things break because it altogether remove access policy. I have to everything via ARM here.

Issue #7613, opened 2018-10-18
Feature/Discussion: provide safety net for `create` commands and make them consistent


ppanyukov:
**Is your feature request related to a problem? Please describe.**

This is prompted by **"https://github.com/Azure/azure-cli/issues/7607: `az keyvault create` silently zaps existing access policy when run on existing vault"**

The crux of the problem is:

- Some/most `create` commands are destructive and will silently do their destructive deed.

- There is no safety net.

- This is not the default behaviour I want in production -- it's too scary.

- Relying on humans not forget to check if resource already exists before 
  running `create` is not the answer in the absence of safety net.

- With some `create` commands like `az storage container create` being
  safe to run at any time, it makes it easier to forget that other commands
  are not safe.

- Destructive commands are inconsistent and undocumented in what they destroy:
  e.g. `az keyvault create` zaps access policies but leaves the secrets intact.
  Practical test seems to be the only way to be 100% sure and it's a poor option.

- It is well within Azure CLI's capabilities to do the "remembering" and
  safety net for us and thus provice a better experience here.

-----
**Describe the solution you'd like**

The goal of the solution is:

- Make Azure CLI much safer to use;
- Bring consistency to the all commands;
- Do the "remembering" work for us humans;
- No nasty surprises.

The proposal is:

**For all safe non-destructive `create` commands:**

- A **safe** `create` command is a command which:

    - Safe to run at any time
    - Never destroys or changes anything if target resource exists
    - Actually creates target resource only if it does not exist
    - Note that if we run `create` with parameters such that
      it would lead to any chages in target resource it is not 
      a safe command.
    - Example: `az group create` (it's safe, right?)

- We can run it at any time just like now.

- Consider for all such commands to have `--fail-on-exist` flag for
  consistency (see `az storage container create`).

**For all non-safe destructive `create` commands:**

- A **non-safe** `create` command is any command which:

    - Destroys all or part of the target resource (see `az keyvault create`)
    - Makes any changes to the target resource if it exists
    - Some normally safe commands may become unsafe depending on paramteres.

- Default behaviour for these should be:

    - Command always fails if the target resource already exists.
    - This reminds people to use `if resource exists; then update; else create` pattern.
    - This provides much needed safety net.

- For legacy and "I know what I'm doing" scenarios:

    - Supply `--force` or similar flag.
    - This will enable to current destructive behaviour for
      compatility/whatever other reasons. 

-----

**Describe alternatives you've considered**

[...]

PR #14209, opened 2020-07-03
{KeyVault} Make `az keyvault create` idempotent

bim-msft:
Fix: #13752 

As the above issue described, currently `az keyvault create` is not idempotent: re-create will lose previous access policies.

**[NOTICE]  We have added a `get` operation before `create`, this breaks the existing playback. So all tests related to `az keyvault create` have to run live again.**

---


yonzhan:
KeyVault

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
```diff
@@ -334,6 +334,11 @@ def create_keyvault(cmd, client,  # pylint: disable=too-many-locals
     from azure.graphrbac.models import GraphErrorException
     from azure.graphrbac import GraphRbacManagementClient
 
+    try:
+        return client.get(resource_group_name=resource_group_name, vault_name=vault_name)
```
yungezz, line 338: if get, will directly return existing keyvault without updating? This might be problem when user creating params is different with existing properties. I think 2 options here
- a nice message. "The resource exists already"
- get firstly, if exists, do put to update

fengzhou-msft: Portal behavior is to disallow the creation if the vault name already exists. Shall we follow it?

bim-msft: Hi @yungezz , @fengzhou-msft Thanks for the good suggestion.
It makes more sense to just disallow the creation, and prompt some messages as well.
Update the resource may cause breaking changes when handling default values. It's better to use `az keyvault update` command to explicitly do the update.

bim-msft:
**[UPDATE]**
No longer raise exceptions if the vault already exists. Instead, updating it using the user-specified parameters (not default values).

```python
        vault = client.get(resource_group_name=resource_group_name, vault_name=vault_name)
        if vault:
            properties = vault.properties
            if tags:
                vault.tags = tags

            if enabled_for_deployment is not None:
                properties.enabled_for_deployment = enabled_for_deployment
            if enabled_for_disk_encryption is not None:
                properties.enabled_for_disk_encryption = enabled_for_disk_encryption
            if enabled_for_template_deployment is not None:
                properties.enabled_for_template_deployment = enabled_for_template_deployment
            if enable_rbac_authorization is not None:
                properties.enable_rbac_authorization = enable_rbac_authorization
            if enable_soft_delete is not None:
                properties.enable_soft_delete = enable_soft_delete
            if enable_purge_protection is not None:
                properties.enable_purge_protection = enable_purge_protection
            if retention_days is not None:
                properties.retention_days = retention_days

            if network_acls or network_acls_ips or network_acls_vnets:
                if cmd.supported_api_version(resource_type=ResourceType.MGMT_KEYVAULT, min_api='2018-02-14'):
                    network_acls = _create_network_rule_set(cmd, bypass, default_action) \
                        if bypass or default_action else \
                        _parse_network_acls(cmd, resource_group_name, network_acls,
                                            network_acls_ips, network_acls_vnets)
                    properties.network_acls = network_acls

            parameters = VaultCreateOrUpdateParameters(location=vault.location,
                                                       tags=vault.tags,
                                                       properties=properties)

            return client.create_or_update(resource_group_name=resource_group_name,
                                           vault_name=vault_name,
                                           parameters=parameters)
```

src/azure-cli/azure/cli/command_modules/batchai/tests/latest/test_batchai_custom.py
```diff
@@ -945,7 +945,7 @@ def test_list_setup_task_files_for_cluster_when_output_on_bfs(self, get_files_fr
                 relative_mount_path='bfs',
                 container_name='container',
                 credentials=None),
-            os.path.join('path/segment/', '.'), 60)
+            os.path.join('path', 'segment', '.'), 60)
```
qwordy, line 948: What's the difference by adding `/.`

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
```diff
@@ -345,6 +345,46 @@ def create_keyvault(cmd, client,  # pylint: disable=too-many-locals
...
+
+            parameters = VaultCreateOrUpdateParameters(location=vault.location,
+                                                       tags=vault.tags,
+                                                       properties=properties)
+
+            return client.create_or_update(resource_group_name=resource_group_name,
+                                           vault_name=vault_name,
+                                           parameters=parameters)
+    except:  # pylint: disable=bare-except
+        pass
```
qwordy, line 386: Should you inform user that something is wrong?

qwordy:
I don't think it's a reasonable request of this issue. He should assign policy again. I suggest abandon this PR.
```
Create a KeyVault with az keyvault create e.g. "az keyvault create --location westus2 --name MyKeyVault --resource-group MyResourceGroup"

Assign an access policy to the KeyVault, either manually in the portal, or via another Azure CLI or PowershellAz command.

Run the "az keyvault create --location westus2 --name MyKeyVault --resource-group MyResourceGroup" command again

The access policy has been removed/deleted
```

src/azure-cli/azure/cli/command_modules/batchai/tests/latest/test_batchai_custom.py
bim-msft: @qwordy The old one only works on Linux-like systems. Any hardcoded slashes or anti-slashes are not recommended.

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
bim-msft: @qwordy Not necessary. An exception here means we didn't find the existing vault, so just create a new one.

bim-msft:
> I don't think it's a reasonable request of this issue. He should assign policy again. I suggest abandon this PR.
> 
> ```
> Create a KeyVault with az keyvault create e.g. "az keyvault create --location westus2 --name MyKeyVault --resource-group MyResourceGroup"
> 
> Assign an access policy to the KeyVault, either manually in the portal, or via another Azure CLI or PowershellAz command.
> 
> Run the "az keyvault create --location westus2 --name MyKeyVault --resource-group MyResourceGroup" command again
> 
> The access policy has been removed/deleted
> ```

@qwordy I think it's the bad user experience (at least 2 customers complained about this):
1. The user created a vault: `az keyvault create --location westus2 --name MyKeyVault --resource-group MyResourceGroup`
2. We assigned some initial access policies to the vault.
3. The user modified the initial access policies.
4. The user ran `az keyvault create --location westus2 --name MyKeyVault --resource-group MyResourceGroup` again.
5. The access policies would be changed back to the original one. (doesn't make sense, the user didn't mean to delete/modify the policies)

src/azure-cli/azure/cli/command_modules/batchai/tests/latest/test_batchai_custom.py
bim-msft: @qwordy I think there's no difference between `{path}` and `{path}/.`, but in order to keep the original code I decided to not change it.

fengzhou-msft: already fixed in this PR: https://github.com/Azure/azure-cli/pull/14239/files#diff-a7bbbe4eef8bb85685e4c7965541d92d

wangzelin007 (approved):

wangzelin007 (commented):
I was wondering do we still need merge this PR since bim-msft has left MS.

jiasli:
By the way, the discussion about `az keyvault create`'s idempotence is at #7613.

PR #14248, opened 2020-07-07
[App Service] fix#-13035- Added validation for az webapp config access-restriction …

Kotasudhakarreddy:
…to avoid adding duplicates ipaddress

**Description**  
#13035 

**Testing Guide**  

try to add the multiple rules with same Ip. Then it should not allow.

sample command: 

**az webapp config access-restriction add -g 13035RG -n 13035-testing --ip-address 208.131.0.0/16 -p 100** 

**History Notes**  

[AppService] - Added validation for az webapp config access-restriction to avoid adding duplicates IP rules 
---


yonzhan:
AppService

Kotasudhakarreddy:
@panchagnula - FYA.

Kotasudhakarreddy:
@panchagnula - please review the changes.

src/azure-cli/azure/cli/command_modules/appservice/access_restrictions.py
```diff
@@ -57,10 +57,18 @@ def add_webapp_access_restriction(
...
+            if rule.ip_address == ip_address:
+                is_exists = True
+                break
+        if not is_exists:
+            rule_instance = IpSecurityRestriction(
+                name=rule_name, ip_address=ip_address,
+                priority=priority, action=action, tag='Default', description=description)
+            access_rules.append(rule_instance)
+        else:
+            raise CLIError('Rule is already exists with the specified criteria')
```
panchagnula, line 71: A few comments here -> I don't think we need the tmp is_exists variable, in Python we can do a lambda filter to check if value exists in object, if we find a value, then we should just raise CLI error saying "Duplicate Rule <RuleName>. already found".

Infact, is there a reason we are not adding this to the validations.py so for add we do the isDuplicateValidation check first before we even enter the execution method here?

Kotasudhakarreddy: Made the code changes using Lambda. Regarding adding validation in validations.py file. Validator is already added to the ip_address parameter to check the format (`c.argument('ip_address', help="IP address or CIDR range", validator=validate_ip_address)`). And it is not allowing allowing me to add second validator to pass. So I have added it in function.

src/azure-cli/azure/cli/command_modules/appservice/access_restrictions.py
```diff
@@ -57,10 +57,14 @@ def add_webapp_access_restriction(
         access_rules.append(rule_instance)
 
     elif ip_address:
-        rule_instance = IpSecurityRestriction(
-            name=rule_name, ip_address=ip_address,
-            priority=priority, action=action, tag='Default', description=description)
-        access_rules.append(rule_instance)
+        is_exists = [(lambda x: x.ip_address == ip_address)(x) for x in access_rules]
```
qwordy, line 60:
```
is_exists = [x.ip_address == ip_address for x in access_rules]
```
It seems lambda is not necessary.

src/azure-cli/azure/cli/command_modules/appservice/_validators.py
```diff
@@ -164,21 +164,38 @@ def validate_asp_sku(cmd, namespace):
...
+def validate_ip_address_existance(cmd, namespace):
+    resource_group_name = namespace.resource_group_name
+    name = namespace.name
+    slot = namespace.slot
+    from ._appservice_utils import _generic_site_operation
+    configs = _generic_site_operation(cmd.cli_ctx, resource_group_name, name, 'get_configuration', slot)
+    access_rules = configs.ip_security_restrictions
+    is_exists = [(lambda x: x.ip_address == namespace.ip_address)(x) for x in access_rules]
+    if True in is_exists:
+        raise CLIError('Rule is already exists with the specified criteria')
```
panchagnula, line 201: IP address '<value entered>' already exists. Cannot add duplicate IP address values.

Kotasudhakarreddy:
closing this as a duplicate of #14486

PR #14269, opened 2020-07-08
{CommandIndex} Support commands with positional arguments

jiasli:
**Description**  
Support commands with positional arguments. 

Instead of doing **exact match** in `command_table`, check if the raw command **starts with** any item in the command table. If so, the item is considered as a match, and the remaining of the command are considered as positional arguments. 

For example, `az find vm create` won't exist in the command table, but the corresponding command should be `az find`.

```
raw command  : az find vm create
command table: az find
remaining    :         vm create
```

The debug log is:

```
> az find "vm create" --debug
Command arguments: ['find', 'vm create', '--debug']

Modules found from index for 'find': ['azure.cli.command_modules.find']
Loading command modules:
Name                  Load Time    Groups  Commands
find                      0.003         1         1
Total (1)                 0.003         1         1
These extensions are not installed and will be skipped: ['azext_ai_examples']
Loading extensions:
Name                  Load Time    Groups  Commands  Directory
ai-did-you-mean-this      0.011         1         1  C:\Users\xxx\.azure\cliextensions\ai-did-you-mean-this
Total (1)                 0.011         1         1
Loaded 2 groups, 2 commands.
Found a match in the command table.
Raw command  : find vm create
Command table: find
remaining    :      vm create
```

**Testing Guide**  
```
az find vm --debug
az find "vm create" --debug
```


yonzhan:
add to S173

PR #14187, opened 2020-07-02
[Core] Fix get_token() issue in msi login and `expiresIn` key error in cloud shell login credentials for track 2 SDK related commands

Juliehzl:
**Description**  

**Testing Guide**  

- Test in Cloud shell as https://github.com/Azure/azure-cli/pull/14027#discussion_r448716269
- `az login --identity` and try track 2 extension

**History Notes**  

[Component Name 1] BREAKING CHANGE: az command a: Make some customer-facing breaking change.  
[Component Name 2] az command b: Add some customer-facing feature.

---

This checklist is used to make sure that common guidelines for a pull request are followed.

- [ ] The PR title and description has followed the guideline in [Submitting Pull Requests](https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules#submitting-pull-requests).

- [ ] I adhere to the [Command Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md).

PR #14180, opened 2020-07-02
{Core} Fix #14175: Load ALWAYS_LOADED_EXTENSIONS correctly

jiasli:
**Description**  
Fix #14175: ai-did-you-mean-this extension is not loaded after a first run

Currently, if no extension corresponds to a command like `az account` in Command Index:

```
"account": [
    "azure.cli.command_modules.profile",
    "azure.cli.command_modules.resource"
],
```

`_update_command_table_from_extensions` won't be called, thus left `ALWAYS_LOADED_EXTENSIONS` ignored.

This PR fixes this issue by always calling `_update_command_table_from_modules` and `_update_command_table_from_extensions`.

**Testing Guide**  

```pwsh
> az extension add -n ai-did-you-mean-this

# If you already have ai-did-you-mean-this installed, delete `~/.azure/commandIndex.json`

# Build command index
> az account --debug

# Use command index
> az account --debug
...
[Thoth]: recommend_recovery_options: version: "2.8.0", command: "account", parameters: "['--debug']", extension: "None"
[Thoth]: User must agree to telemetry before failure recovery recommendations can be presented.
```

PR #14331, opened 2020-07-13
{Perf} Delay importing MSIAuthenticationWrapper and check_version_compatibility

jiasli:
**Description**  
Fix regressions introduced by
- #14180, which unconditionally imports `check_version_compatibility` and `requests`
- #14187, which unconditionally imports `MSIAuthenticationWrapper` and `azure.core.credentials`

This PR delays the imports and brings `az version` to its normal speed. 

**Testing Guide**  
```pwsh
# Before
> Measure-Command {az version}

TotalMilliseconds : 1132.3246

# After
> Measure-Command {az version}

TotalMilliseconds : 627.5202
```

### DIFF
--- a/src/azure-cli-core/azure/cli/core/__init__.py
+++ b/src/azure-cli-core/azure/cli/core/__init__.py
@@ -255,9 +255,6 @@ class MainCommandsLoader(CLICommandsLoader):
              Otherwise, the list will be extended using ALWAYS_LOADED_EXTENSIONS.
              If the extensions in the list are not installed, it will be skipped.
             """
-
-            from azure.cli.core.extension.operations import check_version_compatibility
-
             def _handle_extension_suppressions(extensions):
                 filtered_extensions = []
                 for ext in extensions:
@@ -299,6 +296,7 @@ class MainCommandsLoader(CLICommandsLoader):
 
                 for ext in allowed_extensions:
                     try:
+                        from azure.cli.core.extension.operations import check_version_compatibility
                         check_version_compatibility(ext.get_metadata())
                     except CLIError as ex:
                         # issue warning and skip loading extensions that aren't compatible with the CLI core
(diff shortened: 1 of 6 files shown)


yonzhan:
add to S173

src/azure-cli-core/azure/cli/core/__init__.py
```diff
@@ -299,6 +296,9 @@ def _filter_modname(extensions):
 
                 for ext in allowed_extensions:
                     try:
+                        # Import in the `for` loop because `allowed_extensions` can be []. In such case we
+                        # don't need to import `check_version_compatibility` at all.
+                        from azure.cli.core.extension.operations import check_version_compatibility
```
haroldrandom, line 301: Importing a module in a for loop is not a good practice.
How much can it benefit from importing in this for loop?

jsntcy: You can import module as many times as you want in one Python program, no matter what module it is. Every subsequent import after the first accesses the cached module instead of re-evaluating it. So there is no performance penalty.

---
In reply to: [454774982](https://github.com/Azure/azure-cli/pull/14331#discussion_r454774982) [](ancestors = 454774982)

jiasli: This is a tricky part. I just want to eliminate an extra `if allowed_extensions:`. Python has an [import cache](https://docs.python.org/3/reference/import.html#the-module-cache) which will handle "import in a for loop" nicely.

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -21,8 +21,6 @@
     is_windows, is_wsl
 from azure.cli.core.cloud import get_active_cloud, set_cloud_subscription
 
-from .adal_authentication import MSIAuthenticationWrapper
```
jsntcy, line 24: for patch(), we cannot use relative path for import statement?

jiasli, line 24: @jsntcy, relative import it is not recommended according to [PEP 8](https://www.python.org/dev/peps/pep-0008/#imports):

> Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages)...

Issue #14287, opened 2020-07-09
None type object errors due to not checking object exists


yungezz:
> ### `az feedback` auto-generates most of the information requested below, as of CLI version 2.0.62

**Describe the bug**

This is a typical and frequent error we met. use this issue to track reported cases.

**To Reproduce**

**Expected behavior**

**Environment summary**

**Additional context**

yonzhan:
None type object errors

yungezz:
this issue acts as a tracking record.

yungezz:
https://github.com/Azure/azure-cli/pull/14335

qwordy:
Does checking object existence fix this type of bug? Sometimes it conceals the bug and buries bombs in project.

PR #14335, opened 2020-07-13
{NoneType Error} Fix existing NoneType errors

houk-ms:
**Description**  

This pr is to fix the existing NoneType errors across command groups. The NoneType errors usually happen when a property is used without checking the existance of its father property. The typical error message is
`NoneType object has no attribute xxx`

The related issues includes what are listed in #14287, and the error records in Telemetry.

**Testing Guide**  

**History Notes**


yonzhan:
NoneType Error

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -1070,6 +1070,9 @@ def open_vm_port(cmd, resource_group_name, vm_name, port, priority=900, network_
 
     vm = get_vm(cmd, resource_group_name, vm_name)
     location = vm.location
+    if not vm.network_profile:
+        raise CLIError("Network profile not found for VM '{}'".format(vm_name))
```
qwordy, line 1074: Can we create a VM without network profile?

houk-ms:
#14987, #14986

PR #14336, opened 2020-07-13
{CI} unify azdev installation method

MyronFanQiu:
Our CI use three different ways to install azdev, which cause different version of azdev in our CI system. Refine all these methods to `azdev_setup.yml`.
- azdev_setup.yml
- install_azdev.sh
- pip install azdev in pipelines.yml directly


yonzhan:
add to S173

.azure-pipelines/templates/azdev_setup.yml
```diff
@@ -14,7 +14,7 @@ steps:
       chmod +x env/bin/activate
       . env/bin/activate
 
-      pip install -q azdev
+      pip install -q azdev==0.1.21
```
haroldrandom, line 17: no need to pin verison

linter_exclusions.yml
```diff
@@ -231,6 +231,16 @@ eventgrid event-subscription update:
     endpoint_type:
       rule_exclusions:
       - no_parameter_defaults_for_update_commands
+eventgrid partner topic event-subscription update:
```
MyronFanQiu, line 234: Discussed with service team offline. It would bring a breaking change. Meanwhile, the logic behind three update are same. It need some time to fix these violations.

.azure-pipelines/templates/azdev_setup.yml
MyronFanQiu: Unpin it. However, I think we should pin this version in case of potential bug in azdev causing the broken of our CI.

jsntcy: We can unpin the azdev version, but run Azure CLI CI as a gate test when releasing new azdev to guarantee the azdev release package quality.

PR #14436, opened 2020-07-21
[Config] az config: Add new `config` command module

### LINKED ISSUE #13981: [Configure] A more versatile `az configure`
`az configure` currently only supports setting and listing the `default` section:

```
> az configure -h

Command
    az configure : Manage Azure CLI configuration. This command is interactive.

Arguments
    --defaults -d      : Space-separated 'name=value' pairs for common argument defaults.
    --list-defaults -l : List all applicable defaults.  Allowed values: false, true.
    --scope            : Scope of defaults. Using "local" for settings only effective under current
                         folder.  Allowed values: global, local.  Default: global.
```

For other sections described in [Azure CLI configuration](https://docs.microsoft.com/en-us/cli/azure/azure-cli-configuration?view=azure-cli-latest), users will have to manually edit the `~/.azure/config` file, like 

```ini
[core]
collect_telemetry = no
no_color = yes

[logging]
enable_log_file = no
```

This is inconvenient. 

To improve the user experience, `az configure` needs to 

1. Support set, get and unset
2. Support sections other than `default`

jiasli:
## Description
Resolve #13981

Add new `config` command module. This PR relies on Knack change https://github.com/microsoft/knack/pull/217.

### Set

Disable color with `core.no_color`.

    az config set core.no_color=true

Hide warnings and only show errors with `core.only_show_errors`.

    az config set core.only_show_errors=true

Turn on client-side telemetry.

    az config set core.collect_telemetry=true

Turn on file logging and set its location.

    az config set logging.enable_log_file=true
    az config set logging.log_dir=~/az-logs

Set the default resource group to `myRG`.

    az config set default.group=myRG

Set the default resource group to `myRG` on a local scope.

    az config set default.group=myRG --local

### Get

Get all configurations.

    az config get

Get configurations in `core` section.

    az config get core

Get configuration of key `core.no_color`.
    
    az config get core.no_color

### Unset

Unset configurations of key `core.no_color`.

    az config unset core.no_color


yonzhan:
Config refactor

doc/sphinx/azhelpgen/doc_source_map.json
```diff
@@ -1,5 +1,6 @@
 {
     "az": "src/azure-cli/azure/cli/command_modules/profile/_help.py",
+    "confige": "src/azure-cli/azure/cli/command_modules/config/_help.py",
```
fengzhou-msft, line 3: extra `e`?

jiasli: Oh! Nice catch. Perhaps our CI should check it? @haroldrandom @MyronFanQiu

src/azure-cli/azure/cli/command_modules/config/custom.py
```diff
@@ -0,0 +1,88 @@
...
+
+    if not name:
+        # Only section
+        return items
+
+    # section.option
+    try:
+        return next(x for x in items if x['name'] == name)
+    except StopIteration:
+        raise CLIError("Configuration '{}' is not set.".format(key))
```
arrownj, line 73: Another option for this case is to return empty string. Personally, I prefer empty string for this case.

src/azure-cli/azure/cli/command_modules/config/custom.py
```diff
@@ -0,0 +1,88 @@
...
+        parts = k.split('.', 1)
+
+        if len(parts) == 1:
+            raise CLIError("usage error: [section].[name]")
+
+        section = parts[0]
+        name = parts[1]
+
+        with ScopedConfig(cmd.cli_ctx.config, local):
+            cmd.cli_ctx.config.remove_option(section, name)
```
arrownj, line 88: What will be returned if `key` is not set?

src/azure-cli/azure/cli/command_modules/config/custom.py
jiasli: Empty string means the config is actually set to `''`, which is different from the concept of "not set".

src/azure-cli/azure/cli/command_modules/config/custom.py
jiasli: It will ignore the operation and return `False`. See [`remove_option`](https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.remove_option).

jiasli: However, `configparser.NoSectionError` is ignored by Knack.

https://github.com/microsoft/knack/blob/148173e6fd02bd721920e6b33c9f508b4d584ae9/knack/config.py#L219-L223

```py
            try:
                existed = self.config_parser.remove_option(section, option)
                self.set(self.config_parser)
            except configparser.NoSectionError:
                pass
```

We'd better confirm with PM about the behavior later.

src/azure-cli/azure/cli/command_modules/config/custom.py
arrownj: Make sense to me. Maybe we should also confirm this with PM.

src/azure-cli/azure/cli/command_modules/config/custom.py
arrownj: Agree with your suggestion.

jiasli:
An excerpt from the design spec https://github.com/jiasli/azure-cli-spec/blob/master/az-config.md:

## Additional work - migrate to settings.json

`settings.json` is the latest unwritten convention of Microsoft open source project, like [Visual Studio Code](https://github.com/microsoft/vscode) and [Windows Terminal](https://github.com/microsoft/terminal).

### Visual Studio Code

The config file is at `C:\Users\xxx\AppData\Roaming\Code\User\settings.json`:

```jsonc
{
    "terminal.integrated.shell.windows": "C:\\WINDOWS\\System32\\WindowsPowerShell\\v1.0\\powershell.exe",
    "terminal.integrated.scrollback": 99999,
    "editor.minimap.enabled": false,
    ...
}
```

### Windows Terminal

The config file is at `C:\Users\xxx\AppData\Local\Packages\Microsoft.WindowsTerminal_8wekyb3d8bbwe\LocalState\settings.json`:

```jsonc
{
    "$schema": "https://aka.ms/terminal-profiles-schema",

    "defaultProfile": "{574e775e-4f2a-5b96-ac1e-a2962a402336}",

    "profiles":
    [
        {
            // Make changes here to the powershell.exe profile
            "guid": "{61c54bbd-c2c6-5271-96e7-009a87ff44bf}",
            "name": "Windows PowerShell",
            "commandline": "powershell.exe",
            "hidden": false
        },
        ...
    ],
    ...
}
```

Using `settings.json` will open up more possibilities like "hover" help (controlled by `editor.hover.enabled` in VSCode) using the `"$schema"` property:

image

PR #14438, opened 2020-07-21
{Linter} replace command with show comamnd and custom command with custom show command

MyronFanQiu:
`show` command should use `show_command` or `custom_show_command`.

Compared to `command` and `custom_command`, `show_command` and `custom_show_command` have extra error handler which should be harmless.


src/azure-cli/azure/cli/command_modules/vm/commands.py
```diff
@@ -477,7 +477,7 @@ def load_command_table(self, _):
         g.command('delete', 'delete')
 
     with self.command_group('vm monitor log', log_analytics_data_plane_sdk, client_factory=cf_log_analytics_data_plane) as g:
-        g.custom_command('show', 'execute_query_for_vm', transform=transform_log_analytics_query_output)
+        g.custom_command('show', 'execute_query_for_vm', transform=transform_log_analytics_query_output)  #pylint: disable=show-command
```
qwordy, line 480: why cannot use show command?

MyronFanQiu: It's a data plane operation and not the general show command.

yonzhan:
Linter

Issue #14056, opened 2020-06-22
Changes related to dedicated host group automatic placement


dochung4:
**Resource Provider**

Microsoft.Compute

**Description of Feature or Work Requested**

Today customers can place a VM on a dedicated host, by providing the ARM resource id of the dedicated host in the VM input.
The new feature is to support automatic placement of VMs and VMScaleSets on a dedicated host group. This means in the VM/VMScaleSet input, customers can specify the ARM resource id of the dedicated host group, on which they want their VMs/VMScaleSets to be placed, and Azure will select an appropriate dedicated host under the dedicated host group for each VM and allocate the VM.
There are four major changes related to this feature:
1. "supportAutomaticPlacement" is a new property introduced under DedicatedHostGroup.properties. Customers can specify either true or false. It defaults to true, if no input is provided. If a dedicated host group has automatic placement enabled, VMs or VMScaleSets can be placed on the dedicated host group using automatic placement.

2. GET dedicated host group now supports instance view query parameter. Dedicated host group instance view will return instance view of the dedicated hosts under the dedicated host group.

3. "hostGroup" is a new property introduced under VM.properties and VMScaleSet.properties. The input should be the resource id of the dedicated host group, on which the customer wants his VM/VMScaleSet placed using automatic placement.

4. "assignedHost" is a new property introduced under VMInstanceView and VMScaleSetVMInstanceView. It is the resource id of the dedicated host, on which the queried VM/VMScaleSetVM is placed using atuomatic placement.

Design document: https://microsoft.sharepoint.com/:w:/t/ComputeVM/EV1bTk0XSvNLtvDnLn9PSpoBH2NvlfAFQhduoCSJvatreg?e=6AiqWp

**Minimum API Version Required**

2020-06-01

**Swagger Link**

https://github.com/Azure/azure-rest-api-specs/pull/9684

**Target Date**

August 2020

yungezz:
thanks for @dochung4 for detail info. Just to confirm, is service feature deployed to Prod yet? Or pls share us timeline, then we can start testing and python SDK release, then CLI work.

dochung4:
Would testing in Canary work for you? Everything has been deployed to Canary, except for one small bit, which is expected to be deployed to Canary as well this week. The feature requires subscription whitelisting. I can whitelist subscriptions, if you could provide me subscriptionIds.

I am not completely sure about the timeline for prod and in which regions we are planning to enable first. But we are targeting for preview in August.

PR #14439, opened 2020-07-21
[Compute] Dedicated host group automatic placement

qwordy:
**Description**  

Resolve https://github.com/Azure/azure-cli/issues/14056

1. New parameter `--automatic-placement` in `vm host group create`
```
    --automatic-placement            : Specify whether virtual machines or virtual machine scale
                                       sets can be placed automatically on the dedicated host group.
                                       Automatic placement means resources are allocated on
                                       dedicated hosts, that are chosen by Azure, under the
                                       dedicated host group. The value is defaulted to true when not
                                       provided.  Allowed values: false, true.
```
2. New parameter `--host-group`  in `vmss create`
```
    --host-group                        : Name or ID of dedicated host group that the virtual
                                          machine scale set resides in.
```
3. New command `vm host group get-instance-view`

I will make a **breaking change**. These are parameters from `vm create`. However, `--host-group` is not a parameter in REST API, it is used to complete the info of host if user only tells host name. Now, in the new API version, host group is supported in service.
I plan to abandon old `--host-group`, let `--host` only support ID, and the new `--host-group` represents the dedicated host group that the virtual machine resides in.
```
--host

Name or ID of the dedicated host this VM will reside in. If a name is specified, a host group must be specified via --host-group.

--host-group

Name of the dedicated host group containing the dedicated host this VM will reside in.
```
New spec:
```
        "host": {
          "$ref": "#/definitions/SubResource",
          "description": "Specifies information about the dedicated host that the virtual machine resides in. <br><br>Minimum api-version: 2018-10-01."
        },
        "hostGroup": {
          "$ref": "#/definitions/SubResource",
          "description": "Specifies information about the dedicated host group that the virtual machine resides in. <br><br>Minimum api-version: 2020-06-01. <br><br>NOTE: User cannot specify both host and hostGroup properties."
        },
```

@yungezz @arrownj What do you think about this change? Do you agree?

In the past 30 days, number of invocations:
`vm create`, 474384
`vm create --host-group`, 39
`vm create --host`, 46

> "supportAutomaticPlacement" is a new property introduced under DedicatedHostGroup.properties. Customers can specify either true or false. It defaults to true, if no input is provided. If a dedicated host group has automatic placement enabled, VMs or VMScaleSets can be placed on the dedicated host group using automatic placement.
GET dedicated host group now supports instance view query parameter. Dedicated host group instance view will return instance view of the dedicated hosts under the dedicated host group.
"hostGroup" is a new property introduced under VM.properties and VMScaleSet.properties. The input should be the resource id of the dedicated host group, on which the customer wants his VM/VMScaleSet placed using automatic placement.
"assignedHost" is a new property introduced under VMInstanceView and VMScaleSetVMInstanceView. It is the resource id of the dedicated host, on which the queried VM/VMScaleSetVM is placed using atuomatic placement.

**Testing Guide**  

```
az vm host group create -n hg -c 1 -g fy --automatic-placement -l centraluseuap
az vm host create -n h --host-group hg -d 0 -g fy --sku DSv3-Type1 -l centraluseuap
az vm create -g fy -n vm1 --image centos --size Standard_D4s_v3 --host <host_id> -l centraluseuap
az vm create -g fy -n vm2 --image centos --size Standard_D4s_v3 --host-group hg -l centraluseuap
```

**History Notes**  

---


yonzhan:
Compute

src/azure-cli/azure/cli/command_modules/vm/tests/latest/test_vm_commands.py
```diff
@@ -4046,6 +4047,8 @@ def test_dedicated_host_e2e(self, resource_group, resource_group_location):
         self.assertTrue(instance_view["assetId"])
         self.assertTrue(instance_view["availableCapacity"])
 
+        self.cmd('vm host group get-instance-view -g {rg} -n {host-group}')
```
arrownj, line 4050: add check ?

yungezz:
thanks @qwordy ,  excellent PR description to detail problems and changes!

after learning [host, host group](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/dedicated-hosts) and (service change)[https://github.com/Azure/azure-cli/issues/14056], I propose below change. It's not necessary to be a breaking change, since meaning of host-group param not change: it means the host group this VM would like to go into. Pls have a look and check does it make sense to you, we can have discussion offline also.

My proposal:
- keep host, host-group params.
- change the logic to below
  - if `--host` is name
     -- host-group is name, then it's existing usge: host-group+ host is used to specific host
     -- host-group is id, throw error.

  - if `--host` is id
     -- host-group is name, throw error. host is enough to specific host, no need of host-group.
     -- host-group is id, throw error. same as above

  - if `--host` is None and `--automatic-placement` true
     -- host-group is name, it will be VM.host
     -- host-group is id, it will be VM.host

  - if `--host` is None and `--automatic-placement` false
     -- host-group is name, error
     -- host-group is id, throw error.

src/azure-cli/azure/cli/command_modules/vm/tests/latest/test_vm_commands.py
qwordy: Sure. Testing is not finished yet

qwordy:
Very clear! `--host-group` has multiple meanings now. Will it be a burden for user to understand?

yungezz:
> Very clear! `--host-group` has multiple meanings now. Will it be a burden for user to understand? I need to think about how to write help for `--host-group`.

valid concern :) let's work together to see any better way

PR #14454, opened 2020-07-22
{CI} Pin azdev to latest version

MyronFanQiu:
There is a PR https://github.com/Azure/azure-cli/pull/14438 that we have to release azdev and fix CI at the same time. I don't want to create a broken status for our CI system. Let's pin azdev version first.


MyronFanQiu:
@haroldrandom  We need change Azdev Release Pipeline as well to add one more step to bump version in our CI system.

yonzhan:
CI

Issue #2999, opened 2017-04-26
[Functions] Listing Azure Functions information - Keys


kamil-mrzyglod:
Hi,

is there any possibility(or maybe ongoing work) to manage Azure Functions using Azure CLI? I'm talking mostly about adding/retrieving keys for a specific function or listing URLs in release/deploy scenarios.

yugangw-msft:
@kamil-mrzyglod, we are waiting for service API available. If no ETA, I will submit a PR to support at least `adding/retrieving keys for a specific function`

panchagnula:
@ahmelsayed is this supported using FUNC CLI if so can you share the details here?

chriswue:
Seeing that this has been open for 2.5 years - is there any plan to get traction on this? The func core tools have some basic support for this (see https://github.com/Azure/azure-functions-core-tools/pull/190) but the output is insufficient since I require access to some of the details such as the contents of the function.json

anthonychu:
Related #4041

PR #14465, opened 2020-07-22
[AppService] functionapp: Added ability to manage keys for function apps and functions. Added function-specific commands.

gzuber:
**Description**  

Resolves https://github.com/Azure/azure-cli/issues/4041
Resolves https://github.com/Azure/azure-cli/issues/2999

This PR adds the ability for users to manage their functionapp and function keys. It also allows users to show and delete individual functions. It adds the following commands:

```
az functionapp keys set -g group -n appname --key-name keyname --key-type functionKeys --key-value keyvalue -s slotname
az functionapp keys list -g group -n appname -s slotname
az functionapp keys delete -g group -n appname --key-name keyname --key-type functionKeys -s slotname

az functionapp function show -g group -n appname --function-name functionname
az functionapp function delete -g group -n appname --function-name functionname

az functionapp function keys set -g group -n appname --function-name functionname --key-name keyname --key-value keyvalue -s slotname
az functionapp function keys list -g group -n appname --function-name functionname -s slotname
az functionapp function keys delete -g group -n appname --function-name functionname --key-name keyname -s slotname

where --key-value and --slot are optional wherever used
```

I want to address some of the decisions I made while creating this:
 - Why am I redefining the structure of the KeyInfo class from the python auto-generated code?
   This is unfortunate. I'm not exactly sure how this happened, but the ARM API expects a KeyInfo to look like this:
```
{
    properties: {
        name: MyName,
        value: MyValue
    }
}
```
The python auto-generated code expects a KeyInfo to look like this:
```
{
    name: MyName,
    value: MyValue
}
```
My best guess is that the ARM API is the one at fault here, because based on the [spec provided](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/web/resource-manager/Microsoft.Web/stable/2018-02-01/WebApps.json#L19991), I don't believe `properties` should be there. Either way, fixing it one way or the other would be a breaking change. I think that in the meantime I've come up with the most succinct way of making this work, and it would be very easy to remove if anything changes/is fixed.
 - Why are there both functionapp keys and functionapp function keys?
   I did this because to manage functionapp keys, you _must_ provide a key_type. You _cannot_ provide a key_type for function keys. In order to manage function keys, you _must_ provide a function_name. You _cannot_ provide a function_name for functionapp keys. I originally tried to combine them by having the user provide a function_name for function specific keys, but if the user provides a function_name, they can't provide a key_type and vice-versa. I thought this muddied the waters a bit. We're also introducing a group for function-specific commands so that seemed to work.
 - Why did I add a new example dotnet functionapp?
   sample_dotnet_function is a C# app that would need to be built during deployment. This capability is only available for linux. sample_csx_function_httptrigger is a csx functionapp that will have a function ready to run when deployed.

**Testing Guide**  

I've introduced three new groups of tests that exercise each new command added. You can run all the new tests like so:
```
azdev test FunctionAppKeysTests FunctionAppFunctionKeysTests FunctionAppFunctionTests
```

**History Notes**  

[AppService] az functionapp keys: Added commands that allow users to manage their function app keys
[AppService] az functionapp function: Added commands that allow users to manage their individual functions
[AppService] az functionapp function keys: Added commands that allow users to manage their function keys

---


yonzhan:
AppService

yungezz:
hi @qwordy could you pls help to review this PR? thanks

anthonychu:
Looks good. Thanks @gzuber for the thorough explanation. I can't think of any problems with this.

Will docs automatically be regenerated to reflect the changes?

What are the possible values of `--key-type`?

/cc @fabiocav

gzuber:
I'm not sure what docs would/should be updated for this. I've found [this](https://docs.microsoft.com/en-us/azure/azure-functions/functions-cli-samples?toc=%2Fcli%2Fazure%2Ftoc.json&bc=%2Fcli%2Fazure%2Fbreadcrumb%2Ftoc.json&view=azure-cli-latest) which I'm assuming wouldn't get updated automatically. But it's also not a reference but rather just some examples. I've added examples of using the commands to the help menu.



A `--key-type` can be of type `functionKeys`, `masterKey`, or `systemKey`. That's the way they appear in the response body of the list command, and what the ARM API expects as parameters. It's documented [here](https://github.com/Azure/azure-cli/pull/14465/files#diff-e780e1de20871bbfd717f323d9c53cfaR769) in the help menu. 

Another option would be to have `--key-type` be of value `function`, `master`, or `system` and translate them under the hood. I didn't know if that would be better considering they don't appear that way in response bodies, I'm open to feedback about that.

gzuber:
@qwordy -- Credential scan is failing because there are keys in the HTTP response in the test recordings. This is intentional because it's testing that I can set and query keys. Is there a way to whitelist these credentials so it doesn't fail?

I'm also getting "Request to https://azhackcli.blob.core.windows.net/whl/hack-0.4.2-py2.py3-none-any.whl failed with 403" but I assume this is transient (I just can't find a way to re-run the tests for some reason).

qwordy:
If you want to whitelist credentials, modify this file, azure-cli\scripts\ci\credscan\CredScanSuppressions.json

gzuber:
@qwordy thanks for your help! This should be good to merge

qwordy:
> 
> 
> @qwordy thanks for your help! This should be good to merge

 Code owner review required
Waiting on code owner review from fengzhou-msft and/or haroldrandom.

gzuber:
@fengzhou-msft @haroldrandom 
Whenever you get a chance, would appreciate a review thanks!

PR #14478, opened 2020-07-23
[Extension] Support automatically installing an extension if the extension of a command is not installed

fengzhou-msft:
**Description**  

When a command failed in `parser._check_value` due to a command word is not in the choice list of its previous word, such as `az vm abc` would fail because `abc` is not a choice under `vm`. This failure could be caused by:
1. Typo, e.g. `az accuont show` or use `get` instead of the correct `show`.
3. The command is from an extension but the extension is not installed.

Previously, we cannot tell the difference and the error message is like this:

> az: 'bluprint' is not in the 'az' command group. See 'az --help'. If the command is from an extension, please make sure the corresponding extension is installed. To learn more about extensions, please visit https://docs.microsoft.com/en-us/cli/azure/azure-cli-extensions-overview

This PR adds a search for the unmatched command in all the extension commands if a close match for the wrong command is not found in all module commands (indicating it's not likely a typo). If the command (not including options) matches a complete extension command, based on config, we can automatically install the extension with or without prompt. Then the command will continue to run by default or users can choose to rerun by themselves.

A mapping of all extension commands to extension names will be fetched from a storage account and stored in ~/.azure/extCmdIndex.json. It is valid for 10 days and will only be refreshed when a parse error happens. An example of the file:
```
{
       "aks": {
                "create": "aks-preview",
                "update": "aks-preview",
                "app": {
                    "up": "deploy-to-azure"
                },
                "use-dev-spaces": "dev-spaces"
        },
        ...
}
```

Configuration:
Both config options are under extension section:
1. use_dynamic_install
yes_prompt: it will use dynamic extension install and prompt to user whether they would like to install the needed extension.
yes_without_prompt: it will install the needed extension directly without confirmation from users.
no: it falls back to previous behavior and just throws an error. This is the default for now.

2. run_after_dynamic_install
yes/true: It’s the default and will continue to run the command after extension is installed.
no/false: It will stop when extension is installed and ask the user to rerun the command.

You can also set with environment variables (AZURE_EXTENSION_USE_DYNAMIC_INSTALL, AZURE_EXTENSION_RUN_AFTER_DYNAMIC_INSTALL)and they have higher priority.

**Testing Guide**  

1.set env var to turn on the feature:
export AZURE_EXTENSION_USE_DYNAMIC_INSTALL=yes_promt
2.run a command from an extension not installed to chekc the behavior.
3.run a real error command to chekc the new clear error message. 

---


yonzhan:
Extension

src/azure-cli-core/azure/cli/core/parser.py
```diff
@@ -280,37 +280,157 @@ def parse_known_args(self, args=None, namespace=None):
...
+                    'extension', 'use_dynamic_install', 'no') if cli_ctx else 'no'
+                if use_dynamic_install.lower() not in ['no', 'yes_prompt', 'yes_without_prompt']:
+                    use_dynamic_install = 'no'
+                if use_dynamic_install.lower() != 'no' and not candidates:
+                    # Check if the command is from an extension
+                    from azure.cli.core.util import roughly_parse_command
+                    cmd_list = self.prog.split() + self._raw_arguments
+                    command_str = roughly_parse_command(cmd_list[1:])
+                    ext_name = self._search_in_extension_commands(command_str)
+                    if ext_name:
```
arrownj, line 357: what if `ext_name` is `None` ?

fengzhou-msft: If `ext_name` is `None`, `error_msg` will be `None` and below code will use the `{prog}: '{value}' is not in the '{prog}' command group.` for error message.

fengzhou-msft:
> Is it possible to add test for this change ?

Test added.

src/azure-cli-core/azure/cli/core/parser.py
```diff
@@ -280,37 +280,167 @@ def parse_known_args(self, args=None, namespace=None):
...
+            try:
+                if isinstance(cmd_chain[part], str):
+                    return cmd_chain[part]
+                cmd_chain = cmd_chain[part]
+            except KeyError:
+                return None
+        return None
+
+    def _get_extension_use_dynamic_install_config(self):
+        cli_ctx = self.cli_ctx or (self.cli_help.cli_ctx if self.cli_help else None)
```
haroldrandom, line 341: cli_ctx seems could be `None` and then line 342 would crash.

src/azure-cli-core/azure/cli/core/parser.py
```diff
@@ -280,37 +280,167 @@ def parse_known_args(self, args=None, namespace=None):
...
+        if not EXT_CMD_TREE.data:
+            import requests
+            from azure.cli.core.util import should_disable_connection_verify
+            try:
+                response = requests.get(
+                    'https://azurecliextensionsync.blob.core.windows.net/cmd-index/extensionCommandTree.json',
+                    verify=(not should_disable_connection_verify()),
+                    timeout=300)
+            except Exception as ex:  # pylint: disable=broad-except
+                logger.info("Request failed for extension command tree: %s", str(ex))
```
haroldrandom, line 302: Would log at warning be better?

src/azure-cli-core/azure/cli/core/parser.py
```diff
@@ -280,37 +280,167 @@ def parse_known_args(self, args=None, namespace=None):
...
+                return None
+            if response.status_code == 200:
+                EXT_CMD_TREE.data = response.json()
+                EXT_CMD_TREE.save_with_retry()
+            else:
+                logger.info("Error when retrieving extension command tree. Response code: %s", response.status_code)
```
haroldrandom, line 308: Would log at warning level be better?

src/azure-cli-core/azure/cli/core/parser.py
fengzhou-msft: line 342 has the logic `if cli_ctx else 'no'` to handle `None`, it will return `None`.

src/azure-cli-core/azure/cli/core/parser.py
fengzhou-msft: I will modify in upcoming PR. For now, let's just fallback to previous behavior when something is wrong in this feature.

Issue #14473, opened 2020-07-22
az login hangs on WSL


gxtaillon:
## Describe the bug

Running `az login` hangs, no output is shown. This is on Debian 10 over WSL on Windows 1809.

Running `az login --debug` logs the following link, opening manually and logging in works fine.
`Open browser with url: https://login.microsoftonline.com/common/oauth2/authorize?....`

**Command Name**
`az login`

**Errors:**
None

## To Reproduce:
Steps to reproduce the behavior. Note that argument values have been redacted, as they may contain sensitive information.

- `az login`

## Expected Behavior
The browser opens.

## Environment Summary
```
Linux-4.4.0-17763-Microsoft-x86_64-with-debian-10.4
Python 3.6.10
Installer: DEB

azure-cli 2.9.1

```

yonzhan:
add to S174

jiasli:
I tried Debian WSL on Microsoft Store but couldn't repro.

![image](https://user-images.githubusercontent.com/4003950/88357413-c1034d00-cd9d-11ea-81eb-708ed40e2c9a.png)

The code for opening a browser is at 

https://github.com/Azure/azure-cli/blob/04210197837bbceb547f86df7f086a42c96e0d2a/src/azure-cli-core/azure/cli/core/util.py#L526

Does it work if you run `powershell.exe` in WSL? It should print something like

```
$ powershell.exe
Windows PowerShell
Copyright (C) Microsoft Corporation. All rights reserved.

Try the new cross-platform PowerShell https://aka.ms/pscore6

PS Microsoft.PowerShell.Core\FileSystem::\\wsl$\Debian\home\user3>
```

But indeed, the current `az login` process for auth code flow gives too little information and sometimes leads to confusion, as discussed in https://github.com/MicrosoftDocs/azure-docs-cli/issues/2093.

gxtaillon:
@jiasli Running this spins up the loading icon on the mouse for about half a second, and then nothing happens.

```bash
gtaillon@debian:/mnt/c/Users/gtaillon$ powershell.exe -Command Start-Process "https://google.com"
gtaillon@debian:/mnt/c/Users/gtaillon$ which powershell.exe
/mnt/c/Windows/System32/WindowsPowerShell/v1.0/powershell.exe
```
However this works (from https://github.com/PowerShell/PowerShell/issues/5715#issuecomment-400857048):
```powershell
[...]

PR #14505, opened 2020-07-24
[Profile] az login: Add back the warning that a browser is opened

jiasli:
**Description**  

Revert the change in #10851. The removal of the warning message that a browser is opened leads to confusions: https://github.com/MicrosoftDocs/azure-docs-cli/issues/2093, #14473.

If the browser fails to open (due to some unknown reasons), Azure CLI will hang forever and the user won't know what to do next. 

```
> az login
_
```
With this PR, the warning is shown again:

```
> az login
The default web browser has been opened at https://login.microsoftonline.com/common/oauth2/authorize. Please continue the login in the web browser.
If no web browser is available or if the web browser fails to open, use device code flow with `az login --use-device-code`.
```

**Testing Guide**  
```
az login
```


yonzhan:
add to S173

yonzhan (approved):
LGTM

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -1272,6 +1272,12 @@ def log_message(self, format, *args):  # pylint: disable=redefined-builtin,unuse
         results['no_browser'] = True
         return
 
+    # Emit a warning to inform that a browser is opened.
+    # Only show the path part of the URL and hide the query string.
+    logger.warning("The default web browser has been opened at %s. Please continue the login in the web browser.\n"
+                   "If no web browser is available or if the web browser fails to open, use device code flow "
```
evelyn-ys, line 1278: If the web browser does fail to open, will CLI run into this line or hang over line 1269 `succ = open_page_in_browser(url)` ?

evelyn-ys (approved):

jsntcy (approved):

src/azure-cli-core/azure/cli/core/_profile.py
jiasli: If the web browser fails to open, `open_page_in_browser` should return `False`. This warning is for the scenario where `open_page_in_browser` returned `True`, but the browser is not brought to the foreground for some reason. (This mainly happens on some Linux distributions or MacOS.)

PR #14508, opened 2020-07-24
{CI} Run linter based on diff

MyronFanQiu:
Since we want to enable medium and low rule as well. But we don't plan to suppress them all for those old commands. So we have to run linter based on the diff so that customers won't receive too much noisy message.


yonzhan:
add to S173

PR #14513, opened 2020-07-24
{Role} az ad app credential list: Update help

tikicoder:
clarification.  Currently, the list has the description, List an application's password or certificate credentials.  To me, that implies you can recover the secret/cert using the command.  However, that is not the case from my understanding.  This returns I would say more Metadata around the password or certificate credentials.

**Description**  
Clarification around that the secret/cert is not returned.


yonzhan:
add to S174

src/azure-cli/azure/cli/command_modules/role/_help.py
```diff
@@ -85,7 +85,7 @@
 
 helps['ad app credential list'] = """
 type: command
-short-summary: list an application's password or certificate credentials
+short-summary: list an application's password or certificate credentials (The secret/cert stored is not recoverable)
```
jiasli, line 88:
```suggestion
short-summary: List an application's password or certificate credential metadata. (The content of the password or certificate credential is not retrievable.)
```

evelyn-ys (approved):

fengzhou-msft (approved):

jiasli (approved):

PR #14546, opened 2020-07-28
{Role} Remove redundant --id parameter from example

### LINKED ISSUE #14043: Create operation example refers to ambiguous --id parameter
The second example of `az ad app create` (`create`) command makes a reference to an `--id` parameter. This parameter has no relevance to the `create` command. At least it is missing from the **required** and from **optional** parameters. Not only that, but Azure AD does not support creating application registrations with user provided `id`.  One can only guess, the `--id` is only there to compensate missing id from the role defitniion json (because Roles must have IDs). Or it is there to totally confuse the user reading the documentation. Anyway the second example must be fixed and explanation for `--id` parameter by the `create` operation must be provided.

---
#### Document Details

* ID: 921aa920-c25f-5227-aabb-49116da38fa7
* Version Independent ID: 9b5a53ad-edf8-4d89-f8e6-cde01b7866f6
* Content: [az ad app](https://docs.microsoft.com/en-us/cli/azure/ad/app?view=azure-cli-latest#az-ad-app-create)
* Content Source: [src/azure-cli/azure/cli/command_modules/role/_help.py](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/role/_help.py)
* Service: **active-directory**
* GitHub Login: @rloutlaw

jiasli:
**Description**  
Fix #14043: Create operation example refers to ambiguous `--id` parameter

As `--id` is not a valid parameter of `az ad app create`, it is recognized as `--identifier-uris` due to [Argument abbreviations (prefix matching)](https://docs.python.org/3/library/argparse.html#argument-abbreviations-prefix-matching).

Also because `argparse.py` doesn't check duplicated parameters (same parameter provided twice) in `_StoreAction`,

https://github.com/python/cpython/blob/c82dda1e08c4b74ca24f88d6a549d93108c319cf/Lib/argparse.py#L928-L929

```py
    def __call__(self, parser, namespace, values, option_string=None):
        setattr(namespace, self.dest, values)
```

the second `--identifier-uris https://mytestapp.websites.net` overwrites the first `--id e042ec79-34cd-498f-9d9f-123456781234`. So `--id e042ec79-34cd-498f-9d9f-123456781234` never makes its way into the actual execution of the command.

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/role/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/role/_help.py
@@ -36,7 +36,7 @@ examples:
         }]
   - name: Create an application with a role
     text: |
-        az ad app create --id e042ec79-34cd-498f-9d9f-123456781234 --display-name mytestapp --identifier-uris https://mytestapp.websites.net --app-roles @manifest.json
+        az ad app create --display-name mytestapp --identifier-uris https://mytestapp.websites.net --app-roles @manifest.json
         ("manifest.json" contains the following content)
         [{
             "allowedMemberTypes": [


yonzhan:
Role

PR #14568, opened 2020-07-29
[ARM] az deployment mg/tenant what-if: Add support to management group and tenant level deployment What-If

shenglol:
**Description**  

The PR adds new commands/parameters to support MG-level and tenant-level deployment What-If.

**Testing Guide**  

`az deployment mg what-if -m <mg_id> -f <path_to_deployment_file> -l <location>`
`az deployment tenant what-if -f <path_to_deployment_file> -l <location>`
`az deployment mg create -m <mg_id> -f <path_to_deployment_file> -l <location> -c`
`az deployment tenant create -f <path_to_deployment_file> -l <location> -c`

**History Notes**  

[ARM] az deployment mg/tenant what-if: Add support to management group and tenant level deployment What-If
[ARM] az deployment mg/tenant create: Add --confirm-with-what-if/-c parameter.
[ARM] az deployment mg/tenant create: Add --what-if-result-format/-r parameter.
[ARM] az deployment mg/tenant create: Add --what-if-exclude-change-types/-x parameter.

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/resource/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/resource/_help.py
@@ -284,7 +284,7 @@ parameters:
   - name: --name -n
     short-summary: The deployment name.
   - name: --what-if-result-format -r
-    short-summary: The format of What-If results. Applicable when --confirm-with-what-if is set.
+    short-summary: The format of What-If results. Applicable when `--confirm-with-what-if` is set.
 examples:
   - name: Create a deployment at subscription scope from a remote template file, using parameters from a local JSON file.
     text: >
@@ -419,7 +419,7 @@ parameters:
   - name: --name -n
     short-summary: The deployment name.
   - name: --what-if-result-format -r
-    short-summary: The format of What-If results. Applicable when --confirm-with-what-if is set.
+    short-summary: The format of What-If results. Applicable when `--confirm-with-what-if` is set.
 examples:
   - name: Create a deployment at subscription scope from a remote template file, using parameters from a local JSON file.
     text: >
@@ -584,7 +584,7 @@ parameters:
   - name: --mode
     short-summary: The deployment mode.
   - name: --what-if-result-format -r
-    short-summary: The format of What-If results. Applicable when --confirm-with-what-if is set.
+    short-summary: The format of What-If results. Applicable when `--confirm-with-what-if` is set.
 examples:
   - name: Create a deployment at resource group from a remote template file, using parameters from a local JSON file.
     text: >
@@ -790,7 +790,7 @@ parameters:
   - name: --location -l
     short-summary: The location to store the deployment metadata.
   - name: --what-if-result-format -r
-    short-summary: The format of What-If results. Applicable when --confirm-with-what-if is set.
+    short-summary: The format of What-If results. Applicable when `--confirm-with-what-if` is set.
 examples:
   - name: Create a deployment at management group from a remote template file, using parameters from a local JSON file.
     text: >
@@ -944,7 +944,7 @@ parameters:
   - name: --location -l
     short-summary: The location to store the deployment metadata.
   - name: --what-if-result-format -r
-    short-summary: The format of What-If results. Applicable when --confirm-with-what-if is set.
+    short-summary: The format of What-If results. Applicable when `--confirm-with-what-if` is set.
 examples:
   - name: Create a deployment at tenant scope from a remote template file, using parameters from a local JSON file.
     text: >
--- a/src/azure-cli/azure/cli/command_modules/resource/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/resource/_help.py
@@ -737,6 +737,39 @@ examples:
     text: az deployment mg validate --management-group-id testmg --location WestUS --template-file {template-file}
 """
 
+helps['deployment mg what-if'] = """
+type: command
+short-summary: Execute a deployment What-If operation at management group scope.
+parameters:
+  - name: --parameters -p
+    short-summary: Supply deployment parameter values.
+    long-summary: >
+        Parameters may be supplied from a file using the `@{path}` syntax, a JSON string, or as <KEY=VALUE> pairs. Parameters are evaluated in order, so when a value is assigned twice, the latter value will be used.
+        It is recommended that you supply your parameters file first, and then override selectively using KEY=VALUE syntax.
+  - name: --template-file -f
+    short-summary: The path to the template file.
+  - name: --template-uri -u
+    short-summary: The URI to the template file.
+  - name: --management-group-id -m
+    short-summary: The management group id to create deployment at.
+  - name: --name -n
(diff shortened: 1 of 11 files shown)


src/azure-cli/azure/cli/command_modules/resource/_utils.py
```diff
@@ -5,47 +5,61 @@
...
+_management_group_pattern = (
+    r"^\/?providers\/Microsoft.Management\/managementGroups\/(?P<management_group_id>[\w\d_\.\(\)-]+)"
+)
 _subscription_pattern = r"^\/?subscriptions\/(?P<subscription_id>[a-f0-9-]+)"
 _resource_group_pattern = r"^\/resourceGroups\/(?P<resource_group_name>[-\w\._\(\)]+)"
 _relative_resource_id_pattern = r"^\/providers/(?P<relative_resource_id>.+$)"
 
 
-def parse_resource_id(resource_id):
+def split_resource_id(resource_id):
```
zhoxing-ms, line 16: It is suggested to add some comments and examples to this method to help others understand its logic and function more clearly when using it.

src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -756,6 +789,8 @@
     short-summary: The deployment name.
   - name: --location -l
     short-summary: The location to store the deployment metadata.
+  - name: --what-if-result-format -r
+    short-summary: The format of What-If results. Applicable when --confirm-with-what-if is set.
```
zhoxing-ms, line 793:
```suggestion
    short-summary: The format of What-If results. Applicable when `--confirm-with-what-if` is set.
```

src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -877,6 +943,8 @@
...
```
zhoxing-ms, line 947:
```suggestion
    short-summary: The format of What-If results. Applicable when `--confirm-with-what-if` is set.
```

src/azure-cli/azure/cli/command_modules/resource/_utils.py
shenglol: Thanks for the suggestion, added examples.

src/azure-cli/azure/cli/command_modules/resource/_help.py
shenglol: Fixed.

src/azure-cli/azure/cli/command_modules/resource/_help.py
shenglol: Done.

yonzhan:
ARM

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -347,6 +347,20 @@ def load_arguments(self, _):
...
+                   arg_type=deployment_what_if_exclude_change_types_type,
+                   help="Space-separated list of resource change types to be excluded from What-If results. Applicable when --confirm-with-what-if is set.",
+                   min_api="2019-10-01")
+
+    with self.argument_context('deployment mg what-if') as c:
+        c.argument('deployment_name', arg_type=deployment_create_name_type)
+        c.argument('no_prompt', arg_type=no_prompt)
+        c.argument('result_format', arg_type=deployment_what_if_result_format_type)
+        c.argument('no_pretty_print', arg_type=deployment_what_if_no_pretty_print_type)
+        c.argument('exclude_change_types', arg_type=deployment_what_if_exclude_change_types_type)
```
zhoxing-ms, line 363: Because these parameters are the same in different scopes, can they be configured centrally? 
such as:
```
  for scope in ['mg', 'tenant', 'sub', 'group']:
        with self.argument_context('deployment  {}  what-if'.format(scope)) as c:
                c.argument('deployment_name', arg_type=deployment_create_name_type)
                ......
```

shenglol: Although `deployment group` has a different set of parameters, `sub`, `mg` and `tenant` can be grouped together, and that is also the case for `create`, `validate`, `list`, etc. I also think that we should do some refactoring here, but we'd update other argument contexts, too, so I might leave that to a separate PR.

zhoxing-ms: Sounds good~

src/azure-cli/azure/cli/command_modules/resource/commands.py
```diff
@@ -335,6 +337,8 @@ def load_command_table(self, _):
                          table_transformer=deployment_validate_table_format, exception_handler=handle_template_based_exception)
         g.custom_command('create', 'deploy_arm_template_at_tenant_scope', supports_no_wait=True, validator=process_deployment_create_namespace,
                          table_transformer=transform_deployment, exception_handler=handle_template_based_exception)
+        g.custom_command('what-if', 'what_if_deploy_arm_template_at_tenant_scope', validator=process_deployment_create_namespace,
```
jsntcy, line 340: It's better to define a CLIArgumentType for what-if for reuse instead of duplicate multiple places.

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -347,6 +347,20 @@ def load_arguments(self, _):
         c.argument('handle_extended_json_format', arg_type=extended_json_format_type,
                    deprecate_info=c.deprecate(target='--handle-extended-json-format/-j'))
         c.argument('no_prompt', arg_type=no_prompt)
+        c.argument('confirm_with_what_if', arg_type=deployment_what_if_confirmation_type, min_api="2019-10-01")
+        c.argument('what_if_result_format', options_list=['--what-if-result-format', '-r'],
+                   arg_type=deployment_what_if_result_format_type, min_api="2019-10-01")
+        c.argument('what_if_exclude_change_types', options_list=['--what-if-exclude-change-types', '-x'],
```
jsntcy, line 353: why is this argument named as what_if_exclude_change_types, while the next one is named as exclude_change_types?

src/azure-cli/azure/cli/command_modules/resource/_utils.py
jsntcy: Not sure if you can leverage existing method "parse_resource_id" in azure.mgmt.core as it seems very similar to yours

---
In reply to: [462710200](https://github.com/Azure/azure-cli/pull/14568#discussion_r462710200) [](ancestors = 462710200)

shenglol: It is similar, but it may not be suitable to use it here, because what the `parse_resource_id` method returns is a dictionary with unpredictable number of key/value pairs depending on how many child resource types a resource ID has. If using it, I have to process the dictionary to construct the relative resource ID, which appears to be a harder task.

PR #14605, opened 2020-07-31
{Storage} Fix secondary storage account read access

Juliehzl:
**Description**  

**Testing Guide**  

**History Notes**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/storage/_validators.py
+++ b/src/azure-cli/azure/cli/command_modules/storage/_validators.py
@@ -169,7 +169,7 @@ def validate_client_parameters(cmd, namespace):
         n.sas_token = n.sas_token.lstrip('?')
 
     # account name with secondary
-    if n.account_name.endswith('-secondary'):
+    if n.account_name and n.account_name.endswith('-secondary'):
         n.location_mode = 'secondary'
         n.account_name = n.account_name[:-10]
 
--- a/src/azure-cli/azure/cli/command_modules/storage/_validators.py
+++ b/src/azure-cli/azure/cli/command_modules/storage/_validators.py
@@ -168,6 +168,11 @@ def validate_client_parameters(cmd, namespace):
     if n.sas_token:
         n.sas_token = n.sas_token.lstrip('?')
 
+    # account name with secondary
+    if n.account_name.endswith('-secondary'):
+        n.location_mode = 'secondary'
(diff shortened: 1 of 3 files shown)


src/azure-cli/azure/cli/command_modules/storage/_validators.py
```diff
@@ -168,6 +168,11 @@ def validate_client_parameters(cmd, namespace):
     if n.sas_token:
         n.sas_token = n.sas_token.lstrip('?')
 
+    # account name with secondary
+    if n.account_name.endswith('-secondary'):
```
arrownj, line 172: How does user know this trick?

src/azure-cli/azure/cli/command_modules/storage/_validators.py
```diff
@@ -168,6 +168,11 @@ def validate_client_parameters(cmd, namespace):
...
+        n.location_mode = 'secondary'
+        n.account_name = n.account_name[:-10]
+
```
fengzhou-msft, line 175: Consider this as a temporary solution. Exposing `--location-mode` to users would be more clear.

src/azure-cli/azure/cli/command_modules/storage/_validators.py
Juliehzl: specify --account-name `xxxx-secondary` 
here is request url:
urllib3.connectionpool : Starting new HTTPS connection (3): zuhlrs-secondary.blob.core.windows.net:443

src/azure-cli/azure/cli/command_modules/storage/_validators.py
Juliehzl: because in track 2, the accepted params for client is account url not account name and location mode. Current design could fix with track2 SDk, too.

Juliehzl:
> better to have a test. 😸

sure. will add it later.

src/azure-cli/azure/cli/command_modules/storage/_validators.py
jsntcy: When will this logic be hit?

Juliehzl: when customer specify --account-name as `xxx-secondary`

yonzhan:
Storage

PR #14673, opened 2020-08-05
{App Service} Fix az webapp up not getting correct ASP to use

calvinsID:
**Description**  
Fixes az webapp up not getting correct appservice plan to use. 

**Testing Guide**  
Download a hello world app, and go into the root directory
Create an app with a paid sku: `az webapp up --sku S1 -n name`
Create another app with a free sku: `az webapp up --sku F1 -n name2`

Creating the free sku app should work. Instead, it gives the following error: `Cannot update the site 'name' because it uses alwaysOn feature which is not allowed in the target compute node`

---


yonzhan:
App Service

PR #14448, opened 2020-07-21
[ARM] az ts: Add new commands for template specs ; [ARM] az deployment : Add support for --template-spec -s

detienne20:
**Description**  

This PR enables template specs (ts) commands to create, update, list, show, export, and delete template specs.  Also added is support for template-specs with Az CLI Deployment. Microsoft.Resources is updated to use the latest Python SDK. The latest Python SDK required some functions to take in different parameters and account for different return types. All test cases involving template-spec have been updated (i.e--template-specs are now created dynamically for az deployment tests). 

Note: #14241 has been closed. This PR would include/override the changes made there.  

**Testing Guide**  

az ts

   Create: 
          
          az ts create -g testRG --name TemplateSpecName -l WestUS --display-name 
            "MyDisplayName" --description "Simple template spec"

          az ts create -g testRG -n TemplateSpecName -v 1.0 -l WestUS
              --f templateSpec.json --display-name "SimpleTemplate" --description "Basic 
               template spec"  --version-description "Version of basic template spec"

   Update:
         
        az ts create -g testrg -n MyTemplateSpecName  -v 1.0 -f updatedFile.json
       
        az ts update -s resourceID -f updatedTemplate.json

        az ts update -g ExistingRG -n ExistingName -v 3.0 -f updatedTemplate.json 
            -d "New parent display name"  --description "New parent description" 
            --version-description "New child description"
   
   Show: 
        
        az ts show -s resourceID

        az ts show-g testrg --name TemplateSpecName

        az ts show -g testrg --name TemplateSpecName --version VersionName

   List: 
       
        az ts list

        az ts list -g SpecifiedResourceGroup

        az ts list --subscription SpecifiedSubscription
         
        az ts list -g testRG -n ParentTemplateSpecName

   Export: 
      
       az ts export -g testrg --name TemplateSpecName --version VersionName -- 
         output-folder C:/path/

       az ts export -s resourceID --output-folder C:/path/

       az ts export -g testrg --name TemplateSpecName --output-folder C:/path/
     
   Delete: 
     
       az ts delete --template-spec resourceID

       az ts delete -g MyResourceGroup --name TemplateSpecName

       az ts delete -g MyResourceGroup --name TemplateSpecName --version 
         VersionName

Az Deployments with --template-spec 

   Resource Group:
     
      az deployment group validate --resource-group "resource-group-name" --template-spec 
        resource-id-to-template-spec

      az deployment group create --resource-group "resource-group-name" --template-spec 
        resource-id-to-template-spec
[...]

PR #14678, opened 2020-08-06
{ARM} Bump azure-mgmt-resource SDK from 10.1.0 to 10.2.0

detienne20:
**Description**  

Bump azure-mgmt-resource from 10.1.0 to 10.2.0 
For PR: #14448   

**Cause of the problem**

The method `provider = client.providers.get(namespace)` is called to retrieve the api-version list of all the ResourceTypes in the specified namespace. Whether the return value of the `ProviderResourceType` in the api-version list contains the `default_api_version` field will affect the api-version fetch logic:
1. The `default_api_version` field is not included in the `ProviderResourceType` returned by the older version before `10.2.0` of SDK, so the latest api-version will be taken (stable version is preferred).
2. The SDK with version `10.2.0` returns the `default_api_version` in `ProviderResourceType`, so the logic to fetch api-version has changed, the value of `default_api_version` will be used directly.

**Fix getting resource API version in ARM cmdlt:**  @eladperets 

Latest swagger updates added the 'defaultApiVersion' property to the GET resource providers response.
The old code was trying to use the default API version and only than fall back to the API versions specified in the manifest.
This was wrong because:
1. default API version was not part of the response (until the latest SDK update), so this code was not doing anything
2. default API versions can be null
3. default API version is mostly used for internal purposes and must be included in the list of available API versions.

This change addresses this issue.

**Testing Guide**  

**History Notes**


detienne20:
@eladperets

detienne20:
ping @eladperets

yonzhan:
ARM

src/azure-cli-core/azure/cli/core/commands/arm.py
```diff
@@ -1164,11 +1164,11 @@ def get_arm_resource_by_id(cli_ctx, arm_id, api_version=None):
             from azure.cli.core.parser import IncorrectUsageError
             raise IncorrectUsageError('Resource type {} not found.'.format(resource_type_str))
         try:
-            # if the service specifies, use the default API version
```
MyronFanQiu, line 1167: This change should be reviewed by more folks.

src/azure-cli-core/azure/cli/core/commands/arm.py
```diff
@@ -1164,11 +1164,11 @@ def get_arm_resource_by_id(cli_ctx, arm_id, api_version=None):
...
-            api_version = rt.default_api_version
-        except AttributeError:
-            # if the service doesn't specify, use the most recent non-preview API version unless there is only a
-            # single API version. API versions are returned by the service in a sorted list
+            # Use the most recent non-preview API version unless there is only a
+            # single API version. API versions are returned by the service in a sorted list.
             api_version = next((x for x in rt.api_versions if not x.endswith('preview')), rt.api_versions[0])
```
Juliehzl, line 1169: With such change, if the rp only has preview api versions, it will get an exception because you only try to get latest stable api version. But in previous design if the default version is preview version, it still can get an preview api version.
I will prefer preview api version could be an backup in such scenario.

zhoxing-ms: Yeah, but in fact this is how the current logic is implemented. When `(x for x in rt.api_versions if not x.endswith('preview'))` cannot get a value, `next()` will take `rt.api_versions[0]`, and `rt.api_versions[0]` is the latest preview api-version.

zhoxing-ms:
> Does this change has impacts to extensions which import `resource` SDK ?

@arrownj There should be no impact, because this version of SDK only returns more fields, but does not affect the original logic. 
The reason for this breaking this time is that the logic of method `get_arm_resource_by_id` itself is not correct, so it only affects where it is called.

Issue #14598, opened 2020-07-30
Accessing createdTime info under az resource list


chenlomis:
> ### `az feedback` auto-generates most of the information requested below, as of CLI version 2.0.62

**Describe the bug**
Unable to access info under the createdTime attribute

**To Reproduce**
"So I have been tasked with creating some cleanup scripts for our testing account.  I ran the az resource list and realize there is a createdTime attribute, but they are all null.  Is this something that is being added and just not used?  Is there a flag to get access to that info?  That would be a really great way combine that with createdBy.  You could easily do some automated hey you have these 5 resources over 2 weeks old are they still in use if not please clean them up or they will be cleaned on x day.  Then you could auto delete and use the cli to export to a template and send it to them.  I tried the Get-AzResource and it does not appear to have a created time attribute."

**Expected behavior**
"I have ideas/options that I could do, but I am more intrigued by if the createdTime is something coming out or in preview, or something that goes against the YAGNI Principal (You ain't gonna need it). "

**Environment summary**

**Additional context**

tikicoder:
Since this ticket was created on my behalf please let me know if you have questions.

yonzhan:
add to S174

PR #14704, opened 2020-08-07
[ARM] `az resource list`: Extend the return data of `createdTime`, `changedTime` and `provisioningState`

zhoxing-ms:
Issue: #14598

**Description**  

Extend the return data of `createdTime`, `changedTime` and `provisioningState` for `az resource list`.
These fields are returned before, but there is no value. Now make up these values.
REST API has supported: [link](https://docs.microsoft.com/en-us/rest/api/resources/resources/list#genericresourceexpanded)

**Testing Guide**  

Execute command `az resource list` to see whether `createdTime`, `changedTime` and `provisioningState` returns

**History Notes**  

---


yonzhan:
ARM

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -1776,7 +1776,9 @@ def list_resources(cmd, resource_group_name=None,
     odata_filter = _list_resources_odata_filter_builder(resource_group_name,
                                                         resource_provider_namespace,
                                                         resource_type, name, tag, location)
-    resources = rcf.resources.list(filter=odata_filter)
+
+    expand = "createdTime,changedTime,provisioningState"
+    resources = rcf.resources.list(filter=odata_filter, expand=expand)
```
Juliehzl, line 1781: is there a way to get expand value not hard code? In addition, could we expose the parameter to customer to let customer choose if they want to expand or not?

zhoxing-ms: Because these fields are all returned before, but they are no value. Therefore, making them optional fields can only reduce the output data, such usage scenarios may be rare.

Juliehzl (approved):

PR #14712, opened 2020-08-07
{Monitor} restrict antlr4 import scope

bluca:
antlr4 python bindings are scarcely available, so restrict the
import scope to avoid breaking commands that do not need it
when it is not available

See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=962952


yonzhan:
add to S174

MyronFanQiu:
Thanks for the contribution!

PR #14718, opened 2020-08-08
{Profile} use UTC timezone when converting expiration timestamps to strings

bluca:
datetime.fromtimestamp will use the machine's locale timezone by
default, causing expiration timestamps to vary depending on the
local setting. Use UTC to ensure they are stable.

Among other things, test_get_raw_token from test_profile_custom.py fails if the machine is not using UTC as the timezone


yonzhan:
add to S174

jiasli:
Thank you for the contribution. 

However, forcing the `expiresOn` to be in UTC will introduce a BREAKING CHANGE which we are trying to avoid without absolute necessity. 

I am fixing the test in https://github.com/Azure/azure-cli/pull/15131 and cherry-picked https://github.com/Azure/azure-cli/pull/14718/commits/ea0b891a456f8462ed67a4277b7ed6c74f5962e8 so that you receive credit for this PR.

PR #14734, opened 2020-08-11
[Network] az network dns export: export FQDN for MX, PTR, NS and SRV type instead of relative path

### LINKED ISSUE #14412: Azure DNS zone export missing trailing "." on MX, NS, PTR, SRV record targets
**Describe the bug**
When exporting a DNS zone from Azure DNS, MX/NS/PTR/SRV record type targets don't get a trailing "." appended which means the zone file is incorrect.

**To Reproduce**
az network dns zone create -g <resource group> -n <zone name>
az network dns record-set mx add-record -e contoso.com -p 0 -n mxtest -g <resource group> -z <zone name>
az network dns record-set ns add-record -d ns.contoso.com -n nstest -g <resource group> -z <zone name>
az network dns record-set ptr add-record -d "ptr.contoso.com" -n ptrtest -g <resource group> -z <zone name>
az network dns record-set srv add-record -r 443 -p 0 -t srv.contoso.com -w 0 -n srvtest -g <resource group> -z <zone name>
az network dns zone export -g <resource group> -n <zone name>

The output of the export of those records will look as follows:

mxtest 3600 IN MX 0 contoso.com
nstest 3600 IN NS ns.contoso.com
ptrtest 3600 IN PTR ptr.contoso.com
srvtest 3600 IN SRV 0 0 443 srv.contoso.com

**Expected behavior**
The output should have a trailing "." on those 4 records as below to make them FQDNs instead of relative names. Azure DNS treats all targets in these record types as FQDNs so the CLI export should treat them as such. Without this trailing "." it will cause these record type values to be incorrect when loading this export into another DNS server such as BIND or even when re-importing using "az network dns zone import" by wrongly appending the zone name.

mxtest 3600 IN MX 0 contoso.com.
nstest 3600 IN NS ns.contoso.com.
ptrtest 3600 IN PTR ptr.contoso.com.
srvtest 3600 IN SRV 0 0 443 srv.contoso.com.

**Environment summary**

Windows-10-10.0.19041-SP0
Python 3.6.6
Installer: MSI

azure-cli 2.9.1

**Additional context**
The CNAME and SOA record types seem to behave correctly already. Please make MX, NS, PTR, SRV target values behave as those. Do note there should only be exactly 1 trailing "." at the end. If the existing record contains one already, then we should not append another.

MyronFanQiu:
**Description**  

Fix #14412 

**Testing Guide**  

**History Notes**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/network/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/network/custom.py
@@ -1942,11 +1942,11 @@ def export_zone(cmd, resource_group_name, zone_name, file_name=None):
             elif record_type == 'cname':
                 record_obj.update({'alias': record.cname.rstrip('.') + '.'})
             elif record_type == 'mx':
-                record_obj.update({'preference': record.preference, 'host': record.exchange})
+                record_obj.update({'preference': record.preference, 'host': record.exchange.rstrip('.') + '.'})
             elif record_type == 'ns':
-                record_obj.update({'host': record.nsdname})
+                record_obj.update({'host': record.nsdname.rstrip('.') + '.'})
             elif record_type == 'ptr':
-                record_obj.update({'host': record.ptrdname})
+                record_obj.update({'host': record.ptrdname.rstrip('.') + '.'}})
             elif record_type == 'soa':
                 record_obj.update({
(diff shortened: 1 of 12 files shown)


yonzhan:
Network

PR #14746, opened 2020-08-11
WIP Do not merge: AppService testing environment setup

calvinsID:
**Description**

WIP Do not merge

We would like the convenience to be able to test & work on our AppService/Webapp commands faster, without the need to wait for the deployment of our backend API changes and SDK updates before working on CLI changes. 

In order to do so, we want to check for these two environment variables:
- `AZURE_CLI_PRIVATE_GEO`: The endpoint of our testing environment
- `AZURE_CLI_CLIENT_CERT_PATH`: The path to the certificate used to authenticate to our testing environment

If these two environment variables are set, we want the CLI to make any calls to our testing environment instead of production environment.

We were wondering if this is something we can do?

**Testing Guide**  

**History Notes**  

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/adal_authentication.py
+++ b/src/azure-cli-core/azure/cli/core/adal_authentication.py
@@ -3,6 +3,7 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 
+import os
 import time
 import requests
 import adal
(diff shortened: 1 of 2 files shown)


yonzhan:
Use private geomaster

yungezz:
hi @calvinsID could you pls share more context about the change in description? thanks.

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -482,6 +482,23 @@ def logout_all(self):
...
+                {
+                    "id": private_sub,
+                    "name": "custom sub",
+                    "state": "Enabled",
+                    "user": {
+                        "name": "fakeuser",
+                        "type": "user"
+                    },
+                    "isDefault": True,
+                    "tenantId": "72f988bf-86f1-41af-91ab-2d7cd011db47",
```
yungezz, line 497: psl not hardcode/public the tenant  id

arrownj (commented):
This requirement is very specific, I doubt that whether we need to support from azure-cli-core side. The solution here is straight forward for this specific requirement, however we also need to consider more cases. For example whether the supported API version also in `AZURE_CLI_PRIVATE_GEO` match the default ones. How to teach user to use these environment variables?

PR #14752, opened 2020-08-12
{KeyVault} Add deprecation messages for soft-delete related commands and parameters

bim-msft:
Add deprecation messages to meet the service team's requirements. Texts were copied from the documentation.

`az keyvault create/update -h`
image

`az keyvault delete -h`
image

`az keyvault key delete -h`
image

`az keyvault secret delete -h`
image

`az keyvault certificate delete -h`
image

---


bim-msft:
> the change itself LGTM. we need sync with Jack on when to removing the message. it's little annoying.

Yes, those messages are really lengthy and they should be removed as soon as possible when the service side finish the change.

src/azure-cli/azure/cli/command_modules/keyvault/_params.py
```diff
@@ -104,8 +110,13 @@ class CLIJsonWebKeyOperation(str, Enum):
                                                          'Vnet/subnet pairs or subnet resource ids.')
 
     with self.argument_context('keyvault update') as c:
-        c.argument('enable_soft_delete', arg_type=get_three_state_flag(),
-                   deprecate_info=c.deprecate(expiration='2.11.0', hide='2.11.0'))
+        c.argument('enable_soft_delete', arg_type=get_three_state_flag(), deprecate_info=Deprecated(
```
fengzhou-msft, line 113: Why not use `deprecate_info=c.deprecate(message_func=)`?

bim-msft: Good point, it works!
Actually I decided in the same way before, but I saw the code of `deprecate` function:
image
It seems that `message_func` would be overwritten by the build-in function. So I use `Deprecated` directly. But anyway, this works, I will change the code.

bim-msft: Updated.

fengzhou-msft (approved):

PR #14759, opened 2020-08-12
Add special CLI name.

tibortakacs:
**Description**  
In order to use azure-cli-core in 3rd-party CLIs, the CLI name has to be configured. Getting an `AzCLI` object with default settings, however, is a great Azure CLI Core feature. Therefore, we recommend this backward compatible change to accept `cli_name` with `az` default value in `get_default_cli(...)` function.


yonzhan:
special CLI name

yungezz:
hi @tibortakacs are you using azure cli core as a library?

PR #14802, opened 2020-08-17
{Pywin32} Remove win32 search path patch

### LINKED ISSUE #11613: Improvements to Python search paths on Windows
I was reviewing some CLI sources and noticed this section of code:

https://github.com/Azure/azure-cli/blob/6887c5badd7246fea1c4586e835224d39acfcc4e/src/azure-cli-telemetry/azure/cli/telemetry/__init__.py#L15-L22

The Windows package should be using a `python._pth` file (possibly with the version number in there), which means you can simply add these directories to that file and they'll be put onto `sys.path`.

You also shouldn't need the `%PATH%` modification, but should `import pywin32_bootstrap` [(source)](https://github.com/mhammond/pywin32/blob/master/win32/Lib/pywin32_bootstrap.py) instead. That will also handle the [change on Python 3.8](https://docs.python.org/3.8/whatsnew/3.8.html#bpo-36085-whatsnew) that prevents `%PATH%` being used for DLL resolution, which is presumably what this is attempting.

fengzhou-msft:
**Description**  

Resolve #11613 
`pywin32.pth` is now honored by the embeddable Python in MSI and the patch is no longer needed.

**Testing Guide**  

**History Notes**  

---

### DIFF
--- a/src/azure-cli-telemetry/azure/cli/telemetry/__init__.py
+++ b/src/azure-cli-telemetry/azure/cli/telemetry/__init__.py
@@ -6,22 +6,7 @@
 import sys
 import os
 import subprocess
-
-try:
-    import portalocker
-except ImportError:
-    if os.name != 'nt':
-        raise
-    # To fix the import error of win32con and win32file in Windows. Manually add win32 search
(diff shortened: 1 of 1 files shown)


yonzhan:
Pywin32

PR #14803, opened 2020-08-17
[Upgrade] Add `az upgrade` command

fengzhou-msft:
**Description**  

Add `az upgrade` to update azure-cli on all platforms and with `--all` enabled, extensions will also be updated.

Auto-upgrade can be enabled with `az config set auto-upgrade.enable=yes`(default is `no`). You can also specify whether to update extensions (`auto-upgrade.all`, default is `yes`), show prompt message(`auto-upgrade.prompt`, default is `yes`). The check against the latest version is done through the local cached `~/.azure/versionCheck.json` file. The data is valid for 10 days (the expected wait period for the 3-week release cycle of Azure CLI). The check happens after the execution of a command.

Add `az extension list-versions` command to list all available versions of an extension.

**Testing Guide**  

**History Notes**  

[Upgrade] Add `az upgrade` command to upgrade azure cli and extensions
[Extension] Add `az extension list-versions` command to list all available versions of an extension

---


yonzhan:
az upgrade

yungezz:
hi @arrownj could you pls review this PR? thanks

src/azure-cli-core/azure/cli/core/extension/operations.py
```diff
@@ -356,6 +358,45 @@ def list_available_extensions(index_url=None, show_details=False):
     return results
 
 
+def list_versions(extension_name, index_url=None):
+    index_data = get_index_extensions(index_url=index_url)
+
+    try:
+        exts = index_data[extension_name]
+    except Exception:
+        raise CLIError('Extension {} not found.'.format(extension_name))
```
yungezz, line 367: should this error out or silently swallow? will this func be used other than az upgrade?

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -171,25 +161,56 @@ def _update_latest_from_pypi(versions):
     return versions, success
 
 
+def get_latest_from_github(package_path='azure-cli'):
+    try:
+        import requests
+        git_url = "https://raw.githubusercontent.com/Azure/azure-cli/master/src/{}/setup.py".format(package_path)
```
yungezz, line 167: how to accommodate air-gapped cloud?

src/azure-cli/azure/cli/__main__.py
```diff
@@ -74,3 +74,25 @@ def cli_main(cli, args):
...
+            if VERSIONS[_VERSION_UPDATE_TIME]:
+                import datetime
+                version_update_time = datetime.datetime.strptime(VERSIONS[_VERSION_UPDATE_TIME], '%Y-%m-%d %H:%M:%S.%f')
+                if datetime.datetime.now() > version_update_time + datetime.timedelta(days=10):
+                    get_cached_latest_versions()
+                from distutils.version import LooseVersion
+                if LooseVersion(VERSIONS['versions']['core']['local']) < LooseVersion(VERSIONS['versions']['core']['pypi']):  # pylint: disable=line-too-long
+                    import subprocess
+                    import platform
+                    logger.warning("New Azure CLI version available. Running 'az upgrade' to update automatically.")
```
yungezz, line 90: A warning should be ok right?

src/azure-cli/azure/cli/__main__.py
```diff
@@ -74,3 +74,25 @@ def cli_main(cli, args):
...
+                    update_all = az_cli.config.getboolean('auto-upgrade', 'all', True)
+                    prompt = az_cli.config.getboolean('auto-upgrade', 'prompt', True)
+                    cmd = ['az', 'upgrade', '--all', str(update_all)]
+                    if not prompt:
+                        cmd.append('-y')
+                    subprocess.call(cmd, shell=platform.system() == 'Windows')
```
yungezz, line 96: how to show error to user when subprocess fail?

src/azure-cli/azure/cli/command_modules/util/custom.py
```diff
@@ -27,3 +29,108 @@ def show_version(cmd):  # pylint: disable=unused-argument
...
+        logger.debug("Failed to get the latest version. %s", str(ex))
+    exts = [ext.name for ext in get_extensions(ext_type=WheelExtension)] if update_all else []
+
+    exit_code = 0
+    if update_cli:
+        logger.warning("Your current Azure CLI version is %s. Latest version available is %s.",
+                       local_version, latest_version)
+        from knack.prompting import prompt_y_n
+        if not yes:
+            confirmation = prompt_y_n("Please check the release notes first: https://docs.microsoft.com/"
```
yungezz, line 65: the url is broken to 2 lines?

src/azure-cli/azure/cli/command_modules/util/custom.py
```diff
@@ -27,3 +29,108 @@ def show_version(cmd):  # pylint: disable=unused-argument
...
+        else:
+            logger.warning(UPGRADE_MSG)
+    if exit_code:
+        telemetry.set_failure("CLI upgrade failed.")
+        sys.exit(exit_code)
+
+    for ext_name in exts:
+        try:
+            logger.warning("Checking update for %s", ext_name)
+            subprocess.call(['az', 'extension', 'update', '-n', ext_name], shell=platform.system() == 'Windows')
```
arrownj, line 133: Why `shell=true` for windows platform ?

src/azure-cli/azure/cli/command_modules/util/custom.py
```diff
@@ -27,3 +29,108 @@ def show_version(cmd):  # pylint: disable=unused-argument
...
+        elif installer == 'DOCKER':
+            logger.warning("Exit the container to pull latest image with 'docker pull mcr.microsoft.com/azure-cli' "
+                           "or run 'pip install --upgrade azure-cli' in this container")
+        elif installer == 'MSI':
+            logger.debug("Update azure cli with MSI from https://aka.ms/installazurecliwindows")
+            exit_code = subprocess.call(['powershell.exe', 'Start-Process powershell -Wait -Verb runAs -ArgumentList "Invoke-WebRequest -Uri https://aka.ms/installazurecliwindows -OutFile AzureCLI.msi;Start-Process msiexec.exe -Wait -ArgumentList \'/I AzureCLI.msi\';Remove-Item AzureCLI.msi"'])  # pylint: disable=line-too-long
+        else:
+            logger.warning(UPGRADE_MSG)
+    if exit_code:
+        telemetry.set_failure("CLI upgrade failed.")
```
arrownj, line 127: There will be no telemetry for successful upgrade ?

src/azure-cli/azure/cli/__main__.py
```diff
@@ -74,3 +74,25 @@ def cli_main(cli, args):
     telemetry.set_init_time_elapsed("{:.6f}".format(init_finish_time - start_time))
     telemetry.set_invoke_time_elapsed("{:.6f}".format(invoke_finish_time - init_finish_time))
     telemetry.conclude()
+    try:
+        if sys.argv[1] != 'upgrade' and az_cli.config.getboolean('auto-upgrade', 'enable', False):
```
arrownj, line 78: Why do we put this at the end of CLI executing ? If the user set `auto-upgrade` and not prompt config, however when he run a command, CLI will execute with the old version then upgrade it, right ?

src/azure-cli/azure/cli/command_modules/util/custom.py
```diff
@@ -27,3 +29,108 @@ def show_version(cmd):  # pylint: disable=unused-argument
...
+            if in_cloud_console():
+                raise CLIError("az upgrade is not supported in Cloud Shell.")
+            apt_update_cmd = 'apt-get update'.split()
+            az_update_cmd = 'apt-get install --only-upgrade -y azure-cli'.split()
+            if os.geteuid() != 0:  # pylint: disable=no-member
+                apt_update_cmd.insert(0, 'sudo')
+                az_update_cmd.insert(0, 'sudo')
+            subprocess.call(apt_update_cmd)
+            logger.debug("Update azure cli with '%s'", " ".join(apt_update_cmd))
+            exit_code = subprocess.call(az_update_cmd)
```
arrownj, line 82: What if previous cmd such as `apt-get update`, `apt-get install --only-upgrade ...` fail ? It seems that `exit_code` should accept all `subprocess`'s return code.

src/azure-cli-core/azure/cli/core/extension/operations.py
fengzhou-msft: This function is not used in `az upgrade` now as we always update to the latest extension version. This error happens when users type the wrong extension name.

src/azure-cli-core/azure/cli/core/util.py
fengzhou-msft: If we are going to support `az upgrade` in air-gapped cloud, we may not be able to use system package managers. We could put all the packages in a storage account and download/install the corresponding package based on the user system.

src/azure-cli/azure/cli/command_modules/util/custom.py
fengzhou-msft: Good catch. Added check on the exit code of the first command.

src/azure-cli/azure/cli/__main__.py
fengzhou-msft: Yes. We don't want the user to wait for the upgrade when typing the first command. 

With the prompt, users can check the result of the command first and then type yes to continue or no to decline the upgrade if they would like to run a second command without the wait for upgrade.

With no prompt config, I have suppressed the stdout, so it won't fill up the screen to block the actual result of the command. Besides, if the command result is assigned to a variable, the variable would only contain the actual command result in this case.

src/azure-cli/azure/cli/command_modules/util/custom.py
fengzhou-msft: No summary messsage for success, but result will still be set when command finsihes.

src/azure-cli/azure/cli/command_modules/util/custom.py
fengzhou-msft: `az` cannot be found without `shell=true` on Windows.

src/azure-cli/azure/cli/command_modules/util/custom.py
fengzhou-msft: They are concatenated in output.

src/azure-cli/azure/cli/__main__.py
fengzhou-msft: A warning will not impact the command result for use in script.

src/azure-cli/azure/cli/__main__.py
fengzhou-msft: Error will show in `stderr`.

src/azure-cli-core/azure/cli/core/extension/operations.py
yungezz: ok. I saw it used by list-available-version, so error out is ok. My original point is that az upgrade internal failure should not error out which might impact user in silent upgrading scenario.

src/azure-cli/azure/cli/command_modules/util/custom.py
arrownj: You mean the upgrade result or the command execute result ?

src/azure-cli-core/azure/cli/core/util.py
yungezz: is it necessary to call it out in az upgrade or cloud endpoint discovery design spec?

src/azure-cli/azure/cli/command_modules/util/custom.py
```diff
@@ -27,3 +29,117 @@ def show_version(cmd):  # pylint: disable=unused-argument
...
+        logger.warning("Your current Azure CLI version is %s. %s", local_version, latest_version_msg)
+        from knack.prompting import prompt_y_n
+        if not yes:
+            confirmation = prompt_y_n("Please check the release notes first: https://docs.microsoft.com/"
+                                      "cli/azure/release-notes-azure-cli\nDo you want to continue?", default='y')
+            if not confirmation:
+                telemetry.set_success("Upgrade stopped by user")
+                return
+        installer = os.getenv(_ENV_AZ_INSTALLER)
+        if installer == 'DEB':
```
yungezz, line 72: upper() then equal?

yungezz (approved):

src/azure-cli/azure/cli/command_modules/util/custom.py
fengzhou-msft: Here the command is `az upgrade`, telemetry records the command execute result. If it is triggered in auto-upgrade, there will be 2 telemetry records, 1 for the original command, 1 for `az upgrade`.

arrownj: got it.

src/azure-cli-core/azure/cli/core/extension/operations.py
fengzhou-msft: In silent upgrading scenario (auto-upgrade with no prompt), the exit code and `stdout`of a command will not be affected, but warnings and errors from `az upgrade` will be shown in `stderr`.

src/azure-cli/azure/cli/command_modules/util/custom.py
fengzhou-msft: updated

src/azure-cli-core/azure/cli/core/util.py
fengzhou-msft: yes, will add to cloud endpoint discovery design spec.

src/azure-cli/azure/cli/command_modules/util/custom.py
```diff
@@ -27,3 +29,120 @@ def show_version(cmd):  # pylint: disable=unused-argument
...
+
+        if installer == 'DEB':
+            from azure.cli.core.util import in_cloud_console
+            if in_cloud_console():
+                raise CLIError("az upgrade is not supported in Cloud Shell.")
+            apt_update_cmd = 'apt-get update'.split()
+            az_update_cmd = 'apt-get install --only-upgrade -y azure-cli'.split()
+            if os.geteuid() != 0:  # pylint: disable=no-member
+                apt_update_cmd.insert(0, 'sudo')
+                az_update_cmd.insert(0, 'sudo')
```
zhoxing-ms, line 82: What happens if the current user does not have sudo permission?

src/azure-cli/azure/cli/command_modules/util/custom.py
```diff
@@ -27,3 +29,120 @@ def show_version(cmd):  # pylint: disable=unused-argument
...
+        elif installer == 'DOCKER':
+            logger.warning("Exit the container to pull latest image with 'docker pull mcr.microsoft.com/azure-cli' "
+                           "or run 'pip install --upgrade azure-cli' in this container")
+        elif installer == 'MSI':
+            logger.debug("Update azure cli with MSI from https://aka.ms/installazurecliwindows")
+            exit_code = subprocess.call(['powershell.exe', "Start-Process msiexec.exe -Wait -ArgumentList '/i https://aka.ms/installazurecliwindows'"])  # pylint: disable=line-too-long
+        else:
+            logger.warning(UPGRADE_MSG)
+    if exit_code:
+        telemetry.set_failure("CLI upgrade failed.")
```
zhoxing-ms, line 133: Could we consider record the error information when the upgrade fails in the telemetry?

src/azure-cli/azure/cli/command_modules/util/custom.py
fengzhou-msft: An error will be thrown by the Linux system and shown to the user.

src/azure-cli/azure/cli/command_modules/util/custom.py
fengzhou-msft: Good suggestion. There is an error improvement PR that revamps telemetry. I will work with @houk-ms to improve error tracking here.

src/azure-cli/azure/cli/command_modules/util/custom.py
zhoxing-ms: If the error message is given by the OS, and the message does not confuse the user, then I think it is fine~
It may be better that let the users input their admin password to obtain sudo permissions.

PR #14829, opened 2020-08-18
[AppService] Fix #14788: az webapp create not getting correct webapp when names are substrings

### LINKED ISSUE #14788: Webapp:az webapp create fails because Azure CLI reports app service exists in a different resource group then specified (which is not the case)
## Describe the bug
We can no longer run out deployment scripts against out PROD environment since Azure CLI has been updated to `2.10.1`
**Command Name**
`az webapp create`

**Errors:**
```
The webapp '<app-service>' exists in resource group '<group>_devtest' and does not match the value entered '<group>'. Please re-run command with the correct parameters.
```

## To Reproduce:

- `az webapp create --resource-group {} --name {} --plan {} --only-show-errors`

## Expected Behavior
The command should pass without making any changes since the web app already exists

## Environment Summary
```
Linux-4.19.104-microsoft-standard-x86_64-with-debian-buster-sid
Python 3.6.10
Installer: DEB

azure-cli 2.10.1

```
## Additional Context
The app service we create here already exists in the resource group `<group>` and not in `<group>_devtest`.  `<group>_devtest` is a resource group we use for DEV purposes and there is and there is an app service named  '<app-service>-dev' in there. As of this version of the Azure CLI some check seems to indicate that the app service is in a different resource group then it really is
`az webapp show --resource-group {} --name {}` with the same name an resource group returns the app service object as well

calvinsID:
**Description**  
Fixes #14788 
When there are multiple webapps with names that are substrings of each other, in certain cases `az webapp create` uses the wrong webapp. 

**Testing Guide**  
- Create different webapps with names that are substrings of each other, for ex. "test-webap", "test-weba", "test-webapp" in different resource groups
`az webapp create -g RG1 --plan Plan1 --name test-webap`
`az webapp create -g RG2 --plan Plan2 --name test-weba`
`az webapp create -g RG3 --plan Plan3 --name test-webapp`

- Create the second webapp again. The command should find the correct existing webapp and update the correct one.
`az webapp create -g RG2 --plan Plan2 --name test-weba`

Expected behaviour: Re running the create commands should find the correct existing webapps and update them.
Current behaviour: Re running the create commands uses the wrong webapp from different resource group, and throws error:

`The webapp 'test-webap' exists in resource group 'RG1' and does not match the value entered 'RG2'. Please re-run command with the correct parameters.`

---


yonzhan:
AppService

qwordy (approved):

qwordy:
Interesting bug

PR #14855, opened 2020-08-20
[Core] Error message improvement

houk-ms:
### Description

This PR works to improve the error messages of CLI, which includes the output messages shown to users, messages for Telemetry records, and new error categories. These changes are interdependent to some extent, so it's not very easy to split it into smaller PRs. Beg your patience and understanding for it may be a little large to review. The main changes are:

### 1. Introduce AzCLIError to replace CLIError
**Backgrounds**

Currently, CLIError is widely used in both CLI core and command groups. Working as a wrapper of the service-defined error types, CLIError accepts only an error message, which brings a side-effect that we can not distinguish the different error types and categorize them.

For example, CLIError can be caused by _resource not found_ which is an userfault, and it can also be caused by _service failure_ which is a service error. So when we want to differentiate _UserFault_ from _Service-Side-Error_ to do some analysis using telemetry, CLIError will block us. Explore [here ](https://msit.powerbi.com/groups/de995b1a-16c1-4cc7-a49b-bc9f90bc6acd/reports/b657544c-ab0a-4306-bc7b-5bedab1bb5a9/ReportSectiona83508fba4565f0f7f32?noSignUpCheck=1)to see more CLIError messages.

**This PR Bings**

This PR introduces the new AzCLIError, **while it will not change the original workflow of CLIError**. You can use the AzCLIError to replace your CLIError anywhere in your codes. And by using AzCLIError, you need to specify

> 1. [Required] An error type, defined below in section 2
> 2. [Required] An error message
> 3. [Optional] A recommendation to fix the error

Another benefit AzCLIError brings is that we can provide the same error message for both CLI users and Telemetry records. Previously, there are many cases that we print error messages for users, while the error message details are not recorded in Telemetry, and even the entire record is missing. [Here ](https://github.com/Azure/azure-cli/compare/dev...houk-ms:new-error-schema?expand=1#diff-143200fad5a24bc35d36c8ba3976c6b9R764-R766)is an example.

### 2. Define new error types for better error categorizing

As mentioned in section 1, these new error types can help us differentiate UserFault, Client-Side-Error and Service-Side-Error. The new defined error types are:

```
# userfaults
  CommandNotFoundError
  ArgumentParseError
  ValidationError
  ManualInterrupt

# service side error
  ServiceError

# client side error
  ClientError

# unexpected error
  UnexpectedError
```

### 3. Improve Telemetry

- One record for one command.

This PR sets the exception info into the UserTask type record, which means the UserTask record will have all the needed info for a command execution, no matter the comand fails or not. **The PR now doesn't change the original workflow of Fault type record**, which will be removed from Telemetry records after we confirm the UserTask record work properly.

- Fix the record missing issue.

When a command fails with a 404 error, there is no Telemetry record for the command. This PR fixes the issue.

### 4. New error messages

- New error messages format

As we discussed in previous meeting, the new error messages are applied in this PR. These message changes are mainly for userfaults, like CommandNotFountError and ArgumentParseError. The pattern is
```
ErrorType: ErrorMSG
Try this: ...recommend a command...
Still stuck? ...reference...
```

- Recommend a command when user's command fails

This PR provides the logic to recommend a command for users combining both examples from help files and aladdin recommendations.  It makes sure the recommended command has the same command name with user's input, otherwise, they will not be recommended. If there are multiple recommendations, we compare them and find the most similar one with the user's original inputs.

### Testing Guide

Type a command which can raise an CommandNotFoundError or ArgumentParseError, you will see the new error message formats. A few examples are shown here.

- az grup show

```
CommandNotFoundError: 'grup' is misspelled or not recognized by the system.
Did you mean 'group' ?
Try this: 'az group show --resource-group myresourcegroup'
Still stuck? Run 'az --help' to view all commands or go to 'https://docs.microsoft.com/en-us/cli/azure/reference-index?view=azure-cli-latest' to learn more
```

- az vm create -g rg
```
ArgumentParseError: the following arguments are required: --name/-n
Try this: 'az vm create -g MyResourceGroup -n MyVm --image MyImage'
Still stuck? Run 'az vm create --help' to view all commands or go to 'https://docs.microsoft.com/en-us/cli/azure/reference-index?view=azure-cli-latest' to learn more
```
- az group get -g rg
```
CommandNotFoundError: 'get' is misspelled or not recognized by the system.
Try this: 'az group list'
Still stuck? Run 'az group --help' to view all commands or go to 'https://docs.microsoft.com/en-us/cli/azure/reference-index?view=azure-cli-latest' to learn more
```
- az grup get -g rg
```
CommandNotFoundError: 'grup' is misspelled or not recognized by the system.
Did you mean 'group' ?
Still stuck? Run 'az --help' to view all commands or go to 'https://docs.microsoft.com/en-us/cli/azure/reference-index?view=azure-cli-latest' to learn more
```


yonzhan:
Core

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -60,7 +63,83 @@
 ]
 
 
-def handle_exception(ex):  # pylint: disable=too-many-return-statements
+class AzCLIErrorType(Enum):
```
yungezz, line 66: how about a new error.py

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -60,7 +63,83 @@
...
+    """ AzureCLI error types """
+
+    # userfaults
+    CommandNotFoundError = 'CommandNotFoundError'
+    ArgumentParseError = 'ArgumentParseError'
+    ValidationError = 'ValidationError'
+    ManualInterrupt = 'ManualInterrupt'
+    # service side error
+    ServiceError = 'ServiceError'
```
yungezz, line 75: AzureError sounds better. Azure is azure, service is little amibigous.

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -60,7 +63,83 @@
...
 
-def handle_exception(ex):  # pylint: disable=too-many-return-statements
+class AzCLIErrorType(Enum):
+    """ AzureCLI error types """
+
+    # userfaults
+    CommandNotFoundError = 'CommandNotFoundError'
+    ArgumentParseError = 'ArgumentParseError'
+    ValidationError = 'ValidationError'
+    ManualInterrupt = 'ManualInterrupt'
```
yungezz, line 73: KeyboardInterrupt is clear enough, what's your thoughts?

src/azure-cli-core/azure/cli/core/commands/arm.py
```diff
@@ -758,8 +758,12 @@ def show_exception_handler(ex):
     if getattr(getattr(ex, 'response', ex), 'status_code', None) == 404:
         import sys
         from azure.cli.core.azlogging import CommandLoggerContext
+        from azure.cli.core.util import AzCLIErrorType
+        from azure.cli.core.util import AzCLIError
         with CommandLoggerContext(logger):
-            logger.error(getattr(ex, 'message', ex))
+            az_error = AzCLIError(AzCLIErrorType.ValidationError, getattr(ex, 'message', ex))
+            az_error.print_error()
+            az_error.send_telemetry()
```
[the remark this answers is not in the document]
houk-ms: No, they don't need.

The changes here fix the issue that we have some records missed in Telemetry. The original logic just prints the error and exits.

src/azure-cli-core/azure/cli/core/util.py
houk-ms: Just for consistency, KeyboardInterrupt is good for me. Will use `KeyboardInterrupt ` to replace it.

src/azure-cli-core/azure/cli/core/_help.py
```diff
@@ -183,6 +183,22 @@ def show_help(self, cli_name, nouns, parser, is_group):
             if not nouns:
                 print(UX_SURVEY_PROMPT_COLOR if self.cli_ctx.enable_color else UX_SURVEY_PROMPT)
 
+    def get_examples(self, command, parser, is_group):
```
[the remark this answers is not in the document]
houk-ms: Sure

src/azure-cli-core/azure/cli/core/parser.py
```diff
@@ -50,6 +61,185 @@ def _get_completions(self, comp_words, cword_prefix, cword_prequote, last_wordbr
...
+
+        api_url = 'https://app.aladdin.microsoft.com/api/v1.0/suggestions'
+        headers = {'Content-Type': 'application/json'}
+        correlation_id = telemetry._session.correlation_id  # pylint: disable=protected-access
+        subscription_id = telemetry._get_azure_subscription_id()  # pylint: disable=protected-access
+
+        context = {
+            'versionNumber': version
+        }
+        if telemetry.is_telemetry_enabled():
```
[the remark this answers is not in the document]
houk-ms: Sure, here is in cli-core. For command owners, they will not need to deal with Telemetry.

src/azure-cli-core/azure/cli/core/parser.py
```diff
@@ -50,6 +61,185 @@ def _get_completions(self, comp_words, cword_prefix, cword_prequote, last_wordbr
...
+        try:
+            response = requests.get(
+                api_url,
+                params={
+                    'query': json.dumps(query),
+                    'clientType': 'AzureCli',
+                    'context': json.dumps(context)
+                },
+                headers=headers)
+        except RequestException as ex:
```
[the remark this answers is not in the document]
houk-ms: Good suggestion! Will consider more cases here to make it robust.

src/azure-cli-core/azure/cli/core/parser.py
```diff
@@ -143,18 +333,29 @@ def load_command_table(self, command_loader):
...
-        # Manual recommendations
-        self._set_manual_recommendations(args['message'])
-        # AI recommendations
+        # Get a recommended command from the CommandRecommender
+        command_arguments = self._get_failure_recovery_arguments()
+        recommender = CommandRecommender(*command_arguments, self.cli_ctx)
+        recommender.set_help_examples(self.get_examples(self.prog))
+        recommendation = recommender.recommend_a_command()
+
+        az_error = AzCLIError(AzCLIErrorType.ArgumentParseError, message, command=self.prog)
```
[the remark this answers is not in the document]
houk-ms: It's an overrided function of Argparse, only called when argparse meets argument errors.

src/azure-cli-core/azure/cli/core/parser.py
```diff
@@ -143,18 +333,29 @@ def load_command_table(self, command_loader):
                 _parser=command_parser)
 
     def validation_error(self, message):
-        telemetry.set_user_fault('validation error: {}'.format(message))
-        return super(AzCliCommandParser, self).error(message)
+        az_error = AzCLIError(AzCLIErrorType.ValidationError, message, command=self.prog)
+        az_error.print_error()
+        az_error.send_telemetry()
+        self.exit(2)
```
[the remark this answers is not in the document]
houk-ms: Should be the same. Will double check it.

src/azure-cli-core/azure/cli/core/parser.py
```diff
@@ -400,46 +609,50 @@ def _check_value(self, action, value):  # pylint: disable=too-many-statements, t
...
                                 exit_code = subprocess.call(cmd_list, shell=platform.system() == 'Windows')
-                                telemetry.set_user_fault("Extension {} dynamically installed and commands will be "
-                                                         "rerun automatically.".format(ext_name))
                                 self.exit(exit_code)
                             else:
-                                error_msg = 'Extension {} installed. Please rerun your command.'.format(ext_name)
+                                with CommandLoggerContext(logger):
+                                    reminder = 'Extension {} installed. Please rerun your command.'.format(ext_name)
+                                    logger.error(reminder)
+                                self.exit(2)
```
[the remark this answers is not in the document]
houk-ms: will double check it.

src/azure-cli-core/azure/cli/core/util.py
houk-ms: yeah, will consider it

src/azure-cli-core/azure/cli/core/parser.py
```diff
@@ -50,6 +61,185 @@ def _get_completions(self, comp_words, cword_prefix, cword_prequote, last_wordbr
...
+                    result['SuccessCommand_Parameters'].split(','),\
+                    result['SuccessCommand_ArgumentPlaceholders'].split('♠')
+                recommendation = 'az {} '.format(command)
+                for parameter, placeholder in zip(parameters, placeholders):
+                    recommendation += '{} {} '.format(parameter, placeholder)
+                recommendations.append(recommendation.strip())
+
+        self.aladdin_recommendations.extend(recommendations)
+
+    def recommend_a_command(self):
```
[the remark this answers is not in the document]
houk-ms: sure ~ willl adjust it accordingly.

src/azure-cli-core/azure/cli/core/parser.py
```diff
@@ -50,6 +61,185 @@ def _get_completions(self, comp_words, cword_prefix, cword_prequote, last_wordbr
...
+        except RequestException as ex:
+            logger.debug('Recommendation requests.get() exception: %s', ex)
+
+        recommendations = []
+        if response and response.status_code == HTTPStatus.OK:
+            for result in response.json():
+                # parse the reponse and format the recommendation
+                command, parameters, placeholders = result['SuccessCommand'],\
+                    result['SuccessCommand_Parameters'].split(','),\
+                    result['SuccessCommand_ArgumentPlaceholders'].split('♠')
```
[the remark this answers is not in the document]
houk-ms: yes, confirmed with Aladdin team. They use it as the delimiter.

src/azure-cli/azure/cli/command_modules/batch/tests/latest/test_batch_data_plane_commands.py
```diff
@@ -219,7 +219,7 @@ def test_batch_jobs_and_tasks(self, resource_group, batch_account_name):
 
         # test create job with missing parameters
         self.kwargs['start'] = datetime.datetime.now().isoformat()
-        with self.assertRaises(SystemExit):
+        with self.assertRaises(Exception):
```
arrownj, line 222: Why it's not a AzCLIError ?

houk-ms: For test, the main purpose here is to assert a failure. So, I think Exception is enough for these cases.

AzCLIError could work here and act as a fine-grained assertion. While my concern is that a fine-grained binding may bring extra works for further possible adjustments on error types.

Anyway,  I am open for this.

src/azure-cli/azure/cli/__main__.py
```diff
@@ -47,17 +47,13 @@ def cli_main(cli, args):
 
     exit_code = cli_main(az_cli, sys.argv[1:])
 
-    if exit_code and exit_code != 0:
-        if az_cli.result.error is not None and not telemetry.has_exceptions():
```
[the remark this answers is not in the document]
houk-ms: The design here is: **sending telemetry and printing error messages together**. So you could expect all Telemetry records are sent just after we print an error message to users.

The logics are moved to the exception handler [here](https://github.com/Azure/azure-cli/pull/14855/files#diff-5241dc629edabeefd74a7984150494bbR215-R216)

src/azure-cli-core/azure/cli/core/parser.py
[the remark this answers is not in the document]
houk-ms: Confirmed, it's the same. `super(AzCliCommandParser, self).error(message)` will finally call Argparse's error() method, which will call `sys.exit(2)`

src/azure-cli-core/azure/cli/core/parser.py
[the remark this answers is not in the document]
houk-ms: Confirmed, it's the same.

Original logic will finally call `self.exit(2)` in `_check_value` function.

src/azure-cli-core/azure/cli/core/parser.py
```diff
@@ -400,46 +609,50 @@ def _check_value(self, action, value):  # pylint: disable=too-many-statements, t
                                 import subprocess
                                 import platform
                                 exit_code = subprocess.call(cmd_list, shell=platform.system() == 'Windows')
-                                telemetry.set_user_fault("Extension {} dynamically installed and commands will be "
-                                                         "rerun automatically.".format(ext_name))
```
houk-ms, line 404: @fengzhou-msft We don't need set userfault here, right?

src/azure-cli-core/azure/cli/core/parser.py
```diff
@@ -50,6 +61,185 @@ def _get_completions(self, comp_words, cword_prefix, cword_prequote, last_wordbr
...
+        Call the aladdin service API, parse the response and set the recommendations.
+        """
+
+        import json
+        import requests
+        from requests import RequestException
+        from http import HTTPStatus
+        from azure.cli.core import __version__ as version
+
+        api_url = 'https://app.aladdin.microsoft.com/api/v1.0/suggestions'
```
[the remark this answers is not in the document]
mirdaki: Correct, the service will not exist in the air-gapped clouds. We are working on a way to put the model into air-gapped clouds so clients could download it and query it manually. Until that works, we should have a way to disable pinging the service if there are multiple failures. It could be a timed disabling (so users who loose internet for a time don't have it disabled forever).

src/azure-cli-core/azure/cli/core/command_recommender.py
```diff
@@ -0,0 +1,191 @@
...
+        """
+
+        import json
+        import requests
+        from requests import RequestException
+        from http import HTTPStatus
+        from azure.cli.core import __version__ as version
+
+        api_url = 'https://app.aladdin.microsoft.com/api/v1.0/suggestions'
+        headers = {'Content-Type': 'application/json'}
```
mirdaki, line 54: I posted this in another comment, but we should have a way to disable Aladdin calls for a period of time if it encounters repeated errors. The scenarios I'm thinking about are:

1. A user in an air-gapped cloud will never be able to hit the service (we are working on a way to download the model and query it locally, but that won't involve the service itself), so they should not waste cycles trying to hit an endpoint that does not exist for them.
2. A user experiencing a temporary internet outage will want this to work once they reconnect, so any disabling should have a time limit to prevent the feature from being disabled forever.

houk-ms: Yes, air gapped cloud is a good case here, we need consider to remove their requests for the api.

src/azure-cli-core/azure/cli/core/parser.py
```diff
@@ -400,46 +609,50 @@ def _check_value(self, action, value):  # pylint: disable=too-many-statements, t
...
-                    error_msg = "{prog}: '{value}' is not in the '{prog}' command group. See '{prog} --help'." \
-                        .format(prog=self.prog, value=value)
-                    if use_dynamic_install.lower() == 'no':
-                        extensions_link = 'https://docs.microsoft.com/en-us/cli/azure/azure-cli-extensions-overview'
-                        error_msg = ("{msg} "
-                                     "If the command is from an extension, "
-                                     "please make sure the corresponding extension is installed. "
-                                     "To learn more about extensions, please visit "
-                                     "{extensions_link}").format(msg=error_msg, extensions_link=extensions_link)
+                    error_msg = "'{value}' is misspelled or not recognized by the system.".format(value=value)
```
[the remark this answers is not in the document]
houk-ms: yes, the message is according to our pm's suggestion. Do we need to change one?

src/azure-cli-core/azure/cli/core/command_recommender.py
```diff
@@ -0,0 +1,196 @@
...
+        Call the aladdin service API, parse the response and set the recommendations.
+        """
+
+        import json
+        import requests
+        from requests import RequestException
+        from http import HTTPStatus
+        from azure.cli.core import __version__ as version
+
+        api_url = 'https://app.aladdin.microsoft.com/api/v1.0/suggestions'
```
mirdaki, line 53: I also wanted to add that we are currently working on support for downloading models locally. This should help to provide faster "cached" results both for regular users and let air-gapped users access to the Aladdin suggestions. Once we have a working example of that for `az find`, I will share that with you and we can discuss how best to integrate that with this.

houk-ms: sure thing!

src/azure-cli-core/azure/cli/core/parser.py
```diff
@@ -143,18 +155,29 @@ def load_command_table(self, command_loader):
...
-            logger.error('%(prog)s: error: %(message)s', args)
-        self.print_usage(sys.stderr)
-        # Manual recommendations
-        self._set_manual_recommendations(args['message'])
-        # AI recommendations
+        # Get a recommended command from the CommandRecommender
+        command_arguments = self._get_failure_recovery_arguments()
+        recommender = CommandRecommender(*command_arguments, self.cli_ctx)
+        recommender.set_help_examples(self.get_examples(self.prog))
+        recommendation = recommender.recommend_a_command()
```
jiasli, line 168: Won't this be a duplication of `ai-did-you-mean-this`?

jiasli, line 168: We'd better make a switch to allow the user to turn it off, especially when core is used by other CLI tools like Azure Sphere CLI or used in an air-gapped cloud.

Something like:

```
az config set core.error_recommendation=False
```

yungezz: client configuration may not help in this case. This is quite advanced behavior. Behavior of huge # of installation/users isn't controllable. better azure cli be more defensive in calling into aladdin. 
- set call timeout. no retry
- aladdin call failure will not block cli code flow
- not call aladdin in clouds in air-gapped clouds. hi @mirdaki, is aladdin service accessible from other public clouds like German, Mookcake?

yungezz (approved):

src/azure-cli-core/azure/cli/core/parser.py
mirdaki: Hi @yungezz, as far as I understand, if they can reach out to the public internet, they should be able to use the Aladdin service. I will add there is already a check for the air-gapped clouds, made in this [commit](https://github.com/Azure/azure-cli/pull/14855/commits/06888502590f263b34b6d91627d98dd0aedd6259).

I'm all for giving users more control of their client, but you know there behavior better than I do to know if this is a config item they would use.

christopher-o-toole (changes requested):

src/azure-cli-core/azure/cli/core/command_recommender.py
```diff
@@ -0,0 +1,196 @@
...
+        response = None
+        try:
+            response = requests.get(
+                api_url,
+                params={
+                    'query': json.dumps(query),
+                    'clientType': 'AzureCli',
+                    'context': json.dumps(context)
+                },
+                headers=headers,
```
mirdaki, line 82: Hey @houk-ms, in order to prevent bad actors from DDOS-ing the service we are adding a hashed User ID to the header as `X-UserId`. Could you please do that here? https://github.com/Azure/azure-cli/pull/15126 has an example of what were doing

src/azure-cli-core/azure/cli/core/command_recommender.py
```diff
@@ -0,0 +1,196 @@
...
+                timeout=1)
```
mirdaki, line 83: I've looked at some of the Azure CLI telemetry for the `az find` command from the past month. In the 90th percentile, it can take ~4.5s to get a response. Internally, the services 90th percentile for the same time period is <400ms, so a lot that is just network latency.

I think a timeout of 5s would cover most users without going exceptionally long. What do you think?

mirdaki (changes requested):
I have some updates based off of Aladdin load testing,

src/azure-cli-core/azure/cli/core/command_recommender.py
```diff
@@ -0,0 +1,196 @@
...
+        context = {
+            'versionNumber': version
+        }
+        if telemetry.is_telemetry_enabled():
+            if correlation_id:
+                context['correlationId'] = correlation_id
+            if subscription_id:
+                context['subscriptionId'] = subscription_id
+
+        parameters = [item for item in self.parameters if item not in ['--debug', '--verbose']]
```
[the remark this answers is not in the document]
houk-ms: will add it

src/azure-cli-core/azure/cli/core/azclierror.py
```diff
@@ -0,0 +1,88 @@
...
+                               AzCLIErrorType.ManualInterrupt]:
+            telemetry.set_user_fault(self.error_msg)
+
+        # For failures: service side error, client side error, unexpected error
+        else:
+            telemetry.set_failure(self.error_msg)
+
+        # For unexpected error
+        if self.raw_exception:
+            telemetry.set_exception(self.raw_exception, '')
```
zhoxing-ms, line 88: May I ask if it is possible that `telemetry.set_failure(self.error_msg)` and `telemetry.set_exception(self.raw_exception, '')` will be sent at the same command execution?
If so, will this cause the error message to be repeatedly recorded by telemetry?

zhoxing-ms:
May I ask whether the customer's adoption of the commands recommended by `ai-did-you-mean-this` will be collected?
I personally think that if there is a feedback mechanism to analyze user adoption, it can help us better optimize the recommendation effect.

zhoxing-ms:
By the way, we've recently been working on `az next` project, which also includes the scenario of error handling 
PR Link: [click](https://github.com/hackathon-cli-recommendation/cli-recommendation/blob/master/Docs/CLI_recommendation_design_doc.md#1-solution)
I think we can integrate these similar features in the future, I will discuss with you after I have perfected my project~ 😊

src/azure-cli-core/azure/cli/core/command_recommender.py
```diff
@@ -0,0 +1,196 @@
...
+        from azure.cli.core import __version__ as version
+
+        api_url = 'https://app.aladdin.microsoft.com/api/v1.0/suggestions'
+        headers = {'Content-Type': 'application/json'}
+        correlation_id = telemetry._session.correlation_id  # pylint: disable=protected-access
+        subscription_id = telemetry._get_azure_subscription_id()  # pylint: disable=protected-access
+
+        context = {
+            'versionNumber': version
+        }
```
christopher-o-toole, line 60: Add `errorType` field to here.

i.e.
```Python
context = {
    'versionNumber': version,
    'errorType': errorType
}
```
where error type is the user fault error type.

src/azure-cli-core/azure/cli/core/command_recommender.py
```diff
@@ -0,0 +1,196 @@
...
+                api_url,
+                params={
+                    'query': json.dumps(query),
+                    'clientType': 'AzureCli',
+                    'context': json.dumps(context)
+                },
+                headers=headers,
+                timeout=1)
+        except RequestException as ex:
+            logger.debug('Recommendation requests.get() exception: %s', ex)
```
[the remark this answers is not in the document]
houk-ms: good suggestion, will add it

src/azure-cli-core/azure/cli/core/command_recommender.py
```diff
@@ -0,0 +1,196 @@
...
+        """
+        from azure.cli.core.cloud import CLOUDS_FORBIDDING_ALADDIN_REQUEST
+
+        if self.cli_ctx and self.cli_ctx.cloud \
+           and self.cli_ctx.cloud.name not in CLOUDS_FORBIDDING_ALADDIN_REQUEST:
+            self._set_aladdin_recommendations()
+
+        # all the recommended commands from help examples and aladdin
+        all_commands = self.help_examples + self.aladdin_recommendations
+        all_commands.sort(key=len)
```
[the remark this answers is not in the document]
houk-ms: Got it. Will consider how to choose a better candidate in next CLI release along with the confidence. Currently, the plan is just using the logic here to get some feedbacks.

src/azure-cli-core/azure/cli/core/command_recommender.py
```diff
@@ -0,0 +1,196 @@
...
+
+        filtered_commands = []
+        filtered_choices = []
+        target = ''.join(self.parameters)
+
+        for command in all_commands:
+            # filter out the commands which begins with a different command group
```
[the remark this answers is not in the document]
houk-ms: Make sense. As we disscussed, will wait for the feedbacks from customers and may consider improve it in next CLI release.

src/azure-cli-core/azure/cli/core/command_recommender.py
houk-ms: Added.

src/azure-cli-core/azure/cli/core/command_recommender.py
houk-ms: As we disscussed, 5s seems too long. 

Anyway, I have recorded the service exceptions into Telemetry. Let's wait for the Telemetry records to see if there are many timeouts there, and then may consider adjust it in next release.

src/azure-cli-core/azure/cli/core/azclierror.py
houk-ms: Only for **UnexpectedError**: set_failure and set_exception will both be called. That's for more convenient error categorizing from Telemetry as well as error tracking.

src/azure-cli-core/azure/cli/core/command_recommender.py
```diff
@@ -0,0 +1,196 @@
...
+                headers=headers,
+                timeout=1)
+        except RequestException as ex:
+            logger.debug('Recommendation requests.get() exception: %s', ex)
+
+        recommendations = []
+        if response and response.status_code == HTTPStatus.OK:
+            for result in response.json():
+                # parse the reponse and format the recommendation
+                command, parameters, placeholders = result['SuccessCommand'],\
```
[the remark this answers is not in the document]
houk-ms: > By the way, we've recently been working on `az next` project, which also includes the scenario of error handling
> PR Link: [click](https://github.com/hackathon-cli-recommendation/cli-recommendation/blob/master/Docs/CLI_recommendation_design_doc.md#1-solution)
> I think we can integrate these similar features in the future, so I will discuss with you after I have perfected my project~ 😊

sure!

houk-ms:
> May I ask whether the customer's adoption of the commands recommended by `ai-did-you-mean-this` will be collected?
> I personally think that if there is a feedback mechanism to analyze user adoption, it can help us better optimize the recommendation effect.

Not yet now. Maybe we can discuss with Aladdin team to see if that can help them improve the service and consider then add it in next release if necessary.

src/azure-cli-core/azure/cli/core/command_recommender.py
houk-ms: Added

src/azure-cli-core/azure/cli/core/command_recommender.py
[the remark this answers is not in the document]
houk-ms: In this release, I just added the fallback logic: if we can not find a recommendation with the same command name. Will fallback to user the first recommended command from aladdin service.

src/azure-cli-core/azure/cli/core/parser.py
houk-ms: Has been discussing with Aladdin team, will make it retired if necessary.

houk-ms: - Air-gapped cloud now is not allowed to call the api 
- Service is using hashed UserID to stop repeated api calls

So, let's just keep the current status. We can disscuss whether to add a config in the future if the are new cases.

src/azure-cli-core/azure/cli/core/azclierror.py
zhoxing-ms: Got it~ Thanks

src/azure-cli-core/azure/cli/core/parser.py
fengzhou-msft: We need this message in telemetry to get the numbers for dynamic install with automatic rerun.
Here actually the same command will run twice, I set the first command to be UserFault with the message, the second command which is rerun will have normal result.

I'm OK to not set user fault for the first command, but you need to put the message in resultSummary to distinguish it.

src/azure-cli-core/azure/cli/core/parser.py
```diff
@@ -400,46 +432,50 @@ def _check_value(self, action, value):  # pylint: disable=too-many-statements, t
...
+            recommender = CommandRecommender(command_name_inferred[3:].strip(), params, extension, error_msg, cli_ctx)
+            recommender.set_help_examples(self.get_examples(command_name_inferred))
+            recommended_command = recommender.recommend_a_command()
+            if recommended_command:
+                az_error.set_recommendation("Try this: '{}'".format(recommended_command))
 
+            # remind user to check extensions if we can not find a command to recommend
+            if az_error.error_type == AzCLIErrorType.CommandNotFoundError \
+                    and not az_error.recommendations and self.prog == 'az':
+                az_error.set_recommendation(EXTENSION_REFERENCE)
```
fengzhou-msft, line 471: EXTENSION_REFERENCE should not be shown if dynamic extension install is turned on and found no match for the command in the extension command tree.

src/azure-cli-core/azure/cli/core/parser.py
```diff
@@ -400,46 +432,50 @@ def _check_value(self, action, value):  # pylint: disable=too-many-statements, t
...
-                    }
-                    self._suggestion_msg.append("\nThe most similar choice{s} to '{value}' {verb}:"
-                                                .format(**print_args))
-                    self._suggestion_msg.append('\n'.join(['\t' + candidate for candidate in candidates]))
+            _, params, extension = self._get_failure_recovery_arguments()
+            if candidates:
+                az_error.set_recommendation("Did you mean '{}' ?".format(candidates[0]))
+
+            # recommand a command for user
+            recommender = CommandRecommender(command_name_inferred[3:].strip(), params, extension, error_msg, cli_ctx)
```
christopher-o-toole, line 462: Aladdin expects the command to be passed unchanged.  Autocorrecting the command requires changing the error type as well.

A few examples to clarify what I mean:
* If `az grup` is corrected to `az group` the new error type should be `MissingRequiredSubcommand` rather than `InvalidSubcommand`
* If `az vim create` is corrected to `az vm create` the new error type should be `MissingRequiredParameters` rather than `InvalidSubcommand`.

christopher-o-toole (changes requested):

src/azure-cli-core/azure/cli/core/parser.py
fengzhou-msft: Add `use_dynamic_install == 'no'` in the condition.

src/azure-cli-core/azure/cli/core/parser.py
```diff
@@ -17,13 +17,25 @@
...
 
 logger = get_logger(__name__)
 
+EXTENSION_REFERENCE = ("If the command is from an extension, "
+                       "please make sure the corresponding extension is installed. "
+                       "To learn more about extensions, please visit "
+                       "'https://docs.microsoft.com/en-us/cli/azure/azure-cli-extensions-overview'")
+
+OVERVIEW_REFERENCE = ("Still stuck? Run '{command} --help' to view all commands or go to "
+                      "'https://docs.microsoft.com/en-us/cli/azure/reference-index?view=azure-cli-latest' "
```
fengzhou-msft, line 36: Use a short link.

src/azure-cli-core/azure/cli/core/parser.py
```diff
@@ -400,46 +432,50 @@ def _check_value(self, action, value):  # pylint: disable=too-many-statements, t
...
                                 exit_code = subprocess.call(cmd_list, shell=platform.system() == 'Windows')
-                                telemetry.set_user_fault("Extension {} dynamically installed and commands will be "
-                                                         "rerun automatically.".format(ext_name))
                                 self.exit(exit_code)
                             else:
-                                error_msg = 'Extension {} installed. Please rerun your command.'.format(ext_name)
+                                with CommandLoggerContext(logger):
+                                    reminder = 'Extension {} installed. Please rerun your command.'.format(ext_name)
+                                    logger.error(reminder)
+                                self.exit(2)
```
fengzhou-msft, line 440: Tested the code. The result of telemetry is None when the extension is dynamically installed, whether run_after_extension_installed is true or false.

houk-ms: That's the same problem with the above one.
Will set userfault for both cases.

src/azure-cli-core/azure/cli/core/parser.py
houk-ms: will set userfault

src/azure-cli-core/azure/cli/core/parser.py
```diff
@@ -400,46 +432,50 @@ def _check_value(self, action, value):  # pylint: disable=too-many-statements, t
...
-                if candidates:
-                    print_args = {
-                        's': 's' if len(candidates) > 1 else '',
-                        'verb': 'are' if len(candidates) > 1 else 'is',
-                        'value': value
-                    }
-                    self._suggestion_msg.append("\nThe most similar choice{s} to '{value}' {verb}:"
-                                                .format(**print_args))
-                    self._suggestion_msg.append('\n'.join(['\t' + candidate for candidate in candidates]))
+            _, params, extension = self._get_failure_recovery_arguments()
```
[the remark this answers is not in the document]
houk-ms: Thanks for pointing it out. Will add the action parameter.

src/azure-cli-core/azure/cli/core/parser.py
houk-ms: Modified.

src/azure-cli-core/azure/cli/core/parser.py
houk-ms: Modified.

src/azure-cli-core/azure/cli/core/parser.py
houk-ms: The link is the same with the original code. I suggest we apply the short link in another pr, along with other possible links.

src/azure-cli-core/azure/cli/core/parser.py
houk-ms: This will be noted as a work item to improve. Will keep sync with you.

src/azure-cli-core/azure/cli/core/command_recommender.py
```diff
@@ -0,0 +1,283 @@
...
+                '--only-show-errors': None,
+                '--help': None,
+                '--output': None,
+                '--query': None,
+                '--debug': None,
+                '--verbose': None
+            }
+
+            if parameter_table:
+                for argument in parameter_table.values():
```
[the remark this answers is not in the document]
houk-ms: Work item to improve.

src/azure-cli-core/azure/cli/core/parser.py
houk-ms: Tested.

src/azure-cli-core/azure/cli/core/parser.py
zhoxing-ms: > `az vim create` is corrected to `az vm create`

@christopher-o-toole May I ask why the new error type of this case is `MissingRequiredParameters`?

christopher-o-toole: @zhoxing-ms 

Running `az vm create` yields the following error:

```
> az vm create
az vm create: error: the following arguments are required: --name/-n, --resource-group/-g
```

Therefore the required parameters `--name` and `--resource-group` are missing in this case.

If we take into account which parameters and arguments the user passed, the logic is more complicated. In terms of getting relevant suggestions however, it suffices to set either `ExpectedArgument` or `MissingRequiredParameters` as the error type assuming that the inferred command is what the user intended.

haroldrandom (approved):
Network LGTM

Juliehzl (approved):
LGTM in general. Recommend to use replacer for unexpected request

houk-ms:
**All the changes on test files are removed. All the changes on Azure Pipeline are removed.**
 
Now the method is: disabling aladdin requests in all the testing environments, which is decied by CLI core whether current instance is DummyCLI.

houk-ms:
#15314

PR #14885, opened 2020-08-22
[IoT Hub] Deprecated 'show-connection-string' command

anusapan:
**Description**  
Deprecating command **az iot hub show-connection-string**
image

Now, we have another command group for connection-string from IoT-Extension that can use instead
image

**Testing Guide**  

**History Notes**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/iot/commands.py
+++ b/src/azure-cli/azure/cli/command_modules/iot/commands.py
@@ -8,7 +8,7 @@ from ._client_factory import iot_hub_service_factory
 from ._client_factory import iot_service_provisioning_factory
 from ._client_factory import iot_central_service_factory
 
-JOB_DEPRECATION_INFO = 'IoT Extension (azure-cli-iot-ext) Job commands'
+CS_DEPRECATION_INFO = 'IoT Extension (azure-iot) connection-string command (az iot hub connection-string show)'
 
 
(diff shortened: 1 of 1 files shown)


src/azure-cli/azure/cli/command_modules/iot/commands.py
```diff
@@ -8,7 +8,7 @@
 from ._client_factory import iot_service_provisioning_factory
 from ._client_factory import iot_central_service_factory
 
-JOB_DEPRECATION_INFO = 'IoT Extension (azure-cli-iot-ext) Job commands'
+CS_DEPRECATION_INFO = 'IoT Extension (azure-cli-iot-ext) connection-string commands'
```
digimaun, line 11: The IoT extension is now called "azure-iot". "azure-cli-iot-ext" is deprecated and should not be used anymore.

digimaun, line 11: I think we should also show the equivalent command in the extension so the user doesn't have to look for it `az iot hub connection-string show`

anusapan: Done

anusapan: Done

zhoxing-ms (approved):

PR #14902, opened 2020-08-25
{VM} Fix #14897: display all requirements for vm admin password when it's invalid

### LINKED ISSUE #14897: Display all PWD requirements in error messages?

Yes, frustrated when the error guidance/output only check one condition at a time rather than providing complete holistic guidance
See following example. Is there a way to inform the user both PWD requirements at once?
image

Combining both error messages so people don't hit the same error twice because of the lack of clear requirements the in first error output

No alternatives

**Additional context**
In midst of designing lab study with VM resource
Believe we've had similar frustration in PWD settings in the storage/db space

houk-ms:
**Description**  

This PR fix issue #14897. 

**Before**
When creating a vm with an invalid admin password, the message shows only the current rule the password violates. When user changes the password and try it again, another error may happen showing that the user violates anthoer rule.

**After**
This PR improves the behavior by displaying all the requirements once we found the password is invalid. 
 

**Testing Guide**  

**Before**
```
Admin Password:
Confirm Admin Password:
The password length must be between 12 and 123

Admin Password:
Confirm Admin Password:
Password must have the 3 of the following: 1 lower case character, 1 upper case character, 1 number and 1 special character
```
**After**
```
Your password is invalid for it violates Rule 1
Rule 1: The password length must be betwween 12 and 123
Rule 2: Password must have the 3 of the following: 1 lower case character, 1 upper case character, 1 number and 1 special character
```

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/vm/_validators.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_validators.py
@@ -1034,20 +1034,23 @@ def _validate_admin_password(password, os_type):
     is_linux = (os_type.lower() == 'linux')
     max_length = 72 if is_linux else 123
     min_length = 12
-    error_msg = ("Rule 1: The password length must be between {} and {}\n"
-                 "Rule 2: Password must have the 3 of the following: 1 lower case character, "
-                 "1 upper case character, 1 number and 1 special character").format(min_length, max_length)
-    if len(password) not in range(min_length, max_length + 1):
-        raise CLIError("Your password is invalid for it violates Rule 1\n{}".format(error_msg))
+
     contains_lower = re.findall('[a-z]+', password)
     contains_upper = re.findall('[A-Z]+', password)
     contains_digit = re.findall('[0-9]+', password)
     contains_special_char = re.findall(r'[ `~!@#$%^&*()=+_\[\]{}\|;:.\/\'\",<>?]+', password)
     count = len([x for x in [contains_lower, contains_upper,
                              contains_digit, contains_special_char] if x])
+
     # pylint: disable=line-too-long
+    length_error = "The password length must be between {} and {}.".format(min_length, max_length)
+    complexity_error = "Password must have the 3 of the following: 1 lower case character, 1 upper case character, 1 number and 1 special character."
+    if len(password) not in range(min_length, max_length + 1) and count < 3:
+        raise CLIError("{} {}".format(length_error, complexity_error))
+    if len(password) not in range(min_length, max_length + 1):
+        raise CLIError(length_error)
     if count < 3:
-        raise CLIError("Your password is invalid for it violates Rule 2\n{}".format(error_msg))
+        raise CLIError(complexity_error)
 
 
 def validate_ssh_key(namespace):
--- a/src/azure-cli/azure/cli/command_modules/vm/_validators.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_validators.py
@@ -1043,14 +1043,10 @@ def _validate_admin_password(password, os_type):
                              contains_digit, contains_special_char] if x])
 
     # pylint: disable=line-too-long
-    length_error = "The password length must be between {} and {}.".format(min_length, max_length)
-    complexity_error = "Password must have the 3 of the following: 1 lower case character, 1 upper case character, 1 number and 1 special character."
-    if len(password) not in range(min_length, max_length + 1) and count < 3:
-        raise CLIError("{} {}".format(length_error, complexity_error))
-    if len(password) not in range(min_length, max_length + 1):
-        raise CLIError(length_error)
-    if count < 3:
-        raise CLIError(complexity_error)
+    error_msg = ("The password length must be between {} and {}. And the password must have the 3 of the following: "
+                 "1 lower case character, 1 upper case character, 1 number and 1 special character.").format(min_length, max_length)
+    if len(password) not in range(min_length, max_length + 1) or count < 3:
+        raise CLIError(error_msg)
 
 
 def validate_ssh_key(namespace):
--- a/src/azure-cli/azure/cli/command_modules/vm/_validators.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_validators.py
@@ -1034,7 +1034,7 @@ def _validate_admin_password(password, os_type):
     is_linux = (os_type.lower() == 'linux')
     max_length = 72 if is_linux else 123
     min_length = 12
-    error_msg = ("Rule 1: The password length must be betwween {} and {}\n"
+    error_msg = ("Rule 1: The password length must be between {} and {}\n"
                  "Rule 2: Password must have the 3 of the following: 1 lower case character, "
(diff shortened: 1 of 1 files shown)


src/azure-cli/azure/cli/command_modules/vm/_validators.py
```diff
@@ -1034,9 +1034,11 @@ def _validate_admin_password(password, os_type):
     is_linux = (os_type.lower() == 'linux')
     max_length = 72 if is_linux else 123
     min_length = 12
+    error_msg = ("Rule 1: The password length must be betwween {} and {}\n"
```
houk-ms, line 1037:
```suggestion
    error_msg = ("Rule 1: The password length must be between {} and {}\n"
```

src/azure-cli/azure/cli/command_modules/vm/_validators.py
```diff
@@ -1034,9 +1034,11 @@ def _validate_admin_password(password, os_type):
     is_linux = (os_type.lower() == 'linux')
     max_length = 72 if is_linux else 123
     min_length = 12
+    error_msg = ("Rule 1: The password length must be between {} and {}\n"
+                 "Rule 2: Password must have the 3 of the following: 1 lower case character, "
+                 "1 upper case character, 1 number and 1 special character").format(min_length, max_length)
     if len(password) not in range(min_length, max_length + 1):
-        raise CLIError('The password length must be between {} and {}'.format(min_length,
-                                                                              max_length))
+        raise CLIError("Your password is invalid for it violates Rule 1\n{}".format(error_msg))
```
yungezz, line 1041: LGTM. why not make it simpler. error_msg = Password length must be between {} and {}, and must have the 3 of following....". In CLIError, just message "Your password doesn't meet requirement, pls see password constraint: {}".format(error_msg)

yungezz (approved):

yungezz:
thanks for quick fix!

src/azure-cli/azure/cli/command_modules/vm/_validators.py
houk-ms: I think user knows where is error better when we split the error cases.

qwordy: Maybe we can just show the rule that the user violates instead of printing all rules.

src/azure-cli/azure/cli/command_modules/vm/_validators.py
```diff
@@ -1045,7 +1047,7 @@ def _validate_admin_password(password, os_type):
                              contains_digit, contains_special_char] if x])
     # pylint: disable=line-too-long
     if count < 3:
-        raise CLIError('Password must have the 3 of the following: 1 lower case character, 1 upper case character, 1 number and 1 special character')
+        raise CLIError("Your password is invalid for it violates Rule 2\n{}".format(error_msg))
```
qwordy, line 1050: Maybe we can just show the rule that the user violates instead of printing all rules.

houk-ms: That's exactly what the issue was open for. They want to know all the rules once they failed :(

qwordy: I suggest create a survey to collect users' opinions widely. That's just one user's appeal.

houk-ms: Our PM opened this issue, I think she knows the users well ~

qwordy:
I test it in Portal. Portal will only show the rule that you violates.
image
image

qwordy (approved):

chenlomis:
Thank you all for the great discussion! Was informed of this new issue and thought I could leave a note to help clarify the ask:

The system logic @qwordy laid out makes sense and I like how it also aligns with portal's. If we're merely checking whether the user PWD has passed certain criterion from the CLI side then there's no fault in checking those conditions one by one. 

However from the user perspective we should make clear of all the PWD requirements once they encounter the error to provide guidance and to prevent additional UX papercuts

Here's the scenario --- suppose Bob is looking to set the admin password for his new VM with a short password only to encounter the "the password length must be between 12 and 123" error. When he sees that, he'd only attempt to lengthen the password since he's unaware of any other PWD requirements. Little did he know that he will actually fail once more because the new PWD is highly likely not complicated enough. 

So in this case, there's a very high chance that Bob (and hundreds others like Bob) has to spend 2x time to come up with an appropriate and acceptable password

But if we make all the PWD requirements clear from the very beginning, then Bob can take into account all of those when coming up with an appropriate password once he hits the initial error


So I'm perfectly content if we:

i)  only show the "password too short" error if the password is too short but is complicated enough; 
ii) only show the "password not complicated enough" error if the password is too simple but is long enough;
iii) show both "password too short" and "password not complicated enough" if the pwd violates both

Hope the above makes sense...

arrownj (approved):

arrownj:
👍 💯 👍

arrownj:
I agree with showing all the rules to user from the very begin. Actually the user experiences are different between portal and CLI, so I think we do not need to follow portal's behavior.

However the "Rule 1, Rule 2" message seems a little imperative, suggest to combine them together and make it more friendly.

houk-ms:
Thanks everyone! With these discussions, I think the behavior here is what we agree on
```
Admin Password: abcdef
Confirm Admin Password: abcdef
The password length must be between 12 and 123. Password must have the 3 of the following: 1 lower case character, 1 upper case character, 1 number and 1 special character

Admin Password: aA1
Confirm Admin Password: aA1
The password length must be between 12 and 123

Admin Password: 123456789012
Confirm Admin Password: 123456789012
Password must have the 3 of the following: 1 lower case character, 1 upper case character, 1 number and 1 special character
```

yungezz:
> Thanks everyone! With these discussions, I think the behavior here is what we agree on
> 
> ```
> Admin Password: abcdef
> Confirm Admin Password: abcdef
> The password length must be between 12 and 123. Password must have the 3 of the following: 1 lower case character, 1 upper case character, 1 number and 1 special character
> 
> Admin Password: aA1
> Confirm Admin Password: aA1
> The password length must be between 12 and 123
> 
> Admin Password: 123456789012
> Confirm Admin Password: 123456789012
> Password must have the 3 of the following: 1 lower case character, 1 upper case character, 1 number and 1 special character
> ```

what's the difference between this one and original issue? I agree with @chenlomis as a client tool it's better to show the constraint to user completely.  On portal, it's very easy to show it with UI, user types something, the error shows immediately. while in Azure CLI, user need to run command repeatedly to get all rules.

qwordy:
Current help message for `--admin-password`. Maybe we can put the rules in help message.
```
--admin-password

Password for the VM if authentication type is 'Password'.
```

houk-ms:
So, let's conclude the discussion. Whenever user inputs an invalid password, we will show the error message below.
```
The password length must be between 12 and 123. And the password must have the 3 of the following: 1 lower case character, 1 upper case character, 1 number and 1 special character.
```

PR #14903, opened 2020-08-25
{CI} Example of bypassing show-command linter rule

MyronFanQiu:
`az configure show`, `az cache show` or `az vm monitor show` is not a real show command.
`az functionapp function show` has its own error handler logic


arrownj (approved):
👍

arrownj (commented):
Don't forget to update HISTORY

Juliehzl (approved):

Issue #13846, opened 2020-06-05
az aks get-credentials throws KeyError: 'clusters'


zoomout:
## Describe the bug

**Command Name**
`az aks get-credentials`

**Errors:**
```
'clusters'
Traceback (most recent call last):
python3.8/site-packages/knack/cli.py, ln 215, in invoke
    cmd_result = self.invocation.execute(args)
cli/core/commands/__init__.py, ln 654, in execute
    raise ex
cli/core/commands/__init__.py, ln 718, in _run_jobs_serially
    results.append(self._run_job(expanded_arg, cmd_copy))
cli/core/commands/__init__.py, ln 711, in _run_job
    six.reraise(*sys.exc_info())
lib/python3.8/site-packages/six.py, ln 703, in reraise
    raise value
...
cli/command_modules/acs/custom.py, ln 3204, in _print_or_merge_credentials
    merge_kubernetes_configurations(path, temp_path, overwrite_existing, context_name)
cli/command_modules/acs/custom.py, ln 1184, in merge_kubernetes_configurations
    _handle_merge(existing, addition, 'clusters', replace)
cli/command_modules/acs/custom.py, ln 1120, in _handle_merge
    if existing[key] is None:
KeyError: 'clusters'
```

## To Reproduce:
Steps to reproduce the behavior. Note that argument values have been redacted, as they may contain sensitive information.

- _Put any pre-requisite steps here..._
- `az aks get-credentials --resource-group {} --name {}`

## Environment Summary
```
macOS-10.15.5-x86_64-i386-64bit
Python 3.8.3
Installer: HOMEBREW

azure-cli 2.7.0

```

yonzhan:
aks

atulaggarwal:
Looking at the stack trace, it looks like merging of cluster into to kubeconfi files is failing which could mean most likely kubectl config files are corrupted (normal path to file is ~/.kube/config). 

Can verify the configuration file using steps mentioned in [link](https://kubernetes.io/docs/tasks/tools/install-kubectl/#verifying-kubectl-configuration)? If you find the file to be corrupted, can you delete it and try again?

matthchr:
I'm having this issue too:
```
$ kubectl config view
apiVersion: v1
clusters: null
contexts: null
current-context: ""
kind: Config
preferences: {}
users: null
```

[...]

PR #14914, opened 2020-08-25
[AKS] CLI should tolerate empty kubeconfig

matthchr:
**Description**  
Fix bug #13846

**Testing Guide**  
Run `az aks get-credentials --resource-group <rg> --name <name>` on an AKS cluster while your local kubeconfig is empty:
```
$ kubectl config view
apiVersion: v1
clusters: null
contexts: null
current-context: ""
kind: Config
preferences: {}
users: null
```

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acs/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/custom.py
@@ -1214,7 +1214,7 @@ def _k8s_get_credentials_internal(name, acs_info, path, ssh_key_file, overwrite_
 def _handle_merge(existing, addition, key, replace):
     if not addition.get(key, False):
         return
-    if existing[key] is None:
+    if not existing.get(key):
         existing[key] = addition[key]
         return
 


matthchr:
Not sure what the right component prefix is supposed to be here...

yonzhan:
AKS

src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -1214,7 +1214,7 @@ def _k8s_get_credentials_internal(name, acs_info, path, ssh_key_file, overwrite_
 def _handle_merge(existing, addition, key, replace):
     if not addition.get(key, False):
         return
-    if existing[key] is None:
+    if not existing.get(key):
```
arrownj, line 1217: existing[key] == None / 0 / [] / "" / {} will all trigger this condition, is this the expected behavior ?

matthchr: I'm not an expert on what "valid" vs "invalid" kubeconfigs look like, so I don't know for sure.

I think at the very least a missing key should be treated the same as an empty key (before, a missing key triggered an exception).

I was matching the line above on 1214 which was was doing the same thing (not checking specifically for `== False` or something.

Looking further on in the code, if we don't overwrite `None` / `0` / `""` the code below will blow up I think as we'll try to iterate over it assuming that it's a dict.

arrownj:
Add @rjtsdl to review.

matthchr:
@arrownj @zqingqing1 - is there more I need to do here? If not can we merge this?

arrownj (approved):

PR #14921, opened 2020-08-26
[Config] Use core as the default section

jiasli:
## Description
Use `core` as the default section for `set`/`get`/`unset`.

## Testing Guide

```sh
# Omit `core.`
az config set no_color=true
az config get no_color
az config unset no_color

# Explicitly specify `core.`
az config set core.no_color=true
az config get core.no_color
az config unset core.no_color
```

Note there is a breaking change for `az config get`: If `<name>` is specified, it is treated as `core.<name>` instead of the section name. To get all options under a section, use `<section>.`:

```sh
az config get core.
```

**History Notes**  

[Config] az config set/get: If `section` is unspecified, use `core` as the default section.
[Config] BREAKING CHANGE: az config get: If `<name>` is specified, it is treated as `core.<name>` instead of the section name. To get all options under a section, use `<section>.`.

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/config/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/config/_params.py
@@ -11,20 +11,27 @@ def load_arguments(self, _):
         c.ignore('_subscription')  # ignore the global subscription param
 
     with self.argument_context('config set') as c:
-        c.positional('key_value', nargs='+', help="Space-separated configurations in the form of <section>.<key>=<value>.")
+        c.positional('key_value_pairs', nargs='+',
+                     help="Set options in the form of <section>.<name>=<value>. "
+                          "If `<section>` is omitted, `core` will be used. "
+                          "To set multiple options, separate each key value pair with spaces.")
         c.argument('local', action='store_true', help='Set as a local configuration in the working directory.')
 
     with self.argument_context('config get') as c:
-        c.positional('key', nargs='?', help='The configuration to get. '
-                                            'If not provided, all sections and configurations will be listed. '
-                                            'If `section` is provided, all configurations under the specified section will be listed. '
-                                            'If `<section>.<key>` is provided, only the corresponding configuration is shown.')
+        c.positional('key', nargs='?',
+                     help='The configuration to get. '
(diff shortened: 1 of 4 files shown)


yonzhan:
Config

MyronFanQiu (approved):
approve for linter_exclusion

src/azure-cli/azure/cli/command_modules/config/_params.py
```diff
@@ -11,20 +11,27 @@ def load_arguments(self, _):
         c.ignore('_subscription')  # ignore the global subscription param
 
     with self.argument_context('config set') as c:
-        c.positional('key_value', nargs='+', help="Space-separated configurations in the form of <section>.<key>=<value>.")
+        c.positional('key_value_pairs', nargs='+',
+                     help="Set options in the form of <section>.<name>=<value>. "
+                          "If `<section>` is omitted, `core` will be used. "
```
arrownj, line 16: It's a little strange to me that the omitted value is not `defaults`.

jiasli: If we default section to `defaults`, setting options in `core` section will still be

```
az config set core.output=table
```

This is not very user-friendly. 

On the other hand, setting the default group becomes:

```
az config set group=rg1
```

which may also confuse users as some commands ignore the default resource group, like `az role assignment create`.

https://github.com/Azure/azure-cli/blob/37d15261c516bc095236ddb1155dbcc470daf79b/src/azure-cli/azure/cli/command_modules/role/_validators.py#L98-L101

Also, defaulting section to `core` is also what gcloud adopts in [`gcloud config set`](https://cloud.google.com/sdk/gcloud/reference/config/set):

> SECTION/PROPERTY
> Property to be set. Note that SECTION/ is optional while referring to properties in the core section, i.e., using either core/project or project is a valid way of setting a project. Using section names is required for setting other properties like compute/region. Consult the Available Properties section below for a comprehensive list of properties.

arrownj (approved):

jiasli:
@chenlomis, would you kindly share your opinions on this change?

PR #15020, opened 2020-09-02
remove duplicate rg from mysql

arde0708:
This PR remove the extra resource group that shows up in the output for MYSQL


arde0708:
Incorrectly opened

PR #14588, opened 2020-07-30
{Core} Print traceback to debug log when handle_exception is called

jiasli:
**Description**  

Print traceback to debug log when `azure.cli.core.util.handle_exception` is called. This can help identify where the exception is generated and thrown.

**Testing Guide**  

Notice the location `westu` is wrong (should be `westus`):

```
> az group create -n rg1 -l westu --debug
...
azure.cli.core.util.handle_exception is called with an exception:
Traceback (most recent call last):
  File "D:\cli\env38\lib\site-packages\knack\cli.py", line 215, in invoke
    cmd_result = self.invocation.execute(args)
  File "d:\cli\azure-cli\src\azure-cli-core\azure\cli\core\commands\__init__.py", line 654, in execute
    raise ex
  File "d:\cli\azure-cli\src\azure-cli-core\azure\cli\core\commands\__init__.py", line 718, in _run_jobs_serially
    results.append(self._run_job(expanded_arg, cmd_copy))
  File "d:\cli\azure-cli\src\azure-cli-core\azure\cli\core\commands\__init__.py", line 711, in _run_job
    six.reraise(*sys.exc_info())
  File "D:\cli\env38\lib\site-packages\six.py", line 703, in reraise
    raise value
  File "d:\cli\azure-cli\src\azure-cli-core\azure\cli\core\commands\__init__.py", line 688, in _run_job
    result = cmd_copy(params)
  File "d:\cli\azure-cli\src\azure-cli-core\azure\cli\core\commands\__init__.py", line 325, in __call__
    return self.handler(*args, **kwargs)
  File "d:\cli\azure-cli\src\azure-cli-core\azure\cli\core\__init__.py", line 776, in default_command_handler
    return op(**command_args)
  File "d:\cli\azure-cli\src\azure-cli\azure\cli\command_modules\resource\custom.py", line 1082, in create_resource_group
    return rcf.resource_groups.create_or_update(rg_name, parameters)
  File "d:\cli\env38\lib\site-packages\azure\mgmt\resource\resources\v2020_06_01\operations\_resource_groups_operations.py", line 154, in create_or_update
    raise exp
msrestazure.azure_exceptions.CloudError: Azure Error: LocationNotAvailableForResourceGroup
[...]

PR #15022, opened 2020-09-02
{REST} Capture SSLError in handle_exception

jiasli:
**Description**  

Split from #14588.

In `handle_exception`, capture `requests.exceptions.SSLError` which is generated by directly calling `requests`'s methods, and provide a better error message.

**Testing Guide**  

Enable Fiddler HTTPS decryption and make `az rest` call like:

```
az rest -u 'https://management.azure.com/subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/resourcegroups/fy?api-version=2018-02-01'
```

Output:
> Request failed: Certificate verification failed. This typically happens when using Azure CLI behind a proxy that intercepts traffic with a self-signed certificate. Please add this certificate to the trusted CA bundle. More info: https://docs.microsoft.com/en-us/cli/azure/use-cli-effectively#work-behind-a-proxy.
> 
> Error detail: HTTPSConnectionPool(host='management.azure.com', port=443): Max retries exceeded with url: /subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/resourcegroups/fy?api-version=2018-02-01 (Caused by SSLError(SSLError("bad handshake: Error([('SSL routines', 'tls_process_server_certificate', 'certificate verify failed')])")))


yonzhan:
REST

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -802,19 +804,17 @@ def send_raw_request(cli_ctx, method, url, headers=None, uri_parameters=None,  #
...
-        req = Request(method=method, url=url, headers=headers, params=uri_parameters, data=body)
-        prepped = s.prepare_request(req)
-
-        # Merge environment settings into session
-        settings = s.merge_environment_settings(prepped.url, {}, None, not should_disable_connection_verify(), None)
-        _log_request(prepped)
-        r = s.send(prepped, **settings)
-        _log_response(r)
-    except Exception as ex:  # pylint: disable=broad-except
-        raise CLIError(ex)
```
zhoxing-ms, line 817: May I ask why the CLIError is not wrapped here any more?

jiasli: If we wrap the exception with `CLIError`, the exception type will be lost and 
https://github.com/Azure/azure-cli/blob/a0f8c8ecbec012b07c00441639f5866043b3c218/src/azure-cli-core/azure/cli/core/util.py#L82
won't be hit anymore.

We are leaving the error handing to `handle_exception` instead.

houk-ms (approved):
Looks good. Add reference #14855

src/azure-cli-core/azure/cli/core/util.py
zhoxing-ms: Got it, thx~

zhoxing-ms (approved):
LGTM

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -807,19 +809,17 @@ def send_raw_request(cli_ctx, method, url, headers=None, uri_parameters=None,  #
...
-        s = Session()
-        req = Request(method=method, url=url, headers=headers, params=uri_parameters, data=body)
-        prepped = s.prepare_request(req)
-
-        # Merge environment settings into session
-        settings = s.merge_environment_settings(prepped.url, {}, None, not should_disable_connection_verify(), None)
-        _log_request(prepped)
-        r = s.send(prepped, **settings)
-        _log_response(r)
-    except Exception as ex:  # pylint: disable=broad-except
```
arrownj, line 821: Is it always `SSLError` ?

jiasli: No. We throw any exceptions and let `handle_exception` handle it.

arrownj (approved):

PR #15076, opened 2020-09-08
{Core} Add SessionId for telemetry

houk-ms:
**Description**  

SessionId is used to help us identify whether the users are running their CLI commands in the same terminal. It can help us analysis the user behavior, by which we may be able to cluster the commands and improve CLI error rate measurement. It may also help us separate interactive users from automation scripts.

This PR 
- adds session id  for telemetry
- adds package psutil as azure-cli dependency (except for cygwin)

**Note**

The package **psutil** was remove from CLI dependencies in #11665 because of cygwin incompatibility. In this PR, we reintroduce the package and provide a workaround to avoid installing it in cygwin.

**Known Issue**

For automation script, the sessionIDs of the commands in script will be the same except for the first command. It only occurs on None-windwos OSes and only in automation scripts.

### DIFF
--- a/src/azure-cli-core/azure/cli/core/telemetry.py
+++ b/src/azure-cli-core/azure/cli/core/telemetry.py
@@ -409,9 +409,7 @@ def _get_installation_id():
 
 @decorators.suppress_all_exceptions(fallback_return="")
 def _get_session_id():
-    # This function works as a workaround to get the terminal info: when the difference
-    # of create time between current process and its parent process is larger than a given
-    # threshold, the parent process will be viewed as a terminal process.
+    # As a workaround to get the terminal info as SessionId, this function may not be accurate.
 
     def get_hash_result(content):
         import hashlib
@@ -420,10 +418,12 @@ def _get_session_id():
         hasher.update(content.encode('utf-8'))
         return hasher.hexdigest()
 
-    # Usually, more than one layer of sub-process will be called when excuting a CLI command. While, the create time
-    # of these processes will be very close, usually in several milliseconds. We use 1 second as the threshold here.
-    time_threshold = 1
+    # Usually, more than one layer of sub-process will be started when excuting a CLI command. While, the create time
+    # of these sub-processes will be very close, usually in several milliseconds. We use 1 second as the threshold here.
+    # When the difference of create time between current process and its parent process is larger than the threshold, 
+    # the parent process will be viewed as the terminal process.
     import psutil
+    time_threshold = 1
     process = psutil.Process()
     while process and process.ppid() and process.pid != process.ppid():
         parent_process = process.parent()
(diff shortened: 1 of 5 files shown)


yonzhan:
Core

src/azure-cli-core/azure/cli/core/telemetry.py
```diff
@@ -407,6 +407,33 @@ def _get_installation_id():
...
+    # of these sub-processes will be very close, usually in several milliseconds. We use 1 second as the threshold here.
+    # When the difference of create time between current process and its parent process is larger than the threshold,
+    # the parent process will be viewed as the terminal process.
+    import psutil
+    time_threshold = 1
+    process = psutil.Process()
+    while process and process.ppid() and process.pid != process.ppid():
+        parent_process = process.parent()
+        if parent_process and process.create_time() - parent_process.create_time() > time_threshold:
+            content = '{}{}{}'.format(_get_installation_id(), parent_process.create_time(), parent_process.pid)
```
yungezz, line 431: I thought you mentioned will distinguish by terminal type right?

houk-ms: It turns out to be an invalid method. For the may be the cases where a shell was called in another shell, e.g, powershell will call cmd to execute CLI command. So we can't recognize the target terminal processs only by the name.

I think the method I proposed here is better for practice. It can deal with all kinds of OSes and their different terminals.

fengzhou-msft (approved):

src/azure-cli-core/azure/cli/core/telemetry.py
houk-ms: For automation script, the sessionIDs of the commands in script will be the same except for the first command. It only occurs on None-windwos OSes and only in automation scripts.

arrownj: Okay, you convince me. 👍

arrownj (approved):

pspag:
Hello, not sure if this is the right place to report this, but it seems this change introduced a regression issue which once more breaks building azure-cli with pip, in cygwin, due to the re-introduction of `psutil`. This was previously fixed in #11665. The issue had been originally reported in #9399.

Please let me know if you would like this reported in a new issue.

houk-ms:
@pspag Sorry for the inconvenience. We removed the psutil dependency for Cygwin in setup.py, but it only takes effect when using source installation.  We will schedule to fix the issue as soon as we can.

As a workaround to use the latest CLI in cygwin, try to
1. Download the source package from pypi
2. pip install the source package

pspag:
@houk-ms many thanks for your reply. No problem at all, since I came across this I just wanted to report it as I imagine it will be also affecting others. In the meantime I use 2.11.1 where this feature is not included.

PR #15104, opened 2020-09-09
{Package} remove upper bound in cryptography dependency

bluca:
It does not seem necessary, none of the changes appear to affect azure-cli usage.

https://github.com/pyca/cryptography/blob/master/CHANGELOG.rst

Distributions have already moved past 3.0, so this restriction is
problematic.

---

### DIFF
--- a/src/azure-cli/setup.py
+++ b/src/azure-cli/setup.py
@@ -126,7 +126,7 @@ DEPENDENCIES = [
     'azure-storage-common~=1.4',
     'azure-synapse-accesscontrol~=0.2.0',
     'azure-synapse-spark~=0.2.0',
-    'cryptography>=2.3.1,<3.0.0',
+    'cryptography>=2.3.1',
     'fabric~=2.4',
     'jsmin~=2.2.2',
     'pytz==2019.1',


yonzhan:
Package

yungezz:
hi @fengzhou-msft @bim-msft could you pls help review the change? what's impact to keyvault?

bim-msft:
Hi @bluca thanks for your feedback.

@yungezz I will do a full regression test on KeyVault later to make sure everything goes well.

bluca:
ping

yungezz:
hi @bluca sorry for late response, we're full with Ignite release before. Will test it then

bluca:
> hi @bluca sorry for late response, we're full with Ignite release before. Will test it then

Sure no problem, thanks

src/azure-cli/setup.py
```diff
@@ -126,7 +126,7 @@
     'azure-storage-common~=1.4',
     'azure-synapse-accesscontrol~=0.2.0',
     'azure-synapse-spark~=0.2.0',
-    'cryptography>=2.3.1,<3.0.0',
+    'cryptography>=2.3.1',
```
yungezz, line 129: normal practice in CLI is pin exact major version. what's impact of the change?

bluca: It might be normal practice here, but I'm afraid it's extremely bad practice to randomly pin to whatever is available that day of the week, without any specific reason. There is no incompatibility introduced, so it's unnecessary. The impact is that python3-cryptography is moving on everywhere, and the old deprecated versions are not available anymore, so we have to manually patch this downstream.

https://packages.debian.org/search?keywords=python3-cryptography&searchon=names&suite=all&section=all
https://packages.ubuntu.com/search?keywords=python3-cryptography&searchon=names&suite=all&section=all
https://rpmfind.net/linux/rpm2html/search.php?query=python3-cryptography

fengzhou-msft:
@bluca we've bumped cryptography:
https://github.com/Azure/azure-cli/blob/b791aa1714f01009f321c5ba195ec2dfba9150b1/src/azure-cli-core/setup.py#L50

Due to latest versions relying on rust and breaking installation on platforms like alpine we used for docker image, it's still upper bounded.

bluca:
That's great, thank you!

PR #15126, opened 2020-09-11
{Find} az find: Send User ID hash for rate limiting

mirdaki:
**Description**  
In order to better handle the increased traffic, the Aladdin service is adding hashed User IDs to the request header in order to accurately rate limit clients. These should not interfere with normal usage, just prevent bad actors (most likely run-a-way scripts) from slowing down or breaking the service for other users.

The hashed ID is only used for the rate limiting and is only stored in memory for a temporary amount of time. It is not stored in telemetry.

**Testing Guide**  
No change in command usage.


yonzhan:
az find

yungezz (approved):

PR #14128, opened 2020-06-29
[Profile] az account get-access-token: Show expiresOn for Managed Identity

jiasli:
## Description
Currently, `az account get-access-token` shows `"expiresOn": "N/A"` for Managed Identity:

```
> az login --identity
> az account get-access-token
{
  "accessToken": "eyJ0eXAi...",
  "expiresOn": "N/A",
  "subscription": "0b1f6471-1bf0-4dda-aec3-cb9272f09590",
  "tenant": "54826b22-38d6-4fb2-bad9-b7b93a3e9c5a",
  "tokenType": "Bearer"
}
```

It works for a user account:

```
> az login
> az account get-access-token
{
  "accessToken": "eyJ0eXAi...",
  "expiresOn": "2020-06-29 17:22:16.443545",
  "subscription": "a18897a6-7e44-457d-9260-f2854c0aca42",
  "tenant": "72f988bf-86f1-41af-91ab-2d7cd011db47",
  "tokenType": "Bearer"
}
```

This is because `msrestazure.azure_active_directory.MSIAuthentication`'s token entry has `expires_on`, while ADAL's token entry has `expiresOn`. 

This PRs fixes this by converting the epoch `expires_on` to ISO `expiresOn`:

```
> az account get-access-token
{
  "accessToken": "eyJ0eXAi...",
  "expiresOn": "2020-06-30 06:14:41.000000",
  "subscription": "0b1f6471-1bf0-4dda-aec3-cb9272f09590",
  "tenant": "54826b22-38d6-4fb2-bad9-b7b93a3e9c5a",
  "tokenType": "Bearer"
}
```

## Testing Guide
```
az login --identity
az account get-access-token
```

## More info

In fact, the REST call in ADAL for token retrieval is:

Request:

```http
POST https://login.microsoftonline.com/72f988bf-86f1-41af-91ab-2d7cd011db47/oauth2/token HTTP/1.1
```

Response:

```json
{
    "token_type": "Bearer",
    "scope": "user_impersonation",
    "expires_in": "3595",
    "ext_expires_in": "3595",
    "expires_on": "1593413327",
    "not_before": "1593409427",
    "resource": "https://management.core.windows.net/",
    "access_token": "...",
    "refresh_token": "...",
    "foci": "1"
}
```

ADAL ignores `expires_on`, but calculates `expiresOn` based on the current time and `expires_in`.

[...]

PR #15131, opened 2020-09-11
{Profile} Fix: test_get_raw_token fails on non-UTC machine

jiasli:
**Description**  
Inspired by https://github.com/Azure/azure-cli/pull/14718. Fix: `test_get_raw_token` fails on non-UTC machine.

The test `test_get_raw_token` was introduced in #14128.

When using `az account get-access-token` for managed identity, `expiresOn` will be converted from POSIX to datetime in **local timezone**. For example, `1593497681` is converted to 

- `2020-06-30 06:14:41.000000` on a UTC machine
- `2020-06-30 14:14:41.000000` on a UTC+8 machine

The test `test_get_raw_token fails` is written for a UTC machine on DevOps and this test will fail on a non-UTC machine (such as UTC+8).

This PR patches `fromtimestamp` with `utcfromtimestamp` to force `expiresOn` to be converted to UTC to make the test pass on any machine.


yonzhan:
Profile

Juliehzl (approved):
LGTM

arrownj (approved):
👍

PR #15156, opened 2020-09-14
[SQL VM] `az sql vm show`: Add configuration options to --expand flag

rchlkm:
**Description**  
This PR makes a small addition to the --expand flag.

The --expand flag previously only accepted `*` as an argument. With these changes, the --expand flag accepts the following: AutoBackupSettings, AutoPatchingSettings, KeyCredentialSettings, ServerConfiguratationsManagementSettings. The flag can also accept any space-separated combination of these arguments.

Tests added accordingly.

**Testing Guide**  

`az sql vm show -g <Resource Group Name> -n <SQL VM Name> --expand <Setting1> <Setting2>`

**History Notes**  

---


rchlkm:
@yareyes

yonzhan:
add to S176

src/azure-cli/HISTORY.rst
```diff
@@ -1609,6 +1609,7 @@ Release History
 
 **SQL VM**
 
+* Added more arguments to `az sql vm show --expand` command
```
yareyes, line 1612: I think you need to add again - for the latest version released (Example 2.0.16) a 
**SQL VM **
* Added more arguments to `az sql vm show --expand` command 

Adding on the old history won't show up on the latest released CLI.

yareyes:
> 2.11.0

It should go over here -> 

---
Refers to: src/azure-cli/HISTORY.rst:6 in 42f5c09. [](commit_id = 42f5c09e91a860cf417fe3a54135bf9564499a2a, deletion_comment = False)

src/azure-cli/azure/cli/command_modules/sqlvm/_validators.py
```diff
@@ -108,3 +108,19 @@ def validate_sqlmanagement(namespace):
...
+# pylint: disable=too-many-statements,line-too-long
+def validate_expand(namespace):
+    '''
+    Validates if expand passes accepted values.
+    '''
+    allowed_values = ['*', 'AutoBackupSettings', 'AutoPatchingSettings', 'KeyVaultCredentialSettings', 'ServerConfigurationsManagementSettings']
+
+    if namespace.expand is not None:
+        items = namespace.expand.split(",")
+        if "*" not in items:
```
yareyes, line 122: >"*" [](start = 11, length = 3)

what happens if customer passes something like *,Test,Mess -> What will be the behavior?

rchlkm: Since * is included, the * will override the unacceptable values (Test, Mess) and display all the settings. Same goes for a case like Test,*,Mess.

yareyes: Ok cool

---
In reply to: [488296563](https://github.com/Azure/azure-cli/pull/15156#discussion_r488296563) [](ancestors = 488296563)

yareyes (approved):
:shipit:

src/azure-cli/azure/cli/command_modules/sqlvm/_params.py
```diff
@@ -158,8 +159,8 @@ def load_arguments(self, _):
                    arg_type=get_location_type(self.cli_ctx),
                    validator=get_default_location_from_resource_group)
         c.argument('expand',
-                   help='Get the SQLIaaSExtension configuration settings.',
-                   arg_type=get_enum_type(['*']))
+                   help='Get the SQLIaaSExtension configuration settings. To view all settings, use *. To select only a few, the settings must be comma-separted. Allowed values: *, AutoBackupSettings, AutoPatchingSettings, KeyVaultCredentialSettings, ServerConfigurationsManagementSettings',
+                   validator=validate_expand)
```
Juliehzl, line 163: please show all enum values here.

rchlkm:
@Juliehzl thanks for reviewing this PR. This PR is actually specific to the SQL VM command module, not SQL. I believe the title should be [SQL VM].

src/azure-cli/azure/cli/command_modules/sqlvm/_params.py
rchlkm: The SQL VM RP currently does not have an enum for the SQLIaaSExtension configuration settings. I'm using the validate_expand to compare the user input and allowed values. In the help flag, I added manually added the allowed values to communicate with users which values are allowed.

If I were to implement this with enums, I would need to make changes to the SQL VM RP and update the Python model. Would I need to take these steps? Thanks!

rchlkm: @Juliehzl Is there a command in CLI that supports a combination of allowed values separated by a comma?

In this expand flag, we want users to be able to input one or more values. It seemed that `arg_type` usually only accepts one value. In `validate_expand`, I check the cases when more than one value is inputted.

Juliehzl:
> @Juliehzl Zunli Hu FTE thanks for reviewing this PR. This PR is actually specific to the SQL VM command module, not SQL. I believe the title should be [SQL VM].

updated

src/azure-cli/azure/cli/command_modules/sqlvm/_params.py
Juliehzl: Example
https://github.com/Azure/azure-cli/blob/f67024b48bd0efea4084f028897d7b7a7abf35c9/src/azure-cli/azure/cli/command_modules/storage/_params.py#L247

Juliehzl: discussed online. get_enum_type() is not suitable for the case.
We have another typical design for such case. We add nargs='+' with arg_type=get_enum_type([a,b,c,d]). And then customer could use `--expand a b ` to input [a,b] for expand variable.

rchlkm: Changes using nargs='+' have been committed.

yareyes (approved):

Juliehzl (approved):

PR #15160, opened 2020-09-15
[AppConfig] Support AAD auth for data operations

avanigupta:
**Description**  

App Configuration supports AAD authentication and authorization for all data operations. Since we are now using the [official Python SDK for AppConfiguration](https://github.com/Azure/azure-sdk-for-python/tree/master/sdk/appconfiguration/azure-appconfiguration), we can easily support AAD auth in appconfig CLI module by utilizing the `az login` credentials.
This PR introduces two new arguments for users to authenticate using AAD credentials:

- --auth-mode (login or key; default is key which preserves the current behavior)

- --endpoint (AppConfiguration endpoint like `https://storename.azconfig.io`)

For any existing AppConfig commands, providing the `--name` and `--connection-string` parameters will use key (HMAC) auth for data operations (preserving current behavior). Instead of these two arguments, if users provide `--auth-mode login` and `--endpoint` parameters, AAD credentials will be used to connect to AzureAppConfigurationClient.

**Testing Guide**  

1. Create a new App Configuration with name "aadtestcli".
2. Create a key-value with key "testKey" and any value.
3. Create a new service principal and assign it "App Configuration Data Reader" role in your App Configuration. Wait for about 15 minutes for permissions to propagate.
4. Login to CLI with the service principal credentials.

This read request should succeed:
`az appconfig kv show --key testKey --auth-mode login --endpoint https://aadtestcli.azconfig.io`

This write request should fail with 'Forbidden' error:
`az appconfig kv lock --key testKey --auth-mode login --endpoint https://aadtestcli.azconfig.io`

**History Notes**  

---


yonzhan:
AppConfig

yungezz:
hi @bim-msft could you pls review the PR? thanks.

avanigupta:
Hi @recao, could you please help review this PR?

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -118,3 +128,34 @@ def prep_null_label_for_url_encoding(label=None):
...
+        profile = Profile(cli_ctx=cmd.cli_ctx)
+        # Due to this bug in get_login_credentials: https://github.com/Azure/azure-cli/issues/15179,
+        # we need to manage the AAD scope by passing appconfig endpoint as resource
+        cred, _, _ = profile.get_login_credentials(resource=endpoint)
+        azconfig_client = AzureAppConfigurationClient(credential=cred,
+                                                      base_url=endpoint,
+                                                      user_agent=HttpHeaders.USER_AGENT)
+
+    if not azconfig_client:
+        raise CLIError("Could not get App Configuration client due to insufficient permissions.")
```
shenmuxiaosen, line 159: This check is misleading. AzureAppConfigurationClient will throw exceptions if it can't create the client.

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -6,41 +6,51 @@
...
+    except Exception as ex:
         raise CLIError(
-            'Cannot find the App Configuration {}. Check if it exists in the subscription that logged in. '.format(config_store_name))
+            'Failed to get access keys for the App Configuration "{}". Make sure that the account that logged in has "Contributor" role assigned for this App Configuration store.\n{}'.format(config_store_name, str(ex)))
 
     raise CLIError('Cannot find a read write access key for the App Configuration {}'.format(
         config_store_name))
 
 
 def resolve_resource_group(cmd, config_store_name):
```
shenmuxiaosen, line 42: >resolve_resource_group [](start = 4, length = 22)

nit: As we modified the return, it doesn't only resolve the rg. Maybe call it resolve_store_metadata?

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -6,41 +6,51 @@
...
         access_keys = config_store_client.list_keys(
             resource_group_name, config_store_name)
         for entry in access_keys:
             if not entry.read_only:
                 return connection_string_template.format(endpoint, entry.id, entry.value)
-    except Exception:
+    except Exception as ex:
         raise CLIError(
-            'Cannot find the App Configuration {}. Check if it exists in the subscription that logged in. '.format(config_store_name))
+            'Failed to get access keys for the App Configuration "{}". Make sure that the account that logged in has "Contributor" role assigned for this App Configuration store.\n{}'.format(config_store_name, str(ex)))
```
shenmuxiaosen, line 36: >Contributor [](start = 118, length = 11)

We don't need to specify which role is needed as it can be contributor, reader, owner etc. Maybe just say "Failed to get access keys for the App Configuration "{}". Make sure that the account that logged in has proper role assigned for this App Configuration store.\n{}"

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -6,41 +6,51 @@
...
-        "App Configuration store: {} does not exist".format(config_store_name))
+    try:
+        config_store_client = cf_configstore(cmd.cli_ctx)
+        all_stores = config_store_client.list()
+        for store in all_stores:
+            if store.name.lower() == config_store_name.lower():
+                # Id has a fixed structure /subscriptions/subscriptionName/resourceGroups/groupName/providers/providerName/configurationStores/storeName"
+                return store.id.split('/')[4], store.endpoint
+    except ErrorException as ex:
+        raise CLIError("Failed to get the list of App Configuration stores for the current user. Make sure that the account that logged in has 'Reader' or 'Contributor' role assigned for the required App Configuration store.\n{}".format(str(ex)))
```
shenmuxiaosen, line 51: > 'Reader' or 'Contributor' [](start = 142, length = 26)

"Make sure that the account that logged in has proper role" We can't list all possible roles and we mentioned it failed to list stores. So it indicates that cusomters need to have a role which have resources read access.

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -6,41 +6,51 @@
...
+
+    raise CLIError("Failed to find the App Configuration store '{}'. Make sure that the account that logged in has 'Reader' or 'Contributor' role assigned for this App Configuration store.".format(config_store_name))
```
shenmuxiaosen, line 54: I assume ErrorException indicates the permsision issue? If it throws other error, it has nothing to do with role.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -72,6 +75,13 @@ def load_arguments(self, _):
         c.argument('all_', options_list=['--all'], action='store_true', help="List all items.")
         c.argument('fields', arg_type=fields_arg_type)
         c.argument('sku', help='The sku of App Configuration', arg_type=get_enum_type(['Free', 'Standard']))
+        c.argument('endpoint', help='If auth mode is "login", provide endpoint of the App Configuration in the format "https://storename.azconfig.io". You can configure the default endpoint using `az configure --defaults appconfig_endpoint=<endpoint>`', configured_default='appconfig_endpoint')
```
shenmuxiaosen, line 78: >https://storename.azconfig.io [](start = 119, length = 29)

This may not be true especially when we have the georeplciation. We can just say "provide the endpoint" or  "provide the endpoint which can be retrived by 'appconfig show' command".

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -63,7 +63,10 @@ def load_arguments(self, _):
 
     with self.argument_context('appconfig') as c:
         c.argument('resource_group_name', arg_type=resource_group_name_type)
-        c.argument('name', options_list=['--name', '-n'], id_part='None', help='Name of the App Configuration. You can configure the default name using `az configure --defaults app_configuration_store=<name>`', configured_default='app_configuration_store')
+        c.argument('name', options_list=['--name', '-n'], id_part='None',
+                   help='Name of the App Configuration. If the auth mode is "key", you should have "Contributor" role assigned in order to fetch access keys. ' +
+                   'If the auth mode is "login", you should have "Reader" or "Contributor" role assigned in order to fetch store endpoint. ' +
```
shenmuxiaosen, line 68: I don't think we need this explanation. As we have help message for auth_mode and lots of helpful error messages if they misuse this parameter.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -72,6 +75,13 @@ def load_arguments(self, _):
         c.argument('all_', options_list=['--all'], action='store_true', help="List all items.")
         c.argument('fields', arg_type=fields_arg_type)
         c.argument('sku', help='The sku of App Configuration', arg_type=get_enum_type(['Free', 'Standard']))
+        c.argument('endpoint', help='If auth mode is "login", provide endpoint of the App Configuration in the format "https://storename.azconfig.io". You can configure the default endpoint using `az configure --defaults appconfig_endpoint=<endpoint>`', configured_default='appconfig_endpoint')
+        c.argument('auth_mode', arg_type=get_enum_type(['login', 'key']), configured_default='appconfig_auth_mode', validator=validate_auth_mode,
```
shenmuxiaosen, line 79: Can you add example to use aad auth for some popular commands like import/export?

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -86,7 +96,7 @@ def resolve_connection_string(cmd, config_store_name=None, connection_string=Non
 
     if not string:
         raise CLIError(
-            'Please specify config store name or connection string(suggested).')
+            'If you are using "key" auth mode, please specify config store name or connection string(suggested).')
     return string
```
shenmuxiaosen, line 100: No change needed. Previously this method is called almost for every single command to finalized the connection string. As we start to support aad, this one should only be invoked when resolving connection string is needed. So it must be under “key” auth mode context.

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
avanigupta: ErrorException is only thrown if the logged in account doesn't have access to any config store, i.e. config_store_client.list() returns no config stores.
It's possible that the logged in account has access to some stores, but not the one they are trying to access here. In that case, we need to raise CLI error if the return statement on line 49 was never hit.

---
In reply to: [494700304](https://github.com/Azure/azure-cli/pull/15160#discussion_r494700304) [](ancestors = 494700304)

shenmuxiaosen: You are right. But it is hard to guess the reason, like it can be no access or the store does not exist. 

---
In reply to: [494716795](https://github.com/Azure/azure-cli/pull/15160#discussion_r494716795) [](ancestors = 494716795,494700304)

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
avanigupta: Done

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
avanigupta: Done

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
avanigupta: Done

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
avanigupta: Done

src/azure-cli/azure/cli/command_modules/appconfig/_help.py
```diff
@@ -121,6 +125,8 @@
     text: az appconfig kv export -n MyAppConfiguration -d appconfig --dest-name AnotherAppConfiguration --key * --label * --preserve-labels
   - name: Export all keys and feature flags with all labels to another App Configuration and overwrite destination labels.
     text: az appconfig kv export -n MyAppConfiguration -d appconfig --dest-name AnotherAppConfiguration --key * --label * --dest-label ExportedKeys
+  - name: Export all keys to another App Configuration using  your 'az login' credentials.
+    text: az appconfig kv export -d appconfig -n MyAppConfiguration --auth-mode login --dest-name AnotherAppConfiguration --dest-auth-mode login --key * --label * --preserve-labels
```
shenmuxiaosen, line 129: nit: can we use endpoint in our example for aad auth? Otherwise, the command is more likely to fail and we have to explain the login has to have both ARM and app configuration role.

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -118,3 +125,38 @@ def prep_null_label_for_url_encoding(label=None):
...
+
+def get_appconfig_data_client(cmd, name, connection_string, auth_mode, endpoint):
+    azconfig_client = None
+    if auth_mode == "key":
+        connection_string = resolve_connection_string(cmd, name, connection_string)
+        try:
+            azconfig_client = AzureAppConfigurationClient.from_connection_string(connection_string=connection_string,
+                                                                                 user_agent=HttpHeaders.USER_AGENT)
+        except Exception as ex:
+            raise CLIError("An exception occurred while connecting to App Configuration: {}".format(str(ex)))
```
shenmuxiaosen, line 138: Will AzureAppConfigurationClient.from_connection_string connect with App Configuration? I assume it will just create a client and won't notice failure until it actually sends out requests.

avanigupta: AzureAppConfigurationClient.from_connection_string will raise an exception if the connection string is invalid. There is an additional check for empty segments in connection string that we don't check on CLI side. 
Apart from that, you are right, there won't be any exception until a request is sent out. The error message here is more general because I felt that saying something like "failed to create client" is an implementation detail.

src/azure-cli/azure/cli/command_modules/appconfig/_help.py
avanigupta: sure

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
shenmuxiaosen: If the connection string check in SDK failed, then throw it out directly. No need to wrap in a try-catch. The current error message indicates there is something wrong connecting to App Configuration like network or server error. Instead, it is client side's problem. 

---
In reply to: [496238201](https://github.com/Azure/azure-cli/pull/15160#discussion_r496238201) [](ancestors = 496238201)

avanigupta: To catch the connection string error, we would still need to wrap this call in try-catch block as we are not using the SDK to validate conn string directly.
How about the error message: "Operation failed to due to an exception: {}"

shenmuxiaosen: Whatever exception the SDK throws, it is unexpected and can bubble up to clients. For some known errors, we may react on it and wrap it to CLIError instead of catching all exceptions. 

---
In reply to: [496284698](https://github.com/Azure/azure-cli/pull/15160#discussion_r496284698) [](ancestors = 496284698)

avanigupta: Fixed

avanigupta:
Hi @bim-msft, @jiasli, can you please review this PR?

yungezz:
hi @fengzhou-msft could you pls help to review since @bim-msft is oof? thanks.

shenmuxiaosen (approved):
:shipit:

src/azure-cli/azure/cli/command_modules/appconfig/_help.py
```diff
@@ -121,6 +123,8 @@
     text: az appconfig kv export -n MyAppConfiguration -d appconfig --dest-name AnotherAppConfiguration --key * --label * --preserve-labels
   - name: Export all keys and feature flags with all labels to another App Configuration and overwrite destination labels.
     text: az appconfig kv export -n MyAppConfiguration -d appconfig --dest-name AnotherAppConfiguration --key * --label * --dest-label ExportedKeys
+  - name: Export all keys to another App Configuration using  your 'az login' credentials.
```
fengzhou-msft, line 126:
```suggestion
  - name: Export all keys to another App Configuration using your 'az login' credentials.
```

avanigupta: Done

fengzhou-msft (approved):

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -118,3 +125,38 @@ def prep_null_label_for_url_encoding(label=None):
...
+        profile = Profile(cli_ctx=cmd.cli_ctx)
+        # Due to this bug in get_login_credentials: https://github.com/Azure/azure-cli/issues/15179,
+        # we need to manage the AAD scope by passing appconfig endpoint as resource
+        cred, _, _ = profile.get_login_credentials(resource=endpoint)
+        try:
+            azconfig_client = AzureAppConfigurationClient(credential=cred,
+                                                          base_url=endpoint,
+                                                          user_agent=HttpHeaders.USER_AGENT)
+        except (ValueError, TypeError) as ex:
+            raise CLIError("Operation failed to due to an exception: {}".format(str(ex)))
```
fengzhou-msft, line 160: @jiasli can you help review this part?

avanigupta:
Hi @yonzhan , since Jiashuo is OOF, can someone else help review this PR?

yonzhan:
@avanigupta We are in Chinese National holiday. @jiasli can help review this PR after holiday. Sorry for the inconvenience caused.

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
jiasli: The usage of `get_login_credentials` is correct for now. After https://github.com/Azure/azure-cli/pull/15184 is merged, `resource=endpoint` can be removed and CLI will honor `scopes` provided by the SDK.

jiasli (approved):

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
jiasli, line 160: For the error message, I would recommend using a more specific word, rather than the generic "Operation" to help you troubleshoot the issue, something like:

> Failed to initialize AzureAppConfigurationClient due to an exception:

avanigupta: Done

src/azure-cli/requirements.py3.Darwin.txt
```diff
@@ -3,7 +3,7 @@ antlr4-python3-runtime==4.7.2
 applicationinsights==0.11.9
 argcomplete==1.11.1
 asn1crypto==0.24.0
-azure-appconfiguration==1.1.0
+azure-appconfiguration==1.1.1
```
yungezz, line 6: pls run live test to make sure no breaking change  in sdk version bumping

avanigupta: I've run all live tests and updated test recordings in this PR.

yungezz (approved):

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -118,3 +125,38 @@ def prep_null_label_for_url_encoding(label=None):
...
+                else:
+                    raise CLIError("App Configuration endpoint or name should be provided if auth mode is 'login'.")
+            except Exception as ex:
+                raise CLIError(str(ex) + "\nYou may be able to resolve this issue by providing App Configuration endpoint instead of name.")
+
+        from azure.cli.core._profile import Profile
+        profile = Profile(cli_ctx=cmd.cli_ctx)
+        # Due to this bug in get_login_credentials: https://github.com/Azure/azure-cli/issues/15179,
+        # we need to manage the AAD scope by passing appconfig endpoint as resource
+        cred, _, _ = profile.get_login_credentials(resource=endpoint)
```
jiasli, line 154: #15184 has been merged. You may now safely remove `resource=endpoint`.

Issue #14020, opened 2020-06-18
az keyvault secret list-versions --maxresults has no effect


mungk:
## Describe the bug
The --maxresults parameter does not seem to have any effect. All versions are returned.

**Command Name**
`az keyvault secret list-versions`

**Errors:**
None

## To Reproduce:
Note that argument values have been redacted, as they may contain sensitive information. Use your own key vault name.

Create a secret and give it multiple versions:
- `az keyvault secret set --vault-name YourKeyVault --name VersionTest --value "1"`
- `az keyvault secret set --vault-name YourKeyVault --name VersionTest --value "2"`
- `az keyvault secret set --vault-name YourKeyVault --name VersionTest --value "3"`
- `az keyvault secret set --vault-name YourKeyVault --name VersionTest --value "4"`
- `az keyvault secret set --vault-name YourKeyVault --name VersionTest --value "5"`

List the versions using maxresults parameter:
- `az keyvault secret list-versions --vault-name YourKeyVault --name VersionTest --maxresults 1`

## Observed Behavior:
All versions are returned, even if there are more versions than the maxresults specified:

>[
>  {
>    "attributes": {
>      "created": "2020-06-18T15:48:21+00:00",
>      "enabled": true,
>      "expires": null,
>      "notBefore": null,
>      "recoveryLevel": "Recoverable",
>      "updated": "2020-06-18T15:48:21+00:00"
>    },
>    "contentType": null,
>    "id": "https://YourKeyVault.vault.azure.net/secrets/VersionTest/75773177976b4e63814f8901e111a6c6",
>    "managed": null,
>    "name": "VersionTest",
>    "tags": null
>  },
>  {
>    "attributes": {
>      "created": "2020-06-18T15:52:18+00:00",
>      "enabled": true,
>      "expires": null,
>      "notBefore": null,
>      "recoveryLevel": "Recoverable",
>      "updated": "2020-06-18T15:52:18+00:00"
>    },
>    "contentType": null,
>    "id": "https://YourKeyVault.vault.azure.net/secrets/VersionTest/8e5edb5877324575837138dc142fd3a9",
>    "managed": null,
>    "name": "VersionTest",
>    "tags": {
>      "file-encoding": "utf-8"
>    }
>  },
>  {
>    "attributes": {
>      "created": "2020-06-18T15:48:04+00:00",
>      "enabled": true,
>      "expires": null,
>      "notBefore": null,
>      "recoveryLevel": "Recoverable",
[...]

PR #15165, opened 2020-09-15
[KeyVault] Fix the issue that `--maxresults` does not take effect while listing resources or resource versions

bim-msft:
Fix: #14020 

`--maxresults` does not take effect while listing resources or resource versions. Added a transformer to truncate the result.

e.g. `az keyvault secret list-versions --vault-name {vault} -n {secret} --maxresult 10`

---


src/azure-cli/azure/cli/command_modules/keyvault/_transformers.py
```diff
@@ -14,6 +14,13 @@ def _multi_transformers(output, **command_args):
     return _multi_transformers
 
 
+def keep_max_results(output, **command_args):
```
arrownj, line 17: It seems that there is no logic handle the --maxresults before ?

bim-msft: @arrownj `--maxresults` was directly passed into SDK function, but seems no effect.

fengzhou-msft (approved):
LGTM

PR #15174, opened 2020-09-15
[AppService] functionapp: Fixed an issue where config-zip could throw an exception even if successful

### LINKED ISSUE #15172: Unable to deploy Linux Function App
Cannot deploy a function app using the az CLI (command: az functionapp deployment source config-zip  -g test  -n mahoeger-linux --src ./helloworld.zip).

The deployment fails with 2.7.0 and 2.11.1, but is ok with 2.0.77. I am running on a Mac, using az CLI 2.11.1 and deploying to Linux Consumption function apps that are already deployed (both v2 and v3 sites).

```
mariehoeger$az functionapp deployment source config-zip  -g test  -n mahoeger-linux --src ./helloworld.zip
Unknown error============================== 100.0% ..
```

gzuber:
**Description**  

Resolves #15172 

This resolves an issue where `az functionapp deployment source config-zip` would throw an exception even if successful when pushing to linux consumption apps.

**Testing Guide**  

**History Notes**  

[AppService] az functionapp deployment source config-zip: Fixed an issue where config-zip could throw an exception on success on linux consumption 

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -574,8 +574,7 @@ def upload_zip_to_storage(cmd, resource_group_name, name, src, slot=None):
             raise ex
     except DefaultErrorResponseException as ex:
         if ex.response.status_code != 200:
-            raise ex 
-
+            raise ex
 
 
 def _generic_settings_operation(cli_ctx, resource_group_name, name, operation_name,
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -568,10 +568,14 @@ def upload_zip_to_storage(cmd, resource_group_name, name, src, slot=None):
             client.web_apps.sync_function_triggers_slot(resource_group_name, name, slot)
         else:
             client.web_apps.sync_function_triggers(resource_group_name, name)
-    except CloudError as ce:
+    except CloudError as ex:
         # This SDK function throws an error if Status Code is 200
-        if ce.status_code != 200:
-            raise ce
+        if ex.status_code != 200:
+            raise ex
+    except DefaultErrorResponseException as ex:
+        if ex.response.status_code != 200:
+            raise ex 
+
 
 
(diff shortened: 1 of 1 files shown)


src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -568,10 +568,14 @@ def progress_callback(current, total):
...
-    except CloudError as ce:
+    except CloudError as ex:
         # This SDK function throws an error if Status Code is 200
-        if ce.status_code != 200:
-            raise ce
+        if ex.status_code != 200:
+            raise ex
+    except DefaultErrorResponseException as ex:
+        if ex.response.status_code != 200:
+            raise ex
```
qwordy, line 577: I think your changes will throw more exceptions. Originally, it throws CloudError . Now, it will throw CloudError or DefaultErrorResponseException .

yonzhan:
AppService

src/azure-cli/azure/cli/command_modules/appservice/custom.py
gzuber: Yeah the sdk function `sync_function_triggers` is already throwing a `DefaultErrorResponseException` so that would happen without this change anyways.

What's happening is that the sdk function `sync_function_triggers` expects the rest api call to sync triggers to return a 204. If it doesn't return a 204, it throws a `CloudError` exception. Turns out the rest api call to sync triggers  returns a 200 on success, not a 204 (not sure if this is a docs issue or what but I've tested it and it seems to be the case). So the sdk function `sync_function_triggers` throws an exception because even though it's a success, it's not 204.

Previously, this was a CloudError and we would catch that exception, and it it's a 200 status code we would ignore the exception because it was actually a success. At some point, I'm not sure when or why, the sdk function `sync_function_triggers` started throwing `DefaultErrorResponseException`s instead of `CloudError` exceptions for 200 status codes. This would result in the user seeing an exception from running config-zip and thinking it failed. So this just covers this new case of the sdk function throwing a `DefaultErrorResponseException` on success.

yungezz:
cc @houk-ms for awareness since new error type.

qwordy (approved):

PR #15248, opened 2020-09-22
CLOSED

jorgecotillo:
**Description**  

This PR removes the preview flag from az deployment-scripts commands, and uses the latest "non-GA" api-version which is 2020-10-01.

**Testing Guide**  
Not much changes from before, just added a create and read corresponding to 2020-10-01 api-version.

**History Notes**  
N.A
---


yonzhan:
version update

src/azure-cli/setup.py
```diff
@@ -90,7 +90,7 @@
     'azure-mgmt-iotcentral~=3.0.0',
     'azure-mgmt-iothub~=0.12.0',
     'azure-mgmt-iothubprovisioningservices~=0.2.0',
-    'azure-mgmt-keyvault~=7.0.0b3',
+    'azure-mgmt-keyvault==7.0.0b3',
```
yungezz, line 93: this is pined already. btw, why the change? did you met any problem? thanks.

zhoxing-ms: @jorgecotillo could you please take look at this problem?

src/azure-cli/azure/cli/command_modules/resource/tests/latest/test_resource.py
```diff
@@ -1622,31 +1622,49 @@ def test_subscription_level_what_if_ts(self, resource_group, resource_group_loca
 
 
 class DeploymentScriptsTest(ScenarioTest):
-    @ResourceGroupPreparer(name_prefix='cli_test_deployment_scripts')
+    @ResourceGroupPreparer(name_prefix='cli_test_deployment_scripts', location='brazilsouth')
```
jorgecotillo, line 1625: There's a known issue in ARM that "list requests" when the RP has more than one endpoint it ends up duplicating the resources, brazil south doesn't have multiple endpoints therefore used this region to ensure a list request only returns one resource (in case you wonder why we are using a specific location).

src/azure-cli/setup.py
jorgecotillo: Not sure how this got updated, perhaps I pulled the latest and messed up the version? Anyways, this update is no longer part of my PR :)

jorgecotillo:
@yonzhan this PR is ready to be merged, can you please remove the flag? Thanks!

src/azure-cli/azure/cli/command_modules/resource/tests/latest/test_resource.py
```diff
@@ -1622,31 +1622,49 @@ def test_subscription_level_what_if_ts(self, resource_group, resource_group_loca
...
+            'api_version': '2019-10-01-preview',
             'template_file': os.path.join(curr_dir, 'deployment-scripts-deploy.json').replace('\\', '\\\\'),
         })
 
         count = 0
         self.cmd('deployment-scripts list',
                  checks=self.check("length([?name=='{deployment_script_name}'])", count))
 
-        self.cmd('deployment group create -g {resource_group} -n {deployment_name} --template-file "{template_file}" --parameters scriptName={deployment_script_name}', checks=[
+        self.cmd('deployment group create -g {resource_group} -n {deployment_name} --template-file "{template_file}" --parameters scriptName={deployment_script_name} --parameters apiVersion="{api_version}"', checks=[
```
zhoxing-ms, line 1640: May I ask why we keep the test for the historical api-version?

jorgecotillo: Just to make sure the previous version works, main reason is that in the previous version the identity property is required and in the latest is optional, so just keeping a test to ensure both approaches work, not sure if is an accepted approach or not, happy to remove the historical api-version.

zhoxing-ms: If the user will still call the old api-version in new CLI version, then I think this test can be retained.
If the user has barely used the old api-version in new CLI version, we can delete it~

jorgecotillo: Since we can not specify the api version, I'll delete this. Thanks @zhoxing-ms for the feedback, will provide a new iteration.

jorgecotillo:
@zhoxing-ms closing this PR, messed up the merge, will create a new PR.

PR #15250, opened 2020-09-22
[AKS] Add node image only option for CLI

gtxistxgao:
**Description**  

To GA node image upgrade, we need to add support on CLI. This PR adds the ability to use following commands to upgrade node image:

- Upgrade the node image version of all node pools in a cluster : `az aks upgrade -g resourceGroupName -n clusterName --node-image-only`

- Upgrade one node pool's node image version: `az aks nodepool upgrade -g resourceGroupName --cluster-name clusterName -n nodePoolName --node-image-only`

**Testing Guide**  

**History Notes**  

---


yonzhan:
AKS

src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -2904,12 +2934,25 @@ def aks_agentpool_scale(cmd, client, resource_group_name, cluster_name,
 
 
 def aks_agentpool_upgrade(cmd, client, resource_group_name, cluster_name,
-                          kubernetes_version,
                           nodepool_name,
+                          kubernetes_version='',
+                          node_image_only=False,
                           no_wait=False):
     instance = client.get(resource_group_name, cluster_name, nodepool_name)
```
arrownj, line 2941: These 2 lines can be moved below the `node_image_only` part.

arrownj (commented):
Hi @gtxistxgao , just as other PRs, as this is a new option, could you please also add tests for it ? I think there would be an existing upgrade test case already, maybe we just add add this option in this case.

src/azure-cli/azure/cli/command_modules/acs/custom.py
gtxistxgao: make sense. moved!

gtxistxgao:
> Hi @gtxistxgao , just as other PRs, as this is a new option, could you please also add tests for it ? I think there would be an existing upgrade test case already, maybe we just add add this option in this case.

Hi, test added. There is no existing one so I create 2 more for this.

arrownj (approved):

Issue #13907, opened 2020-06-09
[webapp]:Import App Service Certificates into a Web App using the Azure CLI


tsgenge:
**Is your feature request related to a problem? Please describe.**
We're using App Service Certificates rather than PFX files imported into Certificates in the Key vault. The problem is the key vault "certificate" created during the enrollment process goes into the Secrets category in the Key Vault. Consequently, the **az webapp config ssl import** command doesn't find the certificate in the key vault. 

**Describe the solution you'd like**
I'd like the ssl import command to also check the secrets category in the keyvault if the certificate by the name provided cannot be found. I know, I know, the actual root cause here is the App Service Certificate creation process using the wrong key vault category! But, we can import App Service Certificates into an App Service via the Web UI, so it's be great to be able to do it in the Azure CLI too.

**Describe alternatives you've considered**
We have to export the app service certificate to PFX, then import it into the key vault again, this time to the Certificate type. *sigh*

**Additional context**
We have to do the PFX import every year. What a BORE.

yonzhan:
webapp

Kotasudhakarreddy:
@ThejaChoudary - try to repro the issue.

panchagnula:
@tsgenge APP service certificates do use KV - however you cannot use the Import app service certificate command for App service certificates - we will have separate commands for those in the future. Thanks!

panchagnula:
We don't have an ETA on supporting this - we will add a milestone when we have a timeline to share.

panchagnula:
@singlessis is working currently on this

btardif:
@panchagnula @singlessis  any updates on this one?

panchagnula:
> @panchagnula @singlessis any updates on this one?

Still in PR @btardif!

PR #15273, opened 2020-09-23
[AppService] az webapp config ssl import: Add functionality to import App Service Certificate to a web app, changing parameter names

### LINKED ISSUE #12427: webapp: Import App Service Certificate
Hello,

Let's assume I created an App Service Certificate with the Azure dedicated service, is it possible to import it in the private key certificates of an App Service with CLI, instead of using the Azure Portal?

Thank you

---
#### Document Details

* ID: 2a3f984d-0d48-e075-7e2f-2da5667b15cf
* Version Independent ID: 98e06d47-0ed4-4d8e-ccf5-d3a78e510539
* Content: [az webapp config ssl](https://docs.microsoft.com/en-us/cli/azure/webapp/config/ssl?view=azure-cli-latest#az-webapp-config-ssl-import)
* Content Source: [latest/docs-ref-autogen/webapp/config/ssl.yml](https://github.com/MicrosoftDocs/azure-docs-cli/blob/live/latest/docs-ref-autogen/webapp/config/ssl.yml)
* Service: **app-service-web**
* GitHub Login: @rloutlaw

singlessis:
**Description**  
I am enabling az webapp config ssl import to also import an app service certificate to a webapp. At the moment, it just allows importing a KeyVault cert. With this change, it can be used for both of these. The breaking change here is changing the parameters, making key_vault optional, and changing the name from key_vault_certificate_name to certificate_name
Fixes #12427, #13907

**Testing Guide**  
existing functionality is changed to:
az webapp config ssl import --resource-group rg --name webapp --certificate-name kvCertName --key-vault kvName
added functionality is: 
az webapp config ssl import --resource-group rg --name webapp --certificate-name appServiceCertificate

**History Notes**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/_help.py
@@ -329,12 +329,15 @@ helps['functionapp config ssl import'] = """
 type: command
 short-summary: Import a certificate to a functionapp. The certificate can be from Key Vault, or an App Service Certificate.
 examples:
-  - name: Import an SSL certificate to a function app from Key Vault. Option '--key-vault-certificate-name' has been deprecated, use '--certificate-name' instead
-    text: az functionapp config ssl import --resource-group MyResourceGroup --name MyFunctionApp --key-vault MyKeyVault --certificate-name MyCertificateName
-  - name: Import an SSL certificate to a function app from Key Vault using resource id (typically if Key Vault is in another subscription). Option '--key-vault-certificate-name' has been deprecated, use '--certificate-name' instead
-    text: az functionapp config ssl import --resource-group MyResourceGroup --name MyFunctionApp --key-vault '/subscriptions/[sub id]/resourceGroups/[rg]/providers/Microsoft.KeyVault/vaults/[vault name]' --certificate-name MyCertificateName
+  - name: Import an SSL certificate to a function app from Key Vault. Option '--key-vault-certificate-name' has been deprecated, use '--certificate' instead
+    text: az functionapp config ssl import --resource-group MyResourceGroup --name MyFunctionApp --key-vault MyKeyVault --certificate MyCertificateName
+  - name: Import an SSL certificate to a function app from Key Vault using resource id (typically if Key Vault is in another subscription). Option '--key-vault-certificate-name' has been deprecated, use '--certificate' instead
+    text: az functionapp config ssl import --resource-group MyResourceGroup --name MyFunctionApp --key-vault '/subscriptions/[sub id]/resourceGroups/[rg]/providers/Microsoft.KeyVault/vaults/[vault name]' --certificate MyCertificateName
   - name: Import an App Service Certificate to a function app
-    text: az functionapp config ssl import --resource-group MyResourceGroup --name MyFunctionApp --certificate-name MyCertificateName
+    text: az functionapp config ssl import --resource-group MyResourceGroup --name MyFunctionApp --certificate MyCertificateName
+  - name: Import an App Service Certificate to a function app using the certificate resource id (if the certificate is in a different subscription)
+    text: az functionapp config ssl import --resource-group MyResourceGroup --name MyFunctionApp --certificate '/subscriptions/[sub id]/resourceGroups/[rg]/providers/Microsoft.Web/certificates/[certificate name]'
+
 
 """
 
@@ -1303,12 +1306,14 @@ helps['webapp config ssl import'] = """
 type: command
 short-summary: Import a certificate to a web app. The certificate can be from Key Vault, or an App Service Certificate.
 examples:
-  - name: Import an SSL certificate to a web app from Key Vault. Option '--key-vault-certificate-name' has been deprecated, use '--certificate-name' instead
-    text: az webapp config ssl import --resource-group MyResourceGroup --name MyWebapp --key-vault MyKeyVault --certificate-name MyCertificateName
-  - name: Import an SSL certificate to a web app from Key Vault using resource id (typically if Key Vault is in another subscription). Option '--key-vault-certificate-name' has been deprecated, use '--certificate-name' instead
-    text: az webapp config ssl import --resource-group MyResourceGroup --name MyWebapp --key-vault '/subscriptions/[sub id]/resourceGroups/[rg]/providers/Microsoft.KeyVault/vaults/[vault name]' --certificate-name MyCertificateName
+  - name: Import an SSL certificate to a web app from Key Vault. Option '--key-vault-certificate-name' has been deprecated, use '--certificate' instead
+    text: az webapp config ssl import --resource-group MyResourceGroup --name MyWebapp --key-vault MyKeyVault --certificate MyCertificateName
+  - name: Import an SSL certificate to a web app from Key Vault using resource id (typically if Key Vault is in another subscription). Option '--key-vault-certificate-name' has been deprecated, use '--certificate' instead
+    text: az webapp config ssl import --resource-group MyResourceGroup --name MyWebapp --key-vault '/subscriptions/[sub id]/resourceGroups/[rg]/providers/Microsoft.KeyVault/vaults/[vault name]' --certificate MyCertificateName
   - name: Import an App Service Certificate to a web app
-    text: az webapp config ssl import --resource-group MyResourceGroup --name MyWebapp --certificate-name MyCertificateName
+    text: az webapp config ssl import --resource-group MyResourceGroup --name MyWebapp --certificate MyCertificateName
+  - name: Import an App Service Certificate to a web app using the certificate resource id (if the certificate is in a different subscription)
+    text: az webapp config ssl import --resource-group MyResourceGroup --name MyWebapp --certificate '/subscriptions/[sub id]/resourceGroups/[rg]/providers/Microsoft.Web/certificates/[certificate name]'
 
 """
 
--- a/src/azure-cli/azure/cli/command_modules/appservice/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/_help.py
@@ -1310,7 +1310,7 @@ examples:
     text: az webapp config ssl import --resource-group MyResourceGroup --name MyWebapp --key-vault MyKeyVault --key-vault-certificate-name MyCertificateName
   - name: Import an SSL certificate to a web app from Key Vault using resource id (typically if Key Vault is in another subscription).
     text: az webapp config ssl import --resource-group MyResourceGroup --name MyWebapp --key-vault '/subscriptions/[sub id]/resourceGroups/[rg]/providers/Microsoft.KeyVault/vaults/[vault name]' --certificate-name MyCertificateName
-    text: az webapp config ssl import --resource-group MyResourceGroup --name MyWebapp --key-vault '/subscriptions/[sub id]/resourceGroups/[rg]/providers/Microsoft.KeyVault/vaults/[vault name]' --key-vault-certificate-name MyCertificateName  
+    text: az webapp config ssl import --resource-group MyResourceGroup --name MyWebapp --key-vault '/subscriptions/[sub id]/resourceGroups/[rg]/providers/Microsoft.KeyVault/vaults/[vault name]' --key-vault-certificate-name MyCertificateName
   - name: Import an App Service Certificate to a web app
     text: az webapp config ssl import --resource-group MyResourceGroup --name MyWebapp --certificate-name MyCertificateName
 
--- a/src/azure-cli/azure/cli/command_modules/appservice/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/_help.py
@@ -1305,10 +1305,8 @@ short-summary: Import a certificate to a web app. The certificate can be from Ke
 examples:
   - name: Import an SSL certificate to a web app from Key Vault. Option '--key-vault-certificate-name' has been deprecated, use '--certificate-name' instead
     text: az webapp config ssl import --resource-group MyResourceGroup --name MyWebapp --key-vault MyKeyVault --certificate-name MyCertificateName
-    text: az webapp config ssl import --resource-group MyResourceGroup --name MyWebapp --key-vault MyKeyVault --key-vault-certificate-name MyCertificateName
   - name: Import an SSL certificate to a web app from Key Vault using resource id (typically if Key Vault is in another subscription). Option '--key-vault-certificate-name' has been deprecated, use '--certificate-name' instead
     text: az webapp config ssl import --resource-group MyResourceGroup --name MyWebapp --key-vault '/subscriptions/[sub id]/resourceGroups/[rg]/providers/Microsoft.KeyVault/vaults/[vault name]' --certificate-name MyCertificateName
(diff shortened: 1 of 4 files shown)


yonzhan:
AppService

yungezz:
hi @singlessis this is a breaking change. Normally we'll suggest add new parameter instead of renaming existing parameters. Just want to get your awareness on impact to existing users.

singlessis:
> hi @singlessis this is a breaking change. Normally we'll suggest add new parameter instead of renaming existing parameters. Just want to get your awareness on impact to existing users.

Hi! Yes, we are aware. We think it would be less confusing to share the parameter for name, is there much we have to do for this breaking change? 

Adding @panchagnula so she is in the loop

panchagnula:
> > hi @singlessis this is a breaking change. Normally we'll suggest add new parameter instead of renaming existing parameters. Just want to get your awareness on impact to existing users.
> 
> We're aware it is a breaking, change, but we think it would be less confusing to share the parameter for certificate name, is there much we have to do for this breaking change?
> 
> Adding @panchagnula so she is in the loop

@singlessis we should follow the breaking change protocol, I suggest we add this as a new parameter & mark the other one as being deprecated after release <>. @yungezz can also recommend the right protocol to follow here. Thanks!

src/azure-cli/azure/cli/command_modules/appservice/_params.py
```diff
@@ -227,8 +227,8 @@ def load_arguments(self, _):
             c.argument('settings', nargs='+', help="space-separated app settings in a format of `<name>=<value>`")
             c.argument('setting_names', nargs='+', help="space-separated app setting names")
         with self.argument_context(scope + ' config ssl import') as c:
-            c.argument('key_vault', help='The name or resource ID of the Key Vault')
-            c.argument('key_vault_certificate_name', help='The name of the certificate in Key Vault')
+            c.argument('certificate_name', help='The name of the certificate to be imported')
+            c.argument('key_vault', help='the keyvault id')
```
qwordy, line 231: Can it be a name?

qwordy (approved):
Be careful of breaking changes.

src/azure-cli/azure/cli/command_modules/appservice/_params.py
panchagnula, line 231: for Keyvault the value can be name of the AKV or the resourceID esp if they are different RGs or subscriptions. So please leave the help text as is.

src/azure-cli/azure/cli/command_modules/appservice/_params.py
```diff
@@ -227,8 +227,8 @@ def load_arguments(self, _):
             c.argument('settings', nargs='+', help="space-separated app settings in a format of `<name>=<value>`")
             c.argument('setting_names', nargs='+', help="space-separated app setting names")
         with self.argument_context(scope + ' config ssl import') as c:
-            c.argument('key_vault', help='The name or resource ID of the Key Vault')
-            c.argument('key_vault_certificate_name', help='The name of the certificate in Key Vault')
```
panchagnula, line 231: Don't remove this at this time. Keep it as is to avoid breaking changes.

Make the Certificate_name new parameter and update the help to say name of the certificate in the keyvalut to be imported or name of the app service certificate. For the old parameter follow the breaking change protocol to add an attribute which states that this will be deprecated in future releases and should be updated to use the certificate_name parameter.

panchagnula (changes requested):
Please work with the core CLI folks to handle the breaking changes as per protocol.

src/azure-cli/azure/cli/command_modules/appservice/_params.py
singlessis: In this case, name is used for the webapp name.

src/azure-cli/azure/cli/command_modules/appservice/_params.py
singlessis: I need to chose a release version to put into the documentation, what version do we want?

panchagnula: what documentation you are referring to? the release documentation gets auto generated with the title of the PR

singlessis: discussed offline, updated deprecation version to 2.15.0

src/azure-cli/azure/cli/command_modules/appservice/_params.py
singlessis: @qwordy could you take another look, I made some changes to avoid it being breaking

singlessis:
@Juliehzl could you review this?

src/azure-cli/azure/cli/command_modules/appservice/_params.py
```diff
@@ -226,8 +226,9 @@ def load_arguments(self, _):
             c.argument('settings', nargs='+', help="space-separated app settings in a format of `<name>=<value>`")
             c.argument('setting_names', nargs='+', help="space-separated app setting names")
         with self.argument_context(scope + ' config ssl import') as c:
+            c.argument('certificate_name', help='The name of the certificate to be imported')
+            c.argument('key_vault_certificate_name', help='The name of the certificate in Key Vault', deprecate_info=c.deprecate(expiration='2.15.0'))
```
panchagnula, line 230: Can we update the help to include to use the new argument moving forward instead of the key_vault_certificate_name? @qwordy can you tell us the recommended approach here?

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2354,39 +2355,61 @@ def delete_ssl_cert(cmd, resource_group_name, certificate_thumbprint):
...
-    if not is_valid_resource_id(key_vault):
-        kv_client = get_mgmt_service_client(cmd.cli_ctx, ResourceType.MGMT_KEYVAULT)
-        key_vaults = kv_client.vaults.list_by_subscription()
-        for kv in key_vaults:
-            if key_vault == kv.name:
-                kv_id = kv.id
-                break
-    else:
-        kv_id = key_vault
+
```
panchagnula, line 2367: if a user is using key vault_certificate_name can we show logger.warning to say we are deprecating & to use certificate_name instead?

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2354,39 +2355,61 @@ def delete_ssl_cert(cmd, resource_group_name, certificate_thumbprint):
...
+    if not key_vault_certificate_name and not certificate_name:
+        name_needed_msg = 'You must provide a certificate name, use either: ' \
```
panchagnula, line 2369: Use certificate_name instead only

src/azure-cli/azure/cli/command_modules/appservice/_params.py
qwordy: https://github.com/Azure/azure-cli/blob/a990e9626392e1f2bf8da0e8afefd24c8a518f48/doc/authoring_command_modules/authoring_commands.md#deprecating-commands-and-arguments
Use `redirect=` redirect users to the new argument

panchagnula: @singlessis please follow the documentation for this breaking change.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2354,39 +2355,61 @@ def delete_ssl_cert(cmd, resource_group_name, certificate_thumbprint):
...
+                          'certificate-name or key-vault-certificate-name.'
+        logger.warning(name_needed_msg)
+        return
+
+    if key_vault_certificate_name:
```
panchagnula, line 2374: Not sure if this validation is necessary.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2354,39 +2355,61 @@ def delete_ssl_cert(cmd, resource_group_name, certificate_thumbprint):
...
+                    key_vault=None):
     Certificate = cmd.get_models('Certificate')
     client = web_client_factory(cmd.cli_ctx)
     webapp = client.web_apps.get(resource_group_name, name)
     if not webapp:
         raise CLIError("'{}' app doesn't exist in resource group {}".format(name, resource_group_name))
     server_farm_id = webapp.server_farm_id
     location = webapp.location
-    kv_id = None
-    if not is_valid_resource_id(key_vault):
```
panchagnula, line 2366: Why are we deleting this?

panchagnula (changes requested):
follow the breaking changes documentation to replace with new parameter.

panchagnula:
> follow the breaking changes documentation to replace with new parameter.

@singlessis any updates on this PR?

src/azure-cli/azure/cli/command_modules/appservice/custom.py
singlessis: moved to helper function get_import_kv_details

src/azure-cli/azure/cli/command_modules/appservice/_help.py
```diff
@@ -327,12 +327,17 @@
 
 helps['functionapp config ssl import'] = """
 type: command
-short-summary: Import an SSL certificate to a function app from Key Vault.
+short-summary: Import a certificate to a functionapp. The certificate can be from Key Vault, or an App Service Certificate.
 examples:
-  - name: Import an SSL certificate to a function app from Key Vault.
+  - name: Import an SSL certificate to a function app from Key Vault. Option '--key-vault-certificate-name' has been deprecated, use '--certificate-name' instead
```
panchagnula, line 332: the word SSL certificate was removed above but here we are still using that. Any reason for this difference?

panchagnula: @singlessis can you run az webapp config ssl import --help & share how the results look with your updates?

src/azure-cli/azure/cli/command_modules/appservice/_params.py
```diff
@@ -227,8 +227,9 @@ def load_arguments(self, _):
             c.argument('settings', nargs='+', help="space-separated app settings in a format of `<name>=<value>`")
             c.argument('setting_names', nargs='+', help="space-separated app setting names")
         with self.argument_context(scope + ' config ssl import') as c:
+            c.argument('certificate_name', help='The name of the certificate to be imported')
+            c.argument('key_vault_certificate_name', help='The name of the certificate in Key Vault', deprecate_info=c.deprecate(expiration='2.15.0', target='--key-vault-certificate-name', redirect='--certificate-name'))
```
panchagnula, line 231: 2.15 already released I would make the 2.18.0 just to be safe. Thanks!

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2382,39 +2383,51 @@ def delete_ssl_cert(cmd, resource_group_name, certificate_thumbprint):
...
-        kv_client = get_mgmt_service_client(cmd.cli_ctx, ResourceType.MGMT_KEYVAULT)
-        key_vaults = kv_client.vaults.list_by_subscription()
-        for kv in key_vaults:
-            if key_vault == kv.name:
-                kv_id = kv.id
-                break
+
+    if key_vault_certificate_name:
```
panchagnula, line 2397: Don't understand this part yet? why are we returning here?

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2382,39 +2383,51 @@ def delete_ssl_cert(cmd, resource_group_name, certificate_thumbprint):
...
-                 'vaults/{0}'.format(key_vault, name, resource_group_name, key_vault_certificate_name)
+                 'vaults/{0}'.format(key_vault, name, resource_group_name, cert)
         logger.warning(kv_msg)
         return
 
     kv_id_parts = parse_resource_id(kv_id)
     kv_name = kv_id_parts['name']
     kv_resource_group_name = kv_id_parts['resource_group']
     kv_subscription = kv_id_parts['subscription']
-    cert_name = '{}-{}-{}'.format(resource_group_name, kv_name, key_vault_certificate_name)
```
panchagnula, line 2417: we should still let users use key_vault_certificate_name.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2382,39 +2383,51 @@ def delete_ssl_cert(cmd, resource_group_name, certificate_thumbprint):
...
+    if not key_vault:
```
panchagnula, line 2426: What is the purpose of this?

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2423,12 +2436,48 @@ def import_ssl_cert(cmd, resource_group_name, name, key_vault, key_vault_certifi
...
     return client.certificates.create_or_update(name=cert_name, resource_group_name=resource_group_name,
                                                 certificate_envelope=kv_cert_def)
 
 
+def get_import_kv_details(cmd, client, certificate_name, key_vault):
+    kv_id = None
+    kv_secret_name = certificate_name
+    # if key_vault is not populated, we are attempting to import an App Service Certificate
+    if not key_vault:
+        certs = client.app_service_certificate_orders.list()
```
panchagnula, line 2450: we are not passing any subscription or other parameters here - so what does this list do?

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2382,39 +2383,51 @@ def delete_ssl_cert(cmd, resource_group_name, certificate_thumbprint):
     raise CLIError("Certificate for thumbprint '{}' not found".format(certificate_thumbprint))
 
 
-def import_ssl_cert(cmd, resource_group_name, name, key_vault, key_vault_certificate_name):
+def import_ssl_cert(cmd, resource_group_name, name, certificate_name=None, key_vault_certificate_name=None,
```
panchagnula, line 2386: if this is the only parameter we expect for ASC - then this can ASC resource_id as well since ASC can be in a different subscription than the subscription the app is in.

panchagnula (changes requested):
A few more comments.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
singlessis: I must have misunderstood some comments, I disabled use of key_vault-certificate-name, let me add it back

src/azure-cli/azure/cli/command_modules/appservice/custom.py
singlessis: What exactly? the random int? we decided that it would be best to differentiate the certificates without using webspace, like portal does. Is there another naming convention I should use?

src/azure-cli/azure/cli/command_modules/appservice/custom.py
singlessis: okay, makes sense, I'll change the parameter, i will need to test this

src/azure-cli/azure/cli/command_modules/appservice/custom.py
singlessis: Based on your comment below, this is going to cause a few issues, we might have to make a new client if the ASC is in a different susbcription, will take a look

panchagnula:
@singlessis is this ready for review with all the new updates?

panchagnula:
@singlessis closing this due to inactivity please start new PR after addressing all the feedback. Thanks!

Issue #14627, opened 2020-08-03
Error text verbiage while creating a new cosmosdb using Azure CLI


avanishkdev:
I was trying to create a azure comosdb using CLI. Got it error a few times , but was not able to identify what the issue was. Syntactically looked like the key value pairs in locations is good. After reviewing it several times identified that it was not the format thats incorrect but  one of the property name itself was incorrect,

![image](https://user-images.githubusercontent.com/66087777/89199003-f79a4e00-d57b-11ea-9977-4b993e13c3d1.png)

Would it be possible to display a different error message for this , something like  "Unknown property name {propertyname}". It would have been easier to debug.

![image](https://user-images.githubusercontent.com/66087777/89199182-30d2be00-d57c-11ea-8bef-218c40e8b199.png)

yonzhan:
cosmosdb

MehaKaushik:
Hey @avanishkdev thanks for your feedback. We will investigate further and decide if that is possible. I will update here in sometime

MehaKaushik:
Hi @yonzhan do you know if the CLI team is aware of any in-progress feature development which can potentially address this ASK. This issue is not service-related. It is CLI infra issue.

yonzhan:
@jiasli to help.

jiasli:
This `--locations` is not the global parameter `--location`, so this issue is not related to CLI infrastructure.

https://github.com/Azure/azure-cli/blob/4aee466f6fce5db00db61639545e98712e6263b6/src/azure-cli/azure/cli/command_modules/cosmosdb/_params.py#L54-L57

And the error `usage error: --locations [KEY=VALUE ...]` is thrown at

https://github.com/Azure/azure-cli/blob/17015379924a13b12a48781469c8b0e5021c7c2a/src/azure-cli/azure/cli/command_modules/cosmosdb/actions.py#L40-L43

\+ the author @shurd

shurd:
The improved error messaging is in PR, and we are working on getting it merged.

shurd:
We are still waiting for the PR to be reviewed, but it should be in the next CLI release.

shurd:
Fixed in 2.14.0

PR #15276, opened 2020-09-24
[Cosmos DB] az cosmosdb create/update: Improve error message from incorrect --locations input

shurd:
**Description**  
Updates the error message user sees when supplying invalid key name for the "locations" parameter during update or create. Helps address issue https://github.com/Azure/azure-cli/issues/14627.

Fix bug preventing removing ip range filter via command line (handle empty string). Helps address issue https://github.com/Azure/azure-cli/issues/15291.

Add analytical storage ttl support for SQL containers.

Fixed typo in help text to address https://github.com/Azure/azure-cli/issues/15524

**Testing Guide**  
az cosmosdb update -g {g} -n {n} --locations regionName=EastUS failoverPriorit=0
az cosmosdb update -g {g} -n {n} --ip-range-filter ""

**History Notes**  

[Cosmos DB] az cosmosdb create/update: Improve error message from incorrect --locations input
[Cosmos DB] az cosmosdb sql container create/update: Add --analytical-storage-ttl parameter

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/cosmosdb/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/cosmosdb/_params.py
@@ -154,6 +154,7 @@ def load_arguments(self, _):
         c.argument('conflict_resolution_policy', options_list=['--conflict-resolution-policy', '-c'], type=shell_safe_json_parse, completer=FilesCompleter(), help='Conflict Resolution Policy, you can enter it as a string or as a file, e.g., --conflict-resolution-policy @policy-file.json or ' + SQL_GREMLIN_CONFLICT_RESOLUTION_POLICY_EXAMPLE)
         c.argument('max_throughput', max_throughput_type)
         c.argument('throughput', help='The throughput of SQL container (RU/s). Default value is 400. Omit this parameter if the database has shared throughput unless the container should have dedicated throughput.')
+        c.argument('analytical_storage_ttl', type=int, help='Analytical TTL, when analytical storage is enabled.')
 
 # SQL stored procedure
     with self.argument_context('cosmosdb sql stored-procedure') as c:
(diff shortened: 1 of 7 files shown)


yonzhan:
Cosmos

shurd:
@Juliehzl can you please take a look?

yonzhan:
@Juliehzl will review this PR after Chinese National holiday.

src/azure-cli/azure/cli/command_modules/cosmosdb/_params.py
```diff
@@ -154,6 +154,7 @@ def load_arguments(self, _):
         c.argument('conflict_resolution_policy', options_list=['--conflict-resolution-policy', '-c'], type=shell_safe_json_parse, completer=FilesCompleter(), help='Conflict Resolution Policy, you can enter it as a string or as a file, e.g., --conflict-resolution-policy @policy-file.json or ' + SQL_GREMLIN_CONFLICT_RESOLUTION_POLICY_EXAMPLE)
         c.argument('max_throughput', max_throughput_type)
         c.argument('throughput', help='The throughput of SQL container (RU/s). Default value is 400. Omit this parameter if the database has shared throughput unless the container should have dedicated throughput.')
+        c.argument('analytical_storage_ttl', type=int, help='Analytical TTL, when analytical storage is enabled.')
```
Juliehzl, line 157: Is there a default value in server side?

Juliehzl:
@shurd could you take a look for CI failure?

src/azure-cli/azure/cli/command_modules/cosmosdb/_params.py
shurd: The default value is None/null I believe

shurd:
@Juliehzl The CI has two failures. One is saying the parameter "analytical_storage_ttl" is too long for the commands I just added it on; however, this already exists for several other commands (mongodb collection create/update). Is there a way to suppress this error? Why is CI not marking the other places as problematic.

The other CI error is saying that "cosmosdb collection/database" commands have resource-group-name as a parameter, which is not allowed. These commands are all deprecated, so can we suppress this error?

Juliehzl:
> @Juliehzl Zunli Hu FTE The CI has two failures. One is saying the parameter "analytical_storage_ttl" is too long for the commands I just added it on; however, this already exists for several other commands (mongodb collection create/update). Is there a way to suppress this error? Why is CI not marking the other places as problematic.
> 
> The other CI error is saying that "cosmosdb collection/database" commands have resource-group-name as a parameter, which is not allowed. These commands are all deprecated, so can we suppress this error?

For the first error, you could add options_list=['--analytical-storage-ttl', '-t'] to bypass it, which means you need to add one more short option for the parameter.

Juliehzl:
The second one is not blocking issue. You could not change  anything for it. But it would be better if we could have an ETA to remove these deprecated commands.

Juliehzl (approved):

PR #15290, opened 2020-09-24
[SQL] Adding AAD-only Support for SQL Managed Instances and Servers

jaredwelsh:
Adding support for a new feature to az cli. Azure Active Directory only allows for only AD users to be admins not traditional username/password combinations. This feature has been added to Azure Python SDK and is now being supported in az cli

Using either a managed instance or server with a ad admin run:

az cli sql server ad-only-auth enable
az cli sql server ad-only-auth disable
az cli sql server ad-only-auth get

az cli sql mi ad-only-auth enable
az cli sql mi ad-only-auth disable
az cli sql mi ad-only-auth get

to enable, disable and show AAD only authentication.

### DIFF
--- a/azure-cli2017.pyproj
+++ b/azure-cli2017.pyproj
@@ -1466,15 +1466,6 @@
     <Content Include="azure-cli\setup.cfg" />
   </ItemGroup>
   <ItemGroup>
-    <Interpreter Include="..\..\env2\">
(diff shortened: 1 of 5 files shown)


jaredwelsh:
mistakenly submitted before linking microsoft account.

PR #15308, opened 2020-09-25
{Storage} Fix #15052: show warning according to command arguments

### LINKED ISSUE #15052: 'az storage table create' doesn't have an '--auth-mode' parameter, but mentions it in a warning
## Describe the bug
When running `az storage table create` without an account key, connection string or sas token, the warning message you get is:
```
There is no credential provided in your command and environment, we will query account key for your storage account.
Please provide --connection-string, --account-key or --sas-token as credential, or use `--auth-mode login` if you have required RBAC roles in your command. For more information about RBAC roles in storage, you can see https://docs.microsoft.com/en-us/azure/storage/common/storage-auth-aad-rbac-cli.
Setting corresponding environment variable can avoid inputting credential in your command. Please use --help to get more information.
```
This command succeeds, as it uses the user's credentials. Unfortunately this command doesn't take have an `--auth-mode` parameter, so if you're using that authentication method you are stuck with this warning.

**Command Name**
`az storage queue create`

## To Reproduce:

- `az storage queue create --name {} --account-name {}`

## Expected Behavior
Either the command should have an `--auth-mode` parameter, or there should be no warning for using RBAC.

## Environment Summary
```
Windows-10-10.0.19041-SP0
Python 3.6.8
Installer: MSI

azure-cli 2.10.1 *

```

evelyn-ys:
## Description
Fix #15052 : `az storage table create` doesn't have an `--auth-mode` parameter, but mentions it in a warning
This PR customize warning message according to the arguments each command includes

## Testing Guide
`az storage table list --account-name $ACCOUNT`
 
warning before:
```sh
> az storage table list --account-name xxxx
There are no credentials provided in your command and environment, we will query for the account key inside your storage account.
Please provide --connection-string, --account-key or --sas-token as credentials, or use `--auth-mode login` if you have required RBAC roles in your command. For more information about RBAC roles in storage, visit https://docs.microsoft.com/en-us/azure/storage/common/storage-auth-aad-rbac-cli.
Setting the corresponding environment variables can avoid inputting credentials in your command. Please use --help to get more information.
[]
```
warning after:
```sh
> az storage table list --account-name yssa
There are no credentials provided in your command and environment, we will query for the account key inside your storage account.
Please provide --connection-string or --account-key or --sas-token as credentials.
Setting the corresponding environment variables can avoid inputting credentials in your command. Please use --help to get more information.
[]
```

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/storage/_validators.py
+++ b/src/azure-cli/azure/cli/command_modules/storage/_validators.py
@@ -65,8 +65,30 @@ def _create_token_credential(cli_ctx):
 
     def _cancel_timer_event_handler(_, **__):
         updater.cancel()
+
     cli_ctx.register_event(EVENT_CLI_POST_EXECUTE, _cancel_timer_event_handler)
     return token_credential
+
+
+def _show_no_credentials_warning(namespace, auth_mode_attr=None):
+    credentials_choices = [ca for ca in ['--connection-string', '--account-key', '--sas-token']
+                           if hasattr(namespace, ca[2:].replace('-', '_'))]
+    logger.warning('There are no credentials provided in your command and environment, we will query for the '
+                   'account key inside your storage account. ')
+    if credentials_choices:
+        credentials_choices_str = ' or '.join(credentials_choices)
+        auth_mode_log = '.'
+        if auth_mode_attr:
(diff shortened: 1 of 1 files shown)


yonzhan:
Storage

Issue #15314, opened 2020-09-27
Fix non-actionable code bugs: Core


houk-ms:
This issue is used to track the process of fixing non-actionable errors in CLI core.

houk-ms:
**Error message is suppressed.**

> az adp data-pool create --account-name adptest23 --data-pool-name dp2 -g rptest2 --subscription adp_dev_playground1 --locations name=westeurope --debug
Command arguments: ['adp', 'data-pool', 'create', '--account-name', 'adptest23', '--data-pool-name', 'dp2', '-g', 'rptest2', '--subscription', 'adp_dev_playground1', '--locations', 'name=westeurope', '--debug']
Event: Cli.PreExecute []
Event: CommandParser.OnGlobalArgumentsCreate [<function CLILogging.on_global_arguments at 0x03E8B8E8>, <function OutputProducer.on_global_arguments at 0x0401AD68>, <function CLIQuery.on_global_arguments at 0x0403FD68>]
Event: CommandInvoker.OnPreCommandTableCreate []
Modules found from index for 'adp': ['azext_adp']
Loading command modules:
Name                  Load Time    Groups  Commands
Total (0)                 0.000         0         0
These extensions are not installed and will be skipped: ['azext_ai_examples', 'azext_ai_did_you_mean_this']
Loading extensions:
Name                  Load Time    Groups  Commands  Directory
adp                       0.008         3        12  C:\Users\elarbel\.azure\cliextensions\adp
Total (1)                 0.008         3        12
Loaded 3 groups, 12 commands.
Found a match in the command table.
Raw command  : adp data-pool create
Command table: adp data-pool create
Event: CommandInvoker.OnPreCommandTableTruncate [<function AzCliLogging.init_command_file_logging at 0x04264B28>]
az_command_data_logger : command args: adp data-pool create --account-name {} --data-pool-name {} -g {} --subscription {} --locations {} --debug
metadata file logging enabled - writing logs to 'C:\Users\elarbel\.azure\commands'.
Event: CommandInvoker.OnPreArgumentLoad [<function register_global_subscription_argument.<locals>.add_subscription_parameter at 0x042B93D8>, <function register_global_query_examples_argument.<locals>.register_query_examples at 0x042B9B70>]
Event: CommandInvoker.OnPostArgumentLoad []
[...]

PR #15319, opened 2020-09-27
{Core} Define new error types

houk-ms:
**Description**  

This PR provides
- New error types for error improvement. A series of error classes are defined to replace the original enumerated error types.
- Adjustments for exception handler in CLI core. The original raw exceptions from different sources are wrapped by the new defiined error types.

Base class: `AzCLIError`
Main-categories: `UserFault`, `ClientError`, `ServiceError`
Sub-categories:
```
# Command related error types
class CommandNotFoundError(UserFault):
    """ Command is misspelled or not recognized by AzureCLI. """
    pass


# Argument related error types
class UnrecognizedArgumentError(UserFault):
    """ Argument is misspelled or not recognized by AzureCLI. """
    pass


class RequiredArgumentMissingError(UserFault):
    """ Required argument is not specified. """
    pass


class ConflictArgumentError(UserFault):
    """ Arguments can not be specfied together. """
    pass


class InvalidArgumentValueError(UserFault):
    """ Argument value is not valid. """
    pass


class ArgumentParseError(UserFault):
    """ Fallback of the argument parsing related errors.
    Use this error only when the argument related errors
    can not be classified into the above error types. """
    pass


# Response related error types
class BadRequestError(UserFault):
    """ Bad request from client: 400 error """
    pass


class UnauthorizedError(UserFault):
    """ Unauthorized request: 401 error """


class ForbiddenError(UserFault):
    """ Service refuse to response: 403 error """


class ResourceNotFoundError(UserFault):
    """ Can not find Azure resources: 404 error """
    pass


class AzureInternalError(ServiceError):
    """ Azure service internal error: 5xx error """
    pass


class AzureResponseError(UserFault):
    """ Fallback of the response related errors.
    Use this error only when the response related errors
    can not be classified into the above error types. """


# Request related error types
class AzureConnectionError(UserFault):
    """ Connection issues like connection timeout, aborted or broken. """
    pass


class ClientRequestError(UserFault):
    """ Fallback of the request related errors.
    Error occurs while attempting to make a request to the service. No request is sent.
    """


# Validation related error types
class ValidationError(UserFault):
    """ Fallback of the errors in validation functions.
    Use this error in validation functions only when the error can not be
    classified into the Argument, Request and Response related error types. """
    pass


# CLI internal error type
class CLIInternalError(ClientError):
    """ Azure CLI internal error """
    pass


# Keyboard interrupt error type
class ManualInterrupt(UserFault):
    """ Keyboard interrupt. """
    pass
```

### DIFF
--- a/src/azure-cli-core/azure/cli/core/azclierror.py
+++ b/src/azure-cli-core/azure/cli/core/azclierror.py
@@ -4,85 +4,137 @@
 # --------------------------------------------------------------------------------------------
 
 import sys
-from enum import Enum
 
+import azure.cli.core.telemetry as telemetry
 from knack.util import CLIError
 from knack.log import get_logger
 
 logger = get_logger(__name__)
+# pylint: disable=unnecessary-pass
 
 
-class AzCLIErrorType(Enum):
-    """ AzureCLI error types """
-
-    # userfaults
-    CommandNotFoundError = 'CommandNotFoundError'
-    ArgumentParseError = 'ArgumentParseError'
-    ValidationError = 'ValidationError'
-    ManualInterrupt = 'ManualInterrupt'
-    # service side error
-    ServiceError = 'ServiceError'
-    # client side error
-    ClientError = 'ClientError'
-    # unexpected error
-    UnexpectedError = 'UnexpectedError'
-
-
+# region: Base Layer
+# Base class for all the AzureCLI defined error classes.
+# DO NOT raise the error class here directly in your codes.
 class AzCLIError(CLIError):
-    """ AzureCLI error definition """
+    """ Base class for all the AzureCLI defined error classes. """
 
-    def __init__(self, error_type, error_msg, raw_exception=None, command=None):
-        """
-        :param error_type: The name of the AzureCLI error type.
-        :type error_type: azure.cli.core.util.AzCLIErrorType
-        :param error_msg: The error message detail.
-        :type error_msg: str
-        :param raw_exception: The raw exception.
-        :type raw_exception: Exception
-        :param command: The command which brings the error.
-        :type command: str
-        :param recommendations: The recommendations to resolve the error.
-        :type recommendations: list
-        """
-        self.error_type = error_type
+    def __init__(self, error_msg):
+        # error message
         self.error_msg = error_msg
-        self.raw_exception = raw_exception
-        self.command = command
+        # set recommendations to fix the error if the message is not actionable,
+        # they will be printed to users after the error message, one recommendation per line
         self.recommendations = []
+        # exception trace for the error
+        self.exception_trace = None
         super().__init__(error_msg)
 
     def set_recommendation(self, recommendation):
         self.recommendations.append(recommendation)
 
-    def set_raw_exception(self, raw_exception):
-        self.raw_exception = raw_exception
+    def set_exception_trace(self, exception_trace):
+        self.exception_trace = exception_trace
 
     def print_error(self):
         from azure.cli.core.azlogging import CommandLoggerContext
         with CommandLoggerContext(logger):
-            message = '{}: {}'.format(self.error_type.value, self.error_msg)
+            # print error type and error message
+            message = '{}: {}'.format(self.__class__.__name__, self.error_msg)
             logger.error(message)
-            if self.raw_exception:
-                logger.exception(self.raw_exception)
+            # print exception trace if there is
+            if self.exception_trace:
+                logger.exception(self.exception_trace)
+            # print recommendations to action
             if self.recommendations:
                 for recommendation in self.recommendations:
                     print(recommendation, file=sys.stderr)
 
     def send_telemetry(self):
-        import azure.cli.core.telemetry as telemetry
-        telemetry.set_error_type(self.error_type.value)
+        telemetry.set_error_type(self.__class__.__name__)
+# endregion
 
-        # For userfaults
-        if self.error_type in [AzCLIErrorType.CommandNotFoundError,
-                               AzCLIErrorType.ArgumentParseError,
-                               AzCLIErrorType.ValidationError,
-                               AzCLIErrorType.ManualInterrupt]:
-            telemetry.set_user_fault(self.error_msg)
 
-        # For failures: service side error, client side error, unexpected error
-        else:
-            telemetry.set_failure(self.error_msg)
+# region: Second Layer
+# Main categories of the AzureCLI error types, used for Telemetry alalysis
+class UserFault(AzCLIError):
+    """ Users should be responsible for the errors. """
+    def send_telemetry(self):
+        super().send_telemetry()
+        telemetry.set_user_fault(self.error_msg)
 
-        # For unexpected error
-        if self.raw_exception:
-            telemetry.set_exception(self.raw_exception, '')
+
+class AzureInternalError(AzCLIError):
+    """ Azure Services should be responsible for the errors.  """
+    def send_telemetry(self):
+        super().send_telemetry()
+        telemetry.set_failure(self.error_msg)
+
+
+class CLIInternalError(AzCLIError):
+    """ AzureCLI should be responsible for the errors. """
+    def send_telemetry(self):
+        super().send_telemetry()
+        telemetry.set_failure(self.error_msg)
+        if self.exception_trace:
+            telemetry.set_exception(self.exception_trace, '')
+# endregion
+
+
+# region: Third Layer
+# Sub-categories of the AzureCLI error types, shown to users
+class CommandNotFoundError(UserFault):
+    """ Command is misspelled or not recognized by AzureCLI. """
+    pass
+
+
+class UnrecognizedArgumentError(UserFault):
+    """ Argument is misspelled or not recognized by AzureCLI. """
+    pass
+
+
+class RequiredArgumentMissingError(UserFault):
+    """ Required argument is not specified. """
+    pass
+
+
+class ConflictArgumentError(UserFault):
+    """ Arguments can not be specfied together. """
+    pass
+
+
+class InvalidArgumentValueError(UserFault):
+    """ Argument value is not valid. """
+    pass
+
+
+class ArgumentParseError(UserFault):
+    """ Fallback of the argument parsing related errors.
+    Use this if the argument parsing related errors can not be classified into the above types. """
+    pass
+
+
+class ResourceNotFoundError(UserFault):
+    """ Can not find user specified resource: 404 error """
+    pass
+
+
+class AzureConnectionError(UserFault):
+    """ Connection issues like timeout, aborted or broken. """
+    pass
+
+
+class BadRequestError(UserFault):
+    """ Bad request: 400 error """
+    pass
+
+
+class ValidationError(UserFault):
+    """ Fallback of the errors found in validators.
+    Use this if the validation related errors can not be classified into the above types. """
+    pass
+
+
+class ManualInterrupt(UserFault):
+    """ Keyboard interrupt. """
+    pass
+# endregion
--- a/src/azure-cli-core/azure/cli/core/azclierror.py
+++ b/src/azure-cli-core/azure/cli/core/azclierror.py
@@ -56,6 +56,7 @@ class AzCLIError(CLIError):
 
 # region: Second Layer
 # Main categories of the AzureCLI error types, used for Telemetry alalysis
+# DO NOT raise the error classes here directly in your codes.
 class UserFault(AzCLIError):
     """ Users should be responsible for the errors. """
     def send_telemetry(self):
@@ -63,15 +64,15 @@ class UserFault(AzCLIError):
         telemetry.set_user_fault(self.error_msg)
 
 
-class AzureInternalError(AzCLIError):
+class ServiceError(AzCLIError):
     """ Azure Services should be responsible for the errors.  """
     def send_telemetry(self):
         super().send_telemetry()
         telemetry.set_failure(self.error_msg)
 
 
-class CLIInternalError(AzCLIError):
-    """ AzureCLI should be responsible for the errors. """
+class ClientError(AzCLIError):
+    """ Azure CLI should be responsible for the errors. """
     def send_telemetry(self):
         super().send_telemetry()
         telemetry.set_failure(self.error_msg)
@@ -82,11 +83,14 @@ class CLIInternalError(AzCLIError):
 
 # region: Third Layer
 # Sub-categories of the AzureCLI error types, shown to users
+# Raise the error classes here in your codes
+# Command related error types
 class CommandNotFoundError(UserFault):
     """ Command is misspelled or not recognized by AzureCLI. """
     pass
 
 
+# Argument related error types
 class UnrecognizedArgumentError(UserFault):
     """ Argument is misspelled or not recognized by AzureCLI. """
     pass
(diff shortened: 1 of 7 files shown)


src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -74,78 +73,165 @@ def handle_exception(ex):  # pylint: disable=too-many-return-statements, too-man
...
         else:
             error_msg = "The command failed with an unexpected error. Here is the traceback:"
-            az_error = AzCLIError(AzCLIErrorType.UnexpectedError, error_msg)
-            az_error.set_raw_exception(ex)
+            az_error = azclierror.CLIInternalError(error_msg)
+            az_error.set_exception_trace(ex)
             az_error.set_recommendation("To open an issue, please run: 'az feedback'")
 
+        if isinstance(az_error, azclierror.ResourceNotFoundError):
+            exit_code = 3
```
houk-ms, line 143: Should exit_code be 3 for common resource-not-found cases? Or only for show command?

src/azure-cli-core/azure/cli/core/azclierror.py
```diff
@@ -4,85 +4,178 @@
...
+# Command related error types
+class CommandNotFoundError(UserFault):
+    """ Command is misspelled or not recognized by AzureCLI. """
+    pass
+
+
+# Argument related error types
+class UnrecognizedArgumentError(UserFault):
+    """ Argument is misspelled or not recognized by AzureCLI. """
+    pass
```
qwordy, line 96: Why many empty class definitions?

src/azure-cli-core/azure/cli/core/util.py
qwordy: The definition is ambiguous.
https://github.com/Azure/azure-cli#exit-codes
I think 3 is OK.

src/azure-cli-core/azure/cli/core/azclierror.py
```diff
@@ -4,85 +4,178 @@
...
+            telemetry.set_exception(self.exception_trace, '')
+# endregion
+
+
+# region: Third Layer
+# Sub-categories of the AzureCLI error types, shown to users
+# Raise the error classes here in your codes
+# Command related error types
+class CommandNotFoundError(UserFault):
+    """ Command is misspelled or not recognized by AzureCLI. """
```
qwordy, line 89:
```suggestion
    """ Command is misspelled or not recognized by Azure CLI. """
```

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -74,78 +73,165 @@ def handle_exception(ex):  # pylint: disable=too-many-return-statements, too-man
...
+        return azclierror.ResourceNotFoundError
+    if status_code.startswith('5'):
+        return azclierror.AzureInternalError
+
+    return azclierror.AzureResponseError
+
+
+def is_azure_connection_error(error_msg):
+    error_msg = error_msg.lower()
+    if 'connection error' in error_msg \
```
qwordy, line 221: Pattern matching is dangerous. Can we use error type to decide it?

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -74,78 +73,165 @@ def handle_exception(ex):  # pylint: disable=too-many-return-statements, too-man
...
+            or 'connection broken' in error_msg \
+            or 'connection aborted' in error_msg:
+        return True
+    return False
+
+
+def is_resource_not_found_error(error_msg):
+    error_msg = error_msg.lower()
+    if 'could not be found' in error_msg:
```
qwordy, line 230: Pattern matching is dangerous. Can we use error type to decide it?

qwordy (approved):

src/azure-cli-core/azure/cli/core/util.py
houk-ms: We cann't, the the error here is too general :). For those specific errors and those who have status_code, we already use status_code and error type to categorize it.

src/azure-cli-core/azure/cli/core/util.py
houk-ms: As above.

src/azure-cli-core/azure/cli/core/azclierror.py
```diff
@@ -4,85 +4,178 @@
...
+    def send_telemetry(self):
+        super().send_telemetry()
+        telemetry.set_user_fault(self.error_msg)
+
+
+class ServiceError(AzCLIError):
+    """ Azure Services should be responsible for the errors.  """
+    def send_telemetry(self):
+        super().send_telemetry()
+        telemetry.set_failure(self.error_msg)
```
yungezz, line 71: so in telemetry, service error and client error will be failure?

yungezz: when service error, no need to set exception trace?

src/azure-cli-core/azure/cli/core/azclierror.py
```diff
@@ -4,85 +4,178 @@
...
+    """ Argument is misspelled or not recognized by AzureCLI. """
+    pass
+
+
+class RequiredArgumentMissingError(UserFault):
+    """ Required argument is not specified. """
+    pass
+
+
+class ConflictArgumentError(UserFault):
```
yungezz, line 104: MutuallyExclusiveArgumentError

src/azure-cli-core/azure/cli/core/azclierror.py
```diff
@@ -4,85 +4,178 @@
...
+    can not be classified into the above error types. """
+
+
+# Request related error types
+class AzureConnectionError(UserFault):
+    """ Connection issues like connection timeout, aborted or broken. """
+    pass
+
+
+class ClientRequestError(UserFault):
```
yungezz, line 157: Little ambiguous, does this mean 400 badrequest? or userFault? or Client CLI error?

src/azure-cli-core/azure/cli/core/parser.py
```diff
@@ -168,7 +172,14 @@ def error(self, message):
         recommender.set_help_examples(self.get_examples(self.prog))
         recommendation = recommender.recommend_a_command()
 
-        az_error = AzCLIError(AzCLIErrorType.ArgumentParseError, message, command=self.prog)
+        az_error = ArgumentParseError(message)
```
yungezz, line 175: it seems less reliable to determine error type by text matching in message. Is there way to force message format?

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -74,78 +73,165 @@ def handle_exception(ex):  # pylint: disable=too-many-return-statements, too-man
...
-
-            except (ValueError, KeyError):
-                pass
-
-            az_error = AzCLIError(AzCLIErrorType.ServiceError, error_msg)
+            if extract_http_operation_error_message(ex):
+                error_msg = extract_http_operation_error_message(ex)
+            # by HttpOperationError definition, the message should be with the
+            # format: '({code}) {message}' if there is a status code
+            status_code = ex.__str__().split(')')[0].split('(')[-1]
```
yungezz, line 128: let's be defensive on this. :)

src/azure-cli-core/azure/cli/core/util.py
yungezz: what's the original exit code?

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -74,78 +73,165 @@ def handle_exception(ex):  # pylint: disable=too-many-return-statements, too-man
...
+    try:
+        response = json.loads(ex.response.text)
+        if isinstance(response, str):
+            error = response
+        else:
+            error = response['error']
+        # ARM should use ODATA v4. So should try this first.
+        # http://docs.oasis-open.org/odata/odata-json-format/v4.0/os/odata-json-format-v4.0-os.html#_Toc372793091
+        if isinstance(error, dict):
+            code = "{} - ".format(error.get('code', 'Unknown Code'))
```
yungezz, line 173: 👍

src/azure-cli-core/azure/cli/core/util.py
fengzhou-msft: 3 is good, it's not limited to show command.

src/azure-cli-core/azure/cli/core/util.py
yungezz: I see. So this will be a long maintaining work to add more pattern found in azure response :)

src/azure-cli-core/azure/cli/core/azclierror.py
houk-ms: > so in telemetry, service error and client error will be failure?

yes

houk-ms: > when service error, no need to set exception trace?

yes, exception trace is only set when the error is not caught. If an error was caught, that means we know what happens there,  and I think there is no need to set exception trace any more.

yungezz: is there easy way to distinguish them, such as one field? otherwise we'll need to maintain a list of error types which are client, which are service.

src/azure-cli-core/azure/cli/core/util.py
houk-ms: If there is a different format, the statement will just return the ex.__str__(). It will not break the logic here, while i have to admit maybe there is a better way.

src/azure-cli-core/azure/cli/core/azclierror.py
houk-ms: sure, we can use set_client_error() and set_service_error() to replace set_failure, which should make the result be one of ['Success', 'UserFault', 'ClientError', 'ServiceError'].

src/azure-cli-core/azure/cli/core/util.py
houk-ms: exactly

src/azure-cli-core/azure/cli/core/parser.py
houk-ms: Agree. I have beening considering if there is a better way here, too~

yungezz: can we force some message template at where error throw out?

src/azure-cli-core/azure/cli/core/util.py
houk-ms: The original one is magic, it even returns 'CloudError' as an exit_code and even check it in testing file [here](https://github.com/Azure/azure-cli/pull/15319/files#diff-96f657597aa99f9b338dc445b0d45590L429-L439). Could not understand :(

src/azure-cli-core/azure/cli/core/azclierror.py
houk-ms: We can do some customiztions if necesary later, but now i think it's enough

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -74,78 +73,165 @@ def handle_exception(ex):  # pylint: disable=too-many-return-statements, too-man
...
+    if status_code == '401':
+        return azclierror.UnauthorizedError
+    if status_code == '403':
+        return azclierror.ForbiddenError
+    if status_code == '404':
+        return azclierror.ResourceNotFoundError
+    if status_code.startswith('5'):
+        return azclierror.AzureInternalError
+
+    return azclierror.AzureResponseError
```
fengzhou-msft, line 216: Any tests for server side error?

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -74,78 +73,165 @@ def handle_exception(ex):  # pylint: disable=too-many-return-statements, too-man
         error_msg = getattr(ex, 'message', str(ex))
         exit_code = 1
 
-        if isinstance(ex, AzCLIError):
+        if isinstance(ex, azclierror.AzCLIError):
```
jiasli, line 76: I still don't think `CLIError` should be categorized as `ValidationError`.

For example, if I `raise CLIError()` in [`show_version`](https://github.com/Azure/azure-cli/blob/38b14f949cbd85cd23f62fde355d3ca75584167f/src/azure-cli/azure/cli/command_modules/util/custom.py#L28-L31):

```py
def show_version(cmd):  # pylint: disable=unused-argument
    raise CLIError()
```

It is converted to a `ValidationError` which in fact is not:

```
> az version
ValidationError:
```

The same happens in https://github.com/Azure/azure-cli/issues/15320.

houk-ms: I totally agree. 

CLIError is a very general error type, it can be any kind of error. Let's have some discussions in the error improvement meeting, to see how we deal with CLIError when applying the new error types. One option can be: 

NOT print error_type info for CLIError (also for other uncertain error types) until it is completely replaced.

houk-ms:
Another problem in exception_handler: 
We are extracting error messages from the response instead of the error's formatted error_msg for AzureError, CloudError and HttpResponseError. Pro is we can show more useful info from response, con is hard to maintain and have to deal with all knids of 
abnormal response cases. Here's an example the useful message is suppressed #15314

src/azure-cli-core/azure/cli/core/azclierror.py
```diff
@@ -4,85 +4,178 @@
...
-        if self.raw_exception:
-            telemetry.set_exception(self.raw_exception, '')
+        telemetry.set_error_type(self.__class__.__name__)
+# endregion
+
+
+# region: Second Layer
+# Main categories of the AzureCLI error types, used for Telemetry alalysis
+# DO NOT raise the error classes here directly in your codes.
+class UserFault(AzCLIError):
```
yungezz, line 60: good point from @qwordy . how about `_UserFault` with `_` prefix to hint user not to use those?

src/azure-cli-core/azure/cli/core/azclierror.py
```diff
@@ -4,85 +4,178 @@
...
-        :type raw_exception: Exception
-        :param command: The command which brings the error.
-        :type command: str
-        :param recommendations: The recommendations to resolve the error.
-        :type recommendations: list
-        """
-        self.error_type = error_type
+    """ Base class for all the AzureCLI defined error classes. """
+
+    def __init__(self, error_msg):
```
jiasli, line 22: Better to support `recommendation` in the constructor so that the exception can be created and raised in one line:

```py
raise AzureCLIError(error_msg, recommendation)
```

fengzhou-msft (approved):

src/azure-cli-core/azure/cli/core/util.py
houk-ms: :(

src/azure-cli-core/azure/cli/core/parser.py
houk-ms: No, they are thrown by knack, it's a little hard to customize

src/azure-cli-core/azure/cli/core/azclierror.py
houk-ms: sure

src/azure-cli-core/azure/cli/core/azclierror.py
qwordy: We can use Abstract Base Classes.
https://docs.python.org/3/library/abc.html
https://www.python.org/dev/peps/pep-3119/

qwordy: I don't it is revolved. Reopen the conversation.

andrewchoi5 (approved):

Issue #15320, opened 2020-09-28
az cli  2.12.0, no matter what command I run, always throws " ValidationError:"


theasphaltworld:
> ### `az feedback` auto-generates most of the information requested below, as of CLI version 2.0.62

**Describe the bug**

I can successfully do az login, and select my subscription, but no matters what I do next, like az group list, it throws error in red 
"ValidationError:"

tried with powershell, windows terminal, CMDer

**To Reproduce**

az login, az account set -s
az group list

**Expected behavior**

**Environment summary**

azure-cli                         2.12.0
windows 10 1909

**Additional context**

jiasli:
Could you run the command with `--debug` and share the output?

```
az group list --debug
```

theasphaltworld:
Command arguments: ['group', 'list', '--debug']
Event: Cli.PreExecute []
Event: CommandParser.OnGlobalArgumentsCreate [<function CLILogging.on_global_arguments at 0x033DC8E8>, <function OutputProducer.on_global_arguments at 0x0356AD68>, <function CLIQuery.on_global_arguments at 0x0358FD68>]
Event: CommandInvoker.OnPreCommandTableCreate []
Modules found from index for 'group': ['azure.cli.command_modules.resource']
Loading command modules:
Name                  Load Time    Groups  Commands
resource                  0.009        34       158
Total (1)                 0.009        34       158
These extensions are not installed and will be skipped: ['azext_ai_examples', 'azext_ai_did_you_mean_this']
Loading extensions:
Name                  Load Time    Groups  Commands  Directory
Total (0)                 0.000         0         0
Loaded 34 groups, 158 commands.
Found a match in the command table.
Raw command  : group list
Command table: group list
Event: CommandInvoker.OnPreCommandTableTruncate [<function AzCliLogging.init_command_file_logging at 0x037B4B28>]
az_command_data_logger : command args: group list --debug
metadata file logging enabled - writing logs to 'C:\Users\rchen\.azure\commands'.
Event: CommandInvoker.OnPreArgumentLoad [<function register_global_subscription_argument.<locals>.add_subscription_parameter at 0x03808348>, <function register_global_query_examples_argument.<locals>.register_query_examples at 0x03808A98>]
Event: CommandInvoker.OnPostArgumentLoad []
[...]

PR #15322, opened 2020-09-28
{Core} Extract error message from AdalError

jiasli:
**Description**  
Reported in https://github.com/Azure/azure-cli/issues/15320

If `AdalError` is created directly with

```py
raise AdalError('More than one token matches the criteria. The result is ambiguous.')
```

The original code fails to extract the error message and returns an empty string `''`. Thus, the original error is lost.


yonzhan:
Core

evelyn-ys (approved):

Juliehzl (approved):

jsntcy (approved):

PR #15376, opened 2020-10-01
[AKS] Honor addon names defined in Azure CLI

KaiqiZhang:
**Description**  

AKS server treats addon names and addon config keys as case insensitive, and store them in DB as it is in input. For example, we have both `logAnalyticsWorkspaceResourceID` and `loganalyticsworkspaceresourceid` in our DB and both are valid. However, in the public document we're only giving examples of `logAnalyticsWorkspaceResourceID`. 
This change makes Azure CLI always convert the keys to the name defined in Azure CLI. (The name defined in Azure CLI is the same as public document), so that customer get the right name back.

**Testing Guide**  

1. Create a AKS cluster.
2. Run `az aks enable-addons -a kube-dashboard -g testrg -n testcluster`
3. The name of kube-dashboard addon name should be change from "KubeDashboard" to "kubeDashboard"

**History Notes**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acs/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/custom.py
@@ -2429,8 +2429,12 @@ def _update_addons(cmd, instance, subscription_id, resource_group_name, addons, 
         if addon == 'aciConnector':
             # only linux is supported for now, in the future this will be a user flag
             addon += os_type
-        # addon name is case insensitive
-        addon = next((x for x in addon_profiles.keys() if x.lower() == addon.lower()), addon)
+
+        # honor addon names defined in Azure CLI
+        for key in list(addon_profiles):
+            if key.lower() == addon.lower() and key != addon:
+                addon_profiles[addon] = addon_profiles.pop(key)
+
         if enable:
             # add new addons or update existing ones and enable them
             addon_profile = addon_profiles.get(addon, ManagedClusterAddonProfile(enabled=False))
@@ -2740,8 +2744,9 @@ def _ensure_default_log_analytics_workspace_for_monitoring(cmd, subscription_id,
 
 def _ensure_container_insights_for_monitoring(cmd, addon):
     # Workaround for this addon key which has been seen lowercased in the wild.
-    if 'loganalyticsworkspaceresourceid' in addon.config:
-        addon.config['logAnalyticsWorkspaceResourceID'] = addon.config.pop('loganalyticsworkspaceresourceid')
+    for key in list(addon.config):
+        if key.lower() == 'loganalyticsworkspaceresourceid' and key != 'logAnalyticsWorkspaceResourceID':
+            addon.config['logAnalyticsWorkspaceResourceID'] = addon.config.pop(key)
 
     workspace_resource_id = addon.config['logAnalyticsWorkspaceResourceID']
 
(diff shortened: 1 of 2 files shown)


KaiqiZhang:
@robbiezhang

yonzhan:
AKS

src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -2737,8 +2741,9 @@ def _ensure_default_log_analytics_workspace_for_monitoring(cmd, subscription_id,
 
 def _ensure_container_insights_for_monitoring(cmd, addon):
     # Workaround for this addon key which has been seen lowercased in the wild.
-    if 'loganalyticsworkspaceresourceid' in addon.config:
-        addon.config['logAnalyticsWorkspaceResourceID'] = addon.config.pop('loganalyticsworkspaceresourceid')
+    for key in list(addon.config):
+        if key.lower() == 'loganalyticsworkspaceresourceid' and key != 'logAnalyticsWorkspaceResourceID':
```
yungezz, line 2745: for != part, better to lower() before compare

yungezz (commented):
could you pls add test to cover this?

yungezz:
hi @KaiqiZhang would you pls address comments? today is Code complete date, would you like to catch this release?

KaiqiZhang:
> hi @KaiqiZhang would you pls address comments? today is Code complete date, would you like to catch this release?

Oh, sorry for late. Will do it soon.

src/azure-cli/azure/cli/command_modules/acs/custom.py
KaiqiZhang: Here my intention is to change the keys back to 'logAnalyticsWorkspaceResourceID'. The `key != 'logAnalyticsWorkspaceResourceID'` is to skip if the key is already be 'logAnalyticsWorkspaceResourceID'.
For example, if the key is 'loganalyticsworkspaceresourceid' or 'LogAnalyticsWorkspaceResourceID', the value will be pop out and write back with key 'logAnalyticsWorkspaceResourceID'. But if the key is already 'logAnalyticsWorkspaceResourceID', it skips and nothing changed.

fengzhou-msft: can you assign 'logAnalyticsWorkspaceResourceID' to a variable and use the variable instead of using the hard-coded string in multiple places?

fengzhou-msft (approved):

yungezz:
hi @KaiqiZhang just want to double confirm, this change will not cause existing users' breaking change right?

src/azure-cli/azure/cli/command_modules/acs/custom.py
KaiqiZhang: @fengzhou-msft We're already doing that pattern in azure-cli-extension. I'll change that in another pr.

KaiqiZhang:
> hi @KaiqiZhang just want to double confirm, this change will not cause existing users' breaking change right? little concern because it changed existing output

It's a long story. 2 months ago, we did a breaking change on the server side. All the keys are converted to lower cases. Based on the IcM at that time and discussion in the team, we reverted the change. Some customer cares about the case because they're using json query in their script. With this change in Azure CLI, customers can change the case back if it's necessary for them. (customer can still use different case in Terraform, Postman, etc.)
Back to this question, it's possible that customer see different addon key if disable, re-enable a specific addon. But it aligns to our public document. I suggest we can mention this change in the change log.

PR #15392, opened 2020-10-02
[RDBMS] Bugfix: az postgres flexible-server create Remove hardcoded API version from network client.

arde0708:
The 'az postgres flexible-server create'  command fails with an error as reported in the issue. It happens only when CLI is installed through the installer.

This PR just removes that hardcoded API version.
https://github.com/MicrosoftDocs/azure-docs/issues/63318


yonzhan:
RDBMS

src/azure-cli/azure/cli/command_modules/rdbms/_client_factory.py
```diff
@@ -324,5 +324,4 @@ def resource_client_factory(cli_ctx, **_):
 
 
 def network_client_factory(cli_ctx):
-    from azure.mgmt.network import NetworkManagementClient
-    return get_mgmt_service_client(cli_ctx, NetworkManagementClient, api_version="2018-08-01")
+    return get_mgmt_service_client(cli_ctx, ResourceType.MGMT_NETWORK)
```
Juliehzl, line 327: actually you even don't need to define it here. network module already has the definition and you could just import it.
https://github.com/Azure/azure-cli/blob/0a90d440930c38b2a7fa8ecf6f500cfb55ae1ca4/src/azure-cli/azure/cli/command_modules/network/_client_factory.py#L7-L10

src/azure-cli/azure/cli/command_modules/rdbms/_params.py
```diff
@@ -229,24 +228,29 @@ def _flexible_server_params(command_group):
...
                 c.argument('zone', options_list=['--zone, -z'],
                            help='Availability zone into which to provision the resource.')
             elif command_group == 'mysql':
                 c.argument('tier', default='Burstable',
                            help='Compute tier of the server. Accepted values: Burstable, GeneralPurpose, Memory Optimized ')
                 c.argument('sku_name', default='Standard_B1ms', options_list=['--sku-name'],
-                           help='The name of the compute SKU. Follows the convention Standard_{VM name}. Examples: Standard_B1ms, Standard_D4s_v3 ')
-                c.argument('storage_mb', default='10', options_list=['--storage-size'], type=int, validator=mysql_storage_validator,
+                           help='The name of the compute SKU. Follows the convention Standard_{VM name}. Examples: Standard_B1ms, Standard_E16ds_v4 ')
+                c.argument('storage_mb', default='10', options_list=['--storage-size'], type=int,
```
Juliehzl, line 249:
```suggestion
                c.argument('storage_mb', default=10, options_list=['--storage-size'], type=int,
```
because it is int type.

src/azure-cli/azure/cli/command_modules/rdbms/_params.py
```diff
@@ -229,24 +228,29 @@ def _flexible_server_params(command_group):
...
+                c.argument('sku_name', default='Standard_D2s_v3', options_list=['--sku-name'],
                            help='The name of the compute SKU. Follows the convention Standard_{VM name}. Examples: Standard_D4s_v3 ')
-                c.argument('storage_mb', default='128', options_list=['--storage-size'], type=int, validator=pg_storage_validator,
+                c.argument('storage_mb', default='128', options_list=['--storage-size'], type=int,
                            help='The storage capacity of the server. Minimum is 32 GiB and max is 16 TiB.')
-                c.argument('version', default='12', options_list=['--version'], validator=pg_version_validator,
+                c.argument('backup_retention', default=7, type=int, options_list=['--backup-retention'],
+                           help='The number of days a backup is retained. Range of 7 to 35 days. Default is 7 days.',
+                           validator=retention_validator)
+                c.argument('version', default='12', options_list=['--version'],
```
Juliehzl, line 240: quick question: how can user know which kind of version is accepted？

arde0708: > quick question: how can user know which kind of version is accepted？

Do you mean how they can find out which are the valid versions ?

src/azure-cli/azure/cli/command_modules/rdbms/_params.py
```diff
@@ -301,12 +300,13 @@ def _flexible_server_params(command_group):
             c.argument('tags', tags_type)
             if command_group == 'mysql':
                 c.argument('tier', options_list=['--tier'],
-                           help='Compute tier of the server. Accepted values: Burstable, GeneralPurpose, Memory Optimized ')
+                           help='Compute tier of the server. Accepted values: Burstable, GeneralPurpose, Memory Optimized')
```
Juliehzl, line 303: still prefer getting allowed value from SDK

src/azure-cli/azure/cli/command_modules/rdbms/_params.py
```diff
@@ -301,12 +300,13 @@ def _flexible_server_params(command_group):
...
                 c.argument('sku_name', options_list=['--sku-name'],
-                           help='The name of the compute SKU. Follows the convention Standard_{VM name}. Examples: Standard_B1ms, Standard_D4s_v3 ')
+                           help='The name of the compute SKU. Follows the convention Standard_{VM name}. Examples: Standard_B1ms, Standard_E16ds_v4 ')
```
Juliehzl, line 305: The same

src/azure-cli/azure/cli/command_modules/rdbms/_params.py
```diff
@@ -301,12 +300,13 @@ def _flexible_server_params(command_group):
...
                 c.argument('storage_mb', options_list=['--storage-size'], type=int,
-                           validator=mysql_storage_validator,
                            help='The storage capacity of the server. Minimum is 5 GiB and increases in 1 GiB increments. Max is 16 TiB.')
+                c.argument('backup_retention', type=int, options_list=['--backup-retention'],
```
Juliehzl, line 308: Here is to define the retention days but how could I enable/disable delete retention?

src/azure-cli/azure/cli/command_modules/rdbms/_params.py
```diff
@@ -317,13 +317,14 @@ def _flexible_server_params(command_group):
...
-                c.argument('sku_name', options_list=['--sku-name'], validator=pg_sku_name_validator,
+                c.argument('tier', options_list=['--tier'],
+                           help='Compute tier of the server. Accepted values: Burstable, GeneralPurpose, Memory Optimized')
+                c.argument('sku_name', options_list=['--sku-name'],
                            help='The name of the compute SKU. Follows the convention Standard_{VM name}. Examples: Standard_D4s_v3 ')
                 c.argument('storage_mb', options_list=['--storage-size'], type=int,
-                           validator=pg_storage_validator,
                            help='The storage capacity of the server. Minimum is 32 GiB and max is 16 TiB.')
+                c.argument('backup_retention', type=int, options_list=['--backup-retention'],
+                           help='The number of days a backup is retained. Range of 7 to 35 days. Default is 7 days.', validator=retention_validator)
```
Juliehzl, line 327: you have so many duplicate arguments. could you use CLIArgumentType to abstract them and reuse these type.

src/azure-cli/azure/cli/command_modules/rdbms/_params.py
```diff
@@ -412,15 +413,9 @@ def _flexible_server_params(command_group):
...
-                           help='Compute tier of the server. Accepted values: Burstable, GeneralPurpose, Memory Optimized ')
-                c.argument('sku_name', options_list=['--sku-name'],
-                           help='The name of the compute SKU. Follows the convention'
-                                ' Standard_{VM name}. Examples: Standard_B1ms, Standard_D4s_v3 ')
-            if command_group == 'postgres':
-                c.argument('tier', options_list=['--tier'], validator=tier_validator,
-                           help='Compute tier of the server. Accepted values: Burstable, GeneralPurpose, Memory Optimized ')
+            c.ignore('location')
+            c.ignore('sku_name')
+            c.ignore('tier')
```
Juliehzl, line 418: The three arguments are totally removed in existing command?

src/azure-cli/azure/cli/command_modules/rdbms/_params.py
arde0708: It cannot be disabled. By default it is 7 days. But user can pass a value and override the default value.

src/azure-cli/azure/cli/command_modules/rdbms/_params.py
arde0708: If they pass in a wrong value, they will get a Validation error with the set of acceptable values.

src/azure-cli/azure/cli/command_modules/rdbms/_params.py
arde0708: We are not using these three args for the command currently

Juliehzl (approved):
discussed with @arde0708 online and checked the telemetry for removed arguments (1 sub usage from @arde0708  team).
Approved the PR for catching  the release.

Please fix other comments in the following PR and prefer separating to several PRs for different motivation.

PR #15449, opened 2020-10-08
[AppConfig] Fix querying key-values by comma separated labels

### LINKED ISSUE #15440: App Configuration: Key/values without labels are not returned when using the label parameter
## Describe the bug

The use of `--label \0` has broken in one of the more recent versions of the Azure CLI, causing our CI/CD to fail.

We use the `--label` flag to filter out any key/values that should not be accessible to the application. We use `az appconfig kv list` at deploy-time (in Azure DevOps) to build up a list of environment variables & secrets that should be deployed along with our micro-services (to aks).

Our build script has been failing since of today (last successful deploy was at `2020-10-06T01:58:48.375Z`), complaining that particular environment variables could not be found within Azure App Configuration. I ran the deploy script locally, and it worked just fine, so I updated to the latest version of the Azure CLI (via Homebrew), re-ran the deploy script, and it failed.

I debugged the deploy script, and I can see that the call to `az appconfig kv list` is only returning key/values that have labels. For example, we have one App Configuration resource per environment, and key/values that we'd like all micro-services to have access to have no label, and key/values we only want particular micro-services to have access to have a label like this: `app=custom-tools-service` or `app=text-extraction-service`. We have generic key/values (like `NODE_ENV`) which are shared across all micro-services, thus do not have a label attached, and those are the ones that are not returning.

I have tried a number of things to try and get the correct results returning from the CLI, however nothing I have tried has worked. I have tried re-ordering the labels in the cli. I have tried URL encoding the `\0` label manually. And a number of random other things.

**Command Name**
`az appconfig kv list`

**Errors:**
The command doesn't error, however it does not return any key/values. The App Configuration resource has many key/values without labels.

## To Reproduce:

- `az appconfig kv list --subscription [REDACTED] --name [REDACTED] --label \0`

OR

- `az appconfig kv list --subscription [REDACTED] --name [REDACTED] --label \0,app=custom-tools-service`

## Expected Behavior

All key/values without labels should be returned.

## Environment Summary
```
macOS-10.15.6-x86_64-i386-64bit
Python 3.9.0
Installer: HOMEBREW

azure-cli 2.12.1

Extensions:
front-door 1.0.4

```

avanigupta:
**Description**  
Fix #15440 

**Testing Guide**  

This PR will address the failure to retrieve key-values when multiple labels are provided for filtering, and one of those labels is null label \0:

`az appconfig kv list -n <storename> --label \0,mylabel
`

**History Notes**  

---


shenmuxiaosen (approved):
:shipit:

shenmuxiaosen:
Can you help to modify the command exmaples for using null label. Like add double quotes.

yonzhan:
AppConfig

avanigupta:
I thought about adding that, but my concern is that its very specific to bash environment. We don't explicitly provide examples for bash anywhere else, and I think bash users would already know that any \ character needs to be escaped or enclosed in quotes.

---
In reply to: [705803422](https://github.com/Azure/azure-cli/pull/15449#issuecomment-705803422) [](ancestors = 705803422)

avanigupta:
Hi @bim-msft , this PR is ready for review.

yungezz:
hi @bim-msft could you pls review the PR? thanks.

bim-msft (approved):

bim-msft:
Ready for merging? @avanigupta

avanigupta:
@bim-msft , it's ready to merge. Thanks!

PR #15456, opened 2020-10-09
{Network} Fix dns loading issue in azure stack

Juliehzl:
**Description**  

Fix issue in https://dev.azure.com/azure-sdk/public/_build/results?buildId=557475&view=logs&j=b21ded15-2ef6-5c2b-875f-d00fdf573f12&t=293d8450-9e65-5c81-f34e-5d7d70fd3ffc.

The root cause is that for network module, the default resource type is MGMT_NETWORK and it is applied to `az network dns` if no addition specification.
In previous profile, the issue doesn't happen because MGMT_NETWOR api version is lower than 2018-05-11. But in new 2020-09-01-hybrid profile, it is larger than 2018-05-01 and the command will be loaded but there is no such support in SDK, which will cause error as shown in CI.

**Testing Guide**  

**History Notes**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/network/commands.py
+++ b/src/azure-cli/azure/cli/command_modules/network/commands.py
@@ -635,7 +635,7 @@ def load_command_table(self, _):
     # endregion
 
     # region DNS
-    with self.command_group('network dns', network_dns_reference_sdk) as g:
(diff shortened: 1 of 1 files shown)


yonzhan:
Network

haroldrandom (approved):

PR #14962, opened 2020-08-28
{ARM} Fix the bug for live test

zhoxing-ms:
**Description**  

Fix the bug for live test:

***1. Fixed resource names***
* Errors:
   - test_subscription_level_deployment               *The storage account named armbuilddemo18221 is already taken*
   - test_subscription_level_deployment_old_command   *The storage account named armbuilddemo18221 is already taken.*
   - test_group_deployment_with_large_params   *cli_test_deployment_with_large_paramsxxx is not valid.*
* Solutions:
   - Use the random method to change the resource name to a random value

***2. The subscription ID belonging to service team is used***
* Errors:
   - test_management_group_deployment  *knack.util.CLIError: Azure Error: CrossTenantDeploymentNotPermitted*
   - test_management_group_level_what_if  *knack.util.CLIError: Azure Error: CrossTenantDeploymentNotPermitted*
   - test_resource_policy_management_group  Reader level access is required on the management group or group ancestors.
* Solutions:
   - Change the subscription ID to our own

***3. The check status value is not correct***
* Errors:
   - test_feature_unregister  *'properties.state' doesn't yield expected value 'Unregistered', instead the actual value is 'Unregistering'*
* Solutions:
    - Change to the correct check value

**Testing Guide**  

**History Notes**  

[Component Name 1] BREAKING CHANGE: az command a: Make some customer-facing breaking change.  
[Component Name 2] az command b: Add some customer-facing feature.

---

This checklist is used to make sure that common guidelines for a pull request are followed.

- [x] The PR title and description has followed the guideline in [Submitting Pull Requests](https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules#submitting-pull-requests).

- [x] I adhere to the [Command Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md).

PR #15480, opened 2020-10-12
[ARM] BREAKING CHANGE: Add user confirmation for az ts create

detienne20:
**Description**  

This PR adds a user confirmation flag to az ts create to account for the scenario where a user updates a template spec version via the create command;  in that case the existing template spec version would be overwritten because create is not a patch like operation. 

**Testing Guide**  

{
1. Create template spec version 

ts create -g {rg} -n {template_spec_name} -v 1.0 -l {resource_group_location} -f "{template_file}" -d {display_name} --description {description} --version-description {version_description}

2. Use the create function to  "update"  the template spec in step one. The description should now be null and the template content changed. 

ts create -g {rg} -n {template_spec_name} -v 1.0 -f "{different_template_file}" --yes

}


yonzhan:
ts

src/azure-cli/azure/cli/command_modules/resource/tests/latest/subscription_level_template.json
```diff
@@ -4,7 +4,7 @@
   "parameters": {
     "storageAccountName": {
       "type": "string",
-      "defaultValue": "armbuilddemo1801"
+      "defaultValue": "cliarmbuilddemo1804"
```
zhoxing-ms, line 7: Can we pass in this parameter by random value? If it is written as a fixed value, sometimes it will affect the test results of our live test

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -540,6 +540,7 @@ def load_arguments(self, _):
         c.argument('display_name', arg_type=ts_display_name_type)
         c.argument('description', arg_type=ts_description_type)
         c.argument('version_description', arg_type=ts_version_description_type)
+        c.argument('prompt', options_list=['--yes', '-y'], action='store_true', help='Do not prompt for confirmation')
```
zhoxing-ms, line 543: May I ask if `--yes` is passed in for this parameter, then there will be no prompt, right? If so, is it better to change the parameter name to `no_prompt`? Because the meaning of this name is more clear and accurate.

src/azure-cli/azure/cli/command_modules/resource/tests/latest/subscription_level_template.json
detienne20: I agree. I changed the value since storage account names have to be globally unique and this test repeatedly failed as a result.

zhoxing-ms: You may modify it by referring to this PR: https://github.com/Azure/azure-cli/pull/14962/files#diff-ef876d58099baf761fb6f20c0aa11991830c232a37b137875c103839524b684cR676

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -540,6 +540,7 @@ def load_arguments(self, _):
         c.argument('display_name', arg_type=ts_display_name_type)
         c.argument('description', arg_type=ts_description_type)
         c.argument('version_description', arg_type=ts_version_description_type)
+        c.argument('no_prompt', options_list=['--yes', '-y'], action='store_true', help='Do not prompt for confirmation')
```
Juliehzl, line 543: From your current design, the default behavior will be prompt, which is a breaking change for existing users, especially automation scripts.
If you are sure to make such change, please let me know and mark BREAKING CHANGE

detienne20: That is the intention, title has been updated.

Juliehzl (approved):
LGTM in general. Could you refine the title name with [ARM] not [ts]?

zhoxing-ms (approved):
LGTM

jsntcy (approved):

Issue #12909, opened 2020-04-07
-j/--handle-extended-json-format does not work as expected for mutil-line string containing double quotes


galiacheng:
Hi experts,
I found if an ARM template contains multi-line string containing double quotes, the template deployment will fail, please help to check, thank you.
I would like to pass parameters that contains double quotes to shell script within virtual machine extension, please see variables executeCustomScriptArguments1. The template deployment failed. I deploy the template with az cli flag --handle-extended-json-format, az version is 2.2.0.
My template works if I change the mutil-line string to single line string.
Error:
```
cli.azure.cli.core.util : Invalid control character at: line 1 column 5789 (char 5788)
```
The flag also works for cases listed in [document](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/template-syntax#multi-line-strings), I can make sure my Az CLI supports --handle-extended-json-format.

Attach my template:
```
{
   "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
   "contentVersion": "1.0.0.0",
   "parameters": {
      "_artifactsLocation": {
         "type": "string",
         "metadata": {
            "description": "The base URI where artifacts required by this template are located. When the template is deployed using the accompanying scripts, a private location in the subscription will be used and this value will be automatically generated."
         },
         "defaultValue": "[deployment().properties.templateLink.uri]"
      },
      "_artifactsLocationSasToken": {
         "type": "securestring",
         "metadata": {
            "description": "The sasToken required to access _artifactsLocation.  When the template is deployed using the accompanying scripts, a sasToken will be automatically generated. Use the defaultValue if the staging location is not secured."
         },
         "defaultValue": ""
      },
      "guidValue": {
         "type": "string",
         "defaultValue": "[newGuid()]"
      },
      "adminUsername": {
         "type": "string",
         "metadata": {
            "description": "User name for the Virtual Machine."
         }
      },
      "authenticationType": {
         "type": "string",
         "defaultValue": "password",
         "allowedValues": [
            "sshPublicKey",
            "password"
         ],
         "metadata": {
            "description": "Type of authentication to use on the Virtual Machine. SSH key is recommended."
         }
      },
      "adminPasswordOrKey": {
         "type": "securestring",
         "metadata": {
            "description": "SSH Key or password for the Virtual Machine. SSH key is recommended."
         }
      },
      "acceptOTNLicenseAgreement": {
         "type": "string",
         "metadata": {
[...]

PR #15502, opened 2020-10-13
[ARM] `az deployment`: Fix parse errors with multiline strings

wiktorn:
**Description**  
For more complex multiline strings (such as policies for Azure API Management), `az deploy` will fail with message: ` Failed to parse ... , please check whether it is a valid JSON format`, because multiline string contains block quotes within the string.

This PR adds test case that's failing with current az and naive proposal to improve detection of such situation. I couldn't find a regexp that is solving all cases, that's why I just slightly changed existing one and added check with replacement. The regex is already quite complex and I think it's already on the bridge of being understandable.

**Testing Guide**  
Test cases are provided within PR.

**History Notes**  
[resource] Fix parsing JSON files with multi-line strings  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/resource/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/resource/custom.py
@@ -275,10 +275,8 @@ def _remove_comments_from_json(template, preserve_order=True, file_path=None):
     # It will affect the subsequent multi-line processing logic, so deal with this situation in advance here.
     template = re.sub(r'(^[\t ]*//[\s\S]*?\n)|(^[\t ]*/\*{1,2}[\s\S]*?\*/)', '', template, flags=re.M)
     minified = jsmin(template)
-    # Get rid of multi-line strings. Note, we are not sending it on the wire rather just extract parameters to prompt for values
-    result = re.sub(r'"[^"]*?\n[^"]*?(?<!\\)"', '"#Azure Cli#"', minified, re.DOTALL)
     try:
-        return shell_safe_json_parse(result, preserve_order)
+        return shell_safe_json_parse(minified, preserve_order, strict=False) # use strict=False to allow multiline strings
     except CLIError:
         # Because the processing of removing comments and compression will lead to misplacement of error location,
         # so the error message should be wrapped.
--- a/src/azure-cli/azure/cli/command_modules/resource/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/resource/custom.py
@@ -276,7 +276,7 @@ def _remove_comments_from_json(template, preserve_order=True, file_path=None):
     template = re.sub(r'(^[\t ]*//[\s\S]*?\n)|(^[\t ]*/\*{1,2}[\s\S]*?\*/)', '', template, flags=re.M)
     minified = jsmin(template)
     try:
-        return shell_safe_json_parse(minified, preserve_order, strict=False) # use strict=False to allow multiline strings
+        return shell_safe_json_parse(minified, preserve_order, strict=False)  # use strict=False to allow multiline strings
     except CLIError:
         # Because the processing of removing comments and compression will lead to misplacement of error location,
         # so the error message should be wrapped.
--- a/src/azure-cli-core/azure/cli/core/util.py
+++ b/src/azure-cli-core/azure/cli/core/util.py
@@ -501,15 +501,15 @@ def read_file_content(file_path, allow_binary=False):
     raise CLIError('Failed to decode file {} - unknown decoding'.format(file_path))
 
 
-def shell_safe_json_parse(json_or_dict_string, preserve_order=False):
+def shell_safe_json_parse(json_or_dict_string, preserve_order=False, strict=True):
     """ Allows the passing of JSON or Python dictionary strings. This is needed because certain
     JSON strings in CMD shell are not received in main's argv. This allows the user to specify
     the alternative notation, which does not have this problem (but is technically not JSON). """
     try:
         if not preserve_order:
-            return json.loads(json_or_dict_string)
+            return json.loads(json_or_dict_string, strict=strict)
         from collections import OrderedDict
-        return json.loads(json_or_dict_string, object_pairs_hook=OrderedDict)
+        return json.loads(json_or_dict_string, object_pairs_hook=OrderedDict, strict=strict)
     except ValueError as json_ex:
         try:
             import ast
--- a/src/azure-cli/azure/cli/command_modules/resource/tests/latest/recordings/test_resource_group_deployment.yaml
+++ b/src/azure-cli/azure/cli/command_modules/resource/tests/latest/recordings/test_resource_group_deployment.yaml
@@ -1,24 +1,24 @@
 interactions:
 - request:
-    body: "{\"properties\": {\"parameters\": {\"location\": {\"value\": \"westus\"},
-      \"name\": {\"value\": \"azure-cli-deploy-test-nsg1\"}}, \"mode\": \"Incremental\",
-      template:{\r\n  \"$schema\": \"https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#\",\r\n
-      \ \"contentVersion\": \"1.0.0.0\",\r\n  \"parameters\": {\r\n    \"location\":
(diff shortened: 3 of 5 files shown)


yonzhan:
resource

zhoxing-ms:
Related issue: #12909

zhoxing-ms:
@wiktorn Hi, thank you for your contribution.

However, there are still some problems with the support of service for multi-line, the verification of CLI pass multi-line still can't solve the problem.

1. If it is multi-line of pure describing text, service will break the original format with newline character. For example:
- In ARM template:
```
"meta": {
    "type": "string",
    "metadata": {
        "description": "meta of the network security group."
    },
    "defaultValue": "ARM
    template
    test
    newline"
}
```
- After service processing:
```
"meta": {
    "type": "String",
    "value": "ARM\r\n      template\r\n      test\r\n      newline"
}
```

2. If it is multi-line of parameter values, service may break the resolution of those values.
For example:
```
"defaultValue": "
    [resourceGroup().location]
"
```
It will cause an error in the service:
```
{"error":{"code":"InvalidResourceLocation","message":"The specified location '\r\n        [resourceGroup().location]\r\n      ' is invalid. A location must consist of characters, whitespace, digit, or following symbols '(,)'."}}
```

Please refer to this issue #12909 for the discussion of the service problem.
I suggest that this PR should not merge at this time. After the service solves this problem, this PR can be merged to pass the verification of multi-line in CLI side.

wiktorn:
I gave this second thought and I thing that maybe the best approach to this problem is to use custom json decoder (lets call it azjson) that would respect the rules for Resource templates (comments allowed, multiline strings allowed). Is there any chance to use the same algorithm on both ends?

If such tool would be written, then following template:
```
"meta": {
    "type": "string",
    "metadata": {
        "description": "meta of the network security group."
    },
    "defaultValue": "ARM
    template
    test
    newline"
}`
```
Would indeed have in-memory representation as:
```
"meta": {
    "type": "String",
    "value": "ARM\r\n      template\r\n      test\r\n      newline"
}
```
(modulo platform discrepancies around CRLF/CR). 

Also - taking a look into the json module, maybe far easier approach than this regex magic would be just to pass `strict=False` argument to json parser in `shell_safe_json_parse`? The difference would be, that other control characters would be allowed within template.

zhoxing-ms:
@wiktorn Hi, may I ask what should we do if the multi-line is in the value of parameters?
Such as: 
```
"defaultValue": "
    [resourceGroup().location]
"
```
If it is translated to 
```
defaultValue": "\r\n        [resourceGroup().location]\r\n      "
``` 
, the parsing of this value should be destroyed.

And another factor is that in our design, CLI can't modify the user's template, so it needs to send the template directly to the Service.

wiktorn:
> @wiktorn Hi, may I ask what should we do if the multi-line is in the value of parameters?
> Such as:
> ```
> "defaultValue": "
>     [resourceGroup().location]
> "
> ```

With such templates - do noting. Pass it to Service, and if it will be rejected there - it's perfectly fine as the message is clear (location should not contain whitespace characters) and user can easily fix.

As I understand, you would like multiline strings behave something like that (example in Python):
```python
defaultValue = "" +
    "[resourceGroup().location]" +
    ""
```

But even multiline strings in Python do not behave that way and maintain all the whitespace even though it has semantic meaning in language and for example - all it is possible to remove whitespace that was used to indent nearby code. But this was only implemented (as post-processing) to [docstrings](https://www.python.org/dev/peps/pep-0257/). So I think that Pythons strategy with multiline strings is perfectly fine here too and it's still in line with the documentation. AFAIR there is nothing there, that some magic with whitespace will happen.

Or maybe I'm just totally not understanding the problem that you raise with this. My assumptions are:
* such template fails on Service, because location contains invalid characters
* there is no implementation of ARM Template functions, so cli doesn't parse the value

zhoxing-ms:
> such template fails on Service, because location contains invalid characters

Yes, the value cannot be parsed properly because it contains invalid characters, such as `\r\n      `.

wiktorn:
As I see no follow-up on #12909, can we proceed with this one? Would you like to stay with this approach or you prefer to go with `json.load(..., strict=False)` ?

zhoxing-ms:
Let me try to contact the service team again. I have just summarized the questions into an email to ask them

wiktorn:
@zhoxing-ms any updates on this?

zhoxing-ms:
@wiktorn Hi, as we discussed with the Service team, the current behavior of the service is in line with expectations, so we only need to merge this PR to solve the CLI verification problem (If you are interested in the details of our communication, I can forward the email to you~)

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -287,6 +287,17 @@ def _remove_comments_from_json(template, preserve_order=True, file_path=None):
         raise CLIError("Failed to parse the JSON data, please check whether it is a valid JSON format")
 
 
+def _remove_multiline_from_json(template):
+    result = template
+    # find all text within bock quotes (single-line and multiline)
+    for match in re.finditer(r'((?<!\\)")((?:.?(?!(?<!\\)"))*.?)??"', template, re.DOTALL):
```
zhoxing-ms, line 293: In order to increase the readability of the code, could you please add some comments explaining the regular expression in detail? For example, add some matching samples to it?

wiktorn:
I've created the branch with the other approach (non-strict JSON parsing) which doesn't need any magic regexes:
https://github.com/Azure/azure-cli/compare/dev...wiktorn:multiline_string_nonstrict_parse?expand=1

If you'd still want to continue with present approach, I've added comments with description how this regex works.

wiktorn:
Shall I rebase it on dev? There are conflicts in recordings, as I've recorded them using WSL, also - not sure what is the policy to provide them.

OTOH, as far as I see, there are some other tests failing currently on dev.

zhoxing-ms:
> I've created the branch with the other approach (non-strict JSON parsing) which doesn't need any magic regexes:
> https://github.com/Azure/azure-cli/compare/dev...wiktorn:multiline_string_nonstrict_parse?expand=1
> 
> If you'd still want to continue with present approach, I've added comments with description how this regex works.

It's a nice solution ! I think it's more readable and easier to maintain~
Could you please use it to update this PR?

zhoxing-ms:
> Shall I rebase it on dev? There are conflicts in recordings, as I've recorded them using WSL, also - not sure what is the policy to provide them.
> 
> OTOH, as far as I see, there are some other tests failing currently on dev.

Yes, you need to fetch the latest code from the remote dev branch, then rebase locally before committing this PR

wiktorn:
Code rebased, changed approach to non-strict json parsing.

I haven't updated the test recordings, not sure what to do about that.

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -275,10 +275,8 @@ def _remove_comments_from_json(template, preserve_order=True, file_path=None):
     # It will affect the subsequent multi-line processing logic, so deal with this situation in advance here.
     template = re.sub(r'(^[\t ]*//[\s\S]*?\n)|(^[\t ]*/\*{1,2}[\s\S]*?\*/)', '', template, flags=re.M)
     minified = jsmin(template)
-    # Get rid of multi-line strings. Note, we are not sending it on the wire rather just extract parameters to prompt for values
-    result = re.sub(r'"[^"]*?\n[^"]*?(?<!\\)"', '"#Azure Cli#"', minified, re.DOTALL)
     try:
-        return shell_safe_json_parse(result, preserve_order)
+        return shell_safe_json_parse(minified, preserve_order, strict=False) # use strict=False to allow multiline strings
```
zhoxing-ms, line 279: Please solve the code style issues from Flake8 verification
> custom.py:279:77: E261 at least two spaces before inline comment

```suggestion
        return shell_safe_json_parse(minified, preserve_order, strict=False)  # use strict=False to allow multiline strings
```

zhoxing-ms:
> Code rebased, changed approach to non-strict json parsing.
> 
> I haven't updated the test recordings, not sure what to do about that.

@wiktorn In fact, you need to rerun the test `azdev test test_resource_group_deployment --live` in live mode and then upload the updated yaml file.

zhoxing-ms (approved):

jsntcy (approved):

PR #15538, opened 2020-10-15
[RDBMS] : `az postgres|mariadb|mysql server create` : Updated create experience for Postgres, MySQL and MariaDB - new fields in the output , Introduce new values for `--public` parameter in create command (all,<IP>,<IPRange>,0.0.0.0)

arde0708:
This PR 
- updates the current create experience for Postgres, MySQL and MariaDB services on Azure in order to align it with the experience that we have for flexible servers.
- new tests for the relevant scenarios.
- current test for create command updated
- tests for create with public parameters added.


yonzhan:
RDBMS

src/azure-cli/azure/cli/command_modules/rdbms/_params.py
```diff
@@ -58,7 +57,7 @@ def _complex_params(command_group):
...
             c.argument('ssl_enforcement', arg_type=get_enum_type(['Enabled', 'Disabled']), options_list=['--ssl-enforcement'], help='Enable or disable ssl enforcement for connections to server. Default is Enabled.')
             c.argument('minimal_tls_version', arg_type=get_enum_type(['TLS1_0', 'TLS1_1', 'TLS1_2', 'TLSEnforcementDisabled']), options_list=['--minimal-tls-version'], help='Set the minimal TLS version for connections to server when SSL is enabled. Default is TLSEnforcementDisabled.', validator=tls_validator)
-            c.argument('public_network_access', arg_type=get_enum_type(['Enabled', 'Disabled']), options_list=['--public-network-access', '--public'], help='Enable or disable public network access to server. When disabled, only connections made through Private Links can reach this server. Default is Enabled.')
+            c.argument('public_network_access', options_list=['--public-network-access', '--public'], help='Enable or disable public network access to server. When disabled, only connections made through Private Links can reach this server. Default is Enabled.')
```
Juliehzl, line 60: May I know the reason roving allowed values?

src/azure-cli/azure/cli/command_modules/rdbms/_params.py
```diff
@@ -76,18 +75,19 @@ def _complex_params(command_group):
...
+            c.argument('administrator_login', default=generate_username(), arg_group='Authentication')
+            c.argument('administrator_login_password', arg_group='Authentication')
 
             c.argument('backup_retention', type=int, options_list=['--backup-retention'], help='The number of days a backup is retained. Range of 7 to 35 days. Default is 7 days.', validator=retention_validator)
             c.argument('geo_redundant_backup', arg_type=get_enum_type(['Enabled', 'Disabled']), options_list=['--geo-redundant-backup'], help='Enable or disable geo-redundant backups. Default value is Disabled. Not supported in Basic pricing tier.')
-            c.argument('storage_mb', options_list=['--storage-size'], type=int, help='The storage capacity of the server (unit is megabytes). Minimum 5120 and increases in 1024 increments. Default is 51200.')
+            c.argument('storage_mb', default=51200, options_list=['--storage-size'], type=int, help='The storage capacity of the server (unit is megabytes). Minimum 5120 and increases in 1024 increments. Default is 51200.')
```
Juliehzl, line 85: May I know the change in client side? is it to align with server side as shown in description?

src/azure-cli/azure/cli/command_modules/rdbms/_params.py
```diff
@@ -76,18 +75,19 @@ def _complex_params(command_group):
...
             c.argument('auto_grow', arg_type=get_enum_type(['Enabled', 'Disabled']), options_list=['--auto-grow'], help='Enable or disable autogrow of the storage. Default value is Enabled.')
             c.argument('infrastructure_encryption', arg_type=get_enum_type(['Enabled', 'Disabled']), options_list=['--infrastructure-encryption', '-i'], help='Add an optional second layer of encryption for data using new encryption algorithm. Default value is Disabled.')
             c.argument('assign_identity', options_list=['--assign-identity'], help='Generate and assign an Azure Active Directory Identity for this server for use with key management services like Azure KeyVault.')
 
-            c.argument('location', arg_type=get_location_type(self.cli_ctx), validator=get_default_location_from_resource_group)
+            c.argument('location', arg_type=get_location_type(self.cli_ctx))
```
Juliehzl, line 90: why not get default value from resource group?

src/azure-cli/azure/cli/command_modules/rdbms/_params.py
arde0708: We are aligning the experience to that of Flexible Severs which recently went live in Ignite. As a part of this change, we are allowing more values apart from Enabled and Disabled. We will still continue to support Enabled and Disabled as well. so, this wont be a breaking change.

src/azure-cli/azure/cli/command_modules/rdbms/_util.py
```diff
@@ -35,3 +37,49 @@ def expand(self, dest, model_type, group_name=None, patches=None):
...
+def parse_public_network_access_input(public_network_access):
+    # pylint: disable=no-else-return
+    if public_network_access is not None:
+        parsed_input = public_network_access.split('-')
+        if len(parsed_input) == 1 and str(parsed_input).find('.') != -1:
+            return parsed_input[0], parsed_input[0]
+        elif len(parsed_input) == 2:
+            return parsed_input[0], parsed_input[1]
+        else:
+            raise CLIError('incorrect usage: --public/--public_network_access. Acceptable values are \'all\','
```
Juliehzl, line 52:
```suggestion
            raise CLIError('incorrect usage: --public/--public-network-access. Acceptable values are \'all\','
```

src/azure-cli/azure/cli/command_modules/rdbms/_params.py
arde0708: Yes, for single servers, we are changing the default to 50 GB.

src/azure-cli/azure/cli/command_modules/rdbms/_util.py
```diff
@@ -35,3 +37,49 @@ def expand(self, dest, model_type, group_name=None, patches=None):
...
+                           ' \'enabled\', \'disabled\', \'<startIP>\' and \'<startIP>-<destinationIP>\' '
```
Juliehzl, line 53: could we expose these values in help meassage? otherwise, customer will fail first and they have to know allowed value from error message.

src/azure-cli/azure/cli/command_modules/rdbms/_util.py
```diff
@@ -35,3 +37,49 @@ def expand(self, dest, model_type, group_name=None, patches=None):
...
+                                                                                                   now.minute,
+                                                                                                   now.second)
+    elif start_ip == end_ip:
+        logger.warning('Configuring server firewall rule to accept connections from \'%s\'...', start_ip)
+    else:
+        if start_ip == '0.0.0.0' and end_ip == '255.255.255.255':
+            firewall_name = 'AllowAll_{}-{}-{}_{}-{}-{}'.format(now.year, now.month, now.day, now.hour, now.minute,
+                                                                now.second)
+        logger.warning('Configuring server firewall rule to accept connections from \'%s\' to \'%s\'...', start_ip,
+                       end_ip)
```
Juliehzl, line 76: it will be better if you can abstract this part to one function as validator. In this way you could reuse it in multiple commands.

src/azure-cli/azure/cli/command_modules/rdbms/custom.py
```diff
@@ -87,7 +140,39 @@ def _server_create(cmd, client, resource_group_name, server_name, sku_name, no_w
             location=location,
             tags=tags)
 
-    return client.create(resource_group_name, server_name, parameters)
+    server_result = resolve_poller(
```
Juliehzl, line 143: What does resolve_poller do here? Is create a long running operation here?

src/azure-cli/azure/cli/command_modules/rdbms/_params.py
arde0708: There are two cases for this :
The user can choose whether or not to provide the location. 
 - In case they do, we take up the location they provided. 
 - In case they don't - we look for it in the local context, else, we provision it in the default location (eastus for PG and westus2 for Mysql and MariaDB). When the user does not provide the resource group, we create the resource group as well. 
Having the above validator would mean that the user **always** has to pass the resource group from which the location will be determined. However, based on the current implementation, the user can choose **not** to provide any arguments and we will create it for them. They will just issue the command - az postgres server create and we will create the resource group and in the default location and use that location for any other action that follows.

src/azure-cli/azure/cli/command_modules/rdbms/custom.py
```diff
@@ -87,7 +140,39 @@ def _server_create(cmd, client, resource_group_name, server_name, sku_name, no_w
...
+        firewall_id = create_firewall_rule(cmd, resource_group_name, server_name, start_ip, end_ip, engine_name)
+
+    logger.warning('Make a note of your password. If you forget, you would have to '
+                   'reset your password with \'az %s flexible-server update -n %s -g %s -p <new-password>\'.',
+                   engine_name, server_name, resource_group_name)
+
+    update_local_contexts(cmd, server_name, resource_group_name, location, user)
+
+    if engine_name == 'postgres':
+        return form_response(user, sku, loc, server_id, host, version,
```
Juliehzl, line 164: response transform could be done with transform as shown in 

https://github.com/Azure/azure-cli/blob/7c03a5d292538f9e8ec923f8b01ce8ed473dbb39/src/azure-cli/azure/cli/command_modules/storage/commands.py#L327

src/azure-cli/azure/cli/command_modules/rdbms/custom.py
```diff
@@ -596,3 +681,79 @@ def _get_tenant_id():
...
+
+def check_server_name_availability(check_name_client, server_name, service_name):
+    server_availability = check_name_client.execute(server_name, service_name)
+    if not server_availability.name_available:
+        raise CLIError("The server name '{}' already exists.Please re-run command with some "
+                       "other server name.".format(server_name))
+    return True
+
+
+def update_local_contexts(cmd, server_name, resource_group_name, location, user):
```
Juliehzl, line 749: @arrownj could you help review local context setting here?

src/azure-cli/azure/cli/command_modules/rdbms/tests/latest/test_rdbms_commands.py
```diff
@@ -89,6 +90,62 @@ def test_mysql_server_mgmt(self, resource_group_1, resource_group_2):
...
+
+    @AllowLargeResponse()
+    @ResourceGroupPreparer(parameter_name='resource_group_1')
+    @live_only()
+    def test_mysql_server_mgmt_public_paramter(self, resource_group_1):
+        self._test_server_mgmt_public_paramter('mysql', resource_group_1)
+
+    @AllowLargeResponse()
+    @ResourceGroupPreparer(parameter_name='resource_group_1')
+    @live_only()
```
Juliehzl, line 107: May I know the reason to make it live only here?

src/azure-cli/azure/cli/command_modules/rdbms/_util.py
arde0708: yes, sure. we can update that in the help message.

src/azure-cli/azure/cli/command_modules/rdbms/tests/latest/test_rdbms_commands.py
```diff
@@ -962,13 +1012,9 @@ def _test_replica_mgmt(self, resource_group, skuName, testSkuName, isBasicTier):
                           '--admin-user cloudsa --admin-password SecretPassword123 '
                           '--sku-name {}'
                           .format(database_engine, resource_group, server, skuName),
-                          checks=[
```
Juliehzl, line 965: May i know the reason decreasing properties to check?

src/azure-cli/azure/cli/command_modules/rdbms/custom.py
arde0708: Yes.

src/azure-cli/azure/cli/command_modules/rdbms/tests/latest/test_rdbms_commands.py
arde0708: This is because the firewall name consists of the timestamp( to keep it unique). Recording the test with non-deterministic value like the timestamp would fail the test always.

src/azure-cli/azure/cli/command_modules/rdbms/tests/latest/test_rdbms_commands.py
arde0708: Previously, we used to output all server properties to the user many of which were irrelevant. We have now reduced that to limited information which the user would find useful. replication capacity is one such field which added very little value to the user. So, we removed that from the create output. And since validating that parameter does not really help much, we have removed from validating that field from the replication tests as well.

src/azure-cli/azure/cli/command_modules/rdbms/_util.py
arde0708: Addressed

src/azure-cli/azure/cli/command_modules/rdbms/_util.py
arde0708: But, this is not exactly a validator. This is more of a common operation that multiple commands can use. So added it as an util function. The role of this function is to add a firewall rule based on the start IP and end IP. There is no validation that's happening in this function. It is responsible for creating the firewall rule based on the engine type. Do you still suggest it to be moved to the validator file ?

Juliehzl: I am fine for your current code. Just want to check if it is possible to have more clear code structure.
For the reason I recommend to use validator, it is because it looks like you are validating ip values with configuring firewalls.
Anyway I am fine.

src/azure-cli/azure/cli/command_modules/rdbms/custom.py
Juliehzl: How long does it take for the operation? Does customer need to wait for a long time?
Could we not wait for LRO complete?

src/azure-cli/azure/cli/command_modules/rdbms/custom.py
```diff
@@ -87,7 +140,39 @@ def _server_create(cmd, client, resource_group_name, server_name, sku_name, no_w
...
+
+    if engine_name == 'postgres':
+        return form_response(user, sku, loc, server_id, host, version,
+                             administrator_login_password if administrator_login_password is not None else '*****',
+                             create_postgresql_connection_string(server_name, host, user, administrator_login_password),
+                             infrastructure_encryption, None, firewall_id)
+    # Serves both - MySQL and MariaDB
+    # Create mysql database if it does not exist
+    database_name = DEFAULT_DB_NAME
+    create_database(cmd, resource_group_name, server_name, database_name, engine_name)
```
Juliehzl, line 171: For previous design, _server_create() is only used to create server, but here you add a following operation to create database, right?
Will it break some existing commands which use _server_create()?

src/azure-cli/azure/cli/command_modules/rdbms/custom.py
```diff
@@ -87,7 +140,39 @@ def _server_create(cmd, client, resource_group_name, server_name, sku_name, no_w
...
+    return form_response(user, sku, loc, server_id, host, version,
+                         administrator_login_password if administrator_login_password is not None else '*****',
```
Juliehzl, line 173: avoid positional arguments

src/azure-cli/azure/cli/command_modules/rdbms/custom.py
Juliehzl: table_transform is used to transform for table output. transform is used to format result(response), which is applied to all output format and table_transform also could utilize transformed result.

src/azure-cli/azure/cli/command_modules/rdbms/_util.py
```diff
@@ -35,3 +37,49 @@ def expand(self, dest, model_type, group_name=None, patches=None):
...
+    firewall_client = cf_postgres_firewall_rules(cmd.cli_ctx, None)
+    if db_engine == 'mysql':
+        firewall_client = cf_mysql_firewall_rules(cmd.cli_ctx, None)
+    elif db_engine == 'mariadb':
+        firewall_client = cf_mariadb_firewall_rules(cmd.cli_ctx, None)
+
+    firewall = firewall_client.create_or_update(resource_group_name, server_name, firewall_name, start_ip,
+                                                end_ip).result()
+    return firewall.name
```
arrownj, line 85: why not return the firewall object directly ?

arde0708: We **only** use the firewall name anyway.

src/azure-cli/azure/cli/command_modules/rdbms/custom.py
arde0708: Should not break anything as the new _server_create() **creates a server + creates a database.**.

src/azure-cli/azure/cli/command_modules/rdbms/custom.py
arde0708: It takes around 10 minutes for the operation to complete. Yes, the customer would need to wait otherwise if we return them the serverdetails w/o waiting for the operation to be complete, then they won't be be able to connect to the server if the server creation fails.

src/azure-cli/azure/cli/command_modules/rdbms/custom.py
arde0708: Changed to keyword arguments.

arrownj (approved):

src/azure-cli/azure/cli/command_modules/rdbms/custom.py
arde0708: I noticed that the other examples where transformers are leveraged is when the command functions returns the result of an API call directly. In our case, after the API call , there are a number of other operations that need to be completed based on the result of the API call before returning the result ( L: 146 - 161 ) in custom.py. Hence, I was thinking if it is okay to keep the current implementation unless there is an efficient way to do it even by moving it to the transformer. Thoughts ?

src/azure-cli/azure/cli/command_modules/rdbms/custom.py
arde0708: arrownj approved the PR. Hence, resolving this comment

Juliehzl (commented):
Overall, the PR looks good to me. But I have concern for the new response. Could you help confirm there is no breaking change in your new formatted response? 
If yes, I am fine.
And please take a look at conflict which blocking merge now.

arde0708:
> Overall, the PR looks good to me. But I have concern for the new response. Could you help confirm there is no breaking change in your new formatted response?
> If yes, I am fine.
> And please take a look at conflict which blocking merge now.

It will be a breaking change and I have updated the title to indicate that.

Juliehzl:
HI @arde0708, may I know which kind of breaking change involved? Because there are so many usages for `server create` command, could we make new response backward compatible? For example, appending new properties is fine but changing existing structure will have high risk for automation users.

arde0708:
> HI @arde0708, may I know which kind of breaking change involved? Because there are so many usages for `server create` command, could we make new response backward compatible? For example, appending new properties is fine but changing existing structure will have high risk for automation users.

@Juliehzl Reverted the breaking change.

Juliehzl (approved):

haroldrandom (approved):

PR #15550, opened 2020-10-16
{Core} az login: Capture managed_by_tenants error

jiasli:
**Description**  

If profile `latest` is used for Azure Stack cloud, `az login` will fail:

```powershell
> az cloud register -n redmond --endpoint-resource-manager "https://management.redmond.azurestack.corp.microsoft.com/" 
> az cloud set -n redmond
> az login
...
  File "d:\cli\azure-cli\src\azure-cli-core\azure\cli\core\_profile.py", line 276, in _normalize_properties
    subscription_dict[_MANAGED_BY_TENANTS] = [{_TENANT_ID: t.tenant_id} for t in s.managed_by_tenants]
TypeError: 'NoneType' object is not iterable
```

This is because `latest` should not be accepted by the service, and the service should return `Not Supported`. On the contrary, the service does accept the request and returns an invalid JSON which lacks `managedByTenants`. This behavior doesn't comply with the spec (https://github.com/Azure/azure-rest-api-specs/issues/9567), causing the parsing logic of CLI to fail.

Ideally, the user should run `az cloud set/update -n redmond --profile 2019-03-01-hybrid` to set the profile for the cloud before `az login`. 

This PR wraps the raw exception with better instructions:

```
UserFault: Invalid profile is used for cloud 'redmond'. To configure the cloud profile, 
run `az cloud set --name redmond --profile <profile>(e.g. 2019-03-01-hybrid)`. For more 
information about using Azure CLI with Azure Stack, see https://docs.microsoft.com/en-us
/azure/azure-stack/user/azure-stack-version-profiles-azurecli2#connect-to-azure-stack
```


src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -273,6 +273,16 @@ def _normalize_properties(self, user, subscriptions, is_service_principal, cert_
                 if hasattr(s, 'home_tenant_id'):
                     subscription_dict[_HOME_TENANT_ID] = s.home_tenant_id
                 if hasattr(s, 'managed_by_tenants'):
+                    if s.managed_by_tenants is None:
+                        # managedByTenants is missing from the response. This is a known service issue:
+                        # https://github.com/Azure/azure-rest-api-specs/issues/9567
+                        # pylint: disable=line-too-long
+                        from azure.cli.core.azclierror import UserFault
+                        raise UserFault("Invalid profile is used for cloud '{cloud_name}'. "
```
houk-ms, line 281: Can we define an error sub-class may be InvalidProfileError(UserFault) and raise the specific error? UserFault is designed as an abstract class and is not recommended to be raised directly.

jiasli: This is only a temporary mitigation to the service bug, it should actually be an `Unsupported API version` error.

yonzhan:
Core

arrownj (approved):

src/azure-cli-core/azure/cli/core/_profile.py
arrownj: @houk-ms , you can use `ABC` and `abstractmethod` (`from abc import ABC, abstractmethod`) to define your class and make the parent class can't be instantiated.

houk-ms: @arrownj sure, we can do that.

Juliehzl (approved):

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -273,6 +273,16 @@ def _normalize_properties(self, user, subscriptions, is_service_principal, cert_
...
+                                        "To configure the cloud profile, run `az cloud set --name {cloud_name} --profile <profile>(e.g. 2019-03-01-hybrid)`. "
+                                        "For more information about using Azure CLI with Azure Stack, see "
+                                        "https://docs.microsoft.com/en-us/azure/azure-stack/user/azure-stack-version-profiles-azurecli2#connect-to-azure-stack"
```
fengzhou-msft, line 284:
```suggestion
                                        "https://docs.microsoft.com/azure/azure-stack/user/azure-stack-version-profiles-azurecli2#connect-to-azure-stack"
```

src/azure-cli-core/azure/cli/core/_profile.py
houk-ms: @jiasli In this case, cosidering use `from azure.cli.core.azclierror import UnknownError` to replace UserFault if you don't define a new error type

fengzhou-msft (approved):

src/azure-cli-core/azure/cli/core/_profile.py
jiasli, line 281:
```suggestion
                        from azure.cli.core.azclierror import UnknownError
                        raise UnknownError("Invalid profile is used for cloud '{cloud_name}'. "
```

jiasli: Let's use `CLIError` for now until the error can be classified as `UnsupportedAPIError` after the service is fixed.

jiasli:
This PR has been shipped in Azure CLI 2.15.0.

PR #15553, opened 2020-10-16
{Cloud} Refactor patch method for endpoints when they are not added in ARM metadata yet

fengzhou-msft:
**Description**  

Use general patch methods to get fallback values from hard-coded clouds when the endpoint is not added to ARM metadata yet so that we don't need to keep adding new patch methods.

**Testing Guide**  

Unit test `test_metadata_url_endpoints` covers the test for the change.

**History Notes**  

---


yonzhan:
Cloud

bim-msft (approved):
Attestation and MHSM look good to me.

src/azure-cli-core/azure/cli/core/cloud.py
```diff
@@ -163,42 +163,6 @@ def __getattribute__(self, name):
         return val
 
 
-def _get_ossrdbms_resource_id(cloud_name):
-    ossrdbms_mapper = {
```
arrownj, line 167: Is it because all these endpoints are defined in the `HARD_CODED_CLOUD_LIST` so that we can remove them ?

fengzhou-msft: Yes. The patch methods are used when the endpoint is in `HARD_CODED_CLOUD_LIST` but not in ARM Metadata yet.

arrownj (approved):

src/azure-cli-core/azure/cli/core/cloud.py
```diff
@@ -209,58 +173,18 @@ def _concat_db_server_endpoint(db_prefix):
...
-        'AzureChinaCloud': 'https://api.applicationinsights.azure.cn',
-        'AzureUSGovernment': 'https://api.applicationinsights.us'
-    }
-    return app_insights_resource_id_mapper.get(cloud_name, None)
-
+def _get_endpoint_fallback_value(cloud_name):
+    def _get_cloud_endpoint_fallback_value(endpoint_name):
+        endpoint_mapper = {c.name: c.endpoints.__dict__.get(endpoint_name, None) for c in HARD_CODED_CLOUD_LIST}
+        return endpoint_mapper.get(cloud_name, None)
+    return _get_cloud_endpoint_fallback_value
```
jiasli, line 180: Is this nested function truly necessary? Can't it be simplified as `_get_endpoint_fallback_value(cloud_name, endpoint_name)`?

`_get_endpoint_fallback_value` does no more than saving `cloud_name` as `_get_cloud_endpoint_fallback_value`'s `locals()`.

See https://stackoverflow.com/questions/4020419/why-arent-python-nested-functions-called-closures

PR #14992, opened 2020-08-31
[Compute] Fix a vmss update, disk attach/detach, extension set/delete issue

qwordy:
**Description**  

Resolve https://github.com/Azure/azure-cli/issues/14924
Resolve https://github.com/Azure/azure-cli/issues/15058

Support partial update of vmss to avoid unnecessary permission check.

How to try it. https://github.com/Azure/azure-cli/blob/dev/doc/try_new_features_before_release.md

**Testing Guide**  

**History Notes**  

[Component Name 1] BREAKING CHANGE: az command a: Make some customer-facing breaking change.  
[Component Name 2] az command b: Add some customer-facing feature.

---

This checklist is used to make sure that common guidelines for a pull request are followed.

- [ ] The PR title and description has followed the guideline in [Submitting Pull Requests](https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules#submitting-pull-requests).

- [ ] I adhere to the [Command Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md).

PR #15575, opened 2020-10-19
[Compute] az vm update / az sig image-version update: Support update vm/image-version even it uses a cross tenant image

### LINKED ISSUE #15536: Can't update vm or image-version if reference to a cross tenant image

**Describe the bug**
Here’s a scenario that an internal customer, Lab Services, is having. They have a HOBO scenario.

1. Lab Services has a snapshot sitting in their own subscription.
2. Lab Services uses a service principal that has access to the Lab Services subscription and customers’ subscriptions.
3. The service principal creates a gallery, image definition and image version in the customer’s subscription. The image version is based on the snapshot in Lab Services’ subscription, so it is a cross-tenant scenario.
4. The customer then tries to update the replication details of the image version (az sig image-version update), but because the customer does not have access to Lab Services’ subscription, it fails.

Image version based on a resource in another tenant
- Customer tries to update the replica count
- Customer tries to add / remove a region on the image version

VM based on a SIG image in another tenant
- Customer tries to add a data disk
- I would argue that since it’s not allowed to change the imageReference.id of any VM, then, maybe all az vm update can omit the imageReference.id parameter.

**To Reproduce**

**Expected behavior**

**Environment summary**

**Additional context**

arrownj:
**Description**  
This PR is to fix #15536 

Previously, the user can't update a vm/image-version if the vm/image-version uses a cross tenant image which he/she has no permission to access.

The root cause is that CLI use the create_or_update API to in `update` command to support generic update. The service will check all the properties in the request body even the property value is not changed.

To fix this, we need to set the `storage_profile.image_reference` / `storage_profile.source` to None in the request body if the user is not updating the image property itself.

There is a similar PR #14992 which created by @qwordy .

**Testing Guide**  
Prepare a image, assume its id is <IMGAGE_ID>, and in <subscription_1>

update vm
`az account set -s <subscription_2>`
`az vm create -g <rg> -n <vm_name> --image <IMAGE_ID>`
> login with another user account who doesn't as <subscription_1> permission.

`az vm update -g <rg> -n <vm_name> hardwareProfile.vmSize=Standard_B2`

**History Notes**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/vm/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/custom.py
@@ -1239,6 +1239,7 @@ def show_vm(cmd, resource_group_name, vm_name, show_details=False):
 def get_vm_for_generic_update(cmd, resource_group_name, vm_name):
     client = _compute_client_factory(cmd.cli_ctx)
     vm = client.virtual_machines.get(resource_group_name, vm_name)
+    # To avoid unnecessary permission check of image
     vm.storage_profile.image_reference = None
     return vm
 
@@ -3291,9 +3292,11 @@ def fix_gallery_image_date_info(date_info):
     return date_info
 
 
+# pylint: disable=line-too-long
 def get_image_version_for_generic_update(cmd, resource_group_name, gallery_name, gallery_image_name, gallery_image_version_name):
     client = _compute_client_factory(cmd.cli_ctx)
     version = client.gallery_image_versions.get(resource_group_name, gallery_name, gallery_image_name, gallery_image_version_name)
+    # To avoid unnecessary permission check of image
     version.storage_profile.source = None
     return version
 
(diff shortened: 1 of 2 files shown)


yonzhan:
Compute

qwordy (approved):

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -1236,6 +1236,13 @@ def show_vm(cmd, resource_group_name, vm_name, show_details=False):
         else get_vm(cmd, resource_group_name, vm_name)
 
 
+def get_vm_for_generic_update(cmd, resource_group_name, vm_name):
```
qwordy, line 1239: Maybe you can add some comments like PR description here.

qwordy:
line too long error

PR #15593, opened 2020-10-21
{Compute} Apply UsageError to replace CLIError prefixed with 'Usage error'

houk-ms:
**Description**  

This PR uses the UsageError to replace the all the CLIErrors prefixed by 'Usage error: ' in vm command group, so that
1. the errors can be classified into userfaults
2. we can control whether to print error types in CLI Core instead of in each of the error messages

**Testing Guide**  

No difference from users' experience.

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/vm/_image_builder.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_image_builder.py
@@ -24,6 +24,7 @@ from knack.log import get_logger
 from msrestazure.azure_exceptions import CloudError
 from msrestazure.tools import is_valid_resource_id, resource_id, parse_resource_id
 
+from azure.cli.core.azclierror import UsageError
 from azure.cli.core.commands import cached_get, cached_put
 from azure.cli.core.commands.client_factory import get_subscription_id
 from azure.cli.core.commands.validators import get_default_location_from_resource_group, validate_tags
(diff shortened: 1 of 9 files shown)


yonzhan:
Besides Compute, other services will also do the similar changes, right?

yungezz (approved):

houk-ms:
> Besides Compute, other services will also do the similar changes, right?

Yes, exactly. This is the most simple case where we can take to replace the CLIErrors.

src/azure-cli/azure/cli/command_modules/vm/_image_builder.py
```diff
@@ -326,7 +327,7 @@ def process_img_tmpl_output_add_namespace(cmd, namespace):
     if len(outputs) != 1:
         err = "Supplied outputs: {}".format(outputs)
         logger.debug(err)
-        raise CLIError("Usage error: must supply exactly one destination type to add. Supplied {}".format(len(outputs)))
+        raise UsageError("must supply exactly one destination type to add. Supplied {}".format(len(outputs)))
```
qwordy, line 330:
```suggestion
        raise UsageError("Must supply exactly one destination type to add. Supplied {}".format(len(outputs)))
```

qwordy (changes requested):
Capitalize the first letter in all error messages.

src/azure-cli/azure/cli/command_modules/vm/_vm_utils.py
```diff
@@ -14,6 +14,8 @@
 from knack.log import get_logger
 from knack.util import CLIError
```
qwordy, line 15: Can we remove this line since it is not used any more?

houk-ms: good catch!

houk-ms:
> Capitalize the first letter in all error messages.
You're right, modified.

qwordy (approved):

yungezz:
hi @houk-ms will this change go into this release?

houk-ms:
> hi @houk-ms will this change go into this release?

I prefer not. 

Reason is I am not quite sure whether the changes are necessary. I mean for the existing `CLIError` in our codes, i don't have a clear plan about how to deal with them.

qwordy (approved):

PR #15608, opened 2020-10-22
{Telemetry} Add platform information and some config values

fengzhou-msft:
**Description**  

Currently, we only have platform system name in telemetry, i.e., `windows`, `linux`, `macos`. This PR adds more platform details in a new property in telemetry, e.g., `windows-10-10.0.19041-sp0`, `Linux-5.0.0-1027-azure-x86_64-with-Ubuntu-18.04-bionic`.

This PR also adds all keys in `az config set/unset` commands in telemetry. For `az config set`, if the section or key is in the allow list (mainly for switch type configs), values will also be recorded. We can better analyze users' preference with this data.

 
**Testing Guide**  

1. Run the following command and config keys and values are recorded.
   `az config set auto-upgrade.enable=no extension.run_after_dynamic_install=no`

    > "Context.Default.AzureCLI.debug_info":"ConfigSet: auto-upgrade.enable=no extension.run_after_dynamic_install=no"

2. Run the following command and only config keys are recorded.
    `az config set defaults.group=myGroup storage.account=myAccount`

    > "Context.Default.AzureCLI.debug_info":"ConfigSet: defaults.group=*** storage.account=***"

3. Run the following command and keys are recorded.
    `az config unset auto-upgrade.enable defaults.group`

    > "Context.Default.AzureCLI.debug_info":"ConfigUnset: auto-upgrade.enable defaults.group"

4. A new property `Context.Default.VS.Core.OS.Platform` is recorded for all comamnds.

**History Notes**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/config/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/config/custom.py
@@ -44,12 +44,8 @@ def config_set(cmd, key_value=None, local=False):
                 name = parts[1]
 
                 cmd.cli_ctx.config.set_value(section, name, _normalize_config_value(value))
-                if section in telemetry_allowed_sections_or_keys or key in telemetry_allowed_sections_or_keys:
-                    telemetry_content.append('{}={}'.format(key, value))
-                else:
-                    telemetry_content.append('{}=***'.format(key))
-            if telemetry_content:
-                telemetry.set_debug_info('ConfigSet', ' '.join(telemetry_content))
+                telemetry_content.append((key, section, value))
+            telemetry.set_debug_info('ConfigSet', telemetry_content)
 
 
 def config_get(cmd, key=None, local=False):
--- a/src/azure-cli/azure/cli/command_modules/config/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/config/custom.py
@@ -12,11 +12,6 @@ import azure.cli.core.telemetry as telemetry
 
 logger = get_logger(__name__)
 
-# Put a section or key (section.name) in the set to allow recording the config
-# values in the section or for the key
-telemetry_allowed_sections_or_keys = {'auto-upgrade', 'extension', 'core', 'logging.enable_log_file',
-                                      'output.show_survey_link'}
-
 
 def _normalize_config_value(value):
     if value:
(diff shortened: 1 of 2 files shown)


yonzhan:
allow list is a global whitelist setting maintained by us, right?

fengzhou-msft:
> allow list is a global whitelist setting maintained by us, right?

Yes. If the config value does not contain user-specific information, we can put it in the allow list to gain more insights on users' config preference.

src/azure-cli/azure/cli/command_modules/config/custom.py
```diff
@@ -36,6 +44,12 @@ def config_set(cmd, key_value=None, local=False):
                 name = parts[1]
 
                 cmd.cli_ctx.config.set_value(section, name, _normalize_config_value(value))
+                if section in telemetry_allowed_sections_or_keys or key in telemetry_allowed_sections_or_keys:
```
jiasli, line 47: I would suggest to put this filtering logic into telemetry module to make `config_set` as simple as possible.

fengzhou-msft: As this extra `debug_info` is command specific and in the future we may add other data for other modules. Putting it in telemetry module may need extra handling of each command module and less flexible.

yungezz:
hi @houk-ms could you pls review the PR thanks.

src/azure-cli/azure/cli/command_modules/config/custom.py
```diff
@@ -86,6 +91,7 @@ def config_unset(cmd, key=None, local=False):
 
         with ScopedConfig(cmd.cli_ctx.config, local):
             cmd.cli_ctx.config.remove_option(section, name)
+    telemetry.set_debug_info('ConfigUnset', ' '.join(key))
```
houk-ms, line 94: If user unset a property, why not just remove the property from Telemetry？

houk-ms (approved):

jiasli (approved):

src/azure-cli/azure/cli/command_modules/config/custom.py
fengzhou-msft: To record user behavior of the unset command.

Issue #13497, opened 2020-05-14
--tags option missing for "iot hub" and "iot dps"


cocowalla:
The `--tags` option appears to be missing from the `az iot hub create` and `az iot hub update` commands.

It's not [documented](https://docs.microsoft.com/en-us/cli/azure/iot?view=azure-cli-latest), and trying to use it results in an error `az: error: unrecognized arguments: --key=`.

I'd consider this a bug rather than a feature request, as every other create/update command seems to have a `--tags` option; `--tags` is like part of the create/update contract.

yonzhan:
iot

yiyiguo:
@Azure/azure-iot-cli-triage

c-ryan-k:
@cocowalla @yonzhan @yiyiguo `--tags` should be supported in update/create for both Hub and DPS

vilit1:
Closing the issue due to no response. If you still need help with this issue, please re-open.

PR #15648, opened 2020-10-23
[DPS] Allow tags in IoT DPS create command

anusapan:
**Description**  
Fix partial https://github.com/Azure/azure-cli/issues/13497

**Testing Guide**  

**History Notes**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/iot/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/iot/_params.py
@@ -10,7 +10,9 @@ from azure.cli.core.commands.parameters import (get_location_type,
                                                 file_type,
                                                 get_resource_name_completion_list,
                                                 get_enum_type,
-                                                get_three_state_flag)
+                                                get_three_state_flag,
+                                                tags_type)
+
 from azure.mgmt.iotcentral.models import AppSku
 from azure.mgmt.iothub.models import IotHubSku
 from azure.mgmt.iothubprovisioningservices.models import (IotDpsSku,
@@ -52,6 +54,7 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
     # Arguments for IoT DPS
     with self.argument_context('iot dps') as c:
         c.argument('dps_name', dps_name_type, options_list=['--name', '-n'], id_part='name')
+        c.argument('tags', tags_type)
 
     with self.argument_context('iot dps create') as c:
         c.argument('location', get_location_type(self.cli_ctx),
--- a/src/azure-cli/azure/cli/command_modules/iot/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/iot/custom.py
@@ -84,18 +84,21 @@ def iot_dps_get(client, dps_name, resource_group_name=None):
     return client.iot_dps_resource.get(dps_name, resource_group_name)
 
 
-def iot_dps_create(cmd, client, dps_name, resource_group_name, location=None, sku=IotDpsSku.s1.value, unit=1):
+def iot_dps_create(cmd, client, dps_name, resource_group_name, location=None, sku=IotDpsSku.s1.value, unit=1, tags=None):
     cli_ctx = cmd.cli_ctx
     _check_dps_name_availability(client.iot_dps_resource, dps_name)
     location = _ensure_location(cli_ctx, resource_group_name, location)
     dps_property = IotDpsPropertiesDescription()
     dps_description = ProvisioningServiceDescription(location=location,
                                                      properties=dps_property,
-                                                     sku=IotDpsSkuInfo(name=sku, capacity=unit))
+                                                     sku=IotDpsSkuInfo(name=sku, capacity=unit),
+                                                     tags=tags)
     return client.iot_dps_resource.create_or_update(resource_group_name, dps_name, dps_description)
 
 
-def iot_dps_update(client, dps_name, parameters, resource_group_name):
+def iot_dps_update(client, dps_name, parameters, resource_group_name, tags=None):
+    if tags is not None:
+        parameters.tags = tags
     return client.iot_dps_resource.create_or_update(resource_group_name, dps_name, parameters)
 
 
--- a/src/azure-cli/azure/cli/command_modules/iot/tests/latest/test_iot_dps_commands.py
+++ b/src/azure-cli/azure/cli/command_modules/iot/tests/latest/test_iot_dps_commands.py
@@ -23,10 +23,11 @@ class IoTDpsTest(ScenarioTest):
                          self.check('sku.name', 'S1')])
 
         # Create DPS
-        self.cmd('az iot dps create -g {} -n {}'.format(group_name, dps_name), checks=[
-            self.check('name', dps_name),
-            self.check('location', group_location)
-        ])
+        tags = "key1=value1 key2=value2"
+        self.cmd('az iot dps create -g {} -n {} --tags {}'.format(group_name, dps_name, tags),
+                 checks=[self.check('name', dps_name),
+                         self.check('location', group_location),
+                         self.check('tags', {'key1': 'value1', 'key2': 'value2'})])
 
         # List DPS
         self.cmd('az iot dps list -g {}'.format(group_name), checks=[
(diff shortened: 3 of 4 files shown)


yonzhan:
IotHub

digimaun (approved):

zhoxing-ms:
Could this structure of tags be consistent with the tags of other CLI commands?
For example:
```
az resource tag --ids {resource-id} --tags key1=value1 key2=value2 key3=value3
```
This tag verification can be referred to https://github.com/zhoxing-ms/azure-cli/blob/master/src/azure-cli/azure/cli/command_modules/resource/_params.py#L57

src/azure-cli/azure/cli/command_modules/iot/_params.py
```diff
@@ -59,6 +59,11 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('sku', arg_type=get_enum_type(IotDpsSku),
                    help='Pricing tier for the IoT provisioning service.')
         c.argument('unit', help='Units in your IoT Provisioning Service.', type=int)
+        c.argument(
+            "tags",
+            options_list=["--tags"],
+            help="IoT dps instance tags. Property bag in key-value pairs with the following format: a=b;c=d."
```
zhoxing-ms, line 65: Could this structure of tags be consistent with the tags of other CLI commands?
For example:
```
az resource tag --ids {resource-id} --tags key1=value1 key2=value2 key3=value3
```
This tag verification can be referred to [code link](https://github.com/Azure/azure-cli/blob/master/src/azure-cli/azure/cli/command_modules/resource/_params.py#L101)

digimaun: Hmm I guess we aren't aligned on the tags input pattern since IoT hasn't had tags support in the CLI until recently and with the first IoT service for tags upon creation we went in [this direction](https://docs.microsoft.com/en-us/cli/azure/ext/azure-iot/dt?view=azure-cli-latest#ext_azure_iot_az_dt_create-examples).

digimaun: Ok, we will conform the --tags to be consistent.

digimaun (changes requested):
Make `--tags` consistent and use tags_type from azure.cli.core.commands.parameters

src/azure-cli/azure/cli/command_modules/iot/_params.py
anusapan: Fixed this

src/azure-cli/azure/cli/command_modules/iot/tests/latest/test_iot_dps_commands.py
```diff
@@ -23,10 +23,13 @@ def test_dps_lifecycle(self, group_name, group_location):
...
-        self.cmd('az iot dps create -g {} -n {}'.format(group_name, dps_name), checks=[
+        tags = "key1=value1 key2=value2"
+        dps = self.cmd('az iot dps create -g {} -n {} --tags {}'.format(group_name, dps_name, tags), checks=[
             self.check('name', dps_name),
-            self.check('location', group_location)
-        ])
+            self.check('location', group_location),
+        ]).get_output_in_json()
+
+        assert len(dps['tags']) == 2
```
zhoxing-ms, line 32: Can we check the value of tags instead of checking the number of tags?
Such as:
```
self.check('tags', {'key1': 'value1', 'key2': 'value2'})
```

src/azure-cli/azure/cli/command_modules/iot/tests/latest/test_iot_dps_commands.py
```diff
@@ -43,12 +46,15 @@ def test_dps_lifecycle(self, group_name, group_location):
...
-            self.check('location', group_location),
-            self.check(property_to_update, updated_value)
-        ])
+        updated_dps = self.cmd('az iot dps update -g {} -n {} --tags {} --set {}="{}"'
+                               .format(group_name, dps_name, updated_tags, property_to_update, updated_value),
+                               checks=[self.check('name', dps_name),
+                                       self.check('location', group_location),
+                                       self.check(property_to_update, updated_value)]).get_output_in_json()
+
+        assert len(updated_dps['tags']) == 3
```
zhoxing-ms, line 57: Is updating tags an incremental operation rather than a full coverage operation?

src/azure-cli/azure/cli/command_modules/iot/tests/latest/test_iot_dps_commands.py
anusapan: Fixed

src/azure-cli/azure/cli/command_modules/iot/tests/latest/test_iot_dps_commands.py
anusapan: Yes, Update is an incremental operation

digimaun (approved):

src/azure-cli/azure/cli/command_modules/iot/tests/latest/test_iot_dps_commands.py
zhoxing-ms: For other RP, however, modifying the tags is generally a full coverage operation. Is it necessary to consider consistency with other RPS?
What's more, if a customer wants to delete some tags, what should he do?

src/azure-cli/azure/cli/command_modules/iot/tests/latest/test_iot_dps_commands.py
```diff
@@ -23,10 +23,14 @@ def test_dps_lifecycle(self, group_name, group_location):
...
+        tags = "key1=value1 key2=value2"
+        dps = self.cmd('az iot dps create -g {} -n {} --tags {}'.format(group_name, dps_name, tags), checks=[
             self.check('name', dps_name),
-            self.check('location', group_location)
-        ])
+            self.check('location', group_location),
+            self.check('tags', {'key1': 'value1', 'key2': 'value2'})
+        ]).get_output_in_json()
+
+        assert len(dps['tags']) == 2
```
zhoxing-ms, line 33: Because the tag itself is checked, you can delete this check
```suggestion
```

src/azure-cli/azure/cli/command_modules/iot/tests/latest/test_iot_dps_commands.py
digimaun: Let's add a test which covers removing tags as well.

src/azure-cli/azure/cli/command_modules/iot/tests/latest/test_iot_dps_commands.py
anusapan: Fixed

src/azure-cli/azure/cli/command_modules/iot/tests/latest/test_iot_dps_commands.py
anusapan: Added a scenario for deleting the tags

digimaun: Thank you @anusapan. I think @zhoxing-ms is suggesting most other RP's CLI wise are a full coverage operation for the `--tags` argument. To support a consistent experience for users (at least for common shared attributes) it makes sense for us to go in this direction as well.

anusapan: According to the help content for parameter "--tags", use "" to delete the existing tags.
image

src/azure-cli/azure/cli/command_modules/iot/custom.py
```diff
@@ -47,7 +47,7 @@
 from azure.cli.command_modules.iot.shared import EndpointType, EncodingFormat, RenewKeyType, AuthenticationType
 from ._client_factory import resource_service_factory
 from ._utils import open_certificate, generateKey
-
+import pdb
```
digimaun, line 50: remove import for debugging

src/azure-cli/azure/cli/command_modules/iot/tests/latest/test_iot_dps_commands.py
digimaun: Yea for sure. But my comment is not referring to the delete convention specifically. It's referring to how the current state of this PR supports an incremental update to the tags dictionary `parameters.tags.update(tags)` where-as most other RP's in the CLI don't work this way (zhoxing-ms comment earlier). So to be consistent we should assign the tags attribute every time. If the input is `""` the tags param changes the value to `{}` thereby resetting existing tags.

digimaun (changes requested):
Let's work through these last comments.

Also the change on this PR is for DPS so we should update the title.

src/azure-cli/azure/cli/command_modules/iot/custom.py
anusapan: fixed

src/azure-cli/azure/cli/command_modules/iot/tests/latest/test_iot_dps_commands.py
anusapan: Updated the PR, Now there is no incremental update for the tags, if passing tags parameter value then it will overwrite the existing tags.

zhoxing-ms (approved):
LGTM

digimaun (approved):

Issue #15346, opened 2020-09-29
az deployment group create error message regressed in 2.12.1


jsacapdev:
## Describe the bug

**Command Name**

`az deployment group create`

**Errors:**

```
Command arguments: ['deployment', 'group', 'create', '--name', 'd9764fa9-8de7-4fab-b628-a86f1ab29b25', '--resource-group', 'euw-apipt-dev-rg-03', '--template-file', './azuredeploy.json', '--debug']
Event: Cli.PreExecute []
Event: CommandParser.OnGlobalArgumentsCreate [<function CLILogging.on_global_arguments at 0x7f4ff4e4c950>, <function OutputProducer.on_global_arguments at 0x7f4ff499d6a8>, <function CLIQuery.on_global_arguments at 0x7f4ff4730950>]
Event: CommandInvoker.OnPreCommandTableCreate []
Modules found from index for 'deployment': ['azure.cli.command_modules.resource']
Loading command modules:
Name                  Load Time    Groups  Commands
resource                  0.030        34       158
Total (1)                 0.030        34       158
Loaded 34 groups, 158 commands.
Found a match in the command table.
Raw command  : deployment group create
Command table: deployment group create
Event: CommandInvoker.OnPreCommandTableTruncate [<function AzCliLogging.init_command_file_logging at 0x7f4ff3bb86a8>]
Event: CommandInvoker.OnPreArgumentLoad [<function register_global_subscription_argument.<locals>.add_subscription_parameter at 0x7f4ff3b2fea0>, <function register_global_query_examples_argument.<locals>.register_query_examples at 0x7f4ff3b3dd90>]
Event: CommandInvoker.OnPostArgumentLoad []
Event: CommandInvoker.OnPostCommandTableCreate [<function register_ids_argument.<locals>.add_ids_arguments at 0x7f4ff3b3de18>, <function register_cache_arguments.<locals>.add_cache_arguments at 0x7f4ff3b2fd08>]
Event: CommandInvoker.OnCommandTableLoaded []
Event: CommandInvoker.OnPreParseArgs []
[...]

PR #15658, opened 2020-10-26
[ARM] az deployment: Print out error details for deployment

zhoxing-ms:
Fix: #15346

Since the api-version of resource was upgraded to `2020-06-01`, `validate` became a poller operation, such as:
```
validation_poller = mgmt_client.validate(resource_group_name, deployment_name, deployment)
```
For ARMPolling, CloudError will be thrown based on the Http status code
```
class ARMPolling(PollingMethod):

    ...

    def initialize(self, client, initial_response, deserialization_callback):
        """Set the initial status of this LRO.

        :param initial_response: The initial response of the poller
        :raises: CloudError if initial status is incorrect LRO state
        """
        self._client = client
        self._response = initial_response
        self._operation = LongRunningOperation(initial_response, deserialization_callback, self._lro_options)
        try:
            self._operation.set_initial_status(initial_response)
        except BadStatus:
            self._operation.status = 'Failed'
            raise CloudError(initial_response)
        except BadResponse as err:
            self._operation.status = 'Failed'
            raise CloudError(initial_response, str(err))
        except OperationFailed:
            raise CloudError(initial_response)
```
Those error message of CloudError can't be resolved in depth:
```
The template deployment 'validate_error_template' is not valid according to the validation procedure. The tracking id is 'e7a1400a-e7a8-4ceb-9718-73ec7b263e77'. See inner errors for details.
``` 
So there is a layer of wrapper resolution for CloudError when it appears, the effect is as follows:
```
{"error":{"code":"InvalidTemplateDeployment","message":"The template deployment 'validate_error_template' is not valid according to the validation procedure. The tracking id is 'f017fe6b-7d53-4974-a2dc-ac1802324047'. See inner errors for details.","details":[{"code":"ValidationForResourceFailed","message":"Validation failed for a resource. Check 'Error.Details[0]' for more information.","details":[{"code":"InvalidSkuName","message":"Invalid skuName."}]}]}}
```

Fix: #14978
Subdivide the error types of CLIError for deployment

Two new error types are added:
- Invalid template error type
```
class InvalidTemplateError(UserFault):
    """ Incorrect deployment contents, such as incorrect template files or incorrect deployment parameters """
    pass
```
- Deployment error type (Used when a subdivision error type cannot be distinguished)
```
class DeploymentError(UserFault):
    """ Deployment fail error """
    pass
```


yonzhan:
ARM

houk-ms (commented):
Overall good to me. Leave some suggestions.

Remember update the new error types to [error handling guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/error_handling_guidelines.md).

zhoxing-ms:
> Overall good to me. Leave some suggestions.
> 
> Remember update the new error types to [error handling guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/error_handling_guidelines.md).

OK, thank you for the reminder

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -689,9 +714,13 @@ def _deploy_arm_template_at_tenant_scope(cmd,
     mgmt_client = _get_deployment_management_client(cmd.cli_ctx, plug_pipeline=(template_uri is None and template_spec is None))
 
     if cmd.supported_api_version(min_api='2019-10-01', resource_type=ResourceType.MGMT_RESOURCE_RESOURCES):
+        from msrestazure.azure_exceptions import CloudError
         ScopedDeployment = cmd.get_models('ScopedDeployment')
         deployment = ScopedDeployment(properties=deployment_properties, location=deployment_location)
-        validation_poller = mgmt_client.validate_at_tenant_scope(deployment_name=deployment_name, parameters=deployment)
+        try:
+            validation_poller = mgmt_client.validate_at_tenant_scope(deployment_name=deployment_name, parameters=deployment)
+        except CloudError as cx:
```
Juliehzl, line 722: could you still catch it when track2 SDK is adopted?

Juliehzl:
if possible, please refine the error message to make it actionable.

src/azure-cli/azure/cli/command_modules/resource/custom.py
zhoxing-ms: I guess catch the `HttpResponseError` may be needed, because the processing of `status==400` by the track 2 SDK still does not throw an exception, but returns the result of `DeploymentValidateResult`.  [code link](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/resources/azure-mgmt-resource/azure/mgmt/resource/resources/v2020_06_01/operations/_deployments_operations.py#L1158)

Then, HttpResponseError may be thrown later because ARMPolling (from azure.mgmt.core) is called later:
```
def _raise_if_bad_http_status_and_method(response):
    # type: (ResponseType) -> None
    """Check response status code is valid.

    Must be 200, 201, 202, or 204.

    :raises: BadStatus if invalid status.
    """
    code = response.status_code
    if code in {200, 201, 202, 204}:
        return
    raise BadStatus(
        "Invalid return status {!r} for {!r} operation".format(
            code, response.request.method
        )
    )

def initialize(self, client, initial_response, deserialization_callback):

    ...

    try:
        _raise_if_bad_http_status_and_method(self._initial_response.http_response)
        self._status = self._operation.set_initial_status(initial_response)

    except BadStatus as err:
        self._status = "Failed"
        raise HttpResponseError(response=initial_response.http_response, error=err)
```

I will verify it when I migrate track 2 later. If there is still this problem, we can consider giving suggestions to the Python SDK of track 2 to throw exceptions in this case of `status==400`, right?

zhoxing-ms:
> if possible, please refine the error message to make it actionable.

Because many different errors will be returned here, their error information need to be parsed and redefined according to different situations, so it's not easy to implement and needs some effort~
If there are some user feedback that the error messages of some scenarios are not friendly enough in the future, I will consider redefining the error message for these scenarios~

Do you have any suggestions for this?

jsntcy (approved):

src/azure-cli/azure/cli/command_modules/resource/custom.py
Juliehzl: If you have tests about the exception type, it will be easy to catch the change.

Juliehzl:
> > if possible, please refine the error message to make it actionable.
> 
> Because many different errors will be returned here, their error information need to be parsed and redefined according to different situations, so it's not easy to implement and needs some effort~
> If there are some user feedback that the error messages of some scenarios are not friendly enough in the future, I will consider redefining the error message for these scenarios~
> 
> Do you have any suggestions for this?

besides invalidParameter info in response, consider also adding the related parameter name like `--template` or `--parameter`. but it is not big issue, you could also refine later

src/azure-cli/azure/cli/command_modules/resource/custom.py
zhoxing-ms: Yes, however there are some breaking changes during the initialization of `ResourceManagementClient` when migrating to track 2, which needs further investigation~

houk-ms (approved):

Juliehzl (approved):

jiasli (approved):

PR #15695, opened 2020-10-28
[AppService] functionapp: Added warning if customer tries to switch functionapp from premium to consumption.

### LINKED ISSUE #15694: Add warning if customer tries to switch from premium to consumption functionapp
The command `az functionapp update --plan` gives users the capability to switch a functionapps plan for a different, working app service plan. This could potentially be dangerous if the function is using premium features. This scenario should be handled with at least a warning.

gzuber:
Resolves #15694

**Description**  

This PR involves changes to the `az functionapp update` command:

We're now requiring a user specify a `--force` flag to their command if they're attempting to migrate a functionapp from Premium -> Consumption. This is a potentially dangerous action if their functionapp uses Premium features that are not available in consumption. 

This also adds an error message if the user tries to migrate a functionapp to or from a Linux plan. Migration will only work on Windows at this time.

I also changed the exception handling slightly to add a helpful message if a migration attempt fails from the backend letting the customer know that some plans are incapable of upgrading.

**Testing Guide**  

Most of these changes are warnings and error messages. The rest of these changes will either be tested by the move/migration tests we already have. 

**History Notes**  

[AppService] BREAKING CHANGE: az functionapp update: Migrating a functionapp from Premium to Consumption plans now requires the '--force' flag.
[AppService] az functionapp update: Added error message if functionapp migration involves any plans on Linux.
[AppService] az functionapp update: Added more descriptive error message if functionapp migration fails.

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/commands.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/commands.py
@@ -40,6 +40,7 @@ def ex_handler_factory(creating_plan=False):
 
 def update_function_ex_handler_factory():
     from knack.util import CLIError
+
     def _ex_handler(ex):
         http_error_response = False
         if hasattr(ex, 'response'):
@@ -49,13 +50,13 @@ def update_function_ex_handler_factory():
         if http_error_response:
             try:
                 detail = ('If using \'--plan\', a consumption plan may be unable to migrate '
-                        'to a given premium plan. Please confirm that the premium plan '
-                        'exists in the same resource group and region. Note: Not all '
-                        'functionapp plans support premium instances. If you have verified '
-                        'your resource group and region and are still unable to migrate, '
-                        'please redeploy on a premium functionapp plan.')
+                          'to a given premium plan. Please confirm that the premium plan '
+                          'exists in the same resource group and region. Note: Not all '
+                          'functionapp plans support premium instances. If you have verified '
+                          'your resource group and region and are still unable to migrate, '
+                          'please redeploy on a premium functionapp plan.')
                 ex = CLIError(ex.args[0] + '\n\n' + detail)
-            except Exception: # pylint: disable=broad-except
+            except Exception:  # pylint: disable=broad-except
                 pass
         raise ex
     return _ex_handler
--- a/src/azure-cli/azure/cli/command_modules/appservice/commands.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/commands.py
@@ -32,28 +32,56 @@ def transform_web_list_output(webs):
 
 
 def ex_handler_factory(creating_plan=False):
-    def _polish_bad_errors(ex):
-        import json
-        from knack.util import CLIError
-        try:
-            if 'text/plain' in ex.response.headers['Content-Type']:  # HTML Response
-                detail = ex.response.text
-            else:
-                detail = json.loads(ex.response.text)['Message']
-                if creating_plan:
-                    if 'Requested features are not supported in region' in detail:
-                        detail = ("Plan with linux worker is not supported in current region. For " +
-                                  "supported regions, please refer to https://docs.microsoft.com/"
-                                  "azure/app-service-web/app-service-linux-intro")
-                    elif 'Not enough available reserved instance servers to satisfy' in detail:
-                        detail = ("Plan with Linux worker can only be created in a group " +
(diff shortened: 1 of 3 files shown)


gzuber:
@cachai2 for help with wording the warning message. Part of me wonders if this is enough. It'll warn the user but too late for them to abort. I could see us adding warnings to the help menu or event making the user provide a `--force` flag or something like that. That would be a breaking change but considering how dangerous this is and how rarely it's probably used it might not be a big deal.

yonzhan:
AppService

cachai2:
@gzuber thanks for working on this. I'll take a look and help out with wording as well to understand if we need to add more to ensure users have what they need. 

Also, is there any messaging for consumption to premium migration in cases where it might fail? We should provide some messaging in these cases as well to make sure customers aren't in the dark if their migration doesn't work.

cachai2:
@gzuber I like most of the wording you had, and added a few updates. Please let me know your thoughts on the below for premium to consumption migration:

-----------
"Warning: Moving a functionapp from Premium to Consumption may result in loss of functionality and cause the app to break. 
Please ensure the functionapp is compatible with a Consumption plan and is not using any features only available in Premium.

"Do you still want to migrate the functionapp to a consumption plan?" yes/no
-----------

- I think it would be good to have a proactive check to allow users to decline the migration after seeing the error message but before the actual migration to give them the opportunity to verify. 
- I was also thinking about whether or not to include a link to docs. However, it could be a bit dangerous if anyone does the migration and something breaks that isn't in the docs, so it might be best to leave it out. Let me know your thoughts.

yungezz:
hi @qwordy could you pls help to review? thanks

qwordy (approved):

gzuber:
Thanks for reviewing! Just as an FYI I still have some changes to get in for this PR, hoping to get those pushed soon

src/azure-cli/azure/cli/command_modules/appservice/commands.py
```diff
@@ -32,28 +32,57 @@ def transform_web_list_output(webs):
...
+                              "supported regions, please refer to https://docs.microsoft.com/"
+                              "azure/app-service-web/app-service-linux-intro")
+                elif 'Not enough available reserved instance servers to satisfy' in detail:
+                    detail = ("Plan with Linux worker can only be created in a group " +
+                              "which has never contained a Windows worker, and vice versa. " +
+                              "Please use a new resource group. Original error:" + detail)
+        ex = CLIError(detail)
+    except Exception:  # pylint: disable=broad-except
+        pass
+    return ex
```
gzuber, line 86: My reasoning for refactoring this is because I wanted to take advantage of `_polish_bad_errors` because `az functionapp update` has been using this exception handler. However, I needed to add an additional message in case of an exception but not to all the other commands that use this handler. I believe this should preserve all the other commands' exception handling behavior.

gzuber:
@qwordy/@yungezz this should be ready for any additional review and merge (I added a few things), thanks!

gzuber:
@qwordy / @yungezz this should be ready for any additional review and merge (I added a few things), thanks!

qwordy:
@gzuber Error handling guidelines are updated. CLIError is not recommended to used any more. We suggest use more concrete error type. Thank you for your work.
https://github.com/Azure/azure-cli/blob/dev/doc/error_handling_guidelines.md

gzuber:
@qwordy I changed the Errors for the only command I changed with this PR (`az functionapp update`). I didn't change `ex_handler_factory` because that would change the behavior of most of the commands in appservice. I think it's something we need to do, but I think that should be done in its own separate PR since it's such an extensive change.

Is there any chance we can merge this in time for [S178](https://github.com/Azure/azure-cli/milestone/96)? Please let me know if there are any more changes I need to make on my end, thanks!

qwordy:
> 
> 
> @qwordy I changed the Errors for the only command I changed with this PR (`az functionapp update`). I didn't change `ex_handler_factory` because that would change the behavior of most of the commands in appservice. I think it's something we need to do, but I think that should be done in its own separate PR since it's such an extensive change.
> 
> Is there any chance we can merge this in time for [S178](https://github.com/Azure/azure-cli/milestone/96)? Please let me know if there are any more changes I need to make on my end, thanks!

Thank you for your quick response. It is a tough job to update all existing code. We only require new code to raise more concrete error type now. I'll merge it.

PR #15184, opened 2020-09-16
{Core} Honor scopes specified by Track 2 SDK

jiasli:
Fix #15179

## Issue

In 
- Track 1 SDK, `scopes` (`resource`) is managed by Azure CLI
- Track 2 SDK, `scopes` (`resource`) is managed by SDK

For example, for Track 2 SDK mgmt-plane [`SubscriptionClientConfiguration`](https://github.com/Azure/azure-cli-extensions/blob/120196e953f7b776fa209f4c49838b5cc7af56b0/src/account/azext_account/vendored_sdks/subscription/_configuration.py#L44):

```py
self.credential_scopes = ['https://management.azure.com/.default']
```

For Track 2 SDK data-plane [`AzureAppConfigurationClient`](https://github.com/Azure/azure-sdk-for-python/blob/022bc80f2bc645d8d82d647620376fcea2229522/sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_azure_appconfiguration_client.py#L128):

```py
if aad_mode:
    scope = base_url.strip("/") + "/.default"
```

But in `AdalAuthentication.get_token`, `scopes` is not honored (discarded), resulting in getting a token for a wrong `scopes` (ARM `https://management.core.windows.net/` by default):

https://github.com/Azure/azure-cli/blob/c2b9a2c5cfe34b108b52b49ac50c6802ac4d729f/src/azure-cli-core/azure/cli/core/adal_authentication.py#L61-L67

## Fix

This PR fixes the issue by honoring the `scopes` specified by Track 2 SDK.

### User Identity

```sh
# Track 1 SDK mgmt-plan
> az group show -n fy --debug
DEBUG: AdalAuthentication.signed_session invoked by Track 1 SDK

# Track 2 SDK mgmt-plane
> az account subscription show --id 0b1f6471-1bf0-4dda-aec3-cb9272f09590 --debug
DEBUG: AdalAuthentication.get_token invoked by Track 2 SDK with scopes=('https://management.azure.com/.default',)

# Track 1 SDK data-plane
> az account subscription show --id 0b1f6471-1bf0-4dda-aec3-cb9272f09590 --debug
DEBUG: Profile.get_raw_token invoked with resource='https://storage.azure.com', subscription=None, tenant=None

# Track 2 SDK data-plane
az storage blob list --account-name st0921 -c con1 --auth-mode login --debug
DEBUG: AdalAuthentication.get_token invoked by Track 2 SDK with scopes=('https://storage.azure.com/.default',)
[...]

PR #15698, opened 2020-10-28
{Core} Do not append/strip trailing slash in scope

jiasli:
**Description**  

Fix incorrect logic of `resource_to_scopes` and `scopes_to_resource` introduced by #15184.

According to [_Trailing slash and /.default_](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-permissions-and-consent#trailing-slash-and-default),

- For some services, the trailing slash should **not** appear in the resource URI, like `https://managedhsm.azure.com`, and the corresponding `scope` should be `https://managedhsm.azure.com/.default`
- On the other hand, for some other services, the trailing slash is mandatory, like `https://management.azure.com/`, and the corresponding `scope` should be `https://management.azure.com//.default` - note the double slash!

This PR changes the behavior so that no extra logic is performed on the trailing slash - CLI only appends or strips `/.default` with no fancy logic.

### DIFF
--- a/src/azure-cli-core/azure/cli/core/tests/test_util.py
+++ b/src/azure-cli-core/azure/cli/core/tests/test_util.py
@@ -380,25 +380,37 @@ class TestUtils(unittest.TestCase):
     def test_scopes_to_resource(self):
         from azure.cli.core.util import scopes_to_resource
         # scopes as a list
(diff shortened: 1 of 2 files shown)


yonzhan:
Core

arrownj (approved):
👍

bim-msft (approved):

jsntcy (approved):

PR #15669, opened 2020-10-27
{Core} Migrate azure-cli-core to use azure-mgmt-resource Track 2 SDK

jiasli:
**Description**  

Migrate `azure-cli-core` to use `azure-mgmt-resource` Track 2 SDK.

**Testing Guide**  

```sh
az login --debug

# In sovereign cloud
az cloud set -n AzureChinaCloud
az login --debug

# In Azure Stack
az cloud register -n redmond --endpoint-resource-manager "https://management.redmond.azurestack.corp.microsoft.com/" 
az cloud set -n redmond --profile 2019-03-01-hybrid
az login --debug
```

Issue #15530, opened 2020-10-15
Bump azure-mgmt-resource to Track 2 15.0.0


jiasli:
**Is your feature request related to a problem? Please describe.**

Due to the code change in Track 2 SDK, if a user uses `pip` to install Azure CLI and `azure-mgmt-resource==15.0.0` in the same virtual environment. CLI will stop functioning correctly:

```pwsh
> az login
...
AttributeError: 'SubscriptionClient' object has no attribute 'config'
```

More info: https://github.com/Azure/azure-cli/issues/13995

**Describe the solution you'd like**

Bump the dependency of `azure-mgmt-resource` to Track 2 15.0.0.

**Describe alternatives you've considered**

Install Azure CLI and `azure-mgmt-resource==15.0.0` in separate environments. 

**Additional context**

See #15461, #15429, #14716, #13995, #15460, #13995.

yonzhan:
Bump azure-mgmt-resource

glaubitz:
Is there any preliminary patch for this issue yet?

jiasli:
We are currently working on it, but this migration requires huge amount of effort. 

We will vendor a subscription SDK in `azure-cli-core` temporarily (#15711) in the next sprint so that `azure-cli-core` doesn't have hard dependency on `azure-mgmt-resource`, but the migration on command modules will take several sprints.

Thank you for you understanding.

glaubitz:
OK, thanks a lot for the explanation. I understand the necessary effort, I just wanted to get a rough estimate so that we can inform our users at SUSE.

Thanks!

glaubitz:
It seems that there is another incompatibility with ```azure-mgmt-resource``` with ```azure-cli-core```:

```
glaubitz@suse-laptop:~> az resource delete adriantest
UnrecognizedArgumentError: unrecognized arguments: adriantest
Try this: 'az resource delete --ids </subscriptions/0b1f6471-1bf0-4dda-aec3-111111111111/resourceGroups/MyResourceGroup/providers/Microsoft.Web/sites/MyWebapp>'
Still stuck? Run 'az --help' to view all commands or go to 'https://aka.ms/cli_ref' to learn more
glaubitz@suse-laptop:~> az group delete --name adriantest
CLIInternalError: The command failed with an unexpected error. Here is the traceback:
The operation 'azure.mgmt.resource.resources.v2020_06_01.operations#ResourceGroupsOperations.delete' is invalid.
Traceback (most recent call last):
[...]

PR #15711, opened 2020-10-29
{Core} Vendor track 2 subscription SDK in azure-cli-core

jiasli:
**Description**  
Implement issue #15530: Bump azure-mgmt-resource to Track 2 15.0.0
Replace PR #15669: {Core} Migrate `azure-cli-core` to use `azure-mgmt-resource` Track 2 SDK

Vendor track 2 subscription SDK in `azure-cli-core` so that it doesn't have dependency on `azure-mgmt-resource`, which is required by [`resource`](https://github.com/Azure/azure-cli/blob/dc6500f06032ea0486aa5af1f7efa00a7f94644e/src/azure-cli/azure/cli/command_modules/resource) command module.

In this way, `azure-cli-core` can start using Track 2 `azure-mgmt-resource` SDK before `resource` finishes the migration.

To switch back to using public SDK, simply change `_USE_VENDEROED_SUBSCRIPTION_SDK` to `False`.


yonzhan:
Core

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -303,11 +305,18 @@ def _build_tenant_level_accounts(self, tenants):
...
+            from azure.cli.core.vendored_sdks.subscriptions.models import Subscription
+            SubscriptionType = Subscription
+        else:
+            from azure.cli.core.profiles import ResourceType, get_sdk
+            SubscriptionType, StateType = get_sdk(self.cli_ctx, ResourceType.MGMT_RESOURCE_SUBSCRIPTIONS,
+                                                  'Subscription',
+                                                  'SubscriptionState', mod='models')
         s = SubscriptionType()
-        s.state = StateType.enabled
+        s.state = 'Enabled'
```
jiasli, line 319: Track 2 uses `str` instead of `Enum` as the value when deserializing objects, so we do the same. 

See email: _A breaking change in Enum property's deserialization_

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -449,8 +458,7 @@ def _match_account(account, subscription_id, secondary_key_name, secondary_key_v
 
     @staticmethod
     def _pick_working_subscription(subscriptions):
-        from azure.mgmt.resource.subscriptions.models import SubscriptionState
-        s = next((x for x in subscriptions if x.get(_STATE) == SubscriptionState.enabled.value), None)
+        s = next((x for x in subscriptions if x.get(_STATE) == 'Enabled'), None)
```
jiasli, line 461: Same as https://github.com/Azure/azure-cli/pull/15711/files#r523861234

src/azure-cli-core/setup.py
```diff
@@ -84,12 +86,7 @@
     url='https://github.com/Azure/azure-cli',
     zip_safe=False,
     classifiers=CLASSIFIERS,
-    packages=[
-        'azure.cli.core',
-        'azure.cli.core.commands',
-        'azure.cli.core.extension',
-        'azure.cli.core.profiles',
-    ],
+    packages=find_packages(exclude=["*.tests", "*.tests.*", "tests.*", "tests", "azure", "azure.cli"]),
```
jiasli, line 89: Change the package finding logic so that `vendored_sdks/` are included. Thank @arrownj for the help.

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -77,6 +77,8 @@
 
 _AZ_LOGIN_MESSAGE = "Please run 'az login' to setup account."
 
+_USE_VENDORED_SUBSCRIPTION_SDK = True
```
arrownj, line 80: Do we need to change this to `False` in the future ?

jiasli: Probably not. `azure-cli-core` uses only very limited functionalities from this SDK, so we can always use the fixed version.

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -821,18 +828,19 @@ def __init__(self, cli_ctx, auth_context_factory, adal_token_cache, arm_client_f
...
+            client_type = self._get_subscription_client_class()
             if client_type is None:
+                from azure.cli.core.azclierror import CLIInternalError
                 raise CLIInternalError("Unable to get '{}' in profile '{}'"
                                        .format(ResourceType.MGMT_RESOURCE_SUBSCRIPTIONS, cli_ctx.cloud.profile))
             api_version = get_api_version(cli_ctx, ResourceType.MGMT_RESOURCE_SUBSCRIPTIONS)
+            client_kwargs = _prepare_client_kwargs_track2(cli_ctx)
+            # We don't need to change credential_scopes as 'scopes' is ignored by BasicTokenCredential anyway
             client = client_type(credentials, api_version=api_version,
-                                 base_url=self.cli_ctx.cloud.endpoints.resource_manager)
```
zhoxing-ms, line 834: What api-version of Subscription is used by azure-cli, the corresponding version of vendor SDK is required for azure-cli-core, right?

jiasli: Only `2019-11-01` and `2016-06-01` are used and they are included under `vendored_sdks`.

https://github.com/Azure/azure-cli/blob/0952571c2214693a79850f67d33b3fa6ab0354c0/src/azure-cli-core/azure/cli/core/profiles/_shared.py#L155

https://github.com/Azure/azure-cli/blob/0952571c2214693a79850f67d33b3fa6ab0354c0/src/azure-cli-core/azure/cli/core/profiles/_shared.py#L225

arrownj (approved):

src/azure-cli-core/azure/cli/core/_profile.py
zhoxing-ms: I mean if a new api-version of Subscription is added to azure-cli in the feture, does the vendor SDK in azure-cli-core also need to change to this new api-version, right?

zhoxing-ms (approved):

src/azure-cli-core/azure/cli/core/_profile.py
jiasli: `MGMT_RESOURCE_SUBSCRIPTIONS` is now only used by `azure-cli-core`. We can use it for now, assuming the corresponding `api-version` is vendored correctly.

src/azure-cli-core/azure/cli/core/adal_authentication.py
```diff
@@ -171,3 +171,13 @@ def _try_scopes_to_resource(scopes):
...
+
+
+class BasicTokenCredential:
+    # pylint:disable=too-few-public-methods
+    """A Track 2 implementation of msrest.authentication.BasicTokenAuthentication."""
+    def __init__(self, access_token):
+        self.access_token = access_token
+
+    def get_token(self, *scopes, **kwargs):  # pylint:disable=unused-argument
+        return AccessToken(self.access_token, int(time.time() + 3600))
```
jsntcy, line 183: Can we use a meaningful constant instead of a magic number?

jsntcy (approved):
:shipit:

src/azure-cli-core/azure/cli/core/adal_authentication.py
jiasli: As `self.access_token` is a fixed value, `get_token` cannot refresh the token, thus defeating the purpose of returning a valid `expires_on`. Using `time.time() + 3600` will force SDK to use the provided token, regardless of whether the token is expired or not.

Also, [Subscriptions - List](https://docs.microsoft.com/en-us/rest/api/resources/subscriptions/list) and [Tenants - List](https://docs.microsoft.com/en-us/rest/api/resources/tenants/list) APIs are not long-running operations, so the time of `expires_on` will not be reached anyway.

PR #15719, opened 2020-10-29
{Packaging} Add required python version in various setup.py

con-f-use:
Currently, pip assumes 2.14.0 is the latest version for all python interpreters and installs that version. This causes errors like this on older python versions:

```
  File "<string>", line 1, in <module>
  File "/home/confus/.local/share/virtualenvs/specter-yzp-I12c/lib/python3.2/site-packages/azure/cli/core/commands/__init__.py", line 425
    def cached_put(cmd_obj, operation, parameters, *args, setter_arg_name='parameters', **kwargs):
                                                                        ^
SyntaxError: invalid syntax
```

This probably a systemic problem in more than one of the azure libraries and because older packages with this error are already on pypi.org and other package repositories (debian, suse, ...), it is hard to solve in retrospect. This PR is just to raise awareness for the issue.

### DIFF
--- a/src/azure-cli-core/setup.py
+++ b/src/azure-cli-core/setup.py
@@ -91,10 +91,8 @@ setup(
         'azure.cli.core.profiles',
     ],
     install_requires=DEPENDENCIES,
+    python_requires='>=3.6.0',
     extras_require={
-        ":python_version<'3.4'": ['enum34'],
-        ":python_version<'2.7.9'": ['pyopenssl', 'ndg-httpsclient', 'pyasn1'],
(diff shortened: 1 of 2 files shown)


yonzhan:
required python version

yungezz:
hi @fengzhou-msft could you pls help to review?

fengzhou-msft (approved):

con-f-use:
Please be advised, that similar changes are likely due in other azure packages, too.

haroldrandom (approved):

jiasli (approved):

PR #15760, opened 2020-11-02
{Release} Hotfix: TS Multiline String Support for Template Inputs 

detienne20:
Description

Customer reported template specs issue (Azure/template-specs#37) where the contents of multiline strings were replaced with #Azure CLI#. This was the result of the _remove_comments_from_json function used to process the template input. This function is typically used in deployments for processing template inputs; multiline strings are processed to prompt for parameters. The full functionality is not needed for creating a template spec from a template input therefore the function has been replaced by process_template_spec in _packing_engine.py.

Example:

Template Input:
{
"variables": {
"myVar": "[if(equals(parameters('kind'), ''),
variables('resources')[variables('provider')][variables('resourceType')],
variables('resources')[variables('provider')][variables('resourceType')][parameters('kind')]))]"
}
}

Template Specs Created:
{
"variables": {
"myVar": "#Azure Cli#"
}
}

Testing Guide
az ts create -g resource_group --name Issue37 -v 1.0 -l westus --template-file original.json

--> Multiline strings starting on lines 2644, 2657, and 2661 should not be replaced by "#Azure CLI#"

-->Can compare template spec created with postTsCreation.json (showcases issue).
[Issue37Templates.zip](https://github.com/Azure/azure-cli/files/5476485/Issue37Templates.zip)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -122,7 +122,7 @@ jobs:
   - task: UsePythonVersion@0
     displayName: 'Use Python 3'
     inputs:
-      versionSpec: 3.x
+      versionSpec: 3.8
 
   - task: BatchScript@1
     inputs:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -122,7 +122,7 @@ jobs:
   - task: UsePythonVersion@0
     displayName: 'Use Python 3'
     inputs:
-      versionSpec: 3.8
+      versionSpec: 3.x
 
   - task: BatchScript@1
     inputs:
--- a/src/azure-cli/azure/cli/command_modules/resource/_packing_engine.py
+++ b/src/azure-cli/azure/cli/command_modules/resource/_packing_engine.py
@@ -33,7 +33,7 @@ def process_template(template, preserve_order=True, file_path=None):
     template = re.sub(r'(^[\t ]*//[\s\S]*?\n)|(^[\t ]*/\*{1,2}[\s\S]*?\*/)', '', template, flags=re.M)
     minified = jsmin(template)
 
-    # Get rid of extra spaces.
+    # Remove extra spaces, compress multiline string(s)
     result = re.sub(r'\s\s+', ' ', minified, flags=re.DOTALL)
 
     try:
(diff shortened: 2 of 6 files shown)


yonzhan:
TS Multiline String Support

zhoxing-ms:
May I ask if there is a requirement in the design for template spec that it cannot be modified on the client side, but should be sent directly to the service side?
For ARM template, it is designed like this, so I would like to confirm with you whether template spec has considered this aspect.

detienne20:
> May I ask if there is a requirement in the design for template spec that it cannot be modified on the client side, but should be sent directly to the service side?
> For ARM template, it is designed like this, so I would like to confirm with you whether template spec has considered this aspect.


There is no requirement for this in Template Spec, it (template spec) takes in a valid json object.

src/azure-cli/azure/cli/command_modules/resource/_packing_engine.py
```diff
@@ -94,8 +115,7 @@ def _pack_artifacts(cmd, template_abs_file_path, context):
             TemplateSpecTemplateArtifact = get_sdk(cmd.cli_ctx, ResourceType.MGMT_RESOURCE_TEMPLATESPECS,
                                                    'TemplateSpecTemplateArtifact', mod='models')
             template_content = read_file_content(abs_local_path)
-            sanitized_template = _remove_comments_from_json(template_content)
-            template_json = json.loads(json.dumps(sanitized_template))
+            template_json = json.loads(json.dumps(process_template(template_content)))
```
zhoxing-ms, line 118: > json.loads(json.dumps(

Why serialize and then deserialize the object from `process_template()` ? The data returned by `process_template` has been deserialized~

detienne20: data returned by process_template is a ordered dictionary due to  shell_safe_json_parse, the template_json needs to be a json object.

zhoxing-ms: OK

zhoxing-ms (approved):

Juliehzl (commented):
Could you add test for the scenario?

Juliehzl (approved):

azure-pipelines.yml
```diff
@@ -122,7 +122,7 @@ jobs:
   - task: UsePythonVersion@0
     displayName: 'Use Python 3'
     inputs:
-      versionSpec: 3.x
+      versionSpec: 3.8
```
haroldrandom, line 125: You could revert this file as this PR https://github.com/Azure/azure-cli/pull/15804 has fixed the build issue

jiasli: Then merge from `release` to get a clear branch history.

PR #15779, opened 2020-11-03
{Core} Support file name check when using prefix `@` in parameters

bim-msft:
Background:

There are lots of usages of `get_json_object` in the CLI code base, eg:

```python
c.argument('release_policy', options_list=['--policy'],
                       help='The policy rules under which the key can be exported encoded with JSON. '
                            'Use @{file} to load from a file(e.g. @my_policy.json).',
                       type=get_json_object, is_preview=True)
```

Users can use something like `--policy @"C:\path\file.json"` to specify a file name. If the file name does not exist, CLI will treat the input string as JSON content, and then throw a parser error like this:
image

After the refinement, it will be more actionable:
image

Also added an error type `FileNotFoudError` which inherits from `UserFault`, synced offline with @houk-ms 

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/azclierror.py
+++ b/src/azure-cli-core/azure/cli/core/azclierror.py
@@ -194,7 +194,13 @@ class ManualInterrupt(UserFault):
     pass
 
 
-# Unknow error type
+# File not found error type
+class FileNotFoundError(UserFault):
+    """ File not found. """
+    pass
+
+
+# Unknown error type
 class UnknownError(UserFault):
     """ Reserved for the errors which can not be categorized into the error types above.
     Usually for the very general error type like CLIError, AzureError.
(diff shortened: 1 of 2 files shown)


houk-ms:
@evelyn-ys , please sync with @bim-msft to use the new defined error type `FileNotFoundError`

jiasli:
Please consider using text instead of screenshots in the PR description.

yungezz:
thanks @bim-msft for the fixing!  some comments
- are usage of params of this type `get_json_object` well documented in each command? or is there central way to show the usage?
- `--admin-user-password` is of type `get_json_object`?  THen lots of password related params need to be whitelisted. Is there easy way to find out all occurence? If no, instead of whitelist  params where `@` not meaning filepath, how about whitelist params where `@`  meaning file path to  reduce impact?

PR #15782, opened 2020-11-03
{ARM} Fix live test

zhoxing-ms:
**Description**  

## Fix live test #15623

|  Test   | Error Message  | Reason |
|  ----  | ----  | ----  |
| test_subscription_level_deployment_ts  | The storage account named armbuilddemo1801 is already taken. | Caused by fixed resource name |
| test_resource_group_deployment | Query 'properties.provisioningState' doesn't yield expected value 'Canceled', instead the actual value is 'Running' | Service status updates are not stable |
| test_management_group_deployment  | Query 'properties.provisioningState' doesn't yield expected value 'Canceled', instead the actual value is 'Running' | Service status updates are not stable |
|test_create_template_specs  | Query 'artifacts[0].path' doesn't yield expected value 'artifacts\createResourceGroup.json', instead the actual value is 'artifacts/createResourceGroup.json' | Path matching error caused by the change of operating system separator |
| test_update_template_specs  | Query 'artifacts[0].path' doesn't yield expected value 'artifacts\createResourceGroup.json', instead the actual value is 'artifacts/createResourceGroup.json' | Path matching error caused by the change of operating system separator |
| test_export_template_spec | AssertionError: False is not true | Path matching error caused by the change of operating system separator |
| test_subscription_level_deployment_old_command | Query '[0].name' doesn't yield expected value 'azure-cli-subscription_level_deploymentdvlqd53kz6wht32dt65u3', instead the actual value is 'policy_definition_deploy' | The data obtained by the list operation is uncertain due to the parallel running test |

**Testing Guide**  

**History Notes**  

---


yonzhan:
ARM

src/azure-cli/azure/cli/command_modules/resource/tests/latest/test_resource.py
```diff
@@ -710,6 +710,7 @@ def test_list_template_spec(self, resource_group, resource_group_one, resource_g
         self.cmd('ts delete --template-spec {template_spec_id_rg1} --yes')
 
     @ResourceGroupPreparer(name_prefix='cli_test_template_specs', location='westus')
+    @live_only()
```
zhoxing-ms, line 713: Because the path separator of the operating system for recording yaml files may be different from that of the current operating system, the test should be changed to `@live_only()`

Juliehzl: I think we should not add live_only in such scenario. If there is only path separator issue, we should make the code compatible for different OS, not using hard coded path separator.

zhoxing-ms:
```
self.check('artifacts[0].path', 'artifacts' + os.sep + 'createResourceGroup.json')
```

In fact the scenario why I add `live_only()` is like this:
1. When the test is executed locally on Windows system, the path separator is `\`, and the value recorded in the yaml file is "artifacts\createResourceGroup.json"
2. However, CI runs the test on a Linux system, and the path separator is `/`, so the expected match value is "artifacts/createResourceGroup.json". Therefore, there will be a error that the check does not match.

Because the path separator will be recorded into the yaml file, so the test of record mode in different operating systems will have the problem of mismatch.

jsntcy: Instead of using live_only, can we use os-independent check?@Zhou.Xing@microsoft.com 

---
In reply to: [522156993](https://github.com/Azure/azure-cli/pull/15782#discussion_r522156993) [](ancestors = 522156993)

Juliehzl: using check with `contains` not `equal` also would be fine.

zhoxing-ms: OK, I use the regular expression validation instead

Juliehzl (approved):

PR #15500, opened 2020-10-13
{Core} Remove azext_ai_did_you_mean_this from ALWAYS_LOADED_EXTENSIONS

jiasli:
**Description**  
Remove `azext_ai_did_you_mean_this` from `ALWAYS_LOADED_EXTENSIONS` as it has been retired (https://github.com/Azure/azure-cli-extensions/pull/2402).

PR #12889, opened 2020-04-06
{Aladdin} Add v0 hook to AzCliCommandParser class.

christopher-o-toole:
**Description of PR (Mandatory)**  
Add hook to allow extensions to make suggestions on parser failures.

**Failure Types**
| Failure Description               | Command to Reproduce     | Parse Error Handler             |
|-----------------------------------|--------------------------|---------------------------------|
| Command is not in a command group | az list-av               | AzCliCommandParser._check_value |
| Arguments are required            | az vm                    | AzCliCommandParser.error        |
| Expected one argument             | az login --username      | AzCliCommandParser.error        |
| Unrecognized parameter            | az login --list          | AzCliCommandParser.error        |
| Invalid JMESPath Query            | az vm list --query ".id" | AzCliCommandParser.error        |

PR #15793, opened 2020-11-04
{Core} Remove useless codes in core for ai-did-you-mean-this

houk-ms:
**Description**  

This PR removes the useless codes in core for ai-did-you-mean-this, as the extension was already retired.

- ai-did-you-mean-this is retired: https://github.com/Azure/azure-cli-extensions/pull/2402
- azext_ai_did_you_mean_this is removed from core #15500

### DIFF
--- a/src/azure-cli-core/azure/cli/core/parser.py
+++ b/src/azure-cli-core/azure/cli/core/parser.py
@@ -68,12 +68,6 @@ class AzCompletionFinder(argcomplete.CompletionFinder):
 class AzCliCommandParser(CLICommandParser):
     """ArgumentParser implementation specialized for the Azure CLI utility."""
 
-    @staticmethod
-    def recommendation_provider(version, command, parameters, extension):  # pylint: disable=unused-argument
-        logger.debug("recommendation_provider: version: %s, command: %s, parameters: %s, extension: %s",
-                     version, command, parameters, extension)
-        return []
-
     def __init__(self, cli_ctx=None, cli_help=None, **kwargs):
         self.command_source = kwargs.pop('_command_source', None)
         self._raw_arguments = None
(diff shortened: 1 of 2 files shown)


houk-ms:
@christopher-o-toole @mirdaki for awareness

jiasli (approved):
I think there may be something more to remove. See #12889 for the original change.

jiasli:
I think there may be something more to remove. See #12889 for the original change.

mirdaki:
I have no problem with this.

yungezz (approved):

houk-ms:
Some codes introduce by #12889 are not removed because they are still used in command recomendations

src/azure-cli-core/azure/cli/core/parser.py
```diff
@@ -418,7 +392,6 @@ def _check_value(self, action, value):  # pylint: disable=too-many-statements, t
                     command_str = roughly_parse_command(cmd_list[1:])
                     ext_name = self._search_in_extension_commands(command_str)
                     if ext_name:
-                        caused_by_extension_not_installed = True
```
fengzhou-msft, line 421: Can you override the recommended command if caused_by_extension_not_installed is true? The output is like this if users type n for dynamic extension install.

>  $ az blueprint list
> The command requires the extension blueprint. Do you want to install it now? The command will continue to run after the extension is installed. (Y/n): n
> CommandNotFoundError: The command requires the latest version of extension blueprint. To install, run 'az extension add --upgrade -n blueprint'.
> Try this: 'az account list'
> Still stuck? Run 'az --help' to view all commands or go to 'https://docs.microsoft.com/en-us/cli/azure/reference-index?view=azure-cli-latest' to learn more

fengzhou-msft (approved):

PR #15831, opened 2020-11-07
[RDBMS] : `az mysql flexible-server create --iops <value>` : Allow user to choose IOPS for their SKU.

arde0708:
This PR introduces the support for IOPS feature in MySQL Flexible Servers. The Cx can now pass --iops as a parameter to regulate the IOPS he desired for his/her SKU. The range of values it can take as per the SKU is described in the spec here : [IOPS Limits per SKU](https://microsoft.sharepoint.com/:x:/t/mysqlpms/EX_Wyge2JlhOgygL-YAAxOcBaZi21_yAzCHP09pvdsAS9Q?e=SnEvbq).

The command can be executed using : 
az mysql flexible-server create --iops <value>
az mysql flexible-server create --iops <value> --sku_name <value> --tier <value>
az mysql flexible server update --iops <value>
az mysql flexible-server update --iops <value> --sku_name <value> --tier <value>


yonzhan:
@evelyn-ys @Juliehzl please help with this PR review

src/azure-cli/azure/cli/command_modules/rdbms/_params.py
```diff
@@ -258,6 +258,10 @@ def _flexible_server_params(command_group):
                            help='Server major version.')
                 c.argument('zone', options_list=['--zone, -z'],
                            help='Availability zone into which to provision the resource.')
+                c.argument('iops', type=int, options_list=['--iops'],
```
Juliehzl, line 261: could you use CLIArgumentType to abstract the argument definition and reuse them?

src/azure-cli/azure/cli/command_modules/rdbms/_params.py
```diff
@@ -258,6 +258,10 @@ def _flexible_server_params(command_group):
...
+                           help='Number of IOPS to be allocated for this server. You will get certain amount of free IOPS based '
```
Juliehzl, line 262: Is there an allowed value set for the parameter?

src/azure-cli/azure/cli/command_modules/rdbms/flexible_server_custom_mysql.py
```diff
@@ -193,11 +199,78 @@ def flexible_server_update_custom_func(cmd, instance,
...
+                    else:
+                        iops = min(max_allowed_iops_new_sku, free_iops)
+                        logger.warning('Provisioning the server with %s IOPS...', iops)
+            instance.sku.name = sku_name
+            instance.sku.tier = tier
+        instance.storage_profile.storage_iops = iops
+
+    # pylint: disable=too-many-boolean-expressions
+    if (iops is None and tier is None and sku_name) or (iops is None and sku_name is None and tier):
+        raise CLIError('Argument Error. If you pass --tier, --sku_name is a mandatory parameter and vice-versa.')
```
Juliehzl, line 261: consider the error type in https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/azure/cli/core/azclierror.py

Juliehzl (changes requested):
please add test the your new paramater

src/azure-cli/azure/cli/command_modules/rdbms/flexible_server_custom_mysql.py
evelyn-ys, line 261: From previous code, there seems no restriction that `--sku-name` is a mandatory when passing `--tier`. An existing example shows like this:
https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/rdbms/_helptext_mysql.py#L270-L272

Is this breaking change expected?

src/azure-cli/azure/cli/command_modules/rdbms/_params.py
arde0708: No, it can be any int value.

src/azure-cli/azure/cli/command_modules/rdbms/_params.py
arde0708: Do you mean like the way we do it for local-context variables ? Also, just trying to understand how does that help apart from not being repeated in two commands where they serve as parameters ? Because for none of the parameters (which are common across the services - PG, MySQL and MariaDB) we don't use CLI argument() except in local context varaibles.

src/azure-cli/azure/cli/command_modules/rdbms/flexible_server_custom_mysql.py
arde0708: This wont be a breaking change. Previously - it would have anyways failed without a proper error message like ' Deployment Failed. Correlation ID .....' if one of the params were missing. Now, with this change, it will still continue to fail, but with a meaningful error message and the reason as to why it is failing.

arde0708: Thanks. Updated them to RequiredArgumentMissingError.

evelyn-ys: I didn't know that two params are required not only for create, but also for update. In that case, could you correct the example? I can still see this in help:
```
Examples
    Update a flexible server's sku, using local context for server and resource group.
        az mysql flexible-server update --sku-name Standard_D4s_v3
```

evelyn-ys:
@arde0708 Kind reminder: the **deadline** of code complete for S179 is **12/2/2020**. Please finish refining this PR before 12/2/2020 if you want yo catch S179, or else we may postpone it to S180.

src/azure-cli/azure/cli/command_modules/rdbms/tests/latest/test_rdbms_flexible_commands.py
```diff
@@ -65,18 +65,24 @@ def _get_resource_group(self, **kwargs):
...
-    postgres_location = 'southeastasia'
-    mysql_location = 'southeastasia'
-
-    @AllowLargeResponse()
-    @ResourceGroupPreparer(location=postgres_location)
-    def test_postgres_flexible_server_mgmt(self, resource_group):
-        self._test_flexible_server_mgmt('postgres', resource_group)
+    postgres_location = 'eastus'
+    mysql_location = 'westus2'
+    mysql_iops_location = 'southeastasia'
```
arde0708, line 70: What is the difference between the two ? I am assuming it is because IOPS is deployed in Stage only. But this needs to change to canary as the location because that is where the daily pipeline will run. Also, hold back from checking this PR in until the feature is rolled out in prod. Else Cx will face failures if they try provisioning with iops

DaeunYim:
> please add test the your new paramater

Added the test for it! Thanks :)

src/azure-cli/azure/cli/command_modules/rdbms/flexible_server_custom_mysql.py
DaeunYim: Fixed this help text. Thanks for pointing it out!

src/azure-cli/azure/cli/command_modules/rdbms/tests/latest/test_rdbms_flexible_commands.py
evelyn-ys: >  Also, hold back from checking this PR in until the feature is rolled out in prod. Else Cx will face failures if they try provisioning with iops

@arde0708 Can I mark this pr as draft now? You can bring it back to review stage if the feature is rolled out in prod.

DaeunYim:
Could you open this PR again since the feature will be released to PROD before s182 milestone is released.

yonzhan:
@evelyn-ys @Juliehzl to help with.

evelyn-ys (approved):

DaeunYim:
Can this PR be merged so I can resolve any conflicts when merging [another pending PR](https://github.com/Azure/azure-cli/pull/16693)??

evelyn-ys:
@Juliehzl Can you help approve this?

src/azure-cli/azure/cli/command_modules/rdbms/_params.py
Juliehzl: I mean for `--iops` definition.  You could abstract it as below and then reuse it:
https://github.com/Azure/azure-cli/blob/b4c666c9f9fc07eb98b85949e84184a3eddca9b0/src/azure-cli/azure/cli/command_modules/storage/_params.py#L86-L87

https://github.com/Azure/azure-cli/blob/b4c666c9f9fc07eb98b85949e84184a3eddca9b0/src/azure-cli/azure/cli/command_modules/storage/_params.py#L221

Juliehzl (approved):
LGTM in general. But please consider refining your code as commented in https://github.com/Azure/azure-cli/pull/15831/files#r567540534

Issue #15692, opened 2020-10-28
Error Output Improvement: Clear Requests


houk-ms:
These requests are for error output improvements, which are clear and techinically feasible.

**1. Remove the ErrorType name in error message.**

Currently
```
RequiredArgumentMissingError: the following arguments are required: --name/--resource-group/-n/-g, --location/-l
Try this: 'az group create --location westeurope --resource-group MyResourceGroup --subscription mysubscription'
Still stuck? Run 'az group create --help' to view all commands or go to 'https://docs.microsoft.com/en-us/cli/azure/reference-index?view=azure-cli-latest' to learn more
```
Request
```
The following arguments are required: --name/--resource-group/-n/-g, --location/-l
Try this: 'az group create --location westeurope --resource-group MyResourceGroup --subscription mysubscription'
Still stuck? Run 'az group create --help' to view all commands or go to 'https://docs.microsoft.com/en-us/cli/azure/reference-index?view=azure-cli-latest' to learn more
```

**2. Use users' input values instead of placeholders for arguments if there are**

Currently
```
(env_new) PS C:\Users\houk\Desktop\msws> az group create -n houk
RequiredArgumentMissingError: the following arguments are required: --location/-l
Try this: 'az group create -l westus -n MyResourceGroup'
Still stuck? Run 'az group create --help' to view all commands or go to 'https://docs.microsoft.com/en-us/cli/azure/reference-index?view=azure-cli-latest' to learn more
```
Request
```
(env_new) PS C:\Users\houk\Desktop\msws> az group create -n houk
RequiredArgumentMissingError: the following arguments are required: --location/-l
Try this: 'az group create -l <westus> -n houk'
Still stuck? Run 'az group create --help' to view all commands or go to 'https://docs.microsoft.com/en-us/cli/azure/reference-index?view=azure-cli-latest' to learn more
```
**3. Apply new error output for service related UserFaults**
NOTE: A possible problem could be the recommended command is not useful, because users are using CLI in a right way, they may not need recommended command

Currently
```
(env_new) PS C:\Users\houk\Desktop\msws> az group show -n houk
ResourceNotFoundError: Resource group 'houk' could not be found.
```
Request
```
(env_new) PS C:\Users\houk\Desktop\msws> az group show -n houk
ResourceNotFoundError: Resource group 'houk' could not be found.
Try this: 'az group list'
[...]

Issue #10170, opened 2019-08-06
Normalize param names and ordering in Telemetry lookup


JasonRShaver:
In the core/commands/__init__.py, when loading values into self.cli_ctx.data['safe_params'], we should normalize the names and order.
* Expand short names into the full name: e.g. -n becomes --name
* Order params consistency: e.g. (alpha ordering) --resource-group, --name 

This enables:
* More accurate telemetry
* Ability to translate command+paramset combos to REST endpoint via ARM telemetry (Arun ask)
* Smarter Aladdin models as it gets a more accurate signal

JasonRShaver:
@albaloo

yonzhan:
add to S163.

PR #15841, opened 2020-11-09
{Error Improvement} Failure recommendation enhancement

houk-ms:
**Description**  

This PR enables part of the feature requests in #15692

These features are

**1. Use users' input values instead of placeholders for arguments if there are**
- supports both short and long options
- supports both `key val` and `key=val`
- use `<>` to wrap placeholders

**2. Remove non-neccesary arguments in recommended commands**
- partly supported. If we have multiple candidate commands which could match users' input, we will choose a command without the non-necessary parameters.

Rely on Aladdin service to recommend commands with only necessary parameters

**Testing Guide**  

- for short/long options
Just type `az vm create -n houkVM --image houkImage `, you will get the recommended command with the parameter values replaced with your inputs.
```
Try this: 'az vm create -n houkVM -g <MyResourceGroup> --image houkImage'
```

- for `key val` and `key=val` cases
Just type `az vm create --name houkVM --image=houkImage`, you will get the recommended command with the parameter values replaced with your inputs.

```
Try this: 'az vm create -n houkVM -g <MyResourceGroup> --image houkImage'
```

- Remove non-necessary parameters
Just type `az group create`

Previously, you will get the recommended command
```
Try this: 'az group create --location westeurope --resource-group MyResourceGroup --subscription mysubscription'
```
Now, you will get the recommended command
```
Try this: 'az group create -l <westus> -n <MyResourceGroup>'
```


yonzhan:
@zhoxing-ms for awareness

src/azure-cli-core/azure/cli/core/command_recommender.py
```diff
@@ -211,66 +230,65 @@ def _disable_aladdin_service(self):
...
+
+        return param_mappings
+
+    def _normalize_parameters(self, raw_parameters):
+        """Normalize a parameter list.
+        Get the standard parameter names of the raw parameters, which includes:
+            1. Use long options to replace short options
+            2. Remove the unrecognized parameter names
+        An example: ['-g', 'RG', '-n', 'NAME'] ==> {'--resource-group': 'RG', '--name': 'NAME'}
+        """
```
fengzhou-msft, line 280: Good job! This can be used to solve this [telemetry issue](https://github.com/Azure/azure-cli/issues/10170), just assigned it to you.

src/azure-cli-core/azure/cli/core/command_recommender.py
```diff
@@ -320,3 +338,73 @@ def _get_error_type(self):
...
+                ['-g', 'RG', '--name=NAME'] ==> {'-g': 'RG', '--name': 'NAME'}
+            """
+
+            parameter_kwargs = dict()
+            for index, parameter in enumerate(parameters):
+                if parameter.startswith('-'):
+
+                    param_name, param_val = parameter, None
+                    if '=' in parameter:
+                        pieces = parameter.split('=')
```
yungezz, line 371: there's case param value be empty string "", suggest to be defensive

yungezz (approved):

src/azure-cli-core/azure/cli/core/command_recommender.py
houk-ms: Is it the case like `az vm create -n=`? If so, I think it's aready handled here.

fengzhou-msft (approved):

PR #15847, opened 2020-11-10
[ARM] Add az provider register-mg api

tjegbejimba:
**Description**  

Add capability to register management groups

swagger PR [link](https://github.com/Azure/azure-rest-api-specs/pull/10869)
**Testing Guide**  

provider register-mg --namespace {prov} --management-group-id {mg}
**History Notes**  

---


yonzhan:
ARM

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -1947,6 +1947,25 @@ def unregister_provider(cmd, resource_provider_namespace, wait=False):
...
+    r = rcf.providers.register_at_management_group_scope(resource_provider_namespace, management_group_id)
+    if r.registration_state == 'Registered':
+        return
+
+    if wait:
+        while True:
+            time.sleep(10)
+            rp_info = rcf.providers.get(resource_provider_namespace)
+            if rp_info.registration_state == 'Registered':
+                break
```
zhoxing-ms, line 1962: How about declaring `register_management_group` as a long running operation in swagger in the future? 
Can the logic of judging whether the registration has been completed be replaced by the judgment of `ARMPolling` according to the status code?

zhoxing-ms:
There are some conflicts here. Please resolve them

src/azure-cli/azure/cli/command_modules/resource/tests/latest/test_resource.py
```diff
@@ -643,6 +643,19 @@ def test_provider_registration(self):
...
 
 
+class ProviderMGRegistrationTest(ScenarioTest):
+
+    def test_provider_mg_registration(self):
+
+        self.kwargs.update({'prov': 'Microsoft.ClassicInfrastructureMigrate', 'mg': self.create_random_name('azure-cli-management', 30)})
+
+        result = self.cmd('provider show -n {prov}').get_output_in_json()
+        if result['registrationState'] == 'Unregistered':
```
Juliehzl, line 653: please not use if in test scenarios.
just assert Unregistered status and then register the mg.

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -1947,6 +1947,25 @@ def unregister_provider(cmd, resource_provider_namespace, wait=False):
     _update_provider(cmd.cli_ctx, resource_provider_namespace, registering=False, wait=wait)
 
 
+def register_management_group(cmd, resource_provider_namespace, management_group_id, wait=False):
+    import time
+    rcf = _resource_client_factory(cmd.cli_ctx)
+    r = rcf.providers.register_at_management_group_scope(resource_provider_namespace, management_group_id)
+    if r.registration_state == 'Registered':
```
Juliehzl, line 1954: could you try to utilize the function_update_provider()?
https://github.com/Azure/azure-cli/blob/84eeeb0469054e2e054bc1cb234db371e433e590/src/azure-cli/azure/cli/command_modules/resource/custom.py#L963-L984

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -1947,6 +1947,25 @@ def unregister_provider(cmd, resource_provider_namespace, wait=False):
     _update_provider(cmd.cli_ctx, resource_provider_namespace, registering=False, wait=wait)
 
 
+def register_management_group(cmd, resource_provider_namespace, management_group_id, wait=False):
+    import time
+    rcf = _resource_client_factory(cmd.cli_ctx)
+    r = rcf.providers.register_at_management_group_scope(resource_provider_namespace, management_group_id)
```
Juliehzl, line 1953: I could see register mg is a separate function in SDK. But we still could consider add integrate mg id registration with `az provider register` command by adding one additional parameter `--mg`.
In this way, we could also utilize previous check code _update_provider because the only difference is operation name.

tjegbejimba: I changed the logic so that mg register utilizes the _update_provider function. Do you think that's enough or would you still like to see MG register integrated into `az provider register`?

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -1947,6 +1949,10 @@ def unregister_provider(cmd, resource_provider_namespace, wait=False):
     _update_provider(cmd.cli_ctx, resource_provider_namespace, registering=False, wait=wait)
 
 
+def register_management_group(cmd, resource_provider_namespace, management_group_id, wait=False):
+    _update_provider(cmd.cli_ctx, resource_provider_namespace, registering=True, wait=wait, mg=True, mgID=management_group_id)
```
Juliehzl, line 1953: Hi @tjegbejimba, because we could use one function for registering subscription and management group, could we consider a new design using `az provider register` with `--mg` instead of `az provider register-mg`?
utilizing `az provider register` would be more extensible

zhoxing-ms:
@tjegbejimba  We are going to release new version for CLI tomorrow. Could you please resolve these comments and conflicts?

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -137,6 +137,7 @@ def load_arguments(self, _):
         c.argument('resource_provider_namespace', options_list=['--namespace', '-n'], completer=get_providers_completion_list, help=_PROVIDER_HELP_TEXT)
 
     with self.argument_context('provider register') as c:
+        c.argument('mg', arg_type=management_group_id_type, help="The management group id to register.", required=False, options_list=['--mg'])
```
zhoxing-ms, line 140: `, required=False, options_list=['--mg']` these two parameters don't seem necessary. Could you please try to delete them?

tjegbejimba: I added them so they show up in the --help arguments section for provider register. If I remove them instead of 
`--mg                      : The management group id to register.`
it will read:
`--management-group-id -m [Required] : The management group id to register.`

which isn't what I want since it's the wrong flag and not required

tjegbejimba: I can remove the `arg_type` as well and manually add the options list if that would be a better solution

zhoxing-ms: In fact, `management_group_id_type` doesn't fit your scenario, so please don't use it
Your scenario is:
```
c.argument('mg', help="The management group id to register.")
```
But `management_group_id_type` is defined as
```
management_group_id_type = CLIArgumentType(options_list=['--management-group-id', '-m'], required=True, help='The management group id.')
```

zhoxing-ms: By the way, do we need to consider keeping parameter names consistent? How about naming the parameter as `['--management-group-id', '-m']` to users?

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -992,18 +992,20 @@ def _get_auth_provider_latest_api_version(cli_ctx):
     return api_version
 
 
-def _update_provider(cli_ctx, namespace, registering, wait, accept_terms=None):
+def _update_provider(cli_ctx, namespace, registering, wait, mgID=None, accept_terms=None):
```
zhoxing-ms, line 995: Hump case is not recommended for Python variable names. Please refer to [Style Guide for Python Code](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names)

> Function names should be lowercase, with words separated by underscores as necessary to improve readability.
> Variable names follow the same convention as function names.
> mixedCase is allowed only in contexts where that's already the prevailing style (e.g. threading.py), to retain backwards compatibility.

src/azure-cli/azure/cli/command_modules/resource/_params.py
tjegbejimba: Yeah that makes more sense I'll make that switch

zhoxing-ms (approved):

Juliehzl (approved):

Issue #15805, opened 2020-11-05
Can't resolve Keyvault reference in App Config with managed identity


nilsga:
## Describe the bug
Resolving a keyvault reference through App Config, with a VM managed identity does not work.

**Command Name**
`az appconfig kv list`

**Errors:**
```
'MSIAuthenticationWrapper' object has no attribute '_token_retriever'
```

## To Reproduce:
Steps to reproduce the behavior. 

* Create a VM with a managed identity
* Create a keyvault, and give VM access policy to get and list secrets
* Create an App Config instance
* Add a keyvault reference, say `testsecret/keyvaultref`
* From the VM, run `az appconfig kv list --name <appconfig_name> --key 'testsecret/*' --resolve-keyvault`

## Expected Behavior
The value should be resolved from the keyvault.

## Environment Summary
```
Linux-5.4.0-1031-azure-x86_64-with-debian-buster-sid
Python 3.6.10
Installer: DEB

azure-cli 2.14.0

```
## Additional Context

Stacktrace:

```
msrestazure.azure_active_directory : MSI: Token retrieved
azure.cli.core.util.handle_exception is called with an exception:
Traceback (most recent call last):
  File "/opt/az/lib/python3.6/site-packages/azure/cli/command_modules/appconfig/_kv_helpers.py", line 982, in __resolve_secret
    secret_version=kv_identifier.version)
  File "/opt/az/lib/python3.6/site-packages/azure/keyvault/v7_0/key_vault_client.py", line 1843, in get_secret
    response = self._client.send(request, header_parameters, stream=False, **operation_config)
  File "/opt/az/lib/python3.6/site-packages/msrest/service_client.py", line 336, in send
    pipeline_response = self.config.pipeline.run(request, **kwargs)
  File "/opt/az/lib/python3.6/site-packages/msrest/pipeline/__init__.py", line 197, in run
    return first_node.send(pipeline_request, **kwargs)  # type: ignore
  File "/opt/az/lib/python3.6/site-packages/msrest/pipeline/__init__.py", line 150, in send
    response = self.next.send(request, **kwargs)
  File "/opt/az/lib/python3.6/site-packages/msrest/pipeline/requests.py", line 72, in send
    return self.next.send(request, **kwargs)
  File "/opt/az/lib/python3.6/site-packages/msrest/pipeline/requests.py", line 137, in send
    return self.next.send(request, **kwargs)
[...]

PR #15854, opened 2020-11-10
{Core} Add comments to prevent calling AdalAuthentication._token_retriever

jiasli:
**Description**  

`_token_retriever` is only available for `AdalAuthentication` which is designed for **user or Service Principal credential**. It is not available for **Managed Identity** credential `MSIAuthenticationWrapper`. Therefore, calling `_token_retriever` to retrieve a raw token will fail for Managed Identity credential:

```py
# WRONG!!!
return Profile(cli_ctx=cli_ctx).get_login_credentials(resource)[0]._token_retriever()
```

This has already caused 2 buggy command module and extension: #15805, https://github.com/Azure/azure-cli-extensions/issues/2461

This PR adds comments to prevent calling `AdalAuthentication._token_retriever` in the future by developers.

For more detail including correct ways to get a raw token, see https://github.com/Azure/azure-cli/issues/15805#issuecomment-722914070.


yonzhan:
Core

arrownj (approved):
👍

evelyn-ys (approved):

PR #15870, opened 2020-11-11
[Network] az network dns zone import: Raise FileOperationError instead of FileNotFoundError if zone file doesn't exist

haroldrandom:
**Description**  

Fix the top 1 non-actionable bug (Azure/azure-cli#14984) 
image

Raise FileOperationError instead of FileNotFoundError if zone file doesn't exist.

**Testing Guide**  

**History Notes**  

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/network/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/network/custom.py
@@ -2046,7 +2046,12 @@ def import_zone(cmd, resource_group_name, zone_name, file_name):
     logger.warning("In the future, zone name will be case insensitive.")
     RecordSet = cmd.get_models('RecordSet', resource_type=ResourceType.MGMT_NETWORK_DNS)
 
-    file_text = read_file_content(file_name)
+    from azure.cli.core.azclierror import FileOperationError
+    try:
+        file_text = read_file_content(file_name)
+    except FileNotFoundError as e:
+        raise FileOperationError(e)
+
     zone_obj = parse_zone_file(file_text, zone_name)
 
     origin = zone_name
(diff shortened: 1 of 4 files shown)


src/azure-cli/azure/cli/command_modules/network/tests/latest/test_dns_commands.py
```diff
@@ -59,6 +59,14 @@ def _test_zone(self, zone_name, filename):
         # verify that each record in the original import is unchanged after export/re-import
         self._check_records(records1, records2)
 
+    @live_only()
```
haroldrandom, line 62: Just FYI, why decorated as `live_only`?

Because the HTTP request is never sent and vcr framework won't be able to record anything. However, vcr would think the test without YAML recording should run in live mode which requires login. 

So there is a conflict that CI won't perform and don't contain any login credential. This test is meant to be live only

msyyc (approved):
ok

src/azure-cli/azure/cli/command_modules/network/custom.py
```diff
@@ -2046,7 +2046,12 @@ def import_zone(cmd, resource_group_name, zone_name, file_name):
     logger.warning("In the future, zone name will be case insensitive.")
     RecordSet = cmd.get_models('RecordSet', resource_type=ResourceType.MGMT_NETWORK_DNS)
 
-    file_text = read_file_content(file_name)
+    from azure.cli.core.azclierror import FileOperationError
+    try:
+        file_text = read_file_content(file_name)
+    except FileNotFoundError as e:
```
houk-ms, line 2052: May also catch other types, e.g, `IsADirectoryError` ?

src/azure-cli/azure/cli/command_modules/network/custom.py
```diff
@@ -2046,7 +2046,12 @@ def import_zone(cmd, resource_group_name, zone_name, file_name):
...
+        raise FileOperationError(e)
```
houk-ms, line 2053: Is the raw error message clear enough? Could we customize the error message (and provide some recommendations if necessary)

PR #15875, opened 2020-11-11
{Core} Fix the incorrect token expiration on ADAL tokens

bim-msft:
Currently I'm facing with a token expiration issue, and I think I have found the root cause of this. When using ADAL token retriever, token data will be loaded from the local cache (`~/.azure/accessTokens.json`), incluing two important properties: `expiresIn` and `expiresOn`:

Local cache file:
image

Unfortunatelly, our CLI framework just loads all data directly from the cache without any changes:
image

Based on the observation, our current code is incorrect. It just use `expiresIn + now` to calculate the expiration time, which will be larger than the real one:
https://github.com/Azure/azure-cli/blob/4eb137b2df1a2969cebe1357cbf74ab0490e0709/src/azure-cli-core/azure/cli/core/adal_authentication.py#L77-L79

---


src/azure-cli-core/azure/cli/core/adal_authentication.py
```diff
@@ -75,6 +76,11 @@ def get_token(self, *scopes, **kwargs):  # pylint:disable=unused-argument
         logger.debug("AdalAuthentication.get_token invoked by Track 2 SDK with scopes=%s", scopes)
 
         _, token, full_token, _ = self._get_token(_try_scopes_to_resource(scopes))
+
+        if 'expiresOn' in full_token:
+            return AccessToken(token, int(time.mktime(
+                datetime.datetime.strptime(full_token['expiresOn'], '%Y-%m-%d %H:%M:%S.%f').timetuple())))
```
jiasli, line 82: [`datetime.timestamp()`](https://docs.python.org/3/library/datetime.html#datetime.datetime.timestamp) is an easier way to acquire epoch time.

src/azure-cli-core/azure/cli/core/adal_authentication.py
```diff
@@ -75,6 +76,11 @@ def get_token(self, *scopes, **kwargs):  # pylint:disable=unused-argument
...
+
         try:
             return AccessToken(token, int(full_token['expiresIn'] + time.time()))
```
jiasli, line 85: Still we need to figure out why `expiresIn` gets corrupted in the first place.

src/azure-cli-core/azure/cli/core/adal_authentication.py
bim-msft: Good idea, updated.

src/azure-cli-core/azure/cli/core/adal_authentication.py
bim-msft: @jiasli From my understanding, relative values like `expiresIn` should not be saved and used. Relative values would become meaningless as time goes by. Or we can recalculate it everytime when we use it, but that will be no difference with using the absolute value `expiresOn` directly.

jiasli: The original code is really "carving the boat to seek the sunk sword" (刻舟求剑).

jiasli (approved):

src/azure-cli-core/azure/cli/core/adal_authentication.py
```diff
@@ -75,6 +76,13 @@ def get_token(self, *scopes, **kwargs):  # pylint:disable=unused-argument
         logger.debug("AdalAuthentication.get_token invoked by Track 2 SDK with scopes=%s", scopes)
 
         _, token, full_token, _ = self._get_token(_try_scopes_to_resource(scopes))
+
+        try:
+            expires_on = full_token.get('expiresOn', full_token['expires_on'])
```
jiasli, line 81: `full_token['expires_on']` will be evaluated first causing failure.

```suggestion
            expires_on = full_token['expiresOn']
```

jiasli: This will also break Cloud Shell because `expires_on` is an epoch time and should be used directly. Parsing it will result in failure.

qwordy:
I also met this problem!

PR #15884, opened 2020-11-12
{Core} Fix 404 in long running delete operation

fengzhou-msft:
**Description**

In the response of `DELETE` long running operation, some service (observed in some corner case of [servicebus](https://clitestresultstac.blob.core.windows.net/20201106-033418-010544/servicebus.report.parallel.html), **update**: confirmed in swagger review session that more services have this behavior) may return 404 while `GET`ting operation results when the resource is successfully deleted (normally it should return 200 with no response body). 

To handle the 404 case, this PR adds similar logic to the error handling in `wait --deleted` command:
https://github.com/Azure/azure-cli/blob/05e3b63a3df61aa17cf9e29e0b3d45ffbb3b6188/src/azure-cli-core/azure/cli/core/commands/arm.py#L700-L702

**Testing Guide**

Without this PR, `azdev test test_sb_alias --live` will fail when deleting the secondary namespace.
With this PR, it can succeed.

**History Notes**

---


yonzhan:
Core

arrownj (approved):

src/azure-cli-core/azure/cli/core/commands/__init__.py
```diff
@@ -978,9 +978,11 @@ def __call__(self, poller):
         except ClientException as client_exception:
             from azure.cli.core.commands.arm import handle_long_running_operation_exception
             self.cli_ctx.get_progress_controller().stop()
+            if getattr(client_exception, 'status_code', None) == 404 and 'delete' in self.cli_ctx.data['command']:
```
Juliehzl, line 981: it looks good to me in general  for long running operation delete. but sometimes delete will not be long running operation, and I am not if there is service returning 404 for delete succeed scenario. 
so for better understanding for further change, could we add comment for the hack?

Juliehzl: will you change influence the scenario where resource to delete doesn't exist?

fengzhou-msft: Good point. I will do some tests for the scenario.

fengzhou-msft: Tested and this logic is in the LongRunningOperation class and will only be called when service return a poller for LRO. If delete is not a LRO, the code will not reach here. 

For the case that the resource does not exist when delete, the service will return an error directly instead of a poller. So this change will not impact such case.

404 seems to me is a bug in service side. Let me talk to service team first and see if they can fix on their side.

jiasli:
Is `GET` called on the service itself or the operation status (`Azure-AsyncOperation`/`Location`)?

src/azure-cli-core/azure/cli/core/commands/__init__.py
```diff
@@ -990,11 +990,14 @@ def __call__(self, poller):
...
```
jiasli, line 993: Even if we add this hack, Python SDK and its users will still suffers from this non-standard behavior.

jiasli: Do we have an open issue in https://github.com/Azure/azure-rest-api-specs to track this?

fengzhou-msft: I opened an issue in our repo https://github.com/Azure/azure-cli/issues/15906 for service team to deal with.

fengzhou-msft: Based on a disucssion with swagger reviewers, there are some existing services using 404 in LRO. For new swaggers, there should be validation tasks to catch it but for existing ones as long as the json is not modified with PRs, it will not be caught.

fengzhou-msft:
> Is `GET` called on the service itself or the operation status (`Azure-AsyncOperation`/`Location`)?

The error is thrown during long running operation, so I think it's called on the operation status.


```
        # Delete Namespace - primary
>       self.cmd('servicebus namespace delete --resource-group {rg} --name {namespacenameprimary}')

azure-cli/src/azure-cli/azure/cli/command_modules/servicebus/tests/latest/test_servicebus_alias_commands.py:163: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
azure-cli/src/azure-cli-testsdk/azure/cli/testsdk/base.py:167: in cmd
    return execute(self.cli_ctx, command, expect_failure=expect_failure).assert_with_checks(checks)
azure-cli/src/azure-cli-testsdk/azure/cli/testsdk/base.py:238: in __init__
    self._in_process_execute(cli_ctx, command, expect_failure=expect_failure)
azure-cli/src/azure-cli-testsdk/azure/cli/testsdk/base.py:301: in _in_process_execute
    raise ex.exception
env/lib/python3.8/site-packages/knack/cli.py:233: in invoke
azure-cli/src/azure-cli-core/azure/cli/core/commands/__init__.py:659: in execute
    raise ex
azure-cli/src/azure-cli-core/azure/cli/core/commands/__init__.py:722: in _run_jobs_serially
azure-cli/src/azure-cli-core/azure/cli/core/commands/__init__.py:715: in _run_job
    six.reraise(*sys.exc_info())
env/lib/python3.8/site-packages/six.py:703: in reraise
    raise value
azure-cli/src/azure-cli-core/azure/cli/core/commands/__init__.py:704: in _run_job
    result = LongRunningOperation(cmd_copy.cli_ctx, 'Starting {}'.format(cmd_copy.name))(result)
azure-cli/src/azure-cli-core/azure/cli/core/commands/__init__.py:988: in __call__
    handle_long_running_operation_exception(client_exception)
```

src/azure-cli-core/azure/cli/core/commands/__init__.py
jiasli: I still think we should force service team to fix it, instead of patching it from our side, otherwise Python SDK and its users will still suffers from this non-standard behavior.

houk-ms:
`Soft-delete` and `purge` operation on managed-hsm also suffer from this bad behavior. It just returns 404 when the delete operation completes.

**I think we should have this logic supported in CLI core to handle these cases.** Of course, at the same time, we should report these bad behiviors to service team.

houk-ms (approved):
LGTM

PR #15901, opened 2020-11-13
{Core} Honor expiresOn, instead of expiresIn

jiasli:
## Description

> Refine #15875, #15880

#12952 uses `expiresIn` to calculate the expiration date:

https://github.com/Azure/azure-cli/pull/12952/files#diff-7f489fcc6ae732563b52a550ccc3c32411ec118393eafd449ee9f5ab211ebf48R64

```py
return AccessToken(token, int(full_token['expiresIn'] + time.time()))
```

As tokens are cached, `expiresIn` saved in `~/.azure/accessTokens.json` will always be a fixed value, a little bit smaller or equal to `3600`. Using `full_token['expiresIn'] + time.time()` will result in an expiration time already past the actual expiration time (`expiresOn`).

```
Get token from server        3600s             Expiration
         |------------------------------------------|
               |------------------------------------------|
      Get token from cache                       Computed expiration
```

This way of computing expiration date is no different than "carving a mark on the boat to seek the sunk sword" (刻舟求剑).

## Changes

1. Honor `expiresOn`, instead of  `expiresIn`.
2. Add some token samples, to avoid future overlooked issue caused by different name of keys in the token entry (#14187) - User and service principal token entries have `expiresOn` in local time, while Managed Identity token entries has `expires_on` in epoch time.

## Testing Guide

Test on a local machine:

```sh
az login
az extension add -n account
account subscription show --id 0b1f6471-1bf0-4dda-aec3-cb9272f09590
```

Test on a VM with Managed Identity:

```sh
az login --identity
az extension add -n account
account subscription show --id 0b1f6471-1bf0-4dda-aec3-cb9272f09590
```

To test in Cloud Shell, see #13567.


yonzhan:
Core

qwordy:
Has Bin's PR fixed it?
Can we remove expiresIn field in token file.
When will we support file write protection of token file?

jiasli:
> Has Bin's PR fixed it?

Those PRs still leave some unreachable code which should be removed:
```py
return AccessToken(token, int(full_token['expiresIn'] + time.time()))
```

> Can we remove expiresIn field in token file.

They are from ADAL cache, we prefer not to do special handling to it.

> When will we support file write protection of token file?

After we migrate to MSAL.

src/azure-cli-core/azure/cli/core/adal_authentication.py
```diff
@@ -171,3 +219,12 @@ def _try_scopes_to_resource(scopes):
...
     return scopes_to_resource(scopes)
+
+
+def _timestamp(dt):
+    # datetime.datetime can't be patched:
+    #   TypeError: can't set attributes of built-in/extension type 'datetime.datetime'
+    # So we wrap datetime.datetime.timestamp with this function.
+    # https://docs.python.org/3/library/unittest.mock-examples.html#partial-mocking
+    # https://williambert.online/2011/07/how-to-unit-testing-in-django-with-mocking-and-patching/
+    return dt.timestamp()
```
jiasli, line 230: This function serves the same purpose as `azure.cli.command_modules.profile.custom._fromtimestamp` introduced by #15131.

src/azure-cli-core/azure/cli/core/tests/test_adal_authentication.py
```diff
@@ -4,8 +4,12 @@
 # --------------------------------------------------------------------------------------------
 
 # pylint: disable=line-too-long
+import datetime
 import unittest
-from azure.cli.core.adal_authentication import _try_scopes_to_resource
+import unittest.mock as mock
+from unittest.mock import MagicMock
```
qwordy, line 10: Can be improved.

evelyn-ys (approved):

src/azure-cli-core/azure/cli/core/adal_authentication.py
```diff
@@ -77,17 +75,56 @@ def get_token(self, *scopes, **kwargs):  # pylint:disable=unused-argument
...
+        #     "resource": "https://management.core.windows.net/",
+        #     "accessToken": "eyJ0eXAiOiJKV...",
+        #     "isMRRT": True,
+        #     "_clientId": "22800c35-46c2-4210-b8a7-d8c3ec3b526f",
+        #     "_authority": "https://login.microsoftonline.com/54826b22-38d6-4fb2-bad9-b7b93a3e9c5a"
+        # }
+        if 'expiresOn' in full_token:
+            import datetime
+            expires_on_timestamp = int(_timestamp(
+                datetime.datetime.strptime(full_token['expiresOn'], '%Y-%m-%d %H:%M:%S.%f')))
```
arrownj, line 109: The default timezone here will be the local time zone. Who does set the expiresOn value ? CLI client or the service ? If it is the service, the time zone may be different and the timestamp may be inaccurate.

bim-msft (approved):
LGTM

src/azure-cli-core/azure/cli/core/tests/test_adal_authentication.py
jiasli: How?

src/azure-cli-core/azure/cli/core/adal_authentication.py
jiasli: AAD server sets `expires_in`:

https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow#successful-response-1

```
{
    "access_token": "eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiIsIng1dCI6Ik5HVEZ2ZEstZnl0aEV1Q...",
    "token_type": "Bearer",
    "expires_in": 3599,
    "scope": "https%3A%2F%2Fgraph.microsoft.com%2Fmail.read",
    "refresh_token": "AwABAAAAvPM1KaPlrEqdFSBzjqfTGAMxZGUTdM0t4B4...",
    "id_token": "eyJ0eXAiOiJKV1QiLCJhbGciOiJub25lIn0.eyJhdWQiOiIyZDRkMTFhMi1mODE0LTQ2YTctOD...",
}
```

and ADAL converts it to a local time:

https://github.com/AzureAD/azure-activedirectory-library-for-python/blob/35f9e003bed47936a1df3c6eb696691dc1fa91c3/adal/oauth2_client.py#L183-L187

```py
expires_in = wire_response.get(OAuth2.ResponseParameters.EXPIRES_IN)
if expires_in:
    now = datetime.now()
    soon = timedelta(seconds=expires_in)
    wire_response[OAuth2.ResponseParameters.EXPIRES_ON] = str(now + soon)
```

So `expiresOn` is computed locally and won't hit the "the time zone may be different and the timestamp may be inaccurate" issue.

arrownj (approved):

src/azure-cli-core/azure/cli/core/tests/test_adal_authentication.py
qwordy: Delete 2nd line or delete 1st line and import all used class in mock

src/azure-cli-core/azure/cli/core/adal_authentication.py
```diff
@@ -77,17 +75,56 @@ def get_token(self, *scopes, **kwargs):  # pylint:disable=unused-argument
...
+        #     "expires_on": "1605686811",
+        #     "not_before": "1605682911",
+        #     "resource": "https://management.core.windows.net/",
+        #     "token_type": "Bearer"
+        # }
+        if 'expires_on' in full_token:
             return AccessToken(token, int(full_token['expires_on']))
 
+        from azure.cli.core.azclierror import CLIInternalError
+        raise CLIInternalError("No expiresOn or expires_on is available in the token entry.")
```
Juliehzl, line 126: If there is no expected property in the token entry, should it be AAD or ADAL issue?

jiasli: As I tested, it is never possible for `expiresOn` or `expires_on` to be missing.

jsntcy (approved):

jsntcy:
@jiasli , please fix CI failure

PR #15955, opened 2020-11-17
{RDBMS} Update test servers to be publicly accessible

arde0708:
The motivation of this PR is to update the test servers to be publicly accessible. After every M-train(backend changes for PG and MySQL Flexible Servers), we need to run these tests in the Stage environment in order to validate if these changes are breaking any current functionality. Currently, the test servers are created with private Vnet access and the Stage environment does not support vnet creation. As a result, we are blocked on running these tests in the Stage environment. Every time we run these tests, we have to update the server creation commands to let them have public access instead of private. 

Command usage : 
azdev test rdbms --live


yonzhan:
RDBMS

evelyn-ys (approved):
LGTM

PR #15964, opened 2020-11-18
{Core} Add support for local record executed commands and error messages for `az next`

zhoxing-ms:
**Description**

Related PR: https://github.com/Azure/azure-cli-extensions/pull/2449

There are some decoupling problems that need to be improved, but due to the lack of time at present, a new issue is created to follow up its subsequent refactoring
Issue link: #16770

**Testing Guide**

**History Notes**

---


yonzhan:
Core

src/azure-cli-core/azure/cli/core/__init__.py
```diff
@@ -389,9 +389,17 @@ def _get_extension_suppressions(mod_loaders):
         self.command_group_table.clear()
         self.command_table.clear()
 
+        # Configure the commands that require all modules to be loaded
+        # Such as: "az next" need to support searching the contents of other module commands
+        full_loaded_commands = ['next']
```
jiasli, line 394: Let's put this at the beginning of this file and change it too `SKIP_COMMAND_INDEX_FOR`.

zhoxing-ms: Done

src/azure-cli-core/azure/cli/core/__init__.py
```diff
@@ -36,6 +36,9 @@
 ALWAYS_LOADED_MODULES = []
 # Extensions that will always be loaded if installed. They don't expose commands but hook into CLI core.
 ALWAYS_LOADED_EXTENSIONS = ['azext_ai_examples']
+# Configure the commands that need to skip command index
+# Such as: "az next" need to support searching the contents of other module commands
```
jiasli, line 40:
```suggestion
# `az next` needs to search commands from all modules and extensions.
```

src/azure-cli-core/azure/cli/core/azclierror.py
```diff
@@ -249,4 +251,10 @@ class CLIInternalError(ClientError):
     """ AzureCLI internal error """
     pass
 
+
+# Client error for az next
+class RecommendationError(ClientError):
+    """ The client error raised by "az next", it is needed in "az next" to skip error records """
```
jiasli, line 257:
```suggestion
    """ The client error raised by "az next", it is needed in "az next" to skip error records. """
```

src/azure-cli-core/azure/cli/core/azclierror.py
```diff
@@ -249,4 +251,10 @@ class CLIInternalError(ClientError):
     """ AzureCLI internal error """
     pass
 
+
+# Client error for az next
+class RecommendationError(ClientError):
```
jiasli, line 256: If this error is dedicated to `az next`, maybe consider putting it in `next` extension?

src/azure-cli-core/azure/cli/core/__init__.py
jiasli: We usually use backticks `` ` `` to mark commands.

src/azure-cli-core/azure/cli/core/azclierror.py
```diff
@@ -63,6 +64,7 @@ def print_error(self):
 
     def send_telemetry(self):
         telemetry.set_error_type(self.__class__.__name__)
+        log_latest_error_info(self.error_msg, self.__class__.__name__)
```
jiasli, line 67: Will `log_latest_error_info` be called if telemetry is turned off by setting `core.collect_telemetry=false`?

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -1253,3 +1254,101 @@ def get_parent_proc_name():
         parent_proc_name = _get_parent_proc_name()
         setattr(get_parent_proc_name, "return_value", parent_proc_name)
     return getattr(get_parent_proc_name, "return_value")
+
+
+def log_cmd_history(command, args):
```
jiasli, line 1259: Consider putting this function in a separate file, because `util.py` is already too big containing too many functions with different purposes.

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -1253,3 +1254,101 @@ def get_parent_proc_name():
...
+        az_next_is_installed = get_extension("next")
+    except ExtensionNotInstalledException:
+        az_next_is_installed = False
+
+    if not az_next_is_installed:
+        return
+
+    # For commands that do not require logging, we can pass in the parameter "--no-log" to control
+    if '--no-log' in args:
+        return
```
jiasli, line 1279: Is `--no-log` registered in `next` extension? It won't work without registration:

```
$ az group list --no-log
UnrecognizedArgumentError: unrecognized arguments: --no-log
```

jiasli:
It will be great if an outline of the PR can be provided in the description.

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -17,6 +17,7 @@
 
 import six
 from six.moves.urllib.request import urlopen  # pylint: disable=import-error
+from azure.cli.core._environment import get_config_dir
```
jiasli, line 20: Consider delay this import to the function where it is actually used.

src/azure-cli-core/azure/cli/core/commands/__init__.py
```diff
@@ -518,6 +519,10 @@ def execute(self, args):
         self.cli_ctx.invocation.data['command_string'] = command
         telemetry.set_raw_command_name(command)
 
+        # record execution log when "az next" is installed
+        log_cmd_history(command, args)
+        clean_exception_history(command)
```
jiasli, line 524: Core should not be responsible for calling these 2 functions.

```suggestion
        prepare_for_next(command, args)
```

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -1253,3 +1254,101 @@ def get_parent_proc_name():
...
+    from knack.util import ensure_dir
+    from azure.cli.core.extension import get_extension, ExtensionNotInstalledException
+
+    if command == 'next':
+        return
+
+    # Determine whether "az next" has been installed.
+    # At present, command execution log is only recorded when "az next" is installed
+    try:
+        az_next_is_installed = get_extension("next")
```
jiasli, line 1270: `get_extension` on each command will have performance impact.

src/azure-cli-core/azure/cli/core/__init__.py
```diff
@@ -389,9 +392,14 @@ def _get_extension_suppressions(mod_loaders):
         self.command_group_table.clear()
         self.command_table.clear()
 
+        need_skip_command_index = False
+        if isinstance(args, list) and args:
+            need_skip_command_index = args[0] in SKIP_COMMAND_INDEX_FOR
```
jiasli, line 397: I have made `load_command_table` idempotent. You may load command table again with command index disabled to get the full command table.

```py
# pseudocode
os.environ['AZURE_CORE_USE_COMMAND_INDEX']='True'
load_command_table()
```

zhoxing-ms: OK~

src/azure-cli-core/azure/cli/core/azclierror.py
zhoxing-ms: Yes,  it is only used in `log_latest_error_info` at present. When `log_latest_error_info` is migrated to extension, it will also be migrated

jiasli (approved):

src/azure-cli-core/azure/cli/core/azclierror.py
zhoxing-ms: The reason why the exception log is recorded here is to record the same exception information locally as that sent to the telemetry for subsequent matching.
I will follow up with @houk-ms  to see if there is a more suitable place to record all exceptions uniformly.

src/azure-cli-core/azure/cli/core/util.py
zhoxing-ms: Yes, and `--no-log` is currently only used for internal logic in the `next` extension and will not have the same name as the existing parameters in CLI.

src/azure-cli-core/azure/cli/core/commands/__init__.py
zhoxing-ms: Good suggestion!

src/azure-cli-core/azure/cli/core/util.py
zhoxing-ms: Done

src/azure-cli-core/azure/cli/core/util.py
Juliehzl: have same concern

src/azure-cli-core/azure/cli/core/azclierror.py
houk-ms: This is a valid concern, when setting `core.collect_telemetry=false`. The function `send_telemetry` will be called,  but the telemtry record won't be written to cache file. This is controlled in https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/azure/cli/core/telemetry.py#L235. While in this case `log_latest_error_info` will log the error info. 

Another problem is this is not the only entry point where we log telemetry.

So as we synced offline, we may utilize telemetry/cache file to avoid the problem.

houk-ms (approved):
As planed, we may need some refactoring works in next sprint.

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -1253,3 +1253,104 @@ def get_parent_proc_name():
...
+    except ExtensionNotInstalledException:
+        az_next_is_installed = False
+
+    if not az_next_is_installed:
+        return
+
+    # For commands that do not require logging, we can pass in the parameter "--no-log" to control
+    if '--no-log' in args:
```
Juliehzl, line 1278: As I know, there are settings collect_telemetry and enable_log_file in our config. Will these setting also influence if you will start logging?

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -1253,3 +1253,104 @@ def get_parent_proc_name():
...
+        params = []
+        for arg in args:
+            if arg.startswith('-'):
+                params.append(arg)
+        if params:
+            command_info['arguments'] = params
+
+        lines.append(json.dumps(command_info))
+        if len(lines) > 15:
+            lines = lines[-15:]
```
Juliehzl, line 1308: why we only choose last 15 lines here?

src/azure-cli-core/azure/cli/core/util.py
zhoxing-ms: At present, this method is only available for `az next` to use, so it has a independent control logic in the scenario of `az next`, which will not be affected by these settings.
I'll migrate this method to extension in an event driven way in the next sprint

src/azure-cli-core/azure/cli/core/util.py
zhoxing-ms: Because I had concern that too many commands will be recorded, which may occupy the user's disk space. At present, we actually only use the last two commands for analysis

src/azure-cli-core/azure/cli/core/azclierror.py
zhoxing-ms: Yes, this is a great suggestion. It has been added to issue #16770

fengzhou-msft (approved):

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -1253,3 +1253,104 @@ def get_parent_proc_name():
...
+
+    if not az_next_is_installed:
+        return
+
+    # For commands that do not require logging, we can pass in the parameter "--no-log" to control
+    if '--no-log' in args:
+        return
+
+    if not args or not command:
+        return
```
evelyn-ys, line 1282: Can we combine these early return conditions before determining whether "az next" has been installed as `get_extension` cost a lot.
```suggestion
    if not args or not command or command == 'next' or '--no-log' in args:
        return

    # Determine whether "az next" has been installed.
    # At present, command execution log is only recorded when "az next" is installed
    try:
        get_extension("next")
    except ExtensionNotInstalledException:
        return
```

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -1253,3 +1253,104 @@ def get_parent_proc_name():
...
+
+    if not az_next_is_installed:
+        return
+
+    if not error_info:
+        return
+
+    # az next exception will not be recorded
+    if error_type and error_type == 'RecommendationError':
+        return
```
evelyn-ys, line 1333:
```suggestion
    if not error_info or (error_type and error_type == 'RecommendationError'):
        return
 
    # Determine whether "az next" has been installed.
    # At present, exception log is only recorded when "az next" is installed
    try:
        get_extension("next")
    except ExtensionNotInstalledException:
        return
```
Same as above

src/azure-cli-core/azure/cli/core/util.py
Juliehzl: If we could record command scenario, it would be more useful for analysis. 
Current design is fine for start stage, but hope we could consider the improvement later.

src/azure-cli-core/azure/cli/core/util.py
Juliehzl: I see. Please also consider if user doesn't enable logging in config. In such scenario, we may not also log for `az next` to avoid policy violations.
@fengzhou-msft do you have any idea for logging to file setting?

src/azure-cli-core/azure/cli/core/util.py
zhoxing-ms: Yes~ At present, there are technical challenges in locally determining whether a group of commands are in a scenario, we will consider this optimization in our long-term plan

evelyn-ys (approved):

PR #15775, opened 2020-11-03
{Storage} Add error handling for file related exceptions

evelyn-ys:
## Description ##
This PR works as a part of fixing the none-actionable issues in #14985 

- Fix file related exceptions with `az storage blob upload`, `az storage blob download`, `az storage file upload`, `az storage file download`

- Add new error type: FileOperationError

## Testing Guide ##  
`az storage blob upload -f /file/not/exist -c mycontainer -n myblob`
`az storage blob upload -f /file/with/no/permission -c mycontainer -n myblob`
`az storage blob download -f /dir/not/exist/filename -c mycontainer -n myblob`
and other similar file commands

before this pr, the errors will show as:
```
> az storage blob upload -f E:\test.py -n test.py -c yscontainer --account-name yssa

CLIInternalError: The command failed with an unexpected error. Here is the traceback:
[WinError 3] The system cannot find the path specified: 'E:\\test.py'
Traceback (most recent call last):
  File "D:\workspace\ysenv\lib\site-packages\knack\cli.py", line 215, in invoke
    cmd_result = self.invocation.execute(args)
  File "d:\workspace\azure-cli\src\azure-cli-core\azure\cli\core\commands\__init__.py", line 654, in execute
    raise ex
  File "d:\workspace\azure-cli\src\azure-cli-core\azure\cli\core\commands\__init__.py", line 718, in _run_jobs_serially
    results.append(self._run_job(expanded_arg, cmd_copy))
  File "d:\workspace\azure-cli\src\azure-cli-core\azure\cli\core\commands\__init__.py", line 709, in _run_job
    cmd_copy.exception_handler(ex)
  File "d:\workspace\azure-cli\src\azure-cli\azure\cli\command_modules\storage\__init__.py", line 343, in new_handler
    first(ex)
  File "d:\workspace\azure-cli\src\azure-cli\azure\cli\command_modules\storage\_exception_handler.py", line 9, in file_related_exception_handler
    raise ex
  File "d:\workspace\azure-cli\src\azure-cli-core\azure\cli\core\commands\__init__.py", line 688, in _run_job
    result = cmd_copy(params)
  File "d:\workspace\azure-cli\src\azure-cli-core\azure\cli\core\commands\__init__.py", line 325, in __call__
    return self.handler(*args, **kwargs)
  File "d:\workspace\azure-cli\src\azure-cli-core\azure\cli\core\__init__.py", line 784, in default_command_handler
    return op(**command_args)
[...]

PR #16034, opened 2020-11-25
{Storage} err improvement

evelyn-ys:
## Description ##
This PR is a supplement to #15775 , also works as a part of fixing the none-actionable issues in #14985 
Fix the left uncaught errors with `az storage`

## Testing Guide ##
see comments I left


src/azure-cli/azure/cli/command_modules/storage/__init__.py
```diff
@@ -170,12 +170,14 @@ def register_precondition_options(self):
...
-        self.extra('container_name', required=True)
+        from ._validators import get_not_none_validator
+        self.extra('blob_name', required=True, validator=get_not_none_validator('blob_name'))
+        self.extra('container_name', required=True, validator=get_not_none_validator('container_name'))
         self.extra('timeout', help='Request timeout in seconds. Applies to each call to the service.', type=int)
 
     def register_container_arguments(self):
-        self.extra('container_name', required=True)
+        from ._validators import get_not_none_validator
+        self.extra('container_name', required=True, validator=get_not_none_validator('container_name'))
```
evelyn-ys, line 180: Try `az storage blob list -c '""'  --account-name mystorageaccount` in PowerShell, It will pass an empty string and cause a crush.

src/azure-cli/azure/cli/command_modules/storage/_client_factory.py
```diff
@@ -204,7 +204,13 @@ def cf_blob_service(cli_ctx, kwargs):
...
     sas_token = kwargs.pop('sas_token', None)
     if connection_string:
-        return t_blob_service.from_connection_string(conn_str=connection_string)
+        try:
+            return t_blob_service.from_connection_string(conn_str=connection_string)
+        except ValueError:
+            from azure.cli.core.azclierror import InvalidArgumentValueError
+            raise InvalidArgumentValueError('Invalid connection string',
+                                            recommendation='Try `az storage account show-connection-string` '
+                                                           'to get a valid connection string')
```
evelyn-ys, line 213: Testing Guide
`az storage blob list -c mycontainer --connection-string 'some_invalid_connection_string'`

src/azure-cli/azure/cli/command_modules/storage/_client_factory.py
```diff
@@ -204,7 +204,13 @@ def cf_blob_service(cli_ctx, kwargs):
     token_credential = kwargs.pop('token_credential', None)
     sas_token = kwargs.pop('sas_token', None)
     if connection_string:
-        return t_blob_service.from_connection_string(conn_str=connection_string)
+        try:
+            return t_blob_service.from_connection_string(conn_str=connection_string)
+        except ValueError:
+            from azure.cli.core.azclierror import InvalidArgumentValueError
+            raise InvalidArgumentValueError('Invalid connection string',
```
houk-ms, line 211: Can we also show the invaid connection string in error message?

src/azure-cli/azure/cli/command_modules/storage/_exception_handler.py
```diff
@@ -12,6 +12,8 @@ def file_related_exception_handler(ex):
                                  recommendation='Please make sure you have enough permissions on the file/directory.')
     if isinstance(ex, IsADirectoryError):
         raise FileOperationError(ex, recommendation='File is excepted, not a directory.')
+    if isinstance(ex, NotADirectoryError):
+        raise FileOperationError(ex, recommendation='Directory is excepted, not a file.')
```
houk-ms, line 16:
```suggestion
        raise FileOperationError(ex, recommendation='Directory is expected, not a file.')
```

src/azure-cli/azure/cli/command_modules/storage/_exception_handler.py
```diff
@@ -12,6 +12,8 @@ def file_related_exception_handler(ex):
                                  recommendation='Please make sure you have enough permissions on the file/directory.')
     if isinstance(ex, IsADirectoryError):
         raise FileOperationError(ex, recommendation='File is excepted, not a directory.')
```
houk-ms, line 14:
```suggestion
        raise FileOperationError(ex, recommendation='File is expected, not a directory.')
```

Is this a typo?

src/azure-cli/azure/cli/command_modules/storage/_validators.py
```diff
@@ -1556,3 +1556,13 @@ def progress_callback(acl_changes):
...
     namespace.progress_hook = progress_callback
+
+
+def get_not_none_validator(attribute_name):
+    def validate_not_none(cmd, namespace):
+        attribute = getattr(namespace, attribute_name, None)
+        options_list = cmd.arguments[attribute_name].type.settings.get('options_list')
+        if not attribute:
+            from azure.cli.core.azclierror import InvalidArgumentValueError
+            raise InvalidArgumentValueError('argument {} can not be None'.format('/'.join(options_list)))
```
houk-ms, line 1567: Use capital case ahead of an error message.

yonzhan:
Storage

src/azure-cli/azure/cli/command_modules/storage/_validators.py
houk-ms: Is this case due to users not specifying the parameter? If so, I think we should use something like `Argument {} should be specified.` directly.

`argument {} can not be None` may be a little ambiguous, making users feel they had provided a `None` value for the parameter.

houk-ms (commented):
Leave some comments for error handling.

src/azure-cli/azure/cli/command_modules/storage/_exception_handler.py
evelyn-ys: Yes

src/azure-cli/azure/cli/command_modules/storage/_validators.py
evelyn-ys: Good suggestion.

src/azure-cli/azure/cli/command_modules/storage/_client_factory.py
evelyn-ys: Done

src/azure-cli/azure/cli/command_modules/storage/_client_factory.py
```diff
@@ -204,7 +204,14 @@ def cf_blob_service(cli_ctx, kwargs):
     token_credential = kwargs.pop('token_credential', None)
     sas_token = kwargs.pop('sas_token', None)
     if connection_string:
-        return t_blob_service.from_connection_string(conn_str=connection_string)
+        try:
+            return t_blob_service.from_connection_string(conn_str=connection_string)
+        except ValueError as err:
```
Juliehzl, line 209: when the `ValueError` will be raised in SDK?

src/azure-cli/azure/cli/command_modules/storage/_validators.py
```diff
@@ -1556,3 +1556,13 @@ def progress_callback(acl_changes):
         failed_entries.append(acl_changes.batch_failures)
 
     namespace.progress_hook = progress_callback
+
+
+def get_not_none_validator(attribute_name):
+    def validate_not_none(cmd, namespace):
+        attribute = getattr(namespace, attribute_name, None)
+        options_list = cmd.arguments[attribute_name].type.settings.get('options_list')
+        if not attribute:
```
Juliehzl, line 1565:
```suggestion
        if attribute is None:
```

consider  attribute value is 0 or False

src/azure-cli/azure/cli/command_modules/storage/_validators.py
```diff
@@ -1556,3 +1556,13 @@ def progress_callback(acl_changes):
         failed_entries.append(acl_changes.batch_failures)
 
     namespace.progress_hook = progress_callback
+
+
+def get_not_none_validator(attribute_name):
+    def validate_not_none(cmd, namespace):
+        attribute = getattr(namespace, attribute_name, None)
+        options_list = cmd.arguments[attribute_name].type.settings.get('options_list')
```
Juliehzl, line 1564: are you sure `options_list` always be there? recommend to give default value as fallback.

src/azure-cli/azure/cli/command_modules/storage/_validators.py
```diff
@@ -1556,3 +1556,13 @@ def progress_callback(acl_changes):
         failed_entries.append(acl_changes.batch_failures)
 
     namespace.progress_hook = progress_callback
+
+
+def get_not_none_validator(attribute_name):
```
Juliehzl, line 1561: please add unit test for the validator. see https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/storage/tests/latest/test_storage_validators.py

src/azure-cli/azure/cli/command_modules/storage/_client_factory.py
evelyn-ys: yes

src/azure-cli/azure/cli/command_modules/storage/_validators.py
evelyn-ys: Yes, all arguments should have `options_list`. Even if no `option_list` specified, core will create `option_list` by `dest` since `dest` is mandatory for each argument. [Code in knack](https://github.com/microsoft/knack/blob/dc7b7d857ce3c11b2a3155b81f2470b3b579e7c9/knack/arguments.py#L45-L68):
```Python
class CLICommandArgument(object):


    NAMED_ARGUMENTS = ['options_list', 'validator', 'completer', 'arg_group', 'deprecate_info', 'preview_info',
                       'experimental_info', 'default_value_source']


    def __init__(self, dest=None, argtype=None, **kwargs):
        """An argument that has a specific destination parameter.

        :param dest: The parameter that this argument is for
        :type dest: str
        :param argtype: The argument type for this command argument
        :type argtype: knack.arguments.CLIArgumentType
        :param kwargs: see knack.arguments.CLIArgumentType
        """
        self.type = CLIArgumentType(overrides=argtype, **kwargs)
        if dest:
            self.type.update(dest=dest)


        # We'll do an early fault detection to find any instances where we have inconsistent
        # set of parameters for argparse
        if not self.options.get('dest', False):
            raise ValueError('Missing dest')
        if not self.options_list:  # pylint: disable=access-member-before-definition
            self.options_list = ('--{}'.format(self.options['dest'].replace('_', '-')),)
```

src/azure-cli/azure/cli/command_modules/storage/_validators.py
evelyn-ys: Good catch.
Changed it to `if attribute in (None, '')` since I want to allow 0 or False but not allow empty string.

src/azure-cli/azure/cli/command_modules/storage/_validators.py
evelyn-ys: Done

src/azure-cli/azure/cli/command_modules/storage/_validators.py
```diff
@@ -1556,3 +1556,13 @@ def progress_callback(acl_changes):
...
+    def validate_not_none(cmd, namespace):
+        attribute = getattr(namespace, attribute_name, None)
+        options_list = cmd.arguments[attribute_name].type.settings.get('options_list')
+        if attribute in (None, ''):
+            from azure.cli.core.azclierror import InvalidArgumentValueError
+            raise InvalidArgumentValueError('Argument {} should be specified'.format('/'.join(options_list)))
```
houk-ms, line 1567: this should be a `RequiredArgumentMissingError`, right?

houk-ms (commented):
Error messages look good to me.

src/azure-cli/azure/cli/command_modules/storage/_validators.py
evelyn-ys: Actually they provide the required argument, but the value is None or '' which is invalid

houk-ms (approved):

Juliehzl (approved):

PR #16059, opened 2020-11-26
{Test} Add retry utility

qwordy:
**Description**

Time problem in testing
```
# test delete iaas pool
self.batch_cmd('batch pool delete --pool-id {pool_i} --yes')
self.batch_cmd('batch pool show --pool-id {pool_i} --select "state"').assert_with_checks([
    self.check('state', 'deleting')])
```
The second command will fail if the first command deletes the pool quickly enough.
```
self.cmd('vm simulate-eviction --resource-group {rg} --name {vm2}')
time.sleep(180)
self.cmd('vm get-instance-view --resource-group {rg} --name {vm2}', checks=[
    self.check('name', '{vm2}'),
    self.check('resourceGroup', '{rg}'),
    self.check('length(instanceView.statuses)', 2),
    self.check('instanceView.statuses[0].code', 'ProvisioningState/succeeded'),
    self.check('instanceView.statuses[1].code', 'PowerState/deallocated'),
])
```
This piece of code is not robust. “vm simulate-eviction” command returns before it really takes effect. 180 seconds is an empirical value to wait. Sometimes, the next command fails.
```
self.cmd('az cognitiveservices account identity remove -n {sname} -g {rg}')
E azure.mgmt.cognitiveservices.models._models_py3.ErrorException: (RequestConflict) Cannot modify resource with id '/subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/resourceGroups/clitest.rgbwjubs5hgprklyfdjebasibxhldpdlspiqzpt2zdstmfmkbevhhke5zk3e3psiw77/providers/Microsoft.CognitiveServices/accounts/cs_cli_test_g3ut' because the resource entity provisioning state is not terminal. Please wait for the provisioning state to become terminal and then retry the request.
```
It is a similar problem. The provisioning is not finished.
A good example,
```
for i in range(10):
    time.sleep(15) # sleep() is mocked in replay mode, so just use a large value.
    account = self.cmd('az cognitiveservices account show -n {sname} -g {rg}').get_output_in_json()
    if 'Creating' != account['properties']['provisioningState']:
        break
```
The benefit of this approach is that it can exit fast and doesn’t need to wait for the entire timeout. We can extract a function for it. That is `retry` is TestSDK.

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli-testsdk/azure/cli/testsdk/utilities.py
+++ b/src/azure-cli-testsdk/azure/cli/testsdk/utilities.py
@@ -4,6 +4,7 @@
 # --------------------------------------------------------------------------------------------
 
 import os
+import time
 from contextlib import contextmanager
 
 from azure_devtools.scenario_tests import (create_random_name as create_random_name_base, RecordingProcessor,
@@ -198,3 +199,18 @@ class GeneralNameReplacer(_BuggyGeneralNameReplacer):
                     request.body = body.replace(old, new)
 
         return request
+
+
+def retry(func, timeout, interval=10):
+    """
+    Retry func until success.
+    :param timeout: in seconds
+    :param func: function
+    :param interval: polling interval
+    :return:
+    """
+    for i in range(0, timeout, interval):
+        try:
+            return func()
+        except Exception:  # pylint: disable=broad-except
+            time.sleep(interval)
--- a/src/azure-cli-testsdk/HISTORY.rst
+++ b/src/azure-cli-testsdk/HISTORY.rst
@@ -3,6 +3,12 @@
 Release History
 ===============
 
+0.2.6
++++++
+* Add retry utility
+
+0.2.5
++++++
 * Add resource group name prefix validator in resource group preparer
 
 0.2.4


yonzhan:
@jiasli  for awareness.

src/azure-cli-testsdk/azure/cli/testsdk/utilities.py
```diff
@@ -198,3 +199,18 @@ def process_request(self, request):
...
+    :param timeout: in seconds
+    :param func: function
+    :param interval: polling interval
+    :return:
+    """
+    for i in range(0, timeout, interval):
+        try:
+            return func()
+        except Exception:  # pylint: disable=broad-except
+            time.sleep(interval)
```
jiasli, line 216: This method serves exactly the same functionality as https://github.com/Azure/azure-cli/blob/11347b73999dc9b9d1c6f924fb42e3d20b34a767/src/azure-cli/azure/cli/command_modules/role/tests/util.py#L12-L26

src/azure-cli-testsdk/azure/cli/testsdk/utilities.py
```diff
@@ -198,3 +199,18 @@ def process_request(self, request):
...
+
+def retry(func, timeout, interval=10):
+    """
+    Retry func until success.
+    :param timeout: in seconds
+    :param func: function
+    :param interval: polling interval
+    :return:
+    """
+    for i in range(0, timeout, interval):
```
jiasli, line 212: When `timeout` is reached, the exception is swallowed and can't be thrown out.

src/azure-cli-testsdk/azure/cli/testsdk/utilities.py
```diff
@@ -198,3 +199,18 @@ def process_request(self, request):
                     request.body = body.replace(old, new)
 
         return request
+
+
+def retry(func, timeout, interval=10):
```
jiasli, line 204: Better to have some tests for `retry`.

PR #15780, opened 2020-11-03
[Core] Upgrade azure-mgmt-core to 1.2.1

jsntcy:
**Description**  
Upgrade azure-mgmt-core to 1.2.1 due to codegen's upgrade, see details in #15766 

**Testing Guide**  

**History Notes**  

[Component Name 1] BREAKING CHANGE: az command a: Make some customer-facing breaking change.  
[Component Name 2] az command b: Add some customer-facing feature.

---

This checklist is used to make sure that common guidelines for a pull request are followed.

- [ ] The PR title and description has followed the guideline in [Submitting Pull Requests](https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules#submitting-pull-requests).

- [ ] I adhere to the [Command Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md).

PR #16074, opened 2020-11-28
{Core} Use semver for azure-mgmt-core

jonringer:
**Description**
Trying to package this for nixpkgs, but I get:
```
ERROR: No matching distribution found for azure-mgmt-core==1.2.1 (from azure-cli-core==2.15.1)
```
when it's being given 1.2.2

Azure-sdk generally follows semver, so no need to pin.

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli-core/setup.py
+++ b/src/azure-cli-core/setup.py
@@ -59,7 +59,7 @@ DEPENDENCIES = [
     'requests~=2.22',
     'six~=1.12',
     'pkginfo>=1.5.0.1',
-    'azure-mgmt-core==1.2.1',
+    'azure-mgmt-core~=1.2',
     # Dependencies of the vendored subscription SDK
     # https://github.com/Azure/azure-sdk-for-python/blob/ab12b048ddf676fe0ccec16b2167117f0609700d/sdk/resources/azure-mgmt-resource/setup.py#L82-L86
     'msrest>=0.5.0',
--- a/src/azure-cli-core/setup.py
+++ b/src/azure-cli-core/setup.py
@@ -59,7 +59,7 @@ DEPENDENCIES = [
     'requests~=2.22',
     'six~=1.12',
     'pkginfo>=1.5.0.1',
-    'azure-mgmt-core~=1.2',
+    'azure-mgmt-core>=1.2.0,<2.0.0',
     # Dependencies of the vendored subscription SDK
     # https://github.com/Azure/azure-sdk-for-python/blob/ab12b048ddf676fe0ccec16b2167117f0609700d/sdk/resources/azure-mgmt-resource/setup.py#L82-L86
     'msrest>=0.5.0',


yonzhan:
Core

src/azure-cli-core/setup.py
```diff
@@ -59,7 +59,7 @@
     'requests~=2.22',
     'six~=1.12',
     'pkginfo>=1.5.0.1',
-    'azure-mgmt-core==1.2.1',
```
jiasli, line 62: `azure-mgmt-core` was pinned by #15780. @jsntcy, is it necessary that we should pin it?

We does deliver `azure-mgmt-core` at https://pypi.org/project/azure-mgmt-core/1.2.1/. nixpkgs is not one of our supported delivery platform. @fengzhou-msft

fengzhou-msft: The Azure SDKs are in the transition of track 1 to track 2, there may be unexpected breaking changes. To play safe, we pinned the version for the transition period and a version bump even with the micro version needs full live tests.

Azure CLI relies on too many Azure SDKs and third-party dependencies, it is strongly recommended to use the all-in-one platform packages or install the wheel package in an isolated python environment.

jonringer: > nixpkgs is not one of our supported delivery platform.

Yes, I maintain azure-cli for nixpkgs / NixOS.

> The Azure SDKs are in the transition of track 1 to track 2, there may be unexpected breaking changes. To play safe, we pinned the version for the transition period and a version bump even with the micro version needs full live tests.

This should only be done if there's breaking changes between patch versions. Azure sdk doesn't.

> it is strongly recommended to use the all-in-one platform packages or install the wheel package in an isolated python environment.

This is what I'm providing for nixpkgs. I already applied the correct patch https://github.com/NixOS/nixpkgs/pull/105267/commits/d03b3c4915e557122658ea0698b89ee0c716194e, I'm just upstreaming the changes so that other azure-cli users can benefit from better dependency management.

> To play safe, we pinned the version for the transition period and a version bump even with the micro version needs full live tests.

What I'm referring to is https://github.com/Azure/azure-sdk-for-python/blob/master/doc/dev/mgmt/mgmt_release.md#version in which python sdk is guaranteeing not to break downstream packages until the next the next major bump

cc @lmazuel

fengzhou-msft: > What I'm referring to is https://github.com/Azure/azure-sdk-for-python/blob/master/doc/dev/mgmt/mgmt_release.md#version in which python sdk is guaranteeing not to break downstream packages until the next the next major bump

It's supposed to be, but not always true when developers are not aware of the breaking changes especially with Track 2 adoption. Add @jsntcy who works with this SDK to evaulate if we're confident to follow semver at this stage.

src/azure-cli-core/setup.py
```diff
@@ -59,7 +59,7 @@
...
+    'azure-mgmt-core~=1.2',
```
jiasli, line 62: We can at least use `azure-mgmt-core~=1.2.1` to get our users unblocked.

jiasli: Mgmt SDKs use:

https://github.com/Azure/azure-sdk-for-python/blob/ab12b048ddf676fe0ccec16b2167117f0609700d/sdk/network/azure-mgmt-network/setup.py#L85

```py
'azure-mgmt-core>=1.2.0,<2.0.0',
```

which is equivalent to

```py
'azure-mgmt-core~=1.2'
```

jiasli, line 62:
```suggestion
    'azure-mgmt-core>=1.2.0,<2.0.0',
```

jiasli (approved):

yonzhan (approved):
LGTM

jsntcy (approved):

PR #16081, opened 2020-11-30
[Role] az ad sp create-for-rbac: Deprecate creating Contributor role assignment by default

jiasli:
**Description**
`az ad sp create-for-rbac` creates a **Contributor** role assignment at **subscription** scope if `--skip-assignment` or `--role` is not provided. This is a dangerous behavior and contradicts the [principle of least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege).

This PR adds the warning to **help message** and **command execution** that:

> In a future release, this command will NOT create any role assignment by default. (--skip-assignment will be the default behavior.) If needed, always use --role to explicitly create a role assignment.

**Testing Guide**

```sh
> az ad sp create-for-rbac --help

        By default, this command assigns the 'Contributor' role to the service principal at the
        subscription scope. To reduce your risk of a compromised service principal, use --skip-
        assignment to avoid creating a role assignment, then assign a more specific role and narrow
        the scope to a resource or resource group. See [steps to add a role
        assignment](https://aka.ms/azadsp-more) for more information.

        WARNING: In a future release, this command will NOT create 'Contributor' role assignment by
        default. If needed, always use --role to explicitly create a role assignment.

> az ad sp create-for-rbac
WARNING: In a future release, this command will NOT create 'Contributor' role assignment by default. If needed, 
always use --role to explicitly create a role assignment.
WARNING: Creating 'Contributor' role assignment under scope '/subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590'
WARNING:   Retrying role assignment creation: 1/36
{
  "appId": "bc26a0e0-a351-403a-9605-f97424e35dd2",
  "displayName": "azure-cli-2020-11-30-06-01-01",
  "name": "http://azure-cli-2020-11-30-06-01-01",
  "password": "...",
  "tenant": "54826b22-38d6-4fb2-bad9-b7b93a3e9c5a"
}
```

See email: 
- _az ad sp create-for-rbac: Deprecate creating role assignment by default_
- _Updating az ad sp create-for-rbac behavior and warning_


src/azure-cli/azure/cli/command_modules/role/custom.py
```diff
@@ -1397,7 +1402,7 @@ def _validate_app_dates(app_start_date, app_end_date, cert_start_date, cert_end_
 # pylint: disable=inconsistent-return-statements
 def create_service_principal_for_rbac(
         # pylint:disable=too-many-statements,too-many-locals, too-many-branches
-        cmd, name=None, years=None, create_cert=False, cert=None, scopes=None, role='Contributor',
+        cmd, name=None, years=None, create_cert=False, cert=None, scopes=None, role=None,
```
jiasli, line 1405: Need to detect whether `role` is explicitly set and show the warning if not. 

If `role='Contributor'` is used, it won't be possible to detect if `role` is provided by the user or as the default value. 

See https://stackoverflow.com/a/14749388

src/azure-cli/azure/cli/command_modules/role/custom.py
```diff
@@ -1479,8 +1484,11 @@ def create_service_principal_for_rbac(
 
     # retry while server replication is done
     if not skip_assignment:
+        if not role:
+            role = "Contributor"
+            logger.warning(ROLE_ASSIGNMENT_CREATE_WARNING)
         for scope in scopes:
-            logger.warning('Creating a role assignment under the scope of "%s"', scope)
+            logger.warning("Creating '%s' role assignment under scope '%s'", role, scope)
```
jiasli, line 1491: Echo the role name to be more explicit.

yonzhan:
Role

evelyn-ys:
Shall we mark `--skip-assignment` as deprecated?

jiasli:
> Shall we mark `--skip-assignment` as deprecated this time?

No, as it will be become the **default behavior**, instead of **deprecated**. Marking it as **deprecated** will further confuse the user.

evelyn-ys (approved):

yonzhan (approved):
LGTM

src/azure-cli/azure/cli/command_modules/role/custom.py
```diff
@@ -39,6 +39,11 @@
     "The output includes credentials that you must protect. Be sure that you do not include these credentials in "
     "your code or check the credentials into your source control. For more information, see https://aka.ms/azadsp-cli")
 
+ROLE_ASSIGNMENT_CREATE_WARNING = (
+    "In a future release, this command will NOT create any role assignment by default. (--skip-assignment will be the "
```
dbradish-microsoft, line 43: @jiasli , "not" + "any" is a "double negative".  Change to 

"In a future release, this command will NOT create a role assignment by default..."

dbradish-microsoft (changes requested):

src/azure-cli/azure/cli/command_modules/role/custom.py
jiasli:
```suggestion
    "In a future release, this command will NOT create a role assignment by default. (--skip-assignment will be the "
```

jiasli: Thanks @dbradish-microsoft

dbradish-microsoft (approved):

src/azure-cli/azure/cli/command_modules/role/custom.py
jiasli: Another option is to check `role.is_default`, and show the warning message accordingly. If `role` is assigned a default value, its type will be `knack.validators.DefaultStr` which has `is_default` field indicating whether it is the default value.

PR #16083, opened 2020-11-30
{Upgrade} az upgrade: Handle NoTTYException and fix extension upgrade for homebrew

fengzhou-msft:
**Description**

Handle `NoTTYException`.
Fix the issue that extensions cannot be upgraded after `azure-cli` is upgraded for `homebrew` package.
Add version comparison after upgrade is done.

**Testing Guide**

**History Notes**

---


src/azure-cli/azure/cli/__main__.py
```diff
@@ -72,35 +72,46 @@ def cli_main(cli, args):
...
-        if az_cli.config.getboolean('auto-upgrade', 'enable', False) and \
+        if exit_code == 0 and az_cli.config.getboolean('auto-upgrade', 'enable', False) and \
                 sys.argv[1] != 'upgrade' and (sys.argv[1] != 'extension' or sys.argv[2] != 'update'):
             from azure.cli.core._session import VERSIONS  # pylint: disable=ungrouped-imports
             from azure.cli.core.util import get_cached_latest_versions, _VERSION_UPDATE_TIME  # pylint: disable=ungrouped-imports
             if VERSIONS[_VERSION_UPDATE_TIME]:
                 import datetime
                 version_update_time = datetime.datetime.strptime(VERSIONS[_VERSION_UPDATE_TIME], '%Y-%m-%d %H:%M:%S.%f')
-                if datetime.datetime.now() > version_update_time + datetime.timedelta(days=10):
+                if datetime.datetime.now() > version_update_time + datetime.timedelta(days=11):
```
zhoxing-ms, line 82: > (days=11)

May I ask why the interval of update time need be changed?

src/azure-cli/azure/cli/command_modules/util/custom.py
```diff
@@ -130,19 +136,37 @@ def upgrade_version(cmd, update_all=None, yes=None):  # pylint: disable=too-many
...
         sys.exit(exit_code)
-    # Updating Azure CLI and extension together is not supported in homebrewe package.
+
+    import azure.cli.core
+    import importlib
+    importlib.reload(azure.cli.core)
+    new_version = azure.cli.core.__version__
+
+    if update_cli and new_version == local_version:
+        err_msg = "CLI upgrade failed or aborted."
```
zhoxing-ms, line 150: Why not prompt users `The local version is already the new version and does not need to be updated` here, but report an error?

src/azure-cli/azure/cli/command_modules/util/custom.py
```diff
@@ -130,19 +136,37 @@ def upgrade_version(cmd, update_all=None, yes=None):  # pylint: disable=too-many
...
+        telemetry.set_failure(err_msg)
         sys.exit(exit_code)
-    # Updating Azure CLI and extension together is not supported in homebrewe package.
+
+    import azure.cli.core
+    import importlib
+    importlib.reload(azure.cli.core)
+    new_version = azure.cli.core.__version__
+
+    if update_cli and new_version == local_version:
```
zhoxing-ms, line 149: Why not compare `new_version` and `latest_version` to determine whether the upgrade is successful?

src/azure-cli/azure/cli/__main__.py
fengzhou-msft: Considering our release cycle is 21 days and 2 checks of 10 days may miss a new version while 2 checks of 11 days would guanrantee a discovery of new versions.

src/azure-cli/azure/cli/command_modules/util/custom.py
fengzhou-msft: To avoid issues caused by the case that a new version is released during user's upgrade epspecially for unplanned hot fix release, so the `latest_version` retrieved earlier may be smaller than the actual updated `new_version`, which is a false alarm here.

src/azure-cli/azure/cli/__main__.py
zhoxing-ms: Got it

src/azure-cli/azure/cli/command_modules/util/custom.py
zhoxing-ms: Got it

zhoxing-ms (approved):

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -425,15 +425,16 @@ def show_updates_available(new_line_before=False, new_line_after=False):
     if updates_available_components:
         if new_line_before:
             logger.warning("")
-        show_updates(updates_available_components)
+        show_updates(updates_available_components, only_show_when_updates_available=True)
```
fengzhou-msft, line 428: The method is called in places like `az xx --help`, only show version message when updates available.

arrownj (approved):
LGTM

src/azure-cli/azure/cli/command_modules/util/custom.py
```diff
@@ -130,19 +136,37 @@ def upgrade_version(cmd, update_all=None, yes=None):  # pylint: disable=too-many
...
-    logger.warning("Upgrade finished.")
+        importlib.reload(subprocess)
+
+    if exts:
+        logger.warning("Upgrading extensions")
+    for ext_name in exts:
+        try:
+            logger.warning("Checking update for %s", ext_name)
+            subprocess.call(['az', 'extension', 'update', '-n', ext_name],
+                            shell=platform.system() == 'Windows')
```
Juliehzl, line 165: How the subprocess could be called in homebrew?
It seems that you will call subprocess in homebrew with shell=False, right?

haroldrandom (approved):

src/azure-cli/azure/cli/command_modules/util/custom.py
fengzhou-msft: Right, on macOS, shell=False should be used to call subprocess for the homebrew command to run correctly.

PR #16108, opened 2020-12-02
{RDBMS} Update error message if user passes server name in upper case

arde0708:
This PR updates the error message to show to the user if they pass a server name containing UPPERCASE characters.


yonzhan:
RDBMS

evelyn-ys (approved):
👌

PR #16124, opened 2020-12-02
[AppService] Fixes #16041: az webapp config ssl create results in unknown error

### LINKED ISSUE #16041: WebApp:Creation of an ssl certificate for an App Service ('az webapp config ssl create') results in an 'UnkownError'
**Describe the bug**
The creation of an ssl certificate for an App Service via the command _az webapp config ssl create_ results in an _UnknownError_.

**To Reproduce**
Create a Resource Group, App Service Plan (Linux) and App Service.
Create a CNAME record and add a hostname to the app service:
`
az webapp config hostname add --resource-group {{resource-group-name}} --webapp-name {{app-service-name}} --hostname {{hostname}}
`

When trying to create an ssl certificate an error is returned:
`az webapp config ssl create --hostname {{hostname}} --name {{app-service-name}} --resource-group {{resource-group-name}}`

Result: 'UnknownError: Unknown error'

**Expected behavior**
There should be no error and an ssl certificate should be created in the App Service.

**Environment summary**
Install method: MSI
CLI Version: 2.15.1
OS Version: Windows 10
Shell: cmd.exe

calvinsID:
**Description**
Fixes #16041 

`az webapp config ssl create` results in an unknown error. This is caused by python SDK not set to handle 202 responses meant for polling

- Will make update to SDK to handle 202 response, but in the mean time we can catch when SDK throws error due to 202 response and poll on the certificate to see when its created

**Testing Guide**
- `az webapp config hostname add --resource-group {{resource-group-name}} --webapp-name {{app-service-name}} --hostname {{hostname}}`
- `az webapp config ssl create --hostname {{hostname}} --name {{app-service-name}} --resource-group {{resource-group-name}}`

---


src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2465,8 +2465,26 @@ def create_managed_ssl_cert(cmd, resource_group_name, name, hostname, slot=None)
...
                                 server_farm_id=server_farm_id, password='')
-    return client.certificates.create_or_update(name=hostname, resource_group_name=resource_group_name,
-                                                certificate_envelope=easy_cert_def)
+
+    try:
+        return client.certificates.create_or_update(name=hostname, resource_group_name=resource_group_name,
+                                                    certificate_envelope=easy_cert_def)
+    except DefaultErrorResponseException as ex:
+        if ex.response.status_code == 202:
+            retry_times = 12
```
panchagnula, line 2474: Do we need the re-try lmit I was hoping CLI/client will have a threshold for time out otherwise polling on the operation status using GET should return 200 in <2 mins or non apex domains in my testing with portal for the same

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2465,8 +2465,26 @@ def create_managed_ssl_cert(cmd, resource_group_name, name, hostname, slot=None)
...
+            retry_delay = 5
+            while retry_times > 0:
```
panchagnula, line 2476: this I would check to see while the status of the GET response is not 200.

panchagnula (changes requested):
Not quite what I had in mind - is there a reason we are not using the location URL from the Async API response - does the LongRunningOperation on CLI on do this?

src/azure-cli/azure/cli/command_modules/appservice/custom.py
calvinsID: Updated to use 2 minute timeout

calvinsID:
Currently not using LongRunningOperation since API doesn't support it. Added comment to change this to use LongRunningOperation once supported

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2465,8 +2465,30 @@ def create_managed_ssl_cert(cmd, resource_group_name, name, hostname, slot=None)
...
+                time.sleep(5)
+                r = send_raw_request(cmd.cli_ctx, method='get', url=poll_url)
+
+            if r.status_code == 200:
+                try:
+                    return r.json()
+                except ValueError:
+                    return r.text
+            logger.warning("Managed Certificate creation in progress. Please use the command "
+                           "'az webapp config ssl list' to view your certificate once it is created")
```
panchagnula, line 2489: Just NIT: is there a show command to get just the single cert they should query for - the List might be large & so not useful in cases. if we do single done we should give them fulll command they need to run like az webapp config ssl show -name <certName>

panchagnula (approved):
LGTM - just one minor comment & do we not have any test cases that need to re-recorded?

src/azure-cli/azure/cli/command_modules/appservice/custom.py
calvinsID: Changed

src/azure-cli/azure/cli/command_modules/appservice/custom.py
calvinsID: Added show command and updated msg

calvinsID:
The tests are only for import upload and bind currently, so no need for re-recording

panchagnula:
@qwordy can we get this in S179 please? Thanks

src/azure-cli/azure/cli/command_modules/appservice/_help.py
```diff
@@ -1282,6 +1291,15 @@
     crafted: true
 """
 
+helps['webapp config ssl show'] = """
+type: command
+short-summary: Show the details of an SSL certificate for a web app.
+examples:
+  - name: Show the details of an SSL certificate for a web app. (autogenerated)
+    text: az webapp config ssl show --resource-group MyResourceGroup --certificate-name cname.mycustomdomain.com
```
panchagnula, line 1299: we should update this in the next milestone to use thumprint as optional parameter instead of certname as well since in portal we mostly auto generate certname so users might know the cert name.

qwordy:
https://pypi.org/project/azure-core/
This package has some pollers to use. Maybe you can try them instead of writing it again.

qwordy:
> 
> 
> @qwordy can we get this in S179 please? Thanks

OK

qwordy (approved):

panchagnula:
> > @qwordy can we get this in S179 please? Thanks
> 
> OK

Thank you. 🙏

PR #16149, opened 2020-12-04
[ARM] az ts:  Add support for --tags

detienne20:
**Description**

Template Specs create and update now supports --tags. 

**Testing Guide**

                               **The test cases added cover all of the following**

(1) 
az ts create -g {rg} -n {template_spec_name} -v 1.0 -l {resource_group_location} -f "{tf}" --tag cli-test=test 

     Expected Result(s):

          If the parent template spec does not yet exist, the tag(s) will apply to both the parent (template spec) and the child 
          (template spec version). 

           If the parent template spec exists, only the child (template spec version) will be assigned the tag. Note: A child ( 
           template spec version) can have different tags from it's parent (template spec). Changing the tags of the child does not 
           affect the parent. 

(2) 

az ts create -g {rg} -n {template_spec_name} -v 1.0 -l {resource_group_location} -f "{tf}"

       Expected Result(s): 
            
          If the parent (template spec) does not yet exist, no tags will be applied to the parent or the child (tags={}) 
          
          If the parent (template spec) exists then the new child will inherit the parents' tags or lack of tags. 

          If the child (template spec version) already exists this command will function the same. Create will override the child 
          (template spec version). 

(3a.)

az ts update -g {rg} -n {template_spec_name} -v 1.0 -l {resource_group_location} -f "{tf}" --tags --yes 

az ts update -g {rg} -n {template_spec_name} -v 1.0 -l {resource_group_location} -f "{tf}" --tags "" --yes

az ts update -g {rg} -n {template_spec_name} -v 1.0 -l {resource_group_location} -f "{tf}" --tags  '' --yes 

     Expected Result(s):
          All the commands listed above in #3 should result in tags being removed (tags = {}). When updating a template 
          spec or template spec version tags should only be removed if explicitly empty. 

(3b.) 

az ts update -g {rg} -n {template_spec_name} -v 1.0 -l {resource_group_location} -f "{tf}" --yes 

az ts update -g {rg} -n {template_spec_name} -v 1.0 -l {resource_group_location} -f "{tf}"  --yes

az ts update -g {rg} -n {template_spec_name} -v 1.0 -l {resource_group_location} -f "{tf}" --yes 

     Expected Result(s):
          Tags should be unchanged. 

   


yonzhan:
ARM

src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -2201,11 +2201,11 @@
 short-summary: Create a template spec and or template spec version.
 examples:
   - name: Create a template spec.
-    text: az ts create -g testRG --name TemplateSpecName -l WestUS --display-name "MyDisplayName" --description "Simple template spec"
+    text: az ts create -g testRG --name TemplateSpecName -l WestUS --display-name "MyDisplayName" --description "Simple template spec" --tag key1=value1
```
Juliehzl, line 2204:
```suggestion
    text: az ts create -g testRG --name TemplateSpecName -l WestUS --display-name "MyDisplayName" --description "Simple template spec" --tags key1=value1
```

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -540,6 +540,8 @@ def load_arguments(self, _):
         c.argument('display_name', arg_type=ts_display_name_type)
         c.argument('description', arg_type=ts_description_type)
         c.argument('version_description', arg_type=ts_version_description_type)
+        c.argument('tag', tag_type)
+        c.argument('tags', tags_type)
```
Juliehzl, line 544: what is the difference for tag and tags?

src/azure-cli/azure/cli/command_modules/resource/_validators.py
```diff
@@ -16,7 +16,9 @@
 MSI_LOCAL_ID = '[system]'
 
 
-def _validate_template_input(namespace):
+def process_ts_create_or_update_namespace(namespace):
+    from azure.cli.core.commands.validators import validate_tags
+    validate_tags(namespace)
```
zhoxing-ms, line 21:
```
tags_type = CLIArgumentType(
    validator=validate_tags,
    help="space-separated tags: key[=value] [key[=value] ...]. {}".format(quote_text),
    nargs='*'
)
```

`tags_type` already has the logic to call `validate_tags`, why do you call it repeatedly here?

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -1848,7 +1852,12 @@ def update_template_spec(cmd, resource_group_name=None, name=None, template_spec
         existing_template = rcf.template_spec_versions.get(resource_group_name=resource_group_name, template_spec_name=name, template_spec_version=version)
 
         location = getattr(existing_template, 'location')
-        version_tags = getattr(existing_template, 'tags')
+
+        version_tags = tags
+        if tags is None:  # Do not remove tags if not explicitely empty.
+            version_tags = getattr(existing_template, 'tags')
+        else:
+            version_tags = tags
```
zhoxing-ms, line 1860:
```
        else:
            version_tags = tags
```

The logic is redundant?

src/azure-cli/azure/cli/command_modules/resource/_params.py
detienne20: Customer can use tag for a single tag and tags for multiple.

src/azure-cli/azure/cli/command_modules/resource/_validators.py
detienne20: To set tags in the namespace to the tags dictionary.
Ref: https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/network/_validators.py for similar usage.

zhoxing-ms (approved):

src/azure-cli/azure/cli/command_modules/resource/_params.py
Juliehzl: could we consider that use --tags for both single and multiple tags? I mean if there is one tag, use `--tags key=value1`, which would be more consistent with other commands. `--tag` and `--tags` would make cx confused

detienne20: @Juliehzl the customer can use either one. --tags works for a single tag as well. --tags and --tag map to the same input.  I expanded functionality for --tag since the code for this functionality is already available in the code based and used for other modules (try: az network vnet create --help)

Juliehzl: it seems there is only `--tags` in `az network vnet create`
image

Juliehzl: as you know, `--tags` could support one or multiple tags, so I think it is unnecessary to add `--tag` and it would align with other commands

detienne20: @Juliehzl  Done.

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -540,6 +540,8 @@ def load_arguments(self, _):
         c.argument('display_name', arg_type=ts_display_name_type)
         c.argument('description', arg_type=ts_description_type)
         c.argument('version_description', arg_type=ts_version_description_type)
+        c.argument('tag', tag_type)
```
Juliehzl, line 543: should we remove this argument here?

detienne20: @Juliehzl Done.

src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -2218,11 +2218,11 @@
...
```
Juliehzl, line 2221:
```suggestion
    text: az ts create -g testRG --name TemplateSpecName -l WestUS --display-name "MyDisplayName" --description "Simple template spec" --tags key1=value1
```

src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -2231,12 +2231,15 @@
 examples:
   - name: Update the template content of a template spec or template spec version based on the resource ID.
     text: az ts update --template-spec resourceID -f updatedFile.json
-  - name: Update the display name of a template spec based on the resource ID.
-    text: az ts update --template-spec resourceID --display-name "NewParentDisplayName"
-  - name: Update the description of a template spec version.
-    text: az ts update -g ExistingRG --name ExistingName -v 3.0 --version-description "New description"
+  - name: Update the display name and tag(s) of a template spec based on the resource ID.
+    text: az ts update --template-spec resourceID --display-name "NewParentDisplayName" --tag key1=value1
```
Juliehzl, line 2235:
```suggestion
    text: az ts update --template-spec resourceID --display-name "NewParentDisplayName" --tags key1=value1
```

Juliehzl (approved):

PR #16158, opened 2020-12-04
{RDBMS} Update location against which tests are executed

arde0708:
For the module RDBMS, we have two kind of servers that gets created in the service - servers and flexible servers. We have different commands for both and the backend trains are also different. Given that our backend trains run on a different schedule as compared to the azure-cli train, it is becoming difficult and time consuming for us to validate the commands every time we have to sign off on the CLI commands after either of the backend train is deployed.

Our stage region is ‘southeastasia’. However, our tests are currently targeted to prod --eastus and westus2. This PR changes the location to southeastasia and contains the re-recorded files for the tests.


yonzhan:
RDBMS

src/azure-cli/azure/cli/command_modules/rdbms/tests/latest/test_rdbms_flexible_commands.py
```diff
@@ -829,14 +829,14 @@ def _test_flexible_server_public_access_mgmt(self, database_engine, resource_gro
...
-                         JMESPathCheck('host', '{}.{}.database.azure.com'.format(servers[0], database_engine))])
+                         StringContainCheck(servers[0])])
 
         # Case 2 : Provision a server with public access allowing all azure services
         self.cmd('{} flexible-server create -g {} -n {} --public-access {}'
                  .format(database_engine, resource_group, servers[1], '0.0.0.0'),
                  checks=[JMESPathCheck('resourceGroup', resource_group), JMESPathCheck('skuname', sku_name),
                          StringContainCheck('AllowAllAzureServicesAndResourcesWithinAzureIps_'),
-                         JMESPathCheck('host', '{}.{}.database.azure.com'.format(servers[1], database_engine))])
+                         StringContainCheck(servers[1])])
```
Juliehzl, line 839: why not check suffix any more?

arde0708: Its the actually just the server name that we want to validate. Removed the domain name because once we run it in the pipeline for the Staging env,the domain name will change. The domain name in stage is :azuredbpaccessm3smp5.orcabrci-seas1-a.mscds.com  and in production is : azuredbpaccessm3smp5.mysql.database.azure.com. As a result hardcoding the domains name in the validation would always fail it in the Stage env

src/azure-cli/azure/cli/command_modules/rdbms/tests/latest/test_rdbms_flexible_commands.py
```diff
@@ -127,9 +126,8 @@ def _test_flexible_server_mgmt(self, database_engine, resource_group):
             self.cmd('mysql flexible-server create -g {} -l {} --tier MemoryOptimized --sku-name Standard_E2ds_v4 --public-access none'
                      .format(resource_group, location))
 
-        show_output = self.cmd('{} flexible-server show -g {} -n {}'
-                               .format(database_engine, resource_group, server_name), checks=list_checks).get_output_in_json()
-        self.assertIn('subnetArmResourceId', show_output["delegatedSubnetArguments"])
+        self.cmd('{} flexible-server show -g {} -n {}'
```
Juliehzl, line 129: it seems get_output_in_json() is not used

arde0708: Yes, we are not using it anyways.

src/azure-cli/azure/cli/command_modules/rdbms/tests/latest/test_rdbms_flexible_commands.py
```diff
@@ -862,6 +860,7 @@ def test_mysql_flexible_server_local_context(self, resource_group):
         self._test_flexible_server_local_context('mysql', resource_group)
 
     def _test_flexible_server_local_context(self, database_engine, resource_group):
+        self.cmd('config param-persist on')
```
evelyn-ys, line 863: Shall we turn off `param-persist` after finishing test?

src/azure-cli/azure/cli/command_modules/rdbms/tests/latest/test_rdbms_flexible_commands.py
```diff
@@ -127,9 +126,8 @@ def _test_flexible_server_mgmt(self, database_engine, resource_group):
             self.cmd('mysql flexible-server create -g {} -l {} --tier MemoryOptimized --sku-name Standard_E2ds_v4 --public-access none'
                      .format(resource_group, location))
 
-        show_output = self.cmd('{} flexible-server show -g {} -n {}'
-                               .format(database_engine, resource_group, server_name), checks=list_checks).get_output_in_json()
-        self.assertIn('subnetArmResourceId', show_output["delegatedSubnetArguments"])
+        self.cmd('{} flexible-server show -g {} -n {}'
+                 .format(database_engine, resource_group, server_name), checks=list_checks).get_output_in_json()
```
evelyn-ys, line 130:
```suggestion
        self.cmd('{} flexible-server show -g {} -n {}'
                 .format(database_engine, resource_group, server_name), checks=list_checks)
```

arde0708: Addressed

src/azure-cli/azure/cli/command_modules/rdbms/tests/latest/test_rdbms_flexible_commands.py
arde0708: Turned off

evelyn-ys (approved):

Juliehzl (approved):

PR #16169, opened 2020-12-07
{Role} az role assignment create: Refine unclear help message of `--assignee-object-id`

jiasli:
**Description**
The current description of `--assignee-object-id` is unclear about "graph permission issues":

```
    --assignee-object-id          : Use this parameter instead of '--assignee' to bypass graph
                                    permission issues.
```

Replace it with more explicit description:

```
    --assignee-object-id          : Use this parameter instead of '--assignee' to bypass Graph API
                                    invocation in case of insufficient privileges.
```


yonzhan:
Role

evelyn-ys (approved):

Issue #15944, opened 2020-11-16
2 corrections to the documentation


mossywell:
2 corrections (both in the "create" section. The first is a typo correction. The second is to add clarification.

1. "The resource id of which private enpoint connect to." --> "The resource id of the private endpoint to connect to."

2. "Name or ID of an existing subnet. If name specified, also specify --vnet-name. If you want to use an existing subnet in other resource group or subscription, please provide the ID instead of the name of the subnet." --> "Name or ID of an existing subnet. If the name is specified, also specify --vnet-name. If you want to use an existing subnet in other resource group or subscription, please provide the ID instead of the name of the subnet and do not specify the --vnet-name."

[Enter feedback here]

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 8911e94f-9496-a68d-50c0-b5d536a375a4
* Version Independent ID: cc37b77b-0fa2-991b-e16b-fcdbbf183853
* Content: [az network private-endpoint](https://docs.microsoft.com/en-us/cli/azure/network/private-endpoint?view=azure-cli-latest)
* Content Source: [latest/docs-ref-autogen/network/private-endpoint.yml](https://github.com/MicrosoftDocs/azure-docs-cli/blob/master/latest/docs-ref-autogen/network/private-endpoint.yml)
* Service: **virtual-network**
* GitHub Login: @rloutlaw
* Microsoft Alias: **routlaw**

yonzhan:
network

PR #16174, opened 2020-12-07
[Network] az network private-endpoint create: add more declaration of '--subnet' and '--private-connection-resource-id'

msyyc:
Close https://github.com/Azure/azure-cli/issues/15944

--

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/network/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/network/_params.py
@@ -812,10 +812,10 @@ def load_arguments(self, _):
     with self.argument_context('network private-endpoint') as c:
         c.argument('private_endpoint_name', private_endpoint_name, options_list=['--name', '-n'])
         c.argument('location', get_location_type(self.cli_ctx), validator=get_default_location_from_resource_group)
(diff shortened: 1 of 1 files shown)


yonzhan:
network

haroldrandom (approved):

src/azure-cli/azure/cli/command_modules/network/_params.py
```diff
@@ -812,10 +812,10 @@ def load_arguments(self, _):
     with self.argument_context('network private-endpoint') as c:
         c.argument('private_endpoint_name', private_endpoint_name, options_list=['--name', '-n'])
         c.argument('location', get_location_type(self.cli_ctx), validator=get_default_location_from_resource_group)
-        subnet_help = get_folded_parameter_help_string('subnet', other_required_option='--vnet-name')
+        subnet_help = get_folded_parameter_help_string('subnet', other_required_option='--vnet-name') + ' and do not specify the --vnet-name'
```
jsntcy, line 815: Can we also validate this logic in validator as well?

msyyc: we indeed tell the exclusion in 'help message', but not in the same message block. Maybe the issue owner think it should be more obvious.

PR #16220, opened 2020-12-10
{Style} Support style with color

jiasli:
**Description**

Support style with color.

For a complete demo, see `src/azure-cli/azure/cli/command_modules/util/custom.py` and run `az demo style`.

### DIFF
--- a/src/azure-cli-core/azure/cli/core/style.py
+++ b/src/azure-cli-core/azure/cli/core/style.py
@@ -14,18 +14,17 @@ https://devdivdesignguide.azurewebsites.net/command-line-interface/color-guideli
 For a complete demo, see `src/azure-cli/azure/cli/command_modules/util/custom.py` and run `az demo style`.
 """
 
-import re
+import sys
 from enum import Enum
-from colorama import Fore
 
-current_color = None
+from colorama import Fore
 
 
 class Style(str, Enum):
     PRIMARY = "primary"
     SECONDARY = "secondary"
     IMPORTANT = "important"
-    ACTION = "action"
+    ACTION = "action"  # name TBD
     HYPERLINK = "hyperlink"
     # Message colors
     ERROR = "error"
@@ -48,10 +47,9 @@ THEME = {
 }
 
 
-def print_styled_text(styled):
+def print_styled_text(styled, file=sys.stderr):
     formatted = format_styled_text(styled)
-    import sys
-    print(formatted, file=sys.stderr)
+    print(formatted, file=file)
 
 
 def format_styled_text(styled_text):
--- a/src/azure-cli-core/azure/cli/core/style.py
+++ b/src/azure-cli-core/azure/cli/core/style.py
@@ -1,0 +1,63 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+"""
+Support styled output.
+
+Currently, only color is supported, underline/bold/italic may be supported in the future.
+
+Design spec:
+https://devdivdesignguide.azurewebsites.net/command-line-interface/color-guidelines-for-command-line-interface/
+
+For a complete demo, see `src/azure-cli/azure/cli/command_modules/util/custom.py` and run `az demo style`.
+"""
+
+import re
+from enum import Enum
+from colorama import Fore
+
+current_color = None
+
+
+class Style(str, Enum):
+    PRIMARY = "primary"
+    SECONDARY = "secondary"
+    IMPORTANT = "important"
+    ACTION = "action"
+    HYPERLINK = "hyperlink"
+    # Message colors
+    ERROR = "error"
+    SUCCESS = "success"
+    WARNING = "warning"
+
+
+THEME = {
+    # Style to ANSI escape sequence mapping
+    # https://docs.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences
+    Style.PRIMARY: Fore.LIGHTWHITE_EX,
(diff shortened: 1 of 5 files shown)


yonzhan:
Style

src/azure-cli-core/azure/cli/core/style.py
```diff
@@ -0,0 +1,61 @@
...
+def print_styled_text(styled, file=sys.stderr):
+    formatted = format_styled_text(styled)
+    print(formatted, file=file)
+
+
+def format_styled_text(styled_text):
+    # https://python-prompt-toolkit.readthedocs.io/en/stable/pages/printing_text.html#style-text-tuples
+    formatted_parts = []
+    for text in styled_text:
+        formatted_parts.append(THEME[text[0]] + text[1])
```
houk-ms, line 59: key checking for `THEME`? Or add some error handling logics here?

src/azure-cli/azure/cli/command_modules/util/_help.py
```diff
@@ -45,3 +45,13 @@
 type: command
 short-summary: Upgrade Azure CLI and extensions
 """
+
+helps['demo'] = """
```
houk-ms, line 49: Just curious is this a convention to add a demo command here?  The command is convenient for developers but also visible to users right？

houk-ms (approved):
Good job ~

Can we also provide a simple developer doc later (may in another pr)?

src/azure-cli-core/azure/cli/core/style.py
```diff
@@ -0,0 +1,61 @@
...
+    Style.ACTION: Fore.LIGHTBLUE_EX,
+    Style.HYPERLINK: Fore.LIGHTCYAN_EX,
+    # Message colors
+    Style.ERROR: Fore.LIGHTRED_EX,
+    Style.SUCCESS: Fore.LIGHTGREEN_EX,
+    Style.WARNING: Fore.LIGHTYELLOW_EX,
+}
+
+
+def print_styled_text(styled, file=sys.stderr):
```
zhoxing-ms, line 50: May I ask why the default value of the `file` parameter is `sys.stderr` instead of `sys.stdout`?

jiasli: Only the JSON output should go to `stdout` and the user may pipe it to downstream commands. Per the description of `stderr`:

https://en.wikipedia.org/wiki/Standard_streams#sys.stderr

> This is similar to sys.stdout because it also prints directly to the Console. But the difference is that can be used to print Exceptions and Error messages and also **info/debugging comments**. (Which is why it is called Standard Error). This can be very useful when the stdout is used to write data instead.

zhoxing-ms: OK, got it~

src/azure-cli/azure/cli/command_modules/util/_help.py
jiasli: > Just curious is this a convention to add a demo command here?

No. We don't have such convention. The only existing usage of this pattern is `az self-test`:

https://github.com/Azure/azure-cli/blob/3c54715e05ccc656aaaa31a001485a02fa4ec424/src/azure-cli/azure/cli/command_modules/profile/__init__.py#L32

> The command is convenient for developers but also visible to users right？

It is not visible to the user, as the parent command group has been marked as `hide=True`:

https://github.com/Azure/azure-cli/blob/bc2d0d434fdd410dd510327f60dbd457778d08dd/src/azure-cli/azure/cli/command_modules/util/commands.py#L18-L19

src/azure-cli-core/azure/cli/core/style.py
jiasli: As for `az next`, it is only used interactively, so the stream doesn't really matter as everything is printed to the screen.

jiasli:
> Can we also provide a simple developer doc later (may in another pr)?

Of course. I will add the doc once this feature is stabilized.

zhoxing-ms (approved):
LGTM!
Maybe we can also encapsulate a util to reuse some common pattern prompts in the future, such as: `TASK DONE: Style.SUCCESS + '(✓) Done:' + Style.PRIMARY + ...`

src/azure-cli-core/azure/cli/core/style.py
zhoxing-ms: Yes

src/azure-cli-core/azure/cli/core/style.py
jiasli: Nice catch. Added checks and tests as requested.

jiasli:
> LGTM!
> Maybe we can also encapsulate a util to reuse some common pattern prompts in the future, such as: `TASK DONE: Style.SUCCESS + '(✓) Done:' + Style.PRIMARY + ...`

Nice suggestion. Let's do this until there is some real usage so that the design/quality can be verified.

src/azure-cli-core/azure/cli/core/style.py
```diff
@@ -0,0 +1,74 @@
...
+from enum import Enum
+
+from colorama import Fore
+
+
+class Style(str, Enum):
+    PRIMARY = "primary"
+    SECONDARY = "secondary"
+    IMPORTANT = "important"
+    ACTION = "action"  # name TBD
```
jiasli, line 27: @chenlomis, what style name should we use for light blue?

zhoxing-ms (approved):
LGTM

houk-ms (approved):

src/azure-cli-core/azure/cli/core/style.py
chenlomis: Hi there
Am slightly confused
Did you mean finding an internal short name for the the bright blue?
https://devdivdesignguide.azurewebsites.net/command-line-interface/color-guidelines-for-command-line-interface/#action-colors

PR #16221, opened 2020-12-10
{Test} Fix or skip live tests for cdn, privatedns, reservations

fengzhou-msft:
**Description**

Make some tests record_only as they rely on existing resources in another subscription or require extra permissions such as reservation purchases.

Add retry for reimporting network private-dns zone as we are seeing more failures due to resource privioning not finished before calling the command.

**Testing Guide**

Test results: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=648963&view=artifacts&pathAsName=false&type=publishedArtifacts

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/netappfiles/tests/latest/test_backup_policy_commands.py
+++ b/src/azure-cli/azure/cli/command_modules/netappfiles/tests/latest/test_backup_policy_commands.py
@@ -3,7 +3,7 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 from azure.cli.testsdk import ScenarioTest, ResourceGroupPreparer
-LOCATION = "eastus2euap"
+LOCATION = "eastus2"
 
 
 class AzureNetAppFilesBackupPolicyServiceScenarioTest(ScenarioTest):
--- a/src/azure-cli/azure/cli/command_modules/netappfiles/tests/latest/test_backup_policy_commands.py
+++ b/src/azure-cli/azure/cli/command_modules/netappfiles/tests/latest/test_backup_policy_commands.py
@@ -3,7 +3,7 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 from azure.cli.testsdk import ScenarioTest, ResourceGroupPreparer
-LOCATION = "eastus2"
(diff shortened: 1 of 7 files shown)


src/azure-cli/azure/cli/command_modules/netappfiles/tests/latest/test_backup_policy_commands.py
```diff
@@ -3,7 +3,7 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 from azure.cli.testsdk import ScenarioTest, ResourceGroupPreparer
-LOCATION = "eastus2euap"
+LOCATION = "eastus2"
```
fengzhou-msft, line 6: `eastus2euap` is not avaliable for the service now.

qwordy:
Skip does not fix them. Should we skip test that we can't fix to let data be beautiful?

haroldrandom (approved):

haroldrandom (changes requested):
`test_aro_delete` Could you rerun the test? Or replace the network API version

fengzhou-msft:
> Skip does not fix them. Should we skip test that we can't fix to let data be beautiful?

These tests rely on existing resources of another subscription and when the live test pipeline tests them with our subscription, it would have resource not found error or permission error.

The best way to fix it is to rewrite the tests to create the resources during the test if the resource does not require extra permissions or we apply for relative permissions (we have fixed some tests in this way before). 

The second best way is to create resources in our subscription and lock them for usage in live tests if the creation of such resources is very time-consuming. 

The least we want to do is to skip the test, but it's difficult to fix some tests due to permission issues or complexity. The failure of such tests does not tell us the real issues. I think it's OK to skip them now. When we or service teams have more resources, we can rewrite those tests. I think we will evaulate test coverage sometime in the future.

fengzhou-msft:
> `test_aro_delete` Could you rerun the test? Or replace the network API version to fix CI?

submitted #16226

haroldrandom (approved):

src/azure-cli/azure/cli/command_modules/netappfiles/tests/latest/test_backup_policy_commands.py
fengzhou-msft: reverted the change as the new location has permission errors, will leave it to service team to fix.

PR #16229, opened 2020-12-10
[SQL] az sql mi create/update: Add support for maintenance configuration in managed instance operations

kukislav:
**Description**
 - Add option to create and update managed instance with maintenance configuration (exposing FMW feature)

---


yonzhan:
SQL

kukislav:
I checked the linter pipeline that fails and fixed all issues associated with my command. What should I do about others like this one?
image

kukislav:
This is not customer reported as bot labelled it, this is a new feature.

evelyn-ys:
> I checked the linter pipeline that fails and fixed all issues associated with my command. What should I do about others like this one?
> image

This is fixed in #16297. Please merge the latest 'dev' branch into your branch to pass CI

kukislav:
@evelyn-ys I have updated code and fixed everything. Please review.

src/azure-cli/azure/cli/command_modules/sql/custom.py
```diff
@@ -3637,6 +3638,8 @@ def managed_instance_update(
     if tags is not None:
         instance.tags = tags
 
+    instance.maintenance_configuration_id = maintenance_configuration_id
```
evelyn-ys, line 3641:
```suggestion
    if maintenance_configuration_id is not None:
        instance.maintenance_configuration_id = maintenance_configuration_id
```
We only need to update this when given value is not `None`, right?

src/azure-cli/azure/cli/command_modules/sql/tests/latest/test_sql_commands.py
```diff
@@ -3510,6 +3507,28 @@ def test_sql_managed_instance_mgmt(self):
                                           JMESPathCheck('tags', "{'tagName1': 'tagValue1', 'tagName2': 'tagValue2'}"),
                                           JMESPathCheck('storageAccountType', backup_storage_redundancy_internal)]).get_output_in_json()
 
+        maintenance_configuration_id = '/subscriptions/a295933f-f7f5-4994-a109-8fa51241a5d6/providers/Microsoft.Maintenance/publicMaintenanceConfigurations/SQL_EastUS2EUAP_MI_2'
+        subnet = '/subscriptions/a295933f-f7f5-4994-a109-8fa51241a5d6/resourceGroups/fmwtest/providers/Microsoft.Network/virtualNetworks/vnet-fmwm2/subnets/ManagedInstance'
```
evelyn-ys, line 3511: May I ask why using fixed resources? If these resources are deleted tests will fail then. How about prepare temporary resource before test and delete them when tests finish?

You can take storage module as an example: https://github.com/Azure/azure-cli/blob/4ac1441c518630f3032caf135b9fdfbd9392cbd3/src/azure-cli/azure/cli/command_modules/storage/tests/latest/test_storage_account_scenarios.py#L40-L52 To test `storage account network-rule`, `network vnet` is called to prepare a vnet and subnet

kukislav: Regarding subnet_id this test already uses it as a premade resource I just did the same.

Regarding maintenance configuration id, these are not physical resources, they are fixed and exist publicly for everyone to use them. Via get request one can get details about policy such as next maintenance window start time, duration, days when maintenance occurs, etc. This is real config name `SQL_EastUS2EUAP_MI_2` and sub is different for every user because MRP provider require it that way.

src/azure-cli/azure/cli/command_modules/sql/custom.py
kukislav: I think we don't need this check. If maintenance_configuration_id is not passed the API won't remove this property. We made sure it doesn't. The only way to remove property is to pass a sentinel value like this: `/subscriptions/a295933f-f7f5-4994-a109-8fa51241a5d6/providers/Microsoft.Maintenance/publicMaintenanceConfigurations/SQL_Default`

kukislav: Does this sound reasonable?

evelyn-ys (approved):
Please resolve conflicts before merging

PR #16238, opened 2020-12-10
[ACR] az acr connected-acr create/update/delete/list/show: new feature in beta.

rosanch:
**Description**
New ACR on-prem feature called connected registry. Allows client to create and manager connected registry in their own hardware.
Adds scope-map and token support for Gateway permissions. 

**Testing Guide**
First create a registry and at least one repo. Then run: 
az acr connected-registry create --registry mycloudregistry --name myconnectedregistry --repository "app/hello-world service/mycomponent"
az acr list --registry mycloudregistry -o table

**History Notes**
[ACR scope-map] az acr scope-map update: --add and --remove are deprecated, they are renamed to --add-repo --remove-repo
[ACR scope-map] az acr scope-map create/update: Add support to handle Gateway actions.
[ACR Token]  az acr token create: support added for gateway actions
[ACR connected-registry] az acr connected-registry: New feature for on-prem Azure Container Registry

---


yonzhan:
ACR

src/azure-cli/azure/cli/command_modules/acr/_format.py
```diff
@@ -93,6 +93,55 @@ def agentpool_output_format(result):
...
+def connected_registry_output_format(result):
+    return _output_format(result, _connected_registry_format_group)
+
+
+def connected_registry_list_output_format(result):
+    family_tree = {}
+    for reg in result:
+        parent_id = _get_value(reg, 'parent', 'id')
+        parent_name = '' if parent_id.isspace() else parent_id.split('/connectedRegistries/')[1]
+        name = ' ' + _get_value(reg, 'name')
```
northtyphoon, line 105: what is `' '` prefix used for?

src/azure-cli/azure/cli/command_modules/acr/_format.py
```diff
@@ -93,6 +93,55 @@ def agentpool_output_format(result):
...
+        family_tree[_get_value(reg, 'id')] = {
+            "name": name,
+            "id": _get_value(reg, 'id'),
+            "connectionState": _get_value(reg, 'connectionState'),
```
northtyphoon, line 109: is `conectionState` value used? if used, is the name supposed to be connection_state?

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -528,6 +528,11 @@
   - name: Create a scope map that allows content/write and metadata/read actions for `hello-world` repository, and content/read action for `hello-world-again`.
     text: >
         az acr scope-map create -n MyScopeMap -r MyRegistry --repository hello-world content/write metadata/read --repository hello-world-again content/read --description "Sample scope map."
+  - name: Create a scope map that allows content/write and metadata/read actions for `hello-world` repository, and all gateway actions for `connectedRegistry`.
```
northtyphoon, line 531: I would expect most cases need content/read to do image pull. So I would suggest the sample includes content/read and metedata/write

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -1204,6 +1210,115 @@
         az acr webhook update -n MyWebhook -r MyRegistry --status disabled
 """
 
+# region connected-registry
+helps['acr connected-registry'] = """
+type: group
+short-summary: Manage Edge/On-Prem Container Registries with Azure Container Registries.
```
northtyphoon, line 1217: Are we supposed to use Edge/On-Prem? How about 

Manage connected registry resources with Azure Container Registries?

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -1204,6 +1210,115 @@
...
+examples:
+  - name: Delete a mirror connected registry 'myconnectedregistry' from parent registry 'mycloudregistry' and skip verification.
+    text: |
+        az acr connected-registry delete --registry mycloudregistry 
+            --name myconnectedregistry --yes
+"""
+
+helps['acr connected-registry install-info'] = """
+type: command
+short-summary: Retrives information required for to activate a connected registry.
```
northtyphoon, line 1252: `for to` => required `to` activate

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -381,6 +390,35 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('key_encryption_key', help="Key vault key uri. To enable automated rotation, provide a version-less key uri. For manual rotation, provide a versioned key uri.")
         c.argument('identity', help="client id of managed identity, resource name or id of user assigned identity. Use '[system]' to refer to the system assigned identity")
 
+    with self.argument_context('acr connected-registry') as c:
+        c.argument('registry_name', options_list=['--registry', '-r'], help='The login server of the Cloud ACR registry. Must be the FQDN to support also Azure Stack.', required=True)
+        c.argument('connected_registry_name', options_list=['--name', '-n'], help='Name for the Connected ACR. Name must start with a letter and contain only alphanumeric characters (including ‘_’) optionally separated by a ‘-’. 5 to 40 chars long. Name must be unique under the Cloud ACR hierarchy.', required=True)
```
northtyphoon, line 395: including `_` or `-`?

src/azure-cli/azure/cli/command_modules/acr/_format.py
rosanch: That was to add an space before the '-' so when the users double click in the connected registry name it doesn't select as well all the '-'. And the table format doesn't print spaces so if they where level 0 it would be remove before printing. 
I have a todo to remove if the tree format is not shipped.

src/azure-cli/azure/cli/command_modules/acr/_params.py
northtyphoon, line 395: Connected ACR => connected registry

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -381,6 +390,35 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
...
+        c.argument('resource_group_name', options_list=['--resource-group', '-g'], help='The name of the resource group where the Connected ACR registry will be created. Must be the same as the resource group for the --registry.', required=False)
```
northtyphoon, line 396: ditto

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -381,6 +390,35 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
...
+        c.argument('parent', options_list=['--parent', '-p'], help='The name of the parent Connected Registry.', required=False)
+        c.argument('repositories', options_list=['--repository', '-t'], nargs='+', help='Specifies the repositories that need to be synched to the Connected ACR. It can be in the format [REPO01] [REPO02]...', required=False)
```
northtyphoon, line 398: synched => typo?

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -381,6 +390,35 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
...
+        c.argument('sync_token_name', options_list=['--sync-token'], help='Specifies the sync token needed to be synched to the Connected ACR. It can include access for multiple repositories.', required=False)
```
northtyphoon, line 399: synched => typo?

src/azure-cli/azure/cli/command_modules/acr/_format.py
rosanch: that's just the key name used. I prefer not to change so the code is easier to follow. 
As for the the title name printed, that is in line 315 and it was the last name suggested in the meeting since it was mentioned to use status as misleading. 
If you have any other suggestion, let me know. I'll change the tittle name.

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -381,6 +390,35 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
...
+        c.argument('fresh_install', options_list=['--fresh-install'], help='Creates a new password for the sync token.', required=False)
+        c.argument('no_children', options_list=['--no-children'], help='Used to remove all children from the list.', required=False, action='store_true')
+        c.argument('sync_audit_logs_enabled', options_list=['--sync-audit-logs-enabled'], help='Indicates whether audit log synchronization is enabled. It\'s enabled by default.', required=False, arg_type=get_three_state_flag())
```
northtyphoon, line 402: It is

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -381,6 +390,35 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
...
+
+    with self.argument_context('acr connected-registry create') as c:
+        c.argument('log_level', options_list=['--log-level'], help='Sets the log level for logging on the instance. Accepted log levels are Debug, Information, Warning, Error, and None. ', required=False, default="Information")
```
northtyphoon, line 405: None. ' => extra space

src/azure-cli/azure/cli/command_modules/acr/_help.py
rosanch: I'll change the example to add all the actions.

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -381,6 +390,35 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
...
+        c.argument('mode', options_list=['--mode', '-m'], help='Can be one of the two operating modes: registry or mirror(pull-only mode).', required=False, default="registry")
+        c.argument('client_token_list', options_list=['--client-tokens'], nargs='+', help='Specifies the client access to the repositories in the connected registry. It can be in the format [TOKEN_NAME01] [TOKEN_NAME02]...', required=False)
+        c.argument('sync_window', options_list=['--sync-window', '-w'], help='Required parameter if --sync-schedule is present. Used to determine the schedule duration. Uses ISO 8601 duration format.', required=False)
```
northtyphoon, line 408: what is the default value of sync window?

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -381,6 +390,35 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
...
+        c.argument('sync_schedule', options_list=['--sync-schedule', '-s'], help='Optional parameter to define the sync schedule. Uses cron expression to determine the schedule. If not specified, the instance is considered always online and attempts to sync every minute.', required=False, default="* * * * *")
+        c.argument('sync_message_ttl', options_list=['--sync-message-ttl'], help='Determines how long the sync messages will be kept in the cloud. Uses ISO 8601 duration format.', required=False, default="P2D")
+
+    with self.argument_context('acr connected-registry update') as c:
+        c.argument('log_level', options_list=['--log-level'], help='Sets the log level for logging on the instance. Accepted log levels are Debug, Information, Warning, Error, and None. ', required=False)
```
northtyphoon, line 413: and None. ' => extra space

src/azure-cli/azure/cli/command_modules/acr/_utils.py
```diff
@@ -469,3 +453,125 @@ def prepare_source_location(cmd, source_location, client_registries, registry_na
...
+    scope_map = LongRunningOperation(cmd.cli_ctx)(poller)
+    return scope_map
+
+
+def build_token_id(subscription_id, resource_group_name, registry_name, token_name):
+    return "/subscriptions/{}/resourceGroups/{}".format(subscription_id, resource_group_name) + \
+            "/providers/Microsoft.ContainerRegistry/registries/{}/tokens/{}".format(registry_name, token_name)
+
+
+def get_token_from_id(cmd, token_id):
```
northtyphoon, line 549: do we have a shared utility method to handle the parsing?

src/azure-cli/azure/cli/command_modules/acr/_utils.py
```diff
@@ -469,3 +453,125 @@ def prepare_source_location(cmd, source_location, client_registries, registry_na
...
+    from ._client_factory import cf_acr_tokens
+    from .token import acr_token_show
+    token_client = cf_acr_tokens(cmd.cli_ctx)
+    # SCOPE MAP ID example
+    #/subscriptions/<1>/resourceGroups/<3>/providers/Microsoft.ContainerRegistry/registries/<7>/tokens/<9>'
+    token_info = token_id.lstrip('/').split('/')
+    if len(token_info) != 10:
```
northtyphoon, line 556: Is it a common pattern in cli to parse the resource?

src/azure-cli/azure/cli/command_modules/acr/_params.py
rosanch: I changed the sentence. please take a look

src/azure-cli/azure/cli/command_modules/acr/_format.py
```diff
@@ -93,6 +93,55 @@ def agentpool_output_format(result):
...
+
+def connected_registry_list_output_format(result):
+    family_tree = {}
+    for reg in result:
+        parent_id = _get_value(reg, 'parent', 'id')
+        parent_name = '' if parent_id.isspace() else parent_id.split('/connectedRegistries/')[1]
+        name = ' ' + _get_value(reg, 'name')
+        family_tree[_get_value(reg, 'id')] = {
+            "name": name,
+            "id": _get_value(reg, 'id'),
```
mnltejaswini, line 108: Is this resource ID?

src/azure-cli/azure/cli/command_modules/acr/_params.py
rosanch: From the CLI perspective null(None). Since the sync_window is not needed when the schedule is to sync all the time

src/azure-cli/azure/cli/command_modules/acr/_utils.py
rosanch: We don't do this anywhere else in ACR but I'm not sure if other teams do it or not.
Optionally I could do a request to get the token to verify it exists and get the ID from the response

src/azure-cli/azure/cli/command_modules/acr/_format.py
```diff
@@ -93,6 +93,55 @@ def agentpool_output_format(result):
...
+            "loginServer_host": _get_value(reg, 'loginServer', 'host'),
+            "parent_syncProperties_lastSyncTime": _get_value(reg, 'parent', 'syncProperties', 'lastSyncTime'),
+            "mode": _get_value(reg, 'mode'),
+            "childs": []
+        }
+
+    parents = []
+    for reg in result:
+        parent_id = _get_value(reg, 'parent', 'id')
+        if parent_id.isspace() or parent_id not in family_tree:
```
mnltejaswini, line 121: curious as to what this check as not in family_tree

src/azure-cli/azure/cli/command_modules/acr/_format.py
rosanch: correct

src/azure-cli/azure/cli/command_modules/acr/_format.py
```diff
@@ -93,6 +93,55 @@ def agentpool_output_format(result):
...
+            parents.append(_get_value(reg, 'id'))
```
mnltejaswini, line 122: will this add nulls if the parent is null?

src/azure-cli/azure/cli/command_modules/acr/_format.py
rosanch: so this is for keeping the tree format when the user uses the --parent option. Since the output was filter to only the children of the parent connected registry given, the resources that doesn't have a parent in the family_tree would be the level 0 resources of the tree.

src/azure-cli/azure/cli/command_modules/acr/_format.py
rosanch: _get_value returns an empty string when the parent is null. So it will never be null.

rosanch: But I understand why it might be confusing. I'll change the parents name to roots.

src/azure-cli/azure/cli/command_modules/acr/_format.py
```diff
@@ -93,6 +93,55 @@ def agentpool_output_format(result):
...
+
+def _recursive_format_list_acr_childs(family_tree, parent_id, level=""):
+    connected_registry = family_tree[parent_id]
+    childs = connected_registry['childs']
+    if connected_registry['parent_id'] in family_tree:
+        level += "\t"
+        connected_registry['name'] = level + connected_registry['name']
+    result = [connected_registry]
+    for child_id in childs:
+        result.extend(_recursive_format_list_acr_childs(family_tree, child_id, level))
```
mnltejaswini, line 141: Is this recursive function :)

src/azure-cli/azure/cli/command_modules/acr/_format.py
```diff
@@ -243,6 +292,33 @@ def _agentpool_format_group(item):
...
+    ])
+
+
+def _connected_registry_list_format_group(item):
+    return OrderedDict([
+        ('NAME', _get_value(item, 'name')),
+        ('MODE', _get_value(item, 'mode')),
+        ('CONNECTION STATE', _get_value(item, 'connectionState')),
+        ('PARENT', _get_value(item, 'parent_name')),
+        ('LOGIN SERVER', _get_value(item, 'loginServer_host')),
```
mnltejaswini, line 317: Just curious as to why in the line 303 it _get_value(item, 'loginServer', 'host') and why the format is different in this list _get_value(item, 'loginServer_host')

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -981,9 +987,10 @@
   - name: Create a token which has read permissions on hello-world repository.
     text: >
         az acr token create -n myToken -r MyRegistry --repository hello-world content/read metadata/read
-  - name: Create a token without credentials.
+  - name: Create a token without credentials and all gateway permissions.
     text: >
         az acr token create -n myToken -r MyRegistry --repository hello-world content/read --no-passwords
+            --gateway registry config/read config/write message/read message/write
```
mnltejaswini, line 993: --gateway connectedregistry

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -1204,6 +1211,115 @@
         az acr webhook update -n MyWebhook -r MyRegistry --status disabled
 """
 
+# region connected-registry
+helps['acr connected-registry'] = """
+type: group
+short-summary: Manage Edge/On-Prem Container Registries with Azure Container Registries.
```
mnltejaswini, line 1217: Manage connectedregistries

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -1204,6 +1211,115 @@
...
+"""
+
+helps['acr connected-registry create'] = """
+type: command
+short-summary: Create a connected registry for an Azure Container Registry.
+examples:
+  - name: Create a connected registry in registry mode with access to repos app/hello-world and service/mycomponent. It'll create a sync token and scope-map with the right repo permissions.
+    text: |
+        az acr connected-registry create --registry mycloudregistry --name myconnectedregistry
+            --repository "app/hello-world service/mycomponent"
```
mnltejaswini, line 1227: is " required for repository?

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -1204,6 +1211,115 @@
...
+  - name: Create a mirror connected registry with only read permissions and pass the sync token
```
mnltejaswini, line 1228: as a child to another connected registry

mnltejaswini: Not sure, if we want to provide this option now as it is blocked on RP?/cc @jaysterp

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -1204,6 +1211,115 @@
...
+            --name myconnectedregistry --yes
+"""
+
+helps['acr connected-registry install-info'] = """
+type: command
+short-summary: Retrives information required for to activate a connected registry.
+examples:
+  - name: Prints the values requiered to activate a connected registry in json format
+    text: |
+        az acr connected-registry install --registry mycloudregistry
```
mnltejaswini, line 1256: Is it install or install-info?

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -1204,6 +1211,115 @@
...
+    text: |
+        az acr connected-registry delete --registry mycloudregistry 
+            --name myconnectedregistry --yes
+"""
+
+helps['acr connected-registry install-info'] = """
+type: command
+short-summary: Retrives information required for to activate a connected registry.
+examples:
+  - name: Prints the values requiered to activate a connected registry in json format
```
mnltejaswini, line 1254: typo required

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -1204,6 +1211,115 @@
...
+type: command
+short-summary: Lists all the connected registries under the current parent registry.
+examples:
+  - name: Lists all the connected registries of 'mycloudregistry' in table format.
+    text: >
+        az acr connected-registry list --registry mycloudregistry --output table
+  - name: Lists only the inmediate children of 'mycloudregistry' in expanded form in a table.
+    text: >
+        az acr connected-registry list --registry mycloudregistry --no-children --output table
+  - name: Lists all the offspring of 'myconnectedregistry' in expanded form inside a table.
```
mnltejaswini, line 1284: nit: offsprings

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -1204,6 +1211,115 @@
...
+        az acr connected-registry list --registry mycloudregistry --parent myconnectedregistry --output table
+"""
+
+helps['acr connected-registry list-client-tokens'] = """
+type: command
+short-summary: Lists all the client tokens associated to a specific connected registries.
+examples:
+  - name: Lists all client tokens of 'mymirroracr'.
+    text: >
+        az acr connected-registry list-client-tokens --registry mycloudregistry --name mymirroracr
```
mnltejaswini, line 1295: Wondering, what is the use case for this? wouldn't show connected-registry provide the list?

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -1204,6 +1211,115 @@
...
+  - name: Show all the details of the 'mymirroracr' registry in table form.
+    text: |
+        az acr connected-registry show --registry mycloudregistry --name mymirroracr --output table
+"""
+
+helps['acr connected-registry update'] = """
+type: command
+short-summary: Update a connected registry for an Azure Container Registry.
+examples:
+  - name: Update the connected registry client Tokens.
```
mnltejaswini, line 1311: maybe little more specific update the connected registry to add/remove client tokens?

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -381,6 +390,35 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('key_encryption_key', help="Key vault key uri. To enable automated rotation, provide a version-less key uri. For manual rotation, provide a versioned key uri.")
         c.argument('identity', help="client id of managed identity, resource name or id of user assigned identity. Use '[system]' to refer to the system assigned identity")
 
+    with self.argument_context('acr connected-registry') as c:
+        c.argument('registry_name', options_list=['--registry', '-r'], help='The login server of the Cloud ACR registry. Must be the FQDN to support also Azure Stack.', required=True)
```
mnltejaswini, line 394: what is this Azure Stack reference here?

mnltejaswini: Can we use ACR in CLI or should it be azure container registry in docs?

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -381,6 +390,35 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
...
+        c.argument('log_level', options_list=['--log-level'], help='Sets the log level for logging on the instance. Accepted log levels are Debug, Information, Warning, Error, and None. ', required=False, default="Information")
+        c.argument('mode', options_list=['--mode', '-m'], help='Can be one of the two operating modes: registry or mirror(pull-only mode).', required=False, default="registry")
+        c.argument('client_token_list', options_list=['--client-tokens'], nargs='+', help='Specifies the client access to the repositories in the connected registry. It can be in the format [TOKEN_NAME01] [TOKEN_NAME02]...', required=False)
+        c.argument('sync_window', options_list=['--sync-window', '-w'], help='Required parameter if --sync-schedule is present. Used to determine the schedule duration. Uses ISO 8601 duration format.', required=False)
+        c.argument('sync_schedule', options_list=['--sync-schedule', '-s'], help='Optional parameter to define the sync schedule. Uses cron expression to determine the schedule. If not specified, the instance is considered always online and attempts to sync every minute.', required=False, default="* * * * *")
```
mnltejaswini, line 409: I don't think we sync every minute, do you want to just mention that it will always be connected

src/azure-cli/azure/cli/command_modules/acr/_utils.py
```diff
@@ -469,3 +453,125 @@ def prepare_source_location(cmd, source_location, client_registries, registry_na
...
+        for action in rule[1:]:
+            action = action.lower()
+            if action not in valid_actions:
+                raise CLIError('Invalid action "{}" provided. \nValid actions are {}.'.format(action, valid_actions))
+            actions.append('{}/{}/{}'.format('repositories', repository, action))
+
+    return actions
+
+
+def parse_actions_from_gateways(gateway_actions_list):
```
mnltejaswini, line 493: This code looks similar to parse_actions_from_repositories, wondering if this can be a shared function?

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
```diff
@@ -0,0 +1,494 @@
...
+    ErrorResponseException = cmd.get_models('ErrorResponseException')
+    parent = None
+    if parent_name:
+        try:
+            parent = acr_connected_registry_show(cmd, client, parent_name, registry_name, resource_group_name)
+        except ErrorResponseException as ex:
+            if ex.response.status_code == 404:
+                raise CLIError("The parent connected registry '{}' could not be found.".format(parent_name))
+            raise CLIError(ex)
+        if parent.mode.lower() != ConnectedRegistryModes.REGISTRY.value and parent.mode.lower() != mode.lower():
```
mnltejaswini, line 81: should the parent mode be same as the parent?

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
```diff
@@ -0,0 +1,494 @@
...
+                                  sync_token_name=None,
+                                  client_token_list=None,
+                                  resource_group_name=None,
+                                  mode=None,
+                                  parent_name=None,
+                                  sync_schedule=None,
+                                  sync_message_ttl=None,
+                                  sync_window=None,
+                                  log_level=None,
+                                  sync_audit_logs_enabled=True):
```
mnltejaswini, line 58: should it be defaulted to False? not sure how the default values are passed to this create?

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
```diff
@@ -0,0 +1,494 @@
...
+        cmd, client, connected_registry_name, registry_name, resource_group_name)
+
+    # Add or remove from the current client token id list
+    if add_client_token_list is not None:
+        for i, client_token_name in enumerate(add_client_token_list):
+            add_client_token_list[i] = build_token_id(
+                subscription_id, resource_group_name, registry_name, client_token_name)
+        add_client_token_set = set(add_client_token_list)
+    else:
+        add_client_token_set = set()
```
mnltejaswini, line 153: will set() be null or empty array?

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
```diff
@@ -0,0 +1,494 @@
...
+        for i, client_token_name in enumerate(remove_client_token_list):
+            remove_client_token_list[i] = build_token_id(
+                subscription_id, resource_group_name, registry_name, client_token_name)
+        remove_client_token_set = set(remove_client_token_list)
+    else:
+        remove_client_token_set = set()
+
+    duplicate_client_token = set.intersection(add_client_token_set, remove_client_token_set)
+    if duplicate_client_token:
+        errors = sorted(map(lambda action: action[action.find('/') + 1:], duplicate_client_token))
```
mnltejaswini, line 164: what is this logic?

src/azure-cli/azure/cli/command_modules/acr/_client_factory.py
```diff
@@ -48,16 +49,19 @@ def cf_acr_runs(cli_ctx, *_):
...
 
 def cf_acr_token_credentials(cli_ctx, *_):
-    return get_acr_service_client(cli_ctx, VERSION_2019_05_01_PREVIEW).registries
+    return get_acr_service_client(cli_ctx, VERSION_2020_11_01_PREVIEW).registries
 
 
 def cf_acr_agentpool(cli_ctx, *_):
     return get_acr_service_client(cli_ctx, VERSION_2019_06_01_PREVIEW).agent_pools
+
+def cf_acr_connected_registry(cli_ctx, *_):
```
jaysterp, line 66: @rosanch for consistency can we rename this `cf_acr_connected_registries`?

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
```diff
@@ -0,0 +1,494 @@
...
+        cmd, registry_name, resource_group_name)
+
+    user_confirmation("Are you sure you want to delete the Connected Registry '{}' from '{}'?".format(
+        connected_registry_name, registry_name), yes)
+    try:
+        result = client.delete(resource_group_name, registry_name, connected_registry_name)
+        msg = "Connected registry successfully deleted. Please cleanup your sync tokens and scope maps. " + \
+              "Run the following commands for cleanup: \n\t" + \
+              "az acr token delete -n <sync token> -r {}\n\t".format(registry_name) + \
+              "az acr scope-map delete -n <sync scope map> -r {}\n".format(registry_name) + \
```
mnltejaswini, line 217: can we give their sync token name here?

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
```diff
@@ -0,0 +1,494 @@
...
+
+def acr_connected_registry_list(cmd,
+                                client,
+                                registry_name,
+                                parent=None,
+                                no_children=False,
+                                resource_group_name=None):
+    _, resource_group_name = validate_managed_registry(
+        cmd, registry_name, resource_group_name)
+    connected_registry_list = list(client.list(resource_group_name, registry_name))
```
mnltejaswini, line 253: will this go over all the pages if the list has continuation token?

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
```diff
@@ -0,0 +1,494 @@
...
+        }
+    else:
+        sync_username = "<sync token username>"
+        sync_password = "<sync token password>"
+
+    return {
+        "ACR_REGISTRY_NAME": connected_registry_name,
+        "ACR_REGISTRY_LOGIN_SERVER": connected_registry_localhost,
+        "ACR_SYNC_TOKEN_NAME": sync_token_name,
+        "ACR_SYNC_TOKEN_USERNAME": sync_username,
```
mnltejaswini, line 334: if we give them password, do we need to indicate that it is one time password and should copy? Not sure how the token credentials CLI does this

src/azure-cli/azure/cli/command_modules/acr/_format.py
```diff
@@ -243,6 +292,33 @@ def _agentpool_format_group(item):
...
+        ('LAST SYNC (UTC)', _get_value(item, 'parent_syncProperties_lastSyncTime'))
```
jaysterp, line 318: @rosanch Why do we not also include `LAST ACTIVITY (UTC)` here as well?

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
```diff
@@ -0,0 +1,494 @@
...
+    if not all(action in parent_sync_actions  for action in gateway_actions):
+        raise CLIError(
+            "Parent '{}' sync token '{}'. ".format(parent.name, sync_token.name) +
+            "Ensure the sync token has the following permissions: \n\t{}\n".format(gateway_actions) +
+            "Run the command: az acr scope-map update -n {} -r {} --add-gateway {}".format(
+                connected_registry_name, registry_name, gateway_list
+            ))
+
+
+#TODO delete if not used
```
mnltejaswini, line 434: Is this still needed?

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -528,6 +528,12 @@
   - name: Create a scope map that allows content/write and metadata/read actions for `hello-world` repository, and content/read action for `hello-world-again`.
     text: >
         az acr scope-map create -n MyScopeMap -r MyRegistry --repository hello-world content/write metadata/read --repository hello-world-again content/read --description "Sample scope map."
+  - name: Create a scope map that allows all repository actions for `test`, and all gateway actions for `connectedRegistry`.
```
jaysterp, line 531: @rosanch Do we want to clarify here that these actions are for `connectedRegistry` in `Registry` mode?

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -981,9 +987,10 @@
   - name: Create a token which has read permissions on hello-world repository.
     text: >
         az acr token create -n myToken -r MyRegistry --repository hello-world content/read metadata/read
-  - name: Create a token without credentials.
+  - name: Create a token without credentials and all gateway permissions.
```
jaysterp, line 990: Nit: "Create a token without credentials and with all gateway permissions"

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -1204,6 +1211,115 @@
...
+examples:
+  - name: Delete a mirror connected registry 'myconnectedregistry' from parent registry 'mycloudregistry' and skip verification.
+    text: |
+        az acr connected-registry delete --registry mycloudregistry 
+            --name myconnectedregistry --yes
+"""
+
+helps['acr connected-registry install-info'] = """
+type: command
+short-summary: Retrives information required for to activate a connected registry.
```
jaysterp, line 1252: @rosanch typo: Retrives -> Retrieves

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -1204,6 +1211,115 @@
...
+"""
+
+helps['acr connected-registry deactivate'] = """
+type: command
+short-summary: Deactivate a connected registry from Azure Container Registry.
+examples:
+  - name: Deactivate a connected registry 'myconnectedregistry' and skip verification.
+    text: |
+        az acr connected-registry deactivate --registry mycloudregistry 
+            --name myconnectedregistry --yes
```
jaysterp, line 1271: @rosanch Why do we have a `--yes` flag for this command?

jaysterp: I see, it is for confirmation.

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -1204,6 +1211,115 @@
...
+examples:
+  - name: Update the connected registry client Tokens.
+    text: |
+        az acr connected-registry update --registry mycloudregistry --name myconnectedregistry
+            --remove-client-tokens myTokenName1 --add-client-tokens myTokenName2 myTokenName3
+
+  - name: Update the sync and window time, and disable audit logs of a connected registry.
+    text: |
+        az acr connected-registry create --registry mycloudregistry --name mymirroracr
+            --sync-schedule "0 12 * * *" --sync-window PT4H --sync-audit-logs-enabled false
```
jaysterp, line 1319: @rosanch the `create` example above also used this schedule and window. Can we change them for the `update` example?

src/azure-cli/azure/cli/command_modules/acr/_params.py
jaysterp: @rosanch @northtyphoon @mnltejaswini we do not support `_` or `-` in connected registry resource names. Only alphanumeric chars demonstrated by the pattern `^[a-zA-Z0-9]*$`

src/azure-cli/azure/cli/command_modules/acr/_params.py
jaysterp, line 399: @rosanch "Specifies the sync token used to synchronize the connected registry with its parent."

yonzhan:
ACR

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
```diff
@@ -0,0 +1,494 @@
...
+        raise CLIError("usage error: you need to provide either --sync-token-name or --repository, but not both.")
+    registry, resource_group_name = get_registry_by_name(cmd.cli_ctx, registry_name, resource_group_name)
+    subscription_id = get_subscription_id(cmd.cli_ctx)
+
+    if not registry.data_endpoint_enabled:
+        logger.warning("Registry '%s' dedicated data endpoint is disabled. Enabling...", registry_name)
+        from .custom import acr_update_custom, acr_update_get, acr_update_set
+        acr_update_set(cmd, cf_acr_registries(cmd.cli_ctx), registry_name, resource_group_name,
+                       parameters=acr_update_custom(cmd, acr_update_get(cmd), data_endpoint_enabled=True))
+        logger.info("Registry '%s' dedicated data endpoint successfully enabled.", registry_name)
```
northtyphoon, line 70: >successfully [](start = 59, length = 12)

has been successfully enabled.

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
```diff
@@ -0,0 +1,494 @@
...
+        try:
+            parent = acr_connected_registry_show(cmd, client, parent_name, registry_name, resource_group_name)
+        except ErrorResponseException as ex:
+            if ex.response.status_code == 404:
+                raise CLIError("The parent connected registry '{}' could not be found.".format(parent_name))
+            raise CLIError(ex)
+        if parent.mode.lower() != ConnectedRegistryModes.REGISTRY.value and parent.mode.lower() != mode.lower():
+            raise CLIError("Can't create the registry '{}' with mode '{}'".format(connected_registry_name, mode) +
+                           "when the connected registry parent '{}' mode is '{}'.".format(parent_name, parent.mode) +
+                           "For more information on connected registries, please visit https://aka.ms/acr/onprem.")
```
northtyphoon, line 84: >onprem [](start = 106, length = 6)

@jaysterp what is the aka.ms link we use in rp?

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
```diff
@@ -0,0 +1,494 @@
...
+                                  client,
+                                  connected_registry_name,
+                                  registry_name,
+                                  yes=False,
+                                  resource_group_name=None):
+
+    _, resource_group_name = validate_managed_registry(
+        cmd, registry_name, resource_group_name)
+
+    user_confirmation("Are you sure you want to delete the Connected Registry '{}' from '{}'?".format(
```
northtyphoon, line 210: >Connected Registry [](start = 59, length = 18)

keep the message format consistent in the code `Connected Registry` => `connected registry`

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
```diff
@@ -0,0 +1,494 @@
...
+                                      client,
+                                      connected_registry_name,
+                                      registry_name,
+                                      yes=False,
+                                      resource_group_name=None):
+    _, resource_group_name = validate_managed_registry(
+        cmd, registry_name, resource_group_name)
+    subscription_id = get_subscription_id(cmd.cli_ctx)
+
+    user_confirmation("Are you sure you want to deactivate the Connected Registry '{}' from '{}'?".format(
```
northtyphoon, line 237: > from '{} [](start = 86, length = 9)

from => in ?

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
```diff
@@ -0,0 +1,494 @@
...
+                                no_children=False,
+                                resource_group_name=None):
+    _, resource_group_name = validate_managed_registry(
+        cmd, registry_name, resource_group_name)
+    connected_registry_list = list(client.list(resource_group_name, registry_name))
+    result = []
+    if no_children:
+        if parent:
+            result = [registry for registry in connected_registry_list \
+                if registry.parent.id is not None and registry.parent.id.endswith(parent)]
```
northtyphoon, line 258: >endswith(parent) [](start = 73, length = 16)

can it handle the following cases?

parent
myparent

endswith("/"+parent)

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
```diff
@@ -0,0 +1,494 @@
...
+        parent_registry_endpoint = registry.login_server
+
+    if fresh_install:
+        from ._client_factory import cf_acr_token_credentials
+        from .token import acr_token_credential_generate
+        cred_client = cf_acr_token_credentials(cmd.cli_ctx)
+        poller = acr_token_credential_generate(
+            cmd, cred_client, registry_name, sync_token_name,
+            password1=True, password2=True, resource_group_name=resource_group_name)
+        #   expiration_in_days=expiration_in_days, expiration=expiration)
```
northtyphoon, line 319: remove?

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
```diff
@@ -0,0 +1,494 @@
...
+            cmd, client, parent, registry_name, resource_group_name)
+        parent_registry_endpoint = parent.login_server.host
+    else:
+        parent_registry_endpoint = registry.login_server
+
+    if fresh_install:
+        from ._client_factory import cf_acr_token_credentials
+        from .token import acr_token_credential_generate
+        cred_client = cf_acr_token_credentials(cmd.cli_ctx)
+        poller = acr_token_credential_generate(
```
northtyphoon, line 316: >acr_token_credential_generate [](start = 17, length = 29)

regenerate the password will invalidate existing password and could impact the client using it, do you want to at least print out a warning message if previous password exists?

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
```diff
@@ -0,0 +1,494 @@
...
+            "password1": credentials.passwords[0].value,
+            "password2": credentials.passwords[1].value
+        }
+    else:
+        sync_username = "<sync token username>"
+        sync_password = "<sync token password>"
+
+    return {
+        "ACR_REGISTRY_NAME": connected_registry_name,
+        "ACR_REGISTRY_LOGIN_SERVER": connected_registry_localhost,
```
northtyphoon, line 332: >connected_registry_localhost [](start = 37, length = 28)

may rename to connected_registry_login_server

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
```diff
@@ -0,0 +1,494 @@
...
+                                              client,
+                                              connected_registry_name,
+                                              registry_name,
+                                              resource_group_name=None):
+    _, resource_group_name = validate_managed_registry(
+        cmd, registry_name, resource_group_name)
+    current_connected_registry = acr_connected_registry_show(
+        cmd, client, connected_registry_name, registry_name, resource_group_name)
+    if current_connected_registry.client_token_ids is None:
+        raise CLIError("No Client tokens found: You can update your connected registry to add client tokens.")
```
northtyphoon, line 352: >Client [](start = 27, length = 6)

lowercase client

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
```diff
@@ -0,0 +1,494 @@
...
+
+    mode = mode.lower()
+    if not any(option for option in ConnectedRegistryModes if option.value == mode):
+        raise CLIError("usage error: --mode supports only 'registry' and 'mirror' values.")
+    repository_actions_list = [[repo] + REPO_SCOPES_BY_MODE[mode] for repo in repositories]
+    gateway_actions_list = [[connected_registry_name] + DEFAULT_GATEWAY_SCOPE]
+    try:
+        message = "Created by connected registry sync token: {}"
+        sync_scope_map_name = SYNC_SCOPE_MAP_NAME.format(connected_registry_name)
+        logger.warning("If sync scope map '%s' already exist properties will be overrided", sync_scope_map_name)
```
northtyphoon, line 377: >exist [](start = 55, length = 5)

If sync scope map '%s' already exists, its actions will be overridden

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
```diff
@@ -0,0 +1,494 @@
...
+        sync_scope_map = create_default_scope_map(cmd, resource_group_name, registry_name, sync_scope_map_name,
+                                                  repository_actions_list, gateway_actions_list,
+                                                  scope_map_description=message.format(connected_registry_name),
+                                                  force=True)
+
+        sync_token_name = SYNC_TOKEN_NAME.format(connected_registry_name)
+        logger.warning("If sync token '%s' already exist properties will be overrided", sync_token_name)
```
northtyphoon, line 384: >If sync token '%s' already exist properties will be overrided [](start = 24, length = 61)

If sync token '%s' already exists, it properties will be overwritten

northtyphoon (changes requested):
:clock1:

yonzhan:
ACR

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
rosanch: when the parent is mirror the child can't be registry. Since the parent isn't going to have the permissions the child requires.

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
rosanch: The default value in the CLI spec is that the audit_log_status Should be Enabled.

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
northtyphoon: should we use logger.warning instead of logger.info?

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
rosanch: good catch! this should be rfind. The idea is to display only the token name. No the whole ID and the +1 is to start one character after the '/' match

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
rosanch: I would need to get the connected registry to get the token and then get the token to get the scope map ID and trim it to only pass the scope map name. But yes, it is possible.

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
rosanch: I'm not sure what do you mean. But this method goes through all the connected registries. That is the only way I can build the tree to display it.

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
rosanch: good catch! yes they do. I added a warning

src/azure-cli/azure/cli/command_modules/acr/_format.py
rosanch: We had a meeting to discuss what should be displayed. But I'll bring this up  in next meeting.

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
rosanch: I'll delete all those methods before merging. But there has been a lot of back and forward on the CLI behavior. So I left this methods just in case. We needed them again. But that doesn't seems to be the case.

src/azure-cli/azure/cli/command_modules/acr/_help.py
rosanch: correct. --yes flag is to skip the delete verification.

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
rosanch: No, but is that possible? this is reading from RP response. It's not checking users input.

rosanch: I could add cover that scenario if it's possible.

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
rosanch: I'm not sure if there's a way to check. But every time the password is created we print a warning saying the client should store the password. This follows the same behavior found in the token command

src/azure-cli/azure/cli/command_modules/acr/_utils.py
rosanch: we could but, it was just easier like this to handle different error messages. It didn't seemed worth the trouble. We can save a ifs as well.

src/azure-cli/azure/cli/command_modules/acr/_help.py
rosanch: I thought about it. But I wasn't sure exactly what is the need for Gateway tokens and I'm not sure if there is another application the users might give to this commands.

rosanch:
Hello @fengzhou-msft  @yungezz @yonzhan 
I'm wondering if you could review my pr? We'd like this to be included in the December release.
I sync with @adewaleo to merge this pr after his PR #15975 . Since both prs are affected by the new ACR API. But his changes have impact on the existing testing and I'm only adding new features. 
I'll merge his changes and push them to pass the automated test once his pr gets merged

yungezz:
> Hello @fengzhou-msft @yungezz @yonzhan
> I'm wondering if you could review my pr? We'd like this to be included in the December release.
> I sync with @adewaleo to merge this pr after his PR #15975 . Since both prs are affected by the new ACR API. But his changes have impact on the existing testing and I'm only adding new features.
> I'll merge his changes and push them to pass the automated test once his pr gets merged

hi @rosanch sure.

src/azure-cli-core/azure/cli/core/profiles/_shared.py
```diff
@@ -162,7 +162,7 @@ def default_api_version(self):
             'role_definitions': '2018-01-01-preview',
             'provider_operations_metadata': '2018-01-01-preview'
         }),
-        ResourceType.MGMT_CONTAINERREGISTRY: '2019-12-01-preview',
+        ResourceType.MGMT_CONTAINERREGISTRY: '2020-11-01-preview',
```
yungezz, line 165: pls run full live test to avoid regression

src/azure-cli/azure/cli/command_modules/acr/_format.py
yungezz: same question, properties in list of connected registry is different with above connected registry

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -381,6 +390,35 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('key_encryption_key', help="Key vault key uri. To enable automated rotation, provide a version-less key uri. For manual rotation, provide a versioned key uri.")
         c.argument('identity', help="client id of managed identity, resource name or id of user assigned identity. Use '[system]' to refer to the system assigned identity")
 
+    with self.argument_context('acr connected-registry') as c:
+        c.argument('registry_name', options_list=['--registry', '-r'], help='The login server of the Cloud ACR registry. Must be the FQDN to support also Azure Stack.', required=True)
+        c.argument('connected_registry_name', options_list=['--name', '-n'], help='Name for the connected registry. Name must be between 5 to 40 character long, start with a letter and contain only alphanumeric characters (including ‘_’ or ‘-’). Name must be unique under the Cloud ACR hierarchy.', required=True)
+        c.argument('resource_group_name', options_list=['--resource-group', '-g'], help='The name of the resource group where the connected registry registry will be created. Must be the same as the resource group for the --registry.', required=False)
```
yungezz, line 396: I don't think this is need. This is global parm

yungezz:
hi @rosanch could you pls fix lint/style issue, and look at comments?

src/azure-cli-core/azure/cli/core/profiles/_shared.py
rosanch: done. The change was already done by tosin. He ran all the test and created the new records.
Test pass 100%

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
```diff
@@ -0,0 +1,445 @@
...
+                                  log_level=None,
+                                  sync_audit_logs_enabled=False):
+
+    if bool(sync_token_name) == bool(repositories):
+        raise CLIError("usage error: you need to provide either --sync-token-name or --repository, but not both.")
+    registry, resource_group_name = get_registry_by_name(cmd.cli_ctx, registry_name, resource_group_name)
+    subscription_id = get_subscription_id(cmd.cli_ctx)
+
+    if not registry.data_endpoint_enabled:
+        raise CLIError("Can't create the connected registry '{}' ".format(connected_registry_name,) +
```
northtyphoon, line 64: >format(connected_registry_name,) [](start = 67, length = 32)

there is an extra comma after connected_registry_name

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
```diff
@@ -0,0 +1,445 @@
...
+                                      client,
+                                      connected_registry_name,
+                                      registry_name,
+                                      yes=False,
+                                      resource_group_name=None):
+    _, resource_group_name = validate_managed_registry(
+        cmd, registry_name, resource_group_name)
+    subscription_id = get_subscription_id(cmd.cli_ctx)
+
+    user_confirmation("Are you sure you want to deactivate the Connected Registry '{}' in '{}'?".format(
```
northtyphoon, line 249: >Connected Registry [](start = 63, length = 18)

lower case: connected registry

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -383,6 +392,34 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('key_encryption_key', help="Key vault key uri. To enable automated rotation, provide a version-less key uri. For manual rotation, provide a versioned key uri.")
         c.argument('identity', help="client id of managed identity, resource name or id of user assigned identity. Use '[system]' to refer to the system assigned identity")
 
+    with self.argument_context('acr connected-registry') as c:
+        c.argument('registry_name', options_list=['--registry', '-r'], help='The login server of the Cloud ACR registry. Must be the FQDN to support also Azure Stack.', required=True)
+        c.argument('connected_registry_name', options_list=['--name', '-n'], help='Name for the connected registry. Name must be between 5 to 40 character long, start with a letter and contain only alphanumeric characters (including ‘_’ or ‘-’). Name must be unique under the Cloud ACR hierarchy.', required=True)
+        c.argument('parent', options_list=['--parent', '-p'], help='The name of the parent Connected Registry.', required=False)
```
northtyphoon, line 398: >Connected Registry [](start = 91, length = 18)

parent registry or parent connected registry

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
northtyphoon: I mean if there are two connected registry resources.

connectedRegistries/parent
connectedRegistries/myparent

customer run `az acr connected-registry list --parent parent`, the above code  will return the result from `connectedRegistries/myparent` as well.



---
In reply to: [545675378](https://github.com/Azure/azure-cli/pull/16238#discussion_r545675378) [](ancestors = 545675378)

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
```diff
@@ -0,0 +1,445 @@
...
+        parent = acr_connected_registry_show(
+            cmd, client, parent, registry_name, resource_group_name)
+        parent_registry_endpoint = parent.login_server.host
+    else:
+        parent_registry_endpoint = registry.login_server
+
+    connected_registry_login_server = connected_registry.login_server.host
+    if not connected_registry_login_server:
+        logger.warning("No login server value was provided nor retrieved from the connected registry. "
+                       "Use the argument '--login-server' to pass a value. For more information, "
```
northtyphoon, line 414: the error message need to be updated, there is no --login-server option

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
```diff
@@ -0,0 +1,440 @@
...
+            password1=True, password2=True, resource_group_name=resource_group_name)
+        credentials = LongRunningOperation(cmd.cli_ctx)(poller)
+        sync_username = credentials.username
+        sync_password = {
+            "password1": credentials.passwords[0].value,
+            "password2": credentials.passwords[1].value
+        }
+        logger.warning('Please store your generated credentials safely.')
+    else:
+        sync_username = "<sync token username>"
```
northtyphoon, line 433: >"<sync token username>" [](start = 24, length = 23)

if customer run `info` command, can CLI still provide sync_username?

src/azure-cli/azure/cli/command_modules/acr/connected_registry.py
```diff
@@ -0,0 +1,445 @@
...
+    sync_token_name = connected_registry.parent.sync_properties.token_id.split('/tokens/')[1]
+    if parent_id:
+        parent = parent_id.split('/connectedRegistries/')[1]
+        parent = acr_connected_registry_show(
+            cmd, client, parent, registry_name, resource_group_name)
+        parent_registry_endpoint = parent.login_server.host
+    else:
+        parent_registry_endpoint = registry.login_server
+
+    connected_registry_login_server = connected_registry.login_server.host
```
northtyphoon, line 411: I remember we agree to always provide a place holder string with sample and aks.ms link for connected_registry_login_server install parameter

rosanch: my bad I thought it was only when the value was null.

yungezz (approved):

yungezz:
HI @northtyphoon could you pls approve the PR if all comments addressed? We'll cut build tomorrow morning.

northtyphoon (commented):
LGTM

PR #16241, opened 2020-12-11
[Synapse] Add new cmdlets for synapse sql aad admin and sql audit policy and add update cmdlet for synapse workspace firewall rule

aim-for-better:
**Description**

1. Add new cmdlets az synapse sql ad-admin show, create, update, delete
2. Add new cmdlet az synapse workspace firewall-rule update
3. Add new cmdlets az synapse sql audit-policy show, update

**Testing Guide**

**History Notes**

[Synapse] 1. Add new cmdlets az synapse sql ad-admin show, create, update, delete
2. Add new cmdlet az synapse workspace firewall-rule update
3. Add new cmdlets az synapse sql audit-policy show, update

---


aim-for-better:
Is the linter rule "MEDIUM severity: no_required_location_param" new added?

src/azure-cli/azure/cli/command_modules/synapse/_params.py
```diff
@@ -251,29 +251,44 @@ def load_arguments(self, _):
...
+                       arg_group=policy_arg_group,
+                       help='List of actions and action groups to audit.',
+                       nargs='+')
+            c.argument('retention_days',
+                       arg_group=policy_arg_group,
+                       help='The number of days to retain audit logs.')
+
+    with self.argument_context('synapse sql audit-policy update') as c:
+        c.argument('queue_delay_milliseconds', type=int,
+                   options_list=['--queue-delay-time', '--queue-delay-milliseconds'],
```
sunsw1994, line 281: >'--queue-delay-milliseconds' [](start = 55, length = 28)

longer than threshold 22?

aim-for-better: > > '--queue-delay-milliseconds' [](start = 55, length = 28)
> 
> longer than threshold 22?

I have added another option `--queue-delay-time`.  Now the CI failed in check CLI style but I can not find details. It seems it is not related with this.
image

aim-for-better: fixed the ci error.

aim-for-better:
The only one CI error is: [error]The job running on agent Azure Pipelines 102 ran longer than the maximum time of 10 minutes. For more information, see https://go.microsoft.com/fwlink/?linkid=2077134
It seems that it is not related with our PR.

yonzhan:
Synapse

aim-for-better:
Hi @jsntcy @Juliehzl  Could you please help review this PR? Thanks~

src/azure-cli/azure/cli/command_modules/synapse/_client_factory.py
```diff
@@ -3,6 +3,14 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 
+
+def cf_resource_groups(cli_ctx, subscription_id=None):
+    from azure.cli.core.commands.client_factory import get_mgmt_service_client
+    from azure.cli.core.profiles import ResourceType
+    return get_mgmt_service_client(cli_ctx, ResourceType.MGMT_RESOURCE_RESOURCES,
```
Juliehzl, line 10: could you import the client factory for resource module?
https://github.com/Azure/azure-cli/blob/a3b384eba388bc31003967769ede09a91c162571/src/azure-cli/azure/cli/command_modules/resource/_client_factory.py#L67-L68

src/azure-cli/azure/cli/command_modules/synapse/_params.py
```diff
@@ -251,29 +251,44 @@ def load_arguments(self, _):
...
-                   arg_group=policy_arg_group,
-                   help='List of actions and action groups to audit.',
-                   nargs='+')
-        c.argument('retention_days',
-                   arg_group=policy_arg_group,
-                   help='The number of days to retain audit logs.')
+    for scope in ['synapse sql pool audit-policy', 'synapse sql audit-policy']:
+        with self.argument_context(scope + ' update') as c:
+            _configure_security_or_audit_policy_storage_params(c)
+            c.argument('storage_account_subscription_id', arg_group=storage_arg_group,
```
Juliehzl, line 257: actually I am curious about previous design. If you want to  use storage account in another subscription, why not declare a storage account parameter which would accept both storage account resource id and name? 
In this way, cx could use resource id to access storage account in another subscription and use name for storage account in the same subscription.

src/azure-cli/azure/cli/command_modules/synapse/_params.py
```diff
@@ -251,29 +251,44 @@ def load_arguments(self, _):
...
+                       help='Whether enabling azure monitor target or not.',
+                       arg_type=get_three_state_flag())
+            c.argument('state',
+                       arg_group=policy_arg_group,
+                       help='Auditing policy state',
+                       arg_type=get_enum_type(BlobAuditingPolicyState))
+            c.argument('audit_actions_and_groups',
+                       options_list=['--actions'],
+                       arg_group=policy_arg_group,
+                       help='List of actions and action groups to audit.',
```
Juliehzl, line 273: which kind of actions could be added?

src/azure-cli/azure/cli/command_modules/synapse/_params.py
```diff
@@ -251,29 +251,44 @@ def load_arguments(self, _):
...
+                       nargs='+')
+            c.argument('retention_days',
+                       arg_group=policy_arg_group,
+                       help='The number of days to retain audit logs.')
```
Juliehzl, line 277: type int?

src/azure-cli/azure/cli/command_modules/synapse/_params.py
```diff
@@ -251,29 +251,44 @@ def load_arguments(self, _):
...
+
+    with self.argument_context('synapse sql audit-policy update') as c:
+        c.argument('queue_delay_milliseconds', type=int,
+                   options_list=['--queue-delay-time', '--queue-delay-milliseconds'],
+                   help='The amount of time in milliseconds that can elapse before audit actions are forced to be processed')
+
+    with self.argument_context('synapse sql ad-admin') as c:
+        c.argument('workspace_name', help='The workspace name.')
```
Juliehzl, line 285: if workspace is the target to operate, recommend use `--name` or `-n`.

src/azure-cli/azure/cli/command_modules/synapse/_params.py
```diff
@@ -251,29 +251,44 @@ def load_arguments(self, _):
...
+    for scope in ['create', 'update']:
+        with self.argument_context('synapse sql ad-admin ' + scope) as c:
+            c.argument('login_name', options_list=['--display-name', '-u'],
+                       help='Display name of the Azure AD administrator user or group.')
+            c.argument('object_id', options_list=['--object-id', '-i'],
```
Juliehzl, line 290: question: would these two parameters are both required for the commands?
I am wondering why object id could not determine the admin and we need name here.

src/azure-cli/azure/cli/command_modules/synapse/_params.py
```diff
@@ -282,14 +297,15 @@ def load_arguments(self, _):
...
 
-    with self.argument_context('synapse workspace firewall-rule create') as c:
-        c.argument('start_ip_address', help='The start IP address of the firewall rule. Must be IPv4 format.')
-        c.argument('end_ip_address', help='The end IP address of the firewall rule. Must be IPv4 format. '
-                                          'Must be greater than or equal to startIpAddress.')
+    for scope in ['create', 'update']:
+        with self.argument_context('synapse workspace firewall-rule ' + scope) as c:
+            c.argument('start_ip_address', help='The start IP address of the firewall rule. Must be IPv4 format.')
+            c.argument('end_ip_address', help='The end IP address of the firewall rule. Must be IPv4 format. '
+                                              'Must be greater than or equal to startIpAddress.')
```
Juliehzl, line 308: recommend to add validator for these parameter to make sure the format is right and make sense.

src/azure-cli/azure/cli/command_modules/synapse/commands.py
```diff
@@ -200,13 +211,34 @@ def get_custom_sdk(custom_module, client_factory):
...
+    # Management Plane Commands --Sql Ad-Admin
+    with self.command_group('synapse sql ad-admin', command_type=synapse_workspace_aad_admin_sdk,
+                            custom_command_type=get_custom_sdk('workspacesqlaadadmin',
+                                                               cf_synapse_client_workspace_aad_admins_factory),
+                            client_factory=cf_synapse_client_workspace_aad_admins_factory) as g:
+        g.show_command('show', 'get')
+        g.custom_command('create', 'create_workspace_sql_aad_admin', supports_no_wait=True)
+        g.generic_update_command('update', custom_func_name='update_workspace_sql_aad_admin',
+                                 setter_arg_name='aad_admin_info', supports_no_wait=True)
+        g.command('delete', 'delete')
```
Juliehzl, line 223: please add wait command here

src/azure-cli/azure/cli/command_modules/synapse/operations/sqlpoolblobauditingpolicy.py
```diff
@@ -58,7 +95,16 @@ def sqlpool_blob_auditing_policy_update(
...
+
+def _validate_audit_policy_arguments(state=None, storage_account=None, storage_endpoint=None,
+                                     storage_account_access_key=None, retention_days=None):
+    blob_storage_arguments_provided = any(
+        [storage_account, storage_endpoint, storage_account_access_key, retention_days])
+    if not state and not blob_storage_arguments_provided:
+        raise CLIError('Either state or blob storage arguments are missing')
+
+    if retention_days is not None and (not retention_days.isdigit() or int(retention_days) <= 0):
+        raise CLIError('retention-days must be a positive number greater than zero')
```
Juliehzl, line 107: such validation would be recommended to add as validator and other commands/parameters could utilize the function.
see https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/resource/commands.py#L240

src/azure-cli/azure/cli/command_modules/synapse/operations/workspace.py
```diff
@@ -15,12 +16,14 @@ def list_workspaces(cmd, client, resource_group_name=None):
...
 
 def create_workspace(cmd, client, resource_group_name, workspace_name, storage_account, file_system,
-                     sql_admin_login_user, sql_admin_login_password, location, enable_managed_virtual_network=None,
+                     sql_admin_login_user, sql_admin_login_password, location=None, enable_managed_virtual_network=None,
                      tags=None, no_wait=False):
     identity_type = "SystemAssigned"
     identity = ManagedIdentity(type=identity_type)
     account_url = "https://{}.dfs.{}".format(storage_account, cmd.cli_ctx.cloud.suffixes.storage_endpoint)
     default_data_lake_storage = DataLakeStorageAccountDetails(account_url=account_url, filesystem=file_system)
+    location = location or get_resource_group_location(cmd.cli_ctx, resource_group_name)
```
Juliehzl, line 25: recommend such kind of usage to get location from resource group
https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/storage/_params.py#L212

src/azure-cli/azure/cli/command_modules/synapse/operations/workspacesqlaadadmin.py
```diff
@@ -0,0 +1,36 @@
...
+from ..constant import AdministratorType
+
+
+# Synapse SQL ad-admin
+def create_workspace_sql_aad_admin(cmd, client, resource_group_name, workspace_name, login_name, object_id,
+                                   no_wait=False):
+    """
+    Set a Workspace SQL AD admin.
+    """
+    workspace_client = cf_synapse_client_workspace_factory(cmd.cli_ctx)
```
Juliehzl, line 20: why not use the client in parameter list?

src/azure-cli/azure/cli/command_modules/synapse/_client_factory.py
aim-for-better: > could you import the client factory for resource module?
> https://github.com/Azure/azure-cli/blob/a3b384eba388bc31003967769ede09a91c162571/src/azure-cli/azure/cli/command_modules/resource/_client_factory.py#L67-L68

Hi @Juliehzl , I am afraid that if we import the `cf_resource_groups` from resource module there will be potential issue if there is some changes in resource module. Current status is using the client factory from azure.cli.core. core.

src/azure-cli/azure/cli/command_modules/synapse/_params.py
aim-for-better: > actually I am curious about previous design. If you want to use storage account in another subscription, why not declare a storage account parameter which would accept both storage account resource id and name?
> In this way, cx could use resource id to access storage account in another subscription and use name for storage account in the same subscription.

I agree with you. In fact storage account name is global unique. But in the api definition (please see this [line](https://github.com/Azure/azure-rest-api-specs/blob/6bf460bee8e1d3a837fd0e37b1864cba4725c124/specification/synapse/resource-manager/Microsoft.Synapse/preview/2019-06-01-preview/sqlPool.json#L5481)), there is a property requires subscription id. Per my previous experience, I usually define a related parameter for property in the api. As you see the update cmdlet can update a property.

src/azure-cli/azure/cli/command_modules/synapse/_params.py
aim-for-better: > which kind of actions could be added?
The action seems like a statement: `FAILED_DATABASE_AUTHENTICATION_GROUP 'UPDATE on database::mydb by public'`
The usage is like:
```
az synapse sql pool audit-policy update --name sqlpool --workspace-name testsynapseworkspace --resource-group rg \\
        --actions FAILED_DATABASE_AUTHENTICATION_GROUP 'UPDATE on database::mydb by public'

```

src/azure-cli/azure/cli/command_modules/synapse/_params.py
aim-for-better: > type int?

Fixed.

src/azure-cli/azure/cli/command_modules/synapse/_params.py
aim-for-better: > if workspace is the target to operate, recommend use `--name` or `-n`.

I think it may be confused if we use `--name`. Because the cmdlet is `az synapse **sql** ad-admin`.

src/azure-cli/azure/cli/command_modules/synapse/_params.py
aim-for-better: > question: would these two parameters are both required for the commands?
> I am wondering why object id could not determine the admin and we need name here.

Per my test, object_id is mandatory, the display_name is not and the display_name can be any string. My understanding iis display name is more friendly and  we do not have the logic that get display name by object id **in cli**.

src/azure-cli/azure/cli/command_modules/synapse/commands.py
aim-for-better: > please add wait command here

fixed.

src/azure-cli/azure/cli/command_modules/synapse/operations/workspace.py
aim-for-better: > recommend such kind of usage to get location from resource group
> https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/storage/_params.py#L212

Fixed.

src/azure-cli/azure/cli/command_modules/synapse/operations/sqlpoolblobauditingpolicy.py
aim-for-better: > such validation would be recommended to add as validator and other commands/parameters could utilize the function.
> see https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/storage/_params.py#L351

Hi @Juliehzl , thanks for your comment, I have removed the validation for retention_days. I agree with you, but per my understanding, the validator need to bind with one parameter which means one validator for one parameter, how should we validate multiple parameters in validator?

src/azure-cli/azure/cli/command_modules/synapse/_params.py
aim-for-better: > recommend to add validator for these parameter to make sure the format is right and make sense.

can we keep this? I think client side should not limit customer, backend will handle the format again.

src/azure-cli/azure/cli/command_modules/synapse/_client_factory.py
Juliehzl: but if we have some changes for get_mgmt_service_client, we need to change these two functions.
and resource module is owned by azure cli team and there should be safer.

src/azure-cli/azure/cli/command_modules/synapse/commands.py
```diff
@@ -200,13 +211,35 @@ def get_custom_sdk(custom_module, client_factory):
...
```
Juliehzl, line 223: prefer adding confirmation=true for delete

src/azure-cli/azure/cli/command_modules/synapse/_client_factory.py
aim-for-better: > but if we have some changes for get_mgmt_service_client, we need to change these two functions.
> and resource module is owned by azure cli team and there should be safer.

Ok, do you mean that when you update get_mgmt_service_client  you need to update all the reference? If all the services refer the resource module, it will be easy for you to maintain.  

Now I have removed this in client factory because this is for get resource group's location. Now the location has use the validator "get_default_location_from_resource_group", so we do not need this.  Thanks~

src/azure-cli/azure/cli/command_modules/synapse/commands.py
aim-for-better: > prefer adding confirmation=true for delete

fixed.

src/azure-cli/azure/cli/command_modules/synapse/_params.py
Juliehzl: interesting. In this way, what is the role for display name in your funtion?

aim-for-better:
Hi @jsntcy  Could you please help review this PR? It is almost 5 work days. We hope we can catch S180. Thank you very much~

src/azure-cli/azure/cli/command_modules/synapse/_params.py
aim-for-better: > interesting. In this way, what is the role for display name in your funtion?

The display name provided by customer will be shown in portal.

src/azure-cli/azure/cli/command_modules/synapse/operations/sqlpoolblobauditingpolicy.py
aim-for-better: Hi @Juliehzl , for this comments, could you please give some suggestions? Thanks

src/azure-cli/azure/cli/command_modules/synapse/_params.py
Juliehzl: I see.

src/azure-cli/azure/cli/command_modules/synapse/_params.py
Juliehzl: The display name should be bound with object id and it is pre-defined for the admin, right?
From my previous understanding, the display name is already fixed with object id.
Could your command redefine the display name?

src/azure-cli/azure/cli/command_modules/synapse/commands.py
```diff
@@ -200,13 +211,35 @@ def get_custom_sdk(custom_module, client_factory):
...
+        g.wait_command('wait')
+
+    # Management Plane Commands --Sql audit-policy
+    with self.command_group('synapse sql audit-policy',
+                            command_type=synapse_workspace_managed_sqlserver_blob_auditing_policies_sdk,
+                            custom_command_type=get_custom_sdk('sqlpoolblobauditingpolicy',
+                                                               cf_synapse_client_sqlserver_blob_auditing_policies_factory),
+                            client_factory=cf_synapse_client_sqlserver_blob_auditing_policies_factory) as g:
+        g.show_command('show', 'get')
+        g.generic_update_command('update', custom_func_name='sqlserver_blob_auditing_policy_update')
```
Juliehzl, line 233: only two commands for audit-policy? What about delete, create, list?

src/azure-cli/azure/cli/command_modules/synapse/tests/latest/test_synapse_scenario.py
```diff
@@ -346,6 +347,57 @@ def test_sql_pool_audit_policy(self):
...
+            'storage-account': 'teststorageforsynapse'
+        })
+
+        self.cmd('az synapse sql audit-policy update --state Enabled --storage-account {storage-account} '
+                 '--workspace-name {workspace} --resource-group {rg}')
+
+        self.cmd('az synapse sql audit-policy show --workspace-name {workspace} --resource-group {rg}',
+                 checks=[
+                     self.check('state', 'Enabled')
+                 ])
```
Juliehzl, line 399: 1. could we also check the update command?
2. could we get first, update and then get again to see the value change?
3. please add update for all parameters.

src/azure-cli/azure/cli/command_modules/synapse/tests/latest/test_synapse_scenario.py
```diff
@@ -346,6 +347,57 @@ def test_sql_pool_audit_policy(self):
...
+        self.cmd('az synapse sql ad-admin update --workspace-name {workspace} --resource-group {rg} '
+                 '--display-name {user-email}',
+                 checks=[
+                     self.check('login', self.kwargs['user-email'])
+                 ])
+        # Test delete cmdlet
+        self.cmd('az synapse sql ad-admin delete --workspace-name {workspace} --resource-group {rg} -y')
+        self.cmd('az synapse sql ad-admin show --workspace-name {workspace} --resource-group {rg}', expect_failure=True)
+
+    @record_only()
```
Juliehzl, line 384: why use record only here?

src/azure-cli/azure/cli/command_modules/synapse/tests/latest/test_synapse_scenario.py
```diff
@@ -346,6 +347,57 @@ def test_sql_pool_audit_policy(self):
                      self.check('state', 'Enabled')
                  ])
 
+    @record_only()
```
Juliehzl, line 350: why record only here?

yonzhan:
Please address the comments raised by @Juliehzl before merging this PR.

src/azure-cli/azure/cli/command_modules/synapse/_params.py
aim-for-better: We need to keep consistent with `az sql server ad-admin`.
In synapse `az synapse sql ad-admin create` the display name and object id are both mandatory. And the display name is no fixed, it can be anything, it just display in portal.

image

src/azure-cli/azure/cli/command_modules/synapse/tests/latest/test_synapse_scenario.py
aim-for-better: For this test we need an existing aad object id which is very confidential.

src/azure-cli/azure/cli/command_modules/synapse/tests/latest/test_synapse_scenario.py
aim-for-better: The related test need to create workspace, create sql pool, create storage and assign permissions, it is hard to do in live mode.

src/azure-cli/azure/cli/command_modules/synapse/tests/latest/test_synapse_scenario.py
aim-for-better: > 1. could we also check the update command?
> 2. could we get first, update and then get again to see the value change?
> 3. please add update for all parameters.

Thanks, I have updated the test for `az synapse sql audit-policy` and `az synapse sql pool audit-policy`

src/azure-cli/azure/cli/command_modules/synapse/commands.py
aim-for-better: Here we just to keep consistent with `az sql server`. And I checked for `az synapse sql audit-policy` the dependent api is a long running, and for `az synapse sql pool audit-policy` the dependent api is not long running.
I have add wait command and add `suppport_no_wait=True` for this comments.

src/azure-cli/azure/cli/command_modules/synapse/operations/sqlpoolblobauditingpolicy.py
aim-for-better: I have moved this function to _validators.py and add validator in command level. Thanks for this comments, it is my first time to know that we can add validator in command level.

Juliehzl (approved):

aim-for-better:
> Please address the comments raised by @Juliehzl before merging this PR.

Sure. Thanks

jsntcy (approved):

sunsw1994 (approved):

Issue #16160, opened 2020-12-04
[Compute] Request to Add "enableAutomaticUpgrade" flag for VMExtensions in CLI


chjaress:
**Resource Provider**
Microsoft.Compute (CRP)

**Description of Feature or Work Requested**
Need to add support for "enableAutomaticUpgrade" flag as part of VMExtension model. This bool flag will indicate if the extension should be automatically upgraded by Compute Platform.

**Minimum API Version Required**
2020-06-01

**Swagger Link**
https://github.com/Azure/azure-rest-api-specs/pull/10134

**Target Date**
Jan 31, 2021

yungezz:
ETA end of Jan 31st

yungezz:
spec https://nam06.safelinks.protection.outlook.com/ap/w-59584e83/?url=https%3A%2F%2Fmicrosoft-my.sharepoint-df.com%2F%3Aw%3A%2Fp%2Fmanayar%2FEYTyarAcsjRKkLm2jXL5hHoBjfDeLhONxARBbRmCGrlcWw%3Fe%3Dvqqhxg&data=04%7C01%7Cyungez%40microsoft.com%7C3663e6f8d3bc4062cac108d898b29ca7%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C1%7C637427240217333366%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C1000&sdata=AaHFZt47ukoPIXvaZeyWcTS3qeeQTRMUTFksaN%2BuwyA%3D&reserved=0

PR #16243, opened 2020-12-11
[Compute] vm/vmss extension set: New parameter --enable-auto-upgrade

qwordy:
**Description**

Resolve https://github.com/Azure/azure-cli/issues/16160

New parameter in vm/vmss extension set
```
--enable-auto-upgrade
Indicate the extension should be automatically upgraded by the platform if there is a newer version of the extension available
```
 
Change `--no-auto-upgrade` to `--no-auto-upgrade-minor-version` to avoid confusion.

Usage statistics in 30 days:
592 invocations, `vm extension` with `--no-auto-upgrade`
113 invocations, `vmss extension` with `--no-auto-upgrade`
665640 invocations, `vm extension`
26701 invocations, `vmss extension`

**Testing Guide**

```
az vm/vmss extension set --enable-auto-upgrade true/false
```

**History Notes**

---


yonzhan:
Compute

src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -789,6 +789,7 @@ def load_arguments(self, _):
                    help='Name of the extension.', id_part=None)
         c.argument('force_update', action='store_true', help='force to update even if the extension configuration has not changed.')
         c.argument('extension_instance_name', extension_instance_name_type)
+        c.argument('enable_automatic_upgrade', options_list=['--enable-automatic-upgrade', '-a'], action='store_true', help='Indicate the extension should be automatically upgraded by the platform if there is a newer version of the extension available.')
```
fengzhou-msft, line 792: Is there a way to disable it after it's enabled?

qwordy: `vm extension` does not support `update`. We can `set` it again.

yungezz (changes requested):
could you pls share rationale behind the breaking change

qwordy:
Given the usage statistics, do you think we can make a breaking change? I can revert it if you are worry about the impact to users. @yungezz

src/azure-cli/azure/cli/command_modules/vm/_help.py
```diff
@@ -1391,6 +1391,10 @@
   - name: Add a customScript extension to VM(s) specified by --ids.
     text: |
         az vm extension set -n customScript --publisher Microsoft.Azure.Extensions --ids {vm_id}
+  - name: Add an extension and enable automatic upgrade.
```
yungezz, line 1394: pls use more accurate description from Swagger to make the description clear.  From Swagger
```
the extension should be automatically upgraded by the platform if there is a newer version of the extension available.
```

src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -640,7 +640,10 @@ def load_arguments(self, _):
     # region VM & VMSS Shared
     for scope in ['vm', 'vmss']:
         with self.argument_context(scope) as c:
-            c.argument('no_auto_upgrade', arg_type=get_three_state_flag(), help='If set, the extension service will not automatically pick or upgrade to the latest minor version, even if the extension is redeployed.')
+            c.argument('no_auto_upgrade',
+                       options_list=['--no-auto-upgrade-mi', c.deprecate(target='--no-auto-upgrade', redirect='--no-auto-upgrade-mi')],
```
yungezz, line 644: for options_list, I would suggest to keep sync with Swagger property name so user has consistent context. Since Swagger properties `autoUpgradeMinorVersion` and `enableAutomaticUpgrade` is confusing enough already. I can see the difficult here. `--no-auto-upgrade` is `!autoUpgradeMinorVersion`.  How about `--no-auto-upgrade-minorversion`?

yungezz: hi @qwordy , what's future plan of original param `no-auto-upgrade`, If it keeps there, this will not be a breaking change.

qwordy: Yes. It is really confusing. I suggest keep "no" prefix for it is already here. "--no-auto-upgrade-minorversion" is too long. It can't pass linter. We need a better name to express the meaning clearly and not be very long. How about "--no-aum"?
Yes. The old parameter can still be used. It is not a breaking change for now.

yungezz: Agree that `--no-auto-upgrade-minorversion` is long. While first thing here is clarity. I have one question: If user set `--no-auto-upgrade-mi` to true,  while `--enable-auto-upgrade` true. What will happen when there's new minor version released of extension? WIll `enableAutomaticUpgrade` overwrite `autoUpgradeMinorVersion` value?

yungezz (approved):

PR #16251, opened 2020-12-13
Update custom.py

talalhaddad:
Fix rcf.template_specs.delete If VersionID presented

**Description**
CLI command for delete Template specs is deleting the template specs instead of the provided version alone, else statement adder

**Testing Guide**
az ts delete -g RGName --n TSName -v 1.0 
this was deleting the template specs resource and not the version needed, now will delete the version alone in case version ID provided 

**History Notes**
[Template Specs] az command Delete: Fix version parameter issue.

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/resource/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/resource/custom.py
@@ -1899,7 +1899,8 @@ def delete_template_spec(cmd, resource_group_name=None, name=None, version=None,
             version = None
     if version:
         return rcf.template_specs.delete(resource_group_name=resource_group_name, template_spec_name=name, template_spec_version=version)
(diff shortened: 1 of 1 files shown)


yonzhan:
ARM

jiasli:
This is a bot spamming our repo. 

The change is also against the linter rule. 

https://dev.azure.com/azure-sdk/public/_build/results?buildId=652812&view=logs&j=36dd4138-4d53-5e46-00d9-e5cd9744be05&t=1cf3879c-0a42-5ccd-7693-7a4781d739d8&l=107

```
src/azure-cli/azure/cli/command_modules/resource/custom.py:1900:4: R1705: Unnecessary "else" after "return" (no-else-return)
```

jiasli (changes requested):
Should close.

zhoxing-ms:
It was replaced by #16295

PR #16254, opened 2020-12-14
[Extension] Support extension command prefix match for dynamic install

fengzhou-msft:
**Description**

Previously we only match users' input command with a complete command (not including options), so command like `az blueprint -h` will not be matched with an extension command and dynamic install of the extension will not be triggered.

This PR adds support for command prefix match so that commands like `az blueprint`, `az blueprint -h` will also trigger the dynamic install of the `blueprint` extension. It also handle cases when the command matches the prefix of multiple extensions. It will prompt to let users select an extension to install.

This PR also fixes the false positive of misspelled commands by checking extension command match first. Previously, if a candidate of close match to core module command is found, the command is considered to be misspelled. It increased the preformance but have some false positves.

**Testing Guide**

1. Run `az blueprint -h` and CLI will prompt:

> The command requires the extension blueprint. Do you want to install it now? The command will continue to run after the extension is installed. (Y/n):

2. We don't have cases that will trigger the selection of multiple extensions, you can mock up by adding `{"fake-cmd": "fake-ext"}` to some extension in the `~/.azure/extensionCommandTree.json` (or in your `env/.azure/extensionCommandTree.json`)，for instance under `vm aem`. Then run `az vm aem -h`, the output would be:

> The command requires the latest version of one of the following extensions. You need to pick one to install:
>  [1] aem
>  [2] fake-ext
> Please enter a choice [Default choice(1)]: 1
> The command requires the extension aem. It will be installed first.
>  \- Installing ..
> ...

3. For misspelled case, `azcopy` was considered as a typo for `copy` which is in the main module command. With this PR if you run `az storage azcopy blob delete -c test1 -t a.txt --account-name mydefault`, it should prompt you to install `storage-preview`.

**History Notes**

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/parser.py
+++ b/src/azure-cli-core/azure/cli/core/parser.py
@@ -318,9 +318,8 @@ class AzCliCommandParser(CLICommandParser):
                 return None
         return EXT_CMD_TREE
 
-    def _get_all_extensions(self, cmd_chain, exts):
-        """Find all the extension names in cmd_chain (dict of extension command subtree) and store
-        in exts.
+    def _get_all_extensions(self, cmd_chain, ext_set=None):
+        """Find all the extension names in cmd_chain (dict of extension command subtree).
         An example of cmd_chain may look like (a command sub tree of the 'aks' command group):
         {
             "create": "aks-preview",
@@ -330,13 +329,15 @@ class AzCliCommandParser(CLICommandParser):
             },
             "use-dev-spaces": "dev-spaces"
         }
-        Then the resulting exts is {'aks-preview', 'deploy-to-azure', 'dev-spaces'}
+        Then the resulting ext_set is {'aks-preview', 'deploy-to-azure', 'dev-spaces'}
         """
+        ext_set = set() if ext_set is None else ext_set
         for key in cmd_chain:
             if isinstance(cmd_chain[key], str):
-                exts.add(cmd_chain[key])
+                ext_set.add(cmd_chain[key])
             else:
-                self._get_all_extensions(cmd_chain[key], exts)
+                self._get_all_extensions(cmd_chain[key], ext_set)
+        return ext_set
 
     def _search_in_extension_commands(self, command_str):
         """Search the command in an extension commands dict which mimics a prefix tree.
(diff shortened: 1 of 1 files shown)


yonzhan:
@jiasli for awareness

yungezz (approved):

src/azure-cli-core/azure/cli/core/parser.py
```diff
@@ -346,7 +367,9 @@ def _search_in_extension_commands(self, command_str):
                 cmd_chain = cmd_chain[part]
             except KeyError:
                 return None
-        return None
+        # command_str is prefix of one or more complete commands.
+        all_exts = self._get_all_extensions(cmd_chain)
+        return list(all_exts) if all_exts else None
```
houk-ms, line 372: if the `command_str` starts with a `not-exist-command`, this function will return all the extension names in root cmd_chain, right?

src/azure-cli-core/azure/cli/core/parser.py
```diff
@@ -377,26 +400,41 @@ def _check_value(self, action, value):  # pylint: disable=too-many-statements, t
...
             if not self.command_source:
-                candidates = difflib.get_close_matches(value, action.choices, cutoff=0.7)
-                if candidates:
-                    # use the most likely candidate to replace the misspelled command
-                    args = self.prog.split() + self._raw_arguments
-                    args_inferred = [item if item != value else candidates[0] for item in args]
-                    command_name_inferred = ' '.join(args_inferred).split('-')[0]
-
+                candidates = []
+                args = self.prog.split() + self._raw_arguments
```
houk-ms, line 404: Are there any possibilities that `args` begins with an unknown command name, and then `_search_in_extension_commands` would return all the extension names？

src/azure-cli-core/azure/cli/core/parser.py
fengzhou-msft: In this case, KeyError will be thrown and return None in above code. This function will not be called.

src/azure-cli-core/azure/cli/core/parser.py
fengzhou-msft: I don't think so. `_search_in_extension_commands` will return a list of extension names only when the args matches at least one word (not including the beginning `az`) in extension commands. Otherwise it returns `None`.

houk-ms (approved):

Juliehzl (approved):

PR #16258, opened 2020-12-14
{Style} Refine style framework

jiasli:
## Description

Refine the style framework.

1. Since `BLUE` and `LIGHTBLUE_EX` are not very visible/accessible under `powershell.exe` (see the demo at https://github.com/jiasli/color-demo), they are replaced with `LIGHTWHITE_EX`.

   ℹ This doesn't apply to PowerShell sessions inside modern terminals, see item 5.

   - Bring back `get_parent_proc_name` removed by #11665. #15076 has made sure that **psutil** is not installed under **cygwin**.

2. Honor `cli_ctx.enable_color` so that when color is disabled, it can remove color when formatting styled text. This can eliminate the necessity to make 2 versions of the same string, one with color, one without, like

   https://github.com/Azure/azure-cli/blob/026016e7bf205bfa378f79351f99c9d536efa3d5/src/azure-cli-core/azure/cli/core/commands/constants.py#L35-L37

   In the demo, you may easily turn off color with    
   ```
   az demo style --no-color
   ```

3. Change the color of primary and secondary text:
   - Primary color: Use default **foreground** color instead of `BRIGHT WHITE` so that at least primary text is visible under both dark/light themes
   - Secondary color: Use `BRIGHT BLACK`

4. Add light theme support which uses dark versions of colors. To try it, in light/dark-themed terminals, run
   ```
   az demo style --theme light/dark
   ```

5. Add function `is_modern_terminal` to detect if the terminal is a modern terminal like [VS Code](https://code.visualstudio.com/), [Windows Terminal](https://github.com/microsoft/terminal).

6. `print_styled_text` wraps `print` and can pass `kwargs`.

7. `format_styled_text` supports more formats.

## Testing Guide

```
az demo style
```

`powershell.exe`:

image

`pwsh.exe`:

image


yonzhan:
Style

src/azure-cli-core/setup.py
```diff
@@ -66,6 +66,10 @@
     'azure-common~=1.1',
 ]
 
+# dependencies for specific OSes
+if not sys.platform.startswith('cygwin'):
+    DEPENDENCIES.append('psutil~=5.7')
+
```
jiasli, line 72: Move the dependency of `psutil` to core as it is used by core, not command modules. @houk-ms

chenlomis:
LGTM from the design side as per our latest discussion!

zhoxing-ms:
May I ask could we support `end` parameter of `print()` in the `print_styled_text()` function?

At present, I have a requirement that the message of `input()` should also be colored, such as:
Screenshot 2020-12-28 161054

My previous implementation is:
``` python
print("Please input " + Fore.LIGHTBLUE_EX + param + Fore.RESET + ":", end='')
value = input()
```

Or do we have other better ways to implement this requirement, such as wrapping a specialized method?

jiasli:
@zhoxing-ms, `end` is already supported. Note that `kwargs` is passed to `print` as is.

https://github.com/Azure/azure-cli/blob/fa9985e7f18f5624cd21484404c313368a1909bd/src/azure-cli-core/azure/cli/core/style.py#L82

src/azure-cli-core/azure/cli/core/style.py
```diff
@@ -46,29 +51,129 @@ class Style(str, Enum):
...
+    if isinstance(theme, str):
+        if theme.lower() == 'none':
+            theme = THEME_NONE
+        elif theme.lower() == 'dark':
+            theme = THEME_DARK
+        elif theme.lower() == 'light':
+            theme = THEME_LIGHT
+        else:
+            from azure.cli.core.azclierror import CLIInternalError
+            raise CLIInternalError("Invalid theme. Supported themes: none, dark, light")
```
zhoxing-ms, line 111: May I ask why do we need to support the case that `theme` is string?

By the way, could we consider setting a theme configuration of dict or enum for this mapping relationship instead of `if-else` logic? Such as:
```
THEME = {
    'none': THEME_NONE,
    'dark': THEME_DARK,
    'light': THEME_LIGHT
}
```
I personally think it might be more readable~

src/azure-cli-core/azure/cli/core/tests/test_util.py
```diff
@@ -412,6 +413,36 @@ def test_resource_to_scopes(self):
...
+        parent1.name.return_value = "cmd.exe"
+        parent2.name.return_value = "explorer.exe"
+        self.assertEqual(_get_parent_proc_name(), "cmd.exe")
+
+        # Linux
+        process.name.return_value = "python"
+        parent1.name.return_value = "bash"
+        parent2.name.return_value = "init"
+        parent3.name.return_value = "init"
+        self.assertEqual(_get_parent_proc_name(), "bash")
```
zhoxing-ms, line 444: Out of curiosity, I just want to confirm that Mac and Linux are exactly the same, right?

jiasli: Yes. Mac is also using `bash`, but it doesn't really mater as long as no ancestor process is called `powershell.exe`, `pwsh.exe`. 

However, I think `pwsh` running on Linux can be a more complex scenario. I need to do more testing on that.

src/azure-cli-core/azure/cli/core/style.py
jiasli: > May I ask why do we need to support the case that `theme` is string?

Because everything passed to CLI via the shell will be treated as `str`.

Moreover, the `value`s of an `Enum` are usually basic types like `str` or `int`. Using `dict` as the value, like

```py
class Theme(Enum):
    DARK = THEME_DARK
    LIGHT = THEME_LIGHT
    NONE = THEME_NONE
```

will result in:
```
> az demo style -h

Arguments
    --theme            : The theme to format styled text.  Allowed values: {<Style.PRIMARY:
                         'primary'>: '\x1b[39m', <Style.SECONDARY: 'secondary'>: '\x1b[90m',
                         <Style.IMPORTANT: 'important'>: '\x1b[35m', <Style.ACTION: 'action'>:
                         '\x1b[34m', <Style.HYPERLINK: 'hyperlink'>: '\x1b[36m', <Style.ERROR:
                         'error'>: '\x1b[31m', <Style.SUCCESS: 'success'>: '\x1b[32m',
                         <Style.WARNING: 'warning'>: '\x1b[33m'}, {<Style.PRIMARY: 'primary'>:
                         '\x1b[39m', <Style.SECONDARY: 'secondary'>: '\x1b[90m', <Style.IMPORTANT:
                         'important'>: '\x1b[95m', <Style.ACTION: 'action'>: '\x1b[94m',
                         <Style.HYPERLINK: 'hyperlink'>: '\x1b[96m', <Style.ERROR: 'error'>:
                         '\x1b[91m', <Style.SUCCESS: 'success'>: '\x1b[92m', <Style.WARNING:
                         'warning'>: '\x1b[93m'}, {}.  Default: dark.
```

But the idea of using `Enum` for `Theme` is good. Added as requested:

https://github.com/Azure/azure-cli/blob/6240eafbd6829460b1a1b4a9e4a156d8675f954d/src/azure-cli-core/azure/cli/core/style.py#L67-L77

zhoxing-ms: OK

zhoxing-ms (approved):
LGTM

houk-ms (approved):
@jiasli When do you plan to get this PR merged?

PR #16294, opened 2020-12-16
{Release} Hotfix: TS Supports Deleting a Single Version 

detienne20:
Description: 

Customer reported template specs (ts) issue where attempting to delete a single version of the template spec deleted the template spec and all the versions. This was the result of the delete_template_spec function calling the _template_spec_delete_ operation instead of the _template_spec_versions_ delete operation. 

Note: https://github.com/Azure/azure-cli/pull/16251 should be closed. 

Testing Guide: 

az ts delete -g resource_group --name Issue37 -v 1.0 

az ts list -g resource_group --name Issue37  -> should return the parent template spec and versions (excluding version 1.0)


detienne20:
Wrong Branch.

PR #16302, opened 2020-12-17
{Packaging} Fix version to conform to pip 20.3

fengzhou-msft:
**Description**

Edge build is broken as pip now compares the version in the wheel package name and the version in Metadata. Previously, we only added `.post1` in package name. This PRs changes the version in source code with `.post1` suffix as well when building packages for non-release branch code. This also helps us identify requests from edge builds in telemetry data.

CI is broken as `azure-cli-core` previously pinned `azure-cli-telemetry` to `1.0.6` while in CI `1.0.6.devxxx` is installed. Latest pip will throw an error. This PR changes the dependency version to `azure-cli-telemetry==1.0.6.*`.

**Testing Guide**

**History Notes**

---


scripts/ci/build.sh
```diff
@@ -86,7 +86,7 @@ cat >$testsrc_dir/setup.py <<EOL
 
 from setuptools import setup
 
-VERSION = "1.0.0.dev$version"
+VERSION = "1.0.0.post1"
```
jiasli, line 89: Wouldn't `1.0.1.devN` be more feasible?

yonzhan:
Packaging

scripts/ci/build.sh
fengzhou-msft: Reverted this change for CI (I don't think it matters what version to use in CI as it will not be released.), but refactored to put `dev` inside `$version` and for edge built python wheel `$version` is changed to `postN`.

With the increasing number of unexpected hot fix release, it may not be a good idea to bump version right after a release and use `devN` for edge build versions.

jiasli: > With the increasing number of unexpected hot fix release, it may not be a good idea to bump version right after a release and use `devN` for edge build versions.

I think hotfix PRs will target `release` branch, which still has the last version, like `2.16.0`, even though the `dev` branch is now at `2.17.0`.

haroldrandom (changes requested):
require a test PR/a test result for BuldPythonWheel and Automation Test and

scripts/ci/build.sh
fengzhou-msft: Let's discuss about this in our next release meeting. The CI in `azure-cli-extensions` repo is broken now, let's merge this PR to unblock it first. If we decided to use `dev` version, we can start to do it after the next release.

haroldrandom (approved):

PR #16415, opened 2021-01-02
{RDBMS} : Stabilize tests for live-test runs in pipeline.

arde0708:
This PR introduces some changes required for supporting a pipeline that would monitor the rdbms modules for 
any major issues or breaking changes. The changes include:
1. Update location to Canary - EastUS2 EUAP for all tests. The location needed to be changed from SouthEastAsia as vnet creation is not supported in SouthEastAsia.
2. Update SKU's from *_v4 to *_v3. This is again because *_v4's are not supported in Canary(EastUS2 EUAP) which was previously used in all our tests.
3. Introduce explicit --location parameter in the commands wherever applicable. This is because, not mentioning the explicit location would pick up the default locations which is different from EastUS2 EUAP.
4. Recordings of all tests due to the change in target location of the tests.

Here is the record for last six successful runs of the pipeline:
https://msdata.visualstudio.com/Database%20Systems/_build?definitionId=14128&_a=summary


src/azure-cli/azure/cli/command_modules/rdbms/tests/latest/test_rdbms_flexible_commands.py
```diff
@@ -844,6 +860,7 @@ def _test_flexible_server_public_access_mgmt(self, database_engine, resource_gro
                  checks=NoneCheck())
 
 
+'''
```
arde0708, line 863: This test is intentionally commented out for few reasons. The goal of this test was to run the same commands with local context turned ON. However, the major problem this test was causing is :
1. The test should ideally run the commands and check for successful execution of the command **OR** it should validate the output of the commands. Currently, it is missing out on both. 
2. Since the tests execute in parallel and this test **turns on** the local context, it is inconsistently impacting the other tests caused by a race condition. If the local-context tests get kicked off before other tests, all commands in other tests is starting to pick up the values from local context since local context is turned on in the environment due to this  test. This leads to the failure of other live tests.

@DaeunYim will address these changes and send out a different PR.

yonzhan:
RDBMS

src/azure-cli/azure/cli/command_modules/rdbms/flexible_server_virtual_network.py
```diff
@@ -213,7 +213,7 @@ def create_vnet(cmd, servername, location, resource_group_name, delegation_servi
                                                    VirtualNetwork(name=vnet_name, location=location,
                                                                   address_space=AddressSpace(
                                                                       address_prefixes=[vnet_address_prefix])))
-    delegation = Delegation(name=delegation_service_name, serv_name=delegation_service_name)
+    delegation = Delegation(name=delegation_service_name, service_name=delegation_service_name)
```
evelyn-ys, line 216: We have fix this in #16381, after the horfix is GA, we will merge this change into `dev` branch. So there's no need to modify it again here.

arde0708: Will remove it before check in. Had to include it in the PR for the CI to pass.

evelyn-ys: The hotfix is released. You can merge `dev` branch now😆

arde0708: @evelyn-ys : Reverted the change and merged with dev. Can you take a look and approve/merge this PR whenever you get a chance ?

evelyn-ys (approved):

PR #16446, opened 2021-01-07
[ARM] ts: error handling improvement - use of --template-file without --version prohibited

detienne20:
**Description**

Bug found: az ts create does not update the template pec content when the version is not specified and no useful error message or warning is shown to specify --version is required. 

This PR addresses this bug by throwing an incorrect usage error when --template-file is provided without the --version parameter present.  

**Testing Guide**

az ts create --name TemplateSpecName --resource-group ResourceGroupName --template-file "azuredeploy.json"  should result in incorrect usage error.


yonzhan:
ARM

zhoxing-ms (approved):

PR #16400, opened 2020-12-30
[ARM] Az Deployments : Add support for --query-string

detienne20:
PR is dependent on https://github.com/Azure/azure-cli/pull/16179. 

**Description**

Az deployments now supports  --query-string to be used with --template-uri for linked template deployments; --query-string is a SAS token for a storage container that contains the main template and linked template(s).  Support for the --query-string parameter will facilitate the transition to template specs. Like with template specs linked templates are referenced through the relative path field in template link as illustrated below:
      
        "templateLink":  {
                  "relativePath":"linked_template.json",
                   "contentVersion":"1.0.0.0"
         }

**Testing Guide**

1. Create storage account container and upload files for deployment. 

     az storage account create -n {} -g {} -l {} --sku {} --kind {} 
     az storage container create -n {} --account-name {}
     az storage blob upload -c {} -f "{}" -n {} --account-name {}

2. Generate an SAS token for the entire container (query-string parameter) 

     az storage container generate-sas --account-name {} --account-key {} --name {} --permissions dlrw --expiry {}

3. Generate the url for the main template (template-uri parameter) 

     az storage blob url -c {} -n {} --account-name {}

4. Deploy with --template-uri and --query-string parameter. 

     az deployment group create -g {} --template-uri {blob_url} --query-string "{sas_token}"
     az deployment sub create -g {} --template-uri {blob_url} --query-string "{sas_token}"
     az deployment mg validate  -g {} --template-uri {blob_url} --query-string "{sas_token}"
     az deployment tenant validate  -g {} --template-uri {blob_url} --query-string "{sas_token}"

PR #16447, opened 2021-01-07
[ARM] Az Deployments : Add support for --query-string

detienne20:
Closed : https://github.com/Azure/azure-cli/pull/16400 

**Description**

Az deployments now supports  --query-string to be used with --template-uri for linked template deployments; --query-string is a SAS token for a storage container that contains the main template and linked template(s).  Support for the --query-string parameter will facilitate the transition to template specs. Like with template specs linked templates are referenced through the relative path field in template link as illustrated below:
      
        "templateLink":  {
                  "relativePath":"linked_template.json",
                   "contentVersion":"1.0.0.0"
         }

**Testing Guide**

1. Create storage account container and upload files for deployment. 

     az storage account create -n {} -g {} -l {} --sku {} --kind {} 
     az storage container create -n {} --account-name {}
     az storage blob upload -c {} -f "{}" -n {} --account-name {}

2. Generate an SAS token for the entire container (query-string parameter) 

     az storage container generate-sas --account-name {} --account-key {} --name {} --permissions dlrw --expiry {}

3. Generate the url for the main template (template-uri parameter) 

     az storage blob url -c {} -n {} --account-name {}

4. Deploy with --template-uri and --query-string parameter. 

     az deployment group create -g {} --template-uri {blob_url} --query-string "{sas_token}"
     az deployment sub create -g {} --template-uri {blob_url} --query-string "{sas_token}"
     az deployment mg validate  -g {} --template-uri {blob_url} --query-string "{sas_token}"
     az deployment tenant validate  -g {} --template-uri {blob_url} --query-string "{sas_token}"

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/resource/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/resource/_params.py
@@ -51,6 +51,7 @@ def load_arguments(self, _):
                                                     help="a template file path in the file system")
     deployment_template_uri_type = CLIArgumentType(options_list=['--template-uri', '-u'], help='a uri to a remote template file')
     deployment_template_spec_type = CLIArgumentType(options_list=['--template-spec', '-s'], is_preview=True, min_api='2019-06-01', help="The template spec resource id.")
+    deployment_query_string_type = CLIArgumentType(options_list=['--query-string', '-q'], is_preview=True, help="The query string (a SAS token) to be used with the template-uri in the case of linked templates.")
     deployment_parameters_type = CLIArgumentType(options_list=['--parameters', '-p'], action='append', nargs='+', completer=FilesCompleter(), help='the deployment parameters')
     filter_type = CLIArgumentType(options_list=['--filter'], is_preview=True,
                                   help='Filter expression using OData notation. You can use --filter "provisioningState eq \'{state}\'" to filter provisioningState. '
@@ -262,6 +263,7 @@ def load_arguments(self, _):
         c.argument('template_file', arg_type=deployment_template_file_type)
         c.argument('template_uri', arg_type=deployment_template_uri_type)
         c.argument('template_spec', arg_type=deployment_template_spec_type)
+        c.argument('query_string', arg_type=deployment_query_string_type)
         c.argument('parameters', arg_type=deployment_parameters_type)
 
     with self.argument_context('deployment create') as c:
--- a/src/azure-cli/azure/cli/command_modules/resource/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/resource/custom.py
@@ -428,7 +428,7 @@ def deploy_arm_template_at_subscription_scope(cmd,
                                               deployment_name=None, deployment_location=None,
                                               no_wait=False, handle_extended_json_format=None, no_prompt=False,
                                               confirm_with_what_if=None, what_if_result_format=None,
-                                              what_if_exclude_change_types=None, template_spec=None):
+                                              what_if_exclude_change_types=None, template_spec=None, query_string=None):
     if confirm_with_what_if:
         what_if_deploy_arm_template_at_subscription_scope(cmd,
                                                           template_file=template_file, template_uri=template_uri,
@@ -436,7 +436,7 @@ def deploy_arm_template_at_subscription_scope(cmd,
                                                           deployment_location=deployment_location,
                                                           result_format=what_if_result_format,
                                                           exclude_change_types=what_if_exclude_change_types,
-                                                          no_prompt=no_prompt, template_spec=template_spec)
+                                                          no_prompt=no_prompt, template_spec=template_spec, query_string=query_string)
         from knack.prompting import prompt_y_n
 
         if not prompt_y_n("\nAre you sure you want to execute the deployment?"):
@@ -446,7 +446,7 @@ def deploy_arm_template_at_subscription_scope(cmd,
                                                       template_file=template_file, template_uri=template_uri, parameters=parameters,
                                                       deployment_name=deployment_name, deployment_location=deployment_location,
                                                       validate_only=False, no_wait=no_wait,
-                                                      no_prompt=no_prompt, template_spec=template_spec)
+                                                      no_prompt=no_prompt, template_spec=template_spec, query_string=query_string)
 
 
 # pylint: disable=unused-argument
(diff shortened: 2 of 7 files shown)


detienne20:
@zhoxing-ms @Juliehzl The query string tests are live tests because --query-string is an SAS token which cannot be included in PR (cred scan would fail).

src/azure-cli/azure/cli/command_modules/resource/tests/latest/test_resource.py
```diff
@@ -1012,6 +1012,99 @@ def test_export_template_spec(self, resource_group, resource_group_location):
 
 
 
+class DeploymentTestsWithQueryString(LiveScenarioTest):
```
detienne20, line 1015: @zhoxing-ms @Juliehzl The query string tests are live tests because --query-string is an SAS token which cannot be included in PR (cred scan would fail).

yonzhan:
ARM

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -51,6 +51,7 @@ def load_arguments(self, _):
                                                     help="a template file path in the file system")
     deployment_template_uri_type = CLIArgumentType(options_list=['--template-uri', '-u'], help='a uri to a remote template file')
     deployment_template_spec_type = CLIArgumentType(options_list=['--template-spec', '-s'], is_preview=True, min_api='2019-06-01', help="The template spec resource id.")
+    deployment_query_string_type = CLIArgumentType(options_list=['--query-string', '-q'], is_preview=True, help="The query string (a SAS token) to be used with the template-uri in the case of linked templates.")
```
zhoxing-ms, line 54: Does `--query-string` contain anything other than `sas-token`?
If `--query-string` is `sas-token`, why not name it `sas-token`?

There is `sas-token` parameter in storage module, so I personally think it's better to keep it consistent with storage
Such as:
```
storage file show -s {share} -p {file} --account-name {account} --sas-token
```

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -842,8 +842,18 @@ def _build_preflight_error_message(preflight_error):
...
 
 
+def _prepare_template_uri_with_query_string(template_uri, input_query_string):
+    from six.moves.urllib.parse import urlencode, parse_qs, urlsplit, urlunsplit  # pylint: disable=import-error
+
+    scheme, netloc, path, query_string, fragment = urlsplit(template_uri)  # pylint: disable=unused-variable
+    query_params = parse_qs(input_query_string)
+    new_query_string = urlencode(query_params, doseq=True)
+
+    return urlunsplit((scheme, netloc, path, new_query_string, fragment))
```
zhoxing-ms, line 852: Do we need to consider adding some verifications for `input_query_string` to make sure it is vaild?

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -852,8 +862,16 @@ def _prepare_deployment_properties_unmodified(cmd, template_file=None, template_
     template_obj = None
     on_error_deployment = None
     template_content = None
+
+    if query_string and not template_uri:
+        raise IncorrectUsageError('please provide --template-uri if --query-string is specified')
```
zhoxing-ms, line 867: Do we need to consider adding some verifications for `query_string` to make sure it is vaild?

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -842,8 +842,18 @@ def _build_preflight_error_message(preflight_error):
     return '\n'.join(err_messages)
 
 
+def _prepare_template_uri_with_query_string(template_uri, input_query_string):
+    from six.moves.urllib.parse import urlencode, parse_qs, urlsplit, urlunsplit  # pylint: disable=import-error
+
+    scheme, netloc, path, query_string, fragment = urlsplit(template_uri)  # pylint: disable=unused-variable
+    query_params = parse_qs(input_query_string)
+    new_query_string = urlencode(query_params, doseq=True)
```
zhoxing-ms, line 850: If there is an exception in parsing, does the current error raised meet the expectation? Do we need to consider the logic of adding `try-catch`?

src/azure-cli/azure/cli/command_modules/resource/custom.py
detienne20: This is done on the server side.

src/azure-cli/azure/cli/command_modules/resource/_params.py
detienne20: It does not contain anything other than sas-token, it is named query string in PowerShell so it may be best to keep it consistent (https://github.com/Azure/azure-powershell/pull/13775).

alex-frankel: today it is only used for sasToken, but we'd like to leave open the possibility that something else could be passed besides a sas token. `--query-string` keeps it generic, while still being accurate.

src/azure-cli/azure/cli/command_modules/resource/custom.py
detienne20: I added the try catch for cases I may not have thought of yet but the current errors raised does meet expectations. If the query string is not valid then query_params ={} and is not url encoded and the server side handles any errors very well. 

I did notice that when the url is retrieved to get the template_obj and prompt for parameters, if the template-uri is incorrect then the error message returned is <urlopen error [Errno 11001] getaddrinfo failed> for the current Azure CLI release. I did not make any changes to this.

src/azure-cli/azure/cli/command_modules/resource/_params.py
zhoxing-ms: Got it, thanks~

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -842,8 +842,21 @@ def _build_preflight_error_message(preflight_error):
...
+    from six.moves.urllib.parse import urlencode, parse_qs, urlsplit, urlunsplit  # pylint: disable=import-error
+
+    try:
+        scheme, netloc, path, query_string, fragment = urlsplit(template_uri)  # pylint: disable=unused-variable
+        query_params = parse_qs(input_query_string)
+        new_query_string = urlencode(query_params, doseq=True)
+
+        return urlunsplit((scheme, netloc, path, new_query_string, fragment))
+    except Exception:  # pylint: disable=broad-except
+        raise CLIError('Unable to parse parameter: {} .Make sure the value is formed correctly.'.format(input_query_string))
```
zhoxing-ms, line 855: How about using `InvalidArgumentValueError` in stead of `CLIError`?

src/azure-cli/azure/cli/command_modules/resource/custom.py
zhoxing-ms: > <urlopen error[Errno 11001]getaddrinfo failed>

This error message doesn't seem clear. Should we consider catching the parsing logic of `template_uri`?

detienne20: This actually comes from def _urlretrieve(url):

zhoxing-ms (approved):

Issue #16066, opened 2020-11-27
webapp:Unable to select a runtime version if it includes "space" in its name such as "Java SE"


v-yuwak:
> ### `az feedback` auto-generates most of the information requested below, as of CLI version 2.0.62

**Describe the bug**
As following github, it seems "space" inside runtime version name is considered as delimiter as coded.
https://github.com/Azure/azure-cli/blob/ed1fe586d24b5d78127c0b616cd7e32b765d0bfc/src/azure-cli/azure/cli/command_modules/appservice/custom.py#L2572
As a result, CLI outputs an error as follow which suggests "Java SE" is processed as "Java|SE".
"Windows runtime 'java\|1.8\|Java\|SE\|8' is not supported. Please invoke 'az webapp list-runtimes' to cross check"

**To Reproduce**
Execute the command with a runtime version that includes "space" in the name, such as "az webapp create ~ omitted ~ --runtime 'java | 1.8 | Java SE | 8'.

**Expected behavior**
"Space" inside their runtime version name should be treated as just space so that you can select "Java SE".

**Environment summary**
It occurs on both CloudShell or local Azure CLI, so it looks no matter what OS you use.
My Azure CLI version is "azure-cli": "2.11.0".

**Additional context**
N/A

yungezz:
route to appropriate team

Kotasudhakarreddy:
Able to reproduce the reported issue. Will look in to it further.

panchagnula:
@calvinsID this is due to the change we allow space to be a delimiter for runtime language & version. Can you take a look. Thanks!

PR #16528, opened 2021-01-13
[AppService] BREAKING CHANGE: Fix regression where you can't use a runtime version with a space in the name

calvinsID:
**Description**
Fixes issue #16066 

Before, az webapp up/create allowed you to specify `--runtime` parameter with delimiters "|", " ", and ":".
This parsing caused runtimes with a space (" ") in it to not be parsed correctly.

In this PR, removing ability to use space (" ") as a delimiter so that runtimes with a space in it get parsed correctly again

**Testing Guide**
`az webapp create -n {} -g {} --runtime "'java|1.8|Java SE|8" --os-type Windows` should work

---


panchagnula:
@calvinsID should we remove "" as a delimiter completely or can we just make this stricter.
As in if we have "" & "|" or ":" then we don't assume that " " is a delimiter here but part of the runtime value.

If we see multiple spaces in a case the runtime Value has a space & space is being used as delimiter can we do something to consider the last space in the string to be the delimiter?

calvinsID:
@panchagnula good idea I'll change so that if there are other delimiters first, we ignore the other delimiters. That way if a runtime which has spaces is used, it will work with other delimiters

panchagnula:
> @panchagnula good idea I'll change so that if there are other delimiters first, we ignore the other delimiters. That way if a runtime which has spaces is used, it will work with other delimiters

Sounds good @calvinsID also may be update the help doc to clarify that space is a supported delimiter & add examples on how to use this. Also if a runtime has space in language should't the recommendation be to  enclose them in single or double quotes "Java SE" so if the command is used as so

az webapp create -n {} -g {} --runtime "'Java SE"
we don't assume space to be a delimiter for "language <space>version"

calvinsID:
@panchagnula if the runtime has a space in the language it might be messy to allow the use of space as a delimiter, ex. `'"java 11 "Java SE" 8"'` or `'"java 11 Java SE 8"'`

src/azure-cli-core/azure/cli/core/profiles/_shared.py
```diff
@@ -152,7 +152,7 @@ def default_api_version(self):
         ResourceType.MGMT_RESOURCE_LINKS: '2016-09-01',
         ResourceType.MGMT_RESOURCE_LOCKS: '2016-09-01',
         ResourceType.MGMT_RESOURCE_POLICY: '2019-09-01',
-        ResourceType.MGMT_RESOURCE_RESOURCES: '2020-10-01',
+        ResourceType.MGMT_RESOURCE_RESOURCES: '2020-06-01',
```
calvinsID, line 155: Version bump caused our commands to fail with "APIVersion 2020-10-01 is not available" so I changed it back.. wondering if this will cause any newly added commands that needs the newest api version to fail?

panchagnula:
> @panchagnula if the runtime has a space in the language it might be messy to allow the use of space as a delimiter, ex. `'"java 11 "Java SE" 8"'` or `'"java 11 Java SE 8"'`



> @panchagnula if the runtime has a space in the language it might be messy to allow the use of space as a delimiter, ex. `'"java 11 "Java SE" 8"'` or `'"java 11 Java SE 8"'`

Agreed Please update help with some examples with different delimiters & add a test case for command with runtime with language value only like 'Java SE' i.e if we allow runtime version with language value only where we use default version value? otherwise change looks good to me.

calvinsID:
Added examples in the help text. We currently don't allow language value only like "Java SE" though we can add this in a future PR

src/azure-cli-core/azure/cli/core/profiles/_shared.py
qwordy: It is not an easy job to change API version. All tests that use this package will fail without rerecording them again.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2619,7 +2619,13 @@ def __init__(self, cmd, client, linux=False):
     @staticmethod
     def remove_delimiters(runtime):
         import re
-        runtime = re.split('[| :]', runtime)  # delimiters allowed: '|', ' ', ':'
+        # delimiters allowed: '|', ':', ' '
```
qwordy, line 2622: When a user use hybrid delimiter, what will happen? For example, java|1.8:Java SE 8

qwordy: Should we cancel delimiter ' '? It is confusing. "Java SE" should not be split. I don't think it is a good delimiter because it is a valid character in runtime name.

src/azure-cli-core/azure/cli/core/profiles/_shared.py
calvinsID: My mistake, reinstalled packages and working as expected

src/azure-cli/azure/cli/command_modules/appservice/custom.py
calvinsID: Hybrid delimiter won't work. I've updated to remove space as a delimiter, and docs show use '|' or ':'

panchagnula (approved):
LGTM. Thanks for jumping on this fix so quickly from the time I assigned the issue to you :)

qwordy (approved):

PR #16530, opened 2021-01-14
[ACR] add --allow-trusted-services support

adewaleo:
**Description**

This PR adds the option `--allow-trusted-services` to `az acr create` and `az acr update`. By default ACR allows trusted azure services to access network restricted registries.

**Testing Guide**
```
az acr create -h 
az acr update -h
azdev test test_acr_with_public_network_access
```
**History Notes**

[ACR] az acr create / update: add --allow-trusted-services. This parameter determines whether trusted azure services are allowed to access network restricted registries. The default is to allow.

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acr/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/acr/_params.py
@@ -82,8 +82,10 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         with self.argument_context(scope, arg_group='Network Rule') as c:
             c.argument('default_action', arg_type=get_enum_type(DefaultAction),
                        help='Default action to apply when no rule matches. Only applicable to Premium SKU.')
-            c.argument('public_network_enabled', get_three_state_flag(), help="Allow public network access for the container registry. The Default is to allow public access.")
-            c.argument('allow_trusted_services', get_three_state_flag(), is_preview=True, help="Allow trusted Azure Services to access network restricted registries. For more information, please visit https://aka.ms/acr/trusted-services. The default is to allow.")
+            suffix = " The Default is to allow public access."  if "create" in scope else ""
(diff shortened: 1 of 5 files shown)


src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -82,7 +82,8 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         with self.argument_context(scope, arg_group='Network Rule') as c:
             c.argument('default_action', arg_type=get_enum_type(DefaultAction),
                        help='Default action to apply when no rule matches. Only applicable to Premium SKU.')
-            c.argument('public_network_enabled', get_three_state_flag(), help="Allow public network access for the container registry. The Default is allowed")
+            c.argument('public_network_enabled', get_three_state_flag(), help="Allow public network access for the container registry. The Default is to allow public access.")
+            c.argument('allow_trusted_services', get_three_state_flag(), is_preview=True, help="Allow trusted Azure Services to access network restricted registries. For more information, please visit https://aka.ms/acr/trusted-services. The default is to allow.")
```
fengzhou-msft, line 86: Help message says `The default is to allow`.

For `acr update`, there should be no default values for the options in update commands.
For `acr create`, I don't see the logic to default to allow. Is it handled on service side?

adewaleo: yes, this is the default on the service side. Should I remove this?

fengzhou-msft: You can keep `The default is to allow` for `acr create`, but should remove it for `acr update`. If `allow_trusted_services` is none in update command, then the setting will not be changed.

adewaleo: yes this makes sense. I see what you mean. Thanks for the clarification

fengzhou-msft (approved):

PR #16536, opened 2021-01-14
[AppConfig] BREAKING CHANGE: az appconfig feature filter add: Support adding JSON objects as feature filter parameter values

### LINKED ISSUE #16374: Feature manager filter parameters parsed incorrectly
**Describe the bug**
Hello, I have an issue with Feature Manager stuff (I know that it is in preview).
In particular it is command "az appconfig feature filter add" and "--filter-parameters" part.
Documentation said it is space separated pairs Key=Value but they work not as expected.
1. Any Value treated as string and wrapped into double quotes. So no chance to set any non string value.
2. Array parsed with error: KeyName=[arrayValues] results into 2 keys: 
"KeyName"="[arrayValues"
"]"=""
(closing bracket becomes new key and again array is a string now, moreover if there was an array of strings inner values have NO string quotes anymore)
3. If a value is an Object type then it fails the command entirely. (maybe the command doesn't like curly braces)

**To Reproduce**
Just call this command.
az appconfig feature filter add --feature "FlagName" --label "testEnv" --filter-name "FilterName" --filter-parameters PropertyName="UserName" IsOneOf=["user1","user2","user3","user4"] IntValueCheck=50 BoolValueCheck=true --yes

For object test something like this.
az appconfig feature filter add --feature "FlagName" --label "testEnv" --filter-name "FilterName" --filter-parameters ObjectValueCheck={"a": 50, "b": true}  --yes

**Actual behavior**
"parameters": {
     "PropertyName": "UserName",
     "IsOneOf": "[user1,user2,user3,user4",
     "]": "",
     "IntValueCheck": "50",
     "BoolValueCheck": "true"
    }

**Expected behavior**
"parameters": {
     "PropertyName": "UserName",
     "IsOneOf": ["user1","user2","user3","user4"],
     "IntValueCheck": 50,
     "BoolValueCheck": true
    }
Ability to pass an object should exist.
I would be grateful if you could allow this command to take a json as input.
I think it is the simplest and convenient way to take a json as it is. Because currently I'm parsing json to make space separated pairs which in turn should create the same json as I've parsed.

**Environment summary**
Azure Powershell Task in Azure DevOps 5.179.0, azure-cli 2.11.0, Az-Account 1.9.3

**Additional context**
I've tried it by installing azure cli on local PC, result is the same

avanigupta:
**Description**
Fix #16374 

- **Affected Command:**
`Az appconfig feature filter add -n TestAppConfig --feature Beta –filter-name TestFilter --filter-parameters name1=value1 name2=value2`

- **Current Behavior:**
Always treat the `value` in `--filter-parameters name=value` as string value.

- **New Behavior:**
Accept only _escaped JSON string_ as `value` in `--filter-parameters name=value`, and convert it to the correct data type accordingly.

**Testing Guide**

The following command should fail since the value is not a JSON escaped string:
`az appconfig feature filter add -n TestAppConfig --feature Beta --filter-name TestFilter --filter-parameters stringValue=SomeText`

This modified version should pass:
`az appconfig feature filter add -n TestAppConfig --feature Beta --filter-name TestFilter --filter-parameters stringValue=\"SomeText\"`

**History Notes**

[AppConfig] BREAKING CHANGE: az appconfig feature filter add:
- Support adding JSON objects as feature filter parameter values
- Accept only JSON escaped string as `value` in  `--filter-parameters name=value`
- `name` cannot be empty in `--filter-parameters name=value`
- Sample inputs for common data types:
    - String value: `--filter-parameters stringValue=\"This is some text\"`
    - Numeric value: `--filter-parameters numberValue=75`
    - Array value: `--filter-parameters arrayValue=[1,2,3]`
    - Boolean value: `--filter-parameters boolValue=true`
    - Object value: `--filter-parameters objectValue={\"Test\":{\"Level\":\"Debug\"}}`

---


avanigupta:
Hi @shenmuxiaosen, could you help review this change?

yungezz:
hi @qwordy could you pls help to review?

src/azure-cli/azure/cli/command_modules/appconfig/_validators.py
```diff
@@ -158,9 +154,23 @@ def validate_filter_parameter(string):
...
+                try:
+                    # Ensure that provided value of this filter parameter is valid JSON. Error out if value is invalid JSON.
+                    filter_param_value = json.loads(filter_param_value)
+                except ValueError:
+                    raise CLIError('Filter parameter value must be a JSON escaped string. "{}" is not a valid JSON object.'.format(filter_param_value))
+                result = (comps[0], filter_param_value)
+            else:
+                result = (string, '')
         else:
             logger.warning("Ignoring filter parameter '%s' because parameter name is empty.", string)
```
shenmuxiaosen, line 175: We can consider this is invalid if it is not in "key=value" format (preferred).  So no "=" or "=value" will fail explicitly. 
If you want to have some tolerance here, we should add warning log for both case.

avanigupta: Okay, I think it makes sense to error out on `=` or `=value` input since portal also does the same.

shenmuxiaosen: If the input format is not key=value, in line 172, we can also throw exception instead of treat input as key and value is empty string. 

---
In reply to: [558463404](https://github.com/Azure/azure-cli/pull/16536#discussion_r558463404) [](ancestors = 558463404)

avanigupta: If line 172 condition is met, it means that the input must me something like `--filter-parameters name`. This is something we can support and treat the value as empty string (same behavior as portal).

shenmuxiaosen (approved):
:shipit:

qwordy (approved):

qwordy:
It is not easy to input JSON.

MarshallMoorman:
When is this new feature expected to be released?  Was it included in today's release?

avanigupta:
Hi @MarshallMoorman, this feature did not make it in today's release. It will be included in the next release. You can refer to the [release notes](https://github.com/MicrosoftDocs/azure-docs-cli/blob/master/docs-ref-conceptual/release-notes-azure-cli.md) to see which CLI version will contain this release.

PR #16612, opened 2021-01-21
{Identity} Beta 2.18.0.1

jiasli:
**Description**

- Pass `kwargs` to credentials so that `logging_enable=True` can be passed to credentials to enable unredacted HTTP logging. 
- Hide `--environment` and `--tenant-access` temporarily.

### DIFF
--- a/src/azure-cli/setup.py
+++ b/src/azure-cli/setup.py
@@ -53,7 +53,7 @@ DEPENDENCIES = [
     'antlr4-python3-runtime~=4.7.2',
     'azure-appconfiguration~=1.1.1',
     'azure-batch~=10.0.0',
-    'azure-cli-core=={}.*'.format(".".join(VERSION.split(".")[:3])),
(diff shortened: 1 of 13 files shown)


fengzhou-msft (approved):

yonzhan:
Identity

src/azure-cli/setup.py
```diff
@@ -53,7 +53,7 @@
     'antlr4-python3-runtime~=4.7.2',
     'azure-appconfiguration~=1.1.1',
     'azure-batch~=10.0.0',
-    'azure-cli-core=={}.*'.format(".".join(VERSION.split(".")[:3])),
+    'azure-cli-core=={}'.format(VERSION),
```
jiasli, line 56: Pin to the full version `2.18.0.1`, otherwise `2.18.0` will also match.

fengzhou-msft (approved):

PR #16694, opened 2021-01-27
[Botservice] az bot create/update: Adding Encryption features --cmk-key-url and --encryption-off

munozemilio:
**Description**
In the bot service by default, Microsoft-managed encryption keys are used. For greater flexibility in managing keys or controlling access to your subscription, select Customer-Managed Keys (CMK), also known as Bring Your Own Key. At the moment it is only possible to enable CMK through the portal and ARM templates but there is no way to do it through the CLI. To allow customers to create and update Bot resources and enable/add CMK encryption the argument --cmk-key-url  has been added to az bot create and az bot update which enables CMK encryption and if passed and --encryption-off has been added to az bot update to disable if needed.

**Testing Guide**
Az bot update:
```

// Enables CMK encryption
az bot update --name <BotResourceName> --resource-group <ResourceGroupName> --cmk-key-url <KeyVaultURL>

// Disables CMK encryption. If --encryption-off and --cmk-key-url are bot passed MutuallyExclusiveArgumentError is thrown
az bot update --name <BotResourceName> --resource-group <ResourceGroupName> --encryption-off

```

Az bot create:
```

// Enables CMK encryption if --cmk-key-url passed
az bot create --appid <MicrosoftAccountID> --kind <Registration or WebApp> --name <BotResourceName> --resource-group <ResourceGroupName>  --cmk-key-url <KeyVaultURL>
                
```

**History Notes**

---


munozemilio:
@stevengum FYI

yonzhan:
Botservice

RustyDes (changes requested):

src/azure-cli/azure/cli/command_modules/botservice/custom.py
```diff
@@ -696,6 +703,13 @@ def update(client, resource_group_name, resource_name, endpoint=None, descriptio
     if app_insights_api_key:
         bot_props.developer_app_insights_api_key = app_insights_api_key
 
+    if cmek_key_vault_url is not None:
+        bot_props.cmek_key_vault_url = cmek_key_vault_url
+        bot_props.is_cmek_enabled = True
+
+    if encryption_off is not None:
```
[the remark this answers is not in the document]
stevengum: I raised the same thoughts with @munozemilio offline.

The code is saying that the following scenario is valid: a user provides a KeyVault URL but they might want the encryption to be off... Perhaps temporarily?

The module should instead log a warning to inform the user that they provided a KeyVault URL but also turned the encryption off. This would be a better UX for a user doing the above intentional scenario, otherwise the user needs to run to `az bot update` commands twice to mutate their bot resource to the desired state.

src/azure-cli/azure/cli/command_modules/botservice/webappv4.template.json
```diff
@@ -148,7 +156,7 @@
             ]
         },
         {
-            "apiVersion": "2017-12-01",
+            "apiVersion": "2018-07-12",
```
stevengum, line 159: I don't see a test for the `az bot create -k webapp` flow with the updates in the ARM template. This `apiVersion` doesn't match with when the KeyVault support was added, which was in 2020. (right @RustyDes?)

src/azure-cli/azure/cli/command_modules/botservice/tests/latest/test_bot_commands.py
```diff
@@ -809,6 +809,97 @@ def test_botservice_create_should_remove_invalid_char_from_name_when_webapp(self
...
+        self.cmd(
+            'az keyvault set-policy --name {keyvaultname} --key-permissions get wrapKey unwrapKey --object-id a5c54cac-9ebb-45c5-81e5-29dabf2eb5ee')
+
+        resultAzKey = self.cmd(
+            'az keyvault key create --vault-name {keyvaultname} --name {keyName} --protection software',
+            checks={
+                self.check('key', object)
+            })
+        resultAzKey = resultAzKey.get_output_in_json()
+        cmk_url = resultAzKey['key']['kid']
```
stevengum, line 836: @RustyDes or @swagatmishra2007, does Azure Bot Service do any validation on this KeyVault URL and key when they're provided?

If yes, does ABS look for uniqueness in the URL and/or key? Or can multiple bots use the same values. 

If uniqueness isn't a requirement, the KeyVault values could be provisioned once and reused for each test as a pre-arrangement function.

src/azure-cli/azure/cli/command_modules/botservice/custom.py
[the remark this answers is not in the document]
RustyDes: While the backend code should be fine with the operation, the likelihood that someone tries to do that seems much lower than the likelihood of someone doing that by mistake, so I consider having the error preferable.

src/azure-cli/azure/cli/command_modules/botservice/tests/latest/test_bot_commands.py
RustyDes: There's validation done, but it can be reuse by multiple bots (in fact I could easily imagine a company with more than one bot having a single key that they use for all their bots and possibly some other part of their systems such as their own storage - remember other azure features also support cmk, and it would make sense for the customer to use 1 key for all of them).

stevengum: Thanks @RustyDes! So @munozemilio it could speed up the tests to arrange a single key vault instance with a key than provisioning a new key vault resource for every test.

@yonzhan what do you think on this proposed test refactoring?

yonzhan: @fengzhou-msft could you help with this?

src/azure-cli/azure/cli/command_modules/botservice/_params.py
```diff
@@ -64,6 +64,7 @@ def load_arguments(self, _):
         c.argument('tags', arg_type=tags_type)
         c.argument('language', options_list=['--lang'], arg_type=get_enum_type(SUPPORTED_LANGUAGES), help='The language to be used to create the bot.', arg_group='Web App Bot Specific')
         c.argument('deploy_echo', options_list=['--echo'], arg_type=get_three_state_flag(), help='Deploy an Echo Bot template to the newly created v4 Web App Bot.', arg_group='V4 Bot Templates')
+        c.argument('cmek_key_vault_url', options_list=['--cmk-key-url'], help='The key vault url to enable Customer Managed Keys encryption')
```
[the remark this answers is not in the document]
munozemilio: Changed to '--cmk-key-vault-key-url', '--cmk'

src/azure-cli/azure/cli/command_modules/botservice/custom.py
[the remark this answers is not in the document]
munozemilio: Changed to throw MutuallyExclusiveArgumentError if both args are passed

src/azure-cli/azure/cli/command_modules/botservice/webappv4.template.json
munozemilio: This came from a sample Rusty provided. @RustyDes could you please confirm?

src/azure-cli/azure/cli/command_modules/botservice/tests/latest/test_bot_commands.py
munozemilio: @stevengum I did a refactor to create the Key Vault resource and key during the setup but started to get this error: Key not found. Ensure that the key exist (if it was deleted it may need to be undeleted).
This is not thrown if created within the same test. Given the deadline is almost here I will refactor that once I find a most efficient and successful way to achieve that.

stevengum: Understood, thanks for giving it a go @munozemilio!

src/azure-cli/azure/cli/command_modules/botservice/webappv4.template.json
stevengum: If @RustyDes signed off, then it works for me 👍🏼

stevengum (approved):
All of my concerns were addressed or can be implemented in a later pass, so this LGTM

src/azure-cli/azure/cli/command_modules/botservice/webappv4.template.json
RustyDes: @swagatmishra2007 I got the version from you so you should be the one signing off on that :)

stevengum:
This PR should also fix https://github.com/microsoft/botframework-cli/issues/570 via the version bump in the `azure-mgmt-botservice` SDK. 🎉

src/azure-cli/azure/cli/command_modules/botservice/webappv4.template.json
swagatmishra2007: version should be 2020-06-02

PR #16763, opened 2021-02-02
{Storage} Hornor user input and add logger infomration

Juliehzl:
**Description**

Fix ICM https://portal.microsofticm.com/imp/v3/incidents/details/224946991/home

In previous code, environment variable value for connection string will have high priority compared with account key which is inputted by user input, which will cause issue for customer.

With current change, user input will be the first one to use and if no input value, we will try to get from config.

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/storage/_validators.py
+++ b/src/azure-cli/azure/cli/command_modules/storage/_validators.py
@@ -146,7 +146,7 @@ def validate_client_parameters(cmd, namespace):
         return
 
     # When there is no input for credential, we will read environment variable
-    if not n.connection_string and not n.account_key:
+    if not n.connection_string and not n.account_key and not n.sas_token:
         n.connection_string = get_config_value(cmd, 'storage', 'connection_string', None)
 
     # if connection string supplied or in environment variables, extract account key and name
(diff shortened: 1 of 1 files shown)


src/azure-cli/azure/cli/command_modules/storage/_validators.py
```diff
@@ -144,7 +145,8 @@ def validate_client_parameters(cmd, namespace):
                            ' ,'.join(account_key_args))
         return
 
-    if not n.connection_string:
+    # When there is no input for credential, we will read environment variable
+    if not n.connection_string and not n.account_key:
```
evelyn-ys, line 149:
```suggestion
    if not n.connection_string and not n.account_key and not n.sas_token:
```
If `sas_token` is given, do we need to read from environment variable?

Juliehzl: goo point. Fixed in two places.

src/azure-cli/azure/cli/command_modules/storage/_validators.py
```diff
@@ -157,7 +159,7 @@ def validate_client_parameters(cmd, namespace):
     # otherwise, simply try to retrieve the remaining variables from environment variables
     if not n.account_name:
         n.account_name = get_config_value(cmd, 'storage', 'account', None)
-    if not n.account_key:
+    if not n.account_key and not n.sas_token:
         n.account_key = get_config_value(cmd, 'storage', 'key', None)
     if not n.sas_token:
```
evelyn-ys, line 164:
```suggestion
    if not n.account_key and not n.sas_token:
```
If `account_key` has been retrieved from enviroment variable, then we can omit retrieve `sas_token`, right? But this will make our code ugly😂

Juliehzl: we should not omit sas token although we could get account key from config value. Because user input is first priority.

evelyn-ys (approved):

PR #16845, opened 2021-02-08
[Core] Built in alias for edge build

evelyn-ys:
**Description**
Support alias in command level and argument level

**Testing Guide**
`az sa -h`
`az sa bsp -h`
`az tsi env -h`
`az tsi env create --type longterm`

---


src/azure-cli-core/azure/cli/core/commands/__init__.py
```diff
@@ -112,9 +112,31 @@ def _expand_file_prefix(arg):
...
+        args_transformed = ['az']
+        args_transformed.extend(find_alias_and_transform_args(args[1:], None))
+        return args_transformed
+    return find_alias_and_transform_args(args, None)
+
+
 def _pre_command_table_create(cli_ctx, args):
     cli_ctx.refresh_request_id()
-    return _expand_file_prefixed_files(args)
+    real_args = _process_args_with_built_in_alias(cli_ctx, args)
```
jiasli, line 136: I would recommend calling `_process_args_with_built_in_alias` with an event. If the current even `raise_event` can't satisfy our needs, we can define new event types.

src/azure-cli-core/azure/cli/core/commands/alias_util.py
```diff
@@ -0,0 +1,192 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+import shlex
+
+
+def find_alias_and_transform_args(args, root_alias_nodes):
```
jiasli, line 9: This file can be moved to a dedicated command module `alias`.

src/azure-cli-core/azure/cli/core/commands/built_in_alias
```diff
@@ -0,0 +1,38 @@
+[sa]
+command = storage account
```
jiasli, line 2: I prefer JSON for its extensibility and flexibility.

src/azure-cli-core/setup.py
```diff
@@ -49,6 +49,8 @@
     'colorama~=0.4.1',
     'humanfriendly>=4.7,<10.0',
     'jmespath',
+    'configobj',
```
jiasli, line 52: Python has its own [`configparser`](https://docs.python.org/3/library/configparser.html). We usually avoid using external dependencies whenever possible.

jiasli (approved):

src/azure-cli-core/setup.py
evelyn-ys: I import `configobj` here because `configparser` can't support sub section.
```
[section]
    [[sub section]]
```
But if we use JSON instead of INI, then this dependency can be removed.

yonzhan:
Core

PR #17028, opened 2021-02-22
[Storage] Upgrade storage api version to 2021-01-01

Juliehzl:
**Description**

To support new features, we have to use new package with api version 2021-01-01.

**Testing Guide**

**History Notes**

---


src/azure-cli/azure/cli/command_modules/storage/operations/account.py
```diff
@@ -538,8 +539,18 @@ def update_file_service_properties(cmd, instance, enable_delete_retention=None,
...
+        params['share_delete_retention_policy'] = instance.share_delete_retention_policy
 
-    return instance
+    # set protocol settings
+    if enable_smb_multichannel is not None:
+        instance.protocol_settings = cmd.get_models('ProtocolSettings')()
+        instance.protocol_settings.smb = cmd.get_models('SmbSetting')(
+            multichannel=cmd.get_models('Multichannel')(enabled=enable_smb_multichannel))
+    if instance.protocol_settings.smb.multichannel:
+        params['protocol_settings'] = instance.protocol_settings
```
evelyn-ys, line 551: Why another `if` instead of setting `params['protocal_settings']` within `if enable_smb_multichannel is not None`?

If the origin instance settings is `True` for `protocol_settings.smb.multichannel` and customer don't need to change, shall we still set `params['protocal_settings']` here?

yonzhan:
Storage

src/azure-cli/azure/cli/command_modules/storage/tests/latest/test_storage_account_scenarios.py
```diff
@@ -1063,11 +1066,26 @@ def test_storage_account_file_delete_retention_policy(self, resource_group, stor
             'cmd': 'storage account file-service-properties'
         })
         self.cmd('{cmd} show --account-name {sa} -g {rg}').assert_with_checks(
-            JMESPathCheck('shareDeleteRetentionPolicy', None))
+            JMESPathCheck('shareDeleteRetentionPolicy.enabled', True),
+            JMESPathCheck('shareDeleteRetentionPolicy.days', 7))
```
evelyn-ys, line 1070: Can we add this breaking change into history notes so that customers can be aware?

houk-ms (approved):
LGTM

src/azure-cli/azure/cli/command_modules/storage/_params.py
```diff
@@ -281,6 +281,9 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals, too-many-statem
     with self.argument_context('storage account update', resource_type=ResourceType.MGMT_STORAGE) as c:
         t_tls_version = self.get_models('MinimumTlsVersion', resource_type=ResourceType.MGMT_STORAGE)
         c.register_common_storage_account_options()
+        c.argument('sku', arg_type=get_enum_type(t_sku_name),
+                   help='Note that the SKU name cannot be updated to Standard_ZRS, Premium_LRS or Premium_ZRS, '
+                   'nor can accounts of those SKU names be updated to any other value')
```
fengzhou-msft, line 286: The allowed values still contain Standard_ZRS, Premium_LRS and Premium_ZRS in help message. They should be removed to avoid confusion.

src/azure-cli/azure/cli/command_modules/storage/operations/account.py
Juliehzl: This part code is copied from azure cli extension storage preview as you can see in 
https://github.com/Azure/azure-cli-extensions/blob/19b4eaa8f47bd0c4be5dcdd28acc39956b9bf5e2/src/storage-preview/azext_storage_preview/operations/account.py#L302-L308

I have such code because if previous `protocol_settings.smb.multichannel` is None, I will not add `protocol_setting` in params because for storage account type which doesn't support protocol setting will throw error in server side if we have schema `protocol_setting.smb` in params no matther it is None or non-None.
 For the storage account type which not support smb setting, SRP will still return with protocol_settings.smb although it is None.

And if it is true, resetting it with the same value will not cause any issue.

src/azure-cli/azure/cli/command_modules/storage/tests/latest/test_storage_account_scenarios.py
Juliehzl: sure

src/azure-cli/azure/cli/command_modules/storage/_params.py
Juliehzl: Actually you are right. There are previous sku type because they are still in model definition as you can see in
https://github.com/Azure/azure-sdk-for-python/blob/09521dfb517e0859ec961cae006fb728d787b565/sdk/storage/azure-mgmt-storage/azure/mgmt/storage/v2021_01_01/models/_models_py3.py#L5088.

Here is to align with SDK, so I haven't removed them. But I will check with service team and try to fix from swagger definition. 
For CLI side, I want to keep current deign with SDK first and refine with next SDK generation for the fix.

evelyn-ys (approved):
LGTM

aim-for-better (approved):
Approved from HDInsight part

msyyc (approved):

fengzhou-msft (approved):

PR #17038, opened 2021-02-22
[AppService] az webapp up: Save defaults for "runtime" and "os_type" parameter

calvinsID:
**Description**

For `az webapp up` command, defaults are saved for "group", "sku", "appserviceplan", "location", "web" parameters.
Recently added parameters "runtime" and "os_type", however defaults weren't being saved for these parameters

**Testing Guide**
- Run `az webapp up` with --runtime or --os-type parameter
- On later runs of `az webapp up`, the overrided runtime/os_type parameters aren't being used since they weren't saved in defaults.

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -3851,6 +3851,8 @@ def webapp_up(cmd, name=None, resource_group_name=None, plan=None, location=None
         cmd.cli_ctx.config.set_value('defaults', 'appserviceplan', plan)
         cmd.cli_ctx.config.set_value('defaults', 'location', loc)
         cmd.cli_ctx.config.set_value('defaults', 'web', name)
+        cmd.cli_ctx.config.set_value('defaults', 'runtime', runtime_version)
+        cmd.cli_ctx.config.set_value('defaults', 'os', os_name.lower())
     return create_json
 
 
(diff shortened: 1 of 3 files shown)


yonzhan:
AppService

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -3851,6 +3851,8 @@ def webapp_up(cmd, name=None, resource_group_name=None, plan=None, location=None
         cmd.cli_ctx.config.set_value('defaults', 'appserviceplan', plan)
         cmd.cli_ctx.config.set_value('defaults', 'location', loc)
         cmd.cli_ctx.config.set_value('defaults', 'web', name)
+        cmd.cli_ctx.config.set_value('defaults', 'runtime', runtime_version)
+        cmd.cli_ctx.config.set_value('defaults', 'os', os_name.lower())
```
panchagnula, line 3855: Not sure what the purpose of these are - since runtime & Os are not universal defaults in Azure ?

panchagnula (changes requested):
I am not sure what the purpose of this changes are?
if an app doesn't exist user should provide these values to override the defaults up makes?
if an app exists & command is used for re-deploymenr then we should just honor the values of the current app & try to override anything . Correct?

qwordy:
@calvinsID Is this PR still active? Otherwise, I will close it.

PR #17277, opened 2021-03-11
[Packaging] Add azure-identity dependency

jiasli:
**Description**
RDBMS command module has its own credential logic which uses Azure Identity's `ClientSecretCredential`(https://github.com/Azure/azure-cli/pull/17191#discussion_r591992455).

Install Azure Identity so that CI can pass.

https://dev.azure.com/azure-sdk/public/_build/results?buildId=776163&view=logs&j=36dd4138-4d53-5e46-00d9-e5cd9744be05&t=1cf3879c-0a42-5ccd-7693-7a4781d739d8&l=134

```
src/azure-cli/azure/cli/command_modules/rdbms/_client_factory.py:140:12: E0611: No name 'identity' in module 'azure' (no-name-in-module)
src/azure-cli/azure/cli/command_modules/rdbms/_client_factory.py:140:12: E0401: Unable to import 'azure.identity' (import-error)
```


src/azure-cli/requirements.py3.Darwin.txt
```diff
@@ -5,16 +5,19 @@ argcomplete==1.11.1
 asn1crypto==0.24.0
 azure-appconfiguration==1.1.1
 azure-batch==10.0.0
-azure-cli==2.20.0
 azure-cli-core==2.20.0
 azure-cli-telemetry==1.0.6
+azure-cli==2.20.0
```
jiasli, line 10: Sorted all lines alphabetically.

evelyn-ys (approved):

yonzhan:
Packaging

DaeunYim (approved):
Thanks a lot for taking care of this..!

fengzhou-msft (approved):

Juliehzl (approved):
If they do want to use azure-identity, I could  approve for the change. But I am considering if it could utilize existing msal package for the same thing and all service could find a consistent way, which is better to have.
Anyway, approved from my side.

jiasli:
> ... I am considering if it could utilize existing msal package for the same thing and all service could find a consistent way, which is better to have.

MSAL doesn't expose `get_token` API for Track 2 SDK's usage. In order to make Track 2 SDK working, one must implement `get_token` API by themeselves.

The consistent way is to use the one provided by core (https://github.com/Azure/azure-cli/pull/17191#discussion_r591221240) instead of developing a custom auth mechanism.

DaeunYim:
> > ... I am considering if it could utilize existing msal package for the same thing and all service could find a consistent way, which is better to have.
> 
> MSAL doesn't expose `get_token` API for Track 2 SDK's usage. In order to make Track 2 SDK working, one must implement `get_token` API by themeselves.
> 
> The consistent way is to use the one provided by core ([#17191 (comment)](https://github.com/Azure/azure-cli/pull/17191#discussion_r591221240)) instead of developing a custom auth mechanism.

RDBMS module uses get_mgmt_service_client for default auth but I think the predecessor also implemented a way to override the authentification using credential. I wanted to keep both ways in case users were using this approach....! Thank you so much!

PR #13102, opened 2020-04-21
[Core] fix logging file fd leaking

haroldrandom:
**Description**  
This PR mainly did is to To make CLI instance close logger file after it's invoked, specifically:
1. Fix the bug that would leak file handler and its corresponding fd (Fix https://github.com/Azure/azure-cli/issues/12882, Fix https://github.com/Azure/azure-cli/issues/10435)
2. Fix the edge bug that file logger would write to previous file handler like `test_feedback.py` because of cli never clean up only when it exists.
3. Bump knack to 0.7.0rc4 because need its new feature support

The original discussion list [here](https://github.com/Azure/azure-cli/pull/12971).

With the [PR of knack](https://github.com/microsoft/knack/pull/199/files) which make knack be aware of SystemExit and the timing to raise EVENT_CLI_POST_EXECUTE  is merged, the test should be fine.

**Testing Guide**  
Follow the guide in the previous [PR](https://github.com/Azure/azure-cli/pull/12971), it will show `(deleted)` files when you `lsof` to inspect the opened files of CLI.
With this fix, it won't show again.

**History Notes**  

[Component Name 1] BREAKING CHANGE: az command a: Make some customer-facing breaking change.  
[Component Name 2] az command b: Add some customer-facing feature.

---

This checklist is used to make sure that common guidelines for a pull request are followed.

- [ ] The PR title and description has followed the guideline in [Submitting Pull Requests](https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules#submitting-pull-requests).

- [ ] I adhere to the [Command Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md).

PR #17324, opened 2021-03-16
{Log} Adapt az_command_data_logger to Knack 0.8.0

jiasli:
Adapt to https://github.com/microsoft/knack/pull/228: [Log] `CLILogging.configure` returns as early as possible

## Symptom

`az feedback` no longer works with knack `dev` branch:

```
> azdev test feedback

src\azure-cli\azure\cli\command_modules\feedback\tests\latest\test_feedback.py:130: in _helper_test_log_contents_correct
    self.assertEqual(command_log_files[1].get_command_status(), "SUCCESS")
E   AssertionError: '' != 'SUCCESS'
E   + SUCCESS
```

## Cause

In `pytest` environment, `pytest` will configure the `root` logger to capture all logs. 

- The old knack always forcibly set `root` logger level to `DEBUG`.

- The new knack can detect that and it leaves `pytest` `root` logger as is (https://github.com/microsoft/knack/pull/228).

  In such case, we have to manually set `az_command_data_logger`'s level to `DEBUG`, otherwise it will honor `pytest` `root` logger's level and be `WARNING` by default (when `--log-level` is not set). This makes `az_command_data_logger`'s log be ignored. For more info on how Python logging works, see [`logging.Logger.getEffectiveLevel`](https://docs.python.org/3/library/logging.html#logging.Logger.getEffectiveLevel).

## Additional Context

- `az feedback` relies on `az_command_data_logger` (which are written to `~/.azure/commands/`).

- `az_command_data_logger` is written by `azure.cli.core.util.handle_exception`.

- During testing, `azure.cli.core.util.handle_exception` is mocked by `azure.cli.testsdk.patches.patch_main_exception_handler` so that the original `Exception` can be thrown and asserted.

- However, `azure.cli.testsdk.patches.patch_main_exception_handler` doesn't write to `az_command_data_logger`.

- In the oldest code (#8829), this is worked around by calling the leaked file handler after `execute` finishes.

- As #13102 fixed the leaked file handler, the old code no longer worked. To work around it again, #13102 manually delayed the cleanup for file handler.

- This PR refactors #13102 with a more elegant `mock.patch` on `azure.cli.core.util.handle_exception` again, so that no manual delay of file handler cleanup is necessary.

### DIFF
--- a/src/azure-cli-core/azure/cli/core/azclierror.py
+++ b/src/azure-cli-core/azure/cli/core/azclierror.py
@@ -70,16 +70,16 @@ class AzCLIError(CLIError):
             if self.exception_trace:
                 logger.exception(self.exception_trace)
 
-            # print recommendations to action
-            if self.recommendations:
-                for recommendation in self.recommendations:
-                    print(recommendation, file=sys.stderr)
+        # print recommendations to action
+        if self.recommendations:
+            for recommendation in self.recommendations:
+                print(recommendation, file=sys.stderr)
 
-            if self.aladdin_recommendations:
-                print('\nTRY THIS:', file=sys.stderr)
-                for recommendation, description in self.aladdin_recommendations:
-                    print_styled_text(recommendation, file=sys.stderr)
-                    print_styled_text(description, file=sys.stderr)
+        if self.aladdin_recommendations:
+            print('\nTRY THIS:', file=sys.stderr)
+            for recommendation, description in self.aladdin_recommendations:
+                print_styled_text(recommendation, file=sys.stderr)
+                print_styled_text(description, file=sys.stderr)
 
     def send_telemetry(self):
         telemetry.set_error_type(self.__class__.__name__)
--- a/src/azure-cli-core/azure/cli/core/util.py
+++ b/src/azure-cli-core/azure/cli/core/util.py
@@ -58,7 +58,6 @@ def handle_exception(ex):  # pylint: disable=too-many-locals, too-many-statement
     from jmespath.exceptions import JMESPathError
     from msrestazure.azure_exceptions import CloudError
     from msrest.exceptions import HttpOperationError, ValidationError, ClientRequestError
-    from azure.cli.core.azlogging import CommandLoggerContext
     from azure.common import AzureException
     from azure.core.exceptions import AzureError
     from requests.exceptions import SSLError, HTTPError
--- a/src/azure-cli-core/azure/cli/core/util.py
+++ b/src/azure-cli-core/azure/cli/core/util.py
@@ -69,89 +69,88 @@ def handle_exception(ex):  # pylint: disable=too-many-locals, too-many-statement
     # Print the traceback and exception message
     logger.debug(traceback.format_exc())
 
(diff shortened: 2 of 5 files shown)


src/azure-cli/azure/cli/command_modules/feedback/tests/latest/test_feedback.py
```diff
@@ -192,22 +192,18 @@ def _run_cmd(self, command, checks=None, expect_failure=False):
...
-        # close it manually because we unregister the deinit_cmd_metadata_logging
-        # callback from EVENT_CLI_POST_EXECUTE event
-        cli_ctx.logging.end_cmd_metadata_logging(result.exit_code)
+        def _handle_exception_with_log(ex, *args, **kwargs):
+            with CommandLoggerContext(logger):
+                logger.error(ex)
+            original_handle_exception(*args, **kwargs)
 
+        with mock.patch('azure.cli.core.util.handle_exception', _handle_exception_with_log):
+            result = execute(cli_ctx, command, expect_failure=expect_failure).assert_with_checks(checks)
```
jiasli, line 206: The old code registers `Logger` as a `Handler`. It works simply because both `Logger` and `Handler` expose `handle` method by coincidence. It is never guaranteed to work as such and may break any time the internal logic of `logging` module changes.

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -69,89 +69,88 @@ def handle_exception(ex):  # pylint: disable=too-many-locals, too-many-statement
     # Print the traceback and exception message
     logger.debug(traceback.format_exc())
 
-    with CommandLoggerContext(logger):
```
jiasli, line 72: `CommandLoggerContext` will be called again at:

https://github.com/Azure/azure-cli/blob/4560e9603d93560ededeaa209e807690d130ae9d/src/azure-cli-core/azure/cli/core/azclierror.py#L62-L65

Calling it here is redundant.

jiasli: Besides, no log is written here, so these lines shouldn't be surrounded by `CommandLoggerContext`.

src/azure-cli-core/azure/cli/core/azclierror.py
```diff
@@ -70,16 +70,16 @@ def print_error(self):
...
+        # print recommendations to action
+        if self.recommendations:
+            for recommendation in self.recommendations:
+                print(recommendation, file=sys.stderr)
+
+        if self.aladdin_recommendations:
+            print('\nTRY THIS:', file=sys.stderr)
+            for recommendation, description in self.aladdin_recommendations:
+                print_styled_text(recommendation, file=sys.stderr)
+                print_styled_text(description, file=sys.stderr)
```
jiasli, line 82: No log is written here. The scope of `CommandLoggerContext` should be as small as possible.

yonzhan:
Log

fengzhou-msft (approved):

houk-ms (approved):

evelyn-ys (approved):

Issue #16808, opened 2021-02-04
[Service Team Request] Adding subnet ID and IP address as eligible input for load balancer backend pool commands


erichrt:
**Resource Provider**

Networking

**Description of Feature or Work Requested**
We have added subnet ID and IP address as a way to specify backend address when creating or updating an IP-based backend pool of a load balancer. This is related to this feature: https://github.com/Azure/azure-cli/issues/12937 however, we will start allowing customers to specify by subnet ID in place of vnet ID. 

**Minimum API Version Required**

2020-11-01

**Swagger Link**

https://github.com/Azure/azure-rest-api-specs/tree/network-november-release/specification/network/resource-manager/Microsoft.Network/stable/2020-11-01

**Target Date**

3/15

yonzhan:
network

msyyc:
@erichrt When will '2020-11-01' be merged into 'master' of main repo?

Issue #12937, opened 2020-04-09
[Service Team Request] CLI support needed for New LB backend format


ninzavivek:
**Resource Provider**
Networking (NRP)

**Description of Feature or Work Requested**
CLI changes to support newly added Create/Update operation for backend pool and format.

**Minimum API Version Required**
2020-04-01

**Swagger Link**
Here is the PR link : https://github.com/Azure/azure-rest-api-specs/pull/8993

**Target Date**
May-end ( Manganese end).

ninzavivek:
@gmainar - FYI

MyronFanQiu:
@ninzavivek Hi, I want to confirm one thing. Before this issue and your PR, can customer create backend pool? I cannot do it in Azure CLI.

ninzavivek:
> @ninzavivek Hi, I want to confirm one thing. Before this issue and your PR, can customer create backend pool? I cannot do it in Azure CLI.

ninzavivek:
closed by mistake.

ninzavivek:
> @ninzavivek Hi, I want to confirm one thing. Before this issue and your PR, can customer create backend pool? I cannot do it in Azure CLI.

Yes, earlier it was done as part of whole LB put operation.

ninzavivek:
DEL operation support is also needed - https://github.com/Azure/azure-rest-api-specs/pull/9040

anavinahar:
Can we get an ETA for this?

MyronFanQiu:
@ninzavivek This new feature need 2020-04-01 api-version which has not been released yet. After the release of new sdk, we need to bump version first. Currently, we have another issue with higher priority to solve. If everything goes well, we might be able to catch S169(Official Release: 05/19/2020) release train. If we don't have enough bandwidth in this sprint, we should be able to catch S160, whose release date is 06/02/2020.

anavinahar:
Please confirm as we need to align on preview dates and GA dates accordingly. @MyronFanQiu

yonzhan:
We can set 06/02/2020 as release date.

erichrt:
Hi @yonzhan, @MyronFanQiu thanks for the update on release date. Do we already have a design of the CLI commands? 

I will set up a meeting so we can collaborate on the release of these commands as part of the feature preview

PR #17336, opened 2021-03-17
[Network] `az network lb address-pool` : Support subnet when creating or updating an IP-based backend pool of a load balancer.

msyyc:
**Description**

We have added subnet ID and IP address as a way to specify backend address when creating or updating an IP-based backend pool of a load balancer. This is related to this feature: #12937 however, we will start allowing customers to specify by subnet ID in place of vnet ID.

link issue : https://github.com/Azure/azure-cli/issues/16808
**Testing Guide**

**History Notes**

---


src/azure-cli/azure/cli/command_modules/network/_help.py
```diff
@@ -3337,12 +3337,14 @@
 short-summary: Create an address pool.
 parameters:
   - name: --backend-address
-    short-summary: Backend addresses information for backend address pool. If it's used, --vnet is also required.
+    short-summary: Backend addresses information for backend address pool. If it's used, --vnet is required or subnet is required.
     long-summary: |
-        Usage: --backend-address name=addr1 ip-address=10.0.0.1 --vnet MyVnet
+        Usage1: --backend-address name=addr1 ip-address=10.0.0.1 --vnet MyVnet
+        Usage2: --backend-address name=addr1 ip-address=10.0.0.1 subnet==/subscriptions/000/resourceGroups/MyRg/providers/Microsoft.Network/virtualNetworks/vnet/subnets/subnet1
```
kairu-ms, line 3343: 'Subnet' is the replace of 'vnet', so they should be in the same level. I think you should use '--subnet'.

kairu-ms: BTW, `subnet` is belong to a `vnet`. So if a user provide a vnet name and a subnet name then you can generate the subnet ID. So `--vnet` and `--subnet` should provide both in the case.

src/azure-cli/azure/cli/command_modules/network/custom.py
```diff
@@ -3509,12 +3512,22 @@ def _process_vnet_name_and_id(vnet):
...
+                                                        ip_address=addr['ip_address']) for addr in addresses_pool] if addresses_pool else None
+        except KeyError:
+            raise CLIError('Each backend address must have name, ip-address, (vnet | subnet) information.')
+    else:
+        try:
+            new_addresses = [LoadBalancerBackendAddress(name=addr['name'],
+                                                        virtual_network=VirtualNetwork(id=_process_vnet_name_and_id(addr['virtual_network'])),
+                                                        ip_address=addr['ip_address']) for addr in addresses_pool] if addresses_pool else None
+        except KeyError:
+            raise CLIError('Each backend address must have name, vnet and ip-address information.')
```
kairu-ms, line 3529: CLIError is not a clear error type.

yonzhan:
Network

src/azure-cli/azure/cli/command_modules/network/_actions.py
```diff
@@ -8,6 +8,7 @@
 import argparse
 from knack.util import CLIError
 from azure.cli.core.azclierror import UnrecognizedArgumentError
+from msrestazure.tools import is_valid_resource_id
```
jsntcy, line 11: use `from azure.mgmt.core.tools import is_valid_resource_id` instead as we'll deprecate `msrestazure` in the future.

src/azure-cli/azure/cli/command_modules/network/custom.py
```diff
@@ -3465,17 +3466,25 @@ def set_lb_frontend_ip_configuration(
...
+            resource_group=resource_group_name,
+            namespace='Microsoft.Network',
+            type='virtualNetworks',
+            name=vnet)
+    return vnet
+
+
+def _process_subnet_name_and_id(subnet, vnet, cmd, resource_group_name):
+    if subnet and not is_valid_resource_id(subnet):
+        subnet = _process_vnet_name_and_id(vnet, cmd, resource_group_name) + f'/subnets/{subnet}'
```
kairu-ms, line 3482: what if _process_vnet_name_and_id return None?

msyyc: There is check before here.

kairu-ms (approved):

PR #17301, opened 2021-03-13
{Feedback} Fix feedback minification

gukoff:
The previous implementation had 3 bugs:
- `IndexError` when only 1 line in `error_string` [example: #11889];
- infinite loop when repeatedly can't strip down `error_string`;
- ", ln" instead of ", in" in one matching expression that rendered it useless.

Also the logic itself was "magical": it removed 4 lines at a time, starting from a middle line but if possible (never successfully) from a line before that has an exception start in it...

This change should fix all these problems.

Issue #16386, opened 2020-12-29
`az feedback` truncates the call stack


jiasli:
**Describe the bug**
`az feedback` truncates the call stack of errors, like in #16373.

For example, if `az demo style` raises an `Exception` at

https://github.com/Azure/azure-cli/blob/821c01375e9f3aa513b221116c4f1f44b823d8ba/src/azure-cli/azure/cli/command_modules/util/custom.py#L175

```py
raise Exception()
```

The original call stack is 

```
CLIInternalError: The command failed with an unexpected error. Here is the traceback:

Traceback (most recent call last):
  File "D:\cli\env38\lib\site-packages\knack\cli.py", line 215, in invoke
    cmd_result = self.invocation.execute(args)
  File "d:\cli\azure-cli\src\azure-cli-core\azure\cli\core\commands\__init__.py", line 659, in execute
    raise ex
  File "d:\cli\azure-cli\src\azure-cli-core\azure\cli\core\commands\__init__.py", line 722, in _run_jobs_serially
    results.append(self._run_job(expanded_arg, cmd_copy))
  File "d:\cli\azure-cli\src\azure-cli-core\azure\cli\core\commands\__init__.py", line 715, in _run_job
    six.reraise(*sys.exc_info())
  File "D:\cli\env38\lib\site-packages\six.py", line 703, in reraise
    raise value
  File "d:\cli\azure-cli\src\azure-cli-core\azure\cli\core\commands\__init__.py", line 693, in _run_job
    result = cmd_copy(params)
  File "d:\cli\azure-cli\src\azure-cli-core\azure\cli\core\commands\__init__.py", line 330, in __call__
    return self.handler(*args, **kwargs)
  File "d:\cli\azure-cli\src\azure-cli-core\azure\cli\core\__init__.py", line 784, in default_command_handler
    return op(**command_args)
  File "d:\cli\azure-cli\src\azure-cli\azure\cli\command_modules\util\custom.py", line 176, in demo_style
    raise Exception()
Exception
```

But `az feedback` reports it as

```
Traceback (most recent call last):
env38\lib\site-packages\knack\cli.py, ln 215, in invoke
    cmd_result = self.invocation.execute(args)
azure\cli\core\commands\__init__.py, ln 659, in execute
    raise ex
azure\cli\core\commands\__init__.py, ln 722, in _run_jobs_serially
    results.append(self._run_job(expanded_arg, cmd_copy))
[...]

PR #17353, opened 2021-03-18
[Feedback] Do not minify feedback issue body

jiasli:
(Requires https://github.com/Azure/azure-cli/pull/17324)

## Context

`az feedback` opens an GitHub issue by filling the `body` query parameter. (See [About automation for issues and pull requests with query parameters](https://docs.github.com/en/github/managing-your-work-on-github/about-automation-for-issues-and-pull-requests-with-query-parameters))

```
https://github.com/azure/azure-cli/issues/new?body={}
```

Due to the URL length limitation on some legacy web browsers like IE, CLI minifies the issue body (especially call stack) to fit the URL length into the limitation:

https://github.com/Azure/azure-cli/blob/e3ac647675035d40877b43636843f4f737edcd42/src/azure-cli/azure/cli/command_modules/feedback/custom.py#L34

This sometimes truncates the call stack has caused us many problems while troubleshooting (https://github.com/Azure/azure-cli/issues/16386).

Modern browsers are already accepting URL length exceeding the old limitation (https://stackoverflow.com/a/417184), so the motivation to minify the `error_string` is less reasonable.

The minification process is also very error prone (#17301).

## Change

This PR

1. removes the minification logic
2. tells the user to manually open an issue if the URL exceeds the browser or GitHub's limitation

## Testing Guide

1. Place `raise Exception` after https://github.com/Azure/azure-cli/blob/f7fc07031900e05576ae91951038c7b42cc95c9f/src/azure-cli-core/azure/cli/core/adal_authentication.py#L118
2. Run `az group list` to trigger a very long error stack
3. Run `az feedback`


yonzhan:
Feedback

fengzhou-msft (approved):

houk-ms (approved):

PR #17366, opened 2021-03-19
{Misc} Drop `six` from core

jiasli:
**Description**

Because of the deprecation of Python 2 support (https://github.com/Azure/azure-cli/pull/11363), drop `six`.

### DIFF
--- a/src/azure-cli-core/azure/cli/core/commands/__init__.py
+++ b/src/azure-cli-core/azure/cli/core/commands/__init__.py
@@ -17,7 +17,6 @@ import sys
 import time
 import copy
 from importlib import import_module
-import six
 
 # pylint: disable=unused-import
 from azure.cli.core.commands.constants import (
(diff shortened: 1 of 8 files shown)


jiasli:
There are too many usages in command modules, defer to module owners to remove them.

yonzhan:
Misc

evelyn-ys (approved):

src/azure-cli-core/azure/cli/core/commands/__init__.py
```diff
@@ -713,7 +712,7 @@ def _run_job(self, expanded_arg, cmd_copy):
         except Exception as ex:  # pylint: disable=broad-except
             if cmd_copy.exception_handler:
                 return cmd_copy.exception_handler(ex)
-            six.reraise(*sys.exc_info())
+            raise
```
jsntcy, line 715: `sys.exc_info` can still be captured in exception body?

jiasli: https://docs.python.org/3/reference/simple_stmts.html#the-raise-statement

> If no expressions are present, `raise` re-raises the last exception that was active in the current scope.

jiasli: A stackoverflow post answering the difference of `raise` between Python 3 and Python 2: https://stackoverflow.com/questions/18188563/how-to-re-raise-an-exception-in-nested-try-except-blocks

jsntcy (approved):

msyyc (approved):

PR #17368, opened 2021-03-19
[Packaging] Support Python 3.9

jiasli:
## Description

Close #16611

It is time to onboard Python 3.9.

## Prerequisites

- [x] Bump underlying cryptography-related dependencies to newer versions (#16611)
    - [x] [bcrypt 3.2](https://pypi.org/project/bcrypt/3.2.0/): #17369
    - [x] [PyNaCl 1.4](https://pypi.org/project/PyNaCl/1.4.0/): #17369
    - [x] [cryptography 3.3.1](https://pypi.org/project/cryptography/3.3.1/): #15687  
- [x] `azdev setup`: show error if `pip` command fails (https://github.com/Azure/azure-cli-dev-tools/pull/281)
- [x] `azdev`: support Python 3.9 (https://github.com/Azure/azure-cli-dev-tools/pull/280)
- [x] Knack: support Python 3.9 (https://github.com/microsoft/knack/pull/229)
- [x] Bump Pylint to the latest 2.8.0 (https://github.com/Azure/azure-cli/pull/17861)
- [x] Use `azdev` and `pytest` for full test (#17383)

PR #17383, opened 2021-03-22
{CI} Use `azdev` and `pytest` for full test

jiasli:
## Description

As `nosttests` has actually been deprecated (the repo https://github.com/nose-devs/nose hasn't been updated since 2016), the full test should be moved to `azdev` and `pytest`.

## Known issues

- Coverage tools are installed by the old test pipeline, but is never used:
  https://github.com/Azure/azure-cli/blob/e7186c8160568291edb67169e379343c6312ab7a/scripts/ci/test_automation.sh#L23-L24
- Due to the lack of process isolation (https://github.com/pytest-dev/pytest/issues/6372), `azdev` calls `pytest` to run tests sequentially. This causes the pipeline running time to double compared to that of `nosttests`.


yonzhan:
CI

fengzhou-msft (approved):

azure-pipelines-full-tests.yml
```diff
@@ -13,144 +13,74 @@ pr:
       - '*'
 
 jobs:
-  - job: ExtractMetadata
-    displayName: Extract Metadata
+- job: AutomationTest
+  displayName: Automation Test (Profile Latest)
+  timeoutInMinutes: 120
+  pool:
+    vmImage: 'ubuntu-16.04'
```
qwordy, line 20: Is 16.04 too old? Consider 18.04 or 20.04

azure-pipelines-full-tests.yml
```diff
@@ -13,144 +13,74 @@ pr:
...
+  displayName: Automation Test (Profile 2020-09-01)
+  timeoutInMinutes: 120
+  pool:
+    vmImage: 'ubuntu-16.04'
+  strategy:
+    matrix:
+      Python36:
+        python.version: '3.6'
+      Python38:
+        python.version: '3.8'
```
qwordy, line 44: Where is 3.7?

.azure-pipelines/templates/automation_test.yml
```diff
@@ -1,6 +1,16 @@
+# Azure DevOps task for automation test using azdev and pytest
+
 parameters:
-  pythonVersion: ''
-  profile: ''
+- name: pythonVersion
+  type: string
+  default: ''
```
qwordy, line 6: What does '' mean? How about set default value 3.8?

azure-pipelines-full-tests.yml
jiasli: Will update in https://github.com/Azure/azure-cli/pull/17368 to keep in sync with main pipeline.

azure-pipelines-full-tests.yml
jiasli: We package with Python 3.6 and currently we support up to 3.8. @fengzhou-msft

.azure-pipelines/templates/automation_test.yml
jiasli: `''` means empty string. I copied it as-is. It also appears here:

qwordy (approved):

PR #17391, opened 2021-03-22
{Doc} Change rst to md

jiasli:
## Description
- Contents in `HISTORY.rst` are already using [Markdown syntax for inline code](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#code), like `` `az acr login` ``. The correct [RST syntax for inline code](https://docutils.sourceforge.io/docs/user/rst/quickref.html#inline-markup) should be double backticks, like ``` ``az acr login`` ```. Using Markdown syntax in RST will actually be rendered as "interpreted text" and highlighting gets lost: 
image
- The official document is also using Markdown: https://github.com/MicrosoftDocs/azure-docs-cli/blob/master/docs-ref-conceptual/release-notes-azure-cli.md. This will simplify the release pipeline.

## How it is done

Replace `++++++` with level 2 headings: 
```
Find: (.+)\n\++\n*
Replace: ## $1\n\n
```

Replace `**` with level 3 headings: 

```
Find: ^\*\*(.+)\*\*$
Replace: ### $1
```

### DIFF
--- a/src/azure-cli-core/HISTORY.md
+++ b/src/azure-cli-core/HISTORY.md
@@ -1,123 +1,120 @@
-.. :changelog:
+# Release History
 
-Release History
-===============
+## 2.21.0
 
-2.21.0
-++++++
 * Unify ADAL and MSAL error handler (#17072)
 * Add support for specifying minimum required version of an extension to work with the current core version (#16751)
 
-2.20.0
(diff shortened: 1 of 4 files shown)


yonzhan:
Doc

PR #17475, opened 2021-03-26
{Core} Fix `test_open_page_in_browser` on WSL

jiasli:
**Description**

`src/azure-cli-core/azure/cli/core/tests/test_util.py::TestUtils::test_open_page_in_browser` fails on WSL:

```py
_mock_self = <MagicMock name='open' spec='function' id='139689703216128'>, args = ('http://foo', 2), kwargs = {}
self = <MagicMock name='open' spec='function' id='139689703216128'>
msg = "Expected 'open' to be called once. Called 0 times."

    def assert_called_once_with(_mock_self, *args, **kwargs):
        """assert that the mock was called exactly once and that that call was
        with the specified arguments."""
        self = _mock_self
        if not self.call_count == 1:
            msg = ("Expected '%s' to be called once. Called %s times.%s"
                   % (self._mock_name or 'mock',
                      self.call_count,
                      self._calls_repr()))
>           raise AssertionError(msg)
E           AssertionError: Expected 'open' to be called once. Called 0 times.

env38/lib/python3.8/site-packages/mock/mock.py:925: AssertionError
```

Now fix it!

### DIFF
--- a/src/azure-cli-core/azure/cli/core/util.py
+++ b/src/azure-cli-core/azure/cli/core/util.py
@@ -643,7 +643,7 @@ def open_page_in_browser(url):
         try:
             # https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_powershell_exe
             # Ampersand (&) should be quoted
-            return subprocess.call(['powershell.exe', '-Command', 'Start-Process "{}"'.format(url)])
+            return subprocess.Popen(['powershell.exe', '-Command', 'Start-Process "{}"'.format(url)])
         except OSError:  # WSL might be too old  # FileNotFoundError introduced in Python 3
             pass
     elif platform_name == 'darwin':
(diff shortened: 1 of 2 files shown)


src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -643,7 +643,7 @@ def open_page_in_browser(url):
         try:
             # https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_powershell_exe
             # Ampersand (&) should be quoted
-            return subprocess.call(['powershell.exe', '-Command', 'Start-Process "{}"'.format(url)])
+            return subprocess.Popen(['powershell.exe', '-Command', 'Start-Process "{}"'.format(url)])
```
jiasli, line 646: `Start-Process` is async, so no need to use `call` to wait for the result.

yonzhan:
Core

evelyn-ys (approved):

PR #17476, opened 2021-03-26
[Synapse] Update role assignment/definition related cmdlets

sunsw1994:
**Description**

Update 'az synapse role' commands to support new role base access control feature.  

**Testing Guide**

az synapse role scope -h
az synapse role assignment -h
az synapse role definition -h

**History Notes**

[Synapse] az synapse role scope list:  List all scopes synapse supports.
[Synapse] az synapse role assignment create/list/delete: Adding --scope/--item-type/--item arguments to support manage role assignments based on scope.
[Synapse] az synapse role assignment create/list/delete: Adding --assignee-object-id argument,  it will bypass Graph API and uniquely determine principal object instead of deducing principal object using --assignee argument.
[Synapse] BREAKING CHANGE: az synapse role assignment create: Role names at old version are not allowed, Sql Admin, Apache Spark Admin, Workspace Admin
[Synapse] BREAKING CHANGE: az synapse role assignment create: When --assignee argument can't  uniquely determine the principal object, the command will raise error instead of adding a role assignment for the uncertain principal object. 

---


yonzhan:
Synapse

src/azure-cli/azure/cli/command_modules/synapse/operations/accesscontrol.py
```diff
@@ -4,55 +4,67 @@
...
-    object_id = _resolve_object_id(cmd, assignee, fallback_to_object_id=True)
+def list_role_assignments(cmd, workspace_name, role=None, assignee=None, assignee_object_id=None,
+                          scope=None, item=None, item_type=None):
+    if bool(assignee) and bool(assignee_object_id):
+        raise CLIError('usage error: --assignee STRING | --assignee-object-id GUID')
+
+    if bool(item) != bool(item_type):
+        raise CLIError('usage error: --item-type STRING --item STRING')
+
+    return _list_role_assignment(cmd, workspace_name, role, assignee or assignee_object_id,
```
zesluo, line 23: _list_role_assignments?

src/azure-cli/azure/cli/command_modules/synapse/operations/accesscontrol.py
```diff
@@ -4,55 +4,67 @@
...
 def get_role_assignment_by_id(cmd, workspace_name, role_assignment_id):
-    client = cf_synapse_client_accesscontrol_factory(cmd.cli_ctx, workspace_name)
+    client = cf_synapse_role_assignments(cmd.cli_ctx, workspace_name)
     return client.get_role_assignment_by_id(role_assignment_id)
 
 
 # Delete Synapse Role Assignment
-def delete_role_assignment(cmd, workspace_name, ids=None, assignee=None, role=None):
-    client = cf_synapse_client_accesscontrol_factory(cmd.cli_ctx, workspace_name)
+def delete_role_assignment(cmd, workspace_name, ids=None, assignee=None, assignee_object_id=None, role=None,
```
zesluo, line 48: delete_role_assignment_by_id?

src/azure-cli/azure/cli/command_modules/synapse/operations/accesscontrol.py
```diff
@@ -61,32 +73,96 @@ def delete_role_assignment(cmd, workspace_name, ids=None, assignee=None, role=No
...
+
+def _build_role_scope(workspace_name, scope, item, item_type):
+    if scope:
+        return scope
+
+    if item and item_type:
+        # workspaces/{workspaceName}/bigDataPools/{bigDataPoolName}
+        scope = "workspaces/" + workspace_name + "/" + item_type + "/" + item
+    else:
+        scope = "workspaces/" + workspace_name
```
zesluo, line 163: I just want to mention that these two may be not equal from my memory:  1.scope = "workspaces/" + workspace_name 2. scope = null

sunsw1994: Hi @zesluo  As far as my understand, the default scope is workspaces/{workspace_name}, we can discuss and confirm with @idear1203 .

Thanks for throwing it out.

src/azure-cli/azure/cli/command_modules/synapse/operations/accesscontrol.py
sunsw1994: Will fix it.

src/azure-cli/azure/cli/command_modules/synapse/operations/accesscontrol.py
sunsw1994: Cx can give a set of ids to delete not only one role assignment. Besides, user can delete role assignments by assignee or by role. 

delete_role_assignment_by_id is not friendly to cx.

src/azure-cli/azure/cli/command_modules/synapse/constant.py
```diff
@@ -12,6 +12,9 @@
 SPARK_DOTNET_UDFS_FOLDER_NAME = 'udfs'
 SPARK_SERVICE_ENDPOINT_API_VERSION = '2019-11-01-priview'
 AdministratorType = "activeDirectory"
+ITEM_NAME_MAPPING = {'bigDataPools': '{bigDataPoolName}', 'integrationRuntimes': '{integrationRuntimeName}',
+                     'linkedServices': '{linkedServiceName}', 'credentials': '{credentialName}',
+                     'scopePools': '{scopePoolName}'}
```
idear1203, line 17: >'scopePools': '{scopePoolName}' [](start = 21, length = 31)

let's remove scopePools for now.

src/azure-cli/azure/cli/command_modules/synapse/operations/accesscontrol.py
```diff
@@ -4,55 +4,67 @@
...
+from ..constant import ITEM_NAME_MAPPING
 
 
 # List Synapse Role Assignment
-def list_role_assignments(cmd, workspace_name, role=None, assignee=None):
-    # get role id
-    role_id = _resolve_role_id(cmd, role, workspace_name)
-    # get object_id
-    object_id = _resolve_object_id(cmd, assignee, fallback_to_object_id=True)
+def list_role_assignments(cmd, workspace_name, role=None, assignee=None, assignee_object_id=None,
```
idear1203, line 15: >assignee=None, assignee_object_id=None, [](start = 58, length = 39)

For what reason we have both `--assignee` and `--assignee-object-id`? Can we just combine them as one parameter?

idear1203 (approved):
:shipit:

src/azure-cli/azure/cli/command_modules/synapse/operations/accesscontrol.py
sunsw1994: keep same step with 'az role assignment', I also think `--assignee-object-id` can bypass `--assignee` potential issue. 
`--assignee`  parameter needs to invoke graph api which may meeting time delay issue.

idear1203: Make sense to me.

sunsw1994:
Hi @jsntcy  Could you help to review and merge this PR for Synapse RBAC commands. Thanks a lot.

src/azure-cli/azure/cli/command_modules/synapse/_help.py
```diff
@@ -901,15 +923,24 @@
   - name: Create a role assignment using service principal name.
     text: |-
         az synapse role assignment create --workspace-name testsynapseworkspace \\
-        --role "Sql Admin" --assignee sp_name
```
Juliehzl, line 904: does `Sql Admin` still work?

src/azure-cli/azure/cli/command_modules/synapse/constant.py
```diff
@@ -62,3 +64,17 @@ class SqlPoolConnectionClientAuthenticationType(str, Enum):
...
+    group = "Group"
+    service_principal = "ServicePrincipal"
+
+
+class ItemType(str, Enum):
+    bigDataPools = "bigDataPools"
+    scopePools = "scopePools"
+    integrationRuntimes = "integrationRuntimes"
+    credentials = "credentials"
+    linkedServices = "linkedServices"
```
Juliehzl, line 80: why not define in SDK?

src/azure-cli/azure/cli/command_modules/synapse/operations/accesscontrol.py
```diff
@@ -4,55 +4,67 @@
...
+                          scope=None, item=None, item_type=None):
+    if bool(assignee) and bool(assignee_object_id):
+        raise CLIError('usage error: --assignee STRING | --assignee-object-id GUID')
```
Juliehzl, line 18: Consider the following error type
https://github.com/Azure/azure-cli/blob/3bff786b9e337aa35bb4cc64a09e5600a35663f8/src/azure-cli-core/azure/cli/core/azclierror.py#L152-L154

src/azure-cli/azure/cli/command_modules/synapse/operations/accesscontrol.py
```diff
@@ -4,55 +4,67 @@
...
+
+    if bool(item) != bool(item_type):
```
Juliehzl, line 20: same as above

src/azure-cli/azure/cli/command_modules/synapse/_help.py
```diff
@@ -919,11 +950,11 @@
   - name: Delete role assignments by role and assignee.
     text: |-
         az synapse role assignment delete --workspace-name testsynapseworkspace \\
-        --role "Sql Admin" --assignee sp_name
+        --role "Synapse Administrator" --assignee sp_name
   - name: Delete role assignments by role id/name.
     text: |-
         az synapse role assignment delete --workspace-name testsynapseworkspace \\
-        --role "Sql Admin"
+        --role "Synapse Administrator"
```
evelyn-ys, line 957: Why no new examples for deleting role assignments by scope?

src/azure-cli/azure/cli/command_modules/synapse/commands.py
```diff
@@ -139,8 +139,12 @@ def get_custom_sdk(custom_module, client_factory):
...
         client_factory=None)
 
-    synapse_accesscontrol_sdk = CliCommandType(
-        operations_tmpl='azure.synapse.accesscontrol.operations#AccessControlClientOperationsMixin.{}',
+    synapse_role_assignment_sdk = CliCommandType(
+        operations_tmpl='azure.synapse.accesscontrol.operations#RoleAssignmentsOperations.{}',
+        client_factory=None)
+
+    synapse_role_definitions_sdk = CliCommandType(
+        operations_tmpl='azure.synapse.accesscontrol.operations#RoleAssignmentsOperations.{}',
```
evelyn-ys, line 147:
```suggestion
        operations_tmpl='azure.synapse.accesscontrol.operations#RoleDefinitionsOperations.{}',
```

src/azure-cli/azure/cli/command_modules/synapse/constant.py
```diff
@@ -62,3 +64,17 @@ class SqlPoolConnectionClientAuthenticationType(str, Enum):
...
+
+class PrincipalType(str, Enum):
+    user = "User"
+    group = "Group"
+    service_principal = "ServicePrincipal"
+
+
+class ItemType(str, Enum):
+    bigDataPools = "bigDataPools"
+    scopePools = "scopePools"
```
evelyn-ys, line 77: Why `scopePools` here since it has been removed from `ITEM_NAME_MAPPING`? Should it be removed too?

src/azure-cli/azure/cli/command_modules/synapse/operations/accesscontrol.py
evelyn-ys: I checked that only `az role assignment create` have additional `--assignee-object-id` with `--assignee-principal-type`. `list` and `delete` don't.

May I know why `az synapse role assignment list/delete` need `--assignee-object-id` without `--assignee-principal-type`?

src/azure-cli/azure/cli/command_modules/synapse/_help.py
sunsw1994: No, now Synapse is using new role name. 
Synapse SQL Administrator
Synapse Apache Spark Administrator
...

https://docs.microsoft.com/en-us/azure/synapse-analytics/security/synapse-workspace-synapse-rbac-roles

src/azure-cli/azure/cli/command_modules/synapse/commands.py
```diff
@@ -375,18 +379,22 @@ def get_custom_sdk(custom_module, client_factory):
...
         g.custom_command('delete', 'delete_role_assignment', confirmation=True)
 
-    with self.command_group('synapse role definition', synapse_accesscontrol_sdk,
+    with self.command_group('synapse role definition', synapse_role_definitions_sdk,
                             custom_command_type=get_custom_sdk('accesscontrol', None)) as g:
         g.custom_command('list', 'list_role_definitions')
         g.custom_show_command('show', 'get_role_definition')
 
+    with self.command_group('synapse role scope', synapse_role_definitions_sdk,
+                            custom_command_type=get_custom_sdk('accesscontrol', None)) as g:
```
Juliehzl, line 395: You could declare your client factory here and don't need to define in every fuction.

src/azure-cli/azure/cli/command_modules/synapse/commands.py
```diff
@@ -139,8 +139,12 @@ def get_custom_sdk(custom_module, client_factory):
         operations_tmpl='azure.synapse.spark.operations#SparkBatchOperations.{}',
         client_factory=None)
 
-    synapse_accesscontrol_sdk = CliCommandType(
-        operations_tmpl='azure.synapse.accesscontrol.operations#AccessControlClientOperationsMixin.{}',
+    synapse_role_assignment_sdk = CliCommandType(
+        operations_tmpl='azure.synapse.accesscontrol.operations#RoleAssignmentsOperations.{}',
+        client_factory=None)
```
Juliehzl, line 144: why your client factory is None?

src/azure-cli/azure/cli/command_modules/synapse/_help.py
Juliehzl: In this way, it is breaking change to existing customers and we should try to avoid such breaking. 
To not break customers, you should try to make previous definition work and add deprecation info to redirect to new definition.

src/azure-cli/azure/cli/command_modules/synapse/constant.py
sunsw1994: @Juliehzl  I think this not belong to SDK's scope, we define these item type for CLI using friendly. User can pass item type and CLI code will use this parameter to generate `scope` which is defined in SDK and received by service. 

`PrincipalType` and `ItemType` 's purpose is only to limit the user's input

@idear1203  Do we have a plan to move it to SDK?

src/azure-cli/azure/cli/command_modules/synapse/_help.py
idear1203: Can we just allow this breaking change considering that the module is still in preview? Now Synapse moves to new names, we don't want to allow the old names.

src/azure-cli/azure/cli/command_modules/synapse/constant.py
sunsw1994: Thanks for catching that, will remove it.

src/azure-cli/azure/cli/command_modules/synapse/operations/accesscontrol.py
```diff
@@ -4,55 +4,67 @@
...
+                           resolve_assignee=True, item=None, item_type=None):
+    """Prepare scope, role ID and resolve object ID from Graph API."""
+    if any([scope, item, item_type]):
+        scope = _build_role_scope(workspace_name, scope, item, item_type)
+    role_id = _resolve_role_id(cmd, role, workspace_name)
+    object_id = _resolve_object_id(cmd, assignee, fallback_to_object_id=True) if resolve_assignee else assignee
+    client = cf_synapse_role_assignments(cmd.cli_ctx, workspace_name)
+    role_assignments = client.list_role_assignments(role_id, object_id, scope).value
+    if scope:
+        role_assignments = [x for x in role_assignments if x.scope == scope]
```
evelyn-ys, line 37: Why do we need to filter the results by scope again since `client.list_role_assignments` already take `scope` parameter?

evelyn-ys: Oh I saw the previous comments and got the reason. Better to keep the comments~

src/azure-cli/azure/cli/command_modules/synapse/operations/accesscontrol.py
```diff
@@ -4,55 +4,67 @@
...
+def _list_role_assignments(cmd, workspace_name, role=None, assignee=None, scope=None,
+                           resolve_assignee=True, item=None, item_type=None):
+    """Prepare scope, role ID and resolve object ID from Graph API."""
+    if any([scope, item, item_type]):
+        scope = _build_role_scope(workspace_name, scope, item, item_type)
+    role_id = _resolve_role_id(cmd, role, workspace_name)
+    object_id = _resolve_object_id(cmd, assignee, fallback_to_object_id=True) if resolve_assignee else assignee
+    client = cf_synapse_role_assignments(cmd.cli_ctx, workspace_name)
+    role_assignments = client.list_role_assignments(role_id, object_id, scope).value
+    if scope:
```
evelyn-ys, line 36:
```suggestion
    if scope and role_assignments:
```

src/azure-cli/azure/cli/command_modules/synapse/operations/accesscontrol.py
```diff
@@ -61,32 +73,96 @@ def delete_role_assignment(cmd, workspace_name, ids=None, assignee=None, role=No
...
+    if bool(item) != bool(item_type):
+        raise CLIError('usage error: --item-type STRING --item STRING')
 
+    try:
+        return _create_role_assignment(cmd, workspace_name, role, assignee or assignee_object_id, scope, item,
+                                       item_type, resolve_assignee=(not assignee_object_id),
+                                       assignee_principal_type=assignee_principal_type, assignment_id=assignment_id)
+    except Exception as ex:  # pylint: disable=broad-except
+        if _error_caused_by_role_assignment_exists(ex):  # for idempotent
+            return list_role_assignments(cmd, assignee, role, scope)
```
evelyn-ys, line 97:
```suggestion
            return list_role_assignments(cmd, workspace_name, role=role, assignee=assignee, assignee_object_id=assignee_object_id, scope=scope, item=item, item_type=item_type)
```

src/azure-cli/azure/cli/command_modules/synapse/operations/accesscontrol.py
idear1203: Good catch. I think we should remove the logic, just return raw RP response. @sunsw1994

src/azure-cli/azure/cli/command_modules/synapse/commands.py
sunsw1994: Fixed.

src/azure-cli/azure/cli/command_modules/synapse/_help.py
sunsw1994: Will add it.

src/azure-cli/azure/cli/command_modules/synapse/operations/accesscontrol.py
sunsw1994: Fixed

src/azure-cli/azure/cli/command_modules/synapse/commands.py
sunsw1994: That is a dataplane command, so we need to create client using workspace name which is provided by user from parameter. So the client will be created when CLI received parameter instead of initialization at here. 

For the management commands, client factory is not None.

src/azure-cli/azure/cli/command_modules/synapse/operations/accesscontrol.py
sunsw1994: The server side has not supported list assignments by scope. `scope` parameter will be ignored by server, so I add filter at CLI side.

src/azure-cli/azure/cli/command_modules/synapse/operations/accesscontrol.py
sunsw1994: Fixed

src/azure-cli/azure/cli/command_modules/synapse/operations/accesscontrol.py
sunsw1994: Fixed

src/azure-cli/azure/cli/command_modules/synapse/constant.py
idear1203: Currently we don't have such definitions in SDK. This is a good suggestion.

src/azure-cli/azure/cli/command_modules/synapse/operations/accesscontrol.py
sunsw1994: 1. `--assignee-principal-type` is an optional parameter for  `az role assignment create` , and its code following `az role assignment` 
https://github.com/Azure/azure-cli/blob/50ccb24528eb36529d1432808f3ff385cc4c04aa/src/azure-cli/azure/cli/command_modules/role/custom.py#L146

list and delete api don't support `--assignee-principal-type` parameter

2. `--assignee-object-id` bypass Graph API, it can be sent to server directly without transforming  `assignee` to its object_id. Server only accepted object id in GUID format. `--assignee-principal-type` is not required. 

3. Cx can still using `--assignee 00000000-0000-0000-0000-000000000000`, adding  `--assignee-object-id`  just provide a way for cx to bypass Graph API invocation in case of insufficient privileges and provide assignee more accurately

evelyn-ys (approved):

src/azure-cli/azure/cli/command_modules/synapse/_help.py
sunsw1994: Adding breaking change description at History Notes

[Synapse] BREAKING CHANGE: az synapse role assignment create: Role names at old version are not allowed, Sql Admin, Apache Spark Admin, Workspace Admin

src/azure-cli/azure/cli/command_modules/synapse/operations/accesscontrol.py
```diff
@@ -119,6 +127,11 @@ def _resolve_object_id(cmd, assignee, fallback_to_object_id=False):
                            "If the assignee is a principal id, make sure the corresponding principal is created "
                            "with 'az ad sp create --id {assignee}'.".format(assignee=assignee))
 
+        if len(result) > 1:
```
sunsw1994, line 130: Add breaking change description at History Notes:

[Synapse] BREAKING CHANGE: az synapse role assignment create: When --assignee can't uniquely determine the principal object, the command will raise error instead of adding a role assignment for the uncertain principal object.

sunsw1994:
@evelyn-ys @Juliehzl  Thanks for your comments, these are helpful! I have fixed them and updated history notes with known breaking change information, could you help to review it and merge this PR. Hope we can catch up the latest release.  Thanks a lot！

src/azure-cli/azure/cli/command_modules/synapse/_help.py
Juliehzl: I see. According to the telemetry, there is no much usage for this command as it is in preview. So we could accept the breaking change this time. Please note that we need to avoid breaking change as possible as we can in future design.

src/azure-cli/azure/cli/command_modules/synapse/constant.py
Juliehzl: I see.

src/azure-cli/azure/cli/command_modules/synapse/commands.py
Juliehzl: Actually it doesn't matter because client factory will take effect during command execution where user parameter is already loaded. 
For `-h`, we are using operation_tmpl to get parameter list. 
You could see `storage blob` related command, which is data plane command and need user input to create a client factory. 
It is nice to have if you can refactor your design but as we are close to release date. I could accept current code now.

Juliehzl (approved):

PR #17479, opened 2021-03-26
[AppService] Fixes #17219: Fix ssl bind bug

### LINKED ISSUE #17219: webapp: Hostname does not exist error on ssl bind
## Describe the bug
Starting in 2.20.0, this started throwing errors. (was working in <=2.19). Note that this is for a situation where the app service & app service plan are in different resource groups. 

What is happening is that we are trying to bind an ssl cert to app service with cname of ABC, but instead we get an error that an unrelated hostname (XYZ) does not exist (It actually does exist, making this a doubly-fun error).

It looks like something in `az webapp config ssl bind` is using the *last* ssl cert in the list rather than the one that corresponds to the given app service / cert thumprint. (If I run `az webapp config ssl list`, the hostname XYZ that appears in the error is the last in the list.)

**Command Name**
`az webapp config ssl bind`

**Errors:**
```
Hostname 'XYZ.domain.com' does not exist.
```

## To Reproduce:

Here is how the certificate thumprint is selected:

APP_FQDN=ABC.domain.com

`THUMBPRINT=$(az webapp config ssl list -g $PLAN_RESOURCE_GROUP --query "[?name=='$APP_FQDN'].thumbprint | [0]")`

and here is the command that errors:
`az webapp config ssl bind --resource-group $RESOURCE_GROUP --name $APP_SERVICE --certificate-thumbprint $THUMBPRINT --ssl-type SNI`

Same results if the `--ids` arg is used instead.

## Expected Behavior
Was working in 2.19

## Environment Summary

Note, this is also occurring in multiple Azure DevOps pipelines (just started recently)

```
macOS-10.13.6-x86_64-i386-64bit
Python 3.8.8
Installer: HOMEBREW

azure-cli 2.20.0

Extensions:
front-door 1.0.11
aks-preview 0.5.4
webapp 0.3.1
application-insights 0.1.13

```

calvinsID:
**Description**
Fix `az webapp config ssl bind` using the wrong ssl cert to bind

**Testing Guide**
If app has multiple ssl certs, `az webapp config ssl bind` uses the wrong one

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -2584,14 +2584,14 @@ def _update_ssl_binding(cmd, resource_group_name, name, certificate_thumbprint, 
             if webapp_cert.thumbprint == certificate_thumbprint:
                 found_cert = webapp_cert
     if found_cert:
-        if len(webapp_cert.host_names) == 1 and not webapp_cert.host_names[0].startswith('*'):
+        if len(found_cert.host_names) == 1 and not found_cert.host_names[0].startswith('*'):
             return _update_host_name_ssl_state(cmd, resource_group_name, name, webapp,
-                                               webapp_cert.host_names[0], ssl_type,
+                                               found_cert.host_names[0], ssl_type,
                                                certificate_thumbprint, slot)
 
         query_result = list_hostnames(cmd, resource_group_name, name, slot)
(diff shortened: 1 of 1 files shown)


panchagnula:
@calvinsID thanks for taking care of this quickly. Do we have or can add a test case to catch this scenario?

panchagnula (approved):
LGTM! Just the one question on test case.

qwordy (approved):

PR #17509, opened 2021-03-30
{Core} Quick patch for `managed_by_tenants` missing from response

jiasli:
## ⚠ WARNING

This PR is a **workaround** for an ongoing ARM incident. The ARM service team is investigating with highest priority. We will close this PR once ARM service is fixed.

## Symptom

`az login` fails with

```
Invalid profile is used for cloud 'AzureCloud'. To configure the cloud profile, 
run `az cloud set --name AzureCloud --profile <profile>(e.g. 2019-03-01-hybrid)`. 
For more information about using Azure CLI with Azure Stack, see 
https://docs.microsoft.com/azure-stack/user/azure-stack-version-profiles-azurecli2
```

or (Azure CLI versions smaller than `2.15.0`)

```
azure/cli/core/_profile.py, ln 276, in _normalize_properties
    subscription_dict[_MANAGED_BY_TENANTS] = [{_TENANT_ID: t.tenant_id} for t in s.managed_by_tenants]
TypeError: 'NoneType' object is not iterable
```

## Root Cause Analysis

1. Azure CLI started using [Subscriptions - List](https://docs.microsoft.com/en-us/rest/api/resources/subscriptions/list) API `/subscriptions?api-version=2019-06-01` since Feb 2020 for public cloud `AzureCloud` (#11886). This call is made during `az login` to fetch the subscriptions.

2. Azure Stack returns the response for `/subscriptions?api-version=2019-06-01` with older API version `2016-06-01` format without `managedByTenants` property (https://github.com/Azure/azure-rest-api-specs/issues/9567).

3. Az CLI checks this explicitly and shows the above error message to the user implying "Invalid profile is used for Azure Stack cloud and the profile is set to `latest`, you need to change the API profile to something like `2019-03-01-hybrid`" (#15550).

4. Due to the ARM incident, now in public cloud `AzureCloud`, the `/subscriptions` API's response omits `managedByTenants` in some regions. As a consequence, the lack of `managedByTenants`    
   1. breaks [Azure Lighthouse](https://docs.microsoft.com/en-us/azure/lighthouse/) functionality
   2. makes Azure CLI think the Azure Stack issue (https://github.com/Azure/azure-rest-api-specs/issues/9567) is hit, so the user is presented with the above error message implying you are logged in to Azure Stack. This causes `az login` to fail.

## Changes in this PR

This PR provides a quick patch which bypasses the check on `managedByTenants` and thus bypasses the ARM service incident.

## Apply the quick patch

[...]

PR #17526, opened 2021-03-31
{Core} Hotfix: Remove the check on `managed_by_tenants`

jiasli:
## Description

Hotfix for #17509

1. When `latest` or `2019-06-01` is used for Azure Stack `/subscription`, the REST response doesn't have `managedByTenants`. However, because the REST spec defines it, the `Subscription` object returned by SDK will have `managed_by_tenants` as `None`. In such case, the behavior of this PR differs from the patch provided in #17509:
    - #17509: CLI doesn't return `managedByTenants`, following the **actual REST response**
    - This PR: CLI returns `managedByTenants`, but sets it to `null`, following the **REST spec**

2. When `2019-03-01-hybrid` or `2016-06-01` is used for Azure Stack `/subscription`, the REST response doesn't have `managedByTenants`. Because the REST spec doesn't define it, the `Subscription` object returned by SDK doesn't have `managed_by_tenants`, so CLI ignores it. This is the same as #17509.

## Additional Context

1. When SDK object has `managed_by_tenants` as `None`, CLI can't distinguish between
   - the REST API doesn't return `managedByTenants`
   - the REST API returns `managedByTenants` as `null` (Luckily, this has never been observed.)

2. By disabling this check in the gatekeeper `az login`, CLI may fail even deeper in other commands because the profile / API version is incorrect after all. The handling of unsupported API will be more difficult, but Azure Stack team confirms this is acceptable.


fengzhou-msft (approved):

yonzhan:
Core

yonzhan (approved):
LGTM

bganapa (approved):

evelyn-ys (approved):

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -1376,3 +1364,19 @@ def _get_authorization_code(resource, authority_url):
...
+    Transforms properties from Subscriptions - List 2019-06-01 and later to the subscription dict.
+
+    :param s: subscription object
+    :param s_dict: subscription dict
+    """
+    if hasattr(s, 'home_tenant_id'):
+        s_dict[_HOME_TENANT_ID] = s.home_tenant_id
+    if hasattr(s, 'managed_by_tenants'):
+        if s.managed_by_tenants is None:
+            s_dict[_MANAGED_BY_TENANTS] = None
```
jiasli, line 1380: If ARM returns `"managedByTenants": null`, Azure CLI output will also have `"managedByTenants": null`. This behavior doesn't comply with swagger either as `null` is not an array, so users assuming `managedByTenants` to be an array may also face failure.

**Actually, there is no way to develop omnipotent failure recovery for all kinds of REST API mistakes. Making service comply with the swagger is the best and ultimate solution.**

In this case, 

1. `managedByTenants` property MUST exist
2. `managedByTenants` MUST be an array, not `null`

https://learn.microsoft.com/en-us/rest/api/resources/subscriptions/list?tabs=HTTP

image

PR #17565, opened 2021-04-05
[ARM]Support policy exemption

robga:
**Description**

1. Upgrade azure-mgmt-resource SDK version to 12.1.0
2. Add policy exemption commands
3. Add unit test for policy exemption commands
4. "sku" property is removed from Policy assignment properties in the SDK.
4. Since the SDK version is increased, re-recorded some tests

**Testing Guide**

**History Notes**

---


yonzhan:
ARM

robga:
@zhoxing-ms 
The linter threw an error. 
 FAIL - HIGH severity: faulty_help_example_rule
    Help-Entry: `policy exemption update` - The following example entry indices do not include the command: 1 | 2
I didn't follow what that means. Could you please point it out what I should change?

robga:
@zhoxing-ms , for the only test failure left. 
FAILED src/azure-cli/azure/cli/command_modules/sql/tests/latest/test_sql_commands.py::SqlManagedDatabaseLogReplayScenarionTest::test_sql_midb_logreplay_mgmt
The test seems is setup for the owner's environment. The policy assignment name is hardcoded. I don't have their environment to record the test again. Please advice.

robga:
@zhoxing-ms , I should remove the  'sku' parameter for policy assignment. Is that a breaking change? How should I remove it without breaking any existing usage? 
It's line 190 in the file (src/azure-cli/azure/cli/command_modules/resource/_params.py )        
`c.argument('sku', options_list=['--sku', '-s'], help='policy sku.', arg_type=get_enum_type(['free', 'standard']))`

zhoxing-ms:
@robga Hi, since we plan to complete the migration of track 2 for resource module this month, we will upgrade the SDK version of `azure-mgmt-resource` to `16.0.0` (policy exemption is also included in this SDK version)
Because this PR will have some conflicts with the migration of track 2, so may I ask what is the expected release time for this requirement? *(If it is not an urgent requirement, maybe we can consider merging this PR after the migration of track 2.)*

robga:
Hi, @zhoxing-ms , I want to merged it in ASAP. I hope to catch the 4/7 deadline for this PR.

zhoxing-ms:
> Hi, @zhoxing-ms , I want to merged it in ASAP. I hope to catch the 4/7 deadline for this PR.
> I will really appreciate if you can let my PR merge first. I think you probably much more familiar this project than I do.
> The merge could take me much more time.

OK, we will release the PR in this sprint (04/13/2021)

src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -1760,6 +1760,79 @@
...
+    text: |
+        az policy exemption create -n exemptTestVM \\
+            --exemption-category "Mitigated" \\
+            --resource-group "myResourceGroup"
+  - name: Update a policy exemption in a management group.
+    text: |
+        az policy exemption create -n exemptTestVM \\
+            --exemption-category "Mitigated" \\
+            --scope "/providers/Microsoft.Management/managementGroups/myMG"
+"""
```
zhoxing-ms, line 1834:
```suggestion
  - name: Update a policy exemption in the resource group.
    text: |
        az policy exemption update -n exemptTestVM \\
            --exemption-category "Mitigated" \\
            --resource-group "myResourceGroup"
  - name: Update a policy exemption in a management group.
    text: |
        az policy exemption update -n exemptTestVM \\
            --exemption-category "Mitigated" \\
            --scope "/providers/Microsoft.Management/managementGroups/myMG"
"""
```

zhoxing-ms: > The linter threw an error.
FAIL - HIGH severity: faulty_help_example_rule
Help-Entry: policy exemption update - The following example entry indices do not include the command: 1 | 2
I didn't follow what that means. Could you please point it out what I should change?

@robga This problem is caused by the inconsistency between the command sample and the command. Please fix those samples of this command

zhoxing-ms:
> for the only test failure left.
FAILED src/azure-cli/azure/cli/command_modules/sql/tests/latest/test_sql_commands.py::SqlManagedDatabaseLogReplayScenarionTest::test_sql_midb_logreplay_mgmt
The test seems is setup for the owner's environment. The policy assignment name is hardcoded. I don't have their environment to record the test again. Please advice.

The following error occurred when executing command `az policy assignment show -n {policy_name}` in the test:
```
msrestazure.azure_exceptions.CloudError: Azure Error: PolicyAssignmentNotFound
Message: The policy assignment 'SDOStdPolicyNetwork' is not found.
```
Whether this test needs to add creation logic for policy assignment before this command or use other solutions, @Juliehzl please help to have a look and give some advice~

zhoxing-ms:
> I should remove the 'sku' parameter for policy assignment. Is that a breaking change? How should I remove it without breaking any existing usage?
It's line 190 in the file (src/azure-cli/azure/cli/command_modules/resource/_params.py )
c.argument('sku', options_list=['--sku', '-s'], help='policy sku.', arg_type=get_enum_type(['free', 'standard']))

Yes, it's a breaking change, because deleting parameter `sku` directly may affect the user's existing script.

You can add an deprecate annotation for this parameter to let the user know in advance that this parameter will be deprecated. 
When no user uses this parameter any more (user usage can be evaluated according to Telemetry), then remove the parameter completely.
Please refer to document [deprecating-commands-and-arguments](https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#deprecating-commands-and-arguments) for more details on deprecate parameters

src/azure-cli/azure/cli/command_modules/policyinsights/tests/latest/test_policyinsights_scenario.py
```diff
@@ -364,7 +364,7 @@ def test_policy_insights_remediation_management_group(self):
     @AllowLargeResponse()
     def test_policy_insights_remediation_complete(self):
         self.kwargs.update({
-            'pan': '09d18943ace14946aff83c21',
+            'pan': '98904c39668a4f70804aef09',
```
zhoxing-ms, line 367: May I ask why you need to change this value?

robga: This test test_policy_insights_remediation_complete also requires test environment setup. There are several steps in the comments in the front of the test. Updating the policy assignment value is in the step 4. (The root reason is when we create the assignment on policy UI, the resource name is generated by the UI. We cannot specify the assignment name on the UI)

src/azure-cli/azure/cli/command_modules/resource/_help.py
robga: good catch!

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -220,6 +220,22 @@ def load_arguments(self, _):
...
         c.argument('name', options_list=['--name', '-n'], help='Name of the new policy set definition.')
 
+    with self.argument_context('policy exemption', min_api='2020-09-01', resource_type=ResourceType.MGMT_RESOURCE_POLICY) as c:
+        c.ignore('_subscription')
+        c.argument('name', options_list=['--name', '-n'], completer=get_policy_exemption_completion_list, help='Name of the policy exemption.')
+        c.argument('scope', help='Scope to which this policy exemption applies.')
+        c.argument('disable_scope_strict_match', options_list=['--disable-scope-strict-match', '-i'], action='store_true', help='Include policy exemptions either inherited from parent scope or at child scope.')
+        c.argument('display_name', help='Display name of the policy exemption.')
+        c.argument('description', help='Description of policy exemption.')
+        c.argument('exemption_category', options_list=['--exemption-category', '-e'], help='The policy exemption category of the policy exemption. Possible values are Waiver and Mitigated.')
```
zhoxing-ms, line 230:
```suggestion
        c.argument('exemption_category', options_list=['--exemption-category', '-e'], help='The policy exemption category of the policy exemption. ', arg_type=get_enum_type(['Waiver', 'Mitigated']))
```

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -220,6 +220,22 @@ def load_arguments(self, _):
...
+        c.argument('policy_definition_reference_ids', nargs='+', options_list=['--policy-definition-reference-ids', '-r'], help='The policy definition reference ids to exempt in the initiative (policy set).')
```
zhoxing-ms, line 231: > help='The policy definition reference ids to exempt in the initiative (policy set).'

Why are there `()` outside `policy set` ?

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -2437,6 +2432,96 @@ def update_policy_setdefinition(cmd, policy_set_definition_name, definitions=Non
...
 
 
+def create_policy_exemption(cmd, name, policy_assignment=None, exemption_category=None,
+                            policy_definition_reference_ids=None, expires_on=None,
+                            display_name=None, description=None, resource_group_name=None, scope=None,
+                            metadata=None):
+    if policy_assignment is None:
+        raise CLIError('--policy_assignment is required')
+    if exemption_category is None:
+        raise CLIError('--exemption_category is required')
```
zhoxing-ms, line 2442:
```suggestion
    if policy_assignment is None:
        raise RequiredArgumentMissingError('--policy_assignment is required')
    if exemption_category is None:
        raise RequiredArgumentMissingError('--exemption_category is required')
```

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -2437,6 +2432,96 @@ def update_policy_setdefinition(cmd, policy_set_definition_name, definitions=Non
...
+            resource_group, namespace,
+            parent_resource_path, resource_type, resource_name)
+    elif resource_group:
+        result = policy_client.policy_exemptions.list_for_resource_group(resource_group)
+    elif subscription:
+        result = policy_client.policy_exemptions.list()
+    elif scope:
+        raise CLIError('usage error `--scope`: must be a fully qualified ARM ID.')
+    else:
+        raise CLIError('usage error: --scope ARM_ID | --resource-group NAME')
```
zhoxing-ms, line 2517:
```suggestion
    elif scope:
        raise InvalidArgumentValueError('usage error `--scope`: must be a fully qualified ARM ID.')
    else:
        raise InvalidArgumentValueError('usage error: --scope ARM_ID | --resource-group NAME')
```

src/azure-cli/azure/cli/command_modules/resource/tests/latest/test_resource.py
```diff
@@ -2539,12 +2531,12 @@ def test_resource_policy_management_group(self, resource_group):
...
 
             # Attempt to get a policy definition at an invalid management group scope
-            with self.assertRaises(SystemExit):
+            with self.assertRaises(IncorrectUsageError):
                 self.cmd(self.cmdstring('policy definition show -n "/providers/microsoft.management/managementgroups/myMg/providers/microsoft.authorization/missingsegment"'))
         finally:
             self.cmd('account management-group delete -n ' + management_group_name)
 
-    @record_only()
+    @live_only()
```
zhoxing-ms, line 2539: Why change this test from `@record_only()` to `@live_only()`

src/azure-cli/azure/cli/command_modules/resource/tests/latest/test_resource.py
```diff
@@ -2582,9 +2574,9 @@ def test_resource_policyset_subscription_id(self, resource_group):
         if not self.in_recording:
             with mock.patch('azure.cli.command_modules.resource.custom._get_subscription_id_from_subscription',
                             return_value=MOCKED_SUBSCRIPTION_ID):
-                self.resource_policyset_operations(resource_group, None, '0b1f6471-1bf0-4dda-aec3-cb9272f09590')
+                self.resource_policyset_operations(resource_group, None, 'f67cc918-f64f-4c3f-aa24-a855465f9d41')
         else:
-            self.resource_policyset_operations(resource_group, None, '0b1f6471-1bf0-4dda-aec3-cb9272f09590')
+            self.resource_policyset_operations(resource_group, None, 'f67cc918-f64f-4c3f-aa24-a855465f9d41')
```
zhoxing-ms, line 2579: Please do not replace the tenant id used by the CLI team. If you don't have permission for this tenant, I can help you re-record yaml file

src/azure-cli/azure/cli/command_modules/resource/tests/latest/test_resource.py
```diff
@@ -2701,6 +2691,223 @@ def test_resource_create_policy_assignment_random(self, resource_group, manageme
...
+    @AllowLargeResponse(4096)
+    def test_resource_policyexemption_subscription(self, resource_group):
+        # under playback, we mock it so the subscription id will be '00000000...' and it will match
+        # the same sanitized value in the recording
+        if not self.in_recording:
+            with mock.patch('azure.cli.command_modules.resource.custom._get_subscription_id_from_subscription',
+                            return_value=MOCKED_SUBSCRIPTION_ID):
+                self.resource_policyexemption_operations(resource_group, None, 'f67cc918-f64f-4c3f-aa24-a855465f9d41')
+        else:
+            self.resource_policyexemption_operations(resource_group, None, 'f67cc918-f64f-4c3f-aa24-a855465f9d41')
```
zhoxing-ms, line 2909: Please use the tenant id commonly used by CLI team~
```suggestion
                self.resource_policyexemption_operations(resource_group, None, '0b1f6471-1bf0-4dda-aec3-cb9272f09590')
        else:
            self.resource_policyexemption_operations(resource_group, None, '0b1f6471-1bf0-4dda-aec3-cb9272f09590')
```

src/azure-cli/azure/cli/command_modules/policyinsights/tests/latest/test_policyinsights_scenario.py
zhoxing-ms: Do you mean that the steps to set up the test environment must be on portal?
If so, why can't it be implemented by commands in the test?

robga: Yes, it needs to setup through Portal. I don't know why wasn't it implemented by cmd before. I guess the scenario depends on many async operations and needs a long wait time for backend job could be potentially the blocker

src/azure-cli/azure/cli/command_modules/resource/_params.py
robga: "initiative" is another name for "policy set".

src/azure-cli/azure/cli/command_modules/policyinsights/tests/latest/test_policyinsights_scenario.py
robga: Actually, it described the reason in the comment above: 
```
    # Executing a real remediation requires more time-intensive setup than can be done in a live scenario test.
    # This record_only test executes a real remediation against a known non-compliant policy
```

src/azure-cli/azure/cli/command_modules/resource/tests/latest/test_resource.py
robga: This test will create policy assignment on a specific subscription. Policy assignment uses --scope argument. The mock doesn't work when the subscription comes from --scope. 
This test wasn't run in either live or record before my change.

src/azure-cli/azure/cli/command_modules/resource/tests/latest/test_resource.py
robga: updated. It seems the mock.patch helped the test pass.

src/azure-cli/azure/cli/command_modules/resource/tests/latest/test_resource.py
robga: updated, the mock patch call helped the test pass

src/azure-cli/azure/cli/command_modules/policyinsights/tests/latest/test_policyinsights_scenario.py
zhoxing-ms: I suggest trying to use commands to set up the environment, otherwise it is not easy to maintain and reuse this test in the future

src/azure-cli/azure/cli/command_modules/resource/_params.py
zhoxing-ms: Got it, thanks~

src/azure-cli/azure/cli/command_modules/policyinsights/tests/latest/test_policyinsights_scenario.py
robga: The test pre-requirements of this test needs to create other resources like VM. If we creates those resources on the fly, the test will become the same situation as that we are hitting right now on the SQL test. I think ideally, the command module should not create dependencies on other command modules. E.g. the current broken SQL test is not do any validation on the policy CLI cmd. They just called policy CLI cmd to update their test env to unblock the SQL managed instance creation. I think that is not proper. I think SDK team should discuss about the strategy/restriction of calling other command module in the command module scenario test. The product team may not response to the re-record request in time. It will become un-necessary blocker.

zhoxing-ms: At present, the tests of some modules in CLI depend on the commands of other modules to construct the environment, so it is quite common to call other module commands in testing.
It is a known issue that bumping version for some RP will cause other module tests which rely on this RP to need to be re-recorded.
For the creation of some resources, we can use prepare (such as @ResourceGroupPreparer,@StorageAccountPreparer) to avoid this problem, but VM is not supported at present.
If it's not easy to build a test environment with commands for this case, we can pass this problem first.

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -220,6 +220,22 @@ def load_arguments(self, _):
...
+        c.ignore('_subscription')
+        c.argument('name', options_list=['--name', '-n'], completer=get_policy_exemption_completion_list, help='Name of the policy exemption.')
+        c.argument('scope', help='Scope to which this policy exemption applies.')
+        c.argument('disable_scope_strict_match', options_list=['--disable-scope-strict-match', '-i'], action='store_true', help='Include policy exemptions either inherited from parent scope or at child scope.')
+        c.argument('display_name', help='Display name of the policy exemption.')
+        c.argument('description', help='Description of policy exemption.')
+        c.argument('exemption_category', options_list=['--exemption-category', '-e'], help='The policy exemption category of the policy exemption', arg_type=get_enum_type(['Waiver', 'Mitigated']))
+        c.argument('policy_definition_reference_ids', nargs='+', options_list=['--policy-definition-reference-ids', '-r'], help='The policy definition reference ids to exempt in the initiative (policy set).')
+        c.argument('expires_on', help='"The expiration date and time (in UTC ISO 8601 format yyyy-MM-ddTHH:mm:ssZ) of the policy exemption.')
```
zhoxing-ms, line 232:
```suggestion
        c.argument('expires_on', help='The expiration date and time (in UTC ISO 8601 format yyyy-MM-ddTHH:mm:ssZ) of the policy exemption.')
```

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -220,6 +220,22 @@ def load_arguments(self, _):
     with self.argument_context('policy set-definition create', min_api='2017-06-01-preview', resource_type=ResourceType.MGMT_RESOURCE_POLICY) as c:
         c.argument('name', options_list=['--name', '-n'], help='Name of the new policy set definition.')
 
+    with self.argument_context('policy exemption', min_api='2020-09-01', resource_type=ResourceType.MGMT_RESOURCE_POLICY) as c:
```
Juliehzl, line 223: Could we add Preview for the new command group for one sprint?

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -220,6 +220,22 @@ def load_arguments(self, _):
...
+        c.ignore('_subscription')
+        c.argument('name', options_list=['--name', '-n'], completer=get_policy_exemption_completion_list, help='Name of the policy exemption.')
+        c.argument('scope', help='Scope to which this policy exemption applies.')
+        c.argument('disable_scope_strict_match', options_list=['--disable-scope-strict-match', '-i'], action='store_true', help='Include policy exemptions either inherited from parent scope or at child scope.')
+        c.argument('display_name', help='Display name of the policy exemption.')
+        c.argument('description', help='Description of policy exemption.')
+        c.argument('exemption_category', options_list=['--exemption-category', '-e'], help='The policy exemption category of the policy exemption', arg_type=get_enum_type(['Waiver', 'Mitigated']))
```
Juliehzl, line 230: Is it possible to read from SDK definition for the enum list?

src/azure-cli/azure/cli/command_modules/resource/commands.py
```diff
@@ -387,6 +393,13 @@ def load_command_table(self, _):
         g.custom_show_command('show', 'get_policy_setdefinition')
         g.custom_command('update', 'update_policy_setdefinition')
 
+    with self.command_group('policy exemption', resource_policy_exemptions_sdk, resource_type=ResourceType.MGMT_RESOURCE_POLICY, min_api='2020-09-01') as g:
+        g.custom_command('create', 'create_policy_exemption')
+        g.custom_command('delete', 'delete_policy_exemption')
```
Juliehzl, line 398: please add confirmation=True for delete command

src/azure-cli/azure/cli/command_modules/resource/commands.py
```diff
@@ -387,6 +393,13 @@ def load_command_table(self, _):
...
+        g.custom_command('list', 'list_policy_exemption')
+        g.custom_show_command('show', 'get_policy_exemption')
+        g.custom_command('update', 'update_policy_exemption')
```
Juliehzl, line 401: could we use generic update command for you update? https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#generic-update-commands

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -2437,6 +2431,96 @@ def update_policy_setdefinition(cmd, policy_set_definition_name, definitions=Non
...
+def create_policy_exemption(cmd, name, policy_assignment=None, exemption_category=None,
+                            policy_definition_reference_ids=None, expires_on=None,
+                            display_name=None, description=None, resource_group_name=None, scope=None,
+                            metadata=None):
+    if policy_assignment is None:
+        raise RequiredArgumentMissingError('--policy_assignment is required')
+    if exemption_category is None:
+        raise RequiredArgumentMissingError('--exemption_category is required')
+
+    policy_client = _resource_policy_client_factory(cmd.cli_ctx)
```
Juliehzl, line 2443: could we use cf_policy_exemptions as you defined in command type?

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -2437,6 +2431,96 @@ def update_policy_setdefinition(cmd, policy_set_definition_name, definitions=Non
...
+                                display_name=display_name, description=description, metadata=metadata)
+    createdExemption = policy_client.policy_exemptions.create_or_update(scope, name, exemption)
+    return createdExemption
+
+
+def update_policy_exemption(cmd, name, exemption_category=None,
+                            policy_definition_reference_ids=None, expires_on=None,
+                            display_name=None, description=None, resource_group_name=None, scope=None,
+                            metadata=None):
+    policy_client = _resource_policy_client_factory(cmd.cli_ctx)
```
Juliehzl, line 2458: the same

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -2437,6 +2431,96 @@ def update_policy_setdefinition(cmd, policy_set_definition_name, definitions=Non
...
+        expires_on=expires_on if expires_on is not None else exemption.expires_on,
+        display_name=display_name if display_name is not None else exemption.display_name,
+        description=description if description is not None else exemption.description,
+        metadata=metadata if metadata is not None else exemption.metadata)
+    updatedExemption = policy_client.policy_exemptions.create_or_update(scope, name, parameters)
+    return updatedExemption
+
+
+def delete_policy_exemption(cmd, name, resource_group_name=None, scope=None):
+    policy_client = _resource_policy_client_factory(cmd.cli_ctx)
```
Juliehzl, line 2476: the same

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -2437,6 +2431,96 @@ def update_policy_setdefinition(cmd, policy_set_definition_name, definitions=Non
...
+    scope = _build_policy_scope(policy_client.config.subscription_id,
+                                resource_group_name, scope)
+    policy_client.policy_exemptions.delete(scope, name)
+
+
+def get_policy_exemption(cmd, name, resource_group_name=None, scope=None):
+    policy_client = _resource_policy_client_factory(cmd.cli_ctx)
```
Juliehzl, line 2483: the same

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -2437,6 +2431,96 @@ def update_policy_setdefinition(cmd, policy_set_definition_name, definitions=Non
...
+    scope = _build_policy_scope(policy_client.config.subscription_id,
+                                resource_group_name, scope)
+    return policy_client.policy_exemptions.get(scope, name)
+
+
+def list_policy_exemption(cmd, disable_scope_strict_match=None, resource_group_name=None, scope=None):
+    from azure.cli.core.commands.client_factory import get_subscription_id
+    policy_client = _resource_policy_client_factory(cmd.cli_ctx)
```
Juliehzl, line 2491: the same

src/azure-cli/azure/cli/command_modules/sql/tests/latest/test_sql_commands.py
```diff
@@ -5632,7 +5632,10 @@ def test_sql_mi_failover_mgmt(self):
 
 
 class SqlManagedDatabaseLogReplayScenarionTest(ScenarioTest):
-    @ResourceGroupPreparer(random_name_length=28, name_prefix='clitest-logreplay', location='westcentralus')
+    import unittest
```
Juliehzl, line 5635: issue created in https://github.com/Azure/azure-cli/issues/17603

src/azure-cli/azure/cli/command_modules/policyinsights/tests/latest/test_policyinsights_scenario.py
robga: Let's pass this problem first.

src/azure-cli/azure/cli/command_modules/resource/_params.py
robga: good catch.

src/azure-cli/azure/cli/command_modules/resource/commands.py
robga: Existing policy command don't ask for confirmation. I'd keep the consistency on policy related commands.

src/azure-cli/azure/cli/command_modules/resource/commands.py
robga: As other policy resource, the policy exemption doesn't support patch. And using generic kind of break the consistency.

src/azure-cli/azure/cli/command_modules/resource/_params.py
robga: I don't know. @zhoxing-ms ?

src/azure-cli/azure/cli/command_modules/resource/_params.py
robga: Sure, for the command group preview, the flag is added to the commends.py file. (preview flag in the params is only for the argument preview state.)

src/azure-cli/azure/cli/command_modules/resource/custom.py
robga: cf_policy_exemptions is not defined in this file. And I needs subscription_id from the policy_client object. It doesn't help remove this line.

src/azure-cli/azure/cli/command_modules/resource/custom.py
robga: cf_policy_exemptions is not defined in this file. And I needs subscription_id from the policy_client object. It doesn't help remove this line.

src/azure-cli/azure/cli/command_modules/resource/custom.py
robga: cf_policy_exemptions is not defined in this file. And I needs subscription_id from the policy_client object. It doesn't help remove this line.

src/azure-cli/azure/cli/command_modules/resource/custom.py
robga: cf_policy_exemptions is not defined in this file. And I needs subscription_id from the policy_client object. It doesn't help remove this line.

src/azure-cli/azure/cli/command_modules/resource/custom.py
robga: cf_policy_exemptions is not defined in this file. And I needs subscription_id from the policy_client object. It doesn't help remove this line.

src/azure-cli/azure/cli/command_modules/resource/_params.py
zhoxing-ms: Yes, the enumeration values can be obtained from ExemptionCategory

src/azure-cli/azure/cli/command_modules/resource/commands.py
Juliehzl: then what is the difference between create and update? Generic update command is a CLI solution for operations without PATCH to provide a PATCH-like behavior.

src/azure-cli/azure/cli/command_modules/resource/custom.py
Juliehzl: How policy_client.config.subscription_id get the sub id value then?

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -220,6 +220,22 @@ def load_arguments(self, _):
     with self.argument_context('policy set-definition create', min_api='2017-06-01-preview', resource_type=ResourceType.MGMT_RESOURCE_POLICY) as c:
         c.argument('name', options_list=['--name', '-n'], help='Name of the new policy set definition.')
 
+    with self.argument_context('policy exemption', min_api='2020-09-01', resource_type=ResourceType.MGMT_RESOURCE_POLICY) as c:
+        c.ignore('_subscription')
```
Juliehzl, line 224: as you ignore subscription here, it seems there is no subscription id exposed as parameter any more?

src/azure-cli/azure/cli/command_modules/resource/_params.py
robga: updated.

src/azure-cli/azure/cli/command_modules/resource/commands.py
robga: I'd keep the consistency with other policy commands.

src/azure-cli/azure/cli/command_modules/resource/custom.py
robga: It is the same as other policy commands to get the subscription Id from the context (current selected subscription)

src/azure-cli/azure/cli/command_modules/resource/_params.py
robga: Yes, it is using scope parameter, which is consistent with the policy assignment command.

src/azure-cli/azure/cli/command_modules/resource/custom.py
zhoxing-ms: @robga Actually, the `subscription_id` of ` policy_client.config` comes from `cli_ctx`, why not get `subscription_id` directly from `cli_ctx`?
Such as `cmd.cli_ctx.data['subscription_id']`

robga: OK, I will try it.

zhoxing-ms (approved):
Because there are some nonstandard writing style in the old code that have been referenced in this PR, I approve this PR at first. We could consider refactoring the overall nonstandard part in the future.

Juliehzl (approved):
I still have concern for code style. But as it is close to build, approved to merge the feature. But please refactor the code in future. Next time we will not approve for non-standard code style.

PR #17580, opened 2021-04-06
[Compute] add VMSS Rolling Upgrade opt

### LINKED ISSUE #16038: VMSS Rolling Upgrade opt in flags
**Resource Provider**

Compute Resource Provider (CRP) - > Virtual Machine Scale Sets (VMSS)

VMSS Rolling Upgrade provides an upgrade facility that automatically applies VMSS model changes to the individual instances through a health-based orchestration. 
There are some additional opt-in flags we are introducing for Upgrade policy for VMSS. At the moment we expose these four properties that our customers configure  - 

1.	maxBatchInstancePercent - The maximum percent of total virtual machine instances that will be upgraded simultaneously by the rolling upgrade in one batch. Default: 20% 
2.	maxUnhealthyInstancePercent - The maximum percentage of the total virtual machine instances in the scale set that can be simultaneously unhealthy. Default: 20% 
3.	maxUnhealthyUpgradedInstancePercent - The maximum percentage of upgraded virtual machine instances that can be found to be in an unhealthy state. Default: 20% 
4.	pauseTimeBetweenBatches - The wait time between completing the update for all virtual machines in one batch and starting the next batch. Default: 0 seconds 

In addition to these four, we are introducing two new properties. Note that these are opt-in properties. 

1.	enableCrossZoneUpgrade - Setting this Boolean property will allow VMSS to ignore AZ boundaries when constructing upgrade batches, and only consider Update Domain and maxBatchInstancePercent to determine the batch size. 
2.	prioritizeUnhealthyInstances - Setting this Boolean property will lead to all unhealthy instances in a scale set getting upgraded before any healthy instances. 

2020-12-01

**Swagger Link**

https://github.com/Azure/azure-rest-api-specs/pull/11184
[Rolling upgrade - Prioritize Unhealthy VMs.docx](https://github.com/Azure/azure-cli/files/5594864/Rolling.upgrade.-.Prioritize.Unhealthy.VMs.docx)

**Target Date**

2020-12-20

xfz11:
**Description**
resolve #16038 
**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/vm/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_params.py
@@ -610,17 +610,17 @@ def load_arguments(self, _):
         with self.argument_context(scope) as c:
             c.argument('terminate_notification_time', min_api='2019-03-01',
                        help='Length of time (in minutes, between 5 and 15) a notification to be sent to the VM on the instance metadata server till the VM gets deleted')
-            c.argument('max_batch_instance_percent', type=int, options_list=['--max-bip'], min_api='2020-12-01',
+            c.argument('max_batch_instance_percent', type=int, min_api='2020-12-01',
                        help='The maximum percent of total virtual machine instances that will be upgraded simultaneously by the rolling upgrade in one batch. Default: 20%')
-            c.argument('max_unhealthy_instance_percent', type=int, options_list=['--max-uip'], min_api='2020-12-01',
+            c.argument('max_unhealthy_instance_percent', type=int, min_api='2020-12-01',
                        help='The maximum percentage of the total virtual machine instances in the scale set that can be simultaneously unhealthy. Default: 20%')
-            c.argument('max_unhealthy_upgraded_instance_percent', type=int, options_list=['--max-uuip'], min_api='2020-12-01',
+            c.argument('max_unhealthy_upgraded_instance_percent', type=int, min_api='2020-12-01',
                        help='The maximum percentage of upgraded virtual machine instances that can be found to be in an unhealthy state. Default: 20%')
-            c.argument('pause_time_between_batches', options_list=['--ptb'], min_api='2020-12-01',
+            c.argument('pause_time_between_batches', min_api='2020-12-01',
                        help='The wait time between completing the update for all virtual machines in one batch and starting the next batch. Default: 0 seconds')
-            c.argument('enable_cross_zone_upgrade', options_list=['--enable-czu'], arg_type=get_three_state_flag(), min_api='2020-12-01',
+            c.argument('enable_cross_zone_upgrade', arg_type=get_three_state_flag(), min_api='2020-12-01',
                        help='Set this Boolean property will allow VMSS to ignore AZ boundaries when constructing upgrade batches, and only consider Update Domain and maxBatchInstancePercent to determine the batch size')
-            c.argument('prioritize_unhealthy_instances', options_list=['--prioritize-ui'], arg_type=get_three_state_flag(), min_api='2020-12-01',
+            c.argument('prioritize_unhealthy_instances', arg_type=get_three_state_flag(), min_api='2020-12-01',
                        help='Set this Boolean property will lead to all unhealthy instances in a scale set getting upgraded before any healthy instances')
 
     for scope, help_prefix in [('vmss update', 'Update the'), ('vmss wait', 'Wait on the')]:
--- a/src/azure-cli/azure/cli/command_modules/vm/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_params.py
@@ -610,6 +610,18 @@ def load_arguments(self, _):
         with self.argument_context(scope) as c:
             c.argument('terminate_notification_time', min_api='2019-03-01',
                        help='Length of time (in minutes, between 5 and 15) a notification to be sent to the VM on the instance metadata server till the VM gets deleted')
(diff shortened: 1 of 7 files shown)


src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -3024,6 +3033,29 @@ def update_vmss(cmd, resource_group_name, name, license_type=None, no_wait=False
     if proximity_placement_group is not None:
         vmss.proximity_placement_group = {'id': proximity_placement_group}
 
+    if max_batch_instance_percent is not None or max_unhealthy_instance_percent is not None \
```
yungezz, line 3036: why new condition check? it's better to have comments

src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -610,6 +610,18 @@ def load_arguments(self, _):
...
             c.argument('terminate_notification_time', min_api='2019-03-01',
                        help='Length of time (in minutes, between 5 and 15) a notification to be sent to the VM on the instance metadata server till the VM gets deleted')
+            c.argument('max_batch_instance_percent', type=int, options_list=['--max-bip'], min_api='2020-12-01',
+                       help='The maximum percent of total virtual machine instances that will be upgraded simultaneously by the rolling upgrade in one batch. Default: 20%')
+            c.argument('max_unhealthy_instance_percent', type=int, options_list=['--max-uip'], min_api='2020-12-01',
+                       help='The maximum percentage of the total virtual machine instances in the scale set that can be simultaneously unhealthy. Default: 20%')
+            c.argument('max_unhealthy_upgraded_instance_percent', type=int, options_list=['--max-uuip'], min_api='2020-12-01',
+                       help='The maximum percentage of upgraded virtual machine instances that can be found to be in an unhealthy state. Default: 20%')
+            c.argument('pause_time_between_batches', options_list=['--ptb'], min_api='2020-12-01',
```
houk-ms, line 619: @qwordy Does the short option follow VM's convention?

houk-ms:
Do we have any convention when adding a short option for the parameter? @qwordy 

If not, maybe we should have some.

qwordy:
> 
> 
> Do we have any conventions when adding a short option for the parameter? @qwordy
> 
> If not, maybe we should have some.

It is an opportunity that you can draft one document.

src/azure-cli/azure/cli/command_modules/vm/_params.py
qwordy: VM does not prefer abbreviation based on existing code style.

src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
```diff
@@ -915,7 +918,15 @@ def build_vmss_resource(cmd, name, naming_prefix, location, tags, overprovision,
...
-            'mode': upgrade_policy_mode
+            'mode': upgrade_policy_mode,
+            'rollingUpgradePolicy': {
+                'maxBatchInstancePercent': max_batch_instance_percent,
+                'maxUnhealthyInstancePercent': max_unhealthy_instance_percent,
+                'maxUnhealthyUpgradedInstancePercent': max_unhealthy_upgraded_instance_percent,
+                'pauseTimeBetweenBatches': pause_time_between_batches,
+                'enableCrossZoneUpgrade': enable_cross_zone_upgrade,
+                'prioritizeUnhealthyInstances': prioritize_unhealthy_instances
+            }
```
qwordy, line 929: If they are all not specified, do we need to set these `None`s? Does it break existing VM commands?

xfz11: According to my test, `None `value will not change the properties' original values

qwordy (approved):

houk-ms (approved):

src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -610,6 +610,18 @@ def load_arguments(self, _):
...
+            c.argument('max_unhealthy_upgraded_instance_percent', type=int, min_api='2020-12-01',
+                       help='The maximum percentage of upgraded virtual machine instances that can be found to be in an unhealthy state. Default: 20%')
+            c.argument('pause_time_between_batches', min_api='2020-12-01',
+                       help='The wait time between completing the update for all virtual machines in one batch and starting the next batch. Default: 0 seconds')
+            c.argument('enable_cross_zone_upgrade', arg_type=get_three_state_flag(), min_api='2020-12-01',
+                       help='Set this Boolean property will allow VMSS to ignore AZ boundaries when constructing upgrade batches, and only consider Update Domain and maxBatchInstancePercent to determine the batch size')
+            c.argument('prioritize_unhealthy_instances', arg_type=get_three_state_flag(), min_api='2020-12-01',
+                       help='Set this Boolean property will lead to all unhealthy instances in a scale set getting upgraded before any healthy instances')
```
zhoxing-ms, line 624: @xfz11 Hi, I see that the `min_api` of these parameters is `2020-12-01`, but in fact, these parameters are already supported in version `2020-06-01`.
Swagger link: [code link](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/compute/resource-manager/Microsoft.Compute/stable/2020-06-01/compute.json#L9207 )

So I would like to confirm that those parameters are expected to be supported only after version `2020-12-01`, right? (in this case, profile `2020-09-01-hybrid` does not support those parameters)

PR #17609, opened 2021-04-08
{Telemetry} Add telemetry info of extension even when install fail

fengzhou-msft:
**Description**

Existing code will only add `Context.Default.AzureCLI.ExtensionManagementDetail` after the extension is installed. This PR extracts extension name and version from source url and add to telemetry so that these data will be sent by telemetry even when the install of extension fails.

**Testing Guide**

In a virtual env, downgrade pip: `python -m pip -U pip==18.0`.

Run `az extension add -s https://azurecliext.blob.core.windows.net/release/azure_cli_ml-1.21.0-py3-none-any.whl -y`. 

The install will fail, check the telemetry record in `~/.azure/telemetry/cache`, you should see `"Context.Default.AzureCLI.ExtensionManagementDetail": "azure-cli-ml@1.21.0"`.

**History Notes**

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/extension/operations.py
+++ b/src/azure-cli-core/azure/cli/core/extension/operations.py
@@ -85,6 +85,17 @@ def _validate_whl_extension(ext_file):
     check_version_compatibility(azext_metadata)
 
 
+def _get_extension_info_from_source(source):
+    url_parse_result = urlparse(source)
+    is_url = (url_parse_result.scheme == 'http' or url_parse_result.scheme == 'https')
+    whl_filename = os.path.basename(url_parse_result.path) if is_url else os.path.basename(source)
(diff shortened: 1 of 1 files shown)


msyyc (approved):

src/azure-cli-core/azure/cli/core/extension/operations.py
```diff
@@ -296,6 +307,7 @@ def add_extension(cmd=None, source=None, extension_name=None, index_url=None, ye
     if extension_name:
         cmd_cli_ctx.get_progress_controller().add(message='Searching')
         ext = None
+        set_extension_management_detail(extension_name, version)
```
fengzhou-msft, line 310: `version` may be None, the information will be overwritten if the extension is resolved from index. If before that, the command exits, then the information set here can be useful.

Issue #17625, opened 2021-04-09
Redact tokens from `--debug` log


jiasli:
## Context

1. Starting from `azure-core` 1.13.0, `Authorization` header is now exposed in DEBUG log (https://github.com/Azure/azure-sdk-for-python/pull/17424).
2. Python SDK decided not to redact `x-ms-authorization-auxiliary` header (https://github.com/Azure/azure-sdk-for-python/issues/17271).

Before bumping `azure-core` to 1.13.0, Azure CLI must adapt to `azure-core`'s new behavior.

## Proposed solutions

For tokens in `Authorization` and `x-ms-authorization-auxiliary`:

1. Keep the current behavior and redact tokens.
2. Expose tokens in `--debug` mode like the new `NetworkTraceLoggingPolicy`.
   1. Unconditionally do so, but redact tokens in `--verbose`.
   2. Make a config like `az config set logging.show_tokens=True`.

yonzhan:
Core

PR #17628, opened 2021-04-09
{Packaging} Pin `azure-core` to 1.12.0

jiasli:
**Description**

Temporary fix for https://github.com/Azure/azure-cli/issues/17625.

Starting from `azure-core` 1.13.0, `Authorization` header is now exposed in DEBUG log (https://github.com/Azure/azure-sdk-for-python/pull/17424).

Temporarily pin `azure-core` to 1.12.0 so that `Authorization` header is redacted.

**Testing Guide**
```
az storage account list --debug
```

### DIFF
--- a/src/azure-cli-core/setup.py
+++ b/src/azure-cli-core/setup.py
@@ -46,6 +46,7 @@ DEPENDENCIES = [
     'argcomplete~=1.8',
     'azure-cli-telemetry==1.0.6.*',
     'azure-common~=1.1',
+    'azure-core==1.12.0',
     'azure-mgmt-core>=1.2.0,<2.0.0',
     'colorama~=0.4.1',
     'cryptography>=3.2,<3.4',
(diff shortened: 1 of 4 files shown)


src/azure-cli-core/setup.py
```diff
@@ -46,6 +46,7 @@
     'argcomplete~=1.8',
     'azure-cli-telemetry==1.0.6.*',
     'azure-common~=1.1',
+    'azure-core==1.12.0',
```
jiasli, line 49: Also do this in `setup.py` so that CLI installed with `pip` will also only use `azure-core==1.12.0`.

yonzhan:
Packaging

fengzhou-msft (approved):

evelyn-ys (approved):

jsntcy (approved):

Issue #17551, opened 2021-04-02
[Feature Request] Support `--scope` in `az login`


jiasli:
**Is your feature request related to a problem? Please describe.**

The current Azure CLI unconditionally uses ARM resource ID (e.g.

https://github.com/Azure/azure-cli/blob/3feea02888ea67f033f407174a3a7a340158b81a/src/azure-cli-core/azure/cli/core/cloud.py#L306

) to call `/authorize` API of [authorization code flow](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow).

There are increasing asks for Conditional Access support from multiple service teams (#15220):

- Storage
- App Configuration
- VM SSH

**Describe the solution you'd like**

We should reconsider supporting Conditional Access on `dev` branch. 

In order to do so, `az login` must take `--scope` and use it in `/authorize`.

ℹ ADAL only takes `resource`, so `scope` must be converted back to `resource` using 

https://github.com/Azure/azure-cli/blob/ad30d5595a4e49ff4fa5b2be887b8fac3a1bce79/src/azure-cli-core/azure/cli/core/util.py#L1214-L1228

yonzhan:
support scope

PR #17093, opened 2021-02-25
[Compute] az ssh vm: Automatically launch browser when getting certificate fails

jiasli:
## Description

Fix #16988, #16989
Require #17072

When getting certificate fails for VM SSH, 

1. the original AAD error is shown as a warning
2. browser is automatically launched to open `/authorize` with `resource=https://pas.windows.net/CheckMyAccess/Linux`

### Conditional Access policy for MFA (#16989)

When ARM doesn't require MFA, but VM SSH does, `az ssh vm` calls `/token` and fails with 

```
AADSTS50076: Due to a configuration change made by your administrator, or because you moved to a new location, you must use multi-factor authentication to access 'ce6ff14a-7fdc-4685-bbe0-f6afdfcfa8e0'.
Trace ID: ab8e08ab-b0f3-4c1b-b2b8-85f905da5b00
Correlation ID: 73ad3fe7-524c-4a6f-9e14-b448fa4b8c06
Timestamp: 2021-02-25 11:21:49Z
```

CLI then launches the browser. The browser prompts the user to do MFA. 

### Conditional Access policy for compliance check (#16988)

When ARM doesn't require compliance check, but VM SSH does, `az ssh vm` calls `/token` and fails with 

```
AADSTS53000: Device is not in required device state: compliant. Conditional Access policy requires a compliant device, and the device is not compliant. The user must enroll their device with an approved MDM provider like Intune.
Trace ID: fed27c52-e7f0-4d4b-a6df-51016ec54a00
Correlation ID: 582d7943-5bc3-4fd4-9bb9-f6c7721b131f
Timestamp: 2021-02-25 11:06:44Z
```

CLI then launches the browser. After logging in, the browser gives the user a better error message:

image

### Limitation

- Only user account with [authorization code flow](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow) is supported.
- CLI should selectively launch the browser instead of launching the browser unconditionally for any AAD error.

## Testing Guide
```
az login
az ssh vm -g <group> --vm-name <vm>
```

PR #17738, opened 2021-04-16
[Core] Conditional Access: Show `--scope` for `az login` message when failed to refresh the access token

### LINKED ISSUE #15220: Support Conditional Access
Azure CLI does not support [Conditional Access](https://docs.microsoft.com/en-us/azure/active-directory/conditional-access/) currently. For cases that require MFA for specific resources (VM for example), CLI will not be able to  handle it. We need to define a flow to communicate with user in this case.

**Additional context**

jiasli:
Resolve #15220
Require the solution of #17551

## Context

When ARM doesn't require MFA, but data-plane service does, CLI fails and can't recover.

## Change

When CLI fails because ARM's Refresh Token doesn't have permission to get data-plane service's Access Token (step-up), CLI gives an instruction to re-login with `--scope`:

```
AADSTS50076: Due to a configuration change made by your administrator, or because you moved to a new location, you must use multi-factor authentication to access 'ce6ff14a-7fdc-4685-bbe0-f6afdfcfa8e0'.
Trace ID: 9df367fc-f494-4cb4-8080-525b1a3d2800
Correlation ID: e1801d21-4173-4634-adcc-5c3f238bd9b2
Timestamp: 2021-04-16 12:34:57Z
To re-authenticate, please run:
az login --scope https://pas.windows.net/CheckMyAccess/Linux/.default
If the problem persists, please contact your tenant administrator.
```

## Additional context

This is a non-automated solution to https://github.com/Azure/azure-cli/pull/17093, because we don't want CLI to launch a browser automatically during script execution.


src/azure-cli/azure/cli/command_modules/keyvault/_completers.py
```diff
@@ -4,21 +4,14 @@
 # --------------------------------------------------------------------------------------------
 
 from azure.cli.core.decorators import Completer
-from azure.cli.core._profile import Profile
-
-
-def _get_token(cli_ctx, server, resource, scope):  # pylint: disable=unused-argument
-    return Profile(cli_ctx=cli_ctx).get_login_credentials(resource)[0]._token_retriever()  # pylint: disable=protected-access
```
jiasli, line 11: This `get_token` never works because it doesn't comply with `KeyVaultAuthentication`'s `authorization_callback` signature.

src/azure-cli/azure/cli/command_modules/keyvault/_completers.py
```diff
@@ -4,21 +4,14 @@
...
+from azure.cli.command_modules.keyvault._client_factory import keyvault_data_plane_factory
```
jiasli, line 7: As pointed by #15854, calling the protected method `_token_retriever` is simply wrong. 

Switch to calling `azure.cli.command_modules.keyvault._client_factory.keyvault_data_plane_factory` instead which calls the correct `get_raw_token` method provided by https://github.com/Azure/azure-cli/issues/15805#issuecomment-722914070.

src/azure-cli/azure/cli/command_modules/keyvault/_client_factory.py
```diff
@@ -125,7 +125,7 @@ def _keyvault_mgmt_client_factory(cli_ctx, _):
     return _keyvault_mgmt_client_factory
 
 
-def keyvault_data_plane_factory(cli_ctx, _):
+def keyvault_data_plane_factory(cli_ctx, *_):
```
jiasli, line 128: Support being called as `keyvault_data_plane_factory(cli_ctx)`.

yonzhan:
Core

yonzhan (approved):
Nice

evelyn-ys (approved):

src/azure-cli-testsdk/azure/cli/testsdk/base.py
```diff
@@ -214,6 +214,9 @@ def __init__(self, method_name):
         self.kwargs = {}
         self.test_resources_count = 0
 
+    def setUp(self):
+        patch_main_exception_handler(self)
```
jiasli, line 218: We have to patch `azure.cli.core.util.handle_exception` for `LiveScenarioTest` as well. Otherwise, `AuthenticationError` can't be checked in tests.

kairu-ms (approved):

jiasli:
Some more information I provided in an internal ticket:

## Why token refreshing fails

Consider the following scenario:

1. ARM doesn't require MFA
2. But data-plane services (like MS Graph, Storage) do, which is enforced by Conditional Access policy

Without `--scope` provided, `az login` by default authenticates against ARM `https://management.core.windows.net//.default`. So `az login` is equivalent to

```
az login --scope https://management.core.windows.net//.default
```

If ARM doesn’t have Conditional Access policies enforced, the refresh token won’t be compliant with MS Graph’s Conditional Access policy. However, when the user (possibly through Terraform) requests an access token for MS Graph (step-up refreshing) with

```
az account get-access-token --scope https://graph.microsoft.com//.default
```

the token refreshing request will fail. That’s exactly why the error message is introduced in this PR to instruct the user to re-authenticate interactively.

## Best practice for using Azure CLI non-interactively

The best practice for using Azure CLI (or through Terraform) without any interaction after `az login` is to make `az login` authenticate against the scope that have the most strict Conditional Access policy enforced, e.g., MS Graph:

```
az login --scope https://graph.microsoft.com//.default
```

Then the step-down refreshing (MS Graph -> ARM) is silent.

## 2 possible scenarios

However, there are 2 scenarios we have considered but haven’t solved yet (luckily, haven’t encountered either):

1. What if ARM and MS Graph have non-overlapping sets of Conditional Access policies enforced? In other words, the refresh token for ARM can’t be used to get access tokens for MS Graph and vice versa.
2. On top of that, we can extend this to cross-tenant scenarios - what if different tenants have non-overlapping sets of Conditional Access policies enforced. Then the refresh token can’t be used to do cross-tenant refreshing either.

Providing 2 scopes at the same time? Unfortunately, this doesn't work because only one static resource can be provided:

```
> az account get-access-token --scope https://management.core.windows.net//.default https://graph.microsoft.com//.default
AADSTS70011: The provided request must include a 'scope' input parameter. The provided value for the input parameter
'scope' is not valid. The scope https://graph.microsoft.com//.default https://management.core.windows.net//.default
offline_access openid profile is not valid. static scope limit exceeded.
...
```

Previously we had some discussion on this topic with MSAL team @rayluo. This can fundamentally affect how MSAL’s token refreshing mechanism works, as this would require MSAL to bind refresh tokens to different scopes and tenants as well.

PR #17746, opened 2021-04-17
[ACR] check-health: add support to verify dns routings to private endpoints 

yugangw-msft:


src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -312,6 +312,9 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
 
     with self.argument_context('acr check-health') as c:
         c.argument('ignore_errors', options_list=['--ignore-errors'], help='Provide all health checks, even if errors are found', action='store_true', required=False)
+        c.argument('vnet', options_list=['--vnet'],
+                   help="Virtual network ID to verify the DNS routing of private endpoints. Can be a name if in the same resource group of the registry. You must run the command inside the virtual network",
```
yugangw-msft, line 316: @toddysm, please check out the description and error messages below and see whether it makes sense and friendly to users

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -337,8 +337,75 @@ def _check_registry_health(cmd, registry_name, ignore_errors):
...
+    for pe_id in pe_ids:
+        res = parse_resource_id(pe_id)
+        network_client = get_mgmt_service_client(cmd.cli_ctx, ResourceType.MGMT_NETWORK,
+                                                 subscription_id=res['subscription'])
+        pe = network_client.private_endpoints.get(res['resource_group'], res['name'])
+        if bool(re.match(vnet_of_private_endpoint, pe.subnet.id, re.I)):
+            nic_id = pe.network_interfaces[0].id
+            nic_res = parse_resource_id(nic_id)
+            nic = network_client.network_interfaces.get(nic_res['resource_group'], nic_res['name'])
+            for dns_config in nic.ip_configurations:
```
hkuadithya, line 382: This loop is LEGIT. This way, we'll cover all registry endpoints and data endpoints for a registry PE 👍  💯

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -337,8 +337,75 @@ def _check_registry_health(cmd, registry_name, ignore_errors):
...
+
+    dns_ok = True
+    if not dns_mappings:
+        logger.warning('Registry "%s" doesn\'t have private endpoints in the vnet of "%s".',
+                       registry_name, vnet_of_private_endpoint)
+        return
+
+    for fqdn in dns_mappings:
+        result = socket.gethostbyname(fqdn)
+        if result not in dns_mappings[fqdn]:
```
hkuadithya, line 397: Trying to understand why we're using `not in` here. `result` will contain IP V4 address string right? `dns_mappings[fqdn]` is also string not a list. Any reason for not doing `if result != dns_mappings[fqdn]:`? 

Also, if an invalid/non-existing FQDN is passed to `socket.gethostbyname`, it leads to an error. Should the error be caught?

image

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -337,8 +337,75 @@ def _check_registry_health(cmd, registry_name, ignore_errors):
...
+
+    # retrieve FQDNs for registry and its data endpoint
+    pe_ids = [e.private_endpoint and e.private_endpoint.id for e in registry.private_endpoint_connections]
+    dns_mappings = {}
+    for pe_id in pe_ids:
+        res = parse_resource_id(pe_id)
+        network_client = get_mgmt_service_client(cmd.cli_ctx, ResourceType.MGMT_NETWORK,
+                                                 subscription_id=res['subscription'])
+        pe = network_client.private_endpoints.get(res['resource_group'], res['name'])
+        if bool(re.match(vnet_of_private_endpoint, pe.subnet.id, re.I)):
```
hkuadithya, line 378: Using a regex module might be overkill. Can we do `vnet_of_private_endpoint.lower() == pe.subnet.id.lower()`?

Also, variable name `subnet_of_private_endpoint` might avoid confusions?

src/azure-cli/azure/cli/command_modules/acr/_params.py
hkuadithya: @yugangw-msft Correct me if I'm wrong - please double-check the usage of VNET throughout the code. Users are expected to provide the Subnet Resource ID or (Subnet name and Vnet name). If they provide the VNET ID, the code may not work since subnetID != VnetID.

Vnet Resource ID: `/subscriptions/{subID}/resourceGroups/{resourceGroup}/providers/Microsoft.Network/virtualNetworks/{vnetName}`

Subnet Resource ID: 
`/subscriptions/{subId}/resourceGroups/adupadhyRG/providers/Microsoft.Network/virtualNetworks/{vnetName}/subnets/{subnetName}`

Usually, subnet name is `default`.

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -337,8 +337,75 @@ def _check_registry_health(cmd, registry_name, ignore_errors):
...
+                       registry_name)
+        return
+
+    if not registry.private_endpoint_connections:
+        logger.warning('Registry "%s" doesn\'t have private endpoints to verify DNS routings.', registry_name)
+        return
+
+    if not is_valid_resource_id(vnet_of_private_endpoint):
+        res = parse_resource_id(registry.id)
+        vnet_of_private_endpoint = resource_id(name=vnet_of_private_endpoint, resource_group=res['resource_group'],
```
hkuadithya, line 366: Please double-check and test the code. My thinking is that it will return VNET resource ID, not subnet ID.

Vnet Resource ID: `/subscriptions/{subID}/resourceGroups/{resourceGroup}/providers/Microsoft.Network/virtualNetworks/{vnetName}`

hkuadithya (changes requested):

src/azure-cli/azure/cli/command_modules/acr/_params.py
hkuadithya: For texts, maybe we can use `az network private-endpoint create -h` for inspiration.

image

yugangw-msft: `--subnet`  is more accurate, which I can follow your suggestion.
The context of only matching `vnet` is having 2+ PE in the same vnet (even different subnet) is an anti-pattern we want to capture, as dns routing might become unpredictable.
So, it is your call. Let me know we go with vnet or subnet

src/azure-cli/azure/cli/command_modules/acr/check_health.py
yugangw-msft: Correct, like I mentioned, we mainly care about the vnet

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -337,8 +337,75 @@ def _check_registry_health(cmd, registry_name, ignore_errors):
...
+                                               namespace='Microsoft.Network', type='virtualNetworks',
+                                               subscription=res['subscription'])
+
+    # retrieve FQDNs for registry and its data endpoint
+    pe_ids = [e.private_endpoint and e.private_endpoint.id for e in registry.private_endpoint_connections]
```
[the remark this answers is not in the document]
yugangw-msft: python's `and` is [a bit different](https://stackoverflow.com/questions/19213535/using-and-and-or-operator-with-python-strings) with non boolean value, but I can use your suggestion for better code maintainability

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -337,8 +337,75 @@ def _check_registry_health(cmd, registry_name, ignore_errors):
...
+    dns_mappings = {}
+    for pe_id in pe_ids:
+        res = parse_resource_id(pe_id)
+        network_client = get_mgmt_service_client(cmd.cli_ctx, ResourceType.MGMT_NETWORK,
```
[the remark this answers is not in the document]
yugangw-msft: It is cheap. The reason of initializing each time is PE can be cross tenant which mean the subscription value might be different. We can use dictionary with keying on subscription for cache, but that might be overkill.

src/azure-cli/azure/cli/command_modules/acr/check_health.py
yugangw-msft: will do

src/azure-cli/azure/cli/command_modules/acr/check_health.py
yugangw-msft: Will address both comments

src/azure-cli/azure/cli/command_modules/acr/_params.py
hkuadithya: Makes sense, let's go with VNET.

src/azure-cli/azure/cli/command_modules/acr/check_health.py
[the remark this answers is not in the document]
hkuadithya: Makes sense, I missed the `res['subscription']` part.

src/azure-cli/azure/cli/command_modules/acr/check_health.py
hkuadithya: I see why you used regex module here. `re.match` will return non-null object if vnet resource ID is a part of subnet resource ID. I would do a simple prefix matching to make it more readable but either of the two options is fine.

`pe.subnet.id.lower().startswith(vnet_of_private_endpoint.lower())`

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -337,8 +337,79 @@ def _check_registry_health(cmd, registry_name, ignore_errors):
...
+            result = socket.gethostbyname(fqdn)
+            if result != dns_mappings[fqdn]:
+                err = 'DNS routing to registry "%s" through private IP is incorrect. Expect: %s, Actual: %s'
+                logger.warning(err, registry_name, dns_mappings[fqdn], result)
+                dns_ok = False
+        except Exception as e:  # pylint: disable=broad-except
+            logger.warning('Error resolving DNS for %s. Ex: %s', fqdn, e)
+            dns_ok = False
+
+    if dns_ok:
```
hkuadithya, line 406: Question only: Do you see any value in printing the DNS mappings (for both ok and non-ok scenarios)? 
Do you think It might help customers with self-diagnosis?

```
registry.azurecr.io             10.0.0.1
registry.eastus.data.azurecr.io 10.0.0.2
registry.westus.data.azurecr.io 10.0.0.3
```

src/azure-cli/azure/cli/command_modules/acr/check_health.py
hkuadithya: @yugangw-msft Do you think we should validate if the user has entered a valid VNET resource ID? Right now, we only check if it's a valid resource ID.

For instance, if a user (by mistake) provides a Registry ID instead of VNET ID, the current code flow will log an error:
>Registry ... doesn\t have private endpoints in the vnet of ...

src/azure-cli/azure/cli/command_modules/acr/check_health.py
yugangw-msft: Right now we only have stderr for non structured info(stdout is for json data), I suggest keep it terse so error can be spotted easily. 
However if feedback shows users don't get the error easily, we can incorporate errors in the pretty formating
```
registry.azurecr.io                        10.0.0.1
registry.eastus.data.azurecr.io            10.0.0.2
registry.westus.data.azurecr.io            10.0.0.3
registry.westeurope.data.azurecr.io    (dns routing error)
```

hkuadithya (approved):
Overall LGTM. Last two things.

1. Please confirm the Help text and description for the `--vnet` command line arg
2. Please evaluate if we need to validate if `vnet_resource_id` is a valid VNET resource id. Right now, we only verify if it's a valid ARM resource ID.

yugangw-msft:
@fengzhou-msft, please review, and merge if you are happy

fengzhou-msft (approved):

PR #17750, opened 2021-04-19
{Core} Fix #17708: Refactor user_confirmation

### LINKED ISSUE #17708: Refactor user_confirmation
`user_confirmation` is defined in multiple places with the same (or slightly different) code. Please only keep the method defnition in `core/util.py` and refactor accordingly.

PARADISSEEKR:
**Description**

Fix issue #17708: Refactor user_confirmation
**Testing Guide**

**History Notes**

---


src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -1123,11 +1123,12 @@ def get_az_rest_user_agent():
 
 def user_confirmation(message, yes=False):
     if yes:
-        return
+        return True
```
fengzhou-msft, line 1126: I don't think you need to add a return value. As `False` will raise an exception, if code below where this function gets executed, the result for `user_confirmation()` must be `True`.

src/azure-cli/azure/cli/command_modules/rdbms/flexible_server_custom_common.py
```diff
@@ -124,16 +125,3 @@ def database_delete_func(client, resource_group_name=None, server_name=None, dat
...
-    if yes:
-        return True
-    from knack.prompting import prompt_y_n, NoTTYException
-    try:
-        if not prompt_y_n(message):
-            raise CLIError('Operation cancelled.')
-        return True
-    except NoTTYException:
-        raise CLIError(
-            'Unable to prompt for confirmation as no tty available. Use --yes.')
```
fengzhou-msft, line 139: Check where this function is called. It could be refactored without checking the return value.

src/azure-cli-core/azure/cli/core/util.py
PARADISSEEKR: Calling `core/util.py/user_confirmation()` does not use its return value. After adding the return value, it is exactly the same as the logic in `flexible_server_custom_common.py/user_confirmation()`, so they are merged

qwordy (approved):
Approve appconfig part

src/azure-cli/azure/cli/command_modules/rdbms/flexible_server_custom_common.py
```diff
@@ -106,34 +105,18 @@ def flexible_firewall_rule_update_custom_func(instance, start_ip_address=None, e
...
+        user_confirmation(
             "Are you sure you want to delete the server '{0}' in resource group '{1}'".format(server_name,
-                                                                                              resource_group_name),
-            yes=yes)
-    if confirm:
+                                                                                              resource_group_name), yes=yes)
         try:
             result = client.begin_delete(resource_group_name, server_name, database_name)
         except Exception as ex:  # pylint: disable=broad-except
             logger.error(ex)
```
fengzhou-msft, line 121: Move 118-121 outside of `if not yes:`.

fengzhou-msft (approved):

src/azure-cli/azure/cli/command_modules/rdbms/flexible_server_custom_common.py
```diff
@@ -71,17 +72,15 @@ def firewall_rule_create_func(client, resource_group_name, server_name, firewall
 
 
 def firewall_rule_delete_func(client, resource_group_name, server_name, firewall_rule_name, yes=None):
-    confirm = yes
     result = None
     if not yes:
-        confirm = user_confirmation(
+        user_confirmation(
```
Juliehzl, line 77: why add this change?
I think the following delete operation will be run only when we get positive confirm, so need to check confirmation value as previous design.
Add @DaeunYim for confirmation.

fengzhou-msft: `user_confirmation` will raise exceptions if not positive confirm, so no need to check the return value. Once the function returns successfully, it must be positive confirm.

PARADISSEEKR: If there is no confirmation, `user_confirmation` raise an exception,If confirmed, the following program will execute normally, so there is no effect.

DaeunYim: Checked and confirmed. Thanks @Juliehzl for the heads up.

src/azure-cli/azure/cli/command_modules/rdbms/flexible_server_custom_mysql.py
```diff
@@ -10,15 +10,14 @@
...
-from azure.cli.core.util import CLIError, sdk_no_wait
+from azure.cli.core.util import CLIError, sdk_no_wait, user_confirmation
 from azure.cli.core.local_context import ALL
 from azure.mgmt.rdbms import mysql_flexibleservers
 from ._client_factory import get_mysql_flexible_management_client, cf_mysql_flexible_firewall_rules, \
     cf_mysql_flexible_db, cf_mysql_check_resource_availability
 from ._flexible_server_util import resolve_poller, generate_missing_parameters, create_firewall_rule, \
     parse_public_access_input, generate_password, parse_maintenance_window, get_mysql_list_skus_info, \
     DEFAULT_LOCATION_MySQL
-from .flexible_server_custom_common import user_confirmation
```
Juliehzl, line 21: then should we delete this function code?

src/azure-cli/azure/cli/command_modules/rdbms/flexible_server_custom_common.py
```diff
@@ -7,7 +7,8 @@
 from datetime import datetime
 from knack.log import get_logger
 from knack.util import CLIError
-from knack.prompting import prompt_y_n, NoTTYException
+from azure.cli.core.util import user_confirmation
+
 
 logger = get_logger(__name__)
```
fengzhou-msft, line 13:
```suggestion
from azure.cli.core.util import user_confirmation

logger = get_logger(__name__)
```

Juliehzl (approved):

PR #17816, opened 2021-04-22
[Packaging] Bump bundled python to `3.8.9` and remove more network SDK APIs for MSI

fengzhou-msft:
**Description**

The bundled Python in MSI is bumped to `3.8.9`. The source code of python is moved from ·site-packages· to `python38.zip` (which is the original format in the embeddable python). [lib2to3](https://docs.python.org/3/library/2to3.html) is removed from it as it's not needed.

This PR removes more network SDK API versions with a script to check all used versions in profile and hard-coded ones plus the default versions in `models.py`, and only keep them in MSI. All hard-coded network APIs used in `azure-cli` is refactored into one place in `_shared.py`. Note that all network related extensions are using vendored SDK, so removing a version that is higher than current used versions should not affect this `azure-cli` version working with future extensions that may use the higher version network SDK.

**Testing Guide**

**History Notes**

[Packaging] Bump bundled python to `3.8.9` in MSI.

---


jsntcy:
Looks like not all Network related extensions are using vendored_sdk, although there is vendored_sdk folder, but some places still use Network defined in CLI core. I remember some extensions are broken when Network SDK version is bumped in CLI core.

fengzhou-msft:
> Looks like not all Network related extensions are using vendored_sdk, although there is vendored_sdk folder, but some places still use Network defined in CLI core. I remember some extensions are broken when Network SDK version is bumped in CLI core.
> Please double check.

In azure-cli-extensions repo:

- A search for `def network_client_factory` showed the code all uses customized resource type with vendored SDK.

- A search for `ResourceType.MGMT_NETWORK` showed `spring-cloud` and `ssh` calls `get_mgmt_service_client` to create a network client from the profile defined in azure cli. 

- A search for `NetworkManagementClient` excluding `vendored_sdks` folders showed that they are only used in customized resource type with version from vendored SDK.

So no sign of extensions using a hard-coded API version from azure-cli installed network SDK.

jsntcy (approved):

zhoxing-ms (approved):

Juliehzl (commented):
Have you test the new build with all test scenarios?

fengzhou-msft:
> Have you test the new build with all test scenarios?

Yes. Tested locally with `setup_msi_test.ps1` which is also updated in this PR. The tests for impacted modules all passed. A few tests failed because of some generated test code causing errors like `ValueError: attempted relative import beyond top-level package`. No plan to fix that in this PR.

Juliehzl (approved):

src/azure-cli-core/azure/cli/core/profiles/_shared.py
```diff
@@ -336,6 +336,19 @@ def default_api_version(self):
 }
 
 
+# We should avoid using ad hoc API versions,
+# use the version in a profile as much as possible.
+AD_HOC_API_VERSIONS = {
+    ResourceType.MGMT_NETWORK: {
+        'vm_default_target_network': '2018-01-01',
```
qwordy, line 343: Define a const variable?

src/azure-cli/azure/cli/command_modules/vm/_vm_utils.py
```diff
@@ -28,10 +28,10 @@ def get_target_network_api(cli_ctx):
         necessarily latest, network API version is order to avoid having to re-record every test that uses VM create
         (which there are a lot) whenever NRP bumps their API version (which is often)!
     """
-    from azure.cli.core.profiles import get_api_version, ResourceType
+    from azure.cli.core.profiles import get_api_version, ResourceType, AD_HOC_API_VERSIONS
     version = get_api_version(cli_ctx, ResourceType.MGMT_NETWORK)
     if cli_ctx.cloud.profile == 'latest':
-        version = '2018-01-01'
+        version = AD_HOC_API_VERSIONS[ResourceType.MGMT_NETWORK]['vm_default_target_network']
```
qwordy, line 34: Define a const variable? VM_DEFAULT_TARGET_NETWORK = 'vm_default_target_network'

qwordy (approved):

src/azure-cli/azure/cli/command_modules/vm/_vm_utils.py
fengzhou-msft: Good suggestion, but as this is temp work, eventually this will be removed. Using hardcoded api version should not be allowed or we should support more dimensions in profile definition so it's possible to use different api versions in different modules. Leave this work to profile owner.

PR #17833, opened 2021-04-23
{Core} Work around long-running delete or purge operation with 404 status code

houk-ms:
**Description**

This PR works around the Track2 case of the long-running delete or purge operation with 404 status code when polling completes.

Track1 works around it in this PR https://github.com/Azure/azure-cli/pull/15884

**Testing Guide**

**History Notes**

---


src/azure-cli-core/azure/cli/core/commands/__init__.py
```diff
@@ -984,12 +985,13 @@ def __call__(self, poller):
...
             self.progress_bar.stop()
-            if getattr(client_exception, 'status_code', None) == 404 and 'delete' in self.cli_ctx.data['command']:
-                logger.debug('Service returned 404 on the long-running delete operation. CLI treats it as delete '
-                             'successfully but service should fix this behavior.')
+            if getattr(client_exception, 'status_code', None) == 404 and \
+               ('delete' in self.cli_ctx.data['command'] or 'purge' in self.cli_ctx.data['command']):
+                logger.debug('Service returned 404 on the long-running delete or purge operation. CLI treats it as '
+                             'delete or purge successfully but service should fix this behavior.')
                 return None
             handle_long_running_operation_exception(client_exception)
```
fengzhou-msft, line 996: Please double check whether `handle_long_running_operation_exception` can handle track 2 `HttpREsponseError` as well.

houk-ms: Yes, it could. 
Before this change, the `HttpResponseError` is raised to an outer exception handler.
After this change, it just wraps the exception and throw a CLIError, which i think doesn't matter.

houk-ms: We may add anthor except block to keep the exactly same behavior, but i think it's not really necessary.

houk-ms:
Test cases will fail if we wrap the error in advance rather than raise it. So i just keep the original behavior.

fengzhou-msft (approved):

PR #17857, opened 2021-04-25
{Core} Testing framework enhancement

qwordy:
**Description**

Add an argument `live_only` to `ScenarioTest:cmd` method.

```
def cmd(self, command, checks=None, expect_failure=False, live_only=False):
```

live_only: When it is True, command will be only executed in live mode and will not be recorded.
               It also avoids unnecessary checks of preparation commands in replay mode.
               These checks significantly aggravate the cost of API version upgrade.
               Preparation commands are used to do preparation work such as dependent resource creation so that
               the command to be tested can run.
               Preparation commands are not the targets to be tested.
               They should have their own test cases in their own modules.

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli-testsdk/azure/cli/testsdk/base.py
+++ b/src/azure-cli-testsdk/azure/cli/testsdk/base.py
@@ -162,9 +162,38 @@ class ScenarioTest(ReplayableTest, CheckerMixin, unittest.TestCase):
 
         return uuid.UUID(moniker)
 
-    def cmd(self, command, checks=None, expect_failure=False):
+    def cmd(self, command, checks=None, expect_failure=False, live_only=False):
+        """
+        :param command:
+        :param checks:
+        :param expect_failure:
+        :param live_only: When it is True, command will be only executed in live mode and will not be recorded.
+               It also avoids unnecessary checks of preparation commands in replay mode.
+               These checks significantly aggravate the cost of API version upgrade.
+               Preparation commands are used to do preparation work such as dependent resource creation so that
+               the command to be tested can run.
+               Preparation commands are not the targets to be tested.
+               They should have their own test cases in their own modules.
+        :return:
+        """
         command = self._apply_kwargs(command)
-        return execute(self.cli_ctx, command, expect_failure=expect_failure).assert_with_checks(checks)
+        if live_only:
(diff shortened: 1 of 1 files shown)


src/azure-cli-testsdk/azure/cli/testsdk/base.py
```diff
@@ -162,9 +162,38 @@ def create_guid(self):
...
+                if isinstance(self, LiveScenarioTest):
+                    return self._execute_no_recording(command, checks, expect_failure)
+        else:
+            return execute(self.cli_ctx, command, expect_failure=expect_failure).assert_with_checks(checks)
+
+    def _execute_no_recording(self, command, checks, expect_failure):
+        old_value = self.disable_recording
+        self.disable_recording = True
+        execute_result = execute(self.cli_ctx, command, expect_failure=expect_failure).assert_with_checks(checks)
+        self.disable_recording = old_value
```
jiasli, line 195: Consider using [patch.object](https://docs.python.org/3/library/unittest.mock.html#patch-object) on `self.disable_recording` for simplicity.

src/azure-cli-testsdk/azure/cli/testsdk/base.py
```diff
@@ -162,9 +162,38 @@ def create_guid(self):
...
         command = self._apply_kwargs(command)
-        return execute(self.cli_ctx, command, expect_failure=expect_failure).assert_with_checks(checks)
+        if live_only:
+            try:
+                if self.in_recording:
+                    return self._execute_no_recording(command, checks, expect_failure)
+            except AttributeError:
+                # A test might not have an in_recording attribute. Run live if it is an instance of LiveScenarioTest
+                if isinstance(self, LiveScenarioTest):
+                    return self._execute_no_recording(command, checks, expect_failure)
```
jiasli, line 187: Duplicate with L183. Consider unify them to one `if` statements.

jiasli, line 187: Duplicate with L183. Consider unify them to one `if` statements.

qwordy:
Open question: how to fetch result of a command if it is used in subsequent commands?

PR #17988, opened 2021-05-06
[AppService] `webapp log tail` Fix #17987: logging.warning call with invalid 'end' argument

### LINKED ISSUE #17987: `webapp log tail` throws TypeError
**Describe the bug**

The `appservice` module throws `TypeError` when trying to stream logs.

```shell
[ ~ ]$ az webapp log tail --name mywebapp --resource-group myresourcegroup
Exception in thread Thread-1:
  File "/usr/lib/python3.7/threading.py", line 926, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.7/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/home/nicolas/pyenvs/az/lib/python3.7/site-packages/azure/cli/command_modules/appservice/custom.py", line 2345, in _get_log
    .decode(std_encoding, errors='replace'), end='')  # each line of log has CRLF.
  File "/usr/lib/python3.7/logging/__init__.py", line 1390, in warning
    self._log(WARNING, msg, args, **kwargs)
TypeError: _log() got an unexpected keyword argument 'end'
```
As the error says, `end` [is not a valid keyword](https://docs.python.org/3/library/logging.html#logging.warning) for `_log()` (internal for `logging.warning`). It is a valid keyword for `print` though, which was being used before this change https://github.com/Azure/azure-cli/commit/680f8339ac010a89d4063566fabc5991abc8a4c#diff-76cc90184fe3a5df817ac304d5929231658bef52748a4560dabf51ee18bf8f02L2362:

```diff
-                print(chunk.decode(encoding='utf-8', errors='replace')
-                      .encode(std_encoding, errors='replace')
-                      .decode(std_encoding, errors='replace'), end='')  # each line of log has CRLF.
+                logger.warning(chunk.decode(encoding='utf-8', errors='replace')
+                               .encode(std_encoding, errors='replace')
+                               .decode(std_encoding, errors='replace'), end='')  # each line of log has CRLF.
```

**To Reproduce**

* Create an App Service through Azure Portal
* Login using `az login`
* Run `az webapp log tail --name $WEBAPP --resource-group $RESGROUP` pointing to the app service previosuly created.

**Expected behavior**

The command should stream logs and not fail.

**Environment summary**
* Install Method: `pip`
* Python version: `Python 3.7.10` (I also tried `3.8` and `3.9`)
* OS: Manjaro Linux
* Shell Type: `bash`
* CLI version:
```
[~]$ az --version
azure-cli                         2.23.0

core                              2.23.0
telemetry                          1.0.6
```

npazosmendez:
**Description**

Fixes #17987

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -2342,7 +2342,8 @@ def _get_log(url, user_name, password, log_file=None):
                 # Extra encode() and decode for stdout which does not surpport 'utf-8'
                 logger.warning(chunk.decode(encoding='utf-8', errors='replace')
                                .encode(std_encoding, errors='replace')
-                               .decode(std_encoding, errors='replace'), end='')  # each line of log has CRLF.
+                               .decode(std_encoding, errors='replace')
+                               .rstrip('\n\r'))  # each line of log has CRLF.
     r.release_conn()
 
 


src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2342,7 +2342,8 @@ def _get_log(url, user_name, password, log_file=None):
                 # Extra encode() and decode for stdout which does not surpport 'utf-8'
                 logger.warning(chunk.decode(encoding='utf-8', errors='replace')
                                .encode(std_encoding, errors='replace')
-                               .decode(std_encoding, errors='replace'), end='')  # each line of log has CRLF.
+                               .decode(std_encoding, errors='replace')
+                               .rstrip('\n\r'))  # each line of log has CRLF.
```
npazosmendez, line 2346: `.rstrip('\n\r')` might remove newlines which were intentionally logged by the app, so this is probably not ideal. But I don't know what line terminator is the extra one: `\n`, `\r`, `\r\n`?

From what I've seen, it seems to be `\n`. So replacing `rstrip` with `[:-1]` should do the job, as long as it is guaranteed that there will always be a trailing newline.

qwordy (approved):

Json2CSharp:
when will this fix be live ?

npazosmendez:
> when will this fix be live ?

@Json2CSharp according to [this comment](https://github.com/Azure/azure-cli/issues/18014#issuecomment-839530240), on 05/25/2021.

Issue #17994, opened 2021-05-07
Unable to print special characters to console output


avanigupta:
## Describe the bug

I am from the Azure App Configuration team and I work on the `az appconfig` module. I noticed that when our output contains characters like `汉` or ` 𐐷` , CLI outputs a question mark character instead: ` �`.
Moreover, if we try to print a warning message containing one of these characters, we get an exception. Although this exception doesn't block the operation, it still pollutes the output and cannot be suppressed even if we supply `--only-show-errors` argument.

This is how we print a warning:

```
from knack.log import get_logger
logger = get_logger(__name__)
logger.warning(json.dumps(record, ensure_ascii=False))
```

**Errors:**
This is the exception:
```
--- Logging error ---
Traceback (most recent call last):
  File "C:\Python38\lib\logging\__init__.py", line 1084, in emit
    stream.write(msg + self.terminator)
  File "C:\Python38\lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode character '\u6c49' in position 127: character maps to <undefined>
Call stack:
  File "C:\Python38\lib\runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Python38\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "d:\code\appconfig\azure-cli\src\azure-cli\azure\cli\__main__.py", line 48, in <module>
    exit_code = cli_main(az_cli, sys.argv[1:])
  File "d:\code\appconfig\azure-cli\src\azure-cli\azure\cli\__main__.py", line 35, in cli_main
    return cli.invoke(args)
  File "D:\Code\AppConfig\azure-cli\env\lib\site-packages\knack\cli.py", line 231, in invoke
    cmd_result = self.invocation.execute(args)
  File "d:\code\appconfig\azure-cli\src\azure-cli-core\azure\cli\core\commands\__init__.py", line 650, in execute
    results, exceptions = self._run_jobs_serially(jobs, ids)
  File "d:\code\appconfig\azure-cli\src\azure-cli-core\azure\cli\core\commands\__init__.py", line 720, in _run_jobs_serially
    results.append(self._run_job(expanded_arg, cmd_copy))
  File "d:\code\appconfig\azure-cli\src\azure-cli-core\azure\cli\core\commands\__init__.py", line 691, in _run_job
[...]

PR #17996, opened 2021-05-07
{Core} Always use UTF-8 for command metadata log file encoding

jiasli:
Fix https://github.com/Azure/azure-cli/issues/17994
Adopt https://github.com/microsoft/knack/pull/247

## Context

On Windows with English as the system language (without UTC-8 enabled), the system encoding by default is `cp1252` (Western Europe), and Python will use `cp1252` as the file encoding by default. 

Writing Unicode characters like `汉字` to log file results in error:

```
  File "C:\Users\xxx\AppData\Local\Programs\Python\Python38\lib\logging\__init__.py", line 1084, in emit
    stream.write(msg + self.terminator)
  File "C:\Users\xxx\AppData\Local\Programs\Python\Python38\lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode characters in position 87-88: character maps to <undefined>
```

## Change

This PR forces command metadata log files to use UTF-8 so that `az feedback` works on non-UTF-8 systems as well.

### DIFF
--- a/src/azure-cli-core/azure/cli/core/azlogging.py
+++ b/src/azure-cli-core/azure/cli/core/azlogging.py
@@ -30,7 +30,7 @@ import datetime
 from azure.cli.core.commands.events import EVENT_INVOKER_PRE_CMD_TBL_TRUNCATE
 
 from knack.events import EVENT_CLI_POST_EXECUTE
-from knack.log import CLILogging, get_logger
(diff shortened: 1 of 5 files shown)


yonzhan:
Core

yonzhan (approved):
LGTM

evelyn-ys (approved):

PR #18003, opened 2021-05-07
{Pylint} Ignore `tests` folders

jiasli:
**Description**

Require https://github.com/Azure/azure-cli-dev-tools/pull/301

Due to https://github.com/PyCQA/pylint/issues/2686, `test.*` directories are not ignored. This causes files like `src/azure-cli/azure/cli/command_modules/billing/tests/__init__.py` to be checked which is not desired.

However, `ignore=tests` works because `tests` directories in Azure CLI are actually modules (because they contain `__init__.py`), so `pylint` goes through a different code path of `modutils.get_module_files`:

https://github.com/PyCQA/pylint/blob/8f84dec53d53b8297ef753277fc1d820becd8428/pylint/lint/expand_modules.py#L114-L121

```py
        if has_init or is_namespace or is_directory:
            for subfilepath in modutils.get_module_files(
                os.path.dirname(filepath), ignore_list, list_all=is_namespace
            ):
                if filepath == subfilepath:
                    continue
                if _basename_in_ignore_list_re(
                    os.path.basename(subfilepath), ignore_list_re
```

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -1,5 +1,6 @@
 [MASTER]
-ignore-patterns=test_*,azure_devops_build_*
+ignore=tests,vendored_sdks,privates
+ignore-patterns=test.*,azure_devops_build.*
 reports=no
 
 [MESSAGES CONTROL]
--- a/pylintrc
+++ b/pylintrc
@@ -1,5 +1,5 @@
 [MASTER]
-ignore=tests,vendored_sdks,privates
+ignore=tests,generated,vendored_sdks,privates
 ignore-patterns=test.*,azure_devops_build.*
 reports=no
 
--- a/.flake8
+++ b/.flake8
@@ -20,3 +20,4 @@ exclude =
     build_scripts
     */grammar/
     vendored_sdks
+    tests


pylintrc
```diff
@@ -1,5 +1,6 @@
 [MASTER]
-ignore-patterns=test_*,azure_devops_build_*
+ignore=tests,vendored_sdks,privates
+ignore-patterns=test.*,azure_devops_build.*
```
jiasli, line 3: The original config value `test_*` is wrong ❌. According to http://pylint.pycqa.org/en/latest/technical_reference/features.html

> Files or directories matching the **regex patterns** are skipped. The regex matches against base names, not paths.

It should be regex, instead of wildcard. 

`test_*` is used to match `test________...`. It works because `_*` doesn't match anything at all and only `test` is used to do the regex match.

yonzhan:
Pylint

pylintrc
```diff
@@ -1,5 +1,6 @@
 [MASTER]
-ignore-patterns=test_*,azure_devops_build_*
+ignore=tests,vendored_sdks,privates
```
evelyn-ys, line 2: Why no `generated` like extension repo?

jiasli: Nice catch! Fixed.

jiasli:
```suggestion
ignore=tests,generated,vendored_sdks,privates
```

evelyn-ys (approved):

PR #18012, opened 2021-05-08
{Packaging} Remove `pypiwin32` and bump `pywin32` to 300

jiasli:
### Remove `pypiwin32`

`pypiwin32` is already deprecated: https://pypi.org/project/pypiwin32

Per the source code `setup.py` of `pypiwin32`:

https://files.pythonhosted.org/packages/13/e8/4f38eb30c4dae36634a53c5b2cd73b517ea3607e10d00f61f2494449cec0/pypiwin32-223.tar.gz

```py
from setuptools import setup

setup(
    name='pypiwin32',
    version='223',
    install_requires='pywin32>=223',
)
```

It simply installs `pywin32>=223`, which is already in `requirements.py3.windows.txt`, so we should remove it.

### Bump `pywin32` to 300

`pywin32` 300 follows PEP 440 to import DLLs (https://github.com/mhammond/pywin32/pull/1651) which is a nice change.

### DIFF
--- a/src/azure-cli/requirements.py3.windows.txt
+++ b/src/azure-cli/requirements.py3.windows.txt
@@ -121,11 +121,10 @@ PyGithub==1.38
 PyJWT==1.7.1
 PyNaCl==1.4.0
 pyOpenSSL==19.0.0
-pypiwin32==223
 pyreadline==2.1
 python-dateutil==2.8.0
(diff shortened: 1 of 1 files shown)


yonzhan:
Packaging

evelyn-ys (approved):

PR #18013, opened 2021-05-08
{Pylint} Fix `unnecessary-comprehension`

jiasli:
`unnecessary-comprehension` should be removed according to https://github.com/PyCQA/pylint/issues/2905:

For example:

* `list(iterable)` instead of `[x for x in iterable]`
* `dict(some_dict)` instead of `{k: v for k, v in some_dict}`
* `set(iterable)` instead of `{e for e in iterable}`


src/azure-cli-core/azure/cli/core/extension/__init__.py
```diff
@@ -330,7 +330,7 @@ def get_extensions(ext_type=None):
     elif not isinstance(ext_type, list):
         ext_type = [ext_type]
     for t in ext_type:
-        extensions.extend([ext for ext in t.get_all()])
+        extensions.extend(t.get_all())
```
jiasli, line 333: `extend` accepts an `iterable`, so no need to convert `t.get_all()` to a list first.

- https://docs.python.org/3/library/stdtypes.html#mutable-sequence-types
- https://docs.python.org/3/tutorial/datastructures.html#more-on-lists

src/azure-cli/azure/cli/command_modules/hdinsight/custom.py
```diff
@@ -68,7 +68,7 @@ def create_cluster(cmd, client, cluster_name, resource_group_name, cluster_type,
 
     if component_version:
         # See validator
-        component_version = {c: v for c, v in [version.split('=') for version in component_version]}
+        component_version = dict([version.split('=') for version in component_version])
```
jiasli, line 71: The constructor of `dict` takes `iterables` of 2-element `iternables`.

https://docs.python.org/3/library/stdtypes.html#dict

yonzhan:
Pylint

kairu-ms (approved):

qwordy (approved):

evelyn-ys (approved):

qwordy (approved):

PR #18035, opened 2021-05-11
{Core} Change resource group preparer in testsdk

FumingZhang:
**Description**

Modify resource group preparer, replace the default value of location with the value of 
 'AZURE_CLI_TEST_DEV_RESOURCE_GROUP_LOCATION', so that we can leverage the preparer of testsdk to automatically create and delete resource groups at the **specified location** (*rather than hardcoded default location*) when performing live tests.

Would like to know if there are other ways to change the default value of location in the preparer.

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli-testsdk/azure/cli/testsdk/preparers.py
+++ b/src/azure-cli-testsdk/azure/cli/testsdk/preparers.py
@@ -64,7 +64,10 @@ class ResourceGroupPreparer(NoTrafficRecordingPreparer, SingleValueReplacer):
         self.key = key
 
         self.dev_setting_name = os.environ.get(dev_setting_name, None)
-        self.dev_setting_location = os.environ.get(dev_setting_location, location)
+        # use environment variable to modify the default value of location
+        self.dev_setting_location = os.environ.get(dev_setting_location, None)
+        if self.dev_setting_location is not None and self.dev_setting_location != "":
+            self.location = self.dev_setting_location
 
     def create_resource(self, name, **kwargs):
         if self.dev_setting_name:
--- a/src/azure-cli-testsdk/azure/cli/testsdk/preparers.py
+++ b/src/azure-cli-testsdk/azure/cli/testsdk/preparers.py
@@ -66,7 +66,7 @@ class ResourceGroupPreparer(NoTrafficRecordingPreparer, SingleValueReplacer):
         self.dev_setting_name = os.environ.get(dev_setting_name, None)
         # use environment variable to modify the default value of location
         self.dev_setting_location = os.environ.get(dev_setting_location, None)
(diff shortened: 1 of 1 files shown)


Juliehzl:
Hi @FumingZhang, you could just pass location with expected value to override default one.
https://github.com/Azure/azure-cli/blob/5cf2ac3683ea1cb6103607106c06d6938bc0f61d/src/azure-cli/azure/cli/command_modules/storage/tests/latest/test_storage_account_scenarios.py#L1306

qwordy:
I don't agree with this PR. `location` in constructor is override by environment variable even I set it. I am astonished. https://en.wikipedia.org/wiki/Principle_of_least_astonishment

Juliehzl:
echo @qwordy. Just though about that with your change, all resource group locations will be changed even developers set specific location value in their test cases. Such change will also cause test failures because some tests only can be run in target region.

FumingZhang:
> I don't agree with this PR. `location` in constructor is override by environment variable even I set it. I am astonished. https://en.wikipedia.org/wiki/Principle_of_least_astonishment

The premise of the location being overridden is that you manually set the environment variable 'AZURE_CLI_TEST_DEV_RESOURCE_GROUP_LOCATION', otherwise nothing will change.

FumingZhang:
My requirement is to run test cases in all locations where services are available for testing. As far as I know, there is currently no way to change the location value of the preparer used by the test case when testing through azdev. If there is a way to change the value of location when calling the test case, I would be happy to withdraw this pr.

src/azure-cli-testsdk/azure/cli/testsdk/preparers.py
```diff
@@ -64,7 +64,10 @@ def __init__(self, name_prefix='clitest.rg',
         self.key = key
 
         self.dev_setting_name = os.environ.get(dev_setting_name, None)
-        self.dev_setting_location = os.environ.get(dev_setting_location, location)
+        # use environment variable to modify the default value of location
+        self.dev_setting_location = os.environ.get(dev_setting_location, None)
+        if self.dev_setting_location is not None and self.dev_setting_location != "":
```
kairu-ms, line 69:
```suggestion
        if self.dev_setting_location:
```

FumingZhang:
Hi, any update?

qwordy:
@FumingZhang You can fork Azure CLI and customize it.

FumingZhang:
#18068 
I have just filled another [PR](https://github.com/Azure/azure-cli/pull/18068), and in that one, I create a custom preparer and replace the ResourceGroupPreparer provided by testsdk with it. Is this an acceptable way?

qwordy:
@FumingZhang This is a promising approach.

FumingZhang:
Found a workaround, closing this pr.

PR #17783, opened 2021-04-20
[ARM] Migrate resource to track2 SDK

zhoxing-ms:
**Description**

**This PR upgrades the Python SDK version of `azure-mgmt-resource` from `12.1.0` to `16.1.0`**

**Here are the main changes for resource migration to track2:**
1. LRO operations are renamed to `begin_xxx_xxx`. For example in track1, the name is `create_or_update`; while in track2, the name is `begin_create_or_update`.
2. When `api-version` is less than `2019-10-01`, the parameters of all deployment operations are changed from `DeploymentProperties` to `Deployment`, which is consistent with the SDK after version `2019-10-01`.
3. The objects of input parameters for some methods are further encapsulated:
- `begin_what_if()`: `DeploymentWhatIfProperties` --> `DeploymentWhatIf`
- `begin_what_if_at_subscription_scope()/...`: `DeploymentWhatIfProperties` --> `ScopedDeploymentWhatIf`
- `begin_export_template()`: flattened parameters --> `ExportTemplateRequest`
- `begin_move_resources()`: flattened parameters --> `ResourcesMoveInfo`
- `links_client.create_or_update()`: `ResourceLinkProperties` --> `ResourceLink`
- `tags.create_or_update_at_scope()`: flattened parameters --> `TagsResource`
- `tags.update_at_scope()`: flattened parameters --> `TagsPatchResource`
4. In method `get()` of `ResourcesOperations`, some parameters are passed in through `**kwargs` instead
5. `CloudError` and `HttpOperationError` are replaced by `HttpResponseError`
6. In track 2, the logic corresponding to the class name of Codegen generating operations changes, so `DeploymentOperations` becomes `DeploymentOperationsOperations`
For more details, please refer to this issue: https://github.com/Azure/autorest.python/issues/910#issuecomment-809516272
7. `config` in management client becomes a private variable `_config`.
8. In method `policy_definitions.create_or_update_at_management_group()`, the positions of parameters `management_group`and `parameters` are reversed.
9. `generate_client_request_id` in `ResourceManagementClientConfiguration` no longer needs, SDK automatically add `x-ms-request-id` now through the `RequestIdPolicy`
10. The `accept_language` is no longer supported in `ResourceManagementClientConfiguration`, so the fixed value `en-US` in track 2 is hard coded.
11. The pipeline and policy in management client migrated from `msrest` to `azure-core`. And when doing ARM deployment, the pipeline needs to add custom policy `JsonCTemplatePolicy()` to support JSON templates.
[...]

PR #18047, opened 2021-05-12
{Core} Stop using vendored SDK in core

jiasli:
**Description**

Stop using vendored subscription SDK in core as [`azure-mgmt-resource`](https://pypi.org/project/azure-mgmt-resource/) has been migrated to Track 2 (https://github.com/Azure/azure-cli/pull/17783).


yonzhan:
Core

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -75,7 +75,7 @@
 
 _AZ_LOGIN_MESSAGE = "Please run 'az login' to setup account."
 
-_USE_VENDORED_SUBSCRIPTION_SDK = True
+_USE_VENDORED_SUBSCRIPTION_SDK = False
```
zhoxing-ms, line 78: Since the vendor SDK has been deleted, why do we still keep the logic of this switch?

jiasli: In case there is a track 3 someday. 😏

zhoxing-ms: 😂😂

zhoxing-ms (approved):

evelyn-ys (approved):

PR #18048, opened 2021-05-12
{Pylint} Fix `unrecognized-inline-option`

jiasli:
**Description**

The correct format to suppress a checker is 

```py
# pylint: disable=unused-import
```

These formats are invalid and never worked, but Pylint didn't give error, meaning they are redundant. so we should remove them.


yonzhan:
Pylint

zhoxing-ms (approved):

evelyn-ys (approved):

PR #18057, opened 2021-05-12
[Fish Shell Auto Completions]

mattpasco-microsoft:
**Description**
This adds Fish Shell auto completions with descriptions of sub commands and flags from the Azure CLI. There has been no changes to the actual code or functionality of the Azure CLI.

**Testing Guide**
Since this will not affect the build or release of the Azure CLI. No testing is required.

**History Notes**

[create-fish-complete.py] create fish `complete` statements

---


mattpasco-microsoft:
az-cli-fish

yonzhan:
Fish Shell Auto Completions

src/azure-cli/fish-shell/create_fish_complete.py
```diff
@@ -0,0 +1,134 @@
...
+''')
+
+    is_az = True
+
+    while len(subcommands) != 0:
+
+        # Grab last item in queue (DFS)
+        command = subcommands.pop()
+        # Query back help page
+        response = subprocess.run(command + ['--help'], stdout=subprocess.PIPE).stdout.decode('utf-8')
```
qwordy, line 65: How is the performance? How about we get the candidates in advance and build the index just like modern IDE does.

qwordy:
The demo is awesome.

src/azure-cli/fish-shell/create_fish_complete.py
mattpasco-microsoft: Performance of the python script is not great. It takes on average 20m to generate the completions. The reason I went this route is it requires no code changes when a module or its help doc is changed. Looking at the `help.py` for each command module might be significantly faster but looks like wed still need to do some parsing.

That said once the completions are generated they are faster so I would say its a compile vs runtime speed. And this script only needs to be run once where itll generate a `az.fish` file the user can source.

Here is a snippet of what the python script generates: 
```sh
complete -c az -d "Manage Azure subscription information." -n "__fish_az_no_subcommand" -f -a "account"
complete -c az -d "Manage private registries with Azure Container" -n "__fish_az_no_subcommand" -f -a "acr"
complete -c az -d "Manage Azure Active Directory Graph entities needed" -n "__fish_az_no_subcommand" -f -a "ad"
complete -c az -d "Manage Azure Advisor." -n "__fish_az_no_subcommand" -f -a "advisor"
complete -c az -d "Manage Azure Kubernetes Services." -n "__fish_az_no_subcommand" -f -a "aks"
complete -c az -d "Manage Azure Media Services resources." -n "__fish_az_no_subcommand" -f -a "ams"
complete -c az -d "Manage Azure API Management services." -n "__fish_az_no_subcommand" -f -a "apim"
```

That said if we are concerned about "compile-time" speed, do we have the help docs in a better format that the `help.py` (eg. yaml, json, etc)

qwordy: @mattpasco-microsoft Oh! Sorry, I thought `-h` is running at runtime. I like your solution. We run this file only once.

mattpasco-microsoft: Correct this will only run once, ideally in the build time!

borqosky:
is there any progress with it guys ?

PR #18068, opened 2021-05-13
{AKS} replace ResourceGroupPreparer with custom preparer in aks command test

FumingZhang:
**Description**

[Related PR #18035](https://github.com/Azure/azure-cli/pull/18035)

My requirement is to run test cases in all locations where services are available for testing. As far as I know, there is currently no way to change the default location value of the ResourceGroupPreparer provided by azure.cli.testsdk. Thus I would like to implement a custom preparer to achieve this and it will only be used for testing aks commands (acs module). No impact on other modules. Is this acceptable?

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acs/tests/latest/custom_preparers.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/tests/latest/custom_preparers.py
@@ -8,8 +8,8 @@ from datetime import datetime
 
 from azure.cli.testsdk import CliTestError
 from azure.cli.testsdk.preparers import NoTrafficRecordingPreparer
-from azure_devtools.scenario_tests import SingleValueReplacer
 from azure.cli.testsdk.reverse_dependency import get_dummy_cli
+from azure_devtools.scenario_tests import SingleValueReplacer
 
 
 class AKSCustomResourceGroupPreparer(NoTrafficRecordingPreparer, SingleValueReplacer):
--- a/src/azure-cli/azure/cli/command_modules/acs/tests/latest/custom_preparers.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/tests/latest/custom_preparers.py
@@ -4,61 +4,23 @@
 # --------------------------------------------------------------------------------------------
 
 import os
-from datetime import datetime
 
-from azure.cli.testsdk import CliTestError
-from azure.cli.testsdk.preparers import NoTrafficRecordingPreparer
-from azure.cli.testsdk.reverse_dependency import get_dummy_cli
-from azure_devtools.scenario_tests import SingleValueReplacer
+from azure.cli.testsdk.preparers import ResourceGroupPreparer
 
 
-class AKSCustomResourceGroupPreparer(NoTrafficRecordingPreparer, SingleValueReplacer):
+class AKSCustomResourceGroupPreparer(ResourceGroupPreparer):
     def __init__(self, name_prefix='clitest.rg',
                  parameter_name='resource_group',
                  parameter_name_for_location='resource_group_location', location='westus',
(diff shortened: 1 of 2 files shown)


FumingZhang:
#18035

src/azure-cli/azure/cli/command_modules/acs/tests/latest/custom_preparers.py
```diff
@@ -0,0 +1,62 @@
...
+        self.location = location
+        self.parameter_name = parameter_name
+        self.parameter_name_for_location = parameter_name_for_location
+        self.key = key
+
+        self.dev_setting_name = os.environ.get(dev_setting_name, None)
+        # use environment variable to modify the default value of location
+        self.dev_setting_location = os.environ.get(dev_setting_location, None)
+        if self.dev_setting_location:
+            self.location = self.dev_setting_location
```
fengzhou-msft, line 37:
```suggestion
        if self.dev_setting_location:
            self.location = self.dev_setting_location
        else:
            self.dev_setting_location = location
```
Otherwise it may break tests when `dev_setting_name` is set and `dev_setting_location` is not based on the logic in `create_resource()`

src/azure-cli/azure/cli/command_modules/acs/tests/latest/test_aks_commands.py
```diff
@@ -30,7 +31,7 @@ def _process_sp_name(sp_name):
 class AzureKubernetesServiceScenarioTest(ScenarioTest):
 
     @AllowLargeResponse()
-    @ResourceGroupPreparer(random_name_length=17, name_prefix='clitest', location='westus2')
+    @AKSCustomResourceGroupPreparer(random_name_length=17, name_prefix='clitest', location='westus2')
```
fengzhou-msft, line 34: Why not change the value of `location='westus2'` in original code to use a location that works for the test? Using a environment variable hides the information about the valid location value.

FumingZhang: I'm going to deploy a runner to test aks commands with azure-cli in all locations where services are available and monitor the behavior of that. So I need to change the resource group location programmatically and I do not find other ways to change the default value of this decorator.

FumingZhang: Any suggestions to programmatically change the default value of location in the decorator are extremely welcome! Currently, I only come up with the method of passing an environment variable. If the ```AZURE_CLI_TEST_DEV_RESOURCE_GROUP_LOCATION``` variable is not manually set, it will not have any impact on existing tests.

src/azure-cli/azure/cli/command_modules/acs/tests/latest/custom_preparers.py
```diff
@@ -0,0 +1,64 @@
...
+import os
+from datetime import datetime
+
+from azure.cli.testsdk import CliTestError
+from azure.cli.testsdk.preparers import NoTrafficRecordingPreparer
+from azure.cli.testsdk.reverse_dependency import get_dummy_cli
+from azure_devtools.scenario_tests import SingleValueReplacer
+
+
+class AKSCustomResourceGroupPreparer(NoTrafficRecordingPreparer, SingleValueReplacer):
```
andyliuliming, line 15: I would suggest to inherit from ResourceGroupPreparer.

FumingZhang: Sure, changed to inherit in commit [e802a6b](https://github.com/Azure/azure-cli/pull/18068/commits/e802a6b260c8f179e54a2892dee09f0a90f86c92).

yonzhan:
AKS

PR #18089, opened 2021-05-14
{KeyVault} Support SKR 7.3-preview with vendor SDK

houk-ms:
This PR is used to generate the private package for SKR API version 7.3-preview, with vendored SDK.


yonzhan:
@evelyn-ys for awareness

PR #18092, opened 2021-05-14
[ACR] az acr repository: Add teleportEnabled flag for repository update

shahzzzam:
**Description**

Adds the teleportEnabled functionality to the Repository update. 

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acr/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/acr/_help.py
@@ -488,9 +488,9 @@ examples:
   - name: Update the attributes of the image referenced by digest 'hello-world@sha256:abc123' to disable write operation.
     text: az acr repository update -n MyRegistry --image hello-world@sha256:abc123 --write-enabled false
   - name: Update the attributes of the Repository to disable Teleport feature.
-    text: az acr repository update -n MyRegistry --teleport-enabled false
+    text: az acr repository update -n MyRegistry --repository hello-world --teleport-enabled false
   - name: Update the attributes to enable Teleport feature on all new images pushed to the Repository. Enabling this improves container startup performance on Teleport-enabled hosts.
-    text: az acr repository update -n MyRegistry --teleport-enabled true
+    text: az acr repository update -n MyRegistry --repository hello-world --teleport-enabled true
 """
 
 helps['acr run'] = """
--- a/src/azure-cli/azure/cli/command_modules/acr/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/acr/_help.py
@@ -489,7 +489,7 @@ examples:
     text: az acr repository update -n MyRegistry --image hello-world@sha256:abc123 --write-enabled false
   - name: Update the attributes of the Repository to disable Teleport feature.
     text: az acr repository update -n MyRegistry --teleport-enabled false
-  - name: Update the attributes to enable Teleport feature on all new images in the Repository, which improves container startup performance.
+  - name: Update the attributes to enable Teleport feature on all new images pushed to the Repository. Enabling this improves container startup performance on Teleport-enabled hosts.
     text: az acr repository update -n MyRegistry --teleport-enabled true
 """
 
--- a/src/azure-cli/azure/cli/command_modules/acr/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/acr/_help.py
@@ -487,6 +487,10 @@ examples:
     text: az acr repository update -n MyRegistry --image hello-world:latest --write-enabled false
   - name: Update the attributes of the image referenced by digest 'hello-world@sha256:abc123' to disable write operation.
     text: az acr repository update -n MyRegistry --image hello-world@sha256:abc123 --write-enabled false
+  - name: Update the attributes of the Repository to disable Teleport feature.
+    text: az acr repository update -n MyRegistry --teleport-enabled false
+  - name: Update the attributes to enable Teleport feature on all new images in the Repository, which improves container startup performance.
+    text: az acr repository update -n MyRegistry --teleport-enabled true
 """
 
 helps['acr run'] = """
--- a/src/azure-cli/azure/cli/command_modules/acr/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/acr/_params.py
@@ -131,7 +131,7 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('list_enabled', help='Indicates whether this item shows in list operation results.', arg_type=get_three_state_flag())
         c.argument('read_enabled', help='Indicates whether read operation is allowed.', arg_type=get_three_state_flag())
         c.argument('write_enabled', help='Indicates whether write or delete operation is allowed.', arg_type=get_three_state_flag())
-        c.argument('teleport_enabled', help='Indicates whether Teleport feature is enabled on the repository.', arg_type=get_three_state_flag(), is_preview=True)
+        c.argument('teleport_enabled', help='Indicates whether Teleport feature is enabled on the repository. Please see https://aka.ms/acr/teleport', arg_type=get_three_state_flag(), is_preview=True)
 
     with self.argument_context('acr repository untag') as c:
         c.argument('image', options_list=['--image', '-t'], help="The name of the image. May include a tag in the format 'name:tag'.")
--- a/src/azure-cli/azure/cli/command_modules/acr/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/acr/_params.py
@@ -131,7 +131,7 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('list_enabled', help='Indicates whether this item shows in list operation results.', arg_type=get_three_state_flag())
         c.argument('read_enabled', help='Indicates whether read operation is allowed.', arg_type=get_three_state_flag())
         c.argument('write_enabled', help='Indicates whether write or delete operation is allowed.', arg_type=get_three_state_flag())
-        c.argument('teleport_enabled', help='Indicates whether Teleport feature is enabled on the repository.', arg_type=get_three_state_flag())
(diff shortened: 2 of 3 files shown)


yonzhan:
acr

northtyphoon:
@estebanreyl

shahzzzam:
@yonzhan can you re-run the checks?

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -487,6 +487,10 @@
     text: az acr repository update -n MyRegistry --image hello-world:latest --write-enabled false
   - name: Update the attributes of the image referenced by digest 'hello-world@sha256:abc123' to disable write operation.
     text: az acr repository update -n MyRegistry --image hello-world@sha256:abc123 --write-enabled false
+  - name: Update the attributes of the Repository to disable Teleport feature.
+    text: az acr repository update -n MyRegistry --teleport-enabled false
+  - name: Update the attributes to enable Teleport feature on all new images in the Repository, which improves container startup performance.
```
estebanreyl, line 492: I am not sure about the phrasing here because it sounds like by enabling the flag container startup performance is automatically improved which is not the case always. That's only true when teleport is enabled and the customer runs on a teleport client. Id probably rewrite it as: 

Update the attributes of the Repository to enable Teleport feature on new images pushed to the Repository. Enabling this improves container startup performance on teleport configured clients.

Out of curiosity what happens if we mark something as Teleport Enabled, push some images and then briefly set it to disabled, after which we enable it again, are the previously enabled images expected to be teleport enabled?

estebanreyl (commented):
Pretty straightforward. It looks good to me, there's just one phrase I think we could improve

src/azure-cli/azure/cli/command_modules/acr/_help.py
shahzzzam: Thanks @esteban, I used this description from the swagger

https://github.com/shahzzzam/azure-rest-api-specs/blob/c9220138814fd21766c643c8dcf62d976bb83034/specification/containerregistry/data-plane/Microsoft.ContainerRegistry/preview/2019-08-15/containerregistry.json#L2082

I kept the same for consistency purposes. 

Your description is definitely more accurate that it should only enhance perf on teleport-enabled clients. I will use the term "teleport-enabled hosts" instead of teleport configured clients

shahzzzam: re: enable disable
we delete the shares on disable operation. So once you enable it, it will only teleport new images.

shahzzzam:
@esteban can you approve?

shahzzzam:
@northtyphoon can u please approve?

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -131,6 +131,7 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('list_enabled', help='Indicates whether this item shows in list operation results.', arg_type=get_three_state_flag())
         c.argument('read_enabled', help='Indicates whether read operation is allowed.', arg_type=get_three_state_flag())
         c.argument('write_enabled', help='Indicates whether write or delete operation is allowed.', arg_type=get_three_state_flag())
+        c.argument('teleport_enabled', help='Indicates whether Teleport feature is enabled on the repository.', arg_type=get_three_state_flag())
```
northtyphoon, line 134: Is it possible to mark the parameter as preview? If not, can the help text somehow indicate it's a preview feature?

rosanch: add the flag is_preview=True,

src/azure-cli/azure/cli/command_modules/acr/repository.py
```diff
@@ -265,7 +265,8 @@ def acr_repository_update(cmd,
                           delete_enabled=None,
                           list_enabled=None,
                           read_enabled=None,
-                          write_enabled=None):
+                          write_enabled=None,
+                          teleport_enabled=None):
```
rosanch, line 269: is there no support for create as well?

src/azure-cli/azure/cli/command_modules/acr/_params.py
rosanch: Specify default falue

src/azure-cli/azure/cli/command_modules/acr/repository.py
```diff
@@ -284,6 +285,10 @@ def acr_repository_update(cmd,
         json_payload.update({
             'writeEnabled': write_enabled
         })
+    if teleport_enabled is not None:
+        json_payload.update({
+            'teleportEnabled': teleport_enabled
```
rosanch, line 290: is the value True false? Not enabled disabled?

rosanch: can you add this to an example in testing or add a new command to test it?
You need to regenerate the records after it.

rosanch (changes requested):

src/azure-cli/azure/cli/command_modules/acr/_params.py
shahzzzam: the default value should not be false. We don't want to set it to `false` when the user is changing other properties.

src/azure-cli/azure/cli/command_modules/acr/repository.py
shahzzzam: what do you mean? create repository goes through other API

src/azure-cli/azure/cli/command_modules/acr/repository.py
shahzzzam: It's three state, value can be `true`, `false` or `null`

shahzzzam: What testing? is it unit tests? 
We are going to add integrations tests using our CIs

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -487,6 +487,10 @@
     text: az acr repository update -n MyRegistry --image hello-world:latest --write-enabled false
   - name: Update the attributes of the image referenced by digest 'hello-world@sha256:abc123' to disable write operation.
     text: az acr repository update -n MyRegistry --image hello-world@sha256:abc123 --write-enabled false
+  - name: Update the attributes of the Repository to disable Teleport feature.
+    text: az acr repository update -n MyRegistry --teleport-enabled false
```
northtyphoon, line 491: Can you add the repository parameter to make the sample command complete?

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -131,6 +131,7 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('list_enabled', help='Indicates whether this item shows in list operation results.', arg_type=get_three_state_flag())
         c.argument('read_enabled', help='Indicates whether read operation is allowed.', arg_type=get_three_state_flag())
         c.argument('write_enabled', help='Indicates whether write or delete operation is allowed.', arg_type=get_three_state_flag())
+        c.argument('teleport_enabled', help='Indicates whether Teleport feature is enabled on the repository.', arg_type=get_three_state_flag(), is_preview=True)
```
northtyphoon, line 134: Can you add a link to https://aka.ms/acr/teleport?

src/azure-cli/azure/cli/command_modules/acr/repository.py
```diff
@@ -284,6 +285,10 @@ def acr_repository_update(cmd,
         json_payload.update({
             'writeEnabled': write_enabled
         })
+    if teleport_enabled is not None:
```
northtyphoon, line 288: Does the parameter only apply on repository? If the image parameter is `hello-world:latest`, what is the behavior?

src/azure-cli/azure/cli/command_modules/acr/_help.py
shahzzzam: oops good catch :D

northtyphoon (approved):
LGTM

src/azure-cli/azure/cli/command_modules/acr/repository.py
rosanch: I meant for the CLI. There is a test folder in ACR this is run when you try to submit changes. But as you say probably the CI are enough.

rosanch (approved):

shahzzzam:
@Juliehzl  please merge

src/azure-cli/azure/cli/command_modules/acr/repository.py
```diff
@@ -284,6 +285,12 @@ def acr_repository_update(cmd,
...
+        if not repository:
+            raise CLIError("--teleport-enabled can only be used with --repository.")
```
Juliehzl, line 290: recommend to use error type in azclierror, such as 
**https://github.com/Azure/azure-cli/blob/3eea7fd773722b7a3c40b8115f29a3bad39b70bc/src/azure-cli-core/azure/cli/core/azclierror.py#L157-L161**

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -131,6 +131,7 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('list_enabled', help='Indicates whether this item shows in list operation results.', arg_type=get_three_state_flag())
         c.argument('read_enabled', help='Indicates whether read operation is allowed.', arg_type=get_three_state_flag())
         c.argument('write_enabled', help='Indicates whether write or delete operation is allowed.', arg_type=get_three_state_flag())
+        c.argument('teleport_enabled', help='Indicates whether Teleport feature is enabled on the repository. Please see https://aka.ms/acr/teleport', arg_type=get_three_state_flag(), is_preview=True)
```
Juliehzl, line 134: could you also add test for the parameter?

northtyphoon:
We will revisit it.

PR #18119, opened 2021-05-18
{Pylint} Fix `self-assigning-variable`

jiasli:
Per https://pycodequ.al/docs/pylint-messages/w0127-self-assigning-variable.html, no need to assigned a variable to itself.

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/sqlvm/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/sqlvm/custom.py
@@ -127,13 +127,10 @@ def sqlvm_group_update(instance, domain_fqdn=None, cluster_operator_account=None
 def sqlvm_aglistener_create(client, cmd, availability_group_listener_name, sql_virtual_machine_group_name,
                             resource_group_name, availability_group_name, ip_address, subnet_resource_id,
                             load_balancer_resource_id, probe_port, sql_virtual_machine_instances, port=1433,
-                            public_ip_address_resource_id=None, vnet_name=None):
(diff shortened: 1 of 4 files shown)


yonzhan:
Pylint

idear1203 (approved):
Approve for Synapse side

src/azure-cli/azure/cli/command_modules/sqlvm/custom.py
```diff
@@ -127,13 +127,10 @@ def sqlvm_group_update(instance, domain_fqdn=None, cluster_operator_account=None
 def sqlvm_aglistener_create(client, cmd, availability_group_listener_name, sql_virtual_machine_group_name,
                             resource_group_name, availability_group_name, ip_address, subnet_resource_id,
                             load_balancer_resource_id, probe_port, sql_virtual_machine_instances, port=1433,
-                            public_ip_address_resource_id=None, vnet_name=None):
+                            public_ip_address_resource_id=None, vnet_name=None):  # pylint: disable=unused-argument
```
qwordy, line 130: If `vnet_name` is not used, can we remove it?

jiasli: If we remove it, it will be a break change to the upper-level command `az sql vm group ag-listener create` as `--vnet-name` will be removed. I only fix the linter error, so prefer not to change it.

evelyn-ys (approved):

qwordy (approved):

PR #18203, opened 2021-05-24
[AppService] BREAKING CHANGE: az webapp deployment source config: Remove vsts-cd-manager

### LINKED ISSUE #14821: webapp : Migrate vsts dependency to azure-devops

Currently `appservice` depends on `vsts` which is not maintained anymore. There will be no further support for this package if there are bugs for feature requests. The `vsts` package is migrated to [`azure-devops`](https://github.com/microsoft/azure-devops-python-api).

This is a request to migrate `vsts` to `azure-devops` in CLI side.

**Additional context**

calvinsID:
**Description**
Part of #14821. This removes vsts-cd-manager package only -  still need to remove vsts package

Remove vsts-cd-manager dependency
All old VSTS users were moved to Azure Devops, so the VSTS portion of `az webapp/functionapp deployment source config` should be unused

**Testing Guide**
`az webapp/functionapp deployment source config` shouldn't have the VSTS portion

**History Notes**

[AppService] BREAKING CHANGE: az webapp deployment source config: Remove vsts-cd-manager
[AppService] BREAKING CHANGE: az functionapp deployment source config: Remove vsts-cd-manager

---


src/azure-cli/azure/cli/command_modules/appservice/_params.py
```diff
@@ -256,38 +256,8 @@ def load_arguments(self, _):
...
-                                               'Python 3.6.1 x86']))
-            c.argument('cd_project_url', arg_group='VSTS CD Provider',
-                       help='URL of the Visual Studio Team Services (VSTS) project to use for continuous delivery. URL should be in format `https://<accountname>.visualstudio.com/<projectname>`')
-            c.argument('cd_account_create', arg_group='VSTS CD Provider',
-                       help="To create a new Visual Studio Team Services (VSTS) account if it doesn't exist already",
-                       action='store_true')
-            c.argument('test', arg_group='VSTS CD Provider',
-                       help='Name of the web app to be used for load testing. If web app is not available, it will be created. Default: Disable')
-            c.argument('slot_swap', arg_group='VSTS CD Provider',
-                       help='Name of the slot to be used for deployment and later promote to production. If slot is not available, it will be created. Default: Not configured')
```
panchagnula, line 288: @calvinsID removing all these arguments from the command is a huge breaking change. Are we removing these because the new library doesn't support these? we need to deprecate these & not just remove them completely.

panchagnula (changes requested):
We try to make this not a breaking change to avoid customer impact. Lets wait on the discussion with the core-cli on this.

yonzhan:
@jiasli @Juliehzl to take a look

jiasli:
The PR description seams to be incorrect as there are no `az webapp/functionapp config deployment source` commands:

```
> az webapp config deployment source -h
'deployment' is misspelled or not recognized by the system.

> az functionapp config deployment source -h
'deployment' is misspelled or not recognized by the system.
```

I believe it should be `az webapp/functionapp deployment source`.

src/azure-cli/azure/cli/command_modules/appservice/_params.py
calvinsID: I removed these since these shouldn't currently be used due to VSTS changing to Azure devops. There is a command that functionapps have, `az functionapp devops-pipeline create` which we could make a webapp equivalent, but that command is also in talks to move to an extension due to the vsts package it depends on being too heavyweight

panchagnula: @calvinsID btw we don't need a webapp equivalent for this, this command was created by function apps for something very unique to their area. So I won't worry abt it

panchagnula (approved):
LGTM

jiasli:
The PR description is still incorrect:

```diff
- az webapp/functionapp config deployment source config
+ az webapp/functionapp deployment source config
```

src/azure-cli/azure/cli/command_modules/appservice/_params.py
Juliehzl, line 288: I would recommend you to use deprecate mechanism instead of deleting directly. https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#deprecating-commands-and-arguments

src/azure-cli/azure/cli/command_modules/appservice/_params.py
```diff
@@ -256,38 +256,8 @@ def load_arguments(self, _):
...
             c.argument('repository_type', help='repository type',
-                       arg_type=get_enum_type(['git', 'mercurial', 'vsts', 'github', 'externalgit', 'localgit']))
```
Juliehzl, line 290: the same

src/azure-cli/azure/cli/command_modules/appservice/_params.py
calvinsID: Hi Julie, I added you to an email chain regarding this work item - main reason for removing these commands rather than deprecating is to remove the package dependency

jiasli (approved):

calvinsID:
@Juliehzl would love to get this for S189 if possible, can you pls review and merge as well?

PR #18312, opened 2021-06-01
{Role} `az ad sp create-for-rbac`: Stop setting `identifierUris` on app

jiasli:
## Context

`az ad sp create-for-rbac` creates an app for the service principal. The command specifies `idefntifierUris` on the app:

- if `--name <name>` is provided, `identifierUris` will be set to `http://<name>`
- if `--name <name>` is not provided, `identifierUris` will be set to `http://azure-cli-%Y-%m-%d-%H-%M-%S`

Per https://identitydocs.azurewebsites.net/static/v2/first_party_alert_appiduri_verifieddomain.html, AAD will block app creation with invalid `idefntifierUris` in the near future (planned to be end of June). Valid forms are:

Valid api schemes:

1. api://{appId}
2. api://{tenantId}/{appId}
3. api://{tenantId}/{string}

Other Uri schemes:

1. https://{verifiedCustomerDomain}/{string}
2. https://{string}.{verifiedCustomerDomain}
3. https://{string}.{verifiedCustomerDomain}/{string}

## Changes

`az ad sp create-for-rbac` shouldn't make any assumption and derive `identifierUris` from `--name` as `identifierUris` can be unspecified, so CLI will not use `--name` to derive `identifierUris` to create the app anymore. The app will be created without `identifierUris` to avoid being blocking by AAD.

Before:

```
> az ad sp create-for-rbac --skip-assignment
{
  "appId": "749a8a29-008d-4d1a-b28e-105a086cadc8",
  "displayName": "azure-cli-2021-05-31-08-38-15",
  "name": "http://azure-cli-2021-05-31-08-38-15",    <<<<
  "password": "AYii...",
  "tenant": "54826b22-38d6-4fb2-bad9-b7b93a3e9c5a"
}
```

The `name` property in the output is a property invented by Azure CLI itself (the service principal object doesn't have a `name` property). It maps to `servicePrincipalNames` which is a combination of `idefntifierUris` and `appId`:

```
> az ad sp show --id 749a8a29-008d-4d1a-b28e-105a086cadc8
  ...
  "servicePrincipalNames": [
    "http://azure-cli-2021-05-31-08-38-15",
    "749a8a29-008d-4d1a-b28e-105a086cadc8"
  ],
```

So we switch the value of `name` from `identifierUri` to `appId`. We are deprecating it and will remove it in the future.

After:

```
> az ad sp create-for-rbac --skip-assignment
{
  "appId": "5545b0f9-c100-488c-a55a-02719313f1b2",
  "displayName": "azure-cli-2021-05-31-08-37-30",
  "name": "5545b0f9-c100-488c-a55a-02719313f1b2",    <<<<
  "password": "j4Vu...",
  "tenant": "54826b22-38d6-4fb2-bad9-b7b93a3e9c5a"
}
```

```
> az ad sp show --id 5545b0f9-c100-488c-a55a-02719313f1b2
  ...
  "servicePrincipalNames": [
    "5545b0f9-c100-488c-a55a-02719313f1b2"
  ],
```

Con: This introduces inconsistency:

- `--name` in the input command is used as `displayName`, while
- `name` property in output represents `servicePrincipalNames`

## Idempotence

`az ad sp create-for-rbac` is idempotent when the same `--name` (used as `displayName`) is provided. `az ad sp create-for-rbac --name` can be run multiple times:

- Only the first execution creates new service principal and app
- Following executions update the service principal/app with that `displayName`

This behavior differs from `New-AzADServicePrincipal` which creates a new service principal on each execution.

## Limitation

Many customers are still using older versions of Azure CLI/PowerShell and we do not have control over their migration plan. This BREAKING CHANGE on AAD service side will affect users who are still using older version of Azure CLI/PowerShell.

## Equivalent changes on Azure PowerShell 

- Changes made to [`New-AzADServicePrincipal`](https://docs.microsoft.com/en-us/powershell/module/az.resources/new-azadserviceprincipal?view=azps-6.0.0): https://github.com/Azure/azure-powershell/pull/14952
- Design spec: https://github.com/Azure/azure-powershell-cmdlet-review-pr/issues/895

**History Notes**

[Role] BREAKING CHANGE:  `az ad sp create-for-rbac`: `--name` is now only used as the `displayName` of the app. It is not used to generate `identifierUris` anymore. `name` in the output is now the same as `appID` (`servicePrincipalNames`) and deprecated.


yonzhan:
Role

src/azure-cli/azure/cli/command_modules/role/custom.py
```diff
@@ -42,6 +42,9 @@
     "If needed, use the --role argument to explicitly create a role assignment."
 )
 
+NAME_DEPRECATION_WARNING = \
+    "'name' property in the output is deprecated and will be removed in the future. Use 'appId' instead."
```
evelyn-ys, line 46: Shall we tell users that 'name' property in the output has changed from identifierUris to appId. If they want to get identifierUris, use `az ad app show` instead?

evelyn-ys:
LGTM if all tests pass

src/azure-cli/azure/cli/command_modules/role/custom.py
jiasli: This information is already conveyed in the history. I prefer not to mentioned it here as 

- it is used as an opaque box 
- it will make the already-overly-crowded warning messages even more crowded

yonzhan (approved):

evelyn-ys (approved):

Issue #17223, opened 2021-03-07
UserData for VM and VM ScaleSets (high priority) 


roipmsft:
**Resource Provider**

Compute Resource Provider
**Description of Feature or Work Requested**

https://microsoft.sharepoint.com/:w:/r/teams/ComputeVM/_layouts/15/Doc.aspx?sourcedoc=%7B13AE3881-1B63-4D2E-9334-71E687A4A61D%7D&file=UserDataRequirementsForCLI.docx&action=default&mobileredirect=true
This new UserData feature is essentially a new and improved version of the existing CustomData(https://docs.microsoft.com/en-us/azure/virtual-machines/custom-data). We are trying to catch up with AWS, which already has this feature.
**Minimum API Version Required**

2021-03-01
**Swagger Link**

https://github.com/Azure/azure-rest-api-specs/pull/13220
**Target Date**

We would like to GA on April 30.
This new feature is for Gartner and we are on a tight deadline.

yungezz:
hi @qwordy,  let's check the availability of 2021-03-01 API version and SDK release ETA, then communicate back on CLI plan.

qwordy:
@roipmsft When will the 2021-03-01 API be merged into master branch?

roipmsft:
Api version for 2021-03-01 is already available for CRP in all regions.

UserData for regular VMs is currently rolling out.
UserData for VMSS is not done yet and may not being rolling out until week of April 5.

Swagger PR for UserData (for both VM and VMSS) may not finish until week of April 12, unfortunately.

yungezz:
hi @roipmsft , Swagger change https://github.com/Azure/azure-rest-api-specs/pull/13220#pullrequestreview-633218158 need to be checked in firstly as prerequisite for Azure CLI work. The workflow is: Swagger merged into master -> Python SDK released -> Azure CLI development and release(Azure CLI depends on Python SDK)

zhoxing-ms:
@roipmsft Hi, I would like to ask what is the expected help for `UserData` as a new parameter?
Could I use the description from Swagger
https://github.com/Azure/azure-rest-api-specs/blob/master/specification/compute/resource-manager/Microsoft.Compute/stable/2021-03-01/compute.json#L11047-L11050

yonzhan:
@roipmsft any feedback?

zhoxing-ms:
> There is a max limit to the size of the userData (the same size limit as for existing customData)

[...]

PR #18432, opened 2021-06-10
[Compute] Support user_data for VM and VM Scale Sets

zhoxing-ms:
**Description**

Issue: #17223

- Should allow customer to submit userData as a file.
- User_data must be base64 encoded (CRP will reject any userData that is not based64 encoded).
- CLI can allow customer to pass an empty string as input, it will interpret that as a signal to delete the user_data.
- Allow customers to retrieve value of userData. For that, CLI must make GET request using `$expand=userData`.

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/vm/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/custom.py
@@ -2826,9 +2826,15 @@ def get_vmss(cmd, resource_group_name, name, instance_id=None, include_user_data
     expand = None
     if include_user_data:
         expand = 'userData'
+
     if instance_id is not None:
-        return client.virtual_machine_scale_set_vms.get(resource_group_name, name, instance_id, expand)
-    return client.virtual_machine_scale_sets.get(resource_group_name, name, expand)
+        if cmd.supported_api_version(min_api='2020-12-01', operation_group='virtual_machine_scale_sets'):
+            return client.virtual_machine_scale_set_vms.get(resource_group_name, name, instance_id, expand)
+        return client.virtual_machine_scale_set_vms.get(resource_group_name, name, instance_id)
+
+    if cmd.supported_api_version(min_api='2021-03-01', operation_group='virtual_machine_scale_sets'):
+        return client.virtual_machine_scale_sets.get(resource_group_name, name, expand)
+    return client.virtual_machine_scale_sets.get(resource_group_name, name)
 
 
 def get_vmss_modified(cmd, resource_group_name, name, instance_id=None):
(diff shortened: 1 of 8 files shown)


yonzhan:
Compute

houk-ms (approved):

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -1359,6 +1372,10 @@ def update_vm(cmd, resource_group_name, vm_name, os_disk=None, disk_caching=None
     if license_type is not None:
         vm.license_type = license_type
 
+    if user_data is not None:
+        from azure.cli.core.util import b64encode
+        vm.user_data = b64encode(user_data)
```
evelyn-ys, line 1377: Why not `read_content_if_is_file(user_data)` here?

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -1071,8 +1076,11 @@ def auto_shutdown_vm(cmd, resource_group_name, vm_name, off=None, email=None, we
     return client.global_schedules.create_or_update(resource_group_name, name, schedule)
 
 
-def get_instance_view(cmd, resource_group_name, vm_name):
-    return get_vm(cmd, resource_group_name, vm_name, 'instanceView')
+def get_instance_view(cmd, resource_group_name, vm_name, include_user_data=False):
```
evelyn-ys, line 1079: Why `include_user_data` is added here but ignored in `_params.py`?

src/azure-cli/azure/cli/command_modules/vm/custom.py
zhoxing-ms: Before calling `build_vm_resource()` to build ARM template, method `read_content_if_is_file(user_data)` has been called to get the content of file or strings. [code link](https://github.com/Azure/azure-cli/blob/aa9ac1b2272a8ebd23272075f9bd0068030bf2df/src/azure-cli/azure/cli/command_modules/vm/custom.py#L928
) So here is the content of `user_data` that needs to be sent to REST.
This approach is consistent with `custom_data`.

src/azure-cli/azure/cli/command_modules/vm/custom.py
zhoxing-ms: This is because `get_instance_view()` is not only a custom method called by `vm get-instance-view`, `vm wait` and `vmss wait` commands, but also called by method `get_vm_details()`.
When `get_instance_view()` is used in `get_vm_details()` command, it adds a parameter `include_user_data`, which needs to be passed into `get_instance_view()` method to modify the logic of expand.
Because this parameter is not required for `vm get-instance-view`, `vm wait` and `vmss wait` commands, so it is ignored in these three commands.

evelyn-ys (approved):

PR #18449, opened 2021-06-11
{Compute} Revert "Direct Sharing support for Azure Shared Image Gallery"

zhoxing-ms:
Reverts Azure/azure-cli#18398

As I communicated with @Juliehzl, due to the sudden change of requirement design, we need to revert this PR.


Juliehzl (approved):
Revert as service team request for further ask

PR #18475, opened 2021-06-14
[AppService]: fix #17424 - az appservice plan show does not provide correct exit status

### LINKED ISSUE #17424: webapp:az appservice plan show does not provide correct exit status

**Describe the bug**
`az appservice plan show` does not provide correct exit status

**To Reproduce**
1. execute `az appservice plan show` with arguments describing a non-existent resource, for instance:

`az appservice plan show --resource-group nonesuch-rg-doesntexist --name nonesuch-app-doesntexist`

**Expected behavior**
non-zero exit status, similar to as provided by other `az <resource> show` commands

**Environment summary**
ubunut 18.04 installed from package

Kotasudhakarreddy:
#17424 

**Description**

`az appservice plan show` does not provide correct exit status

**Testing Guide**

`az appservice plan show --resource-group nonesuch-rg-doesntexist --name nonesuch-app-doesntexist`

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -1734,6 +1734,14 @@ def update_app_service_plan(instance, sku=None, number_of_workers=None):
     return instance
 
 
+def show_plan(cmd, resource_group_name, name):
+    client = web_client_factory(cmd.cli_ctx)
+    plan = client.app_service_plans.get(resource_group_name, name)
+    if not plan:
+        raise ResourceNotFoundError("Appservice Plan'{}', is not found on RG '{}'.".format(name, resource_group_name))
+    return plan
+
+
 def update_functionapp_app_service_plan(cmd, instance, sku=None, number_of_workers=None, max_burst=None):
     instance = update_app_service_plan(instance, sku, number_of_workers)
     if max_burst is not None:
--- a/src/azure-cli/azure/cli/command_modules/appservice/commands.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/commands.py
@@ -283,7 +283,7 @@ def load_command_table(self, _):
                          exception_handler=ex_handler_factory(creating_plan=True))
         g.command('delete', 'delete', confirmation=True)
         g.custom_command('list', 'list_app_service_plans')
-        g.show_command('show', 'get')
+        g.custom_show_command('show', 'show_plan')
         g.generic_update_command('update', setter_name='begin_create_or_update', custom_func_name='update_app_service_plan',
                                  setter_arg_name='app_service_plan', supports_no_wait=True,
                                  exception_handler=ex_handler_factory())
--- a/src/azure-cli/azure/cli/command_modules/appservice/tests/latest/recordings/test_webapp_hyperv_e2e.yaml
+++ b/src/azure-cli/azure/cli/command_modules/appservice/tests/latest/recordings/test_webapp_hyperv_e2e.yaml
@@ -13,12 +13,12 @@ interactions:
       ParameterSetName:
       - -g -n --hyper-v --sku
       User-Agent:
-      - AZURECLI/2.24.0 azsdk-python-azure-mgmt-resource/18.0.0 Python/3.8.2 (macOS-10.16-x86_64-i386-64bit)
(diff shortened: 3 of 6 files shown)


yonzhan:
AppService

Kotasudhakarreddy:
@panchagnula please review.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -1734,6 +1734,14 @@ def update_app_service_plan(instance, sku=None, number_of_workers=None):
     return instance
 
 
+def show_plan(cmd, resource_group_name, name):
+    client = web_client_factory(cmd.cli_ctx)
+    plan = client.app_service_plans.get(resource_group_name, name)
+    if not plan:
+        raise ResourceNotFoundError("Appservice Plan'{}', is not found on RG '{}'.".format(name, resource_group_name))
```
Juliehzl, line 1741: https://github.com/Azure/azure-cli/blob/3f230e856a407043102e99b82d4d22f679111a3c/src/azure-cli-core/azure/cli/core/commands/arm.py#L408-L416

If your server could raise exception with status code 404, ResourceNotFound error would be raised. So why do you do the change now?

Kotasudhakarreddy: API is not throwing the exception. But it is returning 404 status code. to handle that I prefer to go with custom exception.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -1734,6 +1734,14 @@ def update_app_service_plan(instance, sku=None, number_of_workers=None):
     return instance
 
 
+def show_plan(cmd, resource_group_name, name):
+    client = web_client_factory(cmd.cli_ctx)
+    plan = client.app_service_plans.get(resource_group_name, name)
+    if not plan:
```
panchagnula, line 1740: @Kotasudhakarreddy have you tried the option we discussed in our meeting - seems like API is throwing a message with 404, but we are not caching this since CLI expects JSON response but this is giving a HTML response. I would like to avoid showing custom message & handle 404 case accurately. Thanks!

panchagnula (changes requested):
Please try the option we discussed in our meeting to handle HTML response from API instead of JSON that CLI expects. Thanks!

src/azure-cli/azure/cli/command_modules/appservice/custom.py
Kotasudhakarreddy: Yes, sisira. I have tried and need to discuss with you on this.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
Kotasudhakarreddy: @Juliehzl- API is not returning any Json response but it is returning html response. please suggest me the best approach of handling html response in CLI.

Juliehzl: As you define 404 status code as non-error in [swagger](https://github.com/Azure/azure-rest-api-specs/blob/c803720c6bcfcb0fcf4c97f3463ec33a18f9e55c/specification/web/resource-manager/Microsoft.Web/stable/2021-01-01/AppServicePlans.json#L140-L142) , the generated python SDK will not raise exception as [here](https://github.com/Azure/azure-sdk-for-python/blob/f89776994d2a0292e6b10b9920f8828d5554143a/sdk/appservice/azure-mgmt-web/azure/mgmt/web/v2021_01_01/operations/_app_service_plans_operations.py#L250-L253) and azure CLI cannot catch the exception for resourceNotFound error.

There are two ways to fix:
1. Change swagger definition, which need to remove 404 part in response as [storage](https://github.com/Azure/azure-rest-api-specs/blob/0c057560083bfdfd50202f0e79cfac8ec054d56a/specification/storage/resource-manager/Microsoft.Storage/stable/2021-04-01/storage.json#L383-L398) and [network](https://github.com/Azure/azure-rest-api-specs/blob/0c057560083bfdfd50202f0e79cfac8ec054d56a/specification/network/resource-manager/Microsoft.Network/stable/2021-02-01/applicationGateway.json#L126-L140).
2. If you don't want to change swagger, you need to catch it in your code. In this way, I would recommend  you check response status code with 404. Currently you will raise resource not error when not plan, which is not accurate enough.

Kotasudhakarreddy: @panchagnula your thoughts please.

PR #18506, opened 2021-06-16
{Packaging} Bump PyJWT to 2.1.0

### LINKED ISSUE #17939: Azure-ci-core has a strict dependency on an very old PyJWT

**Describe the bug**
The setup.py for azure-cli-core explicitly requires `PyJWT==1.7.1` while all the other azure packages are more relaxed with `>=1.0.0,<3`.

https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/setup.py#L61

**To Reproduce**
`pip install azure-cli` alongside more recent packages that need `PyJWT>= 2.0`.

```
pip install azure-cli==2.22.1 PyJWT==2.1.0
... 
ERROR: Cannot install PyJWT==2.1.0 and azure-cli-core==2.22.1 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested PyJWT==2.1.0
    azure-cli-core 2.22.1 depends on PyJWT==1.7.1

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/user_guide/#fixing-conflicting-dependencies
WARNING: You are using pip version 21.0.1; however, version 21.1.1 is available.
```

**Expected behavior**
We should not have a version conflict with newer versions of PyJWT when installing the latest azure cli.

**Environment summary**
Installed with python 3.9.4 on linux and macOS. Pip 21.0.1.

**Additional context**
PyJWT 1.7.1  was released on 2018. This bug relates to #17598.

jiasli:
**Description**

Fix #17939

ADAL already supports PyJWT > 2.0 (https://github.com/AzureAD/azure-activedirectory-library-for-python/pull/246), it is safe for Azure CLI to use PyJWT 2.1.0.

### DIFF
--- a/src/azure-cli-core/azure/cli/core/_profile.py
+++ b/src/azure-cli-core/azure/cli/core/_profile.py
@@ -358,7 +358,7 @@ class Profile:
         token_entry = msi_creds.token
         token = token_entry['access_token']
         logger.info('MSI: token was retrieved. Now trying to initialize local accounts...')
-        decode = jwt.decode(token, verify=False, algorithms=['RS256'])
+        decode = jwt.decode(token, algorithms=['RS256'], options={"verify_signature": False})
         tenant = decode['tid']
 
         subscription_finder = SubscriptionFinder(self.cli_ctx, self.auth_ctx_factory, None)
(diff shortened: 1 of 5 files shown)


src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -358,7 +358,7 @@ def find_subscriptions_in_vm_with_msi(self, identity_id=None, allow_no_subscript
         token_entry = msi_creds.token
         token = token_entry['access_token']
         logger.info('MSI: token was retrieved. Now trying to initialize local accounts...')
-        decode = jwt.decode(token, verify=False, algorithms=['RS256'])
+        decode = jwt.decode(token, algorithms=['RS256'], options={"verify_signature": False})
```
jiasli, line 361: https://pyjwt.readthedocs.io/en/latest/changelog.html?highlight=verify#dropped-deprecated-verify-param-in-jwt-decode

> Dropped deprecated `verify` param in `jwt.decode(...)`
> Use `jwt.decode(encoded, key, options={"verify_signature": False})` instead.

yonzhan:
Packaging

kairu-ms (approved):

evelyn-ys (approved):

PR #18520, opened 2021-06-17
[Keyvault] BREAKING CHANGE: Fix #13752: az keyvault create not idempotent. Creating existing keyvault will fail.

### LINKED ISSUE #13752: az keyvault create not idempotent with access policies

evelyn-ys:
**Description**

Fix #13752 
CLI will set default access_policy when running `az keyvault create`. That will overwrite existing settings if the given vault already exists.
After discussion with customer and keyvault server side engineer, we decided to keep align with what powershell does in such situation: Throw exception if the given `vault_name` exists.

**Testing Guide**

```
> az keyvault create -n existingvault -g testrg
The specified vault: existingvault already exists
```
**Notes:** All module tests containing `keyvault create` need to be re-recorded or changed to `KeyVaultPrepare`

---


yonzhan:
Keyvault

Juliehzl (approved):
Approved for code owner part

kairu-ms (approved):

src/azure-cli/azure/cli/command_modules/appservice/tests/latest/test_webapp_commands.py
```diff
@@ -1196,10 +1196,10 @@ def test_webapp_ssl(self, resource_group, resource_group_location):
 
 class WebappSSLImportCertTest(ScenarioTest):
     @ResourceGroupPreparer(location=WINDOWS_ASP_LOCATION_WEBAPP)
-    def test_webapp_ssl_import(self, resource_group):
+    @KeyVaultPreparer(location=WINDOWS_ASP_LOCATION_WEBAPP, name_prefix='kv-ssl-test', name_len=20)
+    def test_webapp_ssl_import(self, resource_group, key_vault):
```
jiasli, line 1200: `key_vault` is usually used as a single name.

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
```diff
@@ -644,6 +644,15 @@ def create_vault(cmd, client,  # pylint: disable=too-many-locals
                  no_self_perms=None,
                  tags=None,
                  no_wait=False):
+    from azure.core.exceptions import HttpResponseError
+    try:
+        vault = client.get(resource_group_name=resource_group_name, vault_name=vault_name)
+        if vault:
+            raise InvalidArgumentValueError('The specified vault: {} already exists'.format(vault_name))
```
jiasli, line 651: In my perspective, forbidding `create` from being called multiple times is against the Azure API design:

https://docs.microsoft.com/en-us/azure/architecture/microservices/design/api-design#idempotent-operations

> The HTTP specification states that GET, PUT, and DELETE methods must be idempotent.

**Vaults - Create Or Update** is a PUT operation:

https://docs.microsoft.com/en-us/rest/api/keyvault/vaults/create-or-update

```
PUT https://management.azure.com/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.KeyVault/vaults/{vaultName}?api-version=2019-09-01
```

This PR introduces BREAKING CHANGE and we'd better call that out. This script works before but now it breaks:

```
az keyvault create -n myvault -g testrg
az keyvault create -n myvault -g testrg
``` 

Only my 2 cents.

src/azure-cli/azure/cli/command_modules/ams/tests/latest/test_ams_account_encryption_scenarios.py
```diff
@@ -9,7 +9,7 @@
 class AmsEncryptionTests(ScenarioTest):
     @ResourceGroupPreparer(location='centralus')
     @StorageAccountPreparer(parameter_name='storage_account_for_create')
-    @KeyVaultPreparer(location='centralus')
+    @KeyVaultPreparer(location='centralus', additional_params='--enable-purge-protection')
```
jiasli, line 12: The refactoring of `KeyVaultPreparer` could be split into another PR.

src/azure-cli/azure/cli/command_modules/keyvault/tests/hybrid_2018_03_01/test_keyvault_commands.py
```diff
@@ -195,9 +194,9 @@ def test_keyvault_key(self, resource_group):
         self.kwargs['key_file'] = key_file
         self.cmd('keyvault key backup --vault-name {kv} -n {key} --file {key_file}')
         self.cmd('keyvault key delete --vault-name {kv} -n {key}')
-        time.sleep(10)
+        time.sleep(200)
```
jiasli, line 197: Sleeping for 200s makes live test time consuming. Any reason for the change? How about adding some retry logic, like the one in `role` module:

https://github.com/Azure/azure-cli/blob/11347b73999dc9b9d1c6f924fb42e3d20b34a767/src/azure-cli/azure/cli/command_modules/role/tests/util.py#L12-L26

src/azure-cli/azure/cli/command_modules/appservice/tests/latest/test_webapp_commands.py
evelyn-ys: I just keep the origin definition within `KeyVaultPreparer` to make this PR change as small as possible.

src/azure-cli/azure/cli/command_modules/ams/tests/latest/test_ams_account_encryption_scenarios.py
evelyn-ys: It's part of re-recording creating keyvault in other modules.

src/azure-cli/azure/cli/command_modules/keyvault/tests/hybrid_2018_03_01/test_keyvault_commands.py
evelyn-ys: If the deletion has not finished, the later commands will fail.

The retry logic is more elegant, I can refine the code in another PR. For this one, I want it get merged **ASAP** since I have resolved conflicts for 3 times. Tough time😞

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
evelyn-ys: Sure, I'll add the breaking change notes in history

zhoxing-ms (approved):
Approved for code owner part

jiasli (approved):

pascalnaber:
To raise an error when a keyvault already exists is a very bad choice. (just as choice to let it behave the same as powershell)
We are using the az cli for a reason above Powershell. The commands in Powershell are not idempotent.
Better would be to have a look at all other resources in azure regarding the create command.
For example the az storage account create 
You will not get an error that the storage account already exists. It's idempotent.

Please fix this create command to be truly idempotent. Just like all other azure resources that can be created with az.

evelyn-ys:
@pascalnaber You are correct that the keyvault create command isn't idempotent. That's exactly what we argued internally.

The situation is other service creation is idempotent from service side, but keyvault creation is separated to two parts: one from CLI side to set default access policy, another from service side to create with initial settings.

Service won't cover setting default access policy logic during creation then if we want to make the whole process idempotent, client side(both CLI and PSH) will have to add idempotency logic for access policy which should be in server side. Both we and PSH team think it's inappropriate for client to ensure idempotency. That's why we take such design.

Sorry for any inconvenience and hope for you understanding!

PR #18523, opened 2021-06-17
{Pylint} Fix `use-a-generator`

jiasli:
**Description**

http://pylint.pycqa.org/en/latest/technical_reference/features.html

> use-a-generator (R1729)
> Use a generator instead '%s(%s)'. Comprehension inside of 'any' or 'all' is unnecessary. A generator would be sufficient and faster.


yonzhan:
Pylint

evelyn-ys (approved):

Issue #17851, opened 2021-04-24
az acr import needs --no-wait option


rajsell:
az acr import being an async operation, the behavior of the az acr import command is working mimicking it to be a sync operation, it would be great if you can provide the --no-wait option so that if required customers can use that option to wait until the import operation completes.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: c608162e-cf40-9d53-da4a-51f51a0311ea
* Version Independent ID: eccbb543-be68-7de2-d41d-abee55c81452
* Content: [az acr](https://docs.microsoft.com/en-us/cli/azure/acr?view=azure-cli-latest)
* Content Source: [latest/docs-ref-autogen/acr.yml](https://github.com/MicrosoftDocs/azure-docs-cli/blob/master/latest/docs-ref-autogen/acr.yml)
* Service: **container-registry**
* GitHub Login: @rloutlaw
* Microsoft Alias: **routlaw**

yungezz:
route to appropriate team

yugangw-msft:
will consider for v-next

iamjasonp:
To @CarolineNB. Thanks!

stan-sz:
Seems to be related to #1528

PR #18572, opened 2021-06-21
[ACR] az acr import: Support no-wait 

CarolineNB:
**Description**

Fixes https://github.com/Azure/azure-cli/issues/17851

**Testing Guide**
`az acr import -n MyRegistry --source sourceregistry.azurecr.io/sourcerepository:sourcetag --no-wait`

---


yonzhan:
ACR

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -264,6 +264,9 @@
     text: |
         az acr import -n MyRegistry --source sourcerepository:sourcetag -t targetrepository:targettag \\
             -r /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/sourceResourceGroup/providers/Microsoft.ContainerRegistry/registries/sourceRegistry
+  - name: Import an image and return right after starting an import.
```
adewaleo, line 267: Maybe reword in a way like....
```suggestion
  - name: Import an image without waiting for completion.
```
or 
```suggestion
  - name: Import an image. Do not wait for completion.
```
Just some suggestions to consider.

adewaleo: Also you can update the short summary above to add:
```
For large images consider using `--no-wait`.
```

src/azure-cli/azure/cli/command_modules/acr/import.py
```diff
@@ -94,18 +95,23 @@ def acr_import(cmd,
...
+        return result_poller
+
     return _handle_result(cmd, result_poller, source_registry, source_image, registry)
 
 
 def _handle_result(cmd, result_poller, source_registry, source_image, registry):
     from msrestazure.azure_exceptions import ClientException
     try:
-        result = LongRunningOperation(cmd.cli_ctx, 'Importing image...')(result_poller)
+        result = LongRunningOperation(cmd.cli_ctx, 'Importing image...', 'Import has finished')(result_poller)
```
adewaleo, line 113: maybe consider:
```suggestion
        result = LongRunningOperation(cmd.cli_ctx, 'Importing image...', 'Import succeeded')(result_poller)
```
or something else like "Import succeeded" or "Import completed" or just "success".

src/azure-cli/azure/cli/command_modules/acr/import.py
```diff
@@ -94,18 +95,23 @@ def acr_import(cmd,
...
                                               untagged_target_repositories=repository,
                                               mode=ImportMode.force.value if force else ImportMode.no_force.value)
+
     result_poller = client.import_image(
         resource_group_name=resource_group_name,
         registry_name=registry_name,
         parameters=import_parameters)
 
+    if no_wait:
+        return result_poller
```
adewaleo, line 105: Simply returning the poller might not be the best way to go about this. 

The [authoring command modules](https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md) section on supporting `--no-wait`

Mentions the cli core method `azure.cli.core.util.sdk_no_wait` defined here in [util.py](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/azure/cli/core/util.py#L646).

```python
sdk_no_wait(no_wait, func, *args, **kwargs)
```
This method does calls the function you pass with the supplied args and kwargs, and sets/updates the arguments to include `polling: False`. 

Given our code doesn't rely on sdk_no_wait, some approaches come to mind on how we can either use it or copy the behavior of sdk_no_wait when handling result.

src/azure-cli/azure/cli/command_modules/acr/import.py
```diff
@@ -94,18 +95,23 @@ def acr_import(cmd,
...
+
     return _handle_result(cmd, result_poller, source_registry, source_image, registry)
 
 
 def _handle_result(cmd, result_poller, source_registry, source_image, registry):
     from msrestazure.azure_exceptions import ClientException
     try:
-        result = LongRunningOperation(cmd.cli_ctx, 'Importing image...')(result_poller)
```
adewaleo, line 108: Inline with my comment above, we might need to rethink how we approach handle result when we have a wait versus a no wait. There are a few examples of calls to sdk_no_wait here in this [file](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/vm/custom.py#L1013).

src/azure-cli/azure/cli/command_modules/acr/import.py
```diff
@@ -94,18 +95,23 @@ def acr_import(cmd,
...
+        result = LongRunningOperation(cmd.cli_ctx, 'Importing image...', 'Import has finished')(result_poller)
+
     except CLIError as e:
         try:
             # if command fails, it might be because user specified registry twice in --source and --registry
```
adewaleo, line 117: finally let's make sure that with or without no-wait (if user enters input incorrectly) and service rejects the request in the event of no-wait, we show the user the warning below..

adewaleo (changes requested):
Nice work @CarolineNB. some comments and feedback.

src/azure-cli/azure/cli/command_modules/acr/import.py
```diff
@@ -35,8 +35,9 @@ def acr_import(cmd,
                target_tags=None,
                resource_group_name=None,
                repository=None,
-               force=False):
-
+               force=False,
+               no_wait=False,
+               ):
```
jaysterp, line 40: @CarolineNB small style comment: please remove the comma after the last parameter. Also `):` does not need to be on a new line.

Juliehzl (commented):
1. Please add test for --no-wait to make sure it work
2. Then recommend to add wait command to query

src/azure-cli/azure/cli/command_modules/acr/import.py
```diff
@@ -95,45 +96,44 @@ def acr_import(cmd,
...
-        return result_poller
+            return LongRunningOperation(cmd.cli_ctx, 'Importing image...')(result_poller)
 
-    return _handle_result(cmd, result_poller, source_registry, source_image, registry)
+    except CLIError as e:
+        _handle_result(e, cmd, source_registry, source_image, registry)
 
 
-def _handle_result(cmd, result_poller, source_registry, source_image, registry):
+def _handle_result(e, cmd, source_registry, source_image, registry):
```
jaysterp, line 115: @CarolineNB nit: since this class is now an exception handler only, we can rename it to `_handle_exception`.

adewaleo: agreed. Maybe name it `_handle_import_exception`

src/azure-cli/azure/cli/command_modules/acr/tests/latest/test_acr_commands.py
```diff
@@ -245,6 +245,36 @@ def test_acr_create_replication(self, resource_group, resource_group_location):
         # test acr delete
         self.cmd('acr delete -n {registry_name} -g {rg} -y')
 
+    @ResourceGroupPreparer()
+    def test_acr_import_no_wait(self, resource_group):
+        source_registry_name = self.create_random_name("sourceregistrysamesub", 40)
+        registry_name = self.create_random_name("targetregistry", 20)
```
adewaleo, line 251: registry_name isn't being used
```suggestion
```

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -264,7 +264,7 @@
     text: |
         az acr import -n MyRegistry --source sourcerepository:sourcetag -t targetrepository:targettag \\
             -r /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/sourceResourceGroup/providers/Microsoft.ContainerRegistry/registries/sourceRegistry
-  - name: Import an image and return right after starting an import.
+  - name: Import an image without waiting for completion.
```
adewaleo, line 267: @jaysterp it might be overkill, but I wonder if we should hint that this doesn't check for success.
maybe say: 
```suggestion
  - name: Import an image without waiting for successful completion.
```

src/azure-cli/azure/cli/command_modules/acr/tests/latest/recordings/test_check_name_availability.yaml
```diff
@@ -17,8 +17,8 @@ interactions:
       ParameterSetName:
       - -n
       User-Agent:
-      - python/3.7.9 (Windows-10-10.0.19041-SP0) msrest/0.6.18 msrest_azure/0.6.3
-        azure-mgmt-containerregistry/3.0.0rc16 Azure-SDK-For-Python AZURECLI/2.16.0
+      - python/3.8.10 (Windows-10-10.0.19041-SP0) msrest/0.6.21 msrest_azure/0.6.3
```
adewaleo, line 20: is checking in these new recordings necessary? If not, it's okay, but also consider removing unnecessary check-ins to keep PR concise.

adewaleo (commented):
Overall looks good to me. However, left some suggestions and comments.

src/azure-cli/azure/cli/command_modules/acr/import.py
```diff
@@ -35,7 +36,8 @@ def acr_import(cmd,
                target_tags=None,
                resource_group_name=None,
                repository=None,
-               force=False):
+               force=False,
+               no_wait=False):
```
jaysterp, line 40: @CarolineNB please add the comment `# pylint: disable=too-many-locals` to suppress the style warning "Too many local variables". Since these variables are command parameters, we can ignore this error. Please reference [acr_task_create](https://github.com/Azure/azure-cli/blob/aaffbc5cafa4535aadc180136ec094f1725a0f38/src/azure-cli/azure/cli/command_modules/acr/task.py#L43).

src/azure-cli/azure/cli/command_modules/acr/import.py
```diff
@@ -94,41 +96,44 @@ def acr_import(cmd,
...
-    result_poller = client.import_image(
-        resource_group_name=resource_group_name,
-        registry_name=registry_name,
-        parameters=import_parameters)
 
-    return _handle_result(cmd, result_poller, source_registry, source_image, registry)
+    try:
+        if no_wait:
+            return sdk_no_wait(no_wait, client.import_image, resource_group_name, registry_name, import_parameters)
+        else:
```
jaysterp, line 103: @CarolineNB pylint is flagging this if/else with the "Unnecessary "else" after "return"" style error. You can see how to fix this by referencing [no-else-return](https://pheanex.github.io/pylint/#r1705-no-else-return).

Pylint error: https://dev.azure.com/azure-sdk/public/_build/results?buildId=973250&view=logs&jobId=36dd4138-4d53-5e46-00d9-e5cd9744be05&j=36dd4138-4d53-5e46-00d9-e5cd9744be05&t=1cf3879c-0a42-5ccd-7693-7a4781d739d8

src/azure-cli/azure/cli/command_modules/acr/_help.py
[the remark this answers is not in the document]
jaysterp: @adewaleo I like this idea. Since imageImport does not provide a GET API that users could leverage to check their import status, it is a good user experience to warn them of this if they opt out of waiting. We can recommend them in the warning to run `az acr repository show-tags` to confirm the import succeeded (however this does not cover the case where import overwrites tags). @nathana1 what do you think?

`--no-wait=True` will still catch failures in the front end of the request after it lands in our RP and before we send the request to the backend for processing. These errors (request validation, etc) are all thrown before we create the async response. Failure in the backend to actually perform the import will not be surfaced.

jaysterp: @adewaleo I agree with this suggestion. @CarolineNB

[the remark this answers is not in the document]
nathana1: I think adding a warning to the description/documentation would be a good idea.  I don't think we'd want to add a warning to every time the command is run with this flag.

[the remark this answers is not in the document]
adewaleo: yeah I was thinking more along the line of documentation warning. (I do think logging a warning on `--no-wait` imports could also be helpful)

iamjasonp:
Rebase to dev completed with 2nd pair of eyes cc: @jaysterp

jaysterp (approved):

adewaleo (approved):

jaysterp:
@Juliehzl we are ready to merge. Please take a look. Thanks!

Juliehzl (approved):

PR #18622, opened 2021-06-24
[AppService] az appservice domain create: Fix get correct domain agreements

### LINKED ISSUE #18651: WebApp:az appservice domain create: incorrect domain agreements

**Describe the bug**
Incorrect domain agreements used in `az appservice domain create`. This causes domain preflight to fail on the API side

**To Reproduce**
Run `az appservice domain create` without privacy flag. The domain agreement returns privacy & transfer agreements

**Expected behavior**
privacy agreeement should only be shown when privacy is specified, and transfer agreement shouldn't be shown (since we don't support transfer)

calvinsID:
**Description**
Get correct domain agreements for user to acknowledge, in `az appservice domain create` command

Backend is introducing a validation API for domain creates, and domains created through CLI are passing in different agreements than what was expected causing the validation API to fail

Fixes #18651 

**Testing Guide**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/appservice_domains.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/appservice_domains.py
@@ -3,11 +3,12 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 
+from azure.cli.core.azclierror import ValidationError
 from azure.cli.core.commands import DeploymentOutputLongRunningOperation
 from azure.cli.core.commands.client_factory import get_mgmt_service_client
-from azure.mgmt.web.models import NameIdentifier
 from azure.cli.core.profiles import ResourceType
 from azure.cli.core.util import sdk_no_wait, random_string
+from azure.mgmt.web.models import NameIdentifier
 from knack.util import CLIError
 from knack.log import get_logger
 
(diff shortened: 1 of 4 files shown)


yonzhan:
AppService

panchagnula (changes requested):
are there any tests we can add here to identify these issues?

calvinsID:
> are there any tests we can add here to identify these issues?

I don't think we can add a test for `domain create` since we aren't hooked up to GD test endpoints

panchagnula:
> > are there any tests we can add here to identify these issues?
> 
> I don't think we can add a test for `domain create` since we aren't hooked up to GD test endpoints

The question mainly was can we create a test to validate the values returned from the command are correct (like defaults) if possible? If not this LGTM

calvinsID:
> tion mainly was can we create a test to validate the values returned from the command are correct (li

Ah right there is a --dryrun option, let me add tests

panchagnula (approved):
Ship It

src/azure-cli/azure/cli/command_modules/appservice/appservice_domains.py
```diff
@@ -50,6 +50,16 @@ def create_domain(cmd, resource_group_name, hostname, contact_info, privacy=True
     web_client = web_client_factory(cmd.cli_ctx)
     hostname_availability = web_client.domains.check_availability(NameIdentifier(name=hostname))
 
+    if not hostname_availability.available:
+        raise CLIError("Custom domain name '{}' is not available. Please try again "
+                       "with a new hostname.".format(hostname))
+
+    tld = '.'.join(hostname.split('.')[1:])
+    from azure.mgmt.web.models import TopLevelDomainAgreementOption
```
Juliehzl, line 58: please use get_models() as 3.  in https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#multi-api-aware-modules

src/azure-cli/azure/cli/command_modules/appservice/appservice_domains.py
```diff
@@ -50,6 +50,16 @@ def create_domain(cmd, resource_group_name, hostname, contact_info, privacy=True
     web_client = web_client_factory(cmd.cli_ctx)
     hostname_availability = web_client.domains.check_availability(NameIdentifier(name=hostname))
 
+    if not hostname_availability.available:
+        raise CLIError("Custom domain name '{}' is not available. Please try again "
```
Juliehzl, line 54: please use error type in azclierror, e.g.
https://github.com/Azure/azure-cli/blob/3eea7fd773722b7a3c40b8115f29a3bad39b70bc/src/azure-cli-core/azure/cli/core/azclierror.py#L152-L154

panchagnula:
@Juliehzl this PR is needed for S189 as well - can you please sign-off & merge?

PR #18662, opened 2021-06-28
[Compute] Add the hint to suggest users use the standard public IP when creating VM

zhoxing-ms:
**Description**

In the latest profile and the profiles that use the same api-version as latest profile, the default public IP used for VM creation will be expected to be changed from Basic to Standard. 
In order to avoid breaking change which has a big impact on users, we use the hint to guide users to use Standard public IP to create VM in the first stage.

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/vm/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/custom.py
@@ -753,13 +753,10 @@ def create_vm(cmd, vm_name, resource_group_name, image=None, size='Standard_DS1_
     from msrestazure.tools import resource_id, is_valid_resource_id, parse_resource_id
     from azure.cli.core.profiles import AZURE_API_PROFILES
 
-    # In the latest profile and the profiles that use the same api-version as latest profile,
-    # the default public IP used for VM creation will be expected to be changed from Basic to Standard.
-    # In order to avoid breaking change which has a big impact on users,
+    # In the latest profile, the default public IP will be expected to be changed from Basic to Standard.
+    # In order to avoid breaking change which has a big impact to users,
     # we use the hint to guide users to use Standard public IP to create VM in the first stage.
-    api_version_for_latest_profile = AZURE_API_PROFILES['latest'][ResourceType.MGMT_COMPUTE].default_api_version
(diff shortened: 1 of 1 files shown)


yonzhan:
Compute

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -751,6 +751,19 @@ def create_vm(cmd, vm_name, resource_group_name, image=None, size='Standard_DS1_
                                                                 build_vm_windows_log_analytics_workspace_agent)
     from azure.cli.command_modules.vm._vm_utils import ArmTemplateBuilder20190401
     from msrestazure.tools import resource_id, is_valid_resource_id, parse_resource_id
+    from azure.cli.core.profiles import AZURE_API_PROFILES
+
+    # In the latest profile and the profiles that use the same api-version as latest profile,
+    # the default public IP used for VM creation will be expected to be changed from Basic to Standard.
+    # In order to avoid breaking change which has a big impact on users,
+    # we use the hint to guide users to use Standard public IP to create VM in the first stage.
+    api_version_for_latest_profile = AZURE_API_PROFILES['latest'][ResourceType.MGMT_COMPUTE].default_api_version
```
Juliehzl, line 760: for default value change, why do we need to use api verison in latest profile as metric, especially latest profile may change?

Juliehzl: if you don't want to change default value when other profile is used, you only need to check which profile is using. If you don't want to change default value when api version is lower than xxx, you could just use a valid api version instead of a dynamic one.

zhoxing-ms: I have confirmed with @brianlehr that we only need the api-version used in the latest profile to change the default behavior of VM creation, and do not want to affect other profiles such as Azure Stack for the time being

zhoxing-ms: OK, what is the appropriate way to determine which profile to use? I'll change the code here

zhoxing-ms: Done

Juliehzl (approved):

PR #18675, opened 2021-06-29
[App Service] Add support for Linux Consumption and improve how content share name is generated.

### LINKED ISSUE #18252: Function app creation -- add two additional app settings (WEBSITE_CONTENTSHARE and WEBSITE_CONTENTAZUREFILECONNECTIONSTRING) for Linux Consumption apps
During function app creation, please add the following two add settings before sending the request to the backend  for Linux Consumption apps:

"WEBSITE_CONTENTSHARE" - This is the share name. For more information, see 
https://docs.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-shares--directories--files--and-metadata#share-names

"WEBSITE_CONTENTAZUREFILECONNECTIONSTRING" - This is the connection string to the function app's storage account

gzuber:
Resolves #18252 

**Description**

This adds support for powershell on Linux consumption. It also changes the way we generate content share name to make it random. This should guarantee that there are no conflicts if a customer deletes and re-creates an app.

**Testing Guide**

Added checks to existing test `test_functionapp_consumption_linux_powershell`.

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -2982,12 +2982,11 @@ def create_function(cmd, resource_group_name, name, storage_account, plan=None,
     if consumption_plan_location is None and not is_plan_elastic_premium(cmd, plan_info):
         site_config.always_on = True
 
-    # If plan is elastic premium or windows consumption, we need these app settings
-    is_windows_consumption = consumption_plan_location is not None and not is_linux
-    if is_plan_elastic_premium(cmd, plan_info) or is_windows_consumption:
+    # If plan is elastic premium or consumption, we need these app settings
+    if is_plan_elastic_premium(cmd, plan_info) or consumption_plan_location is not None:
         site_config.app_settings.append(NameValuePair(name='WEBSITE_CONTENTAZUREFILECONNECTIONSTRING',
                                                       value=con_string))
-        site_config.app_settings.append(NameValuePair(name='WEBSITE_CONTENTSHARE', value=name.lower()))
+        site_config.app_settings.append(NameValuePair(name='WEBSITE_CONTENTSHARE', value=_get_content_share_name(name)))
 
     create_app_insights = False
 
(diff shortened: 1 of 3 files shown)


gzuber:
This should be ready for review/merge! Was hoping to get this in [S189](https://github.com/Azure/azure-cli/milestone/107)

yonzhan:
functionapp

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2982,12 +2982,11 @@ def create_function(cmd, resource_group_name, name, storage_account, plan=None,
...
 
-    # If plan is elastic premium or windows consumption, we need these app settings
-    is_windows_consumption = consumption_plan_location is not None and not is_linux
-    if is_plan_elastic_premium(cmd, plan_info) or is_windows_consumption:
+    # If plan is elastic premium or consumption, we need these app settings
+    if is_plan_elastic_premium(cmd, plan_info) or consumption_plan_location is not None:
         site_config.app_settings.append(NameValuePair(name='WEBSITE_CONTENTAZUREFILECONNECTIONSTRING',
                                                       value=con_string))
-        site_config.app_settings.append(NameValuePair(name='WEBSITE_CONTENTSHARE', value=name.lower()))
+        site_config.app_settings.append(NameValuePair(name='WEBSITE_CONTENTSHARE', value=_get_content_share_name(name)))
```
panchagnula, line 2989: Curious can this name be an optional property exposed to users so they can offer a name & we can validate it & use it or else generate the random name.

panchagnula (changes requested):
LGTM - apologize I don't know the details of functionapp, so have one question.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
gzuber: Yeah so this change brings content share name in line with what every other tool does (portal doesn't have a field for this, powershell client, etc...). For some reason the CLI wasn't adding that short random string which could potentially cause issues. Talking to Bala, it seems like the sentiment is that if someone wanted to get that specific it's something they could do with ARM templates. It's not that we can't/shouldn't let the customer specify a name, just that I think it would require a larger conversation across tooling

panchagnula (approved):
Thanks for the update

Issue #18327, opened 2021-06-02
Create VM flow broken for "EdgeZone" 


viananth:
> ### `az feedback` auto-generates most of the information requested below, as of CLI version 2.0.62

**Describe the bug**

Creating a VM in EdgeZone using "--edge-zone" parameter creates the vm in edge zone, but creates other resources (like VNet, Nic, Public Ip, etc)  in main region.

**To Reproduce**

1. Install latest Az CLI
2. Run "az vm create -g testcli -n testclivm --image UbuntuLTS --edge-zone microsoftlosangeles1 --public-ip-sku Standard --location westus --debug --generate-ssh-keys"
3. Creates the VM successfully in edge zone. But other resources are created in main region.
4. For example, NIC's Json view from portal should show "ExtendedLocation" as "microsoftlosangeles1", but it doesn't have edgezone information here (with the latest API version from drop down list):

> {
>   "name": "testclivmVMNic",
>   "id": "/subscriptions/baffff89-2c76-424a-af4c-43b7518a3cb8/resourceGroups/viananth-testcli/providers/Microsoft.Network/networkInterfaces/testclivmVMNic",
>   "etag": "W/\"1867f590-9435-4c55-80c9-5d7273c39d3d\"",
>   "location": "westus",
>   "tags": {},
>   "properties": {
>     "provisioningState": "Succeeded",
>     "resourceGuid": "59b8cbe5-8f0e-4d10-8b0b-05345e0070f6",
>     "ipConfigurations": [
>       {
>         "name": "ipconfigtestclivm",
>         "id": "/subscriptions/baffff89-2c76-424a-af4c-43b7518a3cb8/resourceGroups/viananth-testcli/providers/Microsoft.Network/networkInterfaces/testclivmVMNic/ipConfigurations/ipconfigtestclivm",
>         "etag": "W/\"1867f590-9435-4c55-80c9-5d7273c39d3d\"",
>         "type": "Microsoft.Network/networkInterfaces/ipConfigurations",
>         "properties": {
>           "provisioningState": "Succeeded",
>           "privateIPAddress": "10.0.0.4",
>           "privateIPAllocationMethod": "Dynamic",
>           "publicIPAddress": {
>             "id": "/subscriptions/baffff89-2c76-424a-af4c-43b7518a3cb8/resourceGroups/viananth-testcli/providers/Microsoft.Network/publicIPAddresses/testclivmPublicIP"
>           },
>           "subnet": {
[...]

PR #18708, opened 2021-07-01
[Compute] Make dependent resources of VM/VMSS support edge zone

zhoxing-ms:
Issue: #18327

**Description**

Releated PR: #17522

Make dependent resources of VM/VMSS support edge zone

For VM:
- storageAccount 
- publicIPAddress 
- virtualNetwork 
- networkInterface

For VMSS:
- storageAccount 
- publicIPAddress 
- virtualNetwork 
- loadBalancer 

Because the previous fixed api-version of dependent resources is too old, it does not support edge zone. Therefore, we need to upgrade the api-version used to create dependent resources when users pass in `--edge-zone` parameter *(so as to reduce the risk of upgrading api-version, such as the possibility of breaking change in REST)*

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
@@ -65,7 +65,7 @@ def build_output_deployment_resource(key, property_name, property_provider, prop
     return deployment
 
 
-def build_storage_account_resource(_, name, location, tags, sku):
+def build_storage_account_resource(_, name, location, tags, sku, edge_zone=None):
     storage_account = {
         'type': 'Microsoft.Storage/storageAccounts',
         'name': name,
@@ -75,10 +75,16 @@ def build_storage_account_resource(_, name, location, tags, sku):
         'dependsOn': [],
         'properties': {'accountType': sku}
     }
+
+    if edge_zone:
+        storage_account['apiVersion'] = '2021-04-01'
+        storage_account['extendedLocation'] = edge_zone
+
     return storage_account
 
 
-def build_public_ip_resource(cmd, name, location, tags, address_allocation, dns_name, sku, zone, count=None):
+def build_public_ip_resource(cmd, name, location, tags, address_allocation, dns_name, sku, zone, count=None,
+                             edge_zone=None):
     public_ip_properties = {'publicIPAllocationMethod': address_allocation}
 
     if dns_name:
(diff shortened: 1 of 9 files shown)


yonzhan:
Compute

src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
```diff
@@ -172,6 +184,10 @@ def build_nic_resource(_, name, location, tags, vm_name, subnet_id, private_ip_a
             'count': count
         }
 
+    if edge_zone:
+        nic['extendedLocation'] = edge_zone
+        nic['apiVersion'] = '2021-02-01'
```
Juliehzl, line 189: hard-coded api version?

Juliehzl: should we get specified api version from resource type and check if the api version is supported?

zhoxing-ms: Good question! I've thought about this, and my personal opinion is:

Before, the api-versions of dependent resources on are fixed and independent, so it might be better to Keep the api-versions which support edge zone independent.
Otherwise, there will be several problems:
1. If the api-version of dependent resources is upgraded, the VM related tests need to be recorded, which increases the coupling.
2. The api-version used by dependent resources may tend to be stable and fixed. I think this is the reason why the previous api-version used by dependent resources is the lowest version. [code link](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/vm/_template_builder.py#L72)
3. If other profiles also need to support edge zone in the future, if the api-versions of their dependent resources have not been upgraded in that profile, but we need to give higher priority to support VM and VMSS, then we can only bump version for all the dependent resources together. This workload is very heavy and it is not conducive to split.

In addition, similar situations have been dealt with in this way before: [code link](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/vm/_template_builder.py#L147-L155)

As for why not add the check for api-version, it is because `--edge-zone` has the api-version control `min_api='2020-12-01'` [code link](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/vm/_params.py#L103)

That’s just my two cents, do you have any ideas or suggestions for this?

Juliehzl: LGTM then

Juliehzl (approved):

PR #18748, opened 2021-07-06
[App Service] `az appservice ase create`: Command changes for ASE v3 GA

madsd:
**Description**
For ASEv3 GA, there are some changes, most noticeably that External ASE and Zone redundancy is now supported, and the need for private endpoint for inbound traffic is removed and now using an Azure Load Balancer like ASEv2.

ase create-inbound-services will now only create a Private DNS Zone thus the --skip-dns is not relevant. The API for getting inbound IP for ASEv3 is in azure-mgmt-web 3.0.0 and require an SDK upgrade before this can be supported.

**Testing Guide**
az appservice ase create --kind asev3 [--zone-redundant] ...

**History Notes**

[AppService] az appservice ase create: Support for ASEv3 External and Zone redundancy

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/_help.py
@@ -2169,7 +2169,7 @@ helps['appservice ase show'] = """
 
 helps['appservice ase list-addresses'] = """
     type: command
-    short-summary: List VIPs associated with an app service environment.
+    short-summary: List VIPs associated with an app service environment v2.
     examples:
     - name: List VIPs for an app service environments.
       text: az appservice ase list-addresses --name MyAseName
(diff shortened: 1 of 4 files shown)


yonzhan:
AppService

src/azure-cli/azure/cli/command_modules/appservice/_help.py
```diff
@@ -2186,42 +2191,32 @@
       text: |
           az group create -g MyResourceGroup --location westeurope
 
-          az network vnet create -g MyResourceGroup -n MyVirtualNetwork \\
-            --address-prefixes 10.0.0.0/16 --subnet-name MyAseSubnet --subnet-prefixes 10.0.0.0/24
+          az network vnet create -g MyResourceGroup -n MyVirtualNetwork --address-prefixes 10.0.0.0/16 --subnet-name MyAseSubnet --subnet-prefixes 10.0.0.0/24
```
zhoxing-ms, line 2194: I'd like to ask why you delete the line break

src/azure-cli/azure/cli/command_modules/appservice/appservice_environment.py
```diff
@@ -140,34 +141,23 @@ def create_ase_inbound_services(cmd, resource_group_name, name, subnet, vnet_nam
...
-        LongRunningOperation(cmd.cli_ctx)(poller)
-        ase_pe = poller.result()
-        nic_name = parse_resource_id(ase_pe.network_interfaces[0].id)['name']
-        nic = network_client.network_interfaces.get(resource_group_name, nic_name)
-        inbound_ip_address = nic.ip_configurations[0].private_ip_address
-    elif ase.kind.lower() == 'asev2':
-        if ase.internal_load_balancing_mode == 0:
-            raise ValidationError('Private DNS Zone is not relevant for External ASEv2.')
-        ase_vip_info = ase_client.get_vip_info(resource_group_name, name)
-        inbound_ip_address = ase_vip_info.internal_ip_address
```
zhoxing-ms, line 166: Why remove support for `asev3`. Will this cause breaking change to users?

zhoxing-ms:
Could you add some tests to these changes?

src/azure-cli/azure/cli/command_modules/appservice/appservice_environment.py
madsd: The use of private endpoint for ASEv3 was only relevant during preview. For GA we moved to using internal Azure load balancer and therefore the section is not relevant anymore. This command is now only an additional help to setup private DNS zone if Cx would like that. For ASEv3 we have to wait for an update of the SDK to 3.0.0 before we can support this.
So nothing broken. And the command is still flagged as preview.

src/azure-cli/azure/cli/command_modules/appservice/_help.py
madsd: Thought it would make copy/paste experience better, but it did not, so I added them back in.

madsd:
Hi @zhoxing-ms, I have addressed your comments. Thank you for looking into the changes.

src/azure-cli/azure/cli/command_modules/appservice/appservice_environment.py
zhoxing-ms: Got it, thanks~

zhoxing-ms (approved):

PR #18793, opened 2021-07-09
{Redis} Migrate to Track 2 Mgmt SDK

gorandom:
**Description**
This PR upgrades the azure-mgmt-redis library to version 12.0.0 - the new track 2 SDK.

Added new custom command wrappers to ensure backward compatibility.

Added new test cases to improve coverage.


yonzhan:
Redis

src/azure-cli/azure/cli/command_modules/redis/commands.py
```diff
@@ -28,26 +28,26 @@ def load_command_table(self, _):
...
 
     with self.command_group('redis patch-schedule', redis_patch) as g:
-        g.command('create', 'create_or_update')
-        g.command('update', 'create_or_update')
-        g.command('delete', 'delete')
-        g.show_command('show', 'get')
+        g.custom_command('create', 'cli_redis_patch_schedule_create_or_update', client_factory=cf_patch_schedules)
+        g.custom_command('update', 'cli_redis_patch_schedule_create_or_update', client_factory=cf_patch_schedules)
+        g.custom_command('delete', 'cli_redis_patch_schedule_delete', client_factory=cf_patch_schedules)
+        g.custom_show_command('show', 'cli_redis_patch_schedule_get', client_factory=cf_patch_schedules)
```
alfantp, line 46: why do this? We already defined the comman group as redis_patch which corresponds to cf_patch_schedules.

src/azure-cli/azure/cli/command_modules/redis/commands.py
```diff
@@ -28,26 +28,26 @@ def load_command_table(self, _):
 
     with self.command_group('redis', redis_sdk) as g:
         g.custom_command('create', 'cli_redis_create', client_factory=cf_redis)
-        g.command('delete', 'delete', confirmation=True)
+        g.command('delete', 'begin_delete', confirmation=True)
         g.custom_command('export', 'cli_redis_export')
-        g.command('force-reboot', 'force_reboot')
-        g.command('import-method', 'import_data', deprecate_info=g.deprecate(redirect='redis import', hide=True))
-        g.command('import', 'import_data')
+        g.custom_command('force-reboot', 'cli_redis_force_reboot', client_factory=cf_redis)
```
alfantp, line 33: why do this? We already defined the command groups in the beginning so that we dont have to add factory prefix to each commands. Why are we changing this?

src/azure-cli/azure/cli/command_modules/redis/custom.py
```diff
@@ -125,7 +125,21 @@ def cli_redis_create_server_link(cmd, client, resource_group_name, name, server_
...
+    return client.begin_create(resource_group_name, name, cache_to_link.name, params)
+
+
+def cli_redis_patch_schedule_create_or_update(client, resource_group_name, name, schedule_entries):
+    from azure.mgmt.redis.models import RedisPatchSchedule
+    param = RedisPatchSchedule(schedule_entries=schedule_entries)
+    return client.create_or_update(resource_group_name, name, "default", param)
+
+
+def cli_redis_patch_schedule_get(client, resource_group_name, name):
```
alfantp, line 137: Do we need to make these changes too? Ideally if we just define it with the right factory in commands.py, it will auto get from the SDK without a redirection here. Redisrection is required only if you want to do special cases. like first getting a Resource group before passing the name etc

gorandom: The reason for redirection is to avoid a new required param "default".
The new SDK has a mandatory parameter "default" and it can take only one value "default".
I think it is the result of adding the new param in our api spec.
To avoid this mandatory parameter in CLI and to ensure backward compatibility, we have this redirection with that param included implicitly.

src/azure-cli/azure/cli/command_modules/redis/commands.py
gorandom: Modified

jsntcy (approved):
:shipit:

PR #18826, opened 2021-07-13
[Storage] Fix issue #18399: storage account network-rule remove fails to remove subnet but cli returns success and exit code 0 

### LINKED ISSUE #18399: `storage account network-rule remove` fails to remove subnet but cli returns success and exit code 0
**Command Name**
`az storage account network-rule remove`

**Errors:**
No errors. The command succeeds and the CLI reports success.
The network rule removal does not occur though. The request technically fails.

## To Reproduce:

* create a Resource Group which a snakeCase or CamelCase name. It's important that the RG name is a mix of upper/lower chars.
* create a Storage Account
* configure its default network access rule to `deny`
* create a vNet and a Subnet.
* add a storage endpoint to the Subnet
* Update the Storage Account network access rules to allow the Subnet access to the SA.

* run the following command with the necessary variables in the parameters.
`az storage account network-rule remove --resource-group {} --account-name {} --subnet {} --debug`

Test 1:
* with the Resource ID in the `--subnet` parameter, use the correct casing for the RG name in the ID. Note that the Azure CLI returns that the command is successful.
* Use the ARM GUI and inspect the network access rules for the SA.
* Witness that the Subnet resource is still listed with access.

Test 2:
* with the Resource ID in the `--subnet` parameter, use all lower case for the RG name in the ID. Note that the Azure CLI returns that the command is successful.
* Use the ARM GUI and inspect the network access rules for the SA.
* Witness that the Subnet resource has been removed.

## Expected Behavior

Well, thats a good question. Since the same network-rule removal command works with a keyvault, regardless of the casing used in the Resource Group name in the Subnet Resource ID, you could argue that the command for a storage account should also work. 

BUT, you can also argue that the network-rule removal command for the storage account should FAIL because the lowered RG name in the Subnet Resource ID is incorrect. If you agree with this then the Azure CLI should report that failure and not respond that the command was successful.

Feels like there are two errors here.

* something, somewhere is munging the casing of the Resource Group name in the Subnet Resource ID.
* The Azure CLI is reporting a command was successful when in fact it wasnt.

## Environment Summary
```
macOS-11.4-x86_64-i386-64bit
Python 3.8.10
Installer: HOMEBREW

azure-cli 2.24.2

Extensions:
azure-devops 0.18.0
resource-graph 1.1.0
subscription 0.1.4
account 0.2.1
db-up 0.1.13

```

Juliehzl:
**Description**

Fix #18399 
In previous design, when the specified subnet/ip/resource id is not in existing network rules, we will pass without any message.
Here is to raise exception for rules non-existing and provide rules information.

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/storage/operations/account.py
+++ b/src/azure-cli/azure/cli/command_modules/storage/operations/account.py
@@ -528,18 +528,19 @@ def remove_network_rule(cmd, client, resource_group_name, account_name, ip_addre
     rules = sa.network_rule_set
     if subnet:
         if all(not x.virtual_network_resource_id.endswith(subnet) for x in rules.virtual_network_rules):
-            raise InvalidArgumentValueError("Specified subnet {} is not in the storage account network rules. "
-                                            "Found subnets are {}. Note: resource id is case sensitive.".format(
-                subnet, [x.virtual_network_resource_id for x in rules.virtual_network_rules]))
+            raise InvalidArgumentValueError(
+                "Specified subnet {} is not in the storage account network rules. Found subnets are {}. Note: "
+                "resource id is case sensitive.".format(
+                    subnet, [x.virtual_network_resource_id for x in rules.virtual_network_rules]))
         rules.virtual_network_rules = [x for x in rules.virtual_network_rules
                                        if not x.virtual_network_resource_id.endswith(subnet)]
     if ip_address:
         IpRule = cmd.get_models('IPRule')
         ip_rule = IpRule(ip_address_or_range=ip_address)
         if ip_rule not in rules.ip_rules:
-            raise InvalidArgumentValueError("Specified ip address {} is not in the storage account network rules. "
-                                            "Found ip addresses are {}.".format(
-                ip_address, [x.ip_address_or_range for x in rules.ip_rules]))
+            raise InvalidArgumentValueError(
+                "Specified ip address {} is not in the storage account network rules. Found ip addresses are {}.".
+                format(ip_address, [x.ip_address_or_range for x in rules.ip_rules]))
         rules.ip_rules = [x for x in rules.ip_rules if x.ip_address_or_range != ip_address]
 
     if resource_id:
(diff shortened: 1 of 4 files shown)


yonzhan:
Storage

src/azure-cli/azure/cli/command_modules/storage/operations/account.py
```diff
@@ -522,17 +522,36 @@ def add_network_rule(cmd, client, resource_group_name, account_name, action='All
...
-        rules.resource_access_rules = [x for x in rules.resource_access_rules if
-                                       not (x.tenant_id == tenant_id and x.resource_id == resource_id)]
+        ResourceAccessRule = cmd.get_models('ResourceAccessRule')
+        resource_access_rule = ResourceAccessRule(resource_id=resource_id, tenant_id=tenant_id)
+        if resource_access_rule not in rules.resource_access_rules:
+            raise InvalidArgumentValueError(
+                "Specified resource access rule with resource id {} and tenant id {} is not in the storage account "
+                "network rules. Found resource access rules are {}. Note: resource id is case sensitive.".format(
+                    resource_id, tenant_id, [todict(x) for x in rules.resource_access_rules]))
+        rules.resource_access_rules = [x for x in rules.resource_access_rules if x != resource_access_rule]
```
evelyn-ys, line 554: Can we compare the length after filter with the origin length to avoid one more traverse?

Just small suggestion :)

Juliehzl: good point. will update it with new method.

evelyn-ys (approved):

Juliehzl:
leave the PR here to discuss if it is required to introduce the change. because it is not critical issue but the fix will introduce braking change.

src/azure-cli/azure/cli/command_modules/storage/operations/account.py
```diff
@@ -522,17 +522,36 @@ def add_network_rule(cmd, client, resource_group_name, account_name, action='All
...
+    from azure.cli.core.azclierror import InvalidArgumentValueError
+    from knack.util import todict
     sa = client.get_properties(resource_group_name, account_name)
     rules = sa.network_rule_set
     if subnet:
+        if all(not x.virtual_network_resource_id.endswith(subnet) for x in rules.virtual_network_rules):
+            raise InvalidArgumentValueError(
+                "Specified subnet {} is not in the storage account network rules. Found subnets are {}. Note: "
+                "resource id is case sensitive.".format(
+                    subnet, [x.virtual_network_resource_id for x in rules.virtual_network_rules]))
```
evelyn-ys, line 534:
```suggestion
        if all(not x.virtual_network_resource_id.endswith(subnet) for x in rules.virtual_network_rules):
            logger.warning(
                "Specified subnet {} is not in the storage account network rules. Found subnets are {}. Note: "
                "resource id is case sensitive.".format(
                    subnet, [x.virtual_network_resource_id for x in rules.virtual_network_rules]))
            return None
```

PR #18873, opened 2021-07-15
fix example syntax to az deployment group create

gitstua:
AZ CLI Example Fix _az group deployment_ to _az deployment group_

**Description**
The samples use deprecated syntax az group deployment create

**Testing Guide**
run examples

**History Notes**
Non-breaking change - documentation update only
---

- [x ] The PR title and description has followed the guideline in [Submitting Pull Requests](https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules#submitting-pull-requests).

- [x ] I adhere to the [Command Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md).


yonzhan:
fix example syntax

zhoxing-ms:
In fact, the deprecated commands uses their corresponding help, which is by designed here.
Because we need to ensure that the help of the command and the command are consistent, otherwise the user will feel that the help information does not match.
And we already have other guidance prompts for users to use new commands, so we don't need such a way to recommend users to use new commands.

PR #18907, opened 2021-07-17
[App Config] BREAKING CHANGE: `appconfig kv set`: Set value to be empty string if not specified 

shenmuxiaosen:
**Description**
Change the existing behavior for ```appconfig kv set``` command. When creating a new kv, if no```value``` or ```content-type``` specified, then default to be empty string instead of null (existing behavior). This change is to keep consistent with our Portal. 

Same thing for ```appconfig feature set``` command ```description``` argument. 

**Testing Guide**

**History Notes**

[App Config] BREAKING CHANGE: `appconfig kv set`: Set the value of parameter `--value` or `--content-type`  to be empty string if it is not specified
[App Config] BREAKING CHANGE: `appconfig feature set`: Set the value of parameter `--description` to empty string if it is not specified

---


shenmuxiaosen:
@avanigupta Can you help to take a look at this pr?

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -334,23 +334,23 @@ def set_key(cmd,
             raise CLIError("Failed to retrieve key-values from config store. " + str(exception))
 
         if retrieved_kv is None:
-            if content_type and __is_json_content_type(content_type):
+            if __is_json_content_type(content_type):
                 try:
                     # Ensure that provided value is valid JSON. Error out if value is invalid JSON.
-                    value = 'null' if value is None else value
+                    value = '""' if value is None else value
```
avanigupta, line 340: Since this is a JSON key-value, the default value should be null because we don't want to assume that its a string value.

yonzhan:
AppConfiguration

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
shenmuxiaosen: Reverted.

avanigupta (approved):

shenmuxiaosen:
@zhoxing-ms @Juliehzl Can you help to take a look?

zhoxing-ms (approved):

PR #18924, opened 2021-07-19
[ARM] `az deployment group/sub/mg/tenant create`: Add `--what-if` and `--proceed-if-no-change` parameters

### LINKED ISSUE #18481: az deployment * create --confirm-with-what-if asks for confirmation when nothing is changed
**Describe the bug**
When using any `az deployment * create` command with `--confirm-with-what-if`, if there are no changes (excluded using `--what-if-exclude-change-types NoChange Ignore`), Azure CLI still asks to confirm the execution of the deployment.

**To Reproduce**
Deploy any template with `az deployment * create --confirm-with-what-if --what-if-exclude-change-types NoChange Ignore`, then deploy it again without changing anything.

**Expected behavior**
One of two things:
- Either the deployment does not ask for confirmation when "Resource changes: no change." is the what-if output
- A parameter is made available that if there is no change as the what-if output, then do not ask for confirmation and simply proceed

**Environment summary**
Azure CLI v2.24.2 running in CentOS7 & Ubuntu 20.04LTS via WSL2, and in CentOS7 via Docker container on WSL2
OS: Windows 10 Enterprise 20H2

**Additional context**
This is mostly a problem because we have a lot of templates, and a good chunk of them are multistage templates, i.e.: We will run `az deployment * create` with a parameter `-p stage=X` via a custom Bash script that will trigger different sections of our templates based on the stage number. 
This is needed because let's say Stage 2 might create Resource A, which we then need to get some information in Bash to do some manipulations outside of the ARM template OR to pass as additional parameters for later stages (i.e.: Deploy a storage account, use Azure CLI to fetch the Connection String, sanitise it, and pass it as a deployment parameter for a later stage). If those manipulations are not done, then Stage 3 resources will fail or not work properly.

We have up to 4 stages in our templates currently, and sometimes only one of the stages might have changes. Needing to confirm that nothing is being changed every single time is not really ideal in terms of user experience for the rest of our team.

And because of the above requirements, switching to using `az deployment * what-if` is not an option, because resources that would've been created by Stage X won't be available for a later stage, thus failing it.

shenglol:
**Description**
Closes #18481.
Closes https://github.com/Azure/bicep/issues/2923

**Testing Guide**
```
az deployment sub -l westus -f ./deploy.json --what-if
az deployment sub -l westus -f ./deploy.json --confirm-with-what-if --proceed-if-no-change
```

**History Notes**

[ARM] az deployment group/sub/mg/tenant create: Add --what-if parameter for invoking What-If with the deployment create commands.
[ARM] az deployment group/sub/mg/tenant create: Add --proceed-if-no-change parameter to skip confirmation when --confirm-with-what-if is set and there's no changes in What-If results.

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/resource/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/resource/_params.py
@@ -62,11 +62,11 @@ def load_arguments(self, _):
                                 'When the value is true, the prompt requiring users to provide missing parameter will be ignored. The default value is false.')
 
     deployment_what_if_type = CLIArgumentType(options_list=['--what-if', '-w'], action='store_true',
-                                                           help='Instruct the command to run deployment What-If.',
-                                                           min_api='2019-07-01')
+                                              help='Instruct the command to run deployment What-If.',
+                                              min_api='2019-07-01')
     deployment_what_if_proceed_if_no_change_type = CLIArgumentType(options_list=['--proceed-if-no-change'], action='store_true',
-                                                           help='Instruct the command to execute the deployment if the What-If result contains no resource changes. Applicable when --confirm-with-what-if is set.',
-                                                           min_api='2019-07-01')
+                                                                   help='Instruct the command to execute the deployment if the What-If result contains no resource changes. Applicable when --confirm-with-what-if is set.',
+                                                                   min_api='2019-07-01')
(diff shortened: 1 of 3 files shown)


yonzhan:
ARM

zhoxing-ms (approved):

PR #18966, opened 2021-07-22
[IoT] `az iot hub create\update`: File upload and certificate updates

c-ryan-k:
**Description**

* Adds some checks to prevent bad IoT Hub file upload property states regarding managed identities.
* Adds `--fileupload-notification-lock-duration` parameter to `iot hub update` and `iot hub create`
* Deprecates `fileupload-storage-container-uri` parameter as it's not being utilized correctly.
* Changes all certificate upload commands to use base64-encoding of files.

**Testing Guide**

**History Notes**

[IoT] `az iot hub create/update`: Added checks to prevent bad file-upload identity parameters when hub doesn't have identity
[IoT] `az iot hub create/update`: Added `--fileupload-notification-lock-duration` parameter
[IoT] `az iot hub create/update`: Deprecated `fileupload-storage-container-uri` parameter
[IoT] `az iot dps/hub certificate create`: Certificates will now always be uploaded in base64 encoding.

---


yonzhan:
IoT

src/azure-cli/azure/cli/command_modules/iot/custom.py
```diff
@@ -425,13 +426,13 @@ def iot_hub_create(cmd, client, hub_name, resource_group_name, location=None,
...
     identity_based_file_upload = fileupload_storage_authentication_type and fileupload_storage_authentication_type.lower() == AuthenticationType.IdentityBased.value
-    if not identity_based_file_upload and not fileupload_storage_connectionstring and fileupload_storage_container_name:
-        raise RequiredArgumentMissingError('Key-based authentication requires a connection string.')
-    if identity_based_file_upload and not fileupload_storage_container_uri:
-        raise RequiredArgumentMissingError('Identity-based authentication requires a storage container uri (--fileupload-storage-container-uri, --fcu).')
     if not identity_based_file_upload and fileupload_storage_identity:
         raise RequiredArgumentMissingError('In order to set a fileupload storage identity, please set file upload storage authentication (--fsa) to IdentityBased')
-
+    if identity_based_file_upload or fileupload_storage_identity:
+        if fileupload_storage_identity == SYSTEM_ASSIGNED_IDENTITY and not system_identity:
```
digimaun, line 432: nit, the next `if` statement first checks if `fileupload_storage_identity` is a truthy value and then does the value comparison....but this `if` statement skips the truthy check.

src/azure-cli/azure/cli/command_modules/iot/custom.py
```diff
@@ -552,12 +555,33 @@ def update_iot_hub_custom(instance,
...
+        lock_duration = timedelta(seconds=fileupload_notification_lock_duration)
+        instance.properties.messaging_endpoints['fileNotifications'].lock_duration_as_iso8601 = lock_duration
     if fileupload_notification_max_delivery_count is not None:
         count = fileupload_notification_max_delivery_count
         instance.properties.messaging_endpoints['fileNotifications'].max_delivery_count = count
     if fileupload_notification_ttl is not None:
         ttl = timedelta(hours=fileupload_notification_ttl)
         instance.properties.messaging_endpoints['fileNotifications'].ttl_as_iso8601 = ttl
+    # if setting a fileupload storage identity or changing fileupload to identity-based
+    if fileupload_storage_identity or (fileupload_storage_authentication_type and fileupload_storage_authentication_type.lower() == AuthenticationType.IdentityBased.value):
```
digimaun, line 568: Could avoid the `lower()` call in `fileupload_storage_authentication_type.lower()` if `iot/shared.py` attributes had consistent casing with API values.

src/azure-cli/azure/cli/command_modules/iot/custom.py
c-ryan-k: Ok - thanks for pointing this out. Just realized this is because a user can set `--fsa` to enable identityBased file upload - but realized there's a gap in the check here because technically `--fsi` is *not* required (left empty it assumes system assigned identity). I'll update the logic for this.

src/azure-cli/azure/cli/command_modules/iot/custom.py
c-ryan-k: You're right - I think this was done to allow an easier comparison for input casing, can't remember if `get_enum_type` automatically handles input casing or not. Will investigate

c-ryan-k: Ok cool, switching the input type to `get_enum_type(AuthenticationType)` will handle that much better - no more `lower()` checks and it should handle casing / param help as well.

src/azure-cli/azure/cli/command_modules/iot/_utils.py
```diff
@@ -45,10 +45,7 @@ def open_certificate(certificate_path):
     if certificate_path.endswith('.pem') or certificate_path.endswith('.cer'):
         with open(certificate_path, "rb") as cert_file:
             certificate = cert_file.read()
-            try:
-                certificate = certificate.decode("utf-8")
-            except UnicodeError:
-                certificate = base64.b64encode(certificate).decode("utf-8")
+        certificate = base64.b64encode(certificate).decode("utf-8")
```
digimaun, line 48: Just a note to double-check we can cover at least the same cert inputs as before.

digimaun (approved):

src/azure-cli/azure/cli/command_modules/iot/_validators.py
```diff
@@ -53,7 +53,13 @@ def validate_feedback_ttl(ns):
 def validate_feedback_lock_duration(ns):
     if (ns.feedback_lock_duration and
             ns.feedback_lock_duration not in range(5, 301, 1)):
-        raise ArgumentError(None, 'Please specify the duration from 5 to 300 seconds only.')
+        raise ArgumentError(None, 'Please specify the feedback lock duration from 5 to 300 seconds only.')
```
zhoxing-ms, line 56: Please use the specific error type `InvalidArgumentValueError` instead of `ArgumentError`

src/azure-cli/azure/cli/command_modules/iot/_validators.py
```diff
@@ -53,7 +53,13 @@ def validate_feedback_ttl(ns):
...
+
+
+def validate_fileupload_notification_lock_duration(ns):
+    if (ns.fileupload_notification_lock_duration and
+            ns.fileupload_notification_lock_duration not in range(5, 301, 1)):
+        raise ArgumentError(None, 'Please specify the notification lock duration from 5 to 300 seconds only.')
```
zhoxing-ms, line 62: Please use the specific error type `InvalidArgumentValueError` instead of `ArgumentError`

src/azure-cli/azure/cli/command_modules/iot/custom.py
```diff
@@ -813,7 +838,7 @@ def iot_hub_get_stats(client, hub_name, resource_group_name=None):
 
 
 def validate_authentication_type_input(endpoint_type, connection_string=None, authentication_type=None, endpoint_uri=None, entity_path=None):
-    is_keyBased = (AuthenticationType.KeyBased.value == authentication_type.lower()) or (authentication_type is None)
+    is_keyBased = (AuthenticationType.KeyBased.value == authentication_type) or (authentication_type is None)
```
zhoxing-ms, line 841: Why is case sensitive when comparing here?

c-ryan-k: It's not - `authentication_type` has been changed in inputs to be an enum of `AuthenticationType` - and the constant was changed to match the swagger / API value - no more need to convert to `lower()` since they'll be using the enum / constant value.

zhoxing-ms: Got it, thanks~

zhoxing-ms:
@c-ryan-k Could you change the PR title more specifically according to the format of `[Component Name 2] az command b: Add some customer-facing feature`?
  .

c-ryan-k:
> @c-ryan-k Could you change the PR title more specifically according to the format of `[Component Name 2] az command b: Add some customer-facing feature`?
> .

Sure - do you mean the actual title: `[IoT] File upload and certificate updates`
Or just the history notes?

zhoxing-ms:
@c-ryan-k Maybe we can simplify the history description, such as:

[IoT] `az iot hub create/update`: Added checks to prevent bad file-upload identity parameters when hub doesn't have identity
[IoT] `az iot hub create/update`: Added `--fileupload-notification-lock-duration` parameter
[IoT] `az iot hub create/update`: Deprecated `--fileupload-storage-container-uri` parameter
[IoT] `az iot hub/dps certificate create`: Certificates will now always be uploaded in base64 encoding.

What do you think?

c-ryan-k:
@zhoxing-ms sure no problem, just wasn't sure how specific the notes needed to be, I'll update it

c-ryan-k:
Can try to investigate the tests later, but they're passing locally so I'm not sure what's up there

zhoxing-ms (approved):

Issue #18251, opened 2021-05-26
az container exec - does not work again - TypeError: write() argument must be str, not bytes


alexeyshishkin01:
> ### `az feedback` auto-generates most of the information requested below, as of CLI version 2.0.62

**Describe the bug**

az container exec \
> --resource-group dockers \
> --name debiandocker \
> --exec-command "/bin/bash" \
> --container-name debiandocker
The command failed with an unexpected error. Here is the traceback:
write() argument must be str, not bytes
Traceback (most recent call last):
  File "/opt/az/lib/python3.6/site-packages/knack/cli.py", line 231, in invoke
    cmd_result = self.invocation.execute(args)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 657, in execute
    raise ex
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 720, in _run_jobs_serially
    results.append(self._run_job(expanded_arg, cmd_copy))
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 691, in _run_job
    result = cmd_copy(params)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 328, in __call__
    return self.handler(*args, **kwargs)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/command_operation.py", line 121, in handler
    return op(**command_args)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/command_modules/container/custom.py", line 623, in container_exec
    _start_exec_pipe(execContainerResponse.web_socket_uri, execContainerResponse.password)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/command_modules/container/custom.py", line 666, in _start_exec_pipe
    if not _cycle_exec_pipe(ws):
  File "/opt/az/lib/python3.6/site-packages/azure/cli/command_modules/container/custom.py", line 684, in _cycle_exec_pipe
    sys.stdout.write(data)
TypeError: write() argument must be str, not bytes

**To Reproduce**

To reproduce - create Azure Container and run 'az container exec'

**Expected behavior**

Expected behavior - no such issue, connect to container instead

**Environment summary**

root@Alexey-HP:~# az --version
azure-cli                         2.24.0
core                              2.24.0
telemetry                          1.0.6
Python location '/opt/az/bin/python3'
[...]

PR #18384, opened 2021-06-07
{container} Bugfix #15856,#18251,#18275: az container exec - decode received bytes as utf-8 string

bitcloud:
**Description**

The websocket response can be a string for text or bytes for an utf-8 string. utf-8 was not handled properly. This pr adds detection of bytes and decoding as utf-8 string before writing to the console.

Fixes: https://github.com/Azure/azure-cli/issues/15856
Fixes: https://github.com/Azure/azure-cli/issues/18251
Fixes: https://github.com/Azure/azure-cli/issues/18275

**Testing Guide**

- start a container e.g. with `az container create` with a long living container
- try to execute a shell inside the container with `az container exec --exec-command "/bin/bash" ...`
- currently it just exits without an error or shell
- 
**History Notes**

[container] az container exec: decode received bytes as utf-8 string

---

This checklist is used to make sure that common guidelines for a pull request are followed.

- [ ] The PR title and description has followed the guideline in [Submitting Pull Requests](https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules#submitting-pull-requests).

- [ ] I adhere to the [Command Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md).

- [ ] I adhere to the [Error Handling Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/error_handling_guidelines.md).

PR #18974, opened 2021-07-23
decode binary string as utf-8 in win branch

bitcloud:
**Description**
The websocket response can be a string for text or bytes for an utf-8 string. utf-8 was not handled properly. This pr adds detection of bytes and decoding as utf-8 string before writing to the console.
This is the same fix as https://github.com/Azure/azure-cli/pull/18384 applied to the windows code as well.

Fixes: https://github.com/Azure/azure-cli/issues/18251#issuecomment-885091314

**Testing Guide**
- start a container e.g. with `az container create` with a long living container on a Windows system
- try to execute a shell inside the container with `az container exec --exec-command "/bin/bash" ...`
- currently it just exits without an error or shell

**History Notes**
[container] az container exec: decode received bytes as utf-8 string (Windows as well now)

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/container/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/container/custom.py
@@ -641,6 +641,8 @@ def _start_exec_pipe_win(web_socket_uri, password):
 
 
 def _on_ws_msg(ws, msg):
+    if isinstance(msg, bytes):
+        msg = msg.decode('utf-8')
     sys.stdout.write(msg)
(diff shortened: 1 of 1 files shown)


bitcloud:
I couldn't test this patch as I don't have a Windows system at hand. But it should work as it seems to be the same issue we had on the linux/macos code path.

bitcloud:
There seems to be a better PR already on the way: https://github.com/Azure/azure-cli/pull/18909

PR #19146, opened 2021-08-06
[ARM] `az feature registration list/create/delete`: Add new commands to manage resource provider feature registrations

diwudd:
**Description**

Add new az feature registration apis.

**Testing Guide**

Tests have been added/recorded.

**History Notes**

[ARM] az feature registration: Add az feature registration apis

---


yonzhan:
ARM

src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -1124,6 +1124,35 @@
...
+examples:
+  - name: List feature registrations
+    text: az feature registration list
+"""
+
+helps['feature registration create'] = """
+type: command
+short-summary: create a feature registration.
+examples:
+  - name: create the "Shared Image Gallery" feature
```
zhoxing-ms, line 1144:
```suggestion
short-summary: Create a feature registration.
examples:
  - name: Create the "Shared Image Gallery" feature
```

src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -1124,6 +1124,35 @@
...
+    text: az feature registration create --namespace Microsoft.Compute --name GalleryPreview
+"""
+
+helps['feature registration delete'] = """
+type: command
+short-summary: delete a feature registration.
+examples:
+  - name: delete the "Shared Image Gallery" feature
+    text: az feature registration delete --namespace Microsoft.Compute --name GalleryPreview
```
zhoxing-ms, line 1153:
```suggestion
helps['feature registration delete'] = """
type: command
short-summary: Delete a feature registration.
examples:
  - name: Delete the "Shared Image Gallery" feature
    text: az feature registration delete --namespace Microsoft.Compute --name GalleryPreview
```

src/azure-cli/azure/cli/command_modules/resource/_params.py
```diff
@@ -160,6 +160,13 @@ def load_arguments(self, _):
     with self.argument_context('feature list') as c:
         c.argument('resource_provider_namespace', options_list='--namespace', required=False, help=_PROVIDER_HELP_TEXT)
 
+    with self.argument_context('feature registration') as c:
+        c.argument('resource_provider_namespace', options_list='--namespace', required=True, help=_PROVIDER_HELP_TEXT)
+        c.argument('feature_name', options_list=['--name', '-n'], help='the feature name')
+
+    with self.argument_context('feature registration list') as c:
+        c.argument('resource_provider_namespace', options_list='--namespace', required=False, help=_PROVIDER_HELP_TEXT)
```
zhoxing-ms, line 168: `required=False` does not need to be added, `False` is the default value for `required`

src/azure-cli/azure/cli/command_modules/resource/commands.py
```diff
@@ -233,6 +239,14 @@ def load_command_table(self, _):
         g.custom_command('register', 'register_feature')
         g.custom_command('unregister', 'unregister_feature')
 
+    with self.command_group('feature registration', resource_feature_registration_sdk, client_factory=cf_feature_registrations, resource_type=PROFILE_TYPE,
+                            min_api='2021-07-01') as g:
+        feature_table_transform = '{Name:name, RegistrationState:properties.state}'
+        g.custom_command('list', 'list_feature_registrations', table_transformer='[].' + feature_table_transform)
+        g.show_command('show', 'get', table_transformer=feature_table_transform)
+        g.custom_command('create', 'create_feature_registration')
+        g.custom_command('delete ', 'delete_feature_registration')
```
zhoxing-ms, line 248: Please add confirmation for the deletion command `confirmation=True`

src/azure-cli/azure/cli/command_modules/resource/_help.py
diwudd: sorry, what is your suggestion here? I did not see any difference

src/azure-cli/azure/cli/command_modules/resource/_help.py
diwudd: sorry, what is your suggestion here? I did not see any difference

src/azure-cli/azure/cli/command_modules/resource/commands.py
diwudd: removed

src/azure-cli/azure/cli/command_modules/resource/_params.py
diwudd: removed

diwudd: no, I have to add it back as the CI job failed " There is a violation:
 "az feature registration list" is not a valid command.
 the following arguments are required: --namespace."

diwudd: seems the default value is True other than false

src/azure-cli/azure/cli/command_modules/resource/commands.py
diwudd: after adding the confirmation=True, CI task started failing. As I have no idea how to fix it, I have to revert the change.

diwudd: What does confirmation=True do? Why did it break the CI tasks?

src/azure-cli/azure/cli/command_modules/resource/_help.py
zhoxing-ms: The first letter needs to be capitalized

src/azure-cli/azure/cli/command_modules/resource/_help.py
zhoxing-ms: The first letter needs to be capitalized

src/azure-cli/azure/cli/command_modules/resource/commands.py
zhoxing-ms: https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#keyword-argument-reference

> confirmation - During interactive use, will prompt the user to confirm their choice to proceed. Supply a value of True to use the default prompt, or supply a string to use a custom prompt message. If the command is invoked in non-interactive scenarios and the --yes/-y parameter is not supplied, the command will fail.

The purpose of this parameter is to let the user confirm the deletion operation. It should not lead to the failure of CI. Could you show the specific error message when CI fails?

src/azure-cli/azure/cli/command_modules/resource/tests/latest/test_resource.py
```diff
@@ -2156,6 +2156,21 @@ def test_feature_unregister(self):
...
+    
+    @AllowLargeResponse(8192)
+    def test_feature_registration_create(self):
+        self.cmd('feature registration create --namespace Microsoft.Network --name AllowLBPreview', checks=[
+            self.check_pattern('properties.state', 'Registering|Registered')
+        ])
+
+    @AllowLargeResponse(8192)
+    def test_feature_registration_delete(self):
+        self.cmd('feature registration delete --namespace Microsoft.Network --name AllowLBPreview')
```
zhoxing-ms, line 2173: I think the CI issue caused by adding `confirmation=True` is caused by not adding `--yes` to the delete command in this test.
```suggestion
        self.cmd('feature registration delete --namespace Microsoft.Network --name AllowLBPreview' --yes)
```

src/azure-cli/azure/cli/command_modules/resource/_params.py
zhoxing-ms: https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#keyword-argument-reference
> required - See https://docs.python.org/3/library/argparse.html#required. Note that this value is inferred from the function signature depending on whether or not the parameter has a default value. If specified, this will override that value.

zhoxing-ms: This CI issue is caused by you defining the same parameter in the parent command of the upper level and setting it to `required=True`.
Please define the parameter `resource_provider_namespace` separately for `feature registration list` and `feature registration create/delete`, not in the parent command

src/azure-cli/azure/cli/command_modules/resource/_help.py
diwudd: Thanks, addressed

src/azure-cli/azure/cli/command_modules/resource/_help.py
diwudd: Thanks, addressed

src/azure-cli/azure/cli/command_modules/resource/_params.py
diwudd: no, I cannot remove the 'requried=false' as there is top level thing 'feature' which contains 'required=True'.

Even after I remove 'feature registration', the default value will stll be true as the 'feature' has 'required=True'

src/azure-cli/azure/cli/command_modules/resource/commands.py
diwudd: thank you, seems it worked

src/azure-cli/azure/cli/command_modules/resource/tests/latest/test_resource.py
diwudd: yes, it worked with '--yes'

src/azure-cli/azure/cli/command_modules/resource/_params.py
zhoxing-ms: OK

zhoxing-ms (approved):

diwudd:
Hi Xing,

Thank you for approving the PR.

But now the CI failed due to the test failures that are not related to my changes. What should I do?

Thanks,

Di

zhoxing-ms:
@diwudd  This is a issue that the api-version of other modules are inconsistent, we will solve it for you~

zhoxing-ms:
@diwudd It has been fixed~

diwudd:
> @diwudd It has been fixed~

Thank you very much! Could you help merge the PR? I don't have access to merge it.

jsntcy (approved):

Issue #18566, opened 2021-06-21
az upgrade is blocked by policy


chriskuech:
When I run `az upgrade`, I get a prompt saying "The system administrator has set policies to prevent this installation".

Does your installer use unsigned powershell scripts?

yonzhan:
@jiasli for awareness

chriskuech:
Looks like this is only a bug in `az upgrade`.  The msi works fine for upgrading versions with the given group policy.

jiasli:
> Does your installer use unsigned powershell scripts?

No. No script is involved. The lines for installing MSI in `az upgrade` are

https://github.com/Azure/azure-cli/blob/3d14ca21e648f4fc0124132a43156a85a50251cd/src/azure-cli/azure/cli/command_modules/util/custom.py#L133-L135

Does it work if you directly call this command from Command Prompt?

```
powershell.exe -NoProfile Start-Process msiexec.exe -Wait -ArgumentList '/i https://aka.ms/installazurecliwindows'
```

Also, could you open `regedit.exe`, go to `HKLM\Software\Policies\Microsoft\Windows\Installer` and see if `DisableMSI` is configured? If so, is it possible to delete it? 

You may do the same thing in Group Policy Editor: open `gpedit.msc` and check **Local Computer Policy > Computer Configuration > Administrative Templates > Windows Components > Windows Installer > Disable Windows Installer**. If it is enabled, disable it if possible.

### References

- https://docs.microsoft.com/en-us/windows/win32/msi/disablemsi
- https://answers.microsoft.com/en-us/windows/forum/all/this-installation-is-forbidden-by-system-policy/ac1986be-a063-4b7b-834f-8fa56a8c0d04?page=1
- https://kb.vmware.com/s/article/1016563
- https://admx.help/?Category=Windows_8.1_2012R2&Policy=Microsoft.Policies.MSI::DisableMSI

vermegi:
I ended up installing/upgrading by running: 
Invoke-WebRequest -Uri https://aka.ms/installazurecliwindows -OutFile .\AzureCLI.msi; Start-Process msiexec.exe -Wait -ArgumentList '/I AzureCLI.msi /quiet'; rm .\AzureCLI.msi
As described here: https://docs.microsoft.com/en-us/cli/azure/install-azure-cli-windows?tabs=azure-powershell (Microsoft MSI with command). 
[...]

PR #19192, opened 2021-08-11
{Upgrade} Download MSI to a local file first

### LINKED ISSUE #19171: `az upgrade` fails on Windows 10: The system administrator has set policies to prevent this installation.
## Describe the bug

When I run `az upgrade`, the MSI is downloaded and script generated (first progress dialog) but then I get an error shown below. I am on a Microsoft-managed (employee) machine, so there may be additional policies in place.

**Command Name**
`az upgrade`

**Errors:**

image

The MSI log (I have logging enabled with "voicewarmupx") shows:
```
MSI (c) (58:C8) [11:07:07:434]: Using WinHttp to perform URL download
MSI (c) (58:C8) [11:07:07:435]: File path is a URL. Downloading file. . .
MSI (c) (58:C8) [11:07:07:452]: Msi WinHttp: Performing auto proxy detection
MSI (c) (58:C8) [11:07:07:455]: MSI WinHttp: Proxy Settings Proxy: (none) | Bypass: (none) | AccessType: 0
MSI (c) (58:C8) [11:07:08:374]: StatusCode is:[200]
MSI (c) (58:C8) [11:07:08:377]: Downloading https://aka.ms/installazurecliwindows to local file C:\Users\heaths\AppData\Local\Temp\MSIEA5E.tmp
MSI (c) (58:C8) [11:07:12:144]: Machine policy value 'DisableUserInstalls' is 0
MSI (c) (58:C8) [11:07:12:284]: SOFTWARE RESTRICTION POLICY: Verifying package --> 'https://aka.ms/installazurecliwindows' against software restriction policy
MSI (c) (58:C8) [11:07:12:288]: 
MSI (c) (58:C8) [11:07:12:290]: Deleting URL local copy C:\Users\heaths\AppData\Local\Temp\MSIEA5E.tmp.
MSI (c) (58:C8) [11:07:12:295]: MainEngineThread is returning 1625
```

The existing package, 2.26.1, is installed per-machine. The new package, 2.27.0, also has ALLUSERS=1 authored in the Property dialog so this should be allowed. In the past, I was prompted for UAC confirmation, but even now running `az upgrade` from an elevated prompt fails. 

Installing the MSI myself succeeds.

## To Reproduce:

- `az upgrade`

## Expected Behavior

Installs, which it did some time ago (maybe a few months ago).

## Environment Summary
```
Windows-10-10.0.19041-SP0
Python 3.8.9
Installer: MSI

azure-cli 2.26.1 *

Extensions:
azure-devops 0.18.0

```

jiasli:
## Symptom

Fix #19171, #18566

Directly installing from URL may be blocked by policy: https://github.com/Azure/azure-cli/issues/19171. The exact reason for the failure is unclear, and installing MSI from a URL is always problematic.

## Change

Download MSI to a temp folder and install it with `msiexec.exe`. This also gives the user a chance to manually install the MSI in case of `msiexec.exe` failure.

## Testing Guide

In `upgrade_version`, after retrieving the `installer`:

https://github.com/Azure/azure-cli/blob/c100b8026cdb4261b773c44cb0551285734aac84/src/azure-cli/azure/cli/command_modules/util/custom.py#L62

force

```py
local_version = '2.26.0'
installer = 'MSI'
```

so that `_upgrade_on_windows` can be triggered.

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/util/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/util/custom.py
@@ -3,7 +3,10 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 
+import os
+
 from knack.log import get_logger
+from knack.util import CLIError
 
 logger = get_logger(__name__)
 
@@ -32,7 +35,6 @@ def show_version(cmd):  # pylint: disable=unused-argument
 
 
 def upgrade_version(cmd, update_all=None, yes=None):  # pylint: disable=too-many-locals, too-many-statements, too-many-branches, no-member, unused-argument
-    import os
     import platform
     import sys
     import subprocess
@@ -41,7 +43,6 @@ def upgrade_version(cmd, update_all=None, yes=None):  # pylint: disable=too-many
     from azure.cli.core._environment import _ENV_AZ_INSTALLER
     from azure.cli.core.extension import get_extensions, WheelExtension
     from packaging.version import parse
-    from knack.util import CLIError
 
     update_cli = True
     from azure.cli.core.util import get_latest_from_github
@@ -131,8 +132,7 @@ def upgrade_version(cmd, update_all=None, yes=None):  # pylint: disable=too-many
             logger.warning("Exit the container to pull latest image with 'docker pull mcr.microsoft.com/azure-cli' "
                            "or run 'pip install --upgrade azure-cli' in this container")
         elif installer == 'MSI':
-            logger.debug("Update azure cli with MSI from https://aka.ms/installazurecliwindows")
-            exit_code = subprocess.call(['powershell.exe', '-NoProfile', "Start-Process msiexec.exe -Wait -ArgumentList '/i https://aka.ms/installazurecliwindows'"])  # pylint: disable=line-too-long
+            exit_code = _upgrade_on_windows()
         else:
             logger.warning(UPGRADE_MSG)
     if exit_code:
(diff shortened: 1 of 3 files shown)


src/azure-cli/azure/cli/command_modules/util/custom.py
```diff
@@ -175,6 +175,51 @@ def upgrade_version(cmd, update_all=None, yes=None):  # pylint: disable=too-many
                    else auto_upgrade_msg)
 
 
+def _upgrade_on_windows():
+    """Download MSI to a temp folder and install it with msiexec.exe.
+    Directly installing from URL may be blocked by policy: https://github.com/Azure/azure-cli/issues/19171
+    This also gives the user a chance to manually install the MSI in case of msiexec.exe failure.
+    """
+    logger.warning("Updating Azure CLI with MSI from https://aka.ms/installazurecliwindows")
```
heaths, line 183: This was `logger.debug` before. Intentional change?

src/azure-cli/azure/cli/command_modules/util/custom.py
```diff
@@ -175,6 +175,51 @@ def upgrade_version(cmd, update_all=None, yes=None):  # pylint: disable=too-many
...
+    r = requests.get(url, stream=True, verify=(not should_disable_connection_verify()))
+    if r.status_code != 200:
+        raise CLIError("Request to {} failed with {}".format(url, r.status_code))
+
+    # r.url is the real path of the msi, like'https://azcliprod.blob.core.windows.net/msi/azure-cli-2.27.1.msi'
+    file_name = r.url.rsplit('/')[-1]
+    import tempfile
+    tmp_dir = tempfile.mkdtemp()
+    msi_path = os.path.join(tmp_dir, file_name)
+    logger.warning("Downloading MSI to %s", msi_path)
```
heaths, line 211: Similarly, isn't this more of a `logger.debug`?

heaths (approved):
A couple nits, but overall LGTM. Thanks!

src/azure-cli/azure/cli/command_modules/util/custom.py
jiasli: Yes.

I want the command to be more explicit and give the user more information about what it is doing internally. Otherwise, in case of failure, all that the user can see is `CLI upgrade failed`. (No different from `Something occurred`.)

To view the details, the user must specify `--debug` which is definitely not user-friendly.

src/azure-cli/azure/cli/command_modules/util/custom.py
jiasli: Same as above.

evelyn-ys (approved):

calvinhzy (approved):

src/azure-cli/azure/cli/command_modules/util/custom.py
```diff
@@ -131,8 +132,7 @@ def upgrade_version(cmd, update_all=None, yes=None):  # pylint: disable=too-many
             logger.warning("Exit the container to pull latest image with 'docker pull mcr.microsoft.com/azure-cli' "
                            "or run 'pip install --upgrade azure-cli' in this container")
         elif installer == 'MSI':
-            logger.debug("Update azure cli with MSI from https://aka.ms/installazurecliwindows")
-            exit_code = subprocess.call(['powershell.exe', '-NoProfile', "Start-Process msiexec.exe -Wait -ArgumentList '/i https://aka.ms/installazurecliwindows'"])  # pylint: disable=line-too-long
```
jiasli, line 135: There is no need to launch PowerShell and let it launch `msiexec.exe`. Directly launching `msiexec.exe` is sufficient.

Issue #17531, opened 2021-03-31
WebApp:az webapp deployment slot create DOESN'T duplicate all settings from source slot


panmail:
> ### `az feedback` auto-generates most of the information requested below, as of CLI version 2.0.62

**Describe the bug**
The following powershell command doesn't duplicate all settings from the source slot.
`az webapp deployment slot create --name 'newSlotName' --resource-group 'rscGroup' --slot 'test' --configuration-source 'srcSlotName' --subscription 'subID' --debug`

**To Reproduce**
run command above and found that the source slot setting of "ASPNETCORE_ENVIRONMENT" was not duplicated over to the new slot.

**Expected behavior**

**Environment summary**

I ran this in Win10 PS ISE, Azure portal powershell with same result.

**Additional context**

panchagnula:
@panmail thanks for reporting - out of curiosity is this specific setting - "set as slot Setting or not"? I believe only settings marked slot settings should be copied over & not others - portal doesn't follow this rule. Can you please share this info.

@Kotasudhakarreddy  can you or Parveen test this behavior & share your notes as well? Thank you.

panmail:
@panchagnula, I can confirm that if using PS, then 'yes' ONLY non-slot settings are cloned over; however, when using the Azure portal UI, then ALL (slot and non-slot) settings are cloned over.  I would like to see PS to have same behavior as portal.
![image](https://user-images.githubusercontent.com/7851873/113184443-0835bd80-920a-11eb-89d0-94b863925b7e.png)

panchagnula:
Making this duplicate of original issue #6638

Issue #6638, opened 2018-06-22
Parameter --configuration-source does not copy app settings and connection strings


cornem:
I expected the `--configuration-source` parameter on `az webapp deployment slot create` command to copy app settings and connection strings to the new slot, but it doesn't.

Is there another way to achieve this using the CLI? Using the Azure Portal, it is possible to clone the configuration from another slot. It would be really useful if this functionality were also available from the CLI.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: ce9f5bf9-d178-de1a-645b-be6947a7c049
* Version Independent ID: cb7fb876-160c-e6d9-d473-7ac39b930b48
* Content: [az webapp deployment slot](https://docs.microsoft.com/en-us/cli/azure/webapp/deployment/slot?view=azure-cli-latest)
* Content Source: [latest/docs-ref-autogen/webapp/deployment/slot.yml](https://github.com/Azure/azure-docs-cli-python/blob/live/latest/docs-ref-autogen/webapp/deployment/slot.yml)
* Service: **app-service-web**
* GitHub Login: @rloutlaw
* Microsoft Alias: **routlaw**

yugangw-msft:
`--configuration-source` is the way to copy settings over, if it doesn't work then there is a bug. Can you please confirm it is indeed not working. We have a [test](https://github.com/Azure/azure-cli/blob/dev/src/command_modules/azure-cli-appservice/azure/cli/command_modules/appservice/tests/latest/test_webapp_commands.py#L512-L514) to cover this scenario

cornem:
I have narrowed it down a bit further. The difference between the portal and `--configuration-source` is that the portal also clones app settings or configuration strings that are configured as 'slot settings'. The CLI tools apparently ignore slot settings. 

It would probably be helpful if `--configuration-source` worked exactly like the clone functionality in the portal.

panchagnula:
Let me look at this is details & update the issue.

panchagnula:
@cornem you are referring to having settings marked as SlotSettings correct? 
![image](https://user-images.githubusercontent.com/13839372/43231588-e0878602-9021-11e8-8855-af14fa1a75e9.png)
then yes, we should copy over the settings that are marked as Slot Settings if the app that has these values set is used during slot create.

cornem:
Yes, I'm referring to settings marked as slot setting.

panchagnula:
[...]

PR #19260, opened 2021-08-18
[App Service] `az webapp deployment slot create`: configuration source and cloning a slot settings

SharonHart:
When creating a deployment slot and using the `--configuration-source` argument, it is not clear that settings marked as 'slot settings' are not propagated to the newly created deployment slot.
Some issues were opened following the lack of documentation around it: https://github.com/Azure/azure-cli/issues/6638, https://github.com/Azure/azure-cli/issues/17531
 
### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/_params.py
@@ -376,7 +376,11 @@ def load_arguments(self, _):
         c.argument('preserve_vnet', help="preserve Virtual Network to the slot during swap, default to 'true'",
                    arg_type=get_three_state_flag(return_label=True))
     with self.argument_context('webapp deployment slot create') as c:
-        c.argument('configuration_source',
+        c.argument('
+                   
+                   
+                   
+                   ',
                    help="source slot to clone configurations from. Use web app's name to refer to the production slot")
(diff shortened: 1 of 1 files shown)


yonzhan:
App Service

panchagnula:
@SharonHart thanks for PR - however, this is not the right place for the fix - the official documentation of slots should be clarified as well - but if the ask is for CLI to clarify as well, my team will work on the appropriate place to clarify this. Thanks. Closing this for now.

SharonHart:
@panchagnula Thanks for the comment.
Do you mean the official documentation for azure cli, or the general app service documentation for this feature? 
If you meant the later, know that doing the same operation through the UI **does** copy a deployment slot setting.
So either both the CLI and the portal needs to be aligned, or each documented with their specific behaviour. 
This is the reason I opened the PR - the current state is very confusing.

panchagnula:
@SharonHart CLI uses the API to copy the right settings unlike portal client that always copy the full list. If this is the desired behavior CLI can do the same explicit copying I left a note for the feature owner to answer this.
The documentation update won't help in this case (in my opinion) since like you mentioned the original GitHub issue is to support the same behavior as portal - & before doing this I'd like to know from the feature owner if this should be the right behavior. Thanks!

PR #19295, opened 2021-08-20
[ACS] BREAKING CHANGE: `az aks nodepool update`: Change rejecting the ability to use max-surge with node-image-only

charliedmcb:
**Description**

A customer noticed that we do not respect the max-surge parameter when node-image-only is specified. Not an easy case to fix this, so we are instead rejecting this case, and explaining how they can use max surge with it if desired.

**Testing Guide**

**History Notes**

---


src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -4079,6 +4079,13 @@ def aks_agentpool_upgrade(cmd, client, resource_group_name, cluster_name,
             'node version please use the "--node-image-only" option only.'
         )
 
+    if max_surge and node_image_only:
```
paulgmiller, line 4082: comment that upgrade can't accept agentpool put fields like max surge.

paulgmiller (approved):

src/azure-cli/azure/cli/command_modules/acs/custom.py
charliedmcb: Added comment:
`Note: we exclude this option because node image upgrade can't accept nodepools put fields like max surge`

yonzhan:
ACS

src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -4079,6 +4079,14 @@ def aks_agentpool_upgrade(cmd, client, resource_group_name, cluster_name,
             'node version please use the "--node-image-only" option only.'
         )
 
+    # Note: we exclude this option because node image upgrade can't accept nodepool put fields like max surge
+    if max_surge and node_image_only:
+        raise CLIError(
```
FumingZhang, line 4084: I think replacing `CLIError` with `MutuallyExclusiveArgumentError` can more clearly point out the cause of the error.

charliedmcb: Updated. There is another error directly above that is listed as CLIError. However, is a mutually exclusive argument check between kubernetes_version and node_image_only.

Do you want that updated as well, so better to not change the existing error in this PR?

charliedmcb: Also, update the help message. Unsure if that is desired or not, but felt that this mutually exclusive issue was less intuitive

FumingZhang: `CLIError` is an error type that has been deprecated. Although simple and straightforward, it is too general, and it is not possible to distinguish the cause of the error from the user, the client or the server. You can refer to the recommended error types defined in this file [azclierror.py](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/azure/cli/core/azclierror.py).

As for other places where `CLIError` is used, if it is not modified in this PR, just ignore it. I will gradually update these inappropriate error types.

More accurate error types can also help the telemetry and our azcli aks runner/check-in pipleine to determine whether the error encountered is a real issue :smiley:

FumingZhang (approved):

zhoxing-ms:
@charliedmcb Could you please fix the CI issues?

zhoxing-ms (approved):

krsche:
Instead of fixing the bug, the functionality is removed __with a breaking change__? Absolutely not happy...

By the way, why is this considered ACS instead of AKS?

charliedmcb:
@krsche 
Sorry for the inconvenience. As you found it would require support within the API. This would be a significant change. We would still like to support this. However, as it was currently unfunctional (would drop the max-surge value), with no near-term API fix we made the decision that in this case it was worth blocking the combination of the two values so customer would be made aware this scenario was not working as intended. With the current API implementation, it should not have been exposed as a valid combination to begin with. 

> By the way, why is this considered ACS instead of AKS? @charliedmcb @yonzhan

This is legacy where ACS was previously used in place of AKS.

Issue #19342, opened 2021-08-25
--library-requirements issue (az synapse spark pool create)


genegc:
The --library-requirements option does not appear to be working.

**Command Name**
`az synapse spark pool create`

**Errors:**
```
The command failed with an unexpected error. Here is the traceback:
[Errno 2] No such file or directory: 'name: prod-env\ndependencies:\n  - opencensus-ext-azure\n  - Faker'
Traceback (most recent call last):
  File "/opt/az/lib/python3.6/site-packages/knack/cli.py", line 231, in invoke
    cmd_result = self.invocation.execute(args)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 657, in execute
    raise ex
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 720, in _run_jobs_serially
    results.append(self._run_job(expanded_arg, cmd_copy))
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 691, in _run_job
    result = cmd_copy(params)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 328, in __call__
    return self.handler(*args, **kwargs)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/command_operation.py", line 121, in handler
    return op(**command_args)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/command_modules/synapse/operations/sparkpool.py", line 40, in create_spark_pool
    library_requirements_content = read_file_content(library_requirements)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/util.py", line 484, in read_file_content
    with codecs_open(file_path, encoding=encoding) as f:
  File "/usr/bin/../../opt/az/lib/python3.6/codecs.py", line 897, in open
    file = builtins.open(filename, mode, buffering)
FileNotFoundError: [Errno 2] No such file or directory: 'name: prod-env\ndependencies:\n  - opencensus-ext-azure\n  - Faker'

```

## To Reproduce:
I tried the following approaches (one using requirements.yaml file, and one using a requirements.txt file)
az synapse spark pool create --name sparkx1 --workspace-name syn-oea-cisdggv04n --resource-group rg-oea-cisdggv04n --spark-version 3.0 --node-count 3 --node-size Small --min-node-count 3 --max-node-count 10 --library-requirements @./requirements.yml --verbose

[...]

PR #19358, opened 2021-08-26
[Synapse]spark pool remove library requirement

zesluo:
**Description**

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/synapse/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/synapse/_params.py
@@ -115,10 +115,6 @@ def load_arguments(self, _):
                    help='The flag of enabling auto pause.')
         c.argument('delay', arg_group='AutoPause', help='The delay time whose unit is minute.')
 
-        # Environment Configuration
-        c.argument('library_requirements', arg_group='Environment Configuration',
-                   help='The library requirements file.')
-
         # Default Folder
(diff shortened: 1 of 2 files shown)


src/azure-cli/azure/cli/command_modules/synapse/_params.py
```diff
@@ -115,10 +115,6 @@ def load_arguments(self, _):
                    help='The flag of enabling auto pause.')
         c.argument('delay', arg_group='AutoPause', help='The delay time whose unit is minute.')
 
-        # Environment Configuration
```
jsntcy, line 118: Usually before we remove an argument, we add deprecate information as a warning for one or two sprints to notify users that the argument will be deprecated soon instead of a sudden removal which may break users' scripts.

zesluo: Hello @jsntcy , We've already tried it, this argument cannot work at all. https://github.com/Azure/azure-cli/issues/19342

jsntcy: @zesluo, is this a behavior change from service side?

jsntcy: It's a good practice to notify users for the breaking changes beforehand; otherwise when we release this change to users, the scripts with this logic will be broken.

jsntcy (approved):

PR #19381, opened 2021-08-30
{Core} Fix: `open_page_in_browser` corrupts WSL shell

jiasli:
## Symptom

In WSL, when running `az feedback`, CLI calls `open_page_in_browser` to launch a web browser. The shell gets corrupted:

image

Pressing <kbd>ENTER</kbd> doesn't make a line break, and any typed text is not echoed back.

## Cause

If `python` exits before the subprocess `powershell.exe` exits, this can corrupt the WSL shell (https://github.com/microsoft/WSL/issues/7367). 

## Change

Call `wait` on `subprocess.Popen` so that CLI blocks until `powershell.exe` exits.

## Additional information

This issue doesn't happen in `az login` even though `az login` also calls `open_page_in_browser`. This is because `az login` waits for the auth code after launching `powershell.exe`, giving `powershell.exe` enough time to exit.

### DIFF
--- a/src/azure-cli-core/azure/cli/core/util.py
+++ b/src/azure-cli-core/azure/cli/core/util.py
@@ -657,7 +657,8 @@ def open_page_in_browser(url):
         try:
             # https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_powershell_exe
             # Ampersand (&) should be quoted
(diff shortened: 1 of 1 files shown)


yonzhan:
Core

evelyn-ys (approved):

calvinhzy (approved):

PR #19397, opened 2021-08-31
{AKS} Refactor a clean-up part of the create sub-command in acs module

FumingZhang:
**Description**

* Adjust the input of the default test case in the unit tests to the default value corresponding to the parameter in function aks_create.
* Add docstring for all getter functions.
* Break down get_enable_managed_identity_service_principal_and_client_secret into get_service_principal_and_client_secret  and get_enable_managed_identity.
* Remove all parameter intermediates and introduce an additional "read_only" option to bypass dynamic completion and validation to avoid circulation function calls.

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acs/decorator.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/decorator.py
@@ -1429,26 +1429,20 @@ class AKSCreateContext:
         return enable_ahub
 
     # pylint: disable=unused-argument,too-many-statements
-    def get_enable_managed_identity_service_principal_and_client_secret(
+    def get_service_principal_and_client_secret(
         self, **kwargs
-    ) -> Tuple[bool, Union[str, None], Union[str, None]]:
-        """Dynamically obtain the values of enable_managed_identity, service_principal and client_secret according to
-        the context.
+    ) -> Tuple[Union[str, None], Union[str, None]]:
+        """Dynamically obtain the values of service_principal and client_secret according to the context.
 
-        Note: enable_managed_identity will not be decorated into the `mc` object, and changes to the original value
-        will be stored in the intermediates dictionary.
-
-        When service_principal and client_secret are not given and enable_managed_identity is True, dynamic completion
-        will not be triggered. For other cases, dynamic completion will be triggered.
-        When service_principal and client_secret are given, the value of enable_managed_identity will be changed to
-        False.
+        When service_principal and client_secret are not assigned and enable_managed_identity is True, dynamic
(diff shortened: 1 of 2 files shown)


src/azure-cli/azure/cli/command_modules/acs/decorator.py
```diff
@@ -293,14 +330,14 @@ def get_dns_name_prefix(self, enable_validation: bool = False, **kwargs):
...
-            # clean up intermediate if `mc` has been decorated
-            self.remove_intermediate(parameter_name)
             read_from_mc = True
-        elif intermediate is not None:
-            dns_name_prefix = intermediate
         else:
             dns_name_prefix = raw_value
 
+        # skip dynamic completion & validation if option read_only is specified
+        if kwargs.get("read_only"):
```
andyliuliming, line 338: this is new thing?

src/azure-cli/azure/cli/command_modules/acs/decorator.py
```diff
@@ -426,41 +478,47 @@ def get_vm_set_type(self, enable_validation: bool = False, **kwargs):
...
-        # check whether the parameter meet the conditions of dynamic completion
-        if not vm_set_type:
-            dynamic_completion = True
+        # skip dynamic completion & validation if option read_only is specified
+        if kwargs.get("read_only"):
+            return vm_set_type
+
+        # the value verified by the validator may have case problems, and the
+        # "_set_vm_set_type" function will adjust it
+        dynamic_completion = True
```
andyliuliming, line 491: always true?

andyliuliming (approved):

src/azure-cli/azure/cli/command_modules/acs/decorator.py
FumingZhang: Yes, after removing the intermediates of all parameters, in order to reduce the introduction of additional content, we minimally add this option to a few getter functions to avoid state loss and loop calls.

src/azure-cli/azure/cli/command_modules/acs/decorator.py
FumingZhang: The server side is sensitive to the case of the value, but the user may enter a value with the wrong case, which can be corrected here. The reason why `dynamic_completion ` is set to True by default is to be consistent with the behavior of the code in cli.

In addition, if the value is obtained from the mc object, no further modifications will be made (the next line of code will change `dynamic_completion`  to False in this case).

yonzhan:
AKS

zhoxing-ms (approved):

PR #19445, opened 2021-09-03
[ARM] `az deployment`: Remove the log of printing request body from custom policy

zhoxing-ms:
**Description**

Recently, I received an ICM which feed back that the operation of ARM deployment will print the request body after passing in parameter `--verbose`, which may lead to the disclosure of the value for confidential parameters.
Therefore, I submit this PR to fix the issue that it will print the request body in verbose log.
Before
before
After
after

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/resource/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/resource/custom.py
@@ -403,7 +403,6 @@ class JsonCTemplatePolicy(SansIOHTTPPolicy):
 
     def on_request(self, request):
         http_request = request.http_request
-        logger.info(http_request.data)
         if (getattr(http_request, 'data', {}) or {}).get('properties', {}).get('template'):
             template = http_request.data["properties"]["template"]
             if not isinstance(template, JsonCTemplate):


jiasli (approved):

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -403,7 +403,7 @@ class JsonCTemplatePolicy(SansIOHTTPPolicy):
 
     def on_request(self, request):
         http_request = request.http_request
-        logger.info(http_request.data)
+        logger.debug(http_request.data)
```
jiasli, line 406: Won't debug also "disclosure of the value for confidential parameters"?

jiasli (approved):

src/azure-cli/azure/cli/command_modules/resource/custom.py
zhoxing-ms: It is the same as the request body printed in log `cli.azure.cli.core.sdk.policies`, so it's redundant, I deleted it

yonzhan:
ARM

jsntcy (approved):

yonzhan (approved):

Issue #19365, opened 2021-08-26
Progress display uses stderr for output, breaks automation scenarios


wasker:
**Describe the bug**

Any part of Azure CLI that uses "progress controller" to display progress of long running operations outputs to stderr even if there were no errors. This leads to undesired breaking in automation scenarios.

For example, if az CLI is invoked from PowerShell script with $ErrorActionPreference set to "Stop", then even with stderr -> stdout redirection, PowerShell will stop execution of the script due to an "error" reported via progress controller.

See AKS commands for example. There're many places where AKS module would use "progress controller" [with all default options](https://github.com/Azure/azure-cli/blob/da3ff654f80f749525ee5c321c37dc041f1061d2/src/azure-cli/azure/cli/command_modules/acs/custom.py#L489), which results in [stderr used for showing progress](https://github.com/Azure/azure-cli/blob/da3ff654f80f749525ee5c321c37dc041f1061d2/src/azure-cli-core/azure/cli/core/commands/progress.py#L168).

**To Reproduce**

Use `az aks create` command to observe that when AKS updates role assignment for some identity (it does it for multiple identities), stderr is being written to.

Note that this is not limited to AKS. For example storage blob uploads use same mechanism. Storage module ended up adding "--no-progress" flag to disable it.

**Expected behavior**

If there's no error, do not write to stderr.

**Environment summary**

Latest az CLI with all modules up-to-date.

**Additional context**

1. Please change the default of the progress view to use stdout.
2. Please add a GLOBAL flag to disable progress display across the board. Azure CLI is used in automation scenarios, where progress display is both unnecessary and messes up redirected output.

PR #19464, opened 2021-09-07
[Core] Support disable progress bar

zhoxing-ms:
Related issue: #19365

**Description**

There is a issue #19365 about the output of progress bar: when the progress bar uses `stderr` as the output stream, the script execution in the automation scenario may be interrupted, such as:
> For example, if az CLI is invoked from PowerShell script with $ErrorActionPreference set to "Stop", then even with stderr -> stdout redirection, PowerShell will stop execution of the script due to an "error" reported via progress controller.

However, this is by design that progress bar uses `stderr` for output, because users may parse the output data from `stdout`, and our design pattern is that only writes the output data to `stdout`.

So we can have two solutions for this problem:
1. Add a global config by `az config` to allow users to disable progress bar globally in the automation scenario.
2. It is supported to use the parameter `--only-show-errors` to disable progress bar.

**Testing Guide**

**History Notes**

[Core] Add a global config by `az config` to allow users to disable progress bar globally in the automation scenario.
[Core] It is supported to use the parameter `--only-show-errors` to disable progress bar.

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/commands/__init__.py
+++ b/src/azure-cli-core/azure/cli/core/commands/__init__.py
@@ -875,7 +875,11 @@ class LongRunningOperation:  # pylint: disable=too-few-public-methods
         self.poller_done_interval_ms = poller_done_interval_ms
         self.deploy_dict = {}
         self.last_progress_report = datetime.datetime.now()
-        self.progress_bar = progress_bar if progress_bar is not None else IndeterminateProgressBar(cli_ctx)
+
+        self.progress_bar = None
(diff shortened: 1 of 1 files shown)


yonzhan:
Core

jiasli (approved):

wasker:
@yonzhan @jiasli @zhoxing-ms FYI: it seems that core.disable_progress_bar config option is not documented.

zhoxing-ms:
@dbradish-microsoft Could you please help add the information of `core.disable_progress_bar ` to the document? [doc link](https://docs.microsoft.com/en-us/cli/azure/azure-cli-configuration)

PR #19516, opened 2021-09-10
[RDBMS] BREAKING CHANGE: az postgres flexible-server migration: Changed --properties @{filepath} to --properties {filepath}

cpj2195:
**Description**
1. In Create migration command, changed the way to pass properties file from --properties @{filepath} -> --properties filepath
2. Added check in create migration to check if file exists in the given location
3. Added another Migration command to check migration name availability. Test also added
4. Added an Update command parameter to reschedule the migration to start right now.

**Testing Guide**

**History Notes**

[RDBMS] BREAKING CHANGE: `az postgres flexible-server migration`: Change `--properties @{filepath}` to `--properties {filepath}`
[RDBMS] `az postgres flexible-server migration create`: User can pass in filename with double quotes or no quotes and same for absolute paths.
[RDBMS] `az postgres flexible-server migration check-name-availability`: Add a command to check if a migration name is available.
[RDBMS] `az postgres flexible-server migration update`:  Add `--start-data-migration` to reschedule the migration to start right now.

---


yonzhan:
RDBMS

src/azure-cli/azure/cli/command_modules/rdbms/flexible_server_custom_common.py
```diff
@@ -83,12 +85,15 @@ def firewall_rule_create_func(client, resource_group_name, server_name, firewall
 def migration_create_func(cmd, client, resource_group_name, server_name, properties, migration_name=None):
 
     subscription_id = get_subscription_id(cmd.cli_ctx)
-
+    properties_filepath = os.path.join(os.path.abspath(os.getcwd()), properties)
```
evelyn-ys, line 88: What if `properties` is absolute path?

Actually the proper way to define a file_path param in CLI is use `type=file_type, completer=FilesCompleter()`
See example in storage module:
https://github.com/Azure/azure-cli/blob/228fd1c8e749a3fbbf32593e47bdc8ca91ddb9b2/src/azure-cli/azure/cli/command_modules/storage/_params.py#L486-L489

src/azure-cli/azure/cli/command_modules/rdbms/flexible_server_custom_common.py
```diff
@@ -83,12 +85,15 @@ def firewall_rule_create_func(client, resource_group_name, server_name, firewall
...
+    if not os.path.exists(properties_filepath):
+        raise FileOperationError("Properties file does not exist in the given location")
+    with open(properties_filepath, "r") as f:
+        json_data = f.read()
```
evelyn-ys, line 92: Better to utilize existing function `get_file_json` and `shell_safe_json_parse` in `azure.cli.core.util`.
See the same example in storage module:
https://github.com/Azure/azure-cli/blob/228fd1c8e749a3fbbf32593e47bdc8ca91ddb9b2/src/azure-cli/azure/cli/command_modules/storage/operations/account.py#L587-L591

src/azure-cli/azure/cli/command_modules/rdbms/flexible_server_custom_common.py
```diff
@@ -141,6 +146,12 @@ def migration_update_func(cmd, client, resource_group_name, server_name, migrati
...
         properties = "{\"properties\": {\"triggerCutover\": \"true\"} }"
 
+    if start_data_migration is True:
+        if operationSpecified is True:
+            raise MutuallyExclusiveArgumentError("Incorrect Usage: Can only specify one update operation.")
+        operationSpecified = True
+        properties = "{\"properties\": {\"startDataMigration\": \"true\"} }"
+
     if operationSpecified is False:
         raise RequiredArgumentMissingError("Incorrect Usage: Atleast one update operation needs to be specified.")
```
evelyn-ys, line 156:
```suggestion
        raise RequiredArgumentMissingError("Incorrect Usage: At least one update operation needs to be specified.")
```

evelyn-ys:
Can you update **History Notes** part in PR description since there're several changes in this PR and PR title may not able to cover all.

src/azure-cli/azure/cli/command_modules/rdbms/flexible_server_custom_common.py
cpj2195: os.path.join() will join only if properties is a filename and not otherwise. I tested this with properties file being a filename and absolute path and both worked .

src/azure-cli/azure/cli/command_modules/rdbms/flexible_server_custom_common.py
cpj2195: I'll see if the above 2 functions can be used . We want to intentionally remove the use of @ symbol since that is a point of potential failure for our customers. Passing in simple strings is much easier.

src/azure-cli/azure/cli/command_modules/rdbms/flexible_server_custom_common.py
cpj2195: Here, we can specify only 1 update operation . with `At least` means , we can specify more than equal to one. Am i missing something else here?

evelyn-ys (approved):

PR #19531, opened 2021-09-10
[RDBMS] Update list-skus, create command location setting and replica command

DaeunYim:
**Description**

- Change list-skus logic for single availability zone checking
- Added replica validation
- Fixed bug of create command location setting
- private access server restore command network resources to be same as source server's

**Testing Guide**

**History Notes**

---


yonzhan:
RDBMS

src/azure-cli/azure/cli/command_modules/rdbms/_params.py
```diff
@@ -365,11 +365,11 @@ def _flexible_server_params(command_group):
         )
 
         mysql_high_availability_arg_type = CLIArgumentType(
-            arg_type=get_enum_type(['ZoneRedundant', 'SameZone', 'Disabled']),
+            arg_type=get_enum_type(['ZoneRedundant', 'SameZone', 'Disabled', 'Enabled']),
             options_list=['--high-availability'],
             help='Enable (ZoneRedundant or SameZone) or disable high availability feature. '
                  'Default value is Disabled. High availability can only be set during flexible server create time. '
-                 'Allowed values: ZoneRedundant, SameZone, Disabled'
+                 'Allowed values: ZoneRedundant, SameZone, Disabled, Enabled'
```
evelyn-ys, line 372: CLI core will generate `Allowed values` automatically for enum type arguments, you can omit this line.

src/azure-cli/azure/cli/command_modules/rdbms/flexible_server_custom_mysql.py
```diff
@@ -50,7 +50,10 @@ def flexible_server_create(cmd, client,
         cf_availability=cf_mysql_check_resource_availability, cf_private_dns_zone_suffix=cf_mysql_flexible_private_dns_zone_suffix_operations, logging_name='MySQL', command_group='mysql', server_client=client,
         location=location)
 
+    # Process parameters
     server_name = server_name.lower()
+    if high_availability and high_availability.lower() == 'enabled':
+        high_availability = 'ZoneRedundant'
```
evelyn-ys, line 56: Why do you need to add `Enabled` if it's only an alternative of `ZoneRedundant`?

src/azure-cli/azure/cli/command_modules/rdbms/_params.py
DaeunYim: Aha. thanks!

src/azure-cli/azure/cli/command_modules/rdbms/flexible_server_custom_mysql.py
DaeunYim: PMs asked me to make it downward compatible.. so adding back the 'enabled' parameter. What do you think is the best way to deprecate acceptable parameter?

evelyn-ys: Got it. CLI has deprecating mechanism for some argument but no for argument acceptable enums. I guess you can add a warning here to let customer know that `Enabled` is going to be deprecated, use `ZoneRedundant` instead.

DaeunYim: I see. Let me add that! Thank you :)

evelyn-ys (approved):

PR #5328, opened 2018-01-17
Ensure skip cert check when sending out raw requests if related env var is configured

yugangw-msft:
Fix #5099. Fix #5258 
@derekbekoe, I can also update related [extension code](https://github.com/Azure/azure-cli/blob/dev/src/command_modules/azure-cli-extension/azure/cli/command_modules/extension/_index.py#L26). So let me know 
@djyou,  I can update [acr code](https://github.com/Azure/azure-cli/blob/dev/src/command_modules/azure-cli-acr/azure/cli/command_modules/acr/repository.py#L57). So let me know

Moving forward, during code review, for any HTTTPS raw requests outside of Python SDK, we just ensure the new code checks the global flag

- [will do later] The PR has modified HISTORY.rst describing any customer-facing, functional changes. Note that this does not include changes only to help content. (see [Modifying change log](https://github.com/Azure/azure-cli/tree/master/doc/authoring_command_modules#modify-change-log)).

### Command Guidelines

- [na] Each command and parameter has a meaningful description.
- [na] Each new command has a test.

(see [Authoring Command Modules](https://github.com/Azure/azure-cli/tree/master/doc/authoring_command_modules))

PR #19554, opened 2021-09-14
{Misc.} Remove usages of `six`

jiasli:
**Description**

`six` was previously dropped from `azure-cli-core` (https://github.com/Azure/azure-cli/pull/17366). It should be dropped from `azure-cli` as well.

However, as there are countless extensions still using `six`, we still need to install it. This PR removes all usages of `six.

### DIFF
--- a/src/azure-cli-core/setup.py
+++ b/src/azure-cli-core/setup.py
@@ -58,7 +58,6 @@ DEPENDENCIES = [
     'PyJWT>=2.1.0',
     'pyopenssl>=17.1.0',  # https://github.com/pyca/pyopenssl/pull/612
     'requests[socks]~=2.25.1',
-    'six~=1.12',
     'urllib3[secure]>=1.26.5',
 ]
 
--- a/src/azure-cli/azure/cli/command_modules/vm/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/custom.py
@@ -22,7 +22,7 @@ except ImportError:
     from urlparse import urlparse  # pylint: disable=import-error
 
 # the urlopen is imported for automation purpose
-from six.moves.urllib.request import urlopen  # noqa, pylint: disable=import-error,unused-import,ungrouped-imports
+from urllib.request import urlopen
 
 from knack.log import get_logger
 from knack.util import CLIError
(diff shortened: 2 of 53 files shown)


src/azure-cli-core/setup.py
```diff
@@ -58,7 +58,6 @@
     'PyJWT>=2.1.0',
     'pyopenssl>=17.1.0',  # https://github.com/pyca/pyopenssl/pull/612
     'requests[socks]~=2.25.1',
-    'six~=1.12',
```
jiasli, line 61: Remove the dependency from `azure-cli-core`.

src/azure-cli/setup.py
```diff
@@ -144,6 +143,7 @@
     'pytz==2019.1',
     'scp~=0.13.2',
     'semver==2.13.0',
+    'six>=1.10.0',  # six is still used by countless extensions
```
jiasli, line 146: Add `six` to `azure-cli` as it can be used by extensions.

calvinhzy (approved):

yonzhan:
Misc

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -16,13 +16,9 @@
...
 
-try:
-    from urllib.parse import urlparse
-except ImportError:
-    from urlparse import urlparse  # pylint: disable=import-error
-
+from urllib.parse import urlparse
 # the urlopen is imported for automation purpose
-from six.moves.urllib.request import urlopen  # noqa, pylint: disable=import-error,unused-import,ungrouped-imports
+from urllib.request import urlopen  # noqa, pylint: disable=import-error,unused-import,ungrouped-imports
```
jiasli, line 21: `urlopen` is never used by `custom.py` but is patched by test

https://github.com/Azure/azure-cli/blob/87004321ec280e092fc41febf4ada3b58f01f388/src/azure-cli/azure/cli/command_modules/vm/tests/latest/test_vm_image.py#L30

What is exactly "automation purpose"?

zhoxing-ms: This is because in the previous PR https://github.com/Azure/azure-cli/pull/5328, the reference of `urlopen` in the method `load_images_from_aliases_doc()` was modified to use `requests`, but the mock path in `test_read_images_from_alias_doc()` was not synchronously modified.
Screenshot 2021-09-14 172259

Therefore, the import of `from six.moves.urllib.request import urlopen` can be deleted, and we just need to change the mock path to `@mock.patch(azure.cli.command_modules.vm.custom.requests, autospec=True)`

jiasli: We can do this in a separate PR if time permits.

PR #19570, opened 2021-09-15
[Backup] `az backup container register`: Fix refresh container bug

akshayneema:
**Description**

- Fixed the refresh protectable container API call that happens when the protectable container is not found on the first try in az backup container register command. CRI link - [Incident 259315847 : [ARR] [SAP],[DISCOVERY], | [KRKEMA] | [Unable to discovery SAP Hana DB from Azure VM "xub29041431h0" using Azure CLI via ADO pipeline. Getting error message: Container unavailable or already registered.]](https://portal.microsofticm.com/imp/v3/incidents/details/259315847/home)

- Added 'MAB' in the allowed values for backup management type in az backup job/container list command

- Increased the sleep delay while tracking operations to prevent frequent calls.

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/backup/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/backup/_params.py
@@ -140,7 +140,7 @@ def load_arguments(self, _):
 
     with self.argument_context('backup container list') as c:
         c.argument('vault_name', vault_name_type, id_part=None)
-        c.argument('backup_management_type', backup_management_type)
+        c.argument('backup_management_type', arg_type=get_enum_type(allowed_backup_management_types + ["MAB"]), help=backup_management_type_help)
         c.argument('use_secondary_region', action='store_true', help='Use this flag to list containers in secondary region.')
 
     with self.argument_context('backup container unregister') as c:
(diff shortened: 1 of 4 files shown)


yonzhan:
backup

src/azure-cli/azure/cli/command_modules/backup/_params.py
```diff
@@ -140,7 +140,7 @@ def load_arguments(self, _):
 
     with self.argument_context('backup container list') as c:
         c.argument('vault_name', vault_name_type, id_part=None)
-        c.argument('backup_management_type', backup_management_type)
+        c.argument('backup_management_type', arg_type=get_enum_type(allowed_backup_management_types + ["MAB"]), help=backup_management_type_help)
```
zhoxing-ms, line 143: Why not add `MAB` to `allowed_backup_management_types`?

zhoxing-ms:
Could you add some tests for this change?

zhoxing-ms:
Could you add the issue link in the description of this PR?

src/azure-cli/azure/cli/command_modules/backup/_params.py
akshayneema: I'll check with the PM on this and get back.

akshayneema:
> Could you add the issue link in the description of this PR?

Added CRI link in description.

akshayneema:
> Could you add the issue link in the description of this PR?



> Could you add some tests for this change?

It's an internal call that we are making only for newly created SQL/HANA VMs. For testing, we use an already present setup for SQL and HANA cases so adding testing specifically for this won't be possible with the current testing setup we use.

akshayneema:
> > Could you add some tests for this change?
> 
> It's an internal call that we are making only for newly created SQL/HANA VMs. For testing, we use an already present setup for SQL and HANA cases so adding tests specifically for this change won't be possible with the current testing setup we use.

In the future, if we move to an autogenerated setup it will get tested

src/azure-cli/azure/cli/command_modules/backup/_params.py
akshayneema: We only support selected commands for MAB(Azure Backup Agent) backup management type. So, we've only added it in the  arg_type enum wherever necessary.

zhoxing-ms (approved):

Issue #19591, opened 2021-09-17
Install Azure CLI on Alpine Linux


jiasli:
Users frequently face issues while installing Azure CLI on Alpine Linux docker containers (#7437, #8863, #9167, #4352) , this issue serves as an official installation guide. We will consider moving it to the [How to install the Azure CLI official document](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli) if it is proven to be useful.

## Azure CLI official docker image

The [official Azure CLI docker image](https://docs.microsoft.com/en-us/cli/azure/run-azure-cli-docker) is built upon Alpine Linux. We recommend using it whenever possible. You may also take the [Dockerfile](https://github.com/Azure/azure-cli/blob/dev/Dockerfile) as a reference. 

#### `python:alpine`

The [official Python docker image](https://hub.docker.com/_/python) is preferred as the base image.

```console
$ docker run -it --rm python:alpine sh
```

#### `alpine`

If you would like to use the original [official Alpine docker image](https://hub.docker.com/_/alpine), you need to install `pip` and `python`:

```console
$ docker run -it --rm alpine
# apk add py3-pip
```

### Install dependencies

As Alpine Linux is not `manylinux2010` (or greater) compatible, using `pip` to install libraries like `cryptography` will fail without underlying dependencies. Install dependencies with:

```console
# apk add gcc musl-dev python3-dev libffi-dev openssl-dev cargo make
```

See https://cryptography.io/en/latest/installation/#building-cryptography-on-linux

### Install Azure CLI

Simply use `pip` to install Azure CLI:

```console
# pip install azure-cli
```

## Use `Dockerfile`

Here are the `Dockerfile`s you may use as a starting point:

#### `python:alpine` as base image

```dockerfile
FROM python:alpine
RUN apk add gcc musl-dev python3-dev libffi-dev openssl-dev cargo make
RUN pip install azure-cli
CMD sh
```

#### `alpine` as base image

```dockerfile
FROM alpine
RUN apk add py3-pip
RUN apk add gcc musl-dev python3-dev libffi-dev openssl-dev cargo make
RUN pip install azure-cli
CMD sh
```

### Build the image

```console
$ docker build --tag azure-cli-alpine-test .
```

### Run `az` commands

```console
$ docker run -it --rm azure-cli-alpine-test az --version
```

benjamin-goldman:
so we dont think this will work reliably.

jiasli:
> so we dont think this will work reliably.

[...]

PR #19639, opened 2021-09-22
{Packaging} Loosen `cryptography` dependency

### LINKED ISSUE #19635: Unpin cryptography package in azure-cli-core
Would it be possible to unpin the `cryptography` dependency by changing 
`cryptography>=3.2,<3.4`
to
`cryptography~=3.2`
in `/src/azure-cli-core/setup.py`?

This change won't affect great majority of Azure CLI users (who are using prebuilt packages with `cryptography==3.3.2`).

The users who are still using an older OS which doesn't have rust support (e.g. Alpine) can still apply a simple workaround (installing cryptography==3.3.2 before installing Azure-CLI).
If it is important that [install.py](https://azurecliprod.blob.core.windows.net/install.py) works on those platforms without any additional fixes, this script can be modified to install `cryptography==3.3.2` before installing `azure-cli`.

jiasli:
## Description

- Fix #19635

Reasons why we can loose the `cryptography` dependency now:

### For **upper bound**

https://github.com/Azure/azure-cli/pull/15687#discussion_r578898876 set an upper bound for `cryptography` due to https://github.com/pyca/cryptography/issues/5771.

By following https://cryptography.io/en/latest/installation/#alpine, I am now able to install `cryptography` 3.4.8 on Alpine Linux (https://github.com/Azure/azure-cli/issues/19591). There is no need to set upper bound now.

### For **lower bound**

Community packagers ask us not to bump the minimum dependency in `setup.py` as this causes trouble for platforms that doesn't support newer versions of `cryptography` (https://github.com/Azure/azure-cli/pull/15687#issuecomment-878819248). By loosing the dependency on `cryptography`, users or packagers themselves are now responsible for security vulnerabilities in older versions of `cryptography`.

As we still pin the version in `requirements.*.txt`, packages distributed by us still contain `cryptography` which is security-complaint.

### DIFF
--- a/src/azure-cli-core/setup.py
+++ b/src/azure-cli-core/setup.py
@@ -47,7 +47,7 @@ DEPENDENCIES = [
     'argcomplete~=1.8',
     'azure-cli-telemetry==1.0.6.*',
     'azure-mgmt-core>=1.2.0,<1.3.0',     # the preview version of azure-mgmt-core is 1.3.0b1, it cannot fit azure-core >=1.14.0
-    'cryptography~=3.0',
+    'cryptography',
     'humanfriendly>=4.7,<10.0',
     'jmespath',
     'knack~=0.8.2',
--- a/src/azure-cli-core/setup.py
+++ b/src/azure-cli-core/setup.py
@@ -47,7 +47,7 @@ DEPENDENCIES = [
     'argcomplete~=1.8',
     'azure-cli-telemetry==1.0.6.*',
     'azure-mgmt-core>=1.2.0,<1.3.0',     # the preview version of azure-mgmt-core is 1.3.0b1, it cannot fit azure-core >=1.14.0
-    'cryptography>=3.2,<3.4',
+    'cryptography~=3.0',
     'humanfriendly>=4.7,<10.0',
     'jmespath',
     'knack~=0.8.2',


src/azure-cli-core/setup.py
```diff
@@ -47,7 +47,7 @@
     'argcomplete~=1.8',
     'azure-cli-telemetry==1.0.6.*',
     'azure-mgmt-core>=1.2.0,<1.3.0',     # the preview version of azure-mgmt-core is 1.3.0b1, it cannot fit azure-core >=1.14.0
-    'cryptography>=3.2,<3.4',
+    'cryptography~=3.0',
```
jiasli, line 50: Using [Compatible release](https://www.python.org/dev/peps/pep-0440/#compatible-release) so that `cryptograph` is `<4`.

yonzhan:
Packaging

jiasli:
@glaubitz, I have totally removed the lower bound. Will this suit your need?

evelyn-ys (approved):

calvinhzy (approved):

glaubitz:
Is version 3.0 actually the absolute minimum that you need now? That is, is there functionality that you need that is not part of version 2.8?

jiasli:
We don't require >3.0 now. Any version works.

glaubitz:
Then either remove the version requirement or use at least 2.8 which is what SLE-12 and SLE-15 are shipping at the moment.

jiasli:
> Then either remove the version requirement

It is already removed.

PR #19680, opened 2021-09-24
[AKS] `az aks create/update`: Add new parameter `--disable-local-accounts` to support disable local accounts

norshtein:
**Description**

This PR adds ability to disable/enable local accounts for an AKS cluster. The feature is in preview for several months and we're going to GA. This PR simply copies changes in aks-preview extension: https://github.com/Azure/azure-cli-extensions/pull/3266
**Testing Guide**

**History Notes**

---


yonzhan:
AKS

src/azure-cli/azure/cli/command_modules/acs/tests/latest/test_aks_commands.py
```diff
@@ -5640,3 +5640,29 @@ def test_aks_create_with_SP_then_update_to_user_assigned_identity(self, resource
...
+        })
+
+        create_cmd = 'aks create --resource-group={resource_group} --name={name} --ssh-key-value={ssh_key_value} --enable-managed-identity --disable-local-accounts'
+        self.cmd(create_cmd, checks=[
+            self.check('provisioningState', 'Succeeded'),
+            self.check('disableLocalAccounts', True)
+        ])
+
+        # update to enable local accounts
+        self.cmd('aks update --resource-group={resource_group} --name={name} --enable-local-accounts', checks=[
```
andyliuliming, line 5661: better to add some "real validation" to check the local accounts is indeed disabled?

norshtein: I think that should be server behavior's validation, should be covered by something like AKS E2E test, and is out of CLI test scope?

FumingZhang:
Hi @norshtein, could you please add support for this option in `aks create` with the new decorator pattern? https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/acs/decorator.py

src/azure-cli/azure/cli/command_modules/acs/tests/latest/test_aks_commands.py
norshtein: When we create an cluster using option `--disable-local-accounts` by CLI, the CLI's responsibility is to create a cluster with property `disableLocalAccounts: true`, that is the CLI successfully accepts the option and send the request to server and server creates a cluster with desired property. That's all of our expectation for client. As for whether the server respects `disableLocalAccounts: true` when getting credential, it is not related to client. Say one day AKS server has a bug, the static credential can still be fetched even  `disableLocalAccounts: true`, that should not affect CLI's test. If we add server behavior check in CLI test, the server bug will block CLI and CLI has nothing to do with that.

norshtein:
sure, let me take a look.

norshtein:
@FumingZhang I've updated the code, could you please have a review?

src/azure-cli/azure/cli/command_modules/acs/decorator.py
```diff
@@ -3820,6 +3831,20 @@ def set_up_extended_location(self, mc: ManagedCluster) -> ManagedCluster:
             )
         return mc
 
+    def set_up_disable_local_accounts(self, mc: ManagedCluster) -> ManagedCluster:
```
FumingZhang, line 3834: If the newly added option `disable_local_accounts` is not related to other options, is not part of a complex profile and does not need to be set up separately to avoid confusion with other logic, I suppose you could omit this function (`set_up_disable_local_accounts`) and set it to the `ManagedCluster` object directly in `init_mc` (and don't forget to update the docstring/comments 😃).

src/azure-cli/azure/cli/command_modules/acs/tests/latest/test_decorator.py
```diff
@@ -4179,6 +4179,7 @@ def test_construct_default_mc_profile(self):
                 "assign_kubelet_identity": None,
                 "enable_ultra_ssd": False,
                 "edge_zone": None,
+                "disable_local_accounts": False,
```
FumingZhang, line 4182: I would appreciate it if you could also add some cases to test the getter function (I mean `get_disable_local_accounts`, through it quite straightforward for now) under class `AKSCreateContextTestCase`.

src/azure-cli/azure/cli/command_modules/acs/tests/latest/test_aks_commands.py
andyliuliming: yeah, then is there any E2E to cover this in "server" side?

FumingZhang (approved):

src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -2966,6 +2974,16 @@ def aks_update(cmd, client, resource_group_name, name,
             tier="Free"
         )
 
+    if disable_local_accounts and enable_local_accounts:
+        raise CLIError('Cannot specify --disable-local-accounts and --enable-local-accounts '
+                       'at the same time.')
```
zhoxing-ms, line 2979: Could you please use the specific error type `MutuallyExclusiveArgumentError` instead of `CLIError`?

src/azure-cli/azure/cli/command_modules/acs/_params.py
```diff
@@ -319,6 +320,8 @@ def load_arguments(self, _):
         c.argument('enable_managed_identity', action='store_true')
         c.argument('assign_identity', type=str,
                    validator=validate_assign_identity)
+        c.argument('disable_local_accounts', action='store_true')
+        c.argument('enable_local_accounts', action='store_true')
```
zhoxing-ms, line 324: Why do we need two parameters with opposite meanings? Isn't one parameter already able to support the current feature?

src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -2966,6 +2974,16 @@ def aks_update(cmd, client, resource_group_name, name,
...
+
+    if disable_local_accounts:
+        instance.disable_local_accounts = True
+
+    if enable_local_accounts:
+        instance.disable_local_accounts = False
```
zhoxing-ms, line 2985: Should only one parameter meet the requirements? Such as:

```
if disable_local_accounts != None:
     instance.disable_local_accounts = disable_local_accounts
```

src/azure-cli/azure/cli/command_modules/acs/custom.py
norshtein: Thanks for reviewing, updated.

src/azure-cli/azure/cli/command_modules/acs/custom.py
norshtein: Two points:
1. AKS's command has many other similar settings, such as enable/disable public fqdn, enable/disable ahub, etc. This feature uses existing pattern.
2. If I understand it correctly, for `store_true` option, it will be either true or false, and will never be None, default value will be false. In this case, customer must explicitly specify this parameter every time, otherwise their setting can be changed unexpectedly: current `disableLocalAccounts` is true, and if they don't explicitly specify `--disable-local-accounts` to false every time they update the AKS cluster, the setting will be updated to false.

zhoxing-ms: I mean, if you keep only one parameter `disable_local_accounts`, the code here can be simplified to `instance.disable_local_accounts = disable_local_accounts`.
At present, it is designed as two parameters with opposite meanings to do the same thing, which is a little redundant. What do you think of it? @FumingZhang

norshtein: What if customer does not provide this option? always use the default value `false`?

FumingZhang: I guess only one option is not enough for the user to set this property. For option `--disable-local-accounts`, its default value is False. In the **update** command, if the user does not set this option **explicitly** (that is, the value of `disable_local_accounts` is still False), I suppose it does not mean that the user wants to enable local accounts (that is, set `instance.disable_local_accounts = False`), but maintains the original setting.

zhoxing-ms: Could we not update this property when users does not pass in this parameter `--disable-local-accounts`?

norshtein: How to determine whether user passes this property? The `store_true` option will be set to default value `False` if user does not pass the property.

zhoxing-ms: > What if customer does not provide this option? always use the default value false?

@norshtein When using `store_true`, if users do not input this parameter, it's value is `None`. If this parameter is inputed, it's value becomes `True`.

norshtein: > > What if customer does not provide this option? always use the default value false?
> 
> @norshtein When using `store_true`, if users do not input this parameter, it's value is `None`. If this parameter is inputed, it's value becomes `True`.

Are you sure about it? The doc says the default value will be `False` https://docs.python.org/3/library/argparse.html#action

zhoxing-ms: @FumingZhang  I don't think `store_true` is suitable for this scenario and `get_three_state_flag()` may be a better choice.
In this way, the users can explicitly pass in `True` or `False` to enable or disable the property. When users do not pass in this parameter, the value will be `None`, then the property should not be updated

norshtein: Several reasons to keep current option unchanged:
1. The option has been in aks-preview for months, if we change it, when cx switch from aks-preview to Azure CLI, it will be a breaking change, we should avoid it.
2. Semantically, `--enable-xxx` and `disable-xxx` is more clear to cx. `disable-xxx=True` and `disable-xxx=False` can cause confusion to cx.
3. AKS update command has many existing options using `--enable-xxx` and `--disable-xxx`, why only this one is blocked? https://github.com/Azure/azure-cli/pull/19302

zhoxing-ms: > Are you sure about it? The doc says the default value will be False https://docs.python.org/3/library/argparse.html#action

Yes, I confirmed this through code debug.

zhoxing-ms: > 1. The option has been in aks-preview for months, if we change it, when cx switch from aks-preview to Azure CLI, it will be a breaking change, we should avoid it.

OK, I agree with this point

> 2. Semantically, --enable-xxx and disable-xxx is more clear to cx. disable-xxx=True and disable-xxx=False can cause confusion to cx.

The commands of other modules always use only one parameter, such as `--disable-local-auth`, `--enable-session` and so on. So the CLI users are used to this approach. I personally think that two parameters doing the same thing will appear redundant, and the parameters become more and more difficult to find.

> 3. AKS update command has many existing options using --enable-xxx and --disable-xxx, why only this one is blocked?

My point is not only for this parameter, but for all parameters used in this way. I'm sorry I didn't notice this detail before

zhoxing-ms: Because this is to avoid breaking change and keep consistent with other parameter design style of the AKS module, so I think this design is acceptable

zhoxing-ms (approved):

zhoxing-ms:
@norshtein Please address the conflict file~

src/azure-cli/azure/cli/command_modules/acs/tests/latest/test_aks_commands.py
andyliuliming: synced offline. norshtein will file one work item to track this.

andyliuliming (approved):

norshtein:
Resovled conflict. @zhoxing-ms @andyliuliming  @FumingZhang , thanks all for reviewing!

norshtein:
@zhoxing-ms , could you please help merge the PR?

PR #19705, opened 2021-09-27
[Container] BREAKING CHANGE: `az container create`: Removing parameter `--network-profile` and add new parameter `--acr-identity`

joseph-porter:
**Description**
Updating to API Version 07-01-2021, which includes:
- Removal of `network_profile`, now replaced with the `subnet_ids` property
- Add support for MSI authenticated ACR image pull with the new parameter `acr_identity`.

**Testing Guide**
Updated tests with live run of new API version

**History Notes**

[Container] BREAKING CHANGE: `az container create`: Removing `--network-profile` parameter, property no longer supported
[Container] `az container create`: Add parameter `--acr-identity`  for support of MSI authenticated ACR image pull

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/container/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/container/_params.py
@@ -91,6 +91,7 @@ def load_arguments(self, _):
         c.argument('registry_login_server', help='The container image registry login server')
         c.argument('registry_username', help='The username to log in container image registry server')
         c.argument('registry_password', help='The password to log in container image registry server')
+        c.argument('acr_identity', help='The identity with access to the container registry')
 
     with self.argument_context('container create', arg_group='Azure File Volume') as c:
         c.argument('azure_file_volume_share_name', help='The name of the Azure File share to be mounted as a volume')
--- a/src/azure-cli/azure/cli/command_modules/container/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/container/_params.py
@@ -12,7 +12,7 @@ from azure.cli.core.commands.validators import get_default_location_from_resourc
 from azure.mgmt.containerinstance.models import (
     ContainerGroupRestartPolicy, OperatingSystemTypes, ContainerNetworkProtocol)
 from ._validators import (validate_volume_mount_path, validate_secrets, validate_subnet, validate_msi,
-                          validate_gitrepo_directory, validate_network_profile, validate_image)
+                          validate_gitrepo_directory, validate_image)
 
 # pylint: disable=line-too-long
 
@@ -45,11 +45,6 @@ secrets_type = CLIArgumentType(
     validator=validate_secrets,
     help="space-separated secrets in 'key=value' format.",
     nargs='+'
-)
-
-network_profile_type = CLIArgumentType(
-    validator=validate_network_profile,
-    help="The network profile name or id."
 )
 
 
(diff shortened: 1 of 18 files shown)


src/azure-cli/azure/cli/command_modules/container/_params.py
```diff
@@ -85,7 +80,6 @@ def load_arguments(self, _):
         c.argument('identity_role', options_list=['--role'], help="Role name or id the system assigned identity will have")
 
     with self.argument_context('container create', arg_group='Network') as c:
-        c.argument('network_profile', network_profile_type)
```
zhoxing-ms, line 88: Removing this parameter directly will cause breaking change, right? Could we hide this parameter first and use deprecate info to inform users in advance? Such as: `deprecate_info=c.deprecate(redirect='--acr-identity', hide=True)`

src/azure-cli/azure/cli/command_modules/container/_validators.py
```diff
@@ -77,20 +77,7 @@ def validate_subnet(ns):
...
-            ns.network_profile = resource_id(
-                subscription=get_subscription_id(cmd.cli_ctx),
-                resource_group=ns.resource_group_name,
-                namespace='Microsoft.Network', type='networkProfiles',
-                name=ns.network_profile
-            )
+    if (ns.subnet or ns.vnet) and ns.ip_address:
+        raise CLIError('Can not use "--subnet" or "--vnet" with IP address type "Public".')
+    if (ns.subnet or ns.vnet) and ns.dns_name_label:
+        raise CLIError('Can not use "--subnet" or "--vnet" with "--dns-name-label".')
```
zhoxing-ms, line 83: Could you please use the specific error type `MutuallyExclusiveArgumentError` instead of `CLIError`

src/azure-cli/azure/cli/command_modules/container/custom.py
```diff
@@ -438,23 +413,33 @@ def _create_image_registry_credentials(registry_login_server, registry_username,
...
+                try:
+                    registry_username = prompt(msg='Image registry username: ')
+                except NoTTYException:
+                    raise CLIError('Please specify --registry-username in order to use Azure Container Registry.')
+
+            if not registry_password:
+                try:
+                    registry_password = prompt_pass(msg='Image registry password: ')
+                except NoTTYException:
+                    raise CLIError('Please specify --registry-password in order to use Azure Container Registry.')
```
zhoxing-ms, line 438: Pease use the specific error type `RequiredArgumentMissingError` instead of `CLIError`

src/azure-cli/azure/cli/command_modules/container/_params.py
joseph-porter: Yes, it is a breaking change. We cannot deprecate the parameter because this new API version no longer supports the `network_profile` property and uses a different property instead. The two properties are incompatible and could not be used together.

The `acr-identity` parameter is unrelated to the `network_profile` and is for supporting a different feature.

src/azure-cli/azure/cli/command_modules/container/_validators.py
joseph-porter: Changed to more specific error

src/azure-cli/azure/cli/command_modules/container/custom.py
joseph-porter: Changed to more specific error type

src/azure-cli/azure/cli/command_modules/container/_params.py
zhoxing-ms: OK, could we hide `--network-profile` in the help description first, but still allow users to pass in `--network-profile`?
When users passes in `--network-profile`, we will only prompt that `--network-profile` is deprecated, and will not fail the user's execution, nor do other processing.

If so, we can change it as `c.argument('network_profile', deprecate_info=c.deprecate(hide=True))`
If not, you need to add `[BREAKING CHANGE]` to the PR title to indicate that it contains breaking change

zhoxing-ms:
Please change the PR title to the changed content which is visible to users (the modification of api-version is insensitive to users). For details, please refer to document [doc link](https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules#submitting-pull-requests)

src/azure-cli/azure/cli/command_modules/container/_params.py
joseph-porter: No, it cannot be hidden. The `network-profile` has been completely removed from this API/SDK version, so using that property in any way is no longer an option.

I have changed the title to include breaking change.

zhoxing-ms (approved):

PR #19728, opened 2021-09-29
[Compute] `az vm/vmss create`: Hide alias `--data-delete-option` in help

zhoxing-ms:
**Description**

The service team has a requirement to replace the order of aliases `--data-delete-option` and `--data-disk-delete-option` in help.
Since the order of parameter aliases displayed in help in CLI is fixed and sorted alphabetically, so there is no simple way to exchange the order of those two aliases.
Therefore, I adopted another solution: hide the alias `--data-delete-option` which does not need to be displayed to users in help, and it does not affect users who are using the parameter `--data-delete-option` to continue to use it.

**Testing Guide**

**History Notes**

---


yonzhan:
Compute

yonzhan (approved):

PR #19805, opened 2021-10-08
[Role] BREAKING CHANGE: `az ad sp create-for-rbac`: No role assignment will be created by default

jiasli:
**Description**

It has been nearly a year since we deprecated creating **Contributor** role assignment by default (#16081). It is time to officially remove this behavior.

**Testing Guide**
```
az ad sp create-for-rbac
az ad sp create-for-rbac --role Reader
```


evelyn-ys (approved):

zhoxing-ms (approved):

chasewilson (approved):
LGTM

src/azure-cli/azure/cli/command_modules/role/_help.py
```diff
@@ -389,14 +389,10 @@
     [managed identities](https://aka.ms/azadsp-managed-identities) if available to avoid the need to use credentials.
 
 
-    By default, this command assigns the 'Contributor' role to the service principal at the subscription scope.
-    To reduce your risk of a compromised service principal, use --skip-assignment to avoid creating a role assignment,
-    then assign a more specific role and narrow the scope to a resource or resource group.
+    By default, this command does not create role assignment for the service principal.
```
dcaro, line 392:
```suggestion
    By default, this command does not assign any role to the service principal.
```

jiasli: Thanks for the suggestion. Personally I would prefer `role assignment` as this is a specific term, as in `az role assignment create`.

jiasli: Thanks for the suggestion. Personally I would prefer `role assignment` as this is a specific term, as in `az role assignment create`.

jiasli: On a second though, `assign any role to` is definitely much simpler to understand.

calvinhzy (approved):

src/azure-cli/azure/cli/command_modules/role/_params.py
```diff
@@ -87,9 +87,7 @@ def load_arguments(self, _):
         c.argument('scopes', nargs='+')
         c.argument('role', completer=get_role_definition_name_completion_list)
         c.argument('skip_assignment', arg_type=get_three_state_flag(),
-                   help='Skip creating the default assignment, which allows the service principal to access resources under the current subscription. '
-                        'When specified, --scopes will be ignored. You may use `az role assignment create` to create '
-                        'role assignments for this service principal later.')
+                   deprecate_info=c.deprecate(target='--skip-assignment', hide=True), help='No-op.')
```
jiasli, line 90: `--skip-assignment` is now a no-op argument to keep the command backward compatible.

PR #19877, opened 2021-10-13
{testsdk} Ignore MSAL requests while recording

jiasli:
**Description**

[`azure_devtools.scenario_tests.recording_processors.OAuthRequestResponsesFilter`](https://github.com/Azure/azure-python-devtools/blob/aae33c216028960fbf54e70dcb1c579600034054/src/azure_devtools/scenario_tests/recording_processors.py#L132-L141) doesn't recognize MSAL requests due to the URL change in [AAD v2.0](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow).

```py
class OAuthRequestResponsesFilter(RecordingProcessor):
    """Remove oauth authentication requests and responses from recording."""

    def process_request(self, request):
        # filter request like:
        # GET https://login.microsoftonline.com/72f988bf-86f1-41af-91ab-2d7cd011db47/oauth2/token
        import re
        if not re.match('https://login.microsoftonline.com/([^/]+)/oauth2/token', request.uri):
            return request
        return None
```

This results in all https://login.microsoftonline.com requests being recorded, including

- Tenant discovery request: `GET https://login.microsoftonline.com/54826b22-38d6-4fb2-bad9-b7b93a3e9c5a/v2.0/.well-known/openid-configuration`
- Get access token request: `POST https://login.microsoftonline.com/54826b22-38d6-4fb2-bad9-b7b93a3e9c5a/oauth2/v2.0/token`

As it is impractical to update https://github.com/Azure/azure-python-devtools and use its latest version, we need to create a new `RecordingProcessor`. Actually, all requests to https://login.microsoftonline.com should be ignored as it is related to authentication which is patched in playback mode:

https://github.com/Azure/azure-cli/blob/9c648087e5e7bdc20dcefc595085b8180e66582e/src/azure-cli-testsdk/azure/cli/testsdk/patches.py#L62-L76

**Testing Guide**
Run any test in live mode:

```
azdev test test_revoke_sp_for_rbac --live
```


jiasli:
> We should definitely report this issue back to Python SDK team in the future.

Of course. I have reported it at https://github.com/Azure/azure-python-devtools/issues/74.

PR #19884, opened 2021-10-14
{Profile} `az account list`: Make help message more specific about clouds

jiasli:
**Description**

While testing sovereign clouds for MSAL-based CLI, I found the help message of `az account list` vague about clouds.

Refine it to be more specific:

```
> az account list -h

Command
    az account list : Get a list of subscriptions for the logged in account. By default, only
    'Enabled' subscriptions from the current cloud is shown.

Arguments
    --all                           : List all subscriptions from all clouds, rather than just
                                      'Enabled' ones.
```


yonzhan:
Profile

PR #19898, opened 2021-10-15
{Core} Optimize performance for MSAL migration

jiasli:
**Description**

As authentication logic is now split from `Profile`, performance can be further improved by lazy loading `.auth.identity` module.

```
> Measure-Command {python -m azure.cli version}
TotalMilliseconds : 384.9338

> Measure-Command {az version}
TotalMilliseconds : 404.2362
```

Imported modules can be inspected with

```
python -X importtime -m azure.cli version 2>perf.log; tuna .\perf.log
```

image


yonzhan:
performance Optimization

PR #19900, opened 2021-10-15
{Core} Refine error message for MSAL token cache absence

jiasli:
**Description**

Show a better error message for no cache due to MSAL migration.


jamesc:
Looks good

PR #19925, opened 2021-10-18
[Core] `az extension add/upgrade` Add PIP setup-extras

vilit1:
# Proof of Concept for Setup Extras in `az extension add` and `az extension update`
## Motivation and Business Justification
The purpose of this proof of concept is to allow extension developers and their users specify which packages could be installed optionally, as seen with pip's setup_extras. During installation or updating, the user can specify install-setup-extras to show which extra packages the user wants to install.  

In the case of azure-iot, the user can specify uamqp to show that the user wants to install uamqp. This package requires a decent amount of space and is required for only some operations within azure-iot. Thus, users may or may not want to install this package with the azure-iot extension depending on their workload needs. With the current installation, the extension does not install the extra package. Instead, when the user tries to run a command that needs uamqp, the extension checks if the package is installed and asks the user to install it if needed. However, we had a customer who could not install the package when running the extension due to networking limitations. Instead, they wanted the option to install the extension with the extras at the beginning. 

Specifying setup extras during extension installation would simplify the overall installation process for the user. Thus, the user can specify if they want the extra package installed during the initial installation rather than backtracking or manually installing the extra package. 

This feature would be optional and extensions would be able to opt-in as shown in _Appendix A_. Extensions that do not use setup extras will not be impacted.

### Setup Extras in PIP
This is already a feature in pip. By specifying the extra keyword(s) in brackets after the package name or location, the user can install the package and the extras corresponding to those keywords. The extension writer can specify which keywords correspond to specific extra packages to be installed. 

https://pip.pypa.io/en/stable/cli/pip_install/#examples 

https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies 

## Terms
Setup extra keyword or keyword – this refers to the key in the setup extras mapping. For example, in the extension azure-iot, the keyword “uamqp” maps to the package requirement “uamqp~=1.2”. 

### Modified CLI commands 
The modifed commands are az extension add and az extension update. The parameter added to each is  --install-extras. This parameter would take in a space separated list of setup extra keywords, similar to --pip-extra-index-urls. See Expected behavior for setup extras. 

#### az extension add 
Add an extension. 
```
az extension add --name 
                 [--index] 
                 [--pip-proxy] 
                 [--pip-extra-index-urls] 
                 [--install-extras] 
```
##### Example 
Add the azure-iot extension with the setup extra keyword uamqp. 
```
az extension add -n azure-iot --install-extras uamqp 
```

#### az extension update 
Update an extension. 
```
az extension update --name 
                    [--index] 
                    [--pip-proxy] 
                    [--pip-extra-index-urls] 
                    [--install-extras] 
```
#### Example 
Update the azure-iot extension with the setup extra keyword uamqp. 
```
az extension update -n azure-iot --install-extras uamqp 
```

### Keywords in `--install-extras` 
For each keyword provided in the argument `--install-extras`, the associated extra package requirements would be installed, as seen with the example extension in Appendix A – Example Extension.  

A user may enter one or more incorrect keywords. This implementation fails installation or updating if there is one or more incorrect keywords. This yields two possible scenarios as described below. 

#### A user wants to install the extension but enters at least one incorrect keyword 
 For this scenario, the error message tells the user that they have entered a wrong keyword and gives them the possible keywords. If the user would like to know what packages correspond to those keywords, they can use `az extension list-available -d` to see the metadata.  
image

#### A user wants to install the extension but the extension does not support extras 
For this scenario, if the extension does not have any setup extra keywords, the error message would inform the user that the specified version of the extension does not support setup extras. 
image

### Versioning and updating   
A user may have an up-to-date version of the extension and try to reinstall the extension with the extra packages. The current `az extension add` and `az extension update` do not allow the user to reinstall an extension if it is up-to-date, even if the upgrade parameter is specified. 

This implementation would allow the user to reinstall the extension with extra packages only if the upgrade parameter is provided. If the `--install-extras` parameter is provided without the `--upgrade` parameter, the extension will not update and the command will exit with the same behavior as the `az extension add`. 

#### The same behavior is seen if a user uses `az extension add` without `--upgrade` 
image

If the upgrade parameter is provided, the add command calls the update command. In turn, the update command checks if there is a new version of the extension to install and installs it if there is a newer version of the extension. If the extension is already up to date, then the extension is not reinstalled. The implementation with `--install-extras` will reinstall the extension, even if it is already up to date.  

#### Comparison between the behavior when a user uses `az extension add` with `--upgrade`
image

#### Comparison between the behavior when a user uses `az extension update` 
image

If the upgrade parameter and a specific version is provided, the update command will check if the installed version of the extension is the specific version that is provided. If the extension is already the correct version, then the extension is not reinstalled. The implementation with `--install-extras` will reinstall the extension, even if the correct version is installed. 

#### Comparison between the behavior when a user uses `--upgrade` and `--version`
image

 
As seen above, the logger warning message for `az extension add` has been updated to inform the user that setup extras would be installed if the user provides `--install-extras` argument.  

 

### Failed install or update of the extension 
In the case of a failed `az extension add` or `az extension update`, the current method of failure handling is satisfactory and no changes are required. Since the extension installation is localized, the failure will only affect that extension and the clean up will remove the failed installation. Thus, if a user is trying to install or update an extension with setup extras and sees a failure, the failed extension and the setup extras will be removed. 

### Deletion of the extension 
In the case of removing an extension that was added with setup extras, the current method of az extension remove is satisfactory. Similar to how failure is handled, the setup extras will be removed with the extension. 

### Extensions that do not support extras
Extensions that do not support extra would not be affected. As seen in the example _A user wants to install the extension but the extension does not support extras_, trying to use setup extras with an extension that does not support extras would result in an error. For developers who decide to add extras to their extensions, they should follow documentation provided by Pip and see the examples in Appendix A.

## Appendix A - Example extension 

Let an example extension with the name “sample-extension” have the following setup extras mapping: 
```
{ 
'keyword1':'extra packages1', 
'keyword2':'extra packages2', 
'keyword3':'extra packages3', 
} 
```
### With pip 
To install this extension or package without setup extras with pip, a user would run: 
```
pip install sample-extension 
```
To install this extension with all setup extras with pip, a user would run: 
```
pip install sample-extension['keyword1','keyword2','keyword3'] 
```
To install this extension with one setup extra (keyword2) with pip, a user would run: 
```
pip install sample-extension['keyword2'] 
```
 
### With the modified cli extension 
To add this extension without any setup extras, a user would run: 
```
az extension add -n sample-extension 
```
To add this extension with all setup extras, a user would run: 
```
az extension add -n sample-extension --setup-extras keyword1 keyword2 keyword3 
```
To add this extension with one setup extra (keyword2), a user would run: 
```
az extension add -n sample-extension --setup-extras keyword2 
```

### The setup extras are not the clearest in the metadata. 

Currently, the setup extras defined for the extensions are within the requires_dist of the metadata for the extension. This can be accessed with `az extension list-available -d --query 'name of the extension'` as shown below. 
image

Below is an example of an azure-iot version that does not support extras. 
image

In azure-iot version 0.11.0, the uamqp package is shown as an extra: the keyword uamqp would install uamqp~=1.2. This presentation may be confusing to some users, as the extra is nested in run_requires and there is other data that the user may not need. A separate command, as described below, could present the extension extras in a user friendly manner by returning a mapping of the setup extra keywords to the setup extra package requirements. 

Furthermore, as in this implementation, I suggest having an error message that would help the user understand what setup extras are supported. In this image, the user has entered ua, which is an unrecognized setup extra. The error message includes a list of available setup extra keywords that the user can use. If the user would like to know what packages correspond to those keywords, they can use az extension list-available to see the metadata. 

## Appendix C - Help Documentation
Help documentation can be seen by running `az extension add -h` or `az extension update -h`. The `--install-extras` is listed under _Experimental Pip Arguments_.
image

## Appendix D - Testing
Tests can be found in [test_extension_commands.py](src/azure-cli-core/azure/cli/core/extension/tests/latest/test_extension_commands.py). The two test functions added are test_add_extension_install_setup_extras and test_update_extension_install_setup_extras. The tests ensure that:
- with an extension with no metadata, the add extension with setup_extras fails
- with an extension with metadata that does not include all of the given setup_extras,  the add extension with setup_extras fails
- with an extension with metadata that includes all of the given setup_extras, the add extension with setup_extras succeeds and the _run_pip command was run with the correct setup extra syntax.

Manual testing has also been done to see if the uamqp package was installed with the extension. In the manual testing, uamqp was uninstalled using pip, the `az extension add` command was run with setup extras, and presence of uamqp was checked with pip. 

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/extension/__init__.py
+++ b/src/azure-cli-core/azure/cli/core/extension/__init__.py
@@ -189,15 +189,15 @@ class WheelExtension(Extension):
         metadata_extras = []
         extra_const = " extra == "
         try:
-            if ext_file.endswith('.whl'):
-                ext_whl_metadata = pkginfo.Wheel(ext_file)
-            else:
-                raise ValueError()
+            ext_whl_metadata = pkginfo.Wheel(ext_file)
             # Parse out the extras for readability
-            for ext in ext_whl_metadata.requires_dist:
-                if extra_const in ext:
-                    key = ext[ext.find(extra_const) + len(extra_const) + 1: ext.rfind("'")]
-                    metadata_extras.append(key)
+            if len(ext_whl_metadata.provides_extras) > 0:
+                metadata_extras = ext_whl_metadata.provides_extras
+            else:
+                for ext in ext_whl_metadata.requires_dist:
+                    if extra_const in ext:
+                        key = ext[ext.find(extra_const) + len(extra_const) + 1: ext.rfind("'")]
+                        metadata_extras.append(key)
         except ValueError:
             logger.warning('Could not parse metadata to get setup extras. Will install without extras.')
         return metadata_extras


vilit1:
@digimaun POC for add extensions with extras

yonzhan:
Core

jiasli:
1. Please use [markdown syntax](https://docs.github.com/en/github/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax) to format the PR description.
2. Please provide step-by-step testing guide and commands.
3. (From PM point of view) Please provide the business justification about the change, like how many users will benefit from this change, why manual installing is not acceptable, etc.

vilit1:
@jiasli Please look over the pr again

jiasli:
@vilit1, please follow up with our PM in email "Proposed code changes to az extension add and update" internally.

jiasli:
I have read all the PR description and changes. Personally, I feel this is an over engineering as it is only beneficial to only this `azure-iot` extension. We have no other extensions needing the same functionality.

I think there is also a workaround by manually calling `pip install --target {extension_path} uamqp` after `az extension add`.

src/azure-cli-core/azure/cli/core/extension/__init__.py
```diff
@@ -184,6 +184,24 @@ def get_azext_metadata(ext_dir):
                 azext_metadata = json.load(f)
         return azext_metadata
 
+    @staticmethod
+    def get_metadata_extras(ext_file):
+        metadata_extras = []
+        extra_const = "; extra == "
```
jiasli, line 190: This matching pattern may not be accurate. For example, in `requests` library, its `setup.py` has 

https://github.com/psf/requests/blob/v2.27.1/setup.py#L111

```py
        'socks:sys_platform == "win32" and python_version == "2.7"': ['win_inet_pton'],
```

while the `requests-2.27.1.dist-info/METADATA` from the `whl` downloaded from https://pypi.org/project/requests/#files has

```
Requires-Dist: win-inet-pton ; (sys_platform == "win32" and python_version == "2.7") and extra == 'socks'
```

making `; extra ==` fail to match.

src/azure-cli/azure/cli/command_modules/extension/__init__.py
```diff
@@ -64,6 +64,8 @@ def load_arguments(self, command):
                        arg_group='Experimental Pip')
             c.argument('pip_extra_index_urls', options_list=['--pip-extra-index-urls'], nargs='+',
                        help='Space-separated list of extra URLs of package indexes to use. This should point to a repository compliant with PEP 503 (the simple repository API) or a local directory laid out in the same format.', arg_group='Experimental Pip')
+            c.argument('install_setup_extras', options_list=['--install-extras'], nargs='+',
```
jiasli, line 67: The word "install" is basically a synonym of "setup". The naming of `install_setup_extras` seems inappropriate. Why does it have a different name than `--install-extras`?

src/azure-cli-core/azure/cli/core/extension/tests/latest/test_extension_commands.py
```diff
@@ -397,6 +397,69 @@ def test_list_available_extensions_incompatible_cli_version(self):
             self.assertIsInstance(res, list)
             self.assertEqual(len(res), 0)
 
+    def test_add_extension_install_setup_extras(self):
+        """
+        Tests extension addition while specifying --install-setup-extras parameter.
```
jiasli, line 402: `--install-setup-extras` is not defined.

src/azure-cli-core/azure/cli/core/extension/__init__.py
```diff
@@ -184,6 +184,24 @@ def get_azext_metadata(ext_dir):
...
+        try:
+            if ext_file.endswith('.whl'):
+                ext_whl_metadata = pkginfo.Wheel(ext_file)
+            else:
+                raise ValueError()
+            # Parse out the extras for readability
+            for ext in ext_whl_metadata.requires_dist:
```
jiasli, line 197: This is the `METADATA` from https://github.com/Azure/azure-iot-cli-extension/releases/download/v0.14.0/azure_iot-0.14.0-py3-none-any.whl

```
Provides-Extra: uamqp
Requires-Dist: msrestazure (<2.0.0,>=0.6.3)
Requires-Dist: jsonschema (~=3.2.0)
Requires-Dist: azure-iot-device (~=2.5)
Requires-Dist: tqdm (~=4.62)
Requires-Dist: packaging
Requires-Dist: importlib-metadata; python_version < "3.8"
Provides-Extra: uamqp
Requires-Dist: uamqp (~=1.2); extra == 'uamqp'
```

1. Why does `Provides-Extra: uamqp` appear twice? Though it only appears once in https://github.com/Azure/azure-iot-cli-extension/blob/11637cd8a37f8194497ca28ba9518e7b04dc88c9/setup.py#L54
2. Per my understanding, this `get_metadata_extras` function parses extra names from `Requires-Dist`. However, the **extra name** denoted by `extra == ` may not be the same as the actual **package name**. For example, the extra name may be called `ua` while the actual package name is `uamqp`.
3. Why not use `Provides-Extra` to extract the extra names?

jiasli:
In the PR description, there are many occurrences of `“` `”`. These are not valid characters supported by command line.

src/azure-cli-core/azure/cli/core/extension/operations.py
```diff
@@ -180,6 +190,28 @@ def _add_whl_ext(cli_ctx, source, ext_sha256=None, pip_extra_index_urls=None, pi
...
+                "extras.".format(extension_name)
+            )
+        for ext in install_setup_extras:
+            if ext not in extras:
+                raise CLIError(
+                    "The setup extras keyword {} is invalid. The extension {} currently supports "
+                    "these keywords: {}.".format(ext, extension_name, extras)
+                )
+
+        extras = "[{}]".format(",".join(install_setup_extras))
```
jiasli, line 210: This `_check_setup_extras` does more than "check". It actually returns the extras in [PEP 508 convention](https://peps.python.org/pep-0508/#extras), like `requests[security]`.

jiasli:
I have read all the PR description and changes. Personally, I feel this is an over engineering as it is beneficial to only this `azure-iot` extension. We have no other extensions needing the same functionality.

I think there is also a workaround by manually calling `pip install --target {extension_path} uamqp~=1.2` after `az extension add`. 

We can discuss in more detail offline.

src/azure-cli-core/azure/cli/core/extension/__init__.py
vilit1: From `pkginfo.Wheel` the metadata looks like this:
`
{
  "filename": "C:/Users/vilit/Downloads/azure_iot-0.14.0-py3-none-any.whl",
  "metadata_version": "2.0",
  "name": "azure-iot",
  "version": "0.14.0",
  "platforms": [
    "UNKNOWN"
  ],
  "summary": "The Azure IoT extension for Azure CLI.",
  "home_page": "https://github.com/azure/azure-iot-cli-extension",
  "author": "Microsoft",
  "author_email": "iotupx@microsoft.com",
  "license": "MIT",
  "classifiers": [
    "Development Status :: 4 - Beta",
    "Intended Audience :: Developers",
    "Intended Audience :: System Administrators",
    "Programming Language :: Python",
    "Programming Language :: Python :: 3",
    "Programming Language :: Python :: 3.6",
    "Programming Language :: Python :: 3.7",
    "Programming Language :: Python :: 3.8",
    "Programming Language :: Python :: 3.9",
    "Programming Language :: Python :: 3.10",
    "License :: OSI Approved :: MIT License"
  ],
  "requires_python": ">=3.6,<4",
  "requires_dist": [
    "msrestazure (<2.0.0,>=0.6.3)",
    "jsonschema (~=3.2.0)",
    "azure-iot-device (~=2.5)",
    "tqdm (~=4.62)",
    "packaging",
    "importlib-metadata; python_version < \"3.8\"",
    "uamqp (~=1.2); extra == 'uamqp'"
  ],
  "description": "The Azure IoT extension for Azure CLI. Intended for power users and/or automation of IoT solutions at scale.\n\n\n"
}
`

1. Not sure why - seems like a weird bug on the metadata side. Either way, this metadata file is generated _after_ installation. Once you hit this point in the code, the metadata file does not exist yet.
2. In `requires_dist` the extra name is what we want to use, not the actual package name (especially since we can map one extra name to many packages). So, in this case, if the extra name is bar as in `uamqp (~=1.2); extra == 'bar'`, we would want to extract 'bar' so we can install the package with azure_iot-0.14.0-py3-none-any.whl[bar]
3. See 1 - I am looking at the metadata before the package is installed using `pkginfo.Wheel`. This only provides `requires_dist` so I am using that.

src/azure-cli-core/azure/cli/core/extension/__init__.py
vilit1: good catch. Changing to match ` extra ==`

src/azure-cli/azure/cli/command_modules/extension/__init__.py
vilit1: Fixed

src/azure-cli-core/azure/cli/core/extension/tests/latest/test_extension_commands.py
vilit1: fixed

src/azure-cli-core/azure/cli/core/extension/operations.py
vilit1: renamed

src/azure-cli-core/azure/cli/core/extension/__init__.py
jiasli: > Either way, this metadata file is generated after installation.

This is not true, this `METADATA` file is in the wheel https://github.com/Azure/azure-iot-cli-extension/releases/download/v0.14.0/azure_iot-0.14.0-py3-none-any.whl before installation.

`provides_extras` is empty for `azure_iot-0.14.0-py3-none-any.whl`.

image

Meanwhile, `provide_extras` is correct for https://files.pythonhosted.org/packages/2d/61/08076519c80041bc0ffa1a8af0cbd3bf3e2b62af10435d269a9d0f40564d/requests-2.27.1-py2.py3-none-any.whl

image

Looks like it is because `azure_iot` wheel has 

```
Metadata-Version: 2.0
```

but `requests` wheel has

```
Metadata-Version: 2.1
```

See

- https://peps.python.org/pep-0566/
- https://github.com/pypa/wheel/issues/231

vilit1: To add onto the above - the wheel package, which determines the metadata version, needs to be at least 0.31.0 to use metadata version 2.1.

azdev currently enforces wheel == 0.30.0 https://github.com/Azure/azure-cli-dev-tools/blob/4f97af4466e70496b810c811cc094729dabe2a0a/setup.py#L79

This means that most azure extensions will use the older metadata version of 2.0

PR #19929, opened 2021-10-19
{Core} Make `can_launch_browser` return `False` for Docker container on WSL 2

jiasli:
## Description

A workaround for https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/421

Currently in MSAL-based CLI, running `az login` in docker will hang, as **MSAL doesn't throw an exception when it fails to open a web browser**.

On the other hand, in ADAL-based CLI, if CLI fails to open a web browser, it throws `RuntimeError` and falls back to device code flow:

https://github.com/Azure/azure-cli/blob/14cc787d0f58bc649d402b486fdecc5625eee9ac/src/azure-cli-core/azure/cli/core/_profile.py#L199-L205

## Change

This PR makes `can_launch_browser` support Docker container on WSL 2 so that CLI detects if it can launch a browser first before calling `acquire_token_interactive`, so that it won't call `acquire_token_interactive` and hang. 

If no browser is available, use device code instead.

### DIFF
--- a/src/azure-cli-core/azure/cli/core/tests/test_util.py
+++ b/src/azure-cli-core/azure/cli/core/tests/test_util.py
@@ -161,46 +161,51 @@ class TestUtils(unittest.TestCase):
         else:
             webbrowser_open_mock.assert_called_once_with('http://foo', 2)
 
+    @mock.patch('shutil.which', autospec=True)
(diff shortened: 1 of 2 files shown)


yonzhan:
Core

PR #19977, opened 2021-10-21
{Core, Find} az find: Change to Allow Event ID to be Shared with Modules

mirdaki:
**Description**

Being able to include the Event ID with Aladdin telemetry will help the Aladdin team create a more accurate event mapping to the CLI's telemetry. For instance, this will allow us to correctly connect CLI error events with Aladdin telemetry in the "Did you mean" scenario. Connecting to the actual error event will help us better diagnose user errors and understand the context of their problem. That scenario will be added in a PR after this one is merged.

The existing Correlation ID is not unique for each operation (it might be shared with all operations in an `az interactive` session for instance), so that ID is not sufficient.

The Event ID is uniquely created per command operation and is considered non PII information. In addition, we will not collect it if the user disables telemetry. 

**Testing Guide**
N/A

**History Notes**
N/A

---


jiasli:
It will be great if the PR description can include how the event ID is used to "better identify user issues and analyze data in aggregate" with some concrete examples.

jiasli:
> Connecting to the actual error event will help us better diagnose user errors and understand the context of their problem. That scenario will be added in a PR after this one is merged.

I still fail to understand how this works without the PR that will be created later. 😥

src/azure-cli-core/azure/cli/core/telemetry.py
```diff
@@ -105,7 +106,7 @@ def generate_payload(self):
                 props.update(base)
                 props.update(cli)
                 props.update({CORRELATION_ID_PROP_NAME: str(uuid.uuid4()),
-                              'Reserved.EventId': str(uuid.uuid4())})
+                              'Reserved.EventId': _get_or_create_event_id()})
```
jiasli, line 109: It not good programming practice to let an external function like `_get_or_create_event_id` set the instance's attribute (`self.event_id`). Consider using a [method](https://docs.python.org/3/tutorial/classes.html#instance-objects) instead.

jiasli, line 109: It not good programming practice to let an external function like `_get_or_create_event_id` set the instance's attribute (`self.event_id`). Consider using a [method](https://docs.python.org/3/tutorial/classes.html#instance-objects) instead.

jiasli, line 109: It not good programming practice to let an external function like `_get_or_create_event_id` set the instance's attribute (`self.event_id`). Consider using a [method](https://docs.python.org/3/tutorial/classes.html#instance-objects) instead.

mirdaki:
> I still fail to understand how this works without the PR that will be created later. 😥

We didn't include the "Did you mean" change in this PR, because we wanted to keep the PRs small and focused. However, we do have an example of using the Event ID change with `az find` in the current PR.

[Here we pull in the Event ID](https://github.com/Azure/azure-cli/pull/19977/files#diff-5f25482082fac110325955ba166bb8159ee01a00b7a910c3bc8fd691bdaf2f99R100) so it can be added in the request to the server. Without this change, the Event ID pulled at that line would be different than the one used used by the CLI telemetry, which run automatically after the command ends. That's because the previous code generated a new Event ID every time it was called. This PR keeps the same Event ID for the lifetime of the command execution.

src/azure-cli-core/azure/cli/core/telemetry.py
mirdaki: I was following the pattern used by functions like the `_get_session_id()` and `_get_hash_machine_id()` in the file to stay consistent. If you like, I can implement it as a method.

mirdaki: I have switched it to a method

jiasli:
Here is my understand of the actual behavior:

## `correlation_id`

### Interactive

In an interactive session, `correlation_id` keeps the same for multiple commands.

```
az cmd1  <guid1>
az cmd2  <guid1>
```

### Non-interactive

In a non-interactive session, `correlation_id` is different for each command.

```
az cmd1  <guid1>
az cmd2  <guid2>
```


## `event_id`

`event_id` is not related to interactive mode.

### Before this PR

`event_id` is different for each telemetry record:

```
az cmd1
- send to client telemetry  guid1
- send to Aladdin           guid2

az cmd2
- send to client telemetry  guid3
- send to Aladdin           guid4
```

### After this PR

`event_id` is the same for telemetry records within the same command:

```
az cmd1
- send to client telemetry  guid1
- send to Aladdin           guid1

az cmd1
- send to client telemetry  guid2
- send to Aladdin           guid2
```

This allows Aladdin to bind **the record sent to client telemetry** and **the record sent to Aladdin**.

mirdaki:
Yes, that is the behavior we're trying to achieve!

jiasli (approved):

PR #20145, opened 2021-11-02
{AKS} Refactor 8th part of the update sub-command in acs module

FumingZhang:
**Description**

* Refactor the update of windows profile

Unit Test Coverage

- "covered_lines": 1642,
- "num_statements": 1654,
- "percent_covered": 99.27448609431681,
- "missing_lines": 12,
- "excluded_lines": 0

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acs/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/_params.py
@@ -259,7 +259,7 @@ def load_arguments(self, _):
                    '--windows-admin-username'])
         c.argument('windows_admin_password', options_list=[
                    '--windows-admin-password'])
-        c.argument('enable_ahub', options_list=['--enable-ahub'])
+        c.argument('enable_ahub', options_list=['--enable-ahub'], action='store_true')
         c.argument('node_osdisk_diskencryptionset_id', type=str,
                    options_list=['--node-osdisk-diskencryptionset-id', '-d'])
         c.argument('aci_subnet_name')
@@ -315,8 +315,8 @@ def load_arguments(self, _):
         c.argument('auto_upgrade_channel', arg_type=get_enum_type(auto_upgrade_channels))
         c.argument('api_server_authorized_ip_ranges',
                    type=str, validator=validate_ip_ranges)
-        c.argument('enable_ahub', options_list=['--enable-ahub'])
-        c.argument('disable_ahub', options_list=['--disable-ahub'])
+        c.argument('enable_ahub', options_list=['--enable-ahub'], action='store_true')
+        c.argument('disable_ahub', options_list=['--disable-ahub'], action='store_true')
         c.argument('enable_public_fqdn', action='store_true')
         c.argument('disable_public_fqdn', action='store_true')
         c.argument('windows_admin_password', options_list=[
--- a/src/azure-cli/azure/cli/command_modules/acs/decorator.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/decorator.py
@@ -1536,19 +1536,19 @@ class AKSContext:
 
         return self._get_windows_admin_username_and_password()
 
-    def get_windows_admin_username(self) -> Union[str, None]:
-        """Obtain the value of windows_admin_username.
+    def get_windows_admin_password(self) -> Union[str, None]:
+        """Obtain the value of windows_admin_password.
 
         Note: This function is intended to be used in update mode.
 
         :return: string or None
         """
         # read the original value passed by the command
-        windows_admin_username = self.raw_param.get("windows_admin_username")
+        windows_admin_password = self.raw_param.get("windows_admin_password")
 
         # this parameter does not need dynamic completion
         # this parameter does not need validation
-        return windows_admin_username
+        return windows_admin_password
 
     # pylint: disable=unused-argument
     def _get_enable_ahub(
@@ -1565,11 +1565,10 @@ class AKSContext:
         """
         # read the original value passed by the command
         enable_ahub = self.raw_param.get("enable_ahub")
-        # try to read the property value corresponding to the parameter from the `mc` object
-        if self.mc and self.mc.windows_profile:
-            enable_ahub = (
-                self.mc.windows_profile.license_type == "Windows_Server"
-            )
+        # In create mode, try to read the property value corresponding to the parameter from the `mc` object
+        if self.decorator_mode == DecoratorMode.CREATE:
+            if self.mc and self.mc.windows_profile:
+                enable_ahub = self.mc.windows_profile.license_type == "Windows_Server"
 
         # this parameter does not need dynamic completion
         # validation
@@ -1605,9 +1604,7 @@ class AKSContext:
         """
         # read the original value passed by the command
         disable_ahub = self.raw_param.get("disable_ahub")
-        # try to read the property value corresponding to the parameter from the `mc` object
-        if self.mc and self.mc.windows_profile:
-            disable_ahub = self.mc.windows_profile.license_type == "None"
+        # we do not support this option in create mode
 
         # this parameter does not need dynamic completion
         # validation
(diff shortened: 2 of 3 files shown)


yonzhan:
AKS

src/azure-cli/azure/cli/command_modules/acs/decorator.py
```diff
@@ -3498,7 +3576,7 @@ def _get_enable_private_cluster(self, enable_validation: bool = False, **kwargs)
                         raise InvalidArgumentValueError(
                             "--enable-public-fqdn can only be used for private cluster"
                         )
-                    # In fact, there is currently no such option in the create command
+                    # In fact, there is no such option in the create command.
                     raise InvalidArgumentValueError(
```
wangzelin007, line 3580: I was wondering why need two raise here, and the raise info is almost same.

src/azure-cli/azure/cli/command_modules/acs/decorator.py
```diff
@@ -1532,23 +1536,97 @@ def get_windows_admin_username_and_password(
...
+        specified, raise a MutuallyExclusiveArgumentError.
+
         :return: bool
         """
         # read the original value passed by the command
         enable_ahub = self.raw_param.get("enable_ahub")
-        # try to read the property value corresponding to the parameter from the `mc` object
-        if self.mc and self.mc.windows_profile:
-            enable_ahub = self.mc.windows_profile.license_type == "Windows_Server"
+        # In create mode, try to read the property value corresponding to the parameter from the `mc` object
```
AbelHu, line 1568: What does it mean `In create mode`? enable/disable ahub can be used in both create and update. But to change `Windows_Server`, it should be an update operation

src/azure-cli/azure/cli/command_modules/acs/decorator.py
```diff
@@ -5302,6 +5373,29 @@ def update_api_server_access_profile(self, mc: ManagedCluster) -> ManagedCluster
...
+
+        if not mc.windows_profile:
+            raise UnknownError(
+                "Encounter an unexpected error while getting windows profile from the cluster in the process of update."
+            )
+
+        enable_ahub = self.context.get_enable_ahub()
+        disable_ahub = self.context.get_disable_ahub()
+        windows_admin_password = self.context.get_windows_admin_password()
+        if enable_ahub:
```
AbelHu, line 5391: Could `enable_ahub` be `none`? We should not update ahub if neither --enable-ahub and --disable-ahub is used.

src/azure-cli/azure/cli/command_modules/acs/decorator.py
FumingZhang: Currently, we are refactoring `aks create` and `aks update`, and `In create mode` means this only applies to the case of `aks create` but not `aks update`.

I suppose we have the `--enable-ahub` option for both `aks create` and `aks update`, but for option `--disable-ahub`, it is only supported by `aks update`.

FumingZhang: Back to the topic, in the create mode (`aks create`), when determining the value of a parameter, obtaining from `mc` takes precedence over user's input. While in the update mode (`aks update`), we do not obtain the value from `mc`.

src/azure-cli/azure/cli/command_modules/acs/decorator.py
FumingZhang: No, `enable_ahub` could not be `None` since its default value is `False` defined [here](https://github.com/FumingZhang/azure-cli/blob/fuming/acs-update-np8/src/azure-cli/azure/cli/command_modules/acs/custom.py#L2363-L2364) and the user's input will only assign a value other than `None`.

AbelHu: We only set license_type when either is true. You do not change that. LGTM

src/azure-cli/azure/cli/command_modules/acs/decorator.py
AbelHu: Since we do not obtain the value from mc when in the update mode, why do we not use the same for the create model?

src/azure-cli/azure/cli/command_modules/acs/decorator.py
FumingZhang: By the way, I suppose `--enable-ahub` and `--disable-ahub` should be boolean flags instead of normal options which would take an argument? If so, I would like to add `action='store_true'` to the following places to express this clearly (by default, the option will be parsed as an ordinary parameter, and now it is implicitly inferred by the cli through the default value of the parameter in the function of the boolean type). 

- [create: --enable-ahub](https://github.com/FumingZhang/azure-cli/blob/fuming/acs-update-np8/src/azure-cli/azure/cli/command_modules/acs/_params.py#L262)
- [update: --enable-ahub](https://github.com/FumingZhang/azure-cli/blob/fuming/acs-update-np8/src/azure-cli/azure/cli/command_modules/acs/_params.py#L318)
- [update: --disable-ahub](https://github.com/FumingZhang/azure-cli/blob/fuming/acs-update-np8/src/azure-cli/azure/cli/command_modules/acs/_params.py#L319)

AbelHu: They are boolean flags. Please do it.

FumingZhang: Changed in commit [#7b3cb27](https://github.com/Azure/azure-cli/pull/20145/commits/7b3cb2707dbc1b757c771d2d004ef45c7928b73b).

src/azure-cli/azure/cli/command_modules/acs/decorator.py
FumingZhang: In the update mode, the parameter value is not obtained from `mc` because we just need to use the value provided by the user to overwrite the value in `mc`.

But in the create mode, `mc` is completely blank, and we want to assemble the user's input into a complete `mc`. Considering that our parameters have a certain dependency or mutual exclusion relationship, or the parameters will be normalized before putting into `mc`, there are even some parameters that do not need user to provide explicitly, but can be dynamically completed by cli. In order to ensure the consistency of the state and make the value of the parameter more clearly determined, we created the following principle in the create mode, that is, if a parameter has been placed in the `mc` object, then when trying to get the parameter value again, (e.g. cross-validation with other parameters) the value in `mc` shall prevail.

Going back to the example in the windows profile, the current windows related parameters will not be used after being set into `mc`, but in order to keep the form consistent with other parameters, we keep such acquisition method in the create mode.

AbelHu (approved):

src/azure-cli/azure/cli/command_modules/acs/_params.py
```diff
@@ -315,8 +315,8 @@ def load_arguments(self, _):
         c.argument('auto_upgrade_channel', arg_type=get_enum_type(auto_upgrade_channels))
         c.argument('api_server_authorized_ip_ranges',
                    type=str, validator=validate_ip_ranges)
-        c.argument('enable_ahub', options_list=['--enable-ahub'])
-        c.argument('disable_ahub', options_list=['--disable-ahub'])
+        c.argument('enable_ahub', options_list=['--enable-ahub'], action='store_true')
+        c.argument('disable_ahub', options_list=['--disable-ahub'], action='store_true')
```
zhoxing-ms, line 319: Will the change of parameter type cause breaking change to users?

zhoxing-ms (approved):

src/azure-cli/azure/cli/command_modules/acs/_params.py
FumingZhang: By default, the option will be parsed as an ordinary parameter (which should take a string-like argument), and before it was implicitly inferred by the cli through the default value of the parameter in the function of the bool type. Now I am adding action='store_true' to the these places to express this clearly.

Issue #20172, opened 2021-11-03
multiple accounts with same username


Tomnowell:
console output and error message:

The default web browser has been opened at https://login.microsoftonline.com/organizations/oauth2/v2.0/authorize. Please continue the login in the web browser. If no web browser is available or if the web browser fails to open, use device code flow with `az login --use-device-code`. Found multiple accounts with the same username. Please report to us via Github: https://github.com/Azure/azure-cli/issues/new

Error encountered trying to log in to Azure CLI on local system for the first time.  Sorry, still a newb! :D

Not sure exactly how this happened.  I've tried 'merging' the accounts in my account (with a browser, too).  But they still show up separately.  One says it was created by my IT department, and one created by me.  I don't have an IT department, so that can't be totally correct.

yonzhan:
@jiasli for awareness

jiasli:
Could you check https://github.com/Azure/azure-cli/issues/20168#issuecomment-958765085 and share the required information. Many thanks!

jiasli:
Let's work on https://github.com/Azure/azure-cli/issues/20168.

Issue #20168, opened 2021-11-03
Found multiple accounts with same username. 


dylankdreyer78:
When trying to log in using the CLI I get this error:

![image](https://user-images.githubusercontent.com/93645505/140031297-c84248ad-8429-4419-910a-bdb3c9cca3e2.png)

I have obviously made a mess of things here but I don't know how to sort this out now :/ 

Any ideas?

jiasli:
Could you please share some extra information:

1. How is Azure CLI installed? Did you install it following https://docs.microsoft.com/en-us/cli/azure/install-azure-cli-windows?tabs=azure-cli or using `pip`?
2. Could you turn off token encryption first and log in again?
    ```
    az config set core.token_encryption=false
    az login
    ```
    Then share the content of `Account` property in `C:\Users\<username>\.azure\msal_token_cache.json`.

dylankdreyer78:
Thanks so much for pointing me in the right direction. cleared the cache and works 100% now!

jiasli:
@dylankdreyer78, I am glad to know it works for you. Do you mind letting us know what action you have performed? Thanks a lot.

dylankdreyer78:
No problem. I deleted the json files in the .azure directory :)

jiasli:
Hmm... This still didn't reveal why it happened in the first place. Anyway, we will keep monitoring for other occurrences. Thanks for letting us know.

hivyas:
Hi there, I am seeing the same issue. Deleting the cache file and relogging in did not work. I also only see a `msal_token_cache.bin` file, not a .json file.

jiasli:
@hivyas, as shown in the output, these 2 accounts share the same `"home_account_id": "508c96e0-d4ff-4c30-b886-34c377353339.72f988bf-86f1-41af-91ab-2d7cd011db47"`. In such case, MSAL will group them (https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/349). 

Are you using Azure CLI from source code? If so, please verify if your MSAL is update to date and configure all Azure CLI dependencies again.

```
pip list
azdev setup --cli
```

Tomnowell:
Hi jiasli, Thanks for the email.  I'm in a similar situation.  

I Installed Azure CLI from .msi package 

Heres contents of unencrypted account tokens

`"Account": {
[...]

PR #20173, opened 2021-11-03
{Auth} More detailed error message when MSAL returns multiple accounts 

jiasli:
**Description**

Previously we thought `msal.application.ClientApplication.get_accounts` can't return multiple accounts because MSAL groups them (https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/349). That's why we added a double check:

https://github.com/Azure/azure-cli/blob/bd081bb9a4449091bfcce50e2b51292d8a1591d8/src/azure-cli-core/azure/cli/core/auth/msal_authentication.py#L46-L48

But now we do see this happening: https://github.com/Azure/azure-cli/issues/20172, https://github.com/Azure/azure-cli/issues/20168.

Even though the root cause is unknown, its better to expose more information in the error message.

### DIFF
--- a/src/azure-cli-core/azure/cli/core/auth/msal_authentication.py
+++ b/src/azure-cli-core/azure/cli/core/auth/msal_authentication.py
@@ -43,7 +43,7 @@ class UserCredential(PublicClientApplication):
 
         accounts = self.get_accounts(username)
 
-        # MSAL should never return multiple accounts for the same username
+        # Usernames are usually unique. We are collecting corner cases to better understand its behavior.
         if len(accounts) > 1:
             raise CLIError(f"Found multiple accounts with the same username '{username}': {accounts}\n"
                            "Please report to us via Github: https://github.com/Azure/azure-cli/issues/20168")
--- a/src/azure-cli-core/azure/cli/core/auth/msal_authentication.py
+++ b/src/azure-cli-core/azure/cli/core/auth/msal_authentication.py
@@ -43,16 +43,14 @@ class UserCredential(PublicClientApplication):
 
         accounts = self.get_accounts(username)
 
+        # MSAL should never return multiple account for the same username
+        assert len(accounts) <= 1, (f"Found multiple accounts with the same username '{username}': {accounts}\n"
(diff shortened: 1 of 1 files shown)


yonzhan:
Auth

calvinhzy (approved):

src/azure-cli-core/azure/cli/core/auth/msal_authentication.py
```diff
@@ -38,17 +38,20 @@ def __init__(self, client_id, username, **kwargs):
...
 
-        accounts = self.get_accounts(username)
+        # Make sure username is specified, otherwise MSAL returns all accounts
+        assert username, "username must be specified, got {!r}".format(username)
 
-        if not accounts:
-            raise CLIError("User {} does not exist in MSAL token cache. Run `az login`.".format(username))
+        accounts = self.get_accounts(username)
 
+        # MSAL should never return multiple accounts for the same username
```
rayluo, line 46: This statement is not strictly true. [In MSAL, an account's `username` is primarily derived by the `preferred_username` claim from the ID token](https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/1.16.0/msal/token_cache.py#L193), which is [not guaranteed to be unique](https://docs.microsoft.com/en-us/azure/active-directory/develop/id-tokens#payload-claims). I do like the Az CLI error message, though.

```suggestion
        # Usernames are usually unique. We are collecting corner cases to better understand its behavior.
```

rayluo (approved):

evelyn-ys (approved):

src/azure-cli-core/azure/cli/core/auth/msal_authentication.py
```diff
@@ -38,17 +38,20 @@ def __init__(self, client_id, username, **kwargs):
         """
         super().__init__(client_id, **kwargs)
 
-        accounts = self.get_accounts(username)
+        # Make sure username is specified, otherwise MSAL returns all accounts
+        assert username, "username must be specified, got {!r}".format(username)
```
jiasli, line 42: We don't raise `CLIError` because username should never happen. When that indeed happens, it is a bug.

Issue #20155, opened 2021-11-02
Add Instance view to RunCommand


koralski:
**Is your feature request related to a problem? Please describe.**
Current run-command only supports --expand that is not easily discoverable what parameter to pass.
This change applies to:
az vm run-command
az vmss run-command

**Describe the solution you'd like**
Add alternative parameter --status that will be the same as --expand instance-view
This is most common use scenario to track a run command progress, so we need to make it easy

**Describe alternatives you've considered**
Improve documentation/help to show the only option for expand is instance-view, or make this a default parameter if
only --expand is specified

**Additional context**

yonzhan:
Compute

PR #20191, opened 2021-11-04
[Compute] `az vm/vmss run-command show`: Add new parameter `--instance-view` to support tracking the progress of RunCommand

PARADISSEEKR:
**Description**

Add Instance view to RunCommand https://github.com/Azure/azure-cli/issues/20155

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/vm/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_params.py
@@ -804,6 +804,7 @@ def load_arguments(self, _):
     with self.argument_context('vm run-command show') as c:
         c.argument('vm_name', run_cmd_vm_name)
         c.argument('run_command_name', run_cmd_name_type)
+        c.argument('expand', deprecate_info=c.deprecate(hide=True))
         c.argument('status', action='store_true', help='Track the run command progress')
         c.argument('location', arg_type=get_location_type(self.cli_ctx))
         c.argument('command_id', help='The command id.')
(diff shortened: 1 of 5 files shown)


yonzhan:
Compute

src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -804,14 +804,14 @@ def load_arguments(self, _):
     with self.argument_context('vm run-command show') as c:
         c.argument('vm_name', run_cmd_vm_name)
         c.argument('run_command_name', run_cmd_name_type)
-        c.argument('expand', help='The expand expression to apply on the operation.')
+        c.argument('status', action='store_true', help='Track the run command progress')
```
zhoxing-ms, line 807: It is recommended to use `--instance-view` instead of `--status`, because the semantics of `--instance-view` is clearer and can be consistent with the usage of other commands

src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -804,14 +804,14 @@ def load_arguments(self, _):
     with self.argument_context('vm run-command show') as c:
         c.argument('vm_name', run_cmd_vm_name)
         c.argument('run_command_name', run_cmd_name_type)
-        c.argument('expand', help='The expand expression to apply on the operation.')
```
zhoxing-ms, line 807: In addition, please do not delete the previous parameter directly, and deprecate it to avoid breaking change to users

src/azure-cli/azure/cli/command_modules/vm/_params.py
zhoxing-ms: Screenshot 2021-11-15 140722
@PARADISSEEKR We have synced offline

zhoxing-ms (approved):

PR #20261, opened 2021-11-09
[App Service] `az functionapp deployment source config-zip`: Fix returning success when the deployment did not succeed

philipfalkner:
**Description**
Fixes an issue with `az functionapp deployment source config-zip` when Kudu returns a 5xx response code instead of the expected 202.

Before this change:
* If Kudu responds with a 5xx response code to the /api/zipdeploy endpoint, Azure CLI continues anyway, and requests the latest deployment log
* The latest deployment log contains whatever deployment was done _before_ this one, and was likely successful
* So the command overall succeeds (exit code 0)

After this change:
* If Kudu responds with a 5xx response code to the /api/zipdeploy endpoint, Azure CLI immediately ends with an error because the deployment failed

**Testing Guide**
1. Run a command like `az functionapp deployment source config-zip -g MyResourceGroup -n MyFunctionApp --src-file source.zip`
   * Also applies to `az webapp deployment source config-zip` since the two commands share this code.
2. Note in the stderr output the line "WARNING: Deployment endpoint responded with status code 999"
3. If the status code is anything other than 202, the command should have printed an error to stderr and ended with a non-zero exit code.

---


yonzhan:
App Service

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -533,10 +539,10 @@ def enable_zip_deploy(cmd, resource_group_name, name, src, timeout=None, slot=No
...
                        "--setting-names <setting-names> to delete them.".format(deployment_status_url))
 
-    # check the status of async deployment
-    response = _check_zip_deployment_status(cmd, resource_group_name, name, deployment_status_url,
-                                            authorization, timeout)
-    return response
+    # check if an error occured during deployment
+    if res.status_code:
+        raise CLIError("An error occured during deployment. Status Code: {}, Details: {}"
+                       .format(res.status_code, res.text))
```
zhoxing-ms, line 545: Could you please use the specific error type from `azure.cli.core.azclierror` instead of `CLIError` ?

zhoxing-ms:
Could you add some tests for this change?

src/azure-cli/azure/cli/command_modules/appservice/custom.py
philipfalkner: `AzureInternalError` looked like the best option.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -521,6 +521,12 @@ def enable_zip_deploy(cmd, resource_group_name, name, src, timeout=None, slot=No
...
 
+    # check the status of async deployment
+    if res.status_code == 202:
+        response = _check_zip_deployment_status(cmd, resource_group_name, name, deployment_status_url,
+                                                authorization, timeout)
+        return response
+
     # check if there's an ongoing process
     if res.status_code == 409:
         raise CLIError("There may be an ongoing deployment or your app setting has WEBSITE_RUN_FROM_PACKAGE. "
```
wangzelin007, line 532: Same as above.

philipfalkner: Changed to `UnclassifiedUserFault`

wangzelin007 (approved):

zhoxing-ms (approved):

Issue #20269, opened 2021-11-10
Global Arguments stop working in Python 3.9.8


jiasli:
## Symptom

Global Arguments now stop working in Python 3.9.8:

```
> az cloud show --query profile
{
  "endpoints": {
    "activeDirectory": "https://login.microsoftonline.com",
    ...
```

Notice `--query` doesn't take effact.

## Root cause

This is due to the `argparse` breaking change https://github.com/python/cpython/pull/28420 which aims to solve https://bugs.python.org/issue45235. Python 3.9.8 is the first released version with this change. It will spread to all Python versions later.

In Knack, global arguments are registered on all root/intermediate/leaf parsers, and they share the same `dest`. Previously, **leaf-level arguments takes precedence**. After this change, **root-level arguments take precedence**.

We are not sure if this is intended or not, as according to [Semantic Versioning](https://semver.org/), a PATCH version update shouldn't cause breaking change.

Since Azure CLI does some trimming on the command table, using `--query` on the root parser won't work. We have to use Knack directly to demonstrate it:

| command | 3.9.7 | 3.9.8
| - | - | - |
| `python .\knack\examples\exapp2 sample-json --query id` | ✔ | ❌
| `python .\knack\examples\exapp2 --query id sample-json` | ❌ | ✔

## Proposed solution

Knack should only register global arguments on leaf parsers to avoid this issue. We don't allow using global arguments in the middle of a command anyway:

- ✔ `az cloud show --query profile`
- ❌ `az --query profile cloud show`
- ❌ `az cloud --query profile show`

## References

This issue is also affecting other CLI tools utilizing `argparse`: 

- https://github.com/sopel-irc/sopel/issues/2210

A related topic for `--output` was extensively discussed before:

- https://github.com/Azure/azure-cli/issues/7877

jiasli:
This issue is also affecting other CLI tools utilizing `argparse`: https://github.com/sopel-irc/sopel/issues/2210

gvanrossum:
@rhettinger @ambv Can you look at this? It looks like a serious regression in 3.9.8.

jiasli:
Thank you so much @gvanrossum for the prompt response and fix! Really appreciate your help!

jiasli:
For our own tracking purposes: 

- The patch to revert patch on `main` branch: https://github.com/python/cpython/pull/29525
- Backporting to `python:3.9`: https://github.com/python/cpython/pull/29531

jiasli:
[...]

Issue #19857, opened 2021-10-12
Support Python 3.10


jiasli:
## Description

Python 3.10 has been released on 2021-10-04: https://www.python.org/downloads/

Azure CLI needs to support Python 3.10 like how we supported 3.9 (https://github.com/Azure/azure-cli/issues/16611).

## TODOs

- [x] [Knack](https://github.com/microsoft/knack) supports Python 3.10: https://github.com/microsoft/knack/pull/250
- [x] [azdev](https://github.com/Azure/azure-cli-dev-tools) supports Python 3.10: https://github.com/Azure/azure-cli-dev-tools/pull/319
- [x] CI passes with Python 3.10
  - [x] Dependencies like `pytest`, `pylint` are updated to the Python 3.10 compatible version: https://github.com/Azure/azure-cli/pull/20192
  - [x] New `pylint` check failures are fixed or ignored: https://github.com/Azure/azure-cli/pull/20192
  - [x] All test cases pass: https://github.com/Azure/azure-cli/pull/20195

yonzhan:
Python 3.10 support

miqm:
@jiasli - will the Docker image be upgraded to python 3.10.x? currently is 3.9.6 (which is not the latest one though)

jiasli:
@miqm, actually we use Python 3.8 in DEB package and Windows MSI, and Python 3.6 in RPM. 😅

Are there any absolute needs to use Python 3.10 in the docker image? I can certainly prioritize it if they are reasonable.

miqm:
@jiasli - It's not a must, however often people use the CLI image in CI/CD systems and some new custom-made scripts might benefit from newer python features.

In general it would be awesome if you produce a series of azure-cli docker tags so people could choose which kind of base image they would like to use (i.e. python3.9-slim-bullseye, python3.8-alpine, etc.)

jiasli:
@miqm, this is a good suggestion. I have created https://github.com/Azure/azure-cli/issues/21074 to track it.

jiasli:
@miqm, upgrade to Python 3.10 in https://github.com/Azure/azure-cli/pull/21079.

PR #20287, opened 2021-11-11
{CI} Use Python 3.10 as the default and disable Python 3.9 tests

jiasli:
## Description

Python 3.9.8 introduced an unexpected breaking change (https://github.com/python/cpython/pull/28420) that breaks Azure CLI:

- https://github.com/Azure/azure-cli/issues/20269

According to https://bugs.python.org/issue45235, this change will be reverted on 3.9 and withdrawn from future releases for other version like 3.7, 3.8, 3.10.

Azure Pipeline now only has Python 3.9.8, but not old versions like 3.9.7:

- https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/tool/use-python-version
- https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#python
    > #### Python
    > - ...
    > - 3.9.8
    > - 3.10.0

Even if Python releases 3.9.9 which reverts this change, it will take Azure Pipeline weeks to take that version.

To unblock our development and PRs from being merged, we have to use Python 3.10 as the default (thanks to the timely support for Python 3.10 (https://github.com/Azure/azure-cli/issues/19857))  and disable Python 3.9 tests.

### DIFF
--- a/azure-pipelines-full-tests.yml
+++ b/azure-pipelines-full-tests.yml
@@ -24,8 +24,8 @@ jobs:
         python.version: '3.6'
       Python38:
         python.version: '3.8'
-      Python39:
-        python.version: '3.9'
+      # Python39:
+      #   python.version: '3.9'
       Python310:
         python.version: '3.10'
   steps:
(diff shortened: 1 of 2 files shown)


yonzhan:
CI

yonzhan (approved):

evelyn-ys (approved):

calvinhzy (approved):

wangzelin007 (approved):

zhoxing-ms (approved):

Issue #19773, opened 2021-10-04
Support for cross-region copy of incremental snapshots


jamesc:
**Resource Provider**
Microsoft.Compute

**Description of Feature or Work Requested**
Support CopyStart of incremental snapshots of managed disks via CLI. The REST API changes introduces a new option called CopyStart for the creationOption for creating snapshots of managed disks. 
```
{ 
  "id": "/subscriptions/sub/resourceGroups/rg1e47ff76-fc12-483f-9c76-4b74673669ef/providers/Microsoft.Compute/disks/slowDisk",    
  "name": "slowDisk"    
  "type": "Microsoft.Compute/snapshots",    
  "location": "westus",    
  "sku": {    
    "name": "Standard_LRS",    
    "tier": "Standard"    
  },    
  "properties": {    
    "osType": "Windows",    
    "creationData": {    
      "createOption": "CopyStart", 
      "sourceResourceId": "/subscriptions/sub/resourceGroups/resg/providers/Microsoft.Compute/snapshots/mySnapshot
    }
  },    
  "tags": {    
    "department": "Development",    
    "project": "ManagedDisks"    
  }   
  "provisioningState": "Pending",  
  "completionPercent": 75 // readonly: ignored on inputs 
}   
```

**Minimum API Version Required**

**Swagger Link**
* https://github.com/Azure/azure-rest-api-specs/pull/15665
* https://github.com/Azure/azure-rest-api-specs/pull/15665/commits/0f1cd823abe0c4508984619b0fb94387c8dac4fd

Example: https://github.com/rytruong/azure-rest-api-specs/blob/1f561a67f29460998e4ff8020e0691779ff50756/specification/compute/resource-manager/Microsoft.Compute/stable/2021-04-01/examples/CreateASnapshotFromAnExistingSnapshotInDifferentRegion.json

**Target Date**
Ignite

yonzhan:
@zhoxing-ms for awareness

zhoxing-ms:
@jamesc So CLI needs to add a new parameter to support Copy Start, right? Could you provide the design which includes specific commands and parameters that need to be modified?

evelyn-ys:
Thanks @vikramd-ms @ramankumarlive. So we will autodetect `CopyStart` for cross-region scenario and provide an extra `--copy-start` to specify forcibly.

Do both `az disk create` and `az snapshot create` need to change or only `az snapshot create` need this?

ramankumarlive:
@arorasam 

[...]

PR #20306, opened 2021-11-12
[Compute] Fix #20258: `az snapshot create`: Fix creating a snapshot of a Uniform VMSS OS disk

### LINKED ISSUE #20258: Issues when trying to create a snapshot of a Uniform VMSS OS disk in AzureCLI
## Describe the bug

**Command Name**
`az snapshot create`

**Errors:**
```
(ResourceNotFound) The Resource 'Microsoft.Compute/disks/datadiskvmss1_datadiskvmss1_0_OsDisk_1_dca81669e5884c35ad5150dfd854934e' under resource group 'DATADISK' was not found. For more details please go to https://aka.ms/ARMResourceNotFoundFix
Code: ResourceNotFound
Message: The Resource 'Microsoft.Compute/disks/datadiskvmss1_datadiskvmss1_0_OsDisk_1_dca81669e5884c35ad5150dfd854934e' under resource group 'DATADISK' was not found. For more details please go to https://aka.ms/ARMResourceNotFoundFix

```

## To Reproduce:

- `vmssosdisk=$(az vmss show -g {} -n {} --instance-id {} --query storageProfile.osDisk.managedDisk.id -o tsv) //this command DOES pull the correct resourceURI if you run it by itself without turning it into a variable
- `az snapshot create -g {} -n {} --source $vmssosdisk --debug`

## Expected Behavior
I am trying to replicate the steps to create a snapshot of a VMSS instance OS disk in CLI.  The PowerShell steps are listed here: https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-faq#how-do-i-take-a-snapshot-of-a-virtual-machine-scale-set-instance-

## Environment Summary
```
Linux-4.4.0-22000-Microsoft-x86_64-with-debian-bullseye-sid, Ubuntu 20.04.3 LTS
Python 3.6.10
Installer: DEB

azure-cli 2.30.0

Extensions:
interactive 0.4.5
maintenance 1.3.0
resource-graph 2.1.0
vm-repair 0.3.8

```

evelyn-ys:
**Description**

Fix #20258 
When creating a snapshot with a vmss os disk, the disk doesn't really exist. But we shouldn't throw `ResourceNotFound` error. We skip autodetect `copy_start` in this situation.

**Testing Guide**

```
az vmss create -g {} -n {} --image Centos
az vmss show -g {} -n {} --instance-id 1 --query storageProfile.osDisk.managedDisk.id
az snapshot create -g {} -n {} --source {vmssosdiskid}
```

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/vm/_validators.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_validators.py
@@ -1645,7 +1645,7 @@ def process_disk_or_snapshot_create_namespace(cmd, namespace):
                         source_info, _ = _get_disk_or_snapshot_info(cmd.cli_ctx,
                                                                     result['resource_group'],
                                                                     result['name'])
-                    except ResourceNotFoundError:
+                    except Exception:  # pylint: disable=broad-except
                         # There's a chance that the source doesn't exist, eg, vmss os disk.
                         # You can get the id of vmss os disk by
                         #   `az vmss show -g {} -n {} --instance-id {} --query storageProfile.osDisk.managedDisk.id`
--- a/src/azure-cli/azure/cli/command_modules/vm/_validators.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_validators.py
@@ -1641,7 +1641,15 @@ def process_disk_or_snapshot_create_namespace(cmd, namespace):
                 if not source_info:
                     from azure.cli.core.util import parse_proxy_resource_id
                     result = parse_proxy_resource_id(namespace.source_disk or namespace.source_snapshot)
-                    source_info, _ = _get_disk_or_snapshot_info(cmd.cli_ctx, result['resource_group'], result['name'])
+                    try:
+                        source_info, _ = _get_disk_or_snapshot_info(cmd.cli_ctx, result['resource_group'], result['name'])
+                    except ResourceNotFoundError:
+                        # There's a chance that the source doesn't exist, eg, vmss os disk.
+                        # You can get the id of vmss os disk by
+                        #   `az vmss show -g {} -n {} --instance-id {} --query storageProfile.osDisk.managedDisk.id`
+                        # But `az disk show --ids {}` will return ResourceNotFound error
+                        # We don't autodetect copy_start in this situation
+                        return
                 if not namespace.location:
                     get_default_location_from_resource_group(cmd, namespace)
                 # if the source location differs from target location, then it's copy_start scenario
--- a/src/azure-cli/azure/cli/command_modules/vm/_validators.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_validators.py
@@ -1642,7 +1642,9 @@ def process_disk_or_snapshot_create_namespace(cmd, namespace):
                     from azure.cli.core.util import parse_proxy_resource_id
                     result = parse_proxy_resource_id(namespace.source_disk or namespace.source_snapshot)
                     try:
-                        source_info, _ = _get_disk_or_snapshot_info(cmd.cli_ctx, result['resource_group'], result['name'])
+                        source_info, _ = _get_disk_or_snapshot_info(cmd.cli_ctx,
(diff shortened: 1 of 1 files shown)


src/azure-cli/azure/cli/command_modules/vm/_validators.py
```diff
@@ -1641,7 +1641,17 @@ def process_disk_or_snapshot_create_namespace(cmd, namespace):
                 if not source_info:
                     from azure.cli.core.util import parse_proxy_resource_id
                     result = parse_proxy_resource_id(namespace.source_disk or namespace.source_snapshot)
-                    source_info, _ = _get_disk_or_snapshot_info(cmd.cli_ctx, result['resource_group'], result['name'])
+                    try:
+                        source_info, _ = _get_disk_or_snapshot_info(cmd.cli_ctx,
+                                                                    result['resource_group'],
+                                                                    result['name'])
+                    except ResourceNotFoundError:
```
evelyn-ys, line 1648: Should we skip autodetection not only for `ResourceNotFoundError` but also other exceptions? @zhoxing-ms any suggestions?

zhoxing-ms: May I ask if there are other exceptions, can the subsequent process still do `copy_start` normally?

yonzhan:
Compute

jiasli:
> When creating a snapshot with a vmss os disk, the disk doesn't really exist.

Just of curiosity, why? If the disk doesn't exist, how can `az snapshot create` still work? Pardon me for not being an expert of VM.

jiasli:
Some scenario help can definitely prevent this from happening again in the future.

src/azure-cli/azure/cli/command_modules/vm/_validators.py
evelyn-ys: If there are other exceptions, we skip auto-detection for `copy_start`. Then the requests will use `copy` by default. We will leave service to decide whether to succeed or not. From my perspective, the auto-detection is a better-to-have thing. It shouldn't block creating...

evelyn-ys:
> > When creating a snapshot with a vmss os disk, the disk doesn't really exist.
> 
> Just of curiosity:
> 
> * Why?

If you try `az vmss show -g {} -n {} --query storageProfile.osDisk.managedDisk.id`, it returns nothing. Only if you specify `--instance-id`, a disk id will be returned which I think is service's predicted result.

> * If the disk doesn't exist, how can `az snapshot create` still work?

It used to work. Believe me, I'm just as surprised as you.

> * What happens if locations don't match?

Peer the original feature description https://github.com/Azure/azure-cli/issues/19773#issuecomment-940434211, if the locations don't match, we should set `createOption=CopyStart` instead of `createOption=Copy` smartly.

qqnarwhal:
It's because VMSS disks work differently from VM disks. The VMSS disks are platform managed and cannot be accessed by customer using the same disk IP for VM managed disks. This is why the validation step breaks the command for VMSS scenarios.

src/azure-cli/azure/cli/command_modules/vm/_validators.py
zhoxing-ms: OK, got it.
Since the automatic detection is better to have, and if there are other types of exceptions thrown, there will be the risk of blocking creating snapshots. I think we can expand to other exceptions to skip automatic detection.
So this way can be consistent with the original logic to the greatest extent, what do you thik of it?

evelyn-ys, line 1648:
```suggestion
                    except Exception:  # pylint: disable=broad-except
```

evelyn-ys: modified~

wangzelin007 (approved):

zhoxing-ms (approved):

PR #20308, opened 2021-11-12
[Profile] `az logout`, `az account clear`: Remove ADAL token cache file `accessTokens.json`

jiasli:
**Description**

After MSAL migration, we don’t actively delete ADAL token cache file `accessTokens.json` because we don’t want automations which rely on this file to be broken.

If `az logout` or `az account clear` is run explicitly, `accessTokens.json` will be deleted.

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/profile/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/profile/custom.py
@@ -3,6 +3,8 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 
+import os
+
 from knack.log import get_logger
 from knack.prompting import prompt_pass, NoTTYException
 from knack.util import CLIError
@@ -97,6 +99,8 @@ def set_active_subscription(cmd, subscription):
 
 def account_clear(cmd):
     """Clear all stored subscriptions. To clear individual, use 'logout'"""
+    _remove_adal_token_cache()
+
     if in_cloud_console():
         logger.warning(_CLOUD_CONSOLE_LOGOUT_WARNING)
     profile = Profile(cli_ctx=cmd.cli_ctx)
@@ -160,6 +164,8 @@ def login(cmd, username=None, password=None, service_principal=None, tenant=None
 
 def logout(cmd, username=None):
     """Log out to remove access to Azure subscriptions"""
+    _remove_adal_token_cache()
+
     if in_cloud_console():
         logger.warning(_CLOUD_CONSOLE_LOGOUT_WARNING)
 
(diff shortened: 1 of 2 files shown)


src/azure-cli/azure/cli/command_modules/profile/custom.py
```diff
@@ -204,3 +210,18 @@ def check_cli(cmd):
...
+def _remove_adal_token_cache():
+    """Remove ADAL token cache file ~/.azure/accessTokens.json, as it is no longer needed by MSAL-based Azure CLI.
+    """
+    from azure.cli.core._environment import get_config_dir
+    adal_token_cache = os.path.join(get_config_dir(), 'accessTokens.json')
+    try:
+        os.remove(adal_token_cache)
+        logger.warning("ADAL token cache '%s' has been deleted. `az login` no longer generates this file. "
+                       "For more details, see https://docs.microsoft.com/cli/azure/msal-based-azure-cli",
+                       adal_token_cache)
```
rayluo, line 224: Wouldn't such a warning too confusing and scary to normal users? I think it would be fine to just remove it silently. But it is up to you. I'm still approving this PR.

rayluo (approved):

src/azure-cli/azure/cli/command_modules/profile/custom.py
jiasli: @chasewilson, thoughts?

chasewilson: Hmm, would it be a good idea to use logging.info? This would at least provide the information if someone was debugging or using verbose to get more info and for customers using logging would provide a record and reason for the deletion?

rayluo: > Hmm, would it be a good idea to use logging.info? This would at least provide the information if someone was debugging or using verbose to get more info and for customers with logging enabled would provide a record and reason for the deletion?

Makes sense. Could also consider `logging.debug`, then. I am fine either way.

yonzhan:
Profile

yonzhan (approved):

src/azure-cli/azure/cli/command_modules/profile/custom.py
jiasli: This is a one-time execution. After that, this warning won't be shown anymore, so it will get lost forever! 😆

People usually only use `--verbose` and `--debug` after hitting a failure.

rayluo: > This is a one-time execution. After that, this warning won't be shown anymore, so it will get lost forever! 😆
> 
> People usually only use `--verbose` and `--debug` after hitting a failure.

Personally, if I run a product and it emits an unfamiliar but seemingly scary message (all warnings are scary), I would be alerted and then try to repro and see if I need to report back. But then, if my repro attempt can not reproduce that warning, I would be very confuse. This feeling of unpredictability would decrease my trust to the product. (For what it's worth, this was the same thought process when I was testing a new security camera that I bought. Now I'm considering to return it. :-) )

In short, let's at least make it an INFO or DEBUG, not a WARNING. After all, it is an almost expected behavior during `az logout` or `az account clean`, so, there is nothing serious need to be warned.

jamesc: We're adding a new concept here (ADAL) that most customers won't know about.  Maybe saying "Legacy ADAL Token Cache has been cleaned up" to show it's normal and something they probably won't need to act on. If they're actually using it directly, they'll know what we're talking about ?

jiasli: > We're adding a new concept here (ADAL) that most customers won't know about.

This makes sense. Maybe let's just remove the log and save us and users from the fuss.

evelyn-ys (approved):

Issue #17399, opened 2021-03-22
az keyvault certificate list --id


kaletkamk:
> ### `az feedback` auto-generates most of the information requested below, as of CLI version 2.0.62

**Describe the bug**
According to the documentation az keyvault certificate list --id (kv_id) should list all of the certificates. Other parameters should not be required when using --id 
But when you run  az keyvault certificate list --id (kv_id) you get response 'vaultBaseUrl'

```
Marcin@XPS  D:                                                                                               [15:00]
❯ az keyvault certificate list --id /subscriptions/7a77d45b-dbdc-4831-b3d9-e5a9c23fc1af/resourceGroups/AZ-RG-CS-APGv2-NA-Prod-01/providers/Microsoft.KeyVault/vaults/AZ-KV-CS-Cloud-EASTUS-2
'vaultBaseUrl'
Marcin@XPS  D:                                                                                               [15:01]
❯
```
**To Reproduce**
az keyvault certificate list --id (kv_id) 
**Expected behavior**
List of certificates
**Environment summary**
azure-cli                         2.20.0

core                              2.20.0
telemetry                          1.0.6

Extensions:
azure-devops                      0.18.0
azure-firewall                     0.9.0

Python location 'C:\Program Files (x86)\Microsoft SDKs\Azure\CLI2\python.exe'
Extensions directory 'C:\Users\Marcin\.azure\cliextensions'

Python (Windows) 3.6.8 (tags/v3.6.8:3c6b436a57, Dec 23 2018, 23:31:17) [MSC v.1916 32 bit (Intel)]

Legal docs and information: aka.ms/AzureCliLegal

Your CLI is up-to-date.

Please let us know how we are doing: https://aka.ms/azureclihats
and let us know if you're interested in trying out our newest features: https://aka.ms/CLIUXstudy

**Additional context**

yungezz:
hi @houk-ms,  could you pls have a look?

chwarr:
I'm seeing the same behavior in 2.30.0 on Windows.

```text
PS> az keyvault certificate list --id /subscriptions/redacted/resourceGroups/redacted/providers/Microsoft.KeyVault/vaults/redacted  --output json
'vaultBaseUrl'
PS> $LASTEXITCODE
1
```

Here's my  environment summary:

```
Windows-10-10.0.22000-SP0
Python 3.8.9
Installer: MSI

azure-cli 2.30.0

Extensions:
azure-devops 0.18.0

```

[...]

PR #20369, opened 2021-11-16
[Key Vault] az keyvault: Revise --id argument help text

chwarr:
**Description**

For commands that use the `--id` argument to identify the vault/HSM, the
vault/HSM's full URI, like `https://vault-name.vault.azure.net/`, needs
to be passed, not the resource's ARM ID, like
`/subscriptions/.../resourceGroups/.../providers/Microsoft.KeyVault/vaults/vault-name`.

Refine the argument help text to use "Full URI" instead of "Id" to try
to minimize confusion.

**Testing Guide**

This only changes help text. No functionality should be affected.

Fixes https://github.com/Azure/azure-cli/issues/17399

---

   - The existing commands are not following the "-id should be a GUID" guideline. I did not change that.

   - Help text only change.

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/keyvault/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/keyvault/_params.py
@@ -287,7 +287,7 @@ def load_arguments(self, _):
                 try:
                     if cmd in ['list']:
                         c.extra('identifier', options_list=['--id'],
-                                help='Id of the Vault or HSM. '
+                                help='Full URI of the Vault or HSM. '
                                      'If specified all other \'Id\' arguments should be omitted.',
                                 validator=validate_vault_or_hsm)
                     else:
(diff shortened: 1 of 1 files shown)


chwarr:
Based on the discussion in [issue #17399, "az keyvault certificate list --id"](https://github.com/Azure/azure-cli/issues/17399), it appears that operations on the vault/HSM that have an `--id` parameter take the vault/HSM's full URI, not its ARM id. Let me know if any of the commands I've changed do not follow that pattern, and I can update.

yonzhan:
Key Vault

evelyn-ys (approved):

src/azure-cli/azure/cli/command_modules/keyvault/_params.py
```diff
@@ -287,7 +287,7 @@ class CLISecurityDomainOperation(str, Enum):
                 try:
                     if cmd in ['list']:
                         c.extra('identifier', options_list=['--id'],
-                                help='Id of the Vault or HSM. '
+                                help='Full URI of the Vault or HSM. '
```
jiasli, line 290: Personally I think it is anti-convention to name it "URI" in Azure Portal (https://github.com/Azure/azure-cli/issues/17399#issuecomment-969876246), even though it has `https://` prefix.

This was discussed previously at #13067. 

Also see https://danielmiessler.com/study/difference-between-uri-url/

> If the protocol (https, ftp, etc.) is either present or implied for a domain, you should call it a URL—even though it’s also a URI.

jiasli (approved):

Issue #20429, opened 2021-11-19
Refine or deprecate `--refresh` in `az account list`


jiasli:
## Context

`--refresh` in `az account list` was introduced by #4091 on 2017-07-29 and was never maintained ever since then.

It doesn't work with managed identity and Cloud Shell account:

```
$ az login --identity
...
$ az account list --refresh
Refreshing for 'systemAssignedIdentity' failed with an error 'No matched service principal found'. The existing accounts were not modified. You can run 'az login' later to explicitly refresh them
...
```

It adds a lot of complexity given the same effect can be achieved by

```
az account clear
az login
```

The usage of `--refresh` is not very popular. Here is the client telemetry for the last 7 days:

| | invocations | subscriptions |
| - | - | - |
| `az account list --refresh` | 2534 | 287 |
| all commands | 1156167962 | 73166 |

```
RawEventsAzCli
| where EventTimestamp > ago(7d) and EventTimestamp < ago(1h)
| extend exeVersion = tostring(Properties['context.default.vs.core.exeversion'])
| extend clientRequestId = tostring(Properties['context.default.azurecli.clientrequestid'])
| extend result = tostring(Properties['reserved.datamodel.action.result'])
| extend commandName = tostring(Properties['reserved.datamodel.entityname'])
| extend CommandRaw = tostring(Properties['context.default.azurecli.rawcommand'])
| extend paramSet = tostring(Properties['context.default.azurecli.params'])
| extend FaultTypeString = tostring(Properties['reserved.datamodel.fault.typestring'])
| extend FaultMessage = tostring(Properties['reserved.datamodel.fault.exception.message'])
| where paramSet contains '--refresh'
| summarize count(), dcount(AzureSubscriptionId)
```

## Proposals

`--refresh` should either be

- Refined to support managed identity and Cloud Shell account
- Deprecated and removed

## Also see

- https://github.com/Azure/azure-cli/issues/17607
- https://github.com/Azure/azure-cli/issues/20402

yonzhan:
Refine or deprecate --refresh

PR #19853, opened 2021-10-12
[Core] ADAL to MSAL migration

jiasli:
Close #18944

## Description

As [ADAL has been deprecated](https://github.com/AzureAD/azure-activedirectory-library-for-python), Azure CLI needs to be migrated from ADAL to MSAL.

## How CLI authentication works

During initialization:

```
             _get_mgmt_service_client 
             /                      \
          1 /    CredentialAdaptor   \ 2
           v                          v
get_login_credential                SDK client
```

1. The client factory `azure.cli.core.commands.client_factory._get_mgmt_service_client` calls `azure.cli.core._profile.Profile.get_login_credentials` to get an instance of `CredentialAdaptor`.
2. `_get_mgmt_service_client` feeds the `CredentialAdaptor` instance into the SDK client.

During invocation:

```
           Track 1/2 SDK client
                    |
                    v     signed_session/get_token
            CredentialAdaptor
                    |    
                    v     get_token
UserCredential / ServicePrincipalCredential    
```

### `CredentialAdaptor`

1. When Track 2 SDK client makes an HTTP call, `azure.core.pipeline.policies._authentication.BearerTokenCredentialPolicy` calls the `get_token` method on the `CredentialAdaptor`.
2. When Track 1 SDK client makes an HTTP call, `msrest.pipeline.requests.RequestsCredentialsPolicy` calls the `signed_session` method on the `CredentialAdaptor`, `signed_session` calls `get_token`.

### `UserCredential`/`ServicePrincipalCredential`

Since Azure Identity has been dropped for User and Service Principal authentication, the MSAL integration is implemented in `src/azure-cli-core/azure/cli/core/auth/msal_authentication.py`.

There are 2 kinds of credentials:

1. `UserCredential` (inherit from `PublicClientApplication`)
2. `ServicePrincipalCredential` (inherit from `ConfidentialClientApplication`)

They all expose `get_token` so that they implement `azure.core.credentials.TokenCredential`.

## Cross-tenant authentication

Cross-tenant authentication is implemented by simply adding header `x-ms-authorization-auxiliary` by `_prepare_mgmt_client_kwargs_track2`.

There were asks (https://github.com/Azure/azure-sdk-for-python/issues/8313, https://github.com/Azure/azure-sdk-for-python/issues/17764) on Python SDK to implement cross-tenant authentication but it hasn't been implemented. 

[...]

PR #20430, opened 2021-11-19
{Profile} `az account list`: Fix: `--refresh` returns `user` as `null`

### LINKED ISSUE #20402: Service principal context left in inconsistent state after `az account list --refresh`
Repro steps:

```sh
# use a clean configuration directory
export AZURE_CONFIG_DIR="$(pwd)/foo"

az login --service-principal --username ...

# at this point, everything should work; for instance, az keyvault list should return whatever keyvaults your service principal can read
az keyvault list

# refreshing the account list leaves the account configuration in an inconsistent state.
az account list --refresh

# this command fails and prints "Could not retrieve credential from local cache for service principal None. Run `az login` for this service principal."
az keyvault list

# if you look in $AZURE_CONFIG_DIR/azureProfile.json, you will see that the "user" field of every subscription object is null now
```

It seems like a regression in 2.30.0, likely due to #19707

jiasli:
- Fix #20402

## Description

After MSAL migration (https://github.com/Azure/azure-cli/pull/19853), `SubscriptionFinder` no longer maintains `user_id`, so `refresh_accounts` should use existing `user_id` instead of reading from `SubscriptionFinder`.

## Testing Guide

```PowerShell
# PowerShell
# Use a clean config dir if you want
$env:AZURE_CONFIG_DIR='D:\cli\azure2'
az login
az account list --refresh
```

## Also see

- https://github.com/Azure/azure-cli/issues/20429

### DIFF
--- a/src/azure-cli-core/azure/cli/core/_profile.py
+++ b/src/azure-cli-core/azure/cli/core/_profile.py
@@ -633,7 +633,7 @@ class Profile:
                 if not subscriptions:
                     continue
 
-            consolidated = self._normalize_properties(subscription_finder.user_id,
+            consolidated = self._normalize_properties(user_name,
                                                       subscriptions,
                                                       is_service_principal)
             result += consolidated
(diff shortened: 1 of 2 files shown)


yonzhan:
Profile

calvinhzy (approved):

evelyn-ys (approved):

PR #20432, opened 2021-11-19
{Auth} Rename option `token_encryption` to `encrypt_token_cache`

jiasli:
**Description**

Rename option `token_encryption` to `encrypt_token_cache` per the internal discussion email:

- _Naming for the option to disable token encryption_

> While we do store Service Principal credentials, we call the location the **Token Cache** and this would also give an added benefit of an accurate description of what the config is doing. Encrypting the file and not just a token?

⚠ This option is still experimental and subject to change until it is announced at https://docs.microsoft.com/en-us/cli/azure/azure-cli-configuration


yonzhan:
Auth

yonzhan (approved):

chasewilson (approved):

evelyn-ys (approved):

Issue #18528, opened 2021-06-17
az aks create|update --attach-acr makes undue graph api calls that results in `Could not create a role assignment for ACR. Are you an Owner on this subscription?`


duckie:
**Describe the bug**

Running with a managed identity which is **Contributor** on the resource group, **Owner** on the ACR repo, and a cluster with managed identities enabled:

```
az aks update -g $RESOURCE_GROUP -n $CLUSTER_NAME --attach-acr $ACRNAME
```

Results in `Could not create a role assignment for ACR. Are you an Owner on this subscription?`. Though this is [documented as an expected behavior](https://github.com/MicrosoftDocs/azure-docs/issues/64083) of ACR, it is not. It is a bug of Azure CLI that makes a useless call to Graph API to get an information it already has.

Can be reproduced at creation too.

**To Reproduce**

Follow the information given before.

**Expected behavior**

The call should succeed in giving the role assignment to the cluster. Proof by workaround, since the expected result can be achieved by doing this:

```
role_id=$(az role definition list \
  --scope "/subscriptions/$SUBSCRIPTION_ID/resourceGroups/$RESOURCE_GROUP/providers/Microsoft.ContainerRegistry/registries/$ACRNAME" \
  --name AcrPull \
  --query "[0].id" -o tsv)

object_id=$(az aks show \
  -g $RESOURCE_GROUP \
  -n $CLUSTER_NAME \
  --query "identityProfile.kubeletidentity.objectId" -o tsv)

az role assignment create \
  --scope "/subscriptions/$SUBSCRIPTION_ID/resourceGroups/$RESOURCE_GROUP/providers/Microsoft.ContainerRegistry/registries/$ACRNAME" \
  --role "$role_id" \
  --assignee-object-id "$object_id" \
  --assignee-principal-type ServicePrincipal
```

Which is a direct proof that the limitation is on Azure CLI  side and not in ACR product.

**Environment summary**

`azure-cli` version `2.25.0-1` installed with official rpm. 
Running in a machine with image `OpenLogic:CentOS:7_8-gen2`.

**Additional context**

This happens because the code path that makes the role assignment do not care whether the identity is a Service Principal or a Managed Identity. 

We can see it here in `command_modules/acs/custom.py`:

```
def _ensure_aks_acr_role_assignment(cli_ctx,                                                                                                                   
                                    client_id,                                                                                                                 
                                    registry_id,                                                                                                               
                                    detach=False):                                                                                                             
[...]

PR #20477, opened 2021-11-23
{AKS} Avoid unnecessary graph API call and refine error message

norshtein:
**Description**

Two things:
1. For MSI clusters, when attaching ACR, the operation "resolving client ID to object ID" is not necessary. We can directly get the object ID from managed cluster properties. This PR removes this unnecessary operation.
2. Refines the error message when resolving object ID hits error.
**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acs/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/custom.py
@@ -553,8 +553,10 @@ def _add_role_assignment(cmd, role, service_principal_msi_id, is_service_princip
             if ex.message == 'The role assignment already exists.':
                 break
             logger.info(ex.message)
-        except:  # pylint: disable=bare-except
-            pass
+        except CLIError as ex:
+            logger.warning(str(ex))
+        except Exception as ex:  # pylint: disable=broad-except
+            logger.error(str(ex))
         time.sleep(delay + delay * x)
     else:
         return False
@@ -564,7 +566,7 @@ def _add_role_assignment(cmd, role, service_principal_msi_id, is_service_princip
 
 
 def delete_role_assignments(cli_ctx, ids=None, assignee=None, role=None, resource_group_name=None,
-                            scope=None, include_inherited=False, yes=None):
+                            scope=None, include_inherited=False, yes=None, is_service_principal=True):
     factory = get_auth_management_client(cli_ctx, scope)
     assignments_client = factory.role_assignments
     definitions_client = factory.role_definitions
@@ -585,14 +587,14 @@ def delete_role_assignments(cli_ctx, ids=None, assignee=None, role=None, resourc
                               assignments_client.config.subscription_id)
     assignments = _search_role_assignments(cli_ctx, assignments_client, definitions_client,
                                            scope, assignee, role, include_inherited,
-                                           include_groups=False)
+                                           include_groups=False, is_service_principal=is_service_principal)
 
     if assignments:
         for a in assignments:
             assignments_client.delete_by_id(a.id)
 
 
-def _delete_role_assignments(cli_ctx, role, service_principal, delay=2, scope=None):
+def _delete_role_assignments(cli_ctx, role, service_principal, delay=2, scope=None, is_service_principal=True):
     # AAD can have delays in propagating data, so sleep and retry
     hook = cli_ctx.get_progress_controller(True)
     hook.add(message='Waiting for AAD role to delete', value=0, total_val=1.0)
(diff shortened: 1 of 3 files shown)


yonzhan:
acs

wangzelin007:
@norshtein
I have changed title, `{}` means this change is not customer-facing and the message will **NOT** be included in `HISTORY.rst`.

wangzelin007 (approved):

norshtein:
Fixes #18528

@wangzelin007 thanks for reviewing! @FumingZhang @andyliuliming Could you also take a look and help merge the PR? Thanks!

src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -557,8 +557,8 @@ def _add_role_assignment(cmd, role, service_principal_msi_id, is_service_princip
             if ex.message == 'The role assignment already exists.':
                 break
             logger.info(ex.message)
-        except:  # pylint: disable=bare-except
-            pass
+        except Exception as ex:  # pylint: disable=broad-except
+            logger.warning(str(ex))
```
FumingZhang, line 561: Similarly, with the following changes, we can distinguish between cli errors and python errors by log level (warning/error).

```suggestion
        except CLIError as ex:
            logger.warning(str(ex))
        except Exception as ex:  # pylint: disable=broad-except
            logger.error(str(ex))
```

src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -1475,8 +1480,13 @@ def _create_role_assignment(cmd, role, assignee,
...
     # if not use MSI object id.
-    object_id = _resolve_object_id(
-        cmd.cli_ctx, assignee) if resolve_assignee else assignee
+    object_id = assignee
+    if resolve_assignee:
+        try:
+            object_id = _resolve_object_id(cmd.cli_ctx, assignee)
+        except Exception as ex:  # pylint: disable=bare-except
+            msg = "Failed to resolve service principal object ID: " + str(ex)
+            raise CLIError(msg)
```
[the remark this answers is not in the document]
FumingZhang: A screenshot of an example is as follows
Screenshot 2021-11-24 140718

norshtein:
@FumingZhang many thanks for the suggestion! I committed all.

FumingZhang (approved):
LGTM, invite @andyliuliming to take another look

norshtein:
@wangzelin007 @FumingZhang @andyliuliming could you please help merge the PR?

src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -1475,8 +1482,30 @@ def _create_role_assignment(cmd, role, assignee,
...
+                    if internal_error:
+                        internal_error += " - "
+                    internal_error += str(error_reason)
+                if internal_error:
+                    error_msg += "({}) ".format(internal_error)
+            error_msg += ex.message
+            # this should be UserFault or ServiceError, but it is meaningless to distinguish them here
+            raise CLIError(error_msg)
+        except Exception as ex:  # pylint: disable=bare-except
+            raise CLIError(error_msg + str(ex))
```
zhoxing-ms, line 1508: Could you use the specific error type instead of `CLIError`? For more details, please refer to this doc [error_handling_guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/error_handling_guidelines.md)

norshtein: @FumingZhang helped contribute this part of code. I think the comment in L1505 explains why we don't distinguish the error category.

zhoxing-ms: >  this should be UserFault or ServiceError, but it is meaningless to distinguish them here

@FumingZhang Why is it unnecessary to distinguish between `UserFault` and `ServiceError` here?

FumingZhang: We are calling graph client here and this may return a 3xx or 4xx error (maybe 5xx also). We intend to display this error message in the form of warning, so we use `except CLIError` to catch this type of exception [where](https://github.com/Azure/azure-cli/pull/20477/files#diff-f1fc443c5eb2ddee110be5f06110930285573e1debf2640ab83dead88943a9a5R560) it calls this function.

I hope that the type of error can be better described to users, so in the previous suggestion I parsed the exception object raised by graph client and added the error code and reason to the error message. Here is a screenshot of an example of this. Screenshot 2021-11-24 140718

I respect that we should use more accurate error types to describe errors, but in this scenario, I suppose using error codes to distinguish error types will not provide users with more information.

zhoxing-ms: Got it, if this error is not raised out, but is only used to print logs, there is no need to subdivide the error type

norshtein:
@FumingZhang @zhoxing-ms I've rebased the code based on latest dev branch. I think the PR is ready to be merged. Could you please have a look?

zhoxing-ms (approved):

PR #20491, opened 2021-11-24
{App Config} Fix #12668860, #12668922: Fix validations while importing kvset

pratiksanglikar:
**Description**
* Fixes [#12668922](https://msazure.visualstudio.com/Azure%20AppConfig/_workitems/edit/12668922/) Added validations to the command `az appconfig import` when `appconfig/kvset` profile is used and tags are not in string format.
* Fixes [#12668860](https://msazure.visualstudio.com/Azure%20AppConfig/_workitems/edit/12668860/) Handled cases where optional parameters were present in the import file for `az appconfig import` command and `appconfig/kvset` profile was used for import.

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
+++ b/src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
@@ -1107,7 +1107,7 @@ def __validate_import_tags(kv):
     if kv.tags:
         for tag_key, tag_value in kv.tags.items():
             if not isinstance(tag_value, str):
-                logger.warning("The value for the tag '{0}' for key '{1}' is not a string. It will be converted to a string.".format(tag_key, kv.key))
+                logger.warning("The value for the tag '{%s}' for key '{%s}' is not a string. It will be converted to a string.", tag_key, kv.key)
                 return
 
 
--- a/src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
+++ b/src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
@@ -1112,11 +1112,14 @@ def __validate_import_config_setting(config_setting):
         return False
 
     if config_setting.value and not isinstance(config_setting.value, str):
-        logger.warning("The 'value' for the key '{%s}' is not a string. It will be converted to a string.", config_setting.key)
+        logger.warning("The 'value' for the key '{%s}' is not a string. This key-value will not be imported.", config_setting.key)
+        return False
     if config_setting.content_type and not isinstance(config_setting.content_type, str):
-        logger.warning("The 'content_type' for the key '{%s}' is not a string. It will be converted to a string.", config_setting.key)
+        logger.warning("The 'content_type' for the key '{%s}' is not a string. This key-value will not be imported.", config_setting.key)
+        return False
     if config_setting.label and not isinstance(config_setting.label, str):
-        logger.warning("The 'label' for the key '{%s}' is not a string. It will be converted to a string.", config_setting.key)
+        logger.warning("The 'label' for the key '{%s}' is not a string. This key-value will not be imported.", config_setting.key)
+        return False
 
     return __validate_import_tags(config_setting)
 
@@ -1129,7 +1132,8 @@ def __validate_import_tags(kv):
     if kv.tags:
         for tag_key, tag_value in kv.tags.items():
             if not isinstance(tag_value, str):
-                logger.warning("The value for the tag '{%s}' for key '{%s}' is not a string. It will be converted to a string.", tag_key, kv.key)
+                logger.warning("The value for the tag '{%s}' for key '{%s}' is not in a valid format. This key-value will not be imported.", tag_key, kv.key)
+                return False
     return True
 
 
--- a/src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
+++ b/src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
@@ -1123,7 +1123,7 @@ def __validate_import_config_setting(config_setting):
 
 def __validate_import_tags(kv):
     if kv.tags and not isinstance(kv.tags, dict):
-        logger.warning("The value of the 'tags' for key '{%s}' is not an object. This key-value will not be imported.", kv.key)
+        logger.warning("The format of 'tags' for key '%s' is not valid. This key-value will not be imported.", kv.key)
         return False
 
     if kv.tags:
--- a/src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
+++ b/src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
@@ -106,7 +106,6 @@ def validate_import_key(key):
     else:
         logger.warning("Ignoring invalid key ''. Key cannot be empty.")
         return False
-
     return True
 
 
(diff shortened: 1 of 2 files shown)


pratiksanglikar:
@avanigupta, @shenmuxiaosen please take a look!

yonzhan:
App Config

wangzelin007 (approved):

wangzelin007:
@pratiksanglikar 
Please fix the CLI Style error, you can run `azdev style appconfig` to check the error in your local environment.
And add @avanigupta Avani Gupta FTE, @shenmuxiaosen Shuai Wang FTE to **Reviewers**.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -1027,11 +1026,11 @@ def __import_kvset_from_file(client, path, yes):
     if KVSetConstants.KVSETRootElementName not in new_kvset:
         raise FileOperationError("file '{0}' is not in a valid '{1}' format.".format(path, ImportExportProfiles.KVSET))
 
-    kvset_to_import = [ConfigurationSetting(key=kv['key'],
-                                            label=kv['label'],
-                                            content_type=kv['content_type'],
-                                            value=kv['value'],
-                                            tags=kv['tags'])
+    kvset_to_import = [ConfigurationSetting(key=kv['key'] if 'key' in kv else None,
```
shenmuxiaosen, line 1029: >None [](http://example.com/codeflow?start=78&length=4)

Key should never be null.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -1027,11 +1026,11 @@ def __import_kvset_from_file(client, path, yes):
...
+                                            label=kv['label'] if 'label' in kv else None,
+                                            content_type=kv['content_type'] if 'content_type' in kv else None,
+                                            value=kv['value'] if 'value' in kv else None,
+                                            tags=kv['tags'] if 'tags' in kv else None)
                        for kv in new_kvset[KVSetConstants.KVSETRootElementName]]
```
shenmuxiaosen, line 1034: What if kv['tags'] is not a dict?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -1064,8 +1063,9 @@ def __import_kvset_from_file(client, path, yes):
                 continue
         elif not validate_import_key(config_setting.key):
             continue
-
+        __validate_import_tags(config_setting)
         # All validations successful
```
shenmuxiaosen, line 1067: nit: can we add an empty line after the for loop.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
pratiksanglikar: We have that validation down in `validate_import_key` function with a proper error message.
This default value is here just so that the command won't fail with not so helpful error message.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
pratiksanglikar: That's a good point, we don't currently validate the type of these attributes, but we should.

pratiksanglikar: done!

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -1065,8 +1067,9 @@ def __import_kvset_from_file(client, path, yes):
         elif not validate_import_key(config_setting.key):
             continue
```
shenmuxiaosen, line 1068: This part is validating the key. Can we move those checks into __validate_import_config_setting, where label, value, content_type are validated.  Also it should be done before printing preview.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -1103,6 +1106,29 @@ def __validate_import_feature_flag(kv):
...
+        logger.warning("The 'label' for the key '{%s}' is not a string. It will be converted to a string.", config_setting.key)
+
+    return __validate_import_tags(config_setting)
+
+
+def __validate_import_tags(kv):
+    if kv.tags and not isinstance(kv.tags, dict):
+        logger.warning("The value of the 'tags' for key '{%s}' is not an object. This key-value will not be imported.", kv.key)
+        return False
+
```
shenmuxiaosen, line 1124: This looks inconsistent. Why tags format is special that can skip this import? I guess if the format of any properties is wrong, we skip this key value set.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -1103,6 +1106,29 @@ def __validate_import_feature_flag(kv):
...
+        logger.warning("The 'content_type' for the key '{%s}' is not a string. It will be converted to a string.", config_setting.key)
+    if config_setting.label and not isinstance(config_setting.label, str):
+        logger.warning("The 'label' for the key '{%s}' is not a string. It will be converted to a string.", config_setting.key)
+
+    return __validate_import_tags(config_setting)
+
+
+def __validate_import_tags(kv):
+    if kv.tags and not isinstance(kv.tags, dict):
+        logger.warning("The value of the 'tags' for key '{%s}' is not an object. This key-value will not be imported.", kv.key)
```
shenmuxiaosen, line 1122: >object [](http://example.com/codeflow?start=73&length=6)

is not an dictionary?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
shenmuxiaosen: use dict.get(key[, default]) to assign default values.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
pratiksanglikar: The difference is that some errors are correctable vs this is not.
e.g. We expect a value to be a `string`, but if the JSON has it in numeric format, i.e. we expect `value: "56"` but if JSON has `"value": 56`, we can correct and just log the error.
This one can't be corrected since we can't convert a random datatype to a `dict`.

In other places we use the same format of logging a warning and continuing with the best effort, so I think this strategy aligns here.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
pratiksanglikar: In my opinion, `dictionary` is an implementation detail specific to Python that we shouldn't use in a warning message. 

We are importing from a JSON file and we expect a JSON object as a value.
I could potentially make it more generic to cater to cases of importing from other sources, but I'm out of ideas here.
Any suggestions?

shenmuxiaosen: How about "The format of tags for key .... is not valid?".

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
shenmuxiaosen: It is not. Nothing is correctable. We can't simply convert it becasue we can convert it. Json value has type and we need to perserve it. Otherwise, what if the content type is application/json.  The value itself can be number or string. Others like label, content type, tags, we will expect string and dict, which are what we export.

zhoxing-ms (approved):

zhoxing-ms:
@pratiksanglikar Could you please address those comments from @shenmuxiaosen 
Please note that if you can't address these comments today, this PR will not catch up with the release date of this sprint and can only be postponed to the next sprint

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
pratiksanglikar: Moved the logic to the `__validate_import_config_setting` and it now validates before print preview.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
pratiksanglikar: As we discussed offline, failing now when we find datatype other than string.

shenmuxiaosen (approved):
:shipit:

pratiksanglikar:
@zhoxing-ms, The PR has been signed off.
Can you please include this PR in today's release?
cc: @wangzelin007

wangzelin007:
> @zhoxing-ms Xing Zhou FTE, The PR has been signed off. Can you please include this PR in today's release? cc: @wangzelin007 Zelin Wang FTE

DONE

PR #20536, opened 2021-11-29
{App Service} az appservice ase: Fix issue with resource group in new SDK

madsd:
**Description**
Change resource group to be fetched from the resource id of the ASE. The issue occurs if you call commands without explicitly defining the resource group.

**Testing Guide**
az appservice ase list-addresses -n <ase-name>

**History Notes**

---


yonzhan:
App Service

wangzelin007 (approved):

panchagnula:
@madsd like we discussed offline please create a task/issue for ASE to team to update the swagger to include the missing property so that the next SDK drop will have this, instead of CLI/client making the changes to handle this.
This can be a temporary work around for CLI, property missing on SDK/swagger would be a breaking change & if this is not expected, this should be fixed at the source correctly. Thanks!

zhoxing-ms (approved):

panchagnula (approved):
LGTM

PR #20574, opened 2021-12-01
{Packaging} Adapt entry script to Mariner

jiasli:
## Description

See code comments.

## References

- https://github.com/Azure/azure-cli/pull/17491
- https://github.com/Azure/azure-cli/pull/20061

### DIFF
--- a/scripts/release/rpm/azure-cli.spec
+++ b/scripts/release/rpm/azure-cli.spec
@@ -51,14 +51,23 @@ deactivate
 # Fix up %{buildroot} appearing in some files...
 for d in %{buildroot}%{cli_lib_dir}/bin/*; do perl -p -i -e "s#%{buildroot}##g" $d; done;
 
-# Create executable
+# Create executable (entry script)
+
+# For PYTHONPATH:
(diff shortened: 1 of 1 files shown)


jiasli:
A very valuable post: https://stackoverflow.com/questions/3113472/how-to-make-an-rpm-spec-that-installs-libraries-to-usr-lib-xor-usr-lib64-based

Thanks @chalamalasetty

yonzhan:
Packaging

mfrw:
> A very valuable post: https://stackoverflow.com/questions/3113472/how-to-make-an-rpm-spec-that-installs-libraries-to-usr-lib-xor-usr-lib64-based. Thanks @chalamalasetty.
> 
> According to this, if we download a package from https://packages.microsoft.com/yumrepos/azure-cli/ and unzip it, we can see `lib64` is already hardcoded when packaging:
> 
> image
> 
> image

This looks like a `win-win` :)

wangzelin007 (approved):

mfrw (approved):
LGTM.
Thank you @jiasli

Issue #20589, opened 2021-12-02
Empty output for keyvault default policy when using table or tsv


eiffel-fl:
Hi.

I found a bug in azure-cli where default policy is not printed when using table or tsv as output mode:

Here is how to reproduce:

```bash
$ az keyvault certificate get-default-policy -o table

$ az keyvault certificate get-default-policy -o tsv  
                1
```

Best regards.

jiasli:
Thanks for the contribution, but I am not sure if forcing JSON output is a good idea. @evelyn-ys to help take a look.

PR #20590, opened 2021-12-02
{keyvault} Fix empty default policy when using table or tsv

eiffel-fl:
Hi.

This PR fixes bug described in #20589 where output is empty when using table or tsv as output mode.

So, in this PR, I detected if, for this specific subcommand, output mode is either table or tsv and force set json instead.

You can test it using the following (be sure to be logged to azure before):

```bash
$ ./bin/python3 ./src/azure-cli/az keyvault certificate get-default-policy -o table
{
  "issuerParameters": {
    "certificateTransparency": null,
    "name": "Self"
  },
  "keyProperties": {
    "curve": null,
    "exportable": true,
    "keySize": 2048,
    "keyType": "RSA",
    "reuseKey": true
  },
  "lifetimeActions": [
    {
      "action": {
        "actionType": "AutoRenew"
      },
      "trigger": {
        "daysBeforeExpiry": 90
      }
    }
  ],
  "secretProperties": {
    "contentType": "application/x-pkcs12"
  },
  "x509CertificateProperties": {
    "keyUsage": [
      "cRLSign",
      "dataEncipherment",
      "digitalSignature",
      "keyEncipherment",
      "keyAgreement",
      "keyCertSign"
    ],
    "subject": "CN=CLIGetDefaultPolicy",
    "validityInMonths": 12
  }
}
$ ./bin/python3 ./src/azure-cli/az keyvault certificate get-default-policy -o tsv
# Same output than above.
```

Best regards.

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/keyvault/_transformers.py
+++ b/src/azure-cli/azure/cli/command_modules/keyvault/_transformers.py
@@ -4,6 +4,7 @@
 # --------------------------------------------------------------------------------------------
 from collections.abc import Iterable
 from azure.cli.command_modules.keyvault._validators import KeyEncryptionDataType
+from azure.cli.core._output import (get_output_format, set_output_format)
 
 
 def multi_transformers(*transformers):
@@ -19,6 +20,20 @@ def keep_max_results(output, **command_args):
     if maxresults:
         return list(output)[:maxresults]
     return output
+
+
+def force_json_output(result, **command_args):
+    if 'cmd' not in command_args:
+        return result
+
+    ctx = command_args['cmd'].cli_ctx
+
+    output_format = get_output_format(ctx)
+
+    if output_format == 'table' or output_format == 'tsv':
+        set_output_format(ctx, 'json')
+
+    return result
 
 
 def filter_out_managed_resources(output, **command_args):  # pylint: disable=unused-argument
--- a/src/azure-cli/azure/cli/command_modules/keyvault/commands.py
+++ b/src/azure-cli/azure/cli/command_modules/keyvault/commands.py
@@ -12,7 +12,7 @@ from azure.cli.command_modules.keyvault._client_factory import (
     get_client, get_client_factory, Clients, is_azure_stack_profile)
 
 from azure.cli.command_modules.keyvault._transformers import (
-    extract_subresource_name, filter_out_managed_resources,
+    extract_subresource_name, force_json_output, filter_out_managed_resources,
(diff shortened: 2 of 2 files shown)


eiffel-fl:
Oops! I will deal with style issue later and push an updated commit!

jiasli:
Thanks for the contribution, but I am not sure if forcing JSON output is a good idea. @evelyn-ys to help take a look.

eiffel-fl:
Hi.

You are welcome.

I agree with you, the best solution would be to have correct output for table and tsv.
But I found this bug while using `$(az keyvault certificate get-default-policy)` as argument of another `az` command, thus I modified it with `$(az keyvault certificate get-default-policy -o json)` (I have table set as output in my azure config file) and thought it would be better than nothing.

Best regards.

evelyn-ys:
I don't think customer will require table/tsv output since `az keyvault certificate get-default-policy` returns a json result and `az keyvault certificate create --policy` does take a json input. This is a fluent process so we don't implement `-o table/tsv` for `az keyvault certificate get-default-policy`.

If you want a json output just use `-o json` or just omit `-o`, why do you use `-o table/tsv` and then force json output? It looks ambiguous to me.

eiffel-fl:
I think I did not express myself clearly.

My config file for `az` contains, among other parameters, the following:

```bash
$ more ~/.azure/config
[core]
output = table
```

Because when I use `az` I prefer to have thing outputted with this mode as I find it fill my terminal in width rather than in height (like JSON).
So, when I use:

```bash
$ az keyvault certificate get-default-policy

```

I got no output, as my default output mode is table.
For me, this is a bug (small one), which should be fixed as it impedes user experience.

To fix it, we have two choices:
1. Either implement outputting to table mode.
2. Or force JSON as output.

In this PR, I went with the second option as, like you stated, `az keyvault certificate create --policy`  waits a JSON input.

Is it clearer now?

evelyn-ys:
If you have table config as default, then I'll suggest you to call `az keyvault certificate get-default-policy -o json`. With explictly specifying `-o json`, it should show the correct output. It's not a bug for `-o table`. We just didn't implement table output because it will be used nowhere.

eiffel-fl:
I totally understand your point of view about not implementing `-o table`.
I also understand your suggestion about using `az keyvault certificate get-default-policy -o json`.

Nonetheless, this is not a smooth user experience.
If a user types `az keyvault certificate get-default-policy `, gets no output and has to realized by him/herself he/she needs to specify `-o json`.

Issue #8070, opened 2018-12-16
Azure Portal generated passwords with special characters not accepted by az login service principal


michaeldeongreen:
**Describe the bug**
When trying to use a Service Principal password that contains special characters, you get errors (see screenshot attachment)

**To Reproduce**
See screenshot attachment

**Expected behavior**
I am able to login via az login if the correct credentials are supplied

**Environment summary**
latest version of azure-cli, just installed it (2.0.52) and running the commands via Windows Powershell

**Additional context**
The passwords are being generated in the Azure Portal when I create a secret, so I have no control over how the password gets generated.  Would be nice to have an option to specify if special characters are placed in the password or not ie passwordgenerator.net.

See attached to view all the different combinations I have tried with no success.  As you can see in the last attempt, I put the wrong password in ("test"), which has no special characters and authentication failed, which tells me it is the password format that is the issue.

Also, I generated about 10 different secrets and all have special characters and would not work.

![2018_12_16_07_49_17_window](https://user-images.githubusercontent.com/8516636/50054621-89381c00-0109-11e9-8140-3569cd466894.png)

twoo:
same issue here

tjprescott:
The reason is because you are using Powershell and need to properly format your password string accordingly. From your screenshot, you can see that `$?` is being interpreted by the shell (that's why it appears green). Additionally, the `&` character is being interpreted by Powershell as the start of a new command.

Unfortunately, shell-specific processing occurs before the CLI is ever invoked, so we cannot control this behavior. The escape character in powershell is the backtick ` ` ` so you might try that.

Zerg00s:
Similar issue with the **&** character:
When running this from PowerShell: 

```az login -u "denis@contoso.com" --allow-no-subscriptions -p "Pass_&_AY"```

**Error:**
'_AY' is not recognized as an internal or external command,


Escaping & with a backtick does not help.

**Comment by Zerg00s** [comment]
Found a workaround that works in PowerShell:

```
$o365Admin = Get-Credential -Message "Enter Global Office 365 Admin Credentials"
az login -u $o365Admin.UserName --allow-no-subscriptions -p "`"$($o365Admin.GetNetworkCredential().password)`""
```

joshschmitter:
[...]

PR #20626, opened 2021-12-06
{login} double quoting on az login password

furmangg:
This workaround came from https://github.com/Azure/azure-cli/issues/8070

**Description**
With the special characters [?> (not sure which of those, but definitely one of them is the culprit, and probably many others) the only way to get az login for a SPN to succeed is with adding extra double quotes to this string. This wasn't documented so I thought I would propose it.

**Testing Guide**
$password = "<password>"
``az login --service-principal -u <client_id_here> -p "`"$password`"" --tenant <tenant_id_here>``

**History Notes**
[login] suggesting documenting double quotes on az login password to fix issue #8070

---

- [ x ] The PR title and description has followed the guideline in [Submitting Pull Requests](https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules#submitting-pull-requests).

- [ x ] I adhere to the [Command Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md).

- [ x ] I adhere to the [Error Handling Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/error_handling_guidelines.md).

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/profile/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/profile/_help.py
@@ -20,7 +20,7 @@ examples:
       text: az login
     - name: Log in with user name and password. This doesn't work with Microsoft accounts or accounts that have two-factor authentication enabled. Use -p=secret if the first character of the password is '-'.
       text: az login -u johndoe@contoso.com -p VerySecret
-    - name: Log in with a service principal using client secret. Use -p=secret if the first character of the password is '-'.
+    - name: Log in with a service principal using client secret. Use -p=secret if the first character of the password is '-'. For some special characters, including double quotes in the password string is required; in PowerShell this would be -p ``"`"abc[?>def`""``.
       text: az login --service-principal -u http://azure-cli-2016-08-05-14-31-15 -p VerySecret --tenant contoso.onmicrosoft.com
     - name: Log in with a service principal using client certificate.
       text: az login --service-principal -u http://azure-cli-2016-08-05-14-31-15 -p ~/mycertfile.pem --tenant contoso.onmicrosoft.com


src/azure-cli/azure/cli/command_modules/profile/_help.py
```diff
@@ -20,7 +20,7 @@
       text: az login
     - name: Log in with user name and password. This doesn't work with Microsoft accounts or accounts that have two-factor authentication enabled. Use -p=secret if the first character of the password is '-'.
       text: az login -u johndoe@contoso.com -p VerySecret
-    - name: Log in with a service principal using client secret. Use -p=secret if the first character of the password is '-'.
+    - name: Log in with a service principal using client secret. Use -p=secret if the first character of the password is '-'. For some special characters, including double quotes in the password string is required; in PowerShell this would be -p ``"`"abc[?>def`""``.
```
jiasli, line 23: ``` ``"`"abc[?>def`""`` ``` is not valid PowerShell syntax:

```
> echo ``"`"abc[?>def`""``
`"abc[?>def"`
```

Actually, PowerShell quoting issue doesn't only happen here for `az login --password`. It happens everywhere. We have  detailed documentation for this topic: https://github.com/Azure/azure-cli/blob/dev/doc/quoting-issues-with-powershell.md

jiasli:
Thanks a lot for the contribution. However, we don't want to include **specific workaround** for **specific command** running in **specific shell environment**.

Please see

- https://docs.microsoft.com/en-us/cli/azure/use-cli-effectively#pass-arguments
- https://github.com/Azure/azure-cli/blob/dev/doc/quoting-issues-with-powershell.md

PR #20636, opened 2021-12-07
[Core] Enable token cache encryption on MacOS

jiasli:
## Description

Utilize https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/103 to enable token cache encryption on MacOS.

We don't need to provide `account_name` by ourselves. Instead, `msal-extensions` automatically computes `account_name` from `signal_location`.

See code comment for more information.

⚠ **After this change, user will have to re-login so that credential is stored into Keychain.**

## References

Related email discussion: _Naming for service_name and account_name for MacOS token encryption_

### DIFF
--- a/src/azure-cli-core/azure/cli/core/_profile.py
+++ b/src/azure-cli-core/azure/cli/core/_profile.py
@@ -861,7 +861,7 @@ def _create_identity_instance(cli_ctx, *args, **kwargs):
     from .auth.identity import Identity
 
     # Only enable encryption for Windows (for now).
-    fallback = sys.platform.startswith('win32')
+    fallback = sys.platform.startswith('win32') or sys.platform.startswith('darwin')
     # encrypt_token_cache affects both MSAL token cache and service principal entries.
     encrypt = cli_ctx.config.getboolean('core', 'encrypt_token_cache', fallback=fallback)
 
--- a/src/azure-cli-core/azure/cli/core/auth/persistence.py
+++ b/src/azure-cli-core/azure/cli/core/auth/persistence.py
@@ -21,6 +21,8 @@ logger = get_logger(__name__)
 # Files extensions for encrypted and plaintext persistence
 file_extensions = {True: '.bin', False: '.json'}
 
+KEYCHAIN_SERVICE_NAME = 'azure-cli'
+
 
 def load_persisted_token_cache(location, encrypt):
     persistence = build_persistence(location, encrypt)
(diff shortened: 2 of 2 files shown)


src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -861,7 +861,7 @@ def _create_identity_instance(cli_ctx, *args, **kwargs):
     from .auth.identity import Identity
 
     # Only enable encryption for Windows (for now).
```
rayluo, line 863: This comment probably also needs to update to mention macOS now.

By the way, does this comment actually mean to say, "only allow fallback when running on Windows ..."?

src/azure-cli-core/azure/cli/core/auth/persistence.py
```diff
@@ -38,16 +40,27 @@ def build_persistence(location, encrypt):
...
+            logger.debug("Initializing FilePersistenceWithDataProtection.")
             return FilePersistenceWithDataProtection(location)
         if sys.platform.startswith('darwin'):
-            return KeychainPersistence(location, "my_service_name", "my_account_name")
+            # For KeychainPersistence, location is only used as a signal for the credential's last modified time.
+            # The credential is stored in Keychain identified by (service_name, account_name) combination.
+            # msal-extensions automatically computes account_name from signal_location.
+            # https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/103
+            logger.debug("Initializing KeychainPersistence")
+            return KeychainPersistence(location, service_name=KEYCHAIN_SERVICE_NAME)
```
rayluo, line 52: You probably already know that, this line won't work until https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/103 being released. So, please help review and approve my PR there (otherwise I will not give an approval to your PR here :-D)

src/azure-cli-core/azure/cli/core/_profile.py
jiasli: Oh. I forgot to update the comment!

src/azure-cli-core/azure/cli/core/auth/persistence.py
jiasli: I have installed the dev code of `msal-extensions` and it works perfectly

rayluo (approved):

calvinhzy (approved):

yonzhan:
Enable token cache encryption on MacOS

jiasli:
Given we are seeing a plethora of issues with Windows DPAPI encryption (https://github.com/Azure/azure-cli/issues?q=is%3Aissue+label%3A%22common+issue%22+winerror), it may not be a good idea to rush this PR.

rayluo:
> Given we are seeing a plethora of issues with Windows DPAPI encryption (https://github.com/Azure/azure-cli/issues?q=is%3Aissue+label%3A%22common+issue%22+winerror), it may not be a good idea to rush this PR.

Different encryption mechanisms are used on different platforms. All those DPAPI issues will not be observed on macOS platform. On the other hand, we do not know whether there would be "plethora of issues" with macOS's Keychain. @bgavrilms may comment based on the observation of keychain on .Net side.

bgavrilMS:
There is a plethora of issues on Mac as well. They mainly stem from how you package your app, in this case how you package Azure CLI. 

If you package it as Apple like it, with bundle id and via the app store, then you'll have access to the newest KeyChain APIs, which I think work rather well (this is what MSAL iOS uses on Mac, and I discussed it with them and it looks ok). Example of "old" API: `SecKeychainFindGenericPassword`

If however you don't have a bundle id, then you can't use the newest KeyChain APIs and the older ones are not as reliable. Ex: `SecItemAdd`

rayluo:
Indeed [MSAL Extensions for Python is using the old API](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/blob/1.0.0/msal_extensions/osx.py#L95-L97).

jiasli:
> Different encryption mechanisms are used on different platforms.

I understand. I am only saying if encryption problems occur on MacOS, this will heavy the burden we currently have. 

According to our internal email discussion "_Workaround for DPAPI/KeyChain Errors_": https://github.com/Azure/azure-powershell/issues/14478 is one example of KeyChain not working as expected on MacOS:

> The root cause seems to be using the older version of KeyChain APIs.

gattjoe:
Hi @rayluo @jiasli I submitted a PR in MSAL EX to use the modern OS X Keyring API, take a look and let me know if we can help move this forward https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/119

Issue #20666, opened 2021-12-09
GA az storage copy command


amishra-dev:
**Is your feature request related to a problem? Please describe.**
az storage copy uses azcopy which is GA. removing the preview message from this command will give confidence to more people to use it.

**Describe the solution you'd like**
The preview message removed

**Describe alternatives you've considered**

**Additional context**

jiasli:
Thanks @evelyn-ys for handling this issue.

evelyn-ys:
Has planned in this sprint

Issue #20663, opened 2021-12-09
Allow customers to passthrough azcopy flags in az storage copy command


amishra-dev:
**Is your feature request related to a problem? Please describe.**
azcopy has a rich set of options that lets the customer configure many facets of uploads like block size et al. Can we create an option like --passthrough "<azcopy options here>" 
anything within the quote above gets patched through to azcopy as is

**Describe the solution you'd like**
suggested above

**Describe alternatives you've considered**
LEverage azcopy environment variable where we can, but not everything is in environment variables

**Additional context**

jiasli:
Thanks @evelyn-ys for handling this issue.

evelyn-ys:
Has planned in this sprint

evelyn-ys:
Hi @amishra-dev, I have designed `--extra-options` parameter to take all the pass-through arguments like:
```
az storage copy -s source_file -d container_url --extra-options "--block-blob-size 0.25 --check-length"
```
But all the `extra_options` should be quoted with `""` or `''`. And this will cause complication if these extra options contain `""` or `''` themselves. See https://github.com/Azure/azure-cli/pull/20702#issuecomment-992195500

Do you think azcopy options need `""` or `''` often? I'd like to know your thoughts on @jiasli's comments.

evelyn-ys:
Do you mind if I mark the new parameter `--extra-options` as `experimental` or `preview`?

amishra-dev:
--extra-options as experimental sounds good to me. What do you think @zezha-msft?

amishra-dev:
> Hi @amishra-dev Ashish Mishra (AZURE) FTE, I have designed `--extra-options` parameter to take all the pass-through arguments like:
> 
> ```
> az storage copy -s source_file -d container_url --extra-options "--block-blob-size 0.25 --check-length"
> ```
> 
> But all the `extra_options` should be quoted with `""` or `''`. And this will cause complication if these extra options contain `""` or `''` themselves. See [#20702 (comment)](https://github.com/Azure/azure-cli/pull/20702#issuecomment-992195500)
> 
> Do you think azcopy options need `""` or `''` often? I'd like to know your thoughts on @jiasli Jiashuo Li FTE's comments.

@zezha-msft would be the best person to answer that.

zezha-msft:
Quotes might be needed for some flags where the user is free to put any content, some examples:

--metadata string Upload to Azure Storage with these key-value pairs as metadata.
[...]

Issue #20662, opened 2021-12-09
Upgrade AzCopy to the latest in the installer


amishra-dev:
**Is your feature request related to a problem? Please describe.**
AzCopy that ships with azcli is v10.5, there have been multiple perf improvements and features added. kindly upgrade the version

**Describe the solution you'd like**
azcopy move to the latest version 10.13

**Describe alternatives you've considered**
manually replacing the exe but that is not scalable to tell all customers

**Additional context**

jiasli:
Thanks @evelyn-ys for handling this issue.

evelyn-ys:
Has planned in this sprint

Issue #18869, opened 2021-07-15
Unable to pass space-separated values with leading dashes


jiasli:
## Requirement

To run `az synapse spark job submit`, the user wants to pass `--instance`, `trainer` as values to `--arguments`. 

Expected JSON:

```json
"arguments": [
  "--instance",
  "trainer"
]
```

## Symptom

`--arguments` is defined as 

https://github.com/Azure/azure-cli/blob/39c91741b46de99e136e0d5438c4dbe4d672f08e/src/azure-cli/azure/cli/command_modules/synapse/_params.py#L383

So the user tries to pass `--instance`, `trainer` separated by spaces, and the command fails

```sh
# Line breaks for legibility
az synapse spark job submit --name foo --workspace-name personalizer-trainer --spark-pool-name persbackend --main-definition-file foo --main-class-name org.personalizer.backend.PipelineManager --executors 4 --executor-size Medium 
                            --arguments --instance trainer
argument --arguments: expected at least one argument
```

As discussed in https://github.com/Azure/azure-cli/issues/16044#issuecomment-733955065, it is possible to use an equal sign `=` to concatenate the argument name and value for single-value argument: 

```
az keyvault secret set -n mysecret --vault-name mykeyvault --value=-secret
```

> This usage is not documented by `argparse`: https://docs.python.org/3/library/argparse.html, https://docs.python.org/3/howto/argparse.html

However, it doesn't work for multi-value arguments:

```sh
# Line breaks for legibility
az synapse spark job submit --name foo --workspace-name personalizer-trainer --spark-pool-name persbackend --main-definition-file foo --main-class-name org.personalizer.backend.PipelineManager --executors 4 --executor-size Medium 
                            --arguments=--instance trainer
unrecognized arguments: trainer
```

## Workarounds

There are several workaround worth considering:

1. Instead of using a `nargs='+'`, `--arguments` can take **single value** and split it with [`shlex.split`](https://docs.python.org/3/library/shlex.html#shlex.split)

2. Use `nargs=argparse.REMAINDER` to make `--arguments` take all remaining arguments. `azdev` uses this approach for [`--pytest-args`](https://github.com/Azure/azure-cli-dev-tools/blob/ac651115bdb66db69aea8c66c381fa24803f3486/azdev/params.py#L53) and perhaps this is the best way to go. 
   
   However, this usage has been removed from `argparse` document since Python 3.9 (https://bugs.python.org/issue17050):
   
[...]

PR #20702, opened 2021-12-13
[Storage] `az storage copy`: Add positional argument `extra_options` to pass through options to `azcopy`

evelyn-ys:
**Description**

Feature asked by #20662 #20663 #20666 
This PR:
- Adds positional argument `extra_options` with `experimental` tag for flexibility
- Upgrades `azcopy` from `10.8.0` to `10.13.0`
- Removes `preview` tag for `az storage copy`

**Testing Guide**

`az storage copy -s source_file -d destination_container_url -- --block-size-mb=0.25 --check-length`

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/storage/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/storage/_params.py
@@ -1076,7 +1076,7 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals, too-many-statem
         c.argument('cap_mbps', arg_group='Additional Flags', help="Caps the transfer rate, in megabits per second. "
                    "Moment-by-moment throughput might vary slightly from the cap. "
                    "If this option is set to zero, or it is omitted, the throughput isn't capped. ")
-        c.argument('extra_options', arg_group='Additional Flags',
+        c.argument('extra_options', arg_group='Additional Flags', is_experimental=True,
                    help="Other options which will be passed through to azcopy as it is")
 
     with self.argument_context('storage blob copy') as c:
--- a/src/azure-cli/azure/cli/command_modules/storage/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/storage/_params.py
@@ -1076,8 +1076,9 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals, too-many-statem
         c.argument('cap_mbps', arg_group='Additional Flags', help="Caps the transfer rate, in megabits per second. "
                    "Moment-by-moment throughput might vary slightly from the cap. "
                    "If this option is set to zero, or it is omitted, the throughput isn't capped. ")
-        c.argument('extra_options', arg_group='Additional Flags', is_experimental=True,
-                   help="Other options which will be passed through to azcopy as it is")
+        c.positional('extra_options', nargs='+', arg_group='Additional Flags', is_experimental=True,
+                     help="Other options which will be passed through to azcopy as it is. "
+                          "Please put all the extra options after a `--`")
 
     with self.argument_context('storage blob copy') as c:
         for item in ['destination', 'source']:
--- a/src/azure-cli/azure/cli/command_modules/storage/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/storage/_params.py
@@ -1076,6 +1076,8 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals, too-many-statem
         c.argument('cap_mbps', arg_group='Additional Flags', help="Caps the transfer rate, in megabits per second. "
                    "Moment-by-moment throughput might vary slightly from the cap. "
                    "If this option is set to zero, or it is omitted, the throughput isn't capped. ")
+        c.argument('extra_options', arg_group='Additional Flags',
+                   help="Other options which will be passed through to azcopy as it is")
 
     with self.argument_context('storage blob copy') as c:
         for item in ['destination', 'source']:
--- a/src/azure-cli/azure/cli/command_modules/storage/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/storage/_params.py
@@ -1076,7 +1076,7 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals, too-many-statem
         c.argument('cap_mbps', arg_group='Additional Flags', help="Caps the transfer rate, in megabits per second. "
                    "Moment-by-moment throughput might vary slightly from the cap. "
                    "If this option is set to zero, or it is omitted, the throughput isn't capped. ")
-        c.positional('extra_options', nargs='+', arg_group='Additional Flags', is_experimental=True,
+        c.positional('extra_options', nargs='*', is_experimental=True, default=[],
                      help="Other options which will be passed through to azcopy as it is. "
                           "Please put all the extra options after a `--`")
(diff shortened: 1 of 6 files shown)


jiasli:
Unfortunately, there is not easy solution for argument pass-though. Apparently, you are choosing the 1st workaround from https://github.com/Azure/azure-cli/issues/18869.

jiasli:
Unfortunately, there is no easy solution for argument pass-though. Apparently, you are choosing the 1st workaround from https://github.com/Azure/azure-cli/issues/18869. This makes passing quotes `"` `'` complicated as they need to be quoted in shell, such as

```sh
--extra-options "--some-other-parameter \"value with space\""
```

or even worse with single quotes both inside and outside

```sh
--extra-options '--some-other-parameter '"'"'value with space'"'"' something_else'
                 ^^^^^^^^^^^^^^^^^^^^^^^  ^  ^^^^^^^^^^^^^^^^  ^  ^^^^^^^^^^^^^^^  # string content

$ echo '--some-other-parameter '"'"'value with space'"'"' something_else'
--some-other-parameter 'value with space' something_else
```

jiasli:
Things in PowerShell gets worse as now you need to consider escaping 3 times from

1. `az`
2. PowerShell
3. CMD

This becomes unmanageably complicated. 

See https://github.com/Azure/azure-cli/blob/dev/doc/quoting-issues-with-powershell.md

jiasli:
Here is the script I use to debug argument passing in shells:

```py
# echo_args.py

import sys


def echo(args):
    print('[str]')
    for i, arg in enumerate(args):
        print(f"{i}: {arg!s}")

    print()

    print('[repr]')
    for i, arg in enumerate(args):
        print(f"{i}: {arg!r}")


echo(sys.argv)
```

You can invoke it by

```
> python .\echo_args.py --extra-options "--some-other-param \`"value with \\\`"quote\\\`"\`""
[str]
0: .\echo_args.py
1: --extra-options
2: --some-other-param "value with \"quote\""

[repr]
0: '.\\echo_args.py'
1: '--extra-options'
2: '--some-other-param "value with \\"quote\\""'
```

src/azure-cli/azure/cli/command_modules/storage/_params.py
```diff
@@ -1076,6 +1076,8 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals, too-many-statem
         c.argument('cap_mbps', arg_group='Additional Flags', help="Caps the transfer rate, in megabits per second. "
                    "Moment-by-moment throughput might vary slightly from the cap. "
                    "If this option is set to zero, or it is omitted, the throughput isn't capped. ")
+        c.argument('extra_options', arg_group='Additional Flags',
+                   help="Other options which will be passed through to azcopy as it is")
```
jiasli, line 1080: Let's include some examples in the help message.

jiasli: If we can't decide which way from #18869 to go, consider mark it as experimental.

src/azure-cli/azure/cli/command_modules/storage/_params.py
```diff
@@ -1076,6 +1076,8 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals, too-many-statem
         c.argument('cap_mbps', arg_group='Additional Flags', help="Caps the transfer rate, in megabits per second. "
                    "Moment-by-moment throughput might vary slightly from the cap. "
                    "If this option is set to zero, or it is omitted, the throughput isn't capped. ")
+        c.argument('extra_options', arg_group='Additional Flags',
```
evelyn-ys, line 1079:
```suggestion
        c.argument('extra_options', arg_group='Additional Flags', is_experimental=True,
```

src/azure-cli/azure/cli/command_modules/storage/_params.py
```diff
@@ -1076,6 +1076,9 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals, too-many-statem
         c.argument('cap_mbps', arg_group='Additional Flags', help="Caps the transfer rate, in megabits per second. "
                    "Moment-by-moment throughput might vary slightly from the cap. "
                    "If this option is set to zero, or it is omitted, the throughput isn't capped. ")
+        c.positional('extra_options', nargs='*', is_experimental=True, default=[],
```
jiasli, line 1079: I think `[]` is already the default:

```py
import argparse

parser = argparse.ArgumentParser()
parser.add_argument('extra', nargs="*")

print(parser.parse_args(''.split()))

>>> Namespace(extra=[])
```

jiasli:
Copying my comment from https://github.com/Azure/azure-cli/issues/20663#issuecomment-993243242

This topic was intensively discussed before in https://github.com/Azure/azure-cli/issues/18869.

Then I would propose the `--` solution:

```
az storage copy -s source_file -d container_url -- --bla="bla" --foo="foo" --bar="bar"
```

This is a common convention in Linux world:

- https://unix.stackexchange.com/questions/11376/what-does-double-dash-mean
- https://stackoverflow.com/questions/13321458/meaning-of-git-checkout-double-dashes

@chasewilson for awareness. If we go down this path, then other commands can do the same.

chasewilson:
> Copying my comment from [#20663 (comment)](https://github.com/Azure/azure-cli/issues/20663#issuecomment-993243242)
> 
> This topic was intensively discussed before in #18869.
> 
> Then I would propose the `--` solution:
> 
> ```
> az storage copy -s source_file -d container_url -- --bla="bla" --foo="foo" --bar="bar"
> ```
> 
> This is a common convention in Linux world:
> 
> * https://unix.stackexchange.com/questions/11376/what-does-double-dash-mean
> * https://stackoverflow.com/questions/13321458/meaning-of-git-checkout-double-dashes
> 
> @chasewilson for awareness. If we go down this path, then other commands can do the same.

I think this makes a lot of sense all things considered. I'm glad this is experimental as I'm hesitent to make this the standard for similar scenarios. From a command line perspective, it's less than ideal to have a "native" command like this where you have to know another command line tools (azcopy) parameters to make the most of it. 

Is there are reason we don't support wrapping the new flags and parameters. Please forgive me if i'm catching up and I'm not saying we shouldn't implement the current solution, I just want to understand all avenues before committing. 

Another question, are we relying on the native bash parsing for the `--` mangagement? If so, how will we approach from Windows?

Thank you!

jiasli:
> Another question, are we relying on the native bash parsing for the `--` mangagement? If so, how will we approach from Windows?

`--` interpretation is provided by Python's built-in lib `argparse`, so yes, it works on Windows too.

evelyn-ys:
> Is there are reason we don't support wrapping the new flags and parameters.

There has been and will be more and more `azcopy` options in new versions. It's kind of duplicated work for CLI to add all the options and release new version. So we would like pass-through arguments. But if you think it's safer to wrap them all, I can also understand.

src/azure-cli/azure/cli/command_modules/storage/_params.py
evelyn-ys: I didn't know and added manually. Thanks for notifying.

jiasli (approved):

calvinhzy (approved):

PR #20717, opened 2021-12-13
{testsdk} Remove `CommunicationResourcePreparer` from `azure-cli-testsdk`

moirf:
**Description**

Add default connection string required in communication extension in resource preparer file.
**Testing Guide**

NA
**History Notes**

---

### DIFF
--- a/src/azure-cli-testsdk/azure/cli/testsdk/__init__.py
+++ b/src/azure-cli-testsdk/azure/cli/testsdk/__init__.py
@@ -7,8 +7,7 @@ from .scenario_tests import live_only, record_only, get_sha1_hash
 
 from .base import ScenarioTest, LiveScenarioTest, LocalContextScenarioTest
 from .preparers import (StorageAccountPreparer, ResourceGroupPreparer, RoleBasedServicePrincipalPreparer,
-                        KeyVaultPreparer, ManagedApplicationPreparer, VirtualNetworkPreparer, VnetNicPreparer,
-                        CommunicationResourcePreparer)
+                        KeyVaultPreparer, ManagedApplicationPreparer, VirtualNetworkPreparer, VnetNicPreparer)
 from .exceptions import CliTestError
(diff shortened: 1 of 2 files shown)


jiasli:
Please add more context for why this is changed.

jiasli (changes requested):
Please add more context for why this is changed.

beltr0n (approved):

moirf:
> Please add more context for why this is changed.
The test cases for azure-cli-extension, validate the endpoints, for consistency we update the endpoint matching with sanitized.

beltr0n:
> 

I think I'd also like to know, did the previous value work, or is this just a change for consistency? This isn't a new value being added, it's a change to an existing value. I believe for Azure SDK, there's code that checks the value conforms to a certain format.

moirf:
@beltr0n @jiasli we are writing the cli-extension commands for communication, the PR for this is pending here -  https://github.com/Azure/azure-cli-extensions/pull/4206
For running the live mode, the test cases are passed because the account_key( which is actually connection string see line 185) will return the string into correct format, in case of playback mode we need the connection string in same connection string format. since this is not into a correct connection string format, test cases failed as it is looking for an URI into it.

wangzelin007 (approved):

jiasli:
I would suggest not to put `CommunicationResourcePreparer` (from #20552) in `azure-cli-testsdk`, as `azure-cli-testsdk` is only for common preparers. Instead, it should go to you own command module/extension.

moirf:
@jiasli Ok, will revert the changes for [#20552](https://github.com/Azure/azure-cli/pull/20552). And close this PR.

jiasli (approved):

yonzhan (approved):

moirf:
@yonzhan @jiasli can you please merge the branch as I do not have access.

jiasli:
Done. Thanks for notifying me.

Issue #20720, opened 2021-12-13
az - terraform plan - error: EOFError: Ran out of input


vjardin:
## Describe the bug

**Command Name**
`az account get-access-token`

**Errors:**
```
The command failed with an unexpected error. Here is the traceback:
Ran out of input
Traceback (most recent call last):
  File "/opt/az/lib/python3.6/site-packages/knack/cli.py", line 231, in invoke
    cmd_result = self.invocation.execute(args)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 658, in execute
    raise ex
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 721, in _run_jobs_serially
    results.append(self._run_job(expanded_arg, cmd_copy))
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 692, in _run_job
    result = cmd_copy(params)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 328, in __call__
    return self.handler(*args, **kwargs)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/command_operation.py", line 121, in handler
    return op(**command_args)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/command_modules/profile/custom.py", line 77, in get_access_token
    tenant=tenant)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/_profile.py", line 383, in get_raw_token
    credential = self._create_credential(account, tenant)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/_profile.py", line 593, in _create_credential
    return identity.get_user_credential(username_or_sp_id)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/auth/identity.py", line 209, in get_user_credential
    return UserCredential(self.client_id, username, **self._msal_app_kwargs)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/auth/identity.py", line 84, in _msal_app_kwargs
    Identity._msal_http_cache = self._load_msal_http_cache()
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/auth/identity.py", line 113, in _load_msal_http_cache
    persisted_http_cache = pickle.load(f)
EOFError: Ran out of input

```

## To Reproduce:
[...]

PR #20722, opened 2021-12-14
{Core} Retry loading MSAL HTTP cache in case of failure

### LINKED ISSUE #20707: Fail to unpickle `msal_http_cache.bin`: EOFError: Ran out of input
Removed `~/.azure` and tried again. Same issue happens:

```
│ Error: unable to get authorization token for resource manager: Error obtaining Authorization Token from the Azure CLI: Error parsing json result from the Azure CLI: Error waiting for the Azure CLI: exit status 1: ERROR: The command failed with an unexpected error. Here is the traceback:
│ ERROR: Ran out of input
│ Traceback (most recent call last):
│   File "/opt/az/lib/python3.6/site-packages/knack/cli.py", line 231, in invoke
│     cmd_result = self.invocation.execute(args)
│   File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 658, in execute
│     raise ex
│   File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 721, in _run_jobs_serially
│     results.append(self._run_job(expanded_arg, cmd_copy))
│   File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 692, in _run_job
│     result = cmd_copy(params)
│   File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 328, in __call__
│     return self.handler(*args, **kwargs)
│   File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/command_operation.py", line 121, in handler
│     return op(**command_args)
│   File "/opt/az/lib/python3.6/site-packages/azure/cli/command_modules/profile/custom.py", line 77, in get_access_token
│     tenant=tenant)
│   File "/opt/az/lib/python3.6/site-packages/azure/cli/core/_profile.py", line 383, in get_raw_token
│     credential = self._create_credential(account, tenant)
│   File "/opt/az/lib/python3.6/site-packages/azure/cli/core/_profile.py", line 593, in _create_credential
│     return identity.get_user_credential(username_or_sp_id)
│   File "/opt/az/lib/python3.6/site-packages/azure/cli/core/auth/identity.py", line 209, in get_user_credential
│     return UserCredential(self.client_id, username, **self._msal_app_kwargs)
│   File "/opt/az/lib/python3.6/site-packages/azure/cli/core/auth/identity.py", line 84, in _msal_app_kwargs
│     Identity._msal_http_cache = self._load_msal_http_cache()
│   File "/opt/az/lib/python3.6/site-packages/azure/cli/core/auth/identity.py", line 113, in _load_msal_http_cache
│     persisted_http_cache = pickle.load(f)
│ EOFError: Ran out of input
│ To open an issue, please run: 'az feedback'
│
│   with provider["registry.terraform.io/hashicorp/azurerm"],
│   on terraform_config.tf line 13, in provider "azurerm":
│   13: provider "azurerm" {
```

_Originally posted by @edvinas31 in https://github.com/Azure/azure-cli/issues/20252#issuecomment-992255336_

jiasli:
## Symptom

Fix #20707, #20676, #20720

When invoked by Terraform concurrently, one instance of `az` fails with:

```
│   File "/opt/az/lib/python3.6/site-packages/azure/cli/core/auth/identity.py", line 113, in _load_msal_http_cache
│     persisted_http_cache = pickle.load(f)
│ EOFError: Ran out of input
```

This is because a racing condition happens after the cache file `~/.azure/msal_http_cache.bin` is created but not filled yet:

https://github.com/Azure/azure-cli/blob/713b0edca860a8e4037bfc865d847b945fb12886/src/azure-cli-core/azure/cli/core/auth/identity.py#L117-L121

The second instance will see an empty cache file, resulting in `EOFError`.

## Change

This PR adds a retry logic to `_load_msal_http_cache` so that it retries several times in case of failure. If it still fails at last, ignore exceptions and use a fresh MSAL HTTP cache.

An experimental config option `core.use_msal_http_cache` is added to allow the user opt out of MSAL HTTP cache.

## Testing Guide

Sleep for a while here:

https://github.com/Azure/azure-cli/blob/6e431b0bc879eedab1f8542133515a6c11e4097a/src/azure-cli-core/azure/cli/core/auth/identity.py#L124-L128

This will reproduce `EOFError`. Launch another instance of `az` with `--debug` to see `EOFError` is ignored. In reality, the `retry` logic will allow the second `az` to load the cache file successfully.

### DIFF
--- a/src/azure-cli-core/azure/cli/core/auth/identity.py
+++ b/src/azure-cli-core/azure/cli/core/auth/identity.py
@@ -5,9 +5,12 @@
 
 import json
 import os
+import pickle
 import re
+import time
 
 from azure.cli.core._environment import get_config_dir
+from azure.cli.core.decorators import retry
 from msal import PublicClientApplication
 
 from knack.log import get_logger
@@ -45,7 +48,7 @@ class Identity:  # pylint: disable=too-many-instance-attributes
     # It follows singleton pattern so that _secret_file is read only once.
     _service_principal_store_instance = None
 
-    def __init__(self, authority, tenant_id=None, client_id=None, encrypt=False):
+    def __init__(self, authority, tenant_id=None, client_id=None, encrypt=False, use_msal_http_cache=True):
         """
         :param authority: Authentication authority endpoint. For example,
             - AAD: https://login.microsoftonline.com
@@ -58,7 +61,8 @@ class Identity:  # pylint: disable=too-many-instance-attributes
         self.authority = authority
         self.tenant_id = tenant_id
         self.client_id = client_id or AZURE_CLI_CLIENT_ID
-        self.encrypt = encrypt
+        self._encrypt = encrypt
+        self._use_msal_http_cache = use_msal_http_cache
 
         # Build the authority in MSAL style
         self._msal_authority, self._is_adfs = _get_authority_url(authority, tenant_id)
@@ -80,7 +84,7 @@ class Identity:  # pylint: disable=too-many-instance-attributes
         if not Identity._msal_token_cache:
             Identity._msal_token_cache = self._load_msal_token_cache()
 
-        if not Identity._msal_http_cache:
+        if self._use_msal_http_cache and not Identity._msal_http_cache:
             Identity._msal_http_cache = self._load_msal_http_cache()
 
         return {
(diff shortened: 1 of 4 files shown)


src/azure-cli-core/azure/cli/core/decorators.py
```diff
@@ -81,3 +84,25 @@ def _wrapped_func(*args, **kwargs):
                     return fallback_return
         return _wrapped_func
     return _decorator
+
+
+def retry(retry_times=3, interval=0.5, exceptions=Exception):
```
rayluo, line 89: This decorator idea is neat. I may consider backport this to MSAL EX. :-)

Regarding the `exceptionS` parameter, do you also intend to have it support accepting a tuple, such as `exceptions=(ValueError, ZeroDivisionError)`? It might just work, but better test it out (at least manually), if you haven't already.

src/azure-cli-core/azure/cli/core/auth/identity.py
```diff
@@ -100,25 +103,46 @@ def _msal_app(self):
...
+            # We still get exception after retry:
+            # - pickle.UnpicklingError is caused by corrupted cache file, perhaps due to concurrent writes.
+            # - EOFError is caused by empty cache file created by other az instance, but hasn't been filled yet.
+            logger.debug("Failed to load MSAL HTTP cache: %s. Using a fresh one.", ex)
             persisted_http_cache = {}  # Ignore a non-exist or corrupted http_cache
-        atexit.register(lambda: pickle.dump(
-            # When exit, flush it back to the file.
-            # If 2 processes write at the same time, the cache will be corrupted,
-            # but that is fine. Subsequent runs would reach eventual consistency.
-            persisted_http_cache, open(self._http_cache_file, 'wb')))
```
jiasli, line 121: Could it be possible that because `open` is not used with `with` statement, the file content is not being flushed immediately after the write operation, causing the empty file?

src/azure-cli-core/azure/cli/core/auth/identity.py
```diff
@@ -100,25 +103,46 @@ def _msal_app(self):
...
+
+        # When exiting, flush it back to the file.
+        # If 2 processes write at the same time, the cache will be corrupted,
+        # but that is fine. Subsequent runs would reach eventual consistency.
+        atexit.register(self._dump_msal_http_cache)
```
jiasli, line 145: I am thinking perhaps HTTP cache can do the same thing as token cache - use a **last modified time**, so that dumping won't happen too frequently.

src/azure-cli-core/azure/cli/core/decorators.py
jiasli: Yes, it definitely works.

src/azure-cli-core/azure/cli/core/auth/identity.py
rayluo: Unless there is evidence to prove otherwise, I tend to assume the flush would eventually happen during the implicit close, when the entire Python process is shutting down. That being said, I like your new implementation in this PR. Production code should indeed use `with` statement to ensure a timely closure.

src/azure-cli-core/azure/cli/core/auth/identity.py
rayluo: I'll give it some thought to see whether it is possible.

Until then, this PR is good to go.

rayluo (approved):

yonzhan:
Retry loading MSAL HTTP cache

evelyn-ys (approved):

Issue #20732, opened 2021-12-14
Missing dependency of `msrestazure` from `azure-cli-core`


irvokopo:
> ### `az feedback` auto-generates most of the information requested below, as of CLI version 2.0.62

**Describe the bug**

Is there a missing dependency?

**To Reproduce**

In this code CLoudError is imported from msrestazure: https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/azure/cli/core/commands/validators.py#:~:text=from%20msrestazure.azure_exceptions%20import%20CloudError 

**Expected behavior**

Should the 'msrestazure' package be listed as a dependency here: https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/setup.py#:~:text=%5D-,DEPENDENCIES%20%3D%20%5B,%5D,-%23%20dependencies%20for%20specific

**Environment summary**

**Additional context**

jiasli:
@zhoxing-ms, is this line still valid after we migrate to Track 2 for `azure-mgmt-resource`?

https://github.com/Azure/azure-cli/blob/e3fb82480bfd34b6b42f2f5fb541f925a52eebf8/src/azure-cli-core/azure/cli/core/commands/validators.py#L74

zhoxing-ms:
Good catch! I have submitted a PR #20739 to fix this issue

irvokopo:
Hello,
Thank you very much for this!
Would it, maybe make sense to tackle similar usage of CloudError elsewhere as well, for e.g.:
https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/azure/cli/core/util.py#:~:text=exceptions%20import%20JMESPathError-,from%20msrestazure.azure_exceptions%20import%20CloudError,-from%20msrest.exceptions

Thank you again!

jiasli:
Actually we can't, as some SDKs are still on Track 1. Azure CLI needs to catch those `CloudError` so that Track 1 SDKs' exceptions can be correctly handled.

irvokopo:
I see - thank you again for addressing the issue getting back to me - happy to have this closed : )

irvokopo:
P.S.: Just to mention, I've had one other missing-dependency issue of 'packaging', from here:
https://github.com/Azure/azure-cli/blob/d52397ab0f9e80c1f5e68e200ae66b6a272da0ac/src/azure-cli-core/azure/cli/core/__init__.py#:~:text=from%20packaging.version%20import%20parse

[...]

PR #20739, opened 2021-12-15
{Core} Migrate `CloudError` for resource Track 2 SDK

zhoxing-ms:
Fix: #20732

Since the resource module has migrated to Track 2 https://github.com/Azure/azure-cli/pull/17783, `CloudError` should be replaced by `HttpResponseError` in track 2, so submit this PR to modify it

**Description**

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/commands/validators.py
+++ b/src/azure-cli-core/azure/cli/core/commands/validators.py
@@ -71,13 +71,13 @@ def generate_deployment_name(namespace):
 def get_default_location_from_resource_group(cmd, namespace):
     if not namespace.location:
         from azure.cli.core.commands.client_factory import get_mgmt_service_client
-        from msrestazure.azure_exceptions import CloudError
+        from azure.core.exceptions import HttpResponseError
         from knack.util import CLIError
 
         resource_client = get_mgmt_service_client(cmd.cli_ctx, ResourceType.MGMT_RESOURCE_RESOURCES)
         try:
             rg = resource_client.resource_groups.get(namespace.resource_group_name)
-        except CloudError as ex:
+        except HttpResponseError as ex:
             raise CLIError('error retrieving default location: {}'.format(ex.message))
         namespace.location = rg.location  # pylint: disable=no-member
         logger.debug("using location '%s' from resource group '%s'", namespace.location, rg.name)


src/azure-cli-core/azure/cli/core/commands/validators.py
```diff
@@ -71,13 +71,13 @@ def generate_deployment_name(namespace):
...
         from azure.cli.core.commands.client_factory import get_mgmt_service_client
-        from msrestazure.azure_exceptions import CloudError
+        from azure.core.exceptions import HttpResponseError
         from knack.util import CLIError
 
         resource_client = get_mgmt_service_client(cmd.cli_ctx, ResourceType.MGMT_RESOURCE_RESOURCES)
         try:
             rg = resource_client.resource_groups.get(namespace.resource_group_name)
-        except CloudError as ex:
+        except HttpResponseError as ex:
```
jiasli, line 80: Perhaps we don't use try catch here to let `azure.cli.core.parser.AzCliCommandParser.validation_error` handle exceptions, such as `azure.core.exceptions.ResourceNotFoundError`?

zhoxing-ms: OK, makes sense~
Here raise the original specific error seems to be a better choice．

zhoxing-ms: As we discussed offline, we first maintain the latest exit code and error throwing ways (because it seems to work well at present), so we remove the `try-catch` logic first.

If additional information is needed to help locate the problem in the future, we can consider recording the information in the form of error log

src/azure-cli-core/azure/cli/core/commands/validators.py
```diff
@@ -71,15 +71,11 @@ def generate_deployment_name(namespace):
 def get_default_location_from_resource_group(cmd, namespace):
     if not namespace.location:
         from azure.cli.core.commands.client_factory import get_mgmt_service_client
-        from msrestazure.azure_exceptions import CloudError
-        from knack.util import CLIError
 
         resource_client = get_mgmt_service_client(cmd.cli_ctx, ResourceType.MGMT_RESOURCE_RESOURCES)
```
jiasli, line 75: Let's add some comment in the code so that people can know why the exception is not caught in the future.

```suggestion
        # We don't use try catch here to let azure.cli.core.parser.AzCliCommandParser.validation_error
        # handle exceptions, such as azure.core.exceptions.ResourceNotFoundError
        resource_client = get_mgmt_service_client(cmd.cli_ctx, ResourceType.MGMT_RESOURCE_RESOURCES)
```

jiasli (approved):

PR #20748, opened 2021-12-15
[App Service] `az appservice plan update`/`az webapp update` : Add elastic scale support 

StrawnSC:
**Description**

Supports elastic scale for Web Apps by adding `--elastic-scale` and `--max-elastic-worker-count` parameters to `az appservice plan update` as well as `--minimum-elastic-instance-count` and `--prewarmed-instance-count` parameters to `az appservice plan update`
Also refactors a utility method related to app service plan SKUs to have a more appropriate name 

**Testing Guide**

`azdev test test_elastic_scale_plan test_elastic_scale_plan_max_workers`

**History Notes**

[App Service] `az appservice plan update`: Add `--elastic-scale` and `--max-elastic-worker-count` parameters to support elastic scale
[App Service] `az webapp update`: Add `--minimum-elastic-instance-count` and `--prewarmed-instance-count` parameters to support setting instance count

---


StrawnSC:
Not ready for review until I figure out how to get the `--prewarmed-instance-count` parameter to work -- For some reason `SiteConfig` objects have a minimumElasticInstanceCount field, but it's not included in the body of the PUT request when I check with `az webapp update ... --debug`

yonzhan:
App Service

wangzelin007 (approved):

src/azure-cli/azure/cli/command_modules/appservice/_params.py
```diff
@@ -131,6 +131,9 @@ def load_arguments(self, _):
 
     with self.argument_context('appservice plan update') as c:
         c.argument('sku', arg_type=sku_arg_type)
+        c.argument('elastic_scale', arg_type=get_three_state_flag(), is_preview=True, help='Enable or disable automatic scaling. Set to "true" to enable elastic scale for this plan, or "false" to disable elastic scale for this plan. The SKU must be a Premium V2 SKU (P1V2, P2V2, P3V2) or a Premium V3 SKU (P1V3, P2V3, P3V3)')
```
zhoxing-ms, line 134: Could we consider using `get_three_state_flag()` instead of `action='store_true'` ?

StrawnSC: Since this is an update command the parameter needs to be `True` or `False` if the user sets a value for the parameter and `None` if they do not (we don't want elastic scale to be accidentally disabled when someone runs an update command without using `--elastic-scale`) Does using `action='store_true'` allow for the argument to be `None` as well as `True`/`False`?

zhoxing-ms: Got it, thanks~

zhoxing-ms (approved):

PR #20847, opened 2021-12-28
{Core} Introduce `BinaryCache` for MSAL HTTP cache

jiasli:
**Description**

A refinement to https://github.com/Azure/azure-cli/pull/20722.

Instead of dumping MSAL HTTP cache to a file **every time Azure CLI exits**, MSAL HTTP cache is dumped to a file **only when the cache content is changed**. 

This significantly **reduces the number of disk write operations**, thus reduces the possibility of encountering a corrupted HTTP cache file (`pickle.UnpicklingError`, `EOFError`).

But, the limitation is that this solution only works when MSAL updates the HTTP cache with `__setitem__`. If MSAL directly edits the content of cache items, the operation will slip under `BinaryCache`'s radar.

### DIFF
--- a/src/azure-cli-core/azure/cli/core/auth/binary_cache.py
+++ b/src/azure-cli-core/azure/cli/core/auth/binary_cache.py
@@ -1,3 +1,8 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
 import collections.abc as collections
 import pickle
 
--- a/src/azure-cli-core/azure/cli/core/auth/binary_cache.py
+++ b/src/azure-cli-core/azure/cli/core/auth/binary_cache.py
@@ -1,0 +1,79 @@
+import collections.abc as collections
+import pickle
+
+from azure.cli.core.decorators import retry
+from knack.log import get_logger
+
+logger = get_logger(__name__)
+
+
+class BinaryCache(collections.MutableMapping):
+    """
+    Derived from azure.cli.core._session.Session.
+    A simple dict-like class that is backed by a binary file.
+
+    All direct modifications with `__setitem__` and `__delitem__` will save the file.
+    Indirect modifications should be followed by a call to `save`.
+    """
+
+    def __init__(self, file_name):
+        super().__init__()
+        self.filename = file_name
+        self.data = {}
+        self.load()
+
+    @retry()
+    def _load(self):
+        """Load cache with retry. If it still fails at last, raise the original exception as-is."""
+        try:
+            with open(self.filename, 'rb') as f:
+                return pickle.load(f)
+        except FileNotFoundError:
+            # The cache file has not been created. This is expected. No need to retry.
+            logger.debug("%s not found. Using a fresh one.", self.filename)
+            return {}
+
+    def load(self):
+        logger.debug("load: %s", self.filename)
+        try:
+            self.data = self._load()
+        except (pickle.UnpicklingError, EOFError) as ex:
+            # We still get exception after retry:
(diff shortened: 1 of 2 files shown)


src/azure-cli-core/azure/cli/core/auth/binary_cache.py
```diff
@@ -0,0 +1,79 @@
...
+            self.data = self._load()
+        except (pickle.UnpicklingError, EOFError) as ex:
+            # We still get exception after retry:
+            # - pickle.UnpicklingError is caused by corrupted cache file, perhaps due to concurrent writes.
+            # - EOFError is caused by empty cache file created by other az instance, but hasn't been filled yet.
+            logger.debug("Failed to load cache: %s. Using a fresh one.", ex)
+            self.data = {}  # Ignore a non-existing or corrupted http_cache
+
+    @retry()
+    def _save(self):
```
jiasli, line 48: My gut feel tells me to retry the `_save` method, though I don't know what exception will be raised here.

yonzhan:
MSAL HTTP cache refinement

src/azure-cli-core/azure/cli/core/auth/binary_cache.py
jiasli: Wrote an example to constantly simulate concurrent writes which can corrupt a file:

```py
# proc.py

import time
import sys

string_to_write = sys.argv[1]
with open(r"testfile.txt", 'w') as f:
    print(f"{string_to_write}: Sleeping for 5s before writing")
    time.sleep(5)
    f.write(string_to_write)

print(f"{string_to_write}: Finished writing")
```

```py
# caller.py

import subprocess
import time

proc1 = subprocess.Popen(['python', 'proc.py', 'abcdefghijklmnopqrstuvwxyz'])
time.sleep(2)
proc2 = subprocess.Popen(['python', 'proc.py', '0123456789'])

proc1.wait()
proc2.wait()

with open(r"testfile.txt") as f:
    print("Final result:", f.read())
```

Strings `abcdefghijklmnopqrstuvwxyz` and `0123456789` will be written to `testfile.txt` in the order of

```
|---|---|---|---|---|         abcdefghijklmnopqrstuvwxyz
        |---|---|---|---|---| 0123456789
```

Output:

```
python caller.py
abcdefghijklmnopqrstuvwxyz: Sleeping for 5s before writing
0123456789: Sleeping for 5s before writing
abcdefghijklmnopqrstuvwxyz: Finished writing
0123456789: Finished writing
Final result: 0123456789klmnopqrstuvwxyz
```

yonzhan (approved):

src/azure-cli-core/azure/cli/core/auth/binary_cache.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def __init__(self, file_name):
+        super().__init__()
+        self.filename = file_name
+        self.data = {}
+        self.load()
+
+    @retry()
+    def _load(self):
+        """Load cache with retry. If it still fails at last, raise the original exception as-is."""
```
rayluo, line 32: Nit-picking: This method does not currently retry. But that is perhaps good enough. We would just need to rephrase this comment.

rayluo:
> the limitation is that this solution only works when MSAL updates the HTTP cache with `__setitem__`. If MSAL directly edits the content of cache items, the operation will slip under `BinaryCache`'s radar.

No worry. I can confirm that only uses methods defined in `MutableMapping` to change http cache's content. (In fact, I am not aware of any other-than-`__setitem__`-or-`__delitem__` way to directly *edit* the content of a dict. You might be referring to forcefully reset or replace a mapping, such as `old_dict = {"new": "content"}`, but that is no longer *editing* an existing dict. MSAL does not do that, for the exact reason of allowing you to supply your customized mapping.)

I am approving this PR.

rayluo (approved):

evelyn-ys (approved):

calvinhzy (approved):

PR #20850, opened 2021-12-28
[App Config] BREAKING CHANGE: Support app service slots

Jasonra102:
**Description**
When importing or exporting an app configuration and the destination is an app service, the normal functionality is to use the production slot even when a non-production slot is provided as the app service ID. This change will use the provided slot to import / export instead of production. This has been marked as a breaking change since it will react differently than before if given the app service ID for a slot.

**Testing Guide**

**History Notes**

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
+++ b/src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
@@ -1,4 +1,4 @@
-# --------------------------------------------------------------------------------------------
+
 # Copyright (c) Microsoft Corporation. All rights reserved.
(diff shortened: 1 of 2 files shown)


yonzhan:
AppConfig

zhoxing-ms:
@Jasonra102 Could you add some tests for this change?

Jasonra102:
Testing has been added

wangzelin007:
> Testing has been added

Please re-record the related recording file.

Jasonra102:
The recording file has been updated

zhoxing-ms (approved):

wangzelin007 (approved):

PR #20215, opened 2021-11-05
[Core] Support managed identity in App Service container

jiasli:
## Description

- Close https://github.com/Azure/azure-cli/issues/19480
- Close https://github.com/Azure/azure-cli/issues/20196
- Close https://github.com/Azure/azure-cli/issues/20186

For more details, see https://github.com/Azure/azure-cli/issues/19480#issuecomment-961197625

`msrestazure` still uses `2017-09-01` for [managed identity in App Service container](https://docs.microsoft.com/en-us/azure/app-service/overview-managed-identity?tabs=powershell#rest-protocol-examples). The returned `expires_on` is a datetime string like `"11/05/2021 15:18:31 +00:00"`. This differs from `expires_on` returned by [managed identity on Azure VM](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token):

- epoch in str (VM managed identity endpoint 2018-02-01): `"1605238724"`
- datetime in Month/Day/Year str with timezone (App service managed identity endpoint 2017-09-01): `"11/05/2021 15:18:31 +00:00"`

As it is not possible to update `msrestazure` to use the latest `2019-08-01` managed identity endpoint, because `msrestazure` has been out of maintenance. Azure CLI needs to be able to handle such inconsistency in order to support managed identity in App Service container.

## Testing Guide

1. In Azure Portal, create an app service with "quickstart" docker image `mcr.microsoft.com/appsvc/staticsite:latest`
2. Configure system assigned identity for the app service
3. Follow the doc to SSH into the container: https://docs.microsoft.com/en-us/azure/app-service/configure-linux-open-ssh-session

```
az webapp create-remote-connection --subscription 0b1f6471-1bf0-4dda-aec3-cb9272f09590 --resource-group mytest1 -n myweb2

ssh root@127.0.0.1 -p 43147

apt install git
git clone https://github.com/jiasli/azure-cli --branch webapp-mi --depth 1
apt-get install python3-venv
python3 -m venv py
. py/bin/activate
pip install -U pip
pip install azdev
azdev setup -c
az login --identity --debug
```

PR #20874, opened 2021-12-30
{Core} Fix `_normalize_expires_on` on Python 3.6

jiasli:
**Description**

Fix https://github.com/Azure/azure-cli/pull/20215#issuecomment-1002446176

Python 3.6 doesn't recognize timezone `%z` as `+00:00`. It only recognizes `+0000`:

```py
> docker run -it --rm python:3.6

>>> import datetime
>>> datetime.datetime.strptime('12/30/2021 07:53:52 +00:00', '%m/%d/%Y %H:%M:%S %z')
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.6/_strptime.py", line 565, in _strptime_datetime
    tt, fraction = _strptime(data_string, format)
  File "/usr/local/lib/python3.6/_strptime.py", line 362, in _strptime
    (data_string, format))
ValueError: time data '12/30/2021 07:53:52 +00:00' does not match format '%m/%d/%Y %H:%M:%S %z'

>>> datetime.datetime.strptime('12/30/2021 07:53:52 +0000', '%m/%d/%Y %H:%M:%S %z')
datetime.datetime(2021, 12, 30, 7, 53, 52, tzinfo=datetime.timezone.utc)
```

which means `az login --identity` with managed identity in App Service container still doesn't work with DEB, RPM packages, etc.

https://docs.python.org/3/library/datetime.html#technical-detail

> Changed in version 3.7: When the %z directive is provided to the strptime() method, the UTC offsets can have a colon as a separator between hours, minutes and seconds. For example, '+01:00:00' will be parsed as an offset of one hour. In addition, providing 'Z' is identical to '+00:00'.

There is indeed a test `azure.cli.core.auth.tests.test_adal_authentication.TestUtil.test_normalize_expires_on` to verify this:

https://github.com/Azure/azure-cli/blob/af483f52df81f6b14932272e11b84d90d23460d5/src/azure-cli-core/azure/cli/core/auth/tests/test_adal_authentication.py#L13-L14

However, the test is not executed by CI for Python 3.6 due to `azdev` issue

- https://github.com/Azure/azure-cli-dev-tools/issues/318

### DIFF
--- a/src/azure-cli-core/azure/cli/core/auth/adal_authentication.py
+++ b/src/azure-cli-core/azure/cli/core/auth/adal_authentication.py
@@ -96,7 +96,7 @@ def _normalize_expires_on(expires_on):
         # Python 3.6 doesn't recognize timezone as +00:00.
         # These lines can be dropped after Python 3.6 is dropped.
         # https://stackoverflow.com/questions/30999230/how-to-parse-timezone-with-colon
-        if ":" == expires_on[-3]:
+        if expires_on[-3] == ":":
             expires_on = expires_on[:-3] + expires_on[-2:]
 
         # Treat as datetime string "11/05/2021 15:18:31 +00:00"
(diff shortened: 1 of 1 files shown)


evelyn-ys (approved):

calvinhzy (approved):

yonzhan:
Core

PR #20879, opened 2021-12-31
{testsdk} Update @record_only decorator

FumingZhang:
**Description**

For a test case decorated by `@record_only`, to test it under **live mode**, its recording file **would be deleted** (by the init function in class `ReplayableTest`). Besides the warning message says “_This test is excluded from being run live. To force a recording, please remove the recording file._” According to my understanding, from the above message, a new recording file would be generated by **manually deleting** the recording file and then re-executing the test in the **live mode**. However, the actual situation is that the recording file was **automatically deleted**, and regenerating the recording file needs to perform the test in the **recording mode**.

I suppose there should be an option to control whether to delete the recording file automatically, so I opened this PR to add a parameter for the `@record_only` decorator and update the warning message. The changes in the PR will not affect the default behavior of the `@record_only` decorator already in use.

**Testing Guide**

**History Notes**

---


FumingZhang:
@jiasli @zhoxing-ms Could you please help review? Thanks in advance!

yonzhan:
testsdk

zhoxing-ms:
> For a test case decorated by @record_only, to test it under live mode, its recording file would be deleted (by the init function in class ReplayableTest). Besides the warning message says “This test is excluded from being run live. To force a recording, please remove the recording file.”

I believe this is a known issue about `@record_only` in the test framework. I vaguely remember that someone proposed to fix it before?
@jiasli @kairu-ms Could you please take a look at this PR?

src/azure-cli-testsdk/azure/cli/testsdk/scenario_tests/decorators.py
```diff
@@ -16,10 +16,13 @@ def live_only():
         'This is a live only test. A live test will bypass all vcrpy components.')
 
 
-def record_only():
+def record_only(keep_recording_file=False):
+    if keep_recording_file:
+        os.environ[ENV_KEEP_RECORDING_FILE] = 'True'
```
kairu-ms, line 21: If a previous test function set this environment True, when will it be reset as false? If it's not reset, it may pollute the following tests.

FumingZhang:
Legacy PR.

PR #20898, opened 2022-01-04
[App Service] BREAKING CHANGE : `az webapp up` : Remove default argument loading for `--plan`; Fix bug when not providing `--plan` argument

StrawnSC:
**Description**

The command `az webapp up` writes out to and reads from a `.azure/config` file in the current directory. This can cause confusing behavior where for the first time running `az webapp up` without a `--plan` arg, a new App Service Plan is created, but on subsequent runs of the command in the current directory, the plan is reused. There was no documentation of this feature in the help text or any alert via the command line. This can be particularly annoying when the user tries to create an app with a different OS from the default plan's OS (lead to a 'bad request' error). 
To reduce confusion, this default loading behavior was removed for `--plan` in this PR. 

If no `--plan` argument is provided, then the CLI will attempt to create the app in an App Service Plan with matching RG, SKU, location and OS type. I fixed a bug where the OS type was not included in this matching process. A new plan is created if no matching plan is found. 

**Testing Guide**

`azdev test --live test_webapp_up_commands`

**History Notes**

[App Service] BREAKING CHANGE : `az webapp up` : Remove default argument loading for `--plan`
[App Service] `az webapp up` : Fix bug causing existing ASP with wrong OS type to be selected when not providing a `--plan` argument

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -4054,7 +4054,6 @@ def webapp_up(cmd, name=None, resource_group_name=None, plan=None, location=None
     with ConfiguredDefaultSetter(cmd.cli_ctx.config, True):
         cmd.cli_ctx.config.set_value('defaults', 'group', rg_name)
         cmd.cli_ctx.config.set_value('defaults', 'sku', sku)
-        cmd.cli_ctx.config.set_value('defaults', 'appserviceplan', plan)
         cmd.cli_ctx.config.set_value('defaults', 'location', loc)
         cmd.cli_ctx.config.set_value('defaults', 'web', name)
     return create_json
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -3964,7 +3964,8 @@ def webapp_up(cmd, name=None, resource_group_name=None, plan=None, location=None
                                sku=sku,
                                create_rg=_create_new_rg,
                                resource_group_name=rg_name,
-                               plan=plan)
+                               plan=plan,
+                               is_linux=_is_linux)
     dry_run_str = r""" {
                 "name" : "%s",
                 "appserviceplan" : "%s",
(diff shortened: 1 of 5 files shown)


yonzhan:
App Service

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -4053,7 +4054,6 @@ def webapp_up(cmd, name=None, resource_group_name=None, plan=None, location=None
     with ConfiguredDefaultSetter(cmd.cli_ctx.config, True):
```
panchagnula, line 4054: Actually this whole block should be removed - since core-cli changed the logic on how to set local & global defaults.
 
& I also don't understand this fix - Plan like RG & webapp should have some defaults. However, whenever a user passes an argument for any of the parameters that should override the defaults. i.e values passed by user should override the defaults - otherwise we need a default for plan.

panchagnula (changes requested):
Don't understand the reason for Plan not to have defaults. Please remove the block for all but not for one argument over the others. RG, Plan & webapp should have similar experience.

Also, this implementation of local defaults is not consistent with the core-cli's approach anymore so might be good to ignore Setting & using the defaults this way & use the new approach.

StrawnSC:
I need to make significant changes to this PR (see below). I'll open a new one once these changes are ready 
- Re-enable default loading for `--plan`
- When using the default plan value (or a plan that already exists), validate that the OS type agrees with the OS type the user passes in
- In general, return the detailed error message from the server instead of raising a “Bad Request” error
- Include debug text indicating when values (for plan, RG, etc) are being written out as defaults for the current directory
- Add information to the help text describing how defaults are written and loaded as well as the command to change the defaults
- If possible, prevent default from being set if the command fails

src/azure-cli/azure/cli/command_modules/appservice/custom.py
StrawnSC: Are you sure this block should be removed? Whenever I remove this block, the defaults are no longer set, so it seems like it is setting the defaults. Do you know what the new way to set local defaults is?

Issue #20488, opened 2021-11-23
az storage fs directory upload with login authentication fails with MSI aut not yet supported


sgautrin:
## Describe the bug

**Command Name**
`az storage fs directory upload`

**Errors:**
```
The command failed with an unexpected error. Here is the traceback:
MSI auth not yet supported.
Traceback (most recent call last):
  File "/opt/az/lib/python3.6/site-packages/azure/cli/command_modules/storage/azcopy/util.py", line 108, in login_auth_for_azcopy
    token_info = _unserialize_non_msi_token_payload(token_info)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/command_modules/storage/azcopy/util.py", line 144, in _unserialize_non_msi_token_payload
    parsed_authority = urlparse(token_info['_authority'])
KeyError: '_authority'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/az/lib/python3.6/site-packages/knack/cli.py", line 231, in invoke
    cmd_result = self.invocation.execute(args)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 657, in execute
    raise ex
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 720, in _run_jobs_serially
    results.append(self._run_job(expanded_arg, cmd_copy))
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 712, in _run_job
    return cmd_copy.exception_handler(ex)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/command_modules/storage/__init__.py", line 371, in new_handler
    first(ex)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/command_modules/storage/__init__.py", line 370, in new_handler
    raise ex
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 691, in _run_job
    result = cmd_copy(params)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 328, in __call__
    return self.handler(*args, **kwargs)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/command_operation.py", line 121, in handler
    return op(**command_args)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/command_modules/storage/operations/azcopy.py", line 74, in storage_fs_directory_copy
[...]

PR #20917, opened 2022-01-06
[Storage] Stop supporting `--auth-mode login` for `az storage blob sync` and `az storage fs directory upload/download`

evelyn-ys:
**Backgrounds**
Storage module support several kinds of credential for data plane commands and `azcopy` commands:
1) `--account-key` with `--auth-mode key`(which can be omited)
2) `--connection-string`
3) `--sas-token`
4) `--auth-mode login`

`azcopy` has two kinds of auth:
a) SAS token auth
b) AAD auth

When customers provide 1) or 2), we generate a SAS token using the credential they give.
So with 1) 2) 3), we call `azcopy` by SAS token auth.
With 4), we call `azcopy` by AAD auth with login `access_token` and `refresh_token` stored in environment variable.

**Problems**
- After CLI core migrates from ADAL to MSAL, we can't get `refresh_token` any more. `azcopy` related commands fail with `--auth-mode login` (#20488)
- Setting `access_token` and `refresh_token` in env virable is unsafe.

**Changes**
This PR drops `--auth-mode login` support for `azcopy` related commands

PS. Some existing `azcopy` related commands don't support `--auth-mode login` from the very beginning, such as `az storage copy/remove`

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/storage/commands.py
+++ b/src/azure-cli/azure/cli/command_modules/storage/commands.py
@@ -426,7 +426,7 @@ def load_command_table(self, _):  # pylint: disable=too-many-locals, too-many-st
 
     with self.command_group('storage blob', command_type=block_blob_sdk,
                             custom_command_type=get_custom_sdk('azcopy', blob_data_service_factory)) as g:
-        g.storage_custom_command_oauth('sync', 'storage_blob_sync', is_preview=True)
+        g.storage_custom_command('sync', 'storage_blob_sync', is_preview=True)
 
     with self.command_group('storage container', command_type=block_blob_sdk,
                             custom_command_type=get_custom_sdk('blob', blob_data_service_factory)) as g:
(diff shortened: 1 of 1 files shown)


calvinhzy (approved):

jiasli:
I agree `--auth-mode` should be removed for now.

Actually, `azcopy` can support single-sign-on via MSAL and MSAL extensions. In that way, Azure CLI can call `azcopy` with something like

```
az copy --token-cache ~/.azure/msal_token_cache.json --username myname@microsoft.com
```

but I don't think this is currently on `azcopy`'s roadmap.

jiasli (approved):

jiasli:
To give a little bit more info: unlike ADAL, **MSAL now doesn't return refresh token** via `acquire_token_silent_with_error`, which makes the original "we call `azcopy` by AAD auth with login `access_token` and `refresh_token` stored in environment variable" approach invalid.

PR #20935, opened 2022-01-09
[CDN] Fix test failures in test_custom_domain_scenarios.py caused by key vault authentication challenge cache

t-bzhan:
**Description**

Currently, `azdev test cdn` command will fail with the message:

> KeyVaultErrorException('(Unauthorized) Request is missing a Bearer or PoP token.')

The issue is caused by challenge cache conflict  in the key vault authentication process,  test_cdn_custom_domain_byoc_latest and  test_cdn_custom_domain_https_msft used to use the same key vault and they share the challenge cache. When they are launched in the same process, the latter one will reuse the challenge cache and that will cause the previously recorded challenge request to be treated as a regular request in VCR replay model.

image

This PR aims to use different key vault prefixes to bypass the challenge cache issue.

**Testing Guide**

azdev test cdn

**History Notes**

---

- [ x] I adhere to the [Error Handling Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/error_handling_guidelines.md).

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/cdn/tests/latest/byoc_cert_policy.json
+++ b/src/azure-cli/azure/cli/command_modules/cdn/tests/latest/byoc_cert_policy.json
@@ -33,6 +33,16 @@
             "keyCertSign"
         ],
         "subject": "CN=CLIGetDefaultPolicy",
+        "subjectAlternativeNames": {
+            "dnsNames": [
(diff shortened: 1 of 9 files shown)


yonzhan:
CDN

kairu-ms (approved):

PR #20955, opened 2022-01-12
[App Service] `az staticwebapp` : Fix #20830 and remove unneeded properties from staticwebapp output

### LINKED ISSUE #20830: Webapp:staticwebapp - create defaults are invalid
**Describe the bug**

`az staticwebapp create` currently defaults to the following locations:
https://github.com/Azure/azure-cli/blob/86ffba42f91a9082743237286d63d89ef6e76ce4/src/azure-cli/azure/cli/command_modules/appservice/static_sites.py#L343

`output_location` is incorrect. It should default to empty string.

I think the app and api locations should probably default to empty strings as well. @miwebst @simonaco What do you think?

**To Reproduce**

Run `az staticwebapp create` without supplying `--output-location`.

**Expected behavior**

The workflow file created should have an empty `output_location`.

/cc @StrawnSC

StrawnSC:
**Description**

Removes the following properties from the staticwebapp objects returned from the create, show, list, and update commands: 

- allowConfigFileUpdates
- buildProperties
- contentDistributionEndpoint
- keyVaultReferenceIdentity
- kind                                                                                           
- privateEndpointConnections
- repositoryToken
- stagingEnvironmentPolicy
- templateProperties
- sku.capabilities
- sku.capacity
- sku.family
- sku.locations
- sku.size
- sku.skuCapacity

Also changes some default arguments for `az staticwebapp create` to fix #20830

**Testing Guide**

**History Notes**

 
[App Service] `az staticwebapp create` : Change default output location and API location to `None`. Change default app location to "/". Remove unnecessary properties from output 
[App Service] `az staticwebapp show` : Remove unnecessary properties from output 
[App Service] `az staticwebapp list` : Remove unnecessary properties from output 
[App Service] `az staticwebapp update` : Remove unnecessary properties from output 

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/static_sites.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/static_sites.py
@@ -7,7 +7,7 @@ from azure.cli.core.commands.client_factory import get_mgmt_service_client
 from azure.cli.core.util import sdk_no_wait
 
 from azure.cli.core.commands import LongRunningOperation
-from azure.cli.core.azclierror import ValidationError, RequiredArgumentMissingError
+from azure.cli.core.azclierror import ResourceNotFoundError, ValidationError, RequiredArgumentMissingError
 from knack.util import CLIError
 from knack.log import get_logger
 from msrestazure.tools import parse_resource_id
@@ -163,7 +163,7 @@ def delete_staticsite_domain(cmd, name, hostname, resource_group_name=None, no_w
 
 
 def show_identity(cmd, resource_group_name, name):
-    return show_staticsite(cmd, name, resource_group_name).identity
+    return show_staticsite(cmd, name, resource_group_name, format_output=False).identity
 
 
 def assign_identity(cmd, resource_group_name, name, assign_identities=None, role='Contributor', scope=None):
@@ -367,7 +367,7 @@ def update_staticsite_users(cmd, name, roles, authentication_provider=None, user
 
 def create_staticsites(cmd, resource_group_name, name, location,  # pylint: disable=too-many-locals
                        source, branch, token=None,
-                       app_location='.', api_location='.', output_location=None,
+                       app_location="/", api_location=None, output_location=None,
                        tags=None, no_wait=False, sku='Free', login_with_github=False, format_output=True):
     from azure.core.exceptions import ResourceNotFoundError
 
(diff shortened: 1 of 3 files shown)


yonzhan:
App Service

wangzelin007 (approved):

StrawnSC:
Pushing out to next release so we can resolve the issues raised by https://github.com/Azure/azure-cli/issues/20830#issuecomment-1011423996

panchagnula:
@StrawnSC was the ask to remove these from all commands? I would remove this from create response but keep it in show -  otherwise what is the way for users to get back these properites.

src/azure-cli/azure/cli/command_modules/appservice/static_sites.py
```diff
@@ -19,21 +19,48 @@
 logger = get_logger(__name__)
 
 
-def list_staticsites(cmd, resource_group_name=None):
+# remove irrelevant attributes from staticsites printed to the user
+def _format_staticsite(site, format_site=False):
+    if format_site:
```
panchagnula, line 24: Remove these from create response only, not the Show & list - so users have a way to retrieve all the properties using CLI, if needed.

panchagnula (changes requested):
See comments.

src/azure-cli/azure/cli/command_modules/appservice/static_sites.py
StrawnSC: @simonaco originally requested this because the cli "exposes some of our internal architecture that we wouldn’t want customers to interact with" -- so presumably this could apply to list and show as well. 

@simonaco do you have thoughts on this? Were you wanting us to format the output for the show/list command or just the create one?

panchagnula: @StrawnSC lets remove this for create only - this will be considered breaking change for CLI - & as CLI owners we should keep these sorta breaking changes to a minimum. Ideally API should remove these properties if these shouldn't be presented to users. Since CLI is not the only client that would surface this. Thanks

zhoxing-ms:
@panchagnula Can we merge this PR in this sprint?

panchagnula (approved):

zhoxing-ms (approved):

Issue #5105, opened 2017-12-14
MSI: move msi commands to its own command module in knack style


yugangw-msft:
So far, I put the `az identity` command tree inside `azure-cli-vm` command module as i want to avoid the on-going knack conversion. But after the knack release, i will do the work for the new command module. The whole thing is transparent to end users

yugangw-msft:
@tjprescott @williexu, is it worthwhile to do this for 5 tiny commands? I can go either way.

tjprescott:
I suppose it should be done, but there's really no urgency around it.

adewaleo:
@yugangw-msft @tjprescott, it is indeed 5 commands, do we want to move the commands into their own new module?

tjprescott:
@adewaleo @yugangw-msft instead of putting them in their own module, would it make sense to put them in a module that already exists?  For example `azure-cli-role` has to do with authentication, which is what MSI is.  I'd prefer not to put them in a stand-alone module, but if there's a more logical place for them to live besides the VM module that sounds good to me.

yugangw-msft:
yes, a new command module would be too heavy and expensive.  I agree we move them to `azure-cli-role`. MSI essentially is a service principal as a resource. So this makes sense.

jiasli:
As now Azure CLI has command index to accelerate the module load time, moving `az identity` to a separate command module will actually make it faster.

Also, `az ad` has become more and more heavy with Microsoft Graph, including `az identity` in `role` command module also introduces difficulties for command coverage computation.

PR #21041, opened 2022-01-19
{Identity} Move `az identity` command group from `role` to `identity` command module

jiasli:
**Description**

Implemented according to https://github.com/Azure/azure-cli/issues/5105

Move `az identity` command group from `role` to `identity` command module.

Reasons:
- Command test coverage shows `az identity` is not tested. Moving `az identity` command group to a dedicated command module makes adding tests easier.
- Code owners can be assigned more easily and clearly.
- `az identity` no longer bonds to `RoleCommandsLoader` which uses `ResourceType.MGMT_AUTHORIZATION` as `resource_type`: https://github.com/Azure/azure-cli/blob/9c4b39efaea4a94a4f8187e9562e453962adc214/src/azure-cli/azure/cli/command_modules/role/__init__.py#L17-L19
   Using `ResourceType.MGMT_AUTHORIZATION` which is for role assignments/definitions for `az identity` is simply wrong.
- `az identity` uses dedicated `azure-mgmt-msi` SDK, so it is reasonable for it to have its own command module.
- After the introduction of command index, adding command modules is no longer a heavy burden (https://github.com/Azure/azure-cli/issues/5105#issuecomment-1015054927).


yonzhan:
Identity

src/azure-cli/azure/cli/command_modules/role/commands.py
```diff
@@ -209,10 +203,3 @@ def load_command_table(self, _):
         g.command('add', 'add_member')
         g.command('remove', 'remove_member')
         g.custom_command('check', 'check_group_membership', client_factory=get_graph_client_groups)
-
-    with self.command_group('identity', identity_sdk, min_api='2017-12-01') as g:
```
jiasli, line 213: `azure-mgmt-msi==0.2.0` doesn't support multi-API at all. See https://github.com/Azure/azure-sdk-for-python/tree/azure-mgmt-msi_0.2.0/azure-mgmt-msi

This `min_api='2017-12-01'` works because it is checked against `ResourceType.MGMT_AUTHORIZATION` specified for the whole `role` module:

https://github.com/Azure/azure-cli/blob/9c4b39efaea4a94a4f8187e9562e453962adc214/src/azure-cli/azure/cli/command_modules/role/__init__.py#L17-L19

This is certainly wrong.

After removing `min_api='2017-12-01'`, `az identity` will work in `xxx-hybrid` profiles.

doc/sphinx/azhelpgen/doc_source_map.json
```diff
@@ -1,81 +1,82 @@
...
     "eventgrid": "src/azure-cli/azure/cli/command_modules/eventgrid/_help.py",
+    "eventhubs": "src/azure-cli/azure/cli/command_modules/eventhubs/_help.py",
     "extension": "src/azure-cli/azure/cli/command_modules/extension/_help.py",
     "feature": "src/azure-cli/azure/cli/command_modules/resource/_help.py",
+    "feedback": "src/azure-cli/azure/cli/command_modules/feedback/_help.py",
     "find": "src/azure-cli/azure/cli/command_modules/find/_help.py",
+    "functionapp": "src/azure-cli/azure/cli/command_modules/appservice/_help.py",
     "group": "src/azure-cli/azure/cli/command_modules/resource/_help.py",
     "hdinsight": "src/azure-cli/azure/cli/command_modules/hdinsight/_help.py",
+    "identity": "src/azure-cli/azure/cli/command_modules/identity/_help.py",
```
jiasli, line 42: In addition to adding `identity` entry, lines in this file are also sorted.

zhoxing-ms (approved):

wangzelin007 (approved):

evelyn-ys (approved):

src/azure-cli/azure/cli/command_modules/identity/_help.py
```diff
@@ -0,0 +1,21 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+from knack.help_files import helps
+
+helps['identity'] = """
+type: group
+short-summary: Managed Service Identities
```
jiasli, line 10: @zhoxing-ms, we should update the doc because the term MSI is now deprecated. 

https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview

> Managed identities for Azure resources is the new name for the service formerly known as Managed Service Identity (MSI).

kairu-ms (approved):

src/azure-cli/azure/cli/command_modules/identity/_help.py
zhoxing-ms: OK~

Issue #19767, opened 2021-10-03
az iot hub monitor-events fails on a Docker container.


TORIFUKUKaiou:
**Describe the bug**
- `az iot hub monitor-events` fails on a Docker container.
- https://docs.microsoft.com/en-us/cli/azure/run-azure-cli-docker?view=azure-cli-latest

**To Reproduce**

```
$ docker pull mcr.microsoft.com/azure-cli
$ docker run -it mcr.microsoft.com/azure-cli

bash-5.1# az login

bash-5.1# connection_string='HostName=xxxx.azure-devices.net;SharedAccessKeyName=iothubowner;SharedAccessKey=XXXXXXXXXXXXXXXXXXXXX='

bash-5.1# az iot hub monitor-events --login $connection_string

The command requires the extension azure-iot. Do you want to install it now? The command will continue to run after the extension is installed. (Y/n): y
Run 'az config set extension.use_dynamic_install=yes_without_prompt' to allow installing extensions without prompt.
Dependency update (uamqp 1.2) required for IoT extension version: 0.10.17. 
Continue? (y/n) -> y
```

I got Failure updating uamqp.

```
Updating required dependency...
Failure updating uamqp. Aborting...
Collecting uamqp~=1.2
  Downloading uamqp-1.4.2.tar.gz (4.4 MB)
Collecting certifi>=2017.4.17
  Downloading certifi-2021.5.30-py2.py3-none-any.whl (145 kB)
Collecting six~=1.0
  Downloading six-1.16.0-py2.py3-none-any.whl (11 kB)
Building wheels for collected packages: uamqp
  Building wheel for uamqp (setup.py): started
  Building wheel for uamqp (setup.py): finished with status 'error'
  ERROR: Command errored out with exit status 1:
[...]

PR #21051, opened 2022-01-19
{Packaging} Add `gettext` package to Docker image

alersrt:
**Description**

There is no possibility to easily pass environment variables from CI/CD environment to the configs like ARM or YAML templates. However there are some kinds of workarounds based on templates parsing and replacing env vars placeholder by corresponding values. The most convenient way to do it is to use `envsubst`.

As an example we can consider creating the Container Instance from YAML `template.yaml`: 

```yaml
...
properties: # Properties of container group
  ...
  containers: # Array of container instances in the group
    - name: ${CONTAINER_NAME} # Name of an instance
      properties: # Properties of an instance
        image: ${USED_IMAGE} # Container image used to create the instance
       ...
        environmentVariables:
          - name: ENV_VAR_1
            value: ${VALUE_1}
         ...

```

If we will just use the common case:

```bash
az container create -g ContainerInstanceGroup -f template.yaml
```

We will see that value of the `properties.containers[0].name` path will have value `${CONTAINER_NAME}` and it is not the result what we obviously want. As a workaround we can replace env vars placeholders by corresponding values with the next pattern:

```bash
cat template.yaml | envsubst > config.yaml
az container create -g ContainerInstanceGroup -f config.yaml
```

So the `config.yaml` will have correct values instead placeholders.

The `envsubst` util is a part of the `gettext` package and it could be provided in Docker image to allow using this image in CI/CD pipelines without any additional steps like installing this package inside CI/CD pipeline configs or building own images with this package.

**Testing Guide**

[Here][1] you can find the built image.

**History Notes**

{Packaging} Dockerfile: Add `gettext` package.

---

[1]: https://hub.docker.com/repository/docker/userhome/azure-cli

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,6 +38,7 @@ RUN apk add --no-cache bash openssh ca-certificates jq curl openssl perl git zip
  && apk add --no-cache --virtual .build-deps gcc make openssl-dev libffi-dev musl-dev linux-headers \
  && apk add --no-cache libintl icu-libs libc6-compat \
  && apk add --no-cache bash-completion \
+ && apk add --no-cache gettext \
  && update-ca-certificates
 
 ARG JP_VERSION="0.1.3"


yonzhan:
Packaging

jiasli:
Hi @alersrt, thanks a lot for the contribution. Previous there was an attempt to add other libs to the docker image (https://github.com/Azure/azure-cli/issues/19767#issuecomment-954598232), but this will make the docker image even bigger.

Without absolute needs, we add libs to the docker image as less as possible. For now, we would recommend manually installing `gettext` during your script execution, or building your own docker image with Azure CLI docker image as the base image.

Thanks for your understanding.

alersrt:
@jiasli Thanks for your answer! Yes, I understand, so its not a trouble.

PR #21057, opened 2022-01-20
[App Service] BREAKING CHANGE: `az webapp list-runtimes`: Use new stacks APIs

StrawnSC:
**Description**

Changes the commands `az webapp up`, `az webapp create` and `az webapp list-runtimes` to use web app stacks from the API rather than a hardcoded list.
Also does the same for `az functionapp create` and creates a new command `az functionapp list-runtimes`. Note that the command `az functionapp list-runtimes` has more verbose output that it's webapp equivalent because function app stacks availability depends on the functions version (v3, v4, etc). 

Unrelated to the other changes, this also uses the `:` delimiter for `az webapp list-runtimes` instead of `|` to avoid PowerShell issues. 

Currently*, the webapp stacks that will no longer be supported are as follows: 
- windows: "node|10.6", "node|10.10", "node|10.14", "java|7|Tomcat|7.0", "java|7|Tomcat|8.5", "java|7|Tomcat|9.0", "java|7|Java SE|8", "DOTNETCORE|2.1"
- linux: "DOTNETCORE|2.1", "NODE|10.1",  "NODE|10.6", "NODE|10.14", "JBOSSEAP|7.3-java8", "JBOSSEAP|7.3-java11"

*These lists will change over time since the values will be set by the API. 

**Example Usage**
`az webapp list-runtimes` output using no arguments:
Screen Shot 2022-02-08 at 8 38 35 PM

`az webapp list-runtimes` output using `--os` argument: 
Screen Shot 2022-02-08 at 8 40 00 PM

`az functionapp list-runtimes` output without an `--os` argument:
Screen Shot 2022-02-01 at 11 19 55 AM

`az functionapp list-runtimes` output with an `--os` argument:
Screen Shot 2022-02-01 at 11 20 35 AM

**Testing Guide**

`azdev test --live test_webapp_up_commands test_webapp_commands`
`azdev test --live test_functionapp_commands`

**History Notes**

[App Service] BREAKING CHANGE: `az webapp up` : Change supported runtimes 
[App Service] BREAKING CHANGE: `az webapp create` : Change supported runtimes 
[App Service] BREAKING CHANGE: `az webapp list-runtimes` : Add `--os`/`--os-type` argument, change runtimes, change default behavior to return both linux and windows stacks, and deprecate `--linux` argument
[App Service] BREAKING CHANGE: `az functionapp create` : Take runtime names and versions from API instead of hardcoded list 
[App Service] `az functionapp list-runtimes` : Add new command to show function app runtimes, versions, and compatible functions versions 

---


yonzhan:
App Service

panchagnula:
@calvinsID could you review this please? Thanks!

calvinsID:
Awesome, great to see this thanks @StrawnSC! Will review shortly

StrawnSC:
> Awesome, great to see this thanks @StrawnSC! Will review shortly

No prob! By the way, I've gotten the automated tests to pass on this, but I'm still working through some manual testing of `az webapp up` just to make sure everything's good -- if any issues come up I'll have to make some changes to resolve them

StrawnSC:
Converting to a draft until I'm done with changes

src/azure-cli/azure/cli/command_modules/appservice/_params.py
```diff
@@ -170,7 +170,8 @@ def load_arguments(self, _):
         c.argument('slot', options_list=['--slot', '-s'], help='Name of the web app slot. Default to the productions slot if not specified.')
 
     with self.argument_context('webapp list-runtimes') as c:
-        c.argument('linux', action='store_true', help='list runtime stacks for linux based web apps')
+        c.argument('linux', options_list=['--linux', '-l'], action='store_true', help='list runtime stacks for linux based web apps')
+        c.argument('windows', options_list=['--windows', '-w'], action='store_true', help='list runtime stacks for windows based web apps')
```
panchagnula, line 174: @seligj95  & @StrawnSC was this the recommendation from the spec. Instead of adding --windows , can we add --os which will take linux or windows as values & mark --linux as something that to be deprecated in the future . This will ensure simplicity in the usage of command & matches other commands like create & up where we use --os.
Marking the --linux as to be deprecated in future will ensure this is not a total breaking change wrt command signature & we can eventually remove this. Thoughts?

StrawnSC: yeah I think it would make sense to have similar usage to `az webapp up`, which uses an `--os-type` parameter. So that argument would be not required and the default behavior would be to show linux and windows stacks, right?

panchagnula: Correct --os would be optional, exemption of this, default should be both linux & windows. Thanks!

seligj95: agreed with Sisira's thoughts. we should deprecate --linux following the standard process

src/azure-cli/azure/cli/command_modules/appservice/tests/latest/test_webapp_commands.py
```diff
@@ -967,6 +969,7 @@ def test_linux_webapp(self, resource_group):
 
 
 class LinuxWebappSSHScenarioTest(ScenarioTest):
+    @unittest.skip("Flaky test")
```
zhoxing-ms: Why do we skip this test?

zhoxing-ms:
@StrawnSC At present, this PR is not able to catch up with the release of this sprint. We plan to postpone it to the release of next sprint (03-01). Do you think this date is acceptable?

StrawnSC:
> @StrawnSC At present, this PR is not able to catch up with the release of this sprint. We plan to postpone it to the release of next sprint (03-01). Do you think this date is acceptable?

@panchagnula is this timeline okay?

src/azure-cli/azure/cli/command_modules/appservice/tests/latest/test_webapp_commands.py
StrawnSC: This test is flaky and should possibly be removed or rewritten at some point. When I reran the test in live mode, it failed in both this branch and on dev

StrawnSC:
@zhoxing-ms we are good with moving this out to the 2022-03-01 release

StrawnSC:
Converting this to a draft to add an equivalent function app command

StrawnSC:
@zhoxing-ms I'm not sure why the CI tests are failing on the "Build Docker Image" check. Do you know what's going on there?

src/azure-cli/azure/cli/command_modules/appservice/_create_util.py
```diff
@@ -75,31 +72,31 @@ def zip_contents_from_dir(dirPath, lang):
     return zip_file_path
 
 
-def get_runtime_version_details(file_path, lang_name):
+def get_runtime_version_details(file_path, lang_name, is_linux=False, stack_helper=None):
```
calvinsID, line 75: nit: should we make stack_helper a required parameter?

calvinsID:
Thanks so much for doing this, glad to see we're off the hardcoded lists :D

calvinsID (approved):

src/azure-cli/azure/cli/command_modules/appservice/_create_util.py
StrawnSC: yeah that's a good idea. It would just throw an attribute error if `stack_helper` was `None`
While I'm at it I'll also remove the function app stacks JSON file, like I did with the web app stacks file

StrawnSC:
> Thanks so much for doing this, glad to see we're off the hardcoded lists :D

Np! By the way, this may cause some issues with the command `az webapp deployment github-actions add`, although it seems the command was already broken: https://github.com/Azure/azure-cli/issues/20808

StrawnSC:
Also, @panchagnula and @calvinsID please let me know what you think of the output for the new command `az functionapp list-runtimes` (see below). It's more verbose than the webapp equivalent, but this is because:
- the user needs to specify the runtime and the runtime version on different parameters in `az functionapp create`, unlike webapp up/create (which just uses `--runtime <runtime>|<version>`)
- the user needs to pass in the right functions version (ie v3, v4) to `az functionapp create`, and certain stacks are only supported on certain functions versions 

Screen Shot 2022-02-01 at 11 20 35 AM

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -4792,7 +5027,7 @@ def add_github_actions(cmd, resource_group, name, repo, runtime=None, token=None
     else:
         file_name = "{}_{}.yml".format(branch.replace('/', '-'), name.lower())
     dir_path = "{}/{}".format('.github', 'workflows')
-    file_path = "/{}/{}".format(dir_path, file_name)
```
calvinsID, line 4795: Should we also remove the slash in `remove_github_actions` function

panchagnula:
> Also, @panchagnula and @calvinsID please let me know what you think of the output for the new command `az functionapp list-runtimes` (see below). It's more verbose than the webapp equivalent, but this is because:
> 
> * the user needs to specify the runtime and the runtime version on different parameters in `az functionapp create`, unlike webapp up/create (which just uses `--runtime <runtime>|<version>`)
> * the user needs to pass in the right functions version (ie v3, v4) to `az functionapp create`, and certain stacks are only supported on certain functions versions
> 
> If you think there's a better way to show this info, please let me know
> 
> Screen Shot 2022-02-01 at 11 20 35 AM

the response looks good to me for a single os? How does this look for all OS's do we still drop by OS?

panchagnula:
@StrawnSC & @calvinsID ,Regarding DIR path & file paths handling - whatever we do should work across windows/ Linux/ all OS file systems. Expecting the path to be a certain format is incorrect. Whatever fixes you do here ensure - doesn't cause any regressions - I recommend keeping this change targeted to Stacks API only & use another PR,, for unrelated changes/ fixes.

calvinsID:
> > Thanks so much for doing this, glad to see we're off the hardcoded lists :D
> 
> Np! By the way, this may cause some issues with the command `az webapp deployment github-actions add`, although it seems the command was already broken: #20808

Thanks for making the fix for removing the "/". If there are further fixes needed for gh actions (especially if the workflow file needs any changes) I'm thinking we can do those in a later PR

StrawnSC:
@panchagnula here's the output without an `--os` argument. It's cut off, but lower down there's a section for windows stacks. It's essentially the same as the webapps command in how it shows the stacks by OS 

Screen Shot 2022-02-01 at 11 19 55 AM

src/azure-cli/azure/cli/command_modules/appservice/custom.py
StrawnSC: leaving the other one as is per Sisira's comment on keeping this PR focused on stacks

zhoxing-ms:
>  I'm not sure why the CI tests are failing on the "Build Docker Image" check. Do you know what's going on there?

@StrawnSC This is a known issue https://github.com/jmespath/jmespath.terminal/issues/19 related to `jmespath-terminal`, we will plan to drop the dependence on it.

panchagnula (approved):
LGTM

src/azure-cli/azure/cli/command_modules/appservice/_params.py
```diff
@@ -170,8 +172,8 @@ def load_arguments(self, _):
         c.argument('slot', options_list=['--slot', '-s'], help='Name of the web app slot. Default to the productions slot if not specified.')
 
     with self.argument_context('webapp list-runtimes') as c:
-        c.argument('linux', options_list=['--linux', '-l'], action='store_true', help='list runtime stacks for linux based web apps')
```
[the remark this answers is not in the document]
StrawnSC: This is marked as deprecated: 
Screen Shot 2022-02-08 at 2 16 38 PM

I think you may be reviewing an old version of the PR

src/azure-cli/azure/cli/command_modules/appservice/_params.py
```diff
@@ -170,8 +172,8 @@ def load_arguments(self, _):
         c.argument('slot', options_list=['--slot', '-s'], help='Name of the web app slot. Default to the productions slot if not specified.')
 
     with self.argument_context('webapp list-runtimes') as c:
```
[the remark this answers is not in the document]
StrawnSC: The functionapp equivalent has the same params as the webapp one, besides the deprecated `--linux` param (since the functionapp command was newly added)

Screen Shot 2022-02-08 at 2 18 26 PM

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2804,13 +2811,31 @@ def _match_host_names_from_cert(hostnames_from_cert, hostnames_in_webapp):
...
+
+    def __init__(self, cmd, linux=False, windows=False):
         self._cmd = cmd
-        self._client = client
+        self._client = web_client_factory(cmd.cli_ctx, api_version="2021-01-01")
         self._linux = linux
+        self._windows = windows
         self._stacks = []
 
+        # TODO try and get API support for this so it isn't hardcoded
```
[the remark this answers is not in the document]
StrawnSC: This was in the previous implementation of the runtimes handling (see [here](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/appservice/custom.py#L2896)). Windows runtimes are not set with a single attribute like linux runtimes (which uses `linuxFxVersion`), so we have to set them with the proper appsetting key and value. For function apps, the API tells you what appsettings to set, but the webapps API does not  

This is used only for `az webapp up` and `az webapp create`, so if the user deletes the appsetting after the creation of the webapp it will not impact either command.

As far as error handling, the appsetting is only set if the runtime has an entry in the config mapping, so there's no exception if a runtime doesn't have a value in the dict.

StrawnSC:
@zhoxing-ms thanks for the help with the CI issues. Was the credential scan check changed recently? It was passing last week on the test recordings, but now it appears to be failing

zhoxing-ms:
@StrawnSC This is a problem caused by credscan package `Microsoft.Security.CredScan` upgrade. We have submitted a PR #21230 to pin it

StrawnSC:
@zhoxing-ms any chance we could get this merged soon? @panchagnula has another big upcoming change that's blocked until this is in dev

Issue #1073, opened 2016-10-12
[Storage] Resource group should be optional or omitted for most storage management commands


tjprescott:
Since storage accounts have globally unique names, there is no reason you should have to specify a resource group name when performing storage account management operations (except create). In KeyVault, the vault is found using only the name provided it is contained within the subscription. If resource group IS specified, it is treated as a filter (i.e. perform this operation only if the vault with X name is in Y resource group).

johanste:
I'm a bit concerned about this since the storage account name isn't *that* globally unique (it is only unique for a given endpoint for one - I'm pretty sure you can have a storage account with the same name on stack as you have on public azure). Also, there is nothing really contractual about the name being unique - it happens to form the basis of the URLs to access the storage account, but I don't think it would be a breaking change if (hypothetically), a new storage account type used a different suffix than the current one.

tjprescott:
But it is like you said--you don't need the resource group name to identify the storage account--you need the endpoint URL (of which RG isn't a part). Since KeyVault and storage access their resources in the same way (URLs) then it would seem that they should behave consistently.

Right now, RG is optional for KeyVault (in our CLI but also Xplat) and specifying it only makes a request more likely to fail, whereas RG is required for storage. It seems they should behave the same, one way or the other.

derekbekoe:
@williexu

PR #21067, opened 2022-01-20
[App Config] `az appconfig kv import`: Add new parameter `--strict` to support strict import.

pratiksanglikar:
**Description**
`az appconfig kv import` command now supports additional flag `--strict` when importing from a file. When using `--strict` mode, any keyvalues with specific label would be deleted from the configuration store if they are missing from the imported file.

**Testing Guide**
`az appconfig kv import -s file --format json --path test.json -n test --strict --label label1`
This command would import all keyvalues from `test.json` into configuration store `test` and delete all keyvalues with label `label1` that are **missing** from `test.json` from the config store.

**History Notes**
[App Config] `az appconfig kv import`: Add new parameter `--strict` to support strict import.
---


pratiksanglikar:
@avanigupta @shenmuxiaosen please take a look.

yonzhan:
App Config

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -628,9 +635,9 @@ def __print_features_preview(old_json, new_json):
     return True
 
 
-def __print_preview(old_json, new_json):
+def __print_preview(old_json, new_json, strict):
```
avanigupta, line 638: >strict [](http://example.com/codeflow?start=40&length=6)

need a default value for strict argument because this method is invoked during export operation too.

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -101,7 +102,7 @@ def import_config(cmd,
                                       depth=depth,
                                       content_type=content_type)
 
-        if not skip_features:
+        if strict or not skip_features:
```
avanigupta, line 105: >strict or  [](http://example.com/codeflow?start=11&length=10)

So we cant skip features in strict mode?

avanigupta: Oh I remember we discussed not supporting both options together because the usage will be ambiguous.

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -136,21 +137,22 @@ def import_config(cmd,
         src_kvs = __read_kv_from_app_service(
             cmd, appservice_account=appservice_account, prefix_to_add=prefix, content_type=content_type)
 
-    # if customer needs preview & confirmation
-    if not yes:
+    if strict or not yes:
         # fetch key values from user's configstore
         dest_kvs = __read_kv_from_config_store(azconfig_client,
```
avanigupta, line 142: >dest_kvs [](http://example.com/codeflow?start=8&length=8)

For strict mode, we should query `dest_kvs` with the supplied `prefix`, not `SearchFilterOptions.ANY_KEY`. For preview, its okay to fetch all keys, but for deleting, we only want to delete those keys that match the given prefix.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -582,9 +583,9 @@ def __serialize_kv_list_to_comparable_json_list(keyvalues, profile=None):
     return res
 
 
-def __print_features_preview(old_json, new_json):
+def __print_features_preview(old_json, new_json, strict):
```
avanigupta, line 586: >strict [](http://example.com/codeflow?start=49&length=6)

Same here - need a default value

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -1044,15 +1057,20 @@ def __import_kvset_from_file(client, path, yes):
         if __validate_import_config_setting(config_setting):
             kvset_to_import.append(config_setting)
 
+    existing_kvset = __read_kv_from_config_store(client,
+                                                 key=SearchFilterOptions.ANY_KEY,
```
avanigupta, line 1061: >=SearchFilterOptions.ANY_KEY, [](http://example.com/codeflow?start=52&length=29)

Same here, for deleting, we need to match prefix too

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -1044,15 +1057,20 @@ def __import_kvset_from_file(client, path, yes):
...
+                                                 label=SearchFilterOptions.ANY_LABEL)
```
shenmuxiaosen, line 1062: This should be under ``` if strict```

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -171,6 +174,13 @@ def import_config(cmd,
     # append all feature flags to src_kvs list
     src_kvs.extend(src_features)
 
+    # In strict mode, delete kvs with specific label that are missing from the imported file
+    if strict:
+        kvs_to_delete = list(filterfalse(lambda kv: any(kv_import.key == kv.key and label == kv.label
+                                                        for kv_import in src_kvs), dest_kvs))
```
avanigupta, line 180: >dest_kvs [](http://example.com/codeflow?start=83&length=8)

If user wants preview, we call `__discard_features_from_retrieved_kv(dest_kvs)`, which removes all features from `dest_kvs`. So feature flags will not be deleted in strict mode.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -125,6 +126,7 @@ def load_arguments(self, _):
         # bypass cli allowed values limitation
         c.argument('separator', validator=validate_separator, help="Delimiter for flattening the json or yaml file to key-value pairs. Separator will be ignored for property files and feature flags. Supported values: '.', ',', ';', '-', '_', '__', '/', ':' ")
         c.argument('profile', validator=validate_import_profile, arg_type=get_enum_type([ImportExportProfiles.DEFAULT, ImportExportProfiles.KVSET]), help="Import profile to be used for importing the key-values. Options 'depth', 'separator', 'content-type', 'label', 'skip-features' and, 'prefix' are not supported when using '{}' profile.".format(ImportExportProfiles.KVSET))
+        c.argument('strict', validator=validate_strict_import, arg_type=get_three_state_flag(), help="Delete all other key-values in the store with specified prefix and label", is_preview=True)
```
shenmuxiaosen, line 129: >Delete all other key-values in the store with specified prefix and label [](http://example.com/codeflow?start=102&length=72)

I am missing some context here. If strict mode is on, then we will delete all other kvs except for the newly imported ones?

pratiksanglikar: That's partially correct. We will delete all other kvs **with specified prefix and label** except newly imported ones.
i.e.

Consider the following kvs are present in the config store - 

| Key | Label | Value |
|----|----|----|
| TestApp:BackgroundColor | label1 | Blue |
| TestApp:BackgroundColor | label2 | Red |
| TestApp:Font | label1 | 50 |
| OtherKey | None | OtherValue |

And the file contains following keyvalues - 
| Key | Label | Value |
|----|----|----|
| TestApp:BackgroundColor | label1 | Green |

When user imports using `strict` mode, and specifies the label `label1` and prefix `TestApp`, then 
following actions will happen - 
1. The value for TestApp:BackgroundColor with label1 will be updated to `Green`.
1. The key TestApp:Font with `label1` will be deleted.
1. Other key values will remain unchanged.

The final result in the configuration store would be - 
| Key | Label | Value |
|----|----|----|
| TestApp:BackgroundColor | label1 | Green |
| TestApp:BackgroundColor | label2 | Red |
| OtherKey | None | OtherValue |

shenmuxiaosen: Thanks for the clarification. Why we check specific label and prefix. First prefix is not a filter, it will just append to the front of every imported key. So in your example, if the key name in file is ```TestApp:BackgroundColor``` and prefix is ```TestApp```, then the newly imported key will be ```TestAppTestApp:BackgroundColor```. It feels odd to remove other existing kv in the store with same prefix. Or we are targeting some specific scenario?

shenmuxiaosen: Also what will happen for raw import. Will we do any deletions?

zhoxing-ms:
@pratiksanglikar  Could you please address all the comments tomorrow and let @avanigupta @shenmuxiaosen approve it? If not, this PR will not catch up with the release of this sprint and can only be postponed to the next sprint (03-01)

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
pratiksanglikar: > So in your example, if the key name in file is `TestApp:BackgroundColor` and prefix is `TestApp`, then the newly imported key will be `TestAppTestApp:BackgroundColor`.
My bad, let me update the example in the original comment.

> Also what will happen for raw import. Will we do any deletions?
Yes we do deletions in raw import. The difference is that, in the raw import, we will delete all missing kvs.

pratiksanglikar:
@zhoxing-ms, let's take this PR into the next sprint.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
shenmuxiaosen: For normal kv pair import, if no prefix is specified? will we delete all missing kvs with the same label? I am curious why prefix somehow related with the strict mode?

pratiksanglikar: > if no prefix is specified? will we delete all missing kvs with the same label?

Yes, that's the intended behavior.

> I am curious why prefix somehow related with the strict mode?

We are adding a new feature as strict mode for importing from a file, prefix is an existing feature for importing from a file. Since there is no technical limitation that prevents us to support prefix in the strict mode, I am trying to support it.

avanigupta: @shenmuxiaosen we already support the concept of strict mode in our [Push Task](https://docs.microsoft.com/en-us/azure/azure-app-configuration/push-kv-devops-pipeline#parameters) and [Github Actions](https://docs.microsoft.com/en-us/azure/azure-app-configuration/concept-github-action#use-strict-sync). They both support filtering by prefix and label. That's why we want to do the same in CLI.

Basically, customers are encouraged to organize their key-values in a way that they can filter by prefix and/or label. In strict mode, we are offering the same filtering capability.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
shenmuxiaosen: For raw kv, I think the behavior is to delete all kvs not specified in the import file? @pratiksanglikar

avanigupta: Ah yes, you're right. In that case, we need to fetch all keys here, since we don't have any prefix or label filtering for kvset.

pratiksanglikar: @shenmuxiaosen yes, that's correct!

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
pratiksanglikar: We need `existing_kvset` regardless of whether it is strict or not.
The only difference is when it is not strict, we the filter the set to not include missing kvs. (line #1073)

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
pratiksanglikar: I had totally missed this case, now fixed.

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
pratiksanglikar: Good catch!
fixed now.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
shenmuxiaosen: No. We only need to fetch existing kvsets if strict or preview. Otherwise, we are loading the whole store for every single kvset import.

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -136,32 +137,34 @@ def import_config(cmd,
         src_kvs = __read_kv_from_app_service(
             cmd, appservice_account=appservice_account, prefix_to_add=prefix, content_type=content_type)
 
-    # if customer needs preview & confirmation
-    if not yes:
+    if strict or not yes:
         # fetch key values from user's configstore
         dest_kvs = __read_kv_from_config_store(azconfig_client,
-                                               key=SearchFilterOptions.ANY_KEY,
+                                               key=prefix + '*' if prefix else SearchFilterOptions.ANY_KEY,
```
shenmuxiaosen, line 143: >'*' [](http://example.com/codeflow?start=60&length=3)

SearchFilterOptions.ANY_KEY

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -136,32 +137,34 @@ def import_config(cmd,
         src_kvs = __read_kv_from_app_service(
             cmd, appservice_account=appservice_account, prefix_to_add=prefix, content_type=content_type)
 
-    # if customer needs preview & confirmation
-    if not yes:
+    if strict or not yes:
```
avanigupta, line 140: I'm concerned that the if-else branches in this method have gotten too complicated. We might miss some edge cases here and there keeping track of these conditions. Is it possible to simplify it?

pratiksanglikar: I agree with you. Keeping up with various options we support has led to complicated if-else branches.
I would ideally keep only branchng logic here, and use helper methods for separate options. (e.g. what we did for kvset)
But it is a larger change and needs to be scoped out of this PR.

I am open to any suggestions you have to simplify this logic in the meantime.

shenmuxiaosen (approved):
:shipit:

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
avanigupta: Okay, we can refactor this complicated logic in future.

avanigupta (approved):

pratiksanglikar:
@zhoxing-ms, can you please review and let's merge this PR.

zhoxing-ms (approved):

wangzelin007 (approved):

Issue #19743, opened 2021-09-30
az deployment randomly crashes with: ERROR: 'bytes' object has no attribute 'get'


allanmo79:
**Describe the bug**
when Running az deployments in github runners (azure-cli v2.28.0) it randomly crashes with ERROR: 'bytes' object has no attribute 'get'. When in debug i also get the following error description:

DEBUG: cli.azure.cli.core.sdk.policies: Response content:
DEBUG: cli.azure.cli.core.sdk.policies: ***"error":***"code":"InternalServerError","message":"Encountered internal server error. Diagnostic information: timestamp '20210929T131134Z', subscription id '', tracking id 'xxxxxxxxxxxxxx', request correlation id 'xxxxxxxxxxxxxxxxxxx'."***
DEBUG: cli.azure.cli.core.sdk.policies: Request URL: 'https://management.azure.com/providers/Microsoft.Management/managementGroups/ESD-ConnectedS/providers/Microsoft.Resources/deployments/xxxxxxxxxxxxxxxea?api-version=2021-04-01'

***python errors:***
...................
custom.py", line 407, in on_requestAttributeError: 'bytes' object has no attribute 'get'
..................
in handle_template_based_exceptionknack.util.CLIError: 'bytes' object has no attribute 'get'
.................
ERROR: az_command_data_logger: 'bytes' object has no attribute 'get'
DEBUG: cli.knack.cli: Event: Cli.PostExecute [<function AzCliLogging.deinit_cmd_metadata_logging at 0x03FF4460>]

**To Reproduce**
` az deployment sub create --subscription 'xxxxxxxxxxxxxxxxxxxx'  --name (new-guid).guid --location westeurope `
          --template-file '............\main.bicep'  --parameters '...............\myparams.parameters.json'  --debug
`

**Expected behavior**
No errors

**Environment summary**
az
{
  "azure-cli": "2.28.0",
  "azure-cli-core": "2.28.0",
  "azure-cli-telemetry": "1.0.6",
  "extensions": {}
}
'User-Agent': 'AZURECLI/2.28.0 (MSI) azsdk-python-azure-mgmt-resource/19.0.0 Python/3.8.9 (Windows-10-10.0.17763-SP0) GITHUBACTIONS_AzurePowerShellAction_175xxxxxxxxxxxxxxxx6bdb0dd'

also I am using bicep, not arm!

**Additional context**
[...]

PR #21220, opened 2022-02-08
[ARM] `az deployment`: Fix the bug of `'bytes' object has no attribute 'get'` for error handling in retry cases

zhoxing-ms:
Issue: #19743

**Description**

This error `bytes' object has no attribute 'get'` is due to the bug of `JsonCTemplatePolicy` in the retry operation. 
In the case of retry, because the first request has been processed and converted the type of `request.http_request.data` from string to bytes [code link](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/resource/custom.py#L425)  , so there is no need to process `request.http_request.data` again during retry, 

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/resource/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/resource/custom.py
@@ -410,7 +410,15 @@ class JsonCTemplatePolicy(SansIOHTTPPolicy):
 
     def on_request(self, request):
         http_request = request.http_request
-        if (getattr(http_request, 'data', {}) or {}).get('properties', {}).get('template'):
+        request_data = getattr(http_request, 'data', {}) or {}
+
+        # In the case of retry, because the first request has been processed and
+        # converted the type of "request.http_request.data" from string to bytes,
+        # so there is no need to process request object again during retry
+        if isinstance(request_data, bytes):
+            return
+
+        if request_data.get('properties', {}).get('template'):
             template = http_request.data["properties"]["template"]
             if not isinstance(template, JsonCTemplate):
                 raise ValueError()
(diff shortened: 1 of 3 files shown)


yonzhan:
ARM

wangzelin007 (approved):

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -410,7 +410,15 @@ class JsonCTemplatePolicy(SansIOHTTPPolicy):
 
     def on_request(self, request):
         http_request = request.http_request
-        if (getattr(http_request, 'data', {}) or {}).get('properties', {}).get('template'):
+        request_data = getattr(http_request, 'data', {}) or {}
```
jsntcy, line 413: @zhoxing-ms, have we run all tests impacted by this logic?

zhoxing-ms: It mainly runs the tests that affects the scope of resource group. The logic of other scopes (such as: tenant sub mg) calling this policy is the same, so the scope of resource group can basically cover all use cases~

jsntcy (approved):

jiasli:
The logic of `JsonCTemplatePolicy` is hard to understand. We need better comments!

zhoxing-ms:
> It seems JSONSerializer wraps the jsonc in a JsonCTemplate and then JsonCTemplatePolicy extracts the jsonc. Why isn't the jsonc serialization logic done in JSONSerializer?

In fact, it's technically feasible. I've tried it and it works fine. I guess this may be a decision in code design. Maybe this splicing is not appropriate to put it in the serialization logic in code design.
Because this is legacy code, I can't answer this question perfectly. 😂

> Also, The name template_as_bytes is misleading (#10389 (comment)).

I agree! It will be converted to bytes in the subsequent step of `JsonCTemplatePolicy`, but not in the step of `JSONSerializer`. So I submitted a PR #21991 to correct the variable name and add some comments to explain the function of `JsonCTemplatePolicy`

PR #21245, opened 2022-02-10
{Packaging} Build RPM with Mariner 2.0 image

jiasli:
**Description**

According to the mail conversation, Mariner 1.0 is a little outdated. Mariner 2.0 is latest. As public access to cblmariner2preview.azurecr.io has been granted, we should be able to access Mariner 2.0 anonymously.

Email: _Use Mariner 2.0 to build Azure CLI_

### DIFF
--- a/scripts/release/rpm/Dockerfile.mariner
+++ b/scripts/release/rpm/Dockerfile.mariner
@@ -1,6 +1,6 @@
-ARG tag=1.0
+ARG tag=2.0.20211216-amd64
 
-FROM cblmariner.azurecr.io/base/core:${tag} AS build-env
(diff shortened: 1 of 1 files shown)


wangzelin007 (approved):

yonzhan:
Packaging

scripts/release/rpm/Dockerfile.mariner
```diff
@@ -1,6 +1,6 @@
-ARG tag=1.0
+ARG tag=2.0.20211216-amd64
```
mfrw, line 1: I think we should use the 2.0 tag instead pinning to a particular version.
/cc @jslobodzian @nicogbg
```suggestion
ARG tag=2.0
```

calvinhzy (approved):

Issue #18306, opened 2021-05-31
Description for api-id param for az apim api import is incorrect


plzm:
The description for the --api-id param for the 'az apim api import' command is incorrect. It references a version, but it should actually reference the API ID.

This is particularly important when using this command to update an already-existing API with a Swagger file.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 86f27e34-66e0-1d08-f734-26ce29c56c36
* Version Independent ID: d1a421eb-807c-c0bd-cee3-a5aed7335625
* Content: [az apim api](https://docs.microsoft.com/en-us/cli/azure/apim/api?view=azure-cli-latest#az_apim_api_import)
* Content Source: [latest/docs-ref-autogen/apim/api.yml](https://github.com/MicrosoftDocs/azure-docs-cli/blob/master/latest/docs-ref-autogen/apim/api.yml)
* GitHub Login: @rloutlaw
* Microsoft Alias: **routlaw**

yonzhan:
route to service team

miaojiang:
cc @DreamlessA

RakeshMohanMSFT:
This is fixed

PR #21248, opened 2022-02-10
[API Management] apim api import: Update api-id description  #18306

Anand0046:
updated description for api_id argument

**Description**

Pull request resolves the bug #18306 with inaccurate description for api_id argument

**Testing Guide**

**az apim api import --help**: check for the description of api-id _" API revision identifier. Must be unique in the current API
                                          Management service instance. Non-current revision has
                                          ;rev=n as a suffix where n is the revision number."_

**current change** will change the description of api-id to _"API identifier. Must be unique in the current API Management service instance. Non-current revision has ;rev=n as a suffix where n is the revision number."_


yingru97 (approved):

yonzhan:
API Management

kairu-ms (approved):

PR #21284, opened 2022-02-14
[Network] `az network private-endpoint`: Associate IP configurations and ASGs when creating PE

necusjz:
**Description**

RE: CLI Support For Static IP Configurations + Nic Name + ASG:
> Just wanted to follow up for the CLI work. I noticed that on the “Create Private Endpoint” command, there isn’t the ability to also associate an ASG or configure an IP Config. There seems to be a separate command for this.
It might be an oversight on my end, but it should be possible for customers to provide IPs and associate ASGs when they create the PE as well. It is also only possible to add IP Configurations on update when the connected PaaS service is requesting more IP Configs.

**Testing Guide**

**History Notes**

[Component Name 1] BREAKING CHANGE: `az command a`: Make some customer-facing breaking change.
[Component Name 2] `az command b`: Add some customer-facing feature.

---


yonzhan:
Network

wangzelin007 (approved):

PR #21353, opened 2022-02-18
{Core} Additional logic to judge whether `client` is track2

msyyc:
**Description**

Current logic of [is_track2](https://github.com/Azure/azure-cli/blob/67568014e3df1f183b92ee1ad7699b3b86f35dc0/src/azure-cli-core/azure/cli/core/util.py#L631-L637)  is not accurate, especially when the client is patched with hidden signature: https://github.com/Azure/azure-sdk-for-python/pull/23089#discussion_r808768258.

### DIFF
--- a/src/azure-cli-core/azure/cli/core/util.py
+++ b/src/azure-cli-core/azure/cli/core/util.py
@@ -634,7 +634,7 @@ def is_track2(client_class):
     """
     from inspect import getfullargspec as get_arg_spec
     args = get_arg_spec(client_class.__init__).args
-    return "credential" in args
+    return "credential" in args or hasattr(client_class, "_send_request")
 
 
 DISABLE_VERIFY_VARIABLE_NAME = "AZURE_CLI_DISABLE_CONNECTION_VERIFICATION"


jiasli:
Instead of depending on internal files like `_version.py`, attributes like `_send_request` or method signature, we definitely need a more official and solid way to determine whether a client is Track 2 or not.

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -634,7 +634,7 @@ def is_track2(client_class):
     """
     from inspect import getfullargspec as get_arg_spec
     args = get_arg_spec(client_class.__init__).args
-    return "credential" in args
+    return "credential" in args or hasattr(client_class, "_send_request")
```
iscai-msft, line 637: @msyyc I actually shouldn't have hid the signature yesterday, since that hides the signature for end users. I ended up changing it in the PR before you released yesterday. However this does make sense. For our new DPG generations, we expose send_request as public, so we can maybe add that check as well. But I think I did mess up yesterday by hiding the init signature, so I think it's ok for the CLI code to stay as is

msyyc: the PR is discussed about in e-mail: RE: Testing candidates azure-core on CLI testsuite

PR #21362, opened 2022-02-18
[App Config] `az appconfig`: Add new commands - list-deleted, show-deleted, recover, purge

pratiksanglikar:
**Description**
App Configuration now supports soft-delete feature where an App Configuration and it's contents are retained after deleting for a specified period (1 to 7 days) before purging. This PR adds support for the new soft-delete feature.
Changes in this feature - 
1. Add new options `retention-days` and `enable-purge-protection` for command `az appconfig create`.
2. Add new command `az appconfig list-deleted` which will list all deleted but not yet purged App Configurations.
3. Add new command `az appconfig show-deleted` which will show properties of a deleted but not yet purged App Configuration.
4. Add new command `az appconfig recover` to recover a deleted but not yet purged App Configuration.
5. Add new command `az appconfig purge` to purge a deleted store.

**Testing Guide**
1. Create a new App Configuration with Standard SKU and with 5 days of retention period and with purge protection enabled.
`az appconfig create -g MyResourceGroup -n MyAppConfiguration -l westus --sku Standard --retention-days 5 --enable-purge-protection`

2. List all deleted, but not yet purged App Configurations.
    `az appconfig list-deleted`

3. Show properties of a deleted App Configuration named 'sample-app-configuration'.
    `az appconfig show-deleted --name sample-app-configuration`

4. Recover a deleted App Configuration named 'sample-app-configuration'.
    `az appconfig recover --name sample-app-configuration`

5. Purge a deleted App Configuration named 'sample-app-configuration'.
    `az appconfig purge --name sample-app-configuration`

**History Notes**
[App Config] `az appconfig create`: Add new options `retention-days` and `enable-purge-protection`
[App Config] `az appconfig list-deleted`: Add new command to list all deleted but not yet purged App Configurations
[App Config] `az appconfig show-deleted`: Add new command to show properties of a deleted but not yet purged App Configuration
[App Config] `az appconfig recover`: Add new command to recover a deleted but not yet purged App Configuration
[App Config] `az appconfig purge`: Add new command to purge a deleted store

---


pratiksanglikar:
@shenmuxiaosen, @avanigupta please take a look.

yonzhan:
AppConfig

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -34,19 +34,51 @@ def construct_connection_string(cmd, config_store_name):
...
     try:
         config_store_client = cf_configstore(cmd.cli_ctx)
         all_stores = config_store_client.list()
         for store in all_stores:
             if store.name.lower() == config_store_name.lower():
-                # Id has a fixed structure /subscriptions/subscriptionName/resourceGroups/groupName/providers/providerName/configurationStores/storeName"
-                return store.id.split('/')[4], store.endpoint
+                if resource_group is not None:
+                    raise CLIError('Multiple configuration stores found with name {}.'.format(config_store_name))
+                else:
```
wangzelin007, line 46: Unnecessary "else" after "raise" (no-else-raise)
```suggestion
                if resource_group is None:
```

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -34,19 +34,51 @@ def construct_connection_string(cmd, config_store_name):
...
+    resource_group = None
+    location = None
+    try:
+        client = cf_configstore(cmd.cli_ctx)
+        deleted_stores = client.list_deleted()
+        for deleted_store in deleted_stores:
+            if deleted_store.name.lower() == config_store_name.lower():
+                if location is not None:
+                    raise CLIError('Multiple configuration stores found with name {}.'.format(config_store_name))
+                else:
```
wangzelin007, line 69: Unnecessary "else" after "raise" (no-else-raise)
```suggestion
                if location is None:
```

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -89,12 +94,16 @@ def load_arguments(self, _):
         c.argument('enable_public_network', options_list=['--enable-public-network', '-e'], arg_type=get_three_state_flag(), is_preview=True,
                    help='When true, requests coming from public networks have permission to access this store while private endpoint is enabled. When false, only requests made through Private Links can reach this store.')
         c.argument('disable_local_auth', arg_type=get_three_state_flag(), is_preview=True, help='Disable all authentication methods other than AAD authentication.')
+        c.argument('retention_days', arg_type=retention_days_arg_type, is_preview=True)
+        c.argument('enable_purge_protection', arg_type=get_three_state_flag(), is_preview=True, help='Property specifying whether protection against purge is enabled for this App Configuration. Setting this property to true activates protection against purge for this App Configuration and its contents. Enabling this functionality is irreversible.')
```
wangzelin007, line 98: The lengths of all options ['--enable-purge-protection'] are longer than threshold 22. Argument enable_purge_protection must have a short abbreviation.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -89,12 +94,16 @@ def load_arguments(self, _):
...
 
     with self.argument_context('appconfig update') as c:
         c.argument('tags', arg_type=tags_type)
         c.argument('enable_public_network', options_list=['--enable-public-network', '-e'], arg_type=get_three_state_flag(), is_preview=True,
                    help='When true, requests coming from public networks have permission to access this store while private endpoint is enabled. When false, only requests made through Private Links can reach this store.')
         c.argument('disable_local_auth', arg_type=get_three_state_flag(), is_preview=True, help='Disable all authentication methods other than AAD authentication.')
+        c.argument('enable_purge_protection', arg_type=get_three_state_flag(), is_preview=True,
```
wangzelin007, line 105: The lengths of all options ['--enable-purge-protection'] are longer than threshold 22. Argument enable_purge_protection must have a short abbreviation.

src/azure-cli/azure/cli/command_modules/appconfig/_help.py
```diff
@@ -26,6 +26,38 @@
     text: az appconfig create -g MyResourceGroup -n MyAppConfiguration -l westus --enable-public-network --disable-local-auth
 """
 
+helps['appconfig list-deleted'] = """
+type: command
+short-summary: Lists all deleted, but not yet purged App Configurations.
```
zhoxing-ms, line 31:
```suggestion
short-summary: List all deleted, but not yet purged App Configurations.
```

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -34,19 +34,51 @@ def construct_connection_string(cmd, config_store_name):
...
+    endpoint = None
     try:
         config_store_client = cf_configstore(cmd.cli_ctx)
         all_stores = config_store_client.list()
         for store in all_stores:
             if store.name.lower() == config_store_name.lower():
-                # Id has a fixed structure /subscriptions/subscriptionName/resourceGroups/groupName/providers/providerName/configurationStores/storeName"
-                return store.id.split('/')[4], store.endpoint
+                if resource_group is not None:
+                    raise CLIError('Multiple configuration stores found with name {}.'.format(config_store_name))
```
zhoxing-ms, line 45: Could you please use the specific error type from `azure.cli.core.azclierror` instead of `CLIError`?

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -34,19 +34,51 @@ def construct_connection_string(cmd, config_store_name):
...
+def resolve_deleted_store_metadata(cmd, config_store_name):
+    resource_group = None
+    location = None
+    try:
+        client = cf_configstore(cmd.cli_ctx)
+        deleted_stores = client.list_deleted()
+        for deleted_store in deleted_stores:
+            if deleted_store.name.lower() == config_store_name.lower():
+                if location is not None:
+                    raise CLIError('Multiple configuration stores found with name {}.'.format(config_store_name))
```
zhoxing-ms, line 68: Same as above

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -34,19 +34,51 @@ def construct_connection_string(cmd, config_store_name):
...
+                    # configuration_store_id has a fixed structure /subscriptions/subscription_id/resourceGroups/resource_group_name/providers/Microsoft.AppConfiguration/configurationStores/configuration_store_name
+                    resource_group = deleted_store.configuration_store_id.split('/')[4]
+                    location = deleted_store.location
+    except HttpResponseError as ex:
+        raise CLIError("Failed to get the list of deleted App Configuration stores for the current user. Make sure that the account that logged in has sufficient permissions to access the App Configuration store.\n{}".format(str(ex)))
+
+    if resource_group is not None and location is not None:
+        return resource_group, location
+
+    raise CLIError("Failed to find the deleted App Configuration store '{}'.".format(config_store_name))
```
zhoxing-ms, line 79: Same as above

zhoxing-ms:
@pratiksanglikar By the way, this PR doesn't seem to catch up with the release of this sprint and can only be postponed to the next sprint (04-05), do you think this time is acceptable?

pratiksanglikar:
@zhoxing-ms, let's take it to the next sprint.

src/azure-cli/azure/cli/command_modules/appconfig/custom.py
```diff
@@ -49,8 +50,24 @@ def create_configstore(client,
...
+
+    return client.begin_create(resource_group_name, name, configstore_params)
+
 
+def recover_deleted_configstore(cmd, client, name, resource_group_name=None, location=None, yes=False):
+    if resource_group_name is None or location is None:
+        metadata_resource_group, metadata_location = resolve_deleted_store_metadata(cmd, name)
+
+        resource_group_name = resource_group_name if resource_group_name is not None else metadata_resource_group
+        location = location if location is not None else metadata_location
```
shenmuxiaosen, line 65:
```
if resource_group_name is None:
  resource_group_name = metadata_resource_group
```

src/azure-cli/azure/cli/command_modules/appconfig/custom.py
```diff
@@ -49,8 +50,24 @@ def create_configstore(client,
...
                                             public_network_access=public_network_access,
-                                            disable_local_auth=disable_local_auth)
+                                            disable_local_auth=disable_local_auth,
+                                            soft_delete_retention_in_days=retention_days,
+                                            enable_purge_protection=enable_purge_protection)
+
+    return client.begin_create(resource_group_name, name, configstore_params)
+
 
+def recover_deleted_configstore(cmd, client, name, resource_group_name=None, location=None, yes=False):
```
shenmuxiaosen, line 60: >resource_group_name=None, location=None [](http://example.com/codeflow?start=51&length=39)

Add argument explanations in _params.py

pratiksanglikar: `location` and `resource_group_name` are global parameters and they come up in help as well.
Do we still want to include them in `_params.py`?

src/azure-cli/azure/cli/command_modules/appconfig/custom.py
```diff
@@ -49,8 +50,24 @@ def create_configstore(client,
...
+    if resource_group_name is None or location is None:
+        metadata_resource_group, metadata_location = resolve_deleted_store_metadata(cmd, name)
+
+        resource_group_name = resource_group_name if resource_group_name is not None else metadata_resource_group
+        location = location if location is not None else metadata_location
+
+    configstore_params = ConfigurationStore(location=location.lower(),
+                                            sku=Sku(name="Standard"),  # Only Standard SKU stores can be recovered!
+                                            create_mode=CreateMode.RECOVER)
```
shenmuxiaosen, line 69: what is the original store is free? will it be recovered and upgraded? or server will fail the request

pratiksanglikar: You can't recover a free store because the retention period is 0 days.
So it won't be retained to be able to recover.

src/azure-cli/azure/cli/command_modules/appconfig/custom.py
shenmuxiaosen: For recover, the definition of location and resource group name should be a little different right? Like mentioning it is not configurable,  cannot recover to a new group/location.

pratiksanglikar: Correct. I updated the help text now.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
pratiksanglikar: Added a short abbreviation!

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
pratiksanglikar: Added a short abbreviation!

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
pratiksanglikar: Done.

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
pratiksanglikar: Done.

src/azure-cli/azure/cli/command_modules/appconfig/custom.py
pratiksanglikar: I think your suggestion is cleaner.

src/azure-cli/requirements.py3.Darwin.txt
```diff
@@ -19,7 +19,7 @@ azure-keyvault==1.1.0
 azure-loganalytics==0.1.0
 azure-mgmt-advisor==9.0.0
 azure-mgmt-apimanagement==3.0.0
-azure-mgmt-appconfiguration==2.0.0
+azure-mgmt-appconfiguration==2.1.0b2
```
avanigupta, line 22: Is there a stable version that has the changes we need? We normally dont take a dependency on pre-release versions.

src/azure-cli/azure/cli/command_modules/appconfig/custom.py
```diff
@@ -36,7 +35,9 @@ def create_configstore(client,
                        tags=None,
                        assign_identity=None,
                        enable_public_network=None,
-                       disable_local_auth=None):
+                       disable_local_auth=None,
+                       retention_days=7,
+                       enable_purge_protection=None):
```
avanigupta, line 40: >enable_purge_protection=None [](http://example.com/codeflow?start=23&length=28)

Should the default be False?

avanigupta:
>     configstore_params = ConfigurationStore(location=location.lower(),

Should we pass default create mode here?

---
Refers to: src/azure-cli/azure/cli/command_modules/appconfig/custom.py:48 in c27049e. [](commit_id = c27049e9e59bcc0723db6b75a5668194585a3255, deletion_comment = False)

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -34,17 +35,49 @@ def construct_connection_string(cmd, config_store_name):
...
         for store in all_stores:
             if store.name.lower() == config_store_name.lower():
-                # Id has a fixed structure /subscriptions/subscriptionName/resourceGroups/groupName/providers/providerName/configurationStores/storeName"
-                return store.id.split('/')[4], store.endpoint
+                if resource_group is None:
+                    # Id has a fixed structure /subscriptions/subscriptionName/resourceGroups/groupName/providers/providerName/configurationStores/storeName"
+                    resource_group = store.id.split('/')[4]
+                    endpoint = store.endpoint
+                else:
+                    raise ValidationError('Multiple configuration stores found with name {}.'.format(config_store_name))
```
avanigupta, line 50: >raise ValidationError('Multiple configuration stores found with name {}. [](http://example.com/codeflow?start=20&length=72)

Did we decide to throw this error or just return the first match?

pratiksanglikar: We decided to throw an error and have user provide more parameters to uniquely identify the resource.

src/azure-cli/requirements.py3.Darwin.txt
pratiksanglikar: I'm syncing up with @shenmuxiaosen to see if we have any stable version with our required changes.

src/azure-cli/azure/cli/command_modules/appconfig/custom.py
pratiksanglikar: I think it `False` is more robust. Updated.

src/azure-cli/requirements.py3.Darwin.txt
pratiksanglikar: We can use the preview version. As per my understanding with Shuai, the SDK team started marking preview version recently (i.e. `2.1.0bxx`). earlier they just didn't update the versions for minor versions.

shenmuxiaosen: We used to release CLI based on some preview rp version. So I think it should be fine here.

avanigupta: Okay, thanks for confirming!

src/azure-cli/azure/cli/command_modules/appconfig/custom.py
shenmuxiaosen: If we hard code the standard sku here, if the store is actually a free one, what will be the error message from server? Basically CLI just need to error out that free store can't be recovered.

src/azure-cli/azure/cli/command_modules/appconfig/custom.py
```diff
@@ -62,17 +82,36 @@ def delete_configstore(cmd, client, name, resource_group_name=None, yes=False):
...
+
+
 def list_configstore(client, resource_group_name=None):
     response = client.list() if resource_group_name is None else client.list_by_resource_group(resource_group_name)
     return response
 
 
+def list_deleted_configstore(client):
+    response = client.list_deleted()
+    return response
```
shenmuxiaosen, line 100: >response [](http://example.com/codeflow?start=11&length=8)

Does it also work for other output formats like table? If there are some new attributes we want to show as column name, we will need to change in _format.py.

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -34,17 +35,49 @@ def construct_connection_string(cmd, config_store_name):
...
     except HttpResponseError as ex:
-        raise CLIError("Failed to get the list of App Configuration stores for the current user. Make sure that the account that logged in has sufficient permissions to access the App Configuration store.\n{}".format(str(ex)))
+        raise AzureInternalError("Failed to get the list of App Configuration stores for the current user. Make sure that the account that logged in has sufficient permissions to access the App Configuration store.\n{}".format(str(ex)))
```
avanigupta, line 52: >AzureInternalError [](http://example.com/codeflow?start=14&length=18)

We shouldn't throw AzureInternalError as that is meant for  5xx server errors only

avanigupta: How about AzureResponseError?

src/azure-cli/azure/cli/command_modules/appconfig/custom.py
```diff
@@ -49,8 +50,27 @@ def create_configstore(client,
...
+                                            soft_delete_retention_in_days=retention_days,
+                                            enable_purge_protection=enable_purge_protection,
+                                            create_mode=CreateMode.DEFAULT)
+
+    return client.begin_create(resource_group_name, name, configstore_params)
+
 
+def recover_deleted_configstore(cmd, client, name, resource_group_name=None, location=None, yes=False):
+    if resource_group_name is None or location is None:
+        metadata_resource_group, metadata_location = resolve_deleted_store_metadata(cmd, name)
```
avanigupta, line 63: > metadata_resource_group, metadata_location = resolve_deleted_store_metadata(cmd, name) [](http://example.com/codeflow?start=7&length=87)

We need both location and RG for recovering a deleted store. There are three cases here:
1. User provides both location & RG: We use what they provided
2. User does not provide loc or rg: We find the loc and rg by calling list_deleted. Error out if multiple stores with same name.
3. User provides either RG or Location: We try to find the loc and rg (both) and error out if there are multiple matches. 
Can we be more intelligent in the third scenario? Instead of erroring out, can we match the user provided location or rg to figure out which store they wanted?

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -89,12 +94,20 @@ def load_arguments(self, _):
         c.argument('enable_public_network', options_list=['--enable-public-network', '-e'], arg_type=get_three_state_flag(), is_preview=True,
                    help='When true, requests coming from public networks have permission to access this store while private endpoint is enabled. When false, only requests made through Private Links can reach this store.')
         c.argument('disable_local_auth', arg_type=get_three_state_flag(), is_preview=True, help='Disable all authentication methods other than AAD authentication.')
+        c.argument('retention_days', arg_type=retention_days_arg_type, is_preview=True)
+        c.argument('enable_purge_protection', options_list=['--enable_purge_protection', '-p'], arg_type=get_three_state_flag(), is_preview=True, help='Property specifying whether protection against purge is enabled for this App Configuration. Setting this property to true activates protection against purge for this App Configuration and its contents. Enabling this functionality is irreversible.')
```
avanigupta, line 98: >enable_purge_protection [](http://example.com/codeflow?start=63&length=23)

--enable-purge-protection

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -89,12 +94,20 @@ def load_arguments(self, _):
...
 
     with self.argument_context('appconfig update') as c:
         c.argument('tags', arg_type=tags_type)
         c.argument('enable_public_network', options_list=['--enable-public-network', '-e'], arg_type=get_three_state_flag(), is_preview=True,
                    help='When true, requests coming from public networks have permission to access this store while private endpoint is enabled. When false, only requests made through Private Links can reach this store.')
         c.argument('disable_local_auth', arg_type=get_three_state_flag(), is_preview=True, help='Disable all authentication methods other than AAD authentication.')
+        c.argument('enable_purge_protection', options_list=['--enable_purge_protection', '-p'], arg_type=get_three_state_flag(), is_preview=True,
```
avanigupta, line 105: >enable_purge_protection [](http://example.com/codeflow?start=63&length=23)

Same: --enable-purge-protection

src/azure-cli/azure/cli/command_modules/appconfig/custom.py
```diff
@@ -84,7 +123,8 @@ def update_configstore(cmd,
                        encryption_key_version=None,
                        identity_client_id=None,
                        enable_public_network=None,
-                       disable_local_auth=None):
+                       disable_local_auth=None,
+                       enable_purge_protection=False):
```
avanigupta, line 127: >False [](http://example.com/codeflow?start=47&length=5)

I think it makes sense to keep it 'None" here. What if someone enables purge protection, and then runs update command to update something else? We will still pass False to the server and the operation might fail because we are trying to turn off purge protection.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -57,6 +57,11 @@ def load_arguments(self, _):
         type=int,
         help='Maximum number of items to return. Must be a positive integer. Default to 100.'
     )
+    retention_days_arg_type = CLIArgumentType(
+        options_list=['--retention-days'],
+        type=int,
+        help='Number of days to retain the soft delete enabled App Configuration after deleting. Must be a positive integer.'
```
avanigupta, line 63: > Must be a positive integer. [](http://example.com/codeflow?start=96&length=28)

Can we specify that it should be between 0-7?

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -89,12 +94,20 @@ def load_arguments(self, _):
...
                    help='When true, requests coming from public networks have permission to access this store while private endpoint is enabled. When false, only requests made through Private Links can reach this store.')
         c.argument('disable_local_auth', arg_type=get_three_state_flag(), is_preview=True, help='Disable all authentication methods other than AAD authentication.')
+        c.argument('enable_purge_protection', options_list=['--enable_purge_protection', '-p'], arg_type=get_three_state_flag(), is_preview=True,
+                   help='Property specifying whether protection against purge is enabled for this App Configuration. Setting this property to true activates protection against purge for this App Configuration and its contents. Enabling this functionality is irreversible.')
+
+    with self.argument_context('appconfig recover') as c:
+        c.argument('location', options_list=['--location', '-l'], arg_type=get_location_type(self.cli_ctx), help='Location of the deleted App Configuration. Can be viewed using command `az appconfig show-deleted`.')
```
avanigupta, line 109: We should add similar help text for purge and show-deleted commands too. The default message for "location" field is not applicable there.

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/test_appconfig_commands.py
```diff
@@ -127,6 +132,24 @@ def test_azconfig_mgmt(self, resource_group, location):
...
+
+        self.cmd('appconfig recover -n {config_store_name} -y')
+
+        self.cmd('appconfig show -n {config_store_name}',
+            checks=[self.check('name', '{config_store_name}'),
+                    self.check('location', '{rg_loc}')])
+
+        self.cmd('appconfig delete -n {config_store_name} -g {rg} -y')
+
+        self.cmd('appconfig purge -n {config_store_name} -y')
```
avanigupta, line 148: Can you add a test for the list-deleted command too?

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/recordings/test_azconfig_mgmt.yaml
```diff
@@ -1,986 +0,0 @@
-interactions:
```
avanigupta, line 1: Why was this file deleted?

pratiksanglikar: This is actually something wrong with merging of branches. I will look deeper into it.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
pratiksanglikar: None of the other options list it as `--option-name`. e.g. `enable_public_network`.
Why special case for `enable_purge_protection`?

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
pratiksanglikar: Why special case for `enable_purge_protection`?

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
pratiksanglikar: We want the CLI to not be very tightly coupled to the server. If the server changes the range to be 0-30 days, we don't want to have to update the CLI for it.

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/test_appconfig_commands.py
pratiksanglikar: Added.

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/recordings/test_azconfig_mgmt.yaml
pratiksanglikar: Okay, this is a combination of merge conflict and some changes in Azure SDK.
The Azure SDK is not generating this file along with a file for another AAD test anymore.
Seems like there is a bug in Azure SDK or a change in behavior.

We should clarify with the SDK team.

@zhoxing-ms any idea why this might be happening?

src/azure-cli/azure/cli/command_modules/appconfig/custom.py
pratiksanglikar: Good catch!

src/azure-cli/azure/cli/command_modules/appconfig/custom.py
pratiksanglikar: Sure, we can be a little intelligent here.
I added the logic to match location or resource_group when provided so that we don't error out unnecessarily.

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
pratiksanglikar: Sounds good.

src/azure-cli/azure/cli/command_modules/appconfig/custom.py
pratiksanglikar: Good catch. I didn't even know there was a table format.
I have now added a table formatter for deleted stores.

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/recordings/test_azconfig_mgmt.yaml
zhoxing-ms: > This is actually something wrong with merging of branches. I will look deeper into it.
Okay, this is a combination of merge conflict and some changes in Azure SDK.

May I ask which branch do you mean has the conflict? At present, this file is not deleted in the remote dev branch.

zhoxing-ms: This may be due to some exceptions encountered when executing the test in live mode, it caused the test process to exit early, resulting in the deletion of the original yaml file but it hasn't had time to re-record the new yaml file
Could you sync the latest code from the remote dev branch and then re-record the test in live mode?

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
avanigupta: The flip side of that argument is that right now we know that the server will throw an error if a store is created with more than 7 retention days. Shouldn't we try to avoid the extra server call and add argument validation on client side?
This is something I dont expect will change very often on the server side. We might need to change the validation logic just once when server changes the range to be 0-30 days.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
avanigupta: I was referring to the naming inside options_list=['--enable_purge_protection', '-p']

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
pratiksanglikar: I think we would still be in problem because server could change the validation logic any time, but to release a fix in CLI we need to follow Azure CLI timelines.
If it's just CLI, we could match the time when server releases any changes to CLI release timelines, but if multiple clients need to change and have different timelines, then we won't be able to match the release timelines of all clients to the server.

If it is just a message, and no validation, we can still do it as the effects wouldn't be as bad as not being able to provide large retention period from CLI when the server allows it.

I think having the server validate is good option.
We would be documenting this retention period anyways, so users would possibly know acceptable values.
Even if they don't know, making one extra network call is a reasonable tradeoff between potentially having inconsistent behavior between server and CLI.

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/recordings/test_azconfig_mgmt.yaml
pratiksanglikar: @zhoxing-ms, Yes, this file is present in the remote dev branch, I meant I resolved conflicts incorrectly which resulted in committing the deleted file.

I tried your suggestion.
1. Pulled the latest changes from remote dev.
2. Rebased my commits over it.
3. Executed all the tests with `--live`

Following are my observations - 
1. The tests are successful in both `--live` and without `--live`.
2. recording files are not generated for a couple of tests, but are generated for other tests.

Do you know what more steps I could possibly take to get to the root of this issue?

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
pratiksanglikar: Ah, got it! Updated now.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
avanigupta: Okay, sounds good.

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/test_appconfig_commands.py
```diff
@@ -127,6 +132,33 @@ def test_azconfig_mgmt(self, resource_group, location):
...
+                    self.check('purgeProtectionEnabled', '{enable_purge_protection}')])
+
+        deleted_stores = self.cmd('appconfig list-deleted')
+
+        found = False
+        for deleted_store in deleted_stores:
+            if deleted_store['name'] == config_store_name:
+                assert deleted_store['location'] == location
+                found = True
+        assert found
```
avanigupta, line 147: There is an easier way to test this. See this for reference: https://github.com/Azure/azure-cli/blob/90e7b8b2d701168cb8850e93e4fc773eac5b46a8/src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/test_appconfig_commands.py#L1274-L1281

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -34,17 +35,52 @@ def construct_connection_string(cmd, config_store_name):
...
+                    # It should reach here only when the user has provided only name. If they provide either location or resource group, we should be able to uniquely find the store.
+                    raise ValidationError('Multiple configuration stores found with name {}.'.format(config_store_name))
+    except HttpResponseError as ex:
+        raise AzureResponseError("Failed to get the list of deleted App Configuration stores for the current user. Make sure that the account that logged in has sufficient permissions to access the App Configuration store.\n{}".format(str(ex)))
+
+    if resource_group is not None and metadata_location is not None:
+        return resource_group, metadata_location
 
-    raise CLIError("Failed to find the App Configuration store '{}'.".format(config_store_name))
+    raise ResourceNotFoundError("Failed to find the deleted App Configuration store '{}'.".format(config_store_name))
```
avanigupta, line 83: This error could be hit if the store name is correct, but user provided a location or RG that was incorrect. I feel that we should hint at this in the error message. How about this:
"Failed to find the deleted App Configuration store `storeName`. If you think that the store name is correct, please validate all your input parameters again."

src/azure-cli/azure/cli/command_modules/appconfig/custom.py
```diff
@@ -49,8 +50,27 @@ def create_configstore(client,
...
+    if resource_group_name is None or location is None:
+        metadata_resource_group, metadata_location = resolve_deleted_store_metadata(cmd, name, resource_group_name, location)
+
+        if resource_group_name is None:
+            resource_group_name = metadata_resource_group
+        if location is None:
+            location = metadata_location
+
+    configstore_params = ConfigurationStore(location=location.lower(),
+                                            sku=Sku(name="Standard"),  # Only Standard SKU stores can be recovered!
```
avanigupta, line 71: Saw a similar comment by Shuai but can't find it now. What happens if someone is trying to run this command on a free store and we hard code the sku to standard? Will the server throw an error because the hardcoded sku does not match the real sku of this store?

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -34,17 +35,52 @@ def construct_connection_string(cmd, config_store_name):
...
-        raise CLIError("Failed to get the list of App Configuration stores for the current user. Make sure that the account that logged in has sufficient permissions to access the App Configuration store.\n{}".format(str(ex)))
+        raise AzureResponseError("Failed to get the list of App Configuration stores for the current user. Make sure that the account that logged in has sufficient permissions to access the App Configuration store.\n{}".format(str(ex)))
+
+    if resource_group is not None and endpoint is not None:
+        return resource_group, endpoint
+
+    raise ResourceNotFoundError("Failed to find the App Configuration store '{}'.".format(config_store_name))
+
+
+def resolve_deleted_store_metadata(cmd, config_store_name, resource_group_name, location):
```
avanigupta, line 60: Not every method invoking `resolve_deleted_store_metadata` passes the location and resource group. We should add default values here.

src/azure-cli/azure/cli/command_modules/appconfig/_format.py
```diff
@@ -61,6 +65,16 @@ def _configstore_format_group(item):
     ])
```
avanigupta, line 65: I think we should update this format to show some of the new soft-delete properties too.

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/recordings/test_azconfig_mgmt.yaml
zhoxing-ms: > recording files are not generated for a couple of tests, but are generated for other tests.

Which tests do not generate recording files, and are they marked as `@live_only`?

pratiksanglikar: Hi @zhoxing-ms, no, they are not marked as `@live_only`.

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/test_appconfig_commands.py
pratiksanglikar: I don't think we can test this way reliably because -
1. The index at which the deleted store would be at is not fixed because any number of the stores in the same subscription can be getting created or deleted at any moment.
2. The length of deleted stores would not be fixed because of the same reason.

The reliably way to test would be to test whether the recently deleted store is present in the deleted list.

src/azure-cli/azure/cli/command_modules/appconfig/custom.py
pratiksanglikar: Any store can only be recovered if it is within the retention period. The retention period for a free store is 0 days. Since free store won't be retained, executing this command on a free store would fail anyways because you can't recover permanently deleted store.

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
pratiksanglikar: Great suggestion! Thank you!

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/test_appconfig_commands.py
avanigupta: You're right, I forgot that list-deleted works on the subscription level. In that case, should we reduce the retention period to 1 day for all tests? There's a chance that some live test might fail intermittently if a deleted store exists with the same name (even though names are random). 
We have a common method where this can be updated: `_create_config_store`, and a few other 'appconfig create' commands in some tests.

avanigupta: Also, we should add a test for update command with the new --enable-purge-protection argument. Maybe update it to true and then assert that purging throws an error.
Another test would be trying to turn off purge protection after enabling it.

avanigupta:
I'm not able to create a free store with this change. We might need to adjust the config store parameters on create request if the sku is free:

`az appconfig create -n <storeName> -l <location> -g <rg> --sku free`        

(InvalidProperty) The property 'SoftDeleteRetentionInDays' is not valid.
Code: InvalidProperty
Message: The property 'SoftDeleteRetentionInDays' is not valid.
Additional Information:Type: ActivityId
Info: {
    "activityId": "2ed84e0c-2921-437b-a4a7-4809cf573e7e"
}

src/azure-cli/azure/cli/command_modules/appconfig/custom.py
avanigupta: Enabling purge protection on a free store is not failing. The command succeeds as though purge protection was successfully enabled. Even though the retention period is 0 days, it gives the illusion that purge protection has been enabled on a free store.

src/azure-cli/azure/cli/command_modules/appconfig/custom.py
```diff
@@ -95,7 +135,8 @@ def update_configstore(cmd,
     update_params = ConfigurationStoreUpdateParameters(tags=tags,
                                                        sku=Sku(name=sku) if sku else None,
                                                        public_network_access=public_network_access,
-                                                       disable_local_auth=disable_local_auth)
+                                                       disable_local_auth=disable_local_auth,
+                                                       enable_purge_protection=enable_purge_protection)
```
avanigupta, line 139: The way RP handles enable_purge_protection is this:
- default value of enable_purge_protection is false
- update can be called multiple times with enable_purge_protection=false if it is already false (i.e., it wasn't turned on during create)
- update can be called EXACTLY ONCE with enable_purge_protection=true if it was originally false (i.e., it wasn't turned on during create)
- Once enable_purge_protection has been set to true, it cannot accept true or false anymore. RP only accepts None thereafter. If any request is made with true or false, this error is thrown:
```
(InvalidProperty) The property 'EnablePurgeProtection' is not valid.
Code: InvalidProperty
Message: The property 'EnablePurgeProtection' is not valid.
Additional Information:Type: ActivityId
Info: {
    "activityId": "3ddb9188-477c-44be-ac93-3ac5892e02c8"
}
```

This error should not be displayed to users since it doesn't help understand what's wrong. We need to find a way to catch this specific error and display a helpful error message.

src/azure-cli/azure/cli/command_modules/appconfig/custom.py
pratiksanglikar: I think that is something the server should validate.

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/test_appconfig_commands.py
pratiksanglikar: > should we reduce the retention period to 1 day for all tests?
We should definitely do that.

> We should add a test for update command with the new --enable-purge-protection argument. Maybe update it to true and then assert that purging throws an error.
Another test would be trying to turn off purge protection after enabling it.

I am not sure if we should do that. These behaviors are from the server, the CLI shouldn't be testing that out. CLI should test the behavior of CLI. If the CLI had validations for these behaviors, we should have added tests for them.

src/azure-cli/azure/cli/command_modules/appconfig/custom.py
pratiksanglikar: While we should do proper error handling for this case, I believe we need to rethink how RP handles request with `true` value. It should only error out when the value is `false`.

pratiksanglikar:
Now we should be able to create a free store. Didn't know the nuances of how RP works.
We should probably add some tests for creating / updating free stores to catch issues like this early on.

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/recordings/test_azconfig_mgmt.yaml
zhoxing-ms: Could you tell me which tests do not generate recording files?

pratiksanglikar: In the latest run, the only test not generating file is this one -
[test_appconfig_commands.py#L26-L160](https://github.com/pratiksanglikar/azure-cli/blob/prsangli/soft-delete/src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/test_appconfig_commands.py#L26-L160
) and it was run successfully - [test output](https://dev.azure.com/azure-sdk/public/_build/results?buildId=1411036&view=logs&j=15e5eb7c-cf66-5b91-00a2-db9b11b5e0f9&t=845fa2ee-a602-59f5-4e4c-07ee6d0283b9&l=389)
Seems like the issue with other tests has been resolved on it's own.

avanigupta:
> Now we should be able to create a free store. Didn't know the nuances of how RP works. We should probably add some tests for creating / updating free stores to catch issues like this early on.

We allow only one free store per subscription. Unfortunately, we can't add tests for free stores since they might fail depending on which subscription is used for running them. Our internal R&D subscription doesn't have the 1-free-store restriction for testing purposes.

src/azure-cli/azure/cli/command_modules/appconfig/custom.py
pratiksanglikar: I'm a bit hesitant to add redundant validations (adding same validations as the server) in CLI because it tightens the coupling.
If something changes on the server, we land in an inconsistent state and we need to update the CLI client again.

pratiksanglikar: > What I meant was that RP should do this validation so that CLI doesn't have to. But in case RP cannot validate it (maybe some implementation detail on RP side), this error cannot be displayed as it is to users. It doesn't make any sense to the user. In that case, we need to handle the error in CLI as a last resort.

That totally makes sense.

> Server changes will not be reflected to CLI users until we decide to upgrade the CLI to use a newer SDK.
Ah, yes. Thank you for reminding me. I totally forgot that the changes won't be reflected unless we upgrade the SDK.

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/test_appconfig_commands.py
```diff
@@ -24,6 +24,7 @@
 class AppConfigMgmtScenarioTest(ScenarioTest):
 
     @ResourceGroupPreparer(parameter_name_for_location='location')
+    @AllowLargeResponse
```
cxznmhdcxz, line 27: > In the latest run, the only test not generating file is this one - [test_appconfig_commands.py#L26-L160](https://github.com/pratiksanglikar/azure-cli/blob/prsangli/soft-delete/src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/test_appconfig_commands.py#L26-L160) and it was run successfully - [test output](https://dev.azure.com/azure-sdk/public/_build/results?buildId=1411036&view=logs&j=15e5eb7c-cf66-5b91-00a2-db9b11b5e0f9&t=845fa2ee-a602-59f5-4e4c-07ee6d0283b9&l=389) Seems like the issue with other tests has been resolved on it's own.

Hi @pratiksanglikar, this test case didn't actually run since `@AllowLargeResponse()` decorator is not properly used.
```suggestion
    @AllowLargeResponse()
```

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/test_appconfig_commands.py
```diff
@@ -127,6 +132,33 @@ def test_azconfig_mgmt(self, resource_group, location):
 
         self.cmd('appconfig delete -n {config_store_name} -g {rg} -y')
 
+        self.cmd('appconfig show-deleted -n {config_store_name}',
+            checks=[self.check('name', '{config_store_name}'),
+                    self.check('location', '{rg_loc}'),
+                    self.check('purgeProtectionEnabled', '{enable_purge_protection}')])
+
+        deleted_stores = self.cmd('appconfig list-deleted')
```
cxznmhdcxz, line 140:
```suggestion
        deleted_stores = self.cmd('appconfig list-deleted').get_output_in_json()
```

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/recordings/test_azconfig_mgmt.yaml
pratiksanglikar: I fixed the test based on @cxznmhdcxz comment. Now it is working.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -89,12 +94,26 @@ def load_arguments(self, _):
...
+        c.argument('resource_group_name', arg_type=resource_group_name_type, help='Resource group of the deleted App Configuration.')
+
+    with self.argument_context('appconfig show-deleted') as c:
+        c.argument('location', options_list=['--location', '-l'], arg_type=get_location_type(self.cli_ctx), help='Location of the deleted App Configuration.')
+
+    with self.argument_context('appconfig purge') as c:
+        c.argument('location', options_list=['--location', '-l'], arg_type=get_location_type(self.cli_ctx), help='Location of the deleted App Configuration. Can be viewed using command `az appconfig show-deleted`.')
```
zhoxing-ms, line 116: `options_list=['--location', '-l']` has been declared in `get_location_type()`, so there is no need to repeat the explicit declaration

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -89,12 +94,26 @@ def load_arguments(self, _):
...
                    help='When true, requests coming from public networks have permission to access this store while private endpoint is enabled. When false, only requests made through Private Links can reach this store.')
         c.argument('disable_local_auth', arg_type=get_three_state_flag(), is_preview=True, help='Disable all authentication methods other than AAD authentication.')
+        c.argument('enable_purge_protection', options_list=['--enable-purge-protection', '-p'], arg_type=get_three_state_flag(), is_preview=True,
+                   help='Property specifying whether protection against purge is enabled for this App Configuration. Setting this property to true activates protection against purge for this App Configuration and its contents. Enabling this functionality is irreversible.')
+
+    with self.argument_context('appconfig recover') as c:
+        c.argument('location', options_list=['--location', '-l'], arg_type=get_location_type(self.cli_ctx), help='Location of the deleted App Configuration. Can be viewed using command `az appconfig show-deleted`.')
```
zhoxing-ms, line 109: `options_list=['--location', '-l']` has been declared in `get_location_type()`, so there is no need to repeat the explicit declaration

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -89,12 +94,26 @@ def load_arguments(self, _):
...
+        c.argument('resource_group_name', arg_type=resource_group_name_type, help='Resource group of the deleted App Configuration.')
+
+    with self.argument_context('appconfig show-deleted') as c:
+        c.argument('location', options_list=['--location', '-l'], arg_type=get_location_type(self.cli_ctx), help='Location of the deleted App Configuration.')
```
zhoxing-ms, line 113: Same as above

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
zhoxing-ms, line 116: Same as above

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -89,12 +94,26 @@ def load_arguments(self, _):
         c.argument('enable_public_network', options_list=['--enable-public-network', '-e'], arg_type=get_three_state_flag(), is_preview=True,
                    help='When true, requests coming from public networks have permission to access this store while private endpoint is enabled. When false, only requests made through Private Links can reach this store.')
         c.argument('disable_local_auth', arg_type=get_three_state_flag(), is_preview=True, help='Disable all authentication methods other than AAD authentication.')
+        c.argument('retention_days', arg_type=retention_days_arg_type, is_preview=True)
+        c.argument('enable_purge_protection', options_list=['--enable-purge-protection', '-p'], arg_type=get_three_state_flag(), is_preview=True, help='Property specifying whether protection against purge is enabled for this App Configuration. Setting this property to true activates protection against purge for this App Configuration and its contents. Enabling this functionality is irreversible.')
```
zhoxing-ms, line 98: Could we abstract `enable_purge_protection` into `CLIArgumentType`, so that both `appconfig create` and `appconfig update` commands can reuse it?

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -57,6 +57,11 @@ def load_arguments(self, _):
         type=int,
         help='Maximum number of items to return. Must be a positive integer. Default to 100.'
     )
+    retention_days_arg_type = CLIArgumentType(
+        options_list=['--retention-days'],
+        type=int,
+        help='Number of days to retain the soft delete enabled App Configuration after deleting. Must be a positive integer between 0 and 7.'
```
zhoxing-ms, line 63: Do we need to add verification on the client side for it's valid value?

pratiksanglikar: We have intentionally left out the validation for this value on the client side to prevent tight coupling with the server.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
pratiksanglikar: We are using `get_three_state_flag()` for this argument. When moved to `CLIArgumentType`, somehow the arguments are now being passed as a `str` instead of `bool`. This is an added overhead to convert the `str` to `bool` each time.
Also, if someone forgets to convert it to `bool` it would land up in unexplainable errors.
Do you know any workaround for this?

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
zhoxing-ms: OK~

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
zhoxing-ms: Does this problem only occur when it is defined as `CLIArgumentType`? 
If so, do you add the `arg_type=get_three_state_flag()` in the `CLIArgumentType`? Could you show me the code after you abstract it into `CLIArgumentType`?

pratiksanglikar:
```python
enable_purge_protection_arg_type = CLIArgumentType(
        options_list=['--enable-purge-protection', '-p'],
        arg_type=get_three_state_flag(),
        is_preview=True,
        help='Property specifying whether protection against purge is enabled for this App Configuration. Setting this property to true activates protection against purge for this App Configuration and its contents. Enabling this functionality is irreversible.'
    )
```
and usage - 
```python
c.argument('enable_purge_protection', arg_type=enable_purge_protection_arg_type)
```
This returns `str` instead of `bool`.

If you put `type=bool` in the `CLIArgumentType`, it results in a compiler error.

zhoxing-ms: The problem is that `is_preview=True` cannot be supported in `CLIArgumentType` at present, which will affect the resolution of `arg_type=get_three_state_flag()`. So please keep the original design~

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
zhoxing-ms: @pratiksanglikar Could you fix those comments?

pratiksanglikar: Fixed.

pratiksanglikar:
@zhoxing-ms please hold merging this PR. We are awaiting a PR to be merged on our server side.

src/azure-cli/azure/cli/command_modules/appconfig/custom.py
pratiksanglikar: As per the discussion in the scrum, this should be handled by the RP. Joe has a PR out.
We won't need updated SDK because it is an implementation change, not an API change.
We are good to merge this PR.

pratiksanglikar:
@zhoxing-ms this PR is good to merge after @avanigupta approves.

src/azure-cli/azure/cli/command_modules/appconfig/custom.py
```diff
@@ -36,21 +37,49 @@ def create_configstore(client,
...
     public_network_access = None
     if enable_public_network is not None:
         public_network_access = 'Enabled' if enable_public_network else 'Disabled'
 
+    if sku.lower() == 'free' and (enable_purge_protection or retention_days):
+        logger.warning("Options '--enable-purge-protection' and '--retention-days' will be ignored when creating a free store.")
+        retention_days = None
+        enable_purge_protection = None
+    elif sku.lower() == 'standard' and not retention_days:
+        retention_days = 7
```
avanigupta, line 55: The `elif` check can be removed. If no retention days are passed, we dont need to default to 7. Let RP choose the default value.

zhoxing-ms:
@pratiksanglikar Could you please address the linter issue?

zhoxing-ms (approved):

PR #21323, opened 2022-02-16
[Role] BREAKING CHANGE: `az ad sp create-for-rbac`: Stop defaulting `--scopes` to subscription

jiasli:
**Description**

Close #20806

After showing a warning for the behavior change of `--scopes` (https://github.com/Azure/azure-cli/issues/20806, https://github.com/Azure/azure-cli/pull/21322), this PR makes a breaking change to stop defaulting `--scopes` to subscription.

See: MSRC 69728

PR #21474, opened 2022-03-01
[Compute] BREAKING CHANGE: `az vm/vmss create`: Remove the default value `Contributor` of parameter `--role`

zhoxing-ms:
**Description**

As mentioned in PR #20924 and #21326, we need to remove the default value `Contributor` of `--role` to close the security issue: MSRC Case 68900
Screenshot 2022-03-01 172126

**Testing Guide**

Wait for PR #21369 to fix the problems of existing tests, then add tests and re-record related tests

**History Notes**

---


src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -1021,9 +1021,8 @@ def load_arguments(self, _):
     for scope in ['vm create', 'vmss create']:
         with self.argument_context(scope) as c:
             c.argument('identity_role', options_list=['--role'], arg_group='Managed Service Identity',
-                       help='Role name or id the system assigned identity will have. '
-                            'Please note that the default value "Contributor" will be removed in the future version 2.35.0, '
-                            "so please specify '--role' and '--scope' at the same time when assigning a role to the managed identity")
+                       help="Role name or id the system assigned identity will have. "
+                            "Please specify '--role' and '--scope' at the same time when assigning a role to the managed identity")
```
jiasli, line 1025: How about we put this in the command help? Otherwise, we have to have this text for both `--scope` and `--role`.

yonzhan:
Compute

src/azure-cli/azure/cli/command_modules/vm/_params.py
zhoxing-ms: @jiasli Since `az vm create` and `az vmss create` commands have too many other parameters and examples, I think it might be a good idea to add the description to both `--scope` and `--role` parameters

zhoxing-ms: Or maybe we can not include this information in the help description, because we already have verification logic that will tell users that those two parameters need to be passed in at the same time

jiasli: Duplication is bad. `ad sp create-for-rbac` puts this information in command help:

https://github.com/Azure/azure-cli/blob/b86a4ed33317e9b218c84d4c96498f75235df996/src/azure-cli/azure/cli/command_modules/role/_help.py#L392-L393

But I am open to this topic. 😉

zhoxing-ms: The situation of these two commands may be different, `az vm create` command has nearly 100 parameters and too much help information  😂.
Since the `--scope` and `--role` are only one of many usage scenarios, I'm not sure whether it's appropriate to put them in a global position in this case. And the help information printed by this command is too long, so users may not notice the description from the command level every time

In order to avoid duplicated prompts, I personally think maybe I can remove this information in the help description, because we already have the verification to tell users that `--scope` and `--role` parameters need to be passed in at the same time. What do you think of this idea?

src/azure-cli/azure/cli/command_modules/vm/_validators.py
```diff
@@ -1205,14 +1205,30 @@ def _validate_ssh_key_helper(ssh_key_value, should_generate_ssh_keys):
...
 
 def _validate_vm_vmss_msi(cmd, namespace, from_set_command=False):
+
+    # For the creation of VM and VMSS, "--role" and "--scope" should be passed in at the same time
+    # when assigning a role to the managed identity
+    if not from_set_command and namespace.assign_identity is not None:
+        if namespace.identity_scope and not namespace.identity_role:
+            raise ArgumentUsageError(
+                "usage error: '--scope {}' is not applicable as the '--role' is not provided".format(
+                    namespace.identity_scope))
```
jiasli, line 1215: Together with L1232, for simplicity, why don't we say:

> Usage error: To create role assignments, specify both --role and --scopes.

just like https://github.com/Azure/azure-cli/pull/21323

src/azure-cli/azure/cli/command_modules/vm/_validators.py
```diff
@@ -1205,14 +1205,30 @@ def _validate_ssh_key_helper(ssh_key_value, should_generate_ssh_keys):
...
         identities = namespace.assign_identity or []
         from ._vm_utils import MSI_LOCAL_ID
         for i, _ in enumerate(identities):
             if identities[i] != MSI_LOCAL_ID:
                 identities[i] = _get_resource_id(cmd.cli_ctx, identities[i], namespace.resource_group_name,
                                                  'userAssignedIdentities', 'Microsoft.ManagedIdentity')
-        if not namespace.identity_scope and getattr(namespace.identity_role, 'is_default', None) is None:
+
+        if from_set_command:
+            role_is_explicitly_specified = getattr(namespace.identity_role, 'is_default', None) is None
```
jiasli, line 1227: `namespace.identity_role` will never have `is_default` now, won't it?

zhoxing-ms: When `from_set_command` is true, it indicates that the command is `az vm identity assign`. At this time, the `--role` parameter still has the default value `Contributor`. Therefore, when the user does not specify the  `--role`  parameter, this judgment `getattr(namespace.identity_role, 'is_default', None)` will be true, otherwise it will be none.

src/azure-cli/azure/cli/command_modules/vm/_validators.py
```diff
@@ -1205,14 +1205,30 @@ def _validate_ssh_key_helper(ssh_key_value, should_generate_ssh_keys):
...
+        else:
+            role_is_explicitly_specified = namespace.identity_role is not None
+
+        if not namespace.identity_scope and role_is_explicitly_specified:
             raise ArgumentUsageError("usage error: '--role {}' is not applicable as the '--scope' is not provided".
                                      format(namespace.identity_role))
```
zhoxing-ms, line 1233: Here is the error message that `--scope` parameter is not passed in

src/azure-cli/azure/cli/command_modules/vm/_validators.py
zhoxing-ms: In fact, I wanted to be consistent with the error message that the `--scope` parameter was not passed in the original code https://github.com/Azure/azure-cli/pull/21474/files#r826612256

src/azure-cli/azure/cli/command_modules/vm/_validators.py
jiasli: > At this time, the `--role` parameter still has the default value `Contributor`.

This is definitely not secure. 🤔

src/azure-cli/azure/cli/command_modules/vm/_validators.py
jiasli: I mean, I don't like this message either. It is hard to understand.

src/azure-cli/azure/cli/command_modules/vm/_validators.py
```diff
@@ -1210,17 +1210,32 @@ def _validate_ssh_key_helper(ssh_key_value, should_generate_ssh_keys):
...
 
-def _validate_vm_vmss_msi(cmd, namespace, from_set_command=False):
-    if from_set_command or namespace.assign_identity is not None:
+def _validate_vm_vmss_msi(cmd, namespace, is_identity_assign=False):
+
+    # For the creation of VM and VMSS, "--role" and "--scope" should be passed in at the same time
+    # when assigning a role to the managed identity
+    if not is_identity_assign and namespace.assign_identity is not None:
+        if (namespace.identity_scope and not namespace.identity_role) or \
+                (not namespace.identity_scope and namespace.identity_role):
```
jiasli, line 1219: For [operator precedence](https://docs.python.org/3/reference/expressions.html#operator-precedence), not > and > or, so parentheses are not necessary.

zhoxing-ms: Yes, actually I know this. I put parentheses here just for clearer readability~

wangzelin007 (approved):

jiasli (approved):

PR #21509, opened 2022-03-03
{CosmosDB} Correct failover-priority-change example

nachoalonsoportillo:
Set the --failover-policies parameter to some valid values to properly exemplify what user is expected to provide there. Cosmos DB Resource Provider only validates that (1) the total number of elements with unique priority values equals the total number of georeplicas configured for the account, (2) there is one element with priority zero. Priorities don't have do be contiguous, neither they have to be provided in any specific order. And that's the reason why I intentionally set them to 0, 8, and 3.


yonzhan:
CosmosDB

PR #21518, opened 2022-03-03
[App Config] `az appconfig kv export`: Stop throwing error if no key-values are exported to App Service

avanigupta:
**Description**
Fix https://github.com/Azure/AppConfiguration/issues/596

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
+++ b/src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
@@ -463,6 +463,10 @@ def __read_kv_from_app_service(cmd, appservice_account, prefix_to_add="", conten
 
 
 def __write_kv_to_app_service(cmd, key_values, appservice_account):
+    if not key_values:
+        logger.warning('\nSource configuration is empty. No changes will be made.')
+        return
+
     try:
         non_slot_settings = []
         slot_settings = []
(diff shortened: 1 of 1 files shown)


avanigupta:
cc @shenmuxiaosen @pratiksanglikar

pratiksanglikar:
Can we look at other places while we are at it?
I think we may need to add a check in `__write_kv_and_features_to_config_store` as well.
We currently check for `if not key_values and not features`, but there could be a case where `features` are present but not `key_values` that may land us in trouble.

yonzhan:
App Config

avanigupta:
> Can we look at other places while we are at it? I think we may need to add a check in `__write_kv_and_features_to_config_store` as well. We currently check for `if not key_values and not features`, but there could be a case where `features` are present but not `key_values` that may land us in trouble.

Thanks! Updated the check in all 3 export destinations. We should only check for `if not key_values and not features` because if either one is non-empty, we will definitely write something to the destination. This error occurs when there is nothing to write at all.

pratiksanglikar:
There are two cases - 
1. When `key_values` is a `None` 
2. And, when `key_values` is not a `None` but an empty array.

Let's now consider a scenario where `key_values` is a `None` and `features` has a valid value.
The following statement will throw an error because we are trying to extend `None` - 
```python
    if features:
        key_values.extend(__convert_featureflag_list_to_keyvalue_list(features))   # This will throw an error
```
If it is guaranteed that `key_values` and `features` would never be `None` and would always have an empty array, then
the check `if not key_values and not features` is redundant, but if they can be `None`, then we should check for `if key_values` before we try to extend it.

avanigupta:
The check `if not list` will return true if list is None or if list is empty ([reference](https://docs.python.org/3/library/stdtypes.html#truth-value-testing)).
Since the calling method always initializes `key_values` and `features` as empty lists, they will never be None. So the `extend` call cant throw.

pratiksanglikar (approved):
Looks good!

zhoxing-ms (approved):

PR #16242, opened 2020-12-11
[Hint] Post-output hint

jiasli:
**Description**

This PR requires 

- https://github.com/microsoft/knack/pull/224 which adds `EVENT_CLI_SUCCESSFUL_EXECUTE` event
- #16220 which adds a **style** framework to support colorization
- #16258 which refines the **style** framework to support more types of input

Add experimental **post-output hint** which is shown after a command succeeds (except for `az --version` which follows a different code path).

**Testing Guide**

Currently, only `az login` and `az demo hint` is registered with hinters:

```
> az demo hint
This command is implicitly deprecated because command group 'demo' is deprecated and will be removed in a future release.
{
  "cloudName": "AzureCloud",
  "homeTenantId": "54826b22-38d6-4fb2-bad9-b7b93a3e9c5a",
  "id": "0b1f6471-1bf0-4dda-aec3-cb9272f09590",
  "isDefault": true,
  "managedByTenants": [
    {
      "tenantId": "2f4a9838-26b7-47ee-be60-ccc1fdec5953"
    }
  ],
  "name": "AzureSDKTest",
  "state": "Enabled",
  "tenantId": "54826b22-38d6-4fb2-bad9-b7b93a3e9c5a",
  "user": {
    "name": "rolelivetest@azuresdkteam.onmicrosoft.com",
    "type": "user"
  }
}
The hinter can parse the output to show a "projection" of the output, like

       Subscription name: AzureSDKTest
         Subscription ID: 0b1f6471-1bf0-4dda-aec3-cb9272f09590
                    User: rolelivetest@azuresdkteam.onmicrosoft.com

You may then instruct the user to run additional commands:

az account set <X>            Set your default subscription account
az config set output=table    Set your default output to be in table format
az feedback                   File us your latest issue encountered
az next                       Get some ideas on next steps
az upgrade                    Set your default subscription account

You may also show a hyperlink for more detail: https://docs.microsoft.com/cli/azure/
```

**History Notes**

[Hint] az login: Add experimental post-output hint

PR #21520, opened 2022-03-03
[Profile] `az login`: Format output 

StrawnSC:
**Description**

Changes the output for `az login` to be more human-readable / user-friendly in response to internal feedback. 
- the default output format is table instead of JSON
- the only fields included are `isDefault`, `name`, `state`, and `subscriptionId`
- the default subscription is sorted to the top of the table 
- includes debug text to help new users 

The user can still get a json response by running `az login --output json` 

Output before the proposed change: 
Screen Shot 2022-03-03 at 4 47 02 PM

Output after proposed change: 
Screen Shot 2022-03-03 at 4 41 56 PM

**Testing Guide**

`az login`

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/profile/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/profile/_help.py
@@ -14,6 +14,8 @@ long-summary: >-
     By default, this command logs in with a user account. CLI will try to launch a web browser to log in interactively.
     If a web browser is not available, CLI will fall back to device code login.
 
+    Default output type is table (not json).
+
     To login with a service principal, specify --service-principal.
 examples:
     - name: Log in interactively.
--- a/src/azure-cli/azure/cli/command_modules/profile/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/profile/custom.py
@@ -104,7 +104,9 @@ def login(cmd, username=None, password=None, service_principal=None, tenant=None
           identity=False, use_device_code=False, use_cert_sn_issuer=None, scopes=None, client_assertion=None):
     """Log in to access Azure subscriptions"""
 
-    set_output_format(cmd.cli_ctx, "table")
+    global_params = cmd.cli_ctx.data.get("safe_params", [])
+    if "--output" not in global_params and "-o" not in global_params:
+        set_output_format(cmd.cli_ctx, "table")
 
     # quick argument usage check
     if any([password, service_principal, tenant]) and identity:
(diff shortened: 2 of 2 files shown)


yonzhan:
@jiasli for awareness

jiasli:
To be honest, **I personally don't agree with this change**:

- Switching the default output format from `json` to `table` is such a BREAKING CHANGE that all users and scrips relying on the JSON output of `az logout` will fail.
- We tend not to sort the subscription list because the order is returned by ARM and we simply keep it as-is.

For these messages:

image

We tried to implement them via a new feature called **post-output hint** (https://github.com/Azure/azure-cli/pull/16242), but that feature got suspended.

There is also another new feature called `az init` developed by @zhoxing-ms to let the user select the output format interactively.

jiasli (changes requested):
We need (a lot) more discussion about this BREAKING CHANGE.

StrawnSC:
> We need (a lot) more discussion about this BREAKING CHANGE.

@jiasli I definitely agree with this and apologize if this seems like an unsolicited or unneeded change. This work is all very preliminary and I opened a pull request to sharing progress and *not* to try and get these changes merged without more discussion (hence the PR is a draft and marked as do-not-merge). 

@panchagnula directed the work here and can offer more info on the rationale for this change. My understanding is that internal users were confused about the output of `az login` and wanted a better user experience for first-time users

zhoxing-ms:
> There is also another new feature called az init developed by @zhoxing-ms to let the user select the output format interactively.

Yes, this is `az init` related blog: [blog link](https://techcommunity.microsoft.com/t5/azure-tools-blog/streamline-configuring-azure-cli-with-az-init/ba-p/3051810)
`az init` adjusts the same configuration file as `az config set` and is meant to help simplify the configuration process. It's intended to quickly set up global configurations suitable for users' current environment. 
Users can execute the `az extension add -n init` command to install it. Or they can directly execute `az init` command and then complete the automatic installation through dynamic installation.

panchagnula:
> > There is also another new feature called az init developed by @zhoxing-ms to let the user select the output format interactively.
> 
> Yes, this is `az init` related blog: [blog link](https://techcommunity.microsoft.com/t5/azure-tools-blog/streamline-configuring-azure-cli-with-az-init/ba-p/3051810) `az init` adjusts the same configuration file as `az config set` and is meant to help simplify the configuration process. It's intended to quickly set up global configurations suitable for users' current environment. Users can execute the `az extension add -n init` command to install it. Or they can directly execute `az init` command and then complete the automatic installation through dynamic installation.

@zhoxing-ms thanks for sharing this blog. This is a good start. I wanted to call out though from the feedback being received that is mostly control the response object by command. i.e config will change the response type across all CLI commands, & using Set_output_format is for the command to return a subset of the object response & in a better fashion & adding some debug text on what the next steps should be. Ex, in this POC , text showing how to set the default subscription  post login was added.

Issue #21216, opened 2022-02-08
Upgrade API version used for Azure Managed Identity


dcaro:
**Resource Provider**

Microsoft.ManagedIdentity

**Description of Feature or Work Requested**

Update to API version 2018-11-30 using codegen

**Minimum API Version Required**

2018-11-30

**Swagger Link**

https://github.com/Azure/azure-rest-api-specs/blob/476564a1aa6ddb38ec681c9f89d42f00c1becd25/specification/msi/resource-manager/Microsoft.ManagedIdentity/stable/2018-11-30/ManagedIdentity.json 

**Target Date**

To be included in April release

cc: @karavar

yonzhan:
Managed Identity

zhoxing-ms:
@dcaro Hi, since CLI depends on Python SDK, please release the new version of Python SDK `azure-mgmt-msi`.

In fact, we already have a module `identity` to implement the feature of RP `Microsoft.ManagedIdentity`. [code link](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/identity/commands.py#L21-L27)
For this requirement, do we only need to upgrade the api-version for `Microsoft.ManagedIdentity`, or do we also need to add some new commands at the same time?

zhoxing-ms:
> since CLI depends on Python SDK, please release the new version of Python SDK azure-mgmt-msi

At present, `azure-mgmt-msi 1.0.0` already supports api-version `2018-11-30`, so we don't need to release the new Python SDK

PR #21679, opened 2022-03-17
{MSI} Bump up `azure-mgmt-msi` SDK to `6.0.0`

PARADISSEEKR:
**Description**

Upgrade API version used for Azure Managed Identity https://github.com/Azure/azure-cli/issues/21216

**Testing Guide**

**History Notes**

---


yonzhan:
MSI

jiasli:
Ha. This PR further justified the necessity of https://github.com/Azure/azure-cli/pull/21041. 😄

jiasli (approved):

wangzelin007 (approved):

src/azure-cli/azure/cli/command_modules/identity/_params.py
```diff
@@ -17,5 +18,5 @@ def load_arguments(self, _):
         c.argument('resource_name', arg_type=name_arg_type, id_part='name')
 
     with self.argument_context('identity create') as c:
-        c.argument('location', get_location_type(self.cli_ctx))
+        c.argument('location', get_location_type(self.cli_ctx), required=False)
```
zhoxing-ms, line 21: What will happen if the user does not pass in the `--location` parameter?

PARADISSEEKR: Identity can be created in other regions instead of being consistent with the resource group.

zhoxing-ms: Why is the location consistent with the resource group even if `validator=get_default_location_from_resource_group` is not used ?

zhoxing-ms: I see. This logic is implemented in `process_msi_namespace`
```
def process_msi_namespace(cmd, namespace):
    get_default_location_from_resource_group(cmd, namespace)
    validate_tags(namespace)
```

src/azure-cli/azure/cli/command_modules/identity/commands.py
```diff
@@ -18,10 +18,10 @@ def load_command_table(self, _):
...
 
-    with self.command_group('identity', identity_sdk) as g:
-        g.command('create', 'create_or_update', validator=process_msi_namespace)
+    with self.command_group('identity', identity_sdk, client_factory=_msi_user_identities_operations) as g:
+        g.custom_command('create', 'create_identity', validator=process_msi_namespace)
         g.show_command('show', 'get')
         g.command('delete', 'delete')
         g.custom_command('list', 'list_user_assigned_identities')
-        g.command('list-operations', 'list', operations_tmpl='azure.mgmt.msi.operations.operations#Operations.{}',
+        g.command('list-operations', 'list', operations_tmpl='azure.mgmt.msi.operations#Operations.{}',
```
zhoxing-ms, line 26: Due to the existence of non customized commands (such as `delete` and `list-operations`) and their operation path is hard coded as `azure.mgmt.msi.operations#xxx`. 
If `ResourceType.MGMT_MSI: '2018-11-30'` configuration is not added to the old profiles, CLI will report an error when it executes a non customized command in the old profile which without adding ``ResourceType.MGMT_MSI`.
Screenshot 2022-03-30 211508

This is because when loading arguments, a imports the operation path as hard coded path, which will cause the package from operation path cannot be found because the SDK package is a multiple version package.
code link: https://github.com/Azure/azure-cli/blob/45842acafb076d58f66511c46192e039e3f815f0/src/azure-cli-core/azure/cli/core/commands/command_operation.py#L41-L59
Screenshot 2022-03-30 211424

When we add the configuration `ResourceType.MGMT_MSI` in the profile,  the import operation path is the correct path containing the version number in the multiple version SKD
Screenshot 2022-03-30 210707

zhoxing-ms: The relevant processing logic of import operation path: https://github.com/Azure/azure-cli/blob/45842acafb076d58f66511c46192e039e3f815f0/src/azure-cli-core/azure/cli/core/commands/command_operation.py#L41-L59

zhoxing-ms (approved):

evelyn-ys (approved):

src/azure-cli/azure/cli/command_modules/identity/commands.py
zhoxing-ms: We can also add `resource_type=ResourceType.MGMT_MSI` to the `__init__()`, so that the identity module like many other modules which are only supported in the latest profile.
```python
   def __init__(self, cli_ctx=None):
        from azure.cli.core.commands import CliCommandType
        from azure.cli.core.profiles import ResourceType
        identity_custom = CliCommandType(operations_tmpl='azure.cli.command_modules.identity.custom#{}')
        super(IdentityCommandsLoader, self).__init__(cli_ctx=cli_ctx,
                                                     resource_type=ResourceType.MGMT_MSI,
                                                     custom_command_type=identity_custom)
```

However, when the identity module originally used the Python SDK with single version, it can support all the profiles. Therefore, such modification may cause unnecessary breaking changes in the old profiles. 
So we add the `ResourceType.MGMT_MSI` to all profiles to still make `identity` module can support all profiles

PR #21680, opened 2022-03-17
{Packaging} Rename `Dockerfile.*` to `*.dockerfile`

jiasli:
**Description**

File names like `Dockerfile.ubi8` are not valid file names and cannot be recognized by VS Code, PyCharm or GitHub for syntax highlighting.

GitHub sees it as `ubi8` file:

image

These files are renamed to `{platform}.dockerfile` so that they can be syntax highlighted:

GitHub before/after:

image

image

PyCharm before/after:

image

image


yonzhan:
Packaging

calvinhzy (approved):

PR #21743, opened 2022-03-22
{Compute} Clarify that encrypt-format-all is only applicable to Linux VMs

nachoalonsoportillo:


yonzhan:
VM

src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -472,7 +472,7 @@ def load_arguments(self, _):
         c.argument('disk_name', arg_type=name_arg_type, help='The data disk name.')
 
     with self.argument_context('vm encryption enable') as c:
-        c.argument('encrypt_format_all', action='store_true', help='Encrypts-formats data disks instead of encrypting them. Encrypt-formatting is a lot faster than in-place encryption but wipes out the partition getting encrypt-formatted.')
+        c.argument('encrypt_format_all', action='store_true', help='Encrypts-formats data disks instead of encrypting them. Encrypt-formatting is a lot faster than in-place encryption but wipes out the partition getting encrypt-formatted. (Only supported for Linux virtual machines.)')
```
zhoxing-ms, line 475: > (Only supported for Linux virtual machines.)

Could you please explain the background of this modification?

nachoalonsoportillo: Only [Linux extension](https://docs.microsoft.com/en-us/azure/virtual-machines/extensions/azure-disk-enc-linux) supports that encryption operation. [Windows extension](https://docs.microsoft.com/en-us/azure/virtual-machines/extensions/azure-disk-enc-windows) doesn't, and if you specify that parameter for a Windows VM it reports the following error https://github.com/MicrosoftDocs/azure-docs-cli/issues/549.

zhoxing-ms: Got it, thanks~

zhoxing-ms (approved):

PR #21788, opened 2022-03-24
Release docker image for arm64

odidev:
The following file has been created and modified:
Added file .github/workflows/ci.yml and modified Dockerfile to build and push the image for both amd64 and arm64 platforms.

Signed-off-by: odidev odidev@puresoftware.com


yonzhan:
arm64

jiasli:
@odidev, thanks for the contribution. However, due to company policy, we use Azure DevOps for building docker images: 
https://github.com/Azure/azure-cli/blob/6343cb6c50ac482c9e81459357e01e3e38948a9d/azure-pipelines.yml#L251-L263

You may do some research on how to do that in Azure DevOps. I will also do that when time permits.

PR #20918, opened 2022-01-06
[Packaging] Use CentOS Stream 8 to build `el8` RPM package

jiasli:
## Description

- Replace https://github.com/Azure/azure-cli/pull/20852
- Fix https://github.com/Azure/azure-cli/issues/20719

https://github.com/Azure/azure-cli/pull/11204 added `el8` for RHEL 8 but was reverted by https://github.com/Azure/azure-cli/pull/11325.

As CentOS 7's RPM (4.11.3) can't generate SHA256 digests and causes failure on FIPS-enabled RHEL system (https://github.com/Azure/azure-cli/issues/20719), we have to use CentOS Stream 8's RPM (4.14.3) to build.

Now we can see Payload SHA256 digest is calculated as expected:

```
$ docker run -it --rm quay.io/centos/centos:stream8

# curl 'https://artprodcus3.artifacts.visualstudio.com/A0fb41ef4-5012-48a9-bf39-4ee3de03ee35/29ec6040-b234-4e31-b139-33dc4287b756/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL2F6dXJlLXNkay9wcm9qZWN0SWQvMjllYzYwNDAtYjIzNC00ZTMxLWIxMzktMzNkYzQyODdiNzU2L2J1aWxkSWQvMTI3ODY5MC9hcnRpZmFjdE5hbWUveXVtLWNlbnRvc19zdHJlYW040/content?format=file&subPath=%2Fazure-cli-2.32.0-1.el8.x86_64.rpm' -o azure-cli.rpm

# rpm --checksig --verbose azure-cli.rpm
azure-cli.rpm:
    Header SHA256 digest: OK
    Header SHA1 digest: OK
    Payload SHA256 digest: OK
    MD5 digest: OK
```

## References

- CentOS 8 has been deprecated, so we shouldn't use it anymore. See https://wiki.centos.org/About/Product
- CentOS Stream 8 docker image is distributed as `quay.io/centos/centos:stream8` at https://quay.io/repository/centos/centos?tab=tags. See https://wiki.centos.org/FAQ/CentOSStream

PR #21887, opened 2022-03-31
{Packaging} Revert `BuildRpmPackageCentOS7.ArtifactName` to `yum`

jiasli:
**Description**

#20918 changed the RPM artifact name from `yum` to `rpm-centos7` to better reflect its content (an RPM built by CentOS 7), but this requires lots of changes in the release pipeline, as well as the RPM archive path, such as https://azurecliprod.blob.core.windows.net/archive/20220301.49/yum/azure-cli-2.34.1-1.el7.x86_64.rpm

This PR reverts the change and keep the name `yum` for backward compatibility, even though it is a not good name.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -746,7 +746,8 @@ jobs:
     displayName: 'Publish Artifact: rpm'
     inputs:
       TargetPath: $(Build.ArtifactStagingDirectory)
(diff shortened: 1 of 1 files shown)


yonzhan:
Packaging

yonzhan:
Packaging

wangzelin007 (approved):

PR #21941, opened 2022-04-05
[App Config] Fix #14012055: Fix feature flag import for missing description when using 'appconfig/kvset' profile.

pratiksanglikar:
**Description**
- Fixes [#14012055](https://msazure.visualstudio.com/Azure%20AppConfig/_workitems/edit/14012055) Fix feature flag import for missing description when using 'appconfig/kvset' profile.
- Updates help text for a couple of arguments.

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
+++ b/src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
@@ -36,6 +36,7 @@ from._featuremodels import (map_keyvalue_to_featureflag,
 logger = get_logger(__name__)
 FEATURE_MANAGEMENT_KEYWORDS = ["FeatureManagement", "featureManagement", "feature_management", "feature-management"]
 ENABLED_FOR_KEYWORDS = ["EnabledFor", "enabledFor", "enabled_for", "enabled-for"]
+FEATURE_FLAG_PROPERTIES = {'id', 'description', 'enabled', 'conditions'}
 
 
 class FeatureManagementReservedKeywords:
@@ -1127,7 +1128,7 @@ def __validate_import_feature_flag(kv):
     if kv and validate_import_feature(kv.key):
         try:
             ff = json.loads(kv.value)
-            if ff['id'] and ff['description'] and ff['enabled'] and ff['conditions']:
+            if FEATURE_FLAG_PROPERTIES.issubset(ff.keys()):
                 return True
             logger.warning("The feature flag with key '{%s}' is not a valid feature flag. It will not be imported.", kv.key)
         except JSONDecodeError as exception:
--- a/src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
+++ b/src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
@@ -1128,7 +1128,7 @@ def __validate_import_feature_flag(kv):
     if kv and validate_import_feature(kv.key):
         try:
             ff = json.loads(kv.value)
-            if FEATURE_FLAG_PROPERTIES.issubset(ff.keys()):
+            if FEATURE_FLAG_PROPERTIES == ff.keys():
                 return True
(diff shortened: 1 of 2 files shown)


src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -1127,7 +1128,7 @@ def __validate_import_feature_flag(kv):
     if kv and validate_import_feature(kv.key):
         try:
             ff = json.loads(kv.value)
-            if ff['id'] and ff['description'] and ff['enabled'] and ff['conditions']:
+            if FEATURE_FLAG_PROPERTIES.issubset(ff.keys()):
```
avanigupta, line 1131: `issubset` would allow feature flags with invalid properties to be marked as valid. We should do an exact property match here to make sure that only these 4 properties are present (no more, no less). I think the bug exists because `if ff['description']` would return `False` when the description is an empty string. That's the only additional check we need to add here.

pratiksanglikar: > issubset would allow feature flags with invalid properties to be marked as valid. We should do an exact property match here to make sure that only these 4 properties are present (no more, no less).

I agree, that makes sense.

> I think the bug exists because `if ff['description']` would return `False` when the description is an empty string. That's the only additional check we need to add here.

I am not sure I follow. We only need to check for the existance of these properties, and not validate their values. The current implementation will also fail when you have `enabled` property as `False`, but that shouldn't fail the import.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -149,8 +149,8 @@ def load_arguments(self, _):
...
         c.argument('src_name', help='The name of the source App Configuration.')
         c.argument('src_connection_string', validator=validate_connection_string, help="Combination of access key and endpoint of the source store.")
-        c.argument('src_key', help='If no key specified, import all keys by default. Support star sign as filters, for instance abc* means keys with abc as prefix. Key filtering not applicable for feature flags. By default, all feature flags with specified label will be imported.')
-        c.argument('src_label', help="Only keys with this label in source AppConfig will be imported. If no value specified, import keys with null label by default. Support star sign as filters, for instance * means all labels, abc* means labels with abc as prefix.")
+        c.argument('src_key', help='If no key specified, import all keys by default. Support star sign as filters, for instance abc* means keys with abc as prefix. Key filtering not applicable for feature flags. By default, all feature flags with specified label will be imported. This argument is ignored when using `appconfig/kvset` profile.')
```
avanigupta, line 152: This additional help text is not really needed since src_key and src_label are applicable only when the source is AppConfig. In that case, the profile is irrelevant.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: > The current implementation will also fail when you have enabled property as False, but that shouldn't fail the import.

You're right. So essentially, we just need to make sure that these 4 keys exist, their value can be empty/false/whatever.

pratiksanglikar: Done!

avanigupta (approved):

yonzhan:
AppConfig

pratiksanglikar:
Hi @zhoxing-ms, this PR is ready to be merged!

zhoxing-ms (approved):

zhoxing-ms:
@pratiksanglikar Since this is a bug fix, I suggest to write it's PR title into the release notes, so I modified the PR title from `{}` to `[]`. What do you think?

pratiksanglikar:
@zhoxing-ms, I think it is fine to include this in release notes, so `[]` is fine.

Issue #21964, opened 2022-04-06
Installing `k8s-configuration` and `connectedk8s` fails: distutils.errors.DistutilsPlatformError: Microsoft Visual C++ 14.0 or greater is required. 


ponatara:
> ### `az feedback` auto-generates most of the information requested below, as of CLI version 2.0.62

**Describe the bug**
When installing k8s-configuration and connectedk8s extensions using latest version of az cli,  we are seeing the following issue.
Using an earlier version of the az cli works.

![image](https://user-images.githubusercontent.com/13047387/162073469-c7251ef6-d1ea-424a-a0a8-84fdc1b0de5a.png)

![image](https://user-images.githubusercontent.com/13047387/162073844-43f1e0d8-be6e-43a4-af30-4c9096e5d255.png)

**To Reproduce**

**Expected behavior**
CLI-extension addition should succeed

**Environment summary**
az cli version: 2.35.0
Observed in certain Windows envs
Observed on 2 different setups as of now 

**Additional context**

Excerpts from --debug logs:

``
Given no hashes to check 1 links for project 'pycryptodome': discarding no candidates
Collecting pycryptodome==3.9.8
  Created temporary directory: C:\Users\v-vgujjula\AppData\Local\Temp\pip-unpack-nxnyjrrh
  https://files.pythonhosted.org:443 "GET /packages/4c/2b/eddbfc56076fae8deccca274a5c70a9eb1e0b334da0a33f894a420d0fe93/pycryptodome-3.9.8.tar.gz HTTP/1.1" 200 15633268
  Downloading pycryptodome-3.9.8.tar.gz (15.6 MB)
     -------------------------------------- 15.6/15.6 MB 759.2 kB/s eta 0:00:00
  Added pycryptodome==3.9.8 from https://files.pythonhosted.org/packages/4c/2b/eddbfc56076fae8deccca274a5c70a9eb1e0b334da0a33f894a420d0fe93/pycryptodome-3.9.8.tar.gz#sha256=0e24171cf01021bc5dc17d6a9d4f33a048f09d62cc3f62541e95ef104588bda4 (from connectedk8s==1.2.7) to build tracker 'C:\\Users\\v-vgujjula\\AppData\\Local\\Temp\\pip-req-tracker-6brsqoeq'
[...]

PR #21969, opened 2022-04-07
{Core} Ignore `FileNotFoundError` in `rmtree_with_retry`

jiasli:
**Description**

`rmtree_with_retry` now doesn't ignore `FileNotFoundError`. Then though the directory doesn't exist, `rmtree_with_retry` will retry several times (like in https://github.com/Azure/azure-cli/issues/21964), which is definitely not what we want.

This PR ignores `FileNotFoundError` in `rmtree_with_retry`.


yonzhan:
Core

zhoxing-ms (approved):

jsntcy (approved):

PR #22016, opened 2022-04-11
[Compute] `az vm create`: Fix the bug when creating Flex VMSS without SKU and VM profile

zhoxing-ms:
**Description**

This problem comes from ICM. Just as communicating with @fitzgeraldsteele on Teams, CLI needs to fix the following problems:
1. Error handling issue: `'NoneType' object has no attribute 'lower'` when creating Flex VMSS without `--vm-sku` parameter
2. When creating VM/VMSS, if the user does not pass in `--location`, the location from resource group can be used by default *(there is already this logic, but it is not triggered due to this bug)*
3. For creating VMSS, the information of VM profile and SKU should be either passed in or not passed in. Therefore, if users pass in the `--vm-sku` parameter, CLI needs to verify that users must also pass in the `--image` parameter

**Testing Guide**

Because if the `--zone` parameter is added in the test, the recorded file will be very large. Therefore, instead of adding the corresponding test case, upload the screenshot of this case
Screenshot 2022-04-11 113529

**History Notes**

[Compute] `az vm create`: Fix the bug of "NoneType object has no attribute lower" when creating Flex VMSS without `--vm-sku` parameter

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/vm/_validators.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_validators.py
@@ -362,7 +362,7 @@ def _get_storage_profile_description(profile):
 def _validate_location(cmd, namespace, zone_info, size_info):
     if not namespace.location:
         get_default_location_from_resource_group(cmd, namespace)
-        if zone_info:
+        if zone_info and size_info:
             sku_infos = list_sku_info(cmd.cli_ctx, namespace.location)
             temp = next((x for x in sku_infos if x.name.lower() == size_info.lower()), None)
             # For Stack (compute - 2017-03-30), Resource_sku doesn't implement location_info property
@@ -1546,6 +1546,9 @@ def process_vmss_create_namespace(cmd, namespace):
         for param, value in banned_params.items():
             if value is not None:
                 raise ArgumentUsageError(f'usage error: {param} is not supported for Flex mode')
+
+        if namespace.vm_sku and not namespace.image:
+            raise ArgumentUsageError('usage error: please specify the --image when you want to specify the VM SKU')
 
         if namespace.image:
 
--- a/src/azure-cli/azure/cli/command_modules/vm/tests/latest/test_vm_commands.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/tests/latest/test_vm_commands.py
@@ -6491,7 +6491,7 @@ class VMSSOrchestrationModeScenarioTest(ScenarioTest):
 
         self.cmd('ppg create -g {rg} -n {ppg}')
         self.cmd('vmss create -g {rg} -n {vmss} --orchestration-mode Flexible --single-placement-group false '
-                 '--ppg {ppg} --platform-fault-domain-count 3 --admin-username vmtest',
+                 '--ppg {ppg} --platform-fault-domain-count 3',
                  checks=[
                      self.check('vmss.singlePlacementGroup', False),
                      self.check('vmss.platformFaultDomainCount', 3)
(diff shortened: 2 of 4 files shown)


src/azure-cli/azure/cli/command_modules/vm/tests/latest/test_vm_commands.py
```diff
@@ -6491,7 +6491,7 @@ def test_vmss_simple_orchestration_mode(self, resource_group):
 
         self.cmd('ppg create -g {rg} -n {ppg}')
         self.cmd('vmss create -g {rg} -n {vmss} --orchestration-mode Flexible --single-placement-group false '
-                 '--ppg {ppg} --platform-fault-domain-count 3 --admin-username vmtest',
+                 '--ppg {ppg} --platform-fault-domain-count 3',
```
zhoxing-ms, line 6494: Fix the following errors:
> {"status":"Failed","error":{"code":"DeploymentFailed","message":"At least one resource deployment operation failed. Please list deployment operations for details. Please see https://aka.ms/DeployOperations for usage details.","details":[{"code":"BadRequest","message":"{\r\n  \"error\": {\r\n    \"code\": \"BadRequest\",\r\n    \"message\": \"Virtual Machine Scale Set with orchestration mode 'Flexible' and a Virtual Machine profile must use managed disks.\"\r\n  }\r\n}"}]}}

`--admin-username vmtest` was added from PR #21761 by mistake. It will result in adding redundant VM profile when creating an empty VMSS

wangzelin007 (approved):

src/azure-cli/azure/cli/command_modules/vm/_validators.py
```diff
@@ -1547,6 +1547,9 @@ def process_vmss_create_namespace(cmd, namespace):
             if value is not None:
                 raise ArgumentUsageError(f'usage error: {param} is not supported for Flex mode')
 
+        if namespace.vm_sku and not namespace.image:
```
jsntcy, line 1550: Do we need handle the case `if namespace.image and not namespace.vm_sku`?

zhoxing-ms: It has been confirmed with @fitzgeraldsteele  that for creating Flex VMSS, we only need to raise the usage error 
> raise ArgumentUsageError('usage error: please specify the --image when you want to specify the VM SKU')

Because

> For creating VMSS, the information of VM profile and SKU should be either passed in or not passed in. Therefore, if users pass in the --vm-sku parameter, CLI needs to verify that users must also pass in the --image parameter

jsntcy (approved):

fitzgeraldsteele:
Just to be clear, with vmss flex, either:
* no sku object AND no virtualMachineProfile object
* sku object AND virtualMachineProfile object

You can have neither, or both.

AzCli has required customer enter the image parameter, but provided reasonable defaults for all the other parameters. I would expect that same behavior to continue. So in this example, customer picked a sku but not an image. That should fail on any vmss create, since image is required. If customer only picks an image, we should have the same experience as vmss uniform where AzCli has a reasonable default.

Sent from my T-Mobile 5G Device
Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

fitzgeraldsteele:
Can we also check that just specifying --image still works? If that still works, then the other fixes look good to me!

zhoxing-ms:
> Can we also check that just specifying --image still works? If that still works, then the other fixes look good to me!

Yes, we have run relevant tests in `test_quick_create_flexible_vmss`

PR #22119, opened 2022-04-19
[App Config] `az appconfig`: Add new command group `replica` and add new commands in that command group.

pratiksanglikar:
**Description**
App Configuration now supports geo-replication feature. This PR now supports adding new replicas, listing them and deleting the replicas.

Changes in this feature - 
1. Add new command group `replica` - `az appconfig replica`.
2. Add new command `az appconfig replica create` which will create replica.
3. Add new command `az appconfig replica list` which will list replicas of App Configuration.
4. Add new command `az appconfig replica show` which will show the properties of a replica of an App Configuration.
5. Add new command `az appconfig replica delete` which will delete a replica of an App Configuration.

**Testing Guide**
1. Create a new App Configuration.
`az appconfig create -g MyResourceGroup -n MyAppConfiguration -l westus --sku Standard`

2. Create a new replica for the App Configuration.
`az appconfig replica create -g MyResourceGroup -n MyReplicaName -s MyAppConfiguration -l westus`

3. List all replicas of App Configurations.
    `az appconfig replica list -s MyAppConfiguration`

4. Show properties of a replica of App Configuration.
    `az appconfig replica show -n MyReplicaName -s MyAppConfiguration`

5. Delete a replica of App Configuration.
    `az appconfig replica delete -n MyReplicaName -s MyAppConfiguration`

**History Notes**
[App Config] `az appconfig replica`: Add a new command group to manage replicas of an App Configuration.
[App Config] `az appconfig replica create`: Add a new command `az appconfig replica create` to create a replica of an App Configuration.
[App Config] `az appconfig replica show`: Add a new command to show properties of a replica of an App Configuration.
[App Config] `az appconfig replica list`: Add a new command to list all replicas of an App Configuration.
[App Config] `az appconfig replica delete`: Add a new command to delete a replica of an App Configuration.

---


pratiksanglikar:
@shenmuxiaosen, @avanigupta please take a look.
This is still a draft PR, we need to update the API version yet, but let's get the review started.

yonzhan:
App Config

pratiksanglikar:
Reminder to look at the PR.
@shenmuxiaosen, @avanigupta

src/azure-cli/azure/cli/command_modules/appconfig/commands.py
```diff
@@ -64,6 +70,12 @@ def get_custom_sdk(custom_module, client_factory, table_transformer):
         g.show_command('show', 'show_configstore')
         g.show_command('show-deleted', 'show_deleted_configstore', is_preview=True, table_transformer=deleted_configstore_output_format)
 
+    with self.command_group('appconfig replica', configstore_replica_util, is_preview=True) as g:
+        g.command('list', 'list_replica', table_transformer=configstore_replica_output_format)
```
shenmuxiaosen, line 74: > table_transformer=configstore_replica_output_format [](http://example.com/codeflow?start=41&length=52)

Since ```configstore_replica_output_format``` is specified in the command group level in ```configstore_replica_util```, do we still need to add per command?

src/azure-cli/azure/cli/command_modules/appconfig/custom.py
```diff
@@ -243,6 +244,44 @@ def regenerate_credential(cmd, client, name, id_, resource_group_name=None):
...
+        raise ResourceNotFoundError("The replica '{}' for App Configuration '{}' not found.".format(name, store_name))
+
+
+def create_replica(cmd, client, store_name, name, location, resource_group_name=None):
+    if resource_group_name is None:
+        resource_group_name, _ = resolve_store_metadata(cmd, store_name)
+
+    replica_creation_params = Replica(location=location)
+    return client.begin_create(resource_group_name=resource_group_name,
+                               config_store_name=store_name,
```
shenmuxiaosen, line 270: Any known errors create method may throw? Like in the ```show_replica``` method, do you also want to customize some error message? Usually we will use whatever the server returns.

src/azure-cli/azure/cli/command_modules/appconfig/_format.py
```diff
@@ -77,6 +81,17 @@ def _deleted_configstore_format_group(item):
     ])
 
 
+def _configstore_replica_format_group(item):
+    return OrderedDict([
```
shenmuxiaosen, line 85: >OrderedDict [](http://example.com/codeflow?start=11&length=11)

It doesn't have the config store name?

pratiksanglikar: No, it doesn't have the name of the origin store.
Although it would be better to have the origin store name here, it may be unnecessary because the user needs to know the origin store name to view the replicas. So, they would be providing the name int the command.

src/azure-cli/azure/cli/command_modules/appconfig/custom.py
pratiksanglikar: The `ResourceNotFoundError` in `show_replica` is a special case where the error message returned by the server isn't clear, that's the reason I made an exception to override the server's error message with a meaningful one.
Other errors at other places return clear error messages so we don't need to override those error messages.

src/azure-cli/azure/cli/command_modules/appconfig/commands.py
pratiksanglikar: You are right, we don't need to specify the table formatter for individual commands anymore.
Removed it.

src/azure-cli/azure/cli/command_modules/appconfig/custom.py
shenmuxiaosen: Okay, that makes sense.

src/azure-cli/azure/cli/command_modules/appconfig/_format.py
shenmuxiaosen: Most of those attributes are associated with the replica itself. But since we add the source group here, seems like we should also add resource level stuff. What about the json output format. It 1-to-1 maps all attributes?

PR #19442, opened 2021-09-03
{CI} Update `generate_history_notes.py`: Print commit ID and subject

jiasli:
## Description

- Print commit ID and subject in `generate_history_notes.py`
- Accept PR body to be `None`

PR #21309, opened 2022-02-15
[App Service] `az webapp deployment slot create`: Allow overriding container settings

StrawnSC:
**Description**

Allows overriding the container settings `deployment_container_image_name`, `docker_registry_server_password`, and `docker_registry_server_user` from the `configuration_source` slot when creating a new slot. 

Also changes a bunch of instances where `CLIError` is being raised to specific error types  

**Testing Guide**

`azdev test WebappContainerScenarioTests`

**History Notes**

---

This checklist is used to make sure that common guidelines for a pull request are followed.

- [X] The PR title and description has followed the guideline in [Submitting Pull Requests](https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules#submitting-pull-requests).

- [X] I adhere to the [Command Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md).

- [X] I adhere to the [Error Handling Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/error_handling_guidelines.md).

PR #22308, opened 2022-05-09
{Misc.} Replace NBSP with normal ASCII space

jiasli:
**Description**

For unknown reason, https://github.com/Azure/azure-cli/pull/15998 added NBSP ([Non-breaking space U+00A0](https://en.wikipedia.org/wiki/Non-breaking_space)) to `scripts/release/debian/deb_install.sh`

PyCharm:

image

VS Code:

image

This leads to display issue in Edge browser:

https://azurecliprod.blob.core.windows.net/$root/deb_install.sh

image

This PR replaces NBSP with normal ASCII space.

We've seen this issue before:

- https://github.com/Azure/azure-cli/pull/19442
- https://github.com/Azure/azure-cli/pull/8785

### DIFF
--- a/doc/authoring_command_modules/README.md
+++ b/doc/authoring_command_modules/README.md
@@ -176,11 +176,11 @@ History notes are auto-generated based on PR titles and descriptions starting fr
 2. [**Mandatory**] If it's a breaking change, the second part should be `BREAKING CHANGE` followed by a colon. In the case of hotfix, put `Hotfix` in this part. If it's related to fixing an issue, put `Fix #number` in this part. For other cases, this part could be empty.
 3. [**Recommendation**] If the change can be mapped into a command, then the next part could be the command name starting with `az`, followed by a colon.
 4. [**Recommendation**] Use the right verb with **present-tense** in **base form** and **capitalized first letter** to describe what is done:
-    * **Add** for new features.
-    * **Change** for changes in existing functionality.
-    * **Deprecate** for once-stable features removed in upcoming releases.
-    * **Remove** for deprecated features removed in this release.
-    * **Fix** for any bug fixes.
+    * **Add** for new features.
+    * **Change** for changes in existing functionality.
+    * **Deprecate** for once-stable features removed in upcoming releases.
+    * **Remove** for deprecated features removed in this release.
+    * **Fix** for any bug fixes.
 
 Examples of customer-facing change PR title:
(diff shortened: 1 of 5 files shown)


yonzhan:
Packaging

bebound:
There are also some NBSP under `src` path, should we also replace them?

For example:  `src/azure-cli/azure/cli/command_modules/appservice/custom.py`
Screen Shot 2022-05-09 at 12 20 52

jiasli:
> There are also some NBSP under `src` path, should we also replace them?
> 
> For example: `src/azure-cli/azure/cli/command_modules/appservice/custom.py` Screen Shot 2022-05-09 at 12 20 52

Nice catch!

https://github.com/Azure/azure-cli/blob/c2bffb8bef23728919fe8a315ef7553b103ec393/src/azure-cli/azure/cli/command_modules/appservice/custom.py#L2719

https://github.com/Azure/azure-cli/blob/c2bffb8bef23728919fe8a315ef7553b103ec393/src/azure-cli/azure/cli/command_modules/appservice/custom.py#L2730

They are added by #21309.

bebound:
I found more special character in our source code.

`doc/azure2az_commands.rst` and `doc/authoring_command_modules/README.md` contains NBSP
`src/azure-cli/HISTORY.rst` contains NBSP & ZWSP

If this possible to create some rules to prevent these characters being merged?

jiasli:
I searched across the project but have no idea why PyCharm didn't find them. 🤔 I purged the cache, and it works now. Fixed everywhere.

> There are so many unicode whitespace characters. If this is possible to create some rules to prevent these characters being merged?

Surely we can. I am just curious how they are inserted in the first place!

jsntcy:
Use non-breaking spaces instead of regular spaces when you don't want Word to break a line at a certain space. Is it possible some of them are added for specific purpose, for example, avoid to break command into multiple lines?
@jiasli

wangzelin007 (approved):

calvinhzy (approved):

bebound (approved):

jsntcy (approved):

jiasli:
> Is it possible that some of them were added for specific purpose, for example, avoid to break command into multiple lines?

These are not valid scenarios, since line breaks are acceptable. We can resume our investigation once we see more occurrences.

Issue #22439, opened 2022-05-13
Disk Create/Update to PMK


VikasMohandoss:
> ### `az feedback` auto-generates most of the information requested below, as of CLI version 2.0.62

**Related command**
1. az disk update --name CrossSubCMKDisk4 --resource-group vimohand-XTCMK --encryption-type EncryptionAtRestWithPlatformKey
2. az disk create -g vimohand-XTCMK -n CrossSubCMKDisk2 --source CrossSubCMKDisk --encryption-type EncryptionAtRestWithPlatformKey

**Describe the bug**
1. Update encryption type to platform managed keys for a disk encrypted with cmk. I would expect this command to successfully complete. It looks like it is getting treated as cmk and sends a disk encryption set id and thus the error. 
2. Create new disk encrypted with PMK from a disk encrypted with CMK. This succeeds but the new disk is created with CMK.

**To Reproduce**
az feature registration create --namespace Microsoft.Compute --name AllowServerSideEncryptionSettingsChange
+ commands above

**Expected behavior**
1. Disk updated to PMK
2. Disk created with PMK

ETA needed - As soon as possible. Targeted ETA for feature public preview is 5/18

yonzhan:
Compute

zhoxing-ms:
@VikasMohandoss Thanks for creating this issue! 
However, because our work in this sprint has been fully arranged and the release for this sprint will be launched next week, so this requirement cannot catch up with the release of this sprint. Is it acceptable for us to release it in the next sprint **(07-05)**?

PR #22484, opened 2022-05-17
[Compute] `az disk create/update`: Fix `--encryption-type` processing

cxznmhdcxz:
**Related command**
See issue: https://github.com/Azure/azure-cli/issues/22439

Fix following performance:
1. `az disk create`:  if only specified `--encryption-type` but not `--disk-encryption-set`,  `--encryption-type` is actually ignored during processing
2. `az disk update`:  if only specified `--encryption-type EncryptionAtRestWithPlatformKey`, would not remove current disk encryption set(if the disk has it) and only update encryption type, thus may cause response error

**Description**

**Testing Guide**

**History Notes**

[Compute] `az disk create`: Fix the issue that specifying encryption type as `EncryptionAtRestWithPlatformKey` does not take effect when creating a disk
[Compute] `az disk update`: Fix the `(InvalidParameter) Resource xxx encrypted with platform key has disk encryption set id specified` error when updating the encryption type to platform managed keys 

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/vm/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/custom.py
@@ -436,7 +436,7 @@ def list_managed_disks(cmd, resource_group_name=None):
     return client.disks.list()
 
 
-def update_managed_disk(cmd, resource_group_name, instance, size_gb=None, sku=None, disk_iops_read_write=None,
+def update_managed_disk(cmd, resource_group_name, instance, size_gb=None, sku=None, disk_iops_read_write=None,  # pylint: disable=too-many-branches
                         disk_mbps_read_write=None, encryption_type=None, disk_encryption_set=None,
                         network_access_policy=None, disk_access=None, max_shares=None, disk_iops_read_only=None,
                         disk_mbps_read_only=None, enable_bursting=None, public_network_access=None,
--- a/src/azure-cli/azure/cli/command_modules/vm/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/custom.py
@@ -467,11 +467,11 @@ def update_managed_disk(cmd, resource_group_name, instance, size_gb=None, sku=No
                 subscription=get_subscription_id(cmd.cli_ctx), resource_group=resource_group_name,
                 namespace='Microsoft.Compute', type='diskEncryptionSets', name=disk_encryption_set)
         instance.encryption.disk_encryption_set_id = disk_encryption_set
-    if instance.encryption.type != 'EncryptionAtRestWithCustomerKey' and \
-                encryption_type != 'EncryptionAtRestWithCustomerKey':
-        instance.encryption.disk_encryption_set_id = None
     if encryption_type is not None:
         instance.encryption.type = encryption_type
+        if instance.encryption.type != 'EncryptionAtRestWithCustomerKey' and \
+                encryption_type != 'EncryptionAtRestWithCustomerKey':
+            instance.encryption.disk_encryption_set_id = None
     if network_access_policy is not None:
         instance.network_access_policy = network_access_policy
     if disk_access is not None:
--- a/src/azure-cli/azure/cli/command_modules/vm/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/custom.py
@@ -372,10 +372,8 @@ def create_managed_disk(cmd, resource_group_name, disk_name, location=None,  # p
             namespace='Microsoft.Compute', type='diskAccesses', name=disk_access)
 
     encryption = None
-    if disk_encryption_set:
+    if disk_encryption_set or encryption_type:
         encryption = Encryption(type=encryption_type, disk_encryption_set_id=disk_encryption_set)
-    elif encryption_type:
-        encryption = Encryption(type=encryption_type)
 
     disk = Disk(location=location, creation_data=creation_data, tags=(tags or {}),
                 sku=_get_sku_object(cmd, sku), disk_size_gb=size_gb, os_type=os_type, encryption=encryption)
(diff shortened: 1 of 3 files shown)


yonzhan:
Compute

cxznmhdcxz:
Right now this is the request body while trying to update a CMK disk's encryption type to PMK:
```
{"location": "eastus", "tags": {}, "sku": {"name": "Premium_LRS"}, "properties": {"hyperVGeneration": "V1", "creationData": {"createOption": "Empty"}, "diskSizeGB": 1, "diskIOPSReadWrite": 120, "diskMBpsReadWrite": 25, "encryption": {"type": "EncryptionAtRestWithPlatformKey"}, "networkAccessPolicy": "AllowAll", "tier": "P1", "publicNetworkAccess": "Enabled"}}
```


service will return this error:
```
(BadRequest) Cannot change encryption type of disk 'cd' from EncryptionAtRestWithCustomerKey to EncryptionAtRestWithPlatformKey
```

@grizzlytheodore  Is this by design?

wangzelin007 (approved):

zhoxing-ms:
Could you add some tests for those change?

VikasMohandoss:
Hi Xing,

With the feature registration, we should not be getting the error message.
Is this something we could test as well? There is no registration required
in Canary so it would be a good test env.

Thanks!
Vikas

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -373,6 +373,8 @@ def create_managed_disk(cmd, resource_group_name, disk_name, location=None,  # p
     encryption = None
     if disk_encryption_set:
         encryption = Encryption(type=encryption_type, disk_encryption_set_id=disk_encryption_set)
+    elif encryption_type:
+        encryption = Encryption(type=encryption_type)
```
zhoxing-ms, line 377:
```suggestion
    if disk_encryption_set or encryption_type:
        encryption = Encryption(type=encryption_type, disk_encryption_set_id=disk_encryption_set)
```

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -464,8 +466,9 @@ def update_managed_disk(cmd, resource_group_name, instance, size_gb=None, sku=No
                 subscription=get_subscription_id(cmd.cli_ctx), resource_group=resource_group_name,
                 namespace='Microsoft.Compute', type='diskEncryptionSets', name=disk_encryption_set)
         instance.encryption.disk_encryption_set_id = disk_encryption_set
-    if encryption_type is not None:
+    elif encryption_type is not None:
         instance.encryption.type = encryption_type
+        instance.encryption.disk_encryption_set_id = None
```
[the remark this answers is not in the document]
zhoxing-ms: @VikasMohandoss @laurawu19 @vikramd-ms Could you help confirm that when updating the disk, `disk_encryption_set` can be sent to service at the same time only when `encryption_type` is `EncryptionAtRestWithCustomerKey`. 
For other `encryption_type`, `disk_encryption_set` needs to be removed before requesting REST service. May I ask is my understanding right?

zhoxing-ms:
@VikasMohandoss @laurawu19 @vikramd-ms Another question that needs your help to confirm: as mentioned by @cxznmhdcxz, when we update the `encryption_type` of disk from `EncryptionAtRestWithPlatformKey` to `EncryptionAtRestWithPlatformKey` to and clear the original `disk_encryption_set_id` , than service will return the following error message:

> (BadRequest) Cannot change encryption type of disk 'cd' from EncryptionAtRestWithCustomerKey to EncryptionAtRestWithPlatformKey

Does service currently not support updating the `encryption_type` of disk from `EncryptionAtRestWithCustomerKey` to `EncryptionAtRestWithPlatformKey`, right?
If so, does the this error message meet your expectations?

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -464,6 +464,9 @@ def update_managed_disk(cmd, resource_group_name, instance, size_gb=None, sku=No
                 subscription=get_subscription_id(cmd.cli_ctx), resource_group=resource_group_name,
                 namespace='Microsoft.Compute', type='diskEncryptionSets', name=disk_encryption_set)
         instance.encryption.disk_encryption_set_id = disk_encryption_set
+    if instance.encryption.type != 'EncryptionAtRestWithCustomerKey' and \
+                encryption_type != 'EncryptionAtRestWithCustomerKey':
+        instance.encryption.disk_encryption_set_id = None
     if encryption_type is not None:
         instance.encryption.type = encryption_type
```
zhoxing-ms, line 471:
```suggestion
    if encryption_type is not None:
        instance.encryption.type = encryption_type
        if instance.encryption.type != 'EncryptionAtRestWithCustomerKey' and \
                encryption_type != 'EncryptionAtRestWithCustomerKey':
            instance.encryption.disk_encryption_set_id = None
```

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -466,6 +466,9 @@ def update_managed_disk(cmd, resource_group_name, instance, size_gb=None, sku=No
         instance.encryption.disk_encryption_set_id = disk_encryption_set
     if encryption_type is not None:
         instance.encryption.type = encryption_type
+        if instance.encryption.type != 'EncryptionAtRestWithCustomerKey' and \
+                encryption_type != 'EncryptionAtRestWithCustomerKey':
+            instance.encryption.disk_encryption_set_id = None
```
zhoxing-ms, line 471:
```suggestion
        if encryption_type != 'EncryptionAtRestWithCustomerKey':
            instance.encryption.disk_encryption_set_id = None
```

zhoxing-ms (approved):

Issue #19858, opened 2021-10-12
Drop Python 3.6 support


jiasli:
## Description

Python 3.6 is reaching its end of support on 2021-12-23: https://www.python.org/downloads/

Maintaining code which can work perfectly on Python 3.6 and above versions is a maintenance burden:

- https://github.com/Azure/azure-cli/pull/19024#discussion_r678001836

We need to drop Python 3.6 like how we dropped Python 2.7:

- https://github.com/Azure/azure-cli/issues/11362
- https://github.com/Azure/azure-cli/pull/11363

jiasli:
One big problem with CentOS 7 (copied from https://github.com/Azure/azure-cli/pull/20608#issue-1070553808):

- Python 3.6 has been deprecated on 2021-12-23 (https://www.python.org/downloads/), but CentOS official repo currently only has Python 3.6.
- CentOS 8 will be deprecated very soon on 2021-12-31, leaving the old CentOS 7 be the only supported stable distribution (https://wiki.centos.org/About/Product).

Also see:

- https://community.home-assistant.io/t/python-support-on-centos-after-version-3-6-deprecation/146830
- https://thehackernews.com/2021/09/moving-forward-after-centos-8-eol.html
- https://forums.centos.org/viewtopic.php?t=78236&p=328761

jiasli:
Good news is that Red Hat Enterprise Linux (RHEL) 8 now has Python 3.6, 3.8, and 3.9:

- https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_basic_system_settings/assembly_installing-and-using-python_configuring-basic-system-settings
- https://developers.redhat.com/blog/2019/05/07/what-no-python-in-red-hat-enterprise-linux-8

PR #22606, opened 2022-05-25
[Packaging] Use Python 3.9 in RHEL 8's RPM

jiasli:
**Description**

Red Hat Enterprise Linux (RHEL) 8 now has Python 3.6, 3.8, and 3.9: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_basic_system_settings/assembly_installing-and-using-python_configuring-basic-system-settings

Since Python 3.6 has been deprecated (#19858), we use Python 3.9 on RHEL 8.

### DIFF
--- a/scripts/release/rpm/ubi8.dockerfile
+++ b/scripts/release/rpm/ubi8.dockerfile
@@ -6,20 +6,22 @@ FROM registry.access.redhat.com/ubi8/ubi:${tag} AS build-env
 ARG cli_version=dev
 
 RUN yum update -y
-RUN yum install -y wget rpm-build gcc libffi-devel python3-devel openssl-devel make bash diffutils patch dos2unix python3-virtualenv perl
+RUN yum install -y wget rpm-build gcc libffi-devel python39-devel openssl-devel make bash diffutils patch dos2unix perl
 
 WORKDIR /azure-cli
 
 COPY . .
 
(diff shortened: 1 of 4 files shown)


jiasli:
Off-topic: Personally, I don't really like the inconsistency of `python3` package. This consistency causes all the fuss in this PR:

- RHEL 8's `python3` should point to a more modern version, instead of the deprecated Python 3.6.
- Mariner should have specific versions of Python packages like `python39`, so that users can choose which one to install.

scripts/release/rpm/ubi8.dockerfile
```diff
@@ -6,20 +6,22 @@ FROM registry.access.redhat.com/ubi8/ubi:${tag} AS build-env
 ARG cli_version=dev
 
 RUN yum update -y
-RUN yum install -y wget rpm-build gcc libffi-devel python3-devel openssl-devel make bash diffutils patch dos2unix python3-virtualenv perl
```
jiasli, line 9: We use `venv` built-in module now. No need to install `python3-virtualenv`.

bebound:
There are many `sed: command not found` error in  [Rpm Package Mariner 2.0 log](https://dev.azure.com/azure-sdk/29ec6040-b234-4e31-b139-33dc4287b756/_apis/build/builds/1598943/logs/448). Is this a bug of Mariner 2?

bebound (approved):

wangzelin007 (approved):

PR #22607, opened 2022-05-25
[Core] shorthand syntax: Simplify the escape character usage in Single Quotes String

kairu-ms:
**Related command**

**Description**

In previous design, the escape character `/` is hard to use, customers have to double `/` characters when it appeared in Single Quotes String of shorthand syntax.

For example, when customer passes an url `https://bill.microsft.com/blog`, in previous design, it should be `'https:////bill.microsft.com//blog'` in `Full Value` format. It becomes longer than original, which obeys the goal of shorthand syntax.

This PR simplify the escape character usage in [Single Quotes String](https://github.com/Azure/azure-cli/blob/dev/doc/shorthand_syntax.md#single-quotes-string).

The reason we need escape character in Single Quotes String is that we need to distinguish the normal `'` in string or the end `'` to finish Single Quotes String. There's no reason to escape the escape character `/` itself, only `'` character in the Single Quotes String requires escape.

In the new design, we use `'/` to represent escaped `'` in Single Quotes String. The character '/' will only be the escape character when it's after the `'` in Single Quotes String. In other case, it will be normal character. Which means customer can pass an url `https://bill.microsft.com/blog` as what it looks like.

There are two reasons to put `/` after `'`, instead of before it.
- If we use `/'` to escape `'` character, then we have to find a way to represent string end with `/` in Single Quotes String, for example if the string is an url such as `https://azure.microsoft.com/en-us/`，in Single Quotes String it will be like `'https://azure.microsoft.com/en-us/'` and you can see the end of expression is `/'`, which is distinguished with the escape `'`. That's why in previous design, we have to escape `/` itself as well. If we put `/` after `'`, we will never have this issue.
- We choose `/` to escape because it can make commands run both in both Powershell and Bash terminals without any change. Its shape will let eyes focus on the character before it instead of after it. So `'/` is more readable than `/'`.

This PR is a breaking change for customer usage. Fortunately, we don't have any codegen v2 commands published. So it will not break customers. I will limit the extension generated to the new version of azure-cli.

**Testing Guide**

**History Notes**

---

### DIFF
--- a/doc/shorthand_syntax.md
+++ b/doc/shorthand_syntax.md
@@ -6,6 +6,27 @@ Azure cli shorthand syntax can help cli users to pass complicated argument value
 
 The command lines written in shorthand syntax can run in both _Powershell_ and _Bash_ terminals without any change in must situations.
 
+[1. Shorthand syntax formats](#shorthand-syntax-formats)
+
+[1.1. `Full Value` format](#full-value-format)
+
+[1.2. `Partial Value` format](#partial-value-format)
+
+[1.3. Combine with `Full Value` and `Partial Value`](#combine-with-full-value-and-partial-value)
+
+[1.4. use `??` to show help](#use--to-show-help)
+
+[1.5. Pass a `null` Value](#pass-a-null-value)
+
+[2. Single Quotes String](#single-quotes-string)
+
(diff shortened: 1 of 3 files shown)


yonzhan:
Core

calvinhzy (approved):

necusjz (approved):

jiasli (approved):

PR #22612, opened 2022-05-25
{Graph} Remove the warning for Microsoft Graph migration

jiasli:
**Related command**
`az ad`

**Description**
This warning is only needed in the first version after migration - 2.37.0. We should remove it in 2.38.0.


yonzhan:
Graph

Issue #22174, opened 2022-04-25
Migrate modules to Microsoft Graph


jiasli:
## Modules to migrate

| modules           | `az ad` | `az role` | `azure-graphrbac` SDK | Owner                                           | Progress                                                   |
| ----------------- | :-----: | :-------: | :-------------------: | ----------------------------------------------- | ---------------------------------------------------------- |
| acs               |    ✔️    |           |           ✔️           | Service: @FumingZhang <br> CLI: @zhoxing-ms     | ✔️ #22361 Fixed incorrect usage of `identifierUri`, # 22649 |
| ams               |    ✔️    |           |           ✔️           |                                                 | ✔️ #22361, #22703                                           |
| appconfig         |         |     ✔️     |                       | Service: @pratiksanglikar <br> CLI: @zhoxing-ms | ✔️ #22361                                                   |
| appservice        |         |     ✔️     |           ✔️           | Service: @panchagnula <br> CLI: @zhoxing-ms     | ✔️ #22819                                                   |
| aro               |         |           |           ✔️           |                                                 | ✔️ #22549                                                   |
| cosmosdb          |         |     ✔️     |                       |                                                 | ✔️ #22432                                                   |
| deploymentmanager |         |     ✔️     |                       |                                                 | ✔️ Only called when `not is_playback`                       |
| eventhubs         |         |           |  ✔️ via `az keyvault`  |                                                 | ✔️ #22361                                                   |
| hdinsight         |         |           |           ✔️           |                                                 | ✔️ #22503                                                   |
| iot               |         |     ✔️     |                       | Service: @digimaun <br> CLI: @zhoxing-ms        | ✔️ #22262                                                   |
| keyvault          |    ✔️    |     ✔️     |           ✔️           | CLI: @evelyn-ys                                 | ✔️ #22188, #22337                                           |
| lab               |         |           |           ✔️           |                                                 | ✔️ #29889                                                   |
| network           |         |     ✔️     |                       |                                                 | ✔️ Only appear in `@live_only`                              |
| profile           |    ✔️    |           |                       |                                                 | ✔️ Only appear in help/error message                        |
| resource          |    ✔️    |     ✔️     |                       | CLI: @zhoxing-ms @cxznmhdcxz                    | ✔️ #22302                                                   |
| servicebus        |         |           |  ✔️ via `az keyvault`  |                                                 | ✔️ #22361                                                   |
[...]

PR #22649, opened 2022-05-27
{AKS} Graph module migration

FumingZhang:
**Related command**

`az aks create`

**Description**

- Graph module migration.
  - Remove most dependencies on graph SDK.
  - Only keep the functions that query object id by app id, which is used by role assignments.
  - Keep the graph SDK related calls used by the `acs` commands unchanged, and plan to remove the deprecated `acs` command group as a whole during the next breaking change window to eliminate the dependency on the legacy graph SDK.
  - Split the global import of legacy graph SDK in `custom.py` into each function, so that even if CLI removes the dependency, import errors should not occur without using these deprecated commands.
- After the changes in this PR, a service principal will no longer be implicitly created for the user when the user has not provided both app id and secret (through options `--service-principal` and `--client-secret`), instead a `RequiredArgumentMissingError` exception would be raised to prompt the users to create it by themselves.
  - Although the default behavior of aks create is to use managed identity as the cluster's identity, users can still set service principal as the cluster's identity by specifying the --service-principal and --client-secret options. And there is still a [piece of code](https://github.com/Azure/azure-cli/blob/62a01da2b5f2de505e37ed8f444e06c5b8dd5f09/src/azure-cli/azure/cli/command_modules/acs/managed_cluster_decorator.py#L1123) that helps create an [application and service principal](https://github.com/Azure/azure-cli/blob/62a01da2b5f2de505e37ed8f444e06c5b8dd5f09/src/azure-cli/azure/cli/command_modules/acs/_graph.py#L188) when the user only provides --client-secret. But with this graph module migration, the previously used Azure AD Graph API will be replaced by the Microsoft Graph API (see [doc](https://docs.microsoft.com/en-us/graph/migrate-azure-ad-graph-overview), which does not support specifying password credential when creating [application](https://docs.microsoft.com/en-us/graph/api/application-post-applications?view=graph-rest-beta&tabs=http) and [service principals](https://docs.microsoft.com/en-us/graph/api/serviceprincipal-post-serviceprincipals?view=graph-rest-beta&tabs=http), which makes it impossible for us to maintain the previous behavior.
  - Therefore, I propose that when the user does not provide the complete service principal information (i.e., provide the app id and secret at the same time), instead of trying to create an application and service principal for the users, an error is thrown to prompt the users to create it by themselves. In my opinion, this change will not be noticeable to most users of aks, because the default create behavior is to use managed identity, and users who use service principal as identity will generally provide both id and secret. In addition, after this change, we will basically remove the legacy dependency on the graph module on the CLI side, making the code more streamlined and conducive to subsequent maintenance.

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acs/_graph.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/_graph.py
@@ -10,11 +10,10 @@ import time
 import uuid
 
 import dateutil
-from azure.cli.command_modules.acs._client_factory import get_graph_rbac_management_client
+from azure.cli.command_modules.acs._client_factory import get_graph_rbac_management_client, cf_graph_client
 from azure.cli.core.azclierror import AzCLIError
 from azure.graphrbac.models import (
     ApplicationCreateParameters,
-    GetObjectsParameters,
     GraphErrorException,
     KeyCredential,
     PasswordCredential,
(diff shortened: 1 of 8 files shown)


yonzhan:
AKS

FumingZhang:
@weinong, @jiasli Would appreciate if you could help review. Thanks in advance!

src/azure-cli/azure/cli/command_modules/acs/_graph.py
```diff
@@ -3,187 +3,39 @@
...
+        result = list(client.user_list(filter="userPrincipalName eq '{}'".format(assignee)))
+    # find by spn
     if not result:
-        result = list(client.service_principals.list(filter="servicePrincipalNames/any(c:c eq '{}')".format(assignee)))
-    if not result:  # assume an object id, let us verify it
-        result = _get_object_stubs(client, [assignee])
+        result = list(client.service_principal_list(filter="servicePrincipalNames/any(c:c eq '{}')".format(assignee)))
+    # find by display name
+    if not result:
+        result = list(client.service_principal_list(filter="displayName eq '{}'".format(assignee)))
```
andyliuliming, line 26: this is a new logic? (by displayName?)

src/azure-cli/azure/cli/command_modules/acs/_graph.py
```diff
@@ -3,187 +3,39 @@
...
-    logger.info("Finished service principal creation")
-    return service_principal, aad_session_key
-
-
-def _create_client_secret():
-    # Add a special character to satisfy AAD SP secret requirements
-    special_char = "$"
-    client_secret = binascii.b2a_hex(os.urandom(10)).decode("utf-8") + special_char
-    return client_secret
+    return result[0]["id"]
```
andyliuliming, line 35: will it possible that the result is not none, but length is zero.

FumingZhang: No, if this is an empty list, the if statement (empty list is converted to False) at the two lines above will raise an exception.

src/azure-cli/azure/cli/command_modules/acs/_graph.py
FumingZhang: Yes, I added it. Now updated the comments with examples.

jiasli (approved):

zhoxing-ms (approved):

jiasli:
Linking https://github.com/Azure/azure-cli/issues/22174

PR #22685, opened 2022-05-31
[ARM] `az bicep build`: Fix #22621: `--stdout` does not work

### LINKED ISSUE #22621: [Regression] `--stdout` does not work for `az bicep build` in v2.37.0

**Related command**
az bicep build

**Describe the bug**
Since v2.37.0 the `--no-restore` flag has to be passed to `bicep build`, as well as `--stdout` to for the compiled template to be printed.

**To Reproduce**
run `az bicep build --stdout -f <path-to-bicep>` and you'll see not output.  Add `--no-restore` and you will.

**Additional context**
It was introduced, I think, with https://github.com/Azure/azure-cli/commit/215960c2a6f1da41f5e60df041b2a75c52ff2ed0.

Note that previously, when `--stdout` was set, we would print the result of calling `run_bicep_command`. When the restore feature was added, the extra if block that was introduced meant that we only printed the result when `--no-restore` was also set.

shenglol:
**Related command**

**Description**
v2.37.0 has a regression where `--stdout` stops working for `az bicep build`. The PR fixes it.

Closes #22621.

**Testing Guide**
`az bicep build -f main.bicep --stdout`

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/resource/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/resource/custom.py
@@ -3684,7 +3684,7 @@ def build_bicep_file(cmd, file, stdout=None, outdir=None, outfile=None, no_resto
         args += ["--stdout"]
 
     output = run_bicep_command(args)
-    
+
     if stdout:
         print(output)
 
--- a/src/azure-cli/azure/cli/command_modules/resource/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/resource/custom.py
@@ -3678,10 +3678,10 @@ def build_bicep_file(cmd, file, stdout=None, outdir=None, outfile=None, no_resto
         args += ["--outdir", outdir]
     if outfile:
         args += ["--outfile", outfile]
-    if stdout:
-        args += ["--stdout"]
     if no_restore:
         args += ["--no-restore"]
+    if stdout:
+        args += ["--stdout"]
         print(run_bicep_command(args))
         return
     run_bicep_command(args)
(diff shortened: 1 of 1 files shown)


ellismg (approved):
LGTM.

Just wondering - imagine that we only called `run_bicep_command(args)` from a single place (i.e. as we do on line 3687) and just captured the output and then depending on if `stdout` was set we printed the value, otherwise return?

I'm wondering if that would make it less likely, going forward, for issues like this to be introduced?

shenglol:
> LGTM.
> 
> Just wondering - imagine that we only called `run_bicep_command(args)` from a single place (i.e. as we do on line 3687) and just captured the output and then depending on if `stdout` was set we printed the value, otherwise return?
> 
> I'm wondering if that would make it less likely, going forward, for issues like this to be introduced?

Good point. I'll do some refactoring.

yonzhan:
ARM

MrMCake:
@shenglol Please merge this as soon as you can. This bug causes all CARML pipelines running on Ubuntu-20.04 to fail as the 'az bicep build' command does not return any more results.
The workaround of adding '--no-restore' may have other un-intended complications so we may wait for this fix to be merged.

shenglol:
@zhoxing-ms Could you help approve and merge the PR?

shenglol:
> @shenglol Please merge this as soon as you can. This bug causes all CARML pipelines running on Ubuntu-20.04 to fail as the 'az bicep build' command does not return any more results. The workaround of adding '--no-restore' may have other un-intended complications so we may wait for this fix to be merged.

I'd suggest using the workaround below:
```
az bicep restore -f main.bicep
az bicep build -f main.bicep --no-restore
```
Or we can just directly invoke Bicep CLI available at `~/.azure/bin/bicep`.

zhoxing-ms (approved):

PR #22739, opened 2022-06-06
[Graph] Support special characters in object names

### LINKED ISSUE #22700: `az ad group show --group` breaks with group names that use `+` in v2.37
**Related command**
`az ad group show --group <group_name>`

**Describe the bug**
Prior to `v2.37` of the azure-cli, this command would complete successfully:
`az ad group show --group some_group+`
As of `v2.37` that same command returns the following error:
`No group matches the name of 'some_group+'`

If you execute the same command on another group that does not have the `+` in its name:
`az ad group show --group some_group`
...then the command completes successfully.

**To Reproduce**
1. Ensure `az --version` returns a version < 2.37.
2. `az ad group show --group some_group+` - command successful.
3. Upgrade azure-cli to version 2.37.
4. `az ad group show --group some_group+` - command fails.

Further, no amount of shell escaping results in success:
- `az ad group show --group "some_group+"` - command fails.
- `az ad group show --group 'some_group+'` - command fails.
- `az ad group show --group some_group\+` - command fails.

**Expected behavior**
I suspect this behavior is a result of the [recent move from the `Active Directory Graph API` to the `Microsoft Graph API` when upgrading to v2.37 of the azure-cli](https://docs.microsoft.com/en-us/cli/azure/microsoft-graph-migration).  That said, I think there is a reasonable expectation that rules around illegal characters should at least be harmonized between the two graphs before migrating from one to the other.

I found no mention of this behavior in [the documentation I was able to locate on the differences between the two graphs](https://docs.microsoft.com/en-us/graph/migrate-azure-ad-graph-audit-api-use).

**Environment summary**
Experienced in Azure Cloud Shell which recently upgraded azure-cli to v2.37, and also in an installed azure-cli instance on a Manjaro Linux workstation.

jiasli:
**Related command**
`az ad group show`

Close #22700

### Graph service

Even though Graph service recommends encoding query parameters (https://docs.microsoft.com/en-us/graph/query-parameters#encoding-query-parameters), it is not explicit on which characters should be encoded:

> The values of query parameters should be percent-encoded.

```sh
# unencoded
GET https://graph.microsoft.com/v1.0/users?$filter=startswith(givenName, 'J')
# encoded
GET https://graph.microsoft.com/v1.0/users?$filter=startswith(givenName%2C+'J')
```

In the above example, only a subset of reserved characters defined by [rfc3986](https://www.rfc-editor.org/rfc/rfc3986#section-2.2) are encoded, like `,<SPACE>`, but `()'` are not (considered "safe"). Instead, `startswith(givenName, 'J')` should be encoded as

```py
from urllib.parse import quote
print(quote("startswith(givenName, 'J')"))
# startswith%28givenName%2C%20%27J%27%29
# startswith(  givenName,     '  J'  )  <- original
```

Graph service SHOULD explicitly specify which characters should be encoded (https://github.com/microsoftgraph/microsoft-graph-docs/issues/17155).

### CLI

`GraphClient` uses `requests` to make HTTP requests. `requests` uses `urllib3`'s encoding mechanism which treats `,+` as "safe" and doesn't encode them.

https://github.com/urllib3/urllib3/blob/342aff50ff300d96a58e9be22f27fcee771ce98d/src/urllib3/util/url.py#L73-L79

```py
UNRESERVED_CHARS = set(
    "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789._-~"
)
SUB_DELIM_CHARS = set("!$&'()*+,;=")
USERINFO_CHARS = UNRESERVED_CHARS | SUB_DELIM_CHARS | {":"}
PATH_CHARS = USERINFO_CHARS | {"@", "/"}
QUERY_CHARS = FRAGMENT_CHARS = PATH_CHARS | {"?"}
```

Even though Azure CLI 2.37.0 doesn't encode `,`, the API still works:

```
> az ad group list --display-name azure-cli-test1 --debug
cli.azure.cli.core.util: Request URL: "https://graph.microsoft.com/v1.0/groups?$filter=startswith(displayName,'azure-cli-test1')"
```

### SDK

Previously, `graphrbac` SDK encodes all special characters, including ` (),'`.

https://github.com/Azure/msrest-for-python/blob/c97db9df00c7802b1a1af3ab8dcdbf6a5acbbdd7/msrest/serialization.py#L659

```py
                output = quote(str(output), safe='')
```

```
> az ad group show -g azure-cli-test+ --debug
msrest.http_logger: Request URL: 'https://graph.windows.net/54826b22-38d6-4fb2-bad9-b7b93a3e9c5a/groups?$filter=displayName%20eq%20%27azure-cli-test%2B%27&api-version=1.6'

> az ad group list --display-name azure-cli-test+ --debug
msrest.http_logger: Request URL: 'https://graph.windows.net/54826b22-38d6-4fb2-bad9-b7b93a3e9c5a/groups?$filter=startswith%28displayName%2C%27azure-cli-test%2B%27%29&api-version=1.6'
```

This is consistent with the Graph document, but inconsistent with Graph document's example.

### Change

This PR encodes all special characters, like the old SDK does, in order to support special characters such as `+/` in object names.


yonzhan:
Graph

src/azure-cli-testsdk/azure/cli/testsdk/utilities.py
```diff
@@ -182,26 +182,33 @@ def process_response(self, response):
...
-    def __init__(self, test_user, mock_user):
-        self.test_user = test_user
-        self.mock_user = mock_user
+class MSGraphNameReplacer(RecordingProcessor):
+    """If a name appears in URL, it should be percent-encoded, e.g.
+      - test-name+/ is encoded as test-name%2B%2F
+      - example@example.com is encoded as example%40example.com
+
+    Theoretically, GeneralNameReplacer should also follow this pattern, but since we haven't seen any ARM name that
+    is percent-encoded, we only replace percent-encoded names for MS Graph for better test performance.
```
jiasli, line 191: I tried to change `GeneralNameReplacer`'s logic to also replace encoded names, but it turns out to be an overdo since ARM resource names usually don't need to be encoded, so we only do it specially for MS Graph.

jiasli:
Luckily, VCRPY can also match URLs that are percent-encoded, so we don't need to re-record all tests.

evelyn-ys (approved):

calvinhzy (approved):

src/azure-cli-testsdk/azure/cli/testsdk/utilities.py
jsntcy: Do we have test to cover the changes?

jsntcy (approved):

src/azure-cli-testsdk/azure/cli/testsdk/utilities.py
jiasli: `test_special_characters` is exactly for that purpose.

PR #22802, opened 2022-06-09
[ACR] Fix Azure/acr#613: `az acr task run`: Add dockerfile to source upload if context is local directory

oleterno-equinor:
**Related command**

`az acr task run`

**Description**

`az acr task run` will trigger a run of an existing ACR task, but it behaves differently when using a remote or local context. Having a `.dockerignore` which contains `Dockerfile` [is valid docker configuration](https://docs.docker.com/engine/reference/builder/#dockerignore-file:~:text=You%20can%20even%20use%20the%20.dockerignore%20file%20to%20exclude%20the%20Dockerfile%20and%20.dockerignore%20files.%20These%20files%20are%20still%20sent%20to%20the%20daemon%20because%20it%20needs%20them%20to%20do%20its%20job.%20But%20the%20ADD%20and%20COPY%20instructions%20do%20not%20copy%20them%20to%20the%20image.) (scroll down to end of topic). 

When the command is run and a local context is provided, the task run fails because the Dockerfile is not included when the source is uploaded. The behaviour is different if the context is a remote git repository, or when using the `az acr build` command, in which case the task succeeds both if the Dockerfile is present in `.dockerignore` and not. 

This is PR will fix this problem, and the behaviour will be consistent with the `az acr build` command, where the Dockerfile is uploaded when the local source context is uploaded. If `--file` is provided as an argument for `az acr task run`, it will include the Dockerfile when the local source is uploaded. If the `docker_file_path` variable in `prepare_source_location` is a task file or inline yaml, it will not be uploaded, as per the initial behaviour. 

Fix https://github.com/Azure/acr/issues/613

**Testing Guide**

It is expected that the Dockerfile in the task is the same Dockerfile passed in with the `--file` argument. 
Example task.yaml:
```yaml
version: v1.1.0
steps:
  - build: -t testacr.azurecr.io/your-image:tag -f local.Dockerfile .
```
Command:
```sh
az acr task run \
  --name acr-task-name \
  --registry testacr \
  --context /path/to/local/context \
  --file /path/to/local/context/local.Dockerfile
```

**History Notes**

[ACR] `az acr task run`: Add Dockerfile to source upload if context is local directory

---


yonzhan:
ACR

src/azure-cli/azure/cli/command_modules/acr/_validators.py
```diff
@@ -140,3 +140,8 @@ def validate_repository(namespace):
         if ':' in namespace.repository:
             raise InvalidArgumentValueError("Parameter 'name' refers to a repository and"
                                             " should not include a tag or digest.")
+
+
+def validate_docker_file_path(docker_file_path):
+    if not os.path.isfile(docker_file_path):
+        raise CLIError("Unable to find '{}'.".format(docker_file_path))
```
zhoxing-ms, line 147: Could you please use the specific error type `FileOperationError` instead of `CLIError`?

zhoxing-ms:
Could you add some tests for those change?

zhoxing-ms:
Please note that if you do not resolve those before Thursday, this PR will not be able to catch up with the release of this sprint, we can only postpone it to the next sprint **(08-02)**

zhoxing-ms:
@rosanch Could you please review this PR from the perspective of ACR team?

zhoxing-ms:
@oleterno-equinor If you confirm that this error does not affect this PR feature, we can skip this test first to avoid blocking PR from being merged in this sprint. You can solve this problem in the subsequent sprint

oleterno-equinor:
Yes, the error does not affect the feature when running the commands outside the test. I reverted the changes related to the test.

zhoxing-ms (approved):

PR #22861, opened 2022-06-14
{CI} Migrate fabric bot from portal to config-as-code

wangzelin007:
---

Why migrate fabric bot from portal to config-as-code:
- As fabric bot team are preparing to move permanently to "Config As Code", they have hidden the save button to disable the capability to update the configuration in portal.
- We need to update milestone every month, so we need migrate before 2022-07-05.

What will happen after migration?
- The fabricbot.json file in our repo that will always take precedence than config in portal.
- All update need to commit in .github/fabricbot.json.

### DIFF
--- a/.github/fabricbot.json
+++ b/.github/fabricbot.json
@@ -1,0 +1,12009 @@
+{
+  "version": "1.0",
+  "tasks": [
+    {
+      "taskType": "trigger",
+      "capabilityId": "IssueResponder",
+      "subCapability": "IssuesOnlyResponder",
+      "version": "1.0",
+      "config": {
+        "taskName": "Add needs triage label to new issues",
+        "conditions": {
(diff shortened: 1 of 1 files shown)


jiasli:
This is a very big JSON. How do we maintain this file in the future? Manually?

jiasli (approved):

wangzelin007:
> This is a very big JSON. How do we maintain this file in the future? Manually?

Yes manually,
Before migration, we maintained hundreds of tasks on the fabric bot portal.
After migrate to config-as-code, we can use the IDE, which actually will be easier to maintain.

jsntcy (approved):

wangzelin007:
Successful migration to config-as-code:
https://github.com/Azure/azure-cli-extensions/issues/5000

PR #22864, opened 2022-06-14
{Packaging} Bump Python version to 3.10.5

jiasli:
**Description**

Bump Python version to [3.10.5](https://www.python.org/downloads/release/python-3105/).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,9 +3,9 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 #---------------------------------------------------------------------------------------------
 
-ARG PYTHON_VERSION="3.10.4"
+ARG PYTHON_VERSION="3.10"
 
-FROM python:${PYTHON_VERSION}-alpine3.15
+FROM python:${PYTHON_VERSION}-alpine
 
 ARG CLI_VERSION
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -3,9 +3,9 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 #---------------------------------------------------------------------------------------------
 
-ARG PYTHON_VERSION="3.10.4"
+ARG PYTHON_VERSION="3.10"
 
-FROM python:${PYTHON_VERSION}-alpine3.15
+FROM python:${PYTHON_VERSION}-alpine
```
jiasli, line 8: Unpin the version of Python and Alpine Linux, so that we always use the latest and avoid the necessity of frequently bumping versions (https://github.com/Azure/azure-cli/pull/21079).

yonzhan:
Packaging

Dockerfile
bebound: Is there any cache mechanism when pull images?

bebound (approved):

calvinhzy (approved):

Dockerfile
jiasli: Locally yes. If you don't run `docker pull`, existing docker image will be used, but on ADO there is no cache, so it will always pull the latest.

PR #22873, opened 2022-06-15
{CI} Fix the bug that live test does not generate the latest index.html

wangzelin007:
- live test does not generate the latest index.html file.
image
  - storage has a breaking change, need add --overwrite to `az storage blob upload` command to overwrite the old index.html file.
  - The password of the azureclilivetest account needs to be updated regularly, otherwise it will expire.
image


kairu-ms (approved):

yonzhan:
CI

PR #22975, opened 2022-06-23
[Cosmos DB] Add support to create containers with client encryption policy

kr-santosh:
**Related command**
 az cosmosdb sql container create

**Description**
This PR brings in the support for users to configure a client encryption policy on the container. It is similar to any other policy like Indexing policy which can be configured on a container. The policy format consists of "includedPaths" property which takes a list of the following

- path : which is the top level path in the JSON document that the user wants to encrypt.
- clientEncryptionKeyId : the key identifier of the key that is used to encrypt the data.
- encryptionAlgorithm : the encryption algorithm used to encrypt the property value.
- encryptionType : the encryption type used by the encryptionAlgorithm.

Finally it includes a format version for the policy.

**Testing Guide**
az cosmosdb sql container create -g MyResourceGroup -a MyAccount -d MyDatabase -n MyContainer --partition-key-path "/my/path"  --ttl 3000 --throughput "700" --client-encryption-policy '{\"includedPaths\": [{\"path\": \"/SubTotal\",\"clientEncryptionKeyId\": \"key1\",\"encryptionAlgorithm\": \"AEAD_AES_256_CBC_HMAC_SHA256\",\"encryptionType\": \"Deterministic\"}],\"policyFormatVersion\": 2}'

Note:
In order to create a container with a client encryption policy. The key identifier which is included in the path has to be created upfront (which is part of Cosmos DB database resource). The policy cannot be configured without the key created upfront and has to be present in the database. 

Since we cannot support Client Encryption Key creation CLI command in Az Cli I have manually tested the command. The reason why it cannot be supported is because the Client Encryption Key creation process requires an encryption library which currently does not have python support. We currently support it in SDK and Powershell Cmdlets since both have .NET support and the library is available only in C# .NET and Java.

**History Notes**
None

---


yonzhan:
Cosmos DB

src/azure-cli/azure/cli/command_modules/cosmosdb/custom.py
```diff
@@ -540,6 +544,9 @@ def _populate_sql_container_definition(sql_container_resource,
     if unique_key_policy is not None:
         sql_container_resource.unique_key_policy = unique_key_policy
 
+    if client_encryption_policy is not None:
+        _validate_and_populate_client_encryption_policy(sql_container_resource, client_encryption_policy)
```
evelyn-ys, line 548: You can add the validator in parameter definition then just simply pass the value in custom.py

src/azure-cli/azure/cli/command_modules/cosmosdb/custom.py
```diff
@@ -549,6 +556,109 @@ def _populate_sql_container_definition(sql_container_resource,
...
 
+def _validate_and_populate_client_encryption_policy(sql_container_resource, client_encryption_policy):
+    if client_encryption_policy is not None:
+        from azure.cli.core.util import shell_safe_json_parse
+        data = shell_safe_json_parse(json.dumps(client_encryption_policy))
+
+        if "includedPaths" in data:
+            includedPaths = data['includedPaths']
+        else:
+            raise CLIError(None, "includedPaths missing in Client Encryption Policy. Please verify your Client Encryption Policy JSON string")
```
evelyn-ys, line 567: Please use the error type defined in`azure.cli.core.azclierror`, eg, `InvalidArgumentValueError`

kr-santosh: Sure.

kr-santosh:
> Can you add test for the new parameter?

As I have mentioned in the PR description. Since we cannot support Client Encryption Key creation CLI command in Az Cli I have manually tested the command. The reason why it cannot be supported is because the Client Encryption Key creation process requires an encryption library which currently does not have python support. We currently support it in SDK and PowerShell Cmdlets since both have .NET support and the library is available only in C# .NET and Java.

kairu-ms (approved):
LGTM on the network' recordings

evelyn-ys (approved):

PR #23132, opened 2022-07-06
[ACR] BREAKING CHANGE: Update manifest list-referrers to comply with RC1 ORAS spec

mabenedi:
**Related command**
az acr manifest list-referrers

**Description**
Changed command implementation to comply with the new ORAS spec. Behavior to user is unchanged. 

**History Notes**

{acr} CHANGE: `az acr manifest list-referrers`: Update manifest list-referrers to comply with RC1 ORAS spec

---


yonzhan:
acr

northtyphoon (approved):
LGTM

zhoxing-ms:
Is the change introduced by this PR perceived by users? 
If not, we don't need to write the PR title into the history notes, so please change the prefix of the PR title from `[]` to `{}`

mabenedi:
> @mabenedi Is the change introduced by this PR perceived by users?
If not, we don't need to write the PR title into the history notes, so please change the prefix of the PR title from [] to {}

There is one small change perceived by users. The JSON object now nested inside a 'referrers' key. Previously, this key was 'references'

image

zhoxing-ms:
> There is one small change perceived by users. The JSON object now nested inside a 'referrers' key. Previously, this key was 'references'

@mabenedi Please note that modifying the output property name will cause breaking change to users. Is it necessary to modify this name?

mabenedi:
> > There is one small change perceived by users. The JSON object now nested inside a 'referrers' key. Previously, this key was 'references'
> 
> @mabenedi Please note that modifying the output property name will cause breaking change to users. Is it necessary to modify this name?

It is unfortunately necessary in order to comply with the spec. The change is actually happening on the backend. To be honest, it might be considered a bug that it ever said 'references' at all. 

If it makes you feel better our telemetry shows very low usage on this API as this is still a very early preview feature (and much of that usage comes from the OSS ORAS client itself, not Az CLI).

zhoxing-ms (approved):
Since the breaking change has been appeared on the server side and the current usage of this command is very low, so CLI side can be modified directly. 
I modified the PR title to indicate that this change is a breaking change

PR #23205, opened 2022-07-14
[Core] Support specifying `telemetry.push_interval_in_hours` to force push telemetry cache file

evelyn-ys:
**Description**

CLI telemetry will only be uploaded when:
- 10 mins after last sync
- local cache file is larger than 128k and has been rotated with cache.1 cache.2 files

This could cause long time delay for pushing telemetry for customers using CLI not that frequently. It could take several days.

This PR set a threshold for cache file upload. If it has been 24 hours since last push, then cache file will be uploaded even if it's less than 128k.

The threshold can be set manually with `az config set telemetry.push_data_threshold=N`. It only takes effects when N >= 1. Default vaule for N is 24.

---

### DIFF
--- a/src/azure-cli-telemetry/azure/cli/telemetry/components/records_collection.py
+++ b/src/azure-cli-telemetry/azure/cli/telemetry/components/records_collection.py
@@ -8,6 +8,7 @@ import os
 import shutil
 import stat
 import tempfile
+from knack.config import CLIConfig
 
 
 class RecordsCollection:
@@ -35,8 +36,13 @@ class RecordsCollection:
         if not os.path.isdir(folder):
             return
 
-        # sort the cache files base on their last modification time.
+        # Collect all cache.x files. If it has been a long time since last sent, also collect cache file itself.
         candidates = [(fn, os.stat(os.path.join(folder, fn))) for fn in os.listdir(folder) if fn != 'cache']
+        if os.path.exists(os.path.join(folder, 'cache')) and \
+                datetime.datetime.now() - self._last_sent > datetime.timedelta(hours=self._get_threshold_config()):
+            candidates.append(('cache', os.stat(os.path.join(folder, 'cache'))))
+
+        # sort the cache files base on their last modification time.
         candidates = [(fn, file_stat) for fn, file_stat in candidates if stat.S_ISREG(file_stat.st_mode)]
         candidates.sort(key=lambda pair: pair[1].st_mtime, reverse=True)  # move the newer cache file first
 
@@ -65,6 +71,12 @@ class RecordsCollection:
                       ignore_errors=True,
                       onerror=lambda _, p, tr: self._logger.error('Fail to remove file %s', p))
         self._logger.info('Remove directory %s', tmp)
+
+    def _get_threshold_config(self):
+        config = CLIConfig(config_dir=self._config_dir)
+        threshold = config.getint('telemetry', 'push_data_threshold', fallback=24)
+        # the threshold for push telemetry can't be less than 1 hour, default value is 24 hours
+        return threshold if threshold >= 1 else 24
 
     def _read_file(self, path):
         """ Read content of a telemetry cache file and parse them into records. """
--- a/src/azure-cli-telemetry/azure/cli/telemetry/components/records_collection.py
+++ b/src/azure-cli-telemetry/azure/cli/telemetry/components/records_collection.py
@@ -28,6 +28,7 @@ class RecordsCollection:
     def next_send(self):
         return self._next_send
 
+    # pylint: disable=line-too-long
     def snapshot_and_read(self):
         """ Scan the telemetry cache files and move all the rotated files to a temp directory. """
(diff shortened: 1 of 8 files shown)


yonzhan:
I like this: It only takes effects when N >= 1

src/azure-cli-telemetry/azure/cli/telemetry/components/records_collection.py
```diff
@@ -66,6 +71,12 @@ def snapshot_and_read(self):
                       onerror=lambda _, p, tr: self._logger.error('Fail to remove file %s', p))
         self._logger.info('Remove directory %s', tmp)
 
+    def _get_threshold_config(self):
+        config = CLIConfig(config_dir=self._config_dir)
+        threshold = config.getint('telemetry', 'push_data_threshold', fallback=24)
```
jiasli, line 76: I think "threshold" is not very straightforward. How about something like

```suggestion
        threshold = config.getint('telemetry', 'push_data_max_wait_time', fallback=24)
```

src/azure-cli-telemetry/azure/cli/telemetry/components/records_collection.py
```diff
@@ -35,8 +37,11 @@ def snapshot_and_read(self):
         if not os.path.isdir(folder):
             return
 
+        # Collect all cache.x files. If it has been a long time since last sent, also collect cache file itself.
+        include_cache = datetime.datetime.now() - self._last_sent > datetime.timedelta(hours=self._get_threshold_config())
```
jiasli, line 41: We should determine whether we want 

- since last sent, which compares `TELEMETRY_NOTE_NAME`'s content: https://github.com/Azure/azure-cli/blob/be601b7cfe542358f954e180b61b02c98fcfb27b/src/azure-cli-telemetry/azure/cli/telemetry/components/telemetry_note.py#L40-L41, or
- since last check, which compares `TELEMETRY_NOTE_NAME`'s `st_mtime` https://github.com/Azure/azure-cli/blob/be601b7cfe542358f954e180b61b02c98fcfb27b/src/azure-cli-telemetry/azure/cli/telemetry/util.py#L34-L38

evelyn-ys: To clarify, last sent is the record time of last uploaded, last check is the time we check if there's any cache need to be uploaded. So last check can be updated every 10 mins even there's no cache uploaded.

Think about this scenario if we use last check time to compare:
1. `07-14 15:00`, cx runs a cli command. A new record has been stored in cache file while the file size is 6k.
-  Since no telemetry uploaded, so last sent is the time of previous record, let's say it's `07-14 14:30`.
- Last check has been updated to `07-14 15:00`.
3. Then `07-15 14:40`, cs runs another command. Another record has been stored in cache file and the file size grows to 9k.
-  now(`07-15 14:40`) - last check time(`07-14 15:00`) is less than 24 hours, so cache file won't be uploaded
- last sent time is still `07-14 14:30`, last check time is updated to `07-15 14:40`
5. Another 10 mins later, `07-15 14:50`, cs runs another command. Another record has been stored in cache file and the file size grows to 12k.
- now(`07-15 14:50`) - last check time(`07-15 14:40`) is less than 24 hours, so cache file won't be uploaded
- last sent time is still `07-14 14:30`, last check time is updated to `07-15 14:50`
......
......
If customer keeps running cli commands, last check time will keep updating, cache file will never be uploaded until rotated with cache.1

src/azure-cli-telemetry/azure/cli/telemetry/components/records_collection.py
evelyn-ys: How about `push_data_max_period_in_hours`? `time` can be hours, mins or days. I want to make it more clear.

src/azure-cli-telemetry/azure/cli/telemetry/components/records_collection.py
jiasli: Very good analysis! Thanks for clarifying it. 😊

src/azure-cli-telemetry/azure/cli/telemetry/components/records_collection.py
jiasli: Great! This is very good point!

src/azure-cli-telemetry/azure/cli/telemetry/components/records_collection.py
```diff
@@ -27,6 +28,7 @@ def __iter__(self):
     def next_send(self):
         return self._next_send
 
+    # pylint: disable=line-too-long
```
jiasli, line 31: Better not to disable pylint checks but to comply to its check. 😉

jiasli:
Perhaps we can make this changelog available in Core section? Also, we need to bump the version of `azure-cli-telemetry`.

jiasli (approved):

jiasli:
This is a really neat solution. Nice work!

Issue #23160, opened 2022-07-08
az ad sp create-for-rbac --name "GitHub-Actions" --role contributor --scopes /subscriptions/{iuh886c0aa-f9b1-0000-8od7-bf98681r8713} --sdk-auth


duvergonzalez:
## Describe the bug

**Command Name**
az ad sp create-for-rbac --name "GitHub-Actions" --role contributor --scopes /subscriptions/{iuh886c0aa-f9b1-0000-8od7-bf98681r8713} --sdk-auth

**Errors:**
```
Option '--sdk-auth' has been deprecated and will be removed in a future release.
Creating 'contributor' role assignment under scope '/subscriptions/{e906c0aa-f6b1-4341-8bd4-bf98681c8713}'
  Role assignment creation failed.

The command failed with an unexpected error. Here is the traceback:
'e906c0aa-f6b1-4341-8bd4-bf98681c8713'
Traceback (most recent call last):
  File "/opt/az/lib/python3.10/site-packages/knack/cli.py", line 231, in invoke
    cmd_result = self.invocation.execute(args)
  File "/opt/az/lib/python3.10/site-packages/azure/cli/core/commands/__init__.py", line 663, in execute
    raise ex
  File "/opt/az/lib/python3.10/site-packages/azure/cli/core/commands/__init__.py", line 726, in _run_jobs_serially
    results.append(self._run_job(expanded_arg, cmd_copy))
  File "/opt/az/lib/python3.10/site-packages/azure/cli/core/commands/__init__.py", line 718, in _run_job
    return cmd_copy.exception_handler(ex)
  File "/opt/az/lib/python3.10/site-packages/azure/cli/command_modules/role/commands.py", line 54, in graph_err_handler
    raise ex
  File "/opt/az/lib/python3.10/site-packages/azure/cli/core/commands/__init__.py", line 697, in _run_job
    result = cmd_copy(params)
  File "/opt/az/lib/python3.10/site-packages/azure/cli/core/commands/__init__.py", line 333, in __call__
    return self.handler(*args, **kwargs)
  File "/opt/az/lib/python3.10/site-packages/azure/cli/core/commands/command_operation.py", line 121, in handler
    return op(**command_args)
  File "/opt/az/lib/python3.10/site-packages/azure/cli/command_modules/role/custom.py", line 1255, in create_service_principal_for_rbac
    _create_role_assignment(cmd.cli_ctx, role, sp_oid, None, scope, resolve_assignee=False,
[...]

Issue #15739, opened 2020-10-31
Please help - Azure command not working


Anushka-codergirl:
## Describe the bug

**Command Name**
`az ad sp create-for-rbac`

**Errors:**
```
'e73992a4-44cb-40bd-8637-37b110c63102'
Traceback (most recent call last):
Temp\pip-unpacked-wheel-49qlvzn7\knack\cli.py, ln 215, in invoke
cli\core\commands\__init__.py, ln 654, in execute
cli\core\commands\__init__.py, ln 718, in _run_jobs_serially
cli\core\commands\__init__.py, ln 709, in _run_job
cli\command_modules\role\commands.py, ln 69, in graph_err_handler
cli\core\commands\__init__.py, ln 688, in _run_job
cli\core\commands\__init__.py, ln 325, in __call__
...
Temp\pip-unpacked-wheel-kmz4tfp1\msrest\service_client.py, ln 168, in get
Temp\pip-unpacked-wheel-kmz4tfp1\msrest\service_client.py, ln 108, in _request
Temp\pip-unpacked-wheel-kmz4tfp1\msrest\service_client.py, ln 151, in format_url
KeyError: 'e73992a4-44cb-40bd-8637-37b110c63102'
```

## To Reproduce:
Steps to reproduce the behavior. Note that argument values have been redacted, as they may contain sensitive information.

- _Put any pre-requisite steps here..._
- `az ad sp create-for-rbac --name {} --role {} --scopes {} --sdk-auth`

## Environment Summary
```
Windows-10-10.0.18362-SP0
Python 3.6.8
Installer: MSI

azure-cli 2.14.0

```

yonzhan:
add to S179

RakeshMohanMSFT:
This is fixed

jiasli:
We are refining the example in https://github.com/Azure/azure-cli/pull/23206

PR #23206, opened 2022-07-14
{Role} `az ad sp create-for-rbac`: Remove curly brackets from example command

jiasli:
**Related command**
`az ad sp create-for-rbac`

**Description**
The example for `az ad sp create-for-rbac` specifies scope as `/subscriptions/{subscriptionId}/resourceGroups/{resourceGroup1}`. Even though curly brackets denote placeholders in the example, some users may include the curly brackets in the actual scope as `/subscriptions/{e906c0aa-f6b1-4341-8bd4-bf98681c8713}`, which causes failure (#15739, #23160, #23198):

```
KeyError: 'e906c0aa-f6b1-4341-8bd4-bf98681c8713'
```

This PR removes curly brackets to eliminate unnecessary confusion.

Question: Will `$subscription_id` also cause confusion and will some users use it as `/subscriptions/$e906c0aa-f6b1-4341-8bd4-bf98681c8713`?

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/role/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/role/_help.py
@@ -496,9 +496,9 @@ examples:
   - name: Create without role assignment.
     text: az ad sp create-for-rbac
   - name: Create using a custom display name.
-    text: az ad sp create-for-rbac -n "MyApp"
+    text: az ad sp create-for-rbac -n MyApp
   - name: Create with a Contributor role assignments on specified scopes. To retrieve current subscription ID, run `az account show --query id --output tsv`.
-    text: az ad sp create-for-rbac -n "MyApp" --role Contributor --scopes /subscriptions/{subscriptionId}/resourceGroups/{resourceGroup1} /subscriptions/{subscriptionId}/resourceGroups/{resourceGroup2}
+    text: az ad sp create-for-rbac -n MyApp --role Contributor --scopes /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/resourceGroup1 /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/resourceGroup2
   - name: Create using a self-signed certificate.
     text: az ad sp create-for-rbac --create-cert
   - name: Create using a self-signed certificate, and store it within KeyVault.


yonzhan:
Role

jiasli:
Using curly brackets as placeholders also appears in other places:

https://github.com/Azure/azure-cli/blob/630ce428a76374eaf2a76bed3c5c1744813fe509/src/azure-cli/azure/cli/command_modules/role/_help.py#L193-L196

I was wondering what the best way is to convey the idea of "curly brackets are placeholders" in all examples, without the user reading some other documentation.

jiasli:
Besides, REST API doc also uses curly brackets as placeholders, such as

https://docs.microsoft.com/en-us/rest/api/resources/resource-groups/list

image

rayluo:
> Using curly brackets to denote placeholders also appears in other places:
> 
> https://github.com/Azure/azure-cli/blob/630ce428a76374eaf2a76bed3c5c1744813fe509/src/azure-cli/azure/cli/command_modules/role/_help.py#L193-L196
> 
> I was wondering what the best way is to convey the idea of "curly brackets denote placeholders" in all examples, without the user reading some other documentation.

To be fair, when an example shows something like `https://contoso.com/endpoint/{subscriptionId}`, it is not intrinsically clear whether the curly brackets should be part of the input. Can't blame customer for attempting `https://contoso.com/endpoint/{asdf}` first, although we would hope they can try `https://contoso.com/endpoint/asdf` if their first attempt did not work.

Alternatively, we can consider using ALL_CAPS as placeholder, like this `https://contoso.com/endpoint/SUBSCRIPTION_ID`. They still look noticeable, and they have no bracket residual left behind.

Just my two cents.

chasewilson:
I'm fine with attempting something new and seeing if it improves the experience. @jiasli what's your call? We'll go with that and monitor to see if the problem continues.

jiasli:
@chasewilson, which one? The solution in PR or ALL_CAPS?

src/azure-cli/azure/cli/command_modules/role/_help.py
```diff
@@ -513,9 +513,19 @@
   - name: Create without role assignment.
     text: az ad sp create-for-rbac
   - name: Create using a custom display name.
-    text: az ad sp create-for-rbac -n "MyApp"
-  - name: Create with a Contributor role assignments on specified scopes. To retrieve current subscription ID, run `az account show --query id --output tsv`.
-    text: az ad sp create-for-rbac -n "MyApp" --role Contributor --scopes /subscriptions/{subscriptionId}/resourceGroups/{resourceGroup1} /subscriptions/{subscriptionId}/resourceGroups/{resourceGroup2}
```
evelyn-ys, line 518: How about simply replace `{subscriptionId}` with `00000000-0000-0000-0000-000000000000` and replace `{resourceGroup1}` with `myresourcegroup`. The example text would be:
```
text: az ad sp create-for-rbac -n "MyApp" --role Contributor --scopes /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/myresourcegroup1 /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/myresourcegroup2
```

jsntcy:
https://developers.google.com/style/placeholders
Vote for ALL_CAPS

rayluo:
> https://developers.google.com/style/placeholders Vote for ALL_CAPS

FWIW, recently we encountered a support case from elsewhere, that was somehow also caused by unknowingly prepending a `$` in front of

evelyn-ys (approved):

PR #23419, opened 2022-08-02
[App Config] BREAKING CHANGE: Key-value import and export improvements

albertofori:
**Related command**
`az appconfig kv import`
`az appconfig kv export`

**Description**
1. Adds validations to JSON file imports to ensure that only valid JSON objects and arrays are imports. Other JSON primitives will now be rejected. Users importing from a JSON file should provide data formatted as a valid JSON _object_ or _array_. The value can be a string, integer, array or an object. 
2. Prevents issue of dropping some key-values when trying to convert integer keys to an array. Reverts to exporting key-values as objects when keys are not continuous integers starting from 0, or both strings and integer keys are found. This ensures that no keys are dropped. 

**Testing Guide**

**History Notes**

[App Config] BREAKING CHANGE: `az appconfig kv import`: Add validation to JSON file import to ensure that only valid JSON objects are imports
[App Config] BREAKING CHANGE: `az appconfig kv export`: Update the array conversion logic to prevent dropping keys during export

---


yonzhan:
App Configuration

albertofori:
Hi @zhoxing-ms, @yonzhan, 
Please this change does not change any command syntax, however, it updates the behaviour of existing functionality. I would like to confirm if that is considered a breaking change in the CLI just so I could update the hitory accordingly. Thanks!

zhoxing-ms:
@albertofori Yes, if this PR will change the existing business behavior of users, it will cause breaking change.
It is recommended to implement breaking change in the unified breaking change window of CLI *(Microsoft Build and Ignite)*, so as to reduce it's impact on users
For more details about CLI breaking change process, please refer to this doc [breaking_changes.md](https://msazure.visualstudio.com/One/_git/AzureCliTools-Wiki?version=GBzhoxing-test/update-breaking-changes&path=/src/teams_docs/azcli_docs/breaking_changes.md&_a=preview)

albertofori:
> @albertofori Yes, if this PR will change the existing business behavior of users, it will cause breaking change. It is recommended to implement breaking change in the unified breaking change window of CLI _(Microsoft Build and Ignite)_, so as to reduce it's impact on users For more details about CLI breaking change process, please refer to this doc [breaking_changes.md](https://msazure.visualstudio.com/One/_git/AzureCliTools-Wiki?version=GBzhoxing-test/update-breaking-changes&path=/src/teams_docs/azcli_docs/breaking_changes.md&_a=preview)

Thanks a lot, @zhoxing-ms!

shenmuxiaosen:
What do we do when importing a json array as root?

albertofori:
> What do we do when importing a json array as root?

We error out and tell the user to import a json object.  
https://github.com/Azure/azure-cli/pull/23419/commits/b8915aa7dcb6bd7536eb695edd61b5bd1081d8fe#:~:text=if%20not%20isinstance,).__name__))

albertofori:
> @albertofori Yes, if this PR will change the existing business behavior of users, it will cause breaking change. It is recommended to implement breaking change in the unified breaking change window of CLI _(Microsoft Build and Ignite)_, so as to reduce it's impact on users For more details about CLI breaking change process, please refer to this doc [breaking_changes.md](https://msazure.visualstudio.com/One/_git/AzureCliTools-Wiki?version=GBzhoxing-test/update-breaking-changes&path=/src/teams_docs/azcli_docs/breaking_changes.md&_a=preview)

@zhoxing-ms So to confirm, please can this be merged in if there are no further issues or do we have to hold it until the breaking change window?

shenmuxiaosen (approved):
:shipit:

zhoxing-ms:
@albertofori If we merged this PR now, this PR will be added to the current sprint release by our CI pipeline.
So I suggest not to merge it first. Wait until the sprint of Ignite Event *(around October 12, 2022)*, and then please remind us to merge it. What do you think?

albertofori:
> @albertofori If we merged this PR now, this PR will be added to the current sprint release by our CI pipeline. So I suggest not to merge it first. Wait until the sprint of Ignite Event _(October)_, and then please remind us to merge it. What do you think?

@zhoxing-ms Thanks! I will make sure to remind the team to merge this in the next sprint then.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -137,6 +137,9 @@ def __read_with_appropriate_encoding(file_path, format_):
         with io.open(file_path, 'r', encoding=default_encoding) as config_file:
             if format_ == 'json':
                 config_data = json.load(config_file)
+                # Only accept json objects
+                if not isinstance(config_data, dict):
+                    raise ValueError("Json object required but type '{}' was given.".format(type(config_data).__name__))
```
avanigupta, line 142: Should we add this validation for yaml files too?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -879,23 +852,36 @@ def __export_keyvalues(fetched_items, format_, separator, prefix=None):
...
 
 
+def __try_convert_to_arrays(constructed_data, is_root=True):
+    if not (isinstance(constructed_data, dict) and len(constructed_data) > 0):
+        return constructed_data
+
+    # Object cannot be an array if not all keys are numeric
+    if False not in (key.isdigit() for key in constructed_data):
+        is_array = True
+        sorted_data_keys = sorted(int(key) for key in constructed_data)
```
avanigupta, line 869: > sorted(int(key) for key in constructed_data) [](http://example.com/codeflow?start=26&length=45)

Wouldnt constructed_data already be sorted?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -879,23 +852,36 @@ def __export_keyvalues(fetched_items, format_, separator, prefix=None):
...
+
+        # If all keys are digits and in order starting from 0, we convert the dictionary to an array
+        # with indices corresponding to the keys.
+        # We do not try to convert key-values at the root of the object to an array even if they meet this criterion.
+        for index, key in enumerate(sorted_data_keys):
+            if is_root or index != key:
```
avanigupta, line 875: >index != key [](http://example.com/codeflow?start=26&length=12)

We also support converting something like this to a valid array:
```
arr:0='abc'
arr:6='pqr'
arr:7='xyz'
```

This would still flatten to `arr = ["abc","pqr","xyz"]`

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
albertofori: Currently, YAML file data that is not formatted as an object is rejected already. An attempt to import a YAML file with array data throws the error shown below.
image

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
albertofori: Python dictionary keys are not sorted. They remain in the order in which they are added.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
albertofori: We ensure that the indices are in continuous order from 0 to n else we revert to the default export as key-values. This is because a value could be easily inserted into the array by a user if they add, e.g 
arr:4="efg" separately after importing these values.  This changes the indices of values for keys "arr:6" and "arr:7" in the exported array.  However, the indices are guaranteed to stay the same if the values are continuous and one could only append to the end of the array by adding a key continuous with the current order.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: Agreed, I was thinking more about how backend always returns data sorted by key name. So even though dictionaries are not sorted, the data in constructed_data will be sorted since its coming from our backend. But we shouldn't rely on backend sorting and its okay to sort it here just to be sure.

albertofori: Got it!  Thanks for pointing this out. I did not know that the backend returns data sorted by key name. This is not the case with the Portal so I guess I was working under the same assumption here to be safe. Thanks!

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: That's a valid scenario that we have always supported. If someone adds `arr:4="efg"` after importing key-values, the final result after exporting should be `arr = ["abc","efg","pqr","xyz"]`

Consider this scenario - someone imports an array in their config store and then they have to delete some values from the middle of the array. Do we expect them to manually reorder all their key-values just because they deleted one value in the middle? In this scenario they will end up with the example I described above, and we shouldn't convert their array to key-values just because the indices are not contiguous.

albertofori: That is a valid argument, however, int this same scenario, that means we will be changing key names assuming the exported file is imported to a different key-value store. Or there could be duplicates if we import back to the same key-value store for any reason since now a value that was initially `arr:4` will now be `arr:2`.  This will not be easily caught in a user's code if they are accessing a value in a given index. e.g

A user is accessing `arr:2` in `arr = ["abc","efg","pqr","xyz"]` from their code. When we delete this value, the value of key `arr:2` and all other indices after that are going to change, resulting in unexpected behaviour which will be difficult to trace. But if we revert to key-values, the code will just break since the configuration will not have been found.

I think exporting as key-values is the default mode which can be consistently exported/imported so we revert to that to preserve them. But we could discuss this further.

zhoxing-ms:
@albertofori Could you please address those conflicts~

zhoxing-ms:
Could you please add the information of the breaking change and the correct usage recommended in the PR description?

albertofori:
> Could you please add the information of the breaking change and the correct usage recommended in the PR description?
Information on correct usage and the breaking change has been added in point 1. in the PR description.

zhoxing-ms:
@albertofori Could you please resolve those conflicts?

zhoxing-ms (approved):

zhoxing-ms:
@avanigupta Could you help review and approve this PR tomorrow? 
Because the release for Ignite sprint will be launched the day after tomorrow and the breaking change of this PR has been published in the breaking change doc. In order to make this PR catch up with the current breaking change window, if you do not reply in time, we will merge this PR tomorrow evening.

zhoxing-ms:
> [App Config] BREAKING CHANGE: `az appconfig kv import`: Add validation to JSON file import to ensure that only valid JSON objects are imports
[App Config] BREAKING CHANGE: `az appconfig kv export`: Update the array conversion logic to prevent dropping keys during export

@albertofori I modified the PR description to comply with the specification of history notes. Please help to see if it is accurate and meets your expectations

avanigupta (approved):

albertofori:
> > [App Config] BREAKING CHANGE: `az appconfig kv import`: Add validation to JSON file import to ensure that only valid JSON objects are imports
> > [App Config] BREAKING CHANGE: `az appconfig kv export`: Update the array conversion logic to prevent dropping keys during export
> 
> @albertofori I modified the PR description to comply with the specification of history notes. Please help to see if it is accurate and meets your expectations

@zhoxing-ms Please it looks fine. Thanks! By the way, I added a few updates to the first point in the description based on updates that came later in this PR . I hope it still adheres to the required specifications.

PR #23442, opened 2022-08-04
{RDBMS} Add check_name_availability_without_location as fallback in MySQL flex

alanenriqueo:
**Description**
When we want to check if a server name is available for MySQL flexible server, we make a call to the endpoint `/subscriptions/{subscriptionId}/providers/Microsoft.DBforMySQL/locations/{locationName}/checkNameAvailability`. But if it fails with HTTP 403, we need to fallback to the endpoint without location `/subscriptions/{subscriptionId}/providers/Microsoft.DBforMySQL/checkNameAvailability` as the Portal does.

---


yonzhan:
RDBMS

DaeunYim (approved):

PR #23457, opened 2022-08-08
[Compute] Fix #23341: `az vm list-skus`: Fix filtering out VM sizes that are available regionally when they are restricted in all zones

### LINKED ISSUE #23341: az vm list-skus filtering out VM sizes that are available regionally when they are restricted in all zones
**Related command**
az vm list-skus

**Describe the bug**
The current logic for "vm list-skus" is filtering out VM sizes that are available regionally when they are restricted in all zones. Relevant code line is this https://github.com/Azure/azure-cli/blob/b4759064b7e38203ef9d67a581105206a17f513b/src/azure-cli/azure/cli/command_modules/vm/custom.py#L1285

This does not appear to be correct behavior, since a VM size maybe restricted for all 3 zones "zonally" but still be available for a subscription when deployed "regionally".

* When restrictionInfo.type=”Zone” it means it is a zonal restriction for the subscription: i.e., not available to deploy this VM size in the subscription in the zones specified in the restrictionInfo properties (and could be all 3 zones)
* When restrictionInfo.type=”Location” it means it is a regional restriction for the subscription: i.e., not available to deploy this VM size in the subscription in the region even when not specifying a zone

**To Reproduce**
Execute "az vm list-skus" on an Azure subscription that has some VM sizes that are restricted zonally in all 3 zones but are not restricted regionally. Review the output to see that the VMs sizes are not listed.

As expected, all VM sizes are returned when using the `--all` parameter, but the available sizes should also be returned without using `--all` if these sizes are not restricted regionally as well as zonally.

**Expected behavior**
VM sizes that are available for a subscription, either regionally or zonally, should be returned by the "az vm list-skus" command even without including the `--all` parameter.

yanzhudd:
**Description**

In `az vm list-skus` command, fix filtering out VM sizes that are available regionally when they are restricted in all zones
Close #23341

**Testing Guide**

**History Notes**

[Compute] Fix #23341: `az vm list-skus`: Fix filtering out VM sizes that are available regionally when they are restricted in all zones

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/vm/_vm_utils.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_vm_utils.py
@@ -123,6 +123,12 @@ def list_sku_info(cli_ctx, location=None):
 
 
 def is_sku_available(cmd, sku_info, zone):
+    """
+    The SKU is unavailable in the following cases:
+    1. regional restriction and the region is restricted
+    2. parameter --zone is input which indicates only showing skus with availability zones.
+       Meanwhile, zonal restriction and all zones are restricted
+    """
     is_available = True
     is_restrict_zone = False
     is_restrict_location = False
@@ -134,8 +140,6 @@ def is_sku_available(cmd, sku_info, zone):
             if cmd.supported_api_version(max_api='2017-03-30'):
                 is_available = False
                 break
-            # This SKU is only unavailable if all zones are restricted and showing skus supporting availability zones
-            # are enabled or regional restriction and the region is restricted
             if restriction.type == 'Zone' and not (
                     set(sku_info.location_info[0].zones or []) - set(restriction.restriction_info.zones or [])):
                 is_restrict_zone = True
--- a/src/azure-cli/azure/cli/command_modules/vm/_vm_utils.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_vm_utils.py
@@ -122,7 +122,7 @@ def list_sku_info(cli_ctx, location=None):
     return result
 
 
-def is_sku_available(cmd, sku_info):
+def is_sku_available(cmd, sku_info, zone):
     is_available = True
     is_restrict_zone = False
     is_restrict_location = False
@@ -143,7 +143,7 @@ def is_sku_available(cmd, sku_info):
                     sku_info.location_info[0].location in (restriction.restriction_info.locations or [])):
                 is_restrict_location = True
 
-            if is_restrict_location or is_restrict_zone:
+            if is_restrict_location or (is_restrict_zone and zone):
                 is_available = False
                 break
     return is_available
--- a/src/azure-cli/azure/cli/command_modules/vm/_vm_utils.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_vm_utils.py
@@ -122,6 +122,33 @@ def list_sku_info(cli_ctx, location=None):
     return result
 
 
+def is_sku_available(cmd, sku_info):
+    is_available = True
+    is_restrict_zone = False
+    is_restrict_location = False
+    if not sku_info.restrictions:
+        return is_available
+    for restriction in sku_info.restrictions:
+        if restriction.reason_code == 'NotAvailableForSubscription':
+            # The attribute location_info is not supported in versions 2017-03-30 and earlier
+            if cmd.supported_api_version(max_api='2017-03-30'):
+                is_available = False
+                break
+            # This SKU is not available only if zonal restriction and all zones are restricted
+            # and regional restriction and the region is restricted
+            if restriction.type == 'Zone' and not (
+                    set(sku_info.location_info[0].zones or []) - set(restriction.restriction_info.zones or [])):
+                is_restrict_zone = True
+            if restriction.type == 'Location' and (
+                    sku_info.location_info[0].location in (restriction.restriction_info.locations or [])):
+                is_restrict_location = True
+
+            if is_restrict_location and is_restrict_zone:
+                is_available = False
+                break
+    return is_available
+
+
 # pylint: disable=too-many-statements, too-many-branches
 def normalize_disk_info(image_data_disks=None,
                         data_disk_sizes_gb=None, attach_data_disks=None, storage_sku=None,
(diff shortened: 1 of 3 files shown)


zhoxing-ms:
Please add some tests and rerun some tests like #18939

src/azure-cli/azure/cli/command_modules/vm/tests/latest/test_vm_commands.py
```diff
@@ -1561,6 +1561,29 @@ def test_list_compute_skus_partially_unavailable(self):
...
+                if restriction['reasonCode'] != 'NotAvailableForSubscription':
+                    break
+                if restriction['type'] == 'Zone' \
+                        and not (set(result['locationInfo'][0]['zones'] or []) -
+                                 set(restriction['restrictionInfo']['zones'] or [])):
+                    is_restrict_zone = True
+                if restriction['type'] == 'Location' and (
+                        result['locationInfo'][0]['location'] in (restriction['restrictionInfo']['locations'] or [])):
+                    is_restrict_location = True
+                if is_restrict_zone and is_restrict_location:
```
arsenvlad, line 1583: I think it is possible for a VM sku to be restricted in region but not to have zonal restrictions listed. So I think the SKU should be considered restricted if it has a "Location" restriction OR all zones restricted.

yonzhan:
Compute

src/azure-cli/azure/cli/command_modules/vm/tests/latest/test_vm_commands.py
yanzhudd: > I think it is possible for a VM sku to be restricted in region but not to have zonal restrictions listed. So I think the SKU should be considered restricted if it has a "Location" restriction OR all zones restricted.

Thanks. We modified the code again according to your comment. Could you please help to confirm it again?

arsenvlad: I thought about this a bit more and think we may want to bring in Azure Compute team into conversation here to double-check.

If we are filtering out VM sizes from the list based on the restriction types, we may need to make the filtering logic also dependent on the --zone parameter which for this CLI command is described as "--zone: Show skus supporting availability zones.  Allowed values: false, true"

I put together the following table to see if it makes sense to you and to Azure Compute:

| Restriction Types | VM SKU Availability | is_available?  |
|---|---|---|
| Location=null, Zone=null | Yes, regional or any zone | TRUE  |
| Location=null, Zone=1,2 | Yes, regional or zone 3 | TRUE  |
| Location=null, Zone=1,2,3 | Yes, but only regional | if (--zone parameter) then FALSE else TRUE  |
| Location=eastus2, Zone=null | No, since region is restricted all zones are restricted | FALSE  |
| Location=eastus2, Zone=1,2 | No, since region is restricted all zones are restricted | FALSE  |
| Location=eastus2, Zone=1,2,3 | No, since region is restricted all zones are restricted | FALSE  |

yanzhudd: > I thought about this a bit more and think we may want to bring in Azure Compute team into conversation here to double-check.
> 
> If we are filtering out VM sizes from the list based on the restriction types, we may need to make the filtering logic also dependent on the --zone parameter which for this CLI command is described as "--zone: Show skus supporting availability zones. Allowed values: false, true"
> 
> I put together the following table to see if it makes sense to you and to Azure Compute:
> 
> Restriction Types	VM SKU Availability	is_available?
> Location=null, Zone=null	Yes, regional or any zone	TRUE
> Location=null, Zone=1,2	Yes, regional or zone 3	TRUE
> Location=null, Zone=1,2,3	Yes, but only regional	if (--zone parameter) then FALSE else TRUE
> Location=eastus2, Zone=null	No, since region is restricted all zones are restricted	FALSE
> Location=eastus2, Zone=1,2	No, since region is restricted all zones are restricted	FALSE
> Location=eastus2, Zone=1,2,3	No, since region is restricted all zones are restricted	FALSE

Hello @grizzlytheodore, could you please help to confirm this logic about filtering out VM sizes from the list based on the restriction types? Thanks.

Hello @chasewilson, if the relationship mentioned above is confirmed, we would modify the logic about filtering out VM sizes from the list based on the restriction types.

yanzhudd: Hello @grizzlytheodore, could you please help to confirm this logic?

> I thought about this a bit more and think we may want to bring in Azure Compute team into conversation here to double-check.
> 
> If we are filtering out VM sizes from the list based on the restriction types, we may need to make the filtering logic also dependent on the --zone parameter which for this CLI command is described as "--zone: Show skus supporting availability zones. Allowed values: false, true"
> 
> I put together the following table to see if it makes sense to you and to Azure Compute:
> 
> Restriction Types	VM SKU Availability	is_available?
> Location=null, Zone=null	Yes, regional or any zone	TRUE
> Location=null, Zone=1,2	Yes, regional or zone 3	TRUE
> Location=null, Zone=1,2,3	Yes, but only regional	if (--zone parameter) then FALSE else TRUE
> Location=eastus2, Zone=null	No, since region is restricted all zones are restricted	FALSE
> Location=eastus2, Zone=1,2	No, since region is restricted all zones are restricted	FALSE
> Location=eastus2, Zone=1,2,3	No, since region is restricted all zones are restricted	FALSE

@travismaier, could you please help to confirm this logic about filtering out VM sizes from the list based on the restriction types? Thanks.

src/azure-cli/azure/cli/command_modules/vm/_vm_utils.py
```diff
@@ -122,6 +122,34 @@ def _match_location(loc, locations):
...
+        return is_available
+    for restriction in sku_info.restrictions:
+        if restriction.reason_code == 'NotAvailableForSubscription':
+            # The attribute location_info is not supported in versions 2017-03-30 and earlier
+            if cmd.supported_api_version(max_api='2017-03-30'):
+                is_available = False
+                break
+            # This SKU is not available only if zonal restriction and all zones are restricted
+            # and skus supporting availability zones are required to show
+            # or regional restriction and the region is restricted
```
zhoxing-ms, line 139: Could you please refine this comment to make it more readable?

zhoxing-ms (approved):

wangzelin007 (approved):

PR #23469, opened 2022-08-08
msi

heaths:
- Refactor MSI components to improve performance
- Add TODO


yonzhan:
MSI improvement. @jiasli for awareness

build_scripts/windows/azure-cli.wixproj
```diff
@@ -8,6 +8,8 @@
     <WixTasksPath>wixtasks.dll</WixTasksPath>
     <AzureCliSource>artifacts\cli</AzureCliSource>
     <LinkerAdditionalOptions>-fv</LinkerAdditionalOptions>
+    <HarvestDirectorySuppressAllWarnings>5150;5151</HarvestDirectorySuppressAllWarnings>
```
jiasli, line 11: I saw from https://wixtoolset.org/documentation/manual/v3/msbuild/target_reference/harvestdirectory.html that `HarvestDirectorySuppressAllWarnings` should be a boolean:

image

I am wondering if I misunderstand anything.

jiasli:
Thanks a lot @heaths for the help. I downloaded the [built MSI artifact](https://artprodcus3.artifacts.visualstudio.com/A0fb41ef4-5012-48a9-bf39-4ee3de03ee35/29ec6040-b234-4e31-b139-33dc4287b756/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL2F6dXJlLXNkay9wcm9qZWN0SWQvMjllYzYwNDAtYjIzNC00ZTMxLWIxMzktMzNkYzQyODdiNzU2L2J1aWxkSWQvMTc2NjQ0Mi9hcnRpZmFjdE5hbWUvbXNp0/content?format=file&subPath=%2FMicrosoft%20Azure%20CLI.msi) and installed it from scratch on my Surface Book 2 with Windows 11, but it still takes almost 1 minute to install. Maybe I did something wrong.

build_scripts/windows/refactor.xslt
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0"?>
+<!-- Define WiX namespace twice: template matches and selections don't work with default namespace, but default handy for simply emitting elements. -->
+<xsl:stylesheet version="1.0" xmlns="http://schemas.microsoft.com/wix/2006/wi" xmlns:wix="http://schemas.microsoft.com/wix/2006/wi" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
```
jiasli, line 3: I didn't see anything specific to Azure CLI in `refactor.xslt`. Since it is beneficial to all MSI packages, is it a good idea to include it in the WiX Toolset itself?

build_scripts/windows/azure-cli.wixproj
heaths: No, my bad. I copied the wrong one (to avoid typos). Meant to copy the one with "...SpecificWarnings".

build_scripts/windows/refactor.xslt
heaths: Not sure what you mean. Generally, 1 file per component is recommended but not when you have thousands of files/components. That's not common with MSIs, and optimizations like this - though, perhaps more targeted, which isn't really possible when harvesting - are common for MSIs.

As to why it took almost as long, I'd need to see a verbose log. One area of concern was that because of the component refactoring, I had to move `RemoveExistingProducts` earlier which means all the files are removed first then new ones added. If files weren't updated, there's still a larger IO hit. Still, if you most often update most files, this avoids unnecessary time doing CRC checks to see if a file changed.

One idea I had was to conditionally schedule `RemoveExistingProducts` early for older MSI versions (before the refactoring) but later for newer versions (after the refactoring), but that would require a custom action to do so. The first upgrade to the refactored MSI might be longer (still, I'm surprised you didn't see a positive delta) but subsequent ones would be faster.

PR #23473, opened 2022-08-09
[RDBMS] `az postgres flexible-server create/update`: Add `SameZone` for HA in PostgreSQL flexible server

alanenriqueo:
**Related command**
- `az postgres flexible-server create -g testGroup -n testSrv --high-availability SameZone ...`
- `az postgres flexible-server update -g testGroup -n testSrv --high-availability SameZone ...`

**Description**
We need to add `SameZone` as a possible value for argument `--high-availability` in PostgreSQL flexible server, just as MySQL already does.

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/rdbms/flexible_server_custom_postgres.py
+++ b/src/azure-cli/azure/cli/command_modules/rdbms/flexible_server_custom_postgres.py
@@ -50,7 +50,11 @@ def flexible_server_create(cmd, client,
     # Generate missing parameters
     location, resource_group_name, server_name = generate_missing_parameters(cmd, location, resource_group_name,
                                                                              server_name, 'postgres')
+
     server_name = server_name.lower()
+    if high_availability and high_availability.lower() == 'enabled':
+        logger.warning('\'Enabled\' value for high availability parameter will be deprecated. Please use \'ZoneRedundant\' or \'SameZone\' instead.')
+        high_availability = 'ZoneRedundant'
 
     pg_arguments_validator(db_context,
                            server_name=server_name,
@@ -101,8 +105,6 @@ def flexible_server_create(cmd, client,
 
     sku = postgresql_flexibleservers.models.Sku(name=sku_name, tier=tier)
 
-    if high_availability.lower() == "enabled":
-        high_availability = "ZoneRedundant"
     high_availability = postgresql_flexibleservers.models.HighAvailability(mode=high_availability,
                                                                            standby_availability_zone=standby_availability_zone)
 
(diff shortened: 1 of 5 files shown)


yonzhan:
RDBMS

src/azure-cli/azure/cli/command_modules/rdbms/flexible_server_custom_postgres.py
```diff
@@ -298,8 +304,8 @@ def flexible_server_update_custom_func(cmd, client, instance,
     # High availability can't be updated with existing properties
     high_availability_param = postgresql_flexibleservers.models.HighAvailability()
     if high_availability:
-        if high_availability.lower() == "enabled":
```
DaeunYim, line 301: We use list-skus (locationCapabilities) output to dynamically get region information when running the validator. Please add filter to see whether 'SameZone' HAMode is supported in specific region. You might need to modify get_postgres_list_skus_info in flexibleserver_utils file and add the same logic to MySQL if it's missing there too. If there's any validations that needs to be updated, please do so too.

alanenriqueo: I'm already checking the value of the array `supportedHAMode` in location capabilities response, and I modified the validators logic in `_pg_high_availability_validator`.

src/azure-cli/azure/cli/command_modules/rdbms/flexible_server_custom_postgres.py
```diff
@@ -50,7 +50,11 @@ def flexible_server_create(cmd, client,
     # Generate missing parameters
     location, resource_group_name, server_name = generate_missing_parameters(cmd, location, resource_group_name,
                                                                              server_name, 'postgres')
+
     server_name = server_name.lower()
+    if high_availability and high_availability.lower() == 'enabled':
+        logger.warning('\'Enabled\' value for high availability parameter will be deprecated. Please use \'ZoneRedundant\' or \'SameZone\' instead.')
```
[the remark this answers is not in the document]
alanenriqueo: I can't find a way to deprecate only a single possible value for the argument. The closest I could do was this, but the customer will be confused because the whole command `--high-availability` appears as deprecated when it's not:
image

src/azure-cli/azure/cli/command_modules/rdbms/flexible_server_custom_postgres.py
alanenriqueo: I confirmed with both feature owners of MySQL and PostgreSQL and SameZone HAMode is alwayds supported.

src/azure-cli/azure/cli/command_modules/rdbms/flexible_server_custom_postgres.py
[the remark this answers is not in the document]
evelyn-ys: I think warning log is fine since it's not deprecation for the whole parameter but for some enum value of this parameter

DaeunYim (approved):

evelyn-ys (approved):

PR #22798, opened 2022-06-09
[Role] `az ad app/sp update`: Support generic update `--set` on root level

jiasli:
**Related command**
`az ad app update`
`az ad sp update`

**Description**
Close #22758, #22777, #22635

`az ad sp update` has no native arguments, but only generic update arguments:

```
> az ad sp update -h

Command
    az ad sp update : Update a service principal.

Arguments
    --id    [Required] : Service principal name, or object id.

Generic Update Arguments
    --add              : Add an object to a list of objects by specifying a path and key value
                         pairs.  Example: --add property.listProperty <key=value, string or JSON
                         string>.
    --force-string     : When using 'set' or 'add', preserve string literals instead of attempting
                         to convert to JSON.
    --remove           : Remove a property or an element from a list.  Example: --remove
                         property.list <indexToRemove> OR --remove propertyToRemove.
    --set              : Update an object by specifying a property path and value to set.  Example:
                         --set property1.property2=<value>.
```

https://docs.microsoft.com/en-us/cli/azure/microsoft-graph-migration#known-issues documented that generic update arguments are not supported by now. Therefore, `az ad sp update` is basically a no-op.

This PR is an initial attempt to makes generic update work with the new model-less framework by supporting generic update `--set` for `az ad app/sp update` on the root level:

```
ad app update --id {app_id} --set isDeviceOnlyAuthSupported=true
az ad sp update --id 233dd73b-72e3-424a-9367-7588d957267e --set 'tags=["mytag"]'
```

PR #22432, opened 2022-05-13
[Role] `az ad/role`: Azure AD Graph API to Microsoft Graph API migration

jiasli:
## Description

Resolve #12946

According to https://docs.microsoft.com/en-us/graph/migrate-azure-ad-graph-overview

> Azure Active Directory (Azure AD) Graph is deprecated but won't be retired on June 30, 2022 as [previously announced](https://techcommunity.microsoft.com/t5/azure-active-directory-identity/update-your-applications-to-use-microsoft-authentication-library/ba-p/1257363). Listening closely to your feedback about the challenges of migrating such a critical dependency, we're delaying the retirement date through at least the end of this year, 2022. We’ll provide a retirement update mid-calendar year including releasing more tools to help you to migrate your apps.

This PR migrates the underlying API of `az ad` and `az role` from AD Graph API to Microsoft Graph API.

## Testing Guide

```sh
az ad sp create-for-rbac
az ad app create --display-name mytestapp

# Other commands under
az ad ...
az role ...
```

PR #23490, opened 2022-08-10
[Role] `az ad app/sp update`: Add `--parameters` to replace generic update arguments

jiasli:
## Description
During Microsoft Graph migration (https://github.com/Azure/azure-cli/pull/22432), Azure CLI switched to a lightweight model-less `GraphClient` for Microsoft Graph interaction because:

- Autorest for Python doesn't support openapi 3.0
- Azure CLI code-gen doesn't support openapi 3.0
- No Python SDK

Even though this architecture enables Azure CLI to freely call Graph service, it is not compatible with generic update, because the schema of the object is unknown. The best we can do is to allow calling `--set` on root-level properties (https://github.com/Azure/azure-cli/pull/22798). For other unsupported operations `--add`, `--remove`, the current workaround is to use `az rest` which is not very friendly to those who are not familiar with the underlying REST API.

This PR proposes a more powerful way to update a Graph object - using `--parameters` to update Graph objects directly with a JSON, similar to `az rest` but without HTTP protocol details. We experimented with this approach in `az ad app federated-credential create/update` commands and customers' feedback is positive.

```
> az ad app federated-credential update -h

Command
    az ad app federated-credential update : Update application federated identity credential.

Arguments
    --federated-credential-id [Required] : ID or name of the federated identity credential.
    --id                      [Required] : Application's appId, identifierUri, or id (formerly known
                                           as objectId).
    --parameters              [Required] : Parameters for creating federated identity credential.
                                           Should be JSON file path or in-line JSON string. See
                                           examples below for details.
```

Unlike ARM which uses `PUT` for creating and updating, Graph API uses `POST` and `PATCH` for creating and updating, thus not idempotent, such as

- [Create application - POST](https://docs.microsoft.com/en-us/graph/api/application-post-applications)
- [Update application - PATCH](https://docs.microsoft.com/en-us/graph/api/application-update)

This enables us to update the application without getting it first.

For the original generic update argument:

- Using `--set` on root properties are preserved.
- Unsupported generic update argument `--remove` and `--add` are removed. Actually, they stopped working since MS Graph migration (2.37.0) (as documented in [Known issues](https://docs.microsoft.com/en-us/cli/azure/microsoft-graph-migration#known-issues)), and we received very few complaints, so we can assume its usage is not necessary.

## Testing Guide

```
az ad app update --id e042ec79-34cd-498f-9d9f-123456781234 --set groupMembershipClaims=All
az ad app update --id e042ec79-34cd-498f-9d9f-123456781234 --parameters '{"groupMembershipClaims": "All"}'

az ad sp update --id 00000000-0000-0000-0000-000000000000 --set appRoleAssignmentRequired=true
az ad sp update --id 00000000-0000-0000-0000-000000000000 --parameters '{"appRoleAssignmentRequired": true}'
```

## TODO

- `--set` and `--parameters` are only supported on `az ad app/sp update`. Other Graph object type `user`, `group` don't support generic update from the beginning.
- We may also extent `--set` and `--parameters` to `create` commands.


yonzhan:
AD

alexdrenea:
@jiasli this looks great!
Just a couple of comments or potential improvements.
1. can you use `--parameters` to set "deep" properties? i.e. api.acceptMappedClaims? or web.logoutUrl? If yes, maybe add it to the Testing Guide as an example
2. can we enable the `--parameters` flag on create? so that you can create your application in one `az ad app` call and not have to do `create` and `update` as two operations (if MS Graph supports it)

jiasli:
> can you use `--parameters` to set "deep" properties? i.e. api.acceptMappedClaims? or web.logoutUrl? If yes, maybe add it to the Testing Guide as an example

I don't think it will be possible to implement setting "deep" properties, because Azure CLI doesn't keep a model of the app object, so it can't build the JSON only from the path and value that are given.

> can we enable the `--parameters` flag on create? so that you can create your application in one `az ad app` call and not have to do `create` and `update` as two operations (if MS Graph supports it)

Yes, this will be doable. I will revisit this PR when it gets planned.

Issue #20972, opened 2022-01-13
Issue to set password for secret at KeyVauklt by azure CLI


chamir2:
[Enter feedback here]
Hi,
 
Yesterday while trying to change a secret value that contains special characters I encountered an error saying that the command is not recognize.
The RC was the secret had pipeline or “And” sign (| or &) at the secret value and due to it, it tried to run the secret value as a command.
 
To resolve this issue I use `" signs to wrap the variable 

here by command example:

az keyvault secret set --name $secretName --vault-name $KeyVaultName --tags $TagsConverted --value `"$newConnectionString`" --expires $vaultexpires --not-before $vaultstart | Out-Null

First of all I would be happy if it can be fixed to reduce issues for other people.
As well as I understand it, it can be a vulnerability - Please verify.

Thanks,
Amir Chen

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: eeedb9a7-2aed-6727-4cfb-1515809d676e
* Version Independent ID: 2c83180d-22ad-db42-e2eb-f27f11977963
* Content: [az](https://docs.microsoft.com/en-us/cli/azure/reference-index?view=azure-cli-latest&source=docs#feedback)
* Content Source: [latest/docs-ref-autogen/reference-index.yml](https://github.com/MicrosoftDocs/azure-docs-cli/blob/main/latest/docs-ref-autogen/reference-index.yml)
* GitHub Login: @rloutlaw
* Microsoft Alias: **routlaw**

chamir2:
For further information you can contact me by internal teams.
Thanks,
Amir Chen

yonzhan:
keyvault

chamir2:
yes

evelyn-ys:
For you reference: https://github.com/Azure/azure-cli/blob/dev/doc/quoting-issues-with-powershell.md#ampersand--is-interpreted-by-command-prompt

chamir2:
Thanks for your replay
So I understand it’s a known issue.
Would it be fixed?
Any ETA?
Don’t you worry it can be used as a way to remote code execution?

seh-msft:
+1, ran into this today. 

Has to be mishandling of the input value when passing on a password with a pipe `|` character. 

Ex.

```
az keyvault secret set --vault-name SomeVault --name foobar --value abc123|de
```

Yielding the error:

```
'de' is not recognized as an internal or external command,
operable program or batch file.
```

[...]

Issue #15529, opened 2020-10-15
[Shell] Azure CLI receives corrupted arguments in PowerShell


jiasli:
## Describe the bug

When invoking `az` in PowerShell, due to the known issue of PowerShell https://github.com/PowerShell/PowerShell/issues/1995,  arguments passed to Azure CLI may get corrupted. For example, literal double quotes (`"`) are lost:

```pwsh
# PowerShell
> az '{"signInAudience":"AzureADAndMicrosoftAccounts"}' --debug
Command arguments: ['{signInAudience:AzureADAndMicrosoftAccounts}', '--debug']
```

This contradicts the behavior of Bash:

```sh
# Bash
$ az '{"signInAudience":"AzureADAndMicrosoftAccounts"}' --debug
Command arguments: ['{"signInAudience":"AzureADAndMicrosoftAccounts"}', '--debug']
```

## Workaround

Until this issue is fixed by PowerShell, arguments with special characters (`"`, `,`, `&`...) need to be escaped twice, one for CMD, one for PowerShell.

More info: [Quoting issues with PowerShell](https://github.com/Azure/azure-cli/blob/dev/doc/quoting-issues-with-powershell.md)

jiasli:
All similar issues are categorized under the [**Shell - PowerShell**](https://github.com/Azure/azure-cli/issues?q=label%3A%22Shell+-+PowerShell%22) label. By now, received #7054, #8070, #8630, #8827, #9047, #9228, #9742, #10370, #10997, #11003, #1100, #11641, #11668, #12970, #13152, #13208, #13340, #13882, #15512, #15529.

bb-froggy:
The PowerShell issue seems to be fixed in PowerShell 7.3 : https://github.com/PowerShell/PowerShell/issues/1995#issuecomment-1211457683

jiasli:
To dive deeper, the root cause is that **CMD executes the value of `%*`**.

Say `test.cmd` has

```cmd
echo %*
```

And we have `test.py`:

```py
import subprocess
subprocess.run(['test.cmd', 'a&b'])
```

Running `python test.py` can lead to code execution/injection:

```
> python test.py

D:\temp>echo a
a
'b' is not recognized as an internal or external command,
operable program or batch file.
```

This is unimaginable on Linux or with any other programming language where you have to run `eval` or similar functions to execute an argument's value.

[...]

PR #23514, opened 2022-08-11
[Core] Add `az.ps1` entry script for PowerShell

jiasli:
## Context

Fix https://github.com/Azure/azure-cli/issues/20972

If an argument contains special characters like `|` and `&`, even though the character is double quoted, it still gets executed. This is because PowerShell launches `cmd.exe` to execute `az.cmd` script. Double quotes are parsed by PowerShell so not passed to `cmd.exe` during this process. `cmd.exe` sees the raw `|` and `&` and executes them.

However, when `az.cmd` is invoked from `cmd.exe` (Command Prompt), it is launched in-process, and double quotes are preserved, leading to inconsistent behavior between `cmd.exe` and PowerShell.

## Change

This PR adds `az.ps1` entry script for PowerShell, so that PowerShell can natively execute `az.ps1` and no longer launches `cmd.exe` to execute `az.cmd`.

Since PowerShell 7.3, double quotes are now preserved correctly when calling a native executable (https://github.com/PowerShell/PowerShell/issues/1995#issuecomment-1211457683), thus fixing https://github.com/Azure/azure-cli/issues/15529.

## Testing Guide

```
az --debug '"ab"'
cli.knack.cli: Command arguments: ['--debug', '"ab"']

az --debug "a|b"
cli.knack.cli: Command arguments: ['--debug', 'a|b']

az --debug "a&b"
cli.knack.cli: Command arguments: ['--debug', 'a&b']
```


build_scripts/windows/scripts/az.ps1
```diff
@@ -0,0 +1,2 @@
+$env:AZ_INSTALLER="MSI"
+& "$PSScriptRoot\..\python.exe" -IBm azure.cli $args
```
SteveL-MSFT, line 2: This doesn't work correctly with WinPS 5.1, right?  Unfortunately the case of `az '"test"'` by the time the `$args` is processed, you would only get `test` on any PS older than 7.3, and there's no way to know what the original intent.  So perhaps the best thing is to iterate through `$args` and if `|` or `&` is found, throw an error that it would only work with 7.3+?

yonzhan:
add a ps1 file as the entry script to bypass 'PowerShell -> cmd' issue.

build_scripts/windows/scripts/az.ps1
jiasli: Tested in Windows PowerShell 5.1, and indeed double quotes are lost, but luckily `|` or `&` isn't causing any problems to us. :)

```
> az2 --debug "a|b" "a&b" '"ab"'
cli.knack.cli: Command arguments: ['--debug', 'a|b', 'a&b', 'ab']
```

jiasli:
By the way, I am playing around by calling `cmd.exe` (https://ss64.com/nt/cmd.html) myself from Command Prompt and it looks like it has some weird behavior that I need to add extra double quotes (denoted by `^`):

According to https://ss64.com/nt/syntax-esc.html

> To launch a batch script with spaces in the script Path and other parameters, all requiring quotes:
> ```
> CMD /k ""c:\batch files\test.cmd" "Parameter 1 with space" "Parameter2 with space""
>        ^                                                                          ^
> ```

My testing:

`D:\test.cmd` contains 
```
@echo %*
```
```
D:\>cmd /c "D:\test.cmd" "parameter1 parameter2"
'D:\test.cmd" "parameter1' is not recognized as an internal or external command,
operable program or batch file.

D:\>cmd /c ""D:\test.cmd" "parameter1 parameter2""
"parameter1 parameter2"
```

While in Bash

`/home/user2/test.sh` contains

```
echo "$@"
```

```
user2@DESKTOP-A79F1:~$ bash "/home/user2/test.sh" "parameter1 parameter2"
parameter1 parameter2
user2@DESKTOP-A79F1:~$ bash ""/home/user2/test.sh" "parameter1 parameter2""
bash: /home/user2/test.sh parameter1: No such file or directory
```

No weird outer double quotes are needed.

evelyn-ys (approved):

bebound (approved):

jiasli:
I also found another weird PowerShell behavior while writing this PR - PowerShell sometimes even depends on the space location to determine the behavior:

```
> python -c "import sys; print(sys.argv)" '\"a b\"c'
['-c', '"a', 'b"c']
> python -c "import sys; print(sys.argv)" '\"ab\" c'
['-c', '"ab" c']
```

Also consider these commands:

```
> python -c "import sys; print(sys.argv)" '""a b""'
['-c', '"a b"']
> python -c "import sys; print(sys.argv)" '\"a b\"'
['-c', '"a', 'b"']
```

They should be grammatically equivalent, as in Win32 API within double quotes, `""` and `\"` both mean literal double quote sign. However, they are not, because in the second example, PowerShell somehow recognizes double quotes and thinks space is not “on the root level of the string”, so doesn’t surround the argument with double quotes when passing it to `python.exe`, making the argument split.

```
> python -c "import sys; print(sys.argv)" '\"a b\" '
['-c', '"a b" ']
```

In the above example, I put an extra space “on the root level” and PowerShell passed the arguments double-quoted-ly to `python.exe`.

build_scripts/windows/scripts/az.ps1
```diff
@@ -0,0 +1,217 @@
...
+
+# SIG # Begin signature block
+# MIInuQYJKoZIhvcNAQcCoIInqjCCJ6YCAQExDzANBglghkgBZQMEAgEFADB5Bgor
+# BgEEAYI3AgEEoGswaTA0BgorBgEEAYI3AgEeMCYCAwEAAAQQH8w7YFlLCE63JNLG
```
jiasli, line 6: There is one problem with `az.ps1` for `pip`. 

- ESRP code-sign was executed on Azure DevOps's Windows agent. When `git` checks out the source code on Windows, the end of line sequence is `CRLF` (`\r\n`). This is the default of Windows `git`:
   image
   so the signature was generated for the `CRLF` version of `az.ps1`. 
- However, wheels are built on Linux agent, converting `az.ps1`'s end of line sequence to `LF` (`\n`). 
- Even when installing with `pip -e`, `pip` converts `CRLF` to `LF` (for simplicity I guess).

Thus, the signature doesn't match the file content anymore.

src/azure-cli/az.ps1
```diff
@@ -0,0 +1,225 @@
...
+    & "$PSScriptRoot\python.exe" -m azure.cli $args
+}
+else {
+    # Run system python.exe
+    python.exe -m azure.cli $args
+}
+
+# SIG # Begin signature block
+# MIInqgYJKoZIhvcNAQcCoIInmzCCJ5cCAQExDzANBglghkgBZQMEAgEFADB5Bgor
+# BgEEAYI3AgEEoGswaTA0BgorBgEEAYI3AgEeMCYCAwEAAAQQH8w7YFlLCE63JNLG
```
jiasli, line 14: There is one problem with `az.ps1` for `pip`. 

- ESRP code-sign was executed on Azure DevOps's Windows agent. When `git` checks out the source code on Windows, the end of line sequence is `CRLF` (`\r\n`). This is the default of Windows `git`:
   image
   so the signature was generated for the `CRLF` version of `az.ps1`. 
- However, wheels are built on Linux agent, converting `az.ps1`'s end of line sequence to `LF` (`\n`). 
- Even when installing with `pip -e`, `pip` converts `CRLF` to `LF` (for simplicity I guess).

Thus, the signature doesn't match the file content anymore.

a3code:
Hello everyone, could please explain how I should run scripts right now
I am a little bit confused, on azure DevOps I have a script that had run az in the next way:
` Start-Process "az" -ArgumentList $azArgs -NoNewWindow -Wait`
azArgs is a commend to set some settings that is.
and currently I see an error : 
_Start-Process : This command cannot be run due to the error: %1 is not a valid Win32 application_
how do I need to change it to make it work after this fix?
Should I just remove Start-Process ?

yonzhan:
We will revert this script and roll out on 10.12. Everything will work normally by then.

ishepherd:
@jiasli @yonzhan Will you be adding test coverage for all the many issues people are reporting? Looks like a nice opportunity to fill coverage gaps.

PR #23530, opened 2022-08-15
[Core] Allow adding dedicated key for collecting cli telemetry

evelyn-ys:
**Description**

This PR allows command modules & cli extensions to add their own appInsights key to collect all existing cli events.
The cli events sent to CLI's AppInsights are stored in local cache file.
While those sent to dedicated AppInsights will be flushed immediately

**Testing Guide**

We can define a decorator:
```
def add_telemetry_dedicated_key(key=''):
    def _decorator(func):
        @wraps(func)
        def _wrapped_func(*args, **kwargs):
            if key:
                from azure.cli.core import telemetry
                telemetry.add_dedicated_instrumentation_key(key)
            return func(*args, **kwargs)
        return _wrapped_func
    return _decorator
```
And then add the decorator for any command's custom function:
```
@add_telemetry_dedicated_key(key='02b91c82-6729-4241-befc-e6d02ca4fbba')
def create_xxx_func(cmd, client, resource_name,...):
```
Then when we run the specific command, cli events will be sent to dedicated AppInsights along with CLI's AppInsights

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/telemetry.py
+++ b/src/azure-cli-core/azure/cli/core/telemetry.py
@@ -417,10 +417,15 @@ def set_raw_command_name(command):
 
 
 @decorators.suppress_all_exceptions()
-def add_dedicated_instrumentation_key(dedicated_instrumentation_key: set):
+def add_dedicated_instrumentation_key(dedicated_instrumentation_key):
     if not dedicated_instrumentation_key:
         return
-    _session.instrumentation_key.update(dedicated_instrumentation_key)
+
+    from collections import Iterable
+    if isinstance(dedicated_instrumentation_key, str):
+        _session.instrumentation_key.add(dedicated_instrumentation_key)
+    elif isinstance(dedicated_instrumentation_key, Iterable):
+        _session.instrumentation_key.update(dedicated_instrumentation_key)
 
 
 @decorators.suppress_all_exceptions()
--- a/src/azure-cli-core/azure/cli/core/telemetry.py
+++ b/src/azure-cli-core/azure/cli/core/telemetry.py
@@ -421,7 +421,7 @@ def add_dedicated_instrumentation_key(dedicated_instrumentation_key):
     if not dedicated_instrumentation_key:
         return
 
-    from collections import Iterable
+    from collections.abc import Iterable
     if isinstance(dedicated_instrumentation_key, str):
         _session.instrumentation_key.add(dedicated_instrumentation_key)
     elif isinstance(dedicated_instrumentation_key, Iterable):
--- a/src/azure-cli-core/azure/cli/core/telemetry.py
+++ b/src/azure-cli-core/azure/cli/core/telemetry.py
@@ -16,11 +16,11 @@ from collections import defaultdict
 from functools import wraps
 from knack.util import CLIError
 from azure.cli.core import decorators
+from azure.cli.telemetry import DEFAULT_INSTRUMENTATION_KEY
 
 PRODUCT_NAME = 'azurecli'
 TELEMETRY_VERSION = '0.0.1.4'
 AZURE_CLI_PREFIX = 'Context.Default.AzureCLI.'
-DEFAULT_INSTRUMENTATION_KEY = 'c4395b75-49cc-422c-bc95-c7d51aef5d46'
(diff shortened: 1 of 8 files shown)


yonzhan:
This is a good solution to allow adding dedicated key for collecting cli telemetry

asaidabdalla:
I needed to return func(*args,**kwargs) in the decorator func in order for the decorated function to get called. 

```
def add_telemetry_dedicated_key(key=''):
    from functools import wraps
    def _decorator(func):
        @wraps(func)
        def _wrapped_func(*args, **kwargs):
            if key:
                from azure.cli.core import telemetry
                telemetry.add_dedicated_instrumentation_key(key)
            return func(*args, **kwargs)
        return _wrapped_func
    return _decorator
```

evelyn-ys:
> I needed to return func(*args,**kwargs) in the decorator func in order for the decorated function to get called.

Thanks for the correction! I missed that and will update the PR description :)

src/azure-cli-core/azure/cli/core/telemetry.py
```diff
@@ -414,6 +416,18 @@ def set_raw_command_name(command):
     _session.raw_command = command
 
 
+@decorators.suppress_all_exceptions()
+def add_dedicated_instrumentation_key(dedicated_instrumentation_key):
+    if not dedicated_instrumentation_key:
+        return
+
+    from collections import Iterable
```
asaidabdalla, line 424: I was using python 3.10 and this was silently erroring out because collections.Iterable was deprecated. I had to do something like this: 

```
    try:
        from collections.abc import Iterable
    except ImportError:
        from collections import Iterable
```

src/azure-cli-core/azure/cli/core/telemetry.py
```diff
@@ -414,6 +416,18 @@ def set_raw_command_name(command):
...
+@decorators.suppress_all_exceptions()
+def add_dedicated_instrumentation_key(dedicated_instrumentation_key):
+    if not dedicated_instrumentation_key:
+        return
+
+    from collections.abc import Iterable
+    if isinstance(dedicated_instrumentation_key, str):
+        _session.instrumentation_key.add(dedicated_instrumentation_key)
+    elif isinstance(dedicated_instrumentation_key, Iterable):
+        _session.instrumentation_key.update(dedicated_instrumentation_key)
```
zhoxing-ms, line 428: Just out of curiosity, may I ask can we support adding multiple `dedicated_instrumentation_key`?

src/azure-cli-telemetry/azure/cli/telemetry/__init__.py
```diff
@@ -46,10 +48,34 @@ def _start(config_dir):
...
+        logger.info('Begin splitting cli events and extra events, total events: %s', len(events))
+        cli_events = {}
+        client = CliTelemetryClient()
+        for key, event in events.items():
+            if key == DEFAULT_INSTRUMENTATION_KEY:
+                cli_events[key] = event
+            else:
+                extra_event = {key: event}
+                client.add(json.dumps(extra_event), flush=True)
+        client.flush(force=True)
```
zhoxing-ms, line 71: Do we also need to set `client.flush(force=True)` when using the `DEFAULT_INSTRUMENTATION_KEY`?

src/azure-cli-core/azure/cli/core/telemetry.py
evelyn-ys: Just in case they want to send to several appInsights

zhoxing-ms: Got it, thanks~

src/azure-cli-telemetry/azure/cli/telemetry/__init__.py
evelyn-ys: Yes we already have in https://github.com/Azure/azure-cli/blob/48a69dbebb15844afa9e240051a1c4d13ffa6bbb/src/azure-cli-telemetry/azure/cli/telemetry/__init__.py#L83-L86

zhoxing-ms: OK

zhoxing-ms (approved):

calvinhzy (approved):

PR #23602, opened 2022-08-19
[App Service] `az functionapp list-runtimes`: Add linuxFxVersion to output

StrawnSC:
**Related command**

**Description**

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -3213,9 +3213,12 @@ class _FunctionAppStackRuntimeHelper(_AbstractStackRuntimeHelper):
 
         # used for displaying stacks
         def to_dict(self):
-            return {"runtime": self.name,
-                    "version": self.version,
-                    "supported_functions_versions": self.supported_func_versions}
+            d = {"runtime": self.name,
+                 "version": self.version,
+                 "supported_functions_versions": self.supported_func_versions}
+            if self.linux:
+                d["linux_fx_version"] = self.site_config_dict.linux_fx_version
+            return d
 
     def __init__(self, cmd, linux=False, windows=False):
         self.disallowed_functions_versions = {"~1", "~2"}


yonzhan:
App Service

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -3213,9 +3213,12 @@ def __init__(self, name=None, version=None, is_preview=False, supported_func_ver
...
         # used for displaying stacks
         def to_dict(self):
-            return {"runtime": self.name,
-                    "version": self.version,
-                    "supported_functions_versions": self.supported_func_versions}
+            d = {"runtime": self.name,
+                 "version": self.version,
+                 "supported_functions_versions": self.supported_func_versions}
+            if self.linux:
+                d["linux_fx_version"] = self.site_config_dict.linux_fx_version
```
panchagnula, line 3220: @StrawnSC this will display the linuxFxVersion with | which is probably what is supported by the runtime, but in webapps list-runtimes we convert this to : delimiter to help with | being special behavior with PS. I think this is OK (esp since functionapp create takes lang & runtime as 2 separate inputs)but could you share the new output with the function apps PM as well make sure this is what they expect as well. Thanks!

StrawnSC: Sounds good. Are there PMs in particular I should share this with besides Thiago?

panchagnula: Since he is the one requesting this change - I would start with him. Thanks!

zhoxing-ms: @panchagnula @StrawnSC Any update? May I ask if this PR needs to be released in this sprint?

panchagnula: @zhoxing-ms thanks for checking. this is good to go on our end. got confirmation from PM on our end as well. This sprint release please.

panchagnula (commented):
Can you check we don't need to re-record any tests here. would hate to see them fail in nightly build. Otherwise looks good to me. Thanks!

StrawnSC:
> Can you check we don't need to re-record any tests here. would hate to see them fail in nightly build. Otherwise looks good to me. Thanks!

@panchagnula, we should be good here! I just reran the test recordings and everything passes

zhoxing-ms (approved):

zhoxing-ms:
@StrawnSC If there are no other problems with this PR, could you approve it?

panchagnula (approved):
Approved

PR #23607, opened 2022-08-20
[Backup] `az backup vault create/backup-properties set`: Add support for Alert Settings

akshayneema:
**Related command**

**Description**

`az backup vault create/az backup vault backup-properties set`: Add `--classic-alerts` and `--azure-monitor-alerts-for-job-failures` as optional parameters to update alerts setting of the Recovery Services Vault

`az backup protection backup-now`: Add better warning messages and help texts.

**Testing Guide**

**History Notes**
`az backup vault create/az backup vault backup-properties set`: Add `--classic-alerts` and `--azure-monitor-alerts-for-job-failures` as optional parameters to update alerts setting of the Recovery Services Vault
`az backup protection backup-now`: Add better warning messages and help texts

---


yonzhan:
Backup

src/azure-cli/azure/cli/command_modules/backup/_params.py
```diff
@@ -97,13 +97,17 @@ def load_arguments(self, _):
 
     with self.argument_context('backup vault create') as c:
         c.argument('tags', arg_type=tags_type)
+        c.argument('classic_alerts', arg_type=get_enum_type(['Enable', 'Disable']), help='Use this property to specify whether backup alerts from the classic solution should be received.')
+        c.argument('azure_monitor_alerts_for_job_failures', arg_type=get_enum_type(['Enable', 'Disable']), help='Use this property to specify whether built-in Azure Monitor alerts should be received for every job failure.')
```
zhoxing-ms, line 101: This parameter name is too long, which will make it inconvenient for the user to input. Could you change the parameter name to be shorter? Such as `azure_monitor_alerts_for_job_failures` --> `alerts_for_job_failures` ?

src/azure-cli/azure/cli/command_modules/backup/custom_wl.py
```diff
@@ -463,12 +463,16 @@ def backup_now(cmd, client, resource_group_name, vault_name, item, retain_until,
     message = "For SAPHANA and SQL workload, retain-until parameter value will be overridden by the underlying policy"
 
     if retain_until is None:
-        retain_until = datetime.now(timezone.utc) + timedelta(days=30)
-        if backup_type == 'Full':
+        if backup_type.lower() == 'copyonlyfull':
```
zhoxing-ms, line 466: Could you ensure that the value of `backup_type` is not `None`, otherwise exception may be caused here

zhoxing-ms:
Please write the history notes under the `History Notes` part, I have updated it for you

src/azure-cli/azure/cli/command_modules/backup/custom_wl.py
akshayneema: We have a `if backup_type is None:` check above in this method already

zhoxing-ms: Got it, thanks~

src/azure-cli/azure/cli/command_modules/backup/_params.py
```diff
@@ -97,13 +97,17 @@ def load_arguments(self, _):
 
     with self.argument_context('backup vault create') as c:
         c.argument('tags', arg_type=tags_type)
+        c.argument('classic_alerts', arg_type=get_enum_type(['Enable', 'Disable']), help='Use this property to specify whether backup alerts from the classic solution should be received.')
+        c.argument('azure_monitor_alerts_for_job_failures', options_list=['--alerts-for-job-failures'], arg_type=get_enum_type(['Enable', 'Disable']), help='Use this property to specify whether built-in Azure Monitor alerts should be received for every job failure.')
```
zhoxing-ms, line 101:
```suggestion
        c.argument('azure_monitor_alerts_for_job_failures', options_list=['--job-failure-alerts'], arg_type=get_enum_type(['Enable', 'Disable']), help='Use this property to specify whether built-in Azure Monitor alerts should be received for every job failure.')
```

zhoxing-ms, line 101: Could we define those parameter as `arg_type=get_three_state_flag()`? In this way, you can convert them into `Enabled` or `Disabled` according to `True` or `False`

src/azure-cli/azure/cli/command_modules/backup/custom.py
```diff
@@ -111,9 +112,14 @@
...
 
-def create_vault(client, vault_name, resource_group_name, location, tags=None):
+def create_vault(client, vault_name, resource_group_name, location, tags=None, classic_alerts='Enable',
+                 azure_monitor_alerts_for_job_failures='Enable'):
     vault_sku = Sku(name=SkuName.standard)
-    vault_properties = VaultProperties()
+    vault_properties = VaultProperties(
+        monitoring_settings=MonitoringSettings(
+            azure_monitor_alert_settings=AzureMonitorAlertSettings(
+                alerts_for_all_job_failures=azure_monitor_alerts_for_job_failures + 'd'),
```
zhoxing-ms, line 121: Same as https://github.com/Azure/azure-cli/pull/23607/files?file-filters%5B%5D=.py&file-filters%5B%5D=.txt&show-viewed-files=true#r951265946

src/azure-cli/azure/cli/command_modules/backup/_params.py
zhoxing-ms: I suggest using `--job-failure-alerts` as the parameter name, because it is more concise than `--alerts-for-job-failures`, and the naming style can be consistent with `--classic-alerts`

linter_exclusions.yml
```diff
@@ -911,6 +916,9 @@ backup vault backup-properties set:
     hybrid_backup_security_features:
       rule_exclusions:
       - option_length_too_long
+    azure_monitor_alerts_for_job_failures:
+      rule_exclusions:
+      - option_length_too_long
```
zhoxing-ms, line 921: Since we have added a short option alias, I believe we no longer need to modify this file

src/azure-cli/azure/cli/command_modules/backup/_params.py
akshayneema: Hey! Enable/Disable looks like a better choice for us as the same `az backup vault backup-properties set` command has other parameters like `--soft-delete-feature-state` and `--hybrid-backup-security-features` which take Enable/Disable as inputs and we wanted to keep the experience consistent. 

Also, it feels more natural to Enable/Disable alerts rather than setting them to True/False.

zhoxing-ms: OK, make sense

zhoxing-ms (approved):
LGTM. Could you please resolve this conflict?

Issue #23562, opened 2022-08-17
Documentation Incorrect


Red1971:
Doc show workspace as optional..

az synapse managed-private-endpoints delete --pe-name
                                            [--ids]
                                            [--workspace-name]
                                            [--yes]

![image](https://user-images.githubusercontent.com/100938039/185209829-ebaf473b-9b16-4019-b324-77629a0468b8.png)

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: f554620d-874d-97c5-fe88-2f8e6f529001
* Version Independent ID: ccc641f5-c2ac-7937-553d-f5190ad9d49c
* Content: [az synapse managed-private-endpoints](https://docs.microsoft.com/en-us/cli/azure/synapse/managed-private-endpoints?view=azure-cli-latest#az-synapse-managed-private-endpoints-create)
* Content Source: [latest/docs-ref-autogen/synapse/managed-private-endpoints.yml](https://github.com/MicrosoftDocs/azure-docs-cli/blob/main/latest/docs-ref-autogen/synapse/managed-private-endpoints.yml)
* GitHub Login: @rloutlaw
* Microsoft Alias: **routlaw**

yonzhan:
route to CXP team

navba-MSFT:
@Red1971 Thanks for reaching out to us and reporting this issue. We are looking into this issue and we will provide an update.

navba-MSFT:
The above PR has been filed and it is currently under review.

navba-MSFT:
@Red1971 Thanks for reporting this issue. The above PR has been created to track this. Pasting the reply from the CLI Repo team as-is:

>We'll discuss if we should fix this issue inside CLI team.
Even if we need fix it, we should fix it in cli core, but not in specific modules as it's a shared design for all modules (i.e. if user provides --ids, like /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/synapse-cli000001/providers/Microsoft.Synapse/workspaces/clitest000003, then they don't need provide --workspace-name as --ids contains everything including --workspace-name.

You can track the progress in this PR https://github.com/Azure/azure-cli/pull/23620.
There is no ETA to fix this for now. We will update this thread once this issue is addressed.

PR #23620, opened 2022-08-22
{AzureSynapse} fixes Azure/azure-cli#23562

navba-MSFT:
fixes Azure/azure-cli#23562

The below command mentions about having --workspace-name as the optional parameter. But is is a required parameter.

az synapse managed-private-endpoints delete --pe-name
                                            [--ids]
                                            [--workspace-name]
                                            [--yes]

Error:
(--workspace-name | --ids) are required

More Info: https://docs.microsoft.com/en-us/cli/azure/synapse/managed-private-endpoints?view=azure-cli-latest#az-synapse-managed-private-endpoints-create

**Related command**

**Description**

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/synapse/manual/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/synapse/manual/_params.py
@@ -947,7 +947,7 @@ def load_arguments(self, _):
     for scope in ['show', 'create', 'delete']:
         with self.argument_context('synapse managed-private-endpoints ' + scope) as c:
             c.argument('workspace_name', arg_type=workspace_name_arg_type, id_part='name')
-            c.argument('managed_private_endpoint_name', options_list=['--pe-name'], help='The managed private endpoint name.')
+            c.argument('managed_private_endpoint_name', arg_type=workspace_name_arg_type, options_list=['--pe-name'], help='The managed private endpoint name.')
 
     with self.argument_context('synapse managed-private-endpoints list') as c:
         c.argument('workspace_name', arg_type=workspace_name_arg_type)


yonzhan:
AzureSynapse

src/azure-cli/azure/cli/command_modules/synapse/manual/_params.py
```diff
@@ -947,7 +947,7 @@ def load_arguments(self, _):
     for scope in ['show', 'create', 'delete']:
         with self.argument_context('synapse managed-private-endpoints ' + scope) as c:
             c.argument('workspace_name', arg_type=workspace_name_arg_type, id_part='name')
-            c.argument('managed_private_endpoint_name', options_list=['--pe-name'], help='The managed private endpoint name.')
```
jsntcy, line 950: - Hold on, we need discuss how to fix it.
- Regarding this PR, I'm curious what's change for the help message as you add `arg_type=workspace_name_arg_type` for the argument `managed_private_endpoint_name`? I tried, but no change for `--workspace-name`.

navba-MSFT: Currently running `az synapse managed-private-endpoints` command to perform delete operation with only --pe-name fails with below error:
(--workspace-name | --ids) are required

So the `workspace_name_arg_type` parameter should be made **required** here and _not optional._ Also [this](https://docs.microsoft.com/en-us/cli/azure/synapse/managed-private-endpoints?view=azure-cli-latest#az-synapse-managed-private-endpoints-delete) documentation is incorrect.

navba-MSFT: @jsntcy Thanks for looking into this. I shall wait for your reply.

jsntcy: We'll discuss if we should fix this issue inside CLI team. 
Even if we need fix it, we should fix it in cli core, but not in specific modules as it's a shared design for all modules (i.e. if user provides `--ids`, like `/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/synapse-cli000001/providers/Microsoft.Synapse/workspaces/clitest000003`, then they don't need provide `--workspace-name` as `--ids` contains everything including `--workspace-name`.

navba-MSFT: @jsntcy Thanks for the update. Please keep me posted about the progress, so that I can update the OP of this issue https://github.com/Azure/azure-cli/issues/23562 and set the expectations.

jsntcy:
Already discussed in Azure CLI team and decided not to fix it currently as it's not a bug and we need consider how to improve our help message in the future if there are more customers complaining about it.

PR #23712, opened 2022-08-30
{RDBMS} Fix check for data encryption for geo-redundant server

alanenriqueo:
**Related command**
`az mysql flexible-server create/update`

**Description**
Fixing a condition check for servers with geo-redundant backup enabled: if we are going to create/update a geo-redundant server and touch data encryption properties, we need to set both principal keys and backup keys; but if we don't want to touch that properties and instead update other things the check will fail and ask for backup key anyway when it's not intended. So, the fix is just to detect if **principal key is provided but backup key is not provided**, not only check if the backup key is not provided.

---


yonzhan:
RDBMS

evelyn-ys (approved):

PR #23722, opened 2022-08-30
{CosmosDB} Making sure analytical-storage-ttl is applied when it is the only container-specific argument when creating a Cosmos DB Mongo collection

toswedlu:
When creating a Cosmos DB Mongo collection, the parameter 'analytical-storage-ttl' currently isn't added to the HTTP payload if it is the only container-specific parameter.  For example, the following command:

az cosmosdb mongodb -g group-name -a acct-name -d db-name -n coll-name --analytical-storage-ttl -1

will not add the analytical storage ttl to the payload resource and hence won't enable analytical storage on the container.  The issue is that the ttl value wasn't checked for a non-None value when creating the resource.  The ttl will be applied if other container-specific parameters, such as --shard, are also supplied as these parameters are checked for value. 

Added a relevant test to existing Mongo collection creation tests.

---


yonzhan:
CosmosDB

evelyn-ys (approved):

PR #23795, opened 2022-09-06
[App Config] Add support for exporting App Configuration references to App Service

albertofori:
**Related command**
`az appconfig kv import`
`az appconfig kv export`

**Description**
1. Adds support for exporting app configuration settings as references to app service. Can be done by adding the flag `--export-as-reference` or `-r` when exporting to an app service destination.
2. App Configuration references are skipped when importing configuration settings from app service to App Configuration.

**Testing Guide**

**History Notes**

[App Config] `az appconfig kv export`: Support exporting app configuration settings as references to App Service
[App Config] `az appconfig kv import`: Ensure app configuration references are not imported from App Service

---


yonzhan:
App Config

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -281,6 +281,19 @@ def __write_kv_and_features_to_file(file_path, key_values=None, features=None, f
         raise CLIError("Failed to export key-values to file. " + str(exception))
 
 
+# Exported in the format @Microsoft.AppConfiguration(Endpoint=<storeEndpoint>; Key=<kvKey>; Value=<kvValue>; Label=<kvLabel>).
```
avanigupta, line 284: We shouldn't export any 'Value' in the reference

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -518,10 +531,12 @@ def __serialize_kv_list_to_comparable_json_object(keyvalues, level):
...
         for kv in keyvalues:
             kv_json = {'value': kv.value}
+            # Explicitly assign slot settings for comparison.
+            slot_setting = 'false'
             if kv.tags:
-                for tag_k, tag_v in kv.tags.items():
-                    if tag_k == 'AppService:SlotSetting':
-                        kv_json[tag_k] = tag_v
+                slot_setting = kv.tags.get('AppService:SlotSetting', 'false')
+            kv_json['AppService:SlotSetting'] = slot_setting
```
avanigupta, line 538: Should we move this tag to the constants file too?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: I see you're not really adding value in the reference. Can you update the comment?

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -189,6 +189,7 @@ def load_arguments(self, _):
 
     with self.argument_context('appconfig kv export', arg_group='AppService') as c:
         c.argument('appservice_account', validator=validate_appservice_name_or_id, help='ARM ID for AppService OR the name of the AppService, assuming it is in the same subscription and resource group as the App Configuration. Required for AppService arguments')
+        c.argument('export_as_reference', options_list=['--export-as-reference', '-r'], help="Export key-values as App Configuration references (Preview). For more information, see https://docs.microsoft.com/en-us/azure/app-service/app-service-configuration-references ")
```
avanigupta, line 192: We can add preview tag to this argument (no need to add 'Preview' in the help text).

```
c.argument('export_as_reference', is_preview=True, .....)
```

avanigupta: Another thing we can add is the three state flag:
` c.argument('export_as_reference', is_preview=True, arg_type=get_three_state_flag(), .......)`

This enables users to provide true/false/empty values to this argument. You can look at `--enable-public-network` argument for sample usage.

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -132,11 +132,16 @@ def is_valid_connection_string(connection_string):
...
+    return None
+
+
+def get_store_endpoint_from_connection_string(connection_string):
     if is_valid_connection_string(connection_string):
         segments = dict(seg.split("=", 1) for seg in connection_string.split(";"))
-        endpoint = segments.get("Endpoint")
-        if endpoint:
-            return endpoint.split("//")[1].split('.')[0]
+        return segments.get("Endpoint")
```
avanigupta, line 144: Can we return `segments.get("Endpoint", None)`?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -281,6 +281,19 @@ def __write_kv_and_features_to_file(file_path, key_values=None, features=None, f
         raise CLIError("Failed to export key-values to file. " + str(exception))
 
 
+# Exported in the format @Microsoft.AppConfiguration(Endpoint=<storeEndpoint>; Key=<kvKey>; Value=<kvValue>; Label=<kvLabel>).
+# Label is optional
+
+def __map_to_appservice_config_reference(key_value, endpoint):
+    if endpoint is None:
+        raise CLIError("No endpoint provided.")
```
avanigupta, line 289: Instead of raising error, we can log a warning and continue exporting without converting it to reference. This decision can be made even before __map_to_appservice_config_reference is invoked. You are already resolving store endpoint, if you just add a None check over there, we can log warning and export as regular key-values. Thoughts?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -281,6 +281,19 @@ def __write_kv_and_features_to_file(file_path, key_values=None, features=None, f
...
+
+    label = key_value.label
+    key_value.value = AppServiceConstants.APPSVC_CONFIG_REFERENCE_PREFIX + 'Endpoint={0}; Key={1}'.format(
+        endpoint, key_value.key) + ('; Label={0}'.format(label) if label is not None else '') + ')'
```
avanigupta, line 293: `key_value.key` has already been trimmed of the prefix that user might have specified with `--prefix` argument (we do this prefix trimming in `__read_kv_from_config_store`). 
While the AppSetting name is not impacted if the prefix is trimmed, this reference will not work since the trimmed key is not present in AppConfig.

I think this should be the behavior when --prefix and -r both are provided:

Lets say the prefix is "Beta_".

AppConfig key-value:
Key: Beta_Feature
Value: Test

After exporting to AppService:
Key: Feature
Value: @Microsoft.AppConfiguration(Endpoint=mystore.azconfig.io; Key=Beta_Feature)

avanigupta: The simplest solution I can think of is pass the prefix to this method and append it to the value string appropriately.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -416,19 +429,19 @@ def __read_kv_from_app_service(cmd, appservice_account, prefix_to_add="", conten
             cmd, resource_group_name=appservice_account["resource_group"], name=appservice_account["name"], slot=slot)
         for item in settings:
             key = prefix_to_add + item['name']
-            if validate_import_key(key):
+            value = item['value']
+            if validate_import_key(key) and not value.strip().lower().startswith(AppServiceConstants.APPSVC_CONFIG_REFERENCE_PREFIX.lower()):  # Exclude app configuration references.
```
avanigupta, line 433: Can we add a warning log for every AppConfig reference that we are skipping during import? Users may not be expecting it so its better to be more explicit.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
avanigupta: Can you also add a `validator` for this argument?
I think we should throw an error if this argument is used with any destination other than AppService. You can look at this validator for a sample: https://github.com/Azure/azure-cli/blob/f7479671e841d196d3312523c302bab853eb25c2/src/azure-cli/azure/cli/command_modules/appconfig/_validators.py#L226

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -281,6 +281,19 @@ def __write_kv_and_features_to_file(file_path, key_values=None, features=None, f
...
 
 
+# Exported in the format @Microsoft.AppConfiguration(Endpoint=<storeEndpoint>; Key=<kvKey>; Value=<kvValue>; Label=<kvLabel>).
+# Label is optional
+
+def __map_to_appservice_config_reference(key_value, endpoint):
+    if endpoint is None:
+        raise CLIError("No endpoint provided.")
+
+    label = key_value.label
```
avanigupta, line 291: What is the exact value of `key_value.label` when a key-value has no label? Is it `None`?

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
albertofori: `.get` returns `None` by default. Should we still explicitly state it?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
albertofori: You make a good point. Come to think of it, the None check is redundant since we would have already fetched `src_kvs`. That would mean the endpoint would have been resolved successfully else an exception would have been thrown when trying to create `azconfig_client`.
Could I be missing something here? 
If not, would you suggest we consider this case?

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -249,6 +251,12 @@ def export_config(cmd,
     if skip_keyvault:
         src_kvs = [keyvalue for keyvalue in src_kvs if keyvalue.content_type != KeyVaultConstants.KEYVAULT_CONTENT_TYPE]
 
+    if destination == 'appservice' and export_as_reference:
+        if endpoint is None:
+            endpoint = get_store_endpoint_from_connection_string(connection_string) or resolve_store_metadata(cmd, name)[1]
```
avanigupta, line 256: This is where I was thinking we can validate if an endpoint has been resolved or not:
```
endpoint = get_store_endpoint_from_connection_string(connection_string) or resolve_store_metadata(cmd, name)[1]
if endpoint:
     src_kvs = [__map_to_appservice_config_reference(kv, endpoint) for kv in src_kvs]
else:
    logger.warning("Cannot export as AppConfig references because store endpoint could not be resolved. Key-values will be exported as-is.")
```

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: See my other comment for this.

avanigupta: See this comment: https://github.com/Azure/azure-cli/pull/23795#discussion_r977926226

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
albertofori: Thanks for pointing this out. pushed an update to address this.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
albertofori: Since trimming will check whether the prefix exists before removing it, I move the trimming of the prefix to the `__map_to_appservice_config_reference` when the `export_as_reference` flag is indicated.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
albertofori: Update has been pushed for this. Kindly let me know if the wording works.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
albertofori: Yes, it is.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
albertofori: Update pushed for this.

src/azure-cli/azure/cli/command_modules/appconfig/_constants.py
```diff
@@ -16,7 +16,12 @@ class FeatureFlagConstants:
 
 class KeyVaultConstants:
     KEYVAULT_CONTENT_TYPE = "application/vnd.microsoft.appconfig.keyvaultref+json;charset=utf-8"
+
+
+class AppServiceConstants:
+    APPSVC_CONFIG_REFERENCE_PREFIX = "@Microsoft.AppConfiguration("
```
avanigupta, line 22: I just noticed that APPSVC_KEYVAULT_PREFIX doesn't end with `(`. Can we have AppConfig prefix without the ending bracket for consistency?

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/test_appconfig_commands.py
```diff
@@ -937,6 +937,73 @@ def test_appconfig_to_appservice_import_export(self, resource_group, location):
         slot = self.create_random_name(prefix='Slot', length=24)
         self.cmd('webapp deployment slot create -g {} -n {} -s {}'.format(resource_group, webapp_name, slot))
 
+        # App configuration reference tests
+        # Create new key-values
+        entry_key = 'TestKey'
+        entry_value = 'TestValue'
+        entry_label = 'AppServiceReferenceExport'
```
avanigupta, line 944: Do you mind adding a test for null label too? Just to make sure that the reference string is formed as expected.

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -243,12 +245,19 @@ def export_config(cmd,
...
                                           key=key,
                                           label=label if label else SearchFilterOptions.EMPTY_LABEL,
-                                          prefix_to_remove=prefix,
+                                          prefix_to_remove=prefix if not export_as_reference else "",
                                           cli_ctx=cmd.cli_ctx if resolve_keyvault else None)
 
     if skip_keyvault:
         src_kvs = [keyvalue for keyvalue in src_kvs if keyvalue.content_type != KeyVaultConstants.KEYVAULT_CONTENT_TYPE]
 
+    if destination == 'appservice' and export_as_reference:
```
avanigupta, line 254: We can move this block after we have invoked `__discard_features_from_retrieved_kv`. That's because we dont want to call `__map_to_appservice_config_reference` for any feature flags.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -281,6 +281,19 @@ def __write_kv_and_features_to_file(file_path, key_values=None, features=None, f
...
+# Label is optional
+
+def __map_to_appservice_config_reference(key_value, endpoint, prefix):
+    label = key_value.label
+    key_value.value = AppServiceConstants.APPSVC_CONFIG_REFERENCE_PREFIX + 'Endpoint={0}; Key={1}'.format(
+        endpoint, key_value.key) + ('; Label={0}'.format(label) if label is not None else '') + ')'
+
+    if key_value.key.startswith(prefix):
+        key_value.key = key_value.key[len(prefix):]
+    return key_value
```
avanigupta, line 294: In `__write_kv_to_app_service`, we check if the content type is that of KeyVault reference and try to convert the value to AppService format KeyVault reference. But if we have already converted a KeyVault reference to an AppConfig reference here, the check in  `__write_kv_to_app_service` is unnecessarily going to raise an exception while deserializing the value. Although the exception is not thrown and the value is treated as regular string value, we should avoid doing things that will definitely raise exceptions, because catching an exception is expensive.

Can you check if there are any side effects if we clear the content type for all key-values here? I dont think we need the content type while exporting to AppService after this point.

albertofori: I confirmed that the content type is not used when exporting to app service. So this can be changed without any side effects. Technically, the key-value is no more a key-vault reference.  

Another option would be to supply the `export-as-reference` flag to the `__write_kv_to_app_service` method and skip the key-vault reference validation when it is true.

avanigupta: I'd prefer changing the content type, because as you said - its technically not a KeyVault reference anymore.

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -251,21 +251,21 @@ def export_config(cmd,
...
-        return
+            __export_kvset_to_file(file_path=path, keyvalues=src_kvs, yes=yes)
+            return
+
+    if destination == 'appservice' and export_as_reference:
+            if endpoint is None:
+                # Endpoint will not be None as it is already resolved in creating azconfig_client
+                endpoint = get_store_endpoint_from_connection_string(connection_string) or resolve_store_metadata(cmd, name)[1]
+
+            src_kvs = [__map_to_appservice_config_reference(kv, endpoint, prefix) for kv in src_kvs]
```
avanigupta, line 268: Looks like an extra tab space got added to these lines

avanigupta (approved):

zhoxing-ms:
@albertofori Could you please resolve those CI issues?

zhoxing-ms:
> [App Config] Change: az appconfig kv export: Support exporting app configuration settings as references to App Service
[App Config] Change: az appconfig kv import: Ensure app configuration references are not imported from App Service

For ordinary changes that are not breaking changes, we do not need to add the `Change` tag to the PR description/title.

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
albertofori: My bad! Just fixed it.

albertofori:
> @albertofori Could you please resolve those CI issues?

@zhoxing-ms Please I just pushed a fix to address the issues, thanks!

zhoxing-ms (approved):

Issue #23797, opened 2022-09-07
Cannot pass `--web-redirect-uris` multiple URIs with Powershell variable


JackSch:
**Related command**

`az ad app update`

**Describe the bug**

--Using Powershell--
Adding a redirect URI to an app is common practice for us.
`--web-redirect-uris` is rather particular with how values are given to it. As of 2.37 with the migration of `az ad` to Microsoft Graph API, we use the following script to append a single URI to the list and it's worked pretty well:
```
$app = az ad app show --id $appId | ConvertFrom-Json
$redirectUris = $app.web.redirectUris
$redirectUris += $newRedirectUri
$redirectUris = $redirectUris | ForEach-Object {"`"$_`""}
az ad app update --id $appId --web-redirect-uris $redirectUris
```
This has been sufficient until 2.40, where this no longer works. I have not been able to figure out a way to pass multiple URIs in a Powershell variable to `--web-redirect-uris`.

**To Reproduce**

Use Az CLI 2.39:
1. Create a App Registration in AAD. Add two or three redirect URIs.
  a. I.e., `https://site1.example.com/signin-oidc`, `https://site2.example.com/signin-oidc`
3. Assign a new URI to `$newRedirectUri` variable (or similar), and run the script snippet.
4. The command should be successful

`$redirectUris` should appear like so when printed to console:
```
"https://site1.example.com/signin-oidc"
"https://site2.example.com/signin-oidc"
"https://site3.example.com/signin-oidc"
```

Try again with Az CLI 2.40. It no longer works:
`One or more properties contains invalid values.`

**Expected behavior**

At least in Powershell, each URI must be wrapped in quotes. Double quotes works for us because we're using vars. This is achieved with line 4 of the script. Each URI must also be space separated as per the documentation. The `$redirectUris` variable is an _array_, however Powershell appears to just print out each URI sequentially when the command is run. It behaves like so:
`az ad app update --id $AppId --web-redirect-uris "https://site1.example.com/signin-oidc" "https://site2.example.com/signin-oidc" "https://site3.example.com/signin-oidc" "https://site4.example.com/signin-oidc"`
This satisfies the command as of 2.39.

**Environment summary**

Az CLI 2.40, Windows 11, Powershell 7

**Additional context**

There does not appear to be any changes to `az ad` for 2.40 according to the release notes.
The `--set` flag doesn't appear to be much use either. Perhaps I am using it wrong, but that's outside of the scope of this issue.
[...]

PR #23802, opened 2022-09-07
{Core} Entry script: Use `@args` to tolerate incorrectly passing list with `$`

jiasli:
Close https://github.com/Azure/azure-cli/issues/23797

https://github.com/Azure/azure-cli/pull/23514 changed the entry script `az.cmd` to `az.ps1` when executed in PowerShell.

```powershell
& "$PSScriptRoot\..\python.exe" -IBm azure.cli $args
```

Consider we have an array

```
> $mylist = "a", "b"
```

Some user passes the array to `az.ps1` as `$mylist`:

```powershell
> az --debug $mylist
cli.knack.cli: Command arguments: ['--debug', 'a b']
```

This is **incorrect**, as `$args` will become an array of array. Using `$args` to call a native exe will only **splat the outer array** and **concatenate the inner array's items**.

The correct way to a pass a list to `az` is to pass **space-separated arguments**, so in PowerShell, user should use [`@` splatting](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_splatting?view=powershell-7.2) when calling `az.ps1`:

```powershell
> az --debug @mylist
cli.knack.cli: Command arguments: ['--debug', 'a', 'b']
```

Meanwhile, when calling native exe with `@args`, it will even splat internal arrays. We can use a test script `test.ps1` to verify this behavior:

```powershell
echo $args.Length $args.GetType()
echo $args[0].GetType()
python -c "import sys; print(sys.argv)" $args
python -c "import sys; print(sys.argv)" @args
```

```powershell
> ./test.ps1 $mylist $mylist
2

IsPublic IsSerial Name                                     BaseType
-------- -------- ----                                     --------
True     True     Object[]                                 System.Array
True     True     Object[]                                 System.Array
['-c', 'a b', 'a b']
['-c', 'a', 'b', 'a', 'b']

> ./test.ps1 @mylist @mylist
4

IsPublic IsSerial Name                                     BaseType
-------- -------- ----                                     --------
True     True     Object[]                                 System.Array
True     True     String                                   System.Object
['-c', 'a', 'b', 'a', 'b']
['-c', 'a', 'b', 'a', 'b']
```

Interesting thing is that even `@args` can only splat 2 layers and the 3rd layer was concatenated:

```powershell
> ./test.ps1 ($mylist, $mylist)
1

IsPublic IsSerial Name                                     BaseType
-------- -------- ----                                     --------
True     True     Object[]                                 System.Array
True     True     Object[]                                 System.Array
['-c', 'System.Object[] System.Object[]']
['-c', 'a b', 'a b']
```

We can go even deeper:

```powershell
> ./test.ps1 (($mylist, $mylist),($mylist, $mylist))
1

IsPublic IsSerial Name                                     BaseType
-------- -------- ----                                     --------
True     True     Object[]                                 System.Array
True     True     Object[]                                 System.Array
['-c', 'System.Object[] System.Object[]']
['-c', 'System.Object[] System.Object[]', 'System.Object[] System.Object[]']
```


jiasli:
CI task "Check the Format of Pull Request Title and Content" failed because of some example commands in the PR description @wangzelin007:

https://dev.azure.com/azure-sdk/public/_build/results?buildId=1835012&view=logs&j=a8943ac2-38d7-5792-f2a7-5f4fd06db24e&t=3ed51913-4dd7-564f-a8d8-fda07de13946&l=115

```
['-c', 'a b', 'a b']: missing space after ]
                   ↑
```

build_scripts/windows/scripts/az.ps1
```diff
@@ -1,217 +1,217 @@
 $env:AZ_INSTALLER="MSI"
-& "$PSScriptRoot\..\python.exe" -IBm azure.cli $args
+& "$PSScriptRoot\..\python.exe" -IBm azure.cli @args
```
jiasli, line 2: The disadvantage of using `@args` is also obvious, as we can't pass `"a b"` string as `("a", "b")` now.

Issue #23807, opened 2022-09-07
az ad user get-member-groups returns "Specified HTTP method is not allowed for the request target."


gsrx:
Calling az ad user get-member-groups --id "someuserprincipalname" returns error: "Specified HTTP method is not allowed for the request target."

--debug switch shows attempting POST to graph api eg.
https://graph.microsoft.com:443 "POST /v1.0/users/username HTTP/1.1" 405 None

Running on Windows 10
az version
{
  "azure-cli": "2.40.0",
  "azure-cli-core": "2.40.0",
  "azure-cli-telemetry": "1.0.8",
  "extensions": {
    "account": "0.2.1",
    "azure-devops": "0.22.0",
    "datafactory": "0.5.0"
  }
}

yonzhan:
@jiasli for awareness

jiasli:
Could you please share the full URL, but removing your personal information? I doubt something is wrong with the URL.

jiasli:
I think it is because you have `#` in your user principal ID:

```
> az ad user get-member-groups --id 'jiasli_microsoft.com#EXT#@AzureSDKTeam.onmicrosoft.com' --debug
...
cli.azure.cli.core.util: Request URL: 'https://graph.microsoft.com/v1.0/users/jiasli_microsoft.com#EXT%23@AzureSDKTeam.onmicrosoft.com/getMemberGroups'
urllib3.connectionpool: https://graph.microsoft.com:443 "POST /v1.0/users/jiasli_microsoft.com HTTP/1.1" 405 None
```

When the URL was sent to Graph service, characters after `#` are dropped. I will fix it in a PR.

PR #23819, opened 2022-09-08
[AD] Support special characters in user principal name

jiasli:
**Related command**
`az ad user`

**Description**
Close https://github.com/Azure/azure-cli/issues/23807

The Graph API document [directoryObject: getMemberGroups](https://docs.microsoft.com/en-us/graph/api/directoryobject-getmembergroups?view=graph-rest-1.0&tabs=http) doesn't mention `userPrincipalName` should be URL encoded, but it is mentioned in [Get a user](https://docs.microsoft.com/en-us/graph/api/user-get?view=graph-rest-1.0&tabs=http#http-request) API.

> To query a B2B user using the **userPrincipalName**, encode the hash (#) character. That is, replace the `#` symbol with `%23`. For example, `/users/AdeleVance_adatum.com%23EXT%23@contoso.com`.

`@` doesn't need to be encoded. However, similar to what I described in https://github.com/Azure/azure-cli/pull/22739, there is also an inconsistency here between the doc and the old `graphrbac` SDK. SDK also encodes `@`:

```
az ad user get-member-groups --id jiasli_microsoft.com#EXT#@AzureSDKTeam.onmicrosoft.com --debug
...
msrest.http_logger: Request URL: 'https://graph.windows.net/54826b22-38d6-4fb2-bad9-b7b93a3e9c5a/users/jiasli_microsoft.com%23EXT%23%40AzureSDKTeam.onmicrosoft.com?api-version=1.6'
```

This was done by `msrest`:

https://github.com/Azure/msrest-for-python/blob/c97db9df00c7802b1a1af3ab8dcdbf6a5acbbdd7/msrest/serialization.py#L659

```py
                output = quote(str(output), safe='')
```
Here we also encode `@` so that we can reuse `azure.cli.testsdk.utilities.MSGraphNameReplacer`.


yonzhan:
AD

src/azure-cli/azure/cli/command_modules/role/_msgrpah/_graph_client.py
```diff
@@ -364,6 +357,24 @@ def _filter_to_query(filter):
...
+    # https://docs.microsoft.com/en-us/graph/api/user-get
+    # https://docs.microsoft.com/en-us/graph/known-issues#users
+
+    # UPN
+    if '@' in id_or_upn:
+        from urllib.parse import quote
+        encoded_upn = quote(id_or_upn, safe='')
+
+        if id_or_upn.startswith('$'):
+            return f"/users('{encoded_upn}')"
```
evelyn-ys, line 371: Here's a little bit difference from the original implementation. Is it possible that `id_or_upn` doesn't contain `@` but starts with `$`?

jiasli: Nope. If `id_or_upn` doesn't contain `@`, it can only be a GUID.

evelyn-ys (approved):

jiasli:
Many tests fail with 

```
E           /v1.0/users/testuser000002%40azuresdkteam.onmicrosoft.com != /v1.0/users/testuser000002@azuresdkteam.onmicrosoft.com
```

It looks like VCRPY can't match escaped characters in the URL path section.

calvinhzy (approved):

src/azure-cli/azure/cli/command_modules/role/tests/util.py
```diff
@@ -29,3 +30,17 @@ def retry(func, sleep_duration=ROLE_COMMAND_SLEEP_DURATION, max_retry=ROLE_COMMA
...
+    """
+    def __init__(self, test_name, mock_name):
+        self.test_name = test_name
+        self.test_name_encoded = test_name.replace('#', '%23')
+        self.mock_name = mock_name
+        self.mock_name_encoded = mock_name.replace('#', '%23')
+
+    def process_request(self, request):
+        request.uri = request.uri.replace(self.test_name_encoded, self.mock_name_encoded)
+        return request
```
jiasli, line 46: The old `AADGraphUserReplacer` only replaces the domain part

https://github.com/Azure/azure-cli/blob/427fc4d331526541dea7143849a9fa642019f8fe/src/azure-cli/azure/cli/command_modules/role/tests/latest/test_graph.py#L381

https://github.com/Azure/azure-cli/blob/8fc09120c6dd40bf43722c0c53b27a3279b4214c/src/azure-cli-testsdk/azure/cli/testsdk/utilities.py#L160-L168

and leaves the object name renaming job to `GeneralNameReplacer`. This works fine for object names without special characters, but this doesn't work in our current case since our object name contains special characters and is encoded, not only the domain part.

src/azure-cli/azure/cli/command_modules/role/_msgrpah/_graph_client.py
```diff
@@ -364,6 +357,21 @@ def _filter_to_query(filter):
...
+    # https://docs.microsoft.com/en-us/graph/known-issues#users
+
+    # UPN
+    if '@' in id_or_upn:
+        # According to the doc, only encode #, but not @ and $
+        id_or_upn = id_or_upn.replace('#', '%23')
+
+    if id_or_upn.startswith('$'):
+        return f"/users('{id_or_upn}')"
+    return f"/users/{id_or_upn}"
```
jiasli, line 372: @calvinhzy, this is more consistent with your original code. 😉

evelyn-ys (approved):

calvinhzy (approved):

PR #23826, opened 2022-09-08
{RDBMS} Add descriptive error messages when user doesn't have read access to resource

alanenriqueo:
**Related command**
`az mysql/postgres flexible-server create/restore`

**Description**
- When checking the existence of a resource group we perform a HEAD request to `/subscriptions/{subscriptionId}/resourcegroups/{resourceGroupName}`, but if the user doesn't have the permission `Microsoft.Resources/subscriptions/resourceGroups/read` it will only return HTTP 403 Forbidden without explaining the reason why it failed. Hence, we add a descriptive error message.
- When checking the existence of a vnet/subnet we make a GET request to the resource, but we only check if there was an `HttpResponseError` exception. Hence, we need to tell the customer (throw a CLIException) if we receive HTTP 403.

**Testing Guide**
- Create a server in a resource group that doesn't exist yet: `az mysql flexible-server create -g resourceGroup -n serverName`
- Restore a server in a vnet that is in a different resource group, or in a vnet/subnet that doesn't exist yet: `az mysql flexible-server restore -g resourceGroup --source-server sourceServer --name serverName --subnet /subscriptions/{subscriptionId}/resourceGroups/{resourceGroup}/providers/Microsoft.Network/virtualNetworks/{vnet}/subnets/{subnet} --address-prefixes 10.0.0.0/16 --subnet-prefixes 10.0.1.0/24`

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/rdbms/_flexible_server_util.py
+++ b/src/azure-cli/azure/cli/command_modules/rdbms/_flexible_server_util.py
@@ -18,6 +18,7 @@ from msrestazure.tools import parse_resource_id
 from msrestazure.azure_exceptions import CloudError
 from azure.cli.core.util import CLIError
 from azure.cli.core.azclierror import AuthenticationError
+from azure.core.exceptions import HttpResponseError
 from azure.core.paging import ItemPaged
 from azure.cli.core.commands.client_factory import get_subscription_id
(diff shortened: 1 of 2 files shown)


yonzhan:
RDBMS

DaeunYim (approved):
Looks good. Thanks for the fix

evelyn-ys (approved):

PR #22774, opened 2022-06-08
[Core] WAM integration

rayluo:
**Related command**

* `az login`
* And any subsequent command that would obtain a token via MSAL's `acquire_token_silent()`.

**Description**

Close #21201
Require https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/415

This PR is currently a draft to demonstrate how Azure CLI could enable the broker behavior from an MSAL feature branch. 

**Testing Guide**

1. Run the following commands on a Windows 10+ machine
2. Somehow install Azure CLI from this feature branch, conceptually similar to this (however this command does not currently work)
   `pip install git+https://github.com/azure/azure-cli@broker-integration`
3. Then run `az login`, and notice how a pop-up window (rather than a browser) would pop up.
4. Try run some more (all?) different `az foo bar` command to make sure the token returned by broker would work

**History Notes**

---

This checklist is used to make sure that common guidelines for a pull request are followed.

- [ ] The PR title and description has followed the guideline in [Submitting Pull Requests](https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules#submitting-pull-requests).

- [ ] I adhere to the [Command Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md).

- [ ] I adhere to the [Error Handling Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/error_handling_guidelines.md).

PR #23828, opened 2022-09-09
[Core] PREVIEW: Support Web Account Manager (WAM) login on Windows

### LINKED ISSUE #21201: Add WAM support in Azure CLI

A near-term future release of MSAL Python will enable broker-based authentication through Web Account Manager (WAM) on Windows devices. The benefits of authentication through WAM are described [in this document](https://docs.microsoft.com/en-us/azure/active-directory/develop/scenario-desktop-acquire-token-wam#wam-value-proposition) under the "WAM value proposition" heading. 

Upgrade MSAL Python (upon release of the latest version including WAM support) and enable authentication through WAM. 

**Additional context**

@rayluo is currently driving the work to enable WAM support within MSAL Python and the functionality will be ready for testing by mid-February. Ahead of release we would like the Azure CLI team to run tests against WAM functionality / identify any bugs in a feature branch @rayluo will provide.

jiasli:
**Related command**
`az login`

**Description**
Close #21201
Require https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/415
Rework https://github.com/Azure/azure-cli/pull/22774

**Testing Guide**

On Windows, users can opt in WAM login using:

```
az config set core.allow_broker=true
```

Then run

```
az login
```

Instead of launching a web browser, an account selector will be shown to ask the user to select a Windows account:
 
image

**History Notes**

[Core] PREVIEW: Support Web Account Manager (WAM) login on Windows. To opt in, run `az config set core.allow_broker=true`

**Additional information**

WAM will be enabled by default in year 2023.

### DIFF
--- a/src/azure-cli-core/azure/cli/core/_profile.py
+++ b/src/azure-cli-core/azure/cli/core/_profile.py
@@ -766,7 +766,7 @@ class SubscriptionFinder:
                     # The tenant requires MFA and can't be accessed with home tenant's refresh token
                     mfa_tenants.append(t)
                 else:
-                    logger.warning("Failed to authenticate '%s' due to error '%s'", t, ex)
+                    logger.warning("Failed to authenticate %s due to error '%s'", t.tenant_id_name, ex)
                 continue
 
             if not subscriptions:
@@ -856,4 +856,8 @@ def _create_identity_instance(cli_ctx, *args, **kwargs):
     # EXPERIMENTAL: Use core.use_msal_http_cache=False to turn off MSAL HTTP cache.
     use_msal_http_cache = cli_ctx.config.getboolean('core', 'use_msal_http_cache', fallback=True)
 
-    return Identity(*args, encrypt=encrypt, use_msal_http_cache=use_msal_http_cache, **kwargs)
+    # PREVIEW: On Windows, use core.allow_broker=true to use broker (WAM) for authentication.
+    allow_broker = cli_ctx.config.getboolean('core', 'allow_broker', fallback=False)
+
+    return Identity(*args, encrypt=encrypt, use_msal_http_cache=use_msal_http_cache, allow_broker=allow_broker,
+                    **kwargs)
--- a/src/azure-cli-core/azure/cli/core/auth/identity.py
+++ b/src/azure-cli-core/azure/cli/core/auth/identity.py
@@ -136,7 +136,7 @@ class Identity:  # pylint: disable=too-many-instance-attributes
         # Emit a warning to inform that a browser is opened.
         # Only show the path part of the URL and hide the query string.
 
-        def _prompt_launching_ui(ui=None):
+        def _prompt_launching_ui(ui=None, **_):
             if ui == 'browser':
                 logger.warning("A web browser has been opened at %s. Please continue the login in the web browser. "
                                "If no web browser is available or if the web browser fails to open, use device code "
--- a/src/azure-cli-core/azure/cli/core/auth/identity.py
+++ b/src/azure-cli-core/azure/cli/core/auth/identity.py
@@ -53,7 +53,8 @@ class Identity:  # pylint: disable=too-many-instance-attributes
     # It follows singleton pattern so that _secret_file is read only once.
     _service_principal_store_instance = None
 
-    def __init__(self, authority, tenant_id=None, client_id=None, encrypt=False, use_msal_http_cache=True):
+    def __init__(self, authority, tenant_id=None, client_id=None, encrypt=False, use_msal_http_cache=True,
+                 allow_broker=None):
         """
         :param authority: Authentication authority endpoint. For example,
             - AAD: https://login.microsoftonline.com
@@ -68,6 +69,7 @@ class Identity:  # pylint: disable=too-many-instance-attributes
         self.client_id = client_id or AZURE_CLI_CLIENT_ID
         self._encrypt = encrypt
         self._use_msal_http_cache = use_msal_http_cache
+        self._allow_broker = allow_broker
 
         # Build the authority in MSAL style
         self._msal_authority, self._is_adfs = _get_authority_url(authority, tenant_id)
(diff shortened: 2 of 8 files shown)


src/azure-cli-core/azure/cli/core/auth/identity.py
```diff
@@ -132,9 +135,15 @@ def _service_principal_store(self):
...
-                       "with `az login --use-device-code`.", self._msal_app.authority.authorization_endpoint)
+
+        def _prompt_launching_ui(ui=None):
+            if ui == 'browser':
+                logger.warning("A web browser has been opened at %s. Please continue the login in the web browser. "
+                               "If no web browser is available or if the web browser fails to open, use device code "
+                               "flow with `az login --use-device-code`.",
+                               self._msal_app.authority.authorization_endpoint)
+            elif ui == 'broker':
+                logger.warning("Please select the account you want to log in with.")
```
jiasli, line 146: We may want to refine the warning message.

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -856,4 +856,8 @@ def _create_identity_instance(cli_ctx, *args, **kwargs):
     # EXPERIMENTAL: Use core.use_msal_http_cache=False to turn off MSAL HTTP cache.
     use_msal_http_cache = cli_ctx.config.getboolean('core', 'use_msal_http_cache', fallback=True)
 
-    return Identity(*args, encrypt=encrypt, use_msal_http_cache=use_msal_http_cache, **kwargs)
+    # PREVIEW: On Windows, use core.allow_broker=true to use broker (WAM) for authentication.
+    allow_broker = cli_ctx.config.getboolean('core', 'allow_broker', fallback=False)
```
jiasli, line 860: We may want to align the config name with Azure PowerShell @Jacekey23, @chasewilson

yonzhan:
WAM support

src/azure-cli-core/azure/cli/core/auth/identity.py
```diff
@@ -132,9 +135,15 @@ def _service_principal_store(self):
     def login_with_auth_code(self, scopes, **kwargs):
         # Emit a warning to inform that a browser is opened.
         # Only show the path part of the URL and hide the query string.
-        logger.warning("A web browser has been opened at %s. Please continue the login in the web browser. "
-                       "If no web browser is available or if the web browser fails to open, use device code flow "
-                       "with `az login --use-device-code`.", self._msal_app.authority.authorization_endpoint)
+
+        def _prompt_launching_ui(ui=None):
```
rayluo, line 139: Suggest to do it this way for it to be future-proof.
```suggestion
        def _prompt_launching_ui(ui=None, **ignored):
```

rayluo (approved):
Approving with one more suggestion below.

src/azure-cli-core/azure/cli/core/auth/identity.py
jiasli: Sure. Will use the common convention to ignore: `**_`

src/azure-cli-core/azure/cli/core/_profile.py
jiasli: As discussed offline, Azure PowerShell will release separate versions for WAM, instead of using a config option.

evelyn-ys: As mentioned, WAM will be enabled by default in year 2023. By then, should all linux users manually set `core.allow_broker=False` to avoid failure?

chasewilson: Ideally, we don't have the users manually adjust anything to avoid failure. That's not a great experience if the default login is an error. Is it possible to set the configuration appropriately from an environment variable? eg in PowerShell `$IsLinux` or `$IsMacOS`. 

As for if we want to add a parameter like `--allow-broker`, we kind of run into the same issue. If we want to have the broker to be the default flow, we would need to set `--allow-broker` as a default parameter and only if it's Windows. Meaning we are dealing with the same issue. 

Open to more discussion around this as it is a big change to how we handle the flow.

chasewilson: > We may want to align the config name with Azure PowerShell. @Jacekey23 @chasewilson

Could you point me to a complete list of our configurations? Even if it's just the code location where the input is managed. I'd just like to look at all of our configurations and make sure we're following a good format.

rayluo: > As mentioned, WAM will be enabled by default in year 2023. By then, should all linux users manually set `core.allow_broker=False` to avoid failure? Can we add `--allow-broker` parameter for `az login` to distinguish WAM and traditional browser way?

> Ideally, we don't have the users manually adjust anything to avoid failure. That's not a great experience if the default login is an error. Is it possible to set the configuration appropriately from an environment variable? eg in PowerShell `$IsLinux` or `$IsMacOS`.
> 
> As for if we want to add a parameter like `--allow-broker`, we kind of run into the same issue. If we want to have the broker to be the default flow, we would need to set `--allow-broker` as a default parameter and only if it's Windows. Meaning we are dealing with the same issue.
> 
> Open to more discussion around this as it is a big change to how we handle the flow.

Those are all good points. MSAL Python foresaw this difficulty, so, one of the design goal is to save app developers from needing to conditionally enable/disable broker behavior. Our model is, the app developer only needs to opt-in when their app meets those [prerequisites](https://msal-python.readthedocs.io/en/1.20.0b1/#msal.PublicClientApplication.params.allow_broker), and then MSAL Python will take care of those conditions such as the platforms. By the way, that is also why the parameter was named "allow_broker" rather than "enable_broker": after the app allows it, the MSAL will decide whether to actually use it. For example, currently MSAL Python effectively ignores allow_broker when current app is running on non-Windows. (You can try test Azure CLI dev branch on Linux to see how allow_broker=True is a no-op.)

In particular, please do not add "--allow-broker" as a parameter to "az login". The reason being, even though the broker UI is only visible/tangible during an "az login", the broker mechanism is holistic. It will also affect silent token refresh. So, it has to be a global toggle switch.

evelyn-ys (approved):

Issue #24005, opened 2022-09-26
`az` can't be invoked by `Start-Process` since 2.40.0


jiasli:
**Describe the bug**
Before Azure CLI 2.40.0, `az` can be invoked by `Start-Process`:

```
> Start-Process -FilePath az -NoNewWindow -ArgumentList version -Wait
{
  "azure-cli": "2.40.0",
  "azure-cli-core": "2.40.0",
  "azure-cli-telemetry": "1.0.8",
  "extensions": {
    "account": "0.2.3"
  }
}
```

The entry script was migrated from `az.cmd` to `az.ps1` (https://github.com/Azure/azure-cli/pull/23514) in 2.40.0. `az` can no longer be invoked by `Start-Process`:

```
> Start-Process -FilePath az -NoNewWindow -ArgumentList version -Wait
Start-Process: This command cannot be run due to the error: %1 is not a valid Win32 application.
```

This is because `Start-Process` can't invoke a `ps1` script in a separate process.

Issue #23883, opened 2022-09-07
Azure CLI 2.40.0 breaks stdin passing on PowerShell


AzureMarker:
## Describe the bug
We're executing an Azure CLI extension in PowerShell which requires interactive input, and that input is passed through stdin using the following code:
```powershell
$user_in = "..."
$user_in | az arcappliance createconfig vmware --resource-group $resourceGroup --name $resourceName --location $location
``` 

This was working fine until the latest Azure CLI version (2.40.0).

**Command Name**
`az arcappliance createconfig vmware`
Extension Name: arcappliance.
Version: 0.2.16.

(note: we are using an older arcappliance extension version for testing purposes)

**Errors:**
```
ERROR: The command failed with an unexpected error. Here is the traceback:
ERROR: EOF when reading a line
Traceback (most recent call last):
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\knack\cli.py", line 233, in invoke
    cmd_result = self.invocation.execute(args)
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\azure\cli\core\commands\__init__.py", line 663, in execute
    raise ex
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\azure\cli\core\commands\__init__.py", line 726, in _run_jobs_serially
    results.append(self._run_job(expanded_arg, cmd_copy))
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\azure\cli\core\commands\__init__.py", line 697, in _run_job
    result = cmd_copy(params)
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\azure\cli\core\commands\__init__.py", line 333, in __call__
    return self.handler(*args, **kwargs)
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\azure\cli\core\commands\command_operation.py", line 121, in handler
    return op(**command_args)
  File "C:\Program Files\Common Files\AzureCliExtensionDirectory\arcappliance\azext_arcappliance\custom.py", line 193, in create_config
    provider.create_config(infra_config_path, working_dir)
  File "C:\Program Files\Common Files\AzureCliExtensionDirectory\arcappliance\azext_arcappliance\pkg\providers\vmware\provider.py", line 109, in create_config
    vsphere_configuration = self.__prep_for_kva(None, credentials_only=True)
[...]

Issue #23880, opened 2022-09-14
az command returns 0 when failing


fuhuifang:
**Related command**
`az  extension show -n azure-ml-cli`

**Describe the bug**
When azure-ml-cli is not installed, we expect an error with exitcode 1. This works properly for 2.39 and older, but with 2.40, it returns 0.

**To Reproduce**
Run the following python script:
```python
import subprocess
subprocess.run(args=['pwsh', '-EncodedCommand', 'YQB6ACAAZQB4AHQAZQBuAHMAaQBvAG4AIABzAGgAbwB3ACAALQBuACAAYQB6AHUAcgBlAC0AYwBsAGkALQBtAGwA'])
```
Outputs
2.40
![image](https://user-images.githubusercontent.com/32152743/190076962-9fdaf8b0-54d4-4c84-a5e6-37d56c8debad.png)
2.39
![image](https://user-images.githubusercontent.com/32152743/190077035-e0587dbf-35f3-4326-b6cd-3003b4f68b1e.png)

**Expected behavior**
CompletedProcess(args=['pwsh', '-EncodedCommand', 'YQB6ACAAZQB4AHQAZQBuAHMAaQBvAG4AIABzAGgAbwB3ACAALQBuACAAYQB6AHUAcgBlAC0AYwBsAGkALQBtAGwA'], returncode=**1**)

**Environment summary**
Windows, python 3.8

**Additional context**
Similar to https://github.com/Azure/azure-cli/issues/23808

yonzhan:
route to CXP team

yjm607:
Hi Yong, will there be an IcM ticket filed for this issue? It is affecting all customers of the Substrate Intelligence AI platform and the Shrike library.

jiasli:
This looks like some issue with PowerShell.

Say we have the same `test.cmd` and `test.ps1`:

```
python -c "exit(1)"
```

The exit code is correct when they are invoked directly:

```
PS D:\> ./test.cmd ; echo $LASTEXITCODE

D:\>python -c "exit(1)"
1
PS D:\> ./test.ps1 ; echo $LASTEXITCODE
1
```

However, when invoked by `pwsh.exe`, they return different result:

```
PS D:\> pwsh -Command "./test.cmd" ; echo $LASTEXITCODE

D:\>python -c "exit(1)"
1
PS D:\> pwsh -Command "./test.ps1" ; echo $LASTEXITCODE
0
```

If `test.ps1` contains 

```powershell
Exit 1
[...]

PR #24015, opened 2022-09-27
[Core] Revert #23514: Rename entry script `az.ps1` to `azps.ps1`

jiasli:
**Description**
https://github.com/Azure/azure-cli/pull/23514 added `az.ps1` so that when Azure CLI is invoked in PowerShell, `az.ps1` is called instead of `az.cmd`.

However, this change introduced 5 new issues:

**Arguments**
- #23797: PowerShell array `$mylist` is splatted when calling `.cmd`, but not when calling `.ps1`. Instead, the array is concatenated as a single string.
- #24034: PowerShell doesn't include `--` in `$args`

**`stdin`**
- #23883: When calling `.ps1`, PowerShell doesn't automatically connect `stdin` to pipeline.

**Exit code**
- #23880: Exit code is not surfaced when `pwsh.exe` calling `.ps1` calling `.exe`

**Shell interaction**
- #24005: `.cmd` can be invoked by `Start-Process`, but `.ps1` cannot.

This PR reverts the change by renaming `az.ps1` to `azps.ps1` so that users can still use the `.ps1` entry script if they want, but by default, when calling `az`, `az.cmd` will be called as before.

We will work with PowerShell team to see how to fix these issues.


yonzhan:
Users are able to use azpwsh.ps1 on demand.

yonzhan (approved):

chasewilson (commented):
Is there a problem with leaving it `az.ps1`? Is the customer able to specify their entry script by the extension?

jiasli:
> Is there a problem with leaving it `az.ps1`? Is the customer able to specify their entry script by the extension?

Yes. If we don't rename `az.ps1`, it will be the default script when `az` is called from PowerShell, overriding `az.cmd.`

chasewilson (approved):

PR #24020, opened 2022-09-27
[RDBMS]: Switch Back to Online migration by default for PG single to flex migrations using DMS"

cpj2195:
Reverts Azure/azure-cli#21231


yonzhan:
RDBMS

Issue #22605, opened 2022-05-25
[Packaging] Drop Mariner 1.0 support


jiasli:
**Describe the bug**

Since CBL Mariner 2.0 is now GA, Mariner 1.0 will be fully deprecated roughly by the **end of October**.

Azure CLI can drop Mariner 1.0 support in 2.41.0 which is the next breaking change window.

yonzhan:
Packaging

PR #24039, opened 2022-09-28
[Packaging] Drop Mariner 1.0 RPM package

jiasli:
**Description**
Close https://github.com/Azure/azure-cli/issues/22605

Since CBL Mariner 2.0 is now GA, Mariner 1.0 will be fully deprecated roughly by the **end of October**.

We stop building RPM for Mariner 1.0 in 2.41.0 which is a breaking change window.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -600,21 +600,15 @@ jobs:
     vmImage: 'ubuntu-20.04'
   strategy:
     matrix:
-      1.0:
-        image: cblmariner.azurecr.io/base/core
-        tag: 1.0
       2.0:
-        image: mcr.microsoft.com/cbl-mariner/base/core
-        tag: 2.0
+        image: mcr.microsoft.com/cbl-mariner/base/core:2.0
+        artifact: rpm-mariner2.0
   steps:
   - task: Bash@3
     displayName: 'Build Rpm Package: Mariner'
     inputs:
       targetType: 'filePath'
       filePath: scripts/release/rpm/pipeline_mariner.sh
-    env:
-      IMAGE: $(image)
-      TAG: $(tag)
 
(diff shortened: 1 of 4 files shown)


azure-pipelines.yml
```diff
@@ -600,21 +600,15 @@ jobs:
     vmImage: 'ubuntu-20.04'
   strategy:
     matrix:
-      1.0:
-        image: cblmariner.azurecr.io/base/core
-        tag: 1.0
       2.0:
-        image: mcr.microsoft.com/cbl-mariner/base/core
-        tag: 2.0
+        image: mcr.microsoft.com/cbl-mariner/base/core:2.0
```
jiasli, line 604: No need to have a separate variable `tag`. We can combine it with `image`.

azure-pipelines.yml
```diff
@@ -600,21 +600,15 @@ jobs:
...
+        artifact: rpm-mariner2.0
   steps:
   - task: Bash@3
     displayName: 'Build Rpm Package: Mariner'
     inputs:
       targetType: 'filePath'
       filePath: scripts/release/rpm/pipeline_mariner.sh
-    env:
-      IMAGE: $(image)
-      TAG: $(tag)
```
jiasli, line 617: https://learn.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#environment-variables

> System and user-defined variables also get injected as environment variables for your platform. When variables are turned into environment variables, variable names become uppercase, and periods turn into underscores. For example, the variable name `any.variable` becomes the variable name `$ANY_VARIABLE`.

So we don't need to explicitly declare env vars.

wangzelin007 (approved):

yonzhan:
Drop Mariner 1.0

Issue #23015, opened 2022-06-24
azure-cli-core: Test failures with conflicting subparsers in Python 3.11.0b3


major:
**Related command**
Attempting to run tests for `azure-cli-core`.

**Describe the bug**
While running `pytest` to test `azure-cli-core`, many tests fail in Python 3.11.0b3 due to conflicting subparsers. Here is an example:

```
__________________ TestParser.test_parser_error_spellchecker ___________________
[gw0] linux -- Python 3.11.0 /usr/bin/python3

self = <azure.cli.core.tests.test_parser.TestParser testMethod=test_parser_error_spellchecker>

    @mock.patch('importlib.import_module', _mock_import_lib)
    @mock.patch('pkgutil.iter_modules', _mock_iter_modules)
    @mock.patch('azure.cli.core.commands._load_command_loader', _mock_load_command_loader)
    @mock.patch('azure.cli.core.extension.get_extension_modname', _mock_extension_modname)
    @mock.patch('azure.cli.core.extension.get_extensions', _mock_get_extensions)
    def test_parser_error_spellchecker(self):
        cli = DummyCli()
        main_loader = MainCommandsLoader(cli)
        cli.loader = main_loader
    
        cli.loader.load_command_table(None)
    
        parser = cli.parser_cls(cli)
>       parser.load_command_table(cli.loader)

src/azure-cli-core/azure/cli/core/tests/test_parser.py:206: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
src/azure-cli-core/azure/cli/core/parser.py:100: in load_command_table
    command_parser = subparser.add_parser(command_verb,
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = _SubParsersAction(option_strings=[], dest='_subcommand', nargs='A...', const=None, default=None, type=None, choices={'module': 'module'}, required=True, help=None, metavar=None)
name = 'module'
kwargs = {'_command_source': 'azure.cli.core.tests.test_parser', 'cli_help': None, 'conflict_handler': 'error', 'description': None, ...}
aliases = ()

    def add_parser(self, name, **kwargs):
        # set prog from the existing prefix
        if kwargs.get('prog') is None:
[...]

PR #24109, opened 2022-10-05
{Core} Python 3.11: Remove workaround for argparse that breaks on Python 3.11

mdboom:
Fixes the crash inside of argparse when setting up subparsers as described here:

  https://github.com/Azure/azure-cli/issues/23015#issue-1283886050

**Related command**

All `az` commands.

**Description**

First step on the road to Python 3.11 compatibility.

**Testing Guide**

In a Python 3.11 environment, run `az` (or `azdev test`, though this doesn't fix all tests yet, just the vast majority).

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/parser.py
+++ b/src/azure-cli-core/azure/cli/core/parser.py
@@ -4,6 +4,7 @@
 # --------------------------------------------------------------------------------------------
 
 import difflib
+import sys
 
 import argparse
 import argcomplete
@@ -91,8 +92,10 @@ class AzCliCommandParser(CLICommandParser):
 
             command_verb = command_name.split()[-1]
             # To work around http://bugs.python.org/issue9253, we artificially add any new
-            # parsers we add to the "choices" section of the subparser.
-            subparser.choices[command_verb] = command_verb
+            # parsers we add to the "choices" section of the subparser. This workaround
+            # is no longer required in Python 3.11.
+            if sys.version_info < (3, 11):
+                subparser.choices[command_verb] = command_verb
 
             # inject command_module designer's help formatter -- default is HelpFormatter
             fc = metadata.formatter_class or argparse.HelpFormatter


major:
This patch fixes my issues on Fedora 37 + Rawhide (38) with 3.11.0rc2. 👏🏻

yonzhan:
Add @jiasli for awareness

jiasli:
Thanks a lot @mdboom for the contribution. Due to the heavy workload in this sprint, we will check this PR and make sure it is handled in the next sprint.

jacoborus:
> Thanks a lot @mdboom for the contribution. Due to the heavy workload in this sprint, we will check this PR and make sure it is handled in the next sprint.

Hi @jiasli , this problem will affect to every Microsoft Azure client using Fedora, RHEL, or Linux distro subscribed to the Python repos in the following days, as soon as they update their machines. Please, reconsider the move to the next sprint, otherwise, to downgrade python after finding this issue

jiasli:
> this problem will affect to every Microsoft Azure client using Fedora, RHEL, or Linux distro subscribed to the Python repos in the following days, as soon as they update their machines.

I am afraid this is not true. The RPM we built for RHEL has requires a fixed version of Python:

https://github.com/Azure/azure-cli/blob/c24f5fc2daa03cbac2c1269812bdd7684aa5a617/azure-pipelines.yml#L636-L645

Our Ubuntu DEB package bundles its own Python:

https://github.com/Azure/azure-cli/blob/a3aa348058cae300030f6c5b694c837974e9ab8c/scripts/release/debian/build.sh#L32

Also, even Python 3.10 is not available on RHEL yet. It usually takes long time for Linux distributions to support latest Python.

Previously, it usually took us 1~2 sprints (months) to support new Python versions.

> Please, reconsider the move to the next sprint, otherwise, all of these devs will not be able to work with Azure until they find this issue, and then downgrade Python in their machines

This sprint [Oct 2022 (2022-11-01)](https://github.com/Azure/azure-cli/milestone/123) will end very soon. Merging this PR without complete testing is too risky. 

Appreciate your understanding.

jacoborus:
> I am afraid this is not true. The RPM we built for RHEL has requires a fixed version of Python

Ok, let me rephrase it: this problem will affect to every Microsoft Azure client using Fedora, ~~RHEL~~, or any Linux distro subscribed to the Python repos in the following days.

> This sprint [Oct 2022 (2022-11-01)](https://github.com/Azure/azure-cli/milestone/123) will end very soon. Merging this PR without complete testing is too risky. Appreciate your understanding.

I'm sorry, but I don't understand how 2 lines of code that were already planned for this sprint since 3 weeks ago, are suddenly compromising it.

I guess it's because we're just Linux users, even if we pay for M$ Azure...

major:
@jacoborus I've added [this patch](https://src.fedoraproject.org/rpms/azure-cli/blob/rawhide/f/azure-cli.spec#_32) to the long list of patches/hacks for the Fedora azure-cli package and it seems to be working okay. The tests still need work, though.

jacoborus:
Thanks @major

jiasli:
> this problem will affect to every Microsoft Azure client using Fedora, RHEL, or any Linux distro subscribed to the Python repos in the following days.

Even though we are working on support for Fedora, Fedora is not one of the supported OS yet. Azure CLI package for Fedora is mainly maintained by its community.

> I'm sorry, but I don't understand how 2 lines of code that were already planned for this sprint since 3 weeks ago, are suddenly compromising it.

In Azure CLI's development process, there are far more tasks to do in order to support a new programming language version than adding 2 lines. We need to update CI/CD pipeline, thoroughly test all code path, commands and scenarios to ensure the product quality.

Meanwhile, Azure DevOps pipeline doesn't have Python 3.11 available right now:

- https://learn.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#software
- https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#python

image

which is tracked by

- https://github.com/actions/runner-images/issues/6459

So we currently are unable to test our code in our CI/CD pipeline. We will surely continue when the Azure DevOps pipeline is updated to support Python 3.11.

We are working very hard to make Azure CLI a great product, and we truly appreciate your kind understanding.

jiasli:
I did more investigation and below are my findings.

Python 3.11 updated `argparse`'s behavior:

https://docs.python.org/3.11/whatsnew/changelog.html#id50

> bpo-39716: Raise an ArgumentError when the same subparser name is added twice to an `argparse.ArgumentParser`. This is consistent with the (default) behavior when the same option string is added twice to an ArgumentParser.

The PR for this change is https://github.com/python/cpython/pull/18605

Per Git blame,

- `subparser.choices[command_verb] = command_verb` was renamed from `subparser.choices[command_name] = command_name` by https://github.com/Azure/azure-cli/commit/517872af1912ba7a2cd251999e7675546fbd495f
- `subparser.choices[command_name] = command_name` was introduced as early as https://github.com/Azure/azure-cli/commit/2a6f545aee7b29afd90e01af28356507f0f0f606

Running this line makes `subparser.choices` a `str:str` mapping.

In `argparse._SubParsersAction`, `self.choices` is the same `dict` object as `self._name_parser_map`:

https://github.com/python/cpython/blob/v3.10.8/Lib/argparse.py#L1159

```py
        super(_SubParsersAction, self).__init__(
            ...
            choices=self._name_parser_map,
```

Later, in `argparse._SubParsersAction.add_parser`, `self._name_parser_map`'s `command_verb` key is overwritten as a `str:parser` mapping:

https://github.com/python/cpython/blob/v3.10.8/Lib/argparse.py#L1179

```py
self._name_parser_map[name] = parser
```

Before running this line:

image

After running this line:

image

So `subparser.choices[command_verb] = command_verb` is basically a no-op, but it starts to cause failure in Python 3.11 due to the new duplication check.

I am not sure why `subparser.choices[command_verb] = command_verb` was introduced by https://github.com/Azure/azure-cli/commit/2a6f545aee7b29afd90e01af28356507f0f0f606 6 years ago, possibly due to some Python 2 `argparse`'s limitation?

src/azure-cli-core/azure/cli/core/parser.py
```diff
@@ -91,8 +92,10 @@ def load_command_table(self, command_loader):
 
             command_verb = command_name.split()[-1]
             # To work around http://bugs.python.org/issue9253, we artificially add any new
```
jiasli, line 94: I checked http://bugs.python.org/issue9253 mentioned in the comment, but didn't see how it is related to this workaround.

jiasli:
We can think it in another way. Since Python 3.11's `argparse` has no functionality change (only an additional check), if this line should removed for Python 3.11, it should also be removed for Python 3.10-.

jacoborus:
> In Azure CLI's development process, there are far more tasks to....

@jiasli I'm very aware of how the development process works. And still believe that a bug that potentially can waste thousands of hours of your clients worth the time. Even more when it's a bug that was there from previous versions, Python 3.11 just unveiled it

peruzzof:
Just to comment that I am being affected by this error, I have a pipeline with this task (running in microsoft hosted agents):
- task: UsePythonVersion@0
  inputs:
    versionSpec: '3.x'

The workaround was to change from 3.x to 3.10, so I am assuming that Python 3.11 is already supported by Microsoft and a fix is urgent.

There is any other way to prioritize this fix?

PR #24110, opened 2022-10-05
{CosmosDB} Make sure analytical-storage-ttl is applied as only argument for Gremlin graphs

armaansood:
**Active customer bug**
When creating a Cosmos DB Gremlin graph, the parameter 'analytical-storage-ttl' currently isn't added to the HTTP payload if it is the only container-specific parameter. This PR fixes the issue.


yonzhan:
CosmosDB

evelyn-ys (approved):

PR #24111, opened 2022-10-05
{Core} Python 3.11: Remove uses of deprecated inspect.getargspec

mdboom:
inspect.getargspec has been deprecated since 3.0 in favor of its replacement inspect.getfullargspec.

This does change "vendored" code, and I couldn't find where that may require special treatment, if at all.

There is [one other use of `inspect.getargspec`](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/azure/cli/core/util.py#L668) that is retained for backward compatibility with Python 2 that I didn't touch, but in reality that clause could also probably be removed since Python 2 is no longer supported.

**Related command**

All `az` commands.

**Description**

Another step on the road to Python 3.11 compatibility.  (See also #24109)

**Testing Guide**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/keyvault/vendored_sdks/azure_keyvault_t1/key_vault_authentication.py
+++ b/src/azure-cli/azure/cli/command_modules/keyvault/vendored_sdks/azure_keyvault_t1/key_vault_authentication.py
@@ -59,7 +59,7 @@ class KeyVaultAuthBase(AuthBase):
     # for backwards compatibility we need to support callbacks which don't accept the scheme
     def _auth_callback_compat(self, server, resource, scope, scheme):
         return self._user_callback(server, resource, scope) \
-            if len(inspect.getargspec(self._user_callback).args) == 3 \
+            if len(inspect.getfullargspec(self._user_callback).args) == 3 \
             else self._user_callback(server, resource, scope, scheme)
 
     def __call__(self, request):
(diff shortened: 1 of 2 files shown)


yonzhan:
remove deprecated dependencies on test

mdboom:
Speaking to @iritkatriel in another forum, it sounds like the most future-proof option is to use `inspect.signature`.  I'm going to update this PR to use that instead.

mdboom:
> remove deprecated dependencies on test

Sorry, is this a request of me? I'm not sure I understand.

src/azure-cli/azure/cli/command_modules/keyvault/vendored_sdks/azure_keyvault_t1/key_vault_authentication.py
```diff
@@ -59,7 +59,7 @@ def __init__(self, authorization_callback):
     # for backwards compatibility we need to support callbacks which don't accept the scheme
     def _auth_callback_compat(self, server, resource, scope, scheme):
         return self._user_callback(server, resource, scope) \
-            if len(inspect.getargspec(self._user_callback).args) == 3 \
+            if len(inspect.signature(self._user_callback).parameters) == 3 \
```
jiasli, line 62: This is not equivalent to the original code if `self._user_callback` contains `*args` and `**kwargs`, since [`signature`](https://docs.python.org/3.10/library/inspect.html#inspect.signature)'s result includes `*args` (`kind`: `VAR_POSITIONAL`) and `**kwargs` (`kind`: `VAR_KEYWORD`) in `parameters`, but [`getargspec`](https://docs.python.org/3.10/library/inspect.html#inspect.getargspec)'s result has dedicated `varargs`, `keywords` elements for them:

```py
import inspect

def testfunc(a=1, b=2, *args, **kwargs):
    pass

print(len(inspect.getfullargspec(testfunc).args))
print(inspect.getfullargspec(testfunc).args)
print(len(inspect.signature(testfunc).parameters))
print(inspect.signature(testfunc).parameters)
```

Output:

```
2
['a', 'b']
4
OrderedDict([('a', <Parameter "a=1">), ('b', <Parameter "b=2">), ('args', <Parameter "*args">), ('kwargs', <Parameter "**kwargs">)])
```

PR #24126, opened 2022-10-07
[Backup] `az backup restore restore-disks`: Update Cross Zonal Restore behaviour for ZRS vaults and primary region CRR scenarios

zubairabid:
- Adds use_secondary_region to czr validation call
- Allows CZR for all scenarios in ZRS.
- Disallows CZR for zone-pinned primary region CRR restore.

**Related command**

`az backup restore restore-disks`

**Description**

Cross Zonal Restore is now allowed on all ZRS vaults regardless of zone-pinning, and is disallowed for all primary region CRR restores (regardless of zone pinning). Update on CLI end to reflect the same.

**Testing Guide**

- Cross zonal restore should work for
  - Non-zone pinned and zone-pinned VMs in a ZRS vault
  - Zone-pinned VMs in a Cross-region restore, on restoring to the secondary region (whether zones are supported in the secondary region or not - in the case of not, the zone defaults to NoZone)
- Cross zonal restore will not work in any other scenario.

The general command in all cases is `az backup restore restore-disks ...`, where for Cross-zonal restore we add the `--target-zone 1/2/3` parameter, and for cross region restore we use the `--use-secondary-region` in a vault where CRR is enabled.

**History Notes**

`az backup restore restore-disks`: The `--target-zone` argument can now be used with non zone pinned ZRS vaults, and is not allowed for CRR vaults in primary region restore scenario with zone pinned vaults either.

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/backup/_validators.py
+++ b/src/azure-cli/azure/cli/command_modules/backup/_validators.py
@@ -98,15 +98,22 @@ def validate_crr(target_rg_id, rehydration_priority):
         raise MutuallyExclusiveArgumentError("Archive restore isn't supported for secondary region.")
 
 
-def validate_czr(backup_config_response, recovery_point):
+def validate_czr(backup_config_response, recovery_point, use_secondary_region):
     backup_storage_redundancy = backup_config_response.properties.storage_type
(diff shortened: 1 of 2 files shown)


yonzhan:
Backup

zhoxing-ms (approved):

PR #24176, opened 2022-10-12
{AKS} Remove rouge chracters from DCR name and match Prom addon format

bragi92:
**Related command**

**Description**

Remove rouge chracters from DCR name and match Prom addon format

**Testing Guide**

az aks enable-addons -a monitoring -g kaveeshfinalcli2 -n ka_vee_sh_eus--testchars --enable-msi-auth-for-monitoring

**History Notes**

---


yonzhan:
AKS

src/azure-cli/azure/cli/command_modules/acs/addonconfiguration.py
```diff
@@ -268,6 +268,16 @@ def get_existing_container_insights_extension_dcr_tags(cmd, dcr_url):
     return tags
 
 
+def sanitize_name(name):
```
FumingZhang, line 271: Could you please add some unit test cases for this function?

Besides, it seems that this would remove trailing non-alphanumeric characters, but still retains non-alphanumeric characters in the middle part of the name. Specific to the current usage, the modification target is mainly the cluster name, and according to the [restrictions](https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.AKS.Name/), only underscores and hyphens are allowed except for alphanumerics. So, is this expected?

bragi92: Added unit tests. @FumingZhang since its a new file, will I need to add it somewhere for it to be included or is it automatically run in the pipeline?

I've updated the function name to reflect the scenario better too. This is only applicable to the DCR names which have different limitations and based on our naming convention of 'MSPROM-{region_code}-{cluster_name}' and a limit of 44 characters at max in lenght. This is to make sure that we do not accidently end up with any hyphen or underscore at the end (which are allowed in cluster names)

FumingZhang (approved):
LGTM

zhoxing-ms (approved):

PR #24207, opened 2022-10-14
[Backup] Updated subtask duration handling in `backup job show` commands

zubairabid:
**Related command**

`az backup job show`

**Description**

The backup job command returns the details of the main task, along with any subtasks that were run. So far, we only received the details of task start and end times, and duration, for the main task. Now in certain scenarios the subtasks contain this information as well (where they used to be `null` in the case of start and end times, and `0` in the case of duration). As there are no explicit flags of when the subtask duration will be returned, the service returns a value instead of `null` for the start and end times - when the task has not started or has not completed, the value will be a minimum value. We do not want to show customers the minimum value, and during such cases it will be replaced with `null` again when displaying the output.

**Testing Guide**

Run the `az backup job show -n <task name> -v <vault name> -g <resource group>` command for an ongoing backup job task in the `eastuseuap` region.

Without the change, for an ongoing (sub)task, the end time will be `0001-01-01T00:00:00`. With the change, the end time will be `null`. For a subtask that hasn't started, both start and end time will go from `0001-01-01T00:00:00` to `null`.

**History Notes**

[Component Name 1] `az backup job show`: Change subtask start/end time from minimum value to null for ongoing or yet-to-start operation.

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/backup/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/backup/custom.py
@@ -1290,18 +1290,10 @@ def show_job(cmd, client, resource_group_name, vault_name, name, use_secondary_r
         vault_location = vault.location
         azure_region = secondary_region_map[vault_location]
         client = backup_crr_job_details_cf(cmd.cli_ctx)
-        return client.get(azure_region, CrrJobRequest(resource_id=vault.id, job_name=name))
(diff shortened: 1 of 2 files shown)


yonzhan:
Backup

zhoxing-ms:
Could you add some tests for this case?

zubairabid:
> Could you add some tests for this case?

Have added a test case for this change.

zhoxing-ms (approved):

Issue #24213, opened 2022-10-14
`az interactive` fails to start: ModuleNotFoundError: No module named 'prompt_toolkit.interface'


dagnello:
## Describe the bug

**Command Name**
`az interactive`

**Errors:**
```
The command failed with an unexpected error. Here is the traceback:
No module named 'prompt_toolkit.interface'
Traceback (most recent call last):
  File "/opt/homebrew/Cellar/azure-cli/2.41.0/libexec/lib/python3.10/site-packages/knack/cli.py", line 233, in invoke
    cmd_result = self.invocation.execute(args)
  File "/opt/homebrew/Cellar/azure-cli/2.41.0/libexec/lib/python3.10/site-packages/azure/cli/core/commands/__init__.py", line 663, in execute
    raise ex
  File "/opt/homebrew/Cellar/azure-cli/2.41.0/libexec/lib/python3.10/site-packages/azure/cli/core/commands/__init__.py", line 726, in _run_jobs_serially
    results.append(self._run_job(expanded_arg, cmd_copy))
  File "/opt/homebrew/Cellar/azure-cli/2.41.0/libexec/lib/python3.10/site-packages/azure/cli/core/commands/__init__.py", line 697, in _run_job
    result = cmd_copy(params)
  File "/opt/homebrew/Cellar/azure-cli/2.41.0/libexec/lib/python3.10/site-packages/azure/cli/core/commands/__init__.py", line 333, in __call__
    return self.handler(*args, **kwargs)
  File "/opt/homebrew/Cellar/azure-cli/2.41.0/libexec/lib/python3.10/site-packages/azure/cli/core/commands/command_operation.py", line 121, in handler
    return op(**command_args)
  File "/opt/homebrew/Cellar/azure-cli/2.41.0/libexec/lib/python3.10/site-packages/azure/cli/command_modules/interactive/custom.py", line 41, in start_shell
    azext_interactive.start_shell(cmd, style=style)
  File "/Users/davide.agnello/.azure/cliextensions/interactive/azext_interactive/__init__.py", line 10, in start_shell
    from .azclishell.app import AzInteractiveShell
  File "/Users/davide.agnello/.azure/cliextensions/interactive/azext_interactive/azclishell/app.py", line 33, in <module>
    from prompt_toolkit.interface import Application, CommandLineInterface
ModuleNotFoundError: No module named 'prompt_toolkit.interface'

```

## To Reproduce:
[...]

PR #24252, opened 2022-10-18
[Core] Prioritize extension's modle path

taoyama:
**Related command**
az interactive

**Description**
With current implementation, module path for extension is appended by sys.append(). Some module like interactive need old module so we need to prioritize to import modules under extension's directory instead of default directory. This patch will fix problem like #24213

**Testing Guide**
az interactive

---


yonzhan:
interactive

jiasli:
Thanks for the contribution. However, this change was previously proposed by #12778 was rejected due to security concerns.

taoyama:
@jiasli 
Thank you for your reply.
I understand your point but why would we allow loading modules for extension, in the first place?
I think that is the biggest security concern here...isn't it?

Issue #497, opened 2016-07-05
network: 'subnet create' should take NSG 


yugangw-msft:
Xplat and SDK both support it

PR #24279, opened 2022-10-19
[ACR] `acr task update`: Fix logic issue for updating encoded task

wju-MSFT:
This PR changes the logic for task update. It resolves the following issues:

1. The update process missed the `EncodedTaskStepUpdateParameters` to update the values for `EncodedTask`. Issue [#497](https://github.com/Azure/acr/issues/497).

2. If our update cmd has a file arg, then CLI makes the assumption that the step is a FileTaskStep or DuckerBuildTaskStep regardless the existing or intended context [here](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/acr/task.py#L414). If we want to update the file for an `EncodedTask`, we will get an error in [#313](https://github.com/Azure/acr/issues/313). The reason is the following: 
In `TasksController`, we will first update the `exisitingTaskRecord` with all of the updated parameters [here](https://msazure.visualstudio.com/AzureContainerRegistry/_git/Build?path=/src/ACR.BuildRP/src/v18_09_GA/Controllers/TasksController.cs&version=GBmain&line=502&lineEnd=503&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents). As described above, the `EncodedTaskStep` now has been changed to `FileTaskStep`. Then we will call `ValidateContextProperty()` with the updated `FileTaskStep`. If the step were still `EncodedTaskStep`, there should have been no validation for context at all, because of [this](https://msazure.visualstudio.com/AzureContainerRegistry/_git/Build?path=/src/ACR.BuildRP/src/v18_09_GA/Models/Validation.Tasks.cs&version=GBmain&line=531&lineEnd=532&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents) logic. But with the unintended change of task type (from Encoded to File), the validation would fail. 

After the change, the update logics follows the create logics.

To test: run
`az acr task create -r wjucr001 -n task123 --cmd localhost:12900/test:latest -c /dev/null --schedule \*/3\ \*\ \*\ \*\ \* `
`az acr task update -r wjucr001 -n task123 --set foo=bar `
Observe the value is changed.

`az acr task create -r wjucr001 -n task1 -f acb.yaml -c /dev/null`
`acr task update -r wjucr001 -n task1 -f acb.yaml`
Observe no error thrown.

Signed-off-by: Jun <wju@microsoft.com>

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acr/task.py
+++ b/src/azure-cli/azure/cli/command_modules/acr/task.py
@@ -405,6 +405,9 @@ def acr_task_update(cmd,  # pylint: disable=too-many-locals, too-many-statements
     step = task.step
     branch = None
 
+    # If no context is given, use the existing context.
+    # If given context is /dev/null, set it to None.
+    # Otherwise, use the given context.
     if context_path is None:
         context_path = step.context_path
     elif context_path.lower() == ACR_NULL_CONTEXT:
@@ -414,62 +417,23 @@ def acr_task_update(cmd,  # pylint: disable=too-many-locals, too-many-statements
 
     if git_access_token is None:
         git_access_token = step.context_access_token
-    arguments = _get_all_override_arguments(arg, secret_arg)
-    set_values = _get_all_override_arguments(set_value, set_secret)
 
-    if context_path is None:
-        yaml_template = get_yaml_template(
-            cmd_value, timeout, file)
-        import base64
-        step = EncodedTaskStepUpdateParameters(
-            encoded_task_content=base64.b64encode(
-                yaml_template.encode()).decode(),
-            context_path=context_path,
-            context_access_token=git_access_token,
-            values=(set_value if set_value else []) + (set_secret if set_secret else [])
-        )
-    else:
-        if file:
-            if file.endswith(ALLOWED_TASK_FILE_TYPES):
-                step = FileTaskStepUpdateParameters(
-                    task_file_path=file,
-                    values_file_path=values,
-                    context_path=context_path,
-                    context_access_token=git_access_token,
-                    values=(set_value if set_value else []) + (set_secret if set_secret else [])
-                )
-            else:
-                step = DockerBuildStepUpdateParameters(
-                    image_names=image_names,
-                    is_push_enabled=not no_push,
-                    no_cache=no_cache,
-                    docker_file_path=file,
-                    arguments=(arg if arg else []) + (secret_arg if secret_arg else []),
-                    context_path=context_path,
-                    context_access_token=git_access_token,
-                    target=target
-                )
-        elif step:
-            if hasattr(step, 'docker_file_path'):
-                step = DockerBuildStepUpdateParameters(
-                    image_names=image_names,
-                    is_push_enabled=not no_push,
-                    no_cache=no_cache,
-                    docker_file_path=file,
-                    arguments=arguments,
-                    context_path=context_path,
-                    context_access_token=git_access_token,
-                    target=target
-                )
-
-            elif hasattr(step, 'task_file_path'):
-                step = FileTaskStepUpdateParameters(
-                    task_file_path=file,
-                    values_file_path=values,
-                    context_path=context_path,
-                    context_access_token=git_access_token,
-                    values=set_values
-                )
+    step = create_task_step(
+        context_path=context_path,
+        cmd=cmd,
+        file=file,
+        image_names=image_names,
+        values=values,
+        git_access_token=git_access_token,
+        set_value=set_value,
+        set_secret=set_secret,
+        no_push=no_push,
+        no_cache=no_cache,
+        arg=arg,
+        secret_arg=secret_arg,
+        target=target,
+        cmd_value=cmd_value,
+        timeout=timeout)
 
     source_control_type = None
     if context_path:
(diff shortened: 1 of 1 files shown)


yonzhan:
ACR

src/azure-cli/azure/cli/command_modules/acr/task.py
```diff
@@ -401,8 +404,11 @@ def acr_task_update(cmd,  # pylint: disable=too-many-locals, too-many-statements
 
     step = task.step
     branch = None
+
     if context_path is None:
```
huanwu, line 408: Could you add some comments here for the logic?

src/azure-cli/azure/cli/command_modules/acr/task.py
```diff
@@ -401,8 +404,11 @@ def acr_task_update(cmd,  # pylint: disable=too-many-locals, too-many-statements
...
         context_path = step.context_path
```
huanwu, line 409: Is it possible the  step.context_path = '/dev/null'? Should we set context_path to None as well?

src/azure-cli/azure/cli/command_modules/acr/task.py
```diff
@@ -411,46 +417,59 @@ def acr_task_update(cmd,  # pylint: disable=too-many-locals, too-many-statements
...
-            values=set_values
-        )
-    elif file and not file.endswith(ALLOWED_TASK_FILE_TYPES):
-        step = DockerBuildStepUpdateParameters(
-            image_names=image_names,
-            is_push_enabled=not no_push,
-            no_cache=no_cache,
-            docker_file_path=file,
-            arguments=arguments,
+    if context_path is None:
```
huanwu, line 420: Could you wrap the code as a separate method like create_task_step?

src/azure-cli/azure/cli/command_modules/acr/task.py
```diff
@@ -411,46 +417,59 @@ def acr_task_update(cmd,  # pylint: disable=too-many-locals, too-many-statements
...
-                image_names=image_names,
-                is_push_enabled=not no_push,
-                no_cache=no_cache,
-                docker_file_path=file,
-                arguments=arguments,
-                context_path=context_path,
-                context_access_token=git_access_token,
-                target=target
-            )
+    else:
```
huanwu, line 431: In create_task_step, this section code is above the one for encoded task. Could we follow that style? Just make sure the code looks consistent for readability.

src/azure-cli/azure/cli/command_modules/acr/task.py
```diff
@@ -411,46 +417,59 @@ def acr_task_update(cmd,  # pylint: disable=too-many-locals, too-many-statements
...
+        if file:
```
huanwu, line 432: Question, if file is null here, do we need read file from existing task and see if it is null?

src/azure-cli/azure/cli/command_modules/acr/task.py
```diff
@@ -411,46 +417,59 @@ def acr_task_update(cmd,  # pylint: disable=too-many-locals, too-many-statements
...
-    elif file and not file.endswith(ALLOWED_TASK_FILE_TYPES):
-        step = DockerBuildStepUpdateParameters(
-            image_names=image_names,
-            is_push_enabled=not no_push,
-            no_cache=no_cache,
-            docker_file_path=file,
-            arguments=arguments,
+    if context_path is None:
+        yaml_template = get_yaml_template(
+            cmd_value, timeout, file)
```
huanwu, line 422: Question: Is it possible the cmd_value here is null?

src/azure-cli/azure/cli/command_modules/acr/task.py
wju-MSFT: done

src/azure-cli/azure/cli/command_modules/acr/task.py
wju-MSFT: Is it possible the step.context_path = '/dev/null'? 
No, in create: `if context_path.lower() == ACR_NULL_CONTEXT:
        context_path = None` so if it were set to /dev/null during creation, it will be converted to None.

Should we set context_path to None as well?
Yes, given context_path is checked following the same way as create.

src/azure-cli/azure/cli/command_modules/acr/task.py
wju-MSFT: Done.

src/azure-cli/azure/cli/command_modules/acr/task.py
wju-MSFT: It is possible. Then should we use the existing ones? Raised in flux channel for inputs. Will update this section later.

src/azure-cli/azure/cli/command_modules/acr/task.py
```diff
@@ -401,56 +400,41 @@ def acr_task_update(cmd,  # pylint: disable=too-many-locals, too-many-statements
...
+    elif context_path.lower() == ACR_NULL_CONTEXT:
+        context_path = None
     else:
         branch = _get_branch_name(context_path)
 
+    # When update encoded task, existing encoded_task_content in the step cannot be decoded for update.
+    if context_path is None:
+        if cmd_value is not None or timeout is not None or file is not None:
+            raise CLIError(
+                "cannot update the task with change of cmd or timeout or file, please use create instead")
```
huanwu, line 418: We cannot update cmd or timeout or file only if it is encoded task, right?

src/azure-cli/azure/cli/command_modules/acr/task.py
```diff
@@ -401,56 +400,41 @@ def acr_task_update(cmd,  # pylint: disable=too-many-locals, too-many-statements
...
     if context_path is None:
         context_path = step.context_path
+    elif context_path.lower() == ACR_NULL_CONTEXT:
+        context_path = None
     else:
         branch = _get_branch_name(context_path)
 
+    # When update encoded task, existing encoded_task_content in the step cannot be decoded for update.
+    if context_path is None:
+        if cmd_value is not None or timeout is not None or file is not None:
```
huanwu, line 416: I am thinking if customer provide all three of them, we can call get_yaml_template to regenerate the encodedtaskcontent. The logic is similar as create here. If they didn't provide all three, we can just return the error.

wju-MSFT: actually, from what `get_yaml_template` method, we only need cmd_value or file to generate a new yaml_tamplate. Only time we throw an error is when user only wants to update the timeout, which we cannot do because we cannot patch the encoded_task_content, and this is where we are throwing CLIError now.

huanwu (approved):

zhoxing-ms:
@wju-MSFT Could you add some tests for those change?

src/azure-cli/azure/cli/command_modules/acr/task.py
```diff
@@ -401,56 +400,52 @@ def acr_task_update(cmd,  # pylint: disable=too-many-locals, too-many-statements
...
+        if cmd_value is not None or file is not None:
+            yaml_template = get_yaml_template(
+                cmd_value, timeout, file)
+            import base64
+            encoded_task_content = base64.b64encode(
+                yaml_template.encode()).decode()
+        # When only timeout is given,
+        # encoded_task_content cannot be patched
+        elif timeout is not None:
+            raise CLIError(
```
zhoxing-ms, line 427: Could you please use the specific error type `IncorrectUsageError` instead of `CLIError`?

zhoxing-ms:
@wju-MSFT Since we will launch the release of this sprint tomorrow, please resolve all the comments before tomorrow, otherwise this PR will not be able to catch up with the release of this sprint, and it can only be postponed to the next sprint **(12-06)**

wju-MSFT:
> @wju-MSFT Since we will launch the release of this sprint tomorrow, please resolve all the comments before tomorrow, otherwise this PR will not be able to catch up with the release of this sprint, and it can only be postponed to the next sprint **(12-06)**

Will do.

src/azure-cli/azure/cli/command_modules/acr/task.py
wju-MSFT: done

wju-MSFT:
> @wju-MSFT Could you add some tests for those change?

Since we are trying to catch up the sprint on Oct 27, I will add unit tests on another PR. Will also create a new WI for unit tests. @zhoxing-ms

zhoxing-ms (approved):

Issue #11386, opened 2019-11-25
Can't get access token `az account get-access-token` Azure API for FHIR


mallecespedes:
## Summary
When trying to obtain a token for the Azure API for FHIR using Azure CLI the following error appears:

**Command Name**
`az account get-access-token`

**Errors:**
```
400 Client Error: Bad Request for url: http://localhost:50342/oauth2/token
Traceback (most recent call last):
python3.6/site-packages/knack/cli.py, ln 206, in invoke
    cmd_result = self.invocation.execute(args)
cli/core/commands/__init__.py, ln 603, in execute
    raise ex
cli/core/commands/__init__.py, ln 661, in _run_jobs_serially
    results.append(self._run_job(expanded_arg, cmd_copy))
...
python3.6/site-packages/msrestazure/azure_active_directory.py, ln 486, in get_msi_token
    result.raise_for_status()
python3.6/site-packages/requests/models.py, ln 940, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: http://localhost:50342/oauth2/token
```

## To Reproduce:
Open Azure CLI from Azure Portal or sign in with [https://shell.azure.com](url)
Execute following command in CLI:
`token=$(az account get-access-token --resource=https://azurehealthcareapis.com | jq -r .accessToken)`

## Expected Behavior
Save access token in the variable named token for further use. The next step according to the documentation [https://docs.microsoft.com/en-us/azure/healthcare-apis/get-healthcare-apis-access-token-cli](url) is:
`curl -X GET --header "Authorization: Bearer $token" https://<FHIR ACCOUNT NAME>.azurehealthcareapis.com/Patient`

## Environment Summary
```
Linux-4.15.0-1063-azure-x86_64-with-debian-stretch-sid
Python 3.6.5
Shell: bash
azure-cli 2.0.77
Extensions:
healthcareapis 0.1.2
```

haroldrandom:
Thansk for the feedback!  Please @jiasli have a look.

mallecespedes:
> Thansk for the feedback! Please @jiasli have a look.

Is there an update for this issue?

jiasli:
This is the design of Cloud Shell that its MSI restricts `--resource` to only a few well-known ones. 

As a workaround, in Cloud Shell, you may run `az login` to manually log in and run `az account get-access-token --resource=https://azurehealthcareapis.com` again. 

[...]

PR #24332, opened 2022-10-21
[Reference] freshness project: Profile (`az account`, `az login`)

dbradish-microsoft:
This PR is part of the reference refresh project.  This PR focuses on `src/azure-cli/azure/cli/command_modules/profile/_help.py` and any related file containing summaries or examples for these same command groups.

The updates in this PR are intended to freshen the following reference areas:
1. short-summary
2. long-summary
3. examples

**Note:** The testing of all examples is out of scope for this project and is part of the **Content Quality Automation** crew project.

This is work item https://dev.azure.com/mseng/TechnicalContent/_workitems/edit/1998772.


yonzhan:
@jiasli for awareness

src/azure-cli/azure/cli/command_modules/profile/_help.py
```diff
@@ -95,9 +138,23 @@
...
+    - name: Get an access token for a particular scope
+      text: >
+        az account get-access-token --scope NEED EXAMPLE
+        az account get-access-token --scope 
 """
 
 helps['self-test'] = """
 type: command
 short-summary: Runs a self-test of the CLI.
+long-summary: NEED MORE INFORMATION HERE
```
jiasli, line 159: `az self-test` is a hidden command marked as documented. Customers wouldn't see it.

src/azure-cli/azure/cli/command_modules/profile/_help.py
```diff
@@ -9,42 +9,64 @@
...
       text: >
-        az login -u johndoe@contoso.com -p VerySecret
-    - name: Log in with a service principal using client secret. Use -p=secret if the first character of the password is '-'.
+        az login --user johndoe@contoso.com --password VerySecret
+    - name: > 
+        Log in with a password containing a `-` as the first character.    
       text: >
-        az login --service-principal -u http://azure-cli-2016-08-05-14-31-15 -p VerySecret --tenant contoso.onmicrosoft.com
-    - name: Log in with a service principal using client certificate.
+        az login --user johndoe@contoso.com --password secret -myPasswordThatStartsWithAdash
```
jiasli, line 25:
```suggestion
        az login --user johndoe@contoso.com --password=-myPasswordThatStartsWithAdash
```

src/azure-cli/azure/cli/command_modules/profile/_help.py
```diff
@@ -95,9 +138,23 @@
...
+        see `https://learn.microsoft.com/en-us/azure/virtual-network/virtual-network-service-endpoints-overview`.  
+        If you receive a `Failed to connect to MSI...` error, WHAT IS REALLY THE PROBLEM?  Is your resourceID wrong?  Does your resource not allow access tokens?
+      text: >
+        az account get-access-token --resource myResourceID
+        az account get-access-token --resource https://database.windows.net --output tsv
+        az account get-access-token --resource https://management.core.windows.net/ --query accessToken --output tsv
+        
+    - name: Get an access token for a particular scope
+      text: >
+        az account get-access-token --scope NEED EXAMPLE
```
jiasli, line 152:
```suggestion
        az account get-access-token --scope https://management.azure.com//.default
```

src/azure-cli/azure/cli/command_modules/profile/_help.py
```diff
@@ -9,42 +9,64 @@
...
-short-summary: Log in to Azure.
+short-summary: Log into Azure.
 examples:
     - name: Log in interactively.
       text: >
         az login
-    - name: Log in with user name and password. This doesn't work with Microsoft accounts or accounts that have two-factor authentication enabled. Use -p=secret if the first character of the password is '-'.
+    - name: > 
+        Log in with user name and password. This doesn't work with Microsoft accounts or accounts that have two-factor authentication enabled. 
+        Use `-p=secret` if the first character of the password is `-`.
```
jiasli, line 19: We can remove this line if this is mentioned in following examples.

src/azure-cli/azure/cli/command_modules/profile/_help.py
```diff
@@ -9,42 +9,64 @@
...
+long-summary: >
+  For an in-depth overview of working with subscriptions using the Azure CLI, see
+  `https://learn.microsoft.com/en-us/cli/azure/manage-azure-subscriptions-azure-cli`
 """
 
 helps['account clear'] = """
 type: command
 short-summary: Clear all subscriptions from the CLI's local cache.
-long-summary: To clear the current subscription, use 'az logout'.
+long-summary: To clear the current subscription, use `az logout`.
```
jiasli, line 51: Good change! Using backticks is exactly the convention to mark commands.

src/azure-cli/azure/cli/command_modules/profile/_help.py
```diff
@@ -9,42 +9,64 @@
...
-        az login --identity -u /subscriptions/<subscriptionId>/resourcegroups/myRG/providers/Microsoft.ManagedIdentity/userAssignedIdentities/myID
+        az login --identity --user /subscriptions/<subscriptionId>/resourcegroups/myRG/providers/Microsoft.ManagedIdentity/userAssignedIdentities/myID
 """
 
 helps['account'] = """
 type: group
 short-summary: Manage Azure subscription information.
+long-summary: >
+  For an in-depth overview of working with subscriptions using the Azure CLI, see
+  `https://learn.microsoft.com/en-us/cli/azure/manage-azure-subscriptions-azure-cli`
```
jiasli, line 45: We should leave hyperlinks as is without `` ` ``, so that it is clickable in the online help.

src/azure-cli/azure/cli/command_modules/profile/_help.py
```diff
@@ -9,42 +9,64 @@
 
 helps['login'] = """
 type: command
-short-summary: Log in to Azure.
+short-summary: Log into Azure.
```
jiasli, line 12: I think "log in to" has been the convention in Azure CLI for some time, but I am not a native English speaker and I am definitely open to it.

src/azure-cli/azure/cli/command_modules/profile/_help.py
```diff
@@ -9,42 +9,64 @@
...
-    - name: Log in with user name and password. This doesn't work with Microsoft accounts or accounts that have two-factor authentication enabled. Use -p=secret if the first character of the password is '-'.
+    - name: > 
+        Log in with user name and password. This doesn't work with Microsoft accounts or accounts that have two-factor authentication enabled. 
+        Use `-p=secret` if the first character of the password is `-`.
       text: >
-        az login -u johndoe@contoso.com -p VerySecret
-    - name: Log in with a service principal using client secret. Use -p=secret if the first character of the password is '-'.
+        az login --user johndoe@contoso.com --password VerySecret
+    - name: > 
+        Log in with a password containing a `-` as the first character.
```
jiasli, line 23: I don't think we need to add an additional line break `>` here.

src/azure-cli/azure/cli/command_modules/profile/_help.py
```diff
@@ -55,33 +77,54 @@
...
+    - name: >
+        Get information about a subscription using `query`.  
+        To see available query options, first run `az account show --output json`.
+      text: >
+        az account show --query name
+        az account show --query user.name
+        az account show --query tenantId
+    - name: Return results in plain text.
+      text: >
+        az account show --query tenantId -o tsv
```
jiasli, line 103: Personally, I prefer not to include `--query` examples specific to a command 
(in this case `az account show`) as we already have a general documment for `--query`: https://learn.microsoft.com/en-us/cli/azure/query-azure-cli?tabs=concepts%2Cbash

src/azure-cli/azure/cli/command_modules/profile/_help.py
```diff
@@ -9,42 +9,64 @@
...
       text: >
-        az login --identity -u /subscriptions/<subscriptionId>/resourcegroups/myRG/providers/Microsoft.ManagedIdentity/userAssignedIdentities/myID
+        az login --identity --user /subscriptions/<subscriptionId>/resourcegroups/myRG/providers/Microsoft.ManagedIdentity/userAssignedIdentities/myID
 """
 
 helps['account'] = """
 type: group
 short-summary: Manage Azure subscription information.
+long-summary: >
+  For an in-depth overview of working with subscriptions using the Azure CLI, see
```
jiasli, line 44: We use 4 spaces as indent.

src/azure-cli/azure/cli/command_modules/profile/_help.py
```diff
@@ -55,33 +77,54 @@
 helps['account show'] = """
 type: command
 short-summary: Get the details of a subscription.
-long-summary: If no subscription is specified, shows the current subscription.
+long-summary: >
+    If no subscription is specified, information for the current subscription is returned.
+examples:
+    - name: >
+        Get the current default subscription returning results as JSON.  Use `--output yaml` for plain text.
```
jiasli, line 84: JSON is the default, but may be overridden by `defaults.output` config option. `yaml` is for YAML, not plain text.

src/azure-cli/azure/cli/command_modules/profile/_help.py
```diff
@@ -55,33 +77,54 @@
 helps['account show'] = """
 type: command
 short-summary: Get the details of a subscription.
-long-summary: If no subscription is specified, shows the current subscription.
+long-summary: >
+    If no subscription is specified, information for the current subscription is returned.
+examples:
+    - name: >
+        Get the current default subscription returning results as JSON.  Use `--output yaml` for plain text. 
+        You can also get details of the current subscription by using `az account list --query "[?isDefault]"`.
```
jiasli, line 85: It works, yes, but it is too complicated for a simple task like showing the current subscription. I feel we should remove this line and just let users run `az account show`.

src/azure-cli/azure/cli/command_modules/profile/_help.py
```diff
@@ -55,33 +77,54 @@
...
+    - name: Change the active subscription using a subscription name
+      text: >
+        az account set --subscription "My subscription name"
+    - name: Change the active subscription using a subscription ID
+      text: >
+        az account set --subscription 00000000-0000-0000-0000-000000000000
+    - name: Change the active subscription using a variable
+      text: >
+        subscriptionId = "$(az account list --query "[?isDefault].id" -o tsv)"
+        az account set --subscription $subscriptionId
```
jiasli, line 119: This is also a general usage and should not be included in command-specific example.

src/azure-cli/azure/cli/command_modules/profile/_help.py
```diff
@@ -55,33 +77,54 @@
...
 """
 
 helps['account get-access-token'] = """
 type: command
 short-summary: Get a token for utilities to access Azure.
 long-summary: >
-    The token will be valid for at least 5 minutes with the maximum at 60 minutes.
-    If the subscription argument isn't specified, the current account is used.
+    The token will be valid for at least 5 minutes with a maximum of 60 minutes.
+    If the subscription parameter isn't specified, the current account is used.
```
jiasli, line 127: Even though we use word "parameter" and "argument" interchangeably, "argument" is the official word. For example, in the in-tool help:

```
> az group show -h

Command
    az group show : Gets a resource group.

Arguments
    --name --resource-group -g -n [Required] : Name of resource group. You can configure the default
                                               group using `az configure --defaults group=<name>`.

Global Arguments
    --debug                                  : Increase logging verbosity to show all debug logs.
    --help -h                                : Show this help message and exit.
    --only-show-errors                       : Only show errors, suppressing warnings.
    --output -o                              : Output format.  Allowed values: json, jsonc, none,
                                               table, tsv, yaml, yamlc.  Default: json.
    --query                                  : JMESPath query string. See http://jmespath.org/ for
                                               more information and examples.
    --subscription                           : Name or ID of subscription. You can configure the
                                               default subscription using `az account set -s
                                               NAME_OR_ID`.
    --verbose                                : Increase logging verbosity. Use --debug for full
                                               debug logs.
```

In Python, the library to parse arguments are called `argparse`, and the library to do auto-completion is called `argcomplete`.

src/azure-cli/azure/cli/command_modules/profile/_help.py
```diff
@@ -95,9 +138,23 @@
     - name: Get an access token to use with MS Graph API
       text: >
         az account get-access-token --resource-type ms-graph
+    - name: >
+        Get an access token for a particular resource.  For a list of Virtual Network available service endpoints,
```
jiasli, line 142: Single space should suffice.
```suggestion
        Get an access token for a particular resource. For a list of Virtual Network available service endpoints,
```

src/azure-cli/azure/cli/command_modules/profile/_help.py
```diff
@@ -95,9 +138,23 @@
...
+        see `https://learn.microsoft.com/en-us/azure/virtual-network/virtual-network-service-endpoints-overview`.  
+        If you receive a `Failed to connect to MSI...` error, WHAT IS REALLY THE PROBLEM?  Is your resourceID wrong?  Does your resource not allow access tokens?
```
jiasli, line 144: What is the resource ID used? You may be hitting https://github.com/Azure/azure-cli/issues/11386.

src/azure-cli/azure/cli/command_modules/profile/_help.py
jiasli, line 142: You can't get access token for ARM resources, you can only get access token for ARM. This word `resource` here is a concept of AAD/Graph, not that of ARM.

src/azure-cli/azure/cli/command_modules/profile/_help.py
chasewilson: I agree with @jiasli here. The action is to "log in" and I believe should be "log in to".

src/azure-cli/azure/cli/command_modules/profile/_help.py
chasewilson: I guess this depends on if we are working to improve knowledge of our querying ability. @dbradish-microsoft, since this is a larger scale product, do you have insights showing we should be improving our content around queries?

src/azure-cli/azure/cli/command_modules/profile/_help.py
chasewilson: I agree here. We should avoid shell specific examples in reference documentation

src/azure-cli/azure/cli/command_modules/profile/_help.py
dbradish-microsoft: You are correct.

"Login, spelled as one word, is only a noun or an adjective. For example, the information you use to sign in to your email is your login (noun), and the page where you sign in is the login page (adjective). Log in is two words when it functions as a verb. For example, you log in with your login information."

src/azure-cli/azure/cli/command_modules/profile/_help.py
dbradish-microsoft: I feel that for our most used commands, like az account, I have to be careful not to turn our auto-gen content into a tutorial.  :-)  Also, we have a Learn H2 planned so customers can drop down to working How-to guides and tutorials that are more complex in nature.  (Because of this I also feel that we should limit hard-coding HTTPS links in autogen as much as possible.)  Keeping all this in mind, I feel `--query` examples are appropriate for key commands.  There is a balance to be achieved that **leans toward no query examples.**  

@chasewilson, I don't want to replicate [How to query Azure CLI command output using a JMESPath query](https://learn.microsoft.com/en-us/cli/azure/query-azure-cli?view=azure-cli-latest&tabs=concepts%2Cbash) in autogen.  I do want to use `--query` judiciously to help our customers start thinking about using the Azure CLI in multi-step scripts and variables.

Let's revisit this discussion once we get `az accoun`t, `az group1, and `az someService` completed.  I think we'll start to see a good pattern of examples by then, and maybe need to backtrack on `--query` if we think we've taken it too far.  Sound good?

src/azure-cli/azure/cli/command_modules/profile/_help.py
dbradish-microsoft: I was just testing with a random valid `resourceID` which begged the question, `which Azure resources have access tokens available?'

src/azure-cli/azure/cli/command_modules/profile/_help.py
dbradish-microsoft: Hmmm, I like this example!  :-)  Let's talk more about `general usage` and `shell specific` in our next docs:eng checkin.

src/azure-cli/azure/cli/command_modules/profile/_help.py
dbradish-microsoft: removed

src/azure-cli/azure/cli/command_modules/profile/_help.py
dbradish-microsoft: You are making an assumption that the customer hasn't changed their defaults, but I changed it to `--output table` as the majority of customers will have kept system defaults as you assumed.

src/azure-cli/azure/cli/command_modules/profile/_help.py
dbradish-microsoft: Per discussion with @chasewilson , we have agreed to test the following guidelines:

Guidelines:
1.	We can include `--query` and `--output` examples when the usefulness of a command requires it.  (Do not assume that every customer has kept the JSON output default.)
2.	We only provide Bash examples when…
a.	GitHub issues or our internal Power BI report “top error commands” shows that many customers are making the same mistake 

b.	Introductory reference commands.  If someone is new to Azure CLI, what commands will they learn first?
i.	az account
ii.	az group
iii.	az role
iv.	??
We aren’t trying to teach Bash.  We are using a handful of popular commands to show common usage patterns.

What we do for introductory reference commands is not a pattern for Azure service commands (vm, webapp…).

dbradish-microsoft:
closing draft PR as remote is out of sync with local branch.

jiasli:
> closing draft PR as remote is out of sync with local branch.

Actually, you may rebase the local branch and do a force push to overwrite the remote one. It is not necessary to create a new PR. 🙂

PR #24362, opened 2022-10-24
[Role] `az role assignment delete`: If `--ids` is provided, ignore other arguments, instead of raising error

jiasli:
Fix [IcM 341728422](https://portal.microsofticm.com/imp/v3/incidents/details/341728422/home)

**Related command**
`az role assignment delete`

**Description**
If `defaults.group` config option is set via:
```
az config set defaults.group=testrg
```

`az role assignment delete --ids` fails:
```
> az role assignment delete --ids /subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/providers/Microsoft.Authorization/roleAssignments/54df05ac-c9d3-4b71-a005-dec43d813f87
When assignment ids are used, other parameter values are not required
```

**Root Cause**
This command disallows using `--ids` with other arguments:

https://github.com/Azure/azure-cli/blob/98f7865402569ec9a6ae05ccdd0d27d5ea20927c/src/azure-cli/azure/cli/command_modules/role/custom.py#L497-L499

Argument `resource_group_name` is registered at

https://github.com/Azure/azure-cli/blob/d5271c3012201bc575db7e2d20bd8e1cf328dfd5/src/azure-cli-core/azure/cli/core/__init__.py#L492

`resource_group_name_type` has a `configured_default`:

https://github.com/Azure/azure-cli/blob/fa2a2d8ea269b43d8c397f9bd5c43bfc97258876/src/azure-cli-core/azure/cli/core/commands/parameters.py#L242

When `defaults.group` config option is set, `resource_group_name` variable will be populated, causing this failure.

**Proposed solution**
Removing the configured default value for `--resource-group` in `az role assignment delete` will be a BREAKING CHANGE and will break customers who rely on `--resource-group`'s configured default value.

Therefore, this PR changes the behavior so that if `--ids` is provided, it **ignores other arguments, instead of raising error.**

**Testing Guide**
```
az role assignment delete --ids {id} --assignee test --role test --resource-group test --scope test --include-inherit
```


src/azure-cli/azure/cli/command_modules/role/custom.py
```diff
@@ -495,8 +495,11 @@ def delete_role_assignments(cmd, ids=None, assignee=None, role=None, resource_gr
     definitions_client = factory.role_definitions
     ids = ids or []
     if ids:
-        if assignee or role or resource_group_name or scope or include_inherited:
-            raise CLIError('When assignment ids are used, other parameter values are not required')
+        ids_override_args = ['assignee', 'role', 'resource_group_name', 'scope', 'include_inherited']
+        for arg in ids_override_args:
+            if locals()[arg]:
+                logger.warning("option '%s' will be ignored due to use of '--ids'.",
+                               cmd.arguments[arg].type.settings['options_list'][0])
```
jiasli, line 502: This mimics the behavior of core for `--ids` handling:

https://github.com/Azure/azure-cli/blob/ce3a1f1ddea0b59f0a5803bb427c55862ccae72b/src/azure-cli-core/azure/cli/core/commands/arm.py#L265-L268

A difference is that core doesn't show warning if default value is used, but we do here. I feel being explicit is better than being implicit.

yonzhan:
Role

calvinhzy (approved):

PR #24384, opened 2022-10-24
[Reference freshness project] Resource (`az group`, `az account-management`...)

dbradish-microsoft:
This PR is part of the reference refresh project.  This PR focuses on `src/azure-cli/azure/cli/command_modules/resource/_help.py` and any related file containing summaries or examples for these same command groups.

The updates in this PR are intended to freshen the following reference areas:
1. short-summary
2. long-summary
3. examples

**Note:** The testing of all examples is out of scope for this project and is part of the **Content Quality Automation** crew project.

This is work item https://dev.azure.com/mseng/TechnicalContent/_workitems/edit/1998772.

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/resource/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/resource/_help.py
@@ -1176,17 +1176,13 @@ examples:
    - name: >
       Create a resource group but only if it does not exist.
     text: >
-        # run this script in a Docker Container
-        resourceGroup="myResourceGroup"
+        resourceGroup="MyResourceGroup"
         location="eastus"
         if [ $(az group exists --name $resourceGroup --output json) = false ]; then 
           az group create --name $resourceGroup --location $location
         else
-          echo The $resourceGroup resource group already exists.
+          echo A resource group named $resourceGroup already exists in this subscription.
         fi 
-
-        # run this script in Azure Cloud Shell
-        -- Chase Crum to supply.  Also check with az CLI engineering team
 """
 
 helps['group delete'] = """
@@ -1199,18 +1195,17 @@ examples:
    - name: >
       Delete a resource group but only if it exists.  Bypass the confirmation prompt.  Do not wait for the operation to finish.
     text: >
-        resourceGroup=myResourceGroup
-        location=eastus
+        resourceGroup=MyResourceGroup
         if [ $(az group exists --name $resourceGroup --output json) = true ]; then 
           az group delete --name $resourceGroup --yes  --no-wait
         else
-          echo The $resourceGroup resource group does not exist.
+          echo Resource group $resourceGroup does not exist.
         fi 
 """
 
 helps['group deployment'] = """
 type: group
-short-summary: Manage Azure Resource Manager deployments.
+short-summary: Manage Azure Resource Manager (ARM) deployments.
 """
 
 helps['group deployment create'] = """
@@ -1295,9 +1290,13 @@ helps['group list'] = """
 type: command
 short-summary: List resource groups.
 examples:
+  - name: List all resource groups for the current subscription returning results in a table
+    text: >
+        az group list --output table
   - name: List all resource groups located in the West US region.
     text: >
         az group list --query "[?location=='westus']"
+
 """
 
 helps['group lock'] = """
@@ -1307,71 +1306,99 @@ short-summary: Manage Azure resource group locks.
 
 helps['group lock create'] = """
 type: command
-short-summary: Create a resource group lock.
+short-summary: Create a resource group lock. 
+long-summary: Use locks to restrict updates or deletions of a resource group.  You must have WHAT PERMISSIONS? in the resource group to create a lock.
 examples:
-  - name: Create a read-only resource group level lock.
+  - name: Create a read-only resource group level lock with notes.
     text: >
-        az group lock create --lock-type ReadOnly -n lockName -g MyResourceGroup
+        az group lock create --lock-type ReadOnly --name MyLockName --resource-group MyResourceGroup --notes "My note that I want to remember."
+  - name: Create a lock that prohibits deletion of the resource group.
+    text: >
+        az group lock create --lock-type CanNotDelete --name MyLockName --resource-group MyResourceGroup
 """
 
 helps['group lock delete'] = """
 type: command
 short-summary: Delete a resource group lock.
+long-summary: You must have WHAT PERMISSIONS? in the resource group to delete a lock.
 examples:
   - name: Delete a resource group lock
     text: >
-        az group lock delete --name lockName -g MyResourceGroup
+        az group lock delete --name MyLockName --resource-group MyResourceGroup
 """
 
 helps['group lock list'] = """
 type: command
-short-summary: List lock information in the resource-group.
+short-summary: List lock information in a resource-group.
 examples:
-  - name: List out all locks on the resource group level
+  - name: List all locks for a resource group.
     text: >
-        az group lock list -g MyResourceGroup
+        az group lock list --resource-group MyResourceGroup
+  - name: List all locks meeting a filter criteria.
+        az group lock list --filter-string NEED EXAMPLE STRING HERE
+  - name: List all locks meeting a `--query` criteria.
+        az group lock list --query "[?level=='ReadOnly']"  THIS SCRIPT DOES NOT WORK
 """
 
 helps['group lock show'] = """
 type: command
-short-summary: Show the details of a resource group lock
+short-summary: Show the details of a resource group lock.
 examples:
-  - name: Show a resource group level lock
+  - name: Show the details of a resource group lock. The resource group name and lock name are both required if the `--ids` parameter is not used.
     text: >
-        az group lock show -n lockname -g MyResourceGroup
+        az group lock show --name MyLockName --resource-group MyResourceGroup
+  - name: >
+      Show the details of a resource group lock using the Id.  
+      The `--resource-group` and `--name` parameters are not needed in this example as these values exist in the Id.
+      For a list of available lock Ids, run `az group lock list -g MyResourceGroupName -o json`.
+    text: >
+        az group lock show --ids /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/MyResourceGroup/providers/Microsoft.Authorization/locks/MyLockName1
+
 """
 
 helps['group lock update'] = """
 type: command
-short-summary: Update a resource group lock.
+short-summary: Update a resource group lock with new notes, or change the lock type.
+long-summary: You must have WHAT PERMISSIONS? in the resource group to update a lock.
 examples:
   - name: Update a resource group lock with new notes and type
     text: >
         az group lock update --name lockName -g MyResourceGroup --notes newNotesHere --lock-type CanNotDelete
+  - name: Update a resource group lock using the Id.  For a list of available lock Ids, run `az group lock list -g MyResourceGroupName -o json`.
+    text: >
+        az group lock update --ids /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/MyResourceGroup/providers/Microsoft.Authorization/locks/MyLockName --notes MyNewNote
+  - name: Update several resource group locks at the same time using a space delimited list of Ids.
+    text: >
+        az group lock update --ids /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/MyResourceGroup/providers/Microsoft.Authorization/locks/MyLockName1 \\
+            /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/MyResourceGroup/providers/Microsoft.Authorization/locks/MyLockName2 \\
+            --notes MyNewNoteForAllIDs
 """
 
 helps['group update'] = """
 type: command
 short-summary: Update a resource group.
 examples:
-  - name: Update a resource group. (autogenerated)
+  - name: Add tags to a resource group.
     text: |
         az group update --resource-group MyResourceGroup --set tags.CostCenter='{"Dept":"IT","Environment":"Test"}'
-    crafted: true
+  - name: Remove tags from a resource group
+    text: |
+        az group update --resource-group MyResourceGroup --remove tags
+  - name: Add a property to a resource group preserving string literals
+    text: |
+        az group update --resource-group MyResourceGroup --add properties.provisioningState NEED EXAMPLE HERE --force-string
 """
 
 helps['group wait'] = """
 type: command
-short-summary: Place the CLI in a waiting state until a condition of the resource group is met.
+short-summary: Place the Azure CLI in a waiting state until a condition of the resource group is met.
 examples:
-  - name: Place the CLI in a waiting state until a condition of the resource group is met. (autogenerated)
+  - name: Place the Azure CLI in a waiting state until `provisioningState` becomes `succeeded`.
     text: |
         az group wait --created  --resource-group MyResourceGroup
-    crafted: true
-  - name: Place the CLI in a waiting state until a condition of the resource group is met. (autogenerated)
+  - name: Place the Azure CLI in a waiting state as a resource group is deleted.
     text: |
         az group wait --deleted --resource-group MyResourceGroup
-    crafted: true
 """
 
 helps['lock'] = """
@@ -1899,7 +1926,7 @@ examples:
         az policy exemption create -n exemptTestVM \\
             --policy-assignment "/subscriptions/mySubId/providers/Microsoft.Authorization/policyAssignments/limitVMSku" \\
             --exemption-category "Waiver" \\
-            --resource-group "myResourceGroup"
+            --resource-group "MyResourceGroup"
   - name: Create a policy exemption in a management group.
     text: |
         az policy exemption create -n exemptTestVM \\
@@ -1914,7 +1941,7 @@ short-summary: Delete a policy exemption.
 examples:
   - name: Delete a policy exemption.
     text: |
-        az policy exemption delete --name MyPolicyExemption --resource-group "myResourceGroup"
+        az policy exemption delete --name MyPolicyExemption --resource-group "MyResourceGroup"
     crafted: true
 """
 
@@ -1929,7 +1956,7 @@ short-summary: Show a policy exemption.
 examples:
   - name: Show a policy exemption.
     text: |
-        az policy exemption show --name MyPolicyExemption --resource-group "myResourceGroup"
+        az policy exemption show --name MyPolicyExemption --resource-group "MyResourceGroup"
     crafted: true
 """
 
@@ -1945,7 +1972,7 @@ examples:
     text: |
         az policy exemption update -n exemptTestVM \\
             --exemption-category "Mitigated" \\
-            --resource-group "myResourceGroup"
+            --resource-group "MyResourceGroup"
   - name: Update a policy exemption in a management group.
     text: |
         az policy exemption update -n exemptTestVM \\
@@ -2300,7 +2327,7 @@ examples:
     crafted: true
   - name: Update a resource. (autogenerated)
     text: |
-        az resource update --name myresource --resource-group myresourcegroup --resource-type subnets --set tags.key=value
+        az resource update --name myresource --resource-group MyResourceGroup --resource-type subnets --set tags.key=value
     crafted: true
 """
 
--- a/src/azure-cli/azure/cli/command_modules/resource/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/resource/_help.py
@@ -18,7 +18,7 @@ short-summary: Create a subscription lock.
 examples:
   - name: Create a read-only subscription level lock.
     text: >
-        az account lock create --lock-type ReadOnly -n lockName
+        az account lock create --lock-type ReadOnly --name MyLockName
 """
 
 helps['account lock delete'] = """
@@ -27,7 +27,7 @@ short-summary: Delete a subscription lock.
 examples:
   - name: Delete a subscription lock
     text: >
-        az account lock delete --name lockName
+        az account lock delete --name MyLockName
 """
 
 helps['account lock list'] = """
@@ -45,7 +45,7 @@ short-summary: Show the details of a subscription lock
 examples:
   - name: Show a subscription level lock
     text: >
-        az account lock show -n lockname
+        az account lock show --name MyLockName
 """
 
 helps['account lock update'] = """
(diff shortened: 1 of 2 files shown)


yonzhan:
@zhoxing-ms for awareness

src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -54,28 +54,18 @@
...
 type: group
-short-summary: Manage Azure Management Groups.
+short-summary: Manage Azure subscription-level management groups.
+long-summary: For more information on Azure management groups, see [What are Azure management groups?](https://learn.microsoft.com/en-us/azure/governance/management-groups/overview)
 """
 
 helps['account management-group create'] = """
 type: command
 short-summary: Create a new management group.
-long-summary: Create a new management group.
```
dbradish-microsoft, line 68: @zhoxing-ms, as part of general file cleanup, parameter information was removed under the assumptions 1.) this information is not in use here and 2.) parameter information is now in _params.py.  Is this correct?

src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -319,10 +231,10 @@
 type: command
 short-summary: Place the CLI in a waiting state until a deployment condition is met.
 examples:
-  - name: Place the CLI in a waiting state until a deployment condition is met. (autogenerated)
```
dbradish-microsoft, line 322: @zhoxing-ms , please note that all `(autogenerated)` and `crafted: true` have been removed.  Telling our customers that a sample was autogenerated is not helpful to them.

src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -527,160 +384,102 @@
 short-summary: List deployments at resource group.
 examples:
   - name: List deployments at resource group.
-    text: az deployment group list -g testrg
```
dbradish-microsoft, line 530: @zhoxing-ms , please note that _for clarity_, abbreviated reference names have been replaced with full parameter names.  I've also attempted to standardize on a `MyName` object naming convention.

src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -1263,139 +957,185 @@
...
+    text: |
+        az group update --resource-group MyResourceGroup --set tags.Department1='{"Department name":"IT","Cost Center":"tech002"}' --force-string
+  - name: Remove tags from a resource group.
     text: |
-        az group update --resource-group MyResourceGroup --set tags.CostCenter='{"Dept":"IT","Environment":"Test"}'
-    crafted: true
+        az group update --resource-group MyResourceGroup --remove tags
+  - name: Add a property to a resource group.
+    text: |
+        az group update --resource-group MyResourceGroup --add properties.MyPropertyName=MyValue NEED WORKING EXAMPLE HERE
```
dbradish-microsoft, line 1066: @zhoxing-ms Looking at [arm.py#L389-L391](https://github.com/Azure/azure-cli/blob/8779b435ba3d9f920986f7c7990f717b58507a51/src/azure-cli-core/azure/cli/core/commands/arm.py#L389-L391) ...

add_usage = '--add property.listProperty <key=value, string or JSON string>'
set_usage = '--set property1.property2=<value>'
remove_usage = '--remove property.list <indexToRemove> OR --remove propertyToRemove'

... I am intuitively trying to get `--set properties.MyNewPropertyName` to work like `--set tags.MyNewTagName`.  Can you please provide a working example for `--add` and `--set` as I am obviously missing the connection?

image

src/azure-cli/azure/cli/command_modules/resource/_help.py
zhoxing-ms: In fact, the help message of the CLI has four layers, the priority of the help message written in `_help.py` is the highest. For more information, please refer to this doc: [Azure CLI Help System](https://github.com/Azure/azure-cli/blob/dev/doc/authoring_help.md#help-layers). Since there is no repeated help message declared in [_params.py](https://github.com/Azure/azure-cli/blob/22c9d89df193b659179266c405bc713723f74bb5/src/azure-cli/azure/cli/command_modules/resource/_params.py#L594-L596
), so the information here is useful. 

May I ask why you want to remove these help messages?

src/azure-cli/azure/cli/command_modules/resource/_help.py
zhoxing-ms: I'm not sure whether the original design has other purposes besides informing users that these examples are automatically generated by Aladdin. @kairu-ms What do you think of this?

src/azure-cli/azure/cli/command_modules/resource/_help.py
zhoxing-ms: OK, nice work! 👍👍 Thanks so much for your contribution!

src/azure-cli/azure/cli/command_modules/resource/_help.py
zhoxing-ms: For `--set`, this is an example: `az group update -g zhoxing-test --set tags.key=value`
image

zhoxing-ms: For the `--add` parameter, it's usage scenario is to add elements to an array type property. Currently, the resource group does not have such property. This parameter is only added by `generic_update_command` framework, I believe customers have no need to use it, so it should not be necessary to add an example for it

src/azure-cli/azure/cli/command_modules/resource/_help.py
dbradish-microsoft: @zhoxing-ms , I am working under the assumption that 1.) this information is not in use here and 2.) parameter information is now in _params.py. Is this correct?

src/azure-cli/azure/cli/command_modules/resource/_help.py
dbradish-microsoft: @kairu-ms, am I good to go with this type of _help.py file update?

src/azure-cli/azure/cli/command_modules/resource/_help.py
dbradish-microsoft: @zhoxing-ms, is the `--add` parameter then an example of where we should set `is-visible = false` somewhere so it doesn't even show in the autogenerated content?

I'll remove my examples, but I feel that parameters a customer won't (can't?) use just makes noise.  What is your perspective?

src/azure-cli/azure/cli/command_modules/resource/_help.py
zhoxing-ms: > Since there is no repeated help message declared in [_params.py](https://github.com/Azure/azure-cli/blob/22c9d89df193b659179266c405bc713723f74bb5/src/azure-cli/azure/cli/command_modules/resource/_params.py#L594-L596), so the information here is useful.

No, as I mentioned earlier, the current help message comes from `_help.py`.

src/azure-cli/azure/cli/command_modules/resource/_help.py
zhoxing-ms: Good suggestion! I think we can consider hiding the help message of this parameter `--add` through `c.deprecate(hide=True)`

dbradish-microsoft:
@zhoxing-ms , friendly FYI that I've put the parameter information back into the _help.py file.

**Question**, does it matter to you if the `--name:` and sub information is left justified, like this: 
image

The above updates were per @jiasli 's request in PR https://github.com/Azure/azure-cli/pull/24332/files

dbradish-microsoft:
@zhoxing-ms , "Personally, I think there is no problem. In fact, both methods are used in CLI."  

I personally feel the green box above it is more difficult to read.  So, with your feedback, I am not going to put myself in the reformatting business.  :-)  I'll leave well enough alone in this file.

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -1327,7 +1327,7 @@ def update_resource_group(instance, tags=None):
 
 def export_group_as_template(
         cmd, resource_group_name, include_comments=False, include_parameter_default_value=False, resource_ids=None, skip_resource_name_params=False, skip_all_params=False):
-    """Captures a resource group as a template.
+    """Capture a resource group as a template.
```
zhoxing-ms, line 1330: In fact, they are just comments in the code which will not affect the help message. However, these changes themselves look good~

src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -131,59 +117,51 @@
...
   - name: Get a management group with children in all levels of hierarchy.
-    text: >
-        az account management-group show --name GroupName -e -r
+    text: az account management-group show --name MyGroupName --expand --recurse
 """
 
 helps['account management-group subscription'] = """
 type: group
-short-summary: Subscription operations for Management Groups.
+short-summary: Add or remove subscriptions in a management group.
```
zhoxing-ms, line 129: In fact, because it is uncertain whether other operations will be expanded in the future, it is recommended to use a unified and broad description `Manage`
```suggestion
short-summary: Manage subscriptions in a management group.
```

src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -193,17 +171,14 @@
     short-summary: Updates the display name of the management group. If null, no change is made.
   - name: --parent -p
     type: string
-    short-summary: Update the parent of the management group. Can be the fully qualified id or the name of the management group. If null, no change is made.
+    short-summary: Updates the parent of the management group. Can be the fully qualified id or the name of the management group. If null, no change is made.
```
zhoxing-ms, line 174: We need to use the first-person voice in the help message.
```suggestion
    short-summary: Update the parent of the management group. Can be the fully qualified id or the name of the management group. If null, no change is made.
```

src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -294,35 +273,35 @@
...
             --parameters "{ \\"policyName\\": { \\"value\\": \\"policy2\\" }}"
   - name: Create a deployment at subscription scope from a local template, using a parameter file, a remote parameter file, and selectively overriding key/value pairs.
     text: >
-        az deployment create --location WestUS --template-file azuredeploy.json  \\
+        az deployment create --location WestUS --template-file azuredeploy.json \\
             --parameters @params.json --parameters https://mysite/params.json --parameters MyValue=This MyArray=@array.json
   - name: Create a deployment at subscription scope from a template-spec
     text: >
-        az deployment create --location WestUS --template-spec "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/myRG/providers/Microsoft.Resources/templateSpecs/myTemplateSpec/versions/1.0"
+        az deployment create --location WestUS --template-spec "/subscriptions/MySubscriptionID/resourceGroups/myRG/providers/Microsoft.Resources/templateSpecs/myTemplateSpec/versions/1.0"
```
zhoxing-ms, line 285: I'm not sure whether the placeholder `MySubscriptionID` is clearer than the sample value `00000000-0000-0000-0000-000000000000`. The sample value may make it easier for customers to understand the format of the subscription ID

src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -577,16 +553,15 @@
...
+    text: az deployment group validate --resource-group MyResourceGroup --template-file {template-file}
+  - name: Validate whether a template is valid at resource group.
+    text: >
+        az deployment group validate --parameters MyValue=This MyArray=@array.json --resource-group MyResourceGroup --template-file azuredeploy.json
 """
 
 helps['deployment group create'] = """
 type: command
-short-summary: Start a deployment at resource group.
+short-summary: Create a deployment group.
```
zhoxing-ms, line 564: In fact, the feature of this command is to create a resource group level deployment, so the previous description`Start a deployment` may be more appropriate.
```suggestion
short-summary: Start a deployment at resource group scope.
```

src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -610,24 +585,24 @@
...
     text: >
-        az deployment group create --resource-group testrg --name rollout01 \\
-            --template-file azuredeploy.json  --parameters @params.json \\
+        az deployment group create --resource-group MyResourceGroup --name rollout01 \\
+            --template-file azuredeploy.json --parameters @params.json \\
             --parameters https://mysite/params.json --parameters MyValue=This MyArray=@array.json
   - name: Create a deployment at resource group scope from a template-spec
     text: >
-        az deployment group create --resource-group testrg --template-spec "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/testrg/providers/Microsoft.Resources/templateSpecs/myTemplateSpec/versions/1.0"
+        az deployment group create --resource-group MyResourceGroup --template-spec "/subscriptions/MySubscriptionID/resourceGroups/MyResourceGroup/providers/Microsoft.Resources/templateSpecs/myTemplateSpec/versions/1.0"
```
zhoxing-ms, line 605: Same as above

src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -1271,131 +1272,167 @@
...
     text: >
-        az group lock show -n lockname -g MyResourceGroup
+        az group lock show --ids /subscriptions/MySubscriptionID/resourceGroups/MyResourceGroup/providers/Microsoft.Authorization/locks/MyLockName1 \\
+            /subscriptions/MySubscriptionID/resourceGroups/MyResourceGroup/providers/Microsoft.Authorization/locks/MyLockName2 
 """
 
 helps['group lock update'] = """
 type: command
-short-summary: Update a resource group lock.
+short-summary: Update a resource group lock with new notes, or change the lock type.
```
zhoxing-ms, line 1338: I'm not sure whether other properties will be expanded in the future. Do we need to describe the `short-summary` so specifically?

src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -1271,131 +1272,167 @@
...
 examples:
   - name: Update a resource group lock with new notes and type
     text: >
-        az group lock update --name lockName -g MyResourceGroup --notes newNotesHere --lock-type CanNotDelete
+        az group lock update --name MyLockName --resource-group MyResourceGroup --notes newNotesHere --lock-type CanNotDelete
+  - name: Update a resource group lock using the Id.  For a list of available lock Ids, run `az group lock list --resource-group MyResourceGroup -o json`.
```
zhoxing-ms, line 1343:
```suggestion
  - name: Update a resource group lock using the Id. For a list of available lock Ids, run `az group lock list --resource-group MyResourceGroup -o json`.
```

src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -1430,15 +1466,15 @@
...
+            --managedapp-definition-id "/subscriptions/MySubscriptionID/resourceGroups/MyResourceGroup/providers/Microsoft.Solutions/applianceDefinitions/{ApplianceDefinition}"
   - name: Create a managed application of kind 'MarketPlace'. This requires a valid plan, containing details about existing marketplace package like plan name, version, publisher and product.
-    text: |
-        az managedapp create -g MyResourceGroup -n MyManagedApp -l westcentralus --kind MarketPlace \\
-            -m "/subscriptions/{SubID}/resourceGroups/{ManagedResourceGroup}" \\
-            --plan-name ContosoAppliance --plan-version "1.0" --plan-product "contoso-appliance" --plan-publisher Contoso
+    text: >
+        az managedapp create --resource-group MyResourceGroup --name MyManagedApp --location westcentralus --kind MarketPlace \\
+            --management-group-id "/subscriptions/MySubscriptionID/resourceGroups/{ManagedResourceGroup}" \\
+            --plan-name MyAppliancePlan --plan-version "1.0" --plan-product "my-appliance" --plan-publisher MyPlanPublisher
```
zhoxing-ms, line 1477: > MyResourceGroup

Same as above

src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -1537,48 +1569,47 @@
...
-    text: |
+    text: >
         Valid scopes are management group, subscription, resource group, and resource, for example
            management group:  /providers/Microsoft.Management/managementGroups/MyManagementGroup
-           subscription:      /subscriptions/0b1f6471-1bf0-4dda-aec3-111122223333
-           resource group:    /subscriptions/0b1f6471-1bf0-4dda-aec3-111122223333/resourceGroups/myGroup
-           resource:          /subscriptions/0b1f6471-1bf0-4dda-aec3-111122223333/resourceGroups/myGroup/providers/Microsoft.Compute/virtualMachines/myVM
+           subscription:      /subscriptions/MySubscriptionID
+           resource group:    /subscriptions/MySubscriptionID/resourceGroups/MyResourceGroup
+           resource:          /subscriptions/MySubscriptionID/resourceGroups/MyResourceGroup/providers/Microsoft.Compute/virtualMachines/myVM
```
zhoxing-ms, line 1577: Same as above

src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -1771,31 +1797,31 @@
...
-        az policy set-definition create -n computeRequirements \\
-            --definitions "[ { \\"policyDefinitionId \\": \\"/subscriptions/mySubId/providers/ \\
+    text: >
+        az policy set-definition create --name computeRequirements \\
+            --definitions "[ { \\"policyDefinitionId \\": \\"/subscriptions/MySubscriptionID/providers/ \\
                 Microsoft.Authorization/policyDefinitions/storagePolicy\\", \\"groupNames\\": \\
                     [ \\"CostSaving\\", \\"Organizational\\" ] }, { \\"policyDefinitionId\\": \\
-                        \\"/subscriptions/mySubId/providers/Microsoft.Authorization/ \\
+                        \\"/subscriptions/MySubscriptionID/providers/Microsoft.Authorization/ \\
                             policyDefinitions/tagPolicy\\", \\"groupNames\\": [ \\
```
zhoxing-ms, line 1825: > mySubId

Could we consider replacing placeholders with sample values so that users can understand it's format?

src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -1820,29 +1845,28 @@
...
-    text: |
-        az policy set-definition update -n computeRequirements \\
-            --definitions "[ { \\"policyDefinitionId\\": \\"/subscriptions/mySubId/providers/ \\
+    text: >
+        az policy set-definition update --name computeRequirements \\
+            --definitions "[ { \\"policyDefinitionId\\": \\"/subscriptions/MySubscriptionID/providers/ \\
                 Microsoft.Authorization/policyDefinitions/storagePolicy\\", \\"groupNames\\": [ \\
                     \\"CostSaving\\", \\"Organizational\\" ] }, { \\"policyDefinitionId\\": \\
-                        \\"/subscriptions/mySubId/providers/Microsoft.Authorization/ \\
+                        \\"/subscriptions/MySubscriptionID/providers/Microsoft.Authorization/ \\
```
zhoxing-ms, line 1869: Same as above

src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -1858,32 +1882,31 @@
...
+        az policy exemption create --name MyExemptVM \\
+            --policy-assignment "/subscriptions/MySubscriptionID/providers/Microsoft.Authorization/policyAssignments/limitVMSku" \\
             --exemption-category "Waiver"
   - name: Create a policy exemption in the resource group.
-    text: |
-        az policy exemption create -n exemptTestVM \\
-            --policy-assignment "/subscriptions/mySubId/providers/Microsoft.Authorization/policyAssignments/limitVMSku" \\
+    text: >
+        az policy exemption create --name MyExemptVM \\
+            --policy-assignment "/subscriptions/MySubscriptionID/providers/Microsoft.Authorization/policyAssignments/limitVMSku" \\
```
zhoxing-ms, line 1892: > MySubscriptionID

Same as above

src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -1997,30 +2017,30 @@
...
+        az resource create --resource-group myRG --name myApiApp --resource-type Microsoft.web/sites --is-full-object --properties @jsonConfigFile
   - name: Create a web app with the minimum required configuration information.
-    text: |
-        az resource create -g myRG -n myWeb --resource-type Microsoft.web/sites \\
-            --properties "{ \\"serverFarmId\\":\\"/subscriptions/{SubID}/resourcegroups/ \\
-                {ResourceGroup}/providers/Microsoft.Web/serverfarms/{ServicePlan}\\" }"
+    text: >
+        az resource create --resource-group myRG --name myWeb --resource-type Microsoft.web/sites \\
+            --properties "{ \\"serverFarmId\\":\\"/subscriptions/MySubscriptionID/resourcegroups/ \\
+                MyResourceGroup/providers/Microsoft.Web/serverfarms/MyServicePlan\\" }"
```
zhoxing-ms, line 2036: > MySubscriptionID

Same as above

src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -2032,41 +2052,40 @@
...
+        az resource delete --resource-group MyResourceGroup --name MyVm --resource-type "Microsoft.Compute/virtualMachines"
   - name: Delete a web app using a resource identifier.
     text: >
-        az resource delete --ids /subscriptions/0b1f6471-1bf0-4dda-aec3-111111111111/resourceGroups/MyResourceGroup/providers/Microsoft.Web/sites/MyWebapp
+        az resource delete --ids /subscriptions/MySubscriptionID/resourceGroups/MyResourceGroup/providers/Microsoft.Web/sites/MyWebapp
   - name: Delete a subnet using a resource identifier.
     text: >
-        az resource delete --ids /subscriptions/0b1f6471-1bf0-4dda-aec3-111111111111/resourceGroups/MyResourceGroup/providers/Microsoft.Network/virtualNetworks/MyVnet/subnets/MySubnet
+        az resource delete --ids /subscriptions/MySubscriptionID/resourceGroups/MyResourceGroup/providers/Microsoft.Network/virtualNetworks/MyVNet/subnets/MySubnet
   - name: Delete a virtual machine named 'MyVm' by using the latest api-version whether this version is a preview version.
```
zhoxing-ms, line 2062: > MySubscriptionID

Same as above

dbradish-microsoft:
I am giving the methodology for freshening azCLI autogen reference content, and consequently this PR, a necessary ending in order to make way for new beginnings.

PR #24434, opened 2022-10-26
[Network] `az network bastion rdp`: Allow rdp session customization

### LINKED ISSUE #23866: Add opt-in/opt-out for multiple monitors on `az network bastion rdp` command
**Related command**
`az network bastion rdp --name MyBastionHost --resource-group MyResourceGroup --target-resource-id vmResourceId`

While connecting to an Azure resource through Azure Bastion Standard native client, generated RDP file will always use all monitors by default, which is not always desired when working with multiple screens.

Generated RDP file excerpt:
```ini
hostname:s:<serverName>
full address:s:<serverName>
alternate full address:s:<serverName>
use multimon:i:1
serverport:i:3389
gatewaycredentialssource:i:5
gatewayusagemethod:i:1
gatewayprofileusagemethod:i:1
gatewayhostname:s:<bastionPIPName>.bastion.azure.com
gatewayaccesstoken:s:<gatewayAccessToken>
signscope:s:Full Address,Alternate Full Address,GatewayHostname,GatewayUsageMethod,GatewayProfileUsageMethod,GatewayCredentialsSource
signature:s:<AuthSignature>

```

Would be interesting to give option to administrator while connecting to bastion to opt-in or opt-out for the multiple monitor options based in a command switch (eg. `az network bastion rdp --name MyBastionHost --resource-group MyResourceGroup --target-resource-id vmResourceId --use-multimon`)

As generated RDP file comes from managed Remote Desktop Gateway, the solution would be to modify `conn.rdp` in `rdp_bastion_host` function to replace the integer value of `use multimon` to **0** in the presence or absence of a parameter to use multiple monitors, which should be modified in `src/azure-cli/azure/cli/command_modules/network/custom.py`.

As a workaround, whenever needed to use a single monitor, we can cancel connection and then replace value in `conn.rdp` file as snippet below

```powershell
(Get-Content .\conn.rdp).Replace("use multimon:i:1","use multimon:i:0") | Set-Content .\conn.rdp
mstsc .\conn.rdp
```

**Additional context**
N/A

PramodValavala-MSFT:
**Related command**

```sh
az network bastion rdp --name {} --resource-group {} --target-resource-id {}
```

**Description**

When connecting to bastion via RDP using Native Client, the session defaults to using all available monitors. But this may not always be the prefered by customers as mentioned in #23866.

Currently, customers have to disconnect from the session, either manually edit the RDP file or edit it using the Remote Desktop Connection UI, and then reconnect.

This PR adds a flag, `--configure`, which opens the Remote Desktop Connection UI instead of connecting directly, allowing customers to customize the session as required.

resolves #23866 

**Testing Guide**

```sh
az network bastion rdp --name {} --resource-group {} --target-resource-id {} --configure
```

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/network/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/network/_params.py
@@ -2146,6 +2146,7 @@ def load_arguments(self, _):
         c.argument('ssh_key', help='SSH key file location for SSH connections.', options_list=['--ssh-key'])
     with self.argument_context('network bastion rdp') as c:
         c.argument('disable_gateway', arg_type=get_three_state_flag(), help='Flag to disable access through RD gateway.')
+        c.argument('configure', help='Flag to configure RDP session.', action='store_true')
     with self.argument_context('network bastion tunnel') as c:
         c.argument('port', help='Local port to use for the tunneling.', options_list=['--port'])
         c.argument('timeout', help='Timeout for connection to bastion host tunnel.', options_list=['--timeout'])
--- a/src/azure-cli/azure/cli/command_modules/network/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/network/custom.py
@@ -8271,7 +8271,7 @@ def ssh_bastion_host(cmd, auth_type, target_resource_id, resource_group_name, ba
         tunnel_server.cleanup()
 
 
-def rdp_bastion_host(cmd, target_resource_id, resource_group_name, bastion_host_name, resource_port=None, disable_gateway=False):
+def rdp_bastion_host(cmd, target_resource_id, resource_group_name, bastion_host_name, resource_port=None, disable_gateway=False, configure=False):
     from azure.cli.core._profile import Profile
     import os
     from ._process_helper import launch_and_wait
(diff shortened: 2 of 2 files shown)


yonzhan:
Network

src/azure-cli/azure/cli/command_modules/network/_params.py
```diff
@@ -2146,6 +2146,7 @@ def load_arguments(self, _):
         c.argument('ssh_key', help='SSH key file location for SSH connections.', options_list=['--ssh-key'])
     with self.argument_context('network bastion rdp') as c:
         c.argument('disable_gateway', arg_type=get_three_state_flag(), help='Flag to disable access through RD gateway.')
+        c.argument('configure', help='Flag to configure RDP session.')
```
necusjz, line 2149: Do you think it should be `get_three_state_flag` type as `disable_gateway`?

PramodValavala-MSFT: @necusjz The flag here is just to trigger a different behavior - opening the Remote Desktop Connection UI instead of directly connecting to the host - compared to a value that is persisted.

This was my reasoning for not setting the `arg_type` based on my understanding on what the `get_three_state_flag` type means (explicit true or false field). Let me know if my understanding is incorrect here though.

necusjz: @PramodValavala-MSFT Here is the reference https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#registering-flags.

PramodValavala-MSFT: @necusja Thanks for sharing that! This would more be a switch based on the docs, so I will update accordingly.

necusjz (approved):

sgsvetko:
It's good  --configuration option has been added, but it'll be better if rdp session open on one monitor by default, instead each and every time to uncheck "Use all my monitors for the remote session"

ktzyl:
> It's good --configuration option has been added, but it'll be better if rdp session open on one monitor by default, instead each and every time to uncheck "Use all my monitors for the remote session"

Same here. Or --configure can add parameter to set single or all monitor.

PR #24407, opened 2022-10-26
{CI} pytest: `--boxed` argument is deprecated, use `--forked` instead

wangzelin007:
The latest pytest-xdist 3.0.2 (2022-10-25) has [removed](https://pytest-xdist.readthedocs.io/en/latest/changelog.html#removals) `--boxed` argument.
But we still need this functionality, install [pytest-forked](https://pypi.org/project/pytest-forked) separately and use `--forked` instead.
Error log:
```
ERROR: usage: __main__.py [options] [file_or_dir] [file_or_dir] [...]
__main__.py: error: unrecognized arguments: --boxed
  inifile: None
  rootdir: /mnt/vss/_work/1/s/src/azure-cli-core
```
Test screenshot:
image
image

PR #24442, opened 2022-10-27
{CI} Revert #24407 pytest: `--boxed` argument is deprecated, use `--forked` instead

wangzelin007:
Reverts Azure/azure-cli#24407
It seems we run these scripts in docker.

- scripts/release/debian/test_deb_package.py

- scripts/release/homebrew/test_homebrew_package.py

- scripts/release/rpm/test_rpm_package.py

And we need revert #24407 unless we use the latest pytest-xdist in docker images.
image
image


jiasli:
Maybe we should bump `pytest-xdist` instead?

PR #24444, opened 2022-10-27
{CI} Install pytest-forked for docker image

wangzelin007:
The CI pipeline error is as follows:
image
image

We run these test scripts in docker.
- scripts/releases/debian/test_deb_package.py
- scripts/release/homebrew/test_homebrew_package.py
- scripts/releases/rpm/test_rpm_package.py

But I only have the latest pytest-forked installed in azdev(Azure/azure-cli-dev-tools#347), which only works when using the `azdev test` command.
We also need to install the latest pytest-forked in docker as well.
Test screenshot:
image


bebound (approved):

jiasli (approved):

yonzhan:
CI

PR #24447, opened 2022-10-27
{CI} Fix template not found error: Use absolute path for variable template

wangzelin007:
Azure CLI Live Test pipeline will report an error:
```
/scripts/live_test/CLITest.yml: Could not find /scripts/live_test/.azure-pipelines/templates/variables.yml in repository self hosted on https://github.com/ using commit 8444d8686d50a1c1378273efbd05101e02d2c36d. GitHub reported the error, "Not Found"
```
Because the template path uses **relative path by default**, so except for the pipeline script in the root directory, other pipeline scripts cannot find the variable template.
This PR will use absolute path for variable template in all pipeline scripts

image
Test screenshot:

image

### DIFF
--- a/azure-pipelines-full-tests.yml
+++ b/azure-pipelines-full-tests.yml
@@ -13,7 +13,7 @@ pr:
     - '*'
 
 variables:
-- template: ${{variables['Pipeline.Workspace']}}/.azure-pipelines/templates/variables.yml
+- template: ${{ variables.Pipeline.Workspace }}/.azure-pipelines/templates/variables.yml
 
 jobs:
 - job: AutomationTest20200901
--- a/azure-pipelines-full-tests.yml
+++ b/azure-pipelines-full-tests.yml
@@ -13,7 +13,7 @@ pr:
     - '*'
 
 variables:
-- template: .azure-pipelines/templates/variables.yml
+- template: ${{variables['Pipeline.Workspace']}}/.azure-pipelines/templates/variables.yml
 
 jobs:
 - job: AutomationTest20200901
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -13,7 +13,7 @@ pr:
     - '*'
 
 variables:
-- template: $(Pipeline.Workspace)/azure-cli/.azure-pipelines/templates/variables.yml
+- template: ${{variables['Pipeline.Workspace']}}/azure-cli/.azure-pipelines/templates/variables.yml
 
 jobs:
 - job: CheckPullRequest
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -13,7 +13,7 @@ pr:
     - '*'
 
 variables:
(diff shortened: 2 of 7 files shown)


azure-pipelines-full-tests.yml
```diff
@@ -13,7 +13,7 @@ pr:
     - '*'
 
 variables:
-- template: .azure-pipelines/templates/variables.yml
+- template: ${{variables['Pipeline.Workspace']}}/.azure-pipelines/templates/variables.yml
```
jiasli, line 16: Can we use `$(Pipeline.Workspace)`?

yonzhan:
CI

azure-pipelines-full-tests.yml
wangzelin007: If we use `$(Pipeline.Workspace)`, ADO pipeline will report an error:
```
/azure-pipelines.yml: Could not find /$(Pipeline.Workspace)/azure-cli/.azure-pipelines/templates/variables.yml in repository self hosted on https://github.com/ using commit a2544c34c66f64ef3bc97983da82dd4db3595b40. GitHub reported the error, "Not Found"
```

azure-pipelines-full-tests.yml
```diff
@@ -13,7 +13,7 @@ pr:
     - '*'
 
 variables:
-- template: .azure-pipelines/templates/variables.yml
+- template: ${{ variables.Pipeline.Workspace }}/.azure-pipelines/templates/variables.yml
```
wangzelin007, line 16: The template path uses relative path by default.
Actually there is no need to modify the template path in azure-pipelines.yml and azure-pipelines-full-tests.yml.
Just to keep things consistent.

azure-pipelines.yml
```diff
@@ -13,7 +13,7 @@ pr:
     - '*'
 
 variables:
-- template: .azure-pipelines/templates/variables.yml
+- template: ${{ variables.Pipeline.Workspace }}/.azure-pipelines/templates/variables.yml
```
wangzelin007, line 16: The template path uses relative path by default.
Actually there is no need to modify the template path in azure-pipelines.yml and azure-pipelines-full-tests.yml.
Just to keep things consistent.

kairu-ms (approved):

jiasli (approved):

azure-pipelines-full-tests.yml
jiasli: Thanks for the comment to make it clear! 🙂

azure-pipelines-full-tests.yml
jiasli: To add more info, per https://learn.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#understand-variable-syntax, template expression variables (`${{ variables.var }}`) is officially documented.

PR #24582, opened 2022-11-08
{CI} Add extension dependency check

bebound:
**Description**

Inspired by `TestExtensionsLoading`.
This test will check if extension's dependency is conflict with azure-cli. This should avoid issues like https://github.com/Azure/azure-cli/issues/24213
Currently, the extension and its dependency is installed into `~/.azure/cliextensions`. The dependency error is not shown during installation.

This test also finds four incompatible extentions: `attestation`, `cloud-service`, `functionapp`, `serial-console`. 
To pass the test, add them to `ignore_list` in the script.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -474,6 +474,25 @@ jobs:
   - bash: ./scripts/ci/test_extensions.sh
     displayName: 'Load extensions'
 
+- job: TestExtensionsDependency
+  displayName: Test Extensions Dependency
+  condition: succeeded()
+  timeoutInMinutes: 40
+
+  pool:
+    name: ${{ variables.ubuntu_pool }}
+  strategy:
+    matrix:
+      Python310:
+        python.version: '3.10'
+  steps:
+  - task: UsePythonVersion@0
+    displayName: 'Use Python $(python.version)'
+    inputs:
+      versionSpec: '$(python.version)'
+  - bash: ./scripts/ci/test_extensions_dependency.sh
+    displayName: 'Install extensions'
+
 - job: BuildHomebrewFormula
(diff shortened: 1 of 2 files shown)


yonzhan:
One good point to enhance CI

wangzelin007:
Hi @bebound,
I think it is necessary to find if extension's dependency is conflict with azure-cli.
1. I would like to ask, how did we finally solve the `az interactive` issue?
   Whether to upgrade prompt_toolkit to 3.0.31 or uninstall prompt_toolkit directly, if it is uninstalled, how does the module using prompt_toolkit work?

2. What should we do after this test finding out the conflict in PR?
    Is it up to the PR author to resolve the conflict or upgrade the version of the conflicting package or uninstall the conflict package directly?

3. For this four incompatible extentions, what are we going to do but add them to ignore_list in the script?

bebound:
This test can help us to find the broken extension. How to resolve the conflict, it depends.
1. Make sure the dependency change is necessary and there is no alternative. Otherwise, revert the change.
2. If the extension is well maintained, we can notify the owner to update the extension before next azure-cli release.
3. If the extension is not actively developed, we may choose one of these approaches:
    - Fix extension by ourselves.
    - Add it into ignore list. (I'm not sure if we can warn user that the extension is not compatible with cli)

For your first question, the `interactive` issue is because azure-cli installs 3.0.31 as dependency in 2.41.0. `interactive` also installs 1.0.18 in `~/.azure/cliextension/interactive`, but we always load azure-cli library first (https://github.com/Azure/azure-cli/pull/12778). prompt_toolkit 3.0.31 is loaded and interactive fails to run. In `2.42.0`, we fix this by reverting the changes.

I'll check these four extensions later.

PR #24607, opened 2022-11-10
[Packaging] Use 64-bit Python in Windows build

bebound:
Use 64 bit Python in MSI.
The installation path also changes to `Program Files` from `Program Files (x86)`

There are two advantages.
1. We can update cryptography. Related PR: https://github.com/Azure/azure-cli/pull/25690
2. Possible performance improvement due to 64bit packages.


yonzhan:
Packaging

jiasli:
Closes #24607

build_scripts/windows/Product.wxs
```diff
@@ -61,7 +61,7 @@
 
   <Fragment>
     <Directory Id="TARGETDIR" Name="SourceDir">
-      <Directory Id="ProgramFilesFolder">
+      <Directory Id="ProgramFiles64Folder">
```
heaths, line 64: If you're going to dual-build x86 and x64 MSIs, there's a different pattern you need to use for this. I have examples in PowerShell, OpenSSH, et. al.

heaths (commented):
If you're also picking up 64-bit python files, where are the changes for that? Are all the file locations and names just updated?

bebound:
Close as https://github.com/Azure/azure-cli/pull/26640 is merged

Issue #23951, opened 2022-09-20
Service principal is missing in `az ad group owner list` after upgrade to 2.37


floryut:
## Describe the bug

**Command Name**
`az ad group owner list`

**Errors:**
Missing service principal as owner after upgrade to azcli 2.37+
```
$ az ad group owner list --group c40d6c92-ebf1-46ee-9fd1-99685b281394 --out tsv --query '[].[mail,id]'
user1@company.com        7addc286-717c-4590-9ec9-4559d41725f9
user2@company.com        c2ddd1d1-8ae8-42e6-aef4-2d7dd44bce65
user3@company.com        07930e5c-a4fd-4365-a817-1fb436d7c5c3
user4@company.com        aa62e803-6966-4dff-b7f0-0772bf76b0c9
user5@company.com        df2f9469-acc3-4bf9-993e-dfa6872903fe
```

## To Reproduce:
Steps to reproduce the behavior. Note that argument values have been redacted, as they may contain sensitive information.

- Authenticate to azure subscription
- install azcli 2.37 or higher (includes Microsoft Graph API migration)
- `az ad group owner list --group ExampleGroup`

## Expected Behavior
Service principal is display as owner before 2.37
```
$ az ad group owner list --group c40d6c92-ebf1-46ee-9fd1-99685b281394 --out tsv --query '[].[mail,objectId]'
user1@company.com        7addc286-717c-4590-9ec9-4559d41725f9
user2@company.com        c2ddd1d1-8ae8-42e6-aef4-2d7dd44bce65
user3@company.com        07930e5c-a4fd-4365-a817-1fb436d7c5c3
user4@company.com        aa62e803-6966-4dff-b7f0-0772bf76b0c9
user5@company.com        df2f9469-acc3-4bf9-993e-dfa6872903fe
None    1d0bb3be-f638-4b53-ac11-4ddd39e0d990

$ az version
{
  "azure-cli": "2.32.0",
  "azure-cli-core": "2.32.0",
  "azure-cli-telemetry": "1.0.6",
  "extensions": {
    "account": "0.2.2",
    "alertsmanagement": "0.2.2",
    "log-analytics": "0.2.2",
    "monitor-control-service": "0.3.0",
    "resource-graph": "2.1.0",
    "scheduled-query": "0.5.0"
  }
}
```

## Environment Summary
```
Linux-4.4.0-19041-Microsoft-x86_64-with-glibc2.29, Ubuntu 20.04.4 LTS
[...]

Issue #22664, opened 2022-05-30
`az ad group member list` does not list service principals


kevinharing:
**Related command**
 az ad group member list

**Describe the bug**
`az ad group member list` does not list service principals.

**To Reproduce**
Run `az ad group member list --group <group_name>` on a group with members of type `Service Principal`.

**Expected behavior**
Returns all members of the group, including Service Principals.

**Additional context**
```json
{
  "azure-cli": "2.37.0",
  "azure-cli-core": "2.37.0",
  "azure-cli-telemetry": "1.0.6",
  "extensions": {
    "account": "0.2.1",
    "aks-preview": "0.5.10",
    "application-insights": "0.1.13",
    "datafactory": "0.5.0"
  }
}
```

yonzhan:
@jiasli  for awareness

ncook-hxgn:
bump

We've discovered that if we log into azure cloud shell (azure-cli v2.48.1) as the service principal behind our Azure DevOps service connection, we can't use az cli to `add`, `list`, or `check` group members of a security group, even though our service principal owns the group. (Yields `Insufficient privileges to complete the operation.`)

We've also discovered that if we add a managed identity to said group manually, the managed identity isn't listed as a group member by `az ad group list --group '...'`

```bash
#!/bin/bash
# running as the service principal behind an ADO service connection
$principal="xxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx" # where $principal is a system managed identity from a linux web app
$group="yyyyyyyy-yyyy-yyyy-yyyy-yyyyyyyyyyyy" # where $group is an AAD security group with nobody in it owned by a service principal

echo "##[command]az ad group member check --group \"$group\" --member-id \"$principal\""
is_member=$(az ad group member check --group "$group" --member-id "$principal" | jq -r '.value')
if [ "$is_member" = "false" ]; then
    echo "##[command]az ad group member add --group \"$group\" --member-id \"$principal\""
    az ad group member add --group "$group" --member-id "$principal"
else
    echo "##[warning]principal \"$principal\" is already a member of group \"$group\""
fi

echo "##[command]az ad group member list --group \"$group\""
az ad group member list --group "$group"
```

Shaked:
Is there a plan to fix this issue?

FlorentATo:
Could this be reopened @jiasli ? Thank you

valberico:
[...]

PR #24610, opened 2022-11-10
[Role] EXPERIMENTAL: Add config option `graph.api_version` to override default Microsoft Graph API version

jiasli:
**Related command**
`az ad`

**Description**
Workaround for 
- #23951
- #22664

Due to Microsoft Graph API limitation, service principals are not listed in the result of below commands. This limitation is conveyed in those API documents:

- `az ad group owner list`, which calls [List group owners](https://learn.microsoft.com/en-us/graph/api/group-list-owners?view=graph-rest-1.0&tabs=http) API
  > Note: Currently, service principals are not listed as group owners due to the staged rollout of service principals to the Microsoft Graph v1.0 endpoint.
- `az ad group member list`, which calls [List group members](https://learn.microsoft.com/en-us/graph/api/group-list-members?view=graph-rest-1.0&tabs=http) API
  > Currently service principals are not listed as group members due to staged roll-out of service principals on Graph V1.0 endpoint.

This limitation is solved in `beta` API. However, because `az ad` is a GA command module, it should only call GA APIs internally, but not `beta` versions.

This PR adds an experimental config option `graph.api_version` to override default API version so that users can get unblocked without using the more complex `az rest` workaround.

**Testing Guide**
```
az config set graph.api_version=beta
az ad app show --id 0e040bf6-334f-4412-bcf4-1a82b0ebe11b --debug
az config unset graph.api_version
```

**Additional information**
My first thought was to add config option `graph.use_beta_api`, but we will have to add new options if other API versions are introduced, such as `v2.0` or `alpha`, so I switched to this more general option name.

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/role/_msgrpah/_graph_client.py
+++ b/src/azure-cli/azure/cli/command_modules/role/_msgrpah/_graph_client.py
@@ -28,6 +28,10 @@ class GraphClient:
 
     def __init__(self, cli_ctx):
         self._cli_ctx = cli_ctx
+
+        # EXPERIMENTAL: Use config option graph.api_version to override default API version
+        self._api_version = cli_ctx.config.get('graph', 'api_version', fallback=GraphClient.V1_0)
+
         self._scopes = resource_to_scopes(cli_ctx.cloud.endpoints.microsoft_graph_resource_id)
 
         # https://graph.microsoft.com/ (AzureCloud)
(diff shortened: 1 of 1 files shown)


yonzhan:
add config to unblock customers

nis-spiir:
Any updates on getting this merged?  Seems like a tiny/safe change that could make it much nicer for customers due to the "service principal missing as group members"-bug that has been an issue for half a year?

jiasli:
@nis-spiir, I have discussed internally with MS Graph team. The APIs under `beta` should not be used for production use, as they can change behavior or be removed entirely with no advance notice. 

However, this PR will encourage such usage, so we decide not to merge it.

nis-spiir:
I my mind `api_version=beta` is fairly clear that this should be used with care.
Would it help by renaming the variable to something like `dangerously_override_api_version=beta` ?

I don't think the current solution by having people downgrade `az cli` is a good viable solution.
Any public ETA on this bug being fixed in MS Graph?

I would love to help fixing this in az-cli, but seems like that might is not preferred?

PR #24653, opened 2022-11-15
[Packaging] Add CentOS 7 RPM Package

bebound:
In CentOS 7, Python3.8 can be installed from [Software Collections](https://www.softwarecollections.org/en/).
`rh-python38-python` will retire in Jun 2024. (https://access.redhat.com/support/policy/updates/rhscl-rhel7)
We support this distribution again.

RHEL 7 users can install the package directly.
CentOS 7 users need to enable Software Collections with `yum install centos-release-scl` before installing CLI.

### DIFF
--- a/scripts/release/rpm/azure-cli.spec
+++ b/scripts/release/rpm/azure-cli.spec
@@ -5,11 +5,6 @@
 # Turn off python byte compilation
 %global __os_install_post %(echo '%{__os_install_post}' | sed -e 's!/usr/lib[^[:space:]]*/brp-python-bytecompile[[:space:]].*$!!g')
 
-# .el7.centos -> .el7
-%if 0%{?rhel}
-  %define dist .el%{?rhel}
-%endif
-
 # The Python package name for dnf/yum/tdnf, such as python39, python3
 %define python_package %{getenv:PYTHON_PACKAGE}
 # The Python executable name, such as python3.9, python3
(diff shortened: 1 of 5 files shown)


scripts/release/rpm/ubi.dockerfile
```diff
@@ -17,15 +20,15 @@ COPY . .
...
 FROM ${image} AS execution-env
 
 RUN yum update -y
-RUN yum install -y python39
-
+ARG image
+RUN if [ "$image" == "centos:7" ]; then yum install -y centos-release-scl ; fi
 COPY --from=build-env /azure-cli-dev.rpm ./
-RUN rpm -i ./azure-cli-dev.rpm && \
+RUN yum install -y ./azure-cli-dev.rpm && \
```
bebound, line 33: `rpm -i` does not install dependencies but `yum install` does, so `RUN yum install -y python39` is removed

scripts/release/rpm/ubi.dockerfile
```diff
@@ -6,8 +6,11 @@ ARG image=registry.access.redhat.com/ubi8/ubi:8.4
 FROM ${image} AS build-env
 ARG cli_version=dev
 ARG python_package=python39
+ARG python_cmd=python3.9
 
 RUN yum update -y
+ARG image
```
bebound, line 12: `image` is out of scope, declare again. https://docs.docker.com/engine/reference/builder/#scope

scripts/release/rpm/azure-cli.spec
```diff
@@ -5,11 +5,6 @@
 # Turn off python byte compilation
 %global __os_install_post %(echo '%{__os_install_post}' | sed -e 's!/usr/lib[^[:space:]]*/brp-python-bytecompile[[:space:]].*$!!g')
 
-# .el7.centos -> .el7
-%if 0%{?rhel}
-  %define dist .el%{?rhel}
```
bebound, line 10: In RHEL 7.9 `{?dist}` is `.el7_9`.
In CentOS 7, RHEL 8.4, RHEL 9.0.0, it is `.el\d`.

yonzhan:
Packaging

Issue #24494, opened 2022-10-31
Support Python 3.11


jiasli:
## Description

Python 3.11 has been released on 2022-10-24: https://www.python.org/downloads/release/python-3110/

Azure CLI needs to support Python 3.11 like how we supported 3.9 (https://github.com/Azure/azure-cli/issues/16611), 3.10 (https://github.com/Azure/azure-cli/issues/19857).

## TODOs

- [ ] [Knack](https://github.com/microsoft/knack) supports Python 3.11
  - https://github.com/microsoft/knack/pull/275
- [ ] [azdev](https://github.com/Azure/azure-cli-dev-tools) supports Python 3.11
  - `pylint` (https://github.com/Azure/azure-cli-dev-tools/pull/359)
  - `vcrpy` (same as above)
- [ ] Dependencies are updated to the Python 3.11 compatible version
  - `psycopg2` (https://github.com/Azure/azure-cli/pull/24671)
  - `pywin32` (https://github.com/Azure/azure-cli/pull/24982)
  - `cffi` `psutil` (https://github.com/Azure/azure-cli/pull/26842)
- [ ] CI passes with Python 3.11
  - https://github.com/Azure/azure-cli/pull/24497
  - https://github.com/Azure/azure-cli/pull/26685
  - https://github.com/Azure/azure-cli/pull/25370
  - https://github.com/Azure/azure-cli/pull/26923
    - https://github.com/Azure/azure-cli/pull/26991 
    - https://github.com/Azure/azure-cli/pull/27148
    - https://github.com/Azure/azure-cli/pull/27073

jiasli:
`psycopg2` (introduced by #22459) fails to install on Python 3.11:

```
Collecting psycopg2==2.9.3
  Downloading psycopg2-2.9.3.tar.gz (380 kB)
     ------------------------------------- 380.6/380.6 kB 23.1 MB/s eta 0:00:00
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'error'
  error: subprocess-exited-with-error

  python setup.py egg_info did not run successfully.
  exit code: 1

  [23 lines of output]
  running egg_info
  creating C:\Users\jiasli\AppData\Local\Temp\pip-pip-egg-info-4nrehwws\psycopg2.egg-info
  writing C:\Users\jiasli\AppData\Local\Temp\pip-pip-egg-info-4nrehwws\psycopg2.egg-info\PKG-INFO
[...]

PR #24670, opened 2022-11-17
{Packaging} Update pywin32 to support Python 3.11

bebound:
`pywin32` is not compatible with Python3.11, update it to latest version.

Related issue: #24494


yonzhan:
Update pywin32 to support Python 3.11

PR #24699, opened 2022-11-21
[Backup] `az backup vault`: Add new parameter `--public-network-access` to support enabling public network access for the backup vault

zubairabid:
**Related command**

`az backup vault create`

**Description**

(Tested using the recoveryservices SDK version 2022-09-30-preview, will be released with SDK version [to be confirmed]) The PR adds support for the updated recoveryservices SDK that specifies a `public-network-access` flag for vaults, impacting vault creation/editing/display. Vault display (`az backup vault show`) is handled by the existing code, we need to modify behaviour of vault create/delete (Both handled by `az backup vault create`).

**Testing Guide**

`az backup vault show`

- When querying for a vault, under "properties" we should see the "publicNetworkAccess" field.

`az backup vault create`

- With the updated SDK and without the changes, the vault cannot successfully be created: we get an "InvalidRestApiParameter" issue.
- When creating a new vault, by default the value will be "Enable". After creation we can verify this by checking the properties as specified in `az backup vault show` testing steps
- We can also set the flag manually: if we set it to "Disable", the result will be "publicNetworkAccess: Disabled", and likewise for "Enable".
- This can also be used to edit vaults within certain criteria (check: <https://learn.microsoft.com/en-us/azure/backup/private-endpoints-overview#before-you-start>).

**History Notes**

---


src/azure-cli/azure/cli/command_modules/backup/custom.py
```diff
@@ -112,14 +112,26 @@
...
+                 classic_alerts='Enable', azure_monitor_alerts_for_job_failures='Enable'):
     vault_sku = Sku(name=SkuName.standard)
+    if public_network_access is None:
+        # get the existing value of public_network_access so the request is made correctly
+        try:
+            existing_vault_if_any = client.get(resource_group_name, vault_name)
+            existing_vault_public_network_access = existing_vault_if_any.properties.public_network_access
+            # TODO add better validation for existing_vault_public_network_access? It might be invalid.
+            #   Maybe have a list of possible values and iterate through, and if not add a warning to
+            #   contact support? Such as: if public_network_access in [list], <action>, else <warn user>.
```
zubairabid, line 125: @hiaga Can you check this if possible?

hiaga: is the code same for create and update vault in CLI ? why do you need to check here if the vault already exists ?

currently the allowed values are Enabled, Disabled

zubairabid: Yep, the code for update and create are handled by the same call.

Curious about the (unlikely?) situation for where a vault might _somehow_ have some other value than what's currently allowed. In that case, do we want to add a check here? Or we can operate on the assumption that the situation is unlikely, and if it is then we don't need to handle it as it already exists and this code won't be *introducing* the bug, just continuing it (as this is for when the vault already exists and has said value).

hiaga: as this is create and update flow, we only care what user is allowed to pass while creation/updation. So we restrict the user to only be able to pass Enable/Disable else we throw an error saying invalid input. 

now the case where an invalid value exist from service end. we can choose to handle or not handle it, It's upto us, as per service currently allowed values are E, D.

yonzhan:
Backup

zhoxing-ms:
> E           AssertionError: Can't overwrite existing cassette ('/mnt/vss/_work/1/s/src/azure-cli/azure/cli/command_modules/backup/tests/latest/recordings/test_backup_wl_sql_crr.yaml') in your current record mode ('once').
E           No match for the request (<Request (GET) https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/sql-clitest-rg/providers/Microsoft.RecoveryServices/vaults/sql-clitest-vault?api-version=2022-10-01>) was found.
E           Found 4 similar requests with 1 different matcher(s) :
E           
E           1 - (<Request (GET) https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/sql-clitest-rg/providers/Microsoft.RecoveryServices/vaults/sql-clitest-vault?api-version=2022-04-01>).
E           Matchers succeeded : ['method', 'scheme', 'host', 'port', 'path']

Please note that the api-version of `Microsoft.RecoveryServices` has been upgraded from `2022-04-01` to `2022-10-01`. Please re-record the yaml files of failed tests

zubairabid:
Hi @zhoxing-ms , all checks have passed now. Please approve the merge, thanks.

zhoxing-ms (approved):

PR #24735, opened 2022-11-24
{App Service} `az webapp config container show`: Fix accidental side-effect plus make sure this function handles both types of input data, lists and dict

antimirov:
The current implementation only supports dicts while its unit test explicitely passes a list of dicts as one of 6 cases. So it's actually broken in this regard.

**Related command**
N/A

**Description**
I've accidentally noticed a side-effect in the function `_mask_creds_related_appsettings()` - it's modifying the dictionary it's iterating over, so when it's returning a new dict, in fact the input dict is also being modified, which may lead to some failures later. While fixing it, I also noticed that it's supposed to work with both list or dicts and dicts, while not actually doing it.

In fact, this could be a bug, that this function should have only been used for one type of data, and it just accidentally does not break on lists while doing nothing basically, so difficult to spot the problem.

**Testing Guide**
```shell
# this is the command that I originally used
$ az webapp config container show

$ azdev test LinuxWebappScenarioTest --pytest-args "-n0 -s"
```

**History Notes**
N/A

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -1619,10 +1619,13 @@ def _filter_for_container_settings(cmd, resource_group_name, name, settings,
 
 # TODO: remove this when #3660(service tracking issue) is resolved
 def _mask_creds_related_appsettings(settings):
-    for x in [x1 for x1 in settings if x1 in APPSETTINGS_TO_MASK]:
-        settings[x] = None
-    return settings
+    if isinstance(settings, list):
+        return [
+            {**x, **{'value': None if x['name'] in APPSETTINGS_TO_MASK else x['value']}}
+            for x in settings
+        ]
 
+    return {k: None if k in APPSETTINGS_TO_MASK else v for k, v in settings.items()}
 
 def add_hostname(cmd, resource_group_name, webapp_name, hostname, slot=None):
     from azure.mgmt.web.models import HostNameBinding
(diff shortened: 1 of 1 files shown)


yonzhan:
webapp

zhoxing-ms:
@StrawnSC Could you please help review this PR?

haroonf:
Could you add a test for this case? Want to make sure everything is being correctly masked. Would also help to have a test for lists and dicts input.

antimirov:
@haroonf let me see what I can do

zhoxing-ms:
@antimirov Any update?

zhoxing-ms:
Please note that since we will launch the release of this sprint the day after tomorrow. If you cannot address all the comments tomorrow, we will have to postpone the release of this PR to the next sprint (02-07)

zhoxing-ms:
> Could you add a test for this case? Want to make sure everything is being correctly masked. Would also help to have a test for lists and dicts input.

@antimirov Please note that since we will launch the release of this sprint the day after tomorrow. If you cannot address all the comments tomorrow, we will have to postpone the release of this PR to the next sprint **(03-07)**

antimirov:
@zhoxing-ms Sorry guys, I've been busy with some other tasks at my job. I could not finish any of the tests for this PR on time and I don't know when I'll be able to do it. No ETA at the moment. Maybe someone can help here? If you see the code change, you can judge its value. And whether reject this PR or not it's up to you. Apparently, when this code was produced, no tests were required and it was fine back then.

zhoxing-ms:
@antimirov Ok, I postpone the release of this PR to the next sprint. You can refer to document [authoring_tests.md](https://github.com/Azure/azure-cli/blob/dev/doc/authoring_tests.md) to write the tests

@haroonf Could you help antimirov add some relevant tests?

teresaritorto:
Moving this back to Draft for now since this PR has been stale for quite some time and hasn’t seen any recent updates.

Issue #9157, opened 2019-04-19
Tab completion is incredibly slow and picky about formatting


jboulter11:
## Describe the bug
Tab completion is painfully slow.  See repro.

**Command Name**
`az`

**Errors:**

## To Reproduce:

1. Properly configure tab completion for your shell, I'm in zsh but this all holds true in bash as well.
1. Type `az` in the terminal
1. Hit tab, it adds a space instead of giving you tab completion.
1. Hit tab again, wait 1-2 seconds for tab completion to process
1. After a wait, tab completion displays a list of completions.

Another example:
1. Type 'az devops' or 'az boards' in the terminal
1. Hit tab, wait ~5 seconds
1. Get -- no matches found -- output.
1. Hit space, hit tab
1. Wait ~1s get a bunch of completions

## Expected Behavior
A response should be given immediately with no delay and it should not be picky about whether there is a space after the last command.

The entire autocompletion process is way too slow.  Type git and hit tab and you get an _instant_ response with completions with nice formatting.  This has got to be better in the azure cli.  1-2s is too long, but 5 seconds is just shameful especially having it return no results after, only to work after hitting space.

Please let me know if you need any more information to help prioritize this work, feel free to reach out to me via email or teams via my alias: jiboulte.

## Environment Summary
```
Darwin-18.2.0-x86_64-i386-64bit
Python 3.7.3
Shell: bash

azure-cli                         2.0.62

Extensions:
azure-devops                       0.4.0

Python location '/usr/local/Cellar/azure-cli/2.0.62/libexec/bin/python'
Extensions directory '/Users/jiboulte/.azure/cliextensions'

Python (Darwin) 3.7.3 (default, Apr 19 2019, 09:07:31) 
[Clang 10.0.1 (clang-1001.0.46.3)]

```

limingu:
It can't repro on Windows. It may be Mac/Linux only issue.

limingu:
@jboulter11 , can you try other commands (not az) to see if it has the same issue. I think it may the OS issue not az command issue.

jboulter11:
Other commands provide an instant response.  `az` and its nested commands are what is slow, my OS is fine.

adewaleo:
@tjprescott, @limingu, I would be interested in this post build. We should try and imitate powershell. Maddie once mentioned how they cache tab completions...

yonzhan:
add to S164.

nickamon:
[...]

PR #24765, opened 2022-11-30
{Core} Improve autocomplete performance by preventing loading all modules

### LINKED ISSUE #24762: Autocomplete is slow when command is not complete

**Describe the bug**

If I press TAB after `az account cle`, it takes 4seconds to show suggestion `az account clear`. However `az account -h` only takes 2seconds.

The debug log shows that it reloads modules and extensions when run `az account cle --debug`.
```
cli.azure.cli.core: Could not find a match in the command or command group table for 'account cle'. The index may be outdated.
cli.azure.cli.core: Loading command modules:
```

I think we can make reload rules stricter. When ` _ARGCOMPLETE=1`, if `account cle` is not found, check if `account` is matched. If still not matched, reload.

Related issue: https://github.com/Azure/azure-cli/issues/9157

bebound:
**Description**

Close #24762, related issue #9157, related pr #13294
If the command is not complete, for example: `az account sho`, press TAB, cli loads all modules, which takes 1s. It gets worse if user installs many extensions. This make autocomplete slow.
This PR prevents module loading in autocomplete mode.

If command is `az acc`:
1. module begin with `acc` will be returned in `CommandIndex.get()` .
2. `raw_cmd` is 'acc'. Trim last word, it becomes `''`, return `self.command_table`.
```
cli.azure.cli.core: In autocomplete mode, load command begin with: 'acc'
cli.azure.cli.core: Modules found from index for 'acc': ['azure.cli.command_modules.profile', 'azure.cli.command_modules.resource']

cli.azure.cli.core: Could not find a match in the command or command group table for 'acc'
cli.azure.cli.core: In autocomplete mode, try to match trimmed raw cmd: ''
cli.azure.cli.core: Trimmed raw cmd is empty, return command table.
```

If command is `az account sh`:
1. `account` module is returned in `CommandIndex.get()` .
2. `raw_cmd` is `account sh`. Trim last word, it becomes `account` and is in `self.command_group_table`, return `self.command_table`.
```
cli.azure.cli.core: Modules found from index for 'account': ['azure.cli.command_modules.profile', 'azure.cli.command_modules.resource']

cli.azure.cli.core: Could not find a match in the command or command group table for 'account sh'
cli.azure.cli.core: In autocomplete mode, try to match trimmed raw cmd: 'account'
cli.azure.cli.core: Found a match in the command group table for trimmed raw cmd: 'account'.
```

**Testing Guide**
`az acco` and `az account sh` should be faster.
```
$env:_ARGCOMPLETE=1;
Measure-Command {az account sh}
TotalMilliseconds : 567.0021

Remove-Item Env:\_ARGCOMPLETE;
Measure-Command {az account sh}
TotalMilliseconds : 2935.8341
```

---


yonzhan:
Core

src/azure-cli-core/azure/cli/core/__init__.py
```diff
@@ -555,6 +568,15 @@ def get(self, args):
...
         index_modules_extensions = index.get(top_command)
+        if not index_modules_extensions and is_autocomplete():
+            # If user type `az acco`, command begin with `acco` will be matched.
+            # If the user just installed a new extension which contains a command named `acco`, he can't get the
+            # expected suggestion. But it's a rare scenario, and it's okay to ignore.
+            logger.debug("In autocomplete mode, load command begin with: '%s'", top_command)
+            index_modules_extensions = []
+            for command in index:
+                if command.startswith(top_command):
+                    index_modules_extensions += index[command]
```
jiasli, line 579: Maybe put these lines after L597 for simplicity?

src/azure-cli-core/azure/cli/core/__init__.py
```diff
@@ -555,6 +570,15 @@ def get(self, args):
         # Check the command index for (command: [module]) mapping, like
         # "network": ["azure.cli.command_modules.natgateway", "azure.cli.command_modules.network", "azext_firewall"]
         index_modules_extensions = index.get(top_command)
+        if not index_modules_extensions and is_autocomplete():
+            # If user type `az acco`, command begin with `acco` will be matched.
+            # If the user just installed a new extension which contains a command named `acco`, he can't get the
+            # expected suggestion. But it's a rare scenario, and it's okay to ignore.
```
bebound, line 576: This does not happen as `CommandIndex().invalidate()` is called after extension installation/remove/update.

src/azure-cli-core/azure/cli/core/__init__.py
bebound: I put it here as the following code convert `index_modules_extensions` to `index_builtin_modules` and `index_extensions`

src/azure-cli-core/azure/cli/core/__init__.py
```diff
@@ -555,6 +569,13 @@ def get(self, args):
         # Check the command index for (command: [module]) mapping, like
         # "network": ["azure.cli.command_modules.natgateway", "azure.cli.command_modules.network", "azext_firewall"]
         index_modules_extensions = index.get(top_command)
+        if not index_modules_extensions and self.cli_ctx.data['completer_active']:
+            # If user type `az acco`, command begin with `acco` will be matched.
+            logger.debug("In autocomplete mode, load command begin with: '%s'", top_command)
```
jiasli, line 574:
```suggestion
            logger.debug("In autocomplete mode, load commands starting with: '%s'", top_command)
```

Let's be consistent with Python's naming `startswith`: https://docs.python.org/3/library/stdtypes.html#str.startswith

src/azure-cli-core/azure/cli/core/__init__.py
```diff
@@ -449,6 +449,19 @@ def _get_extension_suppressions(mod_loaders):
                 if raw_cmd in self.command_group_table:
                     logger.debug("Found a match in the command group table for '%s'.", raw_cmd)
                     return self.command_table
+                if self.cli_ctx.data['completer_active']:
+                    # If the command is not complete in autocomplete mode, we should match shorter command.
+                    # For example, `account sho` should match `account`.
+                    logger.debug("Could not find a match in the command or command group table for '%s'", raw_cmd)
+                    raw_cmd = ' '.join(raw_cmd.split()[:-1])
```
jiasli, line 456: Please avoid reusing existing name. Instead, introduce a new name to make the code easier to read.

jiasli (approved):

Issue #18298, opened 2021-05-30
Remove colors from Azure CLI text output


Dragas:
I am appalled by the usage of colors to denote types of output in azure cli on bash. Having light background in terminals makes it impossible to read what cli is outputting.
![image](https://user-images.githubusercontent.com/6078508/120118834-4bc28e80-c19d-11eb-9738-f6084b01a644.png)

yonzhan:
@jiasli for awareness

jiasli:
You may turn off color with 

```
az config set core.no_color=true
```

To turn color back on, use

```
az config unset core.no_color
```

You may also switch to the experimental **light** theme:

```
az config set core.theme=light
```

To reset to the default **dark** theme, use

```
az config unset core.theme
```

This feature is not yet documented.

Dragas:
Why is this feature enabled by default to begin with?

jiasli:
Most users of Azure CLI are using a standard dark theme terminal to invoke Azure CLI. We are aiming to provide better readability on a dark theme terminal.

Previously the color for warning is **yellow** and we changed it to **bright yellow** because of the low accessibility of yellow on dark theme terminals, according to user experience study (https://github.com/Azure/azure-cli/pull/16258).

If bright yellow is not readable on PyCharm, this problem should be reported to PyCharm as well to make the visibility of bright yellow better. 

We will also evaluate your feedback and make better documentation regarding this topic as well.

pamelafox:
I am experiencing this issue as well, it is nearly impossible to read the text. I didn't even realize there'd be a command to change the colors, as I've never had to do that with other CLI tools, so I didn't go looking. I did just do a web search now for what I might have searched, and it looks like I wouldn't have found the right command anyway.

I am on a Mac OS X, which defaults to a white terminal, so I think many Mac users would be affected, especially those that are new to develop and environment customization.

Is there some way that you can detect the terminal mode and change the colors accordingly? In my terminal, "defaults read com.apple.Terminal " does give back a JSON of settings with the colors, though you'd have to wrap that in a guard since it wouldn't work cross-OS.

Or perhaps if you detected someone was on a Mac, you could suggest the configure command the first time they run an az command.

charris-msft:
@yonzhan - should we open a new issue to track @pamelafox's feedback about default experience on Mac?

jiasli:
[...]

PR #24804, opened 2022-12-05
{Style} Use light theme by default on MacOS

jiasli:
## Issue

We specify the color for warnings as [ANSI color](https://en.wikipedia.org/wiki/ANSI_escape_code#Colors) **Bright Yellow** without any specification of the actual RGB value:

https://github.com/Azure/azure-cli/blob/8d8d71707d6c60083d008062667cab18d0adaa25/src/azure-cli-core/azure/cli/core/style.py#L73

According to https://github.com/Azure/azure-cli/issues/18298#issuecomment-1263529261, MacOS's terminal by default uses white background and **Bright Yellow** is not readable on that white background.

Something like:

image

Personally, I think it should the terminal's theme's responsibility to make all ANSI colors readable on its background.

## Change

Since MacOS's terminal by default uses white background, we by default uses `light` theme which uses (dark) **Yellow** for warnings:

https://github.com/Azure/azure-cli/blob/8d8d71707d6c60083d008062667cab18d0adaa25/src/azure-cli-core/azure/cli/core/style.py#L85

## Cons

However, a disadvantage is that if the user manually chooses a dark background, or use an IDE like PyCharm with dark theme. There may be other readability issues if we default to `light` theme, such as blue, magenta. 

As there isn't a way to detect the background color, there is no optimal solution. Users still need to run `az config set core.theme=light/dark` to select the color theme or totally disable color with `az config set core.no_color=true`.

### DIFF
--- a/src/azure-cli-core/azure/cli/core/__init__.py
+++ b/src/azure-cli-core/azure/cli/core/__init__.py
@@ -171,11 +171,21 @@ class AzCli(CLI):
     def _configure_style(self):
         from azure.cli.core.util import in_cloud_console
         from azure.cli.core.style import format_styled_text, get_theme_dict, Style
+        import platform
 
-        # Configure Style
+        # Most terminals, like those on Windows, Ubuntu use dark background
+        default_theme = 'dark'
+
+        if in_cloud_console():
+            # Cloud Shell has its own RGB-defined theme
+            default_theme = 'cloud-shell'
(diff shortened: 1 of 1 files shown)


jiasli:
https://stackoverflow.com/questions/58186391/text-displaying-as-bright-yellow-in-terminal

This is the interface for configuring ANSI color's real rendered RGB:

PR #24844, opened 2022-12-08
[Postgres Flexible Server] Update Postgres Flexible server help text - replace single quote with double quote

shliderm:
**Related command**
az postgres flexible-server create --help

**Description**
Replace the single quote with double quote in the --help text for postgres flexible server
when running the command with single quote the command will fail and having single quate in the --help text is confusing for the cli consumer.

**Testing Guide**
az postgres flexible-server create --help


yonzhan:
postgres

PR #24859, opened 2022-12-10
[App Config] `az appconfig snapshot`: New command group to support app configuration snapshots

albertofori:
**Related command**

`az appconfig snapshot`

**Description**

Azure app configuration is adding support for App Configuration Snapshots. This PR adds support for creating, listing, archiving and recovering snapshots.

Changes in this pull request -

- Add new command group `snapshot` - `az appconfig snapshot`.
- Add new command `az appconfig snapshot create` to create a snapshot.
- Add new command `az appconfig snapshot show` to show the properties of an app configuration snapshot.
- Add new command `az appconfig snapshot list` to list snapshots of a given app configuration.
- Add new command `az appconfig snapshot archive` to archive a snapshot.
- Add new command `az appconfig snapshot recover` to recover an archived snapshot.
- Update command `az appconfig kv list` to add new parameter `snapshot` used to list key-values in a snapshot.

**Testing Guide**

- Create a new App Configuration store.
`az appconfig create -g MyResourceGroup  -n MyAppConfiguration -l westus --sku Standard`

- Create a new snapshot for the App Configuration.
`az appconfig snapshot create -n MyAppConfiguration -s MySnapshot --filters '{\"key\": \"app1*\", \"label\": \"dev\"}'`

- Show properties of a snapshot in App Configuration.
`az appconfig snapshot show -n MyAppConfiguration -s MySnapshot`

- List the snapshots of an App Configuration.
`az appconfig snapshot list -n MyAppConfiguration`

- Archive a snapshot.
`az appconfig snapshot archive -n MyAppConfiguration -s MySnapshot`

- Recover a snapshot.
`az appconfig snapshot recover -n MyAppConfiguration -s MySnapshot`

- List key-values.
`az appconfig kv list -n MyAppConfiguration -s MySnapshot`

**History Notes**

[App Config] `az appconfig snapshot create`: Add new command to support creating a snapshot
[App Config] `az appconfig snapshot show`: Add new command to support showing the properties of an app configuration snapshot
[App Config] `az appconfig snapshot list`: Add new command to support listing snapshots of a given app configuration
[App Config] `az appconfig snapshot archive`: Add new command to support archiving a snapshot
[App Config] `az appconfig snapshot recover`: Add new command to support recovering an archived snapshot

---


yonzhan:
App Config

src/azure-cli/azure/cli/command_modules/appconfig/commands.py
```diff
@@ -50,6 +51,11 @@ def load_command_table(self, _):
         client_factory=cf_replicas
     )
 
+    configstore_snapshot_util = CliCommandType(
+        operations_tmpl='azure.cli.command_modules.appconfig.snapshot#{}',
+        table_transformer=configstore_snapshot_output_format
+    )
```
jimmyca15, line 57: How come this one doesn't have client factory like all the others?

src/azure-cli/azure/cli/command_modules/appconfig/_validators.py
```diff
@@ -305,3 +316,41 @@ def validate_export_as_reference(namespace):
...
+                if not isinstance(key_filter_value, str) or len(key_filter_value) < 1:
+                    raise InvalidArgumentValueError("Invalid key filter value. Value must be a non-empty string.")
+
+                if parsed_filter.get("label", None) and not isinstance(parsed_filter["label"], str):
+                    raise InvalidArgumentValueError("Label filter must be a string if specified.")
+
+                filter_parameters.append(parsed_filter)
+
+            except ValueError:
+                raise InvalidArgumentValueError("Parameter must be an escaped JSON. {} is not a valid JSON object.".format(filter_param))
```
jimmyca15, line 354: > Parameter must be an escaped JSON.

I suppose that should say "Parameter must be an escaped JSON object."

src/azure-cli/azure/cli/command_modules/appconfig/commands.py
albertofori: The client factories create a Python SDK client for the given command. Since there is no SDK client implementation for snapshots yet, the client is implemented in `_snapshot_custom_client.py` so this is directly initialized where it is used. This will be updated to include the corresponding client factory when available.

src/azure-cli/azure/cli/command_modules/appconfig/_snapshot_custom_client.py
```diff
@@ -0,0 +1,540 @@
...
+    if_none_match=None,
+    sync_token=None,
+    **kwargs
+):
+    _headers = case_insensitive_dict(kwargs.pop("headers", {}) or {})
+    _params = case_insensitive_dict(kwargs.pop("params", {}) or {})
+
+    api_version = kwargs.pop("api_version", SnapshotConstants.API_VERSION)
+    accept = _headers.pop(
+        "Accept", "application/vnd.microsoft.appconfig.keyset+json, application/json, application/problem+json"
```
jimmyca15, line 60: `application/vnd.microsoft.appconfig.keyset+json` that should only be for the '/keys' endpoint.

"application/json, application/problem+json" should be enough for this purposes.

src/azure-cli/azure/cli/command_modules/appconfig/_snapshot_custom_client.py
```diff
@@ -0,0 +1,540 @@
...
+        params=_params,
+        **kwargs
+    )
+
+
+def _convert_request(request, files=None):
+    data = request.content if not files else None
+    request = HttpRequest(method=request.method, url=request.url, headers=request.headers, data=data)
+    if files:
+        request.set_formdata_body(files)
```
jimmyca15, line 231: We don't use form data for any of our endpoints.

src/azure-cli/azure/cli/command_modules/appconfig/_snapshot_custom_client.py
```diff
@@ -0,0 +1,540 @@
...
+        """
+        timeout = kwargs.pop("timeout", 30)
+        polling_interval = kwargs.pop("polling_interval", 5)
+
+        from datetime import datetime
+
+        def _get_elapsed_time(start_time: datetime):
+            return (datetime.now() - start_time).total_seconds()
+
+        print("Starting...")
```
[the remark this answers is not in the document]
albertofori: This prints to the user. Since the method waits to ensure that the snapshot status is "ready" before completing, the idea is to make sure it does not appear as though the cli is frozen. So there is a "Starting", then "Provisioning" update.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -43,6 +43,12 @@ def load_arguments(self, _):
         validator=validate_feature_query_fields,
         arg_type=get_enum_type(['name', 'key', 'label', 'locked', 'last_modified', 'state', 'description', 'conditions'])
     )
+    snapshot_fields_arg_type = CLIArgumentType(
+        nargs='+',
+        help='Customize output fields for Snapshots',
+        validator=validate_snapshot_query_fields,
+        arg_type=get_enum_type(['name', 'etag', 'retention_period', 'filters', 'status', 'created', 'expires', 'size', 'items_count', 'size', 'items_link'])
```
jimmyca15, line 50: status_code missing.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -79,6 +85,20 @@ def load_arguments(self, _):
...
+        type=str,
+        help='Name of the App Configuration snapshot.',
+        configured_default=None
+    )
+
+    snapshot_filter_arg_type = CLIArgumentType(
+        options_list=['--filters'],
+        validator=validate_snapshot_filters,
+        nargs='+',
+        help='Space-separated key and label filters used to build an app configuration snapshot. These values should escaped JSON objects.'
```
jimmyca15, line 99: > These values should escaped JSON objects

These values should **be** escaped JSON objects

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -326,3 +347,27 @@ def load_arguments(self, _):
 
     with self.argument_context('appconfig replica create') as c:
         c.argument('location', arg_type=get_location_type(self.cli_ctx), help='Location at which to create the replica.')
+
+    with self.argument_context('appconfig snapshot create') as c:
+        c.argument('snapshot_name', arg_type=snapshot_name_arg_type)
+        c.argument('filters', arg_type=snapshot_filter_arg_type)
+        c.argument('composition_type', options_list=['--composition-type'], arg_type=get_enum_type(["all", "group_by_key"]), help='Composition type used in building app configuration snapshots.')
+        c.argument('retention_period', options_list=['--retention-period'], type=int, help='Duration in seconds for which a snapshot can remain archived before expiry. If not specified, retention period defaults to the equivalent of 30 days. (2592000s)')
+        c.argument('tags', arg_type=tags_type, help="Space-separated tags: key[=value] [key[=value] ...].")
```
jimmyca15, line 356: it's only 30 days for standard tier. 7 days for free. I'd omit the actual value and just mention it's optional.

src/azure-cli/azure/cli/command_modules/appconfig/_help.py
```diff
@@ -590,3 +592,66 @@
...
+    type: command
+    short-summary: Create an app configuration snapshot.
+    examples:
+        - name: Create a snapshot MySnapshot of all keys starting with 'Test' in configuration store MyAppConfiguration.
+          text:
+            az appconfig snapshot create -s MySnapshot -n MyAppConfiguration --filters '{\"key\":\"Test*\"}'
+        - name: Create a snapshot MySnapshot of all keys starting with 'abc' and a retention period of 5 minutes.
+          text:
+            az appconfig snapshot create -s MySnapshot -n MyAppConfiguration --filters '{\"key\":\"abc*\"}' --retention-period 300
+        - name: Create a snapshot of all keys starting with 'app/' with the label 'dev' as default, and override the key-values with keys with the label 'prod' if they exist.
```
jimmyca15, line 612: I think the more common example, is the first filter having no label, and then the second filter overriding with "prod"

jimmyca15:
Couple comments. Overall looks great.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
albertofori: Got it, however, I would want to know if there is an upper limit for the number of days a snapshot can be retained.

jimmyca15: Upper limit is 7 days for free store, 90 for standard.

albertofori: I am not sure about this but would it be worth mentioning the limits as is done for config store retention period as shown below? 
image

jimmyca15: I don't see a problem with adding it, but needs to mention the difference between standard and free.

src/azure-cli/azure/cli/command_modules/appconfig/_snapshot_custom_client.py
albertofori: This has been updated, thanks!

src/azure-cli/azure/cli/command_modules/appconfig/_snapshot_custom_client.py
albertofori: Got it! This has been updated.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
albertofori: Thanks for pointing this out. This has been added.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
albertofori: My bad. Thanks! This has been fixed.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
albertofori: Got it. The help message has been updated accordingly.

src/azure-cli/azure/cli/command_modules/appconfig/_help.py
albertofori: This has been updated.

src/azure-cli/azure/cli/command_modules/appconfig/_validators.py
albertofori: This has been updated.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -326,3 +347,27 @@ def load_arguments(self, _):
 
     with self.argument_context('appconfig replica create') as c:
         c.argument('location', arg_type=get_location_type(self.cli_ctx), help='Location at which to create the replica.')
+
+    with self.argument_context('appconfig snapshot create') as c:
+        c.argument('snapshot_name', arg_type=snapshot_name_arg_type)
+        c.argument('filters', arg_type=snapshot_filter_arg_type)
+        c.argument('composition_type', options_list=['--composition-type'], arg_type=get_enum_type(["all", "group_by_key"]), help='Composition type used in building app configuration snapshots.')
+        c.argument('retention_period', options_list=['--retention-period'], type=int, help='Duration in seconds for which a snapshot can remain archived before expiry. A snapshot can be archived for a maximum of 7 days (604,800s) for free tier stores and 90 days (7,776,000s) for standard tier stores.')
```
zhoxing-ms, line 355: It is no need to repeat the declaration of the default option name
```suggestion
        c.argument('composition_type', arg_type=get_enum_type(["all", "group_by_key"]), help='Composition type used in building app configuration snapshots.')
        c.argument('retention_period', type=int, help='Duration in seconds for which a snapshot can remain archived before expiry. A snapshot can be archived for a maximum of 7 days (604,800s) for free tier stores and 90 days (7,776,000s) for standard tier stores.')
```

zhoxing-ms:
> History Notes
[App Config] az appconfig snapshot: Add new command group to support app configuration snapshots

@albertofori We usually recommend that you write the specific description for each new command in the history notes

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -326,3 +347,27 @@ def load_arguments(self, _):
 
     with self.argument_context('appconfig replica create') as c:
         c.argument('location', arg_type=get_location_type(self.cli_ctx), help='Location at which to create the replica.')
+
+    with self.argument_context('appconfig snapshot create') as c:
+        c.argument('snapshot_name', arg_type=snapshot_name_arg_type)
+        c.argument('filters', arg_type=snapshot_filter_arg_type)
+        c.argument('composition_type', options_list=['--composition-type'], arg_type=get_enum_type(["all", "group_by_key"]), help='Composition type used in building app configuration snapshots.')
```
zhoxing-ms, line 354: For enumerated values, we usually recommend using camel case instead of snake case
```suggestion
        c.argument('composition_type', options_list=['--composition-type'], arg_type=get_enum_type(["All", "GroupByKey"]), help='Composition type used in building app configuration snapshots.')
```

zhoxing-ms:
Please note that since we will launch the release of this sprint the day after tomorrow. If you cannot address all the comments tomorrow, we will have to postpone the release of this PR to the next sprint (02-07)

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
albertofori: @zhoxing-ms Snake case was used to maintain consistency with other enumerated value options like selecting [field parameters in output](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/appconfig/_params.py#:~:text=arg_type%3Dget_enum_type(%5B%27key%27%2C%20%27value%27%2C%20%27label%27%2C%20%27content_type%27%2C%20%27etag%27%2C%20%27tags%27%2C%20%27locked%27%2C%20%27last_modified%27%5D)). Since users already specify other options, I believe this would be in line with the usage pattern.

albertofori:
> > History Notes
> > [App Config] az appconfig snapshot: Add new command group to support app configuration snapshots
> 
> @albertofori We usually recommend that you write the specific description for each new command in the history notes

@zhoxing-ms Kindly let me know if the format of the updated History notes works. Thanks!

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
zhoxing-ms: OK, make senses

zhoxing-ms:
> Kindly let me know if the format of the updated History notes works. Thanks!

 The modified format looks good. I have improved it on your basis

zhoxing-ms:
@albertofori Could you please resolve these CI issues?

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -79,6 +85,20 @@ def load_arguments(self, _):
...
+        type=str,
+        help='Name of the App Configuration snapshot.',
+        configured_default=None
+    )
+
+    snapshot_filter_arg_type = CLIArgumentType(
+        options_list=['--filters'],
+        validator=validate_snapshot_filters,
+        nargs='+',
+        help='Space-separated key and label filters used to build an app configuration snapshot. These values should be escaped JSON objects.'
```
avanigupta, line 99: I wasn't sure what the filters should look like just by reading this message. It's a complicated input format that users may need to figure out by looking at usage samples anyway. 
However, I think this may be slightly clearer:
"Space-separated list of escaped JSON objects that represent the key and label filters used to build an App Configuration snapshot."
What do you think?

src/azure-cli/azure/cli/command_modules/appconfig/_snapshot_custom_client.py
```diff
@@ -0,0 +1,513 @@
...
+
+        from datetime import datetime
+        from knack.log import get_logger
+        
+        def _get_elapsed_time(start_time: datetime):
+            return (datetime.now() - start_time).total_seconds()
+
+        logger = get_logger(__name__)
+        
+        logger.warning("Starting...")
```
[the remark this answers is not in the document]
albertofori: This was tried however, the logged data was not shown in the cli as the default verbosity of the cli logs above the level of "information".  I used `warning` according to the guidelines [here ](https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md#coding-practices:~:text=Log%20to%20logger.error()%20or%20logger.warning()%20for%20user%20messages%3B%20do%20not%20use%20the%20print()%20function) since directly using the `print` function is discouraged.

src/azure-cli/azure/cli/command_modules/appconfig/snapshot.py
```diff
@@ -0,0 +1,193 @@
...
+
+    try:
+        snapshot = _client.get_snapshot(snapshot_name, fields=query_fields)
+
+        if not query_fields:
+            return snapshot
+
+        partial_snapshot = {}
+
+        for field in query_fields:
```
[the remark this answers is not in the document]
albertofori: Yes please, if this is not done, all the fields that were not indicated in the filter are shown with null values.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -326,3 +347,27 @@ def load_arguments(self, _):
 
     with self.argument_context('appconfig replica create') as c:
         c.argument('location', arg_type=get_location_type(self.cli_ctx), help='Location at which to create the replica.')
+
+    with self.argument_context('appconfig snapshot create') as c:
+        c.argument('snapshot_name', arg_type=snapshot_name_arg_type)
+        c.argument('filters', arg_type=snapshot_filter_arg_type)
+        c.argument('composition_type', arg_type=get_enum_type(["all", "group_by_key"]), help='Composition type used in building app configuration snapshots.')
```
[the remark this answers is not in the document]
albertofori: Specifying the enum options validates that the supplied value is one of the supplied options and will error out as shown below.
image

Are there additional validations I may be missing here?

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -326,3 +347,27 @@ def load_arguments(self, _):
...
+        c.argument('retention_period', type=int, help='Duration in seconds for which a snapshot can remain archived before expiry. A snapshot can be archived for a maximum of 7 days (604,800s) for free tier stores and 90 days (7,776,000s) for standard tier stores.')
```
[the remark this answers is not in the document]
albertofori: Yes please, retention period is expected to be in seconds.

src/azure-cli/azure/cli/command_modules/appconfig/_snapshotmodels.py
```diff
@@ -0,0 +1,147 @@
...
+    :ivar int size:
+        The size in bytes of the built snapshot.
+    :ivar int items_count:
+        The number of key-values in the built snapshot.
+    :ivar dict[str, str] tags:
+        Dictionary of tags of the snapshot.
+    :ivar str items_link:
+        Endpoint that can be used to query keys-values in a snapshot
+    :ivar int retention_period:
+        Number of days for which an archived snapshot will be kept before being deleted.
```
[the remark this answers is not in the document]
albertofori: This should be seconds, thanks for pointing this out! This has been updated.

src/azure-cli/azure/cli/command_modules/appconfig/snapshot.py
```diff
@@ -0,0 +1,193 @@
...
+        for field in fields:
+            if field == SnapshotQueryFields.ALL:
+                query_fields.clear()
+                break
+            query_fields.append(field.name.lower())
+
+    _client = _get_snapshot_client(cmd, name, connection_string, auth_mode, endpoint)
+
+    try:
+        snapshot = _client.get_snapshot(snapshot_name, fields=query_fields)
```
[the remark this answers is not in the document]
albertofori: You are right, this has been moved above the filter logic as suggested. Thanks!

src/azure-cli/azure/cli/command_modules/appconfig/snapshot.py
```diff
@@ -0,0 +1,193 @@
...
+    query_fields = []
+
+    if fields:
+        for field in fields:
+            if field == SnapshotQueryFields.ALL:
+                query_fields.clear()
+                break
+            query_fields.append(field.name.lower())
+
+    _client = _get_snapshot_client(cmd, name, connection_string, auth_mode, endpoint)
```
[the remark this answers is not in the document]
albertofori: This has been updated, thanks!

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
albertofori: I believe your suggestion explains it more clearly, thanks! This has been updated accordingly.

src/azure-cli/azure/cli/command_modules/appconfig/_snapshotmodels.py
```diff
@@ -0,0 +1,147 @@
...
+            filters=data_dict.get("filters", None),
+            etag=data_dict.get("etag", None),
+            composition_type=data_dict.get("composition_type", None),
+            created=data_dict.get("created", None),
+            expires=data_dict.get("expires", None),
+            size=data_dict.get("size", 0),
+            items_count=data_dict.get("items_count", 0),
+            tags=data_dict.get("tags", {}),
+            items_link=data_dict.get("items_link", None),
+            retention_period=data_dict.get("retention_period", 0)
```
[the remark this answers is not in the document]
albertofori: I guess the default value should be null here since the value is not actually 0. The defaults shown here are  assigned to the snapshot object in the event where the returned JSON from a request does not contain a given property. Eg., when fields are specified.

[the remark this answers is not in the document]
avanigupta: I think null is better than 0, because 0 is a valid value for retention period

avanigupta (approved):

src/azure-cli/azure/cli/command_modules/appconfig/_snapshotmodels.py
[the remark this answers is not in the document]
albertofori: Got it, I guess same could be done for the `items_count` and `size` too.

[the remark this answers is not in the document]
albertofori: This has been updated, thanks!

albertofori:
> @albertofori Could you please resolve these CI issues?

Hello @zhoxing-ms Please we can hold off from merging these changes as the concerned service is yet to go to preview and endpoints are not available in the public cloud yet. This is why unit tests are currently failing.

src/azure-cli/azure/cli/command_modules/appconfig/_snapshot_custom_client.py
```diff
@@ -0,0 +1,595 @@
...
+            if_none_match=if_none_match,
+            **kwargs
+        )
+
+        current_state = self._get_snapshot_create_status(name)
+
+        start_time = datetime.now()
+        while current_state.operation_status.status == ProvisioningStatus.RUNNING:
+            progress.spinner.step(label="Provisioning")
+            if _get_elapsed_time_from(start_time) > timeout:
```
jimmyca15, line 322: Looks like this is a 30 second timeout. Seems a bit short. Is there a way for the customer to set timeout? If not, we should at least allow a long one like ten minutes at least. But, I don't see the point for a built-in timeout on the client side as long as the operation is still running.

src/azure-cli/azure/cli/command_modules/appconfig/_snapshotmodels.py
```diff
@@ -0,0 +1,222 @@
...
+        if not data_dict:
+            return None
+
+        return cls(
+            code=data_dict.get("code", None),
+            message=data_dict.get("message", None)
+        )
+
+
+class SnapshotCreateData:
```
jimmyca15, line 199: I'd recommend to generalize this to OperationStatusResult.

jimmyca15: Or OperationStatusResponse.

src/azure-cli/azure/cli/command_modules/appconfig/_snapshot_custom_client.py
```diff
@@ -0,0 +1,595 @@
...
+        path_arguments={"name": name},
+        if_match=if_match,
+        if_none_match=if_none_match,
+        sync_token=sync_token,
+        params=_params,
+        **kwargs
+    )
+
+
+def build_get_snapshot_create_status(
```
jimmyca15, line 110: Ideally we should have a generic "get operation status" method.

jimmyca15: If we have to special case things for snapshots, then it may indicate that we didn't properly implement the long-running-operation model.

albertofori: he implementations for these are embedded in individual classes and there are no standalone utility functions for this. The implementation was therefore based on the method outlined [here](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/core/azure-core/azure/core/polling/base_polling.py#:~:text=def%20request_status(self%2C%20status_link)%3A)

jimmyca15 (dismissed):

src/azure-cli/azure/cli/command_modules/appconfig/_format.py
```diff
@@ -174,6 +185,23 @@ def _get_value(item, *args):
...
 
 
+def _get_value_by_names(item, property_names):
+    """
+        CLI core converts KeyValue object field names to camelCase (eg: content_type becomes contentType)
+        But when customers specify field filters, we return a dict of requested fields instead of KeyValue object
+        In that case, field names are not converted to camelCase. We need to check for both content_type and contentType
+        We iterate through given names and return the first value that is found
+    """
+    empty_value = ' '
```
avanigupta, line 195: Is empty_value a single space? Can you directly check for empty string in line 200?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -391,6 +404,17 @@ def __read_kv_from_config_store(azconfig_client,
...
+
+    # A request to list kvs of a non-existent snapshot returns an empty result.
+    # We first check if the snapshot exists before returning an empty result.
+    if snapshot and len(retrieved_kvs) == 0:
+        try:
+            _ = AppConfigSnapshotClient(azconfig_client).get_snapshot(name=snapshot)
+
+        except HttpResponseError as exception:
+            if exception.status_code == StatusCodes.NOT_FOUND:
+                raise ResourceNotFoundError("No snapshot with name '{}' was found.".format(snapshot))
```
avanigupta, line 416: What do you think about moving this code inside the `if snapshot` check on line 349? It would keep all snapshot related logic in one place.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -43,6 +43,12 @@ def load_arguments(self, _):
         validator=validate_feature_query_fields,
         arg_type=get_enum_type(['name', 'key', 'label', 'locked', 'last_modified', 'state', 'description', 'conditions'])
     )
+    snapshot_fields_arg_type = CLIArgumentType(
+        nargs='+',
+        help='Customize output fields for Snapshots',
+        validator=validate_snapshot_query_fields,
+        arg_type=get_enum_type(['name', 'etag', 'status_code', 'retention_period', 'filters', 'status', 'created', 'expires', 'size', 'items_count', 'size', 'tags'])
```
avanigupta, line 50: size appears twice in this list

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -340,3 +361,27 @@ def load_arguments(self, _):
 
     with self.argument_context('appconfig replica create') as c:
         c.argument('location', arg_type=get_location_type(self.cli_ctx), help='Location at which to create the replica.')
+
+    with self.argument_context('appconfig snapshot create') as c:
+        c.argument('snapshot_name', arg_type=snapshot_name_arg_type)
+        c.argument('filters', arg_type=snapshot_filter_arg_type)
+        c.argument('composition_type', arg_type=get_enum_type(["key", "key_label"]), help='Composition type used in building app configuration snapshots. If not specified, defaults to key.')
```
avanigupta, line 368:
```suggestion
        c.argument('composition_type', arg_type=get_enum_type(["key", "key_label"]), help='Composition type used in building App Configuration snapshots. If not specified, defaults to key.')
```

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -340,3 +361,27 @@ def load_arguments(self, _):
...
+        c.argument('retention_period', type=int, help='Duration in seconds for which a snapshot can remain archived before expiry. A snapshot can be archived for a maximum of 7 days (604,800s) for free tier stores and 90 days (7,776,000s) for standard tier stores. If specified, retention period must be at least 1 hour (3600s)')
```
avanigupta, line 369: Is there a default retention period? Or is that decided by server?

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -340,3 +361,27 @@ def load_arguments(self, _):
...
+        c.argument('tags', arg_type=tags_type, help="Space-separated tags: key[=value] [key[=value] ...].")
+
+    with self.argument_context('appconfig snapshot show') as c:
+        c.argument('snapshot_name', arg_type=snapshot_name_arg_type)
+        c.argument('fields', arg_type=snapshot_fields_arg_type)
+
+    with self.argument_context('appconfig snapshot list') as c:
+        c.argument('snapshot_name', options_list=['--snapshot-name', '-s'], help='If no name specified, return all snapshots by default. Support star sign as filters, for instance abc* means snapshots with abc as prefix to the name.')
```
avanigupta, line 377: Do we support any type of * regex here, or just prefix matching?

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -340,3 +361,27 @@ def load_arguments(self, _):
...
+        c.argument('status', help='Value used to filter snapshots by their status.')
+        c.argument('top', arg_type=top_arg_type)
+        c.argument('all_', options_list=['--all'], action='store_true', help="List all snapshots.")
```
avanigupta, line 380: We already define `top` and `all_` in the common appconfig section. I dont think we need to define it here again.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -340,3 +361,27 @@ def load_arguments(self, _):
...
+        c.argument('fields', arg_type=snapshot_fields_arg_type)
+
+    with self.argument_context('appconfig snapshot archive') as c:
+        c.argument('snapshot_name', arg_type=snapshot_name_arg_type)
```
avanigupta, line 384: `snapshot_name` can be moved to a common args section for `appconfig snapshot`.
`appconfig snapshot list` can override the common definition since its different.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
avanigupta: composition_type is missing

src/azure-cli/azure/cli/command_modules/appconfig/_snapshotmodels.py
```diff
@@ -0,0 +1,222 @@
...
+        self.composition_type = composition_type
+        self.created = created.isoformat() if isinstance(created, datetime) else str(created)
+        self.expires = expires.isoformat() if isinstance(expires, datetime) else (str(expires) if expires else None)
+        self.size = size
+        self.items_count = items_count
+        self.tags = tags
+        self.retention_period = retention_period
+
+    def __str__(self):
+        return "\netag: " + self.etag + \
```
avanigupta, line 89: If other fields are starting with capital letter, I think etag should also start with capital E.

src/azure-cli/azure/cli/command_modules/appconfig/_format.py
albertofori: Yes, that is possible. It just wasn't easily apparent why I was using a single space so the variable was supposed to be self-explanatory. Would you suggest we do a direct comparison?

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
albertofori: This is decided by the server if no retention period is specified.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
albertofori: It is just prefix matching. Would it be better to re-word `Support star sign as filters` to `Filters support prefix matching`?

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
albertofori: This has been updated, thanks!

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
albertofori: My bad! This has been updated.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
albertofori: This has been updated, thanks!

src/azure-cli/azure/cli/command_modules/appconfig/_snapshotmodels.py
albertofori: This has been updated, thanks!

src/azure-cli/azure/cli/command_modules/appconfig/_format.py
avanigupta: Oh I see we always compared it with single space. I dont remember if that's intentional though. Could you check if it works with empty string instead of single space?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: Right, what I meant was we could check if configsetting_iterable has no items as soon as we call list_snapshot_kv. If there are no items, call get_snapshot to verify if snapshot exists or not.

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
avanigupta: Ok, I see we have the same wording for other params that support filtering. I think its okay to keep the same wording here.

src/azure-cli/azure/cli/command_modules/appconfig/_format.py
albertofori: I tested and did not see any issue using an empty string, however, the default [build_table_output](https://github.com/Azure/azure-cli/blob/fdb9aa742b404a433f207b573eea37341769020c/src/azure-cli-core/azure/cli/core/commands/transform.py#L98C1-L98C30) method uses empty spaces for null values. 
I am thinking it would be good to be consistent with this in case I might be missing something.

avanigupta: Thanks for confirming! Yes, lets keep it as single space and add a comment explaining why its not an empty string.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
albertofori: Got it! This has been updated. 
Since the "configsetting_iterable" is a paged iterator that makes calls to backend to fetch subsequent pages, we only confirm if it is empty by attempting to add to process the first value.
Kindly let me know your thoughts on this.

src/azure-cli/azure/cli/command_modules/appconfig/_format.py
albertofori: This has been updated, thanks!

zhoxing-ms:
> ERROR: Can not find 'appconfig snapshot list --snapshot-name,' test case
ERROR: Please add some scenario tests for the new parameter
ERROR: Or add the parameter with missing_parameter_test_coverage rule in linter_exclusions.yml

Please add test case for the new parameter `--snapshot-name` of `appconfig snapshot list` command

albertofori:
> > ERROR: Can not find 'appconfig snapshot list --snapshot-name,' test case
> > ERROR: Please add some scenario tests for the new parameter
> > ERROR: Or add the parameter with missing_parameter_test_coverage rule in linter_exclusions.yml
> 
> Please add test case for the new parameter `--snapshot-name` of `appconfig snapshot list` command

@zhoxing-ms , please I already have a test case for this but this is still being flagged as missing. https://github.com/Azure/azure-cli/blob/6c0976785b68e538159a3c2dfaa31a2d5c0f174f/src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/test_appconfig_commands.py#L2990

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: You're right, this got too complicated and prone to errors in future. I think your previous approach was better.

albertofori: Got it, I will revert this change then.

albertofori: This has been updated.

avanigupta (approved):

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -340,3 +361,20 @@ def load_arguments(self, _):
...
         c.argument('location', arg_type=get_location_type(self.cli_ctx), help='Location at which to create the replica.')
+
+    with self.argument_context('appconfig snapshot') as c:
+        c.argument('snapshot_name', arg_type=snapshot_name_arg_type)
+
+    with self.argument_context('appconfig snapshot create') as c:
+        c.argument('filters', arg_type=snapshot_filter_arg_type)
+        c.argument('composition_type', arg_type=get_enum_type(["key", "key_label"]), help='Composition type used in building App Configuration snapshots. If not specified, defaults to key.')
+        c.argument('retention_period', type=int, help='Duration in seconds for which a snapshot can remain archived before expiry. A snapshot can be archived for a maximum of 7 days (604,800s) for free tier stores and 90 days (7,776,000s) for standard tier stores. If specified, retention period must be at least 1 hour (3600s)')
+        c.argument('tags', arg_type=tags_type, help="Space-separated tags: key[=value] [key[=value] ...].")
```
zhoxing-ms, line 372: The help message has already been defined in `tags_type`, so there is no need to repeat the definition
```suggestion
        c.argument('tags', arg_type=tags_type)
```

src/azure-cli/azure/cli/command_modules/appconfig/_snapshot_custom_client.py
```diff
@@ -0,0 +1,560 @@
...
+    path_arguments=None,
+    if_match=None,
+    if_none_match=None,
+    sync_token=None,
+    **kwargs
+):
+    _headers = case_insensitive_dict(kwargs.pop("headers", {}) or {})
+    _params = case_insensitive_dict(kwargs.pop("params", {}) or {})
+
+    api_version = kwargs.pop("api_version", SnapshotConstants.API_VERSION)
```
zhoxing-ms, line 57: In addition, if you have new features which are needed to be developed in the future *(especially for atomic commands)*, we strongly recommend that you try our advanced code generation tools [CodeGen - AAZ](https://azure.github.io/aaz-dev-tools/)

This tool is easy to use and can bring you the following benefits:
- It can improve your development efficiency and reduce manual effort.
- It can ensure the style and consistency of code, so it also reduces the burden of PR review.
- It can improve code quality and reduce the occurrence of bugs.
......

If you are interested in it and would like to learn more, please contact our CodeGen expert @kairu-ms

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
albertofori: image
@zhoxing-ms Please the original help message has additional text that is not applicable to the parameter. i.e. `Use "" to clear existing tags.`. This is why we indicate our own custom message to override it.

zhoxing-ms: May I ask why this parameter dose not support using "" to clear existing tags?

albertofori: This command is a `create` command, and will only succeed if a resource with the specified name does not already exist, so no replacement is possible for this command.

zhoxing-ms (approved):

PR #25033, opened 2023-01-04
[ARM] `az bicep`: Add configs to allow disabling version checks and using Bicep CLI from PATH

shenglol:
**Related command**

`az bicep *`

**Description**

The `az bicep` commands download Bicep CLI automatically if it does not exist on `$AZURE_CONFIG_DIR/bin`. This is causing an [GitHub API throttling issue](https://github.com/Azure/bicep/issues/3689) if users uses the Azure CLI ADO task, as each task run creates an empty azure config directory. The PR adds two configurations `bicep.use_binary_from_path` and `bicep.check_version` to allow users to reuse the pre-installed Bicep CLI on the CLI runners and to disable version checks for each `az bicep` command run.

Closes https://github.com/Azure/bicep/issues/9452.

**Testing Guide**
Set `bicep.use_binary_from_path` and `bicep.check_version` and run any `az bicep` commands.

**History Notes**

[ARM] `az bicep`: Add configuration `bicep.use_binary_from_path`. Possible values include `if_running_in_ci` (default) and Booleans
[ARM] `az bicep`: Add configuration `bicep.check_version` that accepts Boolean values. If set to `False`, version checks for Bicep CLI will be disabled

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/resource/_bicep.py
+++ b/src/azure-cli/azure/cli/command_modules/resource/_bicep.py
@@ -133,7 +133,8 @@ def ensure_bicep_installation(release_tag=None, target_platform=None, stdout=Tru
             print(f'Successfully installed Bicep CLI to "{installation_path}".')
         else:
             _logger.info(
-                "Successfully installed Bicep CLI to %s", installation_path,
+                "Successfully installed Bicep CLI to %s",
+                installation_path,
             )
     except IOError as err:
         raise ClientRequestError(f"Error while attempting to download Bicep CLI: {err}")
@@ -196,14 +197,20 @@ def _use_binary_from_path(cli_ctx):
     use_binary_from_path = cli_ctx.config.get("bicep", "use_binary_from_path", "if_running_in_ci").lower()
 
     if use_binary_from_path == "if_running_in_ci":
+        # With if_running_in_ci, GitHub Actions and Azure Pipeline users may expect some delay (usually a few days)
+        # in getting the latest version of Bicep CLI, since the az bicep commands will use the pre-installed Bicep CLI
+        # on the build agents, but the build agents has a different release cycle. The benefit is that the az bicep
+        # commands will not download the Bicep CLI on each pipeline run.
         return _is_running_in_ci()
     if use_binary_from_path in ["1", "yes", "true", "on"]:
+        # Setting the config True forces the az bicep commands to use the Bicep executable added to PATH, which
+        # indicates that the user is intended to manage the Bicep CLI, and version checks will be disabled.
         return True
     if use_binary_from_path in ["0", "no", "false", "off"]:
         return False
 
     _logger.warning(
-        'The configuration value of bicep.use_binary_from_path is invalid: "%s". Possible values include "if_running_in_ci" and Booleans.',  # pylint: disable=line-too-long
+        'The configuration value of bicep.use_binary_from_path is invalid: "%s". Possible values include "if_running_in_ci" (default) and Booleans.',  # pylint: disable=line-too-long
         use_binary_from_path,
     )
 
--- a/src/azure-cli/azure/cli/command_modules/resource/_bicep.py
+++ b/src/azure-cli/azure/cli/command_modules/resource/_bicep.py
@@ -57,9 +57,19 @@ def validate_bicep_target_scope(template_schema, deployment_scope):
         )
 
 
-def run_bicep_command(args, auto_install=True, check_version=True):
+def run_bicep_command(cli_ctx, args, auto_install=True):
+    if _use_binary_from_path(cli_ctx):
+        from shutil import which
+
+        if which("bicep") is None:
+            raise ValidationError('Could not find the "bicep" executable on PATH.')
+
+        return _run_command("bicep", args)
+
     installation_path = _get_bicep_installation_path(platform.system())
     installed = os.path.isfile(installation_path)
+
+    check_version = cli_ctx.config.getboolean("bicep", "check_version", True)
 
     if not installed:
         if auto_install:
@@ -123,8 +133,7 @@ def ensure_bicep_installation(release_tag=None, target_platform=None, stdout=Tru
             print(f'Successfully installed Bicep CLI to "{installation_path}".')
         else:
             _logger.info(
-                "Successfully installed Bicep CLI to %s",
-                installation_path,
+                "Successfully installed Bicep CLI to %s", installation_path,
             )
     except IOError as err:
         raise ClientRequestError(f"Error while attempting to download Bicep CLI: {err}")
@@ -175,6 +184,30 @@ def supports_bicep_publish():
     installation_path = _get_bicep_installation_path(system)
     installed_version = _get_bicep_installed_version(installation_path)
     return semver.compare(installed_version, "0.4.1008") >= 0
+
+
+def _is_running_in_ci():
+    if "GITHUB_ACTIONS" in os.environ or "TF_BUILD" in os.environ:
+        return True
+    return False
+
+
+def _use_binary_from_path(cli_ctx):
+    use_binary_from_path = cli_ctx.config.get("bicep", "use_binary_from_path", "if_running_in_ci").lower()
+
+    if use_binary_from_path == "if_running_in_ci":
+        return _is_running_in_ci()
+    if use_binary_from_path in ["1", "yes", "true", "on"]:
+        return True
+    if use_binary_from_path in ["0", "no", "false", "off"]:
+        return False
+
+    _logger.warning(
+        'The configuration value of bicep.use_binary_from_path is invalid: "%s". Possible values include "if_running_in_ci" and booleans.',  # pylint: disable=line-too-long
+        use_binary_from_path,
+    )
+
+    return False
 
 
 def _load_bicep_version_check_result_from_cache():
--- a/src/azure-cli/azure/cli/command_modules/resource/_bicep.py
+++ b/src/azure-cli/azure/cli/command_modules/resource/_bicep.py
@@ -62,7 +62,7 @@ def run_bicep_command(cli_ctx, args, auto_install=True):
         from shutil import which
 
         if which("bicep") is None:
-            raise ValidationError('Could not find the "bicep" executable on PATH.')
+            raise ValidationError('Could not find the "bicep" executable on PATH. To install Bicep via Azure CLI, set the "bicep.use_binary_from_path" configuration to False and run "az bicep install".')
 
         bicep_version_message = _run_command("bicep", "--version")
 
--- a/src/azure-cli/azure/cli/command_modules/resource/_bicep.py
+++ b/src/azure-cli/azure/cli/command_modules/resource/_bicep.py
@@ -209,7 +209,7 @@ def _bicep_installed_in_ci():
 
 def _use_binary_from_path(cli_ctx):
     use_binary_from_path = cli_ctx.config.get("bicep", "use_binary_from_path", "if_found_in_ci").lower()
-    
+
     _logger.debug('Current value of "use_binary_from_path": %s.', use_binary_from_path)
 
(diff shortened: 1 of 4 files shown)


yonzhan:
ARM

zhoxing-ms (dismissed):

zhoxing-ms:
> The PR adds two configurations bicep.use_binary_from_path and bicep.check_version to allow users to reuse the pre-installed Bicep CLI on the CLI runners and to disable version checks for each az bicep command run.

May I ask how users know about these two new configurations. Is there any related documentation? Or could we consider including their introduction in the help message?

shenglol:
> > The PR adds two configurations bicep.use_binary_from_path and bicep.check_version to allow users to reuse the pre-installed Bicep CLI on the CLI runners and to disable version checks for each az bicep command run.
> 
> May I ask how users know about these two new configurations. Is there any related documentation? Or could we consider including their introduction in the help message?

Ah, I was going to ask the same question. Is adding docs for custom configurations to the help message the preferred way? Do we have an official guideline?

zhoxing-ms:
> [ARM] BREAKING CHANGE (for GitHub Actions and Azure Pipeline users): az bicep: Add configuration bicep.use_binary_from_path. Possible values include if_running_in_ci (default) and Booleans. Setting the config True forces the az bicep commands to use the Bicep executable added to PATH, which indicates that the user is intended to manage the Bicep CLI, and version checks will be disabled. With if_running_in_ci, GitHub Actions and Azure Pipeline users may expect some delay (usually a few days) in getting the latest version of Bicep CLI, since the az bicep commands will use the pre-installed Bicep CLI on the build agents, but the build agents has a different release cycle. The benefit is that the az bicep commands will not download the Bicep CLI on each pipeline run. To keep the old behavior, users may set the value to False.

Since this is not a breaking change release window for CLI (the breaking change window only twice a year: Build sprint and Ignite sprint), do we have any way to avoid this breaking change?

shenglol:
> > [ARM] BREAKING CHANGE (for GitHub Actions and Azure Pipeline users): az bicep: Add configuration bicep.use_binary_from_path. Possible values include if_running_in_ci (default) and Booleans. Setting the config True forces the az bicep commands to use the Bicep executable added to PATH, which indicates that the user is intended to manage the Bicep CLI, and version checks will be disabled. With if_running_in_ci, GitHub Actions and Azure Pipeline users may expect some delay (usually a few days) in getting the latest version of Bicep CLI, since the az bicep commands will use the pre-installed Bicep CLI on the build agents, but the build agents has a different release cycle. The benefit is that the az bicep commands will not download the Bicep CLI on each pipeline run. To keep the old behavior, users may set the value to False.
> 
> Since this is not a breaking change release window for CLI _(the breaking change window only twice a year: **Build sprint** and **Ignite sprint**)_, do we have any way to avoid this breaking change?

I just removed BREAKING CHANGE from the note. Technically, it's not a breaking change on the Azure CLI side.

zhoxing-ms:
> Ah, I was going to ask the same question. Is adding docs for custom configurations to the help message the preferred way? Do we have an official guideline?

Our suggestion is usually to describe the configurations to be used in the help message, or write a concept document for these configurations and then reference the doc link in the help message

zhoxing-ms:
> [ARM] az bicep: Add configuration bicep.use_binary_from_path. Possible values include if_running_in_ci (default) and Booleans. Setting the config True forces the az bicep commands to use the Bicep executable added to PATH, which indicates that the user is intended to manage the Bicep CLI, and version checks will be disabled. With if_running_in_ci, GitHub Actions and Azure Pipeline users may expect some delay (usually a few days) in getting the latest version of Bicep CLI, since the az bicep commands will use the pre-installed Bicep CLI on the build agents, but the build agents has a different release cycle. The benefit is that the az bicep commands will not download the Bicep CLI on each pipeline run. To keep the old behavior, users may set the value to False

This description of history note is too long. It is suggested to simplify it and then write more details into the code comments

shenglol:
@zhoxing-ms Could you take another look? I've addressed the comments.

src/azure-cli/azure/cli/command_modules/resource/_bicep.py
```diff
@@ -57,10 +57,20 @@ def validate_bicep_target_scope(template_schema, deployment_scope):
         )
 
 
-def run_bicep_command(args, auto_install=True, check_version=True):
+def run_bicep_command(cli_ctx, args, auto_install=True):
+    if _use_binary_from_path(cli_ctx):
+        from shutil import which
+
+        if which("bicep") is None:
```
anthony-c-martin, line 64: Is it worth validating here that this can be executed (and also logging the version info to debug output)?

src/azure-cli/azure/cli/command_modules/resource/_bicep.py
```diff
@@ -57,10 +57,20 @@ def validate_bicep_target_scope(template_schema, deployment_scope):
...
+            raise ValidationError('Could not find the "bicep" executable on PATH.')
```
anthony-c-martin, line 65: Should the message explain how the user can configure AzCLI to restore the old behavior?

src/azure-cli/azure/cli/command_modules/resource/_bicep.py
```diff
@@ -177,6 +187,36 @@ def supports_bicep_publish():
     return semver.compare(installed_version, "0.4.1008") >= 0
 
 
+def _is_running_in_ci():
+    if "GITHUB_ACTIONS" in os.environ or "TF_BUILD" in os.environ:
```
anthony-c-martin, line 191: Is there a possibility that these vars will be set but Bicep isn't on the PATH? E.g. if using a non-standard container image.

src/azure-cli/azure/cli/command_modules/resource/_bicep.py
```diff
@@ -177,6 +187,36 @@ def supports_bicep_publish():
...
+    return False
+
+
+def _use_binary_from_path(cli_ctx):
+    use_binary_from_path = cli_ctx.config.get("bicep", "use_binary_from_path", "if_running_in_ci").lower()
+
+    if use_binary_from_path == "if_running_in_ci":
+        # With if_running_in_ci, GitHub Actions and Azure Pipeline users may expect some delay (usually a few days)
+        # in getting the latest version of Bicep CLI, since the az bicep commands will use the pre-installed Bicep CLI
+        # on the build agents, but the build agents has a different release cycle. The benefit is that the az bicep
```
anthony-c-martin, line 202: Do you know how release cycles are managed for these build agents? Is there a way we can get visibility into what versions are pre-installed?

shenglol: The releases are visible at https://github.com/actions/runner/releases and https://github.com/microsoft/azure-pipelines-agent/releases. Based on the historical data they usually get published 2 to 3 times per month.

src/azure-cli/azure/cli/command_modules/resource/_bicep.py
shenglol: `which` ensures the specified file is an executable. Added logging for the Bicep version info.

src/azure-cli/azure/cli/command_modules/resource/_bicep.py
shenglol: Good point. Included an explanation in the error message.

src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -2693,6 +2693,17 @@
...
 short-summary: Bicep CLI command group.
+long-summary: |
+  Bicep CLI command group. There are two configurations that can be set for the command group,
+  including bicep.check_version and bicep.use_binary_from_path.
+
+  bicep.check_version accepts Boolean values and defaults to True. If set to False, version checks for Bicep CLI installed via Azure CLI will be disabled.
+
+  The possible values of bicep.use_binary_from_path are if_found_in_ci and Booleans. The default value is if_found_in_ci.
+
+  With if_found_in_ci, GitHub Actions and Azure Pipeline users may expect some delay (usually a few days) in getting the latest version of Bicep CLI, since the az bicep commands will use the pre-installed Bicep CLI on the build agents, but the build agents has a different release cycle. The benefit is that the az bicep commands will not download the Bicep CLI on each pipeline run.
```
zhoxing-ms, line 2704: > if_found_in_ci

Please do not expose the details of code logic in the help description, which will confuse users.
It is recommended to write these contents into the corresponding code comments.

shenglol: Removed the details from the help message.

zhoxing-ms (approved):

src/azure-cli/azure/cli/command_modules/resource/_bicep.py
```diff
@@ -177,6 +196,38 @@ def supports_bicep_publish():
     return semver.compare(installed_version, "0.4.1008") >= 0
 
 
+def _bicep_installed_in_ci():
+    if "GITHUB_ACTIONS" in os.environ or "TF_BUILD" in os.environ:
+        from shutil import which
+
+        return which("bicep") is not None
```
anthony-c-martin, line 203: Worth adding logging here to distinguish between "env vars not set" & "bicep not found on the PATH"?

src/azure-cli/azure/cli/command_modules/resource/tests/latest/test_resource_bicep.py
```diff
@@ -16,6 +16,10 @@
     _bicep_version_check_file_path,
 )
 from azure.cli.core.azclierror import InvalidTemplateError
+from azure.cli.core.mock import DummyCli
+
+cli_ctx = DummyCli()
+cli_ctx.config.set_value("bicep", "use_binary_from_path", "false")
```
anthony-c-martin, line 22: Is it worth adding tests for some of these new scenarios (reading of config, usage of env vars)?

src/azure-cli/azure/cli/command_modules/resource/_bicep.py
shenglol: Makes sense.

src/azure-cli/azure/cli/command_modules/resource/tests/latest/test_resource_bicep.py
shenglol: Yup, will add tests.

anthony-c-martin (approved):
Looks great!

zhoxing-ms:
@shenglol Could you please fix the code style issue?

src/azure-cli/azure/cli/command_modules/resource/_help.py
```diff
@@ -2693,6 +2693,12 @@
 helps['bicep'] = """
 type: group
 short-summary: Bicep CLI command group.
+long-summary: |
+  Bicep CLI command group. There are two configurations that can be set for the command group, including bicep.check_version and bicep.use_binary_from_path.
+
+  bicep.check_version accepts Boolean values and defaults to True. If set to False, version checks for Bicep CLI installed via Azure CLI will be disabled.
+
+  The possible values of bicep.use_binary_from_path are if_found_in_ci and Booleans. The default value is if_found_in_ci.
```
zhoxing-ms, line 2701: It is suggested to give an example of using `az config set`, which is more helpful for customers to understand.

zhoxing-ms: Such as:
https://github.com/Azure/azure-cli-extensions/blob/c2fa9aff05819e8b311a1614e787170e5ad98146/src/next/azext_next/_help.py#L14-L35

shenglol: Good suggestion. The same code is helpful. I've updated the help message to include examples.

zhoxing-ms (approved):

PR #25160, opened 2023-01-18
{Packaging} Fix #25149: Updated deb_install.sh to use /etc/apt/keyrings

### LINKED ISSUE #25149: Update deb_install.sh to use /usr/share/keyrings?
## Issue

The apt-key command is [deprecated](https://itsfoss.com/apt-key-deprecated/) with Ubuntu 22.04 (except for deletes). 

The `/usr/share/keyrings` directory is the recommended location for package signing keys now, rather than `/etc/apt/keyrings`. 

The [deb_install.sh script](https://github.com/Azure/azure-cli/blob/dev/scripts/release/debian/deb_install.sh) used by [option 1](https://learn.microsoft.com/cli/azure/install-azure-cli-linux?pivots=apt#option-1-install-with-one-command)  would benefit from changing to these newer standards. The [manual commands](https://learn.microsoft.com/cli/azure/install-azure-cli-linux?pivots=apt) are slightly better as they include the `signed-by` location, but again, should that location now be `/usr/share/keyring`?

richeney:
Fixes #25149. Updated deb_install.sh script to match the [manual commands](https://learn.microsoft.com/cli/azure/install-azure-cli-linux?pivots=apt).

* The key is now placed in `/etc/apt/keyrings`
* The /etc/apt/sources.list.d/azure-cli.list file now has the signed-by form, e.g.

    ```text
    deb [arch=amd64 signed-by=/etc/apt/keyrings/microsoft.gpg] https://packages.microsoft.com/repos/azure-cli/ jammy main
    ```

* Also added a check to ensure the script is run as root

---

### DIFF
--- a/scripts/release/debian/deb_install.sh
+++ b/scripts/release/debian/deb_install.sh
@@ -38,6 +38,11 @@ global_consent=0 # Artificially giving global consent after review-feedback. Rem
 
 setup() {
 
+    if [[ $(/usr/bin/whoami) != "root" ]]; then
+        echo "Script $(basename $0) must be run as root."
+        exit 1
+    fi
+
     assert_consent "Add packages necessary to modify your apt-package sources?" ${global_consent}
     set -v
     export DEBIAN_FRONTEND=noninteractive
@@ -45,9 +50,15 @@ setup() {
     apt-get install -y apt-transport-https lsb-release gnupg curl
     set +v
 
+    if [[ ! -d /etc/apt/keyrings ]]; then
+        assert_consent "Create the /etc/apt/keyrings directory?" ${global_consent}
+        mkdir -pm 755 /etc/apt/keyrings
+    fi
+
     assert_consent "Add Microsoft as a trusted package signer?" ${global_consent}
     set -v
-    curl -sL https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > /etc/apt/trusted.gpg.d/microsoft.gpg
+    curl -sL https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > /etc/apt/keyrings/microsoft.gpg
+    chmod go+r /etc/apt/keyrings/microsoft.gpg
     set +v
 
     assert_consent "Add the Azure CLI Repository to your apt sources?" ${global_consent}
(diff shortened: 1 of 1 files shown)


yonzhan:
Packaging

scripts/release/debian/deb_install.sh
```diff
@@ -81,7 +92,7 @@ setup() {
             exit 1
         fi
     fi
-    echo "deb [arch=amd64] https://packages.microsoft.com/repos/azure-cli/ ${CLI_REPO} main" \
+    echo "deb [arch=amd64 signed-by=/etc/apt/keyrings/microsoft.gpg] https://packages.microsoft.com/repos/azure-cli ${CLI_REPO} main" \
```
bebound, line 95: How about using `` arch=`dpkg --print-architecture` ``? We plan to support arm64.

richeney: @bebound , great news about the arm64 support! I have added your suggested change.
 
**I have also added in some additional logic to remove the deprecated key from the main keyring if found.** This will avoid duplicates on any system where they've run the older script and then run this one.

richeney:
@bebound, is there something else that needs to be done to move this forward?

bebound:
I confirm that new script works on Ubuntu 18.04, 20.04, 22.04 and Debian 10, 11.
But should we remove `/etc/apt/trusted.gpg.d/microsoft.gpg`? No one can guarantee that this file is generate by CLI script. (Old version VS Code, Teams and Microsoft Defender also use this file.)
@jiasli for awareness

richeney:
I checked the [vscode install page for Debian](https://code.visualstudio.com/docs/setup/linux) and traced back to when they made the change: <https://github.com/microsoft/vscode-docs/commit/cbdb844f8db819bc2a0e2c4a2756679f6dd2e78d>. 

This PR shows the older commands using the main keyring and older format. If there is an old `/etc/apt/sources.list.d/vscode.list` file (without `signed-by`) then it would be using the main keyring.

I'll remove that logic. Leaving it in the old keyring doesn't seem to cause any problem.

bebound:
Thank you so much. As this is an essential change, we'll do a team review before merging it. Thanks for your patience.

jiasli:
The PR for changing the manual installation step is https://github.com/MicrosoftDocs/azure-docs-cli/pull/3435

PR #25172, opened 2023-01-19
{AD} `az ad`: Update `--filter` argument's help message

jiasli:
**Related command**
`az ad`

**Description**
`--filter` is defined on the whole `az ad` scope, so whenever a command under `az ad` supports `--filter`, the help message shows the same example:

https://github.com/Azure/azure-cli/blob/ce4469321e66be2ef75b2c2721be7e289ab20a84/src/azure-cli/azure/cli/command_modules/role/_params.py#L232

https://github.com/Azure/azure-cli/blob/ce4469321e66be2ef75b2c2721be7e289ab20a84/src/azure-cli/azure/cli/command_modules/role/_params.py#L237

The current example `--filter "displayname eq 'test' and servicePrincipalType eq 'Application'"` only works for `az ad sp list`, but doesn't work for

- `az ad app list`
- `az ad app permission list-grants`
- `az ad user list`
- `az ad group list`

This PR updates the help message of `--filter` to a generic one that suits all commands that supports `--filter`.

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/role/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/role/_params.py
@@ -234,7 +234,9 @@ def load_arguments(self, _):
         c.argument('identifier_uri', help='graph application identifier, must be in uri format')
         c.argument('spn', help='service principal name')
         c.argument('upn', help='user principal name, e.g. john.doe@contoso.com')
-        c.argument('query_filter', options_list=['--filter'], help='OData filter, e.g. --filter "displayname eq \'test\' and servicePrincipalType eq \'Application\'"')
+        c.argument('query_filter', options_list=['--filter'],
+                   help='The $filter OData query parameter to retrieve a subset of a collection. '
+                        'For more information, see https://learn.microsoft.com/graph/filter-query-parameter')
 
     with self.argument_context('ad user') as c:
         c.argument('mail_nickname', help='mail alias. Defaults to user principal name')


src/azure-cli/azure/cli/command_modules/role/_params.py
```diff
@@ -234,7 +234,9 @@ def load_arguments(self, _):
         c.argument('identifier_uri', help='graph application identifier, must be in uri format')
         c.argument('spn', help='service principal name')
         c.argument('upn', help='user principal name, e.g. john.doe@contoso.com')
-        c.argument('query_filter', options_list=['--filter'], help='OData filter, e.g. --filter "displayname eq \'test\' and servicePrincipalType eq \'Application\'"')
+        c.argument('query_filter', options_list=['--filter'],
+                   help='The $filter OData query parameter to retrieve a subset of a collection. '
+                        'For more information, see https://learn.microsoft.com/graph/filter-query-parameter')
```
jiasli, line 239: This topic is quite similar to the one we are having on `--query`: https://github.com/Azure/azure-cli/pull/24633#discussion_r1069071670, https://github.com/Azure/azure-cli/pull/24332#discussion_r1003912453. That is, whether we want to provide concrete examples for generic arguments in each command.

jiasli, line 239: Another thing is **how we can better convey that quoting rules should be considered in different shells**. Especially, when there are spaces in the value, the value should be quoted.

Currently `--query` doesn't mention quoting rules.

```
    --query            : JMESPath query string. See http://jmespath.org/ for more information and
                         examples.
```

yonzhan:
Update --filter help

PR #25178, opened 2023-01-19
Trailing slash on issuer URL causes issues

mluker:
The trailing slash on the issuer URL in the code sample will cause the federated credential to not work when leveraging the azure/login action in a pipeline. The error it throws is "AADSTS70021: No matching federated identity record found for presented assertion".

Removing the trailing slash fixes it and works as expected.

**Description**
Documentation update when using `az ad app federated-credential create` to create a federated credential.

The official GitHub documents also call this out as not having a trailing slash https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#:~:text=(Issuer)%20The%20issuer%20of%20the%20OIDC%20token%3A%20https%3A//token.actions.githubusercontent.com

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/role/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/role/_help.py
@@ -350,7 +350,7 @@ examples:
     ("credential.json" contains the following content)
     {
         "name": "Testing",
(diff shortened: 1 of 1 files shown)


yonzhan:
role

Issue #24836, opened 2022-12-08
`openssl` command not found in Azure CLI 2.43.0 docker image


jiasli:
## Symptom

The latest Azure CLI 2.43.0 docker image doesn't contain `openssl` command anymore.

```
> docker run -it mcr.microsoft.com/azure-cli
0c12bf6c7b2f:/# openssl
bash: openssl: command not found
```

This causes failure if a user's workload relies on `openssl` command.

## Root cause

According to Alpine Linux 3.17's release history:

https://alpinelinux.org/posts/Alpine-3.17.0-released.html

> OpenSSL 3.0 is now the default OpenSSL version. OpenSSL 1.1 is available via the `openssl1.1-compat` package.

`openssl` package in `alpine:3.17` is now OpenSSL 3.0:

```
> docker run -it alpine:3.17
# apk search openssl
...
openssl-3.0.7-r0
```

OpenSSL 3.0 is not compatible with `cryptography` 3.3.2 which Azure CLI currently uses. For compatibility with `cryptography`, we now installs `openssl1.1-compat`:

- https://github.com/Azure/azure-cli/pull/24768

`openssl1.1-compat` installs `/usr/bin/openssl1.1` command:

```
# apk add openssl1.1-compat
# apk info --contents openssl1.1-compat
openssl1.1-compat-1.1.1s-r0 contains:
usr/bin/openssl1.1
# command -v openssl1.1
/usr/bin/openssl1.1
```

Because `openssl` package is not installed anymore, `/usr/bin/openssl` command is not available.

**For the released docker image, Azure CLI (`az` command) is the only tool that is guaranteed to work. The availability of other tools or packages are not guaranteed. If you rely on a certain tool or package such as `openssl`, you have to manually install it with `apk add` before using it.**

However, it is still a good practice for us to explicitly mention such changes. For example:

- https://github.com/Azure/azure-cli/pull/21206

## Proposed solutions

- Install `openssl`, either in the docker image by us or by your own workload. This will make `openssl` command available again, but this may introduce other breaking changes as `openssl` 3.0 may not be backward compatible with `openssl` 1.1, even though they share the same name.
- Switch to using `openssl1.1` in your workload, until you have fully tested `openssl` 3.0 works as expected with your workload.
- Temporarily use Azure CLI 2.42.0 docker image until we add `openssl` in Azure CLI 2.44.0.
[...]

PR #23946, opened 2022-09-20
{Packaging} Trim mgmt SDKs

jiasli:
**Description**

Reduce Azure CLI's package size per

- https://github.com/Azure/azure-cli/issues/7387

Add `trim-sdk.py` script to trim Python SDKs by

- Removing `aio` folders
- Removing unused API versions

PR #25184, opened 2023-01-19
[Packaging] Optimize Dockerfile build

hholst80:
This PR makes the build Docker artifact smaller by including multi-stage build in Dockerfile,
and trim ~250MB from the exported image.

image

Update: The image size is now down to 444MB.

---

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,4 @@
+# syntax=docker/dockerfile:1.4
 #---------------------------------------------------------------------------------------------
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License. See License.txt in the project root for license information.
@@ -5,7 +6,44 @@
 
 ARG PYTHON_VERSION="3.11"
 
-FROM python:${PYTHON_VERSION}-alpine
+FROM python:${PYTHON_VERSION}-alpine AS base
+
+# We don't use openssl (3.0) for now. We only install it so that users can use it.
+# libintl and icu-libs - required by azure devops artifact (az extension add --name azure-devops)
+RUN apk add --no-cache \
+    bash bash-completion openssh-keygen ca-certificates openssl \
+    curl jq git perl zip \
+    libintl icu-libs libc6-compat \
+ && update-ca-certificates
+
+
+FROM base AS builder
+
+ARG JP_VERSION="0.2.1"
+# bash gcc make openssl-dev libffi-dev musl-dev - dependencies required for CLI
+RUN apk add --no-cache --virtual .build-deps gcc make openssl-dev libffi-dev musl-dev linux-headers
+
+WORKDIR azure-cli
+COPY . /azure-cli
+
+# 1. Install azure-cli in /usr/local
+# 2. Run trim_sdk.py to clean up un-used py files: https://github.com/Azure/azure-cli/pull/25801
+# 3. Check that az and az.completion.sh can run
+# 4. Remove __pycache__. It is important that we run this at the end
+RUN ./scripts/install_full.sh
+RUN python3 ./scripts/trim_sdk.py
+RUN /usr/local/bin/az && bash -c "source /usr/local/bin/az.completion.sh"
+RUN find /usr/local/lib/python*/site-packages/azure -name __pycache__ | xargs rm -rf
+
+# Install jp tool
+RUN <<HERE
+  arch=$(arch | sed s/aarch64/arm64/ | sed s/x86_64/amd64/)
+  curl -L https://github.com/jmespath/jp/releases/download/${JP_VERSION}/jp-linux-${arch} -o /usr/local/bin/jp
+  chmod +x /usr/local/bin/jp
+HERE
+
+
+FROM base
 
 ARG CLI_VERSION
 
@@ -25,49 +63,11 @@ LABEL maintainer="Microsoft" \
       org.label-schema.vcs-url="https://github.com/Azure/azure-cli.git" \
       org.label-schema.docker.cmd="docker run -v \${HOME}/.azure:/root/.azure -it mcr.microsoft.com/azure-cli:$CLI_VERSION"
 
-# bash gcc make openssl-dev libffi-dev musl-dev - dependencies required for CLI
-# openssh - included for ssh-keygen
-# ca-certificates
-
-# curl - required for installing jp
-# jq - we include jq as a useful tool
-# pip wheel - required for CLI packaging
-# jmespath-terminal - we include jpterm as a useful tool
-# libintl and icu-libs - required by azure devops artifact (az extension add --name azure-devops)
-
-# We don't use openssl (3.0) for now. We only install it so that users can use it.
-RUN apk add --no-cache bash openssh ca-certificates jq curl openssl perl git zip \
- && apk add --no-cache --virtual .build-deps gcc make openssl-dev libffi-dev musl-dev linux-headers \
- && apk add --no-cache libintl icu-libs libc6-compat \
- && apk add --no-cache bash-completion \
- && update-ca-certificates
-
-ARG JP_VERSION="0.2.1"
-
-RUN arch=$(arch | sed s/aarch64/arm64/ | sed s/x86_64/amd64/) && curl -L https://github.com/jmespath/jp/releases/download/${JP_VERSION}/jp-linux-$arch -o /usr/local/bin/jp \
- && chmod +x /usr/local/bin/jp
-
-WORKDIR azure-cli
-COPY . /azure-cli
-
-# 1. Build packages and store in tmp dir
-# 2. Install the cli and the other command modules that weren't included
-RUN ./scripts/install_full.sh && python ./scripts/trim_sdk.py \
- && cat /azure-cli/az.completion > ~/.bashrc \
- && runDeps="$( \
-    scanelf --needed --nobanner --recursive /usr/local \
-        | awk '{ gsub(/,/, "\nso:", $2); print "so:" $2 }' \
-        | sort -u \
-        | xargs -r apk info --installed \
-        | sort -u \
-    )" \
- && apk add --virtual .rundeps $runDeps
-
-WORKDIR /
-
-# Remove CLI source code from the final image and normalize line endings.
-RUN rm -rf ./azure-cli && \
-    dos2unix /root/.bashrc /usr/local/bin/az
+RUN --mount=from=builder,target=/mnt <<HERE
+  cd /mnt/usr/local
+  cp -ru . /usr/local/
+  ln -s /usr/local/bin/az.completion.sh /etc/profile.d/
+HERE
 
 ENV AZ_INSTALLER=DOCKER
 CMD bash


hholst80:
@microsoft-github-policy-service agree [company="Omnimodular"]

hholst80:

yonzhan:
Packaging

hholst80:
2 weeks of no activity. I am pretty sure I did not break anything with my changes.
(Check the Format of Pull Request Title and Content)  should have been fixed 2 weeks ago.

Dockerfile
```diff
@@ -5,58 +5,82 @@
...
+      org.label-schema.description="The Azure CLI is used for all Resource Manager deployments in Azure." \
+      org.label-schema.url="https://docs.microsoft.com/cli/azure/overview" \
+      org.label-schema.usage="https://docs.microsoft.com/cli/azure/install-az-cli2#docker" \
+      org.label-schema.build-date=$BUILD_DATE \
+      org.label-schema.vcs-url="https://github.com/Azure/azure-cli.git" \
+      org.label-schema.docker.cmd="docker run -v \${HOME}/.azure:/root/.azure -it mcr.microsoft.com/azure-cli:$CLI_VERSION"
+
+COPY --from=builder /azure-cli/az.completion /root/.bashrc
+COPY --from=builder /usr/local /usr/local
+COPY --from=tools /usr/local/bin/jp /usr/local/bin/
```
bebound, line 81: I think this is ambiguous and hard to maintain.
Alternatively, we can modify the build script to install and delete the src in one step.

hholst80: I think that is strictly worse, but I see the point. We can move jp into the builder step and then we have just two COPY commands that would have very little reason for change. We could also do this:

```
COPY --from=builder /usr/local /usr/local
COPY --from=tools /usr/local /usr/local
```

I do not think az.completion should be included in the environment to begin with, nor should jp, but here we are. I only want to make the existing solution smaller. Without sacrificing readability or maintainability.

hholst80: I do not think that az.competion.sh should be in /root is what I mean. The jp binary is a single static executable. If the user wants it inside az-cli they can just inject it at container creation or have it mounted. There is no need to have it bundled with az unless there is a runtime dependency to it from az cli itself.

image

hholst80:
Including the `__pycache__` cleanup, the build image is down to ~704MB (uncompressed) on my local build. From 1.3GB that is worth while. I think the same cleanup (remove __pycache__) should be included in the rpm and deb build as well. Either ship the .py or just the .pyc but not both?

Dockerfile
bebound: I think this line is not safe `COPY --from=builder /usr/local /usr/local`. What if the installed package also modifies another folder. We can't guarantee that everything we need in that folder. (Although it might be true.) 
I prefer using a single RUN command to do the same thing. Like the [CPython dockerfile](https://github.com/docker-library/python/blob/20c8782b7f72091658c18a7347956ec7f7c4010e/3.11/alpine3.17/Dockerfile#L26)

hholst80: In this multistage solution, we do not carry over *any* potential residual components from the build process. That is very hard to achieve with RUN in a single stage. There is always junk left behind and not to mention it is hard to read unmaintainable. There is also observational bias to this commonality of long-winded single-stage builds: nobody dares to change anything because they do not understand the consequences of their changes. Not the best practice, that is not a place where we want to be.

At first, that can seem like a reasonable fear that a generic install script could install something outside the defined install location. To overcome our fear it is enough to open up the install script to see what it does: pip install exclusively. While it could be possible for a setup.py to do something incredibly stupid here the python pip framework is reasonably opinionated on what can and should be done (too opinionated if you ask me). If any files are installed they will be put in the target specified by pip configuration.

Do you feel strong about the /usr/local copy I might be able to fix that. But if you are against the multi-stage I cannot fix that. That is a requirement for a better more maintainable Dockerfile with a smaller build artifact that also can be generated more quickly.

bebound: I'm not against mult-stage build. If CLI is a Node or Go project, I definitely choose the multi-stage build. I'm sure everything I need is the `dist` folder or the binary.

The new dockerfile also remove `bash`, `git` and other packages. If we choose multi-stage build, we also need to figure out what file is changed during installation, which is more difficult than pip installation.

PS: I know we should not keep irrelevant packages into image. In order to prevent issues like #24836, we still keep them for now.

hholst80: Let me know if you would prefer this kind of solution instead.

https://github.com/hholst80/azure-cli/tree/dockerfile-multistage-alt

ozbillwang:
https://github.com/Azure/azure-cli/issues/19591

and add my comments and discoveries.

https://github.com/Azure/azure-cli/issues/19591#issuecomment-1405972500

The reason why it is so big, because, `azure_cli` keeps all version of codes for each component. 

For example, for `az network`, it keeps 33 copes when you install azure_cli in Docker. 

We need know how to install with latest version only.

```
$ pwd
/app/.venv/lib/python3.11/site-packages/azure/mgmt/network

$ ls -l
...
9708	v2018_11_01
9964	v2018_12_01
10180	v2019_02_01
10768	v2019_04_01
11280	v2019_06_01
11640	v2019_07_01
11860	v2019_08_01
11944	v2019_09_01
12172	v2019_11_01
12344	v2019_12_01
12736	v2020_03_01
12924	v2020_04_01
13592	v2020_05_01
14180	v2020_06_01
14420	v2020_07_01
14596	v2020_08_01
14700	v2020_11_01
14880	v2021_02_01
16984	v2022_01_01

$ ls -ld v*|wc -l
33
```

jiasli:
> The reason why it is so big, because, `azure_cli` keeps all version of codes for each component.
> 
> For example, for `az network`, it keeps 33 copes when you install azure_cli in Docker.

We previously trimmed Python SDKs for Windows and Linux packages: https://github.com/Azure/azure-cli/pull/23946. We will apply the same logic to docker images.

ozbillwang:
> > The reason why it is so big, because, `azure_cli` keeps all version of codes for each component.
> > For example, for `az network`, it keeps 33 copes when you install azure_cli in Docker.
> 
> We previously trimmed Python SDKs for Windows and Linux packages: #23946. We will apply the same logic to docker images.

Thanks for the update. Please share the ticket or issue so I can follow up (subsribe it)

hholst80:
We're down to 485 MB now (compressed).  If we keep this PR open long enough, we will eventually get it below 100 MB.  ;-)

hholst80:
If this flies with the CI/CD gods, the size is almost acceptable: 

image

ozbillwang:
> If this flies with the CI/CD gods, the size is almost acceptable: 
> 
> 
> 
> image
> 
> 

yes, you are right. i use it in CICD, thanks for the efforts

ozbillwang:
@hholst80 

Could you show me how to build the image locally?

Should it be fine that I run `docker build -t azure_cli .` under the root of repo ?

jongio:
I just ran 'dive' on this PR and it looks like there's a potential for another 111MB of savings...

image


@hholst80 - Could you also run dive and see what it tells you?

https://github.com/wagoodman/dive

hholst80:
> I just ran 'dive' on this PR and it looks like there's a potential for another 111MB of savings...
> image
> 
> @hholst80 - Could you also run dive and see what it tells you?
> 
> https://github.com/wagoodman/dive

Great catch there. Thank you @jongio. The merge COPY was indeed a bit sloppy. We're up to 98% efficiency as per dive analysis.  I am using buildx extensions here, but that should be no concern I hope. We'll see if the Build CI/CD throws an error or not.

image

idavis:
Do we need all of these installed? Are these just needed for building or do we shell out to them?
```shell
# We don't use openssl (3.0) for now. We only install it so that users can use it.
# libintl and icu-libs - required by azure devops artifact (az extension add --name azure-devops)
RUN apk add --no-cache \
    bash bash-completion openssh-keygen ca-certificates openssl \
    curl jq git perl zip \
    libintl icu-libs libc6-compat \
&& update-ca-certificates
```

The `jp` installation is also hard-coded to amd64. Do we want support for native arm64 such as apple silicon or surface users?

jongio:
@hholst80 - Can you look into using a Chislled image? 

https://ubuntu.com/blog/benefits-chiselled-ubuntu-images-aspnet-eshop-demo

hholst80:
> Do we need all of these installed? Are these just needed for building or do we shell out to them?
> ```shell
> # We don't use openssl (3.0) for now. We only install it so that users can use it.
> # libintl and icu-libs - required by azure devops artifact (az extension add --name azure-devops)
> RUN apk add --no-cache \
>     bash bash-completion openssh-keygen ca-certificates openssl \
>     curl jq git perl zip \
>     libintl icu-libs libc6-compat \
> && update-ca-certificates
> ```
> 
> The `jp` installation is also hard-coded to amd64. Do we want support for native arm64 such as apple silicon or surface users?

I think it is out of scope for this PR to remove or add to existing tooling. Doing so will break between zero, 1, and a million users workflows so it requires some sensitivity and cost benefit analysis etc.

Removing curl broke my own workflows so it is a real problem not just theoretical issue.

This PR takes forever to get across and it's getting better and better. But it will not provide much value until it is merged.

jongio:
@jiasli - I'm wondering if we have any CLI perf tests to see if removing the __pycache__ impacts runtime perf.

Dockerfile
```diff
@@ -1,11 +1,57 @@
...
+COPY . /azure-cli
+
+# 1. Build packages and store in tmp dir
+# 2. Install the cli and the other command modules that weren't included
+# 3. Run trim_sdk.py to clean up un-used py files: https://github.com/Azure/azure-cli/pull/25801
+RUN ./scripts/install_full.sh
+RUN python3 ./scripts/trim_sdk.py
+
+# Normalize line endings.
+RUN dos2unix /usr/local/bin/az /usr/local/bin/az.completion.sh
```
jongio, line 38: I wonder why we have to run dos2unix.  Is that a CI issue?

Dockerfile
```diff
@@ -25,49 +71,18 @@ LABEL maintainer="Microsoft" \
...
     )" \
- && apk add --virtual .rundeps $runDeps
-
-WORKDIR /
-
-# Remove CLI source code from the final image and normalize line endings.
-RUN rm -rf ./azure-cli && \
-    dos2unix /root/.bashrc /usr/local/bin/az
+ && apk add --virtual .rundeps $runDeps \
+ && ln -s /usr/local/bin/az.completion.sh /etc/profile.d/
```
jongio, line 85: Do we not need to remove the ./azure-cli folder?

Dockerfile
```diff
@@ -1,11 +1,57 @@
...
+
+#---------------------------------------------------------------------------------------------
+
+FROM base AS tools
+
+ARG JP_VERSION="0.1.3"
+
+RUN curl -L https://github.com/jmespath/jp/releases/download/${JP_VERSION}/jp-linux-amd64 -o /usr/local/bin/jp \
+ && chmod +x /usr/local/bin/jp
+
```
jongio, line 51: Does this have to be a separate stage?

hholst80: Not really unless it's planned to add additional tooling. These can be built concurrently if placed in separate stages. The serialization will happen only in the copy (now run --mount) step.

Happy to merge together it seems like too big design for something that ultimately might be removed.

Dockerfile
hholst80: This seems like a debt to me. However, now it's isolated to the build step so I'm cool with it. There is however this cost - causing confusion and questions about intent and why.

Dockerfile
hholst80: It's not in the layer. We only copy in the artifacts from /use/local that changed. So no azure-cli folder to remove.

hholst80:
> @jiasli - I'm wondering if we have any CLI perf tests to see if removing the __pycache__ impacts runtime perf.

Compare it to the overhead of starting the container, I'd *assume* it's negligible.

Actually, the total time could be larger, since the image layers need to be unpacked and copied on storage. It *could be* cheaper to just build the byte code on the fly.

Dockerfile
richlander: The repo should enforce a unix line ending standard so that this isn't needed.

Dockerfile
```diff
@@ -25,49 +71,18 @@ LABEL maintainer="Microsoft" \
...
- && apk add --no-cache --virtual .build-deps gcc make openssl-dev libffi-dev musl-dev linux-headers \
- && apk add --no-cache libintl icu-libs libc6-compat \
- && apk add --no-cache bash-completion \
- && update-ca-certificates
-
-ARG JP_VERSION="0.2.1"
-
-RUN arch=$(arch | sed s/aarch64/arm64/ | sed s/x86_64/amd64/) && curl -L https://github.com/jmespath/jp/releases/download/${JP_VERSION}/jp-linux-$arch -o /usr/local/bin/jp \
- && chmod +x /usr/local/bin/jp
+RUN --mount=from=builder,target=/mnt cd /mnt/usr/local && cp -ru . /usr/local/
```
richlander, line 74: I don't grasp why this is needed. What's the scenario where mounting produces a benefit?

Dockerfile
```diff
@@ -1,11 +1,57 @@
...
+RUN ./scripts/install_full.sh
+RUN python3 ./scripts/trim_sdk.py
+
+# Normalize line endings.
+RUN dos2unix /usr/local/bin/az /usr/local/bin/az.completion.sh
+
+# Remove __pycache__
+RUN find /usr/local -name __pycache__ | xargs rm -rf
+
+#---------------------------------------------------------------------------------------------
```
richlander, line 43: I find these lines distracting. In my project, we use two lines of whitespace to separate stages. I think that's better.

https://github.com/dotnet/dotnet-docker/blob/01184dc44bc6084e891bddd06a22584cefc218a3/src/aspnet/8.0/jammy-chiseled/amd64/Dockerfile#L14-L15

hholst80:
> The `jp` installation is also hard-coded to amd64. Do we want support for native arm64 such as apple silicon or surface users?

This was an oversight in a merge from my side.  It is included now, with support for the same architecture as in dev batch.

hholst80:
@bebound what is your final thoughts on the multi-stage build should we keep it as-is (I reduced it to a single build stage and merged the 'tooling' stage with that).  We can reduce it down to a single build stage and remove the .build-deps but I think the result is more ugly:

```Dockerfile
# syntax=docker/dockerfile:1.4
#---------------------------------------------------------------------------------------------
# Copyright (c) Microsoft Corporation. All rights reserved.
#---------------------------------------------------------------------------------------------

ARG PYTHON_VERSION="3.11"

FROM python:${PYTHON_VERSION}-alpine

ARG JP_VERSION="0.2.1"
ARG CLI_VERSION

# Metadata as defined at http://label-schema.org
ARG BUILD_DATE

LABEL maintainer="Microsoft" \
      org.label-schema.schema-version="1.0" \
      org.label-schema.vendor="Microsoft" \
      org.label-schema.name="Azure CLI" \
      org.label-schema.version=$CLI_VERSION \
      org.label-schema.license="MIT" \
      org.label-schema.description="The Azure CLI is used for all Resource Manager deployments in Azure." \
      org.label-schema.url="https://docs.microsoft.com/cli/azure/overview" \
      org.label-schema.usage="https://docs.microsoft.com/cli/azure/install-az-cli2#docker" \
      org.label-schema.build-date=$BUILD_DATE \
      org.label-schema.vcs-url="https://github.com/Azure/azure-cli.git" \
      org.label-schema.docker.cmd="docker run -v \${HOME}/.azure:/root/.azure -it mcr.microsoft.com/azure-cli:$CLI_VERSION"

# We don't use openssl (3.0) for now. We only install it so that users can use it.
# libintl and icu-libs - required by azure devops artifact (az extension add --name azure-devops)
RUN apk add --no-cache \
    bash bash-completion openssh-keygen ca-certificates openssl \
    curl jq git perl zip \
    libintl icu-libs libc6-compat \
 && update-ca-certificates

RUN --mount=target=/mnt --mount=type=cache,target=/root/.cache <<HERE
  apk add --no-cache --virtual .build-deps gcc make openssl-dev libffi-dev musl-dev linux-headers

  mkdir /azure-cli
  cd /mnt
  cp -ar . /azure-cli

  cd /azure-cli
  ./scripts/install_full.sh
  python3 ./scripts/trim_sdk.py

  cd /
  rm -rf /azure-cli

  /usr/local/bin/az && bash -c "source /usr/local/bin/az.completion.sh"
  ln -s /usr/local/bin/az.completion.sh /etc/profile.d/
  find /usr/local/lib/python*/site-packages/azure -name __pycache__ | xargs rm -rf

  arch=$(arch | sed s/aarch64/arm64/ | sed s/x86_64/amd64/)
  curl -L https://github.com/jmespath/jp/releases/download/${JP_VERSION}/jp-linux-${arch} -o /usr/local/bin/jp
  chmod +x /usr/local/bin/jp
  
  apk del --no-cache .build-deps
HERE

ENV AZ_INSTALLER=DOCKER
CMD bash

```

bebound:
I've created a draft pr to remove the unnecessary packages: https://github.com/Azure/azure-cli/pull/27567

Removing `COPY . /azure-cli` can be done by another PR.

Removing `__pycache__` saves the images size, but it increases the container size.
We created a PR which only keeps pyc file: https://github.com/Azure/azure-cli/pull/25801, but we did not merge it.

hholst80:
> Removing pyc file impacts the execution time and increases container size, which makes no sense.

That was not what my benchmark showed though. There was no impact on az --version nor az --help. The container size can be fixed by backing in `PYTHONDONTWRITEBYTECODE=1`.

hholst80:
I would too if I saw that kind result.

1. What py cache files did you remove. Removing the standard lib pycache offers no benefit as they are anyway baked in the base.

2. How and what sub command did you benchmark, and arguments

hholst80:
OK. So the az command raw without any options seems to take a considerable additional time.  There is some 5.5 million lines of Python code so if a significant part of that is pulled in it would of course have a significant impact on the execution time.

hholst80:
@bebound We can close this then? Perhaps a refactor of the Dockerfile could be interesting at a later time, but the primary purpose was to reduce image size.  This has been achieved.

PR #25185, opened 2023-01-20
{CI} Fix azure cli live test send email

wangzelin007:
Abandoning `Twilio SendGrid` for several reasons:
1. `Twilio SendGrid` is not a Microsoft product
2. The `Twilio SendGrid` we created before sends email and returns 202, but the actual sending is not successful, it may be blocked by `Twilio SendGrid`
3. Creating a new `Twilio SendGrid` needs to add a credit card, so it cannot be created successfully
image

After many attempts, I decided to use the azure service `Azure Communication Email` instead of `Twilio SendGrid`
Test screenshot:
image

TODO:
How can we stop Outlook from recognizing this type of mail as external?

### DIFF
--- a/scripts/live_test/CLITest.yml
+++ b/scripts/live_test/CLITest.yml
@@ -464,7 +464,7 @@ jobs:
         ls $(System.ArtifactsDirectory)
         pwd
         commit_id=`git ls-remote https://github.com/Azure/azure-cli.git HEAD`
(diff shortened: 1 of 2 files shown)


yonzhan:
Fix azure cli live test sending email

yonzhan (approved):

kairu-ms (approved):

PR #25223, opened 2023-01-24
[App Service] `az functionapp create`: Add new parameter `--environment` to support setting the name of container app environment

kamperiadis:
**Related command**
az functionapp create

**Description**
As part of the Centauri effort, we need to add the `--environment` parameter to the `az functionapp create` command so that customers can pass the managed AKS environment where the function app will be deployed.

**Testing Guide**
1. Create an App Service plan (`az functionapp plan create -g <resource-group-name> -n <plan-name> --sku S1`)
2. Create an managed AKS environment(`az containerapp env create -n <managed-environment-name> -g <resource-group-name>`)
3. Create a function app in the managed AKS environment (`az functionapp create -g <resource-group-name> -n <function-app-name> --plan <plan-name> -s <storage-account-name> --environment <managed-environment-name>`)

**History Notes**

[App Service] `az functionapp create`: Add new parameter `--environment` to support setting the name of container app environment

---


yonzhan:
App Service

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -352,6 +352,13 @@ def _get_subnet_info(cmd, resource_group_name, vnet, subnet):
     subnet_info["subnet_subscription_id"] = subscription_id
     return subnet_info
 
+def get_managed_environment(cmd, resource_group_name, environment_name):
+    try:
+        appcontainers_client = appcontainers_client_factory(cmd.cli_ctx)
+        return appcontainers_client.managed_environments.get(resource_group_name, environment_name)
+    except Exception as ex:  # pylint: disable=broad-except
+        logger.warning("Retrieving managed environment failed with an exception:'%s'", ex)
+        raise ex
```
kaibocai, line 361:
```suggestion
error_message="Retrieving managed environment failed with an exception: " + ex //something like this
recommendation_message="Please verify your managed environment is valid" //something like this
raise ResourceNotFoundError(error_message, recommendation_message)
```

src/azure-cli/azure/cli/command_modules/appservice/tests/latest/test_functionapp_commands.py
```diff
@@ -508,6 +508,77 @@ def test_functionapp_consumption_linux_powershell(self, resource_group, storage_
...
+
+        self.cmd('functionapp create -g {} -n {} -p {} -s {} --environment {} --runtime dotnet --functions-version 4'
+                 .format(resource_group, functionapp_name, plan_name, storage_account, managed_environment_name)).assert_with_checks([
+                     JMESPathCheck('state', 'Running'),
+                     JMESPathCheck('name', functionapp_name),
+                     JMESPathCheck('kind', 'functionapp,linux'),
+                     JMESPathCheck('hostNames[0]', functionapp_name + '.azurewebsites.net')])
+
+        self.cmd('functionapp config show -g {} -n {}'.format(resource_group, functionapp_name), checks=[
+            JMESPathCheck('linuxFxVersion', 'DOCKER|mcr.microsoft.com/azuredocs/containerapps-helloworld:latest')])
```
kaibocai, line 540: with a normal function app, do we expect a different image here?

kaibocai (approved):
LGTM

src/azure-cli/azure/cli/command_modules/appservice/tests/latest/test_functionapp_commands.py
kamperiadis: Yes, for a normal function app, this value would just be the image that the customer passed. But for function apps deployed on a managed environment, if the customer does not pass any image parameter, we want to use the same default hello world image used by container apps.

kaibocai: so, it's already code in backend that this image will be the default image to be used right.

src/azure-cli/azure/cli/command_modules/appservice/tests/latest/test_functionapp_commands.py
```diff
@@ -508,6 +508,77 @@ def test_functionapp_consumption_linux_powershell(self, resource_group, storage_
...
+        )
+
+        self.cmd('containerapp env create --name {} --resource-group {} --location {}'
+        .format(managed_environment_name, resource_group, LINUX_ASP_LOCATION_WEBAPP)).assert_with_checks([
+                     JMESPathCheck('name', managed_environment_name),
+                     JMESPathCheck('resourceGroup', resource_group),
+                     JMESPathCheck('location', LINUX_ASP_LOCATION_WEBAPP)])
+
+        self.cmd('functionapp create -g {} -n {} -c {} -s {} --environment {} --runtime dotnet --functions-version 4'
+                 .format(resource_group, functionapp_name, WINDOWS_ASP_LOCATION_FUNCTIONAPP, storage_account, managed_environment_name), expect_failure=True)
```
kaibocai, line 581: If we are checking failure for consumption plan, should we exclude the mismatch location failure here? Thanks.

src/azure-cli/azure/cli/command_modules/appservice/tests/latest/test_functionapp_commands.py
kamperiadis: I added the image to the _constants.py file

src/azure-cli/azure/cli/command_modules/appservice/_params.py
```diff
@@ -374,6 +369,26 @@ def load_arguments(self, _):
...
+
+    with self.argument_context('functionapp config container') as c:
+        c.argument('registry_server', options_list=['--registry-server', '-r', c.deprecate(target='--docker-registry-server-url', redirect='--registry-server')],
+                   help='the container registry server url')
+        c.argument('image', options_list=['--image', '-c', '-i', c.deprecate(target='--docker-custom-image-name', redirect='--image')],
+                   help='the container custom image name and optionally the tag name (e.g., <registry-name>/<image-name>:<tag>)')
+        c.argument('registry_username', options_list=['--registry-username', '-u', c.deprecate(target='--docker-registry-server-user', redirect='--registry-username')],
+                   help='the container registry server username')
+        c.argument('registry_password', options_list=['--registry-password', '-p', c.deprecate(target='--docker-registry-server-password', redirect='--registry-password')],
+                   help='the container registry server password')
```
zhoxing-ms, line 390: Why do you want to rename these option names?

src/azure-cli/azure/cli/command_modules/appservice/_params.py
```diff
@@ -703,6 +718,12 @@ def load_arguments(self, _):
     with self.argument_context('functionapp create') as c:
         c.argument('vnet', options_list=['--vnet'], help="Name or resource ID of the regional virtual network. If there are multiple vnets of the same name across different resource groups, use vnet resource id to specify which vnet to use. If vnet name is used, by default, the vnet in the same resource group as the webapp will be used. Must be used with --subnet argument.")
         c.argument('subnet', options_list=['--subnet'], help="Name or resource ID of the pre-existing subnet to have the webapp join. The --vnet is argument also needed if specifying subnet by name.")
+        c.argument('environment', options_list=['--environment'], help="Name of the container app environment.", is_preview=True)
```
zhoxing-ms, line 721: The default option name does not need to be explicitly declared
```suggestion
        c.argument('environment', help="Name of the container app environment.", is_preview=True)
```

src/azure-cli/azure/cli/command_modules/appservice/_params.py
kamperiadis: We want to rename these option names for two reasons:
1. They currently give the impression that only docker is supported, but ACR is also supported
2. We want to be consistent with [container apps commands](https://learn.microsoft.com/en-us/cli/azure/containerapp?view=azure-cli-latest#az-containerapp-create) which have the option names `registry-password`, `registry-username` and `registry-server`. 
At the same time, we want to make sure that there is backward compatibility for the customers who have created scripts that rely on the old naming conventions, which is why we just marked the old option names as deprecated.

zhoxing-ms (approved):

PR #25236, opened 2023-01-25
[Network] `az network dns zone export`: Fix the export to emit all ALIAS records for a particular record set name

Arnavion:
**Related command**
`az network dns zone export`

**Description**
Let's say there are two ALIAS records for the same record set name:

    www 3600 AZURE ALIAS A    /subscriptions/...
    www 3600 AZURE ALIAS AAAA /subscriptions/...

Before this change, export would only emit the last record for any record set name, in this case the AAAA record, because when it processed the second record it would incorrectly clear the array containing the first record.

With this change, the array is only initialized when the first record is processed.

**Testing Guide**
I had this problem with my DNS Zone, so I made the change in my local `azure-cli` installation and verified the export is now correct.

I did not find any checked-in tests for `export_zone`.

---


Arnavion:

yonzhan:
Network

necusjz (approved):

PR #25277, opened 2023-01-30
[Backup] `az backup vault create`: Add new parameter `--immutability-state` to support configuring immutability settings for the backup vault

zubairabid:
**Related command**

`az backup vault create`

**Description**

(Tested using the recoveryservices SDK version 2022-10-01) The PR adds support for the updated recoveryservices SDK that specifies an `immutability-state` flag for vaults, impacting vault creation/editing/display. Vault display (`az backup vault show`) is handled by the existing code, we need to modify behaviour of vault create/delete (Both handled by `az backup vault create`).

**Testing Guide**

`az backup vault show`

When querying for a vault which has immutability-state set to any of the three values, under "properties" we should see:

```
"securitySettings": {
    "immutabilitySettings": {
        "state": "<Value>"
    },
    "multiUserAuthorization": "Invalid",
    "softDeleteSettings": null
},
```

`az backup vault create`

When creating a new vault, we can set the flag manually: if we set it to "Disabled", the result will be "immutabilitySettings.state: Disabled", and likewise for "Unlocked" or "Locked".
This can also be used to edit vaults.

**History Notes**

---


yonzhan:
Backup

zubairabid:
Could we assign this the Jan 2023 milestone? IIUC we need to complete this PR by February 1st for that. Else I can push the changes under <https://github.com/Azure/azure-cli/pull/24699>, which is also related to the same SDK upgrade.

zhoxing-ms:
@zubairabid Don't worry. The release time of the extension is flexible and not limited by the release date. Please remind us to review when your PR is ready and resolves all CI issues

src/azure-cli/azure/cli/command_modules/backup/_params.py
```diff
@@ -293,6 +294,7 @@ def load_arguments(self, _):
 
     with self.argument_context('backup protection disable') as c:
         c.argument('delete_backup_data', arg_type=get_three_state_flag(), help='Option to delete existing backed up data in the Recovery services vault.')
+        c.argument('retain_recovery_points_as_per_policy', arg_type=get_three_state_flag(), options_list=['--retain-recovery-points-as-per-policy', '--retain-as-per-policy'], help='Switch parameter that specifies that existing recovery points should be retained for the duration specified by the backup policy')
```
zhoxing-ms, line 297: Can we define this parameter as `action='store_true'` instead of `arg_type=get_three_state_flag()`

zubairabid: This should work, but it will introduce inconsistencies with previously defined flags that use the `arg_type=get_three_state_flag()` definition. IIUC, `action='store_true'` will break the `az command --flag <value>` workflow and only allow `az command --flag`.

zubairabid: If that's alright, please resolve the conversation and merge.

zhoxing-ms (approved):

PR #25286, opened 2023-01-30
[Feedback] Stop including error messages in the feedback body

jiasli:
**Related command**
`az feedback`

**Description**
Stop including error messages in the feedback body.

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/feedback/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/feedback/custom.py
@@ -68,7 +68,7 @@ _ISSUES_TEMPLATE = """
 
 **Errors:**
 ```
-Insert here the error message you have received. Make sure to remove all sensitive information,
+Paste here the error message you have received. Make sure to remove all sensitive information,
 such as user name, password, credential, subscription ID, etc.
 ```
 


yonzhan:
Az Feedback refinement to fix an issue.

src/azure-cli/azure/cli/command_modules/feedback/custom.py
```diff
@@ -67,9 +67,7 @@
 `{command_name}`
 
 **Errors:**
-```
-{errors_string}
-```
+_Put here the error message you received. Make sure to redact sensitive information, such as personally identifiable information (PII), user name, password, credential, etc._
```
dcaro, line 70:
```suggestion
_Insert here the error message you have received. Make sure to redact any sensitive information, such as user name, password, credential, subscription id, etc._
```

dcaro (approved):
overall lgtm, just a suggestion to improve the message.

src/azure-cli/azure/cli/command_modules/feedback/custom.py
jiasli: I think PII is a well-known term. 🙂 https://www.dol.gov/general/ppii

dbradish-microsoft (changes requested):
I do like @dcaro's rewrite, but I simplified it some more.  There isn't anything wrong with using PII, but what is similar for our readers and language translators?

src/azure-cli/azure/cli/command_modules/feedback/custom.py
```diff
@@ -68,7 +68,8 @@
 
 **Errors:**
 ```
-{errors_string}
+Put here the error message you received. Make sure to redact sensitive information, such as personally
+identifiable information (PII), user name, password, credential, subscription ID, etc.
```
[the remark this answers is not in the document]
jiasli: This line is surrounded by ```` ``` ````, so `_` won't work here. I am also wondering which character we should use as a placeholder.

src/azure-cli/azure/cli/command_modules/feedback/custom.py
jiasli: I choose the word "put" because we already use it in the following section.

https://github.com/Azure/azure-cli/blob/9cf273b8070ba4bdeb17a92bf7223ef39fce94da/src/azure-cli/azure/cli/command_modules/feedback/custom.py#L77

As for "have", do we really have to use present perfect tense? Simple past tense "received" feels good to me. No need to emphasize the "have received" status.

Also, "any" only serves as an emphasis, so it seems a little bit redundant. But, I am not a native speaker, so may be wrong.

dbradish-microsoft (approved):
@jiasli , I see you have kept "Insert" vs your preferred "put".   Maybe "paste" would be even better, but I'm fine with whatever you use.

src/azure-cli/azure/cli/command_modules/feedback/custom.py
```diff
@@ -68,7 +68,8 @@
 
 **Errors:**
 ```
-{errors_string}
+Insert here the error message you have received. Make sure to remove all sensitive information,
```
jiasli, line 71:
```suggestion
Paste here the error message you have received. Make sure to remove all sensitive information,
```

yonzhan (approved):

kairu-ms (approved):

PR #25475, opened 2023-02-16
Add a warning for the deletion command

lalexedwards:
The deletions are permanent so we are working on adding warning for these locations where customer can pull the information to perform the deletions.  Please add the below to any AZ command for DevTest labs that is for deleting:
> [!CAUTION]
> Deleting resources from your lab is permanent, and cannot be undone.

**Related command**

**Description**

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/lab/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/lab/_help.py
@@ -105,6 +105,8 @@ parameters:
 """
 
 helps['lab environment delete'] = """
+[!CAUTION]
+Deleting resources in your lab is permanent, and cannot be undone.
 type: command
 short-summary: Delete an environment from a lab.
 """


yonzhan:
lab

src/azure-cli/azure/cli/command_modules/lab/_help.py
```diff
@@ -105,6 +105,8 @@
 """
 
 helps['lab environment delete'] = """
+[!CAUTION]
+Deleting resources in your lab is permanent, and cannot be undone.
```
jiasli, line 109: This statement almost applies to all Azure resources. I think this is already a common sense.

lalexedwards: @jiasli while I can understand that this may be common sense to some, we should not assume what another person does or does not already know.  This is part of an initiative with my supportability work to prevent more CritSits from being created for this exact problem.  So, with the data that I have for case creation this is not something that is inherently common sense to all.

Issue #23716, opened 2022-08-30
Error upgrading azure cli using Homebrew


elaamrani:
## Describe the bug

**Command Name**
`az upgrade`

**Errors:**
```
Auto upgrade failed. name 'exit_code' is not defined
Traceback (most recent call last):
  File "/opt/homebrew/Cellar/azure-cli/2.38.0/libexec/lib/python3.10/site-packages/knack/cli.py", line 231, in invoke
  File "/opt/homebrew/Cellar/azure-cli/2.38.0/libexec/lib/python3.10/site-packages/azure/cli/core/commands/__init__.py", line 663, in execute
  File "/opt/homebrew/Cellar/azure-cli/2.38.0/libexec/lib/python3.10/site-packages/azure/cli/core/commands/__init__.py", line 726, in _run_jobs_serially
  File "/opt/homebrew/Cellar/azure-cli/2.38.0/libexec/lib/python3.10/site-packages/azure/cli/core/commands/__init__.py", line 697, in _run_job
  File "/opt/homebrew/Cellar/azure-cli/2.38.0/libexec/lib/python3.10/site-packages/azure/cli/core/commands/__init__.py", line 333, in __call__
  File "/opt/homebrew/Cellar/azure-cli/2.38.0/libexec/lib/python3.10/site-packages/azure/cli/core/commands/command_operation.py", line 121, in handler
  File "/opt/homebrew/Cellar/azure-cli/2.38.0/libexec/lib/python3.10/site-packages/azure/cli/command_modules/util/custom.py", line 154, in upgrade_version
  File "/opt/homebrew/Cellar/python@3.10/3.10.5/Frameworks/Python.framework/Versions/3.10/lib/python3.10/importlib/__init__.py", line 168, in reload
ModuleNotFoundError: spec not found for the module 'subprocess'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/homebrew/Cellar/python@3.10/3.10.5/Frameworks/Python.framework/Versions/3.10/lib/python3.10/runpy.py", line 196, in _run_module_as_main
  File "/opt/homebrew/Cellar/python@3.10/3.10.5/Frameworks/Python.framework/Versions/3.10/lib/python3.10/runpy.py", line 86, in _run_code
  File "/opt/homebrew/Cellar/azure-cli/2.38.0/libexec/lib/python3.10/site-packages/azure/cli/__main__.py", line 48, in <module>
[...]

PR #25496, opened 2023-02-17
{Misc} Fix name 'xxx' is not defined error in __main__.py

bebound:
**Description**

In [these issues](https://github.com/Azure/azure-cli/issues?q=is%3Aissue+is%3Aopen+%22name+%27exit_code%27+is+not+defined%22), the error message `Auto upgrade failed. name 'exit_code' is not defined` is useless.

I think `handle_exception` should be able to handle all exceptions, but sometimes it fails (In the above issues, most of them are result from import error). And `knack` raise error in [invoke()](https://github.com/microsoft/knack/blob/11f2454b8e4d977797057231d0850e7935b26fa6/knack/cli.py#L245), and `exit_code` is not defined.
https://github.com/Azure/azure-cli/blob/359a85d804eeac3b7a06407594150faf122977d8/src/azure-cli-core/azure/cli/core/util.py#L56

So I set `exit_code` to None, which makes the logic robust and removes `Name 'exit_code' can be undefined` in Python linter.

I also fix the same error for `az_upgrade_run`, `upgrade_exit_code` and `invoke_finish_time`.

---

### DIFF
--- a/src/azure-cli/azure/cli/__main__.py
+++ b/src/azure-cli/azure/cli/__main__.py
@@ -62,15 +62,12 @@ except SystemExit as ex:  # some code directly call sys.exit, this is to make su
     raise ex
 
 finally:
-    try:
-        # Log the invoke finish time
-        invoke_finish_time = timeit.default_timer()
-        logger.info("Command ran in %.3f seconds (init: %.3f, invoke: %.3f)",
-                    invoke_finish_time - start_time,
-                    init_finish_time - start_time,
-                    invoke_finish_time - init_finish_time)
-    except NameError:
-        pass
+    # Log the invoke finish time
+    invoke_finish_time = timeit.default_timer()
+    logger.info("Command ran in %.3f seconds (init: %.3f, invoke: %.3f)",
+                invoke_finish_time - start_time,
+                init_finish_time - start_time,
+                invoke_finish_time - init_finish_time)
 
     try:
         # check for new version auto-upgrade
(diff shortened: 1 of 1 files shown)


yonzhan:
Misc

src/azure-cli/azure/cli/__main__.py
```diff
@@ -61,15 +62,12 @@ def cli_main(cli, args):
...
 
 finally:
-    try:
-        # Log the invoke finish time
-        invoke_finish_time = timeit.default_timer()
-        logger.info("Command ran in %.3f seconds (init: %.3f, invoke: %.3f)",
-                    invoke_finish_time - start_time,
-                    init_finish_time - start_time,
-                    invoke_finish_time - init_finish_time)
-    except NameError:
```
bebound, line 71: `NameError` is introduced in https://github.com/adewaleo/azure-cli/blob/26a3ddf7330b2958ed0fa40cfb74cd8477013ff7/src/azure-cli/azure/cli/__main__.py#L58-L62.
I think `NameError` is not raised unless `telemetry.start()` fails.

jiasli: Indeed, `NameError` should never be **caught and handled** by us. Instead, it should be **avoided** in the code itself.

src/azure-cli/azure/cli/__main__.py
```diff
@@ -91,6 +89,8 @@ def cli_main(cli, args):
                     update_all = az_cli.config.getboolean('auto-upgrade', 'all', True)
                     prompt = az_cli.config.getboolean('auto-upgrade', 'prompt', True)
                     cmd = ['az', 'upgrade', '--all', str(update_all)]
+                    az_upgrade_run = False
+                    upgrade_exit_code = None
```
jiasli, line 93: PyCharm actually already warns about the `NameError`:

image

jiasli:
While this PR makes auto-upgrade more robust, we still need to figure out why `ModuleNotFoundError: No module named 'jmespath'` happens and makes `handle_exception` crash.

jiasli (approved):

bebound:
Most of the errors result from unofficial package, except https://github.com/Azure/azure-cli/issues/23716, which installs from Homebrew.
I'll investigate it if more users report this.

PR #25541, opened 2023-02-22
[ARM] Fix #25510: `az bicep`: Set bicep.use_binary_from_path to false when installed using Azure CLI

### LINKED ISSUE #25510: CLI 2.45.0 breaks Github Actions that tried to install a specific version of Bicep

**Related command**
Run on Github actions
az bicep install --version <version> 
az bicep version
<outputs latest bicep version. Not installed one>

**Describe the bug**
New behaviour (use_binary_from_path) / if_found_in_ci has broken Bicep pinning by users. Coupled with breaking changes in Bicep versions, this has resulted in broken CI production build / deploys.

**To Reproduce**
From a Github Action
az bicep install --version <version> 
az bicep build --file main.bicep

#metadata in published file is no longer the explicitly installed bicep version

**Expected behavior**
Either document a breaking change (https://github.com/MicrosoftDocs/azure-docs-cli/blob/main/docs-ref-conceptual/release-notes-azure-cli.md) doesn't call this as breaking, or revert it to being intuitive where an explicit az bicep install results in the expected version being used.

**Environment summary**
Github Runner ubuntu:latest

**Additional context**
None. This has broken production builds.

PramodValavala-MSFT:
**Related command**

```sh
az bicep
```

**Description**

The `bicep.use_binary_from_path` config has been recently introduced to using the `bicep` binary available in the path which is especially useful in CI environments where `bicep` is already available and prevent downloading it on every run.

But for customers that rely on `az bicep install --version {}` to install the bicep version of their choice, the default is still to use the one available from the CI environment, which broke customer builds as called out in [this issue](https://github.com/Azure/azure-cli/issues/25510).

This PR includes changes that set the config value to false when `az bicep install` or `az bicep upgrade` is called, which allows subsequent `az bicep` commands to use the manually installed version, which is expected for customers calling this command. Also, there are changes which remove explicit checking for bicep installation in the publish, restore, and generate-params commands.

resolves #25510

**Testing Guide**

```sh
# Install latest bicep through other means and make sure it is in the path
az bicep version # Output should be that of the binary installed

az bicep install --version {} # Use a different version, say v0.3.1
az bicep version # Output should be that of the v0.3.1 binary installed

az config get bicep.use_binary_from_path # value should be false
```

---


PramodValavala-MSFT:
@shenglol This is an extension to a change that you have made recently. Could you please review and let me know what you think? Thanks!

src/azure-cli/azure/cli/command_modules/resource/_bicep.py
```diff
@@ -138,6 +138,8 @@ def ensure_bicep_installation(release_tag=None, target_platform=None, stdout=Tru
 
         os.chmod(installation_path, os.stat(installation_path).st_mode | stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH)
 
+        cli_ctx.config.set_value("bicep", "use_binary_from_path", "false")
```
shenglol, line 141: Could we add a log/warning to notify the user about the config change if `use_binary_from_path` is not set to `false`?

shenglol:
@PramodValavala-MSFT Thanks for the fix. I think changing the config to false when `--version` is specified makes sense. I just added one comment, but otherwise it looks good!

src/azure-cli/azure/cli/command_modules/resource/_bicep.py
PramodValavala-MSFT: Done 👍

src/azure-cli/azure/cli/command_modules/resource/_bicep.py
```diff
@@ -138,6 +138,9 @@ def ensure_bicep_installation(release_tag=None, target_platform=None, stdout=Tru
 
         os.chmod(installation_path, os.stat(installation_path).st_mode | stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH)
 
+        _logger.warning("The configuration value of bicep.use_binary_from_path has been set to 'false'.")
```
shenglol, line 141: Should we only log warning if the current configuration value is not `false`?

PramodValavala-MSFT: I had thought that would never happen after the first time but forgot about the `az bicep upgrade` command. Thanks for catching that!

Done 👍

yonzhan:
@zhoxing-ms for awareness

zhoxing-ms (approved):

ozbillwang:
Seems this PR made trouble, to break our current Azure Pipeline. 

https://github.com/Azure/azure-cli/issues/25710

We have to run a command 
```
az config set bicep.use_binary_from_path=False
```

before `az bicep install` and `az deployment` commands.

@zhoxing-ms @yonzhan 

will you rollback this change or make a fix on it? what's the plan?

shenglol:
@ozbillwang - there is a PR to fix the issue: https://github.com/Azure/azure-cli/pull/25729. We may need to roll out a hotfix for it once the PR is merged.

PR #25564, opened 2023-02-24
{Core} `az login`: Add extra empty line to separate the tenant warning from output

jiasli:
**Related command**
`az login`

**Description**
Currently, if some tenant contains no accessible subscription, Azure CLI shows a warning as below:

image

However, it can confuse the user and make the user think the JSON contains inaccessible tenants.

This PR adds an extra empty line to separate the warning from output.

image

Per my understanding, a better place to show the warning is after the JSON, in the post-output hint (https://github.com/Azure/azure-cli/pull/16242), since if the JSON is very long, it will "flood" the warning message. The user will have to scroll all the way up to see it. However, the post-output hint feature along with its implementation detail is still under discussion.


yonzhan:
Add @dcaro to share some comments for above two proposals.

Issue #24542, opened 2022-11-02
Consider using an improved JMESPath library


springcomp:
More than a feature request, this issue is my formal way to introduce [JMESPath Community](https://jmespath.site/main/).

For many months we at JMESPath Community worked on [brainstorming ideas](https://github.com/jmespath-community/jmespath.spec/discussions) for new features and curated a [reasonably scoped set of new features](https://github.com/jmespath-community/jmespath.spec/discussions?discussions_q=label%3Ajep-candidate) as [our first milestone](https://github.com/jmespath-community/jmespath.spec/discussions/97).

The goal is to pick the JMESPath language were it left off and steward improvements to the specification at a quicker pace with a clear governance.

Amongst the new features being proposed one can find:

- Arithmetic expressions : allowing simple numeric operations.
- String manipulation functions.
- Ability to refer to arbitrary lexical scopes.
- and more.

As we are wrapping up and nearing completion of this first batch of improvements, we are [publishing release candidates](https://pypi.org/project/jmespath-community/) of the Python implementation of the spec.

If those sound like useful features for az CLI, I would like to enquire about the required steps and process to have this library included as a dependency to this repository.

I would be happy to submit a pull request in due course obviously. Please let me know if that is something you would consider.

yonzhan:
@jiasli for awareness

springcomp:
FYI [jmespath-community/1.1.0rc2](https://pypi.org/project/jmespath-community/1.1.0rc2/) has been published.
This marks the final release candidate before publishing the final release version in a few weeks.

springcomp:
FYI, [jmespath-community/1.1.0](https://pypi.org/project/jmespath-community/) has been released.
This marks the official first milestone of JMESPath Community.

Miraj50:
@springcomp I was looking for string manipulation functions for case-insensitive search and it seems it's not available in az cli as of today. From what I understand, these features are already available, but Azure CLI is still on an older version of JMESPath. Am I correct?

@jiasli @yonzhan Is there a plan on upgrading Azure CLI to support these new really helpful features of JMESPath?

springcomp:
@Miraj50 string manipulation functions is available in JMESPath Community.

Case-insensitive search is not, however.
This feature is subject to feedback and acceptance on those two PRs:
- [knack#272](https://github.com/microsoft/knack/pull/272)
- #24542

jiasli:
[...]

PR #21206, opened 2022-02-07
[Packaging] BREAKING CHANGE: Drop `jmespath-terminal` from docker image

jiasli:
**Description**

`jmespath-terminal` no longer installes with latest `setuptools` because it uses very old `urwid==1.2.2` (https://github.com/jmespath/jmespath.terminal/issues/19). This issue was reported in 2018 but per https://pypi.org/project/jmespath-terminal/, the latest version of `jmespath-terminal` was released on Oct 6, 2015.

This causes CI failure and blocks PR (https://github.com/Azure/azure-cli/pull/21057#issuecomment-1031013267):

https://dev.azure.com/azure-sdk/public/_build/results?buildId=1341309&view=logs&j=a3141784-dfa9-5792-78cc-844daee2d774&t=b2b6de04-64e4-5812-4fff-06c02d19f4c2&l=214

```
    error in urwid setup command: use_2to3 is invalid.
    Couldn't build the extension module, trying without it...
    ----------------------------------------
WARNING: Discarding https://files.pythonhosted.org/packages/0c/6a/c6eceee1b52004d791b1249d798e3a6ef280a8ab5645b5e5246e8e73c7cb/urwid-1.2.2.tar.gz#sha256=e122e2dee122314f5626945af4dbe15bf3de9f318c552a4c0b68c1c480852d92 (from https://pypi.org/simple/urwid/). Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
Collecting jmespath-terminal
  Downloading jmespath-terminal-0.2.0.tar.gz (6.2 kB)
Collecting Pygments<=2.0,>=1.6
  Downloading Pygments-2.0-py3-none-any.whl (672 kB)
Collecting jmespath-terminal
  Downloading jmespath-terminal-0.1.1.tar.gz (5.1 kB)
  Downloading jmespath-terminal-0.1.0.tar.gz (3.8 kB)
  Downloading jmespath-terminal-0.0.1.tar.gz (3.4 kB)
ERROR: Cannot install jmespath-terminal==0.0.1, jmespath-terminal==0.1.0, jmespath-terminal==0.1.1, jmespath-terminal==0.2.0 and jmespath-terminal==0.2.1 because these package versions have conflicting dependencies.

The conflict is caused by:
    jmespath-terminal 0.2.1 depends on urwid==1.2.2
    jmespath-terminal 0.2.0 depends on urwid==1.2.2
    jmespath-terminal 0.1.1 depends on urwid==1.2.2
[...]

PR #25590, opened 2023-02-27
Supports JMESPath Community dependencies and assets.

springcomp:
**Description**

This PR addresses #24542 by updating the current dependency on the JMESPath language used in the `--query` command-line argument.

**Testing Guide**

az CLI uses the JMESPath language to power its `--query` command-line argument.
For instance, consider the following current behaviour:

- `` az storage account list --query "[].{location: location, name:name}" ``: 

```json
[
  { "location": "westeurope", "name": "storageaccount-001" },
  { "location": "westeurope", "name": "storageaccount-002" },
  { "location": "northeurope", "name": "storageaccount-003" },
  { "location": "westus", "name": "storageaccount-004" },
  { "location": "westus", "name": "storageaccount-005" }
]
```

Here is a new query that can be performed using JMESPath Community:

```sh
az storage account list \
  --query "group_by([].{loc: location, n:name}, &loc).*.let( {key: [0].loc, values: [*].n}, &{key: key, values: values})|from_items(zip([*].key, [*].values))"
``` 

with the corresponding result:

```json
{
  "westeurope": [ "storageaccount-001", "storageaccount-002" ],
  "northeurope": [ "storageaccount-003" ],
  "westus": [ "storageaccount-004", "storageaccount-005" ]
}
```

While there is definitely a learning curve to JMESPath, this example features most of the improvements we’ve made to the specification.

The full expression can be broken down into chunks:

- `` group_by([].{loc: location, n: name}, &loc) ``: creates a new hash with the subset of interesting `location` and `names` properties from the original object. It uses the new `group_by()` function to split the result by location.
- `` .* ``: is a value projection to create an array that can be iterated upon subsequently.
- `` let( {key: [0].loc, values: [*].n}, &{key:key, values:values}) ``: expression that is evaluated for each item in the projected array. It uses the new `let()` function to take advantage of lexical scoping.
  - Its first argument initialize a new scope with the two `key` and `values` properties that contain the value of the `loc` property from the first array entry – as all entries contain the same location at this stage – and the list of names respectively.
  - Its second argument creates a hash with the two aforementioned properties.
- `` | ``: the pipe stops the projection so that later expressions operate on the full array instead of being evaluated iteratively on each array element.
- `` from_items(zip([*].key, [*].values)) ``: creates the expected result.
  - It uses the new `zip()` function to combine the list of `key` elements and the `values` arrays into a single entity.
  - It then uses the new `from_items()` functions to combine those two lists like for like to produce the resulting object shown above.

**Overview**

This PR is a draft as I would like to make the most effort to help include an updated JMESPath language with an improved feature set to az CLI. However, I’m not sure what would be the reasonable changes to do or if it should be broken into multiple pull requests. I would also likely need guidance over how to best fill the PR description template.

This PR makes the following changes:

- Replaces all mentions of the `jmespath.org` original web site to the new `jmespath.site`.
- Replaces original dependencies over `jmespath` Python package to the new `jmespath-community` Python package.
- Replaces original dependencies over `jmespath-terminal` (`jpterm` helper) Python CLI tool to the new `jmespath-community-terminal` tool in Docker images.
- Replaces original `github.com/jmespath/jp` CLI tool by the new `github.com/jmespath-community/jp` CLI tool in Docker images.
- Updates minimum Python version to `3.7.16` for some Docker image.
- Updates various documentation pages while keeping the same LICENSE requirements.

I’m not sure if all the documentation pages must be updated though.

**JMESPath Community**

This PR aims to include [JMESPath Community](https://jmespath.site/main/) assets into az CLI.

JMESPath Community is a new initiative to bring improvements to and steward the specification going forward,
as the language and its tools no longer appear to be maintained. As its first milestone,
JMESPath Community includes the following improvements to the original specification:

- JEP-11 Lexical Scoping
- JEP-12a Raw String Literals
- JEP-13 Object Manipulation functions
- JEP-14 String functions
- JEP-15 String Slices
- JEP-16 Arithmetic Expressions
- JEP-17 Root Reference
- JEP-18 Grouping
- JEP-19 Standardized Evaluation of Pipe Expressions

JMESPath Community also aims to be an umbrella for specific programming language implementations. It currently hosts the following 100% compliant implementations:

- [jmespath-community (Python)](https://pypi.org/project/jmespath-community/)
- [jmespath-community-terminal (Python)](https://pypi.org/project/jmespath-community-terminal/)
- [go-jmespath (Golang)](https://pkg.go.dev/github.com/jmespath-community/go-jmespath)
- [jp (Golang)](https://github.com/jmespath-community/jp)

I’m also the co-author of the [JMESPath.Net](https://jdevillard.github.io/JmesPath.Net/) implementation 🤐.

For the purpose of az CLI, JMESPath Community is a drop-in replacement for the original `jmespath` Python package. It maintains backwards compatibility with the original language to prevent breaking existing syntax and scripts.

The various built-in Docker images re-introduce `jpterm` which has seen its dependencies updated and [no longer breaks the CI](https://github.com/Azure/azure-cli/pull/21206).

The accompanying `jp` executable has also been brought up to standards and is included in this PR.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,10 +29,10 @@ LABEL maintainer="Microsoft" \
 # openssh - included for ssh-keygen
 # ca-certificates
 
-# curl - required for installing jp
+# wget - required for installing jp
 # jq - we include jq as a useful tool
 # pip wheel - required for CLI packaging
-# jmespath-terminal - we include jpterm as a useful tool
+# jmespath-community-terminal - we include jpterm as a useful tool
 # libintl and icu-libs - required by azure devops artifact (az extension add --name azure-devops)
 
 # We don't use openssl (3.0) for now. We only install it so that users can use it.
@@ -44,10 +44,11 @@ RUN apk add --no-cache bash openssh ca-certificates jq curl openssl openssl1.1-c
  && apk add --no-cache bash-completion \
  && update-ca-certificates
 
-ARG JP_VERSION="0.1.3"
+ARG JP_VERSION="1.1.0"
 
-RUN curl -L https://github.com/jmespath/jp/releases/download/${JP_VERSION}/jp-linux-amd64 -o /usr/local/bin/jp \
- && chmod +x /usr/local/bin/jp
+RUN curl -L https://github.com/jmespath-community/jp/releases/download/v${JP_VERSION}/jp-linux-amd64 -o /usr/local/bin/jp \
+ && chmod +x /usr/local/bin/jp \
+ && pip install --no-cache-dir --upgrade jmespath-community-terminal
 
 WORKDIR azure-cli
 COPY . /azure-cli
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -2262,7 +2262,7 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 
 ---------------------------------------------------------
 
-jmespath 0.10.0 - MIT
+jmespath-community 1.1.0 - MIT
 
 
 Copyright (c) 2013 Amazon.com, Inc.
--- a/src/azure-cli/azure/cli/command_modules/batch/tests/latest/recordings/cli-3.1.11.help.txt
+++ b/src/azure-cli/azure/cli/command_modules/batch/tests/latest/recordings/cli-3.1.11.help.txt
@@ -54,7 +54,7 @@ Global Arguments
     --help -h                     : Show this help message and exit.
     --output -o                   : Output format.  Allowed values: json, jsonc, table, tsv.
                                     Default: json.
-    --query                       : JMESPath query string. See http://jmespath.org/ for more
+    --query                       : JMESPath query string. See http://jmespath.site/main for more
                                     information and examples.
     --verbose                     : Increase logging verbosity. Use --debug for full debug logs.
 
@@ -70,7 +70,7 @@ Global Arguments
     --debug            : Increase logging verbosity to show all debug logs.
     --help -h          : Show this help message and exit.
     --output -o        : Output format.  Allowed values: json, jsonc, table, tsv.  Default: json.
-    --query            : JMESPath query string. See http://jmespath.org/ for more information and
+    --query            : JMESPath query string. See http://jmespath.site/main for more information and
                          examples.
     --verbose          : Increase logging verbosity. Use --debug for full debug logs.
 
@@ -88,7 +88,7 @@ Global Arguments
     --help -h                     : Show this help message and exit.
     --output -o                   : Output format.  Allowed values: json, jsonc, table, tsv.
                                     Default: json.
-    --query                       : JMESPath query string. See http://jmespath.org/ for more
+    --query                       : JMESPath query string. See http://jmespath.site/main for more
                                     information and examples.
     --verbose                     : Increase logging verbosity. Use --debug for full debug logs.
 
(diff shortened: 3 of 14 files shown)


yonzhan:
JMESPath Community

jiasli:
Thank you very much @springcomp for the notification and great work. 

Because of the ever-increasing usage of Azure CLI, we are focusing more and more on maintaining a stable and consistent user experience. https://jmespath.site/main/#libraries mentions the Python version of `jmespath-community` is not _Fully compliant_ with the [original version](https://jmespath.org/) of JMESPath. Even though `enable_legacy_literals=True` is set, we still need extra effort to fully test `jmespath-community` to make sure there is no unexpected breaking changes. 

However, due to other high priority tasks, this has not been planned yet. We will discuss this PR with our Product Managers and see how we should move on.

For now, instructing users to use [`jp`](https://github.com/jmespath-community/jp) instead of Azure CLI's built in `--query` argument sounds like a more plausible solution, as this will defer the decision of using which version of JMESPath to the user.

springcomp:
@jiasli Thank you for your honest feedback.

You are right, JMESPath Community is not fully compliant.

I have made efforts to make sure compatibility is preserved as much as is possible. The unit-tests in AZ CLI are all working fine with the compatibility switch. However, there are still a couple of opinionated differences with the original implementation 😏. For the sake of transparency, these are tracked here:

- [JMESPath Community Compatibility status](https://github.com/microsoft/knack/pull/272)

Instructing users to us [jp](https://github.com/jmespath-community/jp) might be a good interim solution.

Thank you for your time.

src/azure-cli/azure/cli/command_modules/batch/tests/latest/recordings/cli-3.1.11.help.txt
jiasli, line 1: Files under `/recordings` are all generated. There is no need to manually update them. They will be overwritten when running tests.

NOTICE.txt
```diff
@@ -2262,7 +2262,7 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 
 ---------------------------------------------------------
 
-jmespath 0.10.0 - MIT
+jmespath-community 1.1.0 - MIT
```
jiasli, line 2265: Does the [copyright of `jmespath-community`](https://github.com/jmespath-community/python-jmespath/blob/main/LICENSE) still belong to Amazon?

Dockerfile
```diff
@@ -29,10 +29,10 @@ LABEL maintainer="Microsoft" \
 # openssh - included for ssh-keygen
 # ca-certificates
 
-# curl - required for installing jp
+# wget - required for installing jp
```
jiasli, line 32: May I know why `wget` is preferred over `curl`?

NOTICE.txt
springcomp: We have forked the repository, so I’m not sure what is the best practice.
Also, I’m not familiar at all with respect to licenses and copyright 😱.

What do you think in your experience ?

Dockerfile
springcomp: You are right, there is no reason. I thought I had a transient error using CURL but I cannot reproduce.
I pushed a new commit to revert.

src/azure-cli/azure/cli/command_modules/batch/tests/latest/recordings/cli-3.1.11.help.txt
springcomp: Understood. I pushed a change to revert.
I assume these will cancel out when squashing.

PR #25628, opened 2023-03-01
[RDBMS] `az mysql flexible-server update`: Fix `--storage-auto-grow` parameter unable to be set

Zhaoyi61:
**Related command**
 az mysql flexible-server update --storage-auto-grow

**Description**
--storage-auto-grow cannot be set as expected value, due to false instance property name assignment. This PR modify the property name assignment.


yonzhan:
RDBMS

PR #24759, opened 2022-11-29
{CI} Increase TestDebPackages job timeout to 120 minutes

bebound:
**Description**

Some jobs reached the default timeout 60min and failed. Increase default timeout to 120 minutes.
```
The job running on agent pool-ubuntu-2004 20 ran longer than the maximum time of 60 minutes. For more information, see https://go.microsoft.com/fwlink/?linkid=2077134 
```

Ref: [Set timeout in Azure Pipeline](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=azure-devops&tabs=yaml#timeouts)

PR #25640, opened 2023-03-02
{CI} Increase Check CLI Style job timeout to 120 minutes

wangzelin007:
**Description**

image
If no timeoutInMinutes is set, the default timeout of the ADO task is 60 minutes.
Check CLI Style has recently timed out several times, so increased the timeout to 120 minutes.

But it's not a long-term solution：
We found that after migrating to the new org, the test time increased significantly.
The test machines are all single-core, and no SSD storage is used.
image
Running the same tests locally is much faster because tests can be run in parallel.
image

@YanaXu
- If the billing model is based on the total running time, there will be no additional cost for increasing the number of cores.
- And using SSD ( Solid State Driver ) will also greatly increase the speed of some tests, such as installation tests.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -507,7 +507,7 @@ jobs:
 - job: TestExtensionsLoading
   displayName: Test Extensions Loading
   condition: succeeded()
-  timeoutInMinutes: 60
+  timeoutInMinutes: 40
 
   pool:
     name: ${{ variables.ubuntu_pool }}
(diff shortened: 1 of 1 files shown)


bebound (approved):

bebound:
This reminds me of this PR: https://github.com/Azure/azure-cli/pull/24759
The server performance is so poor.

bebound (approved):

Issue #25325, opened 2023-02-01
[App Config] az appconfig kv export: escaped characters exporting to properties format


axdp944:
**Related command**
```az appconfig kv export -d file --path myConfig.properties --format properties```

**Describe the bug**
Using az cli, when exporting one or more entries in Azure app configuration to a properties file, if the entry contains a VALUE w/ certain special characters, these characters will be escaped in the file. 

**To Reproduce**
1. Create an entry in azure app config contains characters ! # : \
2. Run the CLI command above to export this entry into a .properties file format. 
3. View file and observe these characters are escaped with a backslash. 

myConfig.properties
```
appConfigValue=\!\#\:\\
```

**Expected behavior**
The values of any entry exported should render AS-IS into a properties file, just like it is with a yaml or json export. 

**Environment summary**
az-cli v2.44.1
OSX Ventura 13.1

SaurabhSharma-MSFT:
@axdp944 We are looking into it and get back to you for any additional information.

yonzhan:
route to CXP team

SaurabhSharma-MSFT:
@axdp944 I am able to reproduce this and working internally on the same. I will let you know as soon as I have any updates to share. Thanks.

SaurabhSharma-MSFT:
@axdp944 The PR for the fix has been created and it will be merged once approved. Thanks again for pointing this out, I am closing the issue based on this update, please let us know if there is something else that we can help you with.

SaurabhSharma-MSFT:
hi @axdp944 ,
I have been notified that the current behavior is expected based on the guidance on the documentation of properties file shown [here.](https://docs.oracle.com/javase/7/docs/api/java/util/Properties.html#load(java.io.Reader):~:text=The%20key%20contains,as%20described%20above.)

In azure app configuration, we allow leading and trailing whitespaces and the characters that are considered "special" in the .properties format (!, #, etc) for keys, labels, values, etc.
Not escaping these values will result in them being parsed differently in java. Eg. without escaping special characters, a key that starts with the "#" or "!" character will have the whole line treated as a comment in the .properties file and will be ignored when parsed in java code. Trailing and leading spaces will also be trimmed in keys and values, just to name a few.
Also, this breaks round-tripping, i.e, an exported key will change when reimported since the unescaped special characters can result in unpredictable behavior during parsing.

Can you please provide your use case as product team would be interested to understand and can advise you accordingly. Thanks

axdp944:
[...]

PR #25700, opened 2023-03-06
{appconfig} Fix kv export cmdlet to export special characters in .properties file

SaurabhSharma-MSFT:
**Related command**
`az appconfig kv export -d file --path myConfig.properties --format properties`

**Description**
Fix kv export cmdlet to export special characters in the exported .properties file.
Fixes [GitHub Issue 25325](https://github.com/Azure/azure-cli/issues/25325)

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
+++ b/src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
@@ -284,7 +284,7 @@ def __write_kv_and_features_to_file(file_path, key_values=None, features=None, f
                 yaml.safe_dump(exported_keyvalues, fp, sort_keys=False, width=float('inf'))
             elif format_ == 'properties':
                 for key, value in exported_keyvalues.items():
-                    fp.write('{}={}\n'.format(key, value))                
+                    fp.write('{}={}\n'.format(key, value))
     except Exception as exception:
         raise FileOperationError("Failed to export key-values to file. " + str(exception))
 
--- a/src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
+++ b/src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
@@ -283,7 +283,8 @@ def __write_kv_and_features_to_file(file_path, key_values=None, features=None, f
             elif format_ == 'yaml':
                 yaml.safe_dump(exported_keyvalues, fp, sort_keys=False, width=float('inf'))
             elif format_ == 'properties':
-                javaproperties.dump(exported_keyvalues, fp)
+                for key, value in exported_keyvalues.items():
+                    fp.write('{}={}\n'.format(key, value))                
     except Exception as exception:
         raise FileOperationError("Failed to export key-values to file. " + str(exception))
 


yonzhan:
appconfig

zhoxing-ms (approved):

zhoxing-ms:
@albertofori Could you please help review this PR?

albertofori:
Hello @SaurabhSharma-MSFT  please the current behaviour is expected based on the guidance on the documentation of properties file shown [here.](https://docs.oracle.com/javase/7/docs/api/java/util/Properties.html#load(java.io.Reader):~:text=The%20key%20contains,as%20described%20above.)

In azure app configuration, we allow leading and trailing spaces and the special characters for keys, labels, values, etc.
Not escaping these values will result in them being parsed differently in java. Eg. without escaping special characters, a key that starts with the "#" or "!" character will have the whole line treated as a comment in the .properties file and will be ignored when parsed in java code.
Also, this breaks round-tripping, i.e, an key  exported key will change when reimported since the unescaped special characters can result in unpredictable behavior during parsing. 

I would be interested to know the motivation for this change based on the customer request in the linked issue. Any information on the customer's use case could help us advise a way forward.

Thanks!

zhoxing-ms:
@SaurabhSharma-MSFT Could you please take a look at the comment from albertofori ?

SaurabhSharma-MSFT:
Hi @albertofori ,
Please find below the details on customers's use case:

> There are a couple of important counter-points here I want to make to the product team:
> 
> First, I think you are applying the escape rules for the key to the corresponding value as well. This is not what the document you referred to on Java Properties.load states:
> 
> [Java Properties.load method](https://docs.oracle.com/javase/7/docs/api/java/util/Properties.html#load(java.io.Reader))
> 
> The key contains all of the characters in the line starting with the first non-white space character and up to, but not including, the first unescaped '=', ':', or white space character other than a line terminator. All of these key termination characters may be included in the key by escaping them with a preceding backslash character;
> 
> The following example in Java should demonstrate this more clearly using a simple URL as an example:
> 
> config.properties
> 
> myurl=https://127.0.0.1:8080
> my:url=https://127.0.0.1:8080
> my\:url=https://127.0.0.1:8080
> appConfigValue=!#:\\
> App.class
> 
> public class App {
>     public static void main(String[] args) {
>         try (InputStream input = App.class.getClassLoader().getResourceAsStream("config.properties")) {
>             Properties prop = new Properties();
>             prop.load(input);
>             System.out.println(prop.getProperty("myurl"));
>             System.out.println(prop.getProperty("my"));
>             System.out.println(prop.getProperty("my:url"));
>             System.out.println(prop.getProperty("appConfigValue"));
> 
>         } catch (IOException ex) {
>             ex.printStackTrace();
>         }
>     }
> }
> Output:
> 
> https://127.0.0.1:8080
> url=https://127.0.0.1:8080
> https://127.0.0.1:8080
> !#:\
> However, if I have a simple URL value stored in Azure App Config of https://127.0.0.1:8080 and try to export it from the command line as a .properties file, the value returned contains escaped colons.
> 
> https\://127.0.0.1\:8080
> Technically, this might still be acceptable in Java but remember that other languages use .properties for configuration. In my particular use case, I am using az appconfig kv export in a CI/CD pipeline in ADO to export a subset of configuration values and using the exported properties file to create a Kubernetes configmap object using. When importing in this manner, the escaped values are treated as literals in the K8s configmap which will break the service unless I introduce a rather ugly sed command to parse the values.
> 
> kubectl create configmap my-config --from-file=config.properties
> Lastly, its important to point out that the export behavior of the az cli is also inconsistent with the behavior of the Import/Export feature within your own Azure App Configuration UI portal. If you open the Azure portal and add a configuration value of https://127.0.0.1:8080, in your app config instance, then export it to a properties file by navigating to "Import/Export" under "Operations" and select "Properties as the file type, the value will be exported without escaped characters. However, if you do the same thing with the az cli, it will export the value as https\://127.0.0.1\:8080
> 
> That's a big inconsistency in my opinion and would expect the behavior of both Azure export operations between the Portal UI and the CLI to be identical.

Please let me know if you have any other questions.

Thanks
Saurabh

SaurabhSharma-MSFT:
Hi @albertofori, Can you please look into the details shared above by the customer.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -283,7 +283,8 @@ def __write_kv_and_features_to_file(file_path, key_values=None, features=None, f
             elif format_ == 'yaml':
                 yaml.safe_dump(exported_keyvalues, fp, sort_keys=False, width=float('inf'))
             elif format_ == 'properties':
-                javaproperties.dump(exported_keyvalues, fp)
+                for key, value in exported_keyvalues.items():
+                    fp.write('{}={}\n'.format(key, value))
```
albertofori, line 287:
```suggestion
                    fp.write('{}={}\n'.format(javaproperties.escape(key), value))
```

albertofori: @SaurabhSharma-MSFT Thank you for the response and my apologies for the late reply. The response is well understood, however, since there may be some values in the key that may be considered as a separator, we should still escape the key so it is not parsed differently in the properties file. We can keep the value property as-is. 

@mrm9084 @avanigupta Could we please get your thoughts on this, and whether the proposed change is sufficient?

albertofori (changes requested):

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: I agree. We should allow unescaped special characters in the value (which is what the customer is interested in) and escape special characters in the key. 

@albertofori can you help to verify if importing unescaped values might cause problems after this change?
Also, is the behavior same in portal?

albertofori: @avanigupta The only difference that will be observed is that leading white spaces will be trimmed off values. e.g., For the following key-values in a properties file format:
image

The imported key-values will look as shown below:
image

The key `my:url` has the 2 leading spaces of the value trimmed. Trailing spaces are maintained though, as can be seen with the value of key `my`.
In order to maintain leading spaces, we can escape the first leading space.

avanigupta: Does portal have the same behavior? 
I think this is the right behavior because leading whitespaces in the value will be ignored by other parsing libraries too. This will break our round-trip capability, but leading whitespaces in properties file values are not a valid scenario anyway.

avanigupta: Sounds good, we can update the portal to escape keys. Thanks, Albert!

albertofori: @SaurabhSharma-MSFT Could you please, take a look at and accept the suggested changes to this PR?

SaurabhSharma-MSFT: Done

avanigupta (approved):

yonzhan:
Please fix CI issues.

albertofori:
@zhoxing-ms @yanzhudd , could you please help re-run the CI tests for this PR? We would want to proceed to merge this or fix any issues that currently exist.

albertofori (approved):

albertofori:
@zhoxing-ms @yanzhudd Could you please help merge in this fix as it has been approved?

Thanks a lot!

albertofori:
Issue fixed in this [PR](https://github.com/Azure/azure-cli/pull/32010). Closing this out now.

PR #25747, opened 2023-03-10
{CI} Add Test Rpm Package Mariner

bebound:
**Description**

Other builds have this test, add the same test for Mariner.

---

### DIFF
--- a/scripts/release/rpm/test_mariner_in_docker.sh
+++ b/scripts/release/rpm/test_mariner_in_docker.sh
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+
+# This script should be run in a centos7 docker.
+set -exv
+
+export USERNAME=azureuser
+
+tdnf --nogpgcheck install /mnt/rpm/$RPM_NAME -y
+
+tdnf install git gcc python3-devel python3-pip findutils -y
+
+ln -s -f /usr/bin/python3 /usr/bin/python
+time az self-test
+time az --version
+
+cd /azure-cli/
+pip install wheel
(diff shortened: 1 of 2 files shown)


yonzhan:
CI

scripts/release/rpm/test_mariner_in_docker.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
+# This script should be run in a centos7 docker.
+set -exv
+
+export USERNAME=azureuser
+
+tdnf --nogpgcheck install /mnt/rpm/$RPM_NAME -y
+
+tdnf install git gcc python3-devel python3-pip findutils -y
```
bebound, line 10: There are two differences with `test_rpm_in_docker.sh`
1. Mariner does not install `pip` when install `python3.9-devel`, it also does not have `python3.9-pip` package.
2. `dnf` becomes `tdnf`.

I don't want to add too many env variables, so I create a new script for Mariner.

jiasli, line 10: Can this script be reused from `test_rpm_in_docker.sh` by introducing some parameters?

bebound: There are some difference between Mariner and UBI, see https://github.com/Azure/azure-cli/pull/25747#discussion_r1132030505.

Using same script decreases the duplication code but increases the complexity in script.
We've used different dockerfile for Mariner and UBI. We can also use different test scripts.

jiasli (approved):

PR #25758, opened 2023-03-11
[App Config] Ensure the case of imported boolean values does not change for string conversion from file

albertofori:
**Related command**

`az appconfig kv import --source file`

**Description**

Related to [Issue 725](https://github.com/Azure/AppConfiguration/issues/725) where the case of boolean values may be changed during string conversion.

**Testing Guide**

**History Notes**

[App Config] `az appconfig kv import`: Ensure the case of imported boolean values does not change for string conversion from file

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
+++ b/src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
@@ -878,9 +878,9 @@ def __flatten_key_value(key, value, flattened_data, depth, separator):
             if key in flattened_data:
                 logger.debug(
                     "The key %s already exist, value has been overwritten.", key)
-            flattened_data[key] = str(value)
+            flattened_data[key] = value if isinstance(value, str) else json.dumps(value) # Ensure boolean values are properly stringified.
     else:
-        flattened_data[key] = str(value)
+        flattened_data[key] = value if isinstance(value, str) else json.dumps(value)
 
 
 def __export_keyvalue(key_segments, value, constructed_data):
--- a/src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
+++ b/src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
@@ -878,7 +878,7 @@ def __flatten_key_value(key, value, flattened_data, depth, separator):
             if key in flattened_data:
                 logger.debug(
                     "The key %s already exist, value has been overwritten.", key)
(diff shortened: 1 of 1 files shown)


yonzhan:
App Config

albertofori:
@avanigupta Could you please take a look at this fix? Thanks!

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -878,9 +878,9 @@ def __flatten_key_value(key, value, flattened_data, depth, separator):
             if key in flattened_data:
                 logger.debug(
                     "The key %s already exist, value has been overwritten.", key)
-            flattened_data[key] = str(value)
+            flattened_data[key] = value if isinstance(value, str) else json.dumps(value)  # Ensure boolean values are properly stringified.
```
avanigupta, line 881: Is there any change in behavior if the value is a number or null?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -878,9 +878,9 @@ def __flatten_key_value(key, value, flattened_data, depth, separator):
...
     else:
-        flattened_data[key] = str(value)
+        flattened_data[key] = value if isinstance(value, str) else json.dumps(value)
```
avanigupta, line 883: Do we need to check `isinstance(value, str)`? Cant we always do `json.dumps(value)`?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
albertofori: There is no change for behavior with a number, however, null values where being converted to the string "None". 
The `null` value is maintained in this case.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
albertofori: This is because string values will be represented with escaped quotes. e.g.,
For a json value indicated in a file as:
`{ "SomeString": "New value"  }`

The value will be exported to the configuration store with the quotes maintained. 
image

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: Great! I'm surprised we didnt get any complaints about this. Thanks for fixing it!

zhoxing-ms (approved):

PR #25795, opened 2023-03-13
[App Service] `az functionapp config container`: Handle slotConfigNames not supported

kamperiadis:
**Related command**
`az functionapp config container delete`
`az functionapp config container show`

**Description**
For Private Preview, the backend team will not support the slotConfigNames APIs which we call in the `az functionapp config container` commands to obtain/update slot configuration settings.

**Testing Guide**
1. Create function app deployed in a container app environment
2. Run `az functionapp config container set` or `az functionapp config container delete` or `az functionapp config container show`

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -1235,7 +1235,12 @@ def get_site_configs(cmd, resource_group_name, name, slot=None):
 def get_app_settings(cmd, resource_group_name, name, slot=None):
     result = _generic_site_operation(cmd.cli_ctx, resource_group_name, name, 'list_application_settings', slot)
     client = web_client_factory(cmd.cli_ctx)
-    slot_app_setting_names = client.web_apps.list_slot_configuration_names(resource_group_name, name).app_setting_names
+    slot_app_setting_names = None
+    try:
+        slot_app_setting_names = client.web_apps.list_slot_configuration_names(resource_group_name, name) \
+                                       .app_setting_names
+    except:  # pylint: disable=bare-except
+        pass
     return _build_app_settings_output(result.properties, slot_app_setting_names)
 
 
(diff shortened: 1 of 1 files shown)


yonzhan:
App Service

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -1235,7 +1235,12 @@ def get_site_configs(cmd, resource_group_name, name, slot=None):
 def get_app_settings(cmd, resource_group_name, name, slot=None):
     result = _generic_site_operation(cmd.cli_ctx, resource_group_name, name, 'list_application_settings', slot)
     client = web_client_factory(cmd.cli_ctx)
-    slot_app_setting_names = client.web_apps.list_slot_configuration_names(resource_group_name, name).app_setting_names
+    slot_app_setting_names = []
+    try:
+        slot_app_setting_names = client.web_apps.list_slot_configuration_names(resource_group_name, name) \
```
kaibocai, line 1240: so the backend will throw an exception when retrieve the appsetting for slots and from our side we just need to swallow the exception and continue proceed right. Thanks.

kamperiadis: Yep. I was on the fence on whether we should check if the backend threw an exception or if we should add a call to see if the function app is deployed in a managed environment. Both approaches should work. The benefits of just checking if an exception is thrown are that:
- we do not make any additional API calls. 
- once the backend team decides to support this API call, we might not need to make any future changes.
- we do not block execution if the API call fails (not sure if the intended experience was to block execution). We might want to consider maybe showing a warning so customers are not given the wrong idea that their slot configuration was fetched/updated when that is not always the case.
What do you think of the points above? Do you think that instead of checking if the backend threw an exception, we should be checking if the function app is deployed in a container app environment?

kaibocai: Yeah I think we can just add an warning message that should be good enough.

kamperiadis: @kaibocai I've given it some thought and if we show a warning message like 'Unable to fetch slot app settings', we might confuse customers and make them think that there is something they need to manually do on their end. I would say it might be better to check if the app is deployed in a container app environment instead of checking for any exception in general since it might be necessary for other non-Centauri related flows to half execution if the API fails.

PR #25805, opened 2023-03-14
{VM} Clean up Network SDK dependencies

kairu-ms:
**Related command**

**Description**

**Testing Guide**

**History Notes**

---


yonzhan:
VM

src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -728,15 +728,14 @@ def load_arguments(self, _):
...
         c.argument('app_gateway_subnet_address_prefix', help='The subnet IP address prefix to use when creating a new application gateway in CIDR format.')
         c.argument('backend_pool_name', help='Name to use for the backend pool when creating a new load balancer or application gateway.')
         c.argument('backend_port', help='When creating a new load balancer, backend port to open with NAT rules (Defaults to 22 on Linux and 3389 on Windows). When creating an application gateway, the backend port to use for the backend HTTP settings.', type=int)
         c.argument('load_balancer', help='Name to use when creating a new load balancer (default) or referencing an existing one. Can also reference an existing load balancer by ID or specify "" for none.', options_list=['--load-balancer', '--lb'])
-        c.argument('load_balancer_sku', resource_type=ResourceType.MGMT_NETWORK, min_api='2017-08-01', options_list=['--lb-sku'], arg_type=get_enum_type(LoadBalancerSkuName),
+        c.argument('load_balancer_sku', resource_type=ResourceType.MGMT_NETWORK, min_api='2017-08-01', options_list=['--lb-sku'], arg_type=get_enum_type(['Basic', 'Standard', 'Gateway']),
```
PARADISSEEKR, line 738: `LoadBalancerSkuName` has two values(`Basic`, `Standard`) before version `2021-02-01` and three values(`Basic`, `Standard`, `Gateway`) after.

necusjz (approved):

src/azure-cli/azure/cli/command_modules/vm/_params.py
kairu-ms: This change comes from your PR.

necusjz (approved):

src/azure-cli/azure/cli/command_modules/vm/_validators.py
```diff
@@ -981,18 +984,15 @@ def _validate_vm_vmss_create_public_ip(cmd, namespace):
...
-        namespace.public_ip_sku = PublicIPAddressSkuName.standard.value
+        namespace.public_ip_sku = 'Standard'
 
     # Public-IP SKU is only exposed for VM. VMSS has no such needs so far
     if getattr(namespace, 'public_ip_sku', None):
-        if namespace.public_ip_sku == PublicIPAddressSkuName.standard.value:
+        if namespace.public_ip_sku == 'Standard':
             if not namespace.public_ip_address_allocation:
-                namespace.public_ip_address_allocation = IPAllocationMethod.static.value
+                namespace.public_ip_address_allocation = 'Static'
```
zhoxing-ms, line 995: Just out of curiosity, may I ask whether the values defined as enumerations in Swagger can continue to be defined as enumerations in Code Gen?

src/azure-cli/azure/cli/command_modules/vm/_validators.py
```diff
@@ -1635,17 +1640,21 @@ def _validate_vmss_create_load_balancer_or_app_gateway(cmd, namespace):
...
+    if balancer_type == 'application_gateways':
+        return import_aaz_by_profile(cli_ctx.cloud.profile, "network.application_gateway")
+    if balancer_type == 'load_balancers':
+        return import_aaz_by_profile(cli_ctx.cloud.profile, "network.lb")
 
 
 def get_network_lb(cli_ctx, resource_group_name, lb_name):
     from azure.core.exceptions import HttpResponseError
-    network_client = get_network_client(cli_ctx)
+    LBShow = import_aaz_by_profile(cli_ctx.cloud.profile, "network.lb").Show
```
zhoxing-ms, line 1652: Why not use `get_network_client(cli_ctx, 'load_balancers').Show`?
Personally, I think it's more readable

src/azure-cli/azure/cli/command_modules/vm/_validators.py
```diff
@@ -734,28 +734,31 @@ def _validate_vm_vmss_create_vnet(cmd, namespace, for_scale_set=False):
         logger.debug('no subnet specified. Attempting to find an existing Vnet and subnet...')
 
         # if nothing specified, try to find an existing vnet and subnet in the target resource group
-        client = get_network_client(cmd.cli_ctx).virtual_networks
+        VnetList = import_aaz_by_profile(cmd.cli_ctx.cloud.profile, "network.vnet").List
```
zhoxing-ms, line 737: Same as above

src/azure-cli/azure/cli/command_modules/vm/_validators.py
kairu-ms: `get_network_client` is used to get client class in `azure-mgmt-network` sdk. However, `import_aaz_by_profile` can get `aaz` classes not only for `network`. You can get other modules as well. So it can be used to get vmss class like 
```python
    VMSSNicShow = import_aaz_by_profile(cli_ctx.cloud.profile, "vmss.nic").Show
```

src/azure-cli/azure/cli/command_modules/vm/_validators.py
kairu-ms: Answered above.

src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -728,15 +728,16 @@ def load_arguments(self, _):
...
         c.argument('load_balancer', help='Name to use when creating a new load balancer (default) or referencing an existing one. Can also reference an existing load balancer by ID or specify "" for none.', options_list=['--load-balancer', '--lb'])
-        c.argument('load_balancer_sku', resource_type=ResourceType.MGMT_NETWORK, min_api='2017-08-01', options_list=['--lb-sku'], arg_type=get_enum_type(LoadBalancerSkuName),
+        c.argument('load_balancer_sku', resource_type=ResourceType.MGMT_NETWORK, min_api='2017-08-01', options_list=['--lb-sku'], arg_type=get_enum_type(['Basic', 'Standard']),
+                   help="Sku of the Load Balancer to create. Default to 'Standard' when single placement group is turned off; otherwise, default to 'Basic'. The public IP is supported to be created on edge zone only when it is 'Standard'")
+        c.argument('load_balancer_sku', resource_type=ResourceType.MGMT_NETWORK, min_api='2021-02-01', options_list=['--lb-sku'], arg_type=get_enum_type(['Basic', 'Standard', 'Gateway']),
```
zhoxing-ms, line 740: > help="Sku of the Load Balancer to create. Default to 'Standard' when single placement group is turned off; otherwise, default to 'Basic'. The public IP is supported to be created on edge zone only when it is 'Standard'"

May I ask is this help message valid for `load_balancer_sku` with `min_api='2021-02-01'`?

zhoxing-ms: It is easy to be confused if the same parameter is defined twice. Could we add `max_api=xxx` to the first one with `min_api='2017-08-01'` to make the version control information more intuitive

src/azure-cli/azure/cli/command_modules/vm/_validators.py
zhoxing-ms: In fact, the `get_network_client` I want to refer to is the custom method defined in `_validators.py`
```Python
def get_network_client(cli_ctx, balancer_type):
    if balancer_type == 'application_gateways':
        return import_aaz_by_profile(cli_ctx.cloud.profile, "network.application_gateway")
    if balancer_type == 'load_balancers':
        return import_aaz_by_profile(cli_ctx.cloud.profile, "network.lb")
```

src/azure-cli/azure/cli/command_modules/vm/_validators.py
```diff
@@ -1635,17 +1640,21 @@ def _validate_vmss_create_load_balancer_or_app_gateway(cmd, namespace):
...
 
-def get_network_client(cli_ctx):
-    from azure.cli.core.profiles import ResourceType
-    from azure.cli.core.commands.client_factory import get_mgmt_service_client
-    return get_mgmt_service_client(cli_ctx, ResourceType.MGMT_NETWORK)
+def get_network_client(cli_ctx, balancer_type):
+    if balancer_type == 'application_gateways':
+        return import_aaz_by_profile(cli_ctx.cloud.profile, "network.application_gateway")
+    if balancer_type == 'load_balancers':
+        return import_aaz_by_profile(cli_ctx.cloud.profile, "network.lb")
```
zhoxing-ms, line 1647: Mark

src/azure-cli/azure/cli/command_modules/vm/_params.py
kairu-ms: > > help="Sku of the Load Balancer to create. Default to 'Standard' when single placement group is turned off; otherwise, default to 'Basic'. The public IP is supported to be created on edge zone only when it is 'Standard'"
> 
> May I ask is this help message valid for `load_balancer_sku` with `min_api='2021-02-01'`?

Good question. I only keep the help message not changed.

kairu-ms: > It is easy to be confused if the same parameter is defined twice. Could we add `max_api=xxx` to the first one with `min_api='2017-08-01'` to make the version control information more intuitive

Good suggestion, I will add max_api for it.

src/azure-cli/azure/cli/command_modules/vm/_validators.py
kairu-ms: I will remove this function.

kairu-ms: It's the old implementation by @PARADISSEEKR

zhoxing-ms:
May I ask why the current design tends to add the network multi-version folders under each module which depends on it, rather than directly call the multi-version folders from the network module?
I have concerned that these may cause some multi profile code duplication, increase the number of files and package size (may have little actual impact), and may increase the maintenance cost of bug fix in the future.
Maybe I lack some context at present, so I don't consider it comprehensively. I'm looking forward to hearing your design ideas~

src/azure-cli/azure/cli/command_modules/vm/_validators.py
zhoxing-ms: OK~

kairu-ms:
> May I ask why the current design tends to add the network multi-version folders under each module which depends on it, rather than directly call the multi-version folders from the network module? I have concerned that these may cause some multi profile code duplication, increase the number of files and package size (may have little actual impact), and may increase the maintenance cost of bug fix in the future. Maybe I lack some context at present, so I don't consider it comprehensively. I'm looking forward to hearing your design ideas~ 😉

That's a good question. The short answer is to decouple the modules. As you can see the network commands generated for vm module use 2020-01-01 however in network module they are 2022-05-01. That avoids a lot of bump up works. We use compact algorithm to reduce such size increase in packaging stage.

zhoxing-ms (dismissed):

zhoxing-ms:
> That's a good question. The short answer is to decouple the modules. As you can see the network commands generated for vm module use 2020-01-01 however in network module they are 2022-05-01. That avoids a lot of bump up works. We use compact algorithm to reduce such size increase in packaging stage.

OK, LGTM. If Code Gen can support batch upgrade of the dependent RP folders for multiple modules at the same time in the future, this will help us better maintain the additional multi-profile folders

src/azure-cli/azure/cli/command_modules/vm/_validators.py
```diff
@@ -1558,7 +1562,7 @@ def _validate_vmss_create_load_balancer_or_app_gateway(cmd, namespace):
     if balancer_type == 'applicationGateway':
 
         if namespace.application_gateway:
-            client = get_network_client(cmd.cli_ctx, 'application_gateways')
+            client = import_aaz_by_profile(cli_ctx.cloud.profile, "network.application_gateway")
```
kairu-ms, line 1565:
```suggestion
            client = import_aaz_by_profile(cmd.cli_ctx.cloud.profile, "network.application_gateway")
```

necusjz (approved):

Issue #25815, opened 2023-03-15
Initialize Cloud Config reading from ARM metadata with API Version 2022-09-01


bganapa:
**Related command**
az cloud register

**Is your feature request related to a problem? Please describe.**
Azure Stack needs custom endpoints  and many of the azure code bases have hard coded the values in various places depending on the cloud name. This will not work for disconnected custom clouds like Azure Stack as the endpoints are defined by the customer configurations.

We worked with ARM team to provide the custom dataplane endpoints through the ARM metadata with a newer api version 2022-09-01

**Describe the solution you'd like**
CLI cloud config should read the ARM metadata with api version 2022-09-01 instead of the current 1.0  and expose them in CLI Context so that all the extensions could make use of it

**Describe alternatives you've considered**

Other alternatives are hardcoding the values, or making the ARM metadata request with api version 2022-09-01 in CLI extensions like connectedk8s, k8s-extensions, k8s-configuration etc..

**Additional context**
Additional context is shared via internal mail.

The followign are the new fields added with api version 2022-09-01

"suffixes":{
        "relayEndpointSuffix":"servicebus.azs.microsoft.com",
        ...
    },
    dataplaneEndpoints":{
        "arcConfigEndpoint":"https://edge.dp.kubernetesconfiguration.azs.microsoft.com/",
        "arcGlobalNotificationServiceEndpoint":"https://guestnotificationservice.azs.microsoft.com/",
        "arcHybridIdentityServiceEndpoint":"https://his.azs.microsoft.com/",
        "globalContainerRegistryEndpoint":"edgemcr.edgeacr.azs.microsoft.com",
        "arcGuestConfigEndpoint":"https://agentserviceapi.azs.microsoft.com/"
    },

yonzhan:
@jiasli for awareness

PR #25834, opened 2023-03-16
[Cloud] Perform cloud discovery with API Version 2022-09-01 and support `dataplaneEndpoints` for Azure Stack

jiasli:
**Related command**
`az cloud`

**Description**
Close https://github.com/Azure/azure-cli/issues/25815

**Testing Guide**

Populate `dataplaneEndpoints` from metadata:

```
az cloud register -n mycloud --endpoint-resource-manager https://management.azure.com/
```

Populate `dataplaneEndpoints` from local cloud config JSON file:

```
az cloud register -n mycloud --cloud-config @D:\cloud\cloud-config.json
```

### DIFF
--- a/src/azure-cli-core/azure/cli/core/cloud.py
+++ b/src/azure-cli-core/azure/cli/core/cloud.py
@@ -521,7 +521,10 @@ def get_clouds(cli_ctx):
             if option == 'profile':
                 c.profile = config.get(section, option)
             if option.startswith('endpoint_'):
-                setattr(c.endpoints, option.replace('endpoint_', ''), config.get(section, option))
+                raw_value = config.get(section, option)
+                # endpoint_dataplane_endpoints is a JSON. It should be deserialized to a dict.
+                value = json.loads(raw_value) if len(raw_value) >= 1 and raw_value[0] == '{' else raw_value
+                setattr(c.endpoints, option.replace('endpoint_', ''), value)
             elif option.startswith('suffix_'):
                 setattr(c.suffixes, option.replace('suffix_', ''), config.get(section, option))
         if c.profile is None:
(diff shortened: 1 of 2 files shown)


yonzhan:
Cloud enhancement

TheOnlyWei:
Although the main post mentioned that a new nested group called `dataplaneEndpoints` is being supported, there should also be support for new suffixes and endpoints that is not passed by a `cloudconfig.json` file. Currently, Azure CLI is using a deprecated method to populate the hard-coded metadata which should be updated to read all endpoints and suffixes:
https://github.com/Azure/azure-cli/blob/091c6b1c40c19d4076376f9dfa0d4994b291a7c6/src/azure-cli/azure/cli/command_modules/cloud/custom.py#L69
https://github.com/Azure/msrestazure-for-python/blob/8e3def8c9927a349596e575de93df76de7a3a2b6/msrestazure/azure_cloud.py#L204

TheOnlyWei:
@jiasli 
> If these new suffixes and endpoints are not passed by a cloudconfig.json file, how will Azure CLI even know them?
The new endpoints and suffixes will appear in the metadata endpoint URL.

> Reading all endpoints and suffixes is simply against our modeling principal which enforces all endpoints to be an attribute of the cloud object.
Is there a reason there is an exception made for reading all endpoints and suffixes from the cloudconfig.json but not from the metadata endpoint?
For example, the following code just adds all endpoints from cloudconfig.json:
https://github.com/Azure/azure-cli/blob/091c6b1c40c19d4076376f9dfa0d4994b291a7c6/src/azure-cli/azure/cli/command_modules/cloud/custom.py#L46

jiasli:
> Is there a reason there is an exception made for reading all endpoints and suffixes from the cloudconfig.json but not from the metadata endpoint?

I didn't write that code, but I guess it is because `cloudconfig.json` is designed/used specifically for Azure CLI. On the other hand, metadata endpoint is merely a supplemental way to _extract_ information we want.

toefraz:
@jiasli Bumping this.

FYI 2022-09-01 is a breaking change for the metadata endpoint. Prior versions returned a JSON array of cloud endpoints for multiple environments. This version returns a single JSON object for the endpoints associated with the ARM instance being queried.

Is the goal to support both schemas with this for backwards compatibility?

PR #25843, opened 2023-03-16
[RDBMS] Fix validation for availability zone for the replica for postgresql flexible server

ambrahma:
**Related command**
az postgres flexible-server replica create --replica-name testReplicaServer -g testGroup --source-server testserver --location testLocation

**Description**
 Create a read replica 'testReplicaServer' for 'testserver' with public or private access in the specified location if available. Since zone is not passed, it will automatically pick up zone in the replica location which is different from source server, if available.
Prior behavior:
1. Blocks the replica creation in case the location is a single AZ region.
2. Selects the zone to be same as the source server always irrespective of whether that zone is supported in the replica location.
New behavior:
1. Allows replica creation even in case the selected location is a single AZ region.
2. If zone is not passed, it will automatically pick up zone in the replica location which is different from source server, if available.
3. If different zone is not available in the selected location, it will pick up zone same as source server in the replica location if available.
4. If not will set the zone as None, i.e. No preference and not pass the AZ.
5. Also refactored code to fetch data from location capability api and to move it to a separate file

**Testing Guide**
Do not pass AZ during replica creation. Zone should automatically get selected. Also select location as single AZ and now the replica creation must be allowed for single AZ region.

**History Notes**

[RDBMS] BREAKING CHANGE: `az postgres flexible-server replica create`: Fix the behavior of AZ selection in case zone is not passed as parameter
[RDBMS] Fix #368903181: Fix zone selection during creation of replica

---

- [X ] The PR title and description has followed the guideline in [Submitting Pull Requests](https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules#submitting-pull-requests).

- [X ] I adhere to the [Command Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md).

- [X ] I adhere to the [Error Handling Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/error_handling_guidelines.md).


yonzhan:
RDBMS

src/azure-cli/azure/cli/command_modules/rdbms/_helptext_pg.py
```diff
@@ -513,6 +513,11 @@
         --vnet newVnet --subnet newSubnet \\
         --address-prefixes 172.0.0.0/16 --subnet-prefixes 172.0.0.0/24 \\
         --private-dns-zone testDNS.postgres.database.azure.com
+  - name: Create a read replica 'testReplicaServer' for 'testserver' with public or private access \\
+        in the specified location if available. SInce zone is not passed, it will automatically pick up zone in the \\
```
alanenriqueo, line 517: nit: Since

ambrahma:
azp /run

ambrahma:
@evelyn-ys how do I rerun pipeline for tiltle and content fix?

pradeepkintali (approved):

alanenriqueo (approved):

evelyn-ys:
@ambrahma May I know why you need to make the default zone breaking change? Usually we only allow breaking change twice a year and next breaking change window is May release for Build

ambrahma:
> @ambrahma May I know why you need to make the default zone breaking change? Usually we only allow breaking change twice a year and next breaking change window is May release for Build

Hi @evelyn-ys , the behavior which was implemented in the CLI as actually a bug with the CLI implementation. Customers are having 2 different experiences using portal and CLI. WHile the portal behavior is correct for AZ selection, CLI was not implemented correctly in the first place and we missed to catch this issue earlier during GA release for api version 2022-12-01. THis is an actual CRI we received from the customer regarding this issue [Incident 368903181](https://icmcdn.akamaized.net/imp/v3/incidents/details/368903181/home) : [S500] [CSS SEV B] [Prem] [PostgreSQL] [Telstra] - [create geo-redundant replica fails]

evelyn-ys:
@ambrahma so the current behavior will cause 100% failure, right? And after this PR getting merged, it will succeed with correct behavior, right?

ambrahma:
> @ambrahma so the current behavior will cause 100% failure, right? And after this PR getting merged, it will succeed with correct behavior, right?

@evelyn-ys , yes that's right. It will always fail currently. After getting this PR merged, it will succeed with the correct behavior.

evelyn-ys (approved):

PR #25847, opened 2023-03-16
Clearing values between enable/disable static_website

adamsba3:
**Related command**
az storage blob service-properties update --account-name account_name --static-website false

**Description**
Fix issue with enable/disable static website where values are not cleared causing issue below after static was enabled

**Testing Guide**
  Create Blob with static content enabled
  Command line disable static content
  Command line enable static content again

Error:
        "ERROR: XML specified is not syntactically valid.",
        "RequestId:5cca2535-201e-0029-094c-427c79000000",
        "Time:2023-02-16T21:23:46.4764389Z",
        "ErrorCode:InvalidXmlDocument",
        "linenumber:2",
        "lineposition:714",
        "reason:Element IndexDocument is only expected when StaticWebsite/Enabled is enabled

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/storage/operations/blob.py
+++ b/src/azure-cli/azure/cli/command_modules/storage/operations/blob.py
@@ -350,6 +350,10 @@ def set_service_properties_track2(client, parameters, delete_retention=None, del
     if not parameters.minute_metrics.enabled:
         parameters.minute_metrics.include_apis = None
 
+    if not parameters.static_website.enabled:
+        parameters.static_website.index_document = None
+        parameters.static_website.error_document404_path = None
+
     # checks
     policy = kwargs.get('delete_retention_policy', None)
     if policy and policy.enabled and not policy.days:


yonzhan:
storage

adamsba3:
I updated my description thought it would be good now

src/azure-cli/azure/cli/command_modules/storage/operations/blob.py
```diff
@@ -350,6 +350,10 @@ def set_service_properties_track2(client, parameters, delete_retention=None, del
     if not parameters.minute_metrics.enabled:
         parameters.minute_metrics.include_apis = None
 
+    if not parameters.static_website.enabled:
+        parameters.static_website.index_document = None
+        parameters.static_website.error_document404_path = None
```
evelyn-ys, line 355: If necessary, it should be fixed from service side instead of controlling two other params' value through an switch param.

CLI should not have such customization.

adamsba3: Your teams call, but yes this should be fixed on the service side of the API response.

would think that would be an internal discussion on where you want to implement the fix

evelyn-ys: I'll contact service team later.

PR #25864, opened 2023-03-17
{Core} Fix #19183: Improve get GitHub latest version speed if no internet connection

### LINKED ISSUE #19183: docker run mcr.microsoft.com/azure-cli az takes a long time
**Describe the bug**
az takes a very long time(over a minute and a quarter when tested) when executed the first time in the environment with no connection to raw.githubusercontent.com when invoked in docker container 2.26.1

If one were to enter shell:
`docker run mcr.microsoft.com/azure-cli:2.26.1` and run az twice in a row, only the first invocation is slow. Assume contents of /root/.azure/versionsCheck.json does the trick.

**To Reproduce**
in the environment where  raw.githubusercontent.com is not accessible(no response when trying to establish connection)
`docker run mcr.microsoft.com/azure-cli:2.26.1 az`
takes a long time
**Expected behavior**
Expect to timeout within a few seconds. Also it would be useful to have a way to disable check for update

**Environment summary**
Initially reproduce on Windows 10 enterprise host with no direct access to raw.githubusercontent.com

**Workaround**
I am sure there are other options, but adding
`127.0.0.1 raw.githubusercontent.com` to /etc/hosts predictably addresses the issue. Kills ability to update as well.

bebound:
**Description**

Fix #19183
The timeout of checking the latest version from GitHub is 10 seconds. The `az -v` command is extremely slow (20s) if no internet connection.
This PR fixes this issue. It takes about 2s now.

**Testing Guide**

Enable auto upgrade, add `127.0.0.1 raw.githubusercontent.com` in hosts, `time az -v`.

**History Notes**

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/util.py
+++ b/src/azure-cli-core/azure/cli/core/util.py
@@ -314,7 +314,7 @@ def get_latest_from_github(package_path='azure-cli'):
 
 
 def _update_latest_from_github(versions):
-    if not check_connectivity(max_retries=0):
+    if not check_connectivity(url='https://raw.githubusercontent.com', max_retries=0):
         return versions, False
     success = True
     for pkg in ['azure-cli-core', 'azure-cli-telemetry']:
@@ -851,7 +851,7 @@ def find_child_collection(parent, *args, **kwargs):
     return collection
 
 
-def check_connectivity(url='https://raw.githubusercontent.com', max_retries=5, timeout=1):
+def check_connectivity(url='https://azure.microsoft.com', max_retries=5, timeout=1):
     import requests
     import timeit
     start = timeit.default_timer()
--- a/src/azure-cli-core/azure/cli/core/util.py
+++ b/src/azure-cli-core/azure/cli/core/util.py
@@ -851,7 +851,7 @@ def find_child_collection(parent, *args, **kwargs):
     return collection
 
 
-def check_connectivity(url='https://azure.microsoft.com', max_retries=5, timeout=1):
+def check_connectivity(url='https://raw.githubusercontent.com', max_retries=5, timeout=1):
     import requests
     import timeit
     start = timeit.default_timer()


src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -851,7 +851,7 @@ def find_child_collection(parent, *args, **kwargs):
     return collection
 
 
-def check_connectivity(url='https://azure.microsoft.com', max_retries=5, timeout=1):
+def check_connectivity(url='https://raw.githubusercontent.com', max_retries=5, timeout=1):
```
bebound, line 854: This function is only used by `_update_latest_from_github`, we'd better use `raw.githubusercontent.com` to check Internet connection.

yonzhan:
Core

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -314,7 +314,7 @@ def get_latest_from_github(package_path='azure-cli'):
 
 
 def _update_latest_from_github(versions):
-    if not check_connectivity(max_retries=0):
+    if not check_connectivity(url='https://raw.githubusercontent.com', max_retries=0):
```
bebound, line 317: The default URL is azure. We'd better use `raw.githubusercontent.com` to check GitHub connection status.

zhoxing-ms (approved):

src/azure-cli-core/azure/cli/core/util.py
jiasli: 1. Why does using `raw.githubusercontent.com` fails faster?
2. Also, why do we even need this check? Why don't we just check connectivity and call `get_latest_from_github(pkg)` in one shot?

bebound: 1. CLI opens `https://raw.githubusercontent.com/Azure/azure-cli/main/src/{}/setup.py` two times to get version info, the timeout of each openration is 10s.  `check_connectivity` uses `HEAD` to test connection, its timeout is 1s. If `check_connectivity` fails, it skips the url open process.
3. `check_connectivity` can be removed. But it saves time when github.com is not reachable. 1s vs 2*10s.

PR #25894, opened 2023-03-21
{Packaging} Trim all SDKs instead of biggest SDKs

bebound:
**Description**

Trim all sdk in `trim_sdk.py`. This should save 5M disk space.

Before:
e59d5903-943a-4038-b65d-c9951f51448d

After:
image

--------

Live test result does not change much after this PR.
This: Pass: 2547, Fail: 515, Pass rate: 83.18%
dev: Pass: 2510, Fail: 550, Pass rate: 82.03% 

**History Notes**

---

### DIFF
--- a/scripts/trim_sdk.py
+++ b/scripts/trim_sdk.py
@@ -70,6 +70,9 @@ def remove_unused_api_versions(resource_type):
         if resource_type in profile:
             # value is str like '2022-01-01' or SDKProfile
             value = profile[resource_type]
+            if value is None:
+                _LOGGER.info(f'{resource_type}\'s API version is None, skip')
+                return
             if isinstance(value, str):
                 used_api_versions.add(value)
             else:
@@ -108,7 +111,9 @@ def _print_folder_size(folder):
 
 
 def _get_all_sdks_to_trim():
-    resource_types = [k for k, v in AZURE_API_PROFILES['latest'].items() if k.import_prefix.startswith('azure.mgmt')]
+    # azure.mgmt.network has been removed in https://github.com/Azure/azure-cli/pull/25451
+    resource_types = [k for k, v in AZURE_API_PROFILES['latest'].items() if k.import_prefix.startswith('azure.mgmt')
+                      and k.import_prefix != 'azure.mgmt.network']
     return resource_types
 
 
@@ -155,7 +160,7 @@ def main():
     _print_folder_size(mgmt_sdk_dir)
 
     # Removed unused API versions
-    resource_types = _get_biggest_sdks_to_trim()
+    resource_types = _get_all_sdks_to_trim()
 
     for r in resource_types:
         remove_unused_api_versions(r)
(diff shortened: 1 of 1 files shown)


scripts/trim_sdk.py
```diff
@@ -70,6 +70,9 @@ def remove_unused_api_versions(resource_type):
         if resource_type in profile:
             # value is str like '2022-01-01' or SDKProfile
             value = profile[resource_type]
+            if value is None:
+                _LOGGER.info(f'{resource_type}\'s API version is None, skip')
```
bebound, line 74: `ResourceType.MGMT_NETWORK_PRIVATEDNS: None` but we still need it.

scripts/trim_sdk.py
```diff
@@ -108,7 +111,9 @@ def _print_folder_size(folder):
 
 
 def _get_all_sdks_to_trim():
-    resource_types = [k for k, v in AZURE_API_PROFILES['latest'].items() if k.import_prefix.startswith('azure.mgmt')]
+    # azure.mgmt.network has been removed in https://github.com/Azure/azure-cli/pull/25451
+    resource_types = [k for k, v in AZURE_API_PROFILES['latest'].items() if k.import_prefix.startswith('azure.mgmt')
+                      and k.import_prefix != 'azure.mgmt.network']
```
bebound, line 116: `azure.mgmt.network` is removed from dependency in #25451

yonzhan:
Packaging improvement

calvinhzy (approved):

scripts/trim_sdk.py
jiasli: Why is `ResourceType.MGMT_NETWORK_PRIVATEDNS` introduced if its value is `None`?

scripts/trim_sdk.py
jiasli: I don't think we should hard-code `azure.mgmt.network` here. What if another SDK is removed, such as `azure.mgmt.compute`? It won't be a good idea to keep modifying this `scripts/trim_sdk.py`.

scripts/trim_sdk.py
```diff
@@ -155,7 +160,7 @@ def main():
     _print_folder_size(mgmt_sdk_dir)
 
     # Removed unused API versions
-    resource_types = _get_biggest_sdks_to_trim()
+    resource_types = _get_all_sdks_to_trim()
```
jiasli, line 163: The list for top 10 biggest SDKs (L123~L148) can be removed, since we are not using them anymore.

scripts/trim_sdk.py
bebound: @necusjz to explain

necusjz: Adapt to such path conversion logic (line 665 to be exact):
https://github.com/Azure/azure-cli/blob/11c0f3ba64fc29475070a59a04536f19245f71d1/src/azure-cli-core/azure/cli/core/profiles/_shared.py#L657-L671

`azure-mgmt-privatedns` is a kind of "unversioned sdk path".

jiasli (approved):

PR #25962, opened 2023-03-28
{SQL} Revert "`az sql mi server-configuration-option`: New command group to manage server configuration options"

ostojicb:
Reverts Azure/azure-cli#25850

We've discovered a bug in the API, so it's better to delay the release of these cmdlets until the API fix gets deployed


yonzhan:
SQL

evelyn-ys (approved):

PR #25979, opened 2023-03-29
[AppService] `az functionapp delete`: Add a validation to check whether Azure Functions is not in the Azure Container app environments

kaibocai:
**Related command**
az functionapp create

**Description**
As part of the Centauri effort, we need to add the --environment parameter to the az functionapp create command so that customers can pass the managed AKS environment where the function app will be deployed.
We will not consume the --plan and --consumption-plan-location parameters since plans are not relevant for function apps deployed in managed environments. This is why server_farm_id is not set for Centauri.
We will also not consume the --os-type since it will need to be Linux.
The location is set to be the container app environment location.

**Testing Guide**

1. Create an App Service plan (az functionapp plan create -g <resource-group-name> -n <plan-name> --sku S1)

2. Create an managed AKS environment(az containerapp env create -n <managed-environment-name> -g <resource-group-name>)

3. Create a function app in the managed AKS environment (az functionapp create -g <resource-group-name> -n <function-app-name> --plan <plan-name> -s <storage-account-name> --environment <managed-environment-name>)

---


yonzhan:
AppService

zhoxing-ms (approved):

zhoxing-ms (approved):

PR #25997, opened 2023-03-30
[App Service] `az staticwebapp appsettings`: Add `--environment-name` parameter to allow app setting operation on preview environments

joslinmicrosoft:
**Related command**

`az staticwebapp appsettings`

**Description**

In the `az staticwebapp appsettings` command today, we do not provide any way for app setting operations to be made against the non-production preview environments of a static web app. This PR surfaces the `--environment-name` parameter on app setting requests to allow for this.

**Testing Guide**

Added unit tests and also ran commands like:

`az staticwebapp appsettings list -n MyAppName -g MyResourceGroup --environment-name preview`
`az staticwebapp appsettings set -n MyAppName -g MyResourceGroup --environment-name preview --setting-names key1=val1 key2=val2`
`az staticwebapp appsettings delete -n MyAppName -g MyResourceGroup --environment-name preview --setting-names key2 missingkey`

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/_params.py
@@ -1100,6 +1100,7 @@ subscription than the app service environment, please use the resource ID for --
                    help="Validation method for the custom domain.",
                    arg_type=get_enum_type(["cname-delegation", "dns-txt-token"]))
     with self.argument_context('staticwebapp appsettings') as c:
+        c.argument('environment_name', help="Name of the environment of static site")
         c.argument('setting_pairs', options_list=['--setting-names'],
                    help="Space-separated app settings in 'key=value' format. ",
                    nargs='*')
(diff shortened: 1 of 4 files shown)


yonzhan:
App Service

StrawnSC (approved):
LGTM!

src/azure-cli/azure/cli/command_modules/appservice/_params.py
mkarmark, line 1: You probably need a required=False on this param right, because if it isn't included we will assume primary build?

zhoxing-ms: In fact, if the default value of the `environment_name` parameter in the custom method is `None`, then the default setting is `required=False` and there is no need to explicitly add it repeatedly

zhoxing-ms (approved):

PR #26134, opened 2023-04-13
[Core] `aaz`: Support argument prompt input

kairu-ms:
**Related command**

**Description**

Support argument prompt input for AAZCommand args when the value is blank.

Below is the current usage for a command without prompt input for secure arguments like `--password`
image

Now we can support prompt input for those secure arguments like this way
image
The password is not appeared in TTY.

For those arguments which supports prompt input, you can see the following message added in their short summary:
image
Which means if the value of those arguments is blank, the prompt input will be triggered in TTY. 

The there are two prompt input classes: 
- AAZPromptInput: Used for regular arguments. The input will be display in TTY.
- AAZPromptPasswordInput: Used for password kind of arguments. The input will be hidden. And support input confirmation.

image
The first prompt input is for `--name` which use `AAZPromptInput`. The last two prompt input is for `--password` which use `AAZPromptPasswordInput` with confirmation.

**Testing Guide**

**History Notes**

[Core] `aaz`: Support `AAZPasswordArg`
[Core] `aaz`: Support argument prompt input for simple type args

---


yonzhan:
Core

necusjz (approved):

PR #26219, opened 2023-04-21
[Core] `extension`: Install extensions without prompt as default behavior when it's not in a tty

kairu-ms:
**Related command**

**Description**

When we run a command in an extension which is not installed, it will show a prompt like this
image

This will block the script in no-tty environment. This PR use the `yes_without_prompt` mode as default for no-tty environment, so that the missing commands will be installed automatically.

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/extension/dynamic_install.py
+++ b/src/azure-cli-core/azure/cli/core/extension/dynamic_install.py
@@ -121,7 +121,14 @@ def try_install_extension(parser, args):
 
 
 def _get_extension_use_dynamic_install_config(cli_ctx):
-    default_value = 'yes_prompt'
+    from knack.prompting import verify_is_a_tty, NoTTYException
+    try:
+        verify_is_a_tty()
+        # prompt when it's in tty
+        default_value = 'yes_prompt'
+    except NoTTYException:
+        # without prompt when it's not in tty
+        default_value = 'yes_without_prompt'
     use_dynamic_install = cli_ctx.config.get(
         'extension', 'use_dynamic_install', default_value).lower() if cli_ctx else default_value
     if use_dynamic_install not in ['no', 'yes_prompt', 'yes_without_prompt']:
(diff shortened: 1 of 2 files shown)


yonzhan:
Core

jsntcy:
- Add why we need this change in PR description
- Fix CI

src/azure-cli-core/azure/cli/core/extension/dynamic_install.py
```diff
@@ -121,7 +121,14 @@ def try_install_extension(parser, args):
 
 
 def _get_extension_use_dynamic_install_config(cli_ctx):
-    default_value = 'yes_prompt'
+    from knack.prompting import verify_is_a_tty, NoTTYException
```
jsntcy, line 124: Have you tested the change in both tty and non-tty environment?

kairu-ms: Add function tests

src/azure-cli-core/azure/cli/core/tests/test_extension.py
```diff
@@ -230,6 +230,16 @@ def test_ext_compat_with_cli_require_ext_min_version(self):
             self.assertFalse(is_compatible)
             self.assertEqual(min_ext_required, expected_min_ext_required)
 
+    @mock.patch('sys.stdin.isatty', return_value=True)
+    def test_ext_dynamic_install_config_tty(self, _):
+        from azure.cli.core.extension.dynamic_install import _get_extension_use_dynamic_install_config
+        self.assertEqual(_get_extension_use_dynamic_install_config(None), 'yes_prompt')
+
+    @mock.patch('sys.stdin.isatty', return_value=False)
+    def test_ext_dynamic_install_config_tty(self, _):
```
jsntcy, line 239: The name of the test should be `test_ext_dynamic_install_config_no_tty`? otherwise two tests are with the same name `test_ext_dynamic_install_config_tty`

src/azure-cli-core/azure/cli/core/extension/dynamic_install.py
```diff
@@ -121,7 +121,14 @@ def try_install_extension(parser, args):
...
+    try:
+        verify_is_a_tty()
```
jiasli, line 126: Not that `sys.stdin.isatty()` may not be accurate. For example, if a command in run in terminal but redirected, `sys.stdin.isatty()` returns `False`:

```
$ python -c "import sys; print(sys.stdin.isatty())"
True

$ echo "test" | python -c "import sys; print(sys.stdin.isatty())"
False
```

PR #25689, opened 2023-03-06
{Core} Add random_config_dir param in DummyCli

bebound:
**Description**

ARM64 machines are multi-core, and some errors occur because of concurrency, as all tests use same config dir.

I add a `random_config_dir` param in `DummyCli`. If it is True, the config dir will be `~/.azure/dummy_cli_config_dir/xxxxxxx`.

If random config is enabled by default, the config file is created for each test, and the performance decrease a lot.
`Azure.azure-cli Full Test` spends 24 min instead of 16min.

I enable `random_config_dir` for the tests which use `az config` or `az configure`.
Example usage:
``` python
class ConfigTest(ScenarioTest):

    def __init__(self, *arg, **kwargs):
        super().__init__(*arg, random_config_dir=True, **kwargs)
```

```
2023-03-03T11:05:05.5590254Z [gw2] [ 65%] FAILED tests/test_cloud.py::TestCloud::test_remove_known_cloud 
2023-03-03T11:05:05.5590455Z 
2023-03-03T11:05:05.5590884Z =================================== FAILURES ===================================
2023-03-03T11:05:05.5591177Z ______________________ TestCloud.test_remove_known_cloud _______________________
2023-03-03T11:05:05.5592042Z [gw2] linux -- Python 3.10.10 /opt/az/bin/python3
2023-03-03T11:05:05.5592295Z self = <core.tests.test_cloud.TestCloud testMethod=test_remove_known_cloud>
2023-03-03T11:05:05.5592422Z 
2023-03-03T11:05:05.5592577Z     def test_remove_known_cloud(self):
2023-03-03T11:05:05.5592756Z >       cli = DummyCli()
2023-03-03T11:05:05.5592824Z 
2023-03-03T11:05:05.5593120Z /opt/az/lib/python3.10/site-packages/azure/cli/core/tests/test_cloud.py:216: 
2023-03-03T11:05:05.5593362Z _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
2023-03-03T11:05:05.5593706Z /opt/az/lib/python3.10/site-packages/azure/cli/core/mock.py:25: in __init__
2023-03-03T11:05:05.5593932Z     super(DummyCli, self).__init__(
2023-03-03T11:05:05.5594263Z /opt/az/lib/python3.10/site-packages/azure/cli/core/__init__.py:59: in __init__
[...]

PR #26273, opened 2023-04-26
[Container] `az container create`: Add new parameters for container security context for confidential ContainerGroupSku

fnuarnav:
**Related command**

`az container create `

**Description**

This PR:
upgrades azure-mgmt-containerinstance to 10.1.0
Adds Container security context properties for confidential ContainerGroupSku:
- privileged: run the container as a privileged container
- allow-privilege-escalation: allow containers to be run with elevated privileges
- run-as: run container process as a user
- run-as-group: run container process as a user group
- seccomp-profile: base64 encoded seccomp profile
- add-capabilities: list of capabilities to be added
- drop-capabilities: list of capabilities to be dropped

**Testing Guide**

**History Notes**

---


yonzhan:
Container

fnuarnav:
I tried recording the test test_tag_update_by patch under command_modules/resource by running `azdev test test_tag_update_by_patch`, but the test was failing with error "CLIError: Please run az login to setup account",  even after running `az login`, and also when I tried removing the recording so that the test is run during CI tests.

image

I looked at the context and found that a dummy context is being passed to the method, which has SubscriptionId=None.

zhoxing-ms:
@fnuarnav I cannot reproduce your issue locally. Can you try using `az login`  to login first, and then use `az account show` to check the subscription being used and then execute `azdev test test_tag_update_by_patch --live` to re-record the yaml file?

src/azure-cli/azure/cli/command_modules/container/_params.py
```diff
@@ -79,6 +79,13 @@ def load_arguments(self, _):
 
     with self.argument_context('container create', arg_group='Confidential Container Group') as c:
         c.argument('cce_policy', help="The CCE policy for the confidential container group")
+        c.argument('allow_privilege_escalation', options_list=['--allow-escalation'], help="Allows whether a process can gain more privileges than its parent process.", action='store_true')
+        c.argument('privileged', help='The flag to determine if the contianer permissions is elevated to Privileged', action='store_true')
+        c.argument('run_as_user', help="Sets the User GID for the container")
+        c.argument('run_as_group', help="Sets the User UID for the container")
```
zhoxing-ms, line 85: Please use the first-person voice in the help message
```suggestion
        c.argument('allow_privilege_escalation', options_list=['--allow-escalation'], help="Allow whether a process can gain more privileges than its parent process.", action='store_true')
        c.argument('privileged', help='The flag to determine if the contianer permissions is elevated to Privileged', action='store_true')
        c.argument('run_as_user', help="Set the User GID for the container")
        c.argument('run_as_group', help="Set the User UID for the container")
```

src/azure-cli/azure/cli/command_modules/container/_params.py
```diff
@@ -79,6 +79,13 @@ def load_arguments(self, _):
...
+        c.argument('seccomp_profile', help="A base64 encoded string containing the contents of the JSON in the seccomp profile")
+        c.argument('add_capabilities', type=str, nargs='+', help="A List of security context capabilities to be added")
+        c.argument('drop_capabilities', type=str, nargs='+', help="A List of security context capabilities to be dropped")
```
zhoxing-ms, line 88: `type=str` is the default setting, please do not explicitly repeat the declaration
```suggestion
        c.argument('add_capabilities', nargs='+', help="A List of security context capabilities to be added")
        c.argument('drop_capabilities', nargs='+', help="A List of security context capabilities to be dropped")
```

fnuarnav: > @fnuarnav I cannot reproduce your issue locally. Could you try using `az login` to login first, and then use `az account show` to check the subscription being used and then execute `azdev test test_tag_update_by_patch --live` to re-record the yaml file?

I have tried logging in and running the test but still gives the same error.

zhoxing-ms: @fnuarnav May I ask if the subscription output after executing `az login` meets your expectations?

fnuarnav: Yes, after `az login` the subscription output is what I would expect. I was able to run the tests under the container command_module with this subscription.

zhoxing-ms:
@fnuarnav Could you please fix these CI issues?

zhoxing-ms:
@fnuarnav Please note that we are launching the release for this sprint this week. Please resolve all comments by tomorrow, otherwise the release of this PR will have to be postponed to the next sprint (on **07-04**)

zhoxing-ms:
> I tried recording the test test_tag_update_by patch under command_modules/resource by running azdev test test_tag_update_by_patch, but the test was failing with error "CLIError: Please run az login to setup account", even after running az login, and also when I tried removing the recording so that the test is run during CI tests.

I found that the issue with recording tests was caused by PR #25689
```python
class TagScenarioTest(ScenarioTest):

    def __init__(self, *arg, **kwargs):
        super().__init__(*arg, random_config_dir=True, **kwargs)
```
The random `config_dir` can result in the inability to find the corresponding subscription configuration when recording tests in live mode.

zhoxing-ms (approved):

yanzhudd (approved):

zhoxing-ms:
> how did you work around the random_config_dir=True issue, by temporarily removing it?

@jiasli Yes, since it is not related to the business logic we are testing, so I removed it temporarily to quickly unblock our recording

PR #26287, opened 2023-04-28
{Core} `aaz`: Fix importlib deadlocks while loading `requests.models` in multi threads

kairu-ms:
closed: #26272
closed: #26297

**Related command**

**Description**

In python 3.10.10, we encountered import deadlocks in `aaz` LRO which try to load the `requests.models` in two different threads. It's a known issue https://github.com/psf/requests/issues/2925

In this PR, we provide a workaround by importing these modules in the main thread before spawning child threads.

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/aaz/_poller.py
+++ b/src/azure-cli-core/azure/cli/core/aaz/_poller.py
@@ -14,8 +14,7 @@ from azure.core.tracing.common import with_current_context
 from azure.core.tracing.decorator import distributed_trace
 # import LROPoller in main thread to resolve import deadlock between threads in python 3.10.10
 # reference https://github.com/psf/requests/issues/2925 and https://github.com/Azure/azure-cli/issues/26272
-from msrest.polling.poller import LROPoller  # pylint: disable=unused-import
-
+import requests
 
 _LOGGER = logging.getLogger(__name__)
 
--- a/src/azure-cli-core/azure/cli/core/aaz/_poller.py
+++ b/src/azure-cli-core/azure/cli/core/aaz/_poller.py
@@ -12,7 +12,7 @@ from azure.core.polling import NoPolling
 from azure.core.polling.base_polling import LROBasePolling
 from azure.core.tracing.common import with_current_context
 from azure.core.tracing.decorator import distributed_trace
-# import LROPoller in main thread to resolve import deadlock between threads in python 3.10.10
+# import requests in main thread to resolve import deadlock between threads in python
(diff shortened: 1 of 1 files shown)


yonzhan:
Core

src/azure-cli-core/azure/cli/core/aaz/_poller.py
```diff
@@ -12,6 +12,10 @@
 from azure.core.polling.base_polling import LROBasePolling
 from azure.core.tracing.common import with_current_context
 from azure.core.tracing.decorator import distributed_trace
+# import LROPoller in main thread to resolve import deadlock between threads in python 3.10.10
+# reference https://github.com/psf/requests/issues/2925 and https://github.com/Azure/azure-cli/issues/26272
+from msrest.polling.poller import LROPoller  # pylint: disable=unused-import
```
jiasli, line 17: My gut feeling tells me this may not be safe enough, as there may be other dependencies imported by the sub-thread and main thread at the same time.

The SDK implementation is safer because the first `PUT` operation is performed on the main thread, so subsequent polling `GET` operations on the sub-thread will have the full dependency tree cached already.

Besides `LROPoller`, there are also `AzureOperationPoller` and `AzureCoreLROPoller` imported from `msrestazure.azure_operation` and `azure.core.polling` respectively:

https://github.com/Azure/azure-cli/blob/d9accde22cc493d7b4454baf2be8e3707c920920/src/azure-cli-core/azure/cli/core/commands/__init__.py#L1153-L1157

https://github.com/Azure/azure-cli/blob/dae19d0dfcad768f1dc19458f714334bacdf5a4f/src/azure-cli-core/azure/cli/core/util.py#L685-L690

Though I haven't got time to look into their actual usage.

kairu-ms: This line is enough to mitigate the issue. Moving the first request in the main thread will involve a huge change.
We cannot resolve them all because it's python issue https://github.com/python/cpython/issues/91351.

AllyW: Given https://github.com/Azure/azure-cli/issues/26297, do not use command index caching can walk around it. @jiasli

kairu-ms: When loading other modules, some of them will import `requests`. That's why it can mitigate the issue.

necusjz (approved):

PR #26344, opened 2023-05-05
{CI} Optimize pipeline result

wangzelin007:
**Related command**

**Description**

Now we will directly display the test results of all modules, the comment is too long.
So I add a new level `TestName` to optimize the display effect of azure-cli full test: 
image
TODO: 
Support only_show_errors.

**Testing Guide**

**History Notes**

---

### DIFF
--- a/scripts/ci/automation_full_test.py
+++ b/scripts/ci/automation_full_test.py
@@ -381,11 +381,16 @@ def build_pipeline_result():
             "Name": job_name,
             "Details": [
                 {
-                    "Profile": profile,
+                    "TestName": "AzureCLI-FullTest",
                     "Details": [
                         {
-                            "PythonVersion": python_version,
-                            "Details": []
+                            "Profile": profile,
+                            "Details": [
+                                {
+                                    "PythonVersion": python_version,
+                                    "Details": []
+                                }
+                            ]
                         }
                     ]
                 }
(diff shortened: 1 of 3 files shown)


yonzhan:
CI

zhoxing-ms (approved):

yanzhudd (approved):

scripts/ci/automation_full_test.py
```diff
@@ -441,13 +446,13 @@ def get_pipeline_result(test_result_fp, pipeline_result):
...
                     break
-                for i in pipeline_result[unique_job_name]['Details'][0]['Details'][0]['Details']:
+                for i in pipeline_result[unique_job_name]['Details'][0]['Details'][0]['Details'][0]['Details']:
                     if i['Module'] == module:
                         i['Status'] = 'Failed'
                         i['Content'] = build_markdown_content(state, test_case, message, line, i['Content'])
                         break
             else:
-                for i in pipeline_result[unique_job_name]['Details'][0]['Details'][0]['Details']:
+                for i in pipeline_result[unique_job_name]['Details'][0]['Details'][0]['Details'][0]['Details']:
```
jsntcy, line 455: What's the purpose to add `[0]['Details']`?

wangzelin007: image
This is due to the data structure we worked out with the azure-powershell team.

bebound (approved):

scripts/ci/automation_full_test.py
jsntcy: So CLI only has one element for each `Details` property, while there will be more than one element for `Details` property in azure PS?

wangzelin007: Yes, there are multiple elements for Details property in azure PS.
Please note that we also have multiple elements in the innermost layer. 
The reason why the other layers only have a single element is that we use multiple instance tests. 
If these tests are run on the same machine, then our details will also have multiple elements.

jsntcy (approved):

PR #26351, opened 2023-05-05
[App Service] `az functionapp create`: Update the default image to `mcr.microsoft.com` for Centauri

kamperiadis:
**Related command**
`az functionapp create`

**Description**
For Centauri, if the customer does not provide an image, we provide a default value. This default value needs to be updated.

**Testing Guide**
Create a Centauri function app without an image value and then check that the created function app uses the new default image value.

---


kaibocai (approved):
LGTM!

zhoxing-ms (approved):

src/azure-cli/azure/cli/command_modules/appservice/_constants.py
```diff
@@ -95,4 +95,4 @@ def __init__(self):
     'powershell': 'FunctionApp/windows-powershell-functionapp-on-azure.yml',
 }
 
-DEFAULT_CENTAURI_IMAGE = 'azurefunctionstest.azurecr.io/azure-functions/dotnet7-quickstart-demo:1.0'
+DEFAULT_CENTAURI_IMAGE = 'mcr.microsoft.com/azure-functions/dotnet7-quickstart-demo:1.0'
```
zhoxing-ms, line 98: I'd like to confirm that the position of the default image has changed, but the image itself has not changed, right? It won't cause a breaking change for customers, will it?

kamperiadis: Yes, the image itself has not changed. This default image has only been used by private preview customers for Centauri so even if it is a breaking change, we can communicate that they update the image if they were using the previous default image.

zhoxing-ms:
@kamperiadis Could you please resolve this conflict?

src/azure-cli/azure/cli/command_modules/appservice/_constants.py
zhoxing-ms: Got it, thanks

zhoxing-ms (approved):

PR #26354, opened 2023-05-05
[AppService] `az functionapp list`: Fix when resource group is not passed

kamperiadis:
**Related command**
`az functionapp list`

**Description**
Currently, when the `az functionapp list` command is run without a resource group, there are some pagination API calls made that fail with a 502 error. In this PR, we have a workaround for the failed pagination API calls.

This command is needed for Centauri when running the `az functionapp devops-pipeline create` command.

**Testing Guide**
Run `az functionapp list` without any resource group information. 

---


yonzhan:
AppService

zhoxing-ms:
@kaibocai Could you please help review the `functionapp` related PR?

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -997,11 +997,19 @@ def show_app(cmd, resource_group_name, name, slot=None):
...
     if resource_group_name:
         result = list(client.web_apps.list_by_resource_group(resource_group_name))
     else:
-        result = list(client.web_apps.list())
+        resource_groups = rg_client.resource_groups.list()
+        result = []
+        for resource_group in resource_groups:
+            function_apps = list(client.web_apps.list_by_resource_group(resource_group.name))
+            for function_app in function_apps:
+                result.append(function_app)
```
zhoxing-ms, line 1012: Repeated requests for list web_apps based on resource groups may result in poor command performance.

zhoxing-ms: > Currently, when the az functionapp list command is run without a resource group, there are some pagination API calls made that fail with a 502 error.

@kaibocai Could we solve the problem with `client.web_apps.list()` on the server side?

kamperiadis: Thank you for reviewing the PR @zhoxing-ms. We'll get the problem resolved in the server side instead.

Issue #23947, opened 2022-09-20
az upgrade under Windows waits for end of msiexec which wants to update python.exe


lgiuliani80:
**Related command**
az upgrade

**Describe the bug**
az upgrade under Windows downloads and runs the MSI installer. The call to msiexec.exe is synchronous (i.e. it waits for its termination), but during the installation process the MSI wants to update python.exe, which is in use.

**To Reproduce**
az upgrade [under Windows]

**Expected behavior**
Under Windows EXIT PROCESS after calling msiexec: the upgrade will involve UI, so there is no need to keep the console blocked. This way this bug is prevented.

yonzhan:
@jiasli for awareness

miqm:
👍🏻 same here.
I assume it's because Azure CLI is waiting for installer to finish - so the python.exe is still running. So either az cli installer should not require python to exit or the cli after starting installer should exit and hope that user will complete installation.

stijnherreman:
The previous time I tried upgrading this recent install, I tried letting the installer shut down the Python process but it failed to do so. So I continued with the reboot option.
This time I tried to stop the process with Ctrl-C, but the process was unresponsive. Then I killed the Python process through Task Manager, and only then "Terminate batch job (Y/N)?" appeared.

I'd expect anyone on Windows running `az upgrade` is running into this issue, including the people who developed/tested this feature. Does this only occur under a specific environment? I'm using PowerShell Core in Windows Terminal.

---

> cli after starting installer should exit and hope that user will complete installation

@miqm Does it matter to the CLI whether or not the user completes the installation?

KLuuKer:
**bump** still a thing...

PR #26464, opened 2023-05-17
[Packaging] Fix #22741: `az upgrade`: This command becomes non-blocking on Windows

### LINKED ISSUE #22741: Improve the az upgrade locking

**Related command**
az upgrade

**Describe the bug**
Terminal spawns the MSI so the MSI locks Python. If you close the terminal window, the installation will stop.
image

**To Reproduce**
Upgrade within a terminal

**Expected behavior**
Install with no locks.

**Environment summary**
Windows-10-10.0.19043-SP0, 
Python 3.10.4
MSI
Windows 10, Terminal, PowerShell

**Additional context**
none

bebound:
**Description**

Close https://github.com/Azure/azure-cli/issues/22741, close https://github.com/Azure/azure-cli/issues/23947

This PR changes the behavior of `az upgrade` on Windows.
If the Python version changes in new version, the upgrade is blocked.
(This issue can be reproduced by installing 2.44 and running `az upgrade`)
image

In this PR, `az upgrade` exits after MSI installation dialog is opened. So the related file is not opened during upgrade.

Three breaking changes:
1. `az upgrade` becomes non-blocking command, it exit after MSI installation dialog is opened.
2. az does not know if the upgrade is successful, the exit code is always 0 once installation starts.
3. This command does not upgrade extensions if new version is available. Users need to run `az upgrade` again to upgrade extensions.

**Testing Guide**

set this in `upgrade_version`
```python
local_version = '2.50.0'
installer = 'MSI'
```

---


yonzhan:
Packaging

src/azure-cli/azure/cli/command_modules/util/custom.py
```diff
@@ -187,10 +188,12 @@ def _upgrade_on_windows():
...
     logger.warning("Updating Azure CLI with MSI from https://aka.ms/installazurecliwindows")
     tmp_dir, msi_path = _download_from_url('https://aka.ms/installazurecliwindows')
-
     logger.warning("Installing MSI")
     import subprocess
-    exit_code = subprocess.call(['msiexec.exe', '/i', msi_path])
+    subprocess.Popen(['msiexec.exe', '/i', msi_path])
+    logger.warning("Installation started, please wait for a few minutes.")
+    import sys
+    sys.exit(0)
```
jiasli, line 196: As stated in the PR description, this makes us be unable to tell if the upgrade is successful. Better to confirm with PM whether this is acceptable solution. Otherwise, we can develop some "upgrader.exe" written in C.

bebound: We can create a upgrader to update az to prevent running az, but we need to call `az` to map `az upgrade` command to that upgrader.

`az upgrade` is very complicated. It downloads MSI, waits for installation to finish and updates extensions in one command.
I can hardly come up with an idea to fix this issue without breaking change.

jiasli: I thought about it again and feel it is indeed not possible. A GUI application can have such `upgrader.exe` to upgrade the product itself, but this can't be done for a console application.

Consider this invocation chain: `python.exe`(1) -> `upgrader.exe` -> `python.exe`(2)

- To keep the second `python.exe` connected to the original terminal context, every executable on the invocation chain must be alive and wait for its downstream executable to exit. In order words, if `python.exe`(1) subprocesses `upgrader.exe` and exits, then after `upgrader.exe` upgrades `python.exe`, subprocesses `python.exe`(2) and exits, it is no longer possible to connect `python.exe`(2) back to the original terminal context - the output is out of order.
- If the first `python.exe` is alive, `upgrader.exe` cannot update it.

-> A contradiction!

Consider the following code:

```py
import subprocess

p = subprocess.Popen(['python.exe', '-c', 'import time; time.sleep(3); print("awaken")'])
print('Exiting')
```

Output:

```
PS D:\cli\testproj> python.exe main.py
Exiting
PS D:\cli\testproj> awaken
|
```

The first `python.exe` will exits and gives control back to the terminal. The second `python.exe` can print, yes, but it is out of order.

How to solve it? Wait for the subprocess to exit:

```py
import subprocess

p = subprocess.Popen(['python.exe', '-c', 'import time; time.sleep(3); print("awaken")'])
p.wait()
print('Exiting')
```

Output:

```
PS D:\cli\testproj> python.exe main.py
awaken
Exiting
PS D:\cli\testproj> |
```

src/azure-cli/azure/cli/command_modules/util/custom.py
```diff
@@ -187,10 +188,12 @@ def _upgrade_on_windows():
...
 
     if exit_code:
```
jiasli, line 198: These below lines can never be reached, right?

src/azure-cli/azure/cli/command_modules/util/custom.py
bebound: Yes, exactly.

----- 
I come up an idea to break the python dependency of `python.exe(1) -> upgrader.exe`.
We can modify `az.cmd`, if the param is `upgrade`, call upgrader.exe, otherwise, run az.

Furthermore, we can write the upgrade logic in bat, it looks feasible.

src/azure-cli/azure/cli/command_modules/util/custom.py
bebound: Yes, they are not reachable. This is draft PR for test only currently.

src/azure-cli/azure/cli/command_modules/util/custom.py
```diff
@@ -187,10 +188,12 @@ def _upgrade_on_windows():
     """
     logger.warning("Updating Azure CLI with MSI from https://aka.ms/installazurecliwindows")
     tmp_dir, msi_path = _download_from_url('https://aka.ms/installazurecliwindows')
-
     logger.warning("Installing MSI")
     import subprocess
-    exit_code = subprocess.call(['msiexec.exe', '/i', msi_path])
+    subprocess.Popen(['msiexec.exe', '/i', msi_path])
+    logger.warning("Installation started, please wait for a few minutes.")
```
jiasli, line 194:
```suggestion
    logger.warning("Installation started. Please complete upgrade in the opened window.")
```

src/azure-cli/azure/cli/command_modules/util/custom.py
```diff
@@ -188,25 +188,21 @@ def _upgrade_on_windows():
...
     logger.warning("Installing MSI")
-    import subprocess
-    exit_code = subprocess.call(['msiexec.exe', '/i', msi_path])
-
-    if exit_code:
-        logger.warning("Installation Failed. You may manually install %s", msi_path)
-    else:
-        from azure.cli.core.util import rmtree_with_retry
-        logger.warning("Succeeded. Deleting %s", tmp_dir)
-        rmtree_with_retry(tmp_dir)
```
jiasli, line 208: I suggest we find a way to preserve this logic, otherwise the temporary MSI will not be deleted.

bebound: The MSI is in the `tempfile.mkdtemp()` directory (`C:\\Users\\xx\\AppData\\Local\\Temp\\tmp1s_cgvau`), it appears that windows can delete it automatically.
[Manage drive space with Storage Sense](https://support.microsoft.com/en-us/windows/manage-drive-space-with-storage-sense-654f6ada-7bfc-45e5-966b-e24aded96ad5)

src/azure-cli/azure/cli/command_modules/util/custom.py
```diff
@@ -188,28 +188,33 @@ def _upgrade_on_windows():
...
     else:
         msi_url = 'https://aka.ms/installazurecliwindowsx64'
     logger.warning("Updating Azure CLI with MSI from %s", msi_url)
-    tmp_dir, msi_path = _download_from_url(msi_url)
 
-    logger.warning("Installing MSI")
-    import subprocess
-    exit_code = subprocess.call(['msiexec.exe', '/i', msi_path])
+    # Save MSI to ~\AppData\Local\Temp\azure-cli-msi, clean up the folder first
+    msi_dir = Path(tempfile.gettempdir()) / 'azure-cli-msi'
```
jiasli, line 205: Azure CLI doesn't really use `pathlib`. Consider using `os.path` to be consistent with the rest of the code.

src/azure-cli/azure/cli/command_modules/util/custom.py
```diff
@@ -188,28 +188,33 @@ def _upgrade_on_windows():
...
+    rmtree_with_retry(msi_dir)
+    msi_dir.mkdir()
```
jiasli, line 207: Will this line raise exception if `rmtree_with_retry` can't delete the directory successfully?

bebound: `rmtree_with_retry` ignores the `FileNotFoundError` by default, which is my desired behavior.

src/azure-cli/azure/cli/command_modules/util/custom.py
bebound: `pathlib` is the recommended way to deal with path operations now.
> If you’ve never used this module before or just aren’t sure which class is right for your task, [Path](https://docs.python.org/3/library/pathlib.html#pathlib.Path) is most likely what you need.
> For low-level path manipulation on strings, you can also use the [os.path](https://docs.python.org/3/library/os.path.html#module-os.path) module.    --- https://docs.python.org/3/library/pathlib.html

src/azure-cli/azure/cli/command_modules/util/custom.py
jiasli: `rmtree_with_retry` also ingores cannot-delete-folder error, and that's why the directory can still exist and cause error to `mkdir`.

jiasli (approved):

PR #26492, opened 2023-05-19
{CI} azdev style check ignore aaz folder

wangzelin007:
**Related command**

**Description**

`azdev style` ignore aaz folder, reduce the check time from 75min to 35min.
It will still scan the core/aaz folder, as shown in the screenshot:
image
image

**Testing Guide**

**History Notes**

---


yonzhan:
CI

pylintrc
```diff
@@ -1,5 +1,5 @@
 [MASTER]
-ignore=tests,generated,vendored_sdks,privates
+ignore=tests,generated,vendored_sdks,privates,aaz
```
jiasli, line 2: Will this also make `src/azure-cli-core/azure/cli/core/aaz` unchecked?

wangzelin007: I'm just testing first to see if I can improve the test speed.
These parameters are more complicated than we think.
The ignore parameter represents `Files or directories to be skipped. They should be base names, not paths.`
The ignore-patterns parameter represents `Files or directories matching the regular expression patterns are skipped.The regex matches against base names, not paths.`
The ignore-paths parameter represents `Files or directories matching the regular expressions patterns to the ignore-list.`
But I use the pylint 2.11.1 version test and it doesn't work.

jiasli: I think it may be a good idea to rename `azure/cli/command_modules/network/aaz` to `aaz_generated`. It will may linter's config much easier.

wangzelin007: I very much agree with your idea, we should rename all the directories generated by CodeGen to aaz_generated, not only main modules, but also extensions,
But I'm not too sure how much this affects CodeGen.
We can discuss this with @kairu-ms and I'll keep trying to see if there is a way to exclude directories.

pylintrc
```diff
@@ -1,6 +1,7 @@
 [MASTER]
 ignore=tests,generated,vendored_sdks,privates
 ignore-patterns=test.*,azure_devops_build.*
+ignore-paths=.*/command_modules/.*/aaz
```
wangzelin007, line 4: - The `ignore` parameter represents Files or directories to be skipped. They should be base names, not paths.
- The `ignore-patterns` parameter represents Files or directories matching the regular expression patterns are skipped. The regex matches against base names, not paths.
- The `ignore-paths` parameter represents Files or directories matching the regular expressions patterns to the ignore-list.
So we should use `ignore-paths` here.

jsntcy:
Do we also want to apply this change to azure-cli-extensions repo?

wangzelin007:
> Do we also want to apply this change to azure-cli-extensions repo?

azure-cli-extensions repo do not have azdev style check, and we can simple add `azz` to the config since azure-cli-extensoins don't have `/core/aaz` folder either.

kairu-ms (approved):

jsntcy (approved):

necusjz (approved):

evelyn-ys (approved):

PR #23494, opened 2022-08-10
{Packaging} Deprecation notice urllib3[secure]

fyunusa:
### Description

pyOpenSSL and urllib3[secure] are deprecated in the upcoming release (1.26.12)
https://github.com/urllib3/urllib3/issues/2680

PR #26538, opened 2023-05-26
{App Service} Remove usages of deprecated `inject_into_urllib3`

jiasli:
**Description**
A supplementary PR for #23494

According to https://urllib3.readthedocs.io/en/stable/reference/contrib/pyopenssl.html

> This module was relevant before the standard library ssl module supported SNI, but now that we’ve dropped support for Python 2.7 all relevant Python versions support SNI so this module is no longer recommended.

Since we have dropped support for Python 2 (https://github.com/Azure/azure-cli/pull/11363), we should remove usages of deprecated `inject_into_urllib3`. Otherwise, in the future, `urllib3` v2.1.0 will raise an exception when `inject_into_urllib3` is called (https://github.com/urllib3/urllib3/issues/2680).

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -2794,13 +2794,6 @@ def get_scm_site_headers(cli_ctx, name, resource_group_name, slot=None, addition
 
 
 def _get_log(url, headers, log_file=None):
-    import urllib3
-    try:
-        import urllib3.contrib.pyopenssl
-        urllib3.contrib.pyopenssl.inject_into_urllib3()
-    except ImportError:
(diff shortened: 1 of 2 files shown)


yonzhan:
App Service

bebound (approved):

zhoxing-ms (approved):

zhoxing-ms:
@StrawnSC Could you please help review this PR? It would be better if you are willing to do some related tests

jsntcy (approved):

StrawnSC (approved):

StrawnSC:
LGTM -- pulled the code locally and ran the following:
- `az webapp up` with --logs 
- `az webapp log download/tail`
- `az webapp ssh`

PR #26539, opened 2023-05-26
{Keyvault} Remove useless AAD endpoint and honor user input `--subscription` to support cross sub operations

evelyn-ys:
**Description**

In #22432, we migrated AAD Graph to MS Graph. Some related code in Keyvault module was also modified from https://github.com/Azure/azure-cli/blob/1451bd570fd0fe54771cedaecdbd49cdb4b8d69c/src/azure-cli/azure/cli/command_modules/keyvault/custom.py#L681-L689 to https://github.com/Azure/azure-cli/blob/cac3f06cf483d8c81f93c9cd556116e27aff2182/src/azure-cli/azure/cli/command_modules/keyvault/custom.py#L679-L684

Before migration, we need to call `profile.get_login_credentials` to get `cred` and `tenant_id` info.
After migration, we don't need the `cred` anymore but didn't drop `profile.get_login_credentials` just to get `tenant_id` info.

But there's better way to get `tenant_id`, it's a waste to call `profile.get_login_credentials` just for tenant info. Besides, the existing `profile.get_login_credentials` is still passing `resource` with AAD Graph endpoint which is deprecated. This PR aims to clear these useless code.

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/keyvault/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/keyvault/custom.py
@@ -408,7 +408,7 @@ def recover_hsm(cmd, client, hsm_name, resource_group_name, location, no_wait=Fa
 
     # tenantId and sku shouldn't be required
     profile = Profile(cli_ctx=cmd.cli_ctx)
-    tenant_id = profile.get_subscription(subscription=cmd.cli_ctx.data['subscription_id'])[_TENANT_ID]
+    tenant_id = profile.get_subscription(subscription=cmd.cli_ctx.data.get('subscription_id', None))[_TENANT_ID]
 
     # Use 'Recover' as 'create_mode' temporarily since it's a bug from service side making 'create_mode' case-sensitive
     # Will change it back to CreateMode.recover.value('recover') from SDK definition after service fix
@@ -435,7 +435,7 @@ def recover_vault(cmd, client, vault_name, resource_group_name, location, no_wai
     Sku = cmd.get_models('Sku', resource_type=ResourceType.MGMT_KEYVAULT)
     SkuName = cmd.get_models('SkuName', resource_type=ResourceType.MGMT_KEYVAULT)
     profile = Profile(cli_ctx=cmd.cli_ctx)
-    tenant_id = profile.get_subscription(subscription=cmd.cli_ctx.data['subscription_id'])[_TENANT_ID]
+    tenant_id = profile.get_subscription(subscription=cmd.cli_ctx.data.get('subscription_id', None))[_TENANT_ID]
 
     params = VaultCreateOrUpdateParameters(location=location,
                                            properties={'tenant_id': tenant_id,
@@ -604,7 +604,7 @@ def create_hsm(cmd, client,
     ManagedHsmSku = cmd.get_models('ManagedHsmSku', resource_type=ResourceType.MGMT_KEYVAULT)
 
     profile = Profile(cli_ctx=cmd.cli_ctx)
-    tenant_id = profile.get_subscription(subscription=cmd.cli_ctx.data['subscription_id'])[_TENANT_ID]
+    tenant_id = profile.get_subscription(subscription=cmd.cli_ctx.data.get('subscription_id', None))[_TENANT_ID]
 
     properties = ManagedHsmProperties(tenant_id=tenant_id,
                                       enable_purge_protection=enable_purge_protection,
@@ -669,7 +669,7 @@ def create_vault(cmd, client,  # pylint: disable=too-many-locals, too-many-state
 
     profile = Profile(cli_ctx=cmd.cli_ctx)
     graph_client = graph_client_factory(cmd.cli_ctx)
-    subscription = profile.get_subscription(subscription=cmd.cli_ctx.data['subscription_id'])
+    subscription = profile.get_subscription(subscription=cmd.cli_ctx.data.get('subscription_id', None))
     tenant_id = subscription[_TENANT_ID]
 
     # if bypass or default_action was specified create a NetworkRuleSet
--- a/src/azure-cli/azure/cli/command_modules/keyvault/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/keyvault/custom.py
@@ -408,7 +408,7 @@ def recover_hsm(cmd, client, hsm_name, resource_group_name, location, no_wait=Fa
 
     # tenantId and sku shouldn't be required
     profile = Profile(cli_ctx=cmd.cli_ctx)
-    tenant_id = profile.get_subscription()[_TENANT_ID]
+    tenant_id = profile.get_subscription(subscription=cmd.cli_ctx.data['subscription_id'])[_TENANT_ID]
 
     # Use 'Recover' as 'create_mode' temporarily since it's a bug from service side making 'create_mode' case-sensitive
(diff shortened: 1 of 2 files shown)


yonzhan:
Keyvault

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
```diff
@@ -401,15 +401,14 @@ def recover_vault_or_hsm(cmd, client, resource_group_name=None, location=None, v
...
-    from azure.cli.core._profile import Profile
+    from azure.cli.core._profile import Profile, _TENANT_ID
 
     ManagedHsm = cmd.get_models('ManagedHsm', resource_type=ResourceType.MGMT_KEYVAULT)
     ManagedHsmSku = cmd.get_models('ManagedHsmSku', resource_type=ResourceType.MGMT_KEYVAULT)
 
     # tenantId and sku shouldn't be required
     profile = Profile(cli_ctx=cmd.cli_ctx)
-    _, _, tenant_id = profile.get_login_credentials(
-        resource=cmd.cli_ctx.cloud.endpoints.active_directory_graph_resource_id)
```
jiasli, line 412: I suddenly found that `az keyvault recover` supports `--subscription`, but `--subscription` doesn't really work, since its value is stored in `cli_ctx.data['subscription_id']`

https://github.com/Azure/azure-cli/blob/21266e02457263420b62fba1c529f69e28852414/src/azure-cli-core/azure/cli/core/commands/client_factory.py#L70

but is not used when calling `profile.get_login_credentials`.

evelyn-ys: `--subscription` actually works as expected because `az keyvault recover` calls mgmt plane client factory and the client factory takes `cli_ctx.data['subscription_id']` as default subscription: https://github.com/Azure/azure-cli/blob/b527951889402f7bdf79a7782ac6e0c9c9b48139/src/azure-cli-core/azure/cli/core/commands/client_factory.py#L69-L70

jiasli: Even though mgmt-plane client factory is created with the correct subscription, the `tenant_id` retrieved by L411 may be incorrect - the current `tenant_id` may not be the same as the `tenant_id` of `--subscription`.

jiasli, line 412: This is indeed why a token for AD Graph is retrieved.

https://github.com/Azure/msrestazure-for-python/blob/8849f398b6ebd4607de63c2f5d1318f44ec1d822/msrestazure/azure_active_directory.py#L592

```py
class MSIAuthentication(BasicTokenAuthentication):
    def __init__(self, port=50342, **kwargs):
        ....
        self.set_token()
```

When `msrestazure.azure_active_directory.MSIAuthentication` is initialized, it gets an access token immediately, not until `signed_session` or `get_token` is called, so `get_login_credentials` will trigger a web request to get AD Graph token.

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
```diff
@@ -401,15 +401,14 @@ def recover_vault_or_hsm(cmd, client, resource_group_name=None, location=None, v
...
+    tenant_id = profile.get_subscription()[_TENANT_ID]
```
jiasli: A little bit hacky, but `azure.cli.command_modules.role` accesses the private attribute `_config` of the client to retrieve the subscription ID:

https://github.com/Azure/azure-cli/blob/00fe3e551fbc574abfd3e099e8e15e6ecb623a06/src/azure-cli/azure/cli/command_modules/role/custom.py#L65-L66

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
```diff
@@ -436,8 +435,7 @@ def recover_vault(cmd, client, vault_name, resource_group_name, location, no_wai
     Sku = cmd.get_models('Sku', resource_type=ResourceType.MGMT_KEYVAULT)
     SkuName = cmd.get_models('SkuName', resource_type=ResourceType.MGMT_KEYVAULT)
     profile = Profile(cli_ctx=cmd.cli_ctx)
-    _, _, tenant_id = profile.get_login_credentials(
-        resource=cmd.cli_ctx.cloud.endpoints.active_directory_graph_resource_id)
+    tenant_id = profile.get_subscription(subscription=cmd.cli_ctx.data['subscription_id'])[_TENANT_ID]
```
jiasli, line 438: This may result in `KeyError` as `subscription_id` may not always be in `cmd.cli_ctx.data`.

A little bit hacky, but `azure.cli.command_modules.role` accesses the private attribute `_config` of the client to retrieve the subscription ID which is from `cli_ctx.data['subscription_id']`:

https://github.com/Azure/azure-cli/blob/00fe3e551fbc574abfd3e099e8e15e6ecb623a06/src/azure-cli/azure/cli/command_modules/role/custom.py#L65-L66

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
jiasli: Another question, why does the original code choses `active_directory_graph_resource_id` instead of `active_directory_resource_id` (ARM's resource ID)? Is it because the name is confusing?

evelyn-ys: got it! updated.

evelyn-ys: > When `msrestazure.azure_active_directory.MSIAuthentication` is initialized, it gets an access token immediately, not until `signed_session` or `get_token` is called, so `get_login_credentials` will trigger a web request to get AD Graph token.

This explains why AAD call shows up when we create keyvault in Cloudshell sandbox as it's following MSI workflow.

evelyn-ys: Anyway the whole call for `get_login_credentials` has been removed. Doesn't matter which resource id we need to use anymore😉

src/azure-cli/azure/cli/command_modules/keyvault/custom.py
```diff
@@ -401,15 +401,14 @@ def recover_vault_or_hsm(cmd, client, resource_group_name=None, location=None, v
...
+    from azure.cli.core._profile import Profile, _TENANT_ID
 
     ManagedHsm = cmd.get_models('ManagedHsm', resource_type=ResourceType.MGMT_KEYVAULT)
     ManagedHsmSku = cmd.get_models('ManagedHsmSku', resource_type=ResourceType.MGMT_KEYVAULT)
 
     # tenantId and sku shouldn't be required
     profile = Profile(cli_ctx=cmd.cli_ctx)
-    _, _, tenant_id = profile.get_login_credentials(
-        resource=cmd.cli_ctx.cloud.endpoints.active_directory_graph_resource_id)
+    tenant_id = profile.get_subscription(subscription=cmd.cli_ctx.data.get('subscription_id', None))[_TENANT_ID]
```
jiasli, line 411: `None` is the default of `default` in `dict.get`: https://docs.python.org/3/library/stdtypes.html#dict.get

jiasli (approved):

PR #26606, opened 2023-06-05
[RDBMS] az mysql flexible-server import create: Add mysql import create command

nitishsharma25:
**Related command**

az mysql flexible-server import create

**Description**

This PR contains a new import create command which will facilitate migrations from mysql single to flexible servers.
---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/mysql/tests/latest/test_mysql_scenario.py
+++ b/src/azure-cli/azure/cli/command_modules/mysql/tests/latest/test_mysql_scenario.py
@@ -90,7 +90,7 @@ class FlexibleServerMgmtScenarioTest(ScenarioTest):
     @AllowLargeResponse()
     @ResourceGroupPreparer(location=DEFAULT_LOCATION)
     def test_mysql_flexible_server_import_create(self, resource_group):
-        self._test_mysql_flexible_server_import_create('mysql', resource_group)
+        self._test_mysql_flexible_server_import_create_mgmt('mysql', resource_group)
 
     # To run this test live, make sure that your role excludes the permission 'Microsoft.DBforMySQL/locations/checkNameAvailability/action'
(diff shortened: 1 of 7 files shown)


src/azure-cli/azure/cli/command_modules/mysql/tests/latest/test_mysql_scenario.py
```diff
@@ -209,6 +214,35 @@ def _test_flexible_server_mgmt(self, database_engine, resource_group):
 
         self.cmd('{} flexible-server delete -g {} -n {} --yes'.format(database_engine, resource_group, restore_server_name), checks=NoneCheck())
 
+    def _test_mysql_flexible_server_import_create_mgmt(self, database_engine, resource_group):
+        storage_size = 32
+        version = '5.7'
+        location = 'eastus'
+        sku_name = 'Standard_B1ms'
+        tier = 'Burstable'
+        resource_group = 'nitishsharma-group'
```
honghr, line 223: I strongly recommend that you can leverage resource group from function entry. It will be generated random resource group name. Because these cases not only for unit test but also used to end to end testing in pipeline.

honghr: And Can we update the location to reference DEFAULT_LOCATION = "eastus2euap"? If this location cannot support your feature, It's OK for now.

nitishsharma25: I have removed my own rg from the test and I am using the rg from the function entry now. I tried using DEFAULT_LOCATION for test but I faced problems there in creation of single server so using westus2 insted.

honghr (approved):

evelyn-ys (approved):

PR #26617, opened 2023-06-06
[App Service] Fix #21133: `az webapp/functionapp config ssl bind/unbind`: Search for matching certificates in the subscription by App Service Plan Id

### LINKED ISSUE #13929: Webapp:az webapp config ssl bind: Cannot find certificate in other Resource groups
**Describe the bug**
`az webapp config ssl bind` needs `--certificate-ID` parameter.

`az webapp config ssl bind` cannot find certificates located OUTSIDE of App Service Plan resource group.

Put differently, `az webapp config ssl bind` will only search App Service Plan resource group for certificate.

This is a problem when App Service Plan is in "SharedPlan" resource group, a web app is in "MyApp" resource group, and the certificate is located "MyApp" in my resource group.

Additionally: `az webapp config ssl create` will create a certificate in the resource group where the Web App lives, NOT where the App Service Plan lives.

**To Reproduce**
- Create App Service Plan in resource group named "SharedPlan"
- Create Web App in resource group called "MyApp"
- Create/upload a ssl certificate within resource group "MyApp".
- Attempt to bind uploaded certificate to MyApp using `az webapp config ssl bind `
- Observe failure to locate certificate thumbpint.

**Expected behavior**
`az webapp config ssl bind` should be able to bind a certificate located outside of the AppService Plan resource group. AZ Portal is able to successfully bind this certificate.

An easy solution is parameter for `--certificate-ID`, rather than relying on certificate thumbprint lookup.

**Environment summary**
az 2.5.1

**Additional context**
This is a condensed version of #9972, which was erroneously closed.

### LINKED ISSUE #21133: WebApp:az webapp config ssl bind command can not import SSL Certificate from different resource group than WebApp
**Describe the bug**
SSL Certificate imported from GUI (Using KeyVault) for the WebApp "A" residing in ResoureGroup "A", gets stored as "microsoft.web/certificates" object in the Resource group "A" where the WebApp "A" resides. From GUI, you can add the same certificate to another webapp, as GUI recognizes the certificate even you are adding it in WebApp "B" which is in Resource Group "B". This is OK if you have less number of apps.

For large organizations, where you have 100's of WebApps: 
If you try az command "az webapp config ssl bind" to bind the same thumbprint, you get error: **"Certificate for thumbprint 'XXXXXXXXXXXX' not found."**

"microsoft.web/certificates" object gets created for each App Service plan for first import of SSL Certificate from KeyVault, which gets stored in the First WebApp where you performed "Import from Keyvault" operation. This object is not movable.

Multiple, certificate objects gets created for multiple App Service Plan, which is not required.

This stops us to make bulk Certificate Addition on multiple webapps.

**To Reproduce**
1. Create two WebApps A & B in two different resource group, i.e. A & B
2. Create key vault and import SSL certificate in resource group "KeyVault"
3. In Azure Portal, goto WebApp A, import SSL Certificate from Keyvault using the "TLS/SSL settings > Private Key Certificates (.pfx) > Import Key Vault Certificate"
4. You will notice that "microsoft.web/certificates" object gets created in Resource Group "A"
5. Now open Powershell to run Az Command
6. Run following command to add the thumbprint of the certificate to WebApp "B"
`az webapp config ssl bind --certificate-thumbprint "XXXXXXXX" --name B --resource-group B --ssl-type SNI`
7. Command will fail with error : Certificate for thumbprint 'XXXXXXXXX' not found.
This error occurs because Az command tries to search certificate in WebApp's Resource group, which is not the actual location of Certificate.

**Expected behavior**
Option A (Preferred)
1. "microsoft.web/certificates" object should be avoided if possible, and the certificate should be directly fetched from KeyVault.
2. Each app service plan should not have seperate ""microsoft.web/certificates" object, and SSL Certificate should get directly fetched from KeyVault.

Option B
1. If, Option A is going to add huge Cost for Read operation of keyvault, then please store the SSL Certificate object "microsoft.web/certificates" in App Service Plan's "Resource Group"
2. Once the object "microsoft.web/certificates" is created in App Service Plan "Resource Group", you can modify AZ command SSL search location to App Service Plan Resource Group.
3. With this method, all the webapps which are residing in their own resource groups can read the certificate and apply it to webapp using az webapp config ssl bind command. This will not limit apps to be residing in single resource group as of "microsoft.web/certificates" Object.

**Environment summary**
azure-cli                         2.32.0
core                              2.32.0
telemetry                          1.0.6
msal                              1.16.0
azure-mgmt-resource               20.0.0

**Additional context**
This change will help large environments where there are multiple app service plans and individual Resource groups per WebApps.

StrawnSC:
**Related command**

`az webapp config ssl bind`
`az webapp config ssl unbind`
`az functionapp config ssl bind`
`az functionapp config ssl unbind`

**Description**

Fixes #21133 & fixes #13929. Currently the CLI only searches for matching certificates in the web app's resource group and the resource group for the app service plan. This change would add logic to search for a matching cert in the user's subscription as well (filtering on the web app's App Service Plan / serverfarmId)  if no match was found using the previous logic. 

Issue_Close: #21133 #13929

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -3078,7 +3078,7 @@ def _update_ssl_binding(cmd, resource_group_name, name, certificate_thumbprint, 
                 found_cert = webapp_cert
     # search for a cert that matches in the subscription, filtering on the serverfarm
     if not found_cert:
-        sub_certs = client.certificates.list(filter=f"ServerFarmId eq '{webapp.server_farm_id}'", api_version="2021-03-01")
+        sub_certs = client.certificates.list(filter=f"ServerFarmId eq '{webapp.server_farm_id}'")
         found_cert = next(iter([c for c in sub_certs if c.thumbprint == certificate_thumbprint]), None)
     if found_cert:
         if not hostname:
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -3066,14 +3066,20 @@ def _update_ssl_binding(cmd, resource_group_name, name, certificate_thumbprint, 
     webapp_certs = client.certificates.list_by_resource_group(cert_resource_group_name)
 
     found_cert = None
+    # search for a cert that matches in the app service plan's RG
     for webapp_cert in webapp_certs:
         if webapp_cert.thumbprint == certificate_thumbprint:
             found_cert = webapp_cert
+    # search for a cert that matches in the webapp's RG
     if not found_cert:
         webapp_certs = client.certificates.list_by_resource_group(resource_group_name)
         for webapp_cert in webapp_certs:
             if webapp_cert.thumbprint == certificate_thumbprint:
                 found_cert = webapp_cert
+    # search for a cert that matches in the subscription
+    if not found_cert:
+        sub_certs = client.certificates.list()
+        found_cert = next([c for c in sub_certs if c.thumbprint == certificate_thumbprint], None)
     if found_cert:
         if not hostname:
             if len(found_cert.host_names) == 1 and not found_cert.host_names[0].startswith('*'):
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -3076,10 +3076,10 @@ def _update_ssl_binding(cmd, resource_group_name, name, certificate_thumbprint, 
         for webapp_cert in webapp_certs:
             if webapp_cert.thumbprint == certificate_thumbprint:
                 found_cert = webapp_cert
-    # search for a cert that matches in the subscription
+    # search for a cert that matches in the subscription, filtering on the serverfarm
     if not found_cert:
-        sub_certs = client.certificates.list()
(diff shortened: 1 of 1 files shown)


yonzhan:
App Service

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -3066,14 +3066,20 @@ def _update_ssl_binding(cmd, resource_group_name, name, certificate_thumbprint,
...
             found_cert = webapp_cert
+    # search for a cert that matches in the webapp's RG
     if not found_cert:
         webapp_certs = client.certificates.list_by_resource_group(resource_group_name)
         for webapp_cert in webapp_certs:
             if webapp_cert.thumbprint == certificate_thumbprint:
                 found_cert = webapp_cert
+    # search for a cert that matches in the subscription, filtering on the serverfarm
+    if not found_cert:
+        sub_certs = client.certificates.list(filter=f"ServerFarmId eq '{webapp.server_farm_id}'", api_version="2021-03-01")
```
zhoxing-ms, line 3081: > api_version="2021-03-01"

May I ask why do we need to hard code the api-version here?

StrawnSC: good catch, thanks. I didn't mean to leave that in

StrawnSC: Just removed that

zhoxing-ms (approved):

Issue #26387, opened 2023-05-09
Gen2 to Trusted Launch Conversion Support - VM Configuration


AjKundnani:
### Related command

`az vm update`

### Resource Provider

Microsoft.Compute/virtualMachines

## Is your feature request related to a problem? Please describe

No, feature request to add output value, new parameter and add values into existing parameters for following commands to support single-instance VM Gen2 -> Trusted Launch conversion:

- [az vm update](https://learn.microsoft.com/en-us/cli/azure/vm?view=azure-cli-latest#az-vm-update)

#### --security-type (`az vm update`)

Customers can use `az vm update` with new parameter **--security-type** to enable *TrustedLaunch* on existing Gen2 (*HyperVGeneration=V2*) VMs. The new parameter will be aligned with `az vm create --security-type` command.

**Example**: `az vm update --security-type TrustedLaunch`

When above command is executed:

- CLI will validate if VM is Gen2 VM. `HyperVGeneration=V2` using OS Disk attributes.
- CLI will validate `securityType` is currently set to `null` or `TrustedLaunch` using VM attributes. **Note**: Backend will not perform any update if securityType is already set to TrustedLaunch.
- Post execution, the `securityType` of VM will be updated to `TrustedLaunch`

--security-Type Value    |    Description
-|-
TrustedLaunch    |    Changes existing Gen2 VM from Standard (or null) Security type to Trusted Launch enabled.

#### Scenario 1 - Gen2 VMs (Non-TL or TL enabled)

1. Customer runs following command to enable **TrustedLaunch** configuration for existing **Gen2 VM** 

    ```azurecli
    az vm update -n name -g group \
        --security-type TrustedLaunch
    ```
2. Set `--enable-secure-boot: false` and `--enable-vtpm: true` if not specified by end user.

3. CLI submits the request to Compute Resource provider Backend API.

#### Scenario 2 - Gen1 VM

1. Customer runs following command to enable **TrustedLaunch** configuration for existing **Gen1 VM** 

    ```azurecli
    az vm update -n name -g group \
        --security-type TrustedLaunch
    ```
2. OS Disk Validation will return HyperVGeneration = `V1` for Gen1 VMs.

3. CLI generates below error message for end user and terminates:

    `Trusted Launch security configuration can be enabled only with Azure Gen2 VMs. Please visit https://learn.microsoft.com/en-us/azure/virtual-machines/trusted-launch for more details.`

#### Scenario 3 - Confidential VM

[...]

PR #26626, opened 2023-06-07
[Compute] `az vm update`: Add parameter `--security-type` to support VM Gen2 to Trusted Launch conversion

yanzhudd:
**Related command**

`az vm update`

**Description**

Add parameter `--security-type` for `az vm update` command to support VM Gen2 to Trusted Launch conversion
Close: #26387

**Testing Guide**

**History Notes**

---


yonzhan:
Compute

src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -417,6 +417,7 @@ def load_arguments(self, _):
         c.argument('ephemeral_os_disk_placement', arg_type=ephemeral_placement_type,
                    help='Only applicable when used with `--size`. Allows you to choose the Ephemeral OS disk provisioning location.')
         c.argument('enable_hibernation', arg_type=get_three_state_flag(), min_api='2021-03-01', help='The flag that enable or disable hibernation capability on the VM.')
+        c.argument('security_type', arg_type=get_enum_type(["TrustedLaunch"], default=None), min_api='2022-11-01', help='Specify the security type of the virtual machine.')
```
zhoxing-ms, line 420: @AjKundnani Do we currently just need to make the `--security-type` parameter only support `TrustedLaunch`? Or we need to support all security type as `az vm create`?

AjKundnani: @zhoxing-ms - Currently `TrustedLaunch` only because upgrade of existing VMs to ConfidentialVM is not supported.

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -1622,21 +1622,40 @@ def update_vm(cmd, resource_group_name, vm_name, os_disk=None, disk_caching=None
...
+            vm_id_parsed = parse_resource_id(vm.id)
+            disk_id = resource_id(subscription=vm_id_parsed['subscription'],
+                                  resource_group=vm_id_parsed['resource_group'],
                                   namespace='Microsoft.Compute', type='disks', name=os_disk)
-            disk_name = os_disk
+            disk_resource_group, disk_name = vm_id_parsed['resource_group'], os_disk
         vm.storage_profile.os_disk.managed_disk.id = disk_id
         vm.storage_profile.os_disk.name = disk_name
 
+    if security_type is not None and security_type == "TrustedLaunch":
```
zhoxing-ms, line 1645:
```suggestion
    if security_type == "TrustedLaunch":
```

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -1622,21 +1622,40 @@ def update_vm(cmd, resource_group_name, vm_name, os_disk=None, disk_caching=None
...
+        if disk_name is None and vm.storage_profile.os_disk.managed_disk is not None:
+            os_disk_id_parsed = parse_resource_id(vm.storage_profile.os_disk.managed_disk.id)
```
zhoxing-ms, line 1647: > vm.storage_profile.os_disk.managed_disk.id

Could we guarantee that there is no None pointer issue with this variable?

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -1622,21 +1622,40 @@ def update_vm(cmd, resource_group_name, vm_name, os_disk=None, disk_caching=None
...
+            disk_resource_group, disk_name = os_disk_id_parsed['resource_group'], os_disk_id_parsed['name']
+
+        if disk_name is not None:
+            from ._vm_utils import validate_update_vm_trusted_launch_supported
+
+            validate_update_vm_trusted_launch_supported(cmd=cmd, vm=vm, os_disk_resource_group=disk_resource_group,
+                                                        os_disk_name=disk_name)
```
zhoxing-ms, line 1654: > Scenario 1 - Gen2 VMs (Non-TL or TL enabled)
....
Set --enable-secure-boot: false and --enable-vtpm: true if not specified by end user.

We also need to change those default values if they are not specified by customers, right?

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -1622,21 +1622,40 @@ def update_vm(cmd, resource_group_name, vm_name, os_disk=None, disk_caching=None
...
+            if vm.security_profile is None:
+                vm.security_profile = {}
+            vm.security_profile['security_type'] = security_type
```
zhoxing-ms, line 1657: > Scenario 1 - Gen2 VMs (Non-TL or TL enabled)
....
Set --enable-secure-boot: false and --enable-vtpm: true if not specified by end user.

We also need to change those default values if they are not specified by customers, right?

src/azure-cli/azure/cli/command_modules/vm/_params.py
zhoxing-ms: Got it, thanks~

src/azure-cli/azure/cli/command_modules/vm/custom.py
yanzhudd: It seems it is not necessary to specify the default values manually since they are set by the backend side.
@AjKundnani could you please help to confirm this?

src/azure-cli/azure/cli/command_modules/vm/custom.py
yanzhudd: > if disk_name is None and vm.storage_profile.os_disk.managed_disk is not None:

since there is a if statement `vm.storage_profile.os_disk.managed_disk is not None`, the `id` of the managed disk may should not be None.

src/azure-cli/azure/cli/command_modules/vm/custom.py
AjKundnani: @yanzhudd - For Gen2 to TL upgrade yes the backend set the default flag to Secure Boot OFF and vTPM ON.
To be on safer side will recommend setting default at CLI client tool layer as well; if it does not involves too many changes.

yanzhudd: Hi @AjKundnani, may I ask if the default values would be changed frequently?

AjKundnani: @yanzhudd - No default values will not be changed frequently. There'll only be one change required to turn SB ON in future once we've adequate evidence to turn it ON without impacting customers.

zhoxing-ms (approved):

PR #26640, opened 2023-06-09
[Packaging] Support x86 and x64 MSI builds

### LINKED ISSUE #18766: Support Azure CLI 64 Bit

Today we see that Azure CLI MSI installer only supports a 32 bit install of Azure CLI. Many other cloud providers and projects are moving away from supporting 32 bit in favor of 64 bit. In our case, we are building an extension which leverages an SDK that only supports 64 bit. For that reason, for some calls in the extension we run into issues because of 32 bit limitation in CLI

We would like to see a natively supported 64 bit version of Azure CLI available to end users. This is available for AWS and GCP today, so it would be good to do the same in Azure. 

We had previously reached out the Azure CLI team and we were given a temporary workaround is to have Azure CLI point to a 64 bit version of python. I will try this, and see if it resolves our issues, but I believe we can find a more permanent solution.

heaths:
Close #18766

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -206,8 +206,8 @@ jobs:
 
 
   - script: |
-      set /p ARCH=$(Platform)
-      set /p CLI_VERSION=<$(System.ArtifactsDirectory)/metadata/version
+      set ARCH=$(Platform)
+      set CLI_VERSION=<$(System.ArtifactsDirectory)/metadata/version
       set
       
       build_scripts/windows/scripts/build.cmd
@@ -216,16 +216,22 @@ jobs:
   - task: AzureArtifacts.manifest-generator-task.manifest-generator-task.ManifestGeneratorTask@0
     displayName: 'SBOM'
     inputs:
-      BuildDropPath: 'build_scripts/windows/out/$(Platform)'
+      BuildDropPath: 'build_scripts/windows/out/'
 
   - task: PublishPipelineArtifact@0
     displayName: 'Publish Artifact: MSI'
     inputs:
-      TargetPath: 'build_scripts/windows/out/$(Platform)'
-      ArtifactName: msi
+      TargetPath: 'build_scripts/windows/out/'
+      ArtifactName: msi-$(Platform)
 
 - job: TestWindowsMSI
   displayName: Test Windows MSI
+  strategy:
+    matrix:
+      x86:
+        Platform: x86
+      x64:
+        Platform: x64
 
   dependsOn: BuildWindowsMSI
   condition: and(succeeded(), in(variables['Build.Reason'], 'IndividualCI', 'BatchedCI', 'Manual', 'Schedule'))
@@ -242,10 +248,10 @@ jobs:
     displayName: 'Download Build Artifacts'
     inputs:
       TargetPath: '$(Build.ArtifactStagingDirectory)/msi'
-      artifactName: msi
+      artifactName: msi-$(Platform)
 
   - task: PowerShell@2
-    displayName: Install and Load x86 CLI
+    displayName: Install and Load CLI
     inputs:
       targetType: inline
       script: |
@@ -261,53 +267,7 @@ jobs:
         
         $InstallArgs = @(
           "/i"
-          "`"$env:SYSTEM_ARTIFACTSDIRECTORY\msi\x86\Microsoft Azure CLI.msi`""
-          "/q"
-          "/norestart"
-          "/l*v"
-          ".\install_logs.txt"
-        )
-        $pre_installed_version=az version --query '\"azure-cli\"' -o tsv
-        $to_be_installed_version=Get-Content $(System.ArtifactsDirectory)/metadata/version
-        if ($pre_installed_version -eq $to_be_installed_version){
-          # See https://docs.microsoft.com/windows/win32/msi/reinstallmode about options of REINSTALLMODE
-          $reinstall_option="REINSTALL=ALL REINSTALLMODE=emus"
-          $InstallArgs += $reinstall_option
-        }
-        Start-Process "msiexec.exe" -ArgumentList $InstallArgs -Wait -NoNewWindow
-        $install_time=Measure-Command {Start-Process "msiexec.exe" -ArgumentList $InstallArgs -Wait -NoNewWindow} | select -expand TotalSeconds
-        $installed_version=az version --query '\"azure-cli\"' -o tsv
-        if ($installed_version -ne $to_be_installed_version){
-          echo "The MSI failed to install."
-          Exit 1
-        }
-        echo 'Install time(seconds):' $install_time
-        az --version
-        # Test bundled pip with extension installation
-        az extension add -n rdbms-connect
-        az self-test
-        
-        Get-Content .\install_logs.txt
-
-  - task: PowerShell@2
-    # Test install (upgrade of x86) of x64 MSI as well.
-    displayName: Install and Load x64 CLI
-    inputs:
-      targetType: inline
-      script: |
-        if (-NOT ([Security.Principal.WindowsPrincipal][Security.Principal.WindowsIdentity]::GetCurrent()).IsInRole([Security.Principal.WindowsBuiltInRole] "Administrator")) {
-          # Start another Powershell process as Admin and execute this script again
-          $arguments = "& '" +$myinvocation.mycommand.definition + "'"
-          Start-Process powershell -Verb runAs -ArgumentList $arguments
-          # Stop if the PowerShell is not run as Admin
-          Break
-        }
-        # The following are executed by elevated PowerShell
-        az --version
-        
-        $InstallArgs = @(
-          "/i"
-          "`"$env:SYSTEM_ARTIFACTSDIRECTORY\msi\x64\Microsoft Azure CLI.msi`""
+          "`"$env:SYSTEM_ARTIFACTSDIRECTORY\msi\Microsoft Azure CLI.msi`""
           "/q"
           "/norestart"
           "/l*v"
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -207,7 +207,7 @@ jobs:
 
   - script: |
       set ARCH=$(Platform)
-      set CLI_VERSION=<$(System.ArtifactsDirectory)/metadata/version
+      set /p CLI_VERSION=<$(System.ArtifactsDirectory)/metadata/version
       set
       
       build_scripts/windows/scripts/build.cmd
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -289,8 +289,7 @@ jobs:
         
         Get-Content .\install_logs.txt
 
-        - task: PowerShell@2
-
+  - task: PowerShell@2
     # Test install (upgrade of x86) of x64 MSI as well.
     displayName: Install and Load x64 CLI
     inputs:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -186,6 +186,12 @@ jobs:
 
 - job: BuildWindowsMSI
   displayName: Build Windows MSI
+  strategy:
+    matrix:
+      x86:
+        Platform: x86
+      x64:
+        Platform: x64
 
   dependsOn: ExtractMetadata
   condition: succeeded()
@@ -200,6 +206,7 @@ jobs:
 
 
   - script: |
+      set /p ARCH=$(Platform)
       set /p CLI_VERSION=<$(System.ArtifactsDirectory)/metadata/version
       set
       
@@ -209,12 +216,12 @@ jobs:
   - task: AzureArtifacts.manifest-generator-task.manifest-generator-task.ManifestGeneratorTask@0
     displayName: 'SBOM'
     inputs:
-      BuildDropPath: 'build_scripts/windows/out/'
+      BuildDropPath: 'build_scripts/windows/out/$(Platform)'
 
   - task: PublishPipelineArtifact@0
     displayName: 'Publish Artifact: MSI'
     inputs:
-      TargetPath: 'build_scripts/windows/out/'
+      TargetPath: 'build_scripts/windows/out/$(Platform)'
       ArtifactName: msi
 
 - job: TestWindowsMSI
@@ -238,7 +245,7 @@ jobs:
       artifactName: msi
 
   - task: PowerShell@2
-    displayName: Install and Load CLI
+    displayName: Install and Load x86 CLI
     inputs:
       targetType: inline
       script: |
@@ -254,7 +261,54 @@ jobs:
         
         $InstallArgs = @(
           "/i"
-          "`"$env:SYSTEM_ARTIFACTSDIRECTORY\msi\Microsoft Azure CLI.msi`""
+          "`"$env:SYSTEM_ARTIFACTSDIRECTORY\msi\x86\Microsoft Azure CLI.msi`""
+          "/q"
+          "/norestart"
+          "/l*v"
+          ".\install_logs.txt"
+        )
+        $pre_installed_version=az version --query '\"azure-cli\"' -o tsv
+        $to_be_installed_version=Get-Content $(System.ArtifactsDirectory)/metadata/version
+        if ($pre_installed_version -eq $to_be_installed_version){
+          # See https://docs.microsoft.com/windows/win32/msi/reinstallmode about options of REINSTALLMODE
+          $reinstall_option="REINSTALL=ALL REINSTALLMODE=emus"
+          $InstallArgs += $reinstall_option
+        }
+        Start-Process "msiexec.exe" -ArgumentList $InstallArgs -Wait -NoNewWindow
+        $install_time=Measure-Command {Start-Process "msiexec.exe" -ArgumentList $InstallArgs -Wait -NoNewWindow} | select -expand TotalSeconds
+        $installed_version=az version --query '\"azure-cli\"' -o tsv
+        if ($installed_version -ne $to_be_installed_version){
+          echo "The MSI failed to install."
+          Exit 1
+        }
+        echo 'Install time(seconds):' $install_time
+        az --version
+        # Test bundled pip with extension installation
+        az extension add -n rdbms-connect
+        az self-test
+        
+        Get-Content .\install_logs.txt
+
+        - task: PowerShell@2
+
+    # Test install (upgrade of x86) of x64 MSI as well.
+    displayName: Install and Load x64 CLI
+    inputs:
+      targetType: inline
+      script: |
+        if (-NOT ([Security.Principal.WindowsPrincipal][Security.Principal.WindowsIdentity]::GetCurrent()).IsInRole([Security.Principal.WindowsBuiltInRole] "Administrator")) {
+          # Start another Powershell process as Admin and execute this script again
+          $arguments = "& '" +$myinvocation.mycommand.definition + "'"
+          Start-Process powershell -Verb runAs -ArgumentList $arguments
+          # Stop if the PowerShell is not run as Admin
+          Break
+        }
+        # The following are executed by elevated PowerShell
+        az --version
+        
+        $InstallArgs = @(
+          "/i"
+          "`"$env:SYSTEM_ARTIFACTSDIRECTORY\msi\x64\Microsoft Azure CLI.msi`""
           "/q"
           "/norestart"
           "/l*v"
--- a/build_scripts/windows/Product.wxs
+++ b/build_scripts/windows/Product.wxs
@@ -8,18 +8,55 @@
   <?define ProductAuthor = "Microsoft Corporation" ?>
   <?define ProductResources = ".\resources\" ?>
 
+  <?if $(var.Platform) = "x64" ?>
+  <?define AzureCliRegistryGuid = "90b6367e-4e15-4313-af33-a0ef459a56fc" ?>
+  <?define AzureCliSystemPathGuid = "12c8f0dc-0475-4814-93b6-fcbe82e89532" ?>
+  <?define AzureCliVersionGuid = "4df4882f-4604-4cdc-a0b4-bffc4b8c5967" ?>
+  <?define ProgramFilesFolder = "ProgramFiles64Folder" ?>
+  <?define RemoveAzureCLIFolderGuid = "7ce8a521-7429-4ded-9cce-0948e393cc83" ?>
+  <?define RemoveCLIFolderGuid = "59fb416c-91d0-49aa-a185-d4330c435a93" ?>
+  <?define UpgradeCode = "90762fec-9554-4729-a107-c6a8ea316698" ?>
+
+  <?elseif $(var.Platform) = "x86" ?>
+  <?define AzureCliRegistryGuid = "BDEEE50E-70D3-4990-BFF9-FCF8114AA701" ?>
+  <?define AzureCliSystemPathGuid = "3B60CA67-DB92-465A-BC96-3CA6CAEFF41B" ?>
+  <?define AzureCliVersionGuid = "A5647E93-02B7-4CED-802A-DDD7416E0231" ?>
+  <?define ProgramFilesFolder = "ProgramFilesFolder" ?>
+  <?define RemoveAzureCLIFolderGuid = "6AF97701-6A7A-4292-95D6-ED34CD4C97C0" ?>
+  <?define RemoveCLIFolderGuid = "24DA394D-B03D-4623-A0D9-852DB51EF9F4" ?>
+  <?define UpgradeCode = "dff82af0-3f95-4ac9-8efd-948604fdb028" ?>
+
+  <?else ?>
+  <?error Unsupported platform "$(var.Platform)" ?>
+  <?endif ?>
+
   <Product Id="*"
            Name="$(var.ProductName)"
            Language="1033"
            Version="$(var.ProductVersion)"
            Manufacturer="$(var.ProductAuthor)"
-           UpgradeCode="dff82af0-3f95-4ac9-8efd-948604fdb028">
+           UpgradeCode="$(var.UpgradeCode)">
 
     <Package InstallerVersion="200"
              Compressed="yes"
              InstallScope="perMachine" />
 
-    <MajorUpgrade DowngradeErrorMessage="A newer version of [ProductName] is already installed." Schedule="afterInstallExecute" />
+    <Upgrade Id="$(var.UpgradeCode)">
+      <UpgradeVersion Property="WIX_UPGRADE_DETECTED" Maximum="$(var.ProductVersion)" IncludeMinimum="no" MigrateFeatures="yes" />
+      <UpgradeVersion Property="WIX_DOWNGRADE_DETECTED" Minimum="$(var.ProductVersion)" IncludeMinimum="no" OnlyDetect="yes" />
+    </Upgrade>
+    <Condition Message="A newer version of [ProductName] is already installed.">NOT WIX_DOWNGRADE_DETECTED</Condition>
+    <InstallExecuteSequence>
+      <RemoveExistingProducts After="InstallExecute" />
+    </InstallExecuteSequence>
+
+    <!-- New product architectures should upgrade the original x86 product - even of the same version. -->
+    <?if $(var.UpgradeCode) != "dff82af0-3f95-4ac9-8efd-948604fdb028" ?>
+    <Upgrade Id="dff82af0-3f95-4ac9-8efd-948604fdb028">
+      <UpgradeVersion Property="WIX_UPGRADE_DETECTED" Maximum="$(var.ProductVersion)" IncludeMinimum="yes" MigrateFeatures="yes" />
+      <UpgradeVersion Property="WIX_DOWNGRADE_DETECTED" Minimum="$(var.ProductVersion)" IncludeMinimum="no" OnlyDetect="yes" />
+    </Upgrade>
+    <?endif ?>
 
     <Media Id="1" Cabinet="WindowsAzureCLI.cab" EmbedCab="yes" CompressionLevel="high" />
 
@@ -61,7 +98,7 @@
 
   <Fragment>
     <Directory Id="TARGETDIR" Name="SourceDir">
-      <Directory Id="ProgramFilesFolder">
+      <Directory Id="$(var.ProgramFilesFolder)">
         <Directory Id="SDKFOLDER" Name="Microsoft SDKs">
           <Directory Id="AZUREFOLDER" Name="Azure">
             <Directory Id="AZURECLIFOLDER" Name="CLI2">
@@ -78,17 +115,17 @@
     <ComponentGroup Id="AzureCliSettingsGroup">
       <Component Id="RemoveCLIFolder"
                  Directory="DynamicCliDir"
-                 Guid="{24DA394D-B03D-4623-A0D9-852DB51EF9F4}">
+                 Guid="$(var.RemoveCLIFolderGuid)">
         <RemoveFolder Id="DynamicCliDir" On="uninstall" />
       </Component>
       <Component Id="RemoveAzureCLIFolder"
                  Directory="AZURECLIFOLDER"
-                 Guid="{6AF97701-6A7A-4292-95D6-ED34CD4C97C0}">
+                 Guid="$(var.RemoveAzureCLIFolderGuid)">
         <RemoveFolder Id="AZURECLIFOLDER" On="uninstall" />
       </Component>
       <Component Id="AzureCliSystemPath"
                  Directory="DynamicCliDir"
-                 Guid="{3B60CA67-DB92-465A-BC96-3CA6CAEFF41B}">
+                 Guid="$(var.AzureCliSystemPathGuid)">
         <Environment Id="AzureCliAddedToPATH"
                      Name="PATH"
                      Value="[DynamicCliDir]wbin"
@@ -100,7 +137,7 @@
       </Component>
       <Component Id="AzureCliRegistry"
                  Directory="DynamicCliDir"
-                 Guid="{BDEEE50E-70D3-4990-BFF9-FCF8114AA701}">
+                 Guid="$(var.AzureCliRegistryGuid)">
         <RegistryValue Root="HKCU"
                        Key="Software\Microsoft\$(var.ProductName)"
                        Name="installed"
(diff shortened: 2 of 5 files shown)


yonzhan:

build_scripts/windows/scripts/build.cmd
```diff
@@ -13,28 +13,46 @@ if "%CLI_VERSION%"=="" (
...
 set TEMP_SCRATCH_FOLDER=%ARTIFACTS_DIR%\cli_scratch
 set BUILDING_DIR=%ARTIFACTS_DIR%\cli
 set WIX_DIR=%ARTIFACTS_DIR%\wix
 set PYTHON_DIR=%ARTIFACTS_DIR%\Python
 
-set REPO_ROOT=%~dp0..\..\..
+REM Get the absolute directory since we pushd into different levels of subdirectories.
+PUSHD %~dp0..\..\..
+SET REPO_ROOT=%CD%
+POPD
```
jiasli, line 55: This is indeed more elegant than the current implementation. The current implementation sets `REPO_ROOT` to `D:\cli\azure-cli\build_scripts\windows\scripts\..\..\..` and the new approach sets `REPO_ROOT` to `D:\cli\azure-cli`, making it much more readable.

azure-pipelines.yml
```diff
@@ -209,12 +216,12 @@ jobs:
...
     inputs:
-      BuildDropPath: 'build_scripts/windows/out/'
+      BuildDropPath: 'build_scripts/windows/out/$(Platform)'
 
   - task: PublishPipelineArtifact@0
     displayName: 'Publish Artifact: MSI'
     inputs:
-      TargetPath: 'build_scripts/windows/out/'
+      TargetPath: 'build_scripts/windows/out/$(Platform)'
       ArtifactName: msi
```
jiasli, line 225: I am not entirely sure whether artifacts from different `matrix` (matrices) will be merged or overwritten. Let's give it a test.

azure-pipelines.yml
```diff
@@ -238,7 +245,54 @@ jobs:
...
+        }
+        echo 'Install time(seconds):' $install_time
+        az --version
+        # Test bundled pip with extension installation
+        az extension add -n rdbms-connect
+        az self-test
+        
+        Get-Content .\install_logs.txt
+
+        - task: PowerShell@2
```
jiasli, line 292:
```suggestion
  - task: PowerShell@2
```

azure-pipelines.yml
heaths: For posterity (since we talked internally about this briefly), the more changes I have to make to support building both x86 and x64 on a single machine, the more I'm leaning toward not having to support that and having separate artifacts would make more sense. It's all the little "gotchas" that worry me about changing the output paths radically, since that permiates throughout the build and pipelines.

azure-pipelines.yml
```diff
@@ -200,6 +206,7 @@ jobs:
 
 
   - script: |
+      set /p ARCH=$(Platform)
```
bebound, line 209:
```suggestion
      set ARCH=$(Platform)
```

`/p` means get value from user input, thus this env is missing.

build_scripts/windows/scripts/build.cmd
```diff
@@ -72,6 +87,9 @@ if not exist %WIX_DIR% (
     popd
 )
 
+REM Use only downloaded python and ignore machine state
+set PYTHONHOME=%PYTHON_DIR%
+
```
bebound, line 92:
```suggestion
```
Once `PYTHONHOME` is defined.  `azure-cli` will be installed in `PYTHON_DIR`(artifacts\x64\Python\Lib) instead of `BUILDING_DIR` (artifacts\x64\cli\Lib).

build_scripts/windows/Product.wxs
```diff
@@ -8,18 +8,55 @@
...
+    <InstallExecuteSequence>
+      <RemoveExistingProducts After="InstallExecute" />
+    </InstallExecuteSequence>
+
+    <!-- New product architectures should upgrade the original x86 product - even of the same version. -->
+    <?if $(var.UpgradeCode) != "dff82af0-3f95-4ac9-8efd-948604fdb028" ?>
+    <Upgrade Id="dff82af0-3f95-4ac9-8efd-948604fdb028">
+      <UpgradeVersion Property="WIX_UPGRADE_DETECTED" Maximum="$(var.ProductVersion)" IncludeMinimum="yes" MigrateFeatures="yes" />
+      <UpgradeVersion Property="WIX_DOWNGRADE_DETECTED" Minimum="$(var.ProductVersion)" IncludeMinimum="no" OnlyDetect="yes" />
+    </Upgrade>
```
bebound, line 58: This error is raised when build 64-bit MSI
```
  D:\a\_work\1\s\build_scripts\windows\Product.wxs(56): error LGHT0204: ICE61: Upgrade.ActionProperty WIX_UPGRADE_DETECTED may be used in only one record of the Upgrade table. [D:\a\_work\1\s\build_scripts\windows\azure-cli.wixproj]
  D:\a\_work\1\s\build_scripts\windows\Product.wxs(57): error LGHT0204: ICE61: Upgrade.ActionProperty WIX_DOWNGRADE_DETECTED may be used in only one record of the Upgrade table. [D:\a\_work\1\s\build_scripts\windows\azure-cli.wixproj]

```

azure-pipelines.yml
bebound: If `TargetPath` is `build_scripts/windows/out/$(Platform)`, x86 and x64 produce same file in that folder. Only one package will be saved.
I prefer jiasli's solution, it's more concise and less changes to made.

build_scripts/windows/Product.wxs
heaths: Practically it's not an issue, but given it's an ICE *error* I can't suppress it, so I'll have to work around it. FWIW, I can't run ICEs non-elevated, and I don't do development elevated if I can avoid it.

build_scripts/windows/scripts/build.cmd
heaths: I think I'm going to back out all the changes for rehoming the bits anyway. The idea was to make it easy to build packages for either arch on a single machine, but I'm worried about downstream effects in build, publishing, testing, etc. Already the plan was to have separate agents build. A local dev can simply clean the output folder, re-set env vars, and build again.

heaths: I would recommend, though, considering that in a separate PR that fixes multi-target build. Most build systems support this, and in the long run you're probably going to want to if you keep building x86 bits.

azure-pipelines.yml
jiasli: > If `TargetPath` is `build_scripts/windows/out/$(Platform)`, x86 and x64 produce same file in that folder. Only one package will be saved.

Not really. In `build_scripts/windows/azure-cli.wixproj`,

```xml
<OutputPath>out\</OutputPath>
```

is changed to

```xml
<OutputPath>out\$(Platform)\</OutputPath>
```

So built MSIs are separated by `$(Platform)` folders.

azure-pipelines.yml
jiasli: BTW, L210 works because it uses `<` to read from a file.

azure-pipelines.yml
heaths: Keep things separate is quickly becoming a pervasive problem, though. A lot of scripts have to change, and workflows. I'm tempted to go back to a single output location and let separate agents build separate artifacts. I generally like to keep PRs focused on specific problems, and this feels a lot more like a separate effort to support multi-target builds across the build and publish pipeline.

build_scripts/windows/scripts/build.cmd
bebound: `PYTHON_DIR` (artifacts\x64\Python\Lib) is not in MSI package path, so these libraries are not in final MSI.
The build MSI is 15MB, which is much smaller than normal size (50MB).

azure-pipelines.yml
bebound: @jiasli I mean `TargetPath` needs to be set as `build_scripts/windows/out/` if we want to keep both builds in one artifact.

azure-pipelines.yml
```diff
@@ -206,8 +206,8 @@ jobs:
 
 
   - script: |
-      set /p ARCH=$(Platform)
-      set /p CLI_VERSION=<$(System.ArtifactsDirectory)/metadata/version
+      set ARCH=$(Platform)
+      set CLI_VERSION=<$(System.ArtifactsDirectory)/metadata/version
```
bebound, line 210: `/p` is required in line 210, it reads from file.

build_scripts/windows/azure-cli.wixproj
```diff
@@ -21,6 +12,15 @@
...
   </PropertyGroup>
+  <!-- Local WiX -->
+  <PropertyGroup>
+    <LocalWixRoot>artifacts\wix</LocalWixRoot>
+    <WixToolPath>$(LocalWixRoot)</WixToolPath>
+    <WixTargetsPath Condition="Exists('$(WixToolPath)\Wix.targets')">$(WixToolPath)\Wix.targets</WixTargetsPath>
+    <WixTasksPath Condition="Exists('$(WixToolPath)\wixtasks.dll')">wixtasks.dll</WixTasksPath>
+    <AzureCliSource>artifacts\cli</AzureCliSource>
+    <LinkerAdditionalOptions>-fv</LinkerAdditionalOptions>
+  </PropertyGroup>
```
jiasli, line 23: I am a little bit curious about why this `PropertyGroup` is moved here.

build_scripts/windows/azure-cli.wixproj
```diff
@@ -21,6 +12,15 @@
     <WixTargetsPath Condition=" '$(WixTargetsPath)' == '' AND '$(MSBuildExtensionsPath32)' != '' ">$(MSBuildExtensionsPath32)\Microsoft\WiX\v3.x\Wix.targets</WixTargetsPath>
     <WixTargetsPath Condition=" '$(WixTargetsPath)' == '' ">$(MSBuildExtensionsPath)\Microsoft\WiX\v3.x\Wix.targets</WixTargetsPath>
   </PropertyGroup>
+  <!-- Local WiX -->
+  <PropertyGroup>
+    <LocalWixRoot>artifacts\wix</LocalWixRoot>
+    <WixToolPath>$(LocalWixRoot)</WixToolPath>
+    <WixTargetsPath Condition="Exists('$(WixToolPath)\Wix.targets')">$(WixToolPath)\Wix.targets</WixTargetsPath>
+    <WixTasksPath Condition="Exists('$(WixToolPath)\wixtasks.dll')">wixtasks.dll</WixTasksPath>
```
jiasli, line 20: I was wondering why we need to add `Condition="Exists('$(WixToolPath)\Wix.targets')"` here.

build_scripts/windows/azure-cli.wixproj
heaths: To pick up the properties I set above. Could be all in the same PropertyGroup, but I left them separate with your comments for easier identification as to what they do.

build_scripts/windows/azure-cli.wixproj
heaths: It's to support building locally via msbuild instead of your scripts that do more. For quick testing, I just run msbuild from a VS Dev prompt. It won't hurt anything outside of that but will be easier for future edits.

build_scripts/windows/azure-cli.wixproj
```diff
@@ -1,14 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <!-- Local WiX -->
-  <PropertyGroup>
-    <LocalWixRoot>artifacts\wix\</LocalWixRoot>
-    <WixToolPath>$(LocalWixRoot)</WixToolPath>
-    <WixTargetsPath>$(WixToolPath)Wix.targets</WixTargetsPath>
-    <WixTasksPath>wixtasks.dll</WixTasksPath>
```
jiasli, line 8: Shouldn't this be `$(WixToolPath)wixtasks.dll`?

build_scripts/windows/azure-cli.wixproj
jiasli, line 20: Don't we need `$(WixToolPath)\` here?

```suggestion
    <WixTasksPath Condition="Exists('$(WixToolPath)\wixtasks.dll')">$(WixToolPath)\wixtasks.dll</WixTasksPath>
```

bebound (approved):

build_scripts/windows/azure-cli.wixproj
jiasli: https://wixtoolset.org/docs/v3/msbuild/daily_builds/#step-2-modify-your-wixproj-file

image

The example specifies `WixTasksPath` as `$(WixToolPath)wixtasks.dll`, but it also mentions:

> The WixTasksPath property is used inside wix.targets to load WixTasks.dll; its value, if a relative path, must be relative to the wix.targets file.

jiasli, line 20: I guess it's better to follow the convention of the official example and the previous code.

```suggestion
    <LocalWixRoot>artifacts\wix\</LocalWixRoot>
    <WixToolPath>$(LocalWixRoot)</WixToolPath>
    <WixTargetsPath Condition="Exists('$(WixToolPath)Wix.targets')">$(WixToolPath)Wix.targets</WixTargetsPath>
    <WixTasksPath Condition="Exists('$(WixToolPath)wixtasks.dll')">wixtasks.dll</WixTasksPath>
```

build_scripts/windows/Product.wxs
```diff
@@ -8,18 +8,55 @@
...
   <?define ProductResources = ".\resources\" ?>
 
+  <?if $(var.Platform) = "x64" ?>
+  <?define AzureCliRegistryGuid = "90b6367e-4e15-4313-af33-a0ef459a56fc" ?>
+  <?define AzureCliSystemPathGuid = "12c8f0dc-0475-4814-93b6-fcbe82e89532" ?>
+  <?define AzureCliVersionGuid = "4df4882f-4604-4cdc-a0b4-bffc4b8c5967" ?>
+  <?define ProgramFilesFolder = "ProgramFiles64Folder" ?>
+  <?define RemoveAzureCLIFolderGuid = "7ce8a521-7429-4ded-9cce-0948e393cc83" ?>
+  <?define RemoveCLIFolderGuid = "59fb416c-91d0-49aa-a185-d4330c435a93" ?>
+  <?define UpgradeCode = "90762fec-9554-4729-a107-c6a8ea316698" ?>
```
jiasli, line 18: Are these randomly generated GUIDs?

build_scripts/windows/Product.wxs
```diff
@@ -8,18 +8,55 @@
...
+    <Condition Message="A newer version of [ProductName] is already installed.">NOT WIX_DOWNGRADE_DETECTED</Condition>
+    <InstallExecuteSequence>
+      <RemoveExistingProducts After="InstallExecute" />
+    </InstallExecuteSequence>
+
+    <!-- New product architectures should upgrade the original x86 product - even of the same version. -->
+    <?if $(var.UpgradeCode) != "dff82af0-3f95-4ac9-8efd-948604fdb028" ?>
+    <Upgrade Id="dff82af0-3f95-4ac9-8efd-948604fdb028">
+      <UpgradeVersion Property="WIX_X86_UPGRADE_DETECTED" Maximum="$(var.ProductVersion)" IncludeMinimum="yes" MigrateFeatures="yes" />
+      <UpgradeVersion Property="WIX_X86_DOWNGRADE_DETECTED" Minimum="$(var.ProductVersion)" IncludeMinimum="no" OnlyDetect="yes" />
```
jiasli, line 57: Should it fail for `WIX_X86_DOWNGRADE_DETECTED`?

build_scripts/windows/azure-cli.wixproj
heaths: No. The wix targets will canonicalize as needed.

build_scripts/windows/azure-cli.wixproj
heaths: Personally, I've never been a fan of potentially under-normalizing. Many paths - even MSBuild standard paths - may always be separator-terminated while others aren't. It's a cluster of different behaviors. There has almost never (and when it has been a problem, it's because of some third-party EXE invoked by a task) been a problem from including too many directory separators. Better safe than sorry, IMO.

jiasli: Thanks @heaths, then I guess the WiX document should be updated instead. 🤣

build_scripts/windows/Product.wxs
heaths: Yes.

build_scripts/windows/azure-cli.wixproj
```diff
@@ -21,6 +12,15 @@
     <WixTargetsPath Condition=" '$(WixTargetsPath)' == '' AND '$(MSBuildExtensionsPath32)' != '' ">$(MSBuildExtensionsPath32)\Microsoft\WiX\v3.x\Wix.targets</WixTargetsPath>
     <WixTargetsPath Condition=" '$(WixTargetsPath)' == '' ">$(MSBuildExtensionsPath)\Microsoft\WiX\v3.x\Wix.targets</WixTargetsPath>
   </PropertyGroup>
+  <!-- Local WiX -->
+  <PropertyGroup>
+    <LocalWixRoot>artifacts\wix</LocalWixRoot>
+    <WixToolPath>$(MSBuildThisFileDirectory)$(LocalWixRoot)</WixToolPath>
+    <WixTargetsPath Condition="Exists('$(WixToolPath)\Wix.targets')">$(WixToolPath)\Wix.targets</WixTargetsPath>
+    <WixTasksPath Condition="Exists('$(WixToolPath)\wixtasks.dll')">$(WixToolPath)\wixtasks.dll</WixTasksPath>
```
jiasli, line 20: As we discussed offline, if `WixTasksPath` is a relative path, it is relative to `WixTargetsPath`, instead of the current file `build_scripts/windows/azure-cli.wixproj`. If we change `WixTasksPath` to `$(WixToolPath)\wixtasks.dll`, it appends it:

```
error MSB4062: The "HeatDirectory" task could not be loaded from the assembly C:\src\azure-cli\build_scripts\windows\artifacts\wix\artifacts\wix\wixtasks.dll
```

I think it is fine to keep relative path, as it makes the code succinct, doesn't it? It's the document https://wixtoolset.org/docs/v3/msbuild/daily_builds/#step-2-modify-your-wixproj-file that should be updated. 😉

```suggestion
    <WixToolPath>$(LocalWixRoot)</WixToolPath>
    <WixTargetsPath Condition="Exists('$(WixToolPath)\Wix.targets')">$(WixToolPath)\Wix.targets</WixTargetsPath>
    <WixTasksPath Condition="Exists('$(WixToolPath)\wixtasks.dll')">wixtasks.dll</WixTasksPath>
```

heaths: If you look at WixToolPath now, it's absolute. And the condition and value should match. This works exactly as it should now without any guesswork.

jiasli: > And the condition and value should match.

This is a very good point. The fact that the `Condition` and the value don't match as they use different base path is indeed a little bit confusing.

build_scripts/windows/Product.wxs
```diff
@@ -54,14 +93,14 @@
     <!-- Show message to restart any terminals only if the PATH is changed -->
     <CustomAction Id="Set_WIXUI_EXITDIALOGOPTIONALTEXT" Property="WIXUI_EXITDIALOGOPTIONALTEXT" Value="Please close and reopen any active terminal window to use Azure CLI." />
     <InstallUISequence>
-      <Custom Action="Set_WIXUI_EXITDIALOGOPTIONALTEXT" After="CostFinalize">NOT Installed AND NOT WIX_UPGRADE_DETECTED</Custom>
+      <Custom Action="Set_WIXUI_EXITDIALOGOPTIONALTEXT" After="CostFinalize">NOT Installed AND NOT WIX_UPGRADE_DETECTED AND NOT WIX_X86_UPGRADE_DETECTED</Custom>
```
jiasli, line 96: I think `WIX_X86_UPGRADE_DETECTED` does change the `PATH` from `C:\Program Files (x86)\Microsoft SDKs\Azure\CLI2\wbin` to `C:\Program Files\Microsoft SDKs\Azure\CLI2\wbin` , so I guess we do need to show this message?

```suggestion
      <Custom Action="Set_WIXUI_EXITDIALOGOPTIONALTEXT" After="CostFinalize">NOT Installed AND NOT WIX_UPGRADE_DETECTED</Custom>
```

jiasli:
As a side note, as I tested, uninstalling the MSI does remove `C:\Program Files (x86)\Microsoft SDKs\Azure\CLI2\wbin` from system `PATH`. Fascinating!

https://github.com/Azure/azure-cli/blob/b1c7966867343c8dc26ff1bf1df69a452eda8e69/build_scripts/windows/Product.wxs#L89-L100

build_scripts/windows/Product.wxs
```diff
@@ -8,18 +8,57 @@
...
+  <?define RemoveAzureCLIFolderGuid = "6AF97701-6A7A-4292-95D6-ED34CD4C97C0" ?>
+  <?define RemoveCLIFolderGuid = "24DA394D-B03D-4623-A0D9-852DB51EF9F4" ?>
+  <?define UpgradeCode = "dff82af0-3f95-4ac9-8efd-948604fdb028" ?>
+
+  <?else ?>
+  <?error Unsupported platform "$(var.Platform)" ?>
+  <?endif ?>
+
   <Product Id="*"
            Name="$(var.ProductName)"
```
jiasli, line 34: Some programs attach `x86` and `x64` to the product name. Some don't:

image

1. Shall we do the same and change `$(var.ProductName)` too?
2. Shall we allow 32-bit and 64-bit MSIs to be installed side-by-side? I don't feel this is necessary as there can be only one active `az.cmd` on `PATH`.

build_scripts/windows/Product.wxs
```diff
@@ -8,18 +8,57 @@
...
+      <UpgradeVersion Property="WIX_DOWNGRADE_DETECTED" Minimum="$(var.ProductVersion)" IncludeMinimum="no" OnlyDetect="yes" />
+    </Upgrade>
+    <InstallExecuteSequence>
+      <RemoveExistingProducts After="InstallExecute" />
+    </InstallExecuteSequence>
+
+    <!-- New product architectures should upgrade the original x86 product - even of the same version. -->
+    <?if $(var.UpgradeCode) != "dff82af0-3f95-4ac9-8efd-948604fdb028" ?>
+    <Upgrade Id="dff82af0-3f95-4ac9-8efd-948604fdb028">
+      <UpgradeVersion Property="WIX_X86_UPGRADE_DETECTED" Maximum="$(var.ProductVersion)" IncludeMaximum="yes" MigrateFeatures="yes" />
```
jiasli, line 55: The document https://wixtoolset.org/docs/v3/xsd/wix/upgradeversion/ doesn't describe the default value of `IncludeMaximum`:

> Set to "yes" to make the range of versions detected include the value specified in Maximum.

So I assume its default value is `no`.

build_scripts/windows/Product.wxs
heaths: > Shall we do the same and change $(var.ProductName) too?

Sure I can do that.

> Shall we allow 32-bit and 64-bit MSIs to be installed side-by-side? I don't feel this is necessary as there can be only one active az.cmd on PATH.

I don't recommend it. 1) what purpose does it serve? These are EXEs where, unlike libraries e.g., DLLs, architecture doesn't typically matter. 32- and 64-bit code can call any architecture of an EXE. It would be a UX problem because,

1) Depending on the ordering of updates to `PATH`, customer will call the first or last architecture installed. It's not deterministic.
2) Every time there's an update, the ~15-20 minute upgrade of 1 MSI will now be doubled.

And all that for what reason? Again, it's an EXE (well, effectively - an executable, more specifically) so architecture doesn't matter. SxS just unnecessarily complicates things in this case.

build_scripts/windows/Product.wxs
heaths: Yes, but I added defaults here in some cases for easier maintenance. The change that prevented the x64 MSI from removing the x86 was the result of copy-paste and forgetting to change the attribute.

build_scripts/windows/Product.wxs
jiasli: Microsoft products usually use `(x86)` and `(x64)`. May be we can do it the Microsoft way? 😊

image

jiasli (approved):

jiasli:
The upgrade behavior is as below:

- If the architecture doesn't change (installing **x86 MSI on top of x86 MSI**, or installing **x64 MSI on top of x64 MSI**), only version upgrade is allowed; installing the same version or downgrade is not allowed.
- Installing **64-bit MSI on top of 32-bit MS**I will uninstall the 32-bit MSI automatically. Upgrade or installing the same version is allowed; downgrade is not allowed.
- Installing **32-bit MSI on top of 64-bit MSI** will keep both MSI's side-by-side. TODO: We will forbid this behavior in the future.

jiasli:
We have finalized related names to be

| Architecture   | x86                                   | x64                                      |
| -------------- | ------------------------------------- | ---------------------------------------- |
| Product Name   | Microsoft Azure CLI (32-bit)          | Microsoft Azure CLI (64-bit)             |
| MSI file name  | azure-cli-2.x.x.msi                   | azure-cli-2.x.x-x64.msi                  |
| MSI short link | https://aka.ms/installazurecliwindows | https://aka.ms/installazurecliwindowsX64 |

**The product name has been changed from `Microsoft Azure CLI` to `Microsoft Azure CLI (32-bit)` in order to better reflect the installed MSI's architecture and encourage users to migrate to the 64-bit MSI.**

The product name change should not be considered a breaking change, because the official document https://learn.microsoft.com/en-us/cli/azure/install-azure-cli-windows doesn't declare the product name, so the product name is not guaranteed to remain the same across versions.

If you have script relying on the product name `Microsoft Azure CLI`, please consider changing it accordingly.

aluty:
Documentation not yet updated with 32-bit and 64-bit information?
https://learn.microsoft.com/en-us/cli/azure/install-azure-cli-windows?tabs=azure-cli#install-or-update

bebound:
@aluty Yes, the PR is not merged yet: https://github.com/MicrosoftDocs/azure-docs-cli/pull/3759

jiasli:
Hi @heaths, I can give a little bit more context.

`az upgrade` hangs there and waits for the `msiexec.exe` to complete. It currently only works on 32-bit MSI. This issue doesn't happen if you directly install MSI and not through `az upgrade`.

Previously, if Python version is not upgraded during the installation, this Files in Use message will not be shown. However, if Python version is upgraded, this Files in Use message will be shown. We 

> 2\. are you sure you updated Python from 2.50 to 2.51?
The weird thing I mentioned in https://github.com/Azure/azure-cli/pull/26640#issuecomment-1661708400  contradicts the previous behavior and we are not sure why.



We didn't update Python from 2.50 to 2.51

jiasli:
Thanks @heaths for the comment.

> The file in use is Resource Manager

May I know which "Resource Manager" you are referring to? Is it a file?

How `az upgrade` (a `python.exe` process) works is that **it subprocesses `msiexec.exe` and waits for the `msiexec.exe` process to exit**:

https://github.com/Azure/azure-cli/blob/21266e02457263420b62fba1c529f69e28852414/src/azure-cli/azure/cli/command_modules/util/custom.py#L193

It currently only supports 32-bit Azure CLI and 32-bit MSI.

This issue doesn't happen if you directly install MSI and not through `az upgrade`, as there is no `python.exe` running.

Previously before this PR, we have observed:

- if Python version is not upgraded during the installation, this **Files in Use** window will not be shown
- if Python version is upgraded, e.g. [3.10.9 -> 3.10.11](https://www.python.org/downloads/windows/), this **Files in Use** window will be shown

> are you sure you updated Python from 2.50 to 2.51?

We did not upgrade Python from Azure CLI 2.50 to 2.51 and the `python.exe` are exactly the same 3.10.10 in those Azure CLI versions:

https://github.com/Azure/azure-cli/blob/aeb94ed12a46acdd14e7c4e2633908e149f4a1bc/build_scripts/windows/scripts/build.cmd#L28

The weird thing I mentioned in https://github.com/Azure/azure-cli/pull/26640#issuecomment-1661708400  contradicts the previous behavior and we are not sure why.

jiasli:
> Because python.exe is waiting for msiexec.exe to finish, this will happen.

Our previously observation is that if Python version is not upgraded during the installation, this **Files in Use** window will not be shown. That's why we are curious about it.

> Why wait for it? Let it run in basic UI mode and any errors will be surfaced to the user.

This is exactly what https://github.com/Azure/azure-cli/pull/26464 is doing. 

https://github.com/Azure/azure-cli/pull/14803 added the `az upgrade` command. It launches `msiexec.exe` to install the MSI from URL `https://aka.ms/installazurecliwindows`, and then waits for `msiexec.exe` to complete. 

We refined it in #19192 to download the MSI to a temporary folder under `TEMP` (env var) first, and then install that local MSI. For example:

```
> az upgrade
...
Updating Azure CLI with MSI from https://aka.ms/installazurecliwindows
Downloading MSI to C:\Users\xxx\AppData\Local\Temp\tmpr1onzz23\azure-cli-2.51.0.msi
```

Once `az upgrade` knows `msiexec.exe` succeeds, it'll

1. upgrade extensions: https://github.com/Azure/azure-cli/blob/21266e02457263420b62fba1c529f69e28852414/src/azure-cli/azure/cli/command_modules/util/custom.py#L172-L173

2. delete the local MSI: https://github.com/Azure/azure-cli/blob/21266e02457263420b62fba1c529f69e28852414/src/azure-cli/azure/cli/command_modules/util/custom.py#L200

If `az upgrade` kicks off `msiexec.exe` and exits, 

1. it will not have the chance to upgrade extensions. Users will have to run `az upgrade` again to trigger the upgrading extension logic. We can't install extensions first as some extensions may depend on newer versions of Azure CLI core.

2. it will not have the chance to delete the local MSI. Since the MSI is downloaded to a temporary folder under `TEMP`, a new `az upgrade` process won't know where the previous `az upgrade`'s MSI is saved (in the above example, `tmpr1onzz23`), so the MSI will be left there until the user manually runs [Disk Cleanup](https://support.microsoft.com/en-us/windows/disk-cleanup-in-windows-8a96ff42-5751-39ad-23d6-434b4d5b9a68).

heaths:
> Our previously observation is that if Python version is not upgraded during the installation, this Files in Use window will not be shown. That's why we are curious about it.

And that confirms what I said in my [comment above](https://github.com/Azure/azure-cli/pull/26640#issuecomment-1664263893). If the file (component) doesn't need to be upgraded, Restart Manager isn't used for that file because it won't be touched. Once it's upgraded, though, this functionality is working exactly as designed. So you can either make sure (as much as you can) _python.exe_ isn't running (don't wait), or try passing `REBOOT=ReallySuppress`. The latter may or may not work. Alternatively, find a way to create the process for _python.exe_ that opens it with `FILE_SHARE_DELETE` or, if possible, create a temp copy of it e.g., __python.exe_ and exec that.

jiasli:
> make sure (as much as you can) _python.exe_ isn't running (don't wait)

We will do this in https://github.com/Azure/azure-cli/pull/26464.

> find a way to create the process for _python.exe_ that opens it with `FILE_SHARE_DELETE`

There is no fancy logic currently to create the `python.exe` process - `python.exe` is launched by the `az.cmd` entry script:

https://github.com/Azure/azure-cli/blob/9bc9f36ca7c07284c6c4386a544b932fab952e63/build_scripts/windows/scripts/az.cmd#L8

> create a temp copy of it e.g., __python.exe_ and exec that

In order to know whether `_python.exe` succeeds, the original `python.exe` needs to be alive, making it unable to be updated. I provided more detail in https://github.com/Azure/azure-cli/pull/26464#discussion_r1198628626c

heaths:
> In order to know whether _python.exe succeeds, the original python.exe needs to be alive, making it unable to be updated. I provided more detail in https://github.com/Azure/azure-cli/pull/26464#discussion_r1198628626.

Nothing is stopping you from copying the _python.exe_ process to another file and executing that in the same directory, so that all library and package references are resolved correctly. _python.exe_ would be replaced while __python.exe_ (e.g.) continues to run and you can still wait for it. This is a common workaround for cases like this.

jiasli:
> _python.exe_ would be replaced while __python.exe_ (e.g.) continues to run and you can still wait for it.

As far as I know, this is not possible on Windows, because Windows doesn't support replacing a process like [`exec*` on Linux](https://man7.org/linux/man-pages/man3/execlp.3p.html). [`os.exec*` on Windows](https://learn.microsoft.com/en-us/cpp/c-runtime-library/exec-wexec-functions?view=msvc-170) internally calls [`CreateProcess`](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw) to create a new process and exits immediately. The new process will be created as a background process and the control will be given back immediately to the terminal. There's no way of knowing if the new process succeeds or not.

This is of course not what we want - we want the control to be given back to the terminal after `_python.exe` finishes, and we also need to know if `_python.exe` succeeds.

Python exposes `exec*` via [`os.exec*`](https://docs.python.org/3/library/os.html#os.execl), but these functions behave differently on Windows and Linux. This issue has reported multiple times to the Python repo more than 10 years ago, such as https://github.com/python/cpython/issues/63323.

---

Another workaround is to modify the `az.cmd` entry script to install the MSI, but the Return on investment (ROI) seems to be very low (https://github.com/Azure/azure-cli/pull/26464#discussion_r1198634032).

heaths:
Apps using the console subsystem do not return control immediately - only windows subsystem apps. _python.exe_ is compiled to use the console subsystem. Even if it used the windows subsystem, you can use `start /wait python.exe ...` to wait for it and get its `%ERRORLEVEL%` after running. You could always have your batch script copy it and wait on `_python.exe` as well.

The main point is that: if you're going to hold `python.exe` in use, unless you pass `REBOOT=ReallySuppress` to the command line when calling `msiexec.exe` (and I'm not 100% sure that will work; should, but it's been a long time since I've tested that) *and* `python.exe` is getting upgrade, this is exactly the behavior you should expect. This has been in place since Windows Vista or so. Maybe XP. Restart Manager and its use in MSI is not new.

jiasli:
Azure DevOps runner image (Windows) has switched to 64-bit MSI (https://github.com/actions/runner-images/pull/8096). Runners will be updated gradually after 2023-8-25.

PR #26757, opened 2023-06-26
[SQL VM] az sql vm update: Add configuration options for SQL Assessment pre-requisites using AMA agent. Migration from MMA to AMA

ai-abhinav:
az sql vm update --enable-assessment true -g -n --workspace-name <LA workspace's name> --workspace-rg <LA workspace's resource group name> --agent-rg --workspace-sub

--workspace-sub is a new optional param. If specified this workspace is used, else the current workspace is picked as default
--agent-rg is a required parameter. It specifies the resource group in which the dependent resources get installed.
Note:* Considering making --agent-rg optional with a warning that agent resources will be deployed in the specified Resource Group but can be modified using --agent-rg.
--agent-rg is helpful because if respecified for another VM with the same LA workspace - the DCE, DCR get re-used and do not have to be re-created each time.

Added a default value to prevent breaking change.

Todo:Remove prints and add warnings for user messages concerning change.
Fix test errors: tests use dummy subscription 0000-0000* and cannot connect to azure resources through REST API.


yonzhan:

PR #25116, opened 2023-01-12
{AKS} az aks nodepool: Add doc line on how to delete node taints

elliott-king:
**Related command**
az aks nodepool update

**Description**
I added a note to the `--node-taints` option in your docs. 

It is not immediately obvious how to delete node taints. [Kubernetes reference](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/) tells you to add a hyphen at the end, which does not work on the Azure CLI. This behavior is also different from the `labels` argument.

**Testing Guide**
Unneeded, this is a docs change.

**History Notes**

---

This checklist is used to make sure that common guidelines for a pull request are followed.

- [X] The PR title and description has followed the guideline in [Submitting Pull Requests](https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules#submitting-pull-requests).

- [X] I adhere to the [Command Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md).

- [X] I adhere to the [Error Handling Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/error_handling_guidelines.md).

PR #26767, opened 2023-06-27
[Service Fabric] `az sf managed-node-type create`: creates nodetype with 2022-Datacenter vmSku as default

mwesigwaguma:
**Related command**

**Description**
Updating vmSkus to use 2022-Datacenter so that tests are tls compliant for ServiceFabric module

**Testing Guide**
see src\azure-cli\azure\cli\command_modules\help.py for examples and descriptions

**History Notes**
[Service Fabric] `az sf managed-node-type create`: creates nodetype with 2022-Datacenter vmSku as default

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/servicefabric/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/servicefabric/_params.py
@@ -311,7 +311,7 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('vm_size', help='The size of virtual machines in the pool. All virtual machines in a pool are the same size.', default='Standard_D2')
         c.argument('vm_image_publisher', help='The publisher of the Azure Virtual Machines Marketplace image.', default='MicrosoftWindowsServer')
         c.argument('vm_image_offer', help='The offer type of the Azure Virtual Machines Marketplace image.', default='WindowsServer')
-        c.argument('vm_image_sku', help='The SKU of the Azure Virtual Machines Marketplace image.', default='2019-Datacenter')
+        c.argument('vm_image_sku', help='The SKU of the Azure Virtual Machines Marketplace image.', default='2022-Datacenter')
         c.argument('vm_image_version', help='The version of the Azure Virtual Machines Marketplace image. ', default='latest')
         c.argument('capacity', arg_type=capacity)
         c.argument('placement_property', arg_type=placement_property)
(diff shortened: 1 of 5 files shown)


yonzhan:
ServiceFabric

a-santamaria (approved):

src/azure-cli/azure/cli/command_modules/servicefabric/_params.py
```diff
@@ -311,7 +311,7 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('vm_size', help='The size of virtual machines in the pool. All virtual machines in a pool are the same size.', default='Standard_D2')
         c.argument('vm_image_publisher', help='The publisher of the Azure Virtual Machines Marketplace image.', default='MicrosoftWindowsServer')
         c.argument('vm_image_offer', help='The offer type of the Azure Virtual Machines Marketplace image.', default='WindowsServer')
-        c.argument('vm_image_sku', help='The SKU of the Azure Virtual Machines Marketplace image.', default='2019-Datacenter')
+        c.argument('vm_image_sku', help='The SKU of the Azure Virtual Machines Marketplace image.', default='2022-Datacenter')
```
zhoxing-ms, line 314: May I ask will this default value change cause a breaking change to customers?

mwesigwaguma: The parameter will only apply on creation so this should not be a breaking change to customers.

zhoxing-ms: Actually, what I want to confirm is whether it may cause a breaking change to the workflow of the customer's automation scripts. If there is a regression between `2022-Datacenter` and `2019-Datacenter`, then there is a risk of breaking change

mwesigwaguma: @a-santamaria can you help address this question.

a-santamaria: @zhagao-ms there shouldn't be issues with that. We are updating it since azure now recommends 2022 due to its security. In the future probably we would have to change it whenever there is a new version. What is the recommendation for this? is it better to remove the default?

zhoxing-ms: If you currently have evaluated that the change of default value has little impact on customers, then we can directly merge this PR.

mwesigwaguma: You can go ahead and merge

jiasli:
This PR targets the wrong branch, similar to https://github.com/Azure/azure-cli/pull/25116#issuecomment-1420162129.

PR #16857, opened 2021-02-08
[ARM] az deployment group/sub/mg/tenant validate/create/what-if: Support deployment with Bicep files

shenglol:
**Description**

The PR integrates Bicep CLI so that the ARM template deployment commands can deploy Bicep files.

**Testing Guide**

```
az deployment group create -g <rg> -f storage_account.bicep
```

**History Notes**

[ARM] az deployment group/sub/mg/tenant validate/create/what-if: Add support for Bicep files
[ARM] az bicep install: New command for installing Bicep CLI
[ARM] az bicep upgrade: New command for upgrading Bicep CLI
[ARM] az bicep build: New command for building Bicep files
[ARM] az bicep version: New command for showing the current installed version of Bicep CLI
[ARM] az bicep list-versions: New command for showing the available Bicep CLI versions

---

This checklist is used to make sure that common guidelines for a pull request are followed.

- [ ] The PR title and description has followed the guideline in [Submitting Pull Requests](https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules#submitting-pull-requests).

- [ ] I adhere to the [Command Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md).

- [ ] I adhere to the [Error Handling Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/error_handling_guidelines.md).

PR #26797, opened 2023-06-30
{ARM} Fix race condition in bicep unit tests

shenglol:
**Related command**
`az bicep *`

**Description**
~~This main purpose of the PR is to reduce Azure CLI’s packaging maintenance overheads based on @jiasli's feedback https://github.com/Azure/azure-cli/pull/16857#discussion_r1204981875.~~

The PR also fixes race conditions in `test_resource_bicep.py` by having each test uses their own `DummyCli` instance with `random_config_dir` set to `True`.

**Testing Guide**
The existing tests should be sufficient the cover the changes.

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/resource/_bicep.py
+++ b/src/azure-cli/azure/cli/command_modules/resource/_bicep.py
@@ -16,7 +16,8 @@ from contextlib import suppress
 from datetime import datetime, timedelta
 
 import requests
-import semver
+
+from packaging.version import parse, VERSION_PATTERN
 
 from urllib.request import urlopen
 from knack.log import get_logger
@@ -30,8 +31,7 @@ from azure.cli.core.azclierror import (
 )
 from azure.cli.core.util import should_disable_connection_verify
 
-# See: https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
-_semver_pattern = r"(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?"  # pylint: disable=line-too-long
+_version_pattern = re.compile(VERSION_PATTERN, re.VERBOSE | re.IGNORECASE)
 
 # See: https://docs.microsoft.com/azure/azure-resource-manager/templates/template-syntax#template-format
 _template_schema_pattern = r"https?://schema\.management\.azure\.com/schemas/[0-9a-zA-Z-]+/(?P<templateType>[a-zA-Z]+)Template\.json#?"  # pylint: disable=line-too-long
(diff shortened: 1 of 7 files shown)


yonzhan:
ARM

shenglol:
@jiasli @zhoxing-ms Could you take a look?

shenglol:
Well, looks like azure-cli-extensions now has a dependency on `semver` (see https://github.com/Azure/azure-cli-extensions/blob/0b2cbaab7ecd11a8cd99eb5d455045abb22ee9c9/src/fleet/azext_fleet/_validators.py#L5), which causes the failure in CI. Should I add the `semver` dependency back in the requirements files?

zhoxing-ms:
> Well, looks like azure-cli-extensions now has a dependency on semver (see https://github.com/Azure/azure-cli-extensions/blob/0b2cbaab7ecd11a8cd99eb5d455045abb22ee9c9/src/fleet/azext_fleet/_validators.py#L5), which causes the failure in CI. Should I add the semver dependency back in the requirements files?

Okay, this can avoid causing breaking changes to other dependent places

shenglol:
> > Well, looks like azure-cli-extensions now has a dependency on semver (see https://github.com/Azure/azure-cli-extensions/blob/0b2cbaab7ecd11a8cd99eb5d455045abb22ee9c9/src/fleet/azext_fleet/_validators.py#L5), which causes the failure in CI. Should I add the semver dependency back in the requirements files?
> 
> Okay, this can avoid causing breaking changes to other dependent places

Since we cannot remove the `semver` dependency, I restored the _bicep.py to use `semver` instead of `package.versioning`. In the future, if we start to publish pre-releases / private versions for Bicep, the version number may not be compliant with PEP 440.

zhoxing-ms:
@shenglol Actually, I think we can skip deleting B from A in order not to affect other modules/extensions. However, to ensure the consistency in version management, it is still recommended that you use `package.versioning` instead of `semver`

shenglol:
> @shenglol Shenglong Li FTE Actually, I think we can skip removing `semver` from `setup.py` and `requirements.py3.xxx` in order not to affect other modules/extensions. However, to ensure the consistency in version management, it is still recommended that you use `package.versioning` instead of `semver`

To clarify, Bicep CLI is technically using semantic versioning. The version is set by [Nerdbank.GitVersioning](https://github.com/dotnet/Nerdbank.GitVersioning), which produces semver-compatible versions. Currently, for public releases, the version is always in the format of MAJOR.MINOR.PATCH and is also compatible with PEP 440. However, in the future, we might want to publish private or pre-release versions that may contain suffixes (e.g., commit hashes) that are not necessarily compatible with PEP 440.

zhoxing-ms (approved):

zhoxing-ms:
@jiasli Could you please help review this PR?

bebound (approved):

src/azure-cli/azure/cli/command_modules/resource/_bicep.py
```diff
@@ -302,9 +296,9 @@ def _get_bicep_installation_path(system):
     raise ValidationError(f'The platform "{system}" is not supported.')
 
 
-def _extract_semver(text):
+def _extract_version(text):
     semver_match = re.search(_semver_pattern, text)
-    return semver_match.group(0) if semver_match else None
+    return semver.VersionInfo.parse(semver_match.group(0)) if semver_match else None
```
jiasli, line 301: If `semver` is indeed required, we should consider revive https://github.com/Azure/azure-cli/pull/26523.

src/azure-cli/azure/cli/command_modules/resource/tests/latest/test_resource_bicep.py
```diff
@@ -19,20 +20,21 @@
 from azure.cli.core.azclierror import InvalidTemplateError
 from azure.cli.core.mock import DummyCli
 
-cli_ctx = DummyCli()
-
 
 class TestBicep(unittest.TestCase):
+    def setUp(self):
+        self.cli_ctx = DummyCli(random_config_dir=True)
```
jiasli, line 26: This contradicts the convention used in https://github.com/Azure/azure-cli/pull/25689 that `random_config_dir=True` is set in `super().__init__` invocation:

https://github.com/Azure/azure-cli/blob/b223dd9d4379d5c633a9f4dba4941ae791531126/src/azure-cli/azure/cli/command_modules/config/tests/latest/test_config.py#L16-L17

bebound: Its superclass is `unittest.TestCase`, which does not has `random_config_dir` param.

PR #26065, opened 2023-04-07
{Telemetry} Shorten default push interval from 24h to 6h

evelyn-ys:
**Description**

We decide to drop CLI telemetry local cache because of
- CLI telemetry local cache strategy has introduced a lot of troubles
- Learning from PSH and VSCode, we know that they don't have any local cache
- Carl from devdiv data team has confirmed that the application insight's quota is highly customizable.

We don't want to drop local cache by one night so things need to change step by step:
1. Shorten the cache push interval from 24 hours to 6 hours
2. Wait for few CLI versions until the usage of 6hours interval grows up to 50% and see if there's any risks
3. Shorten the cache push interval from 6 hours to 1 hour
4. Repeat step 2
5. Totally remove local cache

This PR is for step 1

---

This checklist is used to make sure that common guidelines for a pull request are followed.

- [ ] The PR title and description has followed the guideline in [Submitting Pull Requests](https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules#submitting-pull-requests).

- [ ] I adhere to the [Command Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md).

- [ ] I adhere to the [Error Handling Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/error_handling_guidelines.md).

PR #24304, opened 2022-10-20
{Core} Log allow_broker configuration in telemetry

evelyn-ys:
**Description**

Fix #23956

This PR adds `allow_broker` configuration in telemetry when it's loaded and used for calling `msal`

**Testing Guide**

Scenario 1: Test telemetry logging when no `allow_broker` is configured
- Run commands:
  `az login`
  `az group show -g {resource_group_name}`

- Open `$home_dir/.azure/telemetry/cache` file and check the record data, you will find `"Context.Default.AzureCLI.AllowBroker": "None"` for both commands

Scenario 2: Test telemetry logging when `allow_broker` is configured as `true`
- Config with `az config set core.allow_broker=true`
- Run commands:
  `az version`
  `az login`
  `az group show -g {resource_group_name}`

- Open `$home_dir/.azure/telemetry/cache` file and check the record data, you will find
  `"Context.Default.AzureCLI.AllowBroker": "None"` for the first command, as `az version` doesn't need auth, no msal call stack
  `"Context.Default.AzureCLI.AllowBroker": "True"` for the other two commands, this configuration is used for calling msal auth

Scenario 3: Test telemetry logging when `allow_broker` is configured as `false`
- Config with `az config set core.allow_broker=false`
- Run commands:
  `az version`
  `az login`
  `az group show -g {resource_group_name}`

- Open `$home_dir/.azure/telemetry/cache` file and check the record data, you will find
  `"Context.Default.AzureCLI.AllowBroker": "None"` for the first command, as `az version` doesn't need auth, no msal call stack
  `"Context.Default.AzureCLI.AllowBroker": "False"` for the other two commands, this configuration is used for calling msal auth

---

This checklist is used to make sure that common guidelines for a pull request are followed.

- [ ] The PR title and description has followed the guideline in [Submitting Pull Requests](https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules#submitting-pull-requests).

- [ ] I adhere to the [Command Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md).

- [ ] I adhere to the [Error Handling Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/error_handling_guidelines.md).

PR #26854, opened 2023-07-10
[Core] Drop client telemetry cache strategy

evelyn-ys:
**Description**

We have reduce client telemetry push interval from 24h to 6h in #26065, from 6h to 1h in #26509

Now it's time to fully drop client telemetry cache strategy. All telemetry records will be uploaded immediately.

**Testing Guide**

Run any cli command and check if there's `$HOME_DIR/.azure/telemetry/cache` file anymore

---

### DIFF
--- a/src/azure-cli-telemetry/azure/cli/telemetry/util.py
+++ b/src/azure-cli-telemetry/azure/cli/telemetry/util.py
@@ -4,41 +4,8 @@
 # --------------------------------------------------------------------------------------------
 
 import os
-import stat
 import logging
 import logging.handlers
-from datetime import datetime
-
-from azure.cli.telemetry.const import TELEMETRY_NOTE_NAME, MANDATORY_WAIT_PERIOD
-
-
-def should_upload(config_dir):
-    """Returns True if it is the right moment to add telemetry.
-    Before the client request a telemetry add process, it can invoke this method to test if it is the right moment.
-    The conditions are:
-        1. The telemetry.txt file doesn't exist; OR
-        2. The telemetry.txt file is a regular file AND there have been enough time passed since last add.
-    """
-    logger = logging.getLogger('telemetry.check')
-
-    telemetry_note_path = os.path.join(config_dir, TELEMETRY_NOTE_NAME)
-    if not os.path.exists(telemetry_note_path):
-        logger.info('Positive: The %s does not exist.', telemetry_note_path)
-        return True
-
-    file_stat = os.stat(telemetry_note_path)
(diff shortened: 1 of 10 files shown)


yonzhan:
Core

calvinhzy (approved):

jiasli:
The removal of telemetry cache makes it difficult to locally verify the sent telemetry record, making the testing guide from https://github.com/Azure/azure-cli/pull/24304 not applicable anymore.

jiasli:
Sending client telemetry in each command slows down the command execution.

Without sending telemetry, on Windows, `az version` takes ~600ms to finish:

```powershell
> az config set core.collect_telemetry=false
> (Measure-Command {az version}).TotalMilliseconds
629.2332
```

But now it becomes ~800ms, with an increase of 33%:

```powershell
> az config set core.collect_telemetry=true
> (Measure-Command {az version}).TotalMilliseconds
822.9069
```

So the sending client telemetry network request takes ~200ms.

The execution time increase on Linux is much more outstanding - almost doubles the execution time (100% increase), as command itself runs faster on Linux, giving more weight to network requests.

```sh
$ az config set core.collect_telemetry=false
$ time az version > /dev/null
real    0m0.230s

$ az config set core.collect_telemetry=true
$ time az version > /dev/null
real    0m0.403s
```


Things get worse if the machine has a bad network connection to the telemetry's Application Insights instance.

evelyn-ys:
But CLI telemetry is sent in separate process:
https://github.com/Azure/azure-cli/blob/07e1b17d4c0f9e4a4cd76fa9ff3838644be5e419/src/azure-cli-telemetry/azure/cli/telemetry/__init__.py#L45

It should not lower the CLI command's execution time

src/azure-cli-telemetry/azure/cli/telemetry/util.py
```diff
@@ -4,41 +4,8 @@
...
-        logger.info('Positive: The %s does not exist.', telemetry_note_path)
-        return True
-
-    file_stat = os.stat(telemetry_note_path)
-    if not stat.S_ISREG(file_stat.st_mode):
-        logger.warning('Negative: The %s is not a regular file.', telemetry_note_path)
-        return False
-
-    modify_time = datetime.fromtimestamp(file_stat.st_mtime)
-    if datetime.now() - modify_time < MANDATORY_WAIT_PERIOD:
```
jiasli, line 35: On a second thought, would it be safer to keep the existing code and let `MANDATORY_WAIT_PERIOD = 0` in order to make the wait period configurable?

PR #26879, opened 2023-07-13
[AppConfig] `az appconfig snapshot list`: Use enums for status parameter

albertofori:
**Related command**

`az appconfig snapshot list`

**Description**

This PR updates the `status` parameter type to use enums since only a defined set of options is valid.

**Testing Guide**

**History Notes**

[AppConfig] `az appconfig snapshot list`: Use enums for status parameter

---


yonzhan:
AppConfig

src/azure-cli/azure/cli/command_modules/appconfig/_snapshot_custom_client.py
```diff
@@ -407,9 +407,11 @@ def list_snapshots(
 
         _headers = kwargs.pop("headers", {}) or {}
 
+        status_filter = ",".join(status) if status else None
+
         initial_request = build_list_snapshots_request(
             name_filter=name,
-            status_filter=status,
+            status_filter=status_filter,
```
avanigupta, line 414: Does passing `None` have the same effect as passing `*` as status filter?

avanigupta (approved):

src/azure-cli/azure/cli/command_modules/appconfig/_snapshot_custom_client.py
avanigupta: I think it may cause confusion if we support * sign here. In all other filters when we support *, we also support prefix filtering. But that's not the case with status filter. So I think it's better to omit the status instead of providing * as an option. We could add more details to the help message though - something like:
`Filter snapshots by their status. If no status specified, return all snapshots by default. `

albertofori: Sounds good! This has been updated, thanks!

avanigupta (approved):

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -98,6 +98,12 @@ def load_arguments(self, _):
         nargs='+',
         help='Space-separated list of escaped JSON objects that represent the key and label filters used to build an App Configuration snapshot.'
     )
+    snapshot_status_arg_type = CLIArgumentType(
+        options_list=['--status'],
+        arg_type=get_enum_type(['archived', 'failed', 'provisioning', 'ready']),
```
zhoxing-ms, line 103: Could we get these enumeration values from Python SDK?

albertofori: @zhoxing-ms I see these status values are similar to what are used in the PollingMethod classes, however, these are also directly defined as strings.

src/azure-cli/azure/cli/command_modules/appconfig/snapshot.py
```diff
@@ -95,7 +95,7 @@ def show_snapshot(cmd,
 
 def list_snapshots(cmd,
                    snapshot_name=SearchFilterOptions.ANY_KEY,
-                   status=SearchFilterOptions.ANY_KEY,
```
zhoxing-ms, line 98: May I ask modifying this default value may cause a breaking change for customers?

albertofori: @zhoxing-ms  No please. The updated default value returns the same result.

zhoxing-ms (approved):

Issue #24127, opened 2022-10-07
CLI - Add Standard as SecurityType Option for backward compatibility


AjKundnani:
### Related command

- [az vm create --name 'vmname' --resource-group 'vm-rgname' --security-type 'Standard'](https://docs.microsoft.com/en-us/cli/azure/vm?view=azure-cli-latest#az-vm-create)
- [az vmss create --name 'vmssname' --resource-group 'rgname' --security-type 'Standard'](https://docs.microsoft.com/en-us/cli/azure/vmss?view=azure-cli-latest#az-vmss-create)
- [az disk create --name 'diskname' -g 'diskrgname' --security-type 'Standard'](https://docs.microsoft.com/en-us/cli/azure/disk?view=azure-cli-latest#az-disk-create)
- [az sig image-definition create --features SecurityType=Standard](https://learn.microsoft.com/en-us/cli/azure/sig/image-definition?view=azure-cli-latest#az-sig-image-definition-create)

### Resource Provider

Microsoft.Compute/virtualMachines

### Overview
For request #23450 and #23614, CLI will be switching defaults to [Trusted Launch VMs](https://docs.microsoft.com/en-us/azure/virtual-machines/trusted-launch). This request is to provide backward compatibility after defaults change such that:
- CLI experience aligns with Azure Portal.
- Allow users to create Non-Trusted Launch VMs for any corner scenarios after defaults are changed to TLVM.

This will allow users to deploy Non-Trusted Launch VMs similar to Azure Portal Experience:
![image](https://user-images.githubusercontent.com/18166395/194509447-7c3a68e1-9aa9-47ec-9067-72d7d0ccd690.png)

>**NOTE**: Default value of Security type in Azure Portal is planned to be "Trusted Launch" in December 2022.

### Change Request
Add **Standard** as supported ENUM value for following command parameters:

#### az vm create & az vmss create
- [az vm create --name 'vmname' --resource-group 'vm-rgname' **--security-type 'Standard'**](https://docs.microsoft.com/en-us/cli/azure/vm?view=azure-cli-latest#az-vm-create)
- [az vmss create --name 'vmssname' --resource-group 'rgname' **--security-type 'Standard'**](https://docs.microsoft.com/en-us/cli/azure/vmss?view=azure-cli-latest#az-vmss-create)

SecurityType Supported Values | Description
-|-
[...]

PR #26892, opened 2023-07-14
[Compute] Add Standard as SecurityType Option for backward compatibility

ReaNAiveD:
**Related command**
- `az vm create`
- `az vmss create`
- `az disk create`
- `az sig image-definition create`

**Description**
Provide `Standard` as SecurityType Option for backward compatibility. The defaults will be changed to Trusted Launch VMs in CLI in the future.
#24127 

**Testing Guide**

- az vm create --name 'vmname' --resource-group 'rgname' --security-type 'Standard'
- az vmss create --name 'vmssname' --resource-group 'rgname' --security-type 'Standard'
- az disk create --name 'diskname' -g 'rgname' --security-type 'Standard'
- az sig image-definition create --features SecurityType=Standard

**History Notes**

---


yonzhan:
Compute

ReaNAiveD:
Hi @zhoxing-ms , should I add such warning in these commands execution?

Below is the prompt that appears when creating vm/vmss without `--public-ip-sku`.

image

zhoxing-ms:
>  should I add such warning in these commands execution?

This is a good question. I think we may need to modify the existing warning log to inform the customer that they can add parameter `--security-type` to keep their current behavior after the implementation of breaking change.

src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
```diff
@@ -626,7 +626,8 @@ def _build_storage_profile():
     if encryption_at_host is not None:
         vm_properties['securityProfile']['encryptionAtHost'] = encryption_at_host
 
-    if security_type is not None:
+    # Provide backward compatibility. The defaults will be changed to Trusted Launch VMs in CLI in the future.
+    if security_type is not None and security_type != "Standard":
```
zhoxing-ms, line 630:
```suggestion
    # The `Standard` is used for backward compatibility. The defaults will be changed to Trusted Launch VMs in CLI in the future.
    if security_type is not None and security_type != "Standard":
```

AjKundnani:
> > should I add such warning in these commands execution?
> 
> This is a good question. I think we may need to modify the existing warning log _(such as `Ignite (November) 2023 onwards "az vm/vmss create" command will deploy Gen2-Trusted Launch VM by default. To know more about the default change and Trusted Launch, please visit https://aka.ms/TLaD`)_ to inform the customer that they can add parameter `--security-type` to keep their current behavior after the implementation of breaking changes. @AjKundnani What do you think of it?

@zhoxing-ms @ReaNAiveD - I am running CLI version 2.50.0 and I see the warning message already in place. Is the ask to retain the warning message? Then yes.

image

Or if the ask is whether warning message should be disabled if end user runs command with `SecurityType = Standard`  then the message should be different to encourage end users for using Trusted Launch.
Draft below:

`Consider upgrading security for your workloads using Azure Trusted Launch VMs. To know more about Trusted Launch, please visit https://docs.microsoft.com/en-us/azure/virtual-machines/trusted-launch`

zhoxing-ms:
@AjKundnani If the customers temporarily do not want to enable Trusted Launch after aware it, because they want to avoid those breaking changes for their bussiness. Do we need to inform customers in the concept document *(such as `https://aka.ms/TLaD`)* about they can use `--security-type 'Standard'` to keep their current behavior?

src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
```diff
@@ -626,8 +629,14 @@ def _build_storage_profile():
...
 
-    if security_type is not None:
+    # The `Standard` is used for backward compatibility to allow customers to keep their current behavior
+    # after changing the default values to Trusted Launch VMs in the future.
+    if security_type is not None and security_type != "Standard":
         vm_properties['securityProfile']['securityType'] = security_type
+    else:
+        logger.warning('Consider upgrading security for your workloads using Azure Trusted Launch VMs. '
+                       'To know more about Trusted Launch, please visit '
+                       'https://learn.microsoft.com/en-us/azure/virtual-machines/trusted-launch.')
```
zhoxing-ms, line 639:
```suggestion
                       'https://aka.ms/TLaD.')
```
We can use the short link uniformly to make notification message more concise. What do you think of it?

zhoxing-ms, line 639: Should this warning log only appear when `security_type == "Standard"`?

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -4615,12 +4621,24 @@ def create_gallery_image(cmd, resource_group_name, gallery_name, gallery_image_n
...
+                    if value == 'Standard':
+                        continue
+                    standard_security = False
                 feature_list.append(GalleryImageFeature(name=key, value=value))
             except ValueError:
                 raise CLIError('usage error: --features KEY=VALUE [KEY=VALUE ...]')
+        if standard_security:
+            logger.warning('Consider upgrading security for your workloads using Azure Trusted Launch VMs. '
+                           'To know more about Trusted Launch, please visit '
+                           'https://learn.microsoft.com/en-us/azure/virtual-machines/trusted-launch.')
```
zhoxing-ms, line 4641:
```suggestion
        for item in features.split():
            try:
                key, value = item.split('=', 1)
                # create Non-Trusted Launch VM Image
                # The `Standard` is used for backward compatibility to allow customers to keep their current behavior
                # after changing the default values to Trusted Launch VMs in the future.
                if key == 'SecurityType' and value == 'Standard':
                        logger.warning('xxx')
                        continue

                feature_list.append(GalleryImageFeature(name=key, value=value))
            except ValueError:
                raise CLIError('usage error: --features KEY=VALUE [KEY=VALUE ...]')
```

src/azure-cli/azure/cli/command_modules/vm/_template_builder.py
AjKundnani: @zhoxing-ms - The short link will be moved to Trusted Launch as default blog post once it's published which will not be relevant to end users who are choosing to not use Trusted Launch configuration. Hence pointing to learn docs to know more about benefits of TL.

ReaNAiveD: > Should this warning log only appear when `security_type == "Standard"`?

Hi @AjKundnani , Should this warning log only appears when the user specifies `security_type` as `Standard` explicitly , and replaces the original message `Ignite (November) 2023 onwards "az vm/vmss create" command will deploy Gen2-Trusted Launch VM by default. To know more about the default change and Trusted Launch, please visit https://aka.ms/TLaD` ? Like below,
image
Or this warning log will always appear if `security_type` is not set or set to `Standard`? Like below.
image

AjKundnani: > > Should this warning log only appear when `security_type == "Standard"`?
> 
> Hi @AjKundnani , Should this warning log only appears when the user specifies `security_type` as `Standard` explicitly , and replaces the original message `Ignite (November) 2023 onwards "az vm/vmss create" command will deploy Gen2-Trusted Launch VM by default. To know more about the default change and Trusted Launch, please visit https://aka.ms/TLaD` ? Like below, image Or this warning log will always appear if `security_type` is not set or set to `Standard`? Like below. image

@ReaNAiveD - Option 1 - When end user specifies `security_type` as `standard` explicitly. Replace the original message.

AjKundnani:
> @AjKundnani If the customers temporarily do not want to enable Trusted Launch after aware it, because they want to avoid those breaking changes for their bussiness. Do we need to inform customers in the concept document _(such as `https://aka.ms/TLaD`)_ about they can use `--security-type 'Standard'` to keep their current behavior?

@zhoxing-ms - Yes in the blog post / concept document we'll publish this guidance. Kindly refer to blog post draft for more details.

src/azure-cli/azure/cli/command_modules/vm/_vm_utils.py
```diff
@@ -558,12 +558,21 @@ def trusted_launch_warning_log(namespace, generation_version, features):
                   'Launch VM by default. To know more about the default change and Trusted Launch, ' \
                   'please visit https://aka.ms/TLaD'
 
+    upgrade_hint = 'Consider upgrading security for your workloads using Azure Trusted Launch VMs. ' \
+                   'To know more about Trusted Launch, please visit ' \
+                   'https://learn.microsoft.com/en-us/azure/virtual-machines/trusted-launch.'
```
zhoxing-ms, line 563: Could we consider placing these identical warning logs into a global static constant? This can avoid duplicate definitions

ReaNAiveD: > Could we consider placing these identical warning logs into a global static constant _(such as a new file `_constants.py`)_? This can avoid duplicate definitions

That makes sense👍

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -4615,6 +4619,13 @@ def create_gallery_image(cmd, resource_group_name, gallery_name, gallery_image_n
         for item in features.split():
             try:
                 key, value = item.split('=', 1)
+                # create Non-Trusted Launch VM Image
+                # The `Standard` is used for backward compatibility to allow customers to keep their current behavior
+                # after changing the default values to Trusted Launch VMs in the future.
+                if key == 'SecurityType' and value == 'Standard':
```
zhoxing-ms, line 4625:
```suggestion
                if key == 'SecurityType' and value == COMPATIBLE_SECURITY_TYPE_VALUE:
```

zhoxing-ms (approved):

PR #26928, opened 2023-07-18
[AppConfig] `az appconfig kv delete/set/set-keyvault`: Add key validations for null or empty space key parameter

albertofori:
**Related command**

`az appconfig kv set`
`az appconfig kv set-keyvault`
`az appconfig kv delete`

**Description**

Adds validations to ensure that key parameters and filters that are empty or whitespace are rejected for kv `set`, `set-keyvault` and `delete` commands.

**Testing Guide**

**History Notes**

[AppConfig] `az appconfig kv delete/set/set-keyvault`: Add key validations for null or empty space keys

---


yonzhan:
AppConfig

src/azure-cli/azure/cli/command_modules/appconfig/_params.py
```diff
@@ -250,7 +251,7 @@ def load_arguments(self, _):
         c.argument('secret_identifier', validator=validate_secret_identifier, help="ID of the Key Vault object. Can be found using 'az keyvault {collection} show' command, where collection is key, secret or certificate. To set reference to the latest version of your secret, remove version information from secret identifier.")
 
     with self.argument_context('appconfig kv delete') as c:
-        c.argument('key', help='Support star sign as filters, for instance * means all key and abc* means keys with abc as prefix.')
+        c.argument('key', validator=validate_non_empty_key, help='Support star sign as filters, for instance * means all key and abc* means keys with abc as prefix.')
```
avanigupta, line 254: Why not use `validate_key` directly? The character restrictions apply for delete command too.

albertofori: You are right. The reason for separating the two was that the delete parameter could also be a filter. But specifying any of the restricted characters without any wildcard should return nothing from the backend anyway since no kv could have been created with such a key.  As you rightly pointed out, it is even rejected by the backend. 

This has been updated, thanks!

avanigupta (approved):

zhoxing-ms (approved):

PR #26980, opened 2023-07-24
{Core} Add `msal_telemetry` to client telemetry

jiasli:
**Related command**
`az login`

**Description**
Incorporate https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/575 and add `msal_telemetry` from MSAL's response to Azure CLI's client telemetry. (Similar to how `core.allow_broker` configuration is added to client telemetry: https://github.com/Azure/azure-cli/pull/24304)

**Testing Guide**
```
az config set core.allow_broker=true
az login
```

Then 

- Continue the login to trigger the success path, or
- Close the window to trigger the error path

### DIFF
--- a/src/azure-cli-core/azure/cli/core/telemetry.py
+++ b/src/azure-cli-core/azure/cli/core/telemetry.py
@@ -71,6 +71,7 @@ class TelemetrySession:  # pylint: disable=too-many-instance-attributes
         self.poll_start_time = None
         self.poll_end_time = None
         self.allow_broker = None
+        self.msal_telemetry = None
 
     def add_event(self, name, properties):
         for key in self.instrumentation_key:
@@ -217,6 +218,7 @@ class TelemetrySession:  # pylint: disable=too-many-instance-attributes
         set_custom_properties(result, 'RegionInput', self.region_input)
         set_custom_properties(result, 'RegionIdentified', self.region_identified)
         set_custom_properties(result, 'AllowBroker', str(self.allow_broker))
+        set_custom_properties(result, 'MsalTelemetry', str(self.msal_telemetry))
 
         return result
 
@@ -457,6 +459,11 @@ def set_region_identified(region_input, region_identified):
 def set_broker_info(allow_broker):
     # whether customer has configured `allow_broker` to enable WAM(Web Account Manager) login for authentication
     _session.allow_broker = allow_broker
+
+
+@decorators.suppress_all_exceptions()
+def set_msal_telemetry(msal_telemetry):
+    _session.msal_telemetry = msal_telemetry
 
 
 @decorators.suppress_all_exceptions()
--- a/src/azure-cli-core/azure/cli/core/telemetry.py
+++ b/src/azure-cli-core/azure/cli/core/telemetry.py
@@ -463,7 +463,8 @@ def set_broker_info(allow_broker):
 
 @decorators.suppress_all_exceptions()
 def set_msal_telemetry(msal_telemetry):
-    _session.msal_telemetry = msal_telemetry
+    if not _session.msal_telemetry:
+        _session.msal_telemetry = msal_telemetry
 
 
 @decorators.suppress_all_exceptions()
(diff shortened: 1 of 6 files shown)


yonzhan:
Core

src/azure-cli-core/azure/cli/core/telemetry.py
```diff
@@ -459,6 +461,11 @@ def set_broker_info(allow_broker):
     _session.allow_broker = allow_broker
 
 
+@decorators.suppress_all_exceptions()
+def set_msal_telemetry(msal_telemetry):
+    _session.msal_telemetry = msal_telemetry
```
jiasli, line 466: There is one problem with this code: If a command makes multiple MSAL requests, the last telemetry record will overwrite previous ones. Or, shall we only keep the first telemetry record?

```suggestion
    if not _session.msal_telemetry:
        _session.msal_telemetry = msal_telemetry
```

rayluo: If there can be one telemetry entry per each "az foo ..." command, yet one such command may potentially make multiple MSAL requests, it is indeed unclear how or whether to handle this one-to-many mapping. I would suggest to just pick one MSAL request to collect its telemetry, preferably the one which happened to fail, but any choice is still better than nothing. CC @MSamWils

src/azure-cli-core/azure/cli/core/telemetry.py
```diff
@@ -217,6 +218,7 @@ def _get_azure_cli_properties(self):
         set_custom_properties(result, 'RegionInput', self.region_input)
         set_custom_properties(result, 'RegionIdentified', self.region_identified)
         set_custom_properties(result, 'AllowBroker', str(self.allow_broker))
+        set_custom_properties(result, 'MsalTelemetry', str(self.msal_telemetry))
```
evelyn-ys, line 221: Is `self.msal_telemetry` string type or other object type? Wondering why need `str()` here

rayluo: Telemetry returned by MSAL Python is a string. It is also OK for Azure CLI to `str(...)`, so that it is more robust.

rayluo (dismissed):

evelyn-ys (approved):

src/azure-cli-core/azure/cli/core/telemetry.py
```diff
@@ -459,6 +461,12 @@ def set_broker_info(allow_broker):
     _session.allow_broker = allow_broker
 
 
+@decorators.suppress_all_exceptions()
+def set_msal_telemetry(msal_telemetry):
+    if not _session.msal_telemetry:
+        _session.msal_telemetry = msal_telemetry
```
jiasli, line 467: If a CLI command makes several MSAL calls, only the first telemetry record is preserved.

The most special command is `az login` which calls `acquire_token_interactive` first, then several subsequent `acquire_token_silent_with_error`s:

1. `acquire_token_interactive`: initial login
2. `acquire_token_silent_with_error`: list tenants
3. `acquire_token_silent_with_error`: list subscriptions in tenant 1
4. `acquire_token_silent_with_error`: list subscriptions in tenant 2
5. ...

If step 1 fails, its telemetry will be sent and `az login` will fail. If step 1 succeeds, its telemetry will be sent, and step 2 and following steps' telemetry will be discarded, no matter if they fail or not.

For long running operation commands such as `az storage account create`, only the MSAL telemetry for the initial `PUT` request is sent. Following MSAL telemetry for `GET` polling requests is discarded.

src/azure-cli-core/azure/cli/core/telemetry.py
jiasli: Other properties don't have the `str()` casting. Even though `str()` makes it more robust, it may hide errors if MSAL no longer returns a `str`, so let's remove it for consistency:

```suggestion
        set_custom_properties(result, 'MsalTelemetry', self.msal_telemetry)
```

PR #27042, opened 2023-07-31
fixed URLs in src/ excluding /test*

dbradish-microsoft:
**Related command**

**Description**

**Testing Guide**

**History Notes**

---


yonzhan:

bebound:
Some test hard code the exception message, and the error message is updated in this PR.
```python
message = 'Trusted Launch security configuration can be enabled only with Azure Gen2 VMs. Please visit https://learn.microsoft.com/en-us/azure/virtual-machines/trusted-launch for more details.'
        with self.assertRaisesRegex(InvalidArgumentValueError, message):
            self.cmd('vm update -g {rg} -n {vm3} --security-type TrustedLaunch')
```
This need to be fixed case by case.

dbradish-microsoft:
> I do not go through all failed test, just find this test hardcodes the exception message. The error message is updated in this PR.
> 
> ```python
> message = 'Trusted Launch security configuration can be enabled only with Azure Gen2 VMs. Please visit https://learn.microsoft.com/en-us/azure/virtual-machines/trusted-launch for more details.'
>         with self.assertRaisesRegex(InvalidArgumentValueError, message):
>             self.cmd('vm update -g {rg} -n {vm3} --security-type TrustedLaunch')
> ```
> 
> This need to be fixed case by case.

Your example is from file` azure-cli/src/azure-cli/azure/cli/command_modules/vm/tests/latest/test_vm_commands.py` which is not in this PR...  
image
...So what you're saying is that even if we aren't editing any files in `/test*`, those tests are still being run and producing errors?

I don't doubt your statement that this will need to be fixed case by case.  I am looking for a way I can get around that by doing something differently.

Why are tests failing with a URL message change?

bebound:
This PR changes the error message, so the test code also needs to be updated.
https://github.com/Azure/azure-cli/blob/1b63d3ccb55d8df754e59c2d0313c3b9cf8da4a0/src/azure-cli/azure/cli/command_modules/vm/_vm_utils.py#L582-L585

All of the failed tests should be re-recorded. If it still fails, then the test code should be checked.

dbradish-microsoft:
> This PR changes the error message, so the test code also needs to be updated.
>...
> All of the failed tests should be re-recorded. If it still fails, then the test code should be checked.

How does this re-recording happen?  The [original hotfix](https://github.com/Azure/azure-cli/pull/25435) only failed on "Check the Format of Pull Request Title and Content".  I am looking for a way to get the same result but with an accepted Title.

bebound:
You need to add `--live` to record the test: https://github.com/Azure/azure-cli-dev-tools#style-linter-check-and-testing. I'm not sure if you have the permission to do this.

dbradish-microsoft:
Closing PR as this fix needs to originate with engineering as there are pipeline build errors I do not control. The update of embedded URLs remains an open task.

PR #27015, opened 2023-07-27
{Release} Upgrade to Azure CLI 2.51.0

azclibot:
Updated Azure CLI version.

PR #16362, opened 2020-12-25
{Release} Upgrade to Azure CLI 2.17.0

azclibot:
Updated Azure CLI version.

PR #27102, opened 2023-08-07
{CI} Fix Python wheels Test

bebound:
**Description**

In this PR, I change the command to install the wheel directly, to make sure it does not install package from pypi

----

When releasing [2.51.0](https://github.com/Azure/azure-cli/pull/27015), Test Python Wheels fails with this error:
```
ERROR: Package 'azure-cli' requires a different Python: 3.7.16 not in '>=3.8.0'
```
That's expected, as 2.51 does not support 3.7. I'm curious why this test does not find this error earlier.

In fact, `pip install -f /mnt/pypi -q azure-cli==$CLI_VERSION.*` never runs as expected.

This command add `/mnt/pypi` into search path, and try to install `azure-cli==$CLI_VERSION.*`. It has two problem:
1. `/mnt/pypi` does not have higher priority, if pypi contains the specific version, it may install from pypi instead of local
  > pip looks for packages in a number of places: on PyPI (if not disabled via --no-index), in the local filesystem, and in any additional repositories specified via --find-links or --index-url. There is no ordering in the locations that are searched. Rather they are all checked, and the “best” match for the requirements (in terms of version number - see [PEP 440](https://peps.python.org/pep-0440/) for details) is selected.  --https://pip.pypa.io/en/stable/cli/pip_install/#finding-packages
2. `$CLI_VERSION.*` includes `$CLI_VERSION`, which can be found from pypi.

Here is the log in python 3.7. I guess it tries to install local version first, which is not compatible with 3.7, then install 2.50 from pypi.
```
pip install -f /down/az install azure-cli==2.50.0.* --no-cache-dir --no-deps
Looking in links: /down/az
Requirement already satisfied: install in /usr/local/lib/python3.7/site-packages (1.3.5)
Processing ./azure_cli-2.50.0.post20230726073400-py3-none-any.whl
Collecting azure-cli==2.50.0.*
  Downloading azure_cli-2.50.0-py3-none-any.whl (11.6 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 11.6/11.6 MB 9.9 MB/s eta 0:00:00
Installing collected packages: azure-cli
Successfully installed azure-cli-2.50.0
```
Here is the log for 3.8. It successfully installs the local wheel.
```
root@a95e41ea4477:/down/az# pip install -f /down/az install azure-cli==2.50.0.* --no-cache-dir --no-deps
Looking in links: /down/az
Collecting install
  Downloading install-1.3.5-py3-none-any.whl (3.2 kB)
Processing ./azure_cli-2.50.0.post20230726073400-py3-none-any.whl
Installing collected packages: install, azure-cli
Successfully installed azure-cli-2.50.0.post20230726073400 install-1.3.5
```


yonzhan:
CI

jiasli:
Why is the log not in English?

> ```
> Linklərdə axtarış edilir: /down/az
> ```

azure-pipelines.yml
```diff
@@ -449,12 +449,12 @@ jobs:
       echo "== Testing pip install on Python 3.9 =="
       docker run \
         --rm -v $PYPI_FILES:/mnt/pypi python:3.9 \
-        /bin/bash -c "ls /mnt/pypi && pip install -f /mnt/pypi -q azure-cli==$CLI_VERSION.* && az self-test && az --version && sleep 5"
```
jiasli, line 452: The test for Python 3.7 was removed by https://github.com/Azure/azure-cli/pull/27015.

`.*` was initially added by https://github.com/Azure/azure-cli/pull/16362/commits/4baa9d8b9c434bc514aa3c65517cb213d272c4f0 from https://github.com/Azure/azure-cli/pull/16362.

I made a test on the built artifact and noticed `*` is not expanded by the shell if no matching file exists:

```
$ curl https://artprodwus21.artifacts.visualstudio.com/A7b238909-6802-4b65-b90d-184bca47f458/5147fa83-336e-44ef-bbe0-c86b8ae86cbb/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL2F6Y2xpdG9vbHMvcHJvamVjdElkLzUxNDdmYTgzLTMzNmUtNDRlZi1iYmUwLWM4NmI4YWU4NmNiYi9idWlsZElkLzgyNDg5L2FydGlmYWN0TmFtZS9weXBp0/content?format=zip -o pypi.zip
$ unzip pypi.zip
$ cd pypi
$ echo azure-*
azure-cli-2.51.0.post20230815081512.tar.gz azure-cli-core-2.51.0.post20230815081512.tar.gz azure-cli-telemetry-1.1.0.post20230815081512.tar.gz
$ CLI_VERSION=2.51.0
$ echo azure-cli==$CLI_VERSION.*
azure-cli==2.51.0.*
```

Since #27001, `2.50.0.post*` on the local disk is not compatible with Python 3.7, so `pip` falls back to a compatible version `2.50.0` from PyPI.

azure-pipelines.yml
```diff
@@ -449,12 +449,12 @@ jobs:
...
+        /bin/bash -c "cd /mnt/pypi && ls && pip install -f ./ azure_cli-$CLI_VERSION*whl && az self-test && az --version && sleep 5"
```
jiasli, line 452: Why cd into `/mnt/pypi`? Is `-f ./` still necessary now?

jiasli, line 452: If we force it to install the `whl` packages, there will be no need to care about `$CLI_VERSION`. I guess we may simply use `azure_cli*.whl` to install all local wheels:

```suggestion
        /bin/bash -c "cd /mnt/pypi && ls && pip install -f ./ azure_cli*.whl && az self-test && az --version && sleep 5"
```

This will include 

```
$ echo azure_cli*.whl
azure_cli-2.51.0.post20230815081512-py3-none-any.whl
azure_cli_core-2.51.0.post20230815081512-py3-none-any.whl
azure_cli_telemetry-1.1.0.post20230815081512-py3-none-any.whl
```

bebound:
> Why is the log not in English but Azerbaijani?
> 
> > ```
> > Linklərdə axtarış edilir: /down/az
> > ```

My OpenAI Translator extension is triggered unexpectedly.
The last two lines
```
Bu mətni Azərbaycan dilinə tərcümə edin, zəhmət olmasa cümlələri izah etməyin, sadəcə tərcümə edin və ya onları olduğu kimi qoyun.```
2.51-dən sonra yerli çark 3.7 ilə uyğun deyil və pypi-da 2.51 yoxdur, bu `Paket 'azure-cli' fərqli bir Python tələb edir: 3.7.16 '>= 3.8.0' deyil` səhvi yaradır.
```
is its prompt "Please translate this text into Azerbaijani language, without explaining the sentences, just translate them as they are.```
After 2.51, it is not compatible with native wheel 3.7 and there is no 2.51 on pypi, which creates an error "Package 'azure-cli' requires a different Python: 3.7.16'>= 3.8.0'"."

azure-pipelines.yml
bebound: I don't want to write full path in `-f` and whl file.

Yes, the `azure_cli-$CLI_VERSION*whl` requires `azure-cli-core-XX.post`, we need to add this folder into additional repositories.

bebound: `pip install azure_cli*.whl` also works, but I want to make this test command close to the real scenario: `pip install azure-cli`.

jiasli (approved):

azure-pipelines.yml
```diff
@@ -449,12 +449,12 @@ jobs:
       echo "== Testing pip install on Python 3.9 =="
       docker run \
         --rm -v $PYPI_FILES:/mnt/pypi python:3.9 \
-        /bin/bash -c "ls /mnt/pypi && pip install -f /mnt/pypi -q azure-cli==$CLI_VERSION.* && az self-test && az --version && sleep 5"
+        /bin/bash -c "cd /mnt/pypi && ls && pip install --find-links ./ azure_cli-$CLI_VERSION*whl && az self-test && az --version && sleep 5"
```
jiasli, line 452: It'll be good to always make `*` match as little as possible:

```suggestion
        /bin/bash -c "cd /mnt/pypi && ls && pip install --find-links ./ azure_cli-*-py3-none-any.whl && az self-test && az --version && sleep 5"
```

PR #27135, opened 2023-08-09
{Core} Fix authentication TypeError

pvaneck:
With the latest version of azure-core, the `BearerTokenCredentialPolicy` now passes in a new keyword argument to `get_token` called `enable_cae`. 

This PR fixes TypeErrors in msal for unexpected keyword arguments due to this keyword being propagated.

Fixes: https://github.com/Azure/azure-cli/issues/27131


yonzhan:
Core

jiasli:
@pvaneck, Thanks for the contribution, but as discussed with @xiangyan99, we will not only pop `tenant_id` and `enable_cae` from `kwargs`, but will make sure only recognized arguments are passed to MSAL.

pvaneck:
@jiasli, Sure, sounds good. Let me close this in favor of a more robust solution.

PR #27137, opened 2023-08-10
Enforce LF-only line endings in git

bluca:
Having some files using CRLF makes things extremely messy when
using tarballs on Linux, as the git repository gets automatically
converted, but the tarball stores the original line ending,
so the two do not match anymore and importing new releases fails.

Enforce that LF-only line endings get committed to the repository,
and add instructions for Windows developers on how to configure
their local git client to checkout in CRLF, so that CRLF never
propagate to the repository but remain only in local checkouts.


yonzhan:

bebound:
I'm not a git guru, I checked the doc after you created this PR. So correct me if I'm wrong.

I guess you want to do this:
1. Every file ends with `lf` in Git index.
2. Windows users still get `crlf` when checkout.

But this PR does not achieve the above target?
1. The bat files end with crlf in `https://api.github.com/repos/azure/azure-cli/tarball/dev`

2. `eof` enforce line ending conversion when checkout, Windows user are affected.
> This attribute marks a path to use a specific line-ending style in the working tree when it is checked out
Set to string value "crlf"
This setting converts the file’s line endings in the working directory to crlf when the file is checked out.
> Set to string value "lf"
This setting uses the same line endings in the working directory as in the index when the file is checked out.
-- https://www.git-scm.com/docs/gitattributes#_eol

I'm using Windows and have `core.autocrlf=true` setting. Here is the test result.
Before:
```
git ls-files --eol src/azure-cli/az.bat
i/lf    w/crlf  attr/text=auto eol=crlf src/azure-cli/az.bat
git ls-files --eol .github/policies/resourceManagement.yml
i/crlf  w/crlf  attr/text=auto          .github/policies/resourceManagement.yml
git ls-files --eol .\src\azure-cli\setup.py
i/lf    w/crlf  attr/text=auto          src/azure-cli/setup.py
```
After checkout to this pr and run `git rm -rf --cached . && git reset --hard HEAD`
```
git ls-files --eol src/azure-cli/az.bat
i/lf    w/lf    attr/text=auto eol=lf   src/azure-cli/az.bat
git ls-files --eol .github/policies/resourceManagement.yml
i/lf    w/lf    attr/text=auto eol=lf   .github/policies/resourceManagement.yml
git ls-files --eol .\src\azure-cli\setup.py
i/lf    w/lf    attr/text=auto eol=lf   src/azure-cli/setup.py
```
The above files are converted to lf.
I prefer to keep the original settings. It makes sense to keep bat files ending with crlf.

bluca:
> 1. I don’t know which tarball you’re using. The bat files end with crlf in `https://api.github.com/repos/azure/azure-cli/tarball/dev`

The ones from https://github.com/Azure/azure-cli/releases and yes having any files with crlf in release tarballs is just broken. That can be allowed to happen if and only if a project is to be used exclusively on Windows, which is not the case here.

> 2. `eof` enforce line ending conversion when checkout, Windows user are affected.
> 
> > This attribute marks a path to use a specific line-ending style in the working tree when it is checked out
> > Set to string value "crlf"
> > This setting converts the file’s line endings in the working directory to crlf when the file is checked out.
> > Set to string value "lf"
> > This setting uses the same line endings in the working directory as in the index when the file is checked out.
> > -- https://www.git-scm.com/docs/gitattributes#_eol
> 
> I'm using Windows and have `core.autocrlf=true` setting. Here is the test result. Before:

Which version of git is that? Also can you try with a fresh clone after changing the global setting? Honestly I don't know why git doesn't make this the default on WIndows, anything else is just broken. I was sure that was the right combination to achieve crlf on checkout on Windows, but maybe there's some other knob to turn?

> The above files are converted to lf. I’m open to change `resourceManagement.yml` to lf, but I prefer to keep the original eol settings. It makes sense to keep bat files ending with crlf.

No files can have crlf, otherwise things are just broken and we need to fork the repository and re-create the tarballs from scratch on Linux, which is a gigantic pain.

bluca:
Maybe the issue is the gitattributes? Perhaps another approach would be to drop that bit, convert once all files, and then add a CI test that enforces no clrf anywhere?

bebound:
Find similar statement from GitHub doc:
> text eol=lf Git will always convert line endings to LF on checkout. You should use this for files that must keep LF endings, even on Windows.  -- https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings

`eol=lf` enforces lF on Windows, so this change breaks bat files on Windows.

------
Currently, only `bat`, `sln` and `resourceManagement.yml` contains CRLF.

If you want to make bat file end with LF, removing `*.bat eol=crlf` in `.gitattributes` meets your needs.
But bat is a Windows only file, I'd like to keep it using CRLF. May I know how bat files affect your workflow on Linux?


``` bash
❯ find . -not -type d -exec file "{}" ";" | grep CRLF
./src/azure-cli/az.bat: DOS batch file, ASCII text, with CRLF line terminators
./src/azure-cli/azure/cli/command_modules/monitor/grammar/autoscale/build_python.bat: ASCII text, with CRLF line terminators
./src/azure-cli/azure/cli/command_modules/monitor/grammar/autoscale/grun.bat: DOS batch file, ASCII text, with CRLF line terminators
./src/azure-cli/azure/cli/command_modules/monitor/grammar/autoscale/run_test.bat: ASCII text, with CRLF line terminators
./src/azure-cli/azure/cli/command_modules/monitor/grammar/metric_alert/build_python.bat: ASCII text, with CRLF line terminators
./src/azure-cli/azure/cli/command_modules/monitor/grammar/metric_alert/grun.bat: DOS batch file, ASCII text, with CRLF line terminators
./src/azure-cli/azure/cli/command_modules/monitor/grammar/metric_alert/run_test.bat: ASCII text, with CRLF line terminators
./src/azure-cli/azure/cli/command_modules/batch/tests/latest/generate_full_cli_help.bat: ASCII text, with CRLF line terminators
./azure-cli2017.sln: Unicode text, UTF-8 (with BOM) text, with CRLF line terminators
./doc/sphinx/make.bat: DOS batch file, ASCII text, with CRLF line terminators
./doc/sphinx/sphinx.sln: Unicode text, UTF-8 (with BOM) text, with CRLF line terminators
./scripts/install_full.bat: ASCII text, with CRLF line terminators
./scripts/ci/dependency_check.bat: ASCII text, with CRLF line terminators
./.github/policies/resourceManagement.yml: Unicode text, UTF-8 text, with very long lines (352), with CRLF line terminators
./azure-cli.sln: Unicode text, UTF-8 (with BOM) text, with CRLF line terminators
./build_scripts/windows/azure-cli.sln: Unicode text, UTF-8 (with BOM) text, with CRLF line terminators
```

bluca:
> But bat is a Windows only file, I'd like to keep it using CRLF. May I know how bat files affect your workflow on Linux?

As mentioned in the PR description:

> Having some files using CRLF makes things extremely messy when
using tarballs on Linux, as the git repository gets automatically
converted, but the tarball stores the original line ending,
so the two do not match anymore and importing new releases fails.

Having any file checked in with crlf in git, for projects that are not exclusively and only Windows-specific, is just broken.

bluca:
Ok I've dropped the settings in .gitattributes, and instead I've added a CI job and converted all the files. This should allow `core.autocrlf=true` to do its job, and stop any crlf from being merged. Can anybody running on Windows check that it works as expected please?

bebound:
The git checkout content on Linux is identical to tarball, I don't see your point.

bluca:
> The git checkout content on Linux is identical to tarball, I don't see your point.

No, it is not when exported and reimported, because the line endings are changed.

> `sln` file is exclusive to visual studio, which is not available on Linux, keep CRLF looks good to me.

No, it does not, checking CRLF files in git is broken and breaks many workflows. They can be added automatically on checkouts as documented.

Issue #19640, opened 2021-09-22
Unofficial Azure CLI packages are distributed by Linux repositories


jiasli:
Community packagers release **unofficial** Azure CLI packages on multiple Linux repositories. These packages are **not maintained by Microsoft**, outdated, and contain various bugs.

### 2.0.81 for Focal 20.04

Package: https://packages.ubuntu.com/source/focal/azure-cli
Issues: https://github.com/Azure/azure-cli/issues?q=is%3Aissue+2.0.81

### 2.18.0 for Hirsute 21.04

Package: https://packages.ubuntu.com/hirsute/python3-azure-cli
Issues: https://github.com/Azure/azure-cli/issues?q=is%3Aissue+2.18.0

### 2.18.0 for Bullseye 11

Package: https://packages.debian.org/bullseye/azure-cli
Issues: https://github.com/Azure/azure-cli/issues?q=is%3Aissue+2.18.0

yonzhan:
feature request

jiasli:
### https://packages.microsoft.com/ is the ONLY source where official Azure CLI packages are distributed.

## Install official Azure CLI

To install the latest official Azure CLI, please follow the official document: https://docs.microsoft.com/en-us/cli/azure/install-azure-cli-linux?pivots=apt

## Uninstall unofficial Azure CLI

If you have already installed the unofficial Azure CLI, you may uninstall it first with 

```
sudo apt remove azure-cli --yes && sudo apt autoremove --yes
```

In a docker container:

```
apt remove azure-cli -y && apt autoremove -y
```

Then install from the official source.

orgads:
On debian testing, when I'm using the automated installed (`curl -sL https://aka.ms/InstallAzureCLIDeb | sudo bash`), it still installs the debian package, because it has a newer version (official is `2.30.0-1~buster`, while debian is `2.30.0-2`).

orgads:
https://github.com/Azure/azure-cli/pull/20557

jiasli:
@bluca, I noticed that recently you have published **azure-cli (2.33.0-2)** for Debian sid: https://packages.debian.org/sid/azure-cli

![image](https://user-images.githubusercontent.com/4003950/153806148-a8eeebf2-2542-470a-8003-3cac717739c0.png)


This makes it impossible to install official Azure CLI 2.33.0 from https://packages.microsoft.com/repos/azure-cli/

[...]

PR #27144, opened 2023-08-10
Snap: refresh and move to core22

gjolly:
**Related command**
All

**Description**

The current version of the `snapcraft.yaml` is very old and uses `classic` confinement. These changes bring the snap to the latest base: `core22` and switch the confinement to strict in order to ease the publication process.

Note that the Azure CLI is not published into the snapstore anymore. This PR creates the per-requisites for republishing the snap into the store.

**Testing Guide**

Build the snap by running `snapcraft`
Install it locally: `snap install --dangerous azure-cli_*.snap`
Verify that the CLI is working as expected: `azure-cli.az xxx`

(since the snap is not named `az` but `azure-cli` the alias is not setup by default)

**History Notes**

Historically, Microsoft was publishing the Azure CLI in the snapstore. It was decided to remove the snap from the store as it wasn't updated and was confusing users who would install the CLI via the snap and end up with an out-dated non-working software. For similarly reasons, Ubuntu decided to remove the upstream Debian package from its repository as the package doesn't contain an up-to-date version of the CLI. Moreover, the debian packages do not seem very well suited for this kind of CLI which are constantly updated and have to follow new API updates.
We've now ended up in a situation where users of Ubuntu cannot natively install the CLI and have to add extra package repositories to their source list in order to get it.

### DIFF
--- a/.snapcraft.yaml
+++ b/.snapcraft.yaml
@@ -3,40 +3,30 @@ summary: Microsoft Azure CLI
 description: |
  A great cloud needs great tools; we're excited to introduce Azure CLI, our
  next generation multi-platform command line experience for Azure.
-version: "release"
-version-script: |
- grep VERSION src/azure-cli/setup.py | head -n 1 | cut -d'"' -f2
-confinement: classic
+confinement: strict
 grade: stable
+base: core22
+adopt-info: azure-cli
+
 apps:
   az:
-    command: usr/bin/python -m azure.cli
-    completer: az.completion
+    command: bin/python3 $SNAP/bin/az
+    completer: az.snap.completion
+    plugs:
+      - network
+      - home
+
 parts:
   azure-cli:
     plugin: python
     source: .
-    python-version: python3
-    build-packages:
-      - build-essential
-      - python3-dev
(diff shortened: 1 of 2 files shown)


yonzhan:

jiasli:
@gjolly, Thanks for the contribution. However, Azure CLI has dropped support for snap long ago. We won't make any changes to snap-related files except removing them. Supported platforms are documented by https://learn.microsoft.com/en-us/cli/azure/install-azure-cli

> For similarly reasons, Ubuntu decided to remove the **upstream Debian package** from its repository as the package doesn't contain an up-to-date version of the CLI.

Could you provide more information regarding the **upstream Debian package**, such as link to it?

> We've now ended up in a situation where users of Ubuntu cannot natively install the CLI and have to add extra package repositories to their source list in order to get it.

This is expected and demanded by Microsoft security policy that all Microsoft Linux packages must be released from https://packages.microsoft.com/.

Some community developers build their own Azure CLI DEB packages, but these packages have been noticed to be buggy (https://github.com/Azure/azure-cli/issues/19640).

PR #27296, opened 2023-08-31
{Core} Sort command loaders to avoid random diff in docs

xmdanni:
**Related command**
Should be all commands when batch export the help files. Examples: `az acr query`, `az ad ds create`, `az mysql down`.

**Description**
We found random output of arguments while batch getting help files, it is because of the random sequence of command loaders. 
image
The random output is introduced by ___for looping a set___, as set in Python is unordered, every execution can lead to different result, and that makes the help files randomly get wrong parameter detail and wrong parameter group info when exporting it in a batch by calling `load_arguments()` in *line 487*. 
After making change to sort the command loader list, the output become steady and correct in my local test. So, we need a steady sequence with the correct order (extension loaders first) here to get a steady and correct output for help files.

For example 1 `az ad ds create`, when export all help files while installed extension `ad`, the parameter "subscription" might be loaded by `azext_ad.DomainServicesResorceProviderCommandsLoader` or `azure.cli.command_modules.role.RoleCommandsLoader`. If the arguments are first updated by `azure.cli.command_modules.role.RoleCommandsLoader`, the parameter "subscription" is lacking; If the arguments are first updated by `azext_ad.DomainServicesResorceProviderCommandsLoader`, the parameter "subscription" will exist and align with the output of `az ad ds create -h`. 
image

For example 2 `az mysql down`, when export all help files while installed extension `db-up`, parameters might be loaded by `azext_db_up.DbupCommandsLoader` or `azure.cli.command_modules.rdbms.RdbmsCommandsLoader`. If the arguments are first updated by `azure.cli.command_modules.rdbms.RdbmsCommandsLoader`, the parameters in help files are:

Command | Parameter | group_name
-- | -- | --
az mysql down | --subscription | Resource Id Arguments
az mysql down | --resource-group -g | Resource Id Arguments
az mysql down | --server-name -s | Resource Id Arguments
az mysql down | --ids | Resource Id Arguments

If the arguments are first updated by `azext_db_up.DbupCommandsLoader`, the parameters align with the output of `az mysql down -h`:

Command | Parameter | group_name
-- | -- | --
az mysql down | --subscription | Global Arguments
az mysql down | --resource-group -g | null
az mysql down | --server-name -s | null

image

For example 3 `az acr query`, when export all help files while installed extension `acrquery`, the parameter "username" and "password" might be loaded by `azure.cli.command_modules.acr.ACRCommandsLoader` or `azext_acrquery.AcrqueryCommandsLoader`. If the arguments are first updated by `azure.cli.command_modules.acr.ACRCommandsLoader`, the option lists in the object of these parameters will be ['--username', '-u'] and ['--password', '-p'], which is from [azure-cli/src/azure-cli/azure/cli/command_modules/acr/_params.py at main · Azure/azure-cli (github.com)](https://github.com/Azure/azure-cli/blob/main/src/azure-cli/azure/cli/command_modules/acr/_params.py#L90), otherwise if the arguments are first updated by `azext_acrquery.AcrqueryCommandsLoader`, the option lists in the object of these parameters will be ['--username'] and ['--password'], which is from [azure-cli-extensions/src/acrquery/azext_acrquery/_params.py at main · Azure/azure-cli-extensions (github.com)](https://github.com/Azure/azure-cli-extensions/blob/main/src/acrquery/azext_acrquery/_params.py#L16). 

For example 4 `az iot hub certificate root-authority set` in extension `azure-iot`, the parameter `--hub-name` will sometimes has `-n` and sometimes not, caused by random loader sequence. 
image

For example 5 `az webapp scan show-result` in extension `webapp`, several parameters change back and forth as below, caused by random loader sequence. 
image

**Testing Guide**

Check the arguments detail in the help objects of the example commands to make sure no random changes with different runs.

**History Notes**

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/__init__.py
+++ b/src/azure-cli-core/azure/cli/core/__init__.py
@@ -489,7 +489,11 @@ class MainCommandsLoader(CLICommandsLoader):
 
         return self.command_table
 
-    def sort_command_loaders(self, command_loaders):
+    @staticmethod
+    def _sort_command_loaders(self, command_loaders):
+        if len(command_loaders) == 1:
+            return list(command_loaders)
+
         module_command_loaders = []
         extension_command_loaders = []
 
@@ -523,7 +527,7 @@ class MainCommandsLoader(CLICommandsLoader):
             command_loaders = self.cmd_to_loader_map.get(command, None)
 
         if command_loaders:
-            command_loaders = self.sort_command_loaders(command_loaders)
+            command_loaders = self._sort_command_loaders(command_loaders)
             for loader in command_loaders:
 
                 # register global args
--- a/src/azure-cli-core/azure/cli/core/__init__.py
+++ b/src/azure-cli-core/azure/cli/core/__init__.py
@@ -490,7 +490,7 @@ class MainCommandsLoader(CLICommandsLoader):
         return self.command_table
 
     @staticmethod
-    def _sort_command_loaders(self, command_loaders):
+    def _sort_command_loaders(command_loaders):
         if len(command_loaders) == 1:
             return list(command_loaders)
 
--- a/src/azure-cli-core/azure/cli/core/__init__.py
+++ b/src/azure-cli-core/azure/cli/core/__init__.py
@@ -491,9 +491,6 @@ class MainCommandsLoader(CLICommandsLoader):
 
     @staticmethod
     def _sort_command_loaders(command_loaders):
-        if len(command_loaders) == 1:
-            return list(command_loaders)
-
         module_command_loaders = []
         extension_command_loaders = []
 
@@ -522,12 +519,13 @@ class MainCommandsLoader(CLICommandsLoader):
             command_loaders = set()
             for loaders in self.cmd_to_loader_map.values():
                 command_loaders = command_loaders.union(set(loaders))
+            # sort command loaders for consistent order when loading all commands for docs generation to avoid random diff
+            command_loaders = self._sort_command_loaders(command_loaders)
             logger.info('Applying %s command loaders...', len(command_loaders))
         else:
             command_loaders = self.cmd_to_loader_map.get(command, None)
 
         if command_loaders:
-            command_loaders = self._sort_command_loaders(command_loaders)
             for loader in command_loaders:
 
                 # register global args
--- a/src/azure-cli-core/azure/cli/core/__init__.py
+++ b/src/azure-cli-core/azure/cli/core/__init__.py
@@ -489,6 +489,25 @@ class MainCommandsLoader(CLICommandsLoader):
 
         return self.command_table
 
+    def sort_command_loaders(self, command_loaders):
+        core_command_loaders = []
+        extension_command_loaders = []
+
+        # Separate core and extension command loaders
+        for loader in command_loaders:
+            if loader.__module__.startswith('azext'):
+                extension_command_loaders.append(loader)
+            else:
+                core_command_loaders.append(loader)
+        
+        # Sort name in each command loader list
+        core_command_loaders.sort(key=lambda loader: loader.__class__.__name__)
+        extension_command_loaders.sort(key=lambda loader: loader.__class__.__name__)
+
+        # Core first, then extension
+        sorted_command_loaders = core_command_loaders + extension_command_loaders
+        return sorted_command_loaders
+
     def load_arguments(self, command=None):
         from azure.cli.core.commands.parameters import (
             resource_group_name_type, get_location_type, deployment_name_type, vnet_name_type, subnet_name_type)
@@ -504,6 +523,7 @@ class MainCommandsLoader(CLICommandsLoader):
             command_loaders = self.cmd_to_loader_map.get(command, None)
 
         if command_loaders:
+            command_loaders = self.sort_command_loaders(command_loaders)
             for loader in command_loaders:
 
                 # register global args
--- a/src/azure-cli-core/azure/cli/core/__init__.py
+++ b/src/azure-cli-core/azure/cli/core/__init__.py
@@ -490,22 +490,22 @@ class MainCommandsLoader(CLICommandsLoader):
         return self.command_table
 
     def sort_command_loaders(self, command_loaders):
-        core_command_loaders = []
+        module_command_loaders = []
         extension_command_loaders = []
 
-        # Separate core and extension command loaders
+        # Separate module and extension command loaders
         for loader in command_loaders:
             if loader.__module__.startswith('azext'):
                 extension_command_loaders.append(loader)
             else:
-                core_command_loaders.append(loader)
+                module_command_loaders.append(loader)
         
         # Sort name in each command loader list
-        core_command_loaders.sort(key=lambda loader: loader.__class__.__name__)
+        module_command_loaders.sort(key=lambda loader: loader.__class__.__name__)
         extension_command_loaders.sort(key=lambda loader: loader.__class__.__name__)
 
-        # Core first, then extension
-        sorted_command_loaders = core_command_loaders + extension_command_loaders
+        # Module first, then extension
+        sorted_command_loaders = module_command_loaders + extension_command_loaders
         return sorted_command_loaders
 
     def load_arguments(self, command=None):
--- a/src/azure-cli-core/azure/cli/core/__init__.py
+++ b/src/azure-cli-core/azure/cli/core/__init__.py
@@ -503,7 +503,7 @@ class MainCommandsLoader(CLICommandsLoader):
                 extension_command_loaders.append(loader)
             else:
                 module_command_loaders.append(loader)
-        
+
         # Sort name in each command loader list
         module_command_loaders.sort(key=lambda loader: loader.__class__.__name__)
         extension_command_loaders.sort(key=lambda loader: loader.__class__.__name__)


yonzhan:

zhoxing-ms:
Please add context in the PR description to help others to understand it's background and purpose~

src/azure-cli-core/azure/cli/core/__init__.py
```diff
@@ -499,7 +499,7 @@ def load_arguments(self, command=None):
             command_loaders = self.cmd_to_loader_map.get(command, None)
 
         if command_loaders:
-            for loader in command_loaders:
+            for loader in sorted(command_loaders, key=lambda x:x.__class__.__name__):
```
jsntcy, line 502: Some questions we can discuss:

- Why does docs depends on the order of command loaders?
- Are there any other solutions which may be better?
- Is there any performance downgrade if we sort command loaders?

jiasli:
Though I haven't looked deeply into this issue, it doesn't seem to happen to the in-tool help.

To isolate the issue, we first disable command index to make CLI load all modules and extensions:

```
az config set core.use_command_index=false
```

Regarding example 1, it is by-design that `az ad app create` from `azure.cli.command_modules.role.RoleCommandsLoader` ignores `--subscription`:

https://github.com/Azure/azure-cli/blob/e8efb791ea2f0ba5e4f172b26f75725fd8aa8079/src/azure-cli/azure/cli/command_modules/role/_params.py#L26

```
> az ad app create -h
...
Global Arguments
    --debug                        : Increase logging verbosity to show all debug logs.
    --help -h                      : Show this help message and exit.
    --only-show-errors             : Only show errors, suppressing warnings.
    --output -o                    : Output format.  Allowed values: json, jsonc, none, table, tsv,
                                     yaml, yamlc.  Default: json.
    --query                        : JMESPath query string. See http://jmespath.org/ for more
                                     information and examples.
    --verbose                      : Increase logging verbosity. Use --debug for full debug logs.
```

And it is also by design that `az ad ds create` from `azext_ad.DomainServicesResorceProviderCommandsLoader` supports `--subscription`:

```
> az ad ds create -h
...
Global Arguments
    --debug                        : Increase logging verbosity to show all debug logs.
    --help -h                      : Show this help message and exit.
    --only-show-errors             : Only show errors, suppressing warnings.
    --output -o                    : Output format.  Allowed values: json, jsonc, none, table, tsv,
                                     yaml, yamlc.  Default: json.
    --query                        : JMESPath query string. See http://jmespath.org/ for more
                                     information and examples.
    --subscription                 : Name or ID of subscription. You can configure the default
                                     subscription using `az account set -s NAME_OR_ID`.
    --verbose                      : Increase logging verbosity. Use --debug for full debug logs.
```

Then let's check the source code:

https://github.com/Azure/azure-cli/blob/ccf2bd499dc75e38010346febde63a0b629c34ee/src/azure-cli-core/azure/cli/core/__init__.py#L492-L502

When `az ad ds create -h` is run, `command` will be `'ad ds create'`. L499 is hit and only one command loader will be loaded, in this case `azext_ad.DomainServicesResourceProviderCommandsLoader`:

image

`azure.cli.command_modules.role.RoleCommandsLoader` will not be loaded and that's why `--subscription` is not ignored.

Question: Can the docs generation pipeline correctly handle different presence of `--subscription` for `az ad app create` and `az ad ds create`? In other words, it is incorrect to show `--subscription` for `az ad app create`, and it is incorrect to hide `--subscription` for `az ad ds create`.

jiasli:
Whether `--subscription` is included in the help message is determined when L696 is called:

https://github.com/Azure/azure-cli/blob/7a4d03f077151be8fc312c9ce82febd1dda118cb/src/azure-cli-core/azure/cli/core/__init__.py#L694-L696

It updates the command's reflection-retrieved arguments based on the information in `master_arg_registry`.

If `azure.cli.command_modules.role.RoleCommandsLoader` has been loaded, the _ignored_ `_subscription` will appear in `master_arg_registry`, under `ad`:

image

When `az ad ds create`'s argument is updated, the ignored `_subscription` takes effect.

On the other hand, if `azure.cli.command_modules.role.RoleCommandsLoader` has not been loaded, `_subscription` won't appear under `master_arg_registry`. When `az ad ds create`'s argument is updated, `_subscription` will be kept.

Even though loading the extension first will solve this specific case, this issue still occurs if things reverse: a command module supports `_subscription` but an extension ignores it. In that case, if the extension is loaded first, the help message for the command module will not have `_subscription` which is wrong.

For example 1, without changing the current command design such as moving `az ad ds` to a dedicated command group `az ds`, the solution is to not ignore `_subscription` on `az ad` which includes `az ad ds`, but on deeper levels - `az ad app/sp/user/group`.

I haven't looked deeply into example 2 and 3. I will take a deeper look next week.

src/azure-cli-core/azure/cli/core/__init__.py
```diff
@@ -499,7 +499,7 @@ def load_arguments(self, command=None):
             command_loaders = self.cmd_to_loader_map.get(command, None)
 
         if command_loaders:
-            for loader in command_loaders:
+            for loader in sorted(command_loaders, key=lambda loader: (loader.__module__, loader.__class__.__name__)):
```
jiasli, line 502: For a command module, such as `role`, 

- `loader.__module__` evaluates to `azure.cli.command_modules.role`
- `loader.__class__.__name__` evaluates to `RoleCommandsLoader`

For an extension, such as `ad`,

- `loader.__module__` evaluates to `azext_ad`
- `loader.__class__.__name__` evaluates to `DomainServicesResourceProviderCommandsLoader`

The sorted list will look like

```
azext_a
azext_b
...
azext_z
azure.cli.command_modules.a
azure.cli.command_modules.b
...
azure.cli.command_modules.z
```

In my opinion, prioritizing extensions over command modules by string comparing `loader.__module__` (`azext_ad` < `azure.cli.command_modules.role`) is a little bit tricky. It is only a coincidence that `azext_` appears before `azure.` Things can get complex if we want to prioritize command modules.

Even if we want to prioritize extensions, it's better to do it programmatically, by either:

1. Prefixing the key tuple with another `0 if loader.__module__.startswith('azext') else 1` representing if `loader.__module__` is an extension and its priority.
2. Extracting extensions and command modules into 2 separate lists, sorting them separately and combining these 2 lists into one.

jiasli:
For example 2, `server_name` is defined under `az mysql` with `id_part='name'`:

https://github.com/Azure/azure-cli/blob/70f70f39cdaff028c111a750a3419d5c60590d9c/src/azure-cli/azure/cli/command_modules/rdbms/_params.py#L50-L52

https://github.com/Azure/azure-cli/blob/70f70f39cdaff028c111a750a3419d5c60590d9c/src/azure-cli/azure/cli/command_modules/rdbms/_params.py#L43

The extension `db_up` alters `server_name` for `az mysql down`:

https://github.com/Azure/azure-cli-extensions/blob/4d67f30a9d4542ca71c801de32d577e0b7b03bc6/src/db-up/azext_db_up/_params.py#L42

```py
            c.argument('server_name', options_list=['--server-name', '-s'], help='Name of the server.')
```

For example 3, `username` and `password` are defined under `az acr` with short name `-u`, `-p`:

https://github.com/Azure/azure-cli/blob/ccf2bd499dc75e38010346febde63a0b629c34ee/src/azure-cli/azure/cli/command_modules/acr/_params.py#L90-L91

The extension `acrquery` alters `username` and `password` for `az acr query`:

https://github.com/Azure/azure-cli-extensions/blob/16217261d3ecb66523ee6e833dcfedda0b3d6449/src/acrquery/azext_acrquery/_params.py#L16-L17

```py
        c.argument('username', help='Registry username')
        c.argument('password', help='Registry password')
```

Notice: Instead of replacing all properties of an argument as a single object, only specified properties are altered:

https://github.com/microsoft/knack/blob/cd590312b9b4ea1402d7a0a76814f6ca4da7fab8/knack/arguments.py#L41

```py
            self.settings.update(**other.settings)
```

So, in example 2, `id_part` is preserved for `az mysql down`; in example 3, `options_list` is preserved for `az acr query`.

All these 3 examples have one thing in common: An argument is defined at an upper level (`az ad`) by a command module and gets altered at a lower level (`az ad ds`) by an extension.

When a specific command from an extension is run, the core will only load the extension, so that the command module's argument definition won't take effect.

However, in the reference doc generation pipeline, all command modules and extensions are loaded, but in a random order. When a command module is loaded and then a corresponding extension is loaded, the command module's argument definition takes effect, affecting the extension's arguments.

src/azure-cli-core/azure/cli/core/__init__.py
```diff
@@ -499,7 +499,7 @@ def load_arguments(self, command=None):
             command_loaders = self.cmd_to_loader_map.get(command, None)
 
         if command_loaders:
-            for loader in command_loaders:
+            for loader in sorted(command_loaders, key=lambda loader: (0 if loader.__module__.startswith('azext') else 1, loader.__module__, loader.__class__.__name__)):
```
jiasli, line 502: I think there is a side effect if we prioritize extensions. Consider the reserved case described in https://github.com/Azure/azure-cli/pull/27296#issuecomment-1878581800:

> A command module supports `_subscription` but an extension ignores it. In that case, if the extension is loaded first, the help message for the command module will not have `_subscription` which is wrong.

src/azure-cli-core/azure/cli/core/__init__.py
```diff
@@ -489,6 +489,25 @@ def _get_extension_suppressions(mod_loaders):
 
         return self.command_table
 
+    def sort_command_loaders(self, command_loaders):
+        core_command_loaders = []
```
jiasli, line 493: We usually call them modules:

```suggestion
        module_command_loaders = []
```

`core` has a special meaning as `azure-cli-core`.

jiasli: Yes, they are in one repo, but `modules` usually refer to `azure.cli.command_modules` (path in repo: `src/azure-cli/azure/cli/command_modules`) while `core` refers to `azure.cli.core` (path in repo: `src/azure-cli-core/azure/cli/core`). They are different packages:

https://github.com/Azure/azure-cli/blob/dc4393dfe5319df8ce4190f12f3f981ff8f49563/src/azure-cli/setup.py#L161

https://github.com/Azure/azure-cli/blob/3a2c33aa525d4718c9c277e2aa3ce5c8c4d50a20/src/azure-cli-core/setup.py#L73

src/azure-cli-core/azure/cli/core/__init__.py
```diff
@@ -489,6 +489,25 @@ def _get_extension_suppressions(mod_loaders):
 
         return self.command_table
 
+    def sort_command_loaders(self, command_loaders):
```
jiasli, line 492: Making it a protected static method sounds more proper:
```suggestion
    @staticmethod
    def _sort_command_loaders(command_loaders):
```

src/azure-cli-core/azure/cli/core/__init__.py
```diff
@@ -504,6 +523,7 @@ def load_arguments(self, command=None):
             command_loaders = self.cmd_to_loader_map.get(command, None)
 
         if command_loaders:
+            command_loaders = self.sort_command_loaders(command_loaders)
```
jiasli, line 526:
```suggestion
            command_loaders = self._sort_command_loaders(command_loaders)
```

src/azure-cli-core/azure/cli/core/__init__.py
```diff
@@ -489,6 +489,25 @@ def _get_extension_suppressions(mod_loaders):
 
         return self.command_table
 
+    def sort_command_loaders(self, command_loaders):
+        module_command_loaders = []
```
jiasli, line 493: A small optimization: skip the sorting if there is only one element:

```suggestion
        if len(command_loaders) == 1:
            return list(command_loaders)

        module_command_loaders = []
```

src/azure-cli-core/azure/cli/core/__init__.py
```diff
@@ -489,6 +489,25 @@ def _get_extension_suppressions(mod_loaders):
...
+                extension_command_loaders.append(loader)
+            else:
+                module_command_loaders.append(loader)
+        
+        # Sort name in each command loader list
+        module_command_loaders.sort(key=lambda loader: loader.__class__.__name__)
+        extension_command_loaders.sort(key=lambda loader: loader.__class__.__name__)
+
+        # Module first, then extension
+        sorted_command_loaders = module_command_loaders + extension_command_loaders
```
jiasli, line 508: The loading order of modules and extensions matters - modules must be loaded first. 

Otherwise, an extension can maliciously or accidentally affect other modules. For example, an extension can declare 

```py
def load_arguments(self, _):
    with self.argument_context('resource') as c:
        c.ignore('_subscription')  # hide global subscription param
```

thus corrupting the global `--subscription` for all `az resource`'s commands.

src/azure-cli-core/azure/cli/core/__init__.py
```diff
@@ -489,6 +489,25 @@ def _get_extension_suppressions(mod_loaders):
...
+        module_command_loaders = []
+        extension_command_loaders = []
+
+        # Separate module and extension command loaders
+        for loader in command_loaders:
+            if loader.__module__.startswith('azext'):
+                extension_command_loaders.append(loader)
+            else:
+                module_command_loaders.append(loader)
+
```
jiasli, line 502: White spaces cause CI error:

```
Running flake8 on modules...
ERROR: /mnt/vss/_work/1/s/src/azure-cli-core/azure/cli/core/__init__.py:502:1: W293 blank line contains whitespace
1     W293 blank line contains whitespace
```

```suggestion

```

src/azure-cli-core/azure/cli/core/__init__.py
```diff
@@ -489,6 +489,29 @@ def _get_extension_suppressions(mod_loaders):
 
         return self.command_table
 
+    @staticmethod
+    def _sort_command_loaders(self, command_loaders):
```
jiasli, line 493: `self` should be removed for `@staticmethod`. Otherwise, it fails:

```
ERROR: MainCommandsLoader._sort_command_loaders() missing 1 required positional argument: 'command_loaders'
  File "/mnt/vss/_work/1/s/env/lib/python3.12/site-packages/knack/cli.py", line 233, in invoke
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/vss/_work/1/s/src/azure-cli-core/azure/cli/core/commands/__init__.py", line 564, in execute
    self.commands_loader.load_arguments(command)
  File "/mnt/vss/_work/1/s/src/azure-cli-core/azure/cli/core/__init__.py", line 530, in load_arguments
    command_loaders = self._sort_command_loaders(command_loaders)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: MainCommandsLoader._sort_command_loaders() missing 1 required positional argument: 'command_loaders'
```

```suggestion
    def _sort_command_loaders(command_loaders):
```

src/azure-cli-core/azure/cli/core/__init__.py
```diff
@@ -489,6 +489,29 @@ def _get_extension_suppressions(mod_loaders):
...
+        module_command_loaders = []
+        extension_command_loaders = []
+
+        # Separate module and extension command loaders
+        for loader in command_loaders:
+            if loader.__module__.startswith('azext'):
+                extension_command_loaders.append(loader)
+            else:
+                module_command_loaders.append(loader)
+
```
jiasli, line 506:
```suggestion

```

jiasli (approved):

src/azure-cli-core/azure/cli/core/__init__.py
```diff
@@ -504,6 +527,7 @@ def load_arguments(self, command=None):
             command_loaders = self.cmd_to_loader_map.get(command, None)
 
         if command_loaders:
+            command_loaders = self._sort_command_loaders(command_loaders)
```
jiasli, line 530: On a second thought, perhaps this line can be moved to L523 so that it only takes effect during ref doc generation?

src/azure-cli-core/azure/cli/core/__init__.py
```diff
@@ -504,6 +527,7 @@ def load_arguments(self, command=None):
             command_loaders = self.cmd_to_loader_map.get(command, None)
```
jiasli, line 527: I am not sure when a `command` can come from multiple `command_loaders`.

jiasli:
Example 5:

In `appservice` modules, `--name` has `id_part`, enabling `--ids` argument:

https://github.com/Azure/azure-cli/blob/0052ebfbeb374246a8dfdb907f93a80fd23cc21d/src/azure-cli/azure/cli/command_modules/appservice/_params.py#L92

https://github.com/Azure/azure-cli/blob/0052ebfbeb374246a8dfdb907f93a80fd23cc21d/src/azure-cli/azure/cli/command_modules/appservice/_params.py#L56-L58

In `webapp` extension, `--name` doesn't have `id_part` on `az webapp scan` scope:

https://github.com/Azure/azure-cli-extensions/blob/bf4b942f8e269f0ee4ca319c9e6333fc1dc4e37b/src/webapp/azext_webapp/__init__.py#L38-L39

```py
        with self.argument_context('webapp scan') as c:
            c.argument('name', options_list=['--name', '-n'], help='Name of the webapp to connect to')
```

AllyW (approved):

PR #27354, opened 2023-09-09
{ARM} Fix bug in deployment serialization logic

anthony-c-martin:
**Related command**
`az deployment`

**Description**
I ran into this problem when implementing #27311, and spent a huge amount of time trying to debug it. What made it especially difficult is that the outgoing HTTP logs when using the `--debug` flag are not aware of the modification.

I found a scenario where we convert the request body from string -> dict with `json.loads`, but then only conditionally change it back to string - meaning that we actually end up sending a completely invalid (non-JSON formatted) request body.

The fix is to simply use an intermediate variable and only write to `request.data` when we have correctly-formatted data.

I've also added some debug logging to hopefully make it easier for someone else to track down if anyone has the misfortune of running into a similar problem in future.

---


yonzhan:
ARM

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -409,24 +409,28 @@ def on_request(self, request):
...
 
             # The 'template' and other properties (such as 'parameters','mode'...) are spliced and encoded into the UTF-8 bytes as the request data
             # The format of the request data is: {"properties": {"parameters": {...}, "mode": "Incremental", template:{\r\n  "$schema": "...",\r\n  "contentVersion": "...",\r\n  "parameters": {...}}}
             # This is not an ordinary JSON format, but it is a JSONC format that service can deserialize
             # If not do this splicing, the request data generated by default serialization cannot be deserialized on the service side.
             # Because the service cannot deserialize the template element: "template": "{\r\n  \"$schema\": \"...\",\r\n  \"contentVersion\": \"...\",\r\n  \"parameters\": {...}}"
-            partial_request = json.dumps(http_request.data)
+            partial_request = json.dumps(modified_data)
             http_request.data = partial_request[:-2] + ", template:" + template + r"}}"
-            http_request.data = http_request.data.encode('utf-8')
```
anthony-c-martin, line 429: I removed the encoding from string -> bytes. Conversion from one to the other feels unexpected, because this pipeline is explicitly only operating on a string value (bytes are ignored on line 407).

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -409,24 +409,28 @@ def on_request(self, request):
 
         # 'request_data' has been dumped into JSON string in set_json_body() when building HttpRequest in Python SDK.
         # In order to facilitate subsequent parsing, it is converted into a dict first
-        http_request.data = json.loads(request_data)
+        modified_data = json.loads(request_data)
 
-        if http_request.data.get('properties', {}).get('template'):
```
shenglol, line 414: Out of curiosity, does this mean that deployment commands never worked if `properties` only contain `templateLink` 🤔?

shenglol (approved):

src/azure-cli/azure/cli/command_modules/resource/custom.py
anthony-c-martin: It works today because of the `plug_pipeline` parameter:

https://github.com/Azure/azure-cli/blob/a9a87c7dbc7a50ce93c7f753b814e3991a49328e/src/azure-cli/azure/cli/command_modules/resource/custom.py#L491

It's not very discoverable that this is linked to `JsonCTemplatePolicy`'s special-casing of "templateLink". In my case, I was trying to support a `.bicepparam` file linked directly to a template spec, so I ran into this bug because neither `template_uri` nor `template_spec` were being set.

My guess is that someone else ran into the same bug and added the `plug_pipeline` workaround rather than fixing the root issue.

anthony-c-martin:
@zhoxing-ms please could you take a look?

zhoxing-ms:
As this PR involves modifications to the core logic of ARM deployment, it is recommended that you rerun all relevant tests in live mode first

anthony-c-martin:
> As this PR involves modifications to the core logic of ARM deployment, it is recommended that you rerun all relevant tests in live mode first

I've run the relevant tests, and also broken out the tests owned by my team into a separate file.

zhoxing-ms:
I am so sorry, because we have been super busy with the top priority project Azure Copilot recently, and the end of September is the Mid-Autumn Festival holiday in China, I am afraid we do not have enough time and bandwidth to review this PR.
If your PR is not very urgent, we would like to postpone it to the next sprint **(the release date is 11-15)** for review, do you think this time is acceptable?

anthony-c-martin:
> I am so sorry, because we have been super busy with the top priority project Azure Copilot recently, and the end of September is the Mid-Autumn Festival holiday in China, I am afraid we do not have enough time and bandwidth to review this PR. If your PR is not very urgent, we would like to postpone it to the next sprint **(the release date is 11-15)** for review, do you think this time is acceptable?

That's fine by me! This change is really to fix developer experience - I don't think it'll have any customer impact because we already have workarounds for the bug in place.

src/azure-cli/azure/cli/command_modules/resource/custom.py
```diff
@@ -409,24 +409,29 @@ def on_request(self, request):
 
         # 'request_data' has been dumped into JSON string in set_json_body() when building HttpRequest in Python SDK.
         # In order to facilitate subsequent parsing, it is converted into a dict first
-        http_request.data = json.loads(request_data)
+        modified_data = json.loads(request_data)
```
zhoxing-ms, line 412: I would like to confirm that this is just replacing the variable from `http_request.data` to `modified_data` to improve the readability of the code, but there is no actual logical modification, right?
If so, could you please rerecord the `DeploymentTest` tests in live mode? Because the modified code here is too core and versatile, I suggest that we conduct do live tests to prevent unexpected situations

anthony-c-martin: This was deliberate - there **is a logical modification** - to fix a bug. The earlier logic was essentially doing this:

```python
http_request.data = json.loads(request_data)
if http_request.data.get('properties', {}).get('template'):
    ...
    http_request.data = http_request.data.encode('utf-8')
```

The bug that this PR is fixing is that if the `if` condition doesn't match (e.g. if there's no `.properties.template`), `http_request.data` is left in an invalid state.

My fix is essentially making sure that we only modify `http_request.data` inside the `if` condition, and only with valid properties.

anthony-c-martin: If you review the changes, you should find that this is **the only** logical change (aside from adding logging).

Everything else is a refactor, to use intermediate variables instead of using `http_request.data` as if it were an intermediate variable.

anthony-c-martin: I've re-run `azdev test --live DeploymentTest` and checked in the result.

zhoxing-ms: Make sense! Thanks for your careful and thoughtful contribution

zhoxing-ms:
> I've run the relevant tests, and also broken out the tests owned by my team into a separate file.

Could you recording some related tests in live mode, and then upload them to this PR?

anthony-c-martin:
> @anthony-c-martin Could you recording some related tests in live mode, and then upload them to this PR?

I actually ended up reverting this change as it was causing frequent merge conflicts. Will update the PR description.

zhoxing-ms:
@anthony-c-martin OK, since the `az group deployment create` command in test `test_group_deployment_lite` is a deprecated command, could you please also re-record the yaml file for `test_resource_group_deployment` test?

zhoxing-ms:
Please note that we are launching the release for this sprint this week. Please resolve all comments by Tomorrow, otherwise the release of this PR will have to be postponed to the next sprint **(on 12-05)**

anthony-c-martin:
> @anthony-c-martin OK, since the `az group deployment create` command in test `test_group_deployment_lite` is a deprecated command, could you please also re-record the yaml file for `test_resource_group_deployment` test?

Done!

zhoxing-ms (approved):

PR #27360, opened 2023-09-11
[ARM] `az deployment`: Add support for determining type of parameters whose definition uses a $ref

jeskew:
**Related command**
`az deployment`

**Description**
Resolves https://github.com/Azure/bicep/issues/11747.

ARM recently added support for reusing parameter schemas within a template. As part of this change, a template parameter is no longer guaranteed to have a `type` property but may have a `$ref` property instead. This PR will follow `$ref` pointers to find the base type of a parameter, which allows users to supply json-formatted parameters as arguments, even if those parameters use type references.


yonzhan:
ARM

jeskew:
@zhoxing-ms Gentle ping on this.

zhoxing-ms:
@jeskew Could you please add some ScenarioTest for this PR? 
In addition, since this PR modification affects the core command `az deployment`, it will have a wide range of impacts. Could you help re-record all the tests related to `az deployment` in live mode?

jeskew:
@zhoxing-ms This change shouldn't affect the behavior of any existing test, so there should be no need to re-record. (Are the live tests not executed as part of CI?)

Also, could you point me towards any documentation on the CLI's scenario tests and how to add to them? I added several tests to the suites within the `src/azure-cli/azure/cli/command_modules/resource/tests` folder but couldn't find a separate test suite to augment.

zhoxing-ms:
@jeskew OK, so you confirm that the modification of this PR does not affect the requests to the ARM service, only the client's parsing of template parameters, right?

zhoxing-ms:
@jeskew Could you please resolve these conflicts?

zhoxing-ms:
@jeskew Could you please resolve these CI issues? Please note that we are launching the release for this sprint this week. Please resolve all comments by Tomorrow, otherwise the release of this PR will have to be postponed to the next sprint **(on 12-05)**

zhoxing-ms (approved):

PR #27407, opened 2023-09-15
[App Service] `az functionapp create`: Add new parameter `--workspace` to support creating workspace-based app insights components

kamperiadis:
**Related command**
`az functionapp create`

**Description**
Currently, during the function app creation flow, we create a classic app insights component. However, this will no longer be supported in February 2024 so we have to create workspace-based app insights components instead. The Azure Portal already does this so we need to implement the same behavior in the Azure CLI.

When creating a function app in the Azure Portal, a resource group named `DefaultResourceGroup-<RegionCode>` is created and then a workspace named `DefaultWorkspace-{SubscriptionId>-<RegionCode>` is created inside that default resource group. Then that default workspace is attached to the application insights component that gets created.

Relevant documentation: https://learn.microsoft.com/en-us/azure/azure-monitor/app/convert-classic-resource

**Testing Guide**
1. Create function app with `az functionapp create` command. I have also added a new param `--workspace` that accepts the name of an existing log analytics workspace resource to be used for the application insights component that will be created.
2. Check that the app insights component has a workspace resource associated with it. This can be confirmed either through the Azure Portal or by running `az monitor app-insights component show` and looking for `workspaceResourceId`

---


yonzhan:
AppService

khkh-ms (approved):
LGTM

yanzhudd:
Please fix the CI.

kamperiadis:
@yanzhudd I have fixed the CI failed items. Can you please review the PR again? Thank you.

src/azure-cli/azure/cli/command_modules/appservice/tests/latest/test_functionapp_commands.py
```diff
@@ -1082,8 +1082,9 @@ def test_functionapp_with_default_distributed_tracing(self, resource_group, stor
         self.cmd('functionapp create -g {} -n {} -p {} -s {} --runtime java --functions-version 4'
                  .format(resource_group, functionapp_name, plan, storage_account))
 
-        app_set = self.cmd('functionapp config appsettings list -g {} -n {}'.format(resource_group,
-                                                                                    functionapp_name)).get_output_in_json()
+        self.cmd('functionapp config appsettings set -g {} -n {} --settings FOO=BAR'.format(resource_group, functionapp_name))
```
kamperiadis, line 1085: For this `test_functionapp_with_default_distributed_tracing` unit test, if I run it `--live` then I do not get any errors but for some reason, if I do not have this line, then it does not find the appsetting in L1088 even though it is listed in the recording. Do you know why this succeeds when running the test `--live` but then fails when rerunning the recording without `--live`?

zhoxing-ms: In addition, we will launch the release for Ignite Sprint the day after tomorrow. If this issue is only related to testing and does not block merge PR, then we can merge this PR first and investigate the testing issue in the future

kamperiadis: @zhoxing-ms @yanzhudd Thank you. Yes, I do not think this should block merging the PR. The issue I have is only related to testing when using the recording (even though the recording has the app settings I am validating for), but it works when I test it manually or when I run the tests in `--live` mode. As such, can we please merge the PR? Thank you.

zhoxing-ms (approved):

PR #27476, opened 2023-09-27
[Profile] `az account get-access-token`: Return `expires_on` as POSIX timestamp

### LINKED ISSUE #19700: `az account get-access-token`: Use epoch `expiresOn`/`expires_on`
## Context

Currently, the `expiresOn` property in the output of `az account get-access-token` is a string representing the local time:

```
{
  "accessToken": "eyJ0eXAiOiJ...",
  "expiresOn": "2021-09-27 16:16:30.921462",  <<<
  "subscription": "0b1f6471-1bf0-4dda-aec3-cb9272f09590",
  "tenant": "54826b22-38d6-4fb2-bad9-b7b93a3e9c5a",
  "tokenType": "Bearer"
}
```

This inherits from the ADAL behavior: https://github.com/AzureAD/azure-activedirectory-library-for-python/blob/35f9e003bed47936a1df3c6eb696691dc1fa91c3/adal/oauth2_client.py#L187

This causes various time computation problems and difficulties, especially for [daylight saving time](https://en.wikipedia.org/wiki/Daylight_saving_time), such as 

- https://github.com/Azure/azure-sdk-for-python/pull/14331
- PEP 495 -- Local Time Disambiguation: https://www.python.org/dev/peps/pep-0495/

Other tools or services are already using an integer to represent the [epoch time](https://en.wikipedia.org/wiki/Unix_time):

- Azure Python SDK uses integer `expires_on`: https://github.com/Azure/azure-sdk-for-python/blob/844e16db0abc553ab1adf104128cbf0e223af189/sdk/core/azure-core/azure/core/credentials.py#L14
- MSAL internally uses integer `expires_on`: https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/db6f001060e0adcc1edccac1a86ec05bd59f15f0/msal/token_cache.py#L176
- Managed identity's result uses integer `expires_on`: https://github.com/Azure/azure-cli/blob/14cc787d0f58bc649d402b486fdecc5625eee9ac/src/azure-cli-core/azure/cli/core/adal_authentication.py#L116

This approach is universal and eliminates all necessary complexity.

## Proposal

There are possible approaches:

1. Make a BREAKING CHANGE to convert the `expiresOn` in the output of `az account get-access-token` to an integer representing the epoch time. This can be done during MSAL migration: https://github.com/Azure/azure-cli/issues/18944
2. Introduce another property `expires_on`. This will not be a BREAKING CHANGE, but introduces **naming inconsistency** as other properties uses camelCase, but `expires_on` is snake_case.
3. Introduce another properties like `expiresOnEpoch`, but naming like this is non-conventional.

jiasli:
Close #19700

**Related command**
`az account get-access-token`

**Description**
Returning `expiresOn` as a local datetime is a legacy from the ADAL age. It causes lots of unnecessary complexities. See https://github.com/Azure/azure-cli/issues/19700 for detailed explanation.

This PR makes `az account get-access-token` return `expires_on` as POSIX timestamp which has been supported long ago by Azure CLI Core during the MSAL migration (https://github.com/Azure/azure-cli/pull/19853).

However, as shown in https://github.com/Azure/azure-cli/issues/19700#issuecomment-1729040872, downstream applications that only support 32-bit integer precision timestamp will encounter error in year 2038, but that's not something we can control.

**Testing Guide**
```
az account get-access-token
```

Before:

```
{
  "accessToken": "...",
  "expiresOn": "2023-10-31 21:59:10.000000",
  "subscription": "...",
  "tenant": "...",
  "tokenType": "Bearer"
}
```

After:

```
{
  "accessToken": "...",
  "expiresOn": "2023-10-31 21:59:10.000000",
  "expires_on": 1698760750,
  "subscription": "...",
  "tenant": "...",
  "tokenType": "Bearer"
}
```

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/profile/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/profile/_help.py
@@ -71,6 +71,10 @@ short-summary: Get a token for utilities to access Azure.
 long-summary: >
     The token will be valid for at least 5 minutes with the maximum at 60 minutes.
     If the subscription argument isn't specified, the current account is used.
+
+
+    In the output, `expires_on` represents a POSIX timestamp and `expiresOn` represents a local datetime.
+    It is recommended for downstream applications to use `expires_on` because it is in UTC.
 examples:
     - name: Get an access token for the current account
       text: >
--- a/src/azure-cli/azure/cli/command_modules/profile/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/profile/custom.py
@@ -69,7 +69,7 @@ def get_access_token(cmd, subscription=None, resource=None, scopes=None, resourc
     result = {
         'tokenType': creds[0],
         'accessToken': creds[1],
-        # 'expires_on': creds[2].get('expires_on', None),
+        'expires_on': creds[2]['expires_on'],
         'expiresOn': creds[2]['expiresOn'],
         'tenant': tenant
     }
--- a/src/azure-cli/azure/cli/command_modules/profile/tests/latest/test_profile_custom.py
+++ b/src/azure-cli/azure/cli/command_modules/profile/tests/latest/test_profile_custom.py
@@ -37,7 +37,15 @@ class ProfileCommandTest(unittest.TestCase):
         cmd = mock.MagicMock()
         cmd.cli_ctx = DummyCli()
 
-        get_raw_token_mock.return_value = (['bearer', 'token123', {'expiresOn': '2100-01-01'}], 'sub123', 'tenant123')
+        timestamp = 1695270561
+        datetime_local = '2023-09-21 04:29:21.000000'
+
+        token_entry = {
+            'accessToken': 'token123',
+            'expires_on': timestamp,
+            'expiresOn': datetime_local
+        }
+        get_raw_token_mock.return_value = (('bearer', 'token123', token_entry), 'sub123',  'tenant123')
 
         result = get_access_token(cmd)
 
@@ -46,7 +54,8 @@ class ProfileCommandTest(unittest.TestCase):
         expected_result = {
             'tokenType': 'bearer',
             'accessToken': 'token123',
-            'expiresOn': '2100-01-01',
+            'expires_on': timestamp,
+            'expiresOn': datetime_local,
             'subscription': 'sub123',
             'tenant': 'tenant123'
         }
@@ -55,8 +64,7 @@ class ProfileCommandTest(unittest.TestCase):
         # assert it takes customized resource, subscription
         resource = 'https://graph.microsoft.com/'
         subscription_id = '00000001-0000-0000-0000-000000000000'
-        get_raw_token_mock.return_value = (['bearer', 'token123', {'expiresOn': '2100-01-01'}], subscription_id,
-                                           'tenant123')
+        get_raw_token_mock.return_value = (('bearer', 'token123', token_entry), subscription_id, 'tenant123')
         result = get_access_token(cmd, subscription=subscription_id, resource=resource)
         get_raw_token_mock.assert_called_with(mock.ANY, resource, None, subscription_id, None)
 
@@ -67,12 +75,13 @@ class ProfileCommandTest(unittest.TestCase):
 
         # test get token with tenant
         tenant_id = '00000000-0000-0000-0000-000000000000'
-        get_raw_token_mock.return_value = (['bearer', 'token123', {'expiresOn': '2100-01-01'}], None, tenant_id)
+        get_raw_token_mock.return_value = (('bearer', 'token123', token_entry), None, tenant_id)
         result = get_access_token(cmd, tenant=tenant_id)
         expected_result = {
             'tokenType': 'bearer',
             'accessToken': 'token123',
-            'expiresOn': '2100-01-01',
+            'expires_on': timestamp,
+            'expiresOn': datetime_local,
             'tenant': tenant_id
         }
         self.assertEqual(result, expected_result)


yonzhan:
Profile

cataggar (approved):

yonzhan (approved):

src/azure-cli/azure/cli/command_modules/profile/_help.py
```diff
@@ -71,6 +71,11 @@
 long-summary: >
     The token will be valid for at least 5 minutes with the maximum at 60 minutes.
     If the subscription argument isn't specified, the current account is used.
+
+
+    In the output, `expires_on` represents a POSIX timestamp and `expiresOn` represents a local datetime.
+    It is recommended for downstream applications to use `expires_on`, because `expiresOn` cannot express "fold"
+    when Daylight Saving Time ends.
```
jiasli, line 78: I am not sure whether we should give more explanation on the "fold" concept. For example, add a link to https://peps.python.org/pep-0495/.

dbradish-microsoft: @jiasli, I have read the two links to the Issues you provided in "Conversation".
1 - What is the engineering's team SOP for providing links to URLs outside of Azure in our source code help? What method are you using to check for and correct broken links? I do not suggest deviating from your defined best practices.
2 - I generally lean toward providing all the clarity I can.  That would include your full 2nd sentence plus the link: "It is recommended ... end. For more information on "fold", see [PEP 495 – Local Time Disambiguation](https://peps.python.org/pep-0495/)."

From a doc's perspective, we have 10 docs that use `az account get-access-token` 26 times. I was looking to see if anyone did anything with `expiresOn` but they are only showing it in ```ouput code blocks.

dbradish-microsoft (changes requested):
Requested change: Add the PEP 495 link if it aligns with engineering URL best practices.

src/azure-cli/azure/cli/command_modules/profile/_help.py
jiasli: > What is the engineering's team SOP for providing links to URLs outside of Azure in our source code help?

We don't have Standard Operating Procedure (SOP) for URLs outside of Microsoft. If a link is broken, then it is broken.

jiasli: Then I would suggest we link to our GitHub issue which we can control the link and the content:

```suggestion
    In the output, `expires_on` represents a POSIX timestamp and `expiresOn` represents a local datetime.
    It is recommended for downstream applications to use `expires_on`. For more information, see
    https://github.com/Azure/azure-cli/issues/19700.
```

dbradish-microsoft: @jiasli / @Dcaro, I would prefer to take our customer to PEP 495 and risk the broken link instead of a Github issue. We do have a broken links report that we can use to monitor problems that become autogenerated content.

evelyn-ys (approved):

src/azure-cli/azure/cli/command_modules/profile/_help.py
jiasli: I noticed we currently don't have a dedicated conceptual doc for `az account get-access-token`. https://learn.microsoft.com/en-us/cli/azure/msal-based-azure-cli is the only place where `az account get-access-token` is mentioned. I prefer we don't make the help message too verbose for now until we get user feedback regarding whether more information is needed for the "fold" concept:

```suggestion
    In the output, `expires_on` represents a POSIX timestamp and `expiresOn` represents a local datetime.
    It is recommended for downstream applications to use `expires_on`, because it is in UTC.
```

If we do need to add more info to the doc, we can add something like:

> Starting from Azure CLI 2.54.0, `az account get-access-token` returns `expires_on` besides `expiresOn` for the token expiration time. `expires_on` represents a POSIX timestamp and `expiresOn` represents a local datetime. It is recommended for downstream applications to use `expires_on`, because it is in UTC. On the other hand, `expiresOn` cannot express "fold" when Daylight Saving Time ends. This can cause problems in countries or regions where Daylight Saving Time is adopted. For more information on "fold", see [PEP 495 – Local Time Disambiguation](https://peps.python.org/pep-0495/).

dbradish-microsoft: @jiasli, I've created a work item to add this information to [manage-azure-subscriptions-azure-cli](https://learn.microsoft.com/en-us/cli/azure/manage-azure-subscriptions-azure-cli). Also I've made a 2nd work item to notify the ms.authors of the 41 conceptual docs that contain this command.

Removing both link options is fine with me. The purpose of this message is to provide a recommendation to the customer, not explain "fold", time zones, and daylight savings time. Succinct is better.

Please remove the comma before the ", because".

src/azure-cli/azure/cli/command_modules/profile/_help.py
```diff
@@ -71,6 +71,10 @@
 long-summary: >
     The token will be valid for at least 5 minutes with the maximum at 60 minutes.
     If the subscription argument isn't specified, the current account is used.
+
+
+    In the output, `expires_on` represents a POSIX timestamp and `expiresOn` represents a local datetime.
+    It is recommended for downstream applications to use `expires_on`, because it is in UTC.
```
jiasli, line 77:
```suggestion
    It is recommended for downstream applications to use `expires_on` because it is in UTC.
```

jiasli, line 77:
```suggestion
    It is recommended for downstream applications to use `expires_on` because it is in UTC.
```

dbradish-microsoft (approved):

evelyn-ys (approved):

bebound (approved):

jiasli:
This PR has been released in Azure CLI 2.54.0.

Whether using `expiresOn` causes failure really depends on how the downstream program uses the access token and `expiresOn` returned by Azure CLI.

One thing is for sure: The access token returned by `az account get-access-token` is guaranteed to be valid at the time of invocation. Downstream program can use it without validating `expiresOn`. `expiresOn` is used by downstream program that has a caching mechanism.

`expiresOn` faces Daylight Saving Time problem (https://github.com/Azure/azure-cli/issues/19700#issuecomment-1727797158):

1. `expiresOn` can only express the first hour when Daylight Saving Time ends, not the folded(second) hour. For example, `1699172100`/`2023-11-05 01:15:00-07:00`/`2023-11-05 08:15:00 UTC` is expressed as `2023-11-05 01:15:00`.
2. When spacetime treads into the folded hour (`1699175700`/`2023-11-05 01:15:00-08:00`/`2023-11-05 09:15:00 UTC`), the `expiresOn` still represents a time 1 hour earlier (`2023-11-05 01:15:00`) due to the lack of time zone and fold information. Downstream program will think the token has expired and request a new one. 
3. However, the `expiresOn` of the new token returned by Azure CLI is still 1 hour earlier (`2023-11-05 01:15:00`), downstream program will think the new token has also expired. It may either fail or go into an infinite loop.

If you want better compatibility with old versions of Azure CLI, you may allow falling back to `expiresOn`, but the downstream program would suffer from the Daylight Saving Time problem. If you want robustness and code simplicity, only use `expires_on` and err out when it is missing.

Moazzem-Hossain-pixel (commented):
Need support

PR #27489, opened 2023-09-27
{docs} reference guidelines - removed markdown syntax not supported in this repo

dbradish-microsoft:
- reformatted checklist UI as [!div class="checklist"] is not supported in this repo
- fixed site-relative links not supported in this repo
- Substituted contractions where possible
- Slight changes to `Minimum content` H2 syntax


yonzhan:
docs

dbradish-microsoft:
#sign-off

doc/reference_doc_guidelines.md
```diff
@@ -188,23 +186,19 @@ examples:
                --ssh-key-values @key-file
   ```
 
-- Not all examples can be shell agnostic due to the quoting differences between Bash and PowerShell. For examples that only run in Bash, provide a second example for PowerShell. If you are testing examples in PowerShell, here are some considerations:
-  - [Quoting issues with PowerShell](https://github.com/Azure/azure-cli/blob/dev/doc/quoting-issues-with-powershell.md)
```
jsntcy, line 192: Did you remove them because they are already at the beginning of this file?

jsntcy:
Overall LGTM.

doc/reference_doc_guidelines.md
dbradish-microsoft: @jsntcy, yes

jsntcy (approved):

Issue #26433, opened 2023-05-12
Disabling instance discovery or validate authority


TheOnlyWei:
**Related command**
`az login`

**Is your feature request related to a problem? Please describe.**
`az login` currently doesn't allow disabling instance discovery or validate authority required for some on-premise environments.

**Some investigation info:**
The msal used by latest Azure CLI is 1.20.0, which allows disabling of instance_discovery:
https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/14cbf591948ffb23eb64a675d81e4fa35112d1d9/msal/application.py#L180

I searched the azure cli repo and the only references to instance discovery and validate authority are in the msal package above that was imported.

Also found this issue on this topic:
https://github.com/Azure/azure-cli/issues/12938#issuecomment-611996399
From what jiasli said, it seems that disabling instance discovery or validate authority isn't supported yet for Azure CLI, but that was 3 years ago.

azure-identity Python package introduced validate_authority for version 1.9.0b1 (2022-03-08), which was changed to instance_discovery in 1.13.0b1 (2023-01-10):
https://pypi.org/project/azure-identity/#description
But from the code trace of an error caused by instance_discovery or validate_authority being true:
```
vmadmin@vmadmin-Virtual-Machine:~/Downloads$ az login
The command failed with an unexpected error. Here is the traceback:
invalid_instance: The authority you provided, https://login.devfabric.azs.microsoft.com/organizations, is not whitelisted. If it is indeed your legit customized domain name, you can turn off this check by passing in validate_authority=False
Traceback (most recent call last):
  File "/opt/az/lib/python3.10/site-packages/knack/cli.py", line 233, in invoke
    cmd_result = self.invocation.execute(args)
  File "/opt/az/lib/python3.10/site-packages/azure/cli/core/commands/__init__.py", line 663, in execute
    raise ex
  File "/opt/az/lib/python3.10/site-packages/azure/cli/core/commands/__init__.py", line 726, in _run_jobs_serially
    results.append(self._run_job(expanded_arg, cmd_copy))
  File "/opt/az/lib/python3.10/site-packages/azure/cli/core/commands/__init__.py", line 697, in _run_job
    result = cmd_copy(params)
  File "/opt/az/lib/python3.10/site-packages/azure/cli/core/commands/__init__.py", line 333, in __call__
    return self.handler(*args, **kwargs)
[...]

PR #27494, opened 2023-09-28
[Core] Support disabling instance discovery

jiasli:
Close https://github.com/Azure/azure-cli/issues/26433

**Related command**
`az login`

**Description**
Some Azure Stack cloud uses an AAD endpoint that is not well known. The endpoint can't be discovered with `https://login.microsoftonline.com/common/discovery/instance` either.

Running `az login` in such cloud will result in a failure:

```
ValueError: invalid_instance: The authority you provided, https://login.devfabric.azs.microsoft.com/
organizations, is not whitelisted. If it is indeed your legit customized domain name, you can turn 
off this check by passing in validate_authority=False
```

**Testing Guide**

Register a cloud with private AAD endpoint `https://login.devfabric.azs.microsoft.com/` and log in:

```
az cloud register -n myazurestack --endpoint-resource-manager https://management.azure.com/ --endpoint-active-directory https://login.devfabric.azs.microsoft.com/
az cloud set -n myazurestack
az login --debug
```

`az login` fails with

```
ValueError: invalid_instance: The authority you provided, https://login.devfabric.azs.microsoft.com/
organizations, is not whitelisted. If it is indeed your legit customized domain name, you can turn 
off this check by passing in validate_authority=False
```

Then disable instance discovery with `az config`:

```sh
az config set core.instance_discovery=false
# To unset it:
# az config unset core.instance_discovery
```

Instance discovery can also be disabled using env var:

```sh
# Bash
export AZURE_CORE_INSTANCE_DISCOVERY=false
# PowerShell
$env:AZURE_CORE_INSTANCE_DISCOVERY="false"
```

`az login` should now fail with

```
requests.exceptions.ConnectionError: HTTPSConnectionPool(host='login.devfabric.azs.microsoft.com', port=443): 
Max retries exceeded with url: /organizations/v2.0/.well-known/openid-configuration (Caused by 
NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x000001F5D27F3100>: Failed to establish a 
new connection: [Errno 11001] getaddrinfo failed'))
```

This is expected as we can't resolve `login.devfabric.azs.microsoft.com` from public internet.

**History Notes**
[Core] Support disabling instance discovery by running `az config set core.instance_discovery=false`

### DIFF
--- a/src/azure-cli-core/azure/cli/core/_profile.py
+++ b/src/azure-cli-core/azure/cli/core/_profile.py
@@ -861,5 +861,8 @@ def _create_identity_instance(cli_ctx, *args, **kwargs):
     from .telemetry import set_broker_info
     set_broker_info(allow_broker=allow_broker)
 
+    # PREVIEW: In Azure Stack environment, use core.instance_discovery=false to disable MSAL's instance discovery.
+    instance_discovery = cli_ctx.config.getboolean('core', 'instance_discovery', True)
+
     return Identity(*args, encrypt=encrypt, use_msal_http_cache=use_msal_http_cache, allow_broker=allow_broker,
-                    **kwargs)
+                    instance_discovery=instance_discovery, **kwargs)
--- a/src/azure-cli-core/azure/cli/core/auth/identity.py
+++ b/src/azure-cli-core/azure/cli/core/auth/identity.py
@@ -54,7 +54,7 @@ class Identity:  # pylint: disable=too-many-instance-attributes
     _service_principal_store_instance = None
 
     def __init__(self, authority, tenant_id=None, client_id=None, encrypt=False, use_msal_http_cache=True,
-                 allow_broker=None):
+                 allow_broker=None, instance_discovery=None):
         """
         :param authority: Authentication authority endpoint. For example,
             - AAD: https://login.microsoftonline.com
@@ -70,6 +70,7 @@ class Identity:  # pylint: disable=too-many-instance-attributes
         self._encrypt = encrypt
         self._use_msal_http_cache = use_msal_http_cache
         self._allow_broker = allow_broker
+        self._instance_discovery = instance_discovery
 
         # Build the authority in MSAL style
         self._msal_authority, self._is_adfs = _get_authority_url(authority, tenant_id)
@@ -98,6 +99,7 @@ class Identity:  # pylint: disable=too-many-instance-attributes
             "authority": self._msal_authority,
             "token_cache": Identity._msal_token_cache,
             "http_cache": Identity._msal_http_cache,
+            "instance_discovery": self._instance_discovery,
             # CP1 means we can handle claims challenges (CAE)
             "client_capabilities": None if "AZURE_IDENTITY_DISABLE_CP1" in os.environ else ["CP1"]
         }


yonzhan:
Core

jiasli:
The error shown when `instance_discovery` is enabled still needs some polishing.

Currently, the error is raised by MSAL:

```
ValueError: invalid_instance: The authority you provided, https://login.devfabric.azs.microsoft.com/
organizations, is not whitelisted. If it is indeed your legit customized domain name, you can turn 
off this check by passing in validate_authority=False
```

Problems:

1. It is a `ValueError`. Without a MSAL-specific error type, it is difficult for Azure CLI to capture and handle it: https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/482

3. According to MSAL developer
   > Indeed, `instance_discovery` and `validate_authority` have some overlap, but not identical. The future is to use `instance_discovery`.
   
   MSAL may need to adjust the error message and suggest the user to use `instance_discovery`.

4. We need to translate the error message into Azure CLI terminology:
   > you can turn off this check by running `az config set core.instance_discovery=false`

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -861,5 +861,8 @@ def _create_identity_instance(cli_ctx, *args, **kwargs):
     from .telemetry import set_broker_info
     set_broker_info(allow_broker=allow_broker)
 
+    # PREVIEW: In Azure Stack environment, use core.instance_discovery=false to disable MSAL's instance discovery.
+    instance_discovery = cli_ctx.config.getboolean('core', 'instance_discovery', True)
```
rayluo, line 865:
```suggestion
    instance_discovery = cli_ctx.config.getboolean('core', 'instance_discovery', None)
```

I would recommend let the default value match [MSAL's default value](https://msal-python.readthedocs.io/en/latest/#msal.ClientApplication.params.instance_discovery), rather than hardcoding a `True`. The benefit is, when/if MSAL changes the default behavior to `False` in the future, Azure CLI can automatically pick up the new behavior.

jiasli:
Hi @keystroke, thanks for the suggestion. We did consider making `instance_discovery` a per-cloud setting with our PMs and related Azure Stack teams. This would require changing the cloud schema by introducing a new field in `az cloud register` and `az cloud show`:

https://github.com/Azure/azure-cli/blob/3403856cac8f85b3e0ba65bc5c6fa6185e326027/src/azure-cli-core/azure/cli/core/cloud.py#L262-L275

However, we didn't come to a conclusion on this. Let me loop you in the internal email thread.

> Though, I would suggest folks never use an alias with AzureCLI unless it permanently caches the result of the instance discovery after the first run (why setup AZ cloud to perform this fetch every single time?).

Could you kindly provide more information on "alias", like which commands you run and what the expected behavior is?

keystroke:
@jiasli the "alias" feature for instance discovery is to configure cloud with different authority endpoint (I think primarily for b2c / enterprise flows and such) so folks may use "login.contoso.com" instead "login.microsoftonline.com" for the "base authority" if they've configured that in AAD. This also feeds into MSAL under the hood sharing tokens for caching so it could see the user has signed-in to Visual Studio or some other Microsoft product already, using one such alias, and thus has a token for a specific AAD instance already cached, and then MSAL can recognize that "login.contoso.com" is an alias for the target region and use the cached token.

e.g. You can give MSAL "login.microsoft.com" and see it make this call and get the following response (note the aliases returned):
https://login.microsoft.com/common/discovery/instance?api-version=1.1&authorization_endpoint=https://login.microsoft.com/organizations/oauth2/v2.0/authorize
```json
{
    "tenant_discovery_endpoint": "https://login.microsoft.com/organizations/v2.0/.well-known/openid-configuration",
    "api-version": "1.1",
    "metadata": [
        {
            "preferred_network": "login.microsoftonline.com",
            "preferred_cache": "login.windows.net",
            "aliases": [
                "login.microsoftonline.com",
                "login.windows.net",
                "login.microsoft.com",
                "sts.windows.net"
            ]
        },
        {
            "preferred_network": "login.partner.microsoftonline.cn",
            "preferred_cache": "login.partner.microsoftonline.cn",
            "aliases": [
                "login.partner.microsoftonline.cn",
                "login.chinacloudapi.cn"
            ]
        },
        {
            "preferred_network": "login.microsoftonline.de",
            "preferred_cache": "login.microsoftonline.de",
            "aliases": [
                "login.microsoftonline.de"
            ]
        },
        {
            "preferred_network": "login.microsoftonline.us",
            "preferred_cache": "login.microsoftonline.us",
            "aliases": [
                "login.microsoftonline.us",
                "login.usgovcloudapi.net"
            ]
        },
        {
            "preferred_network": "login-us.microsoftonline.com",
            "preferred_cache": "login-us.microsoftonline.com",
            "aliases": [
                "login-us.microsoftonline.com"
            ]
        }
    ]
}
```

However, given @rayluo comment about default behavior changing (and my doubts as to whether this sort of feature is actually used in the real world for such customers who have a given mix of these on-prem and public "aliased" clouds) having the config setting be device-wide is perhaps fine despite my original comment (and if the default value changes then the config would not need to be disabled at all). I suspect this would be used in other contexts, not with azure cli which is more likely to be setup with the "generic" cloud instance enums even in the case of these b2c / enterprise flows. Identity team may have more insight / telemetry to suggest otherwise.

That is to say, if this disabling config cannot easily be made part of the specific cloud instance config, the global / process-wide switch may be sufficient.

src/azure-cli-core/azure/cli/core/_profile.py
jiasli: `getboolean` doesn't support `None` as fallback. The fallback value must be a `bool`: https://github.com/microsoft/knack/blob/e496c9590792572e680cb3ec959db175d9ba85dd/knack/config.py#L146-L147

jiasli:
I agree it may be rare for users to have a mix of **on-prem clouds** and **aliased public clouds** in the same Azure CLI context. For now, let's make `instance_discovery` a global setting so that we can quickly unblock Azure Stack use cases.

src/azure-cli-core/azure/cli/core/_profile.py
rayluo: Well, 5 years ago, I used to also design boolean flags as boolean type (such as [MSAL's old `validate_authority`](https://github.com/AzureAD/microsoft-authentication-library-for-python/blame/dev/msal/application.py#L182)). Nowadays, [all new boolean-like parameters accept a `None` as default value](https://github.com/AzureAD/microsoft-authentication-library-for-python/blame/dev/msal/application.py#L194-L196). In other words, the *"default: let the upstream library do whatever it chooses to do"* becomes a legit input. The benefit of this new approach is that it no longer forces the downstream libraries/apps to pick a side, when they shouldn't.

Perhaps you can consider building a new `getboolean2(...)` which allows `None` as default, and potentially transition to replace the old `getboolean()` with the new one. `None` is equivalent to `False` when a real boolean is needed, anyway.

jiasli: This requires changing the underlying framework Knack which is a little bit heavy. Let's keep it the way it is (for now).

evelyn-ys (dismissed):

bebound (dismissed):

bebound (approved):

evelyn-ys (approved):

PR #27544, opened 2023-10-09
{Core} Remove restrictions for parameters `zones_type` and `zone_type` to support more than 3 zones

yanzhudd:
**Related command**

**Description**

Service team expects the removal for the following reason (mail: AZ CLI Zone 4 Support):

> More than 3 AZ’s (in short: >3AZ’s) is a company-wide initiative. As a company, we’re working to enable Azure regions to expand in size beyond the limits that a size and distance of 3 zones allows, and allow for wider expansion. This allows Azure to bring more capacity into a region, in a form of additional Availability Zones, yet preserving promises that we as a company have made to external users. Those promises are in a form of resiliency but also latency governing communication within a single zone.
>  
> Of course, the initiative can only be successful if Azure services are able to take advantage of the new capacity, and allow end users (3P’s) to use that capacity. That’s why each service team has been reviewing their architecture and adjusting where needed to support 4th++ Availability Zones (4th, 5th, 6th), and enabling this functionality in their UI/UX. While we’re testing this, we’ve realized that client CLI’s has its own logic that hardcoded 3 zones permanently.
>
> We need your help to remove the hardcoded value of 3 zones in a core CLI module, and allow the client module to pass through any zone number, letting the service RP Api’s make a decision. 

**Testing Guide**

**History Notes**

---


yonzhan:
Core

yanzhudd:
Hi @jiasli, could you please help review this change? @Jacekey23 has awared this change and reviewed this PR.

Jacekey23:
Notes:
the change is relatively minor and just synced with Yan about timeline. Targeted for current sprint if no underlying impact. If there is, we may plan it before early Dec, Thx.

jiasli:
> Hi @jiasli, could you please help review this change? @Jacekey23 has awared this change and reviewed this PR.

I have little knowledge regarding this change. @zhoxing-ms is also the core owner and can approve this PR.

jiasli (dismissed):

zhoxing-ms:
Please let the server team also review this PR and confirm the following two questions:
1. If the CLI side removes the verification of zone restrictions, is the error message provided by the server side user-friendly and meets expectations when the customer enters the illegal zones?
2. There are multiple Resource Providers that support edge zones. Can the restrictions of zones on the CLI side be removed for all the Resource Providers?

yanzhudd:
> Please let the server team also review this PR and confirm the following two questions:
> 
> 1. If the CLI side removes the verification of zone restrictions, is the error message provided by the server side user-friendly and meets expectations when the customer enters the illegal zones?
> 2. There are multiple Resource Providers that support edge zones. Can the restrictions of zones on the CLI side be removed for all the Resource Providers?

Since PowerShell and Terraform have no restriction for the edge zones, I think it makes sense to remove the restriction.

zhoxing-ms:
> Since PowerShell and Terraform have no restriction for the edge zones, I think it makes sense to remove the restriction.

OK, sounds good

zhoxing-ms:
In fact, I see that other modules also have restrictions on zones, and they do not use the unified code in the azure-cli-core. Can they be removed together?
such as: https://github.com/Azure/azure-cli/blob/5bca3f5a64ebed43a826e0c668857bbe7604d8c8/src/azure-cli/azure/cli/command_modules/network/_params.py#L59-L63C11

yanzhudd:
> In fact, I see that other modules also have restrictions on zones, and they do not use the unified code in the azure-cli-core. Can they be removed together? such as: https://github.com/Azure/azure-cli/blob/5bca3f5a64ebed43a826e0c668857bbe7604d8c8/src/azure-cli/azure/cli/command_modules/network/_params.py#L59-L63C11

yeah, this change only applies to the modules which import `zones_type` or `zone_type` from core directly. 
In my opinion, this change is sufficient for the current need (Compute and AKS modules), it is not necessary to change all modules which do not use unified code of the core module. We could change the logic from customed defining to importing core code directly case by case in the future. How do you think?

necusjz (approved):

Issue #23614, opened 2022-08-22
CLI | Trusted Launch Default - New Az VM OS Disk


AjKundnani:
### Related command

- [az disk create --name 'diskname' -g 'diskrgname'](https://docs.microsoft.com/en-us/cli/azure/disk?view=azure-cli-latest#az-disk-create)

### Resource Provider

Microsoft.Compute/disks

### Description of Feature or Work Requested

# Trusted Launch as Default

As part of making compute layer secure for customers to deploy workloads, request is to make [Trusted Launch](https://docs.microsoft.com/en-us/azure/virtual-machines/trusted-launch) (GA Date: November 2021) as default deployment option in CLI. Following changes will need to be made for same:

>**NOTE**: Below scenarios are applicable only for **OS Disks** created **FromImage** and **LUN** is not set by end user, i.e., if command is used to create *Data Disk*, below scenarios and change requests do not apply.

## Scenario 1 - User creates OS disk from PIR image using `az disk create --image-reference`

Present Behavior | Updated Behavior post change
-|-
New disk with create option FromImage:<ol><li>User runs command `az disk create -n diskname -g diskrgname --image-reference 'urnofPIRimage'`<li>OS Disk with image reference and Premium_LRS SKU is created.</li></ol> | Updated default parameters and sequence:<ol><li>User runs command `az disk create -n diskname -g diskrgname --image-reference 'urnofPIRimage'`<li>CLI queries image capabilities using API [Virtual Machines Images GET API](https://docs.microsoft.com/en-us/rest/api/compute/virtual-machine-images/get?tabs=HTTP)<ul><li>For Gen2, use property **hyperVGeneration** with value of **V2**</li></ul><li>Conditions:<ul><li>If PIR image supports **hyperVGeneration = v2**, enable following parameters by default:<ul><li>--hyper-v-generation with value **V2**<li>--security-type with value **TrustedLaunch**</li></ul><li>Else, Disk will get deployed per user inputs. If securityType is not set to _TrustedLaunch_, Notification message should be printed for user. **Notification Message**:  `Consider upgrading security for your workloads using Azure Trusted Launch VMs. To know more about Trusted Launch, please visit https://aka.ms/TrustedLaunch`</li></ul></ol>

## Scenario 2 - User creates OS disk from ACG image using `az disk create --gallery-image-reference`

**No change** due to potential impact surface due to ACG not being TrustedLaunchSupported and Managed Image does not support Trusted Launch.
[...]

PR #27620, opened 2023-10-18
[Compute] BREAKING CHANGE: `az disk create`: Support creating disk with Gen2 and TLVM as default

yanzhudd:
**Related command**

`az disk create`

**Description**

Close: #23614

**Testing Guide**

**History Notes**

---


yonzhan:
Compute

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -420,64 +416,27 @@ def create_managed_disk(cmd, resource_group_name, disk_name, location=None,  # p
...
-                        ['TrustedLaunchSupported', 'TrustedLaunchAndConfidentialVmSupported'] \
-                        and security_type != 'TrustedLaunch':
-                    # will set default value of security_type
-                    logger.warning(log_message)
+                    hyper_v_generation = 'V2'
+                # set default value of security_type
+                if not security_type:
+                    security_type = 'TrustedLaunch'
+                if security_type != 'TrustedLaunch':
+                    logger.warning(UPGRADE_SECURITY_HINT)
```
zhoxing-ms, line 428: > will set default value of security_type

In this case, we only need to print the warning logs and no longer modify the default values, right?

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -420,64 +416,27 @@ def create_managed_disk(cmd, resource_group_name, disk_name, location=None,  # p
...
-                        and 'SecurityType' in gallery_image_info.features \
-                        and gallery_image_info.features['SecurityType'] in \
-                        ['TrustedLaunchSupported', 'TrustedLaunch', 'TrustedLaunchAndConfidentialVmSupported'] \
-                        and security_type != 'TrustedLaunch':
-                    # will set default value of security_type
-                    logger.warning(log_message)
+        if not security_type:
+            security_type = 'Standard'
+        if security_type != 'TrustedLaunch':
+            logger.warning(UPGRADE_SECURITY_HINT)
```
zhoxing-ms, line 439: We no longer need to modify the default values based on the conditions of `hyper_v_generation` and security_type`, right?

zhoxing-ms:
Could you re-record some related tests? Modifying the default behavior will not affect the tests in playback mode, but there may also be workflow issues caused by the breaking changes which cannot be detected by playback mode

src/azure-cli/azure/cli/command_modules/vm/custom.py
yanzhudd: Hi @zhoxing-ms this change has been confirmed with Ajay:
https://github.com/Azure/azure-cli/issues/23614#issuecomment-1764369004

src/azure-cli/azure/cli/command_modules/vm/custom.py
yanzhudd: Hi @zhoxing-ms  this change has been confirmed with Ajay:
https://github.com/Azure/azure-cli/issues/23614#issuecomment-1764369004

yanzhudd:
> Could you re-record some related tests? Modifying the default behavior will not affect the tests in playback mode, but there may also be workflow issues caused by the breaking changes which cannot be detected by playback mode

addressed

zhoxing-ms (approved):

PR #27628, opened 2023-10-18
[App Service] BREAKING CHANGE: Redact output on deployment source config

kamperiadis:
**Related command**
`az functionapp deployment source config`
`az webapp deployment source config`

**Description**
Currently, these two commands return a secret as part of the output. In this PR, we are redacting the secret to prevent accidental security leaks through CI logs.

**Testing Guide**
1. Create functionapp (`az functionapp create`)
2. Configure deployment (`az functionapp deployment source config`)

Reference document for object returned by the deployment source config commands: https://learn.microsoft.com/en-us/rest/api/appservice/web-apps/get-source-control#sitesourcecontrol

---


yanzhudd:
does it possible to add live test for this change?

kamperiadis:
> does it possible to add live test for this change?

It does not seem possible to add a live test for this change since there would be a dependency on an outside repository for which I would need to add GitHub actions credentials. This is probably why we do not even have live tests for adding GitHub actions.

yanzhudd (approved):

Issue #27641, opened 2023-10-19
Live test broken with Python 3.11


jiasli:
https://github.com/Azure/azure-cli/pull/25055 uses `self._outcome.errors` to check the test result:

https://github.com/Azure/azure-cli/blob/674625c388b36ad684e3531bec2c67cde6e00018/src/azure-cli-testsdk/azure/cli/testsdk/scenario_tests/base.py#L166

However, Python 3.11 made a change in https://github.com/python/cpython/pull/28180 that no longer buffers the errors in `_Outcome`. The errors are immediately sent to `result` and `errors` attribute is removed from `_Outcome`, causing failure during Azure CLI's live run:

```py
> azdev test test_role_assignment_scenario --live
...
self = <azure.cli.command_modules.role.tests.latest.test_role.RoleAssignmentScenarioTest testMethod=test_role_assignment_scenario>
args = ()

    def _save_recording_file(self, *args):  # pylint: disable=unused-argument
        if self.in_recording and self.cassette.dirty:
            self.cassette._save()  # pylint: disable=protected-access
>           if self._outcome.errors or self._outcome.skipped:   # pylint: disable=protected-access
E           AttributeError: '_Outcome' object has no attribute 'errors'

src\azure-cli-testsdk\azure\cli\testsdk\scenario_tests\base.py:166: AttributeError
```

https://stackoverflow.com/a/39606065/2199657 suggests to check `self._outcome.result.errors` and claims this approach is compatible with `pytest`, but it is actually not. When invoked by `pytest`, `_pytest.unittest.TestCaseFunction` is passed to `unittest.case.TestCase.__call__` as `result`. As `_pytest.unittest.TestCaseFunction` doesn't have `errors` attribute, `pytest` fails:

```
E       AttributeError: 'TestCaseFunction' object has no attribute 'errors'
```

`unittest.case._addError` calls `addError` on the `result` object, making retrieving the errors from `result` test-runner-dependent: https://github.com/python/cpython/blame/2258d6cfa2fcd0c290328e81b68aff25abf8c3d8/Lib/unittest/case.py#L74, https://github.com/python/cpython/blame/2258d6cfa2fcd0c290328e81b68aff25abf8c3d8/Lib/unittest/case.py#L99

jiasli:
https://github.com/Azure/azure-cli/pull/27649 disabled the temporary recording file functionality, thus bypassing this issue, but let's keep it open until we figure out an elegant way of retrieving the test result from a cleanup function.

PR #27649, opened 2023-10-20
{testsdk} Fix live test on Python 3.11

jiasli:
A temporary fix for https://github.com/Azure/azure-cli/issues/27641

After this change, in Python 3.11, even if a test fails, the temp recording file replaces the original recording file. This is the original behavior before https://github.com/Azure/azure-cli/pull/25055.

### DIFF
--- a/src/azure-cli-testsdk/azure/cli/testsdk/scenario_tests/base.py
+++ b/src/azure-cli-testsdk/azure/cli/testsdk/scenario_tests/base.py
@@ -163,15 +163,20 @@ class ReplayableTest(IntegrationTestBase):  # pylint: disable=too-many-instance-
     def _save_recording_file(self, *args):  # pylint: disable=unused-argument
         if self.in_recording and self.cassette.dirty:
             self.cassette._save()  # pylint: disable=protected-access
-            if self._outcome.errors or self._outcome.skipped:   # pylint: disable=protected-access
+
+            # Since Python 3.11, unittest.case._Outcome no longer has 'errors' attribute.
+            # The below statements only work in Python <= 3.10.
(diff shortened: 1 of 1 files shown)


yonzhan:
testsdk

kairu-ms (approved):

wangzelin007 (approved):

src/azure-cli-testsdk/azure/cli/testsdk/scenario_tests/base.py
```diff
@@ -163,15 +163,20 @@ def tearDown(self):
...
-                # replace recording_file by temp_recording_file
-                if os.path.exists(self.recording_file):
-                    os.remove(self.recording_file)
-                os.rename(self.temp_recording_file, self.recording_file)
+                return
+
+            # replace recording_file by temp_recording_file
+            if os.path.exists(self.recording_file):
+                os.remove(self.recording_file)
+            os.rename(self.temp_recording_file, self.recording_file)
```
jiasli, line 179: To answer @bebound's question, I replace `else` with `return` only to reduce one level of nesting. It is only a code style:

- https://stackoverflow.com/questions/9267643/if-return-vs-if-else-efficiency
- https://stackoverflow.com/questions/9191388/it-is-more-efficient-to-use-if-return-return-or-if-else-return

bebound: Haha, I'm asking because the `return` gives me a sense of intentional skipping of logic(`extra_c()`).

For example:
``` python
if A:
    condition_a()
    return
condition_b()
extra_c()
```

Issue #24753, opened 2022-11-28
Creating role assignment must require all assignee, role definition and scope to be specified explicitly


jiasli:
**Related command**
`az ad sp create-for-rbac`
`az role assignment create`
Other commands related to role assignment creation, such as `az vm identity assign`.

## Context

`az ad sp create-for-rbac` stopped creating role assignment for newly created service principal if `--role` (#19805) and `--scopes` (#21323) are not specified.

The current behavior of `az role assignment create` is to default `--scope` to the current subscription. If `--scope` is not specified and

- if `--resource-group` is not specified, the scope is set to the current subscription
- if `--resource-group` is specified, the scope is set to the resource group under the current subscription

**This behavior is against the Principle of Least Privilege and should be modified to align with `az ad sp create-for-rbac`.**

## Proposal

For _any_ Azure CLI command, **creating role assignment must require all assignee, role definition and scope to be specified explicitly. No default value should be used.**

For `az role assignment create`, this will be a breaking change. Similar to what we previously did for `az ad sp create-for-rbac`'s `--role` (#19805) and `--scopes` (#21323), there are 2 steps:

1. Show a warning to instruct the user to always explicitly specify scope.
2. Make scope mandatory.

## Discussion

One open discussion is **whether we should drop `--resource-group` and `--subscription` from `az role assignment create`**.

`--scope` is the all-in-one argument covering subscription level, resource group level and resource level.

Subscription level:

- `--scope /subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590`
- `--subscription 0b1f6471-1bf0-4dda-aec3-cb9272f09590`

Resource group level:

- `--scope /subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/resourceGroups/mygroup`
- `--subscription 0b1f6471-1bf0-4dda-aec3-cb9272f09590 --resource-group mygroup`

Resource level:

- `--scope /subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/resourceGroups/ZYTEST/providers/Microsoft.Compute/virtualMachines/myvm`
- Not supported by individual arguments

Using individual argument `--resource-group` and `--subscription` has several disadvantages:

- It encourages over-privileged assignment on subscription or resource group level as fine-grained resource-level assignment is not supported.
[...]

PR #24755, opened 2022-11-28
{Role} `az role assignment create`: Show warning if `--scope` argument is not specified

jiasli:
**Description**
Similar to #20965
A temporary warning for #24753

For `az role assignment create`, `--scope` defaults to the subscription. Without explicit consent from the user on the `--scope`, this behavior is considered Elevation of Privilege.

This PR adds a warning if `--scope` is not specified:

```
> az role assignment create --assignee 5963f50c-7c43-405c-af7e-53294de76abd --role Reader --resource-group clitest.rgizxewh7npwdc63xpadmmi73hvq5ftcza2acf5wiyulxnjy5fgujpvvqll63kndkru
Starting from Azure CLI 2.47.0, --scopes argument will become required for creating role assignments. Please explicitly specify --scopes.
{
  "canDelegate": null,
  "condition": null,
  "conditionVersion": null,
  "description": null,
  "id": "/subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/resourceGroups/clitest.rgizxewh7npwdc63xpadmmi73hvq5ftcza2acf5wiyulxnjy5fgujpvvqll63kndkru/providers/Microsoft.Authorization/roleAssignments/3a36df35-93c4-4e5b-b605-61e9f9a0f1fc",
  "name": "3a36df35-93c4-4e5b-b605-61e9f9a0f1fc",
  "principalId": "5963f50c-7c43-405c-af7e-53294de76abd",
  "principalName": "admin2@AzureSDKTeam.onmicrosoft.com",
  "principalType": "User",
  "resourceGroup": "clitest.rgizxewh7npwdc63xpadmmi73hvq5ftcza2acf5wiyulxnjy5fgujpvvqll63kndkru",
  "roleDefinitionId": "/subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/providers/Microsoft.Authorization/roleDefinitions/acdd72a7-3385-48ef-bd42-f606fba81ae7",
  "roleDefinitionName": "Reader",
  "scope": "/subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/resourceGroups/clitest.rgizxewh7npwdc63xpadmmi73hvq5ftcza2acf5wiyulxnjy5fgujpvvqll63kndkru",
  "type": "Microsoft.Authorization/roleAssignments"
}
```

[...]

PR #27651, opened 2023-10-20
[Role] BREAKING CHANGE: `az role assignment create`: `--scope` is now a required argument

jiasli:
Close https://github.com/Azure/azure-cli/issues/24753

**Related command**
`az role assignment create`

**Description**
To adhere to the Principle of Least Privilege and avoid over-scoped role assignment on the subscription, `--scope` is now a required argument.

A warning for this breaking change was announced in https://github.com/Azure/azure-cli/pull/24755.

See https://github.com/Azure/azure-cli/issues/24753 for detailed explanation.

The corresponding change on `az vm/vmss` is https://github.com/Azure/azure-cli/pull/27657.

**Testing Guide**
```sh
az role assignment create --assignee-object-id xxx --scope /subscriptions/xxx/resourceGroups/xxx --role reader

az role assignment create --assignee-object-id xxx --role reader
# This fails: the following arguments are required: --scope
```

**History Notes**
[Role] BREAKING CHANGE: `az role assignment create`: `--scope` is now a required argument.
[Role] BREAKING CHANGE: `az role assignment create`: Remove `--resource-group` argument.

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/role/_validators.py
+++ b/src/azure-cli/azure/cli/command_modules/role/_validators.py
@@ -84,11 +84,13 @@ def process_assignment_namespace(cmd, namespace):  # pylint: disable=unused-argu
     non_empty_msg = "usage error: {} can't be an empty string. Either omit it or provide a non-empty string."
 
     for arg in non_empty_args:
-        if getattr(namespace, arg) == "":
+        if getattr(namespace, arg, None) == "":
             # Get option name, like resource_group_name -> --resource-group
             option_name = cmd.arguments[arg].type.settings['options_list'][0]
             raise CLIError(non_empty_msg.format(option_name))
 
-    resource_group = namespace.resource_group_name
-    if namespace.scope and resource_group and getattr(resource_group, 'is_default', None):
-        namespace.resource_group_name = None  # drop configured defaults
+    # `az role assignment create` doesn't support resource_group_name
+    if hasattr(namespace, "resource_group_name"):
+        resource_group = namespace.resource_group_name
+        if namespace.scope and resource_group and getattr(resource_group, 'is_default', None):
+            namespace.resource_group_name = None  # drop configured defaults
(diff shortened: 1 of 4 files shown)


yonzhan:
Role

src/azure-cli/azure/cli/command_modules/role/_validators.py
```diff
@@ -84,11 +84,13 @@ def process_assignment_namespace(cmd, namespace):  # pylint: disable=unused-argu
...
+        if getattr(namespace, arg, None) == "":
             # Get option name, like resource_group_name -> --resource-group
             option_name = cmd.arguments[arg].type.settings['options_list'][0]
             raise CLIError(non_empty_msg.format(option_name))
 
-    resource_group = namespace.resource_group_name
-    if namespace.scope and resource_group and getattr(resource_group, 'is_default', None):
-        namespace.resource_group_name = None  # drop configured defaults
+    # `az role assignment create` doesn't support resource_group_name
+    if hasattr(namespace, "resource_group_name"):
```
evelyn-ys, line 93: We can mention removing `--resource-group` in history notes. This is also a breaking change customers need to pay attention

evelyn-ys (approved):

src/azure-cli/azure/cli/command_modules/role/_validators.py
jiasli: Nice catch! Added to the **History Notes** section.

src/azure-cli/azure/cli/command_modules/role/_validators.py
```diff
@@ -84,11 +84,13 @@ def process_assignment_namespace(cmd, namespace):  # pylint: disable=unused-argu
     non_empty_msg = "usage error: {} can't be an empty string. Either omit it or provide a non-empty string."
 
     for arg in non_empty_args:
-        if getattr(namespace, arg) == "":
+        if getattr(namespace, arg, None) == "":
```
bebound, line 87: Is there any reason to add the default value for `getattr`?

jiasli: Good question. After removing `resource_group_name` from `az role assignment create`, `resource_group_name` will not exist in `namespace`. If we don't add a default value here, `getattr` will fail for `resource_group_name` under `az role assignment create`.

bebound (approved):

PR #27674, opened 2023-10-24
{AppConfig} `az appconfig kv export`: Fix feature flag deserialization. Add Feature flag property constants

albertofori:
**Related command**

`az appconfig kv export`
`az appconfig feature`

**Description**

Handle deserialization of new `requirement_type` property in feature flags.
Adds feature flag property constants for consistency.

**Testing Guide**

**History Notes**

{AppConfig} `az appconfig kv export`: Fix feature flag deserialization. Add Feature flag property constants

---


yonzhan:
AppConfig

Eskibear (commented):
If unsupported properties are specified in a feature flag (just like `requirement_type`), it should ignore instead of throwing errors. In the future we might have more properties for feature variants for example, and I would expect it to work as usual. Can we add some test cases to guard the behavior?

ReaNAiveD:
> If unsupported properties are specified in a feature flag (just like `requirement_type`), it should ignore instead of throwing errors. In the future we might have more properties for feature variants for example, and I would expect it to work as usual. Can we add some test cases to guard the behavior?

Hi @albertofori , is there any update on the test cases?

Please note that we are launching the release for this sprint this week. Please resolve all comments by 11/07/2023 10:00 UTC, otherwise the release of this PR will have to be postponed to the next sprint (on 12-05).

albertofori:
> If unsupported properties are specified in a feature flag (just like `requirement_type`), it should ignore instead of throwing errors. In the future we might have more properties for feature variants for example, and I would expect it to work as usual. Can we add some test cases to guard the behavior?

This has been updated accordingly. However, the idea is not to ignore the unsupported property, but rather make sure that unsupported properties are not lost during import/export, however, these are not processed in any way. This will ensure that we only add validation/ processing logic whenever a new property is added and not have to individually add parsing logic for each property unless special conversion is needed.

albertofori:
> > If unsupported properties are specified in a feature flag (just like `requirement_type`), it should ignore instead of throwing errors. In the future we might have more properties for feature variants for example, and I would expect it to work as usual. Can we add some test cases to guard the behavior?
> 
> Hi @albertofori , is there any update on the test cases?
> 
> Please note that we are launching the release for this sprint this week. Please resolve all comments by 11/07/2023 10:00 UTC, otherwise the release of this PR will have to be postponed to the next sprint (on 12-05).

@ReaNAiveD Thanks for the reminder. There are a few additional changes that need to be made to other commands, hence, I am marking this as a draft now and it is fine to go out in the next sprint.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -710,14 +713,13 @@ def __export_features(retrieved_features, naming_convention):
...
-                    feature_filter = {}
-                    feature_filter["Name"] = filter_.name
-                    if filter_.parameters:
-                        feature_filter["Parameters"] = filter_.parameters
-                    feature_state[feature_reserved_keywords.enabledfor].append(feature_filter)
+
+                for condition_key, condition in feature.conditions.items():
+                    if condition_key == FeatureFlagConstants.CLIENT_FILTERS:
+                        feature_state[feature_reserved_keywords.enabledfor] = [filter_.to_dict() for filter_ in condition]
+                        continue
```
avanigupta, line 720: We can use if/else statement instead of loop control statements like `continue`.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -766,10 +776,10 @@ def __convert_feature_dict_to_keyvalue_list(features_dict, enabled_for_keyword):
                                 break
 
                             filter_param = val.get("parameters", {})
-                            new_val = {'name': val["name"]}
+                            new_val = {FeatureFlagConstants.FILTER_NAME: val["name"]}
```
avanigupta, line 779: RHS can also use constants for `name` and `parameters`

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/import_features_random_conditions.json
```diff
@@ -0,0 +1,28 @@
...
+        },
+        {
+          "Name": "Production Tests",
+          "Parameters": {
+            "StartTime": "2019-09-02T00:00:00Z",
+            "EndTime": "2019-11-01T00:00:00Z"
+          }
+        }
+      ],
+      "requirement_type": "All",
```
avanigupta, line 24: RequirementType should follow the same naming convention as EnabledFor and FeatureManagement. 
.NET FeatureManagement library expects PascalCase: https://github.com/albertofori/azure-cli/blob/36786c51a5d42ff1b0a197e72a2c12261ff6743b/src/azure-cli/azure/cli/command_modules/appconfig/_params.py#L189

albertofori: This has been updated, thanks!

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
albertofori: The idea was to differentiate between the key names in the feature dictionary from the file and the actual property names in the feature flag value. But since they are the same, it might actually be confusing to explicitly specify one over the other. This has been updated, thanks!

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
albertofori: This has been updated, thanks!

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/import_features_random_conditions.json
```diff
@@ -21,8 +21,8 @@
           }
         }
       ],
-      "requirement_type": "All",
-      "random_condition": "random_value"
+      "RequirementType": "All",
```
avanigupta, line 24: Could you add a new test for validating RequirementType?

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/import_features_random_conditions.json
avanigupta: I think we need to update the code to support different naming conventions for requirement type.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -744,32 +745,40 @@ def __convert_feature_dict_to_keyvalue_list(features_dict, enabled_for_keyword):
...
-                        feature_flag_value.conditions = {'client_filters': v[enabled_for_keyword]}
-                    except KeyError:
+                    feature_flag_value.conditions = {}
+                    enabled_for_found = False
+
+                    for condition, condition_value in v.items():
+                        if condition == enabled_for_keyword:
+                            feature_flag_value.conditions[FeatureFlagConstants.CLIENT_FILTERS] = condition_value
+                            enabled_for_found = True
+                            continue
```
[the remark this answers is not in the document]
albertofori: This has been updated, thanks!

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/import_features_random_conditions.json
albertofori: New tests have been added to validate the "Any/All" values when importing feature flags

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/import_features_random_conditions.json
albertofori: Thanks for pointing this out. It has been updated accordingly.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -228,14 +196,14 @@ def __read_features_from_file(file_path, format_):
     try:
         config_data = __read_with_appropriate_encoding(file_path, format_)
         found_feature_section = False
-        for index, feature_management_keyword in enumerate(FEATURE_MANAGEMENT_KEYWORDS):
+        for keywords in FeatureManagementReservedKeywords.ALL:
```
avanigupta, line 199: nit: singular keyword?

avanigupta: But then it points to multiple keywords within the single naming convention. I feel its slightly confusing.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -744,32 +721,43 @@ def __convert_feature_dict_to_keyvalue_list(features_dict, enabled_for_keyword):
...
-                    if feature_flag_value.conditions["client_filters"]:
+                    feature_flag_value.conditions = {}
+                    enabled_for_found = False
+
+                    for condition, condition_value in v.items():
+                        if condition == feature_management_keywords.enabled_for:
+                            feature_flag_value.conditions[FeatureFlagConstants.CLIENT_FILTERS] = condition_value
+                            enabled_for_found = True
+                        elif condition == feature_management_keywords.requirement_type:
+                            if condition_value not in ("All", "Any"):
```
avanigupta, line 732: Did we decide to keep the case sensitive comparison for Any/All?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: I think we can use singular 'keyword' here. `keyword.feature_management` still makes more sense because the different "keywords" (like feature_management) are just properties of one instance of "keyword".

albertofori: > But then it points to multiple keywords within the single naming convention. I feel its slightly confusing.

The idea was to have all the keywords corresponding to a naming convention together 
 since first we were passing `enabledFor` keyword to `__convert_feature_dict_to_keyvalue_list`.
 But we now have to pass more than just that keyword. So this was to make it easy to extend in the future without having to add another argument to the method.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
albertofori: Actually I see in the provider that it is not a case-sensitive evaluation. https://github.com/microsoft/FeatureManagement-Dotnet/blob/aafe0c95a54a32b376d718580db9bc82c7567d33/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs#L200C66-L200C140

I will update this accordingly.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
albertofori: Right, and I see the source of the confusion since you are iterating through keywords, it makes semantic sense to use the singular "keyword" as the variable.

How about calling the new type `KeywordSet`? 
Then we have `for keywordset in FeatureManagementReservedKeywords.ALL`

Would this be clearer?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
```diff
@@ -219,7 +187,7 @@ def __read_features_from_file(file_path, format_):
     config_data = {}
     features_dict = {}
     # Default is PascalCase, but it will always be overwritten as long as there is a feature section in file
-    enabled_for_keyword = ENABLED_FOR_KEYWORDS[0]
+    feature_management_keywords = FeatureManagementReservedKeywords.get_keywords()
```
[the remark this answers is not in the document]
albertofori: This has been updated, thanks!

src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py
avanigupta: that sounds good!

avanigupta (approved):

albertofori:
@zhoxing-ms @yanzhudd Could you please take a look at, and help merge in this PR if all looks good?
Thanks!

yanzhudd (approved):

PR #27676, opened 2023-10-24
[ACR] Add new command to check if specific blob is available to pull from registry

SakoPak:
**Description**
- Add new command to check if blobs within given image can be pulled from repository using command `az acr check-health --name myRegistry --image hello-world@sha256@abc123`

**Testing Guide**
- Verify if blobs in image 'sha256:abc123' can be pulled from the repository: `az acr check-health -n MyRegistry --image hello-world@sha256:abc123`
- Verify if blobs referenced by tag 'latest' can be pulled from the repository: `az acr check-health -n MyRegistry -t hello-world:latest`

**History Notes**
[ACR] `az acr check-health --name --image`: Add new command to check if blobs in given image can be pulled from registry

---


yonzhan:
ACR

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -49,6 +49,12 @@
   - name: Gets health state of the environment, without stopping on first error.
     text: >
         az acr check-health --ignore-errors
+  - name: Verifies if specific blob identified by image 'sha256:abc123' is available to pull from the repository.
```
mabenedi, line 52: Maybe use artifact instead of blob here?

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -416,28 +415,135 @@ def _check_private_endpoint(cmd, registry_name, vnet_of_private_endpoint):  # py
     else:
         raise CLIError('DNS routing verification failed')
 
+# Validate inputs for --repository and --image
+def validate_repository_and_image(cmd, registry_name, repository_name, image):
```
mabenedi, line 419: do we want to allow cx to use username and password to auth? do we want to allow optional rg param to avoid armcache?

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -416,28 +415,135 @@ def _check_private_endpoint(cmd, registry_name, vnet_of_private_endpoint):  # py
...
+    try:
+        request_data_from_registry(
+            http_method='get',
+            login_server=login_server,
+            path=path,
+            username=username,
+            password=password,
+            json_payload=None,
+            file_payload=None,
+            manifest_headers={'Accept': 'application/vnd.docker.distribution.manifest.v2+json'},
```
mabenedi, line 447: I believe this will only work for that specific type of manifest. Is this intended or do we want this to work with all types of manifests ie oci artifacts? We have some examples in docker utils to follow or functions you can just use

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -416,28 +415,135 @@ def _check_private_endpoint(cmd, registry_name, vnet_of_private_endpoint):  # py
...
+            params=image)
+    except CLIError as e:
+        # Error is due to the image tag not being found on manifest
+        if "manifest tagged by" in str(e):
```
mabenedi, line 451: what if im looking by digest and the digest isn't found? Wouldn't this logic give me REPOSITORY_INPUT_ERROR_MESSAGE?

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -416,28 +415,135 @@ def _check_private_endpoint(cmd, registry_name, vnet_of_private_endpoint):  # py
...
+            ', application/vnd.oci.image.manifest.v1+json'
+            ', application/vnd.oci.image.index.v1+json'
+            ', application/vnd.docker.distribution.manifest.v2+json'
+            ', application/vnd.docker.distribution.manifest.list.v2+json'},
+        auth=(username, password))
+    
+    manifest = response.json()
+
+    # Get the digest of the smallest blob for performance purposes
+    def get_smallest_blob_digest(manifest):
```
mabenedi, line 496: why are we defining a function inside of another function for it to only be used once? Either define it as a helper function the top level so it can be reused by other functions or just write the code directly in this function

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -416,28 +415,135 @@ def _check_private_endpoint(cmd, registry_name, vnet_of_private_endpoint):  # py
...
+    request_url = 'https://' + login_server + '/v2/' + repository_name + '/blobs/' + digest
+    
+    logger.debug(add_timestamp("Sending a HTTP GET request to {}".format(request_url)))
+    
+    response = requests.get(request_url, auth=(username, password))
+    
+    if response.status_code < 400:
+        print_pass("Blob identified by image '{}' is available to pull from registry '{}'".format(image, login_server))
+    else:
+        raise CLIError("'Blob identified by image '{}' cannot be pulled from registry '{}'".format(image, login_server))
```
mabenedi, line 515: Do we want to give the user the response status code or the error text returned from backend?

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -26,7 +26,8 @@
 RECOMMENDED_NOTARY_VERSION = "0.6.0"
 NOTARY_VERSION_REGEX = re.compile(r'Version:\s+([.\d]+)')
 DOCKER_PULL_WRONG_PLATFORM = 'cannot be used on this platform'
-
+IMAGE_INPUT_ERROR_MESSAGE = "The specified image tag was not found. Please check image name for --image."
```
mabenedi, line 29: Not a huge fan of "Please check image name for --image." Can we word this better? Same with repository below.

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -416,28 +415,135 @@ def _check_private_endpoint(cmd, registry_name, vnet_of_private_endpoint):  # py
...
     if in_cloud_console:
         logger.warning("Environment checks are not supported in Azure Cloud Shell.")
     else:
         _get_docker_status_and_version(ignore_errors, yes)
         _get_cli_version()
-
+    
     _check_registry_health(cmd, registry_name, ignore_errors)
 
+    if image and repository_name:
```
mabenedi, line 540: Why do we do validate repository and image before check registry health and then check blob exists after? Intuitively I would think you'd want to check them both back to back, I would think after checking registry health since if there's a network issue or something then you definitely won't be able to hit the repo

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -416,28 +415,135 @@ def _check_private_endpoint(cmd, registry_name, vnet_of_private_endpoint):  # py
...
 def acr_check_health(cmd,  # pylint: disable useless-return
                      vnet=None,
                      ignore_errors=False,
                      yes=False,
-                     registry_name=None):
+                     registry_name=None,
+                     repository_name=None,
+                     image=None):
     from azure.cli.core.util import in_cloud_console
+
```
mabenedi, line 526: You're doing get_registry_by_name 3 times (once in  validate_repository_and_image,  once _check_blob_exists, once in_check_registry_health).

You're doing  get_access_credentials twice (once in  validate_repository_and_image,  once _check_blob_exists). 

Can we just do this once in acr_check_health and pass the info through to the sub functions? This will both cut down on unnecessary network calls and reused code.

src/azure-cli/azure/cli/command_modules/acr/check_health.py
SakoPak: Limiting it to a specific type of manifest was not intended, thanks for catching that, will modify this!

src/azure-cli/azure/cli/command_modules/acr/check_health.py
SakoPak: Yes, good point, I changed the ordering of these methods so check_registry_health is first.

src/azure-cli/azure/cli/command_modules/acr/check_health.py
SakoPak: After consulting with Jason and Roy, I will be implementing this differently to say that image is not found, and then suggest an image they might have intended to enter. This gives helpful specific information in case user entered image name incorrectly with typo or accidentally pasted the wrong image name. If incorrect repo is entered, a repo not found error will come from another method and error message will come from the backend, I realized I don't need to check for it again.

src/azure-cli/azure/cli/command_modules/acr/_help.py
zhoxing-ms, line 52: Please use the first person voice in the help message
```suggestion
  - name: Verify if specific blob identified by image 'sha256:abc123' is available to pull from the repository.
```

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -49,6 +49,12 @@
...
+    text: >
+        az acr check-health -n MyRegistry -r MyRepository --image sha256:abc123
+  - name: Verifies if specific blob identified by tag 'latest' is available to pull from the repository.
```
zhoxing-ms, line 55:
```suggestion
  - name: Verify if specific blob identified by tag 'latest' is available to pull from the repository. 
```

zhoxing-ms:
Please address this CI issues

zhoxing-ms:
@SakoPak Please note that we are launching the release for this sprint this week. Please resolve all comments and conflicts by Tomorrow, otherwise the release of this PR will have to be postponed to the next sprint (on 12-05)

SakoPak:
> @SakoPak Please note that we are launching the release for this sprint this week. Please resolve all comments and conflicts by Tomorrow, otherwise the release of this PR will have to be postponed to the next sprint (on 12-05)

This PR can wait until next sprint as it's still in progress, thank you for the heads up.

src/azure-cli/azure/cli/command_modules/acr/_help.py
SakoPak: Consulted with Esteban on this one, will be sticking with the word 'blob' as the command is checking if blobs within a given image can be pulled from current environment, but will update description.

SakoPak: Yes, noted and changes made, thank you

src/azure-cli/azure/cli/command_modules/acr/check_health.py
SakoPak: Refactored this so that get_registry_by_name and get_access_credentials are in acr_check_health and eliminating duplicate calls, thanks for the fix!

src/azure-cli/azure/cli/command_modules/acr/check_health.py
SakoPak: Added resource-group name to avoid armcache.

src/azure-cli/azure/cli/command_modules/acr/check_health.py
SakoPak: If status code is 400 and above, the returned message will come from backend.

src/azure-cli/azure/cli/command_modules/acr/_help.py
```diff
@@ -49,6 +49,12 @@
   - name: Gets health state of the environment, without stopping on first error.
     text: >
         az acr check-health --ignore-errors
+  - name: Verify that blobs within a given image 'sha256:abc123' can be downloaded from the repository.
+    text: >
+        az acr check-health -n MyRegistry -r MyRepository --image sha256:abc123
```
rosanch, line 54: why is only image using the long name? isn't -t the same?

src/azure-cli/azure/cli/command_modules/acr/_params.py
```diff
@@ -434,6 +434,10 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('ignore_errors', options_list=['--ignore-errors'], help='Provide all health checks, even if errors are found', action='store_true', required=False)
         c.argument('vnet', options_list=['--vnet'],
                    help="Virtual network ID so to run this command inside a VNET to verify the DNS routing to private endpoints", required=False)
+        c.argument('registry_name', options_list=['--name', '-n'], help="The name of the registry.")
+        c.argument('repository_name', options_list=['--repository', '-r'], help="The target repository namespace such as 'ubuntu'.")
+        c.argument('image', options_list=['--image','-t'],
```
rosanch, line 439: We already have a definition for this variable. Is there a reason for this to be custom?
Same goes for the other added parameters

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -72,11 +72,60 @@ def _subprocess_communicate(command_parts, shell=False):
     return output, warning, stderr, succeeded
 
 
+# Retrieve blob identified by digest - GET {url}/v2/{name}/blobs/{digest}
+def _get_blob(login_server, repository_name, digest):
```
rosanch, line 76: is this API only being used in check health? It feels it might be used by multiple files if so let's have in one place

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -418,28 +448,157 @@ def _check_private_endpoint(cmd, registry_name, vnet_of_private_endpoint):  # py
     else:
         raise CLIError('DNS routing verification failed')
 
+# Validate --image input
+def validate_image(cmd,
```
rosanch, line 452: is this for a new command or a new argument? Either way please follow existing design pattern in the CLI

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -342,18 +374,16 @@ def _check_registry_health(cmd, registry_name, ignore_errors):
                 _handle_error(CMK_MANAGED_IDENTITY_ERROR.format_error_message(registry_name), ignore_errors)
 
 
-def _check_private_endpoint(cmd, registry_name, vnet_of_private_endpoint):  # pylint: disable=too-many-locals, too-many-statements
+def _check_private_endpoint(cmd,
+                            registry_name,
+                            registry,
```
rosanch, line 379: why do you need registry and registry name?

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -294,28 +342,12 @@ def _get_endpoint_and_token_status(cmd, login_server, ignore_errors):
     print_pass("Fetch refresh token for registry '{}'".format(login_server))
     print_pass("Fetch access token for registry '{}'".format(login_server))
 
-
-def _check_registry_health(cmd, registry_name, ignore_errors):
+def _check_registry_health(cmd,
+                           registry_name,
+                           registry,
```
rosanch, line 347: Why do you need registry and registry name?

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -294,28 +342,12 @@ def _get_endpoint_and_token_status(cmd, login_server, ignore_errors):
...
+                           login_server,
+                           ignore_errors):
     from azure.cli.core.profiles import ResourceType
-    if registry_name is None:
```
rosanch, line 300: why is the code being removed?

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -72,11 +72,60 @@ def _subprocess_communicate(command_parts, shell=False):
...
+                       username,
+                       password,
+                       image):
+    from difflib import get_close_matches
+    from .manifest import _get_manifest_path
+    from .repository import(
+        acr_repository_show_tags,
+        _obtain_data_from_registry)
+
+    if 'sha256' not in image:
```
rosanch, line 93: define constants

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -72,11 +72,60 @@ def _subprocess_communicate(command_parts, shell=False):
...
+        tags = acr_repository_show_tags(
```
rosanch, line 94: why list all tags? wouldn't it be better to just check for the given image?

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -72,11 +72,60 @@ def _subprocess_communicate(command_parts, shell=False):
...
+            password=password)
+        # Get a list of tags that have at least 80% similarity with the inputted image name
+        matches = get_close_matches(image, tags, n=1, cutoff=0.8)
+
+        if matches:
+            raise CLIError("Tag '{}' not found. Did you mean to input '{}'?".format(image, matches[0]))
+        return
+
+    # If digest is used, get all digests
+    image_digest_list = _obtain_data_from_registry(
```
rosanch, line 109: same here why not just pull the information from the service? I think this whole section is overcomplicating the issue. You can make that request regardless of Tag or Digest being the image format and get the image information

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -418,28 +448,157 @@ def _check_private_endpoint(cmd, registry_name, vnet_of_private_endpoint):  # py
...
+            path=_get_v2_manifest_path(repository_name, image),
+            username=username,
+            password=password,
+            json_payload=None,
+            file_payload=None,
+            manifest_headers=get_manifest_authorization_header(username, password),
+            params=image)
+
+    except RegistryException as e:
+        if e.status_code == 404:
```
rosanch, line 480: what is the intention behind this?

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -418,28 +448,157 @@ def _check_private_endpoint(cmd, registry_name, vnet_of_private_endpoint):  # py
...
+            _check_image_match(cmd,
+                               registry_name,
+                               repository_name,
+                               login_server,
+                               username,
+                               password,
+                               image)
+        raise CLIError("{} Please check if image/tag was inputted correctly.".format(str(e)))
+
+# Validate if blob layer can be pulled
```
rosanch, line 490: what do you mean by "can be pulled"?

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -418,28 +448,157 @@ def _check_private_endpoint(cmd, registry_name, vnet_of_private_endpoint):  # py
...
+    manifest_url = 'https://{}/v2/{}/manifests/{}'.format(login_server, repository_name, image)
+
+    manifest_response = requests.get(
+        manifest_url,
+        headers=get_manifest_authorization_header(username, password))
+
+    manifest = manifest_response.json()
+
+    # Get the digest of the smallest blob for performance purposes
+    smallest_blob = min(manifest['layers'], key=lambda layer: layer['size'])
```
rosanch, line 511: would this work any given manifest? Did you try with a Docker Manifest list?

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -418,28 +448,157 @@ def _check_private_endpoint(cmd, registry_name, vnet_of_private_endpoint):  # py
...
 # General command
 def acr_check_health(cmd,  # pylint: disable useless-return
                      vnet=None,
                      ignore_errors=False,
                      yes=False,
-                     registry_name=None):
+                     registry_name=None,
+                     repository_name=None,
+                     image=None,
+                     resource_group_name=None):
```
rosanch, line 537: isn't resource group by default available in all commands?

rosanch: also, could you please reorder the arguments by priority please?

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -418,28 +448,157 @@ def _check_private_endpoint(cmd, registry_name, vnet_of_private_endpoint):  # py
...
+
+    # Connectivity
+    try:
+        registry, _ = get_registry_by_name(cmd.cli_ctx, registry_name, rg)
+        login_server = registry.login_server.rstrip('/')
+    except CLIError:
+        from ._docker_utils import get_login_server_suffix
+        suffix = get_login_server_suffix(cmd.cli_ctx)
+
+        if not suffix:
```
rosanch, line 554: why? add comment please

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -418,28 +448,157 @@ def _check_private_endpoint(cmd, registry_name, vnet_of_private_endpoint):  # py
...
+            from ._errors import LOGIN_SERVER_ERROR
+            _handle_error(LOGIN_SERVER_ERROR.format_error_message(registry_name), ignore_errors)
+            return
+
+    # Get the access credentials for the registry
+    login_server, username, password = get_access_credentials(
```
rosanch, line 560: isn't the pattern to add username and password for the costumer as option for this type?

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -418,28 +448,157 @@ def _check_private_endpoint(cmd, registry_name, vnet_of_private_endpoint):  # py
...
+    # Then check if blob layer can be pulled
+    if (image and repository_name):
+        validate_image(cmd,
+                       registry_name,
+                       repository_name,
+                       login_server,
+                       username,
+                       password,
+                       image)
+        _check_blob(repository_name,
```
rosanch, line 592: if both functions are only used here couldn't they be just one function that validates the image?

src/azure-cli/azure/cli/command_modules/acr/tests/latest/test_acr_check_health_commands.py
```diff
@@ -0,0 +1,41 @@
...
+            'vnet': None,
+            'repository': repository,
+            'tag': tag,
+            'image': '{}:{}'.format(repository, tag),
+            'reason': 'OK',
+            'source_reg_id': '/subscriptions/dfb63c8c-7c89-4ef8-af13-75c1d873c895/resourcegroups/resourcegroupdiffsub/providers/Microsoft.ContainerRegistry/registries/sourceregistrydiffsub',
+        })
+
+        # Create the registry
+        result = self.cmd('acr create --name {registry} --sku {sku} --resource-group {rg}').get_output_in_json()
```
rosanch, line 30: was there no check health test before?

rosanch (changes requested):
Overall looks good, I left some comments, please do not resolve questions. I'll do after reading the response.
Thanks

src/azure-cli/azure/cli/command_modules/acr/_params.py
SakoPak: @rosanch Looks like majority of the commands actually use 'registry_name', options_list=['--registry', '-r'] instead of ['--name', '-n']. This '--name' was being used prior to my changes, and is also being used on line 82:
image

I can update the check-health command to use ''--registry, -r " if that is the most current version.

Also, the already existing variable, 'image_by_tag_or_digest_type', uses [--image, 't'] but it is defined as "The name of the image. May include a tag in the format 'name:tag' or digest in the format 'name@digest'." It combines the repository name with tag or digest.

The reason I defined --image differently is because with this PR, the check-health command asks user for the repository name, then the image or tag separately and '--image, -t' in this case means either the digest itself ("sha256:123") or the tag ("latest").  I asked for repo name separately as it is used in the APIs for 'get blob' and 'get manifest'.

Alternatively, I could use '--digest' and '--tag' to make it more precise, or refactor code to use '--image' as previously defined and parse out the repo name in the code.
There were concerns that users would not always quickly recall the image digest and that '--image' might be a better blanket term. Open to suggestions and adding @estebanreyl and @AntonioVT  to this comment as they had some input into this.

src/azure-cli/azure/cli/command_modules/acr/check_health.py
SakoPak: It's not being used in other files, but I could put it in '._docker_utils.py' so it could be found and re-used more easily.

src/azure-cli/azure/cli/command_modules/acr/check_health.py
SakoPak: @rosanch - will update to 'def _validate_image'. Is this what you are referring to?

src/azure-cli/azure/cli/command_modules/acr/check_health.py
SakoPak: 'Registry' is an object passed down from 'get_registry_by_name() on line 548.  The other methods like '_check_private_endpoint' were using 'get_registry_by_name()" and accessing properties from 'registry' so I changed it so that 'get_registry..." is called once and then passes the registry properties down.

 'Registry_name' is the name of the registry itself.

image

src/azure-cli/azure/cli/command_modules/acr/tests/latest/test_acr_check_health_commands.py
SakoPak: No, a test for check health didn't exist.

src/azure-cli/azure/cli/command_modules/acr/check_health.py
SakoPak: @rosanch - originally had this whole thing in one huge function but was advised to break it up into methods. The validate image is just checking to make sure the image data entered is valid and then runs '_check_image_match()' to suggest image name in case there is user error/typo. The check blob is calling an API to pull a layer of the blob to see if it can be downloaded. Wanted to separate so that it's more readable and reusable.

src/azure-cli/azure/cli/command_modules/acr/check_health.py
SakoPak: @rosanch  In this method, I am not checking for one given image - I am seeing if there is a possible match with the tag that the user gave me (the given image) and comparing it with the list of tags from their repo. To find a possible match, I would need the comprehensive list. Maybe I am not understanding your question?

src/azure-cli/azure/cli/command_modules/acr/check_health.py
SakoPak: This edit was added in response to above comment by Matt where he pointed out that get_registry_by_name and get_access_credentials were being used multiple times in different methods. So I am calling it once the acr_check_health and passing the info into the subfunctions. I will add comments so this is clear.

src/azure-cli/azure/cli/command_modules/acr/check_health.py
SakoPak: @rosanch sorry not clear on what your question means? So I am getting access credentials here and passing it down to the methods '_check_blob' and '_validate_image' so that I can use it in 'get_manifest_authorization_header' when I make the GET requests.

src/azure-cli/azure/cli/command_modules/acr/check_health.py
SakoPak: @rosanch ' can be pulled' == 'can be downloaded'

src/azure-cli/azure/cli/command_modules/acr/check_health.py
SakoPak: @rosanch So intention behind this was to first check if the image data given by user was valid before calling the API to check for the blob - is it even an image that exists in the repo? If it's a 404 and requested data does not exist then we check to see if it was user error and a possible typo and run '_check_image_match' so they get a more useful error message back. If there's a better way to do this, please let me know.

src/azure-cli/azure/cli/command_modules/acr/check_health.py
SakoPak: @rosanch - not certain about the resource group question, I thought I needed to include it in order to use in 'get_resource_group_by_registry_name' method on line 542?

SakoPak: and yes can reorder the args

src/azure-cli/azure/cli/command_modules/acr/check_health.py
SakoPak: 'Registry' is returned from the get_registry_by_name method, and the registry properties are then accessed by the subfunctions. 

Resgistry_name is just the name of the registry

src/azure-cli/azure/cli/command_modules/acr/check_health.py
SakoPak: @rosanch Yes I guess I don't need to check for tag or digest, it can just be one request. I think I was trying to customize the return message according to what was inputted but it's unnecessary

src/azure-cli/azure/cli/command_modules/acr/check_health.py
SakoPak: Yes thank you

src/azure-cli/azure/cli/command_modules/acr/_help.py
rosanch: my point is if you already support -t then we should use that in the example instead of the longer nameing.

src/azure-cli/azure/cli/command_modules/acr/_params.py
rosanch: Then I don't think it should be name image. Let's keep consistency normally we request for the repo or the image not both and we add validation for it. If you want to use a different format, we can't call it the same. For example, if it is only tag or digest, we can call it reference as shown in our API documentation. 
https://learn.microsoft.com/en-us/rest/api/containerregistry/manifests/get?view=rest-containerregistry-2019-08-15&tabs=HTTP
If you don't want to introduce a new pattern to the CLI and get in touch with the PM's to validate it. Then please use the existing image/repo pattern we use in all other commands

src/azure-cli/azure/cli/command_modules/acr/check_health.py
rosanch: nah it's ok. Whoever wants to reuse it can do that change.

src/azure-cli/azure/cli/command_modules/acr/check_health.py
rosanch: correct thanks

src/azure-cli/azure/cli/command_modules/acr/check_health.py
rosanch: I do like the optimization however, can't we remove the registry name argument then. 
Or are you making it optional (either one of the 2)?

src/azure-cli/azure/cli/command_modules/acr/check_health.py
rosanch: I see you're correct. Thanks for clarifying

src/azure-cli/azure/cli/command_modules/acr/check_health.py
rosanch: I meant add a comment explaining what are you validation with the "if not suffix" it's not clear at first glance by looking at the code

src/azure-cli/azure/cli/command_modules/acr/check_health.py
rosanch: I mean why do we not allow the costumer to pass their own credentials as an option?

src/azure-cli/azure/cli/command_modules/acr/check_health.py
rosanch: it's ok break it up. However, if we're always bound to both couldn't one be step inside the other instead? Because it seems every time you validate the image, you'll check the Blob as well. 
If you want to keep it as is I suggest changing the method names: _validate_image_name & _validate_image_layers

src/azure-cli/azure/cli/command_modules/acr/tests/latest/test_acr_check_health_commands.py
rosanch: thanks for adding coverage

src/azure-cli/azure/cli/command_modules/acr/check_health.py
rosanch: Good optimization, thanks

src/azure-cli/azure/cli/command_modules/acr/check_health.py
rosanch: that's all good. But that should be a step we do ONLY if we didn't find the image correct? If the input was a valid there's not need to add suggestions.

src/azure-cli/azure/cli/command_modules/acr/check_health.py
rosanch: My bad, when can it not be pulled it's what I meant.

src/azure-cli/azure/cli/command_modules/acr/_help.py
SakoPak: @rosanch Yes, there is already an example using '-t' below it
image

src/azure-cli/azure/cli/command_modules/acr/_params.py
SakoPak: @rosanch - yes, so then the command would be rewritten as  'acr check-health --registry myregistry --image hello-world@sha256:123 (or -t hello-world:latest). I will check with the PMs if we can change to '--registry, -r' instead of '--name, -n' or if they'd like to keep it to '--name, -n'. @terencet-dev 

question though - in the new artifact streaming command, we are using '--name' for registry. I'm guessing this is because the command also uses '--repository' and it would be potentially confusing?

src/azure-cli/azure/cli/command_modules/acr/check_health.py
SakoPak: Sure, sounds good, will add comment here. This particular block of code was already existing, I just moved its location. Agree that there is a lack of explanation, thank you!

src/azure-cli/azure/cli/command_modules/acr/check_health.py
SakoPak: Yes good point, I will add username and password as an option, thank you!

src/azure-cli/azure/cli/command_modules/acr/check_health.py
SakoPak: Yes, you are correct, I can just use the 'registry' object and the registry name is accessible already within that; will change this, thank you.

src/azure-cli/azure/cli/command_modules/acr/_params.py
SakoPak: Consulted with Terence and we will keep it to '--name, -n' to maintain consistency with the heavily used commands like 'acr create', 'acr login', 'acr import' etc. that use '--name' for the registry name.

src/azure-cli/azure/cli/command_modules/acr/check_health.py
SakoPak: The method _check_image_match is only called if the input is not found (line 489).

src/azure-cli/azure/cli/command_modules/acr/check_health.py
SakoPak: Correct, the _check_image_match only gets called when the image data is not found.

src/azure-cli/azure/cli/command_modules/acr/check_health.py
SakoPak: Will add a comment, thanks!

src/azure-cli/azure/cli/command_modules/acr/check_health.py
SakoPak: If blob layer can't be downloaded, it could be for a variety of reasons (ie. requested data not found, unauthorized, insufficient permissions etc.) and if that's the case, the server error message should return a status code and message that specifies the reason.

src/azure-cli/azure/cli/command_modules/acr/check_health.py
SakoPak: Good point, I would need to check if the manifest is a manifest list first by checking the media type layer, then find the digest of the smallest blob layer. Will rewrite and test this, thank you.

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -418,28 +484,125 @@ def _check_private_endpoint(cmd, registry_name, vnet_of_private_endpoint):  # py
     else:
         raise CLIError('DNS routing verification failed')
 
+# Validate --image input
+# Validate if image layer can be pulled
+def _validate_image_name(repository_name,
```
estebanreyl, line 489: Nit. _validate_image_in_registry is probably more accurate since you are checking for the repository@digest combo.

estebanreyl (commented):
Added a few comments, mostly on edge cases to consider, overall lgtm

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -71,12 +71,99 @@ def _subprocess_communicate(command_parts, shell=False):
...
+    manifest_url = 'https://{}/v2/{}/manifests/{}'.format(login_server, repository_name, digest)
+
+    manifest_response = requests.get(
+        manifest_url,
+        headers=get_manifest_authorization_header(username, password))
+
+    manifest = manifest_response.json()
+
+    # Pull the smallest blob layer in the image to test if the image can be pulled
+    # First, check if the manifest is a Docker Manifest List
```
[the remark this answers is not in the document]
SakoPak: Sure makes sense since it gives better context for the whole method, thanks for the feedback!

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -71,12 +71,99 @@ def _subprocess_communicate(command_parts, shell=False):
...
+    if manifest['mediaType'] == 'application/vnd.docker.distribution.manifest.list.v2+json':
```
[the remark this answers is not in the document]
SakoPak: Thank you for this input, I will include this in the spec as well!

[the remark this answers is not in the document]
SakoPak: ok that's a good catch, thanks for letting me know about this edge case. I can change it to parse Media Type from the response content-type header.

src/azure-cli/azure/cli/command_modules/acr/check_health.py
```diff
@@ -71,12 +71,99 @@ def _subprocess_communicate(command_parts, shell=False):
...
+            smallest_blobs.append(smallest_blob)
+
+        # Find the smallest blob across all image manifests
+        smallest_blob = min(smallest_blobs, key=lambda blob: blob['size'])
+
+    else:
+        # If manifest is not a Docker Manifest List, find the smallest blob in the given single manifest
+        smallest_blob = min(manifest['layers'], key=lambda layer: layer['size'])
+
+    smallest_digest = smallest_blob['digest']
```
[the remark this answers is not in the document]
SakoPak: ok that's a great point, I will factor in this edge case and have it pull the manifest like a regular blob, thanks!

src/azure-cli/azure/cli/command_modules/acr/check_health.py
SakoPak: True, I will change this so it's a bit more precise and informative, thank you!

zhoxing-ms:
@SakoPak Could you please resolve these CI issues?

toddysm:
@SakoPak is this still needed?

teresaritorto:
Closing due to lack of activity

Issue #25675, opened 2023-03-03
[Collect feedback - important] Settings in cloud for Azure CLI


jsntcy:
### **Why do we need this feature?**
Currently the settings for Azure CLI is only in our local machine, the customers may want to share the settings across their machines so they are always working with their favorite settings.

### **What do we provide?**
There will be two new arguments `--enable-settings-sync` and `--disable-settings-sync` for `az config` to allow customers to turn on settings sync or turn off settings sync.

### **How do we achieve this?**
We'll store settings in cloud (centralized storage or customers' own storage) and sync settings across different machines per customer's account.

### **We’ll be asking you for feedback on this feature. Your participation and feedback are very important to us -- I hope you will take the time to help us improve Azure CLI which a lot of people use every day.** 

# **If you like this feature, please help click**👍
# **If you have any suggestions, please help comment under this issue.**

Thanks in advance
On behalf of Azure CLI team

yonzhan:
roaming profile feature

rguptar:
This would be an awesome feature addition! [atuin](https://github.com/ellie/atuin) offers something similar albeit for shell history; it would be interesting to compare.

Not directly related to settings sync, but it would be nice to also have profiles that users can switch between. The commands would look something like this:
`az config profile create dev` // creates profile named dev
`az config profile activate dev` // activates profile named dev
`az config set cloud <cloud>` // sets the default cloud for the active profile
`az config set subscription <subscription id>` // sets the default subscription for the active profile
`az config show` // shows the active profile's config
`az config profile list` // lists all profiles

jsntcy:
@rguptar, thanks for your feedback.
What do you mean for "profiles"?
Is it the same as the profiles in `az cloud list-profiles` (where the profiles are predefined in Azure CLI)?

rguptar:
@jsntcy I'm not familiar with `az cloud list-profiles` and honestly could not find much with a search. Piecing what I can from this [doc](https://learn.microsoft.com/en-us/azure-stack/user/azure-stack-version-profiles-azurecli2?view=azs-2206&tabs=ad-win), they seems to be per-cloud settings the user can set; I could not find a way to create a new profile at least via the cli.

The profiles I am suggesting are user-centric. Let's say a user has four subscriptions, corresponding to personal, dev, and prod azurepublic, prod azureusgovernment. Profiles offer a convenient way to switch between them. Each profile would save the following information (and potentially more):
- cloud
`az config set cloud <cloud>`
[...]

PR #27726, opened 2023-10-31
{Core} Replace `allow_broker` with `enable_broker_on_windows`

jiasli:
**Related command**
`az login`

**Description**
This PR adopts the MSAL change https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/613 where MSAL will deprecate `allow_broker` and use `enable_broker_on_windows`.

**Testing Guide**
```
az config set core.enable_broker_on_windows=true
az login
az config unset core.enable_broker_on_windows
az login
```

### DIFF
--- a/src/azure-cli-core/azure/cli/core/telemetry.py
+++ b/src/azure-cli-core/azure/cli/core/telemetry.py
@@ -70,7 +70,7 @@ class TelemetrySession:  # pylint: disable=too-many-instance-attributes
         self.suppress_new_event = False
         self.poll_start_time = None
         self.poll_end_time = None
-        self.allow_broker = None
+        self.enable_broker_on_windows = None
         self.msal_telemetry = None
         self.secrets_detected = None
         self.user_agent = None
@@ -221,7 +221,7 @@ class TelemetrySession:  # pylint: disable=too-many-instance-attributes
         set_custom_properties(result, 'ShowSurveyMessage', str(self.show_survey_message))
         set_custom_properties(result, 'RegionInput', self.region_input)
         set_custom_properties(result, 'RegionIdentified', self.region_identified)
-        set_custom_properties(result, 'AllowBroker', str(self.allow_broker))
+        set_custom_properties(result, 'EnableBrokerOnWindows', str(self.enable_broker_on_windows))
         set_custom_properties(result, 'MsalTelemetry', self.msal_telemetry)
         set_custom_properties(result, 'SecretsWarning', _get_secrets_warning_config())
         set_custom_properties(result, 'SecretsDetected', self.secrets_detected)
@@ -462,9 +462,9 @@ def set_region_identified(region_input, region_identified):
 
 
 @decorators.suppress_all_exceptions()
-def set_broker_info(allow_broker):
-    # whether customer has configured `allow_broker` to enable WAM(Web Account Manager) login for authentication
-    _session.allow_broker = allow_broker
+def set_broker_info(enable_broker_on_windows):
+    # Log the value of `enable_broker_on_windows`
+    _session.enable_broker_on_windows = enable_broker_on_windows
 
 
 @decorators.suppress_all_exceptions()
--- a/src/azure-cli-core/azure/cli/core/_profile.py
+++ b/src/azure-cli-core/azure/cli/core/_profile.py
@@ -856,13 +856,14 @@ def _create_identity_instance(cli_ctx, *args, **kwargs):
     # EXPERIMENTAL: Use core.use_msal_http_cache=False to turn off MSAL HTTP cache.
     use_msal_http_cache = cli_ctx.config.getboolean('core', 'use_msal_http_cache', fallback=True)
 
-    # PREVIEW: On Windows, use core.allow_broker=true to use broker (WAM) for authentication.
-    allow_broker = cli_ctx.config.getboolean('core', 'allow_broker', fallback=False)
+    # PREVIEW: On Windows, use core.enable_broker_on_windows=true to use broker (WAM) for authentication.
+    enable_broker_on_windows = cli_ctx.config.getboolean('core', 'enable_broker_on_windows', fallback=False)
     from .telemetry import set_broker_info
-    set_broker_info(allow_broker=allow_broker)
+    set_broker_info(enable_broker_on_windows)
 
     # PREVIEW: In Azure Stack environment, use core.instance_discovery=false to disable MSAL's instance discovery.
     instance_discovery = cli_ctx.config.getboolean('core', 'instance_discovery', True)
 
-    return Identity(*args, encrypt=encrypt, use_msal_http_cache=use_msal_http_cache, allow_broker=allow_broker,
+    return Identity(*args, encrypt=encrypt, use_msal_http_cache=use_msal_http_cache,
+                    enable_broker_on_windows=enable_broker_on_windows,
                     instance_discovery=instance_discovery, **kwargs)
--- a/src/azure-cli-core/azure/cli/core/auth/identity.py
+++ b/src/azure-cli-core/azure/cli/core/auth/identity.py
@@ -54,7 +54,7 @@ class Identity:  # pylint: disable=too-many-instance-attributes
     _service_principal_store_instance = None
 
     def __init__(self, authority, tenant_id=None, client_id=None, encrypt=False, use_msal_http_cache=True,
-                 allow_broker=None, instance_discovery=None):
+                 enable_broker_on_windows=None, instance_discovery=None):
         """
         :param authority: Authentication authority endpoint. For example,
             - AAD: https://login.microsoftonline.com
@@ -69,7 +69,7 @@ class Identity:  # pylint: disable=too-many-instance-attributes
         self.client_id = client_id or AZURE_CLI_CLIENT_ID
         self._encrypt = encrypt
         self._use_msal_http_cache = use_msal_http_cache
-        self._allow_broker = allow_broker
+        self._enable_broker_on_windows = enable_broker_on_windows
         self._instance_discovery = instance_discovery
 
         # Build the authority in MSAL style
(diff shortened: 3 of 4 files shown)


yonzhan:
Core

jiasli:
After some further consideration, I am now not inclined to expose a unified `enable_broker` on the CLI side. Instead, CLI should expose the same config options as MSAL does - separate `enable_broker_on_windows`, `enable_broker_on_mac` config options for different platforms.

**Reasons**

1. Some users may share the same `~/.azure/config` file across platforms, possibly by:
    - mounting the `~/.azure` folder to a docker container
    - copy-and-pasting the `~/.azure` folder, the `~/.azure/config` file or the content of `~/.azure/config` to another computer
    
    In the future, this usage may become more common if we implement config roaming (https://github.com/Azure/azure-cli/issues/25675). With only one `enable_broker`, there will be no way to have different broker configurations for different platforms. For example, if I turn on broker on Windows and copy the `config` file to macOS, CLI on macOS will also have broker enabled, but that may not be what I intend.

2. With only one `enable_broker`, we may face an issue similar to what was described in https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/510#discussion_r1008477274. A user would likely run into an unintended behavior change when we support broker on a new platform in the future. For example:
    * D-day: CLI does not yet support broker on macOS, but a user somehow uses `enable_broker=True` on macOS. As CLI and its first-party app don't support broker on macOS yet, CLI tells MSAL to fallback to browser by passing `enable_broker_on_mac=False`. CLI would work today on macOS.
    * D+1 day: CLI and its first-party app starts to support broker on macOS (but broker is not turned on by default yet).
    * D+2 day: The user makes no other change, but upgrades to the latest CLI. `az login` will start to trigger broker code path unexpectedly.

    This issue can be avoided on D-day by raising an error when setting `enable_broker` on macOS before CLI supports broker, but the user experience is not good, especially if the config is copied from somewhere else.

rayluo:
> not inclined to expose a unified `enable_broker` on the CLI side. Instead, CLI should expose the same config options as MSAL does - separate `enable_broker_on_windows`, `enable_broker_on_mac` config options for different platforms.
> 
> **Reasons**
> 
> 1. Some users may share the same `~/.azure/config` file across platforms
> ...
>
>
> 2. With only one `enable_broker`, we may face an issue similar to what was described ...
> D-day ... D+1 day ... D+2 day ...

I may agree with the choice, but not for the reasons that you brought up.

MSAL and Azure CLI face different challenges. MSAL as a library can never know whether the calling app meets the prerequisite (i.e., registering the specific redirect URI) of a given platform, so, MSAL has to provide a per-platform opt-in flag, to avoid that "D-day, D+1 day, D+2 day" situation. (BTW, can you share me a link of where I mentioned a similar example before? I want to double check it.)

Azure CLI as an app could go with only one `enable_broker` parameter, by hardcoding the "enable_broker"-to-"enable_broker_on_windows" mapping today, and harcoding the "enable_broker"-to-both-"...on_windows" and "...on_mac" logic at a later time. That way, a roamed configuration would not cause an older version of Azure CLI to accidentally enable broker on Mac. So, roaming is not a concern.

In particular,
> For example, if I turn on broker on Windows and copy the config file to macOS, CLI on macOS will also have broker enabled, but that may not be what I intend.
> ...
> upgrades to the latest CLI. az login will start to trigger broker code path unexpectedly.

I do not think a user with `enable_broker=True` config on Windows would dislike seeing a broker experience on Mac, as long as it works smoothly.

---

The reason that a new `enable_broker_on_mac` could be a good idea, is because previously in the era of "there was only one broker and it is on Windows", Azure CLI did not have to expose an "enable_broker" parameter at all, but still wisely chose to provide such a fine-grained option during the beta testing period. So, perhaps a dedicate "enable_broker_on_mac" this time would also give us flexibility to toggle it on-and-off during its beta test.

jiasli:
> MSAL as a library can never know whether the calling app meets the prerequisite (i.e., registering the specific redirect URI) of a given platform

This is true and this "D-day, D+1 day, D+2 day" situation is more severe on MSAL, as it raises an error, but on CLI side, even though there is no error, it is still a behavior change.

> BTW, can you share me a link of where I mentioned a similar example before? I want to double check it.

The link is right in https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/510#discussion_r1008477274: https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/6882?_a=files

> That way, a roamed configuration would not cause an older version of Azure CLI to accidentally enable broker on Mac. So, roaming is not a concern.

I am actually not concerned about roaming across different CLI versions but roaming across different platforms for the same CLI version.

> I do not think a user with `enable_broker=True` config on Windows would dislike seeing a broker experience on Mac, as long as it works smoothly.

How about multiple users on different platforms? Consider your colleagues have `enable_broker=True` config on Windows and you copied their config to your macOS. You may not want to use broker right now.

> So, perhaps a dedicate "enable_broker_on_mac" this time would also give us flexibility to toggle it on-and-off during its beta test.

This is my opinion as well. More flexibility usually means more versatility and customizability (but sometimes also more complexity).

jiasli:
Telemetry-wise, another disadvantage of a combined `enable_broker` is that we have to correlate `enable_broker` with platform information to understand broker's usage. On the other hand, `enable_broker_on_windows` and `enable_broker_on_mac` allow us to get fine-grained information more easily and directly.

In my opinion, the complexity introduced for a combined `enable_broker` outweighs the benefit of it.

jiasli:
> The reason that a new `enable_broker_on_mac` could be a good idea, is because previously in the era of "there was only one broker and it is on Windows", Azure CLI did not have to expose an "enable_broker" parameter at all, but still wisely chose to provide such a fine-grained option during the beta testing period. So, perhaps a dedicate "enable_broker_on_mac" this time would also give us flexibility to toggle it on-and-off during its beta test.

On Windows, after we make WAM the default login method, `allow_broker`/`enable_broker_on_windows` will only be a switch to opt out. Most users won't use it at all. We definitely don't want to add additional complexity for this rarely used feature.

Also, after we kick off beta testing for broker support on mac, `enable_broker_on_mac` will provide us a dedicated switch, besides the `enable_broker_on_windows` switch or shared `allow_broker`.

jiasli:
When we start the Mac broker beta test, do we allow toggle whether broker on Mac is enabled, independent of whether the broker-on-Windows (i.e., WAM) is enabled or not? If so, we will need dedicated flags for Win and Mac, at least for now.

This means a user will indeed be exposed to the details of configuring broker behavior differently per platform, regardless of how we name it. Specifically, the difference is really just:

1. setting `enable_broker_on_windows=true` & `enable_broker_on_mac=false` (vice versa) in one same configuration file which can be reused on different platforms.
2. setting `allow_broker=true or false` differently per platform. The user would need to maintain 2 configuration files and deploy the right one to the right platform.
 
Given that the per-platform discrepancy is already there (for better or worse), we think style 1 is more user friendly.

src/azure-cli-core/azure/cli/core/telemetry.py
```diff
@@ -221,7 +221,7 @@ def _get_azure_cli_properties(self):
         set_custom_properties(result, 'ShowSurveyMessage', str(self.show_survey_message))
         set_custom_properties(result, 'RegionInput', self.region_input)
         set_custom_properties(result, 'RegionIdentified', self.region_identified)
-        set_custom_properties(result, 'AllowBroker', str(self.allow_broker))
+        set_custom_properties(result, 'EnableBrokerOnWindows', str(self.enable_broker_on_windows))
```
jiasli, line 224: Initially I forgot to rename `self.allow_broker` to `self.enable_broker_on_windows`. This results in the below error in debug log:

```
cli.azure.cli.core.decorators: Suppress exception:
  File "d:\cli\azure-cli\src\azure-cli\azure\cli\__main__.py", line 62, in <module>
    raise ex
  File "d:\cli\azure-cli\src\azure-cli\azure\cli\__main__.py", line 55, in <module>
    sys.exit(exit_code)
SystemExit: 0

During handling of the above exception, another exception occurred:

  File "d:\cli\azure-cli\src\azure-cli-core\azure\cli\core\decorators.py", line 79, in _wrapped_func
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "d:\cli\azure-cli\src\azure-cli-core\azure\cli\core\telemetry.py", line 110, in generate_payload
    cli = self._get_azure_cli_properties()
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "d:\cli\azure-cli\src\azure-cli-core\azure\cli\core\telemetry.py", line 224, in _get_azure_cli_properties
    set_custom_properties(result, 'AllowBroker', str(self.allow_broker))
                                                     ^^^^^^^^^^^^^^^^^
AttributeError: 'TelemetrySession' object has no attribute 'allow_broker'

telemetry.main: Split cli events and extra events failure: the JSON object must be str, bytes or bytearray, not NoneType
```

However, the CI doesn't find this. @evelyn-ys

evelyn-ys (approved):

bebound:
Should we add this to changelog? As we have mentioned `core.allow_broker` in doc: https://learn.microsoft.com/en-us/cli/azure/authenticate-azure-cli-web-account-manager
Just to confirm, will we also update the related doc?

src/azure-cli-core/azure/cli/core/telemetry.py
evelyn-ys: good catch. We should add telemetry check in CI too

jiasli:
> Should we add this to changelog? As we have mentioned `core.allow_broker` in doc: https://learn.microsoft.com/en-us/cli/azure/authenticate-azure-cli-web-account-manager Just to confirm, will we also update the related doc?

Nice catch and thanks for the reminder. As `allow_broker`'s default value will be changed to `True` in https://github.com/Azure/azure-cli/pull/28085, users won't need to explicitly set it anymore. The new switch `enable_broker_on_windows` will only be used to disable WAM. This is mentioned in the changelog in https://github.com/Azure/azure-cli/pull/28085.

I will submit another PR to update https://learn.microsoft.com/en-us/cli/azure/authenticate-azure-cli-web-account-manager.

bebound (approved):

PR #27839, opened 2023-11-14
{CI} Fix full test issue

wangzelin007:
The order of modules returned by get_path_table() is not fixed.
However, when generating all test jobs, the order of self.jobs is guaranteed by `self.jobs = sorted(jobs.items(), key=lambda item: -item[1])`, so this issue is not exposed.
But self.jobs only records cli modules before 2022/04/20.
Newly added modules will be assigned to each instance based on the average test time of existing modules.
This may result in a module not being tested by any instance.

For example:
The test load allocated on instance1 is 100 minutes.
The test load assigned on instance2 is 105 minutes
If there is another test, it will definitely be allocated to instance1 first, because the greedy algorithm will give priority to instances with lower load.
Assume that there are two modules at this time, one mysql and one containerapp, both of which take 6 minutes( The avg test time). 
If their order is not fixed, the following situation will occur

Instance1 thinks the tests run are as follows:
instance1: 100+mysql
instance2: 105 + containerapp

But instance2 thinks the tests run are as follows:
instance1: 100+containerapp
instance2: 105+mysql

This resulted in
instance1 thinks it is running mysql and instance2 will run containerapp,
instance2 also thinks that it is running mysql and instance1 will run containerapp,

Finally the containerapp will not be tested by any instance.

image

**Related command**

**Description**

**Testing Guide**

**History Notes**

---

### DIFF
--- a/scripts/ci/automation_full_test.py
+++ b/scripts/ci/automation_full_test.py
@@ -526,7 +526,7 @@ class AutomaticScheduling(object):
         result = get_path_table()
         # only get modules and core, ignore extensions
         self.modules = {**result['mod'], **result['core']}
-        logger.info(self.modules)
+        logger.info(json.dumps(self.modules, indent=2))
 
     def get_extension_modules(self):
         out = subprocess.Popen(['azdev', 'extension', 'list', '-o', 'tsv'], stdout=subprocess.PIPE)
@@ -570,7 +570,7 @@ class AutomaticScheduling(object):
             self.works[idx][k] = v
         # instance_idx: 1~n, python list index: 0~n-1
         self.instance_idx -= 1
-        logger.info(self.works)
+        logger.info(json.dumps(self.works, indent=2))
         return self.works[self.instance_idx]
 
     def run_instance_modules(self, instance_modules):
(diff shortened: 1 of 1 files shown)


yonzhan:
CI

scripts/ci/automation_full_test.py
```diff
@@ -525,7 +525,17 @@ def __init__(self):
...
+        result_mod = result['mod']
+        result_core = result['core']
+
+        # make sure the dictionary is in order, otherwise job assignments will be random.
+        ordered_mod = OrderedDict(sorted(result_mod.items()))
+        ordered_core = OrderedDict(sorted(result_core.items()))
+
+        # merge two ordered dictionaries
+        self.modules = OrderedDict(list(ordered_mod.items()) + list(ordered_core.items()))
+        logger.info(json.dumps(self.modules, indent=2))
```
bebound, line 538: Why not use `OrderedDict(sorted((result_mod | result_core).items()))`?
Does the sequence of `mod` and `core` matters?

wangzelin007: You are right, thank you !

bebound (approved):

Issue #7012, opened 2018-08-09
Blob SAS generated with CLI does not seem to work


zezha-msft:
Please refer to [here](https://github.com/Azure/azure-storage-python/issues/485) for the details.

I validated that there's no problem in the SDK. Thanks!

nicolaibaralmueller:
Can be closed

PR #27887, opened 2023-11-21
[AKS] BREAKING CHANGE: `az aks create`: Updates to option `--enable-managed-identity`

FumingZhang:
**Related command**

`az aks create`

**Description**

- Resolve issue [#7012](https://github.com/Azure/azure-cli-extensions/issues/7012).
- Specifying `--enable-managed-identity` and `--service-principal`/`--client-secret` at the same time will cause a `MutuallyExclusiveArgumentError`
- Change the default value of option `--enable-managed-identity` from `True` to `False`
- When options `--service-principal` and `--client-secret` are not specified at the same time, CLI will backfill the value of `--enable-managed-identity` to `True` to maintain the same behavior as before (that is, create an cluster with managed system assigned identity by default)

**Testing Guide**

**History Notes**

[AKS] BREAKING CHANGE: `az aks create`: Specifying `--enable-managed-identity` and `--service-principal`/`--client-secret` at the same time will cause a `MutuallyExclusiveArgumentError`
[AKS] BREAKING CHANGE: `az aks create`: Change the default value of option `--enable-managed-identity` from `True` to `False`
[AKS] `az aks create`: When options `--service-principal` and `--client-secret` are not specified at the same time, CLI will backfill the value of `--enable-managed-identity` to `True` to maintain the same behavior as before (that is, create an cluster with managed system assigned identity by default)

---


yonzhan:
AKS

matthchr:
Do we actually want to change this?
> Change the default value of option --enable-managed-identity from True to False

I think it's OK that this default is true? Rest of the change seems reasonable to me though.

FumingZhang:
> Do we actually want to change this?
> 
> > Change the default value of option --enable-managed-identity from True to False
> 
> I think it's OK that this default is true? Rest of the change seems reasonable to me though.

If the default is kept as `True`, I guess there is no way to distinguish whether the user has specified this option, and thus cannot raise MutuallyExclusiveArgumentError

FumingZhang:
> > [AKS] BREAKING CHANGE: az aks create: Specifying --enable-managed-identity and --service-principal/--client-secret at the same time will cause a MutuallyExclusiveArgumentError
> > [AKS] BREAKING CHANGE: az aks create: Change the default value of option --enable-managed-identity from True to False
> 
> Since this PR contains multiple breaking changes, could we release it in the next breaking change window (next Build Event sprint)?

marked as draft

PR #14509, opened 2020-07-24
{Profile} az login: Refine error message when tenant is not found

jiasli:
**Description**  

Resolve #14096

In `az login`, when `--tenant` is not a valid GUID, CLI will try to query the tenant information from AAD - [Fetch the OpenID Connect metadata document](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-protocols-oidc#fetch-the-openid-connect-metadata-document) (#10418).

If the tenant is not found, CLI shows error

```
> az login -t non-existing-tenant.onmicrosoft.com
usage: az login [-h] [--verbose] [--debug] [--only-show-errors]
                [--output {json,jsonc,yaml,yamlc,table,tsv,none}]
                [--query JMESPATH] [--username USERNAME] [--password PASSWORD]
                [--service-principal] [--tenant TENANT]
                [--allow-no-subscriptions] [-i] [--use-device-code]
                [--use-cert-sn-issuer]
az login: error: 'issuer'
```

This is not intuitive. 

This PR exposes the real error message:

```
> az login -t non-existing-tenant.onmicrosoft.com
Failed to resolve tenant 'non-existing-tenant.onmicrosoft.com'.

Error detail: {"error":"invalid_tenant","error_description":"AADSTS90002: Tenant 'non-existing-tenant.onmicrosoft.com' 
not found. This may happen if there are no active subscriptions for the tenant. Check to make sure you have the correct 
tenant ID. Check with your subscription administrator.\r\nTrace ID: 94a33bd9-6457-4f89-89f2-1b9144120200\r\nCorrelation 
ID: 652bdb39-211a-423c-aadf-95e4c6401159\r\nTimestamp: 2020-07-24 07:57:38Z","error_codes":[90002],
"timestamp":"2020-07-24 07:57:38Z","trace_id":"94a33bd9-6457-4f89-89f2-1b9144120200","correlation_id":
"652bdb39-211a-423c-aadf-95e4c6401159","error_uri":"https://login.microsoftonline.com/error?code=90002"}
```

**Testing Guide**  
```sh
# Valid tenant 
az login -t azuresdkteam.onmicrosoft.com

# Invalid tenant
az login -t non-existing-tenant.onmicrosoft.com
```

PR #27892, opened 2023-11-21
{Profile} Use v2.0 authority to resolve tenant ID

jiasli:
**Related command**
`az login`

**Description**
Azure CLI migrated from ADAL to MSAL (https://github.com/Azure/azure-cli/pull/19853) which is powered by AAD v2.0. The tenant resolution logic (added by https://github.com/Azure/azure-cli/pull/10418) should do the same.

https://github.com/Azure/azure-cli/pull/14509 added unit tests `azure.cli.command_modules.profile.tests.latest.test_profile_custom.ProfileCommandTest.test_login_validate_tenant` for `azure.cli.command_modules.profile._validators.validate_tenant`.

**Testing Guide**
`az login --tenant microsoft.onmicrosoft.com`

**Additional information**
Even though MSAL has special logic for handling ADFS:

https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/09d5ff9a71a51c998b83b356047230677825a622/msal/authority.py#L106-L112

```py
            tenant_discovery_endpoint = (
                'https://{}:{}{}{}/.well-known/openid-configuration'.format(
                    self.instance,
                    443 if authority.port is None else authority.port,
                    authority.path,  # In B2C scenario, it is "/tenant/policy"
                    "" if tenant == "adfs" else "/v2.0" # the AAD v2 endpoint
                    ))
```

This not our concern here since on Azure CLI side, `adfs` is part of `active_directory_endpoint` and `tenant` is discarded:

https://github.com/Azure/azure-cli/blob/b00483c1a1cd17053af5c483f62b11c829f4c27d/src/azure-cli-core/azure/cli/core/auth/identity.py#L360-L377

`validate_tenant` shouldn't be hit at all.

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/profile/_validators.py
+++ b/src/azure-cli/azure/cli/command_modules/profile/_validators.py
@@ -12,20 +12,20 @@ logger = get_logger(__name__)
 def validate_tenant(cmd, namespace):
     """
     Make sure tenant is a GUID. If domain name is provided, resolve to GUID.
-    https://docs.microsoft.com/azure/active-directory/develop/v2-protocols-oidc#fetch-the-openid-connect-metadata-document
+    https://learn.microsoft.com/en-us/entra/identity-platform/v2-protocols-oidc#find-your-apps-openid-configuration-document-uri
     """
     from azure.cli.core.util import is_guid
     if namespace.tenant is not None and not is_guid(namespace.tenant):
         import requests
         active_directory_endpoint = cmd.cli_ctx.cloud.endpoints.active_directory
-        url = '{}/{}/.well-known/openid-configuration'.format(active_directory_endpoint, namespace.tenant)
+        url = '{}/{}/v2.0/.well-known/openid-configuration'.format(active_directory_endpoint, namespace.tenant)
         response = requests.get(url, verify=not should_disable_connection_verify())
 
         if response.status_code != 200:
             from knack.util import CLIError
             raise CLIError("Failed to resolve tenant '{}'.\n\nError detail: {}".format(namespace.tenant, response.text))
 
-        # Example issuer: https://sts.windows.net/72f988bf-86f1-41af-91ab-2d7cd011db47/
+        # Example issuer: https://login.microsoftonline.com/72f988bf-86f1-41af-91ab-2d7cd011db47/v2.0
         tenant_id = response.json()['issuer'].split("/")[3]
         logger.debug('Resolved tenant domain name %s to GUID %s', namespace.tenant, tenant_id)
         namespace.tenant = tenant_id


yonzhan:
Profile

src/azure-cli/azure/cli/command_modules/profile/_validators.py
```diff
@@ -12,20 +12,20 @@
 def validate_tenant(cmd, namespace):
     """
     Make sure tenant is a GUID. If domain name is provided, resolve to GUID.
-    https://docs.microsoft.com/azure/active-directory/develop/v2-protocols-oidc#fetch-the-openid-connect-metadata-document
+    https://learn.microsoft.com/en-us/entra/identity-platform/v2-protocols-oidc#find-your-apps-openid-configuration-document-uri
```
rayluo, line 15: The other day I learn this trick from a Content Team writer. They said, most (if not all) of the learn.microsoft.com pages shall omit the `/en-us` part from the url, so that the website will automatically redirect the reader into a page written with their browser's preferred language.
```suggestion
    https://learn.microsoft.com/entra/identity-platform/v2-protocols-oidc#find-your-apps-openid-configuration-document-uri
```

src/azure-cli/azure/cli/command_modules/profile/_validators.py
```diff
@@ -12,20 +12,20 @@
...
     """
     from azure.cli.core.util import is_guid
     if namespace.tenant is not None and not is_guid(namespace.tenant):
         import requests
         active_directory_endpoint = cmd.cli_ctx.cloud.endpoints.active_directory
-        url = '{}/{}/.well-known/openid-configuration'.format(active_directory_endpoint, namespace.tenant)
+        url = '{}/{}/v2.0/.well-known/openid-configuration'.format(active_directory_endpoint, namespace.tenant)
```
rayluo, line 21: If all the functionality of this `validate_tenant(...)` function is merely detect the existence of that "well known" url and then extract the tenant_id, I believe this change is technically unnecessary.

That being said, now that you noticed this issue and send out this PR, I agree that we can make this change.

rayluo (approved):

src/azure-cli/azure/cli/command_modules/profile/_validators.py
jiasli: Thanks for the suggestion. Actually, I am well aware of this (omitting `en-us`). This is actually a requirement for customer-facing URLs. However, this URL here appears in code comment which is only for developers, so the English version should be the single source of truth.

src/azure-cli/azure/cli/command_modules/profile/_validators.py
jiasli: > If all the functionality of this `validate_tenant(...)` function is merely detect the existence of that "well known" url and then extract the tenant_id, I believe this change is technically unnecessary.

True, but I am only trying to align with MSAL and reduce the invocation on v1 endpoint. I don't want to get a warning from Microsoft Entra ID some day in the future that Azure CLI is still depending on v1 endpoint to do its work.

evelyn-ys (approved):

bebound (approved):

Issue #26682, opened 2023-06-15
[Feature Request] Replace entry script `az.cmd` with `az.exe` on Windows


jiasli:
**Is your feature request related to a problem? Please describe.**

Currently on Windows, a cmd file is used as the entry point for Azure CLI: `C:\Program Files (x86)\Microsoft SDKs\Azure\CLI2\wbin\az.cmd`

This causes several problems:

1. https://github.com/Azure/azure-cli/issues/15529. There was an attempt https://github.com/Azure/azure-cli/pull/23514 to add `az.ps1` for PowerShell, but that caused more issues and got reverted by https://github.com/Azure/azure-cli/pull/24015

3. Pressing <kbd>Ctrl</kbd>+<kbd>C</kbd> can't terminate `az` commands immediately. Instead, `Terminate batch job (Y/N)?` is shown:
![image](https://github.com/Azure/azure-cli/assets/4003950/37cdd13b-f611-4f45-b3a9-ef0eb6794dd8)

4. `az.cmd` can't be invoked by running `az` in WSL, Cygwin or Git Bash. The `az` Bash entry script introduced by https://github.com/Azure/azure-cli/pull/13197 caused more issues: https://github.com/Azure/azure-cli/issues/13573

**Describe the solution you'd like**

Replace entry script `az.cmd` with `az.exe` on Windows.

**Additional context**

The current `az.cmd` implementation requires Azure CLI to bundle a Python executable `C:\Program Files (x86)\Microsoft SDKs\Azure\CLI2\python.exe`. This Python interpreter is considered not secure by some restrictions and organizations, because it allows using build-in modules such as [`ctypes`](https://docs.python.org/3/library/ctypes.html) to execute arbitrary native code, bypassing enforcements such as [Windows Defender Application Control (WDAC)](https://learn.microsoft.com/en-us/windows/security/threat-protection/windows-defender-application-control/wdac-and-applocker-overview).

yonzhan:
Thank you for opening this issue, we will look into it.

bebound:
Adding `entry_point` in `setup.py` can create `az.exe`, which solves 1&2, but `python.exe` is still preserved.
PyInstaller seems to be the ultimate solution.

bebound:
This is also a prerequisite for MSIX.

PR #27911, opened 2023-11-24
{Packaging} Support Windows ZIP package

### LINKED ISSUE #22462: [Feature Request] Release Azure CLI as ZIP packages

Currently Azure CLI on Windows is released as an MSI package. This has several problems:

- The MSI requires admin permission (https://github.com/Azure/azure-cli/issues/15381), which makes it impossible to install Azure CLI as a normal user.
- The installation location is fixed to `C:\Program Files (x86)\Microsoft SDKs\Azure\CLI2`, making it impossible to customize the installation location.
- ZIP allows installing multiple versions side-by-side
- Some corporate firewall/policy may block downloading `.exe`, `.msi` (https://github.com/Azure/azure-cli/issues/22447) 
- Updating with MSI is slow and takes several minutes.

We could release Azure CLI as a portable ZIP packages, like PowerShell releases its ZIP packages at https://github.com/PowerShell/PowerShell/releases.

**Additional context**

As there is no official way to sign a ZIP, we can only use SHA256 to compute the hash of the ZIP and publish the hash along with the ZIP, but there is no way to sign it with Microsoft certificate.

bebound:
**Description**

Close #22462
Release zip package so that CLi can be used without admin privilege.

### Difference between MSI build
- Only keep `az.cmd` entry script
- Rename `wbin` to `bin`
- x64 only

### Package size
The uncompressed folder is 283MB
zip with `deflate` algorithm:  **90MB**
zip with `LZMA` algorithm: **77MB** (Windows File Explorer does not support it)

MSI is packed to a cab file and zipped with `LZX` algorighm: **67MB**.

If I compressed it to 7z, it's only **37MB**, which is due to the [Solid compression](https://en.wikipedia.org/wiki/Solid_compression). If I disable solid compression, the size become 72MB.

Unfortunately, zip does not support solid compression. Since it can be opened without installing any software, let's keep using zip.

### 7z is much faster than `Compress-Archive` and `Expand-Archive`

`Compress-Archive` and `Expand-Archive` take 5 and 10 minutes respectively. `7z` is much faster, completing the task in just several seconds.

Doc is https://github.com/MicrosoftDocs/azure-docs-cli/pull/4045


yonzhan:
Packaging

build_scripts/windows/scripts/build.cmd
```diff
@@ -137,18 +144,25 @@ if %errorlevel% neq 0 goto ERROR
...
 mkdir %BUILDING_DIR%\wbin
-copy %REPO_ROOT%\build_scripts\windows\scripts\az.cmd %BUILDING_DIR%\wbin\
-copy %REPO_ROOT%\build_scripts\windows\scripts\azps.ps1 %BUILDING_DIR%\wbin\
-copy %REPO_ROOT%\build_scripts\windows\scripts\az %BUILDING_DIR%\wbin\
+if "%TARGET%"=="msi" (
+    copy %REPO_ROOT%\build_scripts\windows\scripts\az.cmd %BUILDING_DIR%\wbin\
+    copy %REPO_ROOT%\build_scripts\windows\scripts\azps.ps1 %BUILDING_DIR%\wbin\
+    copy %REPO_ROOT%\build_scripts\windows\scripts\az %BUILDING_DIR%\wbin\
+) else (
+    copy %REPO_ROOT%\build_scripts\windows\scripts\az_zip.cmd %BUILDING_DIR%\wbin\az.cmd
```
bebound, line 164: `azps.ps1` and `az` is not included in zip as they cause other problems. See https://github.com/Azure/azure-cli/issues/26682

build_scripts/windows/scripts/build.cmd
```diff
@@ -193,14 +207,19 @@ for /d %%d in ("azure*.dist-info") do (
...
-if %errorlevel% neq 0 goto ERROR
 
-echo Building MSI...
-msbuild /t:rebuild /p:Configuration=Release /p:Platform=%ARCH% %REPO_ROOT%\build_scripts\windows\azure-cli.wixproj
+if "%TARGET%"=="msi" (
+    echo Building MSI...
+    msbuild /t:rebuild /p:Configuration=Release /p:Platform=%ARCH% %REPO_ROOT%\build_scripts\windows\azure-cli.wixproj
+) else (
+    echo Building ZIP...
+    ren %BUILDING_DIR% "Azure CLI"
```
bebound, line 216: Use `Azure CLI` as root folder of zip package.
PS: The MSI installation path is `C:\Program Files\Microsoft SDKs\Azure\CLI2`.

build_scripts/windows/scripts/build.cmd
```diff
@@ -122,7 +130,6 @@ set PYTHON_EXE=%PYTHON_DIR%\python.exe
 robocopy %PYTHON_DIR% %BUILDING_DIR% /s /NFL /NDL
 
 set CLI_SRC=%REPO_ROOT%\src
-%BUILDING_DIR%\python.exe -m pip install --no-warn-script-location --force-reinstall pycparser==2.18
```
bebound, line 125: `pycparser==2.19` is installed later.

build_scripts/windows/scripts/build.cmd
```diff
@@ -193,14 +209,21 @@ for /d %%d in ("azure*.dist-info") do (
...
-echo Building MSI...
-msbuild /t:rebuild /p:Configuration=Release /p:Platform=%ARCH% %REPO_ROOT%\build_scripts\windows\azure-cli.wixproj
+if "%TARGET%"=="msi" (
+    echo Building MSI...
+    msbuild /t:rebuild /p:Configuration=Release /p:Platform=%ARCH% %REPO_ROOT%\build_scripts\windows\azure-cli.wixproj
+) else (
+    echo Building ZIP...
+    REM Rename zip root folder to "Azure CLI"
+    ren %BUILDING_DIR% "Azure CLI"
+    REM Use LZMA compression to reduce the size of the zip file.
```
bebound, line 220: zip with default `deflate` algorithm:  90MB
zip with `LZMA` algorithm: 77MB

src/azure-cli/azure/cli/command_modules/util/custom.py
```diff
@@ -140,6 +140,13 @@ def upgrade_version(cmd, update_all=None, yes=None):  # pylint: disable=too-many
                            "or run 'pip install --upgrade azure-cli' in this container")
         elif installer == 'MSI':
             _upgrade_on_windows()
+        elif installer == 'ZIP':
+            if platform.architecture()[0] == '32bit':
+                zip_url = 'https://aka.ms/installazurecliwindowszip'
+            else:
+                zip_url = 'https://aka.ms/installazurecliwindowszipx64'
+            logger.warning("Please download the latest ZIP from %s, delete the old installation folder and extract the "
+                           "new version to the same location", zip_url)
```
bebound, line 149: When extracting a folder with the same name, the default action is to merge, which can lead to unexpected issues.
Deleting the old folder is required.

build_scripts/windows/scripts/build.cmd
```diff
@@ -43,6 +48,7 @@ set ARTIFACTS_DIR=%~dp0..\artifacts
 mkdir %ARTIFACTS_DIR%
 set TEMP_SCRATCH_FOLDER=%ARTIFACTS_DIR%\cli_scratch
 set BUILDING_DIR=%ARTIFACTS_DIR%\cli
+set ZIP_DIR=%ARTIFACTS_DIR%\Azure CLI
```
jiasli, line 51: A folder name with space (`Azure CLI`) causes lots of trouble when the full path is being invoked from PowerShell.

For example, the MSI installs the Azure CLI executable at `C:\Program Files\Microsoft SDKs\Azure\CLI2\wbin\az.cmd`.

It can't be directly executed because of the space in the path:

```
> C:\Program Files\Microsoft SDKs\Azure\CLI2\wbin\az.cmd
C:\Program: The term 'C:\Program' is not recognized as a name of a cmdlet, function, script file, or executable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
```

If we quote it, PowerShell treats it as a string and echoes it:

```
> "C:\Program Files\Microsoft SDKs\Azure\CLI2\wbin\az.cmd"
C:\Program Files\Microsoft SDKs\Azure\CLI2\wbin\az.cmd
```

The correct way to invoke the executable with its full path is:

```
> & "C:\Program Files\Microsoft SDKs\Azure\CLI2\wbin\az.cmd"

Welcome to Azure CLI!
---------------------
Use `az -h` to see available commands or go to https://aka.ms/cli.
```

Of course, we don't what such complex way of invoking `az.cmd`.

azure-pipelines.yml
```diff
@@ -228,6 +229,45 @@ jobs:
       TargetPath: 'build_scripts/windows/out/'
       ArtifactName: msi-$(Platform)
 
+- job: BuildWindowsZIP
+  displayName: Build Windows ZIP
+  strategy:
+    matrix:
+      x64:
+        Platform: x64
```
jiasli, line 237: I agree we should only release 64-bit ZIP as 32-bit is on its way to deprecation.

azure-pipelines.yml
```diff
@@ -259,6 +299,34 @@ jobs:
     inputs:
       filePath: build_scripts\windows\scripts\test_msi_installation.ps1
 
+- job: TestZipPackage
```
jiasli, line 302: The name `TestZipPackage` doesn't tell whether we are testing the installation or contents in the package (unit tests, scenario tests, etc).

I would suggest we align with `TestMsiInstallation`:

```suggestion
- job: TestZipInstallation
```

build_scripts/windows/scripts/azps_zip.ps1
jiasli, line 1: I am not sure if we should add this file as `azps.ps1` is known to be very buggy: #24015.

build_scripts/windows/scripts/build.cmd
jiasli: Where?

build_scripts/windows/scripts/build.cmd
```diff
@@ -137,18 +144,27 @@ if %errorlevel% neq 0 goto ERROR
...
-copy %REPO_ROOT%\build_scripts\windows\scripts\az.cmd %BUILDING_DIR%\wbin\
-copy %REPO_ROOT%\build_scripts\windows\scripts\azps.ps1 %BUILDING_DIR%\wbin\
-copy %REPO_ROOT%\build_scripts\windows\scripts\az %BUILDING_DIR%\wbin\
+if "%TARGET%"=="msi" (
+    copy %REPO_ROOT%\build_scripts\windows\scripts\az.cmd %BUILDING_DIR%\wbin\
+    copy %REPO_ROOT%\build_scripts\windows\scripts\azps.ps1 %BUILDING_DIR%\wbin\
+    copy %REPO_ROOT%\build_scripts\windows\scripts\az %BUILDING_DIR%\wbin\
+) else (
+    copy %REPO_ROOT%\build_scripts\windows\scripts\az.cmd %BUILDING_DIR%\wbin\
+    powershell -Command "(Get-Content '%BUILDING_DIR%\wbin\az.cmd') -replace 'SET AZ_INSTALLER=MSI','SET AZ_INSTALLER=ZIP' | Set-Content '%BUILDING_DIR%\wbin\az.cmd'"
```
jiasli, line 165: This string manipulation can be fragile if the contents of `%BUILDING_DIR%\wbin\az.cmd` changes. I suggest we create a separate `az_zip.cmd`, `az_msi.cmd` and rename then to `az.cmd` when packaging.

build_scripts/windows/scripts/build.cmd
```diff
@@ -193,14 +210,21 @@ for /d %%d in ("azure*.dist-info") do (
...
 
-echo Building MSI...
-msbuild /t:rebuild /p:Configuration=Release /p:Platform=%ARCH% %REPO_ROOT%\build_scripts\windows\azure-cli.wixproj
+if "%TARGET%"=="msi" (
+    echo Building MSI...
+    msbuild /t:rebuild /p:Configuration=Release /p:Platform=%ARCH% %REPO_ROOT%\build_scripts\windows\azure-cli.wixproj
+) else (
+    echo Building ZIP...
+    REM Rename zip root folder to "Azure CLI"
+    ren %BUILDING_DIR% "Azure CLI"
```
jiasli, line 220:
```
set BUILDING_DIR=%ARTIFACTS_DIR%\cli
set ZIP_DIR=%ARTIFACTS_DIR%\Azure CLI
```

`ren %BUILDING_DIR% "Azure CLI"` renames `BUILDING_DIR` to `%ARTIFACTS_DIR%\Azure CLI`. Let's not hard code `Azure CLI` but use a variable.

jiasli:
Windows File Explorer doesn't support extracting `LZMA`-compressed ZIP and fails with `0x80004005`:

image

This ZIP file is created by 

```
& "C:\Program Files\7-Zip\7z.exe" a -tzip -mm=LZMA msi_compressed.zip msi
```

jiasli:
[`PowerShell-7.4.1-win-x64.zip`](https://github.com/PowerShell/PowerShell/releases/download/v7.4.1/PowerShell-7.4.1-win-x64.zip) downloaded from https://github.com/PowerShell/PowerShell/releases/tag/v7.4.1 also uses `Deflate`:

```
& "C:\Program Files\7-Zip\7z.exe" l -slt .\PowerShell-7.4.1-win-x64.zip | Select-Object -First 30
...
Method = Deflate
```

build_scripts/windows/scripts/build.cmd
```diff
@@ -193,14 +210,21 @@ for /d %%d in ("azure*.dist-info") do (
...
+    REM Use LZMA compression to reduce the size of the zip file.
+    "%ProgramFiles%\7-Zip\7z.exe" a -tzip -m0=LZMA "%OUTPUT_DIR%\Microsoft Azure CLI.zip" "%ZIP_DIR%"
```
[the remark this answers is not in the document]
bebound: You're right.
According to the doc, `-m0` should only works in 7z and I should use `-mm`.
However, it works. I'll report this to 7z...

bebound:
> Windows File Explorer doesn't support extracting `LZMA`-compressed ZIP and fails with `0x80004005`:

Well spotted!
Windows explorer can **list** the content of LZMA zip but not able to **uncompress** it. (That's weird...)
We have to revert to `Deflate`.

build_scripts/windows/scripts/build.cmd
bebound: https://github.com/Azure/azure-cli/blob/6fa7aa45084d251a5b66b4d3c4dad1883b94d902/build_scripts/windows/scripts/build.cmd#L138

build_scripts/windows/scripts/azps_zip.ps1
bebound: I'm also inclined to remove it.

build_scripts/windows/scripts/build.cmd
```diff
@@ -43,6 +48,7 @@ set ARTIFACTS_DIR=%~dp0..\artifacts
 mkdir %ARTIFACTS_DIR%
 set TEMP_SCRATCH_FOLDER=%ARTIFACTS_DIR%\cli_scratch
 set BUILDING_DIR=%ARTIFACTS_DIR%\cli
+set ZIP_ROOT_FOLDER="AzureCLI"
```
jiasli, line 51: This will be the first time we introduce the term `AzureCLI`. Other alternatives can be `azure-cli`, `az`. 

However, I downloaded several other Microsoft-owned apps, like

- PowerShell: https://github.com/PowerShell/PowerShell/releases/download/v7.4.1/PowerShell-7.4.1-win-x64.zip from https://github.com/PowerShell/PowerShell/releases/tag/v7.4.1
- VS Code: https://code.visualstudio.com/sha/download?build=stable&os=win32-x64-archive from https://code.visualstudio.com/Download.
- PsTools: https://download.sysinternals.com/files/PSTools.zip from https://learn.microsoft.com/en-us/sysinternals/downloads/pstools

All put their content in the root folder.

build_scripts/windows/scripts/build.cmd
jiasli: Also see https://github.com/Azure/azure-cli/pull/27196#discussion_r1303879323

build_scripts/windows/scripts/build.cmd
```diff
@@ -137,18 +144,25 @@ if %errorlevel% neq 0 goto ERROR
...
 mkdir %BUILDING_DIR%\wbin
-copy %REPO_ROOT%\build_scripts\windows\scripts\az.cmd %BUILDING_DIR%\wbin\
-copy %REPO_ROOT%\build_scripts\windows\scripts\azps.ps1 %BUILDING_DIR%\wbin\
-copy %REPO_ROOT%\build_scripts\windows\scripts\az %BUILDING_DIR%\wbin\
+if "%TARGET%"=="msi" (
+    copy %REPO_ROOT%\build_scripts\windows\scripts\az_msi.cmd %BUILDING_DIR%\wbin\az.cmd
+    copy %REPO_ROOT%\build_scripts\windows\scripts\azps.ps1 %BUILDING_DIR%\wbin\
+    copy %REPO_ROOT%\build_scripts\windows\scripts\az %BUILDING_DIR%\wbin\
+) else (
+    copy %REPO_ROOT%\build_scripts\windows\scripts\az_zip.cmd %BUILDING_DIR%\wbin\az.cmd
```
jiasli, line 164: Other Microsoft-owned apps mentioned in https://github.com/Azure/azure-cli/pull/27911#discussion_r1454996530 put their executables such as `Code.exe`, `pwsh.exe` under the root folder. Perhaps we can do the same?

But this approach has its own problem: adding Azure CLI's directory to `PATH` will also put `python.exe` under `PATH`. If a user accidentally executes `python.exe`, that may corrupt Azure CLI's virtual environment.

jiasli: Why do we use `wbin`, instead of `bin`?

jiasli: Oh. The explanation is right in the comment: 

https://github.com/Azure/azure-cli/blob/04d82da1bf0a319ed38a9eabe7716119f50e402d/build_scripts/windows/scripts/build.cmd#L147-L148

bebound: Hmm, this package is for Windows, `w` seems redundant.

dcaro: > Other Microsoft-owned apps mentioned in [#27911 (comment)](https://github.com/Azure/azure-cli/pull/27911#discussion_r1454996530) put their executables such as `Code.exe`, `pwsh.exe` under the root folder. Perhaps we can do the same and put `az.cmd` under the root folder?
> 
> But this approach has its own problem: adding Azure CLI's folder to `PATH` will also put `python.exe` under `PATH`. If a user accidentally executes `python.exe`, that may corrupt Azure CLI's virtual environment.

Adding python in the path could also create instability if the platform (for example a custom build agent) is using python for other purposes.

I think that `wbin` will be confusing and the information it adds is not necessary.

bebound: @dcaro I can change the folder to `bin` in zip package.
Should we also do this in MSI? We never mention `wbin` in doc, and it's automatically added into PATH during MSI installation. This change should not affect users unless they use full path to run `az`.

dcaro: let's leave the folder name in the msi as is for now, as you mention, it is automatically added to path and transparent to users.

build_scripts/windows/scripts/build.cmd
```diff
@@ -137,18 +143,27 @@ if %errorlevel% neq 0 goto ERROR
...
 REM Remove pywin32 help file to reduce size.
 del %BUILDING_DIR%\Lib\site-packages\PyWin32.chm
 
-echo Creating the wbin (Windows binaries) folder that will be added to the path...
-mkdir %BUILDING_DIR%\wbin
-copy %REPO_ROOT%\build_scripts\windows\scripts\az.cmd %BUILDING_DIR%\wbin\
-copy %REPO_ROOT%\build_scripts\windows\scripts\azps.ps1 %BUILDING_DIR%\wbin\
-copy %REPO_ROOT%\build_scripts\windows\scripts\az %BUILDING_DIR%\wbin\
+if "%TARGET%"=="msi" (
+    REM Creating the wbin (Windows binaries) folder that will be added to the path...
```
bebound, line 157: If I use `echo` instead of `REM`, the `(Windows binaries) folder` will be parsed and `folder was unexpected at this time.` error is raised.
Repro script:
``` cmd
echo () ok
if 'kk'=='kk' (
    echo () ok2
)
```
```
() ok
ok2 was unexpected at this time.
```

jiasli: If you truly need to use `echo`, you may quote the sentence:

```cmd
echo "Creating the wbin (Windows binaries) folder that will be added to the path..."
```

bebound: Exactly. (Generally, `echo` is able to print multiple word without the quote: https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/echo)

In this instance, since the line functions more as a comment, so I changed it to `REM`.

evelyn-ys (approved):

jiasli:
Downloaded from https://azcliprod.blob.core.windows.net/zip/azure-cli-2.57.0-x64.zip and the ZIP package works like a charm! 🎉

jiasli:
Another great benefit of the ZIP package is that it allows much easier edge build release.

Currently, in order to install an MSI edge build, I have to

1. Uninstall the official MSI
2. Install the edge buld MSI

But with ZIP, I can simply extract it and run `bin/az.cmd`!

PR #28059, opened 2023-12-19
{Compute} `az sig update`: Set SoftDeletePolicy to null if no `--soft-delete-policy`

### LINKED ISSUE #27992: Az sig update SoftDeletePolicy keep to be null if user does not specify
**Related command**
az sig update

Yes.

Current scenario: user does 'az sig update' without '--soft-delete-policy', CLI will still pass in SoftDeletePolicy to the backend if the Get result has SoftDeletePolicy.

Ask: When user does 'az sig update' without '--soft-delete-policy', CLI do not pass in SoftDeletePolicy to the backend(set it as null). If user specify '--soft-delete-policy' in CLI, then CLI pass in whatever user passed to backend.

Otherwise we would get icm like https://portal.microsofticm.com/imp/v3/incidents/details/446631906/home

Described above.

An alternative is that we remove all DM of softDeletePolicy, but that could be a huge effort.

ReaNAiveD:
**Related command**

az sig update

**Description**

Close #27992 

When user does 'az sig update' without '--soft-delete-policy', CLI do not pass in SoftDeletePolicy to the backend(set it as null). If user specify '--soft-delete-policy' in CLI, then CLI pass in whatever user passed to backend.

The service team has patch semantics, which means that if the SoftDeletePolicy input is null, the service will not clear this attribute, but will keep it unchanged. The logic would be deleted after the feature is public preview/GA. 

**Testing Guide**

**History Notes**

---


yonzhan:
Compute

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -4574,6 +4574,8 @@ def update_image_galleries(cmd, resource_group_name, gallery_name, gallery, perm
             gallery.soft_delete_policy.is_soft_delete_enabled = soft_delete
         else:
             gallery.soft_delete_policy = {'is_soft_delete_enabled': soft_delete}
+    else:
+        gallery.soft_delete_policy = None
```
zhoxing-ms, line 4578: Because this is a special temporary logic, could we add comments to explain it:
This is a workaround to solve historical legacy issues, send `None` to the service will let service not modify this property. We can delete this logic when the service no longer checks AFEC in the future.

zhoxing-ms:
Please resolve the conflict

src/azure-cli/azure/cli/command_modules/vm/custom.py
ReaNAiveD: Done

zhoxing-ms (approved):

PR #28096, opened 2023-12-26
{CI} Update policy config

wangzelin007:
**Related command**

**Description**

1. Use `\b` for more precise matching to avoid misjudgments.
2. Disable using the install keyword to match install-related issues, as the misjudgment rate is too high.

**Testing Guide**

**History Notes**

---

### DIFF
--- a/.github/policies/resourceManagement.yml
+++ b/.github/policies/resourceManagement.yml
@@ -3067,7 +3067,7 @@ configuration:
             pattern: '\b([Ii][Oo][Tt])\b'
             isRegex: True
         - bodyContains:
-            pattern: '\b([Ii][Oo][Tt])\b'
+            pattern: '\b(az iot)\b'
             isRegex: True
       - or:
         - isAction:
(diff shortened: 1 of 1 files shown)


yonzhan:
CI

jsntcy:
@wangzelin007, please add more details in description, for example, why we need this change? how we do it?

wangzelin007:
Why we need to add `\b` ?
`\b` represents the word boundary in regular expressions, and more precise matching can avoid misjudgments.
Here's an example.
We have the two commands `az storage` and `az storagesync`
If we use `az storage`, will match both
If we use `\b(az storage)\b`, it will only match az storage.
image

.github/policies/resourceManagement.yml
```diff
@@ -4477,27 +4477,27 @@ configuration:
...
-      then:
-      - addLabel:
-          label: Auto-Assign
-      - addLabel:
-          label: Installation
-      - assignTo:
-          users:
-          - jiasli
-      description: '[install] auto assign labels and users based on issue description.'
+#    - if:
```
wangzelin007, line 4480: The misjudgment rate of using the install keyword to match install-related issues is too high, so this task is temporarily disabled.

jsntcy: So for issues really related to installation, we will not assign label automatically after this change?

wangzelin007: Yes, without using keywords, we have no better way to distinguish whether it is an install-related issues.

jsntcy (approved):

PR #28150, opened 2024-01-08
[Backup] `az backup restore restore-disks`: Add support for disk restore in edge-zone backups and support for TWN + TNW CRR

zubairabid:
**Related command**

`az backup restore restore-disks`

**Description**

 Added the `--restore-to-edge-zone` three-state-flag parameter for `restore-disks`. When this parameter is provided, and for OriginalLocation restores, the extendedLocation property of the trigger restore request body is populated with the equivalent details from the recovery point.

**Testing Guide**

For a VM deployed to an edge zone, backed up to a vault, we can trigger a restore with:
- `--restore-mode OriginalLocation`, or
- `--restore-to-edge-zone`
and we should be able to back up successfully. If we inspect request body, it will contain an ExtendedLocation element.

Automated tests have not been written as the testing subscription does not have edge zones enabled yet, but it has been tested manually.

**History Notes**

[Backup] Add support for disk restore in edge-zone backups
[Backup] Support for Taiwan North and Taiwan North-West Cross region restore

---


src/azure-cli/azure/cli/command_modules/backup/custom_wl.py
```diff
@@ -135,8 +135,8 @@ def register_wl_container(cmd, client, vault_name, resource_group_name, workload
     param = ProtectionContainerResource(properties=properties)
 
     # Trigger register and wait for completion
-    result = client.register(vault_name, resource_group_name, fabric_name, container_name, param,
-                             cls=cust_help.get_pipeline_response)
+    result = client.begin_register(vault_name, resource_group_name, fabric_name, container_name, param,
+                                   cls=cust_help.get_pipeline_response, polling=False).result()
```
zhoxing-ms, line 139: May I ask why you need to disable the logic of polling in the code?

zubairabid: The SDK was changed to return LROPoller instead, so I made the change to parse it correctly.

zhoxing-ms: Yes, what I mean is what problems would there be if using polling?

zhoxing-ms: You can try to obtain results from `LROPoller` in the following ways

https://github.com/Azure/azure-cli/blob/2caee1484074c595a836080aa4586b651b40380a/src/azure-cli/azure/cli/command_modules/network/custom.py#L4158

zubairabid: > Yes, what I mean is what problems would there be if using polling?

I'm not entirely sure, honestly. I'm just following the pattern we've been using previously in our code:

https://github.com/Azure/azure-cli/blob/2caee1484074c595a836080aa4586b651b40380a/src/azure-cli/azure/cli/command_modules/backup/custom.py#L1403

zhoxing-ms: Okay, you can keep the original code style for now. If you have a need for the progress bar, you can consider changing the code style to the one we recommend in the future.

zhoxing-ms (approved):

PR #28208, opened 2024-01-16
{App Service} Clean error messages relating to ASP creation

amberrenton:
**Related command**

```az
az appservice plan create
```

**Description**

Cleans outdated error messages related to App Service Plan creation. We no longer append "Plan with Linux worker can only be created in a group which has never contained a Windows worker, and vise versa." because this is no longer true (and has not been true for 2+ years).

We also make a Linux-specific error more generic in `_client_factory.py`, as nothing guarantees that this is only hit while creating a Linux App Service Plan.

**Testing Guide**

I was able to reproduce the original error message with the following commands:

```az
az group create -n testgroup2 -l westus3
az appservice plan create --name windowstestplan --sku p1v3 -g testgroup2 --zone-redundant --number-of-workers 10
az appservice plan create --name hypervtestplan --sku p1v3 -g testgroup2 --zone-redundant --hyper-v --number-of-workers 20
```

image

Re-running the same commands with the changes from this PR, we instead see the following output:

image

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/_client_factory.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/_client_factory.py
@@ -13,13 +13,8 @@ def ex_handler_factory(creating_plan=False, no_throw=False):
             detail = json.loads(ex.response.text)['Message']
             if creating_plan:
                 if 'Requested features are not supported in region' in detail:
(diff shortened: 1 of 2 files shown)


yonzhan:
App Service

jvano (approved):
Looks good!

zhoxing-ms (approved):

PR #28257, opened 2024-01-26
{AppService} `az functionapp config container set`: Check if dapr config is null before updating settings

kamperiadis:
**Related command**

**Description**
When customers create Centauri function apps through the Portal, the daprConfig is set to null. If customers then try to update the daprConfig, workload profile name, cpu or memory through the CLI with the command `az functionapp config container set`, they will get an exception because we do not currently check that daprConfig is not null. This PR makes sure that daprConfig is not null. 

**Testing Guide**
1. Create Centauri function app through the Portal
2. Attempt to update the workload profile name, cpu and memory. You will get a failure without the changes in this PR. 

---


yonzhan:
AppService

khkh-ms (approved):

yanzhudd (approved):

PR #28291, opened 2024-02-01
{AppService} `az functionapp create`: Updating DAPR config to match experience in Portal

kamperiadis:
**Related command**
`az functionapp create`

**Description**
1. Make HTTP read buffer size and HTTP max request size default to 4 if no values were provided 
2. Set daprConfig to null if it is disabled
3. Throw exception to warn customer that if daprConfig is disabled, the other dapr-related attributes are not valid

**Testing Guide**
1. `az functionapp create ... --enable-dapr false` should set daprConfig to null
2. `az functionapp create ... --enable-dapr true` requires --dapr-app-id to be passed (validation is thrown from the backend)
3. `az functionapp create ... --enable-dapr true --dapr-app-id` will create the daprConfig
4.`az functionapp config container set --enable-dapr false` will set daprConfig to null
5. `az functionapp config container set --enable-dapr true --dapr-app-id` will create the daprConfig
6. `az functionapp config container set --enable-dapr None` will keep the same enable/disable value. It is only when daprConfig is enabled that the dapr-related attributes can be updated 

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -3903,7 +3903,7 @@ def update_functionapp_polling(cmd, resource_group_name, name, functionapp):
     updated_functionapp = json.dumps(
         {
             "properties": {
-                "daprConfig": None if functionapp.dapr_config is None else {
+                "daprConfig": {"enabled": False} if functionapp.dapr_config is None else {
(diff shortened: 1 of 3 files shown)


yonzhan:
AppService

arroyc (approved):

yonzhan:
Please fix CI issues.

zhoxing-ms (approved):

PR #28575, opened 2024-03-18
[NetAppFiles] Bump ANF to api-version 2023-07-01

audunn:
**Related command**

**Description**

**Testing Guide**

**History Notes**

[NetAppFiles] `az netappfiles`: Update to api-version 2023-07-01

---


yonzhan:
NetAppFiles

audunn:
aaz https://github.com/Azure/aaz/pull/440

audunn:
The breaking change flagged is due to that the last version, that is still not released to public btw, will be our first version that is generated using aaz from swagger.  In that version we found that it exposes a property on volume called `network-features` in `volume update` command as it is picked up from swagger.  

In previous hand-writter versions volume update was a `PATCH` and in the `PATCH` body this property does not exist.  That is beause this property cannot be set in update senario only during volume creation due to networking requrements and complications.   The generated command uses the azure cli convention of doing a GET/PUT.
The workaround is for customers can use query-networksiblingset and update-networksibling set to change this for all volumes on the same networking sibling set (volumes can share nic).

This can cause error if customers would try to use this paramter and that is why we are removing it now.

Also `network-features` can also only be set to `Basic` or `Standard` by customers.

src/azure-cli/azure/cli/command_modules/netappfiles/aaz/latest/netappfiles/volume/_update.py
```diff
@@ -327,13 +327,6 @@ def _build_arguments_schema(cls, *args, **kwargs):
             help="Specifies whether LDAP is enabled or not for a given NFS volume.",
             nullable=True,
         )
-        _args_schema.network_features = AAZStrArg(
-            options=["--network-features"],
-            arg_group="Properties",
-            help="Basic network, or Standard features available to the volume.",
-            nullable=True,
-            enum={"Basic": "Basic", "Basic_Standard": "Basic_Standard", "Standard": "Standard", "Standard_Basic": "Standard_Basic"},
-        )
```
kairu-ms, line 336: There's a breaking change which removed the '--network-features` argument?

kairu-ms (approved):

PR #28726, opened 2024-04-10
[App Service] `az functionapp create`: Remove unnecessary app settings for flex function apps

kamperiadis:
**Related command**
`az functionapp create`

**Description**
- The backend logic has blocked the creation of flex function apps with the FUNCTIONS_EXTENSION_VERSION app setting. Therefore, we need to make sure that this app setting is not added in the creation flow. 
 - Additionally, we should not be setting distributed tracing on by default for flex function apps. 
 
**Testing Guide**
- `az functionapp create` for a flex app no longer adds the FUNCTIONS_EXTENSION_VERSION app setting
- `az functionapp create` for a java flex app no longer adds the APPLICATIONINSIGHTS_ENABLE_AGENT app setting

**History Notes**
- [AppService] `az functionapp create`: Remove FUNCTIONS_EXTENSION_VERSION app setting for flex function apps
- [AppService] `az functionapp create`: Remove APPLICATIONINSIGHTS_ENABLE_AGENT app setting for flex function apps
- [AppService] `az functionapp create`: Remove AzureWebJobsDashboard app setting for flex function apps
- [AppService] `az functionapp create` and `az functionapp list-flexconsumption-locations`: Make request to geo regions API
- [AppService] `az functionapp create` and `az functionapp vnet-integration add`: Register subscription to Microsoft.App RP for flex function app vnet integration
- [AppService] `az functionapp create`: Delete newly created resources specific to flex if the app creation fails

---


yonzhan:
AppService

src/azure-cli/azure/cli/command_modules/appservice/tests/latest/test_functionapp_commands.py
```diff
@@ -31,7 +31,7 @@
 WINDOWS_ASP_LOCATION_CHINACLOUD_WEBAPP = 'chinaeast'
 
 
-class FunctionappACRScenarioTest(ScenarioTest):
+class FunctionappACRScenarioTest(LiveScenarioTest):
```
kamperiadis, line 34: Had to turn these tests into LiveScenarioTests since they work when run with the `--live` flag but when using the recordings, somehow `azdev test` fails even though the API responses in the yaml files actually show our checks are correct.

kamperiadis: These tests are not actually impacted by the changes in custom.py so I am not sure why they suddenly started failing.

src/azure-cli/azure/cli/command_modules/appservice/tests/latest/test_functionapp_commands.py
```diff
@@ -1500,7 +1523,7 @@ def test_functionapp_without_default_distributed_tracing(self, resource_group, s
                         kp['name'] for kp in app_set])
 
 
-class FunctionAppWithAppInsightsDefault(ScenarioTest):
+class FunctionAppWithAppInsightsDefault(LiveScenarioTest):
```
zhoxing-ms, line 1526: May I ask why do you need to mark this test as `LiveScenarioTest`?

src/azure-cli/azure/cli/command_modules/appservice/tests/latest/test_functionapp_commands.py
zhoxing-ms: Could you please show me the error message when this test is `ScenarioTest` ?

kamperiadis: This is the error message:
`azure.cli.testsdk.exceptions.JMESPathCheckAssertionError: Query '[?name=='DOCKER_REGISTRY_SERVER_USERNAME']|[0].value' doesn't yield expected value 'functionappacrtest000004', instead the actual value is 'None'.`

It appears it is trying to use the response from the first `config appsettings list` call instead of the second call.

zhoxing-ms:
> [AppService] az functionapp create: Remove FUNCTIONS_EXTENSION_VERSION app setting for flex function apps
[AppService] az functionapp create: Remove APPLICATIONINSIGHTS_ENABLE_AGENT app setting for flex function apps
[AppService] az functionapp create: Remove AzureWebJobsDashboard app setting for all function app SKUs

I would like to confirm if these changes will cause a breaking change for customers?

src/azure-cli/azure/cli/command_modules/appservice/tests/latest/test_functionapp_commands.py
zhoxing-ms: Out of curiosity, could you confirm that running with the dev branch (excluding the code of this PR) also has the same issue? If you can confirm that this issue was not introduced by this PR, then we can skip this issue first and fix it when you have time conditioning to investigate and solve this problem in the future

kamperiadis:
> > [AppService] az functionapp create: Remove FUNCTIONS_EXTENSION_VERSION app setting for flex function apps
> > [AppService] az functionapp create: Remove APPLICATIONINSIGHTS_ENABLE_AGENT app setting for flex function apps
> > [AppService] az functionapp create: Remove AzureWebJobsDashboard app setting for all function app SKUs
> 
> I would like to confirm if these changes will cause a breaking change for customers?

These changes would not cause a breaking change for customers.

src/azure-cli/azure/cli/command_modules/appservice/tests/latest/test_functionapp_commands.py
kamperiadis: @zhoxing-ms Just made changes. We really just need to remove the AzureWebJobsDashboard app setting for flex apps. After making that change and rerunning the tests, we no longer have issues when they are `ScenarioTest` so I switched them back to `ScenarioTest` from `LiveScenarioTest`.

zhoxing-ms (approved):

Issue #27771, opened 2023-11-05
Update to application gateway address pool reverts unrelated previous app gateway address pool update


agra6475:
### Describe the bug

Using azure devops to update multiple backend pools of a single application gateway. 
Updates to the application gateway are done concurrently.

#### Job 1
One job executed this command to update a backend pool.

```powershell
az network application-gateway address-pool update `
    --gateway-name someName `
    --name oneBackend `
    --resource-group someResourceGroup `
    --servers someNewPrivateIP
```
The command completes at 7:59:40
![image](https://github.com/Azure/azure-cli/assets/149966774/07f22d7d-58f4-4c7d-bf65-83e40b6555a1)
And the provisioning state of the application gateway changes as expected
![image](https://github.com/Azure/azure-cli/assets/149966774/f59c064c-96fe-4351-ad25-c012f1e81a64)

At  7:59:45 the provisioning state changes again as the app gateway has completed the update
![image](https://github.com/Azure/azure-cli/assets/149966774/42b92deb-7595-4463-a1b4-8dd2097ef7d6)

#### Job 2
At 7:59:51 the following command completes
```powershell
az network application-gateway address-pool update `
  --gateway-name someName  `
  --name aDifferentbackend `
  --resource-group someResourceGroup `
  --servers someNewPrivateIP
```
As expected it updates `aDifferentbackend`
![image](https://github.com/Azure/azure-cli/assets/149966774/abb9a11e-dae2-4cfb-8339-95f230d16b8d)

What is not expected is that it reverts the change to `oneBackend`
![image](https://github.com/Azure/azure-cli/assets/149966774/568390a8-c1a0-434d-a3a1-770d5074e6e8)

### Related command

Job 1 command
```powershell
az network application-gateway address-pool update `
    --gateway-name someName `
    --name oneBackend `
    --resource-group someResourceGroup `
    --servers someNewPrivateIP
```

Job 2 command
```powershell
az network application-gateway address-pool update `
  --gateway-name someName  `
  --name aDifferentbackend `
  --resource-group someResourceGroup `
  --servers someNewPrivateIP
```

### Errors

No error happened during these commands.

### Issue script & Debug output

Not able to reproduce it.
Only has the output from the first command.

```json
{
  "backendAddresses": [
    {
      "ipAddress": "10.0.0.7"
    }
  ],
  "etag": "W/\"3fe05e13-ff6d-4907-89a9-f2a3651b05b0\"",
[...]

PR #28777, opened 2024-04-18
[Network] Fix #28615: `az network application-gateway address-pool update`: Race condition in concurrent scenario

### LINKED ISSUE #28615: Az Cli Gateway updates roll back each other
### Describe the bug

When updating an Application Gateway with Azure CLI, under specific timings, two calls might end up with one of them overwriting the other.

The problem is an interaction between Application Gateway and the default retry logic of the Azure CLI. When a system returns a retryable error, Azure CLI will retry the exact same command once more. This does not work for Application Gateway though, as it returns a 429 - Retryable error if two commands happen at the same time (or close to). Because Application Gateway takes the entire state of the resource in its Put command, retrying will always delete the change that competed with it and caused the 429 in the first place. In other words, getting this retryable error and handling it will revert changes.
 
Solution suggestions:

- Azure CLI is changed so that it retrieves the new status of the application gateway and does its change to it again before retrying
- The Application Gateway is changed to send a non-retryable error and handling is left to customers as Azure CLI gives an error
- The put endpoint in application gateway is changed to take the e-tag the change is based on and reject changes based on outdated settings.
- Azure CLI implements a command that allows disabling retries

A major problem in the current solution is that Azure CLI completely hides the error from us - we do not, without --debug enabled, get any indication that there has been any trouble and everything passes correctly.

The below graph shows how the issue happened when two pipelines (running independently of each other) ended up with one removing the change the other had made. The graph should be viewed as a timeline, with items further down on the graph happening later.

Annotation 2024-03-22 125804

### Related command

az network application-gateway address-pool update

### Errors

No error messages are received by default. With --debug on, the error is shown as a 429 - RetryableError response.

### Issue script & Debug output

Command (note that this works fine outside of the timing issues)
` $verificationAddressPool = az network application-gateway address-pool update --gateway-name $appGatewayName  --name $verificationAddressPoolName --resource-group $AppGatewayResourceGroup  --servers $privateEndpointIp `

Adding a debug log to this would be cumbersome and nearly unreadable as it would include several debug logs that would need to be compared to each other to understand the issue. See the above graph instead.

### Expected behavior

Two sequential updates that ends with both solutions applied

### Environment Summary

azure-cli                         2.58.0

core                              2.58.0
telemetry                          1.1.0

Extensions:
azure-devops                       1.0.0

Dependencies:
msal                              1.26.0
azure-mgmt-resource             23.1.0b2

Python location '/opt/az/bin/python3'
Extensions directory '/opt/az/azcliextensions'

Python (Linux) 3.11.7 (main, Feb 29 2024, 02:08:19) [GCC 11.4.0]

### Additional context

We have tried with `az network application-gateway wait --resource-group $ResourceGroup --name $AppGatewayName --updated` but this does not work as an update is not in progress before both commands finish their get call

necusjz:
**Related command**

**Description**

Fix: https://github.com/Azure/azure-cli/issues/28615
Fix: https://github.com/Azure/azure-cli/issues/27771

**Testing Guide**

**History Notes**

---


yonzhan:
Network

agra6475:
This should also fix #27771

kairu-ms (approved):

PR #28778, opened 2024-04-18
{Core} Proof-of-Concept: Support OIDC token refreshing on GitHub Actions

### LINKED ISSUE #28708: Azure CLI task fails: AADSTS700024: Client assertion is not within its valid time range
Acquiring access token with expired OIDC token fails with:

> ERROR: AADSTS700024: Client assertion is not within its valid time range. Current time: 2024-04-05T23:01:54.2089203Z, assertion valid from 2024-04-05T22:40:41.0000000Z, expiry time of assertion 2024-04-05T22:50:41.0000000Z. Review the documentation at https://docs.microsoft.com/azure/active-directory/develop/active-directory-certificate-credentials

As the error indicates, the OIDC token is only valid for 10 minutes. After it is passed to `az login` via `--federated-token`, Azure CLI cannot get a new OIDC token after the OIDC token expires.

This is the designed v1 behavior of OIDC token support (#19853).

However, as Azure DevOps task **AzureCLI@2** (https://github.com/microsoft/azure-pipelines-tasks/pull/17633) and GitHub Action **azure/login@v2** (https://github.com/Azure/login/pull/147) have supported OIDC token authentication, and it is recommended to use [workload identity federation](https://learn.microsoft.com/en-us/entra/workload-id/workload-identity-federation), this limitation is becoming more prevailing.

## Possible solutions

1. OIDC token provider such as Azure DevOps or GitHub should provide an option to control the expiry time of the OIDC token to make it at least as long as the task duration.
2. Design and implement a v2 solution that uses a managed-identity-like interface which allows MSAL/Azure CLI to refresh OIDC token.

## References

- https://github.com/Azure/login/issues/180
- https://github.com/Azure/login/issues/372
- IcM 490937309
- IcM 491234676
- Email: Workload identity federation in Azure Pipelines fails with AADSTS700024

jiasli:
Fix https://github.com/Azure/azure-cli/issues/28708

**Related command**
`az login`

**Description**
MSAL reserved an undocumented interface which allows using a callback function to get `client_assertion`, instead of supplying a raw OIDC token.

https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/a8476fbe665356aae60e65296337d9eff86d0d66/msal/oauth2cli/oauth2.py#L198-L201

```py
            _data["client_assertion"] = encoder(
                self.client_assertion()  # Do lazy on-the-fly computation
                if callable(self.client_assertion) else self.client_assertion
                )
```

As a Proof-of-Concept, this PR uses this interface to achieve OIDC token refreshing.

Some interesting finds: 

When a task is run on 2024-04-18 10:35:31, the `ACTIONS_ID_TOKEN_REQUEST_TOKEN` has below claims:

```jsonc
  "nbf": 1713435930,  // 2024-04-18 10:25:30
  "exp": 1713458730   // 2024-04-18 16:45:30
```

This means the token starts 10min before the task and expires after 6h10min. After the token expires, no ID token can be retrieved. Therefore, even if we can refresh ID token, the maximum execution time of a task is around 6h.

References:
- https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-cloud-providers#requesting-the-jwt-using-environment-variables
- https://github.com/actions/toolkit/blob/5a736647a123ecf8582376bdaee833fbae5b3847/packages/core/src/oidc-utils.ts
- https://github.com/Azure/azure-cli/issues/28708#issuecomment-2049014471
- https://github.com/github/docs/issues/32573

**Testing Guide**
[Only for development and debugging] 

1. To get all environment variables on the agent, add this Bash command `env | base64` to bypass GitHub Actions' credential redaction. Then decode it with VS Code.
2. You have to use `sleep 1800` to keep the task alive, otherwise, the OIDC token request will be rejected after the task finishes.


yonzhan:
Core

src/azure-cli-core/azure/cli/core/auth/msal_authentication.py
```diff
@@ -138,3 +138,22 @@ def get_token(self, *scopes, **kwargs):
...
         check_result(result)
         return build_sdk_access_token(result)
+
+
+def get_id_token_on_github():
+    import os
+    from urllib.parse import quote
+    import requests
+    token = os.environ['ACTIONS_ID_TOKEN_REQUEST_TOKEN']
+    url = os.environ['ACTIONS_ID_TOKEN_REQUEST_URL']
```
jiasli, line 148: Use this workflow file to collect environment variables:

```yaml
on: [push, workflow_dispatch]

name: AzureCLISample

permissions:
  id-token: write
  contents: read

jobs:
  build-and-deploy:
    runs-on: ubuntu-latest
    steps:
    - name: Azure CLI script
      uses: azure/CLI@v1
      with:
        azcliversion: latest
        inlineScript: |          
          set -ex
          env | base64
```

jarpoole:
@jiasli any chance this could get some more love?

HarleyB123:
@jiasli Is there an update on this? Or has a feature been created elsewhere that assists with the management/refreshing of OIDC tokens in GitHub without scopes/running az login as a background process (which has it's race condition issues)?

PR #28813, opened 2024-04-23
[AKS] `az aks create/update`: Add new parameter `--enable-cost-analysis` to enable exporting Kubernetes namespace and deployment details to the Cost Analysis views

matthchr:
**Related command**
az aks create
az aks update

**Description**
Add --enable-cost-analysis and --disable-cost-analysis to az aks cli

**Testing Guide**

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acs/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/custom.py
@@ -630,6 +630,8 @@ def aks_create(
     no_wait=False,
     aks_custom_headers=None,
     node_public_ip_tags=None,
+    # metrics profile
+    enable_cost_analysis=False,
 ):
     # DO NOT MOVE: get all the original parameters and save them as a dictionary
     raw_parameters = locals()
@@ -779,6 +781,9 @@ def aks_update(
     yes=False,
     no_wait=False,
     aks_custom_headers=None,
+    # metrics profile
+    enable_cost_analysis=False,
+    disable_cost_analysis=False,
 ):
     # DO NOT MOVE: get all the original parameters and save them as a dictionary
     raw_parameters = locals()
--- a/src/azure-cli/azure/cli/command_modules/acs/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/_help.py
@@ -537,6 +537,9 @@ parameters:
   - name: --revision
     type: string
     short-summary: Azure Service Mesh revision to install.
+  - name: --enable-cost-analysis
+    type: bool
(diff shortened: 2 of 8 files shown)


yonzhan:
AKS

FumingZhang (approved):
LGTM

Queued [live test](https://dev.azure.com/msazure/CloudNativeCompute/_build/results?buildId=92176145&view=results) to validate the change.

andyliuliming (approved):

matthchr:
@andyliuliming - can we merge this now?

src/azure-cli/azure/cli/command_modules/acs/custom.py
```diff
@@ -779,6 +781,9 @@ def aks_update(
     yes=False,
     no_wait=False,
     aks_custom_headers=None,
+    # metrics profile
+    enable_cost_analysis=False,
+    disable_cost_analysis=False,
```
zhoxing-ms, line 786: These two default values are both False, which seems contradictory. Could they both be None?

matthchr: It's not really contradictory. If you are neither enabling or disabling this feature, then the default behavior is to leave it alone (its value stays whatever it was before).

There are many other pairs of enable/disables with the same semantics in the AKS CLI, see for example:
```
    enable_image_cleaner=False,
    disable_image_cleaner=False,
    enable_keda=False,
    disable_keda=False,
    enable_vpa=False,
    disable_vpa=False,
    enable_disk_driver=False,
    disable_disk_driver=False,
    enable_file_driver=False,
    disable_file_driver=False,
    enable_workload_identity=False,
    disable_workload_identity=False,
    enable_snapshot_controller=False,
    disable_snapshot_controller=False,
    enable_azure_keyvault_kms=False,
    disable_azure_keyvault_kms=False,
```

(there are more).

Throwing `None` in the mix doesn't really clarify things, as `None` behaves just like `False` would and so there becomes no difference between `enable=false, disable=false` and `enable=None, disable=None`.

FumingZhang: As a supplement, aks command group usually provides a pair of options for setting on or off for a feature (`--enable-xxx`/`--disable-xxx`). Based on the previous discussion, to maintain the consistency of similar options, AKS PM decided not to use the three-state flag, so `False` here does not really mean `False` (~~`enable=False` -> `disable`~~), but means `None`.

zhoxing-ms: Got it, thanks~

PR #28853, opened 2024-04-29
{testsdk} Copy command index to random config dir

jiasli:
Temporally bypass https://github.com/Azure/azure-cli/issues/28848
Final fix should be done by https://github.com/Azure/azure-cli/pull/28849

**Description**
Running `test_containerapp_manualjob_withsecret_crudoperations_e2e` and `test_hdinsight_application` sequentially fails with 

```
$ azdev test test_containerapp_manualjob_withsecret_crudoperations_e2e test_hdinsight_application --series -a -rP
...
self = <azure.cli.core._session.Session object at 0x7f15c1b06410>

    def save(self):
        if self.filename:
>           with open(self.filename, 'w', encoding=self._encoding) as f:
E           FileNotFoundError: [Errno 2] No such file or directory: '/home/user2/.azure/dummy_cli_config_dir/mvDaOhlAGfu4OvI6/commandIndex.json'

src/azure-cli-core/azure/cli/core/_session.py:54: FileNotFoundError
```

As explained in https://github.com/Azure/azure-cli/issues/28848#issuecomment-2081967874, even after `test_containerapp_manualjob_withsecret_crudoperations_e2e` finishes and deletes the random config dir, `test_hdinsight_application` still tries to initialize command index in the same random config dir, because of the polluted env var `AZURE_CONFIG_DIR` (https://github.com/Azure/azure-cli/pull/28849#issue-2267820600).

This PR copies command index to random config dir first so that it can be loaded during `azure.cli.core.AzCli.__init__`:

https://github.com/Azure/azure-cli/blob/f98ba8062f08bf5b7460260c56916bd3074bf9e5/src/azure-cli-core/azure/cli/core/__init__.py#L82

which is called during `azure.cli.testsdk.base.ScenarioTest.__init__`:

https://github.com/Azure/azure-cli/blob/2750c65ea77638ecf699243555a5f2f7a1314ed4/src/azure-cli-testsdk/azure/cli/testsdk/base.py#L86

Then command index will not be initialized again by test methods.

**Testing Guide**

```
azdev test test_containerapp_manualjob_withsecret_crudoperations_e2e test_hdinsight_application --series -a -rP
```

PR #28673, opened 2024-04-02
{Core} Fix `random_config_dir` when `AZURE_CONFIG_DIR` is set

bebound:
Resolve #28668

Knack ignores the `config_dir` param when `AZURE_CONFIG_DIR` env is set, and other functions may call `get_config_dir` directly.
So this PR patches `AZURE_CONFIG_DIR` env when `random_config_dir=true` during test

PR #28857, opened 2024-04-30
{testsdk} Stop patching `AZURE_CONFIG_DIR` immediately after calling `AzCli.__init__`

jiasli:
**Description**
This PR fixes `patch.dict` pollution: `azure.cli.testsdk.base.ScenarioTest.__init__` of each test class is run at pytest's collecting staging, in the same process, but `tearDown` is executed after test methods have been finished. As no test method has been run yet, `patch.dict(os.environ, {'AZURE_CONFIG_DIR': config_dir})` is not stopped, thus polluting other test classes' `__init__` methods at collecting stage and other test methods at running stage.

Consider

```
azdev test test_containerapp_manualjob_withsecret_crudoperations_e2e test_hdinsight_application --live --series -a -rP
```

Even though https://github.com/Azure/azure-cli/pull/28853 bypasses the command index initialization and makes playback test work, during live test, after `test_containerapp_manualjob_withsecret_crudoperations_e2e` finishes, the random config dir is deleted. Because of the `patch.dict` pollution, `test_hdinsight_application` uses the same random config dir, but can't find `msal_token_cache.json` and fails:

```
        if not accounts:
>           raise CLIError("User '{}' does not exist in MSAL token cache. Run `az login`.".format(username))
E           knack.util.CLIError: User 'xxx@microsoft.com' does not exist in MSAL token cache. Run `az login`.

src/azure-cli-core/azure/cli/core/auth/msal_authentication.py:58: CLIError
```

This PR doesn't make things worse than #28673 as currently the `AZURE_CONFIG_DIR` env var is already polluted when running test methods. For example:

```
Collecting:
Class A -> patch AZURE_CONFIG_DIR=random_config_dir_A
Class B -> patch AZURE_CONFIG_DIR=random_config_dir_B

Running:
test_method_A -> see AZURE_CONFIG_DIR=random_config_dir_B
```

When running `test_method_A`, `AZURE_CONFIG_DIR` is `random_config_dir_B`.

**Testing Guide**
```
azdev test test_containerapp_manualjob_withsecret_crudoperations_e2e test_hdinsight_application --live --series -a -rP
```

### DIFF
--- a/src/azure-cli-core/azure/cli/core/mock.py
+++ b/src/azure-cli-core/azure/cli/core/mock.py
@@ -25,12 +25,14 @@ class DummyCli(AzCli):
         from knack.completion import ARGCOMPLETE_ENV_NAME
         from knack.util import ensure_dir
 
+        env_patch = None
+
         if random_config_dir:
             config_dir = os.path.join(GLOBAL_CONFIG_DIR, 'dummy_cli_config_dir', random_string())
             # Knack prioritizes the AZURE_CONFIG_DIR env over the config_dir param, and other functions may call
             # get_config_dir directly. We need to set the env to make sure the config_dir is used.
-            self.env_patch = patch.dict(os.environ, {'AZURE_CONFIG_DIR': config_dir})
-            self.env_patch.start()
+            env_patch = patch.dict(os.environ, {'AZURE_CONFIG_DIR': config_dir})
+            env_patch.start()
 
             # Always copy command index to avoid initializing it again
             files_to_copy = ['commandIndex.json']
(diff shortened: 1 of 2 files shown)


jiasli:
A workaround I am currently using is to run these tests with different `pytest` instances/processes:

```sh
module="test_containerapp_manualjob_withsecret_crudoperations_e2e test_hdinsight_application"

for m in $module; do
    echo Running azdev for $m
    azdev test --series --no-exitfirst $m
    azdev test --series --no-exitfirst --live --lf $m
done
```

bebound:
Hmm, I believe the live mode works without env
 patch.

PR #28892, opened 2024-05-06
[App Service] `az functionapp scale config always-ready`: Set alwaysReady property to empty array if it is null

khkh-ms:
**Related command**
- az functionapp scale config always-ready

**Description**
For Flex Function App, it sets alwaysReady property to empty array when it is null. It is to avoid the exception which is thrown if the mentioned property is not iterable. 

**Testing Guide**

Create a function app
`az functionapp create -g <resource-group> -n <function-app-name> -s flexeastorage --functions-version 4 --debug --runtime python --runtime-version 3.11 --flexconsumption-location '"eastasia"' --debug --deployment-storage-auth-type storageAccountConnectionString`

Update the scale config of the function app
`az functionapp scale config always-ready set -g <resource-group> -n <function-app-name> --settings http=1 functionfunction1=1`


yonzhan:
App Service

amamounelsayed (approved):

zhoxing-ms (approved):

PR #28660, opened 2024-03-29
{ARM} Bump `MGMT_RESOURCE_SUBSCRIPTIONS` to 2022-12-01

jiasli:
**Related command**
`az login`

**Description**
Bump `MGMT_RESOURCE_SUBSCRIPTIONS` to 2022-12-01 in order to consume `defaultDomain` and `displayName` properties from [Tenants - List](https://learn.microsoft.com/en-us/rest/api/resources/tenants/list?view=rest-resources-2022-12-01&tabs=HTTP) API.

PR #28910, opened 2024-05-07
{Profile} Login experience v2

jiasli:
**Related command**
`az login`

**Description**
Interactive commands now prompt the user to interactively select the default subscription and tenant:
- `az login`: Auth code flow/WAM
- `az login --use-device-code`: Device code flow

Automation commands are not affected:
- `az login --username --password`: Username password flow
- `az login --service-principal`: Service principal authentication
- `az login --identity`: managed identity authentication

**Testing Guide**
`az login`
`az login --use-device-code`

Turn off:
```
az config set core.login_experience_v2=off
az login
```

**Additional Information**
Require #28660

**History Notes**

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/_profile.py
+++ b/src/azure-cli-core/azure/cli/core/_profile.py
@@ -24,10 +24,17 @@ logger = get_logger(__name__)
 _IS_DEFAULT_SUBSCRIPTION = 'isDefault'
 _SUBSCRIPTION_ID = 'id'
 _SUBSCRIPTION_NAME = 'name'
+
 # Tenant of the token which is used to list the subscription
 _TENANT_ID = 'tenantId'
-# Home tenant of the subscription, which maps to tenantId in 'Subscriptions - List REST API'
-# https://docs.microsoft.com/en-us/rest/api/resources/subscriptions/list
+
+# More information on the token tenant. Maps to properties in 'Tenants - List' REST API
+# https://learn.microsoft.com/en-us/rest/api/resources/tenants/list
+_TENANT_DEFAULT_DOMAIN = 'tenantDefaultDomain'  # defaultDomain
+_TENANT_DISPLAY_NAME = 'tenantDisplayName'  # displayName
+
+# Home tenant of the subscription. Maps to tenantId in 'Subscriptions - List' REST API
+# https://learn.microsoft.com/en-us/rest/api/resources/subscriptions/list
 _HOME_TENANT_ID = 'homeTenantId'
 _MANAGED_BY_TENANTS = 'managedByTenants'
 _USER_ENTITY = 'user'
@@ -91,17 +98,27 @@ def _get_cloud_console_token_endpoint():
     return os.environ.get('MSI_ENDPOINT')
 
 
-def _attach_token_tenant(subscription, tenant):
-    """Attach the token tenant ID to the subscription as tenant_id, so that CLI knows which token should be used
+def _attach_token_tenant(subscription, tenant, tenant_id_description=None):
+    """Attach the token tenant information to the subscription. CLI uses tenant_id to know which token should be used
     to access the subscription.
 
     This function supports multiple APIs:
-      - v2016_06_01's Subscription doesn't have tenant_id
-      - v2019_11_01's Subscription has tenant_id representing the home tenant ID. It will mapped to home_tenant_id
+      - v2016_06_01: Subscription doesn't have tenant_id
+      - v2022_12_01:
+        - Subscription has tenant_id representing the home tenant ID, mapped to home_tenant_id
+        - TenantIdDescription has default_domain, mapped to tenant_default_domain
+        - TenantIdDescription has display_name, mapped to tenant_display_name
     """
     if hasattr(subscription, "tenant_id"):
         setattr(subscription, 'home_tenant_id', subscription.tenant_id)
     setattr(subscription, 'tenant_id', tenant)
+
+    # Attach tenant_default_domain, if available
+    if tenant_id_description and hasattr(tenant_id_description, "default_domain"):
+        setattr(subscription, 'tenant_default_domain', tenant_id_description.default_domain)
+    # Attach display_name, if available
+    if tenant_id_description and hasattr(tenant_id_description, "display_name"):
+        setattr(subscription, 'tenant_display_name', tenant_id_description.display_name)
 
 
 # pylint: disable=too-many-lines,too-many-instance-attributes,unused-argument
@@ -133,6 +150,7 @@ class Profile:
               use_device_code=False,
               allow_no_subscriptions=False,
               use_cert_sn_issuer=None,
+              show_progress=False,
               **kwargs):
         """
         For service principal, `password` is a dict returned by ServicePrincipalAuth.build_credential
@@ -159,6 +177,11 @@ class Profile:
                 identity.login_with_service_principal(username, password, scopes=scopes)
 
         # We have finished login. Let's find all subscriptions.
+        if show_progress:
+            message = ('Retrieving subscriptions for the selection...' if tenant else
+                       'Retrieving tenants and subscriptions for the selection...')
+            print(f"\n{message}")
+
         if user_identity:
             username = user_identity['username']
 
@@ -735,6 +758,7 @@ class SubscriptionFinder:
         mfa_tenants = []
 
         client = self._create_subscription_client(credential)
+        # https://learn.microsoft.com/en-us/rest/api/resources/tenants/list
         tenants = client.tenants.list()
 
         for t in tenants:
@@ -756,7 +780,9 @@ class SubscriptionFinder:
             specific_tenant_credential = identity.get_user_credential(username)
 
             try:
-                subscriptions = self.find_using_specific_tenant(tenant_id, specific_tenant_credential)
+
+                subscriptions = self.find_using_specific_tenant(tenant_id, specific_tenant_credential,
+                                                                tenant_id_description=t)
             except AuthenticationError as ex:
                 # because user creds went through the 'common' tenant, the error here must be
                 # tenant specific, like the account was disabled. For such errors, we will continue
@@ -802,12 +828,17 @@ class SubscriptionFinder:
                 logger.warning("%s", t.tenant_id_name)
         return all_subscriptions
 
-    def find_using_specific_tenant(self, tenant, credential):
+    def find_using_specific_tenant(self, tenant, credential, tenant_id_description=None):
+        """List subscriptions that can be accessed from a specific tenant.
+        If called from find_using_common_tenant, tenant_id_description is TenantIdDescription retrieved from
+        'Tenants - List' REST API. If directly called, tenant_id_description is None.
+        """
         client = self._create_subscription_client(credential)
+        # https://learn.microsoft.com/en-us/rest/api/resources/subscriptions/list
         subscriptions = client.subscriptions.list()
         all_subscriptions = []
         for s in subscriptions:
-            _attach_token_tenant(s, tenant)
+            _attach_token_tenant(s, tenant, tenant_id_description=tenant_id_description)
             all_subscriptions.append(s)
         self.tenants.append(tenant)
         return all_subscriptions
@@ -840,6 +871,11 @@ def _transform_subscription_for_multiapi(s, s_dict):
     """
     if hasattr(s, 'home_tenant_id'):
         s_dict[_HOME_TENANT_ID] = s.home_tenant_id
+    if hasattr(s, 'tenant_default_domain'):
+        s_dict[_TENANT_DEFAULT_DOMAIN] = s.tenant_default_domain
+    if hasattr(s, 'tenant_display_name'):
+        s_dict[_TENANT_DISPLAY_NAME] = s.tenant_display_name
+
     if hasattr(s, 'managed_by_tenants'):
         if s.managed_by_tenants is None:
             s_dict[_MANAGED_BY_TENANTS] = None
--- a/src/azure-cli-core/azure/cli/core/telemetry.py
+++ b/src/azure-cli-core/azure/cli/core/telemetry.py
@@ -70,11 +70,13 @@ class TelemetrySession:  # pylint: disable=too-many-instance-attributes
         self.suppress_new_event = False
         self.poll_start_time = None
         self.poll_end_time = None
-        self.enable_broker_on_windows = None
-        self.msal_telemetry = None
         self.secrets_detected = None
         self.secret_keys = None
         self.user_agent = None
+        # authentication-related
+        self.enable_broker_on_windows = None
+        self.msal_telemetry = None
+        self.login_experience_v2 = None
 
     def add_event(self, name, properties):
         for key in self.instrumentation_key:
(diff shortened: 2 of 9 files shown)


yonzhan:
Profile

src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -498,9 +528,93 @@ def _match_account(account, subscription_id, secondary_key_name, secondary_key_v
...
+        from knack.prompting import prompt, NoTTYException
+
+        # Keep prompting until the user inputs a valid index
+        while True:
+            try:
+                select_index = prompt('Select a subscription and tenant (Type a number or Enter for no changes): ')
+            except NoTTYException:
+                # This is a good example showing interactive and non-TTY are not contradictory
+                logger.warning("No TTY to select the default subscription.")
+                return active_one
```
jiasli, line 597: If no TTY is available, should we still print the subscription table or the echo message?

src/azure-cli/azure/cli/command_modules/profile/_subscription_selector.py
```diff
@@ -0,0 +1,108 @@
...
+# --------------------------------------------------------------------------------------------
+
+from azure.cli.core._profile import (_SUBSCRIPTION_NAME, _SUBSCRIPTION_ID, _TENANT_DISPLAY_NAME, _TENANT_ID,
+                                     _IS_DEFAULT_SUBSCRIPTION)
+from knack.log import get_logger
+
+logger = get_logger(__name__)
+
+
+class SubscriptionSelector:
```
jiasli, line 13: Some refactors are made in commit https://github.com/Azure/azure-cli/pull/28910/commits/5884ecb533b877d9ca3c8ac58d72bd4b7bc5685f: 

1. Part of `Profile._interactively_select_subscription` is split into `_format_subscription_table` so that it can be unit-tested.
2. `Profile._interactively_select_subscription` and `_format_subscription_table` are extracted into a `SubscriptionSelector` class.
3. `SubscriptionSelector` is moved to `profile` module to reduce the change on core.

src/azure-cli/azure/cli/command_modules/profile/custom.py
```diff
@@ -145,7 +162,21 @@ def login(cmd, username=None, password=None, service_principal=None, tenant=None
...
-        use_cert_sn_issuer=use_cert_sn_issuer)
+        use_cert_sn_issuer=use_cert_sn_issuer,
+        show_progress=select_subscription)
+
+    # No JSON output if interactive account selection is used
+    if select_subscription:
+        from ._subscription_selector import SubscriptionSelector
+        from azure.cli.core._profile import _SUBSCRIPTION_ID
+
+        selected = SubscriptionSelector(subscriptions)()
```
jiasli, line 173: `subscriptions` here only contain subscriptions accessible by this user, provided to `az login`. Should we also show subscriptions from other users as candidates?

src/azure-cli/azure/cli/command_modules/profile/tests/latest/test_subscription_selector.py
```diff
@@ -0,0 +1,127 @@
...
+        from azure.cli.core.style import format_styled_text
+
+        # We have to force format_styled_text to use 'dark' theme, otherwise other tests in the same process
+        # may change format_styled_text.theme to 'none' and format_styled_text will not print color, leading to
+        # unmatched result. For example, this command fails:
+        #   azdev test test_account_show test_format_subscription_table
+        # create=True is required as running this test alone won't set format_styled_text.theme. patch.object
+        # will fail because of the missing theme attribute. For example, this command fails:
+        #   azdev test test_format_subscription_table
+        with mock.patch.object(format_styled_text, 'theme', 'dark', create=True):
```
jiasli, line 89: This mock can also be achieved by deleting the `theme` attribute, just like running `test_format_subscription_table` alone. Pytest's [`monkeypatch`](https://docs.pytest.org/en/latest/how-to/monkeypatch.html) fixture has a [`monkeypatch.delattr(obj, name, raising=True)`](https://docs.pytest.org/en/latest/reference/reference.html#pytest.MonkeyPatch.delattr) method for this purpose, but I can't find a way to do this with [`unittest.mock`](https://docs.python.org/3/library/unittest.mock.html).

jiasli: We can write some custom logic to achieve this:

```py
        is_attr_set = False
        original_attr_value = None
        if hasattr(format_styled_text, 'theme'):
            is_attr_set = True
            original_attr_value = format_styled_text.theme
            delattr(format_styled_text, 'theme')

        ...

        if is_attr_set:
            format_styled_text.theme = original_attr_value
```

or even wrap it as a context manager:

```py
class delattr_patch:
    def __init__(self, object, attr):
        self.object = object
        self.attr = attr
        self.is_attr_set = False
        self.original_attr_value = None

    def __enter__(self):
        if hasattr(self.object, self.attr):
            self.is_attr_set = True
            self.original_attr_value = getattr(self.object, self.attr)
            delattr(self.object, self.attr)

    def __exit__(self, exc_type, exc_value, traceback):
        if self.is_attr_set:
            setattr(self.object, self.attr, self.original_attr_value)
```

Then we can use it as 

```py
        with delattr_patch(format_styled_text, 'theme'):
            ...
```

src/azure-cli-core/azure/cli/core/_profile.py
dcaro: We probably want to skip the selection process in the case there is no TTY available.

src/azure-cli/azure/cli/command_modules/profile/custom.py
dcaro: If the user does not have access to the subscription, why should be show it?

src/azure-cli-core/azure/cli/core/_profile.py
jiasli: I agree, but should we still show the interactive interface, or switch to the previous behavior of returning a JSON?

jiasli: If `stdin` is not a TTY, the user may be running `echo 1 | az login`. This is unlikely as the user can't know beforehand which subscription will be the one they want to select, and the number can change.

If `stdout` is not a TTY, the user may be running `az login | grep`. Not showing the JSON output will be a breaking change in this case.

Therefore, we decide to show interactive subscription selection only when both `stdin` and `stdout` are TTY.

src/azure-cli/azure/cli/command_modules/profile/_subscription_selector.py
```diff
@@ -0,0 +1,99 @@
...
+            if is_default:
+                self._active_one = sub
+
+            # Trim subscription name if it is too long
+            subscription_name = sub[_SUBSCRIPTION_NAME]
+            if len(subscription_name) > subscription_name_length_limit:
+                subscription_name = subscription_name[:subscription_name_length_limit - 3] + '...'
+
+            row = {
+                'No': f'[{index_str}]' + (' ' + self.DEFAULT_ROW_MARKER if is_default else ''),
```
bebound, line 48: Why `No` is not highlighted?

src/azure-cli/azure/cli/command_modules/profile/custom.py
```diff
@@ -136,6 +151,9 @@ def login(cmd, username=None, password=None, service_principal=None, tenant=None
         from azure.cli.core.auth.identity import ServicePrincipalAuth
         password = ServicePrincipalAuth.build_credential(password, client_assertion, use_cert_sn_issuer)
 
+    select_subscription = interactive and sys.stdin.isatty() and sys.stdout.isatty() and \
+        cmd.cli_ctx.config.getboolean('core', 'login_experience_v2', fallback=True)
```
bebound, line 155: This is long logic, how about creating a function.

If `service_principal` is not None, the `select_subscription` is also enabled? This is conflict with PR description.

src/azure-cli/azure/cli/command_modules/profile/custom.py
```diff
@@ -136,6 +151,14 @@ def login(cmd, username=None, password=None, service_principal=None, tenant=None
         from azure.cli.core.auth.identity import ServicePrincipalAuth
         password = ServicePrincipalAuth.build_credential(password, client_assertion, use_cert_sn_issuer)
 
+
+    login_experience_v2 = cmd.cli_ctx.config.getboolean('core', 'login_experience_v2', fallback=True)
+    # Send login_experience_v2 config to telemetry
+    from azure.cli.core.telemetry import set_login_experience_v2
+    set_login_experience_v2(login_experience_v2)
+
+    select_subscription = interactive and sys.stdin.isatty() and sys.stdout.isatty() and login_experience_v2
```
jiasli, line 160: Should we also send `select_subscription` to telemetry? `login_experience_v2` stands for the config value, while 
`select_subscription` stands for whether interactive selection is actually invoked. @dcaro

jiasli, line 160: Should we also send `select_subscription` to telemetry? `login_experience_v2` stands for the config value, while `select_subscription` stands for whether interactive selection is actually invoked. @dcaro

src/azure-cli-core/azure/cli/core/telemetry.py
```diff
@@ -70,11 +70,13 @@ def __init__(self, correlation_id=None, application=None):
...
         self.poll_end_time = None
-        self.enable_broker_on_windows = None
-        self.msal_telemetry = None
         self.secrets_detected = None
         self.secret_keys = None
         self.user_agent = None
+        # authentication-related
+        self.enable_broker_on_windows = None
+        self.msal_telemetry = None
+        self.login_experience_v2 = None
```
jiasli, line 79: Group these properties together for better readability.

src/azure-cli-core/azure/cli/core/telemetry.py
```diff
@@ -222,11 +224,13 @@ def _get_azure_cli_properties(self):
...
         set_custom_properties(result, 'RegionIdentified', self.region_identified)
-        set_custom_properties(result, 'EnableBrokerOnWindows', str(self.enable_broker_on_windows))
-        set_custom_properties(result, 'MsalTelemetry', self.msal_telemetry)
         set_custom_properties(result, 'SecretsWarning', _get_secrets_warning_config())
         set_custom_properties(result, 'SecretsDetected', str(self.secrets_detected))
         set_custom_properties(result, 'SecretKeys', ','.join(self.secret_keys or []))
+        # authentication-related
+        set_custom_properties(result, 'EnableBrokerOnWindows', str(self.enable_broker_on_windows))
+        set_custom_properties(result, 'MsalTelemetry', self.msal_telemetry)
+        set_custom_properties(result, 'LoginExperienceV2', str(self.login_experience_v2))
```
jiasli, line 233: Should we force `str` values? `self.enable_broker_on_windows` is also a bool and we convert it to `str` (#24304). Can we use `bool` as it is? @evelyn-ys

jiasli:
After #26854, there is no easy way to verify the telemetry record sent. I have to add a breakpoint at

https://github.com/Azure/azure-cli/blob/07e1b17d4c0f9e4a4cd76fa9ff3838644be5e419/src/azure-cli-telemetry/azure/cli/telemetry/__init__.py#L78

and check `cli_payload` contains `"Context.Default.AzureCLI.LoginExperienceV2": "True"`. I think it'll be better to have some other mechanism to monitor the telemetry record sent, such as logging it in the debug log. @evelyn-ys

src/azure-cli/azure/cli/command_modules/profile/_subscription_selector.py
jiasli: UX's design.

src/azure-cli/azure/cli/command_modules/profile/custom.py
jiasli: Even if the current user doesn't have access, another user used in a previous `az login` has. Running `az account set` will automatically switch to the other user.

src/azure-cli/azure/cli/command_modules/profile/custom.py
jiasli: True, but not that complex as it is still a one-liner.

jiasli: > This is long logic, how about creating a function?

True, but not that complex as it is still only one line.

> If `service_principal` is not None, the `select_subscription` is also enabled? This conflicts with PR description.

`service_principal` must be used together with `username`:

https://github.com/Azure/azure-cli/blob/3b20c9ad3ac93da8a707198f530397c09412279a/src/azure-cli/azure/cli/command_modules/profile/custom.py#L127-L128

`interactive` is `True` only when `username` is not provided:

https://github.com/Azure/azure-cli/blob/3b20c9ad3ac93da8a707198f530397c09412279a/src/azure-cli/azure/cli/command_modules/profile/custom.py#L141-L148

In other words, when `username` is provided, `interactive` must be `False`.

bebound: How about `az login --identity`?

src/azure-cli/azure/cli/command_modules/profile/custom.py
```diff
@@ -30,6 +31,20 @@
...
 
+LOGIN_ANNOUNCEMENT = (
+    "[Announcements]\n"
+    "With the new Azure CLI login experience, you can select the subscription you want to use more easily. "
+    "Learn more about it and its configuration at https://go.microsoft.com/fwlink/?linkid=2271236\n\n"
+    "If you encounter any problem, please open an issue at https://aka.ms/azclibug\n"
+)
+
+LOGIN_OUTPUT_WARNING = (
+    "[WARNING] The login output has been updated. Please be aware that it no longer displays the full list of "
```
jiasli, line 42: To align with `[Announcements]`:

```suggestion
    "[Warning] The login output has been updated. Please be aware that it no longer displays the full list of "
```

src/azure-cli/azure/cli/command_modules/profile/_subscription_selector.py
jiasli: `No` is also highlighted in the latest code.

src/azure-cli/azure/cli/command_modules/profile/custom.py
jiasli: It is guarded by 

https://github.com/Azure/azure-cli/blob/4f151800433c8d2cdfc4dbcab4e99da7217a203e/src/azure-cli/azure/cli/command_modules/profile/custom.py#L131-L136

src/azure-cli/azure/cli/command_modules/profile/custom.py
jiasli: We'll consider adding telemetry for `select_subscription` in the future.

src/azure-cli-core/azure/cli/core/telemetry.py
jiasli: `bool` may not appear correctly in the final telemetry. Using `str` is more robust.

bebound (approved):

yonzhan (approved):

evelyn-ys (approved):

jiasli:
Equivalent Azure PowerShell PR: https://github.com/Azure/azure-powershell/pull/24755

PR #28913, opened 2024-05-08
[Core] BREAKING CHANGE: `az extension add/update` and `az upgrade`: Set default `--allow-preview` value from `True` to be `False`

AllyW:
**Related command**

**Description**

**Testing Guide**

**History Notes**

[Extension] BREAKING CHANGE: `az extension add/update`: Set default `--allow-preview` value from `True` to be `False`
[Upgrade] BREAKING CHANGE: `az upgrade`: Set default `--allow-preview` value from `True` to be `False`

---

### DIFF
--- a/scripts/ci/test_extensions.sh
+++ b/scripts/ci/test_extensions.sh
@@ -23,7 +23,8 @@ exit_code=0
 # Disable arcappliance arcdata connectedk8s: https://github.com/Azure/azure-cli/pull/20436
 # Disable k8s-extension temporarily: https://github.com/Azure/azure-cli-extensions/pull/6702
 # Disable alias temporarily: https://github.com/Azure/azure-cli/pull/27717
-ignore_list='azure-cli-ml fzf arcappliance arcdata connectedk8s k8s-extension alias'
+# hybridaks is going to be deprecated: https://github.com/Azure/azure-cli/pull/29838
+ignore_list='azure-cli-ml fzf arcappliance arcdata connectedk8s k8s-extension alias hybridaks'
 
 for ext in $output; do
     echo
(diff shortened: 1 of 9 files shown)


yonzhan:
Core

scripts/ci/test_extensions.sh
```diff
@@ -34,7 +34,7 @@ for ext in $output; do
     fi
 
     echo "Verifying extension:" $ext
-    az extension add -n $ext
+    az extension add -n $ext --allow-preview True
```
bebound, line 37: How many extensions do not follow the latest version number convention?

The test script should also use the default config. We can make an exception list for outdated extensions.

bebound:
The PR title does not follow the convention. We need to put command name at the beginning of the title.

src/azure-cli-core/azure/cli/core/extension/operations.py
```diff
@@ -306,12 +306,6 @@ def check_version_compatibility(azext_metadata):
 def add_extension(cmd=None, source=None, extension_name=None, index_url=None, yes=None,  # pylint: disable=unused-argument, too-many-statements
                   pip_extra_index_urls=None, pip_proxy=None, system=None,
                   version=None, cli_ctx=None, upgrade=None, allow_preview=None):
-    if allow_preview is None:
```
kairu-ms, line 309: Please add error log to instruct users add `--allow-preview` if an extension has preview version only but users forget adding `--allow-preview` argument.

kairu-ms (approved):

AllyW:
After cli released, customers will get error message as below for extension modules which do not have stable versions.
image

scripts/ci/test_extensions.sh
AllyW: The latest version schema and cli core code is compatible with the version number of all extension modules. 

There are 94 extension modules that only has preview releases (which would be failed when customers use new cli to install extension, and if it fails, the message like following will be prompted to users.) 

image

PR #28955, opened 2024-05-13
{IoT} Fix DPS policy key regeneration when setting empty strings as keys

c-ryan-k:
**Related command**

`az iot dps policy update`

**Description**

Currently the service rejects empty key values of `""` as invalid input, this sends the correct `None` / `null` values to the service for these values and will successfully regenerate keys for that policy.

**Testing Guide**

Force service to regenerate primary key:
`az iot dps policy update -g {group} --dps-name {dps} --pn {policy} --primary-key ""`

Force service to regenerate secondary key:
`az iot dps policy update -g {group} --dps-name {dps} --pn {policy} --secondary-key ""`

**History Notes**

---


yonzhan:
IoT

src/azure-cli/azure/cli/command_modules/iot/custom.py
```diff
@@ -198,8 +198,12 @@ def iot_dps_policy_update(
         if policy.key_name == access_policy_name:
             if primary_key is not None:
                 policy.primary_key = primary_key
```
digimaun, line 200: Is there a help example for this?

src/azure-cli/azure/cli/command_modules/iot/_params.py
```diff
@@ -96,8 +96,16 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
                        arg_type=get_enum_type(AccessRightsDescription),
                        help='Access rights for the IoT Hub Device Provisioning Service. '
                             'Use space-separated list for multiple rights.')
-            c.argument('primary_key', help='Primary SAS key value.')
-            c.argument('secondary_key', help='Secondary SAS key value.')
+
+    with self.argument_context('iot dps policy create') as c:
+        c.argument('primary_key', help='Primary SAS key value.')
```
digimaun, line 101: Should we append, if not provided will be generated by the service?

src/azure-cli/azure/cli/command_modules/iot/_help.py
```diff
@@ -65,6 +65,9 @@
   - name: Update shared access policy 'MyPolicy' in an Azure IoT Hub Device Provisioning Service instance with EnrollmentWrite right
     text: >
         az iot dps policy update --dps-name MyDps --resource-group MyResourceGroup --policy-name MyPolicy --rights EnrollmentWrite
+  - name: Regenerate keys for access policy 'MyPolicy' by updating keys to empty values
+    text: >
+        az iot dps policy update --dps-name MyDps --resource-group MyResourceGroup --policy-name MyPolicy --primary-key "" --secondary-key ""
```
vilit1, line 70: are we trying to remove or renew keys here? cause if renew then wouldn't naming like `--renew-primary` or whatever be better?

there should be some key renewal stuff in iot hub

digimaun: IoT Hub control plane has a dedicated `policy renew-key` command, but in this case it is around enabling the more generic `dps policy update` to allow service renewal of keys.

vilit1: we could also make the `--primary-key` value optional so if a person just specifies `--primary-key` then the service will regenerate it for them (rather than awkwardly adding in "")

src/azure-cli/azure/cli/command_modules/iot/_params.py
c-ryan-k: That's a good idea, this way we're consistent :)

src/azure-cli/azure/cli/command_modules/iot/_help.py
c-ryan-k: @vilit1 The way it's currently coded I don't see a good way to differentiate between an empty value: `--primary-key` and not passing the arg at all. If update chooses to accept `None` as a value - it will then regenerate the keys on every update if you don't provide them to the command. I think for now it's ok to use the empty string as a way to say "I am providing this value but I want it to be empty" as opposed to "I have not provided this value and don't want it to change".

digimaun (approved):

yanzhudd:
Hi @c-ryan-k 
since we are in a releasing process and the code is frozen during this process, this PR cannot be merged right now. We will merge this PR after the process is completed.

yanzhudd (approved):

PR #28963, opened 2024-05-14
{RDBMS} Update Create help text to show setting version for pg flex server creation to 16

nasc17:
**Related command**
`postgres flexible-server create`

**Description**
Default version is 13 which is very old. Want users to move to 16 when provisioning server.
Updated help to reflect this change of PG16 as default.
Update test to use PG16 during creation and perform MVU with PG16

**Testing Guide**
Manually tested
}

**History Notes**


yonzhan:
RDBMS

nasc17:
@evelyn-ys  @calvinhzy  All checks have passed please help review

evelyn-ys:
It's kind of breaking change to update parameter's default value. CLI only allows breaking changes twice a year, one for Build event and one for Ignite event. Do you mind waiting until Ignite?

nasc17:
> It's kind of breaking change to update parameter's default value. CLI only allows breaking changes twice a year, one for Build event and one for Ignite event. Do you mind waiting until Ignite?

Hi, is there a forum where I could try getting approval for allowing this breaking change. 
We have new version of Postgres each year - meaning we are already 3-4 years behind.

evelyn-ys (approved):

PR #28250, opened 2024-01-25
{Packaging} Use deb822 format in DEB installation script

konstruktoid:
**Description**

This PR:
- replaces the One-Line-Style format with the deb822 format, which will be the default in Ubuntu 24.04. See eg https://discourse.ubuntu.com/t/spec-apt-deb822-sources-by-default/29333 and https://repolib.readthedocs.io/en/latest/deb822-format.html
- performs basic shell linting
- is related to, and aligns with, https://github.com/MicrosoftDocs/azure-docs-cli/pull/4106

This checklist is used to make sure that common guidelines for a pull request are followed.

- [X] The PR title and description has followed the guideline in [Submitting Pull Requests](https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules#submitting-pull-requests).

- [X] I adhere to the [Command Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md).

- [X] I adhere to the [Error Handling Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/error_handling_guidelines.md).

PR #28990, opened 2024-05-20
{Packaging} Fix DEB install script

### LINKED ISSUE #28984: Install script breaks Github action in 2.61
### Describe the bug

We are running the install script at `https://aka.ms/InstallAzureCLIDeb` in a github action pipeline like this:
```
      - name: Install Azure CLI
        run: |
          sudo apt update && sudo apt-get install curl -y
          curl -sL https://aka.ms/InstallAzureCLIDeb | sudo bash
```

this morning this was broken, and it seems like the reason is because the gpg installation breaks the script with
```
gpg: cannot open '/dev/tty': No such device or address
curl: (23) Failed writing body
Error: Process completed with exit code 2.
```

I did some digging by enabling a tty and it seems the gpg installation wants to ask for an okay to overwrite the file:
`File '/etc/apt/keyrings/microsoft.gpg' exists. Overwrite? (y/N)`

which I assume is due to the path changes in:
`https://github.com/Azure/azure-cli/commit/655b05b15204ef8a5c76b7a5cd78cb489058530a`

### Related command

https://aka.ms/InstallAzureCLIDeb

### Errors

gpg: cannot open '/dev/tty': No such device or address
curl: (23) Failed writing body
Error: Process completed with exit code 2.

### Issue script & Debug output

not possible

### Expected behavior

the install script installs az cli without error

### Environment Summary

github action standard runner, before installation

### Additional context

_No response_

bebound:
Resolve #28984.

There is [no tty in GitHub runner](https://github.com/actions/runner/issues/241). This error is raised when run `gpg --dearmor -o /etc/apt/keyrings/microsoft.gpg`:
`gpg: cannot open '/dev/tty': No such device or address`

The reason is that this file already exists in the agent:
```
ls -la /etc/apt/keyrings/
total 20
drwxr-xr-x 2 root root 4096 May 16 11:42 .
drwxr-xr-x 8 root root 4096 May 16 11:10 ..
-rw-r--r-- 1 root root 2825 May 16 11:31 github_git-lfs-archive-keyring.gpg
-rw-r--r-- 1 root root 1200 May 16 11:42 kubernetes-apt-keyring.gpg
-rw-r--r-- 1 root root  641 May 16 11:20 microsoft.gpg
```

And `gpg` wants to ask: `File '/etc/apt/keyrings/microsoft.gpg' exists. Overwrite? (y/N)`

This PR reverts to the save method from the previous commit, which overwrites the file.

Related PR: #28250

Tested in this workflow: https://github.com/bebound/starter-workflows/actions/runs/9152933588

### DIFF
--- a/scripts/release/debian/deb_install.sh
+++ b/scripts/release/debian/deb_install.sh
@@ -49,7 +49,7 @@ setup() {
     set -v
     mkdir -p /etc/apt/keyrings
     curl -sLS https://packages.microsoft.com/keys/microsoft.asc |
-      gpg --dearmor -o /etc/apt/keyrings/microsoft.gpg
+      gpg --dearmor > /etc/apt/keyrings/microsoft.gpg
     chmod go+r /etc/apt/keyrings/microsoft.gpg
     set +v
 


yonzhan:
Packaging

scripts/release/debian/deb_install.sh
```diff
@@ -49,7 +49,7 @@ setup() {
     set -v
     mkdir -p /etc/apt/keyrings
     curl -sLS https://packages.microsoft.com/keys/microsoft.asc |
-      gpg --dearmor -o /etc/apt/keyrings/microsoft.gpg
+      gpg --dearmor > /etc/apt/keyrings/microsoft.gpg
```
jiasli, line 52: An alternative is to use `gpg --dearmor -o /etc/apt/keyrings/microsoft.gpg --yes`, but switching to the old redirection `>` is safer.

jiasli (approved):

jiasli:
The same change should be made on https://learn.microsoft.com/en-us/cli/azure/install-azure-cli-linux?pivots=apt.

bebound:
The document is used in an interactive environment, and the user can handle it when `File '/etc/apt/keyrings/microsoft.gpg' exists. Overwrite? (y/N)` appears.

PR #6716, opened 2018-07-03
Proposal: Add Dockerfile.spot?

derekbekoe:
Support creating 'Spots' directly from PRs in the repo.

PR #6070, opened 2018-04-10
added support for windows container via nanoserver

yorek:
added dockerfile to support windows containers using nanoserver image
Closes https://github.com/Azure/azure-cli/issues/6027

PR #29195, opened 2024-06-18
{Misc.} Remove unused dockerfile

bebound:
`Dockerfile.spot` was added in https://github.com/Azure/azure-cli/pull/6716.
It's used by [vscode-spot](https://github.com/derekbekoe/vscode-spot) extension, which was archived years ago.

`Dockerfile.nanoserver` was added in https://github.com/Azure/azure-cli/pull/6070 to support Windows container, but it has never been released. Now, Linux containers have become the mainstream.


yonzhan:
Misc.

Issue #13188, opened 2020-04-26
`az login` with managed identity indicating system assigned when the identity is user assigned


jiasli:
## Symptom

When using a single **user assigned identity** for `az login --identity`, it says the logged in `user` is a **system assigned identity**. 

The default identity is the identity returned when the `client_id`, `object_id` and `mi_res_id` parameters are not provided to IMDS. This will be the system assigned identity when present but can also be a user assigned identity if 

- there is no system assigned identity, and
- only one user assigned identity. 

`az login` is assuming the default identity used is a system assigned even if it is a user assigned.

The system identity is off since the creation of the VM, but each time I use the command I got the following:

```
C:\Users\xxx>az login --identity --allow-no-subscriptions
[
    {
        "environmentName": "AzureCloud",
        "id": "72f988bf-86f1-41af-91ab-2d7cd011db47",
        "isDefault": true,
        "name": "N/A(tenant level account)",
        "state": "Enabled",
        "tenantId": "72f988bf-86f1-41af-91ab-2d7cd011db47",
        "user": {
            "assignedIdentityInfo": "MSI",
            "name": "systemAssignedIdentity",
            "type": "servicePrincipal"
        }
    }
]
```

## Explanation

In the current implementation, identity type is deduced from whether `identity_id` is provided. When you run `az login --identity` without `--username`, CLI will assume it is a system assigned identity, instead of confirming from the response.

https://github.com/Azure/azure-cli/blob/764332b65bd8758696df36921084c033d82e5ff9/src/azure-cli-core/azure/cli/core/_profile.py#L357

## Solution

Extract the resource ID of managed identity from the access token's `xms_mirid` property to decide the correct type of managed identity.

Get the access token with 

```
curl 'http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&resource=https%3A%2F%2Fmanagement.azure.com%2F' -H Metadata:true -s
```

Then decode the access token with https://jwt.ms/.

System assigned identity:

```json
{
    "typ": "JWT",
    "alg": "RS256",
    "x5t": "CtTuhMJmD5M7DLdzD2v2x3QKSRY",
    "kid": "CtTuhMJmD5M7DLdzD2v2x3QKSRY"
}.{
    "aud": "https://management.azure.com/",
    "iss": "https://sts.windows.net/54826b22-38d6-4fb2-bad9-b7b93a3e9c5a/",
    "iat": 1587900154,
    "nbf": 1587900154,
[...]

PR #5140, opened 2017-12-20
Finalize Knack Branch. Re-enable all Tests.

tjprescott:
Closes #5127.  Incorporates all of the latest changes from dev.

PR #29198, opened 2024-06-18
{Profile} Clarify that an ID is required for user-assigned managed identity

jiasli:
**Related command**
`az login --identity`

**Description**
Close https://github.com/Azure/azure-cli/issues/13188 

If a VM is assigned with

- no system-assigned managed identity, and
- only one user-assigned managed identity

`az login --identity` will use the user-assigned managed identity for authentication and mark it as `systemAssignedIdentity`. 

This is considered a bad practice and should be avoided. As this behavior is baked into Azure VM's IMDS endpoint, Azure CLI as a client cannot change it, but we can clarify it in our help message.


yonzhan:
Profile

src/azure-cli/azure/cli/command_modules/profile/_help.py
```diff
@@ -24,10 +28,10 @@
       text: az login --service-principal -u http://azure-cli-2016-08-05-14-31-15 -p VerySecret --tenant contoso.onmicrosoft.com
     - name: Log in with a service principal using client certificate.
       text: az login --service-principal -u http://azure-cli-2016-08-05-14-31-15 -p ~/mycertfile.pem --tenant contoso.onmicrosoft.com
-    - name: Log in using a VM's system-assigned managed identity.
+    - name: Log in with a system-assigned managed identity.
       text: az login --identity
-    - name: Log in using a VM's user-assigned managed identity. Client or object ids of the service identity also work.
-      text: az login --identity -u /subscriptions/<subscriptionId>/resourcegroups/myRG/providers/Microsoft.ManagedIdentity/userAssignedIdentities/myID
+    - name: Log in with a user-assigned managed identity. You must specify the client ID, object ID or resource ID of the user-assigned managed identity with `--username`.
+      text: az login --identity --username /subscriptions/<subscriptionId>/resourcegroups/myRG/providers/Microsoft.ManagedIdentity/userAssignedIdentities/myID
```
bgavrilMS, line 34: According to Azure SDK, most app developers will prefer to specify a clientID here (a GUID). ResourceID is very exotic scenario, for example when you automate the provisioning of some hosts.

jiasli: Good catch! Changed as suggested.

rayluo: I also agree with Bogdan's suggestion on adding an example of specifying an ID, which you did.

I think it is a bit weird to provide a user-assigned managed identity (UAMI)'s GUID via a parameter named `username`. The word "username" would intuitively give me an impression that it would expect something like "JohnDoe" or "JohnDoe@contoso.com". That being said, if using `--username` to supply a UAMI's client_id is an established usage pattern, I would understand that it is too late to change it now.

Just a FYI. This PR looks good to me overall.

jiasli: > That being said, if using `--username` to supply a UAMI's client_id is an established usage pattern, I would understand that it is too late to change it now.

Exactly. It has been used since 2017 (https://github.com/Azure/azure-cli/pull/5140).

evelyn-ys (approved):

bebound (approved):

bebound (dismissed):

evelyn-ys (approved):

bebound (approved):

PR #29212, opened 2024-06-20
{Misc.} Add cgmanifest.json

bebound:
Component Detection can only detect dependencies in `setup.py`.
Add jp to mainifest so that we can generate NOTICE through the Component Governance UI

Ref:
https://docs.opensource.microsoft.com/tools/cg/component-detection/cgmanifest/
https://docs.opensource.microsoft.com/tools/cg/legal/notice/#generate-notice-through-the-component-governance-ui


yonzhan:
Misc.

jiasli (approved):

Issue #28972, opened 2024-05-15
Example missing an argument - az postgres flexible-server firewall-rule create


perrutbruno:
### Describe the bug

The examples for az postgres flexible-server firewall-rule create section, allowing public access for Azure services does not include the --rule-name argument which is required for this to work. I found [this Issue](https://github.com/Azure/azure-cli/issues/24757) from two years ago that pointed the same problem, and hence, I got how to make the command work.

Without the argument, the command fails with this error: _expected string or bytes-like object_

### Related command

az postgres flexible-server firewall-rule create

### Errors

expected string or bytes-like object

### Issue script & Debug output

cli.knack.cli: Command arguments: ['postgres', 'flexible-server', 'firewall-rule', 'create', '--resource-group', 'proreforma-prog-sg', '--name', 'proreforma-teste-db', '--start-ip-address', '217.123.78.38', '--debug']
cli.knack.cli: __init__ debug log:
Enable color in terminal.
cli.knack.cli: Event: Cli.PreExecute []
cli.knack.cli: Event: CommandParser.OnGlobalArgumentsCreate [<function CLILogging.on_global_arguments at 0x7fa852f7d360>, <function OutputProducer.on_global_arguments at 0x7fa852e73f40>, <function CLIQuery.on_global_arguments at 0x7fa852ec11b0>]
cli.knack.cli: Event: CommandInvoker.OnPreCommandTableCreate []
cli.azure.cli.core: Modules found from index for 'postgres': ['azure.cli.command_modules.rdbms']
cli.azure.cli.core: Loading command modules:
cli.azure.cli.core: Name                  Load Time    Groups  Commands
cli.azure.cli.core: rdbms                     0.200        55       230
cli.azure.cli.core: Total (1)                 0.200        55       230
cli.azure.cli.core: These extensions are not installed and will be skipped: ['azext_ai_examples', 'azext_next']
cli.azure.cli.core: Loading extensions:
cli.azure.cli.core: Name                  Load Time    Groups  Commands  Directory
cli.azure.cli.core: Total (0)                 0.000         0         0
cli.azure.cli.core: Loaded 55 groups, 230 commands.
cli.azure.cli.core: Found a match in the command table.
cli.azure.cli.core: Raw command  : postgres flexible-server firewall-rule create
cli.azure.cli.core: Command table: postgres flexible-server firewall-rule create
[...]

PR #29224, opened 2024-06-21
[RDBMS] `az postgres flexible-server firewall-rule create`: Correct firewall rule name and ip range validators

nasc17:
**Related command**
`az postgres flexible-server firewall-rule create`

**Description**
Help defines that --rule-name is not required argument. If name is omitted, default name will be chosen for firewall name. 
Validator method was return error when arg was not passed. Corrected to only do validation check if rule name arg has value

Addressing bug
#28972 

Bug fix in ip_address_validator to check that only digits are passed in IPV4 format

**Testing Guide**
Manually
Firewall name validator
PS C:\Users\nasc\azureCLI\azure-cli> az postgres flexible-server firewall-rule create --resource-group g --name n --start-ip-address 100.0.00.000
Configuring server firewall rule to accept connections from '100.0.00.000'...
{
  "endIpAddress": "100.0.00.000",
  "id": "/subscriptions/s/resourceGroups/g/providers/Microsoft.DBforPostgreSQL/flexibleServers/n/firewallRules/**FirewallIPAddress_2024-6-21_17-39-47**",
  "name": "**FirewallIPAddress_2024-6-21_17-39-47**",
  "resourceGroup": "g",
  "startIpAddress": "100.0.00.000",
  "systemData": null,
  "type": "Microsoft.DBforPostgreSQL/flexibleServers/firewallRules"
}

IP range validator
az postgres flexible-server firewall-rule create --resource-group g --name n --start-ip-address 10.20.11a1.22                                                       
Incorrect value for ip address. Ip address should be IPv4 format. Example: 12.12.12.12. 

**History Notes**
`az postgres flexible-server firewall-rule create`: Correct default name creation when --rule-name is not passed by user. Fix IP range validation.

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/rdbms/validators.py
+++ b/src/azure-cli/azure/cli/command_modules/rdbms/validators.py
@@ -597,6 +597,8 @@ def virtual_endpoint_name_validator(ns):
 
 
 def firewall_rule_name_validator(ns):
+    if not ns.firewall_rule_name:
+        return
     if not re.search(r'^[a-zA-Z0-9][-_a-zA-Z0-9]{1,126}[_a-zA-Z0-9]$', ns.firewall_rule_name):
         raise ValidationError("The firewall rule name can only contain 0-9, a-z, A-Z, \'-\' and \'_\'. "
                               "Additionally, the name of the firewall rule must be at least 3 characters "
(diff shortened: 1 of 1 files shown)


yonzhan:
RDBMS

evelyn-ys (approved):

Issue #20958, opened 2022-01-12
`az postgres flexible-server create` creates default flexibleserverdb database


marlowp:
## Describe the bug

**Command Name**
`az postgres flexible-server create`

**Errors:**
There doesn't appear to be a way to stop the default **flexibleserverdb** database from being created. The `--database-name` parameter allows you to specify the database name but doesn't allow you to omit the creation of it in the first instance.

## To Reproduce:
Steps to reproduce the behavior. Note that argument values have been redacted, as they may contain sensitive information.

- `az postgres flexible-server create --name {} --resource-group {} --location {} --admin-user {} --admin-password {} --tier {} --sku-name {} --public-access {} --version {} --tags {} --yes`

## Expected Behavior
If the `--database-name` parameter is not provided then I'd expect it to not create a database, this would match the behaviour found in `az postgres server create`. Or a separate parameter is available that allows me to explicitly decide not to create a new database on server creation.

## Environment Summary
```
Windows-10-10.0.18362-SP0
Python 3.8.9
Installer: MSI

azure-cli 2.32.0

Extensions:
aks-preview 0.5.51
application-insights 0.1.14
azure-devops 0.22.0

Dependencies:
msal 1.16.0
azure-mgmt-resource 20.0.0
```

yonzhan:
route to service team

nachoalonsoportillo:
@marlowp Thanks for reporting this issue. PR with the fix has been merged now. It will be surfaced in upcoming version of Azure CLI.

@yonzhan Please, close this one as resolved. Thanks!

PR #29227, opened 2024-06-22
[RDBMS] `az postgres flexible-server create`: Add argument `--create-default-database` to support disabling default database creation

nasc17:
**Related command**
postgres flexible-server create

**Description**
Add a new --create-default-database parameter which defaults to true, so that users who don't like this: behavior can change it, without breaking other existing scenarios.

Fix for #20958 

**Testing Guide**
Manually


yonzhan:
RDBMS

evelyn-ys (approved):

PR #29240, opened 2024-06-24
{RDBMS} Add upcoming breaking change warning for default version value being changed to 16

nasc17:
**Related command**
postgres flexible-server create

**Description**
Default version is 13 which is very old. Want users to move to 16 when provisioning server.

**Testing Guide**
Manually
az postgres flexible-server create -g nasc-runner -n nasc-delete-cli624 --admin-user arcadmin --admin-password Testing123
**In the next release, the default value for the PostgreSQL server major version will be updated to 16**
Checking the existence of the resource group 'nasc-runner'...
Resource group 'nasc-runner' exists ? : True
Detected current client IP : 108.6.37.172


yonzhan:
RDBMS

evelyn-ys (approved):

PR #29275, opened 2024-06-28
{Packaging} Vendor `azure-mgmt-datalake-store`

bebound:
Homebrew formula fails to build because of this package.
It raises a `error: metadata-generation-failed` error. See https://github.com/Azure/azure-cli/issues/27047#issuecomment-2191203916

Vendor it like https://github.com/Azure/azure-cli/pull/27163

PR #29292, opened 2024-07-02
{Misc.} Update CODEOWNERS

bebound:
Add more owner as yu is OOF, and we need to catch the release.
Related PR: https://github.com/Azure/azure-cli/pull/29275


yonzhan:
Misc.

yonzhan (approved):

PR #29388, opened 2024-07-15
{AKS} `az aks create`: do not set `network_plugin` based on the Python SDK

tyler-lloyd:
**Related command**

az aks create

**Description**

the CLI should not own defaulting the network
profile properties. Defaulting should be owned
completely by AKS RP, so if a customer does
not set the `network_plugin` field then the CLI
should treat it as `None`.

Today, the default provided from the SDK
is `kubenet` but in the future, the AKS RP
will change this default value when omitted.
This means that other clients (ARM, bicep,
etc.) that do not set `network-plugin` will
get a different behavior than the CLI because
those clients won't set anything but the CLI
will continue to include the default from the
SDK.

**Testing Guide**

az aks create - without providing a `--network-plugin` should succeed and AKS RP will set the default.

**History Notes**

[AKS] `az aks create`: Not set the `--network-plugin` based on the default from the Python SDK

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acs/managed_cluster_decorator.py
+++ b/src/azure-cli/azure/cli/command_modules/acs/managed_cluster_decorator.py
@@ -2316,10 +2316,10 @@ class AKSManagedClusterContext(BaseAKSContext):
         if enable_validation:
             (
                 pod_cidr,
-                service_cidr,
-                dns_service_ip,
-                docker_bridge_address,
-                network_policy,
+                _,
+                _,
+                _,
+                _,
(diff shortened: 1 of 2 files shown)


yonzhan:
AKS

tyler-lloyd:
related PR where we've already done this in the `aks-preview` extension: https://github.com/Azure/azure-cli-extensions/pull/7766

FumingZhang:
Requeued another [live test](https://dev.azure.com/msazure/CloudNativeCompute/_build/results?buildId=98236519&view=results)

FumingZhang (approved):
LGTM

yanzhudd (approved):

Issue #29421, opened 2024-07-19
CI fails because of setuptools change


bebound:
MarkupSafe 2.0.1 and pycparser 2.19 fails to build.

I seems that it tried to build source code with wheel 0.30.0, but the source code is not compatible with wheel 0.30.0.
```
2024-07-18T03:38:39.8924845Z Building wheels for collected packages: MarkupSafe, pycparser
2024-07-18T03:38:39.8925107Z   Building wheel for MarkupSafe (setup.py): started
2024-07-18T03:38:39.8925480Z   Building wheel for MarkupSafe (setup.py): finished with status 'error'
2024-07-18T03:38:39.8925810Z   error: subprocess-exited-with-error
2024-07-18T03:38:39.8925999Z   
2024-07-18T03:38:39.8926307Z   × python setup.py bdist_wheel did not run successfully.
2024-07-18T03:38:39.8926577Z   │ exit code: 1
2024-07-18T03:38:39.8926831Z   ╰─> [27 lines of output]
2024-07-18T03:38:39.8927061Z       Traceback (most recent call last):
2024-07-18T03:38:39.8927300Z         File "<string>", line 2, in <module>
2024-07-18T03:38:39.8927649Z         File "<pip-setuptools-caller>", line 34, in <module>
2024-07-18T03:38:39.8928138Z         File "/tmp/pip-install-dviqlpkb/markupsafe_202a0ce88dd04bccaa8b4ccc00947489/setup.py", line 61, in <module>
2024-07-18T03:38:39.8928412Z           run_setup(True)
2024-07-18T03:38:39.8928847Z         File "/tmp/pip-install-dviqlpkb/markupsafe_202a0ce88dd04bccaa8b4ccc00947489/setup.py", line 41, in run_setup
2024-07-18T03:38:39.8929122Z           setup(
2024-07-18T03:38:39.8929525Z         File "/mnt/vss/_work/1/s/env/lib/python3.11/site-packages/setuptools/__init__.py", line 106, in setup
2024-07-18T03:38:39.8929826Z           return distutils.core.setup(**attrs)
2024-07-18T03:38:39.8930040Z                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2024-07-18T03:38:39.8930463Z         File "/mnt/vss/_work/1/s/env/lib/python3.11/site-packages/setuptools/_distutils/core.py", line 171, in setup
2024-07-18T03:38:39.8930758Z           ok = dist.parse_command_line()
[...]

PR #27717, opened 2023-10-30
{CI} Disable `alias` extension test

bebound:
**Description**

This PR temporarily fixes the `Test Extensions Loading Python311` error, similar PR: https://github.com/Azure/azure-cli/pull/20436
```
2023-10-30T02:31:59.9288865Z ERROR: az_command_data_logger: cannot import name 'soft_unicode' from 'markupsafe' (/opt/az/azcliextensions/alias/markupsafe/__init__.py)
2023-10-30T02:31:59.9292703Z Traceback (most recent call last):
2023-10-30T02:31:59.9296297Z   File "/opt/hostedtoolcache/Python/3.11.6/x64/lib/python3.11/site-packages/knack/cli.py", line 233, in invoke
2023-10-30T02:31:59.9297243Z     cmd_result = self.invocation.execute(args)
2023-10-30T02:31:59.9299944Z                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2023-10-30T02:31:59.9301257Z   File "/opt/hostedtoolcache/Python/3.11.6/x64/lib/python3.11/site-packages/azure/cli/core/commands/__init__.py", line 663, in execute
2023-10-30T02:31:59.9307162Z     raise ex
2023-10-30T02:31:59.9324332Z   File "/opt/hostedtoolcache/Python/3.11.6/x64/lib/python3.11/site-packages/azure/cli/core/commands/__init__.py", line 726, in _run_jobs_serially
2023-10-30T02:31:59.9325597Z     results.append(self._run_job(expanded_arg, cmd_copy))
2023-10-30T02:31:59.9326044Z                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2023-10-30T02:31:59.9327178Z   File "/opt/hostedtoolcache/Python/3.11.6/x64/lib/python3.11/site-packages/azure/cli/core/commands/__init__.py", line 697, in _run_job
2023-10-30T02:31:59.9328177Z     result = cmd_copy(params)
2023-10-30T02:31:59.9328424Z              ^^^^^^^^^^^^^^^^
2023-10-30T02:31:59.9329467Z   File "/opt/hostedtoolcache/Python/3.11.6/x64/lib/python3.11/site-packages/azure/cli/core/commands/__init__.py", line 333, in __call__
2023-10-30T02:31:59.9330507Z     return self.handler(*args, **kwargs)
[...]

PR #29422, opened 2024-07-19
{Packaging} Remove Jinja2, MarkupSafe and pycparser from requirements.txt

bebound:
`Jinja2` and `MarkupSafe` are not used in CLI.
`pycparser` is a dependency of `cffi`.
We don't need to include them in requirements.txt

Resolve https://github.com/Azure/azure-cli/issues/29421

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/feedback/tests/latest/test_feedback.py
+++ b/src/azure-cli/azure/cli/command_modules/feedback/tests/latest/test_feedback.py
@@ -29,6 +29,7 @@ from azure.cli.testsdk.reverse_dependency import get_dummy_cli
 logger = logging.getLogger(__name__)
 
 
+@unittest.skip('alias extension is not working: https://github.com/Azure/azure-cli/issues/29422')
 class TestCommandLogFile(ScenarioTest):
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


yonzhan:
Packaging

jiasli:
I searched the whole repo and neither `MarkupSafe` nor `pycparser` is directly used. Maybe we can remove them?

bebound:
Yes. In this PR I wrote that jinja2 and markupsafe can be removed and I totally forget it.  https://github.com/Azure/azure-cli/pull/27717

src/azure-cli/azure/cli/command_modules/feedback/tests/latest/test_feedback.py
```diff
@@ -29,6 +29,7 @@
 logger = logging.getLogger(__name__)
 
 
+@unittest.skip('alias extension is not working: https://github.com/Azure/azure-cli/issues/29422')
```
bebound, line 32: After removing `markupsafe`, the `alias` extension does not work anymore. Previously, the CLI's `markupsafe` was loaded; now it loads the outdated alias's `markupsafe` package.
```
az alias create -n foo123 -c bar123
Unable to load extension 'alias: cannot import name 'soft_unicode' from 'markupsafe' (C:\Users\kk\.azure\cliextensions\alias\markupsafe\__init__.py)'. Use --debug for more information.
```

jiasli (approved):

PR #29433, opened 2024-07-22
{Packaging} Use latest setuptools in Alpine docker image

bebound:
We always use latest `setuptools` in packages, see https://github.com/Azure/azure-cli/pull/27196.
This PR updates the `setuptools` and `pip` in Alpine docker image.

For Python<3.12, the docker images uses pinned `setuptools`. In 3.12, it uses the latest version by default.
> ENV PYTHON_SETUPTOOLS_VERSION 65.5.1    --https://github.com/docker-library/python/blob/ec14acd6a0830012b4330429bc173247a375857e/3.11/alpine3.20/Dockerfile#L136

For now, the docker-library team does not want to bump the `setuptools` version because of the compatibility concern: https://github.com/docker-library/python/issues/942

**History Notes**

[Core] Resolve CVE-2024-6345

### DIFF
--- a/alpine.dockerfile
+++ b/alpine.dockerfile
@@ -36,7 +36,7 @@ ARG JP_VERSION="0.2.1"
 
 RUN --mount=type=bind,target=/azure-cli,source=./,rw apk add --no-cache ca-certificates bash bash-completion libintl icu-libs libc6-compat jq openssh-keygen \
     && apk add --no-cache --virtual .build-deps gcc musl-dev linux-headers libffi-dev curl \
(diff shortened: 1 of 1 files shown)


yonzhan:
Packaging

yonzhan (approved):

jiasli:
Isn't #27196 only affect Homebrew? Why Alpine also requires the latest `setuptools`?

bebound:
> Isn't #27196 only affecting Homebrew? Why does Alpine also require the latest `setuptools`?

Actually, all packages contain `setuptools` package, but they always install the latest version (except macOS?). Alpine's `setuptools` is pinned to 65.5.1

PR #28556, opened 2024-03-13
{Core} Bump MSAL to 1.27.0 and pin `pymsalruntime` to 0.14.0

jiasli:
**Related command**
`az login`

**Description**
MSAL 1.27.0 (https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/669) contains several improvements.

The most significant one is to tolerate ID token time errors (https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/657) which will solve 2 very common issues:

- RuntimeError: 0. The ID token is not yet valid. (https://github.com/Azure/azure-cli/issues/20388)
- RuntimeError: 9. The ID token already expires. (https://github.com/Azure/azure-cli/issues/23223)

Issue #15005, opened 2020-09-01
[Enhancement Proposal] User and tenant isolation


jiasli:
## Issue

Currently, Azure CLI uses `subscriptionId` as the primary key to distinguish between accounts (stored in `~/.azure/azureProfile.json`). 

https://github.com/Azure/azure-cli/blob/58c9f78be2db3b7097558cbe76bc8d1ca1d304d6/src/azure-cli-core/azure/cli/core/_profile.py#L399-L405

This causes problem when

1. `az login` is run twice with different accounts that have access to the same subscriptions. The subscriptions listed during the second `az login` will _overwrite_ subscriptions from the first `az login`.
2. A subscription can be accessed in multiple tenants. Only the first occurrence is preserved and all subsequent occurrences are discarded. (#11886)
    https://github.com/Azure/azure-cli/blob/58c9f78be2db3b7097558cbe76bc8d1ca1d304d6/src/azure-cli-core/azure/cli/core/_profile.py#L924-L935

## Proposal

1. CLI should use a **combined primary key** or **3-layer structure** which consists of `username + tenant_id + subscription_id`.
2. CLI should support switching users with `az account set --username`.
2. CLI should support switching tenants with `az account set --tenant`.
3. The selected subscription should be saved for each user and each tenant, so that after switching users/tenants, the selected subscription is restored.

yonzhan:
login Enhancement

miqm:
@yonzhan - any ETA when this will be implemented?

jiasli:
Thank you @miqm for your interest in feature. 😊 We will consider this feature in future planning with our PM @chasewilson.

jiasli:
Now I more and more prefer using a combined primary key or 3-layer structure which consists of `username` + `tenant_id` + `subscription_id`.

Thus, we can keep the current interface of `az account set --subscription`. If the subscription can't be uniquely identified, `--tenant` or `username` can be introduced and used.

cveld:
@jiasli any progress?

PR #29441, opened 2024-07-22
[Core] `az logout`: Remove service principal access tokens from token cache

jiasli:
**Related command**
`az logout`

**Description**
Based on https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/666 from MSAL 1.27.0 (adopted by https://github.com/Azure/azure-cli/pull/28556).

**Testing Guide**
```sh
# Disable token cache encryption so that we can directly check the content of ~/.azure/msal_token_cache.json
az config set core.encrypt_token_cache=false

az login --service-principal --username xxx --password xxx --tenant xxx
# Log out current account
az logout

az login --service-principal --username xxx --password xxx --tenant xxx
# Log out by SP client ID
az logout --username xxx
```

### DIFF
--- a/src/azure-cli-core/azure/cli/core/auth/identity.py
+++ b/src/azure-cli-core/azure/cli/core/auth/identity.py
@@ -11,7 +11,7 @@ import sys
 from azure.cli.core._environment import get_config_dir
 from knack.log import get_logger
 from knack.util import CLIError
-from msal import PublicClientApplication
+from msal import PublicClientApplication, ConfidentialClientApplication
 
 # Service principal entry properties
 from .msal_authentication import _CLIENT_ID, _TENANT, _CLIENT_SECRET, _CERTIFICATE, _CLIENT_ASSERTION, \
@@ -203,8 +203,9 @@ class Identity:  # pylint: disable=too-many-instance-attributes
     def login_in_cloud_shell(self, scopes):
         raise NotImplementedError
 
-    def logout_user(self, user):
-        accounts = self._msal_app.get_accounts(user)
+    def logout_user(self, username):
+        # If username is an SP client ID, it is ignored
+        accounts = self._msal_app.get_accounts(username)
         for account in accounts:
             self._msal_app.remove_account(account)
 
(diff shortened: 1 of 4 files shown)


yonzhan:
Core

src/azure-cli-core/azure/cli/core/auth/identity.py
```diff
@@ -218,12 +219,21 @@ def logout_all_users(self):
...
+        cca = ConfidentialClientApplication(client_id, **self._msal_app_kwargs)
+        cca.remove_tokens_for_client()
+
+        # Step 2: Remove SP from SP store
+        self._service_principal_store.remove_entry(client_id)
 
     def logout_all_service_principal(self):
         # remove service principal secrets
+        # TODO: As MSAL provides no interface to get all service principals in its token cache, this method can't
+        #   clear all service principals' access tokens from MSAL token cache.
```
jiasli, line 236: I left a comment at https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/666#issuecomment-2242548486

src/azure-cli-core/azure/cli/core/auth/identity.py
```diff
@@ -218,12 +219,21 @@ def logout_all_users(self):
...
+    def logout_service_principal(self, client_id):
+        # If client_id is a username, it is ignored
+
+        # Step 1: Remove SP from MSAL token cache
+        # Note that removing SP access tokens shouldn't rely on SP store
+        cca = ConfidentialClientApplication(client_id, **self._msal_app_kwargs)
+        cca.remove_tokens_for_client()
+
+        # Step 2: Remove SP from SP store
+        self._service_principal_store.remove_entry(client_id)
```
jiasli, line 231: I originally wrote the code as 

```py
    def logout_service_principal(self, client_id):
        try:
            entry = self._service_principal_store.load_entry(client_id, self.tenant_id)
        except CLIError as ex:
            logger.info('%s %s is possibly a user account.', ex, client_id)
            return

        # Step 1: Remove service principal access tokens from MSAL token cache.
        sp_auth = ServicePrincipalAuth(entry)
        cred = ServicePrincipalCredential(sp_auth, **self._msal_app_kwargs)
        cred.remove_tokens_for_client()

        # Step 2: Remove service principal secrets
        self._service_principal_store.remove_entry(client_id)
```

Then I realized removing an SP from token cache doesn't require `client_credential`, so there is no need to read from `self._service_principal_store` and create a fully-functioning `ServicePrincipalCredential`.

jiasli:
There is one edge case caused by the "subscription ID as primary key" issue (https://github.com/Azure/azure-cli/issues/15005). If sp1 and sp2 have access to the same subscription, after running `az login` for sp1, running `az login` for sp2 will overwrite the `subscription-tenant-user` login context of sp1, making CLI lose track of sp1 in its profile.

To solve that, the user must either

- Remember sp1's client ID and use it in `az login --username {sp1_client_id}`
- Run `az account clear` which purges the whole token cache

The same happens to user accounts too.

bebound (dismissed):

evelyn-ys (approved):

bebound (approved):

Issue #25860, opened 2023-03-17
Migrate Managed Identity's implementation from `msrestazure` to MSAL


jiasli:
**Related command**
`az login --identity`

**Is your feature request related to a problem? Please describe.**
In order to support Hybrid Runbook Worker VM's Managed Identity (https://github.com/Azure/azure-cli/issues/22243), Azure CLI has to use the new Managed Identity endpoint API version`2019-08-01`. Since `msrestazure` has been deprecated, we have concluded that MSAL will support Managed Identity authentication.

**Describe the solution you'd like**
Once MSAL finishes its implementation for Managed Identity (https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/480), Azure CLI needs to migrates its Managed Identity's implementation from `msrestazure` to MSAL.

**Describe alternatives you've considered**
Use Azure Identity's Managed Identity's implementation, but the dependency on Azure Identity has been removed as explained in https://github.com/Azure/azure-cli/pull/19853#issuecomment-1011763038.

**Additional context**
Some old issues should be discussed or fixed during the migration:

- https://github.com/Azure/azure-cli/issues/20460: Supporting Track 1 SDK adds lots of development effort. However, Storage and KeyVault command module still depends on Track 1 data-plane SDKs.
- https://github.com/Azure/azure-cli/issues/13188

yonzhan:
Managed Identity's implementation migration

PR #29452, opened 2024-07-24
[Core] Allow opt-in retry of MSI auth on connection or timeout error

doubleyewdee:
**Related command**
Many commands utilize the MSI wrapper.

**Description**
This PR provides environment variables to enable retrying retryable errors when performing MSI authentication. Three variables are provided:
- $AZURE_CLI_MSI_RETRY_ENABLED (default off)
- $AZURE_CLI_MSI_RETRY_COUNT (default 3)
- $AZURE_CLI_MSI_RETRY_DELAY (default 15)

This is desirable because we have observed many issues where we use the az CLI within ACI instances to perform operations which fail due to seemingly transient network issues. We believe retrying on (e.g.) 'broken pipe' errors or timeouts is both safe and will ensure fewer deployment fails. We have observed this dozens of times in the last two to three weeks and believe many other teams are negatively impacted.

**Testing Guide**
Unclear -- this code's unit tests all override the class being modified. However, this code should be easily testable by simply setting $AZURE_CLI_MSI_RETRY_ENABLED to `true` and utilizing any commands which require MSI auth.

**History Notes**
[Core] Add environment variables to control attempts to retry MSI authentication. $AZURE_CLI_MSI_RETRY_ENABLED = 'true' enables retries, and $AZURE_CLI_MSI_RETRY_COUNT and $AZURE_CLI_MSI_RETRY_DELAY control the number of retries, and seconds between retries respectively.

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/auth/adal_authentication.py
+++ b/src/azure-cli-core/azure/cli/core/auth/adal_authentication.py
@@ -10,7 +10,7 @@ import requests
 from knack.log import get_logger
 from msrestazure.azure_active_directory import MSIAuthentication
 
-from .util import AccessToken, _normalize_scopes, scopes_to_resource
+from .util import _normalize_scopes, scopes_to_resource, AccessToken
 
 logger = get_logger(__name__)
 
@@ -32,9 +32,9 @@ class MSIAuthenticationWrapper(MSIAuthentication):
             if in_cloud_console():
                 # Use MSAL to get VM SSH certificate
                 import msal
+                from .util import check_result, build_sdk_access_token
 
                 from .identity import AZURE_CLI_CLIENT_ID
-                from .util import build_sdk_access_token, check_result
                 app = msal.PublicClientApplication(
                     AZURE_CLI_CLIENT_ID,  # Use a real client_id, so that cache would work
                     # TODO: This PoC does not currently maintain a token cache;
@@ -86,15 +86,14 @@ class MSIAuthenticationWrapper(MSIAuthentication):
             retry_delay = int(os.getenv(MSI_RETRY_DELAY_ENVIRONMENT_VARIABLE, MSI_RETRY_DELAY_DEFAULT))
 
         import traceback
-
-        from azure.cli.core.azclierror import (AzureConnectionError,
-                                               AzureResponseError)
+        from azure.cli.core.azclierror import AzureConnectionError, AzureResponseError
 
         attempts = 0
         error_msg = None
         while attempts < max_attempts:
             try:
                 super().set_token()
+                error_msg = None
                 break
             except requests.exceptions.HTTPError as err:
                 logger.debug('throw requests.exceptions.HTTPError when doing MSIAuthentication: \n%s',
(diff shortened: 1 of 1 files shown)


yonzhan:
Core

jiasli:
We plan to migrate Managed Identity's implementation from `msrestazure` to MSAL (https://github.com/Azure/azure-cli/issues/25860). You may submit a feature request at https://github.com/AzureAD/microsoft-authentication-library-for-python.

doubleyewdee:
@jiasli what's the timeframe for the ADAL migration? We're having pretty widespread issues today with (e.g.) broken pipe errors across many pipelines that force manual retries/user intervention, and I believe that this is widespread. The item you linked is open and from several months ago. I am concerned that a migration timeline (along with adding resilience in the ADAL Python library) will lead to significant delays in addressing this behavior.

jiasli:
@doubleyewdee, we no longer rely on ADAL and will drop it soon.

Even if we migrate to MSAL for managed identity authentication, if MSAL doesn't add the retry mechanism, Azure CLI won't have it either.

If this issue is urgent, I would recommend contacting ACI service team and let them investigate the network issue. Managed Identity should be reliable. It is not really the client's responsibility to retry. cc @rayluo

jiasli:
> I would push back a little that clients shouldn't also participate in retries. It's not actually possible to have a 100% reliable network (either client or server side), and retries are routinely valuable for tail reliability

Sorry I didn't make it clear. I don't mean a client shouldn't retry, but the client shouldn't take the main responsibility.

> Back to my question, what's the timeline for the MSAL library migration?

We are working on https://github.com/Azure/azure-cli/issues/25860, but there is currently so specific ETA yet. I will bring this ask to our Product Managers.

> When should we be on the lookout for behavior here and whether we might want to ask for retry options in the MSAL lib?

My personal opinion is that MSAL lib is a better place to add the retry logic, as other tools using MSAL can benefit from it.

doubleyewdee:
@jisali thanks for the updates and linked item! we'll wait for the MSAL migration and follow the items you've linked.

rayluo:
> My personal opinion is that MSAL lib is a better place to add the retry logic, as other tools using MSAL can benefit from it.

For the record, the default retry logic does already happen in MSAL lib. It is just that "how (many times) to retry" configuration is owned by calling app, because there is unlikely to be a one-size-fits-all configuration here. For that reason, MSAL also exposes an API to customize that retry behavior, it is just that the API is not in the form of `retry_count` (or `retry_delay`, for that matter).

doubleyewdee:
@rayluo I'd like to get this exposed in a manner roughly similar to what this PR suggested (not clear if delay/etc need to be supported vs. simply supplying a count) but will need to wait for the actual az MSAL migration it seems.

My team is going to attempt to wrap all our `az` calls with retries in the interim, but we're being bitten very badly here by some combination of Azure-owned things (e.g. in certain regions we have 10%+ failure rates here at times) so I think we need a holistic solution for both our 1P and 3P customers that is superior to what's on offer today.

PR #29503, opened 2024-07-29
{Core} Add cli subprocess

AllyW:
**Related command**

**Description**

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/cli_subprocess.py
+++ b/src/azure-cli-core/azure/cli/core/cli_subprocess.py
@@ -7,7 +7,6 @@
 
 import subprocess
 import re
-from functools import wraps
 from knack.log import get_logger
 
 
@@ -53,59 +52,34 @@ def subprocess_arg_mask(args, kwargs):
 
 def subprocess_kwarg_mask(kwargs):
     if kwargs.get("shell", False):
-        logger.warning("Removed shell=True for cli processor")
+        logger.warning("Removed shell=True for cli subprocess")
         kwargs["shell"] = False
 
 
-def cli_subprocess_func_mask(func):
-    @wraps(func)
-    def wrapper(*args, **kwargs):
-        subprocess_arg_type_check(*args, kwargs)
-        subprocess_arg_mask(*args, kwargs)
-        subprocess_kwarg_mask(kwargs)
-        return func(*args, **kwargs)
-    return wrapper
+def cli_subprocess_pre_parser(args, kwargs):
+    subprocess_arg_type_check(args, kwargs)
+    subprocess_arg_mask(args, kwargs)
+    subprocess_kwarg_mask(kwargs)
 
 
-def cli_subprocess_popen_init_mask(init_func):
-    @wraps(init_func)
-    def wrapper(self, *args, **kwargs):
-        subprocess_arg_type_check(*args, kwargs)
-        subprocess_arg_mask(*args, kwargs)
-        subprocess_kwarg_mask(kwargs)
-        return init_func(self, *args, **kwargs)
-
-    return wrapper
-
-
-def cli_subprocess_decorator(target):
-    if isinstance(target, type):
-        if '__init__' in target.__dict__:
-            target.__init__ = cli_subprocess_popen_init_mask(target.__init__)
-        return target
-    else:
-        return cli_subprocess_func_mask(target)
-
-
-@cli_subprocess_decorator
 def run(*popenargs, **kwargs):
     """Run command with arguments and remove shell=True
 
     The other arguments are the same as for the subprocess.run.
     """
+    cli_subprocess_pre_parser(*popenargs, kwargs)
     return subprocess.run(*popenargs, **kwargs)
 
 
-@cli_subprocess_decorator
 def call(*popenargs, **kwargs):
     """Run command with arguments and remove shell=True
 
     The other arguments are the same as for the subprocess.call.
     """
+    cli_subprocess_pre_parser(*popenargs, kwargs)
     return subprocess.call(*popenargs, **kwargs)
 
 
-@cli_subprocess_decorator
 def check_call(*popenargs, **kwargs):
     """Run command with arguments and remove shell=True
 
@@ -113,21 +87,22 @@ def check_call(*popenargs, **kwargs):
 
     check_call(["ls", "-l"])
     """
+    cli_subprocess_pre_parser(*popenargs, kwargs)
     return subprocess.check_call(*popenargs, **kwargs)
 
 
-@cli_subprocess_decorator
 def check_output(*popenargs, **kwargs):
     """Run command with arguments, remove shell=True, and return its output, as subprocess.check_output.
     """
+    cli_subprocess_pre_parser(*popenargs, kwargs)
     return subprocess.check_output(*popenargs, **kwargs)
 
 
-@cli_subprocess_decorator
 class CliPopen(subprocess.Popen):
     """
     Construct subprocess.Popen with masked shell kwargs to avoid security vulnerability
     """
 
     def __init__(self, *args, **kwargs):
+        cli_subprocess_pre_parser(*args, kwargs)
         super().__init__(*args, **kwargs)
--- a/src/azure-cli-core/azure/cli/core/cli_subprocess.py
+++ b/src/azure-cli-core/azure/cli/core/cli_subprocess.py
@@ -16,7 +16,7 @@ __all__ = [
     "call",
     "check_call",
     "check_output",
-    ]
+]
 
 logger = get_logger(__name__)
 
--- a/src/azure-cli-core/azure/cli/core/cli_subprocess.py
+++ b/src/azure-cli-core/azure/cli/core/cli_subprocess.py
@@ -3,6 +3,8 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


yonzhan:
Core

src/azure-cli-core/azure/cli/core/decorators.py
```diff
@@ -111,3 +114,33 @@ def _wrapped_func(*args, **kwargs):
...
+    return wrapper
+
+
+def cli_subprocess_decorator(target):
+    if isinstance(target, type):
+        if '__init__' in target.__dict__:
+            target.__init__ = cli_subprocess_popen_init_mask(target.__init__)
+        return target
+    else:
+        return cli_subprocess_func_mask(target)
```
kairu-ms, line 146: Why need decorators? I think its clear directly call the check functions  before call `subprocess.*`

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -1356,3 +1358,33 @@ def should_encrypt_token_cache(cli_ctx):
...
+    if not enable_arg_mask:
+        return
+    if not isinstance(args, list):
+        return
+    for i, val in enumerate(args):
+        args[i] = re.sub(ARG_MASK_CHAR, "", val)
+
+
+def subprocess_kwarg_mask(kwargs):
+    if kwargs.get("shell", False):
```
jsntcy, line 1388: Typo? should be "True"?

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -1356,3 +1358,33 @@ def should_encrypt_token_cache(cli_ctx):
...
+        logger.warning("Removed shell=True for cli processor")
+        kwargs["shell"] = False
```
jiasli, line 1390: Please don't expose these internal implementations. `src/azure-cli-core/azure/cli/core/util.py` should expose as few methods as possible.

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -1356,3 +1358,33 @@ def should_encrypt_token_cache(cli_ctx):
...
+def subprocess_arg_mask(args, kwargs):
+    enable_arg_mask = kwargs.get("enable_arg_mask", None)
+    if enable_arg_mask is not None:
+        del kwargs["enable_arg_mask"]
+    if not enable_arg_mask:
+        return
+    if not isinstance(args, list):
+        return
+    for i, val in enumerate(args):
+        args[i] = re.sub(ARG_MASK_CHAR, "", val)
```
jiasli, line 1384: Replacing `ARG_MASK_CHAR` is fragile. Consider using `shlex.quote`: https://docs.python.org/3/library/shlex.html#shlex.quote

src/azure-cli-core/azure/cli/core/cli_subprocess.py
```diff
@@ -0,0 +1,67 @@
...
+logger = get_logger(__name__)
+
+
+@cli_subprocess_decorator
+def run(*popenargs, **kwargs):
+    """Run command with arguments and remove shell=True
+
+    The other arguments are the same as for the subprocess.run.
+    """
+    return subprocess.run(*popenargs, **kwargs)
```
jiasli, line 30: When subprocessing `azcopy` and `bicep`, high-level APIs are used.

- `azcopy` (though it is using an [older high-level API](https://docs.python.org/3/library/subprocess.html#older-high-level-api)): https://github.com/Azure/azure-cli/blob/f1d068ece6789d8e77f3a1ee7c140b12982298a8/src/azure-cli/azure/cli/command_modules/storage/azcopy/util.py#L105

- `bicep`: https://github.com/Azure/azure-cli/blob/cff8d9bad3e39aa69ab9ca108693c66088d615bc/src/azure-cli/azure/cli/command_modules/resource/_bicep.py#L327-L331

`azdev` only uses `subprocess.call`: https://github.com/Azure/azure-cli-dev-tools/blob/db1212ac069ce2f4149315b5860c76fc1499b736/azdev/utilities/command.py#L34

Exposing only one function for `subprocess.call` may be sufficient.

src/azure-cli-core/azure/cli/core/cli_subprocess.py
```diff
@@ -0,0 +1,67 @@
...
+    check_call(["ls", "-l"])
+    """
+    return subprocess.check_call(*popenargs, **kwargs)
+
+
+@cli_subprocess_decorator
+def check_output(*popenargs, **kwargs):
+    """Run command with arguments, remove shell=True, and return its output, as subprocess.check_output.
+    """
+    return subprocess.check_call(*popenargs, **kwargs)
```
jiasli, line 57:
```suggestion
    return subprocess.check_output(*popenargs, **kwargs)
```

src/azure-cli-core/azure/cli/core/tests/test_cli_subprocess.py
```diff
@@ -0,0 +1,70 @@
...
+            cmd = ["cmd.exe", "/c"] + cmd
+        output = cli_subprocess.call(cmd)
+        self.assertEqual(int(output), 0, "unexpected output when run cmd in call")
+
+    def test_cli_subprocess_shell(self):
+        cmd = ["echo", "ab;echo 123"]
+        if platform.system().lower() == "windows":
+            cmd = ["cmd.exe", "/c"] + cmd
+        process = cli_subprocess.CliPopen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True,
+                                          enable_arg_mask=True)
```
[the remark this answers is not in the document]
AllyW: Cause arg mask itself is optional. By default it's false and in most scenarios it's necessary.

src/azure-cli-core/azure/cli/core/util.py
AllyW: As the shlex doc said,  "The shlex module is only designed for Unix shells. ", so I planned to put it in cli subprocess doc later for guiding users to split and quote their cmd and args themselves according to the platform.

shell=False and arg list is safe enough for avoid cmd injection.

src/azure-cli-core/azure/cli/core/cli_subprocess.py
AllyW: I think it's ok to expose these commonly used three older high-level APIs cause subprocess itself does this. And it's easier for developers to adjust their current code snippet.

src/azure-cli-core/azure/cli/core/util.py
bebound, line 1390: If we believe `shell=True` is unnecessary, why not raise an error to fail the tests and let the code owner to fix it?

AllyW: We can but that raises a developers' second-check issue, which actually can be easily eliminated. I prefer this to be a by-default behaviour for clisubprocess so that it is a one-time effort for developers to adopt it.

kairu-ms (approved):

src/azure-cli-core/azure/cli/core/util.py
jiasli: I agree with @bebound that developers should be guided to follow the best practice, instead of silently changing the parameters.

src/azure-cli-core/azure/cli/core/cli_subprocess.py
```diff
@@ -0,0 +1,108 @@
...
+    if skip_arg_type_check:
+        return
+    if isinstance(args, list):
+        return
+    from azure.cli.core.azclierror import ArgumentUsageError
+    raise ArgumentUsageError("cli_subprocess args should be a list of sequence")
+
+
+def subprocess_arg_mask(args, kwargs):
+    enable_arg_mask = kwargs.get("enable_arg_mask", None)
```
jiasli, line 42: `enable_arg_mask` and `skip_arg_type_check` are hidden in `kwargs`. In that case, it should be included in the doc string. We can't expect developers to look into the source code to know them.

src/azure-cli-core/azure/cli/core/cli_subprocess.py
```diff
@@ -0,0 +1,108 @@
...
+        kwargs["shell"] = False
+
+
+def cli_subprocess_pre_parser(args, kwargs):
+    subprocess_arg_type_check(args, kwargs)
+    subprocess_arg_mask(args, kwargs)
+    subprocess_kwarg_mask(kwargs)
+
+
+def run(*popenargs, **kwargs):
```
jiasli, line 65: `subprocess.run` only exposes one positional argument `args`, so using `*popenargs` is unnecessary.

src/azure-cli-core/azure/cli/core/cli_subprocess.py
```diff
@@ -0,0 +1,108 @@
...
+
+logger = get_logger(__name__)
+
+ARG_MASK_CHAR = re.compile(r'[$;\s]')
+
+
+def subprocess_arg_type_check(args, kwargs):
+    skip_arg_type_check = kwargs.get("skip_arg_type_check", None)
+    if skip_arg_type_check is not None:
+        del kwargs["skip_arg_type_check"]
```
jiasli, line 32: It should be 

```py
if "skip_arg_type_check" in kwargs:
    del kwargs["skip_arg_type_check"]
```

Otherwise, if `skip_arg_type_check` is `None`, it will not be deleted and passed to `subprocess`.

PR #29515, opened 2024-07-30
{Misc.} Remove deprecated `assertEquals`

bebound:
`assertEquals` is removed in 3.12. It's replaced by `assertEqual`

Ref:

https://docs.python.org/3/whatsnew/3.12.html#id3


yonzhan:
Misc.

wangzelin007 (approved):

zhoxing-ms (approved):

PR #29681, opened 2024-08-08
{doc} Add subprocess security guidance doc

AllyW:
**Related command**

**Description**

**Testing Guide**

**History Notes**

---


yonzhan:
doc

doc/cli_subprocess_guidelines.md
```diff
@@ -0,0 +1,110 @@
...
+cmd = ["git", user_input]
+```
+
+#### Avoid use `shell=True` with user_inputs
+
+When using subprocess module, avoid `shell=True` argument when it comes with cmd containing user inputs. When python executes a command with `shell=True`, the system shell will interpret any string passed in and if this string contains user input without proper sanitization, an attacker can inject malicious commands, as demonstrated in the very beginning of this doc.
+
+
+## Summary
+Ensuring the safety of cli from command injection under subprocess calling requires an in-depth understanding of these vulnerabilities and also proactive measures to counteract potential exploits. Cli developers can either apply the three security practices, if applicable, when using builtin `subprocess`, or use a more centralized func `run_cmd` cli provided, to safeguard cli modules from command injection attack and for future more accessible security enforcements.
```
jsntcy, line 110: Recommend them to use the centralized function `run_cmd` first, if it cannot meet their requirement, then they can write their own function with applying the three security practices.

jsntcy, line 110: Avoid to use abbreviation `func`, use `function` instead which is more straightforward.

doc/cli_subprocess_guidelines.md
```diff
@@ -0,0 +1,110 @@
+# Azure CLI Subprocess Guidelines
+
+In certain cli modules, there are scenarios that need to call a subsystem to run commands outside cli, like getting kubectl info in aks, or deployment setup in mysql using `git` and `gh`, through python built-in [subprocess](https://docs.python.org/3/library/subprocess.html) module. Despite its simplicity and versatility, ensuring the security of applying it can be challenging in different platforms under different circumstance, and it is error-prone for developers to neglect security best practices during development.
```
jiasli, line 3: We usually use upper case "CLI".

doc/cli_subprocess_guidelines.md
```diff
@@ -0,0 +1,110 @@
...
+
+
+## Insecure Use Of Subsystem Commands Under Subprocess Module
+
+
+### Dynamic Command Construction Under Risk
```
jiasli, line 9: No need to capitalize the first letter of each word.

```suggestion
### Dynamic command construction under risk
```

doc/cli_subprocess_guidelines.md
```diff
@@ -0,0 +1,110 @@
...
+subprocess.run(cmd, shell=True)
+```
+
+```commandline
+input a git command to run: --version;echo aa
+git version 2.34.1
+aa
+```
+
+This is a simple example for demonstrating the side effects in python's subsystem improper usage. And it's common for cli developers to build and execute commands dynamically from users' input in a more complicated way. When constructing and executing commands in subprocess through `shell=True`, it exposes a big security vulnerability to potential malicious users outside.
```
jiasli, line 26: subsystem -> subprocess?

https://docs.python.org/3/library/subprocess.html#older-high-level-api doesn't mention the term "subsystem" at all.

jiasli, line 26: subsystem -> subprocess?

https://docs.python.org/3/library/subprocess.html doesn't mention the term "subsystem" at all.

doc/cli_subprocess_guidelines.md
jiasli, line 3: Why using subsystem instead of subprocess? I don't think "subsystem" stands for the scenario this document is talking about.

https://docs.python.org/3/library/subprocess.html doesn't mention the term "subsystem" at all.

doc/cli_subprocess_guidelines.md
```diff
@@ -0,0 +1,110 @@
...
+There are several aspects of security practices that developers need to have in mindset to safeguard their cli modules from command injection attacks.
+
+### Cli Centralized Subsystem Executing
+
+Azure cli provides a centralized function `run_cmd` adapted from official `subprocess.run`, with necessary argument covered and illegal input blocking enforced. 
+
+What developers need to do is:
+1) `from azure.cli.core.util import run_cmd`
+2) replace `subprocess.run` (or `Popen` or `check_call` or `check_output` or `call`) with `run_cmd`.
+3) construct cmd args as array like: [executable, arg0, arg1, arg2, ...]
```
jiasli, line 40:
```suggestion
1. `from azure.cli.core.util import run_cmd`
2. replace `subprocess.run` (or `Popen` or `check_call` or `check_output` or `call`) with `run_cmd`.
3. construct cmd args as array like: `[executable, arg0, arg1, arg2, ...]`
```

doc/cli_subprocess_guidelines.md
```diff
@@ -0,0 +1,110 @@
...
+
+`run_cmd` will add necessary argument type checks and process the input and output the same way as `subprocess.run`, and block potential risks from commands constructed from user input.
+Below is an example for `run_cmd` use case:
+
+```commandline
```
jiasli, line 45:
```suggestion
```py
```

doc/cli_subprocess_guidelines.md
```diff
@@ -0,0 +1,110 @@
...
+output = run_cmd(cmd, check=False, capture_output=True)
+if output.returncode != 0:
+      raise CLIInternalError('Command execution failed, command is: '{}', error message is: {}'.format(cmd, output.stderr))
+return output.stdout
+```
+
+All various kinds of `subprocess` Popen calling use cases can be easily adjusted into `run_cmd` with security risks processed and eliminated in this centralized function.
+
+Besides that, users might need to know some parts of the accessibility in both `run_cmd` and `subprocess`
+1) when calling shell built-in cmds, like `dir` or `echo`, using `shell=True` **in windows platform**, `subprocess` implicitly uses `cmd.exe`, while `run_cmd` asks developers to provide the `cmd.exe` as executable file specifically in the arg list's first item, like `["cmd.exe", "/c", "echo", "abc"]`
```
jiasli, line 70:
```suggestion
1) When calling shell built-in cmds, like `dir` or `echo`, using `shell=True` **in windows platform**, `subprocess` implicitly uses `cmd.exe`, while `run_cmd` asks developers to provide the `cmd.exe` as executable file specifically in the arg list's first item, like `["cmd.exe", "/c", "echo", "abc"]`
```

doc/cli_subprocess_guidelines.md
```diff
@@ -0,0 +1,110 @@
...
+2) if developers want to find an easy way to split their current cmd string into list, **for unix-like platforms**, developers can apply [`shlex.split`](https://docs.python.org/3/library/shlex.html#shlex.split) for quick access. But a prepared cmd statement is still more recommended (for more info about prepared cmd statement, please read below sections).
+3) it might be not that obvious to find target command's executable file **in windows platform**, a tool developer can use is `shutil.which` that gives the executable file path in windows system, like `shutil.which(git)`. The cmd `git --version` can be adjusted as `[shutil.which(git), "--version"]`. Please provide the corresponding executable path in target platforms.
```
jiasli, line 72: Could you share more information on this? I think `subprocess` will look for the executable on `PATH`.

doc/cli_subprocess_guidelines.md
jiasli: Will it be clearer to use "subprocess" instead of "subsystem"? I don't think "subsystem" stands for the scenario this document is talking about.

https://docs.python.org/3/library/subprocess.html doesn't mention the term "subsystem" at all.

jiasli, line 3: Will it be clearer to use "subprocess" instead of "subsystem"? I don't think "subsystem" stands for the scenario this document is talking about.

https://docs.python.org/3/library/subprocess.html doesn't mention the term "subsystem" at all.

doc/cli_subprocess_guidelines.md
AllyW: Yes, subprocess searches executable under `PATH`, as the same way as `shutil.which`. 

In windows platform, cmd-running can be `cmd arg0 arg2..` or `cmd.exe arg0 arg1...`. Just want to give a hint to users how to check the absolute path of corresponding executable file in case they are confused with it. 

Doc adjusted.

doc/cli_subprocess_guidelines.md
AllyW: `subsystem` is taken as a synonym or a phase in the meaning nearby to `subprocess` here.

doc/cli_subprocess_guidelines.md
```diff
@@ -0,0 +1,112 @@
...
+2. if developers want to find an easy way to split their current cmd string into list, **for unix-like platforms**, developers can apply [`shlex.split`](https://docs.python.org/3/library/shlex.html#shlex.split) for quick access. But a prepared cmd statement is still more recommended (for more info about prepared cmd statement, please read below sections). 
+3. if developers want to locate the target command's executable file, a tool developers can use is `shutil.which` that gives the full executable file path in system, like `shutil.which(git)` returns the full `git.exe` path in windows platform `C:\\Program Files\\Git\\cmd\\git.EXE`. 
+4. if the target cmd is `az`-related, like `az group show --name xxxx`, please use internal corresponding function call `cli_ctx.invoke(["az", "group", "show", "--name", "xxx"])` to get the target information.
+
+
+### Best practices in subprocess use cases
+
+
+The following sections discuss some secure coding conventions that, when implemented, can help protect CLI applications from command injection vulnerabilities when calling subsystems through `subprocess`.
```
jiasli, line 80:
```suggestion
The following sections discuss some secure coding conventions that, when implemented, can help protect CLI applications from command injection vulnerabilities when using `subprocess` module.
```

doc/cli_subprocess_guidelines.md
```diff
@@ -0,0 +1,112 @@
...
+user_input = input("input a git command to run: ")
+cmd = f"git {user_input}"
+subprocess.run(cmd, shell=True)
+```
+
+```python
+input a git command to run: --version;echo aa
+git version 2.34.1
+aa
+```
```
jiasli, line 24: This is not Python code.

AllyW: right, it is a mock for the console input and output of running python code above

doc/cli_subprocess_guidelines.md
```diff
@@ -0,0 +1,112 @@
...
+cmd = ["git", user_input]
+```
+
+#### Avoid use `shell=True` with user_inputs
+
+When using subprocess module, avoid `shell=True` argument when it comes with cmd containing user inputs. When python executes a command with `shell=True`, the system shell will interpret any string passed in and if this string contains user input without proper sanitization, an attacker can inject malicious commands, as demonstrated in the very beginning of this doc.
+
+
+## Summary
+Ensuring the safety of Azure CLI from command injection under subprocess calling requires an in-depth understanding of these vulnerabilities and also proactive measures to counteract potential exploits. CLI developers can apply the three security practices, if applicable, when using builtin `subprocess`, but it's recommended to use the centralized function `run_cmd` CLI provided, to safeguard CLI modules from command injection attack and for future more accessible security enforcements.
```
dcaro, line 112: Should we have a more prescriptive approach and require CLI developers to use the `run_cmd` for any subprocess operation to meet our security standards?

AllyW: We can add a checking rule on those unsupervised subprocess call and give cli core's `run_cmd` as the substitution, with this doc provided as assistance.

jiasli (approved):

PR #29689, opened 2024-08-08
{Util} Introduce a centralized function for subprocessing

jiasli:
**Related command**

**Description**
An alternative solution to https://github.com/Azure/azure-cli/pull/29503.

Judging by how `subprocess` is currently used:

https://github.com/Azure/azure-cli/blob/f1d068ece6789d8e77f3a1ee7c140b12982298a8/src/azure-cli/azure/cli/command_modules/storage/azcopy/util.py#L105

https://github.com/Azure/azure-cli/blob/cff8d9bad3e39aa69ab9ca108693c66088d615bc/src/azure-cli/azure/cli/command_modules/resource/_bicep.py#L327-L331

https://github.com/Azure/azure-cli/blob/f4cf2963264145541cc10628af1743e2c93d669f/src/azure-cli/azure/cli/command_modules/mysql/_util.py#L388

https://github.com/Azure/azure-cli/blob/91aace5f851b5e70504efbb89079bb6170d09e74/src/azure-cli/azure/cli/command_modules/rdbms/_flexible_server_util.py#L331

we only need to expose limited arguments:

- `args` (positional)
- `stdin` (keyword)
- `stdout` (keyword)
- `check` (keyword)
- `env` (keyword)

It is not necessary to expose all `*args` and `**kwargs` from `subprocess.run`. 

This PR introduces a centralized function `cmd` for subprocessing with a more explicit and simplified signature.

Internally, to adhere to https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module, `cmd` uses the recommended function `subprocess.run`.

According to https://docs.python.org/3/library/subprocess.html#older-high-level-api, these old APIs are not exposed:

- `call`
- `check_call`
- `check_output`

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/util.py
+++ b/src/azure-cli-core/azure/cli/core/util.py
@@ -1356,3 +1356,16 @@ def should_encrypt_token_cache(cli_ctx):
     encrypt = cli_ctx.config.getboolean('core', 'encrypt_token_cache', fallback=fallback)
 
     return encrypt
+
+
+def run_cmd(args, *, capture_output=False, timeout=None, check=False, encoding=None, env=None):
+    """Run command in a subprocess. It reduces (not eliminates) shell injection by forcing args to be a list
+    and shell=False. Other arguments are keyword-only. For their documentation, see
+    https://docs.python.org/3/library/subprocess.html#subprocess.run
+    """
+    if not isinstance(args, list):
+        raise CLIError("Invalid args. args must be a list.")
+
+    import subprocess
+    return subprocess.run(args, capture_output=capture_output, timeout=timeout, check=check,
+                          encoding=encoding, env=env)
--- a/src/azure-cli/azure/cli/command_modules/resource/_bicep.py
+++ b/src/azure-cli/azure/cli/command_modules/resource/_bicep.py
@@ -28,7 +28,7 @@ from azure.cli.core.azclierror import (
     ClientRequestError,
     InvalidTemplateError,
 )
-from azure.cli.core.util import should_disable_connection_verify
+from azure.cli.core.util import should_disable_connection_verify, run_cmd
 
 # See: https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
 _semver_pattern = r"(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?"  # pylint: disable=line-too-long
@@ -324,20 +324,20 @@ def _extract_version(text):
 
 
 def _run_command(bicep_installation_path, args, custom_env=None):
-    process = subprocess.run(
+    process = run_cmd(
         [rf"{bicep_installation_path}"] + args,
-        stdout=subprocess.PIPE,
-        stderr=subprocess.PIPE,
+        capture_output=True,
+        encoding='utf-8',
         env=custom_env)
 
     try:
         process.check_returncode()
-        command_warnings = process.stderr.decode("utf-8")
+        command_warnings = process.stderr
         if command_warnings:
             _logger.warning(command_warnings)
-        return process.stdout.decode("utf-8")
+        return process.stdout
     except subprocess.CalledProcessError:
-        stderr_output = process.stderr.decode("utf-8")
+        stderr_output = process.stderr
         errors = []
 
         for line in stderr_output.splitlines():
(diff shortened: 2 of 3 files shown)


yonzhan:
Util

src/azure-cli/azure/cli/command_modules/rdbms/_flexible_server_util.py
```diff
@@ -318,18 +318,16 @@ def _resolve_api_version(client, provider_namespace, resource_type, parent_path)
...
+        process = cmd(commands, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
     if process.returncode:
         logger.warning(process.stderr.read().strip().decode('UTF-8'))
 
 
 def run_subprocess_get_output(command):
-    commands = command.split()
-    process = subprocess.Popen(commands, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-    process.wait()
+    command = shlex.split(command)
```
jiasli, line 329: Using a normal `split` is wrong, as it doesn't take quotes into consideration.

src/azure-cli/azure/cli/command_modules/mysql/_util.py
```diff
@@ -374,19 +375,18 @@ def _resolve_api_version(client, provider_namespace, resource_type, parent_path)
 
 
 def run_subprocess(command, stdout_show=None):
+    command = shlex.split(command)
```
jiasli, line 378: The best practice is to construct `command` as a `list` from the very beginning.

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -1356,3 +1356,23 @@ def should_encrypt_token_cache(cli_ctx):
     encrypt = cli_ctx.config.getboolean('core', 'encrypt_token_cache', fallback=fallback)
 
     return encrypt
+
+
+def run_cmd(args, *, capture_output=False, check=False, env=None):
```
jiasli, line 1361: `cmd` is a name widely used in the code base, so we use a more unique name.

src/azure-cli/azure/cli/command_modules/mysql/_util.py
```diff
@@ -374,19 +374,18 @@ def _resolve_api_version(client, provider_namespace, resource_type, parent_path)
...
+        process = run_cmd(command, capture_output=True)
     if process.returncode:
         logger.warning(process.stderr.read().strip().decode('UTF-8'))
 
 
 def run_subprocess_get_output(command):
-    commands = command.split()
-    process = subprocess.Popen(commands, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-    process.wait()
+    command = shlex.split(command)
```
jiasli, line 387: Using a normal `split` is wrong, as it doesn't take quotes into consideration.

AllyW:
The listed keyword (stdin, stdout, check, env ) is not enough for all the usecases in cli, like

 `universal_newlines` and `timeout` in `acs` module here: https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/acs/custom.py#L295-L298, 

and `text` in https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/acs/custom.py#L1772. 

If not all use cases need to be covered, we'll discuss which ones are supported by cli.

src/azure-cli/azure/cli/command_modules/mysql/_util.py
AllyW: Does the POSIX mode need to be mentioned therefore developers may notice the potential difference when splitting their cmds?

Agreed on constructing the list of cmd at the beginning

jiasli: The `command` string is constructed by `mysql` module itself:

https://github.com/Azure/azure-cli/blob/9119731bf5940981c3fe1f65ae495699cdb4b019/src/azure-cli/azure/cli/command_modules/mysql/_util.py#L416

The developer of `mysql` should be aware they should follow POSIX convention to build the `command` string.

jiasli:
> `universal_newlines` and `timeout` in `acs` module here: https://github.com/Azure/azure-cli/blob/bb022828d41467ab8ac423c452bae74f5ef9a249/src/azure-cli/azure/cli/command_modules/acs/custom.py#L295-L297

`universal_newlines` is deprecated according to https://docs.python.org/3/library/subprocess.html#subprocess.run

> The `universal_newlines` argument is equivalent to `text` and is provided for backwards compatibility.

`timeout` can be added to `run_cmd`.

> and `text` in https://github.com/Azure/azure-cli/blob/bb022828d41467ab8ac423c452bae74f5ef9a249/src/azure-cli/azure/cli/command_modules/acs/custom.py#L1772

I prefer not to complicate `run_cmd` too much. It is ok to let the caller `decode`s by itself:

jiasli:
> `universal_newlines` and `timeout` in `acs` module here: https://github.com/Azure/azure-cli/blob/bb022828d41467ab8ac423c452bae74f5ef9a249/src/azure-cli/azure/cli/command_modules/acs/custom.py#L295-L297

`universal_newlines` is deprecated according to https://docs.python.org/3/library/subprocess.html#subprocess.run

> The `universal_newlines` argument is equivalent to `text` and is provided for backwards compatibility.

`timeout` can be added to `run_cmd`.

> and `text` in https://github.com/Azure/azure-cli/blob/bb022828d41467ab8ac423c452bae74f5ef9a249/src/azure-cli/azure/cli/command_modules/acs/custom.py#L1772

I prefer not to complicate `run_cmd` too much. It is ok to let the caller `decode` the output by itself, such as when calling `azcopy`:

https://github.com/Azure/azure-cli/blob/bd27913d0237806e2a4f2bf0d24a78351c58f63a/src/azure-cli/azure/cli/command_modules/storage/azcopy/util.py#L45-L46

jiasli:
Superseded by #29745

PR #29696, opened 2024-08-08
improve exception message on JMESPathChecks when actual result evaluates to false

tobiasb-ms:
**Related command**
Testing issue, so `azdev test` commands.

**Description**
When running scenario tests that use checks, like:
```
self.cmd(f"az some_command --output json", checks=[
    self.check("thing", True)
])
```

If the actual result in the check (in this case the value of `thing`) evaluates to `False` and does not match the expected result, the message in the exception that is raised is misleading. For example:

`````
>           raise JMESPathCheckAssertionError(self._query, self._expected_result, 'None',
                                              execution_result.output)
E           azure.cli.testsdk.exceptions.JMESPathCheckAssertionError: Query 'thing' doesn't yield expected value 'True', instead the actual value is 'None'. Data:
E           {
E             "otherThing": "value",
E             "thing": false
E           }
```

Here you can see that `thing` clearly is not `None`, despite the exception message.

This happens because in [JMESPatchCheck.__call__](https://github.com/Azure/azure-cli/blob/c43b13643970803326f46a675da91fcb23970ec6/src/azure-cli-testsdk/azure/cli/testsdk/checkers.py#L32-L37), the code that raises the error has this check:

```
if not equals:
    if actual_result:
        raise JMESPathCheckAssertionError(self._query, self._expected_result, actual_result,
                                          execution_result.output)
    raise JMESPathCheckAssertionError(self._query, self._expected_result, 'None',
                                      execution_result.output)
```

If `actual_result` is something like `False`, `""`, `[]`, `0`, or anything else would not go into the `if` clause, then the message will say the value is `None` rather than what it really is, which is confusing and can make debugging more difficult.

To fix this, I've simply removed the `if actual_result` check, since `None` will be formatted correctly if passed in to the error constructor anyway.

**Testing Guide**
`azdev test` passes all tests.

I also added unit tests for this, which can be run with
```
pytest src/azure-cli-testsdk/azure/cli/testsdk/scenario_tests/tests/
```

**History Notes**
{azure-cli-testsdk} Fix error message for certain failed tests.

---

### DIFF
--- a/src/azure-cli-testsdk/azure/cli/testsdk/checkers.py
+++ b/src/azure-cli-testsdk/azure/cli/testsdk/checkers.py
@@ -30,11 +30,8 @@ class JMESPathCheck(object):  # pylint: disable=too-few-public-methods
             equals = actual_result == self._expected_result \
                 or str(actual_result).lower() == str(self._expected_result).lower()
         if not equals:
-            if actual_result:
-                raise JMESPathCheckAssertionError(self._query, self._expected_result, actual_result,
-                                                  execution_result.output)
-            raise JMESPathCheckAssertionError(self._query, self._expected_result, 'None',
-                                              execution_result.output)
+            raise JMESPathCheckAssertionError(self._query, self._expected_result, actual_result,
+                                                execution_result.output)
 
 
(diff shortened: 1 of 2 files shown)


yonzhan:

Issue #5184, opened 2018-01-02
Homebrew Installed Azure CLI not working - ValueError: field 6 out of range (need a 48-bit value)


markread007:
Having installed azure-cli via homebrew I get the following python errors whenever I try and run a command.  e.g. if I perform az login, or az --help, or other commands I get the following errors:-

$ az login
field 6 out of range (need a 48-bit value)
Traceback (most recent call last):
  File "/usr/local/Cellar/azure-cli/2.0.23/libexec/lib/python3.6/site-packages/azure/cli/main.py", line 36, in main
    cmd_result = APPLICATION.execute(args)
  File "/usr/local/Cellar/azure-cli/2.0.23/libexec/lib/python3.6/site-packages/azure/cli/core/application.py", line 141, in execute
    self.refresh_request_id()
  File "/usr/local/Cellar/azure-cli/2.0.23/libexec/lib/python3.6/site-packages/azure/cli/core/application.py", line 269, in refresh_request_id
    self.session['headers']['x-ms-client-request-id'] = str(uuid.uuid1())
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/uuid.py", line 606, in uuid1
    clock_seq_hi_variant, clock_seq_low, node), version=1)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/uuid.py", line 168, in __init__
    raise ValueError('field 6 out of range (need a 48-bit value)')
ValueError: field 6 out of range (need a 48-bit value)

### Environment summary

Install Method (e.g. pip, interactive script, apt-get, Docker, MSI, edge build) / CLI version (`az --version`) / OS version / Shell Type (e.g. bash, cmd.exe, Bash on Windows)  

I installed via home-brew, having initially tried to install using the curl method, however this failed to build.  The home-brew version installed fine, but just doesn't work.

$ az --version
azure-cli (2.0.23)

acr (2.0.17)
acs (2.0.22)
advisor (0.1.0)
appservice (0.1.22)
backup (1.0.3)
batch (3.1.7)
batchai (0.1.3)
billing (0.1.6)
cdn (0.0.10)
cloud (2.0.10)
cognitiveservices (0.1.9)
command-modules-nspkg (2.0.1)
configure (2.0.12)
consumption (0.2.0)
container (0.1.15)
core (2.0.23)
cosmosdb (0.1.15)
dla (0.0.15)
dls (0.0.18)
eventgrid (0.1.5)
[...]

PR #5644, opened 2018-02-22
Bug fix for 'ValueError: field 6 out of range (need a 48-bit value)'

derekbekoe:
Closes https://github.com/Azure/azure-cli/issues/5184

---

This checklist is used to make sure that common guidelines for a pull request are followed.

### General Guidelines

- [x] The PR has modified HISTORY.rst describing any customer-facing, functional changes. Note that this does not include changes only to help content. (see [Modifying change log](https://github.com/Azure/azure-cli/tree/master/doc/authoring_command_modules#modify-change-log)).

PR #29706, opened 2024-08-12
{Core} Remove uuid workaround code

bebound:
This code was introduced in https://github.com/Azure/azure-cli/pull/5644 to fix https://github.com/Azure/azure-cli/issues/5184.

Since it's fixed in upstream https://github.com/python/cpython/pull/5254 a long time ago, we don't need it anymore.


yonzhan:
Core

jiasli:
This PR will simplify https://github.com/Azure/azure-cli/pull/29888.

PR #29717, opened 2024-08-15
{Telemetry} Fix telemetry lost due to concurrency issue

evelyn-ys:
**Description**

This PR is for fixing telemetry lost due to concurrency issue.

At the stage when we still had telemetry local cache, here's the process of uploading telemetry:
1. save cli records in telemetry cache file
2. check telemetry note file
2.1 if it's not enough time since last telemetry uploading time, exit
2.2 if it has been long enough since last telemetry uploading time, start a subprocess for uploading
3. within the uploading subprocess
3.1 check telemetry note file again to ensure it has been enough time since last telemetry sent time
3.2 get `last_sent_time` from telemetry note file
3.3 read all cli records from telemetry cache file where `record_time > last_sent_time`
3.4 remove cache file as it has been loaded in 3.3
3.5 uploading all records from 3.3
3.6 update `last_sent_time` for telemetry note file

Since https://github.com/Azure/azure-cli/pull/26854, we have dropped telemetry cache. So step 2 and step 3.1 are removed. We won't check if it has been "enough time" since last uploading, we do the uploading every time. So here's the new process:
1. save cli records in telemetry cache file
2. start a subprocess for uploading
3. within the uploading subprocess
3.1 get `last_sent_time` from telemetry note file
3.2 read all cli records from telemetry cache file where `record_time > last_sent_time`
3.3 remove cache file as it has been loaded in 3.2
3.4 uploading all records from 3.3
3.5 update `last_sent_time` for telemetry note file

But this enlarges concurrency issue as we increased the frequency. Operations on two files may encounter concurrency issue, one is the telemetry cache file (step 1, step 3.2, step 3.3), another is telemetry note file (step 3.1, step 3.5):
- Imagine a subprocess for command A and main process for command B are executing at the same time and the real execution order is subprocess A 3.2 -> main process B -> subprocess A 3.3, then we will lose the cli record for command B because it's not loaded in 3.2 but deleted in 3.3
- Then for telemetry note file, there's lock mechanism. If subprocess A and subprocess B are both acquiring a lock to access telemetry note file, only one will success, the other will fail with `portallocker.AlreadyLocked` error, see https://github.com/Azure/azure-cli/blob/ad96375e3940108bb011e4efb51be0cd748af4a7/src/azure-cli-telemetry/azure/cli/telemetry/components/telemetry_note.py#L72-L76 https://github.com/Azure/azure-cli/blob/ad96375e3940108bb011e4efb51be0cd748af4a7/src/azure-cli-telemetry/azure/cli/telemetry/__init__.py#L109-L112 The failed one will exit uploading directly which means we may also lose the telemetry

So the fix also comes from two part:
- for the telemetry cache file, we log in separate cache file for each cli execution to avoid reading and writing for the same file
- for the telemetry note file, as we have already dropped the last sent time check, we don't need telemetry not file either, so just simply remove step 3.1 and step 3.5

**Testing Guide**

Run a cli script which contains many cli commands
Check $HOME/.azure/logs/telemetry.log and see if all command records have been uploaded

---

### DIFF
--- a/src/azure-cli-telemetry/azure/cli/telemetry/__init__.py
+++ b/src/azure-cli-telemetry/azure/cli/telemetry/__init__.py
@@ -42,8 +42,8 @@ def _start(config_dir, cache_dir):
             kwargs['stdout'] = subprocess.DEVNULL
             kwargs['stderr'] = subprocess.STDOUT
 
-    subprocess.Popen(**kwargs)
-    logger.info('Return from creating process')
+    process = subprocess.Popen(**kwargs)
+    logger.info('Return from creating process %s', process.pid)
 
 
 def save(config_dir, payload):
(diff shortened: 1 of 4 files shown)


yonzhan:
Telemetry

MoChilia (approved):
Tested on GitHub Actions. Everything looks great, and telemetry is fully recorded.

PR #29737, opened 2024-08-20
{Packaging} Use Azure Linux 3 GA image

bebound:
Use GA image to replace preview image.

Ref: https://mcr.microsoft.com/en-us/product/azurelinux/base/core/about


yonzhan:
Packaging

wangzelin007 (approved):

PR #29628, opened 2024-08-01
{Packaging} Add warning message for Alpine-based image

jiasli:
Add global warning message to drive the transition to Azure Linux based image.

PR #29741, opened 2024-08-20
{Packaging} Drop Alpine docker image

bebound:
Alpine is deprecated in https://github.com/Azure/azure-cli/pull/29628
We also removed it in pipeline.


yonzhan:
Packaging

jiasli:
Should we remove `alpine.dockerfile` too?

azure-pipelines.yml
```diff
@@ -324,83 +324,6 @@ jobs:
...
-- job: BuildDockerImageAlpine
-  continueOnError: true
-  displayName: Build Docker Image Alpine
-  strategy:
-    matrix:
-      ${{ each arch in parameters.architectures }}:
-        ${{ arch.name }}:
-          pool: ${{ arch.pool }}
-          artifactName: docker-${{ arch.value }}
-          dockerfile: alpine.dockerfile
```
jiasli, line 336: Should we remove `alpine.dockerfile` too?

bebound: I don't remove it because `centos7.dockerfile` and `fedora.dockerfile` are kept.
I prefer to remove it as it can be restored in git.

jiasli: Then let's do it.

jiasli (approved):

PR #29866, opened 2024-09-09
{C} Use Azure Artifacts feeds in release branch

bebound:
Use Azure Artifacts feeds in release branch build process: PyPI, MSI, ZIP, DEB, and RPM

The public pipeline still uses PyPI as it's hard to keep the CFS feed token secret.

Ref:
- [CFS PyPI Configuration](https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-docs/secure-supply-chain/secure-supply-chain-analysis-task/cfs-detectors/pypi/cfs-pypi)
- [Central Feed Services in opensource project](https://docs.opensource.microsoft.com/security/cfs/)


yonzhan:
Packaging

yonzhan (approved):

scripts/release/rpm/pipeline_azurelinux.sh
```diff
@@ -16,6 +16,7 @@ docker build \
     --target build-env \
     --build-arg cli_version=${CLI_VERSION} \
     --build-arg image=${IMAGE} \
+    --secret id=PIP_INDEX_URL \
```
bebound, line 19: `--build-arg` also works, but I prefer `--secret` as the index url contains credential and secret does not record its value in image.

Ref: https://docs.docker.com/build/building/secrets/

wangzelin007 (approved):

azure-pipelines.yml
```diff
@@ -205,6 +205,10 @@ jobs:
       TargetPath: '$(Build.ArtifactStagingDirectory)/metadata'
       artifactName: metadata
 
+  - task: PipAuthenticate@1
+    displayName: 'Pip Authenticate'
+    inputs:
+      artifactFeeds: 'public/azure-cli-feed'
```
bebound, line 211: This task set `PIP_INDEX_URL` so that this feed can be used later.

Ref: https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/pip-authenticate-v1?view=azure-pipelines

wangzelin007 (approved):

azure-pipelines.yml
```diff
@@ -6,6 +6,8 @@ trigger:
   branches:
     include:
     - '*'
+    exclude:
+    - 'release'
```
bebound, line 10: The public pipeline does not have the permission to run `PipAuthenticate@1` task, so skip it. This also saves some resource during CLI release.

scripts/release/rpm/pipeline_azurelinux.sh
jiasli: It is interesting that `env` can be omitted:

https://docs.docker.com/build/building/secrets/#sources

> When you use secrets from environment variables, you can omit the `env` parameter to bind the secret to a file with the same name as the variable.

scripts/release/rpm/pipeline_azurelinux.sh
```diff
@@ -11,11 +11,15 @@ set -exv
 
 CLI_VERSION=`cat src/azure-cli/azure/cli/__main__.py | grep __version__ | sed s/' '//g | sed s/'__version__='// |  sed s/\"//g`
 
+# PIP_INDEX_URL env must exist in `docker build --secret`, use an empty string if it doesn't exist.
+export PIP_INDEX_URL=${PIP_INDEX_URL}
```
bebound, line 15: Error message without empty env: `ERROR: failed to stat PIP_INDEX_URL: stat PIP_INDEX_URL: no such file or directory`

wangzelin007 (dismissed):

azure-pipelines.yml
```diff
@@ -205,6 +207,11 @@ jobs:
       TargetPath: '$(Build.ArtifactStagingDirectory)/metadata'
       artifactName: metadata
 
+  - task: PipAuthenticate@1
+    condition: eq(variables['Build.SourceBranch'], 'refs/heads/release')
+    displayName: 'Pip Authenticate'
+    inputs:
+      artifactFeeds: ${{ variables.AZURE_ARTIFACTS_FEEDS }}
```
jiasli, line 214: Other tasks use [macro syntax](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#macro-syntax-variables) as `$(variables.AZURE_ARTIFACTS_FEEDS)`. Does [template expression syntax](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#template-expression-syntax) work here

bebound: You're right, the variable defined in ADO UI is empty when use `${{ variables.AZURE_ARTIFACTS_FEEDS }}`. I've changed to `$(AZURE_ARTIFACTS_FEEDS)`

jiasli (approved):

wangzelin007 (approved):

PR #29880, opened 2024-09-10
{Core} Register Upcoming Breaking Change on Empty Command Group

ReaNAiveD:
**Related command**

**Description**

If a developer creates a subcommand or subcommand group without creating its parent command group, the command group is registered as `{}` in the `command_group_table`. Previously, when announcing breaking changes on the command group, this would result in registering breaking changes for all its subcommands and subcommand groups, leading to an inconsistent display with `deprecate_info`.

This PR introduces the creation of an `AzCommandGroup` object in registering breaking changes, which will be registered in the `command_group_table` in such cases. This change ensures a consistent display of breaking changes and improves the overall management of command groups.

**Testing Guide**

As an example, you can create a file `_breaking_change.py` in `dla` command module and add the following lines to it.

```python
from azure.cli.core.breaking_change import register_command_group_deprecate
register_command_group_deprecate('dla')
```

Then delete the following line in `command.py` file.

```python
    with self.command_group('dla', is_preview=True):
        pass
```

Finally, run `az dla -h`. You should see the warning displaying correctly.

**History Notes**

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/breaking_change.py
+++ b/src/azure-cli-core/azure/cli/core/breaking_change.py
@@ -250,8 +250,10 @@ class BreakingChange(abc.ABC):
         elif self.is_command_group(cli_ctx):
             command_group = cli_ctx.invocation.commands_loader.command_group_table[self.command_name]
             if not command_group:
-                self._register_to_direct_sub_cg_or_command(cli_ctx, self.command_name, self.to_tag(cli_ctx))
-            else:
+                loader = self.find_suitable_command_loader(cli_ctx)
+                command_group = loader.command_group(self.command_name)
+                cli_ctx.invocation.commands_loader.command_group_table[self.command_name] = command_group
+            if command_group:
                 command_group.group_kwargs['deprecate_info'] = \
                     self.appended_status_tag(cli_ctx, command_group.group_kwargs.get('deprecate_info'),
                                              self.to_tag(cli_ctx))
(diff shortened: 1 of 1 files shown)


yonzhan:
Core

zhoxing-ms (approved):

jiasli:
This issue was discovered from https://github.com/Azure/azure-cli/pull/29880.

After removing

```py
    with self.command_group('dla', is_preview=True):
        pass
```

`register_command_group_deprecate` no longer works.

src/azure-cli-core/azure/cli/core/breaking_change.py
```diff
@@ -250,8 +250,10 @@ def register(self, cli_ctx):
         elif self.is_command_group(cli_ctx):
             command_group = cli_ctx.invocation.commands_loader.command_group_table[self.command_name]
             if not command_group:
-                self._register_to_direct_sub_cg_or_command(cli_ctx, self.command_name, self.to_tag(cli_ctx))
-            else:
+                loader = self.find_suitable_command_loader(cli_ctx)
```
kairu-ms, line 253: Will multiple modules founded for one command group?

ReaNAiveD: If multiple modules are found for one command group, the method `find_suitable_command_loader` will always choose the first one to create the command group. This is because our goal is simply to create a command group object. The command groups created by different command loaders through `CommandLoader.command_group` are almost identical.

kairu-ms (approved):

PR #29888, opened 2024-09-11
{Misc.} Refactor `__main__.py` and `cli_main()`

jiasli:
**Description**
Refactor `__main__.py` and `cli_main()`:

1. Add `__name__ == "__main__"` detection so that `azure.cli.__main__` can be imported and `cli_main()` can be invoked by other scripts.
2. The telemetry logic is moved into `cli_main()` so that calling `cli_main()` will send the client telemetry.

Example of using `cli_main()`:

```py
from azure.cli.__main__ import cli_main, get_default_cli

cli = get_default_cli()
cli_main(cli, ['version'])

# retrieve the output
print(cli.result.result)

cli = get_default_cli()
cli_main(cli, ['group', 'show', '--help'])
```

### DIFF
--- a/src/azure-cli/azure/cli/__main__.py
+++ b/src/azure-cli/azure/cli/__main__.py
@@ -33,97 +33,98 @@ logger = get_logger(__name__)
 
 
 def cli_main(cli, args):
-    return cli.invoke(args)
+    telemetry.set_application(cli, ARGCOMPLETE_ENV_NAME)
 
-
-az_cli = get_default_cli()
-
-telemetry.set_application(az_cli, ARGCOMPLETE_ENV_NAME)
-
-# Log the init finish time
-init_finish_time = timeit.default_timer()
-exit_code = None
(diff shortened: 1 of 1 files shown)


yonzhan:
Misc.

jiasli:
Thank you so much @online. This is great advice. I also feel this `__main__.py` file is too heavy and against the guideline:

https://docs.python.org/3/library/__main__.html#id1

> The content of `__main__.py` typically isn’t fenced with an `if __name__ == '__main__'` block. Instead, those files are kept short and import functions to execute from other modules.

The initial motivation of this PR is to aid https://github.com/Azure/azure-cli/pull/29899 so that client telemetry can be included in sub-command executions, but this PR is merely a prototype.

We will revisit this PR if we receive enough request for client telemetry in those sub-command executions.

online:
@jiasli you are welcome, bro 🤝

Issue #29828, opened 2024-09-03
Bicep stopped working on azure-cli:latest Docker image (2.64.0)


dallmair:
### Describe the bug

We're using Bicep to deploy our Azure resources. That worked fine up to Azure CLI 2.63.0 Docker image, but since today with 2.64.0 it fails. Please note we're using `azure-cli:latest`, so we expected some of our own stuff to break when moving from Alpine to CBL Mariner 2.0. However, we did not expect Bicep to break.

### Related command

`az deployment sub create --subscription $SUBSCRIPTION_ID --name $NAME --location "North Europe" --template-file ./main.bicep`

### Errors

```
ERROR: Process terminated. Couldn't find a valid ICU package installed on the system. Please install libicu (or icu-libs) using your package manager and try again. Alternatively you can set the configuration flag System.Globalization.Invariant to true if you want to run with no globalization support. Please see https://aka.ms/dotnet-missing-libicu for more information.
   at System.Environment.FailFast(System.String)
   at System.Globalization.GlobalizationMode+Settings..cctor()
   at System.Globalization.GlobalizationMode+Settings.get_Invariant()
   at System.Globalization.GlobalizationMode.get_Invariant()
   at System.Globalization.CultureData.CreateCultureWithInvariantData()
   at System.Globalization.CultureData.get_Invariant()
   at System.Globalization.CultureInfo..cctor()
   at System.Globalization.CultureInfo.get_InvariantCulture()
   at System.Globalization.NumberFormatInfo.get_InvariantInfo()
   at System.Reflection.AssemblyNameParser.ParseVersion(System.String)
   at System.Reflection.AssemblyNameParser.Parse()
   at System.Reflection.AssemblyNameParser.Parse(System.String)
   at System.Reflection.AssemblyName..ctor(System.String)
   at System.Reflection.TypeNameParser.ResolveAssembly(System.String)
   at System.Reflection.TypeNameParser.GetType(System.String, System.ReadOnlySpan`1<System.String>, System.String)
   at System.Reflection.TypeNameParser+NamespaceTypeName.ResolveType(System.Reflection.TypeNameParser ByRef, System.String)
   at System.Reflection.TypeNameParser.Parse()
   at System.Reflection.TypeNameParser.GetTypeReferencedByCustomAttribute(System.String, System.Reflection.RuntimeModule)
   at System.Reflection.CustomAttributeTypedArgument.ResolveType(System.Reflection.RuntimeModule, System.String)
[...]

PR #29897, opened 2024-09-12
Enable bicep in globalization invariant mode for Mariner image

anthony-c-martin:
**Related command**
`az bicep`, `az deployment`, `az stack`

**Description**
Fix for #29828 to conditionally enable invariant globalization for Bicep running inside the Azure CLI docker image.

**Testing Guide**
See changes to `azure-pipelines.yml`

**History Notes**

[ARM] `az bicep`, `az deployment`, `az stack`: Conditionally enable invariant globalization for Bicep running inside the Azure CLI docker image

---


yonzhan:

azure-linux.dockerfile
```diff
@@ -10,5 +10,8 @@ FROM $IMAGE
 # jq: It's widely used for parsing JSON output in Azure CLI and has a small size. See https://github.com/Azure/azure-cli/issues/29830
 RUN --mount=type=bind,target=/azure-cli.rpm,source=./docker-temp/azure-cli.rpm tdnf install ca-certificates jq /azure-cli.rpm -y && tdnf clean all && rm -rf /var/cache/tdnf
 
+# See https://github.com/Azure/azure-cli/issues/29828 for background on this
+ENV AZ_BICEP_GLOBALIZATION_INVARIANT=1
```
bebound, line 14: `ENV AZ_INSTALLER=DOCKER` is sufficient to detect the Docker environment.

anthony-c-martin: While that's technically true, I think it's clearer to define in a separate environment variable. If someone wants to build an image on top of this one with `libicu` installed, they would just need to unset `AZ_BICEP_GLOBALIZATION_INVARIANT`.

There may also be scenarios outside of this container image where it's useful for someone to be able to set this env variable.

Or we may want to distinguish between a "mariner-based docker image" and a "non mariner-based docker image", and change the value of `AZ_BICEP_GLOBALIZATION_INVARIANT` but set `AZ_INSTALLER` to `DOCKER` in both cases.

zhoxing-ms:
@shenglol Could you please help review this Bicep related PR?

azure-pipelines.yml
```diff
@@ -413,7 +413,7 @@ jobs:
         echo "== Test docker image =="
         
         docker load < $TAR_FILE
-        docker run $IMAGE_NAME /bin/bash -c "time az self-test && time az --version && sleep 5"
+        docker run $IMAGE_NAME /bin/bash -c "time az self-test && time az --version && time az bicep version && sleep 5"
```
bebound, line 416: This fails as bicep is not installed.

anthony-c-martin: Do you think it's OK to install Bicep as part of this script? I'd really like to verify the end-to-end works here, just to avoid any new compatibility issues in the future.

E.g.
```
docker run $IMAGE_NAME /bin/bash -c "time az self-test && time az --version && time az bicep install && time az bicep version && sleep 5"
```

bebound: This task is a general test and not meant for specialized testing. All tests should put into the `tests` folder.

In this case, .NET ensures programs can run without `icu` using `DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=1`. I think adding a test to ensures the `az bicep version` still works with this env is sufficient.

Ref: https://learn.microsoft.com/en-us/dotnet/core/runtime-config/globalization#invariant-mode

anthony-c-martin: Makes sense. In this case I wasn't so worried about globalization - was more just thinking about what else could go wrong in the future if other capabilities are stripped out of the container image.

bebound (approved):

zhoxing-ms (approved):

zhoxing-ms:
@shenglol Could you please help review the Bicep related PR as well?

shenglol (approved):
LGTM!

PR #29922, opened 2024-09-18
{Packaging} Add `pip` to dependency

bebound:
Homebrew stops install `pip` in Python 3.12 venv(https://github.com/Homebrew/brew/pull/16306), which is needed in `az extension add`.

Ref: https://github.com/orgs/Homebrew/discussions/5600

### DIFF
--- a/src/azure-cli-core/setup.py
+++ b/src/azure-cli-core/setup.py
@@ -58,6 +58,7 @@ DEPENDENCIES = [
     'msrestazure~=0.6.4',
     'packaging>=20.9',
     'paramiko>=2.0.8,<4.0.0',
+    'pip',  # Homebrew does not install pip in venv, `az extension add` needs it
     'pkginfo>=1.5.0.1',
     # psutil can't install on cygwin: https://github.com/Azure/azure-cli/issues/9399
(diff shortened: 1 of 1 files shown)


yonzhan:
Packaging

src/azure-cli-core/setup.py
```diff
@@ -58,6 +58,7 @@
     'msrestazure~=0.6.4',
     'packaging>=20.9',
     'paramiko>=2.0.8,<4.0.0',
+    'pip',  # Homebrew does not install pip in venv, `az extension add` needs it
```
bebound, line 61: `extension` module is implemented in core, so I add it to `azure-cli-core`

bebound:
This PR is only for Homebrew. In other distros, `pip` is a prerequisite for installing azure-cli.

However, brew calls `pip` with `--no-deps`, I have to manually add `setuptools` and `pip` into resource. 

https://github.com/Homebrew/brew/blame/cd651094878e88cb710bc24e253bff8b3595258e/Library/Homebrew/formula.rb#L1855

Ref: see https://github.com/Homebrew/homebrew-core/pull/194884#issuecomment-2428467302

PR #29972, opened 2024-09-25
{CI} Use macOS-14 agent

bebound:
Test homebrew task fails because macos12 is deprecated.
```
Error: You are using macOS 12.
We (and Apple) do not provide support for this old version.
It is expected behaviour that some formulae will fail to build in this old version.
It is expected behaviour that Homebrew will be buggy and slow.
```
Homebrew can't be installed on macOS 12: https://github.com/Homebrew/brew/pull/18323

Ref:
- https://dev.azure.com/azclitools/public/_build/results?buildId=192519&view=logs&j=cc48ae93-a624-59ff-8a3e-2ebdb4b89e58&t=acd97aa3-1ffd-5990-e2c3-4ea47f198c07
- https://github.com/orgs/Homebrew/discussions/5603

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -138,7 +138,7 @@ jobs:
   displayName: 'Verify src/azure-cli/requirements.*.Darwin.txt'
   condition: succeeded()
   pool:
-    vmImage: 'macOS-12'
+    vmImage: ${{ variables.macos_pool }}
 
   steps:
   - task: UsePythonVersion@0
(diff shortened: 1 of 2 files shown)


yonzhan:
CI

azure-pipelines.yml
```diff
@@ -138,7 +138,7 @@ jobs:
   displayName: 'Verify src/azure-cli/requirements.*.Darwin.txt'
   condition: succeeded()
   pool:
-    vmImage: 'macOS-12'
+    vmImage: ${{ variables.macos_pool }}
```
bebound, line 141: > vmImage string.
Name of the VM image you want to use; valid only in the Microsoft-hosted pool.

Ref: https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/pool?view=azure-pipelines#pool-name-demands-vmimage

wangzelin007 (approved):

jiasli (approved):

PR #29973, opened 2024-09-25
{Core} Add `arm_tools.py`

jiasli:
**Description**
`parse_resource_id`, `is_valid_resource_id`, `resource_id` were initially from `msrestazure.tools`. They were later migrated to `azure.mgmt.core.tools`.

This PR puts them in `azure.cli.core.arm_tools` so that command modules take no direct dependency on `azure.mgmt.core.tools`.

For now, this `azure.cli.core.arm_tools` is a proxy to `azure.mgmt.core.tools`. If there is a Track 3 SDK in the future, the migration would also be easier.

As Azure CLI is moving towards decoupling from SDK, CLI should eventually provide its own implementations.


yonzhan:
Core

PR #30062, opened 2024-10-10
{Core} Only pass `data` to MSAL as `kwargs`

jiasli:
**Description**
Azure CLI currently passes `kwargs` received by `get_token()` directly to MSAL. Since SDK is adding more keyword arguments, such as `enable_cae` (https://github.com/Azure/azure-sdk-for-python/pull/37358), this will cause failure as `enable_cae` is not recognized by MSAL.

This PR changes the behavior so that only `data` is passed to MSAL as `kwargs`.

Also see https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/755

**Testing Guide**
1. Follow https://learn.microsoft.com/en-us/entra/identity/devices/howto-vm-sign-in-azure-ad-linux to enable connecting via Microsoft Entra ID and OpenSSH
2. `az extension add --name ssh`
3. `az ssh vm -n xxx -g xxx`

### DIFF
--- a/src/azure-cli-core/azure/cli/core/auth/credential_adaptor.py
+++ b/src/azure-cli-core/azure/cli/core/auth/credential_adaptor.py
@@ -57,12 +57,13 @@ class CredentialAdaptor:
     def get_token(self, *scopes, **kwargs):
         logger.debug("CredentialAdaptor.get_token: scopes=%r, kwargs=%r", scopes, kwargs)
 
-        # SDK azure-keyvault-keys 4.5.0b5 passes tenant_id as kwargs, but we don't support tenant_id for now,
-        # so discard it.
-        kwargs.pop('tenant_id', None)
+        # Discard unsupported kwargs: tenant_id, enable_cae
+        filtered_kwargs = {}
+        if 'data' in kwargs:
+            filtered_kwargs['data'] = kwargs['data']
 
         scopes = _normalize_scopes(scopes)
-        token, _ = self._get_token(scopes, **kwargs)
+        token, _ = self._get_token(scopes, **filtered_kwargs)
         return token
 
     def get_auxiliary_tokens(self, *scopes, **kwargs):


yonzhan:
Core

evelyn-ys (approved):

src/azure-cli-core/azure/cli/core/auth/credential_adaptor.py
```diff
@@ -57,12 +57,13 @@ def signed_session(self, session=None):
...
-        # so discard it.
-        kwargs.pop('tenant_id', None)
+        # Discard unsupported kwargs: tenant_id, enable_cae
+        filtered_kwargs = {}
+        if 'data' in kwargs:
+            filtered_kwargs['data'] = kwargs['data']
 
         scopes = _normalize_scopes(scopes)
-        token, _ = self._get_token(scopes, **kwargs)
+        token, _ = self._get_token(scopes, **filtered_kwargs)
```
rayluo, line 66: As of this PR, it looks like we don't need that `filtered_kwargs` at all.

```suggestion
        token, _ = self._get_token(scopes, data=kwargs.get("data"))
```

Open question:

Currently there are multiple `get_token(..., **kwargs)` function scattering [here](https://github.com/Azure/azure-cli/blob/azure-cli-2.65.0/src/azure-cli-core/azure/cli/core/auth/credential_adaptor.py#L57) and [there](https://github.com/Azure/azure-cli/blob/azure-cli-2.65.0/src/azure-cli-core/azure/cli/core/auth/msal_credentials.py#L53). If/since we say - via this PR - that only `data` is an expected input parameter, why don't we take one step further and refactor those functions to `get_token(..., data=None)`?

Even if we somehow still want to accept and then ignore arbitrary parameters, a more readable pattern is `def get_token(..., data=None, **ignored)` and then we do NOT relay the `ignored` into underlying functions.

jiasli: The reason why I don't use `data=kwargs.get("data")` is because this will cause `data` always to be passed, even as `None`. I want to make sure `data` is passed only when the caller of `get_token` passes `data`.

As for not making `data` a public argument, this is because Azure CLI Core is also used by other people or projects, I don't want to make `data` a public argument before MSAL does the same (https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/755).

rayluo: > The reason why I don't use `data=kwargs.get("data")` is because this will cause `data` always to be passed, even as `None`. I want to *make sure `data` is passed only when the caller of `get_token` passes `data`*.

Fair enough. In that case, I would suggest you put that reasoning as a comment at the line of `filtered_kwargs`, to prevent a future maintainer refactoring it to the simpler way.


> As for not making `data` a public argument, this is because Azure CLI Core is also used by other people or projects, I don't want to make `data` a public argument before MSAL does the same ([AzureAD/microsoft-authentication-library-for-python#755](https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/755)).

Good point. This PR is ready to go as-is, without https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/755 (If we are really going to expose that in MSAL, we will revisit its naming.)

rayluo (approved):
Approved with suggested changes.

src/azure-cli-core/azure/cli/core/auth/credential_adaptor.py
jiasli: > I would suggest you put that reasoning as a comment at the line of `filtered_kwargs`, to prevent a future maintainer refactoring it to the simpler way.

I am the only one maintaining this part. Let's keep it as is, as I will change it soon.

bebound: One silly question, are `tenant_id` and `data` the only known `kwargs` that might be passed to `get_token`?
`def get_token(self, *scopes, **kwargs)`

jiasli: We only support `scopes` and `data`, but not `tenant_id`, `enable_cae`. That's why anything else in `kwargs` is discarded.

bebound (approved):

src/azure-cli-core/azure/cli/core/auth/credential_adaptor.py
```diff
@@ -57,12 +57,13 @@ def signed_session(self, session=None):
     def get_token(self, *scopes, **kwargs):
         logger.debug("CredentialAdaptor.get_token: scopes=%r, kwargs=%r", scopes, kwargs)
 
-        # SDK azure-keyvault-keys 4.5.0b5 passes tenant_id as kwargs, but we don't support tenant_id for now,
-        # so discard it.
-        kwargs.pop('tenant_id', None)
+        # Discard unsupported kwargs: tenant_id, enable_cae
+        filtered_kwargs = {}
+        if 'data' in kwargs:
+            filtered_kwargs['data'] = kwargs['data']
```
jiasli, line 63: `claims` is a supported keyword argument of `UserCredential.get_token`:

https://github.com/Azure/azure-cli/blob/95ef12647b830f758b73db547e8166ef9a673113/src/azure-cli-core/azure/cli/core/auth/msal_credentials.py#L55

Not passing `claims` breaks CAE silent re-authentication. Interestingly, no issue is reported. 🤔

PR #30163, opened 2024-10-23
[Core] BREAKING CHANGE: `az extension add/update` and `az upgrade`: Install stable version for extension by default, use preview if unapplicable

AllyW:
**Related command**

**Description**

This pr changes the default value of `allow-preview` from `True` to `False` during `az extension add/update`, `az upgrade` and dynamic installation. It's based on the discussion of pr https://github.com/Azure/azure-cli/pull/28913

**Testing Guide**

**History Notes**

[Extension] BREAKING CHANGE: `az extension add/update`: Set default `--allow-preview` value from `True` to be `False` for extensions installation and remain `True` for extensions without stable releases
[Upgrade] BREAKING CHANGE: `az upgrade`: Set default `--allow-preview` value from `True` to be `False` for extensions installation and remain `True` for extensions without stable releases

---


yonzhan:

wangzelin007 (approved):

wangzelin007:
@AllyW 
Base on my understanding:
The installation won't fail because there is fallback logic. If no stable version is available, we will install the latest preview version. The only impact is that some extensions that default to installing the preview version may, due to this change, default to installing the stable version instead.

AllyW:
> @AllyW Base on my understanding: The installation won't fail because there is fallback logic. If no stable version is available, we will install the latest preview version. The only impact is that some extensions that default to installing the preview version may, due to this change, default to installing the stable version instead.

Yes, that's the **breaking** part of this pr. For now we added a pre-announcement for it.  Warning log as below should have been displayed to users since like a year ago.
image

src/azure-cli-core/azure/cli/core/extension/_resolve.py
```diff
@@ -104,6 +104,12 @@ def resolve_from_index(extension_name, cur_version=None, index_url=None, target_
     if not candidates:
         raise NoExtensionCandidatesError(f"No extension found with name '{extension_name}'")
 
+    if not allow_preview:
+        stable_candidates = list(filter(is_stable_from_metadata, candidates))
+        if len(stable_candidates) == 0:
+            logger.warning("No stable version of '%s' to install. Preview versions allowed", extension_name)
+            allow_preview = True
+
```
bebound, line 112: Two questions:
1. The following logic can be removed?
https://github.com/Azure/azure-cli/blob/7432856f7f16aec72abc9985558c5e7972ca2f40/src/azure-cli-core/azure/cli/core/extension/_resolve.py#L124-L130
2. If user has pass `--allow-preview false`, we still set it to True. It's kind of weird. A better solution is set default value to None, if the value is None, we can set it to True. If it's False, we keep it.

AllyW: Yes I agree the fallback logic here for those extension modules who do not have any stable versions is a little bit odd. 

Previously I'm thinking add a new config entry for this behavior so that, if needed, this behavior can be ignored. But that seems adding details on it.

However, what we want for all extension modules, is installing stable versions by default, not allow preview versions. So setting `allow-preview` to be `None` and resetting it to be `True` for `None` is not what we designed it for.

src/azure-cli-core/azure/cli/core/extension/dynamic_install.py
```diff
@@ -110,10 +110,9 @@ def _get_extension_run_after_dynamic_install_config(cli_ctx):
...
 
 def _get_extension_allow_preview_install_config(cli_ctx):
-    default_value = True
+    default_value = False
     if cli_ctx and cli_ctx.config.get('extension', 'dynamic_install_allow_preview', None) is None:
-        logger.warning("Preview version of extension is enabled by default for extension installation now. "
-                       "Will be disabled in future release. ")
+        logger.warning("Preview version of extension is disabled by default for extension installation. ")
         logger.warning("Please run 'az config set extension.dynamic_install_allow_preview=true or false' "
                        "to config it specifically. ")
```
evelyn-ys, line 117: Better to mention this config in `allow_preview` argument help msg as well

bebound:
`Set default --allow-preview value from True to be False` does not explain the real logic, I suggest using `install stable version by default unless there is no stable version is available`.

src/azure-cli-core/azure/cli/core/extension/dynamic_install.py
AllyW: `extension.dynamic_install_allow_preview` from config is designed for dynamic install.
`allow-preview` works specifically under `az extension add/update` and `az upgrade`.

bebound (approved):

evelyn-ys (approved):

kairu-ms (approved):

PR #30225, opened 2024-10-30
[Packaging] Drop Python 3.8 support

bebound:
Drop Python 3.8 because it was end of support on 2024-10-07: https://devguide.python.org/versions/

Drop 3.7 PR: https://github.com/Azure/azure-cli/pull/26855

### DIFF
--- a/doc/install_linux_prerequisites.md
+++ b/doc/install_linux_prerequisites.md
@@ -6,11 +6,11 @@ Some native Linux packages are required when installing the CLI with:
 - Interactive installation script
 - `pip`
 
-Current supported Python versions are Python 3.8 ~ 3.10.
+Current supported Python versions are Python 3.9 ~ 3.12.
 
 The commands to run to install the dependencies for some common distributions are listed below.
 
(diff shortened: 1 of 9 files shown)


yonzhan:

doc/install_linux_prerequisites.md
```diff
@@ -20,13 +20,3 @@ Install the latest Python 3.9 available in the software repo.
...
-
-### SUSE OpenSUSE 13.2
-Install Python 3.8+ if needed.
-```
-sudo zypper refresh && sudo zypper --non-interactive install gcc libffi-devel python-devel openssl-devel
-```
-
-### Flatcar
-
-Python is installed in the Azure-specific distribution of Flatcar, but is installed into the non-standard location `/usr/share/oem/python/bin/python`.
```
bebound, line 32: We've dropped SUSE， ref: https://learn.microsoft.com/en-us/cli/azure/install-azure-cli-linux?branch=live&pivots=zypper#before-you-begin

We don't create package for flatcar, I don't see a reason to mention it here.

PS: This doc is about install CLI with `pip`, which is not recommended. I'm considering removing it.

jiasli (approved):

PR #30376, opened 2024-11-18
[AppConfig] `appconfig feature list/show and appconfig kv import/export`: Support microsoft feature management schema

ChristineWanjau:
**Related command**

- appconfig kv import
- appconfig kv export
- appconfig feature show
- appconfig feature list

**Description**
This PR updates the import and export of features to [support new feature management schema](https://github.com/microsoft/FeatureManagement/blob/main/Schema/FeatureManagement.v2.0.0.schema.json).
The currently supported schema is the [Feature Managment Dotnet Schema](https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/schemas/FeatureManagement.Dotnet.v1.0.0.schema.json).

This PR also adds the use of a new environment variable called [AZURE_APPCONFIG_COMPATIBILITY_MODE] when exporting to a file for backward compatibility for users. This ensures our systems can process data in both old and new formats, avoiding disruptions for users of the previous version and maintaining functionality during the transition.

Default compatibility mode to true.
If compatibility mode true:
1.	Export classic feature flags using the legacy schema (Dotnet schema currently being used)
2.	Export variants feature flags using the new feature management schema
3.	If both variants and classic feature flags exist in the export, have both schemas exist in the file:
Example,
{
    "FeatureManagement":{
         // old schema
    }
    "feature_management":{
         "feature_flags": [
             // new schema
          ]
    }
}
If compatibility mode is false:
1.	All feature flags will be exported using the new feature management schema
    "feature_management":{
         "feature_flags": [
             // new schema
          ]
    }

[Link to design doc](https://microsoft.sharepoint.com/:w:/t/azconfig/EWyBrhokfbxAnTxQLqhSHhIBLaRr1vgILAaQEF2Q9emtLg?e=6Qt3kx)

**Testing Guide**

**History Notes**

[App Config] `az appconfig kv import`: Support microsoft feature management schema
[App Config] `az appconfig kv export`: Support microsoft feature management schema
[App Config] `az appconfig kv export`: Introduce a new environment variable called `AZURE_APPCONFIG_COMPATIBILITY_MODE` when exporting to a file for backward compatibility for users
[App Config] `az appconfig feature show`: Support microsoft feature management schema
[App Config] `az appconfig feature list`: Support microsoft feature management schema

---


yonzhan:
AppConfig

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -11,6 +11,27 @@
...
+from ._kv_helpers import (
+    __convert_featureflag_list_to_keyvalue_list,
+    __discard_features_from_retrieved_kv,
+    __read_kv_from_config_store,
+    __write_kv_and_features_to_config_store,
+    __print_restore_preview
+)
+from ._kv_import_helpers import (
+    __import_kvset_from_file,
+    __delete_configuration_setting_from_config_store,
```
albertofori, line 30: Should this be part of import helpers?

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -33,13 +54,7 @@
...
 from ._utils import get_appconfig_data_client, prep_label_filter_for_url_encoding, resolve_store_metadata, get_store_endpoint_from_connection_string, is_json_content_type
 
-from ._kv_helpers import (__read_kv_from_file, __read_features_from_file,
-                          __write_kv_and_features_to_file, __read_kv_from_config_store,
-                          __write_kv_and_features_to_config_store,
-                          __discard_features_from_retrieved_kv, __read_kv_from_app_service,
-                          __write_kv_to_app_service, __print_restore_preview,
-                          __convert_featureflag_list_to_keyvalue_list, __export_kvset_to_file,
-                          __import_kvset_from_file, __delete_configuration_setting_from_config_store, __map_to_appservice_config_reference)
+
```
albertofori, line 57: nit: extra line

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -50,36 +50,71 @@ class FeatureFlagValue:
...
-
-    def __init__(self,
-                 id_,
-                 description=None,
-                 enabled=None,
-                 conditions=None,
-                 display_name=None):
+    :ivar FeatureAllocation allocation:
+        Determines how variants should be allocated for the feature to various users.
+    :ivar list FeatureVariant[] variants:
```
albertofori, line 55: Is this to signify a list of lists?

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -50,36 +50,71 @@ class FeatureFlagValue:
...
+            ),
         }
 
+        if hasattr(self, FeatureFlagConstants.DISPLAY_NAME):
+            featureflagvalue[FeatureFlagConstants.DISPLAY_NAME] = self.display_name
+
+        if hasattr(self, FeatureFlagConstants.ALLOCATION):
+            featureflagvalue[FeatureFlagConstants.ALLOCATION] = (
+                custom_serialize_allocation(self.allocation)
+            )
```
albertofori, line 103: Is there a reason for the extra brackets?

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -50,36 +50,71 @@ class FeatureFlagValue:
...
+        telemetry=None,
+    ):
         default_conditions = {FeatureFlagConstants.CLIENT_FILTERS: []}
 
         self.id = id_
         self.description = "" if description is None else description
         self.enabled = enabled if enabled else False
         self.conditions = conditions if conditions else default_conditions
-        self.display_name = display_name
+        if allocation is not None:
```
albertofori, line 78: Is there a reason why we do not just set these properties to `None`?

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -129,112 +186,586 @@ def __repr__(self):
...
+            )
+
+        return json.dumps(featurevariant, indent=2, ensure_ascii=False)
+
+
+class FeatureUserAllocation:
+    """
+    Feature user allocation class.
+
+    :ivar str Variant:
```
albertofori, line 302:
```suggestion
    :ivar str variant:
```

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -129,112 +186,586 @@ def __repr__(self):
...
+        The name of the variant to use if the user allocation matches the current user.
+    :ivar list Users:
```
albertofori, line 304:
```suggestion
    :ivar list users:
```

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -129,112 +186,586 @@ def __repr__(self):
...
+        }
+
+        return json.dumps(featureallocationuser, indent=2, ensure_ascii=False)
+
+
+class FeatureGroupAllocation:
+    """
+    Feature group allocation class.
+
+    :ivar str Variant:
```
albertofori, line 347:
```suggestion
    :ivar str variant:
```

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -129,112 +186,586 @@ def __repr__(self):
...
+        The name of the variant to use
+        if the group allocation matches a group the current user is in.
+    :ivar list Groups:
```
albertofori, line 350:
```suggestion
    :ivar list groups:
```

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -129,112 +186,586 @@ def __repr__(self):
...
+        if not variant or not users:
+            raise ValidationError(
+                "User variant allocation must contain required '%s' and '%s' attributes: \n%s"
+                % (FeatureFlagConstants.VARIANT, FeatureFlagConstants.USERS, json.dumps(user_allocation, indent=2, ensure_ascii=False))
+            )
+
+        return cls(variant=variant, users=users)
+
+    def __repr__(self):
+        featureallocationuser = {
```
albertofori, line 335: Can we use snake case here?

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -288,12 +842,16 @@ def map_keyvalue_to_featureflagvalue(keyvalue):
             FeatureFlagConstants.ID,
             FeatureFlagConstants.DESCRIPTION,
             FeatureFlagConstants.ENABLED,
-            FeatureFlagConstants.CONDITIONS}
+            FeatureFlagConstants.CONDITIONS,
+        }
         if valid_fields != feature_flag_dict.keys():
```
albertofori, line 847: I would advise we change this to rather ensure that all the fields are in the feature flag dictionary and not that the two are exactly equal. We could break CLI if user adds a field we consider unknown.

How about we find the intersection of the expected and received fields and confirm that this is equal to the expected fields?

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
ChristineWanjau: the delete function is only being using in` kv import ` . I have separated the files depending on where its used. `kv_import_helpers `are functions used in kv import, `kv_export_helpers `are functions used in export and `kv_helpers `are functions used in multiple areas in kv import, kv export and etc.

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: My thinking was not to add new properties to a flag that doesn't have the new properties. If we set them to None the feature flag value would have these properties set to null in its value e.g `{id : "test", description" : "this is a description", ... allocation: null, telemetry: null }`

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: Updated

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: Updated

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: Updated

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: Updated

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: Updated

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: No need for this. Removed

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: No should be a list. This has been updated

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
ChristineWanjau: Updated

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
albertofori: In my opinion, we should be clear about what properties a class has, else we might end up with `AttributeError` when using the object in other parts of the code without adding a `hasattr` check. 

If anything at all, we can write custom serialization logic that will be passed to the `default` kwarg of `json.dumps` to avoid having `none` for certain properties. Thoughts?

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -306,60 +845,142 @@ def map_keyvalue_to_featureflagvalue(keyvalue):
...
-                    params = client_filter.get(FeatureFlagConstants.FILTER_PARAMETERS, {})
+                    params = lowercase_filter.get(
+                        FeatureFlagConstants.FILTER_PARAMETERS, {}
+                    )
                     client_filters_list.append(FeatureFilter(name, params))
                 else:
-                    logger.warning("Ignoring this filter without the %s attribute:\n%s",
-                                   FeatureFlagConstants.FILTER_NAME,
-                                   json.dumps(client_filter, indent=2, ensure_ascii=False))
+                    raise ValidationError(
```
albertofori, line 856: Just to confirm, we are breaking on an invalid filter instead of ignoring the filter, right?

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -306,60 +845,142 @@ def map_keyvalue_to_featureflagvalue(keyvalue):
             for client_filter in client_filters:
                 # If there is a filter, it should always have a name
                 # In case it doesn't, ignore this filter
-                name = client_filter.get(FeatureFlagConstants.FILTER_NAME)
+                lowercase_filter = {k.lower(): v for k, v in client_filter.items()}
+                name = lowercase_filter.get(FeatureFlagConstants.NAME)
```
albertofori, line 849: nit: Can we have explicit default values here like in other places?

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -306,60 +845,142 @@ def map_keyvalue_to_featureflagvalue(keyvalue):
...
+                ):
+                    raise ValidationError(
+                        f"Feature '{feature_name}' must have an any/all requirement type."
+                    )
+                conditions[FeatureFlagConstants.REQUIREMENT_TYPE] = requirement_type
+
+        # Allocation
+        allocation = feature_flag_dict.get(FeatureFlagConstants.ALLOCATION, None)
+        if allocation:
+            feature_flag_dict[FeatureFlagConstants.ALLOCATION] = (
```
albertofori, line 879: nit: extra brackets

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -306,60 +845,142 @@ def map_keyvalue_to_featureflagvalue(keyvalue):
...
+                    json.dumps(allocation, ensure_ascii=False)
+                )
+            )
+
+        # Variants
+        variants = feature_flag_dict.get(FeatureFlagConstants.VARIANTS, None)
+        variant_list = []
+        if variants:
+            for variant in variants:
+                feature_variant = FeatureVariant.convert_from_json(
```
albertofori, line 890: Do we use this method anywhere outside `featuremodels`?

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -129,112 +183,570 @@ def __repr__(self):
...
+    :ivar list users:
+        Collection of users where if any match the current user, the variant specified in the user allocation is used.
+    """
+
+    def __init__(self, variant, users):
+        self.variant = variant
+        self.users = users
+
+    @classmethod
+    def convert_from_json(cls, json_string):
```
albertofori, line 306: Correct me if I am wrong but it seems  in places where we are using `convert_from_json`, we are always converting some dictionary to json string first, and then we parse the string again in this method. Can these methods directly accept a dict?

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -129,112 +183,570 @@ def __repr__(self):
...
+
+    :ivar str variant:
+        The name of the variant to use if the calculated percentile for the current user falls in the provided range.
+    :ivar number from:
+        The lower end of the percentage range for which this variant will be used.
+    :ivar number to:
+        The upper end of the percentage range for which this variant will be used.
+    """
+
+    def __init__(self, variant, From, to):
```
albertofori, line 395:
```suggestion
    def __init__(self, variant, from_, to):
```

I would suggest adding a trailing underscore like we do with id [here](https://github.com/Azure/azure-cli/blob/a3dc72bc3bd881a57cd856e124362ac6e99178da/src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py#L64)

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -129,112 +183,570 @@ def __repr__(self):
...
+        allocation_user = allocation_json.get(FeatureFlagConstants.USER, None)
+
+        # Convert all users to FeatureUserAllocation object
+        if allocation_user:
+            allocation_user_list = []
+            for user in allocation_user:
+                feature_user_allocation = FeatureUserAllocation.convert_from_json(
+                    json.dumps(user, ensure_ascii=False)
+                )
+                if feature_user_allocation:
```
albertofori, line 520: can `convert_from_json` return `None`?

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -129,112 +183,570 @@ def __repr__(self):
...
+        Determines how variants should be allocated for the feature to various users based on percentile.
+    """
+
+    def __init__(
+        self,
+        user=None,
+        group=None,
+        percentile=None,
+        default_when_enabled=None,
+        default_when_disabled=None,
```
albertofori, line 474: Should we add the `seed` property for allocation?

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -129,112 +183,570 @@ def __repr__(self):
...
+            content_type=FeatureFlagConstants.FEATURE_FLAG_CONTENT_TYPE,
+            tags={},
+        )
 
         set_kv.locked = featureflag.locked
         set_kv.last_modified = featureflag.last_modified
 
     except ValueError as exception:
-        error_msg = "Exception while converting feature flag to key value: {0}\n{1}".format(featureflag.key, exception)
+        error_msg = (
```
albertofori, line 720: nit: extra brackets

ChristineWanjau: Removed. I formatted the code using Vs code it seems it added the extra brackets. Will remove all extra brackets.

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: Yes we should. Updated

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: No we check if the property is null first before calling the function

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: Makes sense Updated

src/azure-cli/azure/cli/command_modules/appconfig/_kv_export_helpers.py
```diff
@@ -0,0 +1,441 @@
...
+    naming_convention="pascal",
+):
+    if not key_values and not features:
+        logger.warning("\nSource configuration is empty. No changes will be made.")
+        return
+
+    try:
+        exported_keyvalues = __export_keyvalues(key_values, format_, separator, None)
+        if features and not skip_features:
+            compatibility_mode = os.environ.get(
```
avanigupta, line 114: Did we consider using `az configure --defaults` instead of environment variable for this?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_export_helpers.py
```diff
@@ -0,0 +1,441 @@
...
+                "AZURE_APPCONFIG_FM_COMPATIBILITY_MODE", True
```
avanigupta, line 115: Suggest `AZURE_APPCONFIG_FM_COMPATIBLE` since the value is true/false.

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: We should still be ignoring the filter because if throwing on an invalid filter would be a breaking change

src/azure-cli/azure/cli/command_modules/appconfig/_kv_export_helpers.py
ChristineWanjau: Yes, we saw it was better to have the `compatibility mode` as an environment variable instead of an argument because its  much less "sticky" we could easily remove it compared to having an argument once users move to using the new fm schema completely.

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: Updated to always accept dict

src/azure-cli/azure/cli/command_modules/appconfig/_kv_export_helpers.py
avanigupta: I dont think `az configure --defaults` can be considered an argument. But I agree that its a more permanent solution than the short term environment variable option.

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
albertofori: So `if feature_user_allocation` should not be needed, right?

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: Added default values

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: Yes you are right its not needed.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_export_helpers.py
```diff
@@ -0,0 +1,441 @@
...
+            exported_dict = {
+                new_ms_featuremanagement_keyword: {feature_flags_keyword: []}
+            }
+            # retrieved_features is a list of FeatureFlag objects
+            for featureflag in retrieved_features:
+                feature_entry = __export_feature_to_new_ms_schema(featureflag)
+                exported_dict[new_ms_featuremanagement_keyword][
+                    feature_flags_keyword
+                ].append(feature_entry)
+
```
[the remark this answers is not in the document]
ChristineWanjau: I would prefer to keep it separate because we would get different kind of schemas when `compatibility_mode `is set to True. For example if the naming convention is underscore then the legacy schema would be nested within the new schema.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_export_helpers.py
```diff
@@ -0,0 +1,441 @@
...
+        raise CLIError("Fail to export key-values." + str(exception))
+
+
+def __export_features(retrieved_features, naming_convention, compatibility_mode):
+    new_ms_featuremanagement_keyword = (
+        FeatureManagementReservedKeywords.UNDERSCORE.feature_management
+    )
+    feature_flags_keyword = FeatureFlagConstants.FEATURE_FLAGS_KEY
+    try:
+        compatibility_mode = str(compatibility_mode).lower() in ["true"]
```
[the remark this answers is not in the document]
ChristineWanjau: No reason. This has been updated to use equality comparison

src/azure-cli/azure/cli/command_modules/appconfig/_kv_export_helpers.py
```diff
@@ -0,0 +1,441 @@
...
+
+    try:
+        exported_keyvalues = __export_keyvalues(key_values, format_, separator, None)
+        if features and not skip_features:
+            compatibility_mode = os.environ.get(
+                "AZURE_APPCONFIG_FM_COMPATIBLE", True
+            )
+            exported_features = __export_features(
+                features, naming_convention, compatibility_mode
+            )
```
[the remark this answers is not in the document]
ChristineWanjau: You're right. This has been updated to evaluate in `__export_features`.

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -129,86 +167,502 @@ def __repr__(self):
...
+        variants_list - List of FeatureVariant objects
+
+    Return:
+        JSON serializable List
+    """
+    featurevariants = []
+    for variant in variants_list:
+        variant_dict = {}
+        variant_dict[FeatureFlagConstants.NAME] = variant.name
+        if variant.configuration_value:
```
avanigupta, line 621: configuration_value can be false, empty string or null, and all of these are valid values that should be added to the feature flag. This `if` condition will block such values.

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -94,23 +114,37 @@ class FeatureFlag:
...
-                 description=None,
-                 conditions=None,
-                 locked=None,
-                 last_modified=None):
+        Dictionary that contains client_filters List
+    :ivar dict allocation:
+        Determines how variants should be allocated for the feature to various users.
+    :ivar list FeatureVariant variants:
+        The list of variants defined for this feature.
+    :ivar dict {string, bool} telemetry:
```
avanigupta, line 129: telemetry is dictionary of string and object (enabled property is bool, and metadata property is a dictionary of string key and string value)

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -129,86 +167,502 @@ def __repr__(self):
...
+            percentile.__dict__ for percentile in allocation.percentile
+        ]
+
+    if allocation.default_when_enabled:
+        featureallocationdict[FeatureFlagConstants.DEFAULT_WHEN_ENABLED] = allocation.default_when_enabled
+
+    if allocation.default_when_disabled:
+        featureallocationdict[FeatureFlagConstants.DEFAULT_WHEN_DISABLED] = allocation.default_when_disabled
+
+    return featureallocationdict
```
avanigupta, line 604: Do we need to serialize seed here?

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -129,86 +167,502 @@ def __repr__(self):
...
+            allocation=featureflag.allocation,
+            variants=featureflag.variants,
+            telemetry=featureflag.telemetry,
+        )
+
+        set_kv = KeyValue(
+            key=featureflag.key,
+            label=featureflag.label,
+            value=json.dumps(
+                feature_flag_value, default=lambda o: {k: v for k, v in o.__dict__.items() if v is not None}, ensure_ascii=False
```
avanigupta, line 661: Why did we change the lambda here?

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -218,23 +672,28 @@ def map_featureflag_to_keyvalue(featureflag):
...
-        Helper Function to convert KeyValue object to FeatureFlag object for display
+    """
+    Helper Function to convert KeyValue object to FeatureFlag object for display
 
-        Args:
-            keyvalue - KeyValue object to be converted
-            show_conditions - Boolean for controlling whether we want to display "Conditions" or not
+    Args:
+        keyvalue - KeyValue object to be converted
+        show_conditions - Boolean for controlling whether we want to display "Conditions" or not
```
avanigupta, line 691: Does the show_conditions flag control variants/allocation/telemetry too?

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -249,14 +708,20 @@ def map_keyvalue_to_featureflag(keyvalue, show_conditions=True):
     if filters and state == FeatureState.ON:
         state = FeatureState.CONDITIONAL
```
avanigupta, line 709: What is our strategy for handling "conditional" state? Is it still controlled by the presence of conditions only? What if the state is enabled but it has variants?

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -249,14 +708,20 @@ def map_keyvalue_to_featureflag(keyvalue, show_conditions=True):
...
-    feature_flag = FeatureFlag(feature_name,
-                               keyvalue.key,
-                               keyvalue.label,
-                               state,
-                               feature_flag_value.description,
-                               conditions,
-                               keyvalue.locked,
-                               keyvalue.last_modified)
+    feature_flag = FeatureFlag(
+        name=feature_name,
```
avanigupta, line 712: Should we rename this 'name' field to 'id'?

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -288,12 +753,16 @@ def map_keyvalue_to_featureflagvalue(keyvalue):
             FeatureFlagConstants.ID,
             FeatureFlagConstants.DESCRIPTION,
             FeatureFlagConstants.ENABLED,
-            FeatureFlagConstants.CONDITIONS}
+            FeatureFlagConstants.CONDITIONS,
```
avanigupta, line 756: We need to add all the new fields to `valid_fields` dict

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/import_multiple_feature_sections.json
```diff
@@ -45,4 +45,4 @@
       ]
     }
   }
-  }
```
avanigupta, line 48: Can you revert the file? It only differs in whitespace

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
avanigupta: Why did our tests not catch this with the new code?

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/test_appconfig_commands.py
```diff
@@ -951,6 +951,7 @@ def test_azconfig_import_export(self, resource_group, location):
         import_features_alt_syntax_file_path = os.path.join(TEST_DIR, 'import_features_alt_syntax.json')
         import_features_random_conditions_file_path = os.path.join(TEST_DIR, 'import_features_random_conditions.json')
         import_features_invalid_requirement_type_file_path = os.path.join(TEST_DIR, 'import_features_invalid_requirement_type.json')
+        os.environ['AZURE_APPCONFIG_FM_COMPATIBLE'] = 'True'
```
avanigupta, line 954: If the test suite is running tests in parallel, environment variables may get mixed up between tests. When setting up tests that depend on environment variables, we need to make sure that those tests never run in parallel. Is that something we need to do for CLI tests?

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: I was wondering whether to add the new fields there because they aren't required fields. The message logged says `            logger.debug(
                "'%s' feature flag is missing required values or it contains ",
                keyvalue.key +
                "unsupported values. Setting missing value to defaults and ignoring unsupported values\n",
            ) ` . If a flag that doesn't have any of the new fields, it will log this message although the fields aren't required. Do we want to show this message in such a situation?

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: Yes we could I see no issue with that. Although this is what we had before I am wondering if there was a reason why we had it as `Name`

ChristineWanjau: This is the FeatureFlag representation that is displayed to the user. The other FeatureFlagValue class has it as `id`

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: Currently we just have it as it was if there are filters and state of the feature flag is enabled then the state is conditional. Should we have it as conditional if there are variants?

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: Currently it just controls the conditions. I am wondering if we should hide the other new fields. If we want to hide the new fields maybe we should rename the argument to something else other than `show_conditions`?

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: My thinking was not to add new properties to a flag that doesn't have the new properties. If we set them to None the feature flag value would have these properties set to null in its value e.g {id : "test", description" : "this is a description", ... allocation: null, telemetry: null } so we added the lambda to remove the null values in the feature flag object.

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/import_multiple_feature_sections.json
ChristineWanjau: Updated

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: You are right. Updated

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -129,86 +167,504 @@ def __repr__(self):
             "Locked": self.locked,
             "Description": self.description,
             "Last Modified": self.last_modified,
-            "Conditions": custom_serialize_conditions(self.conditions)
+            "Conditions": custom_serialize_conditions(self.conditions),
+            "Allocation": custom_serialize_allocation(self.allocation),
+            "Variants": custom_serialize_variants(self.variants),
+            "Telemetry": self.telemetry,
```
avanigupta, line 173: Does telemetry need custom serialization?

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -129,86 +167,504 @@ def __repr__(self):
...
+        default_when_disabled = allocation_dict.get(
+            FeatureFlagConstants.DEFAULT_WHEN_DISABLED, None
+        )
+        default_when_enabled = allocation_dict.get(
+            FeatureFlagConstants.DEFAULT_WHEN_ENABLED, None
+        )
+
+        allocation = cls(
+            default_when_enabled=default_when_enabled,
+            default_when_disabled=default_when_disabled,
```
avanigupta, line 492: I dont see the seed value being assigned here

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
avanigupta: Does that mean we always add default value for conditions since we initialize it to `{client_filters: []}` ? Even when feature flag contains only variants?

avanigupta: I think we should remove the default conditions value while writing new feature flags. While reading, its okay to support reading the default value for backwards compatibility.

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
avanigupta: Yes, we should hide the details of other fields too. Is this user configurable? Maybe we can use something like `show_all_details`

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
avanigupta: ok it makes sense if this is for display. lets keep using "name"

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: The concept of `state `seems to only exist in the CLI and its only used for display. It makes sense to me to have it consistent with portal as well.

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
avanigupta: We need to update the condition `if valid_fields != feature_flag_dict.keys()`

With the old syntax, feature flag had to have only these required fields. Nothing was optional and nothing extra was supported.
With the new syntax, we know what fields we can support, and we also know what fields are required. So we need to modify the condition and error message to reflect that. Maybe even a boolean flag to track if the current feature flag is v1 or v2?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
```diff
@@ -0,0 +1,958 @@
...
+        )
+        if ms_feature_management_keyword in config_data:
+            if found_dotnet_schema and any(
+                key != FeatureFlagConstants.FEATURE_FLAGS_KEY
+                for key in config_data[ms_feature_management_keyword].keys()
+            ):
+                raise FileOperationError(
+                    "Data contains an already defined section with the key %s."
+                    % (dotnet_schema_keyword)
+                )
```
avanigupta, line 171: If I understood correctly, this condition ensures that if dotnet schema exists in file AND "feature_management" section is also present, that section cannot contain any key other than "feature_flags".
Why don't we just ignore the sections that are not "feature_flags"?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
```diff
@@ -0,0 +1,958 @@
...
+            if (
+                FeatureFlagConstants.FEATURE_FLAGS_KEY
+                in config_data[ms_feature_management_keyword]
+            ):
+                found_ms_fm_schema = True
+
+            if any(
+                key != FeatureFlagConstants.FEATURE_FLAGS_KEY
+                for key in config_data[ms_feature_management_keyword].keys()
+            ):
```
avanigupta, line 182: The same check is repeated here. Can you move it to a common local variable?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
```diff
@@ -0,0 +1,958 @@
...
+def __read_features_from_msfm_schema(feature_flags_list):
+    feature_flags = []
+    for feature in feature_flags_list:
+        if validate_import_feature(feature[FeatureFlagConstants.ID]):
+            feature_id = feature[FeatureFlagConstants.ID]
+
+            new_feature = FeatureFlagValue(
+                id_=str(feature.get(FeatureFlagConstants.ID)),
+                enabled=_convert_boolean_value(
+                    feature.get(FeatureFlagConstants.ENABLED, None), feature_id
```
avanigupta, line 305: if "enabled" is not present, default to false

avanigupta: That's true for all boolean properties in FM schema

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
```diff
@@ -0,0 +1,958 @@
...
+
+            if allocation := feature.get(FeatureFlagConstants.ALLOCATION, None):
+                new_feature.allocation = FeatureAllocation.convert_from_dict(allocation)
+
+            if variants := feature.get(FeatureFlagConstants.VARIANTS, None):
+                new_feature.variants = []
+                for variant in variants:
+                    new_variant = FeatureVariant.convert_from_dict(variant)
+                    if new_variant:
+                        new_feature.variants.append(new_variant)
```
avanigupta, line 357:
```suggestion
                for variant in variants:
                    new_feature.variants.append(FeatureVariant.convert_from_dict(variant))
```

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
```diff
@@ -0,0 +1,958 @@
...
+                            raise ValidationError(
+                                "Feature '{0}' must have an any/all requirement type. \n".format(
+                                    str(k)
+                                )
+                            )
+                        feature_flag_value.conditions[
+                            FeatureFlagConstants.REQUIREMENT_TYPE
+                        ] = condition_value
+                    else:
+                        feature_flag_value.conditions[condition] = condition_value
```
avanigupta, line 404: We dont support any fields other than "EnabledFor" and "RequirementType". So this should either be an error or be ignored

ChristineWanjau: It looks like we support random conditions from this test https://github.com/Azure/azure-cli/blob/1bc8703035aea9d51f254dd17ca31956a55a84cb/src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/test_appconfig_commands.py#L1057

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
ChristineWanjau: Yes if dotnet schema exists and "feature_management" section is also present and it contains other keys other than "feature_flags" this means that there is another dotnet schema already defined within the "feature_management" section. The idea is we don't want the dotnet schema defined twice so we throw because it's an invalid schema.

ChristineWanjau: We only support these scenarios:
```
featureManagement: {
//dotnet schema
},
feature_management: {
    feature_flags: [   //new schema]
}
```

```
FeatureManagement: {
//dotnet schema
},
feature_management: {
    feature_flags: [   //new schema]
}
```
```
feature-management: {
//dotnet schema
},
feature_management: {
    feature_flags: [   //new schema]
}
```

```
feature_management: {
    feature_flags: [
   //new schema
    ],
  //dotnet schema
}
```

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/test_appconfig_commands.py
ChristineWanjau: This is a good point. I have taken a look at how the tests run, it seems the tests run sequentially. The tests are grouped into classes and within the class we have functions with the tests. When a test within the function fails the class is marked as failed. It looks like we only queue the next function in a given class after the current running test in the class has passed/has been skipped. 

Discussed with @albertofori as well. We think we can at least group tests that depend on the same value of the environment variable together, please let me know what you think.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
```diff
@@ -0,0 +1,958 @@
...
+    """
+    Convert the value to a boolean if it is a string.
+
+    :param Union[str, bool] enabled: Value to be converted.
+    :return: Converted value.
+    :rtype: bool
+    """
+    if isinstance(enabled, bool):
+        return enabled
+    if enabled.lower() == "true":
```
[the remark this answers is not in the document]
ChristineWanjau: I don't think it will throw, although if we set enabled to a default value False. I think we should be good.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
ChristineWanjau: Updated

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
ChristineWanjau: Set default to False

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
```diff
@@ -0,0 +1,958 @@
...
+                    ] = requirement_type
+
+            if allocation := feature.get(FeatureFlagConstants.ALLOCATION, None):
+                new_feature.allocation = FeatureAllocation.convert_from_dict(allocation)
+
+            if variants := feature.get(FeatureFlagConstants.VARIANTS, None):
+                new_feature.variants = []
+                for variant in variants:
+                    new_variant = FeatureVariant.convert_from_dict(variant)
+                    if new_variant:
```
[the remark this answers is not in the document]
ChristineWanjau: I don't think it will be None. Updated

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
ChristineWanjau: Updated

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: Yes we will always add default conditions when writing new flags. I am wondering if this would be a breaking change if we remove adding default conditions to the flags, maybe there are users who already expect this behavior.

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: Should be assigned. Updated

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: Yes Updated

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
avanigupta: For V1, we should add default conditions. But for V2, we can avoid all defaults. There's no need to have empty conditions defined for a variant. Similarly, no need for empty variants or allocation or telemetry when conditions are specified. While reading, we can be lenient and parse default empty values. But during writing, we should be stricter and not write null or empty values.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
avanigupta: I see. So this condition is checking the last scenario from your example. Thanks!

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
avanigupta: Do we ignore or populate those random fields? I dont see the test file that contain expected output so I'm not sure which one it was.

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -129,86 +167,565 @@ def __repr__(self):
...
+
+    Return:
+        JSON serializable dictionary
+
+    """
+
+    feature_telemetry_dict = {}
+
+    if telemetry.enabled:
+        feature_telemetry_dict[FeatureFlagConstants.TELEMETRY] = telemetry.enabled
```
albertofori, line 686: telemetry.enabled is boolean, we will not add the enabled property if it is false. Is this intended?

albertofori, line 686:
```suggestion
        feature_telemetry_dict[FeatureFlagConstants.ENABLED] = telemetry.enabled
```

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -129,86 +167,565 @@ def __repr__(self):
...
+        telemetry - FeatureTelemetry object
+
+    Return:
+        JSON serializable dictionary
+
+    """
+
+    feature_telemetry_dict = {}
+
+    if telemetry.enabled:
```
albertofori, line 685: `enabled` is a boolean so if it is false, we might end up not writing the property at all. Is this intended?

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -129,86 +167,565 @@ def __repr__(self):
...
+        Return:
+            FeatureAllocation object
+        """
+        default_when_disabled = allocation_dict.get(
+            FeatureFlagConstants.DEFAULT_WHEN_DISABLED, None
+        )
+        default_when_enabled = allocation_dict.get(
+            FeatureFlagConstants.DEFAULT_WHEN_ENABLED, None
+        )
+
```
albertofori, line 489: Should we also read the seed here and pass it to the initialization of the allocation object?

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
albertofori: I agree with getting rid of the `state` property altogether in exchange for the boolean `enabled`. This will then eliminate any confusion that the `conditional` state currently causes.

avanigupta: This would technically be a breaking change if someone has a dependency on the output field "state". But I don't think we can avoid that. The current output makes no sense in context of variants. Maybe we can call it out in the release notes?

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
avanigupta: We generally avoid writing null/default values. But for telemetry, portal always adds telemetry property with enabled=false even though it has the same effect as not having the telemetry property at all. Was there a specific reason why we may want to add telemetry with enabled=false for every feature flag?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
ChristineWanjau: I think it just retain it as it is. In the test, we compare if the imported kvs equal to the exported kvs, meaning we should have retained the random condition. In the code it doesn't look like we ignore or populate the random fields we don't do any validation, we just validate client_filters. https://github.com/Azure/azure-cli/blob/1bc8703035aea9d51f254dd17ca31956a55a84cb/src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py#L741

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: I think calling it out in the release notes is a good idea. I would like to suggest having this in another PR and release it during the breaking change window.

avanigupta: When is the next breaking change release planned?
If we dont make the change here, I'm not sure what the alternative is for "State" property with variants.

ChristineWanjau: I think the breaking change would around May. I am thinking we could have the state as 
`"Variant-based"`. Still open to other suggestions. What do you think? Although what if a flag has both variants and filters, what should we show..

avanigupta: It is a bit confusing. 
Maybe we can keep the output with "State" if compatibility mode is set to true. If compatibility is false, it means users have explicitly opted for v2 schema and we can show a different output format. The new format can show the same fields we display in the portal. And once we make V2 schema default, we can fully adopt the new output.

If we agree on that, we can keep showing "Conditional" state whenever a flag is not explicitly enabled or disabled. So any combination of conditions/variants/allocation will make the flag conditional.

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
albertofori: Actually, there is no strong reason for this, but I was just looking at the situation where we might actually be removing the `enabled` property when telemetry is disabled. 

This is how it looks in the portal:
image

I am wondering if exporting, then re-importing this ff would remove the `enabled` property?

ChristineWanjau: I get your point. A user might have set the value as enabled false explicitly, but still the evaluation of the flag will still be the same as without the enabled=false property. I am also wondering when reading a flag with let's say:

```
{
	"id": "DisabledFeature",
	"description": "",
	"enabled": false,
        "conditions": {}
        "telemetry": {}
}
```

Do we remove conditions and telemetry or leave it as is?

avanigupta: With V1 so far, we have always written default conditions with empty `client_filters` list even if flag is explicitly disabled or enabled. But with V2, we need to rethink that design since we have many new fields and it doesn't make sense to keep empty default values for all of them. That decision is crucial while creating feature flags in order to maintain a consistent user experience across all our clients. 
For import/export, I think we can ignore null/empty fields. 

Discard `description`, `conditions` and `telemetry` in this situation:
```
{
	"id": "DisabledFeature",
	"description": "",
	"enabled": false,
        "conditions": {}
        "telemetry": {}
}
```

Discard `description` and `conditions`, but preserve `telemetry` in this situation:
```
{
	"id": "DisabledFeature",
	"description": "",
	"enabled": false,
        "conditions": {}
        "telemetry": {
             "enabled": false,
         }
}
```

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: Updated to add default values only for v1. For v2 we avoid setting defaults for conditions and other new properties introduced.

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: This has been updated.

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -129,86 +167,565 @@ def __repr__(self):
...
+                   FeatureFlagConstants.TO,
+                   json.dumps(percentile_allocation_dict, indent=2, ensure_ascii=False))
+            )
+
+        return cls(variant=variant, from_=percentile_from, to=perecentile_to)
+
+    def __repr__(self):
+        feature_percentile_allocation = {
+            FeatureFlagConstants.VARIANT: self.variant,
+            FeatureFlagConstants.FROM: self.From,
```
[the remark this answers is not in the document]
ChristineWanjau: This has been updated.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
```diff
@@ -0,0 +1,958 @@
...
+
+
+def __read_features_from_msfm_schema(feature_flags_list):
+    feature_flags = []
+    for feature in feature_flags_list:
+        if validate_import_feature(feature[FeatureFlagConstants.ID]):
+            feature_id = feature[FeatureFlagConstants.ID]
+
+            new_feature = FeatureFlagValue(
+                id_=str(feature.get(FeatureFlagConstants.ID)),
```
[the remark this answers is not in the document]
ChristineWanjau: Yes, this has been updated

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
```diff
@@ -0,0 +1,958 @@
...
+                config_data = javaproperties.load(config_file)
+                logger.debug(
+                    "Importing feature flags from a properties file is not supported. If properties file contains feature flags, they will be imported as regular key-values."
+                )
+
+    return config_data
+
+
+def __read_features_from_file(file_path, format_):
+    config_data = {}
```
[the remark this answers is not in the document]
ChristineWanjau: Yes we can remove this. Updated

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: This has also been updated to only show state when compatibility mode is true and show enabled if compatibility mode is false.

src/azure-cli/azure/cli/command_modules/appconfig/_diff_utils.py
```diff
@@ -153,15 +154,37 @@ def serialize(obj):
...
+            feature = map_keyvalue_to_featureflag(obj, hide_enabled=compatibility_mode)
             # name
             feature_json = {'feature': feature.name}
             # state
-            feature_json['state'] = feature.state
+            if hasattr(feature, 'state'):
+                feature_json['state'] = feature.state
+            # enabled
+            if hasattr(feature, 'enabled'):
+                feature_json['enabled'] = feature.enabled
```
avanigupta, line 172: When we display enabled property instead of state, we must also display the number of filters/variants and if telemetry is enabled.

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -224,142 +758,301 @@ def map_featureflag_to_keyvalue(featureflag):
...
+                "unsupported values. Unsupported values will be ignored.\n",
+            )
+
+        # check if feature flag is version v1 or v2.
+        # We always default values for v1 feature flags but avoid defaults for v2 feature flags.
+        if (
+            FeatureFlagConstants.ALLOCATION in feature_flag_dict or
+            FeatureFlagConstants.VARIANTS in feature_flag_dict or
+            FeatureFlagConstants.TELEMETRY in feature_flag_dict
+        ):
```
avanigupta, line 868: version can be v2 even if all these properties are none.

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -224,142 +758,301 @@ def map_featureflag_to_keyvalue(featureflag):
...
+                    FeatureFlagConstants.REQUIREMENT_TYPE_ANY,
+                ):
+                    raise ValidationError(
+                        f"Feature '{feature_name}' must have an any/all requirement type."
+                    )
+                conditions[FeatureFlagConstants.REQUIREMENT_TYPE] = requirement_type
+
+            # Backend returns conditions: {client_filters: None} for flags with no conditions.
+            # No need to write empty conditions to key-values.
+            if conditions.get(FeatureFlagConstants.CLIENT_FILTERS, None) is None and conditions.get(FeatureFlagConstants.REQUIREMENT_TYPE, None) is None:
```
avanigupta, line 922: you could simplify this:
```
if client_filters or requirement_type:
    feature_flag_dict[FeatureFlagConstants.CONDITIONS] = conditions
elif feature_flag_version == FeatureFlagVersion.V1:
     feature_flag_dict[FeatureFlagConstants.CONDITIONS] = default_conditions
```

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -939,6 +989,25 @@ def is_feature_flag(kv):
     return False
 
 
+def convert_boolean_value(enabled, feature_name):
```
avanigupta, line 1000: We can make this method generic for converting any boolean to string.
`def convert_string_to_bool(value, feature_name)`

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -224,142 +758,301 @@ def map_featureflag_to_keyvalue(featureflag):
...
-        if valid_fields != feature_flag_dict.keys():
-            logger.debug("'%s' feature flag is missing required values or it contains ", keyvalue.key +
-                         "unsupported values. Setting missing value to defaults and ignoring unsupported values\n")
+            FeatureFlagConstants.CONDITIONS,
+            FeatureFlagConstants.ALLOCATION,
+            FeatureFlagConstants.VARIANTS,
+            FeatureFlagConstants.TELEMETRY,
+            FeatureFlagConstants.DISPLAY_NAME
+        }
+        if valid_fields.union(feature_flag_dict.keys()) == valid_fields:
```
albertofori, line 855: Should this be `!=`?

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -90,125 +114,635 @@ class FeatureFlag:
...
+
+        Args:
+            dictionary - {string, Any}
+
+        Return:
+            FeatureTelemetry object
+        """
+        enabled = None
+        telemetry_enabled = feature_telemetry_dict.get(FeatureFlagConstants.ENABLED, None)
+        if telemetry_enabled is not None:
```
avanigupta, line 585:
```suggestion
        if telemetry_enabled:
```

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -90,125 +114,635 @@ class FeatureFlag:
...
+
+    Return:
+        JSON serializable Dictionary
+    """
     featurefilterdict = {}
 
     for key, value in conditions_dict.items():
         if key == FeatureFlagConstants.CLIENT_FILTERS:
-            featurefilterdict[key] = [feature_filter.__dict__ for feature_filter in value]
+            if value is not None:
```
avanigupta, line 616:
```suggestion
            if value:
```

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -224,142 +758,301 @@ def map_featureflag_to_keyvalue(featureflag):
...
+            FeatureFlagConstants.ALLOCATION in feature_flag_dict or
+            FeatureFlagConstants.VARIANTS in feature_flag_dict or
+            FeatureFlagConstants.TELEMETRY in feature_flag_dict
+        ):
+            feature_flag_version = FeatureFlagVersion.V2
+
+        feature_name = feature_flag_dict.get(FeatureFlagConstants.ID, "")
+        if feature_flag_version == FeatureFlagVersion.V1:
+            description = feature_flag_dict.get(FeatureFlagConstants.DESCRIPTION, "")  # assign empty string as default
+            feature_flag_dict[FeatureFlagConstants.DESCRIPTION] = description
```
albertofori, line 874:
```suggestion
            feature_flag_dict[FeatureFlagConstants.DESCRIPTION] = feature_flag_dict.get(FeatureFlagConstants.DESCRIPTION, "")  # assign empty string as default
```

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -224,142 +758,301 @@ def map_featureflag_to_keyvalue(featureflag):
...
+
+        # Allocation
+        allocation = feature_flag_dict.get(FeatureFlagConstants.ALLOCATION, None)
+        if allocation:
+            feature_flag_dict[FeatureFlagConstants.ALLOCATION] = FeatureAllocation.convert_from_dict(allocation)
+
+        # Variants
+        variants = feature_flag_dict.get(FeatureFlagConstants.VARIANTS, None)
+        variant_list = []
+        if variants:
```
albertofori, line 937:
```suggestion

        if variants:
            variant_list = []
```

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
```diff
@@ -224,142 +758,301 @@ def map_featureflag_to_keyvalue(featureflag):
...
+            for variant in variants:
+                feature_variant = FeatureVariant.convert_from_dict(variant)
+                if feature_variant:
```
albertofori, line 940: Is there a condition under which `convert_from_dict` can retun `None`?
Should we rather check if `variant` is not `None` first before trying to convert?

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: Yeah thats true. But I don't think there is another way to tell if a flag is v2 aside from checking the new properties.

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: `telemetry_enabled` could be `false` and we don't want to remove it if its explictly defined as False

src/azure-cli/azure/cli/command_modules/appconfig/_diff_utils.py
ChristineWanjau: Yes we currently display variants/filters and telemetry for both scenarios with state property and enabled property.

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: I was wondering a case where we have `conditions: {client_filters: [] } ` I was thinking we would want to retain this, if we just check it as suggested it might be removed.

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: Yes, we should. This has been updated

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: Updated

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: Updated

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: Yes, nice catch! Updated

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: Sure, this has been updated

src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py
ChristineWanjau: Makes sense. Updated

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
```diff
@@ -0,0 +1,933 @@
...
+    # pylint: disable=too-many-nested-blocks
+    key_values = []
+    feature_flags = []
+
+    try:
+        if dotnet_schema_keyword:
+            for keywordset in FeatureManagementReservedKeywords.ALL:
+                if keywordset.feature_management == dotnet_schema_keyword:
+                    feature_management_keywords = keywordset
+                    break
```
albertofori, line 218: Can we use `get_keywords` define here? https://github.com/ChristineWanjau/azure-cli/blob/d7f33f535ecf78553cd3d1cd5986d5d78924460b/src/azure-cli/azure/cli/command_modules/appconfig/_featuremodels.py#L1055-L1056

ChristineWanjau: I don't think there is a need to do that since the keywords are already defined in the keywordset.

albertofori (approved):

avanigupta (approved):

ChristineWanjau:
Hello @zhoxing-ms this PR is ready to be merged.

yanzhudd:
Since there are multiple changes in this PR, it is better to seperate the changelog into small ones and list in the History Notes section, e.g.,
[App Config] `az appconfig kv import`: Support microsoft feature management schema
...

yanzhudd (approved):

PR #30413, opened 2024-11-25
{CI} Fix Codegen Coverage

wangzelin007:
**Related command**

Due to the inconsistent versions of jinja2 required by aosm and alias, and the random order of extension installations, installing a lower version of jinja2 can result in the following error:

```
  File "/mnt/vss/_work/1/s/env/lib/python3.11/site-packages/jinja2/filters.py", line 13, in <module>
    from markupsafe import soft_unicode
ImportError: cannot import name 'soft_unicode' from 'markupsafe' (/mnt/vss/_work/1/s/env/lib/python3.11/site-packages/markupsafe/__init__.py)
```
Therefore, i want to enforce the use of a higher version of jinja2 for codegen coverage.
image

**Description**

**Testing Guide**

**History Notes**

---

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1173,7 +1173,7 @@ jobs:
       # CLI repo only
       azdev statistics list-command-table CLI --statistics-only
       # CLI + EXT repos
-      pip show MarkupSafe
+      pip install jinja2 -U
       azdev statistics list-command-table --statistics-only > /tmp/codegen_report.json
       python s/scripts/ci/codegen_report.py
     env:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1173,6 +1173,7 @@ jobs:
       # CLI repo only
       azdev statistics list-command-table CLI --statistics-only
       # CLI + EXT repos
+      pip show MarkupSafe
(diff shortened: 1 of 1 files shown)


yonzhan:

bebound:
Looks good as a workaround.

This task has bad design. It should not install **all extensions** in the same env and expects every extension works fine. We'll likely encounter another dependency conflict in the future.

bebound (approved):

AllyW (approved):

wangzelin007:
> Looks good as a workaround.
> 
> This task has bad design. It should not install **all extensions** in the same env and expects every extension works fine. We'll likely encounter another dependency conflict in the future.

We need to install all extensions to calculate code generation coverage.
If the problem occurs again, consider not installing the extension that uses the lower version of the dependency.

AllyW:
> > Looks good as a workaround.
> > This task has bad design. It should not install **all extensions** in the same env and expects every extension works fine. We'll likely encounter another dependency conflict in the future.
> 
> We need to install all extensions to calculate code generation coverage. If the problem occurs again, consider not installing the extension that uses the lower version of the dependency.

I feel the reliable way is to install, calculate and then uninstall each extension modules one by one and finally sum up the whole result from extension modules.

PR #30587, opened 2024-12-27
{Role} `az role definition list`: Use `roleName` filter when `--name` is a non-GUID string

jiasli:
**Related command**
`az role definition list`

**Description**
Close https://github.com/Azure/azure-cli/issues/30579

With the ever-increasing number of built-in roles, the list returned by [Role Definitions - List](https://learn.microsoft.com/en-us/rest/api/authorization/role-definitions/list?view=rest-authorization-2022-04-01&tabs=HTTP) API is getting bigger and bigger, making `az role definition list` command slower (https://github.com/Azure/azure-cli/issues/30579).

This PR leverages `roleName` filter to limit the number of role definitions returned by Role Definitions - List API.

Actually, Azure CLI has been using `roleName` as a filter to call Role Definitions - List API in `_resolve_role_id` since https://github.com/Azure/azure-cli/pull/508:
 
https://github.com/Azure/azure-cli/blob/e8efb791ea2f0ba5e4f172b26f75725fd8aa8079/src/azure-cli/azure/cli/command_modules/role/custom.py#L616

though the behavior was not documented. We use `roleName` filter in `az role definition list` too.

**Testing Guide**
```
az role definition list --name Reader
az role definition list --name acdd72a7-3385-48ef-bd42-f606fba81ae7
```

Issue #30471, opened 2024-12-05
[Feature Request] `az role assignment list/delete`: Make `--scope` a required argument


jiasli:
https://github.com/Azure/azure-cli/pull/27651 made `--scope` a required argument for `az role assignment create`.

I propose making `--scope` a required argument for `az role assignment list/delete` too, as explained below:

### `az role assignment list`

Even though the output of this command contains no secret, it contains PII - the email address of the assignee:

```
  {
    "id": "/subscriptions/xxx/providers/Microsoft.Authorization/roleAssignments/xxx",
    "principalId": "xxx",
    "principalName": xxx#EXT#@xxx.onmicrosoft.com,
    "principalType": "User",
    ...
  }
```

Without specifying `--scope`, this command lists all role assignments under the current subscription. The PII can be saved into build logs, such as GitHub Action logs. Therefore, we should ask the user to explicitly specify the scope under which role assignments should be listed to avoid showing PII as much as possible.

### `az role assignment delete`

https://github.com/Azure/azure-cli/issues/13577 made a very nice metaphor: This command is an insane footgun. Without specifying `--scope`, it deletes all role assignment under the current subscription. We should ask the user to explicitly specify the scope under which role assignments should be deleted to avoid expectedly nuking the whole subscription.

Making `--scope` a required argument aligns all `create`, `list` and `delete` commands to have a consistent interface.

yonzhan:
Thank you for opening this issue, we will look into it.

PR #30593, opened 2024-12-30
[Role] Add `az role definition show` command

jiasli:
**Related command**
`az role definition show`

**Description**
For `az role definition list`, https://github.com/Azure/azure-cli/pull/30587 started to use `roleName` filter when `--name` is a non-GUID string, but if the provided `--name` is a GUID, this command still uses a client-side filter which is inefficient. 

`name` cannot be used as a `filter` and the API fails with:

```
Code: UnsupportedFilter
Message: The filter 'name eq 'Reader'' is not supported. Supported filters are either roleName eq '{value}' or type eq 'BuiltInRole|CustomRole'.
```

We can internally switch to [Role Definitions - Get](https://learn.microsoft.com/en-us/rest/api/authorization/role-definitions/get?view=rest-authorization-2022-04-01&tabs=HTTP) API when `--name`  is a GUID and convert the result to a list, but this introduces complexity within `az role definition list` commands, as it will have 2 code paths and need to handle 404 error returned by the [Role Definitions - Get](https://learn.microsoft.com/en-us/rest/api/authorization/role-definitions/get?view=rest-authorization-2022-04-01&tabs=HTTP) API.

Having a dedicated `az role definition show` command is more consistent with other Azure CLI commands and the underlying REST API.

**Testing Guide**
```sh
# Show the 'Reader' role definition with its name (GUID).
az role definition show --scope /subscriptions/00000000-0000-0000-0000-000000000000 --name acdd72a7-3385-48ef-bd42-f606fba81ae7

# Show the 'Reader' role definition with its resource ID.
az role definition show --id /subscriptions/00000000-0000-0000-0000-000000000000/providers/Microsoft.Authorization/roleDefinitions/acdd72a7-3385-48ef-bd42-f606fba81ae7
```

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/role/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/role/_params.py
@@ -369,8 +369,23 @@ def load_arguments(self, _):
         c.argument('yes', options_list=['--yes', '-y'], action='store_true', help='Currently no-op.')
 
     with self.argument_context('role definition') as c:
-        c.argument('role_definition_id', options_list=['--name', '-n'], help='the role definition name')
         c.argument('custom_role_only', arg_type=get_three_state_flag(), help='custom roles only(vs. build-in ones)')
         c.argument('role_definition', help="json formatted content which defines the new role.")
+
+    with self.argument_context('role definition list') as c:
+        c.argument('name', arg_type=name_arg_type, completer=get_role_definition_name_completion_list,
+                   help="Matches the role definition's name (GUID) or roleName (e.g. 'Reader') property. "
+                        "If a GUID is provided, for better performance, use `az role definition show` command.")
+
+    with self.argument_context('role definition show') as c:
+        c.argument('name', arg_type=name_arg_type, help="The role definition's name (GUID)")
+        c.argument('role_id', options_list=['--id'],
+                   help='The fully qualified role definition ID. Use the format, '
+                        '/subscriptions/{guid}/providers/Microsoft.Authorization/roleDefinitions/{roleDefinitionId} '
+                        'for subscription level role definitions, or '
+                        '/providers/Microsoft.Authorization/roleDefinitions/{roleDefinitionId} '
+                        'for tenant level role definitions.')
+
+    with self.argument_context('role definition delete') as c:
         c.argument('name', arg_type=name_arg_type, completer=get_role_definition_name_completion_list,
                    help="Matches the role definition's name (GUID) or roleName (e.g. 'Reader') property.")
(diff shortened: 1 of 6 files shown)


yonzhan:

src/azure-cli/azure/cli/command_modules/role/_params.py
```diff
@@ -370,6 +370,14 @@ class PrincipalType(str, Enum):
 
     with self.argument_context('role definition') as c:
         c.argument('role_definition_id', options_list=['--name', '-n'], help='the role definition name')
+        c.argument('role_id', options_list=['--id'],
```
jiasli, line 373: The `--ids` argument provided by Azure CLI core relies on `azure.mgmt.core.tools.parse_resource_id` 

https://github.com/Azure/azure-cli/blob/b46f3440137ead38dd7d7b8674cffd573b191768/src/azure-cli-core/azure/cli/core/commands/arm.py#L324

which parses the resource ID into parts:

```
        - subscription:            Subscription id
        - resource_group:          Name of resource group
        - namespace:               Namespace for the resource provider (i.e. Microsoft.Compute)
        - type:                    Type of the root resource (i.e. virtualMachines)
        - name:                    Name of the root resource
        - child_namespace_{level}: Namespace for the child resource of that level
        - child_type_{level}:      Type of the child resource of that level
        - child_name_{level}:      Name of the child resource of that level
        - last_child_num:          Level of the last child
        - resource_parent:         Computed parent in the following pattern: providers/{namespace}\
        /{parent}/{type}/{name}
        - resource_namespace:      Same as namespace. Note that this may be different than the \
        target resource's namespace.
        - resource_type:           Type of the target resource (not the parent)
        - resource_name:           Name of the target resource (not the parent)
```

However, it can't parse `scope`, so we have to define our own `--id`.

src/azure-cli/azure/cli/command_modules/role/custom.py
```diff
@@ -71,6 +71,20 @@ def list_role_definitions(cmd, name=None, resource_group_name=None, scope=None,
     return _search_role_definitions(cmd.cli_ctx, definitions_client, name, [scope], custom_role_only)
 
 
+def show_role_definition(cmd, scope=None, name=None, role_id=None):
+    if not any((scope, name, role_id)):
+        raise CLIError('Usage error: Provide --scope and --name, or --id')
+    if not role_id and not (name and scope):
+        raise CLIError('Usage error: Provide both --scope and --name')
```
jiasli, line 78: `--scope` should be a required argument to align with `az role assignment create` and `az ad sp create-for-rbac`. Also see https://github.com/Azure/azure-cli/issues/30471.

evelyn-ys (approved):

bebound (approved):

PR #30606, opened 2025-01-03
{Compute} `az sig share`: Migrate command group to aaz based

cxznmhdcxz:
**Related command**

`az sig share add/reset/remove/enable-community/wait`

**Description**

Migrate command group to aaz based code
aaz pr https://github.com/Azure/aaz/pull/667

**Testing Guide**

**History Notes**

---


yonzhan:

yanzhudd:
Please remove the unused codes from custom.py.
Also, if this command group isn't supported in other profiles, please remove the related codes from the azure_stack folder as well.

yanzhudd:
Please check if the detected breaking changes would actually break customers' usage or not:
image

src/azure-cli/azure/cli/command_modules/vm/_params.py
```diff
@@ -1414,13 +1397,6 @@ def load_arguments(self, _):
         c.argument('marker', arg_type=marker_type)
         c.argument('show_next_marker', action='store_true', help='Show nextMarker in result when specified.')
 
-    with self.argument_context('sig share enable-community') as c:
-        c.argument('gallery_name', type=str, help='The name of the Shared Image Gallery.', id_part='name')
-        c.argument('subscription_ids', nargs='+', help='A list of subscription ids to share the gallery.')
-        c.argument('tenant_ids', nargs='+', help='A list of tenant ids to share the gallery.')
```
yanzhudd, line 1420: It seems these two params are removed from the `sig share enable-community` command and not added back in the new code implementation. Does it meet your expectations?

src/azure-cli/azure/cli/command_modules/vm/operations/sig_share.py
```diff
@@ -0,0 +1,185 @@
...
+            options=["--tenant-ids"],
+            help="A list of tenant ids to share the gallery with.",
+        )
+        tenant_ids = args_schema.tenant_ids
+        tenant_ids.Element = AAZStrArg()
+
+        args_schema.operation_type._required = False
+        args_schema.operation_type._registered = False
+        args_schema.groups._required = False
+        args_schema.groups._registered = False
```
yanzhudd, line 47: I'm not sure if we could omit `_required=False` if we set the param to `_registered=False` or not.
```suggestion
        args_schema.operation_type._registered = False
        args_schema.groups._registered = False
```

src/azure-cli/azure/cli/command_modules/vm/operations/sig_share.py
```diff
@@ -0,0 +1,185 @@
...
+
+        return args_schema
+
+    def pre_operations(self):
+        args = self.ctx.args
+        if not args.subscription_ids and not args.tenant_ids:
+            raise CLIError('At least one of subscription ids or tenant ids must be provided.')
```
yanzhudd, line 54: It is better to narrow down the error type, as `CLIError` is too vague.

cxznmhdcxz:
> Please check if the detected breaking changes would actually break customers' usage or not: image

`op_type` is not visible to users in released CLI, I handle it in `pre_operation` of aaz code, and there's no change to the usage.

src/azure-cli/azure/cli/command_modules/vm/_params.py
cxznmhdcxz: `sig share` commands call the same url, but for `sig share enable-community` these two params are not required and not handled, so I didn't add them

src/azure-cli/azure/cli/command_modules/vm/operations/sig_share.py
cxznmhdcxz: updated, had to keep `operation_type._required=False` because it's a required one in aaz generated code

src/azure-cli/azure/cli/command_modules/vm/_params.py
yanzhudd: you mean these two params are not used at all in `sig share enable-community` command, right?

src/azure-cli/azure/cli/command_modules/vm/operations/sig_share.py
cxznmhdcxz: replaced with `ArgumentUsageError`

yanzhudd:
> > Please check if the detected breaking changes would actually break customers' usage or not: image
> 
> `op_type` is not visible to users in current released CLI, I handle it in `pre_operation` of aaz code, and there's no change to the usage.

although it is handled in `pre_operation` of aaz code, customers cannot choose what to input for that param. It would break customers' usage. I think it is better to define that param, albeit without using `hide=True`, rather than setting it by ourselves.

yanzhudd:
please resolve the CI issue.

yanzhudd:
Please rerun some related tests in live mode to verify the changes.

yanzhudd (approved):

PR #30668, opened 2025-01-16
[Core] BREAKING CHANGE: Change get_config_dir to honor XDG spec

HStry:
**Description**

This commit modifies the `get_config_dir` function to honor the XDG Base Directory Specification in accordance with https://specifications.freedesktop.org/basedir-spec/latest/, and falls back to a dot-dir in the user's home directory if the relevant environment variables and home directories do not exist.

**Testing Guide**

On running the az cli, it should now refer to '${XDG_CONFIG_HOME}/azure', or '${HOME}/.config/azure'.
---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/_environment.py
+++ b/src/azure-cli-core/azure/cli/core/_environment.py
@@ -10,4 +10,20 @@ _ENV_AZ_BICEP_GLOBALIZATION_INVARIANT = 'AZ_BICEP_GLOBALIZATION_INVARIANT'
 
 def get_config_dir():
     import os
-    return os.getenv('AZURE_CONFIG_DIR', None) or os.path.expanduser(os.path.join('~', '.azure'))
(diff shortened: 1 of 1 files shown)


yonzhan:

HStry:

HStry:
> This is a devastating breaking change.
Yesterday, I indeed found out that it still made the ~/.azure directory after adding this change to my local setup, so yes, as it is, this is definitely breaking.
However it should not be. The XDG spec exists, is backwards compatible, and has been adopted by pretty much all distributions, for a reason. It's rather sad that large projects like Azure appear not to follow it, but I'd need a much longer time to figure out all the locations where this plugin has hard-coded the ~/.azure directory in order to adapt this.

And maybe have the first version check if the ~/.azure directory exists, and if so, use that regardless of what the XDG_CONFIG_HOME variable says.

PR #30469, opened 2024-12-05
[Role] `az role assignment list/delete`: Add `--assignee-object-id`

jiasli:
**Related command**
`az role assignment list`
`az role assignment delete`

**Description**
Close https://github.com/Azure/azure-cli/issues/30436
Fix #20349, #30428

With more encouragement on managed identity and more strict permission/access control, supporting `--assignee-object-id` to bypass Graph query is a necessary feature for role assignment’s management experience.

Similar to https://github.com/Azure/azure-cli/pull/5273 which added `--assignee-object-id` to `az role assignment create` to bypass Graph query, this PR adds `--assignee-object-id` to `az role assignment list/delete`.

**Testing Guide**
```
az role assignment list --assignee
az role assignment list --assignee-object-id
az role assignment delete --assignee
az role assignment delete --assignee-object-id
```

**History Notes**

[Role] `az role assignment list/delete`: Add `--assignee-object-id` argument. Use this argument instead of `--assignee` to bypass Microsoft Graph query

PR #30693, opened 2025-01-22
[Role] `az role assignment list`: Add `--fill-principal-name` to allow skipping filling `principalName` property

jiasli:
**Related command**
`az role assignment list`

**Description**
Fix https://github.com/Azure/azure-cli/pull/30469#issuecomment-2519867564

After getting role assignments from REST API, `az role assignment list` queries Microsoft Graph to fill the `principalName` property:

- For user: `userPrincipalName`
- For service principal: `ServicePrincipalNames[0]`
- For group: `displayName`

The motivation is to provide a display name of the assignee, but when the logged-in account has no permission, `az role assignment list` will show a warning:

https://github.com/Azure/azure-cli/blob/4d6d480acabfd017641a222451253b54ba011f0f/src/azure-cli/azure/cli/command_modules/role/custom.py#L300

To make things worse, if the machine has no network access to Microsoft Graph, the whole command will fail.

This PR adds `--fill-principal-name` argument to allow skipping filling `principalName` property.

**Testing Guide**
```
az role assignment list --role Reader --scope /subscriptions/00000000-0000-0000-0000-000000000000 --fill-principal-name false
```

**Additional information**
I deliberately choose `--fill-principal-name` instead of `--no-fill-principal-name`, because this gives us the possibility to turn off filling `principalName` by default. If we choose `--no-fill-principal-name`, turning on filling `principalName` will become `--no-fill-principal-name false` which is awkward.


yonzhan:

evelyn-ys (approved):

bebound (approved):

PR #30746, opened 2025-02-03
[Compute] `az vm disk attach-detach·`: Add new command to attach/detach data disk from a VM

nitinjainms:
**Related command**
az vm disk attach-detach-data-disk

**Description**
New API is defined to attach and detach data disk from a Virtual Machine. This PR adds CLI command for the same API.

**Testing Guide**

Following commands removes an existing data disk from a VM:
```
az vm disk attach-detach-data-disk --resource-group myResourceGroup --vm-name myVM --data-disks-to-detach [0].disk-id={disk1_id}
```

**History Notes**

---


yonzhan:

yanzhudd:
May I ask why need to add these commands? Since there are already commands to attach/detach disk to/from VM:
https://github.com/Azure/azure-cli/blob/97d67087aaabd792f74d7406367429d1ab204827/src/azure-cli/azure/cli/command_modules/vm/commands.py#L363-L365

nitinjainms:
> May I ask why need to add these commands? It seems there are already commands to attach/detach disk to/from VM:
> 
> https://github.com/Azure/azure-cli/blob/97d67087aaabd792f74d7406367429d1ab204827/src/azure-cli/azure/cli/command_modules/vm/commands.py#L363-L365

This command targets the new AttachDetach API which is recently introduced. As per this API customer need to send only the disk information which needs to be attached or detached.
Whereas existing command uses PUT VM API where customer need to specify the complete disk goal state i.e. all the disk needs to be specified which are required to be attached to the VM and already attached disk missing from the request will be detached.

New API was introduced to simplify the customer use case to attach and detach the disk without specifying the complete goal state and hence the respective CLI command for the same.

nitinjainms:
> Is there any update for the above comment? BTW, it is better to add an example for the command instead of just add a rule exclusion of missing_command_example.

I already added example in following PR
https://github.com/Azure/aaz/pull/683

Let me know if there is any other place where I should be adding similar example.

src/azure-cli/azure/cli/command_modules/vm/aaz/latest/vm/disk/_attach_detach_data_disk.py
```diff
@@ -11,6 +11,10 @@
 from azure.cli.core.aaz import *
 
 
+@register_command(
+    "vm disk attach-detach-data-disk",
+    is_preview=True,
+)
```
yanzhudd, line 17: The aaz repo is just the CodeGen model of CLI, and the logic of this CLI repo is actually what the users can use and see. You need to re-generate codes from the CodeGen UI, and the example will be display in this file.

yanzhudd:
> This command targets the new AttachDetach API which is recently introduced. As per this API customer need to send only the disk information which needs to be attached or detached.
> Whereas existing command uses PUT VM API where customer need to specify the complete disk goal state i.e. all the disk needs to be specified which are required to be attached to the VM and already attached disk missing from the request will be detached.
> 
> New API was introduced to simplify the customer use case to attach and detach the disk without specifying the complete goal state and hence the respective CLI command for the same.

May I ask what's the new API? Because it seems the API of the `az vm disk attach-detach-data-disk` command is:
image
which was defined in an old PR.

nitinjainms:
> > This command targets the new AttachDetach API which is recently introduced. As per this API customer need to send only the disk information which needs to be attached or detached.
> > Whereas existing command uses PUT VM API where customer need to specify the complete disk goal state i.e. all the disk needs to be specified which are required to be attached to the VM and already attached disk missing from the request will be detached.
> > New API was introduced to simplify the customer use case to attach and detach the disk without specifying the complete goal state and hence the respective CLI command for the same.
> 
> May I ask what's the new API? Because it seems the API of the `az vm disk attach-detach-data-disk` command is: image which was defined in an old PR.

That's correct. This is the API for disk attach-detach-data-disk. API was updated in swagger for public preview, but CLI command was not added for the same.

yanzhudd:
Please exclude the changes which are not related to this PR.

nitinjainms:
> Please exclude the changes which are not related to this PR.

Done.

yanzhudd (approved):

nitinjainms:
No more required, as new API was already added via https://github.com/Azure/azure-cli/pull/30236

PR #29631, opened 2024-08-01
[Core] Drop Track 1 SDK authentication

jiasli:
**Description**
Close https://github.com/Azure/azure-cli/issues/20460

After migrating the last Track 1 SDK `azure-batch` to Track 2 (#30501), Azure CLI no longer depends on any Track 1 SDK. This PR drops Track 1 SDK authentication support.

But before this, there are several incorrect/incomplete Track 2 migrations that must be fixed:

- https://github.com/Azure/azure-cli/pull/30785
- https://github.com/Azure/azure-cli/pull/30786
- https://github.com/Azure/azure-cli/pull/30787
- https://github.com/Azure/azure-cli/pull/30788
- https://github.com/Azure/azure-cli/pull/30798
- https://github.com/Azure/azure-cli/pull/30799

They are all caused by passing `resource` via `get_login_credentials()`, instead of passing credential scopes when creating the client. To make things worse, different data-plane SDKs have different implementations and forms of credential scopes.

PR #30787, opened 2025-02-07
{Batch} Set `audience` when creating `LogsQueryClient`

jiasli:
**Related command**

**Description**
The migration to `azure-monitor-query` SDK (https://github.com/Azure/azure-cli/pull/30501) is incomplete - it doesn't set `audience` when creating `LogsQueryClient`.

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/vm/_client_factory.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/_client_factory.py
@@ -130,10 +130,10 @@ def cf_log_analytics_data_plane(cli_ctx, _):
     from azure.monitor.query import LogsQueryClient
     from azure.cli.core._profile import Profile
     profile = Profile(cli_ctx=cli_ctx)
-    cred, _, _ = profile.get_login_credentials(
-        resource=cli_ctx.cloud.endpoints.log_analytics_resource_id)
+    cred, _, _ = profile.get_login_credentials()
     api_version = 'v1'
-    return LogsQueryClient(cred, endpoint=cli_ctx.cloud.endpoints.log_analytics_resource_id + '/' + api_version)
+    return LogsQueryClient(cred, endpoint=cli_ctx.cloud.endpoints.log_analytics_resource_id + '/' + api_version,
+                           audience=cli_ctx.cloud.endpoints.log_analytics_resource_id)
 
 
 def cf_disk_encryption_set(cli_ctx, _):
--- a/src/azure-cli/azure/cli/command_modules/vm/azure_stack/_client_factory.py
+++ b/src/azure-cli/azure/cli/command_modules/vm/azure_stack/_client_factory.py
@@ -130,10 +130,10 @@ def cf_log_analytics_data_plane(cli_ctx, _):
     from azure.monitor.query import LogsQueryClient
     from azure.cli.core._profile import Profile
     profile = Profile(cli_ctx=cli_ctx)
-    cred, _, _ = profile.get_login_credentials(
-        resource=cli_ctx.cloud.endpoints.log_analytics_resource_id)
+    cred, _, _ = profile.get_login_credentials()
     api_version = 'v1'
-    return LogsQueryClient(cred, endpoint=cli_ctx.cloud.endpoints.log_analytics_resource_id + '/' + api_version)
+    return LogsQueryClient(cred, endpoint=cli_ctx.cloud.endpoints.log_analytics_resource_id + '/' + api_version,
+                           audience=cli_ctx.cloud.endpoints.log_analytics_resource_id)
 
 
 def cf_disk_encryption_set(cli_ctx, _):


yonzhan:

src/azure-cli/azure/cli/command_modules/vm/_client_factory.py
```diff
@@ -130,10 +130,10 @@ def cf_log_analytics_data_plane(cli_ctx, _):
     from azure.monitor.query import LogsQueryClient
     from azure.cli.core._profile import Profile
     profile = Profile(cli_ctx=cli_ctx)
-    cred, _, _ = profile.get_login_credentials(
-        resource=cli_ctx.cloud.endpoints.log_analytics_resource_id)
+    cred, _, _ = profile.get_login_credentials()
     api_version = 'v1'
-    return LogsQueryClient(cred, endpoint=cli_ctx.cloud.endpoints.log_analytics_resource_id + '/' + api_version)
+    return LogsQueryClient(cred, endpoint=cli_ctx.cloud.endpoints.log_analytics_resource_id + '/' + api_version,
+                           audience=cli_ctx.cloud.endpoints.log_analytics_resource_id)
```
jiasli, line 136: The naming of `audience` doesn't align with Track 2 mgmt SDKs where its equivalent is called `credential_scopes`.

Also, according to `azure.monitor.query._helpers.get_authentication_policy`

```py
def get_authentication_policy(credential: TokenCredential, audience: str) -> BearerTokenCredentialPolicy:
    """Returns the correct authentication policy"""
    if credential is None:
        raise ValueError("Parameter 'credential' must not be None.")
    scope = audience.rstrip("/") + "/.default"
```

the value of `scopes` is a `str` without the `/.default` suffix, unlike `credential_scopes`'s value which is a `list[str]`.

jiasli, line 136: When `audience` is not provided, it falls back to the domain of the URL according to `azure.monitor.query._logs_query_client.LogsQueryClient.__init__`

```py
    def __init__(self, credential: TokenCredential, **kwargs: Any) -> None:
        endpoint = kwargs.pop("endpoint", "https://api.loganalytics.io/v1")
        if not endpoint.startswith("https://") and not endpoint.startswith("http://"):
            endpoint = "https://" + endpoint
        parsed_endpoint = urlparse(endpoint)
        audience = kwargs.pop("audience", f"{parsed_endpoint.scheme}://{parsed_endpoint.netloc}")
        self._endpoint = endpoint
        auth_policy = kwargs.pop("authentication_policy", None)
        self._client = MonitorQueryClient(
            credential=credential,
            authentication_policy=auth_policy or get_authentication_policy(credential, audience),
```

So passing `audience` is not really necessary.

evelyn-ys: Shall we still add `audience` on client creation? Maybe we can just remove `resource` parameter when calling `Profile.get_login_credentials()`

src/azure-cli/azure/cli/command_modules/vm/_client_factory.py
```diff
@@ -130,10 +130,10 @@ def cf_log_analytics_data_plane(cli_ctx, _):
     from azure.monitor.query import LogsQueryClient
     from azure.cli.core._profile import Profile
     profile = Profile(cli_ctx=cli_ctx)
-    cred, _, _ = profile.get_login_credentials(
-        resource=cli_ctx.cloud.endpoints.log_analytics_resource_id)
+    cred, _, _ = profile.get_login_credentials()
```
zhoxing-ms, line 133: I currently lack some domain knowledge about Auth, may I ask why the resource attribute can be removed for `profile.get_login_credentials`?

jiasli: This is a very good question. I wrote a detailed explanation at https://github.com/Azure/azure-cli/pull/29631#discussion_r1947487038.

src/azure-cli/azure/cli/command_modules/vm/_client_factory.py
jiasli: Setting `audience` may not be necessary, but it will be necessary if the URL's domain is not the same as the audience in certain clouds.

For example, for [Dogfood cloud](https://eng.ms/docs/products/arm/rp_onboarding/testing/dogfood), ARM's resource ID is `https://management.core.windows.net/`, but its domain is `https://api-dogfood.resources.windows-int.net/`.

src/azure-cli/azure/cli/command_modules/vm/_client_factory.py
zhoxing-ms: Great! thanks for the detailed explanation

evelyn-ys (approved):

zhoxing-ms (approved):

PR #30905, opened 2025-02-27
{LTS} Do not add `post` suffix for branches starting with `release`

jiasli:
**Description**
The `pypi` build artifact of ec3172dd54637aadb24ab78cdfb7c7b956d26cd2 contains the `post` suffix:

image

These packages are incorrectly published to PyPI: https://pypi.org/project/azure-cli/2.66.1.post20250227030620/#files

image

We should not add `post` suffix for branches starting with `release` including `release-lts-2.66`.

**Testing Guide**
Create a new branch `release-lts-2.66-test` and trigger the release pipeline. Make sure the artifacts does not contain `post` suffix.

### DIFF
--- a/scripts/release/pypi/build.sh
+++ b/scripts/release/pypi/build.sh
@@ -22,7 +22,7 @@ pip list
 
 script_dir=`cd $(dirname $BASH_SOURCE[0]); pwd`
 
-if [[ "$branch" != "release" ]]; then
+if [[ ! $branch =~ ^release ]]; then
     . $script_dir/../../ci/version.sh post`date -u '+%Y%m%d%H%M%S'`
 fi
 


scripts/release/pypi/build.sh
```diff
@@ -22,7 +22,7 @@ pip list
 
 script_dir=`cd $(dirname $BASH_SOURCE[0]); pwd`
 
-if [[ "$branch" != "release" ]]; then
+if [[ ! $branch =~ ^release ]]; then
```
jiasli, line 25: To test this regex match:

```sh
branch=dev
if [[ ! $branch =~ ^release ]]; then
    echo not a release branch
fi
# output: not a release branch

branch=myfeature
if [[ ! $branch =~ ^release ]]; then
    echo not a release branch
fi
# output: not a release branch

branch=release
if [[ ! $branch =~ ^release ]]; then
    echo not a release branch
fi

branch=release-lts-2.66
if [[ ! $branch =~ ^release ]]; then
    echo not a release branch
fi
```

jiasli, line 25: Double quotes in `"$branch"` is not necessary. The official Bash doc doesn't quote the first operand:

https://www.gnu.org/software/bash/manual/bash.html#index-_005b_005b

```sh
[[ $line =~ ^"initial string" ]]
```

This is also confirmed by https://unix.stackexchange.com/questions/500828/why-use-double-quotes-in-a-test

From the output of `set -x`, we can see there is no difference:

```sh
set -x
branch="with a space"

if [[ $branch =~ ^release ]]; then
    echo matched
fi
# output: + [[ with a space =~ ^release ]]

if [[ "$branch" =~ ^release ]]; then
    echo matched
fi
# output: + [[ with a space =~ ^release ]]
```

yonzhan:
LTS

bebound (approved):

Issue #30897, opened 2025-02-26
[Reference feedback]: az afd endpoint purge | `--domains` parameter does not accept multiple entries


dbradish-microsoft:
### Type of issue

Code doesn't work

### Reference command name

az afd endpoint purge

### Feedback

This GitHub issue is being created from UUF [#374646](https://dev.azure.com/msft-skilling/Content/_workitems/edit/374646).
********************************************************
### Verbatim:
The --domains parameter for az afd endpoint purge is inaccurate. This parameter doesn't accept multiple entries. I suspect the issue raised to add this didn't add the functionality to the CLI, just to the documentation: https://github.com/Azure/azure-cli/issues/29801

### Verbatim Id:
8038c060-d9ea-ef11-9341-6045bd05ffec

### UUF Triage App Feedback URL:
https://aka.ms/UUFTriageApp?pageviewid=dcaa1444-244b-3ca4-a1da-7ef93382f428

### Page URL

https://learn.microsoft.com/en-us/cli/azure/afd/endpoint

### Content source URL

https://github.com/MicrosoftDocs/azure-docs-cli/blob/live/latest/docs-ref-autogen/afd/endpoint.yml

yonzhan:
Thank you for opening this issue, we will look into it.

FrickinCode:
Came across this exact issue 2 days ago myself.  After a bunch of trial and error, I found the format of "[www.contoso.com,www.contoso1.com]" to be successful.  Enclosed in square brackets, comma separated, with no spaces after commas.  So the functionality is there, but the documentation is just out of date.

Ptnan7:
https://github.com/Azure/azure-cli/pull/30933

SazniMohamed:
> "[www.contoso.com,www.contoso1.com]"

After spending hours debugging due to poorly written and misleading Azure documentation, I finally found the answer thanks to @FrickinCode 

Microsoft — this doc needs serious attention. The current explanation is vague and omits critical behavior details, which leads to confusion and wasted time. Please update it to reflect real-world usage and gotchas.

PR #30933, opened 2025-03-04
{afd} Fix 30897 - change docs content

Ptnan7:
Fix https://github.com/Azure/azure-cli/issues/30897


FrickinCode:
This reverts a doc update to show an accurate example.  Also, doesn't properly update the example command in the line above where this update is done.

necusjz (approved):

PR #30975, opened 2025-03-07
{Core} Extension Breaking Change Announcement Instruction

ReaNAiveD:
**Related command**

**Description**

* Add Pre-announcement instruction for breaking changes of extensions.
* Support a plain text target version, which is used for date or month.

**Testing Guide**

**History Notes**

---


yonzhan:

doc/how_to_introduce_breaking_changes.md
```diff
@@ -29,28 +29,57 @@ You could find the next Breaking Change Release plan in our [milestones](https:/
...
-* CLI Owned Module
-  * Service Team should create an Issue that requests CLI Team to create the pre-announcement several sprints ahead Breaking Change Window. The issue should include the label `Breaking Change`. The CLI team will look at the issue and evaluate if it will be accepted in the next breaking change release.
+* **CLI Owned Module**
+  * Service Team should create an Issue that requests CLI Team to create the pre-announcement at least **2** sprints ahead of Breaking Change Window. The issue should include the label `Breaking Change`. The CLI team will look at the issue and evaluate if it will be accepted in the next breaking change release.
     * Please ensure sufficient time for CLI Team to finish the pre-announcement.
-  * The pre-announcement should be released ahead of Breaking Change Window.
-* Service Owned Module
-  * Service Team should create a Pull Request that create the pre-announcement several sprints ahead Breaking Change Window.
-  * The pre-announcement should be released ahead of Breaking Change Window.
+  * The pre-announcement should be released at least **2** sprints ahead of Breaking Change Window.
```
Jacekey23, line 68: usually I suppose sprint will vary across team, but monthly basis is often adopted.

If we put as 1-month policy above(Line 32), we might keep it aligned here, rather than 2 sprints, right?

doc/how_to_introduce_breaking_changes.md
```diff
@@ -29,28 +29,57 @@ You could find the next Breaking Change Release plan in our [milestones](https:/
...
+* **Service Owned Module**
+  * Service Team should create a Pull Request that adds the pre-announcement at least **2** sprints ahead of Breaking Change Window.
+  * The pre-announcement should be released at least **2** sprints ahead of Breaking Change Window.
```
Jacekey23, line 71: and here

doc/how_to_introduce_breaking_changes.md
```diff
@@ -72,7 +101,7 @@ You can then pre-announce breaking changes for different command groups or comma
 from azure.cli.core.breaking_change import register_required_flag_breaking_change, register_default_value_breaking_change, register_other_breaking_change
 
 register_required_flag_breaking_change('bar foo', '--name')
-register_default_value_breaking_change('bar foo baz', '--foobar', 'A', 'B')
+register_default_value_breaking_change('bar foo baz', '--foobar', 'A', 'B', target_version='May 2025')
```
zhoxing-ms, line 104: May I ask if our target version is currently using the date style?

ReaNAiveD: I support using plain text in this PR, so that we could support date style. This will be beneficial for an upcoming breaking change in an extension, which could occur in the coming months.

I don't believe we need to ensure consistency in the target version style, as version numbers are unpredictable for both developers and customers. While they know the version, they don't know when it will be released. Therefore, it is more flexible to support both date and version formats.

zhoxing-ms: Yes, I agree with your~ But what I mean is that since we all use date styles, is it necessary to have a unified style to avoid some people writing the wrong date or inconsistent date styles? Such as some people write `May 2025` while others write `2025-05`、`05/2025`.....

ReaNAiveD: It is not easy to constraint. I think it's ok to keep the flexibility of a field with any str.
Usually, developers would follow the example in doc. And I will also add instructions about date format in doc.

zhoxing-ms (approved):

doc/how_to_introduce_breaking_changes.md
zhoxing-ms: OK, LGTM

kairu-ms (approved):

PR #30990, opened 2025-03-08
[App Service] Fix: #21721: `az webapp config storage-account add`: Add validation for non-existent FileShare

### LINKED ISSUE #21721: Adding to an AppService a mount to an inexistent Azure FileShare doesn't produce an error (but breaks the appService instance)

**Describe the bug**
The following command should return an error when <FILE_SHARE_NAME> doesn't exist in the target storage account:

`az webapp config storage-account add -g <RESSOURCE_GROUP_NAME> -n <APP_SERVICE_NAME> --custom-id "test-az-cli" --storage-type AzureFiles --account-name <STORAGE_ACCOUNT_NAME> --share-name <FILE_SHARE_NAME> --access-key <STORAGE_ACCOUNT_ACCESS_KEY> --mount-path <MOUNT_PATH_ON_APP_SERVICE_FS> --debug`

**When ran on an existing appService, the above command put the appService in a semi-broken state where Kudu become.unavailable and deployments are not working anymore**

**To Reproduce**

- Create a storage account
- Create an appService instance
- run the above command with the right parameters *and* a non-existent fileshare name.
- try to deploy some code to the app, or try to ssh into it (webSSH), or try to look at deployment logs, or try to access Kudu UI (`https://<appServiceInstanceName>.scm.azurewebsites.net` ) : everything should fail

**Expected behavior**

- `az webapp config storage-account add ` should fail / return an error when specifying an inexistent File_share.
- the endpoint of Azure management API `https://management.azure.com/subscriptions/<subscriptionID>/resourceGroups/<RG_NAME>/providers/Microsoft.Web/sites/<APP_NAME>/config/azurestorageaccounts/add` should return an error as well

**Environment summary**

- AppService Linux (Node in our case)
- we use terraform to provision our infra (but the problem was reproduced with azure-cli)

**Additional context**

N/A

seligj95:
**Related command**

az webapp config storage-account add

**Description**

Fixes #21721 

**Testing Guide**

The following command now returns an error when <FILE_SHARE_NAME> doesn't exist in the target storage account:

az webapp config storage-account add -g <RESSOURCE_GROUP_NAME> -n <APP_SERVICE_NAME> --custom-id "test-az-cli" --storage-type AzureFiles --account-name <STORAGE_ACCOUNT_NAME> --share-name <FILE_SHARE_NAME> --access-key <STORAGE_ACCOUNT_ACCESS_KEY> --mount-path <MOUNT_PATH_ON_APP_SERVICE_FS>

**History Notes**

---


yonzhan:
Please fix CI issues

yanzhudd (approved):

zhoxing-ms:
@seligj95 I want to confirm that the newly added validation will not cause any breaking changes to customers, right?

seligj95:
> @seligj95 I want to confirm that the newly added validation will not cause any breaking changes to customers, right?

@zhoxing-ms - I wouldn't consider this a breaking change. This adds validation so that customers won't break their apps. If you add a FileShare that doesn't exist, the app stops working. So I added validation to check if the FileShare exists before it's added to ensure this command won't lead to a broken app. This might break some customer's pipelines if they are adding FileShares that don't exist for some reason, but I think this change is necessary to ensure that doesn't happen. If you think this is a breaking change, please let me know. Otherwise, I think it's fine to proceed.

zhoxing-ms:
Make sense, thanks for your detailed explain~ @seligj95

ksaffarian:
Hi @seligj95, @zhoxing-ms 
This fix has broken our pipeline. We mount a AzureBlob storage type to webapp. However, the command is showing error that share does not exist. I believe the new validation is not taking the mount type into account.
Please revert or publish a fix.

FYI. this azure pipeline task https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/azure-cli-v2?view=azure-pipelines downloads latest cli version and is now broken for us.

ksaffarian:
Hi @seligj95 , @navba-MSFT , @0gust1
This fix has broken our pipeline. We mount a AzureBlob storage type to webapp. However, the command is showing error that share does not exist. I believe the new validation is not taking the mount type into account.
Please revert or publish a fix.

FYI. this azure pipeline task https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/azure-cli-v2?view=azure-pipelines downloads latest cli version and is now broken for us.

ksaffarian:
additionally, it is expecting the file share to exist in the same resource group as the app. In our case, the storage account is in a different resource group. So, even if I add a dummy file share to bypass the validation, it does not work.

seligj95:
@zhoxing-ms can we roll this back? i need to dig deeper into the solution for this

PR #25959, opened 2023-03-28
[Core] Use MSAL for managed identity authentication

jiasli:
**Related command**
`az login`

**Description**
Close #25860, #22677

https://github.com/Azure/azure-cli/pull/31092 added support for MSAL managed identity. It can be opted in by setting `core.use_msal_managed_identity=true` config option.

This PR makes `core.use_msal_managed_identity=true` the default, but `core.use_msal_managed_identity` is kept so that it is possible to revert to the old `msrestazure` implementation by setting `core.use_msal_managed_identity=false`.

**Testing Guide**
```sh
# system-assigned
az login --identity

# user-assigned
az login --identity --client-id xxx
az login --identity --object-id xxx
az login --identity --resource-id /subscriptions/xxx/resourcegroupsxxx/providers/microsoft.managedidentity/userassignedidentities/xxx

az group list
az account get-access-token
```

**History Notes**

[Core] Use MSAL for managed identity authentication. To revert to the old `msrestazure` implementation, run `az config set core.use_msal_managed_identity=false` or set environment variable `AZURE_CORE_USE_MSAL_MANAGED_IDENTITY=false`

PR #29637, opened 2024-08-02
[Core] Use MSAL for Cloud Shell authentication

jiasli:
**Related command**
`az login --identity`

**Description**
This PR is separated from https://github.com/Azure/azure-cli/pull/25959.

Use MSAL for Cloud Shell authentication (https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/420).

**Testing Guide**
Install Azure CLI in Cloud Shell:

```
git clone https://github.com/jiasli/azure-cli/
cd azure-cli
git switch cloud-shell-auth
python3 -m venv env
. env/bin/activate
pip install azdev
azdev setup -c
```

Log in:

```
az login --identity
```

Some basic testing:

```
az account get-access-token
az group show -n nonexist
```

Test getting SSH certificate:

```
az extension add -n ssh
az ssh vm --resource-group xxx --name xxx --debug
```

Test mgmt-plane Track 1 SDK:

```
az dls account list -g $RG_NAME
```

Test data-plane Track 1 SDK:

```sh
# Change to 2019-03-01-hybrid profile
az cloud set -n AzureCloud --profile 2019-03-01-hybrid
# Create a storage account and then test data plane command
az storage account create -n $SA_NAME -g $RG_NAME
az storage container list --account-name $SA_NAME --auth-mode login
# Revert to latest profile
az cloud set -n AzureCloud --profile latest

# Create a managed hsm
az keyvault create --hsm-name $HSM_NAME -g $RG_NAME -l uksouth --administrators $USER_OBJECT_ID --retention-days 7
# Prepare 3 cert files
openssl req -newkey rsa:2048 -nodes -keyout cert_0.key -x509 -days 365 -out cert_0.cer
openssl req -newkey rsa:2048 -nodes -keyout cert_1.key -x509 -days 365 -out cert_1.cer
openssl req -newkey rsa:2048 -nodes -keyout cert_2.key -x509 -days 365 -out cert_2.cer
# Test data plane command
az keyvault security-domain download --hsm-name $HSM_NAME --sd-wrapping-keys cert_0.cer cert_1.cer cert_2.cer --sd-quorum 2 --security-domain-file mhsm-SD.json
```

PR #31014, opened 2025-03-12
{Auth} Clarify `Identity` class's boundary

jiasli:
**Related command**
`az login`

**Description**
Part of #25959

`Identity` will never be used for managed identity authentication, as managed identity inherently differs from user or service principal:

- User or service principal talks to Microsoft Entra eSTS, so it requires an `authority`. Also, because the network request is expensive, a local token cache is used. 
- Managed identity talks to IMDS on the local machine, so no `authority` is required. As IMDS has its own token cache, a local token cache is not necessary.

Of course, it is possible to make `authority` optional and `Identity` work for managed identity, but this requires more effort and blurs `Identity`'s boundary. According to https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-auth-code-flow, the public-facing name of eSTS is "Microsoft identity platform", so calling this class `Identity` and limit it to user or service principal authentication is accurate and reasonable.

This PR clarifies `Identity` class's boundary and deletes unused methods.

### DIFF
--- a/src/azure-cli-core/azure/cli/core/auth/identity.py
+++ b/src/azure-cli-core/azure/cli/core/auth/identity.py
@@ -41,10 +41,7 @@ logger = get_logger(__name__)
 
 
 class Identity:  # pylint: disable=too-many-instance-attributes
-    """Class to manage identities:
-        - user
-        - service principal
-        - TODO: managed identity
+    """Manage user or service principal identities and log into Microsoft identity platform.
     """
 
     # MSAL token cache.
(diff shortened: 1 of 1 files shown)


yonzhan:

evelyn-ys (approved):

src/azure-cli-core/azure/cli/core/auth/identity.py
```diff
@@ -199,12 +196,6 @@ def login_with_service_principal(self, client_id, credential, scopes):
         entry = sp_auth.get_entry_to_persist()
         self._service_principal_store.save_entry(entry)
 
-    def login_with_managed_identity(self, scopes, identity_id=None):  # pylint: disable=too-many-statements
-        raise NotImplementedError
-
-    def login_in_cloud_shell(self, scopes):
-        raise NotImplementedError
```
evelyn-ys, line 206: Why do we need to remove `login_in_cloud_shell` Since cloud_shell is a special user identity in msal?

jiasli: I don't agree with MSAL's design, see https://github.com/Azure/azure-cli/pull/29637#issuecomment-2350937911, https://github.com/Azure/azure-cli/pull/29637#discussion_r1998291449 for detailed explanations.

I want to keep the calling pattern for `Identity` consistent - `authority` should always be a required parameter.

For Cloud Shell authentication, the CLI command is `az login --identity`. It **is** a managed identity credential, no matter what token is returned. Including `CloudShellCredential` in `Identity` blurs its boundary.

PR #31019, opened 2025-03-12
[App Service] `az functionapp list-flexconsumption-locations`: Add `--details` and `--runtime` parameters to provide more details

khkh-ms:
**Related command**
az functionapp list-flexconsumption-locations

**Description**
Add --details and --runtime parameter to az functionapp list-flexconsumption-locations. The --details parameter will provide more details about the locations including the supported runtimes, and the --runtime parameter will filter the locations based on the runtime.

**Testing Guide**

- az functionapp list-flexconsumption-locations --details
- az functionapp list-flexconsumption-locations --runtime java --details

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/_params.py
@@ -232,6 +232,9 @@ subscription than the app service environment, please use the resource ID for --
     with self.argument_context('functionapp list-flexconsumption-locations') as c:
         c.argument('zone_redundant', arg_type=get_three_state_flag(),
                    help='Filter the list to return only locations which support zone redundancy.', is_preview=True)
+        c.argument('details', arg_type=get_three_state_flag(),
+                   help='Include the runtime details of the regions.', is_preview=True)
+        c.argument('runtime', help="limit the output to just the specified runtime", is_preview=True)
 
     with self.argument_context('webapp deleted list') as c:
         c.argument('name', arg_type=webapp_name_arg_type, id_part=None)
--- a/src/azure-cli/azure/cli/command_modules/appservice/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/_params.py
@@ -232,7 +232,7 @@ subscription than the app service environment, please use the resource ID for --
     with self.argument_context('functionapp list-flexconsumption-locations') as c:
         c.argument('zone_redundant', arg_type=get_three_state_flag(),
                    help='Filter the list to return only locations which support zone redundancy.', is_preview=True)
-        c.argument('details', arg_type=get_three_state_flag(),
+        c.argument('show_details', options_list=['--show-details'], arg_type=get_three_state_flag(),
                    help='Include the runtime details of the regions.', is_preview=True)
         c.argument('runtime', help="limit the output to just the specified runtime", is_preview=True)
 
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -6125,10 +6125,10 @@ def get_subscription_locations(cli_ctx):
     return [item.name for item in result]
 
 
-def list_flexconsumption_locations(cmd, zone_redundant=False, details=False, runtime=None):
+def list_flexconsumption_locations(cmd, zone_redundant=False, show_details=False, runtime=None):
     client = web_client_factory(cmd.cli_ctx)
 
-    if runtime and not details:
+    if runtime and not show_details:
         raise ArgumentUsageError(
             '--runtime is only valid with --details parameter. '
             'Please try again without the --details parameter.')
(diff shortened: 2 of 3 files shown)


src/azure-cli/azure/cli/command_modules/appservice/_params.py
```diff
@@ -232,6 +232,9 @@ def load_arguments(self, _):
     with self.argument_context('functionapp list-flexconsumption-locations') as c:
         c.argument('zone_redundant', arg_type=get_three_state_flag(),
                    help='Filter the list to return only locations which support zone redundancy.', is_preview=True)
+        c.argument('details', arg_type=get_three_state_flag(),
```
kamperiadis, line 235: Can you rename this to `show-details`? I think just `details` is a bit confusing.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -1810,13 +1810,18 @@ def list_function_app_runtimes(cmd, os_type=None):
...
-                              .format(runtime))
-    return runtimes
+def list_flex_function_app_runtimes(cmd, location, runtime, ignore_error=False):
+    try:
+        runtime_helper = _FlexFunctionAppStackRuntimeHelper(cmd, location, runtime)
+        runtimes = [r for r in runtime_helper.stacks if runtime == r.name]
+        if not runtimes:
+            raise ValidationError("Runtime '{}' not supported for function apps on the Flex Consumption plan."
+                                  .format(runtime))
+        return runtime_helper.stacks
```
kamperiadis, line 1820: This is a bit confusing in that we consume the runtime param but when it is passed, we do not just return info specific to the runtime. I thought the ticket was to just modify the `az functionapp list-flexconsumption-locations` command? This would modify the `az functionapp list-flexconsumption-runtimes` command.

kamperiadis (changes requested):

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -1810,13 +1810,18 @@ def list_function_app_runtimes(cmd, os_type=None):
...
+    except Exception:
+        if ignore_error:
+            return []
+        raise  # Rethrow the caught exception
```
kamperiadis, line 1824: Why are we adding this? If we throw the validation exception, we shouldn't just ignore it. This command was meant to only work for a specific runtime. That's why we throw the ValidationError.

kamperiadis (changes requested):

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -6132,18 +6130,44 @@ def get_subscription_locations(cli_ctx):
...
-def list_flexconsumption_locations(cmd, zone_redundant=False):
+def list_flexconsumption_locations(cmd, zone_redundant=False, details=False, runtime=None):
+    client = web_client_factory(cmd.cli_ctx)
+
+    if runtime and not details:
+        raise ArgumentUsageError(
+            '--runtime is only valid with --details parameter. '
+            'Please try again without the --details parameter.')
+
+    regions = client.list_geo_regions(sku="FlexConsumption")
```
kamperiadis, line 6141: Does this make the API call to 2023-01-01? I thought it wasn't and that was why we had to the raw API call below? Otherwise, I can't remember anymore why we were just doing the raw API call. 

```
geo_regions_api = 'subscriptions/{}/providers/Microsoft.Web/geoRegions?sku=FlexConsumption&api-version=2023-01-01'
request_url = cmd.cli_ctx.cloud.endpoints.resource_manager + geo_regions_api.format(sub_id)
flex_regions = send_raw_request(cmd.cli_ctx, "GET", request_url).json()['value']
```

kamperiadis (changes requested):

src/azure-cli/azure/cli/command_modules/appservice/custom.py
khkh-ms: Addressed it.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
khkh-ms: Detail requests on certain staging locations throws bad request. Will be returning empty/null for that situation.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
khkh-ms: It is now making a call to 2023-01-01.

yanzhudd:
please note that Azure CLI will have a code freeze on 03/25/2025 07:00 UTC for the upcoming release. If you want to catch this release train, address the commands ASAP, otherwise it has to be postponed to next sprint (05-06).

khkh-ms:
Looks like pipeline is failing. A number of builds are failing on `Test Extensions Loading Python312`.

kamperiadis (approved):

yonzhan:
Please fix CI issues

yanzhudd (approved):

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -6132,18 +6125,52 @@ def get_subscription_locations(cli_ctx):
...
 
 
-def list_flexconsumption_locations(cmd, zone_redundant=False):
+def list_flexconsumption_locations(cmd, zone_redundant=False, show_details=False, runtime=None):
+    client = web_client_factory(cmd.cli_ctx)
+
+    if runtime and not show_details:
+        raise ArgumentUsageError(
+            '--runtime is only valid with --details parameter. '
+            'Please try again without the --details parameter.')
```
nzthiago, line 6134: Should these now say "show_details" instead of "details"?

PR #31029, opened 2025-03-13
{Microsoft Entra ID} `az ad`: Ignore global `--subscription` in subgroups

jiasli:
**Related command**
`az ad`

**Description**
Fix https://github.com/Azure/azure-cli/pull/27296#issuecomment-1878581800

Currently, the global `--subscription` argument is ignored on `az ad` command group (#6516).

When generating reference doc, all `command_loaders` are loaded:

https://github.com/Azure/azure-cli/blob/dc4393dfe5319df8ce4190f12f3f981ff8f49563/src/azure-cli-core/azure/cli/core/__init__.py#L497-L502

If `role` command module is loaded before the `ad` extension, commands under `az ad ds` (https://github.com/Azure/azure-cli-extensions/pull/2842) will not have `--subscription`.

This PR moves `c.ignore('_subscription')` to subgroups to avoid affecting other groups under `az ad`, including `az ad ds`:

```
> az ad -h
...
Subgroups:
    app            : Manage Microsoft Entra applications.
    group          : Manage Microsoft Entra groups.
    signed-in-user : Show graph information about current signed-in user in CLI.
    sp             : Manage Microsoft Entra service principals.
    user           : Manage Microsoft Entra users.
```

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/role/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/role/_params.py
@@ -23,13 +23,13 @@ JSON_PROPERTY_HELP = "Should be JSON file path or in-line JSON string. See examp
 # pylint: disable=too-many-statements
 def load_arguments(self, _):
     with self.argument_context('ad') as c:
-        c.ignore('_subscription')  # hide global subscription param
         c.argument('owner_object_id', help="owner's object id")
         c.argument('show_mine', action='store_true', help='list entities owned by the current user')
         c.argument('include_all', options_list='--all', action='store_true',
                    help='list all entities, expect long delay if under a big organization')
 
     with self.argument_context('ad app') as c:
+        c.ignore('_subscription')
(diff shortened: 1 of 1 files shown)


yonzhan:

jiasli:
This issue doesn't happen for in-tool help, because when a specific `command` is provided, in this case `ad ds create`, only the command loader related to the command is loaded:

https://github.com/Azure/azure-cli/blob/dc4393dfe5319df8ce4190f12f3f981ff8f49563/src/azure-cli-core/azure/cli/core/__init__.py#L504

For example, `az ad ds create` is registered by `azext_ad.DomainServicesResourceProviderCommandsLoader`, so only this loader will be loaded.

Even though `c.ignore('_subscription')` happens in `azure.cli.command_modules.role.RoleCommandsLoader.load_arguments`, it is not executed at all.

evelyn-ys (approved):

PR #31123, opened 2025-03-25
{CI} Revert to Ubuntu 20.04 on ARM agent

bebound:
**Description**

Partially revert https://github.com/Azure/azure-cli/pull/30996.

Arm agent raises `docker: Error response from daemon: open /sys/fs/cgroup/user.slice/user-1000.slice/cgroup.controllers: no such file or directory` when run docker command. Revert to Ubuntu 20.04 to unblock release.

Ref: https://dev.azure.com/azclitools/public/_build/results?buildId=230090&view=logs&jobId=f970374f-0f25-5e15-cb2e-539daf27ea54&j=f970374f-0f25-5e15-cb2e-539daf27ea54&t=04e949c8-23b2-574b-c689-3d7f03b1f36c

PR #31158, opened 2025-03-31
{CI} Use Ubuntu 24.04 on ARM agent

bebound:
**Description**

This PR enables `pool-ubuntu-latest-arm64` pool, which uses Ubuntu 24.04

This PR also changes rootless docker to normal docker.

Related PR:
- https://github.com/Azure/azure-cli/pull/30996
- https://github.com/Azure/azure-cli/pull/31123

### DIFF
--- a/.azure-pipelines/templates/variables.yml
+++ b/.azure-pipelines/templates/variables.yml
@@ -2,5 +2,5 @@ variables:
   ubuntu_pool: 'pool-ubuntu-2204'
   ubuntu_multi_core_pool: 'pool-ubuntu-latest-multi-core'
   windows_pool: 'pool-windows-2019'
-  ubuntu_arm64_pool: 'ubuntu-arm64-2004-pool'
+  ubuntu_arm64_pool: 'pool-ubuntu-latest-arm64'
   macos_pool: 'macOS-14'
--- a/scripts/ci/install_docker.sh
+++ b/scripts/ci/install_docker.sh
@@ -5,11 +5,7 @@ if [[ $(dpkg --print-architecture) == "amd64" ]]; then
     echo "Docker is already installed on AMD64"
     exit 0
 fi
-# https://docs.docker.com/engine/security/rootless/
-/bin/bash -c "$(curl -fsSL https://get.docker.com)"
-sudo apt-get install -y uidmap
-dockerd-rootless-setuptool.sh install
-export XDG_RUNTIME_DIR=/home/cloudtest/.docker/run
-PATH=/usr/bin:/sbin:/usr/sbin:$PATH dockerd-rootless.sh &
-sleep 5
-docker context use rootless
+
+# https://docs.docker.com/engine/install/ubuntu/#install-using-the-convenience-script
+curl -fsSL https://get.docker.com/ | sh
+sudo chmod 666 /var/run/docker.sock
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -30,7 +30,7 @@ parameters:
     pool: pool-ubuntu-latest-multi-core
   - name: ARM64
     value: arm64
-    pool: ubuntu-arm64-2004-pool
+    pool: pool-ubuntu-latest-arm64
 
 jobs:
 - job: CheckPullRequest


yonzhan:

scripts/ci/install_docker.sh
```diff
@@ -5,11 +5,7 @@ if [[ $(dpkg --print-architecture) == "amd64" ]]; then
     echo "Docker is already installed on AMD64"
     exit 0
 fi
-# https://docs.docker.com/engine/security/rootless/
```
bebound, line 8: Not able to install rootless docker in 22.04 and 24.04.

Ubuntu use cgroup v2 after 22.04. Docker keeps raising `docker: Error response from daemon: open /sys/fs/cgroup/user.slice/user-1000.slice/cgroup.controllers: no such file or directory` when run docker, but `/sys/fs/cgroup/user.slice/` exists. I guess `user-1000` is logged in and this folder does not exist.

scripts/ci/install_docker.sh
```diff
@@ -5,11 +5,7 @@ if [[ $(dpkg --print-architecture) == "amd64" ]]; then
...
-sudo apt-get install -y uidmap
-dockerd-rootless-setuptool.sh install
-export XDG_RUNTIME_DIR=/home/cloudtest/.docker/run
-PATH=/usr/bin:/sbin:/usr/sbin:$PATH dockerd-rootless.sh &
-sleep 5
-docker context use rootless
+
+# https://docs.docker.com/engine/install/ubuntu/#install-using-the-convenience-script
+curl -fsSL https://get.docker.com/ | sh
+sudo chmod 666 /var/run/docker.sock
```
bebound, line 11: Use `chmod 666` to ensure `docker` can be used without `sudo` like runner-image agent.

In runner-image, this is achieved by adding user to `docker` group: https://github.com/actions/runner-images/blob/86ef9be8745af78c1e11cce2921532217216c90b/images/ubuntu/scripts/build/install-docker.sh#L49

But we can't re-login for it to take effect in the pipeline.

.azure-pipelines/templates/variables.yml
```diff
@@ -2,5 +2,5 @@ variables:
   ubuntu_pool: 'pool-ubuntu-2204'
   ubuntu_multi_core_pool: 'pool-ubuntu-latest-multi-core'
   windows_pool: 'pool-windows-2019'
-  ubuntu_arm64_pool: 'ubuntu-arm64-2004-pool'
+  ubuntu_arm64_pool: 'pool-ubuntu-latest-arm64'
```
jiasli, line 5: I remember we decided not to use the name `latest`?

scripts/ci/install_docker.sh
jiasli: `runner-image` is a general concept. I believe you are trying to mention [Microsoft-hosted agents](https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops#use-a-microsoft-hosted-agent).

jiasli: I prefer we stick to the example at https://docs.docker.com/engine/install/ubuntu/#install-using-the-convenience-script, instead of piping to `| sh` in case something unexpected happens:

image

jiasli: > Use `chmod 666` to ensure `docker` can be used without `sudo` like runner-image agent.

Any reference for this?

scripts/ci/install_docker.sh
jiasli: The document https://docs.docker.com/engine/security/rootless/ still exists and is valid. Is `user-1000.slice`-not-existing an ADO-specific issue?

.azure-pipelines/templates/variables.yml
bebound: Using the latest version saves effort by avoiding changes to the pool reference in CI and release pipelines, but it loses the ability to revert back if we encounter some issue.

How about this, we create a temp pool with newer image when we need to update the agent, then test it in the main repository. If everything works fine, we'll update the latest pool with the newer image. This could eliminate the issue we met during the 20.04 deprecation.

@wangzelin007 for awareness

jiasli (approved):

scripts/ci/install_docker.sh
bebound: Yes. Some say `user-1000.slice` is created when a user logs in or has a running process, but I can't find any reference.

PR #31262, opened 2025-04-14
{Test} Add email replacer for cli scenario test

evelyn-ys:
**Related command**

**Description**

During recording tests, we may unintentionally record real email addresses in recording yaml files. This PR adds EmailAddressReplacer for ScenarioTests during live run to replace all microsoft email addresses with test@example.com

Please note that this replacer will only work during live run, so no existing recording files will be affected

**Testing Guide**

See the new test recording file `test_keyvault_private_endpoint_connection.yaml` in commit
To try by yourself, run `azdev test $TEST_NAME --live` with any test that may record real email address and see the new recording file

**History Notes**

---

### DIFF
--- a/src/azure-cli-testsdk/azure/cli/testsdk/utilities.py
+++ b/src/azure-cli-testsdk/azure/cli/testsdk/utilities.py
@@ -4,6 +4,7 @@
 # --------------------------------------------------------------------------------------------
 
 import os
+import re
 from contextlib import contextmanager
 
 from .scenario_tests import (create_random_name as create_random_name_base, RecordingProcessor)
@@ -207,6 +208,31 @@ class MSGraphNameReplacer(RecordingProcessor):
     def process_response(self, response):
         if response['body']['string']:
             response['body']['string'] = response['body']['string'].replace(self.test_name, self.mock_name)
+        return response
+
+
+class EmailAddressReplacer(RecordingProcessor):
+    """Replace email address like xxx@microsoft.com with test@example.com"""
+
+    EMAIL_REPLACEMENT = 'test@example.com'
+
+    def process_request(self, request):
+        return request
+
+    def process_response(self, response):
+        if response['body']['string']:
+            body = _byte_to_str(response['body']['string'])
+            pattern = r'[\w.%#+-]+[%40|@|_]microsoft.com'
+            index = 0
+            replaced_body = ''
+            for match in re.finditer(pattern, body):
+                start = match.start()
+                end = match.end()
+                replaced_body += body[index:start] + self.EMAIL_REPLACEMENT
+                index = end
+            if index < len(body):
+                replaced_body += body[index:]
+            response['body']['string'] = replaced_body
         return response
 
 
--- a/src/azure-cli-testsdk/azure/cli/testsdk/utilities.py
+++ b/src/azure-cli-testsdk/azure/cli/testsdk/utilities.py
@@ -216,23 +216,30 @@ class EmailAddressReplacer(RecordingProcessor):
 
     EMAIL_REPLACEMENT = 'test@example.com'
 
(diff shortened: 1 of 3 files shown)


yonzhan:

src/azure-cli-testsdk/azure/cli/testsdk/utilities.py
```diff
@@ -210,6 +211,38 @@ def process_response(self, response):
...
+
+    def _replace_email_address(self, text):
+        pattern = r'[\w.%#+-]+[%40|@|_]microsoft.com'
+        index = 0
+        replaced_text = ''
+        for match in re.finditer(pattern, text):
+            start = match.start()
+            end = match.end()
+            replaced_text += text[index:start] + self.EMAIL_REPLACEMENT
+            index = end
```
jiasli, line 227: It it possible to use [`re.sub`](https://docs.python.org/3/library/re.html#re.sub)?

src/azure-cli-testsdk/azure/cli/testsdk/utilities.py
```diff
@@ -210,6 +211,38 @@ def process_response(self, response):
         return response
 
 
+class EmailAddressReplacer(RecordingProcessor):
+    """Replace email address like xxx@microsoft.com with test@example.com"""
+
+    EMAIL_REPLACEMENT = 'test@example.com'
+
+    def _replace_email_address(self, text):
+        pattern = r'[\w.%#+-]+[%40|@|_]microsoft.com'
```
jiasli, line 220: Any source for this regex?

jiasli, line 220: For `test_microsoft.com#EXT#@AzureSDKTeam.onmicrosoft.com`, this regex only replaces the `test_microsoft.com` part. Should we consider replacing the full string `test_microsoft.com#EXT#@AzureSDKTeam.onmicrosoft.com` to avoid showing the tenant domain too?

evelyn-ys: The source is me😂 Any issue with this pattern?

evelyn-ys: If I loose the regex to recognize all string in the format of `xxx@xxx.xx`, then `test_microsoft.com#EXT#@AzureSDKTeam.onmicrosoft` can be replaced, but there will be many mismatch for example `ssh@10.172` and the recording file can't work.

I'm thinking that as long as the microsoft part is replaced, then the full string will be invalid so I restricted the pattern to recognize `[@|%40|_]microsoft.com` only

src/azure-cli-testsdk/azure/cli/testsdk/utilities.py
evelyn-ys: We are in a loop for iterating `text` so I'm afraid use `re.sub` to update `text` itself may cause trouble

evelyn-ys: Stupid me🤣

src/azure-cli-testsdk/azure/cli/testsdk/utilities.py
jiasli: I am curious how `%#+-` are chosen. Could you provide some examples? I understand there are email addresses like `Isaac.Newton@microsoft.com` which contains `[\w.]`.

evelyn-ys: for example, `v-jingszhang@microsoft.com` contains `-`. But for `%#+`, I don't have any sample on top of my mind

jiasli (approved):

PR #31323, opened 2025-04-23
{Role} Clean up linter exclusion rules

jiasli:
**Related command**
`az role`
`az ad`

**Description**
Clean up linter exclusion rules.

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/role/_params.py
+++ b/src/azure-cli/azure/cli/command_modules/role/_params.py
@@ -96,6 +96,7 @@ def load_arguments(self, _):
                         "'2017-12-31'). Default value is one year after current time")
         c.argument('key_value', arg_group='keyCredential',
                    help='the value for the key credentials associated with the application')
+        # Even in `az ad app update`, key_type and key_usage need to have default values if key_value is specified
         c.argument('key_type', arg_group='keyCredential',
                    help='the type of the key credentials associated with the application',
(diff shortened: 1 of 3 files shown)


yonzhan:

bebound (approved):

src/azure-cli/azure/cli/command_modules/role/_params.py
```diff
@@ -263,7 +264,6 @@ def load_arguments(self, _):
     with self.argument_context('ad user') as c:
         c.ignore('_subscription')
         c.argument('mail_nickname', help='mail alias. Defaults to user principal name')
-        c.argument('force_change_password_next_login', arg_type=get_three_state_flag(), help='Require the user to change their password the next time they log in. Only valid when --password is specified')
```
zhoxing-ms, line 266: May I aks removing this parameter looks like a breaking change, right? If so, why not indicate it in the PR title?

jiasli: Good question! `force_change_password_next_login` has been renamed to `force_change_password_next_sign_in`. `force_change_password_next_sign_in` is no longer used.

zhoxing-ms (approved):

src/azure-cli/azure/cli/command_modules/role/_params.py
zhoxing-ms: Got it, thanks

PR #31371, opened 2025-04-27
[Storage] ip addresses with host bits set should be validated

eric-therond:
**Related command**

```
az storage account network-rule add -g xxx --account-name xxx --ip-address 4.213.28.114/30
```

**Description**

It is possible through the azure portal to add ip addresses that have the host bits set but the az storage cli command doesn't accept that. 

When an existing account storage has ip addresses with host bits set (added through the portal for instance) and then az storage account network-rule is used to add / remove ip addresses, the command will always fail, even if the ip addresses passed in parameter are valids and don't have host bit sets.

```
az storage account network-rule add -g xxx --account-name xxx --ip-address 137.135.190.48
The command failed with an unexpected error. Here is the traceback:
4.213.28.114/30 has host bits set
```

Here 4.213.28.114/30 is an existing ip address added through the portal.

Similar error in the past in another component:
https://github.com/Azure/azure-cli/pull/25031

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/storage/_validators.py
+++ b/src/azure-cli/azure/cli/command_modules/storage/_validators.py
@@ -2341,7 +2341,7 @@ def validate_ip_address(namespace):
     if not ip_address:
         return
 
-    ip_address_networks = [ip_network(ip) for ip in ip_address]
+    ip_address_networks = [ip_network(ip, False) for ip in ip_address]
     for idx, ip_address_network in enumerate(ip_address_networks):
         for idx2, ip_address_network2 in enumerate(ip_address_networks):
             if idx == idx2:
--- a/src/azure-cli/azure/cli/command_modules/storage/operations/account.py
+++ b/src/azure-cli/azure/cli/command_modules/storage/operations/account.py
@@ -709,8 +709,8 @@ def add_network_rule(cmd, client, resource_group_name, account_name, action='All
         for ip in ip_address:
             to_modify = True
             for x in rules.ip_rules:
-                existing_ip_network = ip_network(x.ip_address_or_range)
-                new_ip_network = ip_network(ip)
+                existing_ip_network = ip_network(x.ip_address_or_range, False)
+                new_ip_network = ip_network(ip, False)
                 if new_ip_network.overlaps(existing_ip_network):
                     logger.warning("IP/CIDR %s overlaps with %s, which exists already. Not adding duplicates.",
                                    ip, x.ip_address_or_range)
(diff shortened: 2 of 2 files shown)


yonzhan:

calvinhzy:
Please add tests to this, I have tried creating 4.213.28.114/30 in portal and it was also not allowed. 4.213.28.112/30 is allowed.

eric-therond:
@calvinhzy look at this test case:

1/ Add x.x.x.x/x (IP address with host bits) by any means (in my case 4.213.28.114/30 was added via Terraform/API and not via the portal, sorry for the mistake).

2/ Now that the previous IP address has been added, try adding or removing any kind of IP (with or without host bits set) with the CLI tool, an error will occur, the storage network-rule configuration is no longer manageable via the CLI tool.

calvinhzy:
Hi @eric-therond, usually CLI follows portal user behavior and it seems they only allow the starting ip address with the netmask so 4.213.28.112/30 is allowed. Adding @necusjz on network module behavior.

necusjz:
> Hi @eric-therond, usually CLI follows portal user behavior and it seems they only allow the starting ip address with the netmask so 4.213.28.112/30 is allowed. Adding @necusjz on network module behavior.

afaik, we only support start ip.

eric-therond:
I found a solution in my Terraform/API script to only add "valid" IPs addresses to the storage account. So for me it's ok now.

I think there is still an inconsistency with the portal.

Let's say we have a storage account with the following configuration:

image

You can see that `4.213.28.114/30` is not valid but was added with success via Terraform/API.

Now let's say I want to add an IP, any kind ok IP, to the above existing configuration, for instance:  `176.141.98.215`

With the portal it works, but with the cli it fails:

```
az storage account network-rule add -g xxx --account-name xxxx --ip-address 176.141.98.215
The command failed with an unexpected error. Here is the traceback:
4.213.28.114/30 has host bits set
  File "/opt/homebrew/Cellar/azure-cli/2.71.0/libexec/lib/python3.12/site-packages/knack/cli.py", line 233, in invoke
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/azure-cli/2.71.0/libexec/lib/python3.12/site-packages/azure/cli/core/commands/__init__.py", line 666, in execute
    raise ex
  File "/opt/homebrew/Cellar/azure-cli/2.71.0/libexec/lib/python3.12/site-packages/azure/cli/core/commands/__init__.py", line 734, in _run_jobs_serially
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/azure-cli/2.71.0/libexec/lib/python3.12/site-packages/azure/cli/core/commands/__init__.py", line 703, in _run_job
    result = cmd_copy(params)

```

PR #31061, opened 2025-03-18
{Profile} `az login`: Announce `--username` breaking change for managed identity

jiasli:
**Related command**
`az login --identity --username`

**Description**
Migrate #30525 to the new breaking change announcement framework: https://github.com/Azure/azure-cli/blob/dev/doc/how_to_introduce_breaking_changes.md

**Testing Guide**
```
az login --identity --username 111
```

PR #31015, opened 2025-03-12
[Profile] BREAKING CHANGE: `az login`: Drop `--username` for managed identity authentication

jiasli:
**Related command**
`az login --identity`

**Description**
Follow-up of #30525, #31061
Part of #25959

**Testing Guide**
This should return an error:
```
az login --identity --username
TODO
```

**History Notes**

[Profile] BREAKING CHANGE: `az login`: `--username` no longer accepts user-assigned managed identity ID. Explicitly specify `--client-id`, `--object-id` or `--resource-id` instead

PR #31453, opened 2025-05-12
{Profile} `az login`: Add check for `--username` for managed identity authentication

jiasli:
**Related command**
`az login`

**Description**
https://github.com/Azure/azure-cli/pull/31015/ dropped `--username` for managed identity authentication. If `--username` is provided, it is silently ignored. This may lead to confusion.

This PR makes CLI fail when `--username` is provided.

The breaking change announcement for `--username` (#31061) is also removed.

**Testing Guide**
```
> az login --identity --username 111
Passing the managed identity ID with --username is no longer supported. Use --client-id, --object-id or --resource-id instead.
```


yonzhan:
Add check for `--username` for managed identity authentication

evelyn-ys (approved):

Issue #16459, opened 2021-01-08
[Enhancement Proposal] Support "bring your own access token"


jiasli:
**Is your feature request related to a problem? Please describe.**

We have received several feature requests that the user would like to provide their own access token, without interacting with AAD.

**Describe the solution you'd like**

1. `az login` should support either
   - `--access-token` argument which accepts an access token with [Subscriptions - List](https://docs.microsoft.com/en-us/rest/api/resources/subscriptions/list) permission:
     ```
     az login --access-token <access_token>
     ```
   - both `--access-token` (may not have [Subscriptions - List](https://docs.microsoft.com/en-us/rest/api/resources/subscriptions/list) permission) and `--subscription` which explicitly specifies the default subscription (also see #14933):
     ```
     az login --access-token <access_token> --subscription <subscription_id>
     ```
2. Each `az` command should support a global argument `--access-token` which can be used together with `--subscription` to invoke ARM request:
   ```
   az group list --access-token <access_token> --subscription <subscription_id>
   ```
   I previously made a prototype: https://github.com/jiasli/azure-cli/pull/12

yonzhan:
Enhancement Proposal

jiasli:
Azure CLI can take advantage of `TokenCredential` if https://github.com/Azure/azure-sdk/issues/2189 is implemented.

richeney:
That would be cool. especially if those `--subscription` and `--access-token` switches supported [Azure defaults](https://docs.microsoft.com/en-us/cli/azure/azure-cli-configuration#cli-configuration-values-and-environment-variables). You could then set env vars, e.g. `export AZURE_DEFAULTS_ACCESS_TOKEN=$mytoken`. 

I often have scrips that export AZURE_DEFAULTS_LOCATION and AZURE_DEFAULTS_GROUP to save including `--location` and `--resource-group` on every CLI command.

bganapa:
If you are looking for this option to run few cli commands parallelly with different user context, the option of AZURE_CONFIG_DIR may be helpful 
https://docs.microsoft.com/en-us/cli/azure/use-cli-effectively#concurrent-builds

jiasli:
## Limitations

1. The access token usually expires after 1 hour. Users will have to log in again after it expires.
2. The access token has a specific audience claim (`aud`), which means it can only be used on one resource. For example, if an ARM's access token is provided, Storage or Key Vault data-plane commands will fail as the audience doesn't match the resource.
[...]

Issue #31133, opened 2025-03-26
Install Azure CLI from source code


jiasli:
When a new feature is merged into `dev` branch, you may test its functionality by installing Azure CLI from source code:

### Windows

Prerequisite:

1. Install latest Python 3.12 from https://www.python.org/downloads/windows/
2. Install latest Git from https://git-scm.com/downloads

Run below commands:

```ps1
git clone https://github.com/Azure/azure-cli --depth 1
python -m venv azure-cli-env
. azure-cli-env\Scripts\activate.ps1
python.exe -m pip install --upgrade pip
pip install --no-deps --editable azure-cli/src/azure-cli-core azure-cli/src/azure-cli azure-cli/src/azure-cli-telemetry
pip install --requirement azure-cli/src/azure-cli/requirements.py3.windows.txt

# You may run az commands now
az --version
az login

# After testing, deactivate the virtual environment and delete these folders
deactivate
Remove-Item -Recurse -Force azure-cli, azure-cli-env
```

### Linux

Prerequisite: Install Python using the package manager of your Linux distribution.

Run below commands:

```sh
git clone https://github.com/Azure/azure-cli --depth 1
python3 -m venv azure-cli-env
. azure-cli-env/bin/activate
python -m pip install --upgrade pip
pip install --no-deps --editable azure-cli/src/azure-cli-core azure-cli/src/azure-cli azure-cli/src/azure-cli-telemetry
pip install --requirement azure-cli/src/azure-cli/requirements.py3.Linux.txt

# You may run az commands now
az --version
az login

# After testing, deactivate the virtual environment and delete these folders
deactivate
rm -rf azure-cli azure-cli-env
```

yonzhan:
Thank you for opening this issue, we will look into it.

jiasli:
This instruction originates from https://github.com/Azure/azure-cli/issues/16573#issuecomment-2488223313.

Issue #10241, opened 2019-08-14
Enable authentication via environment variables


garymoon:
**Is your feature request related to a problem? Please describe.**
I'm concerned that I cannot use the CLI without having my credentials written to disk in the clear.

**Describe the solution you'd like**
I would like the CLI to accept credentials from the environment (e.g. `AZURE_TENANT_ID`, `AZURE_CLIENT_ID`, `AZURE_CLIENT_SECRET`).

**Describe alternatives you've considered**
I've yet to find a way to authenticate that doesn't involve secrets being written to `~/.azure`.

**Additional context**
I came across [this](https://github.com/Azure/azure-cli/issues/9282) issue which seems to describe the same thing, but was closed following an apparent misunderstanding. While the AWS CLI does indeed write secrets in the clear to disk when configured via `aws configure`, it also allows for [configuration via environment variables](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html).

I've also found examples in both the Ruby SDK [here](https://github.com/Azure/azure-sdk-for-ruby#option-1---environment-variables), and the Go SDK [here](https://github.com/Azure/azure-sdk-for-go#more-authentication-details) where authentication via environment variables is enabled.

This issue is particularly problematic in a dockerized environment where it's common for the home directory to be mapped on the host.

jiasli:
Hi @garymoon, thank you for raising this issue. We will consider this as a feature request and discuss internally.

garymoon:
@jiasli Thank you for the update. Please also be sure to consider the implications of unintended credential leakage also. IMO this issue represents a security concern.

jiasli:
**Python SDK - Azure Identity** also supports environment variables [here](https://github.com/Azure/azure-sdk-for-python/tree/master/sdk/identity/azure-identity#environment-variables). Maybe we can implement this independently of Azure Identity integration. Even after the Azure Identity integration, this won't be a breaking change.

kostyay:
Any update on this? It will be super useful for us as well.

jiasli:
We are currently implementing this feature in the Azure Identity Integration process and it will be included in the next Beta release. But we can't provide a solid ETA yet.

etaham:
Any update?

jiasli:
This will be released in the next Beta version which uses MSAL. Hopefully it can satisfy your needs.

aliensouls:
Year 2021  Q2

[...]

PR #31575, opened 2025-05-29
[Core] Support bring-your-own access token

jiasli:
**Related command**

**Description**
Close https://github.com/Azure/azure-cli/issues/16459

Set `AZURE_CLI_ACCESS_TOKEN` environment variable to let Azure CLI 

1. bypass its own authentication context created by `az login`
2. use the specified access token

The caller is responsible for providing a valid access token satisfying the below requirements:

1. still be valid and not expired
2. has the correct `scope`

Azure CLI doesn't not perform any check on the access token, so the error may vary depending on the service response.

Optionally, set

- `AZURE_CLI_SUBSCRIPTION_ID` if the command requires a subscription ID, such as `az group list` which internally calls [Resource Groups - List](https://learn.microsoft.com/en-us/rest/api/resources/resource-groups/list?view=rest-resources-2021-04-01)
- `AZURE_CLI_TENANT_ID` if the command requires a tenant ID, such as `az keyvault create` which internally calls [Vaults - Create Or Update](https://learn.microsoft.com/en-us/rest/api/keyvault/keyvault/vaults/create-or-update?view=rest-keyvault-keyvault-2022-07-01&tabs=HTTP)

For more details on authentication via environment variables, see https://github.com/Azure/azure-cli/issues/10241.

**Testing Guide**
Install Azure CLI from source code: https://github.com/Azure/azure-cli/issues/31133

```sh
export AZURE_CLI_ACCESS_TOKEN=<access_token>
export AZURE_CLI_SUBSCRIPTION_ID=<subscription_id>
export AZURE_CLI_TENANT_ID=<tenant_id>

# No `az login` is required

# Confirm Azure CLI is using environment variable credential
az account show

az group list
```

**History Notes**

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/_profile.py
+++ b/src/azure-cli-core/azure/cli/core/_profile.py
@@ -49,6 +49,8 @@ _STATE = 'state'
 _USER_TYPE = 'type'
 _USER = 'user'
 _SERVICE_PRINCIPAL = 'servicePrincipal'
+_ACCESS_TOKEN_IDENTITY_TYPE = 'accessToken'
+_ACCESS_TOKEN_IDENTITY_NAME = 'ACCESS_TOKEN_ACCOUNT'
 _SERVICE_PRINCIPAL_CERT_SN_ISSUER_AUTH = 'useCertSNIssuerAuth'
 _TOKEN_ENTRY_USER_ID = 'userId'
 _TOKEN_ENTRY_TOKEN_TYPE = 'tokenType'
@@ -60,6 +62,11 @@ _USER_ASSIGNED_IDENTITY = 'userAssignedIdentity'
 _ASSIGNED_IDENTITY_INFO = 'assignedIdentityInfo'
 
 _AZ_LOGIN_MESSAGE = "Please run 'az login' to setup account."
+
+
+_AZURE_CLI_SUBSCRIPTION_ID = 'AZURE_CLI_SUBSCRIPTION_ID'
+_AZURE_CLI_TENANT_ID = 'AZURE_CLI_TENANT_ID'
+_AZURE_CLI_ACCESS_TOKEN = 'AZURE_CLI_ACCESS_TOKEN'
 
 
 def load_subscriptions(cli_ctx, all_clouds=False, refresh=False):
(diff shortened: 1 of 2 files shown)


src/azure-cli-core/azure/cli/core/_profile.py
```diff
@@ -62,6 +64,11 @@
 _AZ_LOGIN_MESSAGE = "Please run 'az login' to setup account."
 
 
+_AZURE_CLI_SUBSCRIPTION_ID = 'AZURE_CLI_SUBSCRIPTION_ID'
+_AZURE_CLI_TENANT_ID = 'AZURE_CLI_TENANT_ID'
+_AZURE_CLI_ACCESS_TOKEN = 'AZURE_CLI_ACCESS_TOKEN'
```
jiasli, line 69: I explicitly choose `AZURE_CLI_` prefix instead of `AZURE_` to avoid conflicting with other client tools, such as Azure SDK.

ThomasShih:
Thank you for this @jiasli! Just wondering how this would interact with `az login` as per originally proposed in https://github.com/Azure/azure-cli/issues/16459?

The reason why I'm asking is on our team, we are using sops to decrypt/encrypt from key vault. We would like to use the `az login` cli  to set up `~/.msal_token_cache.json`. We were originally dead in the water with trying the solution from https://github.com/Azure/login/issues/362#issuecomment-1788288204 because we can't guarantee az sdk and az cli version compatibility, but this pr might change things!

Please tell me we might be able to do something like `AZURE_CLI_ACCESS_TOKEN=<token> az login` so that it gets set up!

jiasli:
@ThomasShih, supporting bring-your-own access token will not require `az login`, because the access token will not be persisted to disk.

> We would like to use the `az login` cli to set up `~/.msal_token_cache.json`.

Please do not follow this path. As described in https://github.com/Azure/azure-cli/pull/19853#issuecomment-984440373, `msal_token_cache.json` is private to Azure CLI, and you should not read from this file. 

> Please tell me we might be able to do something like `AZURE_CLI_ACCESS_TOKEN=<token> az login` so that it gets set up!

You can directly read the access token from `AZURE_CLI_ACCESS_TOKEN` if it is available to your application, but we may change the name of this environment variable in the future.

Issue #31591, opened 2025-06-03
UserWarning: pkg_resources is deprecated as an API


MmAtBosch:
### Describe the bug

With the newest version of azure cli 2.74.0, our pipelines fail, as there is a warning printed to standard error:

### Related command

az storage account show-connection-string --name "$saName" --resource-group "$saGroupName" -o tsv 

### Errors

`/opt/az/lib/python3.12/site-packages/azure/multiapi/storagev2/fileshare/__init__.py:1: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.`

### Issue script & Debug output

/opt/az/lib/python3.12/site-packages/azure/multiapi/storagev2/fileshare/__init__.py:1: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  __import__('pkg_resources').declare_namespace(__name__)
##[debug]Agent environment resources - Disk: / Available 106948.53 MB out of 148675.55 MB, Memory: Used 2788.00 MB out of 15990.00 MB, CPU: Usage 28.51%
Storage connection string successfully retrieved, saving it to keyvault
##[debug]Agent environment resources - Disk: / Available 106948.32 MB out of 148675.55 MB, Memory: Used 2808.00 MB out of 15990.00 MB, CPU: Usage 4.72%

##[debug]Process exited with code 0 and signal null for tool '/usr/bin/bash'
##[debug]STDIO streams have closed and received exit code 0 and signal null for tool '/usr/bin/bash'
##[error]/opt/az/lib/python3.12/site-packages/azure/multiapi/storagev2/fileshare/__init__.py:1: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  __import__('pkg_resources').declare_namespace(__name__)
[...]

PR #31616, opened 2025-06-09
{Misc.} Remove `pkg_resources`

bebound:
**Description**

`pkg_resources` is deprecated and importing it in setuptools 80.9.0 raises this warning:
```
[root@1197773e3b76 /]# python3.12 -c 'import pkg_resources'
<string>:1: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
```

This PR removes its usage.

Related issue: https://github.com/Azure/azure-cli/issues/31591

### DIFF
--- a/doc/sphinx/azhelpgen/__init__.py
+++ b/doc/sphinx/azhelpgen/__init__.py
@@ -2,5 +2,3 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
-import pkg_resources
-pkg_resources.declare_namespace(__name__)
--- a/doc/sphinx/cligroup/__init__.py
+++ b/doc/sphinx/cligroup/__init__.py
@@ -2,5 +2,3 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
-import pkg_resources
-pkg_resources.declare_namespace(__name__)
--- a/tools/automation/verify/verify_packages.py
+++ b/tools/automation/verify/verify_packages.py
@@ -12,7 +12,7 @@ import glob
 import filecmp
 import logging
 import unittest
-from pkg_resources import working_set
+from importlib.metadata import distributions
 
 import automation.utilities.path as automation_path
 from automation.utilities.const import COMMAND_MODULE_PREFIX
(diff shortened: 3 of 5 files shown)


yonzhan:

doc/sphinx/azhelpgen/__init__.py
```diff
@@ -2,5 +2,3 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
-import pkg_resources
-pkg_resources.declare_namespace(__name__)
```
bebound, line 6: Not sure why it declares namespace package. This does not look like a namespace package at all.

doc/sphinx/cligroup/__init__.py
```diff
@@ -2,5 +2,3 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
-import pkg_resources
-pkg_resources.declare_namespace(__name__)
```
bebound, line 6: The `doc\sphinx\` folder appears related to documentation generation but hasn't been updated in a long time.

Confirmed with doc team, they only use `doc\sphinx\azhelpgen\doc_source_map.json`

tools/automation/verify/verify_packages.py
```diff
@@ -12,7 +12,7 @@
 import filecmp
 import logging
 import unittest
-from pkg_resources import working_set
+from importlib.metadata import distributions
```
bebound, line 15:
```
def distributions(**kwargs):
    """Get all ``Distribution`` instances in the current environment.

    :return: An iterable of ``Distribution`` instances.
    """
    return Distribution.discover(**kwargs)
```

Distribution doc: https://docs.python.org/3/library/importlib.metadata.html#distributions

jiasli: `importlib.metadata.distributions` is not even documented by https://docs.python.org/3.15/library/importlib.metadata.html

tools/automation/verify/verify_packages.py
```diff
@@ -48,7 +48,7 @@ def test_azure_cli_installation(self):
     def test_azure_cli_module_installation(self):
         expected_modules = set([n for n, _ in automation_path.get_command_modules_paths(include_prefix=True)])
 
-        installed_command_modules = [dist.key for dist in list(working_set) if dist.key.startswith(COMMAND_MODULE_PREFIX)]
+        installed_command_modules = [dist.metadata['Name'] for dist in distributions() if dist.metadata['Name'].startswith(COMMAND_MODULE_PREFIX)]
```
jiasli, line 51: `dist`'s type is `importlib.metadata.PathDistribution`, but this class is not documented by https://docs.python.org/3.15/library/importlib.metadata.html either.

Another way to access the name of the distribution is `dist.name`. Actually, `importlib.metadata.Distribution.name` is defined as

```py
    @property
    def name(self):
        """Return the 'Name' metadata for the distribution package."""
        return self.metadata['Name']
```

bebound:
After this PR, there is no need to install `setuptools` in CLI package for main repo: https://github.com/Azure/azure-cli/pull/27196

I'm unsure if any extensions still depend on it.

PR #31623, opened 2025-06-10
{CI} Update homebrew to latest version during formula test

bebound:
**Description**

Recently, azure-cli's formula added a new config `no_autobump! because: :requires_manual_review`, which is only compatible with homebrew>=4.5.4.

```
echo == Install azure-cli.rb formula ==
brew install --build-from-source $SYSTEM_ARTIFACTSDIRECTORY/homebrew/azure-cli.rb
== Install azure-cli.rb formula ==
Error: azure-cli: 'because' argument should use valid symbol or a string!
```

Update homebrew to the latest version in test script.

Ref:
- https://github.com/Homebrew/homebrew-core/pull/225001/files#diff-4defc7c88b431d79bf10a50aab619bb242b9d03673b3f0192b0651e5f078d975
- https://github.com/Homebrew/brew/pull/20021
- https://dev.azure.com/azclitools/public/_build/results?buildId=249293&view=logs&j=ebe8970d-a8af-5d7e-4086-8f4ce0be006b&t=3f22f999-770c-5d06-3bb8-1a936f5d8384

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -654,9 +654,9 @@ jobs:
 
   - bash: |
       set -ev
-      # Force relink python@3.xx in Homebrew to resolve the conflict with pre-installed python 3.xx on macOS-12 image
-      # See: https://github.com/Azure/azure-cli/issues/29054
-      brew unlink python@$3.12 && brew link --overwrite python@$3.12
+      # Force relink python@3.xx in Homebrew to resolve the conflict with pre-installed python 3.xx on macOS
+      # See: https://github.com/actions/runner-images/issues/9966
+      brew unlink python@3.12 && brew link --overwrite python@3.12
       brew unlink python@3.13 && brew link --overwrite python@3.13
       
       # update homebrew to latest version in case the formula is not compatible with the current homebrew version
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -659,6 +659,9 @@ jobs:
       python_version=3.12
       brew unlink python@$python_version && brew link --overwrite python@$python_version
       
+      # update homebrew to latest version in case the formula is not compatible with the current homebrew version
+      brew update
+      
       echo == Remove pre-installed azure-cli ==
       brew uninstall azure-cli
       
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -656,8 +656,8 @@ jobs:
       set -ev
       # Force relink python@3.xx in Homebrew to resolve the conflict with pre-installed python 3.xx on macOS-12 image
       # See: https://github.com/Azure/azure-cli/issues/29054
-      python_version=3.12
-      brew unlink python@$python_version && brew link --overwrite python@$python_version
+      brew unlink python@$3.12 && brew link --overwrite python@$3.12
+      brew unlink python@3.13 && brew link --overwrite python@3.13
       
       # update homebrew to latest version in case the formula is not compatible with the current homebrew version
       brew update


yonzhan:

azure-pipelines.yml
```diff
@@ -659,10 +659,16 @@ jobs:
...
       
+      # update homebrew to latest version in case the formula is not compatible with the current homebrew version
+      brew update
+      
       echo == Remove pre-installed azure-cli ==
       brew uninstall azure-cli
       
       echo == Install azure-cli.rb formula ==
+      # Do not upgrade installed dependents to avoid unrelated installation failure
+      export HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=1
```
bebound, line 670: `brew install` triggers some unrelated dependency bump if `HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK` is not set.
In this case, brew tries to update `php` too, but fails to start it.
```
==> Caveats
Bash completion has been installed to:
  /usr/local/etc/bash_completion.d

(CLI installation finished)

==> php
To enable PHP in Apache add the following to httpd.conf and restart Apache:
    LoadModule php_module /usr/local/opt/php/lib/httpd/modules/libphp.so

    <FilesMatch \.php$>
        SetHandler application/x-httpd-php
    </FilesMatch>

Finally, check DirectoryIndex includes index.php
    DirectoryIndex index.php index.html

The php.ini and php-fpm.ini file can be found in:
    /usr/local/etc/php/8.4/

To start php now and restart at login:
  brew services start php
Or, if you don't want/need a background service you can just run:
  /usr/local/opt/php/sbin/php-fpm --nodaemonize

##[error]Bash exited with code '1'.

```

Ref:
https://docs.brew.sh/Manpage#install-options-formulacask-
https://dev.azure.com/azclitools/public/_build/results?buildId=250839&view=logs&jobId=ebe8970d-a8af-5d7e-4086-8f4ce0be006b&j=ebe8970d-a8af-5d7e-4086-8f4ce0be006b&t=3f22f999-770c-5d06-3bb8-1a936f5d8384

azure-pipelines.yml
```diff
@@ -656,13 +656,17 @@ jobs:
       set -ev
       # Force relink python@3.xx in Homebrew to resolve the conflict with pre-installed python 3.xx on macOS-12 image
       # See: https://github.com/Azure/azure-cli/issues/29054
-      python_version=3.12
-      brew unlink python@$python_version && brew link --overwrite python@$python_version
+      brew unlink python@$3.12 && brew link --overwrite python@$3.12
+      brew unlink python@3.13 && brew link --overwrite python@3.13
```
bebound, line 660: `python@3.13` is a dependency of azure-cli:
`==> Installing dependencies for azure-cli: libgit2, sqlite, python@3.13, z3, llvm, rust and python@3.12`

`python@3.13.4` [just released](https://github.com/Homebrew/homebrew-core/blob/3c4ff4f035b0e4315500311940feec7f66dd7f4c/Formula/p/python@3.13.rb), and brew tries to update it. We need to manually overwrite it too. 

Dependency chain `cryptography` -> `rust` -> `z3` -> `python@3.13`

jiasli: Why is there a dollar sign in `$3.12`?

bebound: Typo. Weird, the task still succeeds. This line's exit code is 1.
```
brew unlink python@$3.12 && brew link --overwrite python@$3.12
Error: No such keg: /usr/local/Cellar/python@.12
```

--------------

The `&&` can suppress `set -e`:

``` bash
set -e
false && false
echo "wat"

>>> wat
```

azure-pipelines.yml
```diff
@@ -654,10 +654,13 @@ jobs:
 
   - bash: |
       set -ev
-      # Force relink python@3.xx in Homebrew to resolve the conflict with pre-installed python 3.xx on macOS-12 image
-      # See: https://github.com/Azure/azure-cli/issues/29054
-      python_version=3.12
-      brew unlink python@$python_version && brew link --overwrite python@$python_version
+      # Force relink python@3.xx in Homebrew to resolve the conflict with pre-installed python 3.xx on macOS
+      # See: https://github.com/actions/runner-images/issues/9966
```
bebound, line 658: One year later, the problem remains unresolved. Update the link to the runner-image issue.

azure-pipelines.yml
jiasli: See https://stackoverflow.com/questions/25794905/why-does-set-e-true-false-true-not-exit

jiasli: In `false && false`, the first `false` will not cause `set -e` to fail, and the second `false` is not executed. `true && false` will cause `set -e` to fail.

See

- https://stackoverflow.com/questions/25794905/why-does-set-e-true-false-true-not-exit
- https://stackoverflow.com/questions/40532462/why-doesnt-set-e-cause-a-failure-with-false-false-true

bebound:
The agent is using macOS 14 `20250609.1326` now, which has homebrew 4.5.4.

Close this PR.

Ref: https://github.com/actions/runner-images/blob/a0e2400abd3bc3584080d24a2cc2807f99de41cd/images/macos/macos-14-Readme.md

PR #31741, opened 2025-07-01
[Core] Add global policy arguments `--acquire-policy-token` and `--change-reference` to support Change Safety feature

evelyn-ys:
**Related command**

All cli commands except for `show/list` commands

**Description**

Azure Policy now supports fetching data from external endpoints to make compliance decisions. For resources which are applied with such new policy, they will require an extra [acquirePolicyToken](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/resources/resource-manager/Microsoft.Authorization/stable/2025-03-01/policyTokens.json) request before creating/updating/deleting operations.

This PR adds two global arguments for all cli commands except for `list`/`show`:
- `--acquire-policy-token` to allow an extra `acquirePolicyToken` request
- `--change-reference` as `acquirePolicyToken` request parameter if the new policy defines Change Safety validation as external endpoint

It's also supported to turn on/off this feature by
1. run cli command `az config set core.enable_policy_token=true/false`
2. set environment variable `AZURE_CORE_ENABLE_POLICY_TOKEN=True/False`

By default the feature flag is `false`

**Testing Guide**

1. Apply a new Policy which requires Change Safety validation for deletion on your resource
2. Prepare change safety resource and save change reference
4. Delete your resource directly - failure expected
5. Delete your resource with `--acquire-policy-token` - failure expected
6. Delete your resource with `--acquire-policy-token --change-reference $ChangeReferenceId` - success

**History Notes**

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/commands/arm.py
+++ b/src/azure-cli-core/azure/cli/core/commands/arm.py
@@ -219,7 +219,7 @@ def register_global_policy_argument(cli_ctx):
                 'action': ChangeReferenceAction,
             }
             acquire_policy_token_kwargs = {
-                'help': 'Acquiring policy token automatically for this resource operation',
+                'help': 'Acquiring an Azure Policy token automatically for this resource operation',
                 'arg_group': 'Global Policy',
                 'nargs': 0,
                 'action': AcquirePolicyTokenAction,
--- a/src/azure-cli-core/azure/cli/core/sdk/policies.py
+++ b/src/azure-cli-core/azure/cli/core/sdk/policies.py
@@ -145,10 +145,6 @@ def get_custom_hook_policy(cli_ctx):
                 policy_token = response_content.get('token', None)
                 if not policy_token:
                     raise ServiceError(f"No token returned. Response:{acquire_policy_token_response.content}")
-            elif acquire_policy_token_response.status_code == 202:
-                # TODO: Handle async token acquisition after Service is ready
-                raise NotImplementedError("Asynchronous policy token acquisition is not supported in current Azure CLI."
-                                          " Please upgrade and retry.")
         except Exception as ex:
             raise ServiceError(f"Failed to acquire policy token, exception: {ex}")
         if policy_token:


yonzhan:

src/azure-cli-core/azure/cli/core/sdk/policies.py
```diff
@@ -103,3 +103,60 @@ def on_request(self, request):
         super().on_request(request)
         from azure.cli.core.telemetry import set_user_agent
         set_user_agent(request.http_request.headers[self._USERAGENT])
+
+
+# pylint: disable=line-too-long
+def get_custom_hook_policy(cli_ctx):
```
jiasli, line 109: Ideally, this policy should be implemented by Python SDK. Otherwise, SDK users will have to implement this policy by themselves, which is too complicated.

evelyn-ys: True. SDK team need long period to update all their mgmt plane SDKs. So we have to make it by ourselves to unblock CLI support in short term for this feature.

But I'm keeping following the communication and collaboration with SDK team

src/azure-cli-core/azure/cli/core/commands/arm.py
```diff
@@ -187,6 +187,52 @@ def resource_exists(cli_ctx, subscription, resource_group, name, namespace, type
...
+                'action': ChangeReferenceAction,
+            }
+            acquire_policy_token_kwargs = {
+                'help': 'Acquiring an Azure Policy token automatically for this resource operation',
+                'arg_group': 'Global Policy',
+                'nargs': 0,
+                'action': AcquirePolicyTokenAction,
+            }
+            command.add_argument('_change_reference', '--change-reference', **change_reference_kwargs)
+            command.add_argument('_acquire_policy_token', '--acquire-policy-token', **acquire_policy_token_kwargs)
```
jiasli, line 228: Nitpicking: `change_reference_kwargs` and `acquire_policy_token_kwargs` are not reused. Maybe putting them inline for better readability?

I guess it is from

https://github.com/Azure/azure-cli/blob/05a5d8fa878d8ae4d2ee8a1294feedc8af83fedb/src/azure-cli-core/azure/cli/core/commands/arm.py#L230-L242

https://github.com/Azure/azure-cli/blob/05a5d8fa878d8ae4d2ee8a1294feedc8af83fedb/src/azure-cli-core/azure/cli/core/commands/arm.py#L376-L387

src/azure-cli-core/azure/cli/core/commands/arm.py
```diff
@@ -187,6 +187,52 @@ def resource_exists(cli_ctx, subscription, resource_group, name, namespace, type
...
+
+            change_reference_kwargs = {
+                'help': 'The related change reference ID for this resource operation',
+                'arg_group': 'Global Policy',
+                'action': ChangeReferenceAction,
+            }
+            acquire_policy_token_kwargs = {
+                'help': 'Acquiring an Azure Policy token automatically for this resource operation',
+                'arg_group': 'Global Policy',
+                'nargs': 0,
```
jiasli, line 224: According to https://docs.python.org/3/library/argparse.html#nargs, `nargs=0` can indeed be used, but `store_true` is preferred: https://docs.python.org/3/library/argparse.html#action

src/azure-cli-core/azure/cli/core/commands/arm.py
```diff
@@ -187,6 +187,52 @@ def resource_exists(cli_ctx, subscription, resource_group, name, namespace, type
...
+        class AcquirePolicyTokenAction(argparse.Action):  # pylint:disable=too-few-public-methods
+
+            def __call__(self, parser, namespace, value, option_string=None):
+                # save change reference to CLI context
+                cmd = getattr(namespace, 'cmd', None) or getattr(namespace, '_cmd', None)
+                cmd.cli_ctx.data['_acquire_policy_token'] = True
+
+        for command in command_table.values():
+            if command.name.split()[-1] in ['list', 'show']:
+                continue
```
jiasli, line 214: If we only exclude `list` and `show`, commands such as `az ad sp create-for-rbac`, `az role assignment list-changelogs` will also have these arguments, which is invalid.

It is indeed difficult to know which commands support `--acquire-policy-token` solely judging from the command name.

src/azure-cli-core/azure/cli/core/commands/arm.py
```diff
@@ -187,6 +187,52 @@ def resource_exists(cli_ctx, subscription, resource_group, name, namespace, type
...
+        command_table = kwargs.get('commands_loader').command_table
+
+        if not command_table:
+            return
+
+        class ChangeReferenceAction(argparse.Action):  # pylint:disable=too-few-public-methods
+
+            def __call__(self, parser, namespace, value, option_string=None):
+                # save change reference to CLI context
+                cmd = getattr(namespace, 'cmd', None) or getattr(namespace, '_cmd', None)
```
jiasli, line 202: I guess this line is from

https://github.com/Azure/azure-cli/blob/4a15044c9d14c0ea7d207f5defbecffec19831bd/src/azure-cli-core/azure/cli/core/commands/__init__.py#L1521

Why can there be both `cmd` and `_cmd`? Shouldn't there be only one?

src/azure-cli-core/azure/cli/core/commands/arm.py
jiasli, line 1: The last time I reviewed this file was 6 years ago (https://github.com/Azure/azure-cli/pull/11321). 😔

src/azure-cli-core/azure/cli/core/commands/arm.py
evelyn-ys: Because  I need to store the value to `cli_ctx.data['_acquire_policy-token']` instead of to `namespace.acquire_policy_token`  so I've defined `action: AcquirePolicyTokenAction` already. That's why I'm not using `action: store_true`.

I defined `nargs=0` and set the value always to `True` in `AcquirePolicyTokenAction` to achieve the same as `store_true` action 😉

src/azure-cli-core/azure/cli/core/commands/arm.py
evelyn-ys: I have the same question with you haha. But I don't dare to change the code

src/azure-cli-core/azure/cli/core/commands/arm.py
evelyn-ys: That's exactly the same discussion I had with service team and they said they want the most common GET operations being excluded. So I only handled show and list commands

bebound (approved):

jiasli (approved):

kairu-ms (approved):

isra-fel:
Hi @evelyn-ys I appreciate the detailed PR description and discussion with every reviewer! I just have one question:

Since `--change-reference` is necessary to enable this scenario, but `--acquire-policy-token` doesn't play an important role, how about we simplify the UX by combining them in one?

evelyn-ys:
@isra-fel  `--change-reference` is not required. There're many policies that can be applied to resources and change safety is just one of them. For all other policies, `--change-reference` is not needed at all.

isra-fel:
Thanks @evelyn-ys I understand now `--change-reference` is for one scenario.
I was trying to say if we can remove `--acquire-policy-token` because it can be implied by `--change-reference` (and possibly other parameters in the future).

evelyn-ys:
@isra-fel There're scenarios that only need `--acquire-policy-token`, no extra parameters needed. So there has to be at least a `--acquire-policy-token`

Issue #31529, opened 2025-05-21
az keyvault key show command returns different attribute names and format since v2.73.0 upgrade


fidelcasto:
### Describe the bug

Using previous versions of Azure CLI, the 'az keyvault key show' command returned a set of properties including the 'expires' attribute.  Since updating to version 2.73.0, the property is called 'exp'.  I also noticed the dates are not displayed as before also.  Is this expected behavior?  If so, shouldn't there be any release notes regarding this?

### Related command

az keyvault key show

### Issue script & Debug output

attributes    : @{created=1747845872; enabled=True; exp=1811522587; exportable=False; hsmPlatform=0;
                recoverableDays=90; recoveryLevel=Recoverable; updated=1747845872}

### Expected behavior

attributes    : @{created=2025-05-21T16:44:32+00:00; enabled=True; expires=2027-05-28T16:43:07+00:00;
                exportable=False; hsmPlatform=0; notBefore=; recoverableDays=90; recoveryLevel=Recoverable;
                updated=2025-05-21T16:44:32+00:00}

### Environment Summary

azure-cli                         2.73.0

core                              2.73.0
telemetry                          1.1.0

Extensions:
account                            0.2.5

Dependencies:
msal                              1.32.3
azure-mgmt-resource               23.3.0

Python location 'C:\Program Files (x86)\Microsoft SDKs\Azure\CLI2\python.exe'
Config directory 'C:\Users\...]\.azure'
Extensions directory 'C:\Users\[...]\.azure\cliextensions'

Python (Windows) 3.12.8 (tags/v3.12.8:2dc476b, Dec  3 2024, 19:07:15) [MSC v.1942 32 bit (Intel)]

yonzhan:
Thank you for opening this issue, we will look into it.

evelyn-ys:
Thanks for reporting. We will fix the date formatting.

PR #30339, opened 2024-11-14
{Core} Add dict transformation for typespec generated SDKs

evelyn-ys:
**Related command**

**Description**

SDK models have changed when we move from swagger generated to typespec generated. This cause issues for `knack.util.todict` to do the formatter over command output. We call https://github.com/Azure/azure-cli/blob/eb55296b4f3280b0df1d557de99f6f9b828180b3/src/azure-cli-core/azure/cli/core/commands/__init__.py#L718 to convert the command result for easier format. Here's how [todict](https://github.com/microsoft/knack/blob/b1164b57d777711fd3954cf11f18ffb06ab1ca5b/knack/util.py#L133-L156) is implemented in `knack`:
```Python
def todict(obj, post_processor=None):  # pylint: disable=too-many-return-statements
    """
    Convert an object to a dictionary. Use 'post_processor(original_obj, dictionary)' to update the
    dictionary in the process
    """
    if isinstance(obj, dict):
        result = {k: todict(v, post_processor) for (k, v) in obj.items()}
        return post_processor(obj, result) if post_processor else result
    if isinstance(obj, list):
        return [todict(a, post_processor) for a in obj]
    if isinstance(obj, Enum):
        return obj.value
    if isinstance(obj, (date, time, datetime)):
        return obj.isoformat()
    if isinstance(obj, timedelta):
        return str(obj)
    if hasattr(obj, '_asdict'):
        return todict(obj._asdict(), post_processor)
    if hasattr(obj, '__dict__'):
        result = {to_camel_case(k): todict(v, post_processor)
                  for k, v in obj.__dict__.items()
                  if not callable(v) and not k.startswith('_')}
        return post_processor(obj, result) if post_processor else result
    return obj
```
Currently working with **swagger generated models**, it runs into `hasattr(obj, '__dict__')` and transforms `obj.__dict__.items()` one by one if the item isn't private starting with `_`.

But now for **typespec generated models**, a new [base model](https://github.com/mccoyp/azure-sdk-for-python/blob/000f7af897937c8146dea5da28cd74dd4ef0f16e/sdk/keyvault/azure-keyvault-securitydomain/azure/keyvault/securitydomain/_model_base.py#L493) is designed to store all data in `__dict__['_data']`. In such case, we won't get anything by iterating `__dict__` as before. The new model provides a `as_dict` func instead to transform the model object to dict.

Here's an example for **swagger generated models**(old) and **typespec generated models**(new)
```
SecurityDomainUploadStatus (old)
[...]

PR #31775, opened 2025-07-08
{Core} Adopt serialization func from azure-core for CLI's `todict` formatting

evelyn-ys:
**Related command**

All cli commands

**Description**

Previously we added dict transformation for typespec generated SDK models in https://github.com/Azure/azure-cli/pull/30339, we called typespec generated SDK model's `as_dict` func to get each attribute name and attribute value. But this still causes issues like https://github.com/Azure/azure-cli/issues/31529 reported when
- the attribute is renamed during SDK generation
- the attribute has some formatter defined in the SDK model

We can take the `expired` attribute from [Keyvault SDK](https://github.com/Azure/azure-sdk-for-python/blob/d85023107fade43aa9a05f46ce4c1ae8b73597ed/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_generated/models/_models.py#L410-L412) as an example:
```Python
    expires: Optional[datetime.datetime] = rest_field(
        name="exp", visibility=["read", "create", "update", "delete", "query"], format="unix-timestamp"
    )
```
The `expires` is **renamed** from `exp` in http response and has `unix-timestamp` **time format**. By calling `as_dict`
- we are expecting to get `{'expires': '2027-05-28T16:43:07+00:00'}` (from **swagger** generated SDK model)
- but we actually get `{'exp': 1811522587}` (from **typespec** generated SDK model).

**The renaming and formatting was not applied to `as_dict` and this behavior applies for all typespec generated SDK models.**

To resolve this breaking change, SDK team added related serialization support in azure-core with https://github.com/Azure/azure-sdk-for-python/pull/41445, https://github.com/Azure/azure-sdk-for-python/pull/41571. They now have two functions to handle both **swagger** generated SDK and **typespec** generated SDK:
- `is_generated_model` to identify auto generated SDK models
- `attribute_list` to list all attribute names which are after renaming
- we can call `getattr(obj, attr)` for the attribute we get from `attribute_list` to get the attribute value which is after formatting

This PR changes the dict formatting for typespec generated SDK models with new `attribute_list` func. We didn't adopt `is_generated_model` because we don't want to change swagger generated SDK models' dict formatting to avoid any regressions.

**Testing Guide**

**History Notes**

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/util.py
+++ b/src/azure-cli-core/azure/cli/core/util.py
@@ -633,7 +633,7 @@ def todict(obj, post_processor=None):
     """
     from datetime import date, time, datetime, timedelta
     from enum import Enum
-    from azure.core.serialization import is_generated_model, attribute_list
+    from azure.core.serialization import attribute_list
     if isinstance(obj, dict):
         result = {k: todict(v, post_processor) for (k, v) in obj.items()}
         return post_processor(obj, result) if post_processor else result
@@ -647,10 +647,12 @@ def todict(obj, post_processor=None):
         return str(obj)
     # This is the only difference with knack.util.todict because for typespec generated SDKs
     # The base model stores data in obj.__dict__['_data'] instead of in obj.__dict__
+    # The way to detect if it's a typespec generated model is to check the private `_is_model` attribute
     # azure-core provided new function `attribute_list` to list all attribute names
     # so that we don't need to use raw __dict__ directly
-    if hasattr(obj, 'as_dict') and not hasattr(obj, '_attribute_map'):
-        result = {to_camel_case(attr): todict(getattr(obj, attr), post_processor) for attr in attribute_list(obj) if hasattr(obj, attr)}
+    if getattr(obj, "_is_model", False):
+        result = {to_camel_case(attr): todict(getattr(obj, attr), post_processor)
+                  for attr in attribute_list(obj) if hasattr(obj, attr)}
         return post_processor(obj, result) if post_processor else result
     if hasattr(obj, '_asdict'):
         return todict(obj._asdict(), post_processor)
--- a/src/azure-cli-core/azure/cli/core/util.py
+++ b/src/azure-cli-core/azure/cli/core/util.py
@@ -645,17 +645,12 @@ def todict(obj, post_processor=None):
         return obj.isoformat()
     if isinstance(obj, timedelta):
         return str(obj)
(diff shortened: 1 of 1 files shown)


yonzhan:

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -644,11 +645,17 @@ def todict(obj, post_processor=None):
...
+    # This is the only difference with knack.util.todict
+    # azure-core has provided
+    #   - `is_generated_model` to identify auto generated SDK models
+    #   - `attribute_list` to list all attribute names
+    # no matter it's track1/track2/swagger generated/typespec generated SDKs
+    # no matter what transformation (renaming/formatting/...) the attribute has during auto generation
+    if is_generated_model(obj):
+        result = {}
+        for attr in attribute_list(obj):
+            if hasattr(obj, attr):
```
evelyn-ys: We need to check whether the attr exist because we may import SDK model and then do some customization like keyvault certificate L122-127 https://github.com/Azure/azure-cli/blob/bb2b31a832033e1b6e1343cc49b8ec023cd537fb/src/azure-cli/azure/cli/command_modules/keyvault/custom.py#L69-L128

Some attributes are deleted

src/azure-cli-core/azure/cli/core/util.py
```diff
@@ -646,9 +647,12 @@ def todict(obj, post_processor=None):
         return str(obj)
     # This is the only difference with knack.util.todict because for typespec generated SDKs
     # The base model stores data in obj.__dict__['_data'] instead of in obj.__dict__
-    # We need to call obj.as_dict() to extract data for this kind of model
-    if hasattr(obj, 'as_dict') and not hasattr(obj, '_attribute_map'):
-        result = {to_camel_case(k): todict(v, post_processor) for k, v in obj.as_dict().items()}
+    # The way to detect if it's a typespec generated model is to check the private `_is_model` attribute
+    # azure-core provided new function `attribute_list` to list all attribute names
+    # so that we don't need to use raw __dict__ directly
+    if getattr(obj, "_is_model", False):
```
kairu-ms, line 653: How did it handle the nested case? When the internal property is also an object？

evelyn-ys: we will call `todict` recursively

kairu-ms (approved):

PR #31792, opened 2025-07-10
[CosmosDB] `az cosmosdb restore`: Skipping client side restore validation check

dsapaliga:
**Related command**
az cosmosdb restore

**Description**
This is a simple fix to remove the validation check for restorable accounts when the customer uses the powershell restore command. Couple of reason for removing this:
- Due to the mismatch in location format given as input vs how we validate the permissions.
- Avoid timeout during this validation check for large account restores.

This change is already part of cli-preview, updating this for cli stable now.
cli-preview PR - https://github.com/Azure/azure-cli-extensions/pull/8533/files

**Testing Guide**

az cosmosdb restore -n "" -g "" -a "" --restore-timestamp 2025-07-21T11:51:19Z  --source-backup-location "westus" --location "northcentralus"

**History Notes**

[CosmosDB] `az cosmosdb restore`: Remove restore validations which would cause timeouts for large restore and incorrect error messages

---

### DIFF
--- a/src/azure-cli/HISTORY.rst
+++ b/src/azure-cli/HISTORY.rst
@@ -3,12 +3,6 @@
 Release History
 ===============
 
-2.75.1
-++++++
-**Cosmos DB**
-
-* `az cosmosdb restore`: Remove restore validations which would cause timeouts for large restore and incorrect error messages
-
 2.75.0
 ++++++
 


yonzhan:

dsapaliga:
@yonzhan Please let me know if this is the correct approach. Since this change is only client side and does not require API/swagger changes, wanted to check this in for the next milestion of 07/29 and deployment on 08/05.
- How do I update the version of the cli so the customer can use the newer version?
- Any other steps to be performed on my end before check in?

evelyn-ys:
@dsapaliga Please fix CI failure

evelyn-ys:
@microsoft-github-policy-service agree company="Microsoft"

dsapaliga:
@yonzhan can you help run the pipelines again, pushed a new iteration, thank you.

dsapaliga:
Please help run the pipelines and also suggest a fix for this error of version update in HISTORY, if this is not the right way, how do i update the version?

image

src/azure-cli/HISTORY.rst
```diff
@@ -3,6 +3,12 @@
 Release History
 ===============
 
+2.75.1
+++++++
+**Cosmos DB**
```
calvinhzy: Remove this whole section as this should be auto generated by CI.

vinhms (approved):

src/azure-cli/azure/cli/command_modules/cosmosdb/custom.py
```diff
@@ -1864,61 +1862,6 @@ def cli_cosmosdb_restore(cmd,
     if target_restorable_account is None:
         raise CLIError("Cannot find a database account with name {} that is online at {}".format(account_name, restore_timestamp))
```
AgarwalVivek, line 1864: nit: wondering if we should log skipping restorable resources check. inorder to use that feature, please go to previous version? We can mention each come with its own cost? Feel free to close if this would be a overdo.

AgarwalVivek (approved):

src/azure-cli/azure/cli/command_modules/cosmosdb/custom.py
dsapaliga: For long term I am not sure if it would work that well when there are multiple versions built on top of this. 
Typically customer does not care if we do the restorable check before calling our backend or after. So, for the customer its not really a feature. 
It is a mechanism we built to fail quickly for incorrect restore requests. Leaving it as is.

evelyn-ys (approved):

PR #31824, opened 2025-07-16
[Container] `az container create`: Remove default values for container group to support standby pool reuse scenario

drishtipeshwani09:
**Related command**
The changes are related to this command - `az container create`

**Description**

- Ideally, we should not be initializing any default values on our end for the standby pool reuse scenario case which was introduced in API Version - 2024-05-01-preview. 
- However, a bug was identified where we are still setting up values for properties like ports, protocol and ip address which has been fixed. 
- This is an expected behavior when standby pool profile id is provided. 

**Testing Guide**
Currently, while we try to reuse a cg from a given standby pool with delegated subnet, we get the below error - 

Code: BadRequest
Message: ContainerGroup properties other than config map are not allowed

This change would fix that, now if a customer runs the following command for example, they should be able to reuse a cg successfully - 

`az container create --resource-group rg --name ctn --container-group-profile-id {cgprofile-id} --container-group-profile-revision {revision-number} --standby-pool-profile-id {standbypoolprofile-id} --vnet {vnet-name/id} --subnet {subnet-name/id}`

**History Notes**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/container/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/container/custom.py
@@ -142,9 +142,18 @@ def create_container(cmd,
 
     standby_pool_profile_reference = _create_standby_pool_profile_reference(standby_pool_profile_id=standby_pool_profile_id, fail_container_group_create_on_reuse_failure=fail_container_group_create_on_reuse_failure)
 
-    ports = ports or [80]
-    protocol = protocol or ContainerGroupNetworkProtocol.tcp
+    # No default values need to be set for the standbypool reuse scenario
+    if standby_pool_profile_id != None:
+        vnet_address_prefix = None
+        subnet_address_prefix = None
+        
+    ports = ports
+    protocol = protocol
 
+    if standby_pool_profile_id == None:
+        ports = [80]
+        protocol = ContainerGroupNetworkProtocol.tcp
+    
     config_map = _create_config_map(config_map)
 
     container_resource_requirements = _create_resource_requirements(cpu=cpu, memory=memory)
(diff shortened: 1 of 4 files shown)


yonzhan:

zhoxing-ms:
Please add some test cases for this PR change

src/azure-cli/azure/cli/command_modules/container/custom.py
```diff
@@ -142,8 +142,17 @@ def create_container(cmd,
...
 
-    ports = ports or [80]
-    protocol = protocol or ContainerGroupNetworkProtocol.tcp
+    # No default values need to be set for the standbypool reuse scenario
+    if standby_pool_profile_id is not None:
+        vnet_address_prefix = None
+        subnet_address_prefix = None
+
+    ports = ports or None
+    protocol = protocol or None
```
zhoxing-ms: May I ask why design this assignment logic?

drishtipeshwani09: Thanks for pointing this out!! There was a mistake in this logic, I have modified this such that only when the standby pool profile id is null, and the port/protocol variables are null as well, we want to assign the variables default values, otherwise they will remain null.

src/azure-cli/azure/cli/command_modules/container/custom.py
```diff
@@ -142,8 +142,14 @@ def create_container(cmd,
...
-    ports = ports or [80]
-    protocol = protocol or ContainerGroupNetworkProtocol.tcp
+    # No default values need to be set for the standbypool reuse scenario
+    if standby_pool_profile_id is not None:
+        vnet_address_prefix = None
+        subnet_address_prefix = None
+
+    if standby_pool_profile_id is None:
+        ports = ports or [80]
+        protocol = protocol or ContainerGroupNetworkProtocol.tcp
```
zhoxing-ms:
```suggestion
    # No default values need to be set for the standbypool reuse scenario
    if standby_pool_profile_id is not None:
        vnet_address_prefix = None
        subnet_address_prefix = None
    else:
        ports = ports or [80]
        protocol = protocol or ContainerGroupNetworkProtocol.tcp
```

zhoxing-ms (approved):

jsntcy (approved):

PR #31861, opened 2025-07-23
[App Config] `az appconfig kv import`: Support importing key-values from AKS ConfigMap

RichardChen820:
**Related command**
az appconfig kv import --source aks

**Description**
Support importing key-values from AKS ConfigMap to provide a better user experience for migrating to use AppConfiguration as centralized configuration management in AKS.

**Testing Guide**
- Import from AKS by name
```
az appconfig kv import --source aks --aks-cluster <cluster-name> --configmap-name myconfigmap
```

- Import key values from a configmap in a specific namespace from AKS by ARM ID and
```
az appconfig kv import --source aks --aks-cluster <cluster-resource-id> --configmap-name myconfigmap --configmap-namespace mynamespace
```

- Import key values from a file-style configmap in a specific namespace with flattening.
```
az appconfig kv import --source aks --aks-cluster <clustername> --configmap-name myconfigmap --configmap-namespace mynamespace --format json --separator __
```

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/test_appconfig_kv_import_export_commands.py
+++ b/src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/test_appconfig_kv_import_export_commands.py
@@ -1239,12 +1239,8 @@ class AppConfigToAppConfigImportExportScenarioTest(ScenarioTest):
 
 
 
-class AppConfigKubernetesConfigMapImportScenarioTest(ScenarioTest):
+class AppConfigKubernetesConfigMapImportLiveScenarioTest(LiveScenarioTest):
     """Test suite for importing key-values from Kubernetes ConfigMaps using AKS RunCommand."""
-
-    def __init__(self, *args, **kwargs):
-        kwargs["recording_processors"] = kwargs.get("recording_processors", []) + [CredentialResponseSanitizer()]
-        super().__init__(*args, **kwargs)
 
     @AllowLargeResponse()
     @ResourceGroupPreparer(parameter_name_for_location='location')


yonzhan:

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
```diff
@@ -550,7 +551,40 @@ def __read_kv_from_file(
...
+
+    # convert to KeyValue list
+    key_values = []
+    for k, v in flattened_data.items():
+        if validate_import_key(key=k):
+            key_values.append(KeyValue(key=k, value=v))
+    return key_values
+
+
+def flatten_config_data(config_data, format_, content_type, prefix_to_add="", depth=None, separator=None):
```
albertofori: I would suggest we place a `__` in front of it per the convention in CLI

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
```diff
@@ -715,6 +743,160 @@ def __read_kv_from_app_service(
...
+        aks_client = cf_managed_clusters(cmd.cli_ctx)
+
+        # Command to get the ConfigMap and output it as JSON
+        command = f"kubectl get configmap {configmap_name} -n {namespace} -o json"
+
+        # Execute the command on the cluster
+        result = aks_runcommand(cmd, aks_client, aks_cluster["resource_group"], aks_cluster["name"], command_string=command)
+
+        if hasattr(result, 'logs') and result.logs:
+            if not hasattr(result, 'exit_code') or result.exit_code != 0:
```
albertofori: Is it possible that there won't be an exit code?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
```diff
@@ -715,6 +743,160 @@ def __read_kv_from_app_service(
...
+        prefix_to_add (str): Prefix to add to each key
+        content_type (str): Content type to apply to the key-values
+        format_ (str): Format of the data in the ConfigMap (e.g., "json", "yaml")
+        depth (int): Maximum depth for flattening hierarchical data
+        separator (str): Separator for hierarchical keys
+
+    Returns:
+        list: List of KeyValue objects
+    """
+    key_values = []
```
albertofori, line 831: Is this assignment needed?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
```diff
@@ -713,6 +740,156 @@ def __read_kv_from_app_service(
...
+                prefix_to_add=prefix_to_add,
+                depth=depth,
+                separator=separator
+            )
+
+            key_values = []
+            for k, v in flattened_data.items():
+                if validate_import_key(key=k):
+                    key_values.append(KeyValue(key=k, value=v))
+            return key_values
```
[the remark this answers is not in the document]
albertofori: I agree with this comment . Moreover, should the `return key_values` line be unindented?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
```diff
@@ -715,6 +743,160 @@ def __read_kv_from_app_service(
...
+            try:
+                configmap_data = json.loads(result.logs)
+
+                # Extract the data section which contains the key-value pairs
+                kvs = __extract_kv_from_configmap_data(
+                    configmap_data, content_type, prefix_to_add, format_, depth, separator)
+
+                key_values.extend(kvs)
+            except json.JSONDecodeError:
+                raise ValidationError(
```
albertofori: Is it possible for this JSON error to occur if we are getting output as JSON? 
Also, I would rather we use ValueError instead since the parsed payload is a response from the server and not user input

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
```diff
@@ -715,6 +743,160 @@ def __read_kv_from_app_service(
...
+    Returns:
+        list: List of KeyValue objects
+    """
+    try:
+        key_values = []
+        from azure.cli.command_modules.acs.custom import aks_runcommand
+        from azure.cli.command_modules.acs._client_factory import cf_managed_clusters
+
+        # Get the AKS client from the factory
+        cmd.cli_ctx.data['subscription_id'] = aks_cluster["subscription"]
```
albertofori, line 784: Does this mean we are allowing the imports from a cluster in a different subscription? I would expect this to be scoped to the user's subscription.

ChristineWanjau: Why should we limit it to the user's subscription?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
```diff
@@ -715,6 +743,160 @@ def __read_kv_from_app_service(
...
+        cmd.cli_ctx.data['safe_params'] = None
```
ChristineWanjau: Just curious why do we have to set the safe_params to None?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
albertofori: Actually, I think my comment can be ignored. 
I was initially under the impression that we only take cluster name as input, but user can also pass the ARM ID , similar to app service. So I guess there need for the restriction. 
Thanks for calling this out @ChristineWanjau !

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -176,6 +181,25 @@ def import_config(cmd,
         src_kvs = __read_kv_from_app_service(
             cmd, appservice_account=appservice_account, prefix_to_add=prefix, content_type=content_type)
 
+    elif source == 'aks':
+        if separator:
+            # If separator is provided, use max depth by default unless depth is specified.
```
ChristineWanjau, line 186: Should we let a user know this that we use max depth by default if depth is not provided maybe somewhere in the  params.py?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
ChristineWanjau: Correct me if I am wrong, aren't safe params like the global params e.g --debug or --output? Would setting them to None here affect the intended output?

src/azure-cli/azure/cli/command_modules/appconfig/_constants.py
```diff
@@ -144,6 +144,7 @@ class CompareFields:
 CompareFieldsMap = {
     "appconfig": (CompareFields.CONTENT_TYPE, CompareFields.VALUE, CompareFields.TAGS),
     "appservice": (CompareFields.VALUE, CompareFields.TAGS),
+    "aks": (CompareFields.VALUE, CompareFields.TAGS),
```
ChristineWanjau: Should we also compare the content type?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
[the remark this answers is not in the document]
RichardChen820: I agree, updated.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
RichardChen820: updated

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
RichardChen820: This needs input from AKS team @yonzhan, I would expect it always has exit_code in result, but not 100% sure, so I handle it in this way here.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
RichardChen820: It needs more input from AKS @yonzhan as well, for now, I handle the exception incase there's any unknown case.

src/azure-cli/azure/cli/command_modules/appconfig/_constants.py
RichardChen820: I think it should comply with the "appservice", what's the reason that appservice ignores the content type?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
RichardChen820: When we call the `aks_runcommand`, we must eliminate the influence of any parameters passed in by the customer in order to achieve stable and consistent format of result. Otherwise, it could lead to unpredictable outcomes that are difficult to handle.

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
RichardChen820: Do you mean [this](https://github.com/Azure/azure-cli/blob/8c1327851f0b180891f579cf07c7b169e9c7cccd/src/azure-cli/azure/cli/command_modules/appconfig/_params.py#L245)?

src/azure-cli/azure/cli/command_modules/appconfig/_constants.py
ChristineWanjau: I am not sure why we don't have it for appservice...Maybe @albertofori knows. I was just thinking since we can add content type to the kv maybe we should compare it?

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
ChristineWanjau: Yes this works! Hadn't seen it.

src/azure-cli/azure/cli/command_modules/appconfig/_constants.py
albertofori: To provide some background on the use of `CompareFields`, this specifies the applicable properties in the **destination** with which we compare the source values to determine whether a selected key-value for import/export will be added, deleted, or already exists and therefore will be updated accordingly. 

So for example, when exporting to app service, there is no concept of a content-type, there is only key, value, and a tag that specifies the slot setting. So when exporting, we only focus on those properties to determine the diff.

That being said, there are two things of note:
1. This conversation exposes an already existing issue with the current import functionality, i.e., we determine the properties to compare based on the source of the import config instead of the destination which, in the case of import, should always be `appconfig`. 
2. Since we are only importing from aks, we do not need to add compare fields entry for it as it is never a destination. In the `import_config` function, let's always specify the compare fields for the `appconfig`, not source.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
albertofori, line 784: In the event where the store and cluster are in different subscriptions, and we don't restore the original subscription value, wouldn't subsequent calls within the command that depend on `cli_ctx`, target the AKS cluster's subscription?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
albertofori: Same question as [this](https://github.com/Azure/azure-cli/pull/31861#discussion_r2246518923) comment; Perhaps, we should restore the safe params as they may also be used by subsequent calls within the import operation.

albertofori: I agree that we should pass down a different set of safe_params or remove ones that are not applicable, however, I believe it would be good to at least pass down the `--debug` parameter if user specifies it, as this will show the underlying details that might be of interest to the customer. We would not want to miss showing  information on the call made to aks when user has set the `--debug` flag.

Now that I think about it, is it practical to deep copy the context and edit it as you need so we don't need to restore any settings?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
albertofori: Maybe we can choose to error out if exit code is not 0. I don't think we should throw an exception if we don't find an error code. 

Thoughts?

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
RichardChen820: I agree, let's keep it, although it could reveal the debug details even if I clean the `--debug` param, not sure how it works, but let's do the thing in the right way.

src/azure-cli/azure/cli/command_modules/appconfig/_constants.py
RichardChen820: I'm a little bit confused, why `CompareFields` specifies the applicable properties in the destination? I think it's also applicable to the source.

Let's forget about AKS first, imagine import from app service, we also need to determine what fields to compare, since if a key-value has the same value and tags as what's in appconfig, we treat them equality and don't write it.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
RichardChen820: Fair enough, let's tolerate no exit_code

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
RichardChen820: The cli_ctx will not be used afterward since the appconfig client is already created before we modified the subscription. But from the function design perspective, this function should not have a side effect on the original data. I agree with your viewpoint, let's do the right thing.

src/azure-cli/azure/cli/command_modules/appconfig/_constants.py
albertofori: To clarify, I mean, the fields to compare should be dictated by the fields that are written in the destination. In your example, the destination is app config since it’s import. That means we also need to consider the user-specified content type in the command and compare that with that of the existing key-value else we might end up considering a key-value equal to one in app service and not write it though the content type user said  should be applied is different from what is already in the store

albertofori: But when exporting to app service, all we care about to find the key-values that already exist, is the value and the tags. App service does not care about or define content type in configs so there is no way to even compare with existing keys in the destination. So the fields to compare should always be dictated by fields present in where you are writing  the kvs to, hence the destination

RichardChen820: Is that what you mean that the code having an existing issue,  since currently we don't compare the content type in this appservice->appconfig scenario?

RichardChen820: Currently we compare in the same way for otherService -> appconfig and appconfig->otherService. Do you mean we need to differentiate them?

albertofori: Yes. Other service -> app config (import) should always compare all fields.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
```diff
@@ -715,6 +743,170 @@ def __read_kv_from_app_service(
...
+
+                    key_values.extend(kvs)
+                except json.JSONDecodeError:
+                    raise ValueError(
+                        f"The result from ConfigMap {configmap_name} could not be parsed. {result.logs.strip()}"
+                    )
+            else:
+                raise AzureResponseError(f"{result.logs.strip()}")
+        else:
+            raise AzureResponseError("Unable to get the ConfigMap.")
```
albertofori, line 815: Just curious, what is returned when there is no data in the ConfigMap? Would it be `{"logs": {}}`

src/azure-cli/azure/cli/command_modules/appconfig/_constants.py
RichardChen820: Does App Service configuration have the concept of tags? I assume it doesn’t. I'm asking because it is a bit tricky to me

Suppose a user exports config from App Configuration to App Service—information like tags and contentType would be lost during the process. When importing back, the user would have to explicitly know what the original tag and contentType values were in App Configuration and include them during import to ensure they are not overwritten.

From a user experience perspective, this doesn’t sound very reasonable.

RichardChen820: I think this issue needs further discussion. But since we don't support export to AKS, we should compare the value, tags, content type from day 0 for AKS

albertofori: > Does App Service configuration have the concept of tags? I assume it doesn’t. I'm asking because it is a bit tricky to me

They do not, the only concept applicable there is slotsetting, which we store in a tag during import.


> Suppose a user exports config from App Configuration to App Service—information like tags and contentType would be lost during the process. When importing back, the user would have to explicitly know what the original tag and contentType values were in App Configuration and include them during import to ensure they are not overwritten.

Sorry but I don't understand how this would be an issue. The concern is not about round-tripping between import /export, it is about doing the right comparison during import, especially when user is importing using an automated script. E.g., You have imported a kv from app service. You later go on to change the content-type of this kv. But then user wants to import kvs from app service again and set a common content-type  which is different from the one you set, this kv will not be overwritten if the value in app service has not changed. I do not think this is the right behaviour.

> I think this issue needs further discussion. But since we don't support export to AKS, we should compare the value, tags, content type from day 0 for AKS

I agree... Let's discuss this further offline. For now, keep the AKS entry in the compare field map with all fields. The update, if any needed, can be added in a follow-up PR.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
RichardChen820: aks_runcommand result for configMap does not exist

```
{'additional_properties': {}, 'id': '1234d4bebe9e4a1581ff23d07155edb1', 'provisioning_state': 'Succeeded', 'exit_code': 1, 'started_at': datetime.datetime(2025, 8, 4, 6, 0, 31, tzinfo=<isodate.tzinfo.Utc object at 0x00000221D5B87290>), 'finished_at': datetime.datetime(2025, 8, 4, 6, 0, 32, tzinfo=<isodate.tzinfo.Utc object at 0x00000221D5B87290>), 'logs': 'Error from server (NotFound): configmaps "configmap3" not found\n', 'reason': None}
```

RichardChen820: aks_runcommand result for configMap with no data

```
{'additional_properties': {}, 'id': '03ce213b8d4e473a92d82ed8b40e7f79', 'provisioning_state': 'Succeeded', 'exit_code': 0, 'started_at': datetime.datetime(2025, 8, 4, 5, 53, 26, tzinfo=<isodate.tzinfo.Utc object at 0x000001A838517350>), 'finished_at': datetime.datetime(2025, 8, 4, 5, 53, 27, tzinfo=<isodate.tzinfo.Utc object at 0x000001A838517350>), 'logs': '{\n    "apiVersion": "v1",\n    "kind": "ConfigMap",\n    "metadata": {\n        "annotations": {\n            "kubectl.kubernetes.io/last-applied-configuration": "{\\"apiVersion\\":\\"v1\\",\\"data\\":null,\\"kind\\":\\"ConfigMap\\",\\"metadata\\":{\\"annotations\\":{},\\"name\\":\\"configmap-demo\\",\\"namespace\\":\\"default\\"}}\\n"\n        },\n        "creationTimestamp": "2025-05-29T15:14:33Z",\n        "name": "configmap-demo",\n        "namespace": "default",\n        "resourceVersion": "280166617",\n        "uid": "68774c57-72a6-4052-9a94-99244d7cefc5"\n    }\n}\n', 'reason': None}
```

src/azure-cli/azure/cli/command_modules/appconfig/_constants.py
ChristineWanjau: I think Richard’s point about round-tripping make sense. If I'm onboarding into App Configuration from App service, and App Service doesn’t support content-type, I probably wouldn’t bother adding it to my config—because it’s not going to be used anyway. (I am assuming a user would now be pulling config from App config to their app service)

So in that case, it feels more practical to just compare the value and tags during import. If those are the only things App Service actually cares about, then that’s what we should focus on.

We could take a look at how often content type is being added in configs from app service. If it’s not showing up much, that’s even more reason to keep the comparison simple for App Service.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
```diff
@@ -715,6 +743,160 @@ def __read_kv_from_app_service(
...
+        format_ (str): Format of the data in the ConfigMap (e.g., "json", "yaml")
+        depth (int): Maximum depth for flattening hierarchical data
+        separator (str): Separator for hierarchical keys
+
+    Returns:
+        list: List of KeyValue objects
+    """
+    key_values = []
+
+    if 'data' not in configmap_data:
```
albertofori: Maybe refactor this to `if not config_map.get("data", None)` to handle the case where there is a "data" property but the value is `None`

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
albertofori: Thanks!

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
RichardChen820: Umm...in Kubernetes design philosophy, it doesn't explicit `null` field in returned object, especially in built-in object types like `ConfigMap`, so I don't expect it returns data in `None`. 

Although, I'm okay with this way which tolerate more situation if you think it's better :)

albertofori: Sounds good to me. Just wanted to make sure we do not hit a potential null ref when accessing the data items below

src/azure-cli/azure/cli/command_modules/appconfig/_constants.py
albertofori: Fair point, but I think it comes down to this question:
If a user has a kv with the same key as a config in App Service, and during import they specify a content type to be applied to the imported kvs, as long as the values of the config in App Service and the existing one in AppConfig match, should we consider them equal regardless of content type? In the event where user specifies `ignore-match`, this would also mean we do not overwrite that kv even if the content type is different from what user specified in the command.

yanzhudd:
please fix the CI issues.

RichardChen820:
@yonzhan What's the difference between the instances? https://dev.azure.com/azclitools/public/_build/results?buildId=265753&view=results

RichardChen820:
@yanzhudd @yonzhan could you please re-run the ADO pipeline to check if the failure has been fixed?

RichardChen820:
@yanzhudd The test was supposed to be fixed, could you please do a favor to try it again?

RichardChen820:
@yonzhan @yanzhudd Sorry for not so familiar with the vcr test. Just fixed the error, could you please trigger it again? Hope it's the last try.

RichardChen820:
@yonzhan @yanzhudd I mark the test to run live only, it will not fail

RichardChen820:
@yonzhan @yanzhudd Could you please help trigger the ADO CI again?

RichardChen820:
@yonzhan @yanzhudd It passes, please review this PR

PR #31865, opened 2025-07-23
[AppService] `az functionapp flex-migration`: Create commands to migrate CV1 apps to Flex

kamperiadis:
**Related command**
`az functionapp flex-migration list` 
`az functionapp flex-migration start`

**Description**
Linux Consumption customers have to run these steps manually in order to migrate their apps to Flex Consumption: https://learn.microsoft.com/en-us/azure/azure-functions/migration/migrate-plan-consumption-to-flex?tabs=azure-cli%2Clinux%2Csystem-assigned%2Ccontinuous%2Ctraces-table

With the `az functionapp flex-migration list` command, customers will get a list of the Linux Consumption apps that are eligible for the migration to Flex. 
Additionally, with the `az functionapp flex-migration start` command, customers will be able to create a Flex Consumption app from their Linux Consumption (CV1) app`. The command will migrate the relevant CV1 configurations to the Flex app. After running this migration command, customers will just need to deploy their code content, test it and then delete the original CV1 app. 

**Testing Guide**

---

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/appservice/_help.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/_help.py
@@ -982,6 +982,33 @@ type: command
 short-summary: List available built-in stacks which can be used for function apps on the Flex Consumption plan.
 """
 
+helps['functionapp flex-migration'] = """
+type: group
+short-summary: Manage migration of Linux Consumption function apps to the Flex Consumption plan.
+"""
+
+helps['functionapp flex-migration start'] = """
+type: command
+short-summary: Create a Flex Consumption app with the same settings as the provided Linux Consumption function app.
+examples:
+  - name: Migrate a Linux Consumption function app to the Flex Consumption plan.
+    text: >
+        az functionapp flex-migration start --source-name MyLinuxConsumptionApp --source-resource-group MyLinuxConsumptionResourceGroup --name MyFunctionApp --resource-group MyResourceGroup --storage-account MyStorageAccount
+
+  - name: Migrate a Linux Consumption function app to the Flex Consumption plan without migrating managed identity configurations.
+    text: >
+        az functionapp flex-migration start --source-name MyLinuxConsumptionApp --source-resource-group MyLinuxConsumptionResourceGroup --name MyFunctionApp --resource-group MyResourceGroup --storage-account MyStorageAccount --skip-managed-identities
+"""
+
+helps['functionapp flex-migration list'] = """
+type: command
+short-summary: List all Linux Consumption function apps that are eligible for migration to the Flex Consumption plan.
+examples:
+  - name: List all Linux Consumption function apps that are eligible for migration to the Flex Consumption plan.
+    text: >
+        az functionapp flex-migration list
+"""
+
 helps['functionapp plan'] = """
 type: group
 short-summary: Manage App Service Plans for an Azure Function
--- a/src/azure-cli/azure/cli/command_modules/appservice/custom.py
+++ b/src/azure-cli/azure/cli/command_modules/appservice/custom.py
@@ -954,6 +954,678 @@ def remove_remote_build_app_settings(cmd, resource_group_name, name, slot):
             logger.warning("App settings may not be propagated to the SCM site")
 
 
+def _is_linux_consumption_function_app(cmd, site):
+    web_client = get_mgmt_service_client(cmd.cli_ctx, WebSiteManagementClient)
+
+    if site.kind != 'functionapp,linux':
+        return False
+
+    if not is_valid_resource_id(site.server_farm_id):
+        return False
+
+    try:
+        parsed_plan_id = parse_resource_id(site.server_farm_id)
+        plan_info = web_client.app_service_plans.get(parsed_plan_id['resource_group'], parsed_plan_id['name'])
+        if plan_info is None:
+            return False
+        return plan_info.sku.tier.lower() == 'dynamic'
+    except Exception:  # pylint: disable=broad-except
+        return False
+
+
+def list_flex_migration_candidates(cmd):
+    from azure.cli.core.commands.client_factory import get_subscription_id
+
+    subscription_id = get_subscription_id(cmd.cli_ctx)
+    web_client = get_mgmt_service_client(cmd.cli_ctx, WebSiteManagementClient)
+
+    logger.warning("Searching for function apps under the subscription '%s' that are eligible for Flex "
+                   "Consumption migration...\n", subscription_id)
+
+    all_sites = list(web_client.web_apps.list())
+    eligible_sites = []
+    ineligible_sites = []
+
+    flex_regions = [region['name'] for region in list_flexconsumption_locations(cmd)]
+
+    for site in all_sites:
+        if not _is_linux_consumption_function_app(cmd, site):
+            continue
+
+        try:
+            if validate_flex_migration_eligibility_for_linux_consumption_app(cmd, site, flex_regions):
+                site_entry = {
+                    'name': site.name,
+                    'resource_group': site.resource_group,
+                }
+
+                has_slots = len(list_slots(cmd, site.resource_group, site.name)) > 0
+
+                if has_slots:
+                    slots_warning = (f"The site '{site.name}' has slots configured. This will not block migration, "
+                                     f"but please note that slots are not supported in Flex Consumption.")
+                    site_entry['note'] = slots_warning
+
+                eligible_sites.append(site_entry)
+
+        except Exception as e:  # pylint: disable=broad-except
+            ineligible_sites.append({
+                'name': site.name,
+                'resource_group': site.resource_group,
+                'reason': str(e)
+            })
+
+    return {
+        'eligible_apps': eligible_sites,
+        'ineligible_apps': ineligible_sites
+    }
+
+
+def validate_flex_migration_eligibility_for_linux_consumption_app(cmd, site, flex_regions):
+    # Validating that the site is in a Flex Consumption-supported region
+    normalized_site_location = _normalize_location(cmd, site.location)
+    if normalized_site_location not in flex_regions:
+        raise ValidationError("The site '{}' is not in a region supported in Flex Consumption. "
+                              "Please see the list regions supported in Flex Consumption by running az functionapp "
+                              "list-flexconsumption-locations".format(site.name))
+
+    # Validating that the site is using a Flex Consumption-supported runtime
+    site_config = get_site_configs(cmd, site.resource_group, site.name)
+    linux_fx_version = getattr(site_config, 'linux_fx_version', None)
+    runtime_info = _get_functionapp_runtime_info_helper(cmd, linux_fx_version, None, None, True)
+    runtime = runtime_info['app_runtime']
+    runtime_version = runtime_info['app_runtime_version']
+
+    runtime_helper = _FlexFunctionAppStackRuntimeHelper(cmd, normalized_site_location, runtime)
+    runtime_helper.resolve(runtime, runtime_version)
+
+    # Validating that the site does not have SSL bindings configured
+    for ssl_state in site.host_name_ssl_states or []:
+        if ssl_state.ssl_state != 'Disabled':
+            raise ValidationError("The site '{}' is using TSL/SSL certificates. "
+                                  "TSL/SSL certificates are not supported in Flex Consumption.".format(site.name))
+
+    # Validating that the site does not have WEBSITE_LOAD_CERTIFICATES app setting configured
+    app_settings = get_app_settings(cmd, site.resource_group, site.name)
+    for setting in app_settings:
+        if setting['name'] == 'WEBSITE_LOAD_CERTIFICATES':
+            raise ValidationError("The site '{}' has the WEBSITE_LOAD_CERTIFICATES app setting configured. "
+                                  "Certificate loading is not supported in Flex Consumption.".format(site.name))
+
+    # Validating that the site has triggers supported in Flex Consumption
+    functions = list_functions(cmd, site.resource_group, site.name)
+    unsupported_blob_triggers = []
+
+    for function in functions:
+        bindings = function.config.get('bindings', [])
+        for binding in bindings:
+            if binding.get('type', None) == 'blobTrigger' and binding.get('source', None) != 'EventGrid':
+                unsupported_blob_triggers.append(function.name)
+
+    if unsupported_blob_triggers:
+        function_list = '\n'.join(unsupported_blob_triggers)
+        raise ValidationError("The site '{}' has blob storage trigger(s) that don't use Event Grid "
+                              "as the source:\n{}\nFlex Consumption only supports Event Grid-based blob triggers. "
+                              "Please convert these triggers to use Event Grid or replace them with Event Grid "
+                              "triggers before migration.".format(site.name, function_list))
+
+    return True
+
+
+def get_storage_account_from_functionapp(cmd, resource_group_name, name):
+    from azure.cli.command_modules.storage.operations.account import list_storage_accounts
+
+    storage_account_name = None
+    app_settings = get_app_settings(cmd, resource_group_name, name)
+    for setting in app_settings:
+        if setting['name'] == 'AzureWebJobsStorage':
+            for part in setting['value'].split(';'):
+                if part.startswith('AccountName='):
+                    storage_account_name = part.split('=')[1]
+                    break
+
+        if setting['name'] == 'AzureWebJobsStorage__accountName':
+            storage_account_name = setting['value']
+            break
+
+        if setting['name'] == 'AzureWebJobsStorage__blobServiceUri':
+            match = re.match(r'https?://([^.]+)\.blob\.core\.windows\.net', setting['value'])
+            if match:
+                storage_account_name = match.group(1)
+                break
+
+    if not storage_account_name:
+        raise ResourceNotFoundError("Unable to obtain storage account name from app settings for function app '{}'. "
+                                    .format(name))
+
+    storage_accounts_rg = list_storage_accounts(cmd, resource_group_name)
+    for storage_account in storage_accounts_rg:
+        if storage_account.name == storage_account_name:
+            return storage_account.id
+
+    storage_accounts_sub = list_storage_accounts(cmd)
+    for storage_account in storage_accounts_sub:
+        if storage_account.name == storage_account_name:
+            return storage_account.id
+
+    raise ResourceNotFoundError("Storage account '{}' referenced by function app '{}' was not found in subscription."
+                                .format(storage_account_name, name))
+
+
+def migrate_consumption_to_flex(cmd, source_resource_group, source_name, resource_group, name, storage_account=None,
+                                maximum_instance_count=None, skip_managed_identities=False,
+                                skip_access_restrictions=False, skip_storage_mount=False, skip_hostnames=False,
+                                skip_cors=False):
+
+    web_client = get_mgmt_service_client(cmd.cli_ctx, WebSiteManagementClient)
+
+    # Validate that the app is eligible for Flex Consumption migration
+    logger.warning("Validating that the app '%s' is eligible for Flex Consumption migration...", source_name)
+    flex_regions = [region['name'] for region in list_flexconsumption_locations(cmd)]
+    source = web_client.web_apps.get(source_resource_group, source_name)
+
+    if not _is_linux_consumption_function_app(cmd, source):
+        raise ValidationError("The site '{}' is not on a Linux Dynamic (Consumption) plan. Flex Consumption "
+                              "migration is only supported for Function Apps on Linux Consumption plans."
+                              .format(source.name))
+
+    if validate_flex_migration_eligibility_for_linux_consumption_app(cmd, source, flex_regions):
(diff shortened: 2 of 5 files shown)


yonzhan:

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -5023,8 +5645,7 @@ def create_flex_app_service_plan(cmd, resource_group_name, name, location, zone_
         location=location,
         sku=sku_def,
         reserved=True,
-        kind="functionapp",
-        name=name
+        kind="functionapp"
```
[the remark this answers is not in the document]
kamperiadis: A warning is thrown otherwise. The webapps team also had to fix this: https://github.com/Azure/azure-cli/commit/3bd985d0efe4d6a023265eedfb34a70cd974cad1

src/azure-cli/azure/cli/command_modules/appservice/_help.py
```diff
@@ -982,6 +982,37 @@
...
+
+helps['functionapp flex-migration start'] = """
+type: command
+short-summary: Create a Flex Consumption app with the same settings as the provided Linux Consumption function app.
+examples:
+  - name: Migrate a Linux Consumption function app to the Flex Consumption plan.
+    text: >
+        az functionapp flex-migration start --source-name MyLinuxConsumptionApp --source-resource-group MyLinuxConsumptionResourceGroup --name MyFunctionApp --resource-group MyResourceGroup --storage-account MyStorageAccount
+
+  - name: Migrate a Linux Consumption function app to the Flex Consumption plan without configuring managed identity configurations.
```
mathewc: suggestion: maybe reword "... without configuring managed identity configurations", perhaps ".. without migrating managed identity configurations"

src/azure-cli/azure/cli/command_modules/appservice/_help.py
```diff
@@ -982,6 +982,37 @@
...
+
+helps['functionapp flex-migration list'] = """
+type: command
+short-summary: List all Linux Consumption function apps that are eligible for migration to the Flex Consumption plan.
+examples:
+  - name: List all Linux Consumption function apps that are eligible for migration to the Flex Consumption plan.
+    text: >
+        az functionapp flex-migration list
+
+  - name: List all Linux Consumption function apps that are eligible for migration to the Flex Consumption plan, ignoring if they have slots configured. Please note that slots are not supported in the Flex Consumption plan.
```
mathewc: suggestion: "... to the Flex Consumption plan, skipping any with slots configured."

src/azure-cli/azure/cli/command_modules/appservice/_help.py
```diff
@@ -982,6 +982,37 @@
...
+    text: >
+        az functionapp flex-migration list --ignore-slots
```
mathewc: I wonder if the param shouldn't be --skip-slots, since you used the "skip" wording for managed-identities already

src/azure-cli/azure/cli/command_modules/appservice/_help.py
```diff
@@ -982,6 +982,37 @@
...
+short-summary: Manage migration of Linux Consumption function apps to the Flex Consumption plan.
+"""
+
+helps['functionapp flex-migration start'] = """
+type: command
+short-summary: Create a Flex Consumption app with the same settings as the provided Linux Consumption function app.
+examples:
+  - name: Migrate a Linux Consumption function app to the Flex Consumption plan.
+    text: >
+        az functionapp flex-migration start --source-name MyLinuxConsumptionApp --source-resource-group MyLinuxConsumptionResourceGroup --name MyFunctionApp --resource-group MyResourceGroup --storage-account MyStorageAccount
```
mathewc, line 996: Is storage-account required? If not provided, would you use the same storage account that the source uses?

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -954,6 +954,621 @@ def remove_remote_build_app_settings(cmd, resource_group_name, name, slot):
...
+        return False
+
+
+def list_flex_migration_candidates(cmd, ignore_slots=False):
+    from azure.cli.core.commands.client_factory import get_subscription_id
+
+    subscription_id = get_subscription_id(cmd.cli_ctx)
+    web_client = get_mgmt_service_client(cmd.cli_ctx, WebSiteManagementClient)
+
+    logger.warning("Searching for function apps under the subscription '%s' that are eligible for Flex "
```
mathewc: I know you told me that it appears ALL logs are being written using logger.warning, but that doesn't really make sense if you want to be able to have warnings stand out in output. So I guess we need general CLI guidance on why we shouldn't be using logger.info here, reserving warning/error for lines we want to stand out to the customer

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -954,6 +954,621 @@ def remove_remote_build_app_settings(cmd, resource_group_name, name, slot):
...
+        logger.warning("Source app '%s' is eligible for Flex migration", source_name)
+
+    source_site_configs = get_site_configs(cmd, source_resource_group, source_name)
+    source_linux_fx_version = getattr(source_site_configs, 'linux_fx_version', None)
+    source_runtime_info = _get_functionapp_runtime_info_helper(cmd, source_linux_fx_version, None, None, True)
+    source_runtime = source_runtime_info['app_runtime']
+    source_runtime_version = source_runtime_info['app_runtime_version']
+
+    # Create flex function app
+    logger.warning("\nCreating FlexConsumption function app '%s' in resource group '%s'...", name, resource_group)
```
mathewc: Most places you use "Flex Consumption", here you're using "FlexConsumption", and in some places you use just "Flex". Recommend making this consistent.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -954,6 +954,621 @@ def remove_remote_build_app_settings(cmd, resource_group_name, name, slot):
...
+            'WEBSITE_VNET_ROUTE_ALL',
+            'APPLICATIONINSIGHTS_CONNECTION_STRING',
+            'AZUREWEBJOBSDASHBOARD'
+        }
+
+        migrated_app_settings = []
+        for setting in source_app_settings:
+            setting_name = setting['name'].upper()
+            if setting_name not in excluded_settings and not setting_name.startswith('AZUREWEBJOBSSTORAGE'):
+                migrated_app_settings.append(f"{setting['name']}={setting['value']}")
```
mathewc, line 1258: this isn't logging setting values to command line right? Those could be sensitive secrets - not sure what the CLI policy is here

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -954,6 +954,621 @@ def remove_remote_build_app_settings(cmd, resource_group_name, name, slot):
...
+        source_scm_basic_auth_enabled = basic_auth_supported(cmd.cli_ctx, source_name, source_resource_group)
+
+        if source_scm_basic_auth_enabled:
+
+            CsmPublishingCredentialsPoliciesEntity = cmd.get_models("CsmPublishingCredentialsPoliciesEntity")
+            csmPublishingCredentialsPoliciesEntity = CsmPublishingCredentialsPoliciesEntity(allow=True)
+            _generic_site_operation(cmd.cli_ctx, resource_group, name,
+                                    'update_scm_allowed', None, csmPublishingCredentialsPoliciesEntity)
+
+            logger.warning("Successfully enabled SCM basic publishing credentials")
```
mathewc: you're using "enabled" rather than migrated on this and the next log, but others you're using "migrate"

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -954,6 +954,621 @@ def remove_remote_build_app_settings(cmd, resource_group_name, name, slot):
...
+    except Exception as e:  # pylint: disable=broad-except
+        logger.error("Failed to migrate basic publishing credentials policies with error %s. "
+                     "Run 'az resource update' to configure them manually", str(e))
+
+
+def _migrate_cors_settings(cmd, source_site_configs, source_name, resource_group, name):
+    logger.warning("\nMigrating CORS settings from source function app '%s' to target function app '%s'...",
+                   source_name, name)
+    try:
+        source_cors_settings = source_site_configs.cors
```
mathewc, line 1338: I'm no Python developer, but the spacing looks off in the following block of code

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -2455,18 +3070,7 @@ def update_site_configs(cmd, resource_group_name, name, slot=None, number_of_wor
         if max_replicas is not None:
             setattr(configs, 'function_app_scale_limit', max_replicas)
         return update_configuration_polling(cmd, resource_group_name, name, slot, configs)
-    try:
-        return _generic_site_operation(cmd.cli_ctx, resource_group_name, name, 'update_configuration', slot, configs)
-    except Exception as ex:  # pylint: disable=broad-exception-caught
```
mathewc, line 2460: Why the removal of this exception handler block?

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -5542,19 +6159,16 @@ def create_functionapp(cmd, resource_group_name, name, storage_account, plan=Non
                                subnet_name=subnet_info["subnet_name"],
                                subnet_service_delegation=FLEX_SUBNET_DELEGATION if flexconsumption_location else None)
         subnet_resource_id = subnet_info["subnet_resource_id"]
-        vnet_route_all_enabled = True
```
mathewc, line 5539: I don't have context on these changes - they don't seem related?

src/azure-cli/azure/cli/command_modules/appservice/custom.py
kamperiadis: No. We are only logging the names below, not the values

src/azure-cli/azure/cli/command_modules/appservice/custom.py
kamperiadis: For context, this exception handling was added as part of this commit: https://github.com/Azure/azure-cli/commit/ef54889112c5ff4009ea653fdbf91f4fe88a8abf

And this function is run with the `az webapp config set` command. 

I do not think it is right that we are catching Conflict and Bad Request exceptions. I think that if either of those two exceptions are simply logged instead of thrown back to the client when run via a script in a pipeline, customers might miss that there were issues with updating the site config. Even for the purposes of this PR, the caller of this function should know via a thrown exception that something went wrong, and the caller should decide how to handle such exception.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
kamperiadis: @zhoxing-ms Can you provide guidance on this?

According to the guidance document [here](https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md#general-patterns) "Log to logger.error() or logger.warning() for user messages". The next guideline is "Use the appropriate logging level for printing strings" but the example just highlights the distinction between using the logger vs returning the string as opposed to logger.info vs logger.warning. 

When I switch to make these logs use logger.info instead of logger.warning, I do not see the logs unless I use the `--verbose` flag. It seems to be due to [this](https://github.com/microsoft/knack/blob/b1164b57d777711fd3954cf11f18ffb06ab1ca5b/knack/log.py#L153) dependency we use for the logging which will only show messages from warning logs and above, unless the --verbose and --debug flags are passed.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
kamperiadis: I figured "enabled" would make more sense in this context since it is just a single boolean property that we are enabling as opposed to a set of properties. What do you think?

src/azure-cli/azure/cli/command_modules/appservice/custom.py
kamperiadis: The reason why I made these changes is because otherwise we get the "Readonly attribute name will be ignored in class" warning since vnet_route_all_enabled is only part of the SiteConfig class ([code ref](https://github.com/Azure/azure-sdk-for-python/blob/afce6c3a2c8c1c47688e8d0c896da9eac07477b6/sdk/appservice/azure-mgmt-web/azure/mgmt/web/v2024_11_01/models/_models_py3.py#L20941)) It is not part of the Site class: https://github.com/Azure/azure-sdk-for-python/blob/afce6c3a2c8c1c47688e8d0c896da9eac07477b6/sdk/appservice/azure-mgmt-web/azure/mgmt/web/v2024_11_01/models/_models_py3.py#L19638

src/azure-cli/azure/cli/command_modules/appservice/_help.py
kamperiadis: Just implemented the changes to support this and I updated the help info for the storage account param to reflect this.

src/azure-cli/azure/cli/command_modules/appservice/_help.py
kamperiadis: The slots param was removed as per Thiago's suggestion.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -954,6 +954,678 @@ def remove_remote_build_app_settings(cmd, resource_group_name, name, slot):
...
+                has_slots = len(list_slots(cmd, site.resource_group, site.name)) > 0
+
+                if has_slots:
+                    slots_warning = (f"The site '{site.name}' has slots configured. This will not block migration, "
+                                     f"but please note that slots are not supported in Flex Consumption.")
+                    site_entry['note'] = slots_warning
+
+                eligible_sites.append(site_entry)
+
+        except ValidationError as e:
```
khkh-ms: We should also handle other generic exceptions. Currently, if any site throws an error that isn't a ValidationError, it completely prevents the customer from viewing the list of eligible apps.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -954,6 +954,678 @@ def remove_remote_build_app_settings(cmd, resource_group_name, name, slot):
...
+    web_client = get_mgmt_service_client(cmd.cli_ctx, WebSiteManagementClient)
+
+    logger.warning("Searching for function apps under the subscription '%s' that are eligible for Flex "
+                   "Consumption migration...\n", subscription_id)
+
+    all_sites = list(web_client.web_apps.list())
+    eligible_sites = []
+    ineligible_sites = []
+
+    flex_regions = [region['name'] for region in list_flexconsumption_locations(cmd)]
```
khkh-ms, line 989: We are not listing the consumption apps which are in non-flex regions in the `ineligible_sites` list. Little explanation of why some apps did make to both lists might help.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -954,6 +954,678 @@ def remove_remote_build_app_settings(cmd, resource_group_name, name, slot):
...
+    source = web_client.web_apps.get(source_resource_group, source_name)
+
+    if not _is_linux_consumption_function_app(cmd, source):
+        raise ValidationError("The site '{}' is not on a Linux Dynamic (Consumption) plan. Flex Consumption "
+                              "migration is only supported for Function Apps on Linux Consumption plans."
+                              .format(source.name))
+
+    if validate_flex_migration_eligibility_for_linux_consumption_app(cmd, source, flex_regions):
+        slots = list_slots(cmd, source_resource_group, source_name)
```
khkh-ms, line 1133: Why not add `slots` check in `validate_flex_migration_eligibility_for_linux_consumption_app` method and add the warning there? so that you don't have duplicated logic of slots.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
kamperiadis: We are listing consumption apps which are in non flex regions in `ineligible_sites`. We are iterating over `all_sites` which is the list of all apps in the sub. Then in 992 we are skipping non consumption apps. We obtain `flex_regions` and pass it to the `validate_flex_migration_eligibility_for_linux_consumption_app` so that we just get the list once for all the apps. And in that validation function in line 1027 we throw an exception if the site is not in a flex region. That exception will be added as the reason in the `ineligible_apps` result

src/azure-cli/azure/cli/command_modules/appservice/custom.py
kamperiadis: Because for `az functionapp flex-migration list` command, Thiago wants the slot warning to be in 'note' for the eligible_site. Whereas for the `az functionapp flex-migration start` command, we just want to log it.

src/azure-cli/azure/cli/command_modules/appservice/custom.py
kamperiadis: Done

khkh-ms (approved):

src/azure-cli/azure/cli/command_modules/appservice/custom.py
zhoxing-ms: If you want notification message to be displayed by default even without using parameter `--verbose`, you can try using `print()`

kamperiadis: Isn't that against what the guidelines outlined here: https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md#general-patterns

> Log to logger.error() or logger.warning() for user messages; do not use the print() function

src/azure-cli/azure/cli/command_modules/appservice/tests/latest/test_functionapp_commands.py
```diff
@@ -809,6 +809,539 @@ def test_functionapp_update_enabled_dapr(self, resource_group, storage_account):
         ])
 
 
+class FunctionAppFlexMigrationTest(LiveScenarioTest):
```
zhoxing-ms, line 812: May I ask why do you need to mark these tests as `LiveScenarioTest`?

kamperiadis: Much like what happened for this other PR https://github.com/Azure/azure-cli/pull/28001, because there are resources created on the customer's behalf (e.g. server farms, storage accounts, etc.), when I switch to using ScenarioTest, the resources that were generated in the yaml recording are different in the next run and it can't find a match. The tests run just fine when I set they run in live mode.  

When using ScenarioTest, I get success the first time I run the tests. But then when I run those same tests again using the yaml recordings, I see things like this:
image


 vcr.errors.CannotOverwriteExistingCassetteException: Can't overwrite existing cassette ('q:\\src\\azure-cli\\src\\azure-cli\\azure\\cli\\command_modules\\appservice\\tests\\latest\\recordings\\test_functionapp_flex_migration_site_configuration.yaml') in your current record mode ('once').
E               No match for the request (<Request (PUT) https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/clitest.rg000002/providers/Microsoft.Web/serverfarms/ASP-clitestrg000002-0c0b?api-version=2024-11-01>) was found.
E               Found 8 similar requests with 1 different matcher(s) :
E
E               1 - (<Request (PUT) https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/clitest.rg000001/providers/Microsoft.Web/sites/srcfunc000005?api-version=2024-11-01>).
E               Matchers succeeded : ['method', 'scheme', 'host', 'port', '_custom_request_query_matcher']
E               Matchers failed :
E               path - assertion failure :
E               /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/clitest.rg000002/providers/Microsoft.Web/serverfarms/ASP-clitestrg000002-0c0b != /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/clitest.rg000001/providers/Microsoft.Web/sites/srcfunc000005
E
E               2 - (<Request (PUT) https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/clitest.rg000001/providers/Microsoft.Web/sites/srcfunc000005/config/appsettings?api-version=2024-11-01>).
E               Matchers succeeded : ['method', 'scheme', 'host', 'port', '_custom_request_query_matcher']
E               Matchers failed :
E               path - assertion failure :
E               /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/clitest.rg000002/providers/Microsoft.Web/serverfarms/ASP-clitestrg000002-0c0b != /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/clitest.rg000001/providers/Microsoft.Web/sites/srcfunc000005/config/appsettings

zhoxing-ms: Got it, thanks

zhoxing-ms (approved):

Issue #29640, opened 2024-08-02
Support Python 3.13


Goldziher:
Python 3.13 is in the works and will be released in the nearish future. It would be great to have it supported. 

## TODOs
- aaz-dev supports 3.13
  - https://github.com/Azure/aaz-dev-tools/pull/495
- azdev supports 3.13
  - https://github.com/Azure/azure-cli-dev-tools/pull/527
- Dependencies are updated to the Python 3.13 compatible version
  - https://github.com/Azure/azure-cli/pull/31822
  - https://github.com/Azure/azure-cli/pull/31823
- CI passes with Python 3.13
  - aaz warning: `FutureWarning: functools.partial will be a method descriptor in future Python versions; wrap it in staticmethod() if you want to preserve the old behavior`
  - https://github.com/Azure/azure-cli/pull/31895
  - https://github.com/Azure/azure-cli/pull/31928
  - https://github.com/Azure/azure-cli-extensions/pull/9056
- Extension compatible with Python 3.13
  - `webpubsub`: `ModuleNotFoundError: No module named 'cgi'` https://github.com/Azure/azure-cli-extensions/pull/8993 
  - `azure-sphere`: `from pathlib import Path, os` `ImportError: cannot import name 'os' from 'pathlib'` https://github.com/Azure/azure-cli-extensions/pull/9025

yonzhan:
Thank you for opening this issue, we will look into it.

Goldziher:
Thank you! i would like to point out that typing is a major point of improvement for the v3.12 and 3.13. Especially the new generic syntax etc.

XhmikosR:
Any news about this? Fedora Workstation 41 ships with Python 3.13 and there are failures with v2.62.

dmilith:
It's 3.13.1 already, but this still fails like this:

```
ERROR: Ignored the following yanked versions: 0.13.11, 0.13.13, 0.14.1
[...]

PR #31264, opened 2025-04-14
[Packaging] Bump Python to 3.12 on RHEL and CentOS Stream

bebound:
**Description**

This PR bumps the Python dependency to 3.12. **Some extensions may need to be reinstalled if they become non-functional**: `az extension remove -n xxx && az extension add -n xxx`

After this PR, Mariner 2.0 is the only one relies on Python 3.9, the rest of the distros use 3.12.

Issue #27673, opened 2023-10-24
Support Python 3.12


chenrui333:
## Description
Python 3.12 has been released on 2023-10-02: https://www.python.org/downloads/release/python-3120/

Azure CLI needs to support Python 3.12 like how we supported 3.10 (#19857), 3.11 (#24494).

## TODOs
- [x] [Knack](https://github.com/microsoft/knack) supports Python 3.12
  - https://github.com/microsoft/knack/pull/279
- [x] [azdev](https://github.com/Azure/azure-cli-dev-tools) supports Python 3.12
  - https://github.com/Azure/azure-cli-dev-tools/pull/458
- [x] Dependencies are updated to the Python 3.12 compatible version
  - `pymsalruntime` (dependency of `msal[broker]`) https://github.com/Azure/azure-cli/pull/28556
  - https://github.com/Azure/azure-cli/pull/29466
  - https://github.com/Azure/azure-cli/pull/29486
- [ ] CI passes with Python 3.12
  - Remove `distutils` https://github.com/Azure/azure-cli/pull/28796
  - https://github.com/Azure/azure-cli/pull/27196
  - https://github.com/Azure/azure-cli/pull/29504
  - https://github.com/Azure/azure-cli/pull/29515
  - https://github.com/Azure/azure-cli/pull/29517
  - Remove `assertDictContainsSubset` and `assertNotEquals` #28194
    - https://github.com/Azure/azure-cli/pull/29120
    - https://github.com/Azure/azure-cli/pull/29121
  - https://github.com/Azure/azure-cli/issues/28829
  - https://github.com/Azure/azure-cli/pull/29373
  - https://github.com/Azure/azure-cli/pull/29465
  - https://github.com/Azure/azure-cli/pull/29887
  - https://github.com/Azure/azure-cli-extensions/pull/8010
  - https://github.com/Azure/azure-cli-extensions/pull/8116

yonzhan:
Thank you for opening this issue, we will look into it.

chenrui333:
I find there is no open issue to support py3.12, create this one for the team.

borjamunozf:
Could be this the explanation why installing azure-cli with Python 3.12 throw this error complining about pymsalruntime?

```
building 'pymsalruntime.pymsalruntime' extension
      error: Microsoft Visual C++ 14.0 or greater is required. Get it with "Microsoft C++ Build Tools": https://visualstudio.microsoft.com/visual-cpp-build-tools/
      [end of output]

[...]

PR #31895, opened 2025-07-31
[Packaging] Support Python 3.13

bebound:
**Description**

- #29640

Changes:
- Update `setup.py` to support 3.13
- Add Python 3.13 full test
- Change default Python version to 3.13 in CI


yonzhan:

scripts/ci/test_extensions.sh
```diff
@@ -61,7 +62,7 @@ az self-test --debug
 if [ $? != 0 ]
 then
     exit_code=1
-    echo "Failed to verify:" $ext
+    echo "Failed to verify"
```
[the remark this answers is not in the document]
bebound: `$ext` is meaningless, it always points to last extension.

azure-pipelines.yml
```diff
@@ -468,10 +468,10 @@ jobs:
   displayName: Test Python Wheels
   strategy:
     matrix:
-      Python39:
-        python.version: '3.9'
```
jiasli, line 472: We should not drop Python 3.9's test as we need to make sure our code works on Python 3.9

scripts/ci/test_extensions.sh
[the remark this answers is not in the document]
jiasli: I guess this line is copy-and-pasted from L55 without too much thinking.

jiasli:
No need to bump pylint and updating linter rules? We did a lot to support Python 3.12: https://github.com/Azure/azure-cli/issues/27673.

azure-pipelines.yml
bebound: We only test the Python which is used in official package.
Currently, RPM used 3.12, DEB, macOS and Windows use 3.13.

bebound:
> No need to bump pylint and updating linter rules? We did a lot to support Python 3.12: #27673.

There no major version bump in Pylint. Azdev is using the latest version.
>          'pylint<4',
> Source: https://github.com/Azure/azure-cli-dev-tools/blob/5130a24020ca0a47055e914a58a4af67d9eb98ad/setup.py#L76

azure-pipelines.yml
jiasli: We should have removed 3.9's test in https://github.com/Azure/azure-cli/pull/31264.

PR #31973, opened 2025-08-21
{Core} `aaz`: Wrap `functools.partial` in `staticmethod()` to remove FutureWarning

RenSilvaAU:
**Related command**
`az command`  

**Description**  
  
Added `staticmethod` to `functools.partial` to ensure compatibility with Python 3.13.  
This change addresses a breaking behavior introduced in Python 3.13 where `functools.partial` no longer supports static method binding without explicit declaration.  
The update ensures continued functionality across supported Python versions and prevents runtime errors in affected modules.

**Testing Guide**  
  
Run unit tests for modules using `functools.partial` as a static method.  
class MyClass:
    @staticmethod
    def my_method(x):
        return x * 2

partial_func = staticmethod(functools.partial(MyClass.my_method, 5))
assert partial_func() == 10
Example:  

import functools

### DIFF
--- a/src/azure-cli-core/azure/cli/core/aaz/_command.py
+++ b/src/azure-cli-core/azure/cli/core/aaz/_command.py
@@ -327,13 +327,13 @@ def register_command_group(
         helps[name] = yaml.safe_dump(cls.AZ_HELP)
 
         if is_preview:
-            cls.AZ_PREVIEW_INFO = staticmethod(partial(PreviewItem, 
+            cls.AZ_PREVIEW_INFO = staticmethod(partial(PreviewItem,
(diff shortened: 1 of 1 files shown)


yonzhan:

bebound:
Please fix the PR title and lint error.

Original error message when run `az self-test`: `FutureWarning: functools.partial will be a method descriptor in future Python versions; wrap it in staticmethod() if you want to preserve the old behavior`

Here is the related PR about this change: https://github.com/python/cpython/pull/121092

bebound (approved):

kairu-ms (approved):

PR #32079, opened 2025-09-08
{AAZ} Added functionality to AAZ MCP Server

a0x1ab:
**Related command**
NA

**Description**
This PR consists of several changes for the internally used AAZ MCP server to assist 1P teams to generate CLI cmdlets using AAZ through the VSCode Copilot Chat.

**Testing Guide**
NA

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -3,9 +3,7 @@ FROM mcr.microsoft.com/devcontainers/python:3.12
 # Copy uv binaries for virtual environment management
 COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
 
-# Install git and mcp, create /workspaces-src with correct ownership
-RUN apt-get update && apt-get install -y git \
-    && pip install --no-cache-dir mcp \
+RUN pip install --no-cache-dir mcp \
     && mkdir -p /workspaces-src \
     && chown -R vscode:vscode /workspaces-src
 
--- a/tools/aaz-flow/helpers.py
+++ b/tools/aaz-flow/helpers.py
@@ -21,17 +21,18 @@ paths = {
 }
 
 async def fetch_available_services():
-    """Fetch available services from azure-rest-api-specs repository."""
-    url = "https://api.github.com/repos/Azure/azure-rest-api-specs/contents/specification"
-    try:
-        response = requests.get(url, timeout=10)
-        response.raise_for_status()
-    except requests.RequestException:
+    """Retrieve available services by parsing local azure-rest-api-specs/specification directory"""
+    spec_path = os.path.join(paths["swagger_path"], "specification")
+    if not os.path.exists(spec_path):
+        # Fallback to common services if local path is missing
         return ["storage", "compute", "network", "keyvault", "monitor"]
 
-    directories = [item['name'] for item in response.json() if item['type'] == 'dir']
-    directories.sort()
-    return directories
+    try:
+        directories = [d for d in os.listdir(spec_path) if os.path.isdir(os.path.join(spec_path, d))]
+        directories.sort()
+        return directories
+    except Exception:
+        return ["storage", "compute", "network", "keyvault", "monitor"]
 
 async def validate_paths(ctx: Context) -> dict:
     """Validate and get correct paths for required directories."""
--- a/tools/aaz-flow/pyproject.toml
+++ b/tools/aaz-flow/pyproject.toml
@@ -5,8 +5,7 @@ description = "The MCP server utilized for code generation in Azure CLI (AAZ)."
 readme = "README.md"
 requires-python = ">=3.12"
 dependencies = [
-    "fastmcp>=2.11.3",
-    "mcp>=1.13.1",
+    "fastmcp>=2.11.3"
 ]
 
 [tool.setuptools]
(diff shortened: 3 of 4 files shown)


yonzhan:

yanzhudd:
Hi @a0x1ab 
may I ask why we need to modify the devcontainer folder?

yanzhudd:
Could you please narrow down the changes to include only what is necessary?

.devcontainer/devcontainer.json
```diff
@@ -1,17 +1,12 @@
...
-  "workspaceFolder": "/workspaces",
-  "onCreateCommand": "uv venv .venv --seed",
-  "postCreateCommand": "REPO_NAME=$(basename $GITHUB_REPOSITORY) && cat $REPO_NAME/.devcontainer/init.sh >> ~/.bashrc && mkdir .vscode && cp $REPO_NAME/.devcontainer/mcp.json .vscode/",
-  "hostRequirements": {
-    "cpus": 16,
-    "memory": "64gb",
-    "storage": "128gb"
-  }
-}
+  "workspaceFolder": "/workspaces/azure-cli",
```
necusjz, line 10: the 4 folders (azure-cli, azure-cli-extensions, azure-rest-api-specs，aaz) should at the same level. why do we need to change it?

.devcontainer/Dockerfile
```diff
@@ -1,7 +1,13 @@
 FROM mcr.microsoft.com/devcontainers/python:3.12
 
-# Copy binaries from the official
+# Copy uv binaries for virtual environment management
 COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
 
-# Remove after production
-RUN pip install --no-cache-dir mcp
+# Install git and mcp, create /workspaces-src with correct ownership
+RUN apt-get update && apt-get install -y git \
```
necusjz: i’m worried it might affect the container’s build performance. what is the specific purpose of running it?

.devcontainer/devcontainer.json
a0x1ab: So the previous configuration of the devcontainer does not work on my machine - the workspace configuration and host requirements both fail and need manual intervention to work. I have moved the workspace configuration to the init  script and it is now done automatically.

tools/aaz-flow/helpers.py
```diff
@@ -0,0 +1,290 @@
...
+paths = {
+    "aaz": os.getenv("AAZ_PATH", "/workspaces-src/aaz"),
+    "cli": "/workspaces/azure-cli",
+    "cli_extension": os.getenv("CLI_EXTENSION_PATH", "/workspaces-src/azure-cli-extensions"),
+    "swagger_path": os.getenv("SWAGGER_PATH", "/workspaces-src/azure-rest-api-specs")
+}
+
+async def fetch_available_services():
+    """Fetch available services from azure-rest-api-specs repository."""
+    url = "https://api.github.com/repos/Azure/azure-rest-api-specs/contents/specification"
```
yanzhudd: As we discussed earlier, I believe it’s better to retrieve the available service names by parsing the folder names in the local azure-rest-api-specs repository, rather than fetching them from remote data.

tools/aaz-flow/pyproject.toml
```diff
@@ -5,5 +5,9 @@ description = "The MCP server utilized for code generation in Azure CLI (AAZ)."
 readme = "README.md"
 requires-python = ">=3.12"
 dependencies = [
-    "mcp>=1.12.0",
+    "fastmcp>=2.11.3",
+    "mcp>=1.13.1",
```
necusjz: do we need both of them? is using only `fastmcp` enough now?

.devcontainer/Dockerfile
a0x1ab: The update for apt packages allows us to make sure we also retrieve the latest version of all the packages involved, since container build usually a one-time operation (unless changes are added) in terms of a devcontainer usability approach, it should not bring significant start-time impact.

tools/aaz-flow/helpers.py
a0x1ab: Yes, I can add this in as a change - this is a good point!

tools/aaz-flow/pyproject.toml
a0x1ab: Yes, I can remove the mcp dependency and run a check - I think fastmcp by itself should be enough but I'll double check

.devcontainer/Dockerfile
necusjz: i’m mainly concerned about the initial (one-time) performance. which package exactly do we want to keep at the latest version?

actually, `git` is installed by default, why we need to install it again here?

"Entities should not be multiplied unnecessarily."

.devcontainer/devcontainer.json
yanzhudd: What environment did you test the changes? You can create a devcontainer and test in it.
image

necusjz: perhaps we need to follow the same level structure. although our devcontainer is in azure-cli, users usually contribute code to azure-cli-extensions and aaz, which belong to the same abstraction layer.

tools/aaz-flow/pyproject.toml
a0x1ab: Updated and marking as resolved

tools/aaz-flow/helpers.py
a0x1ab: Updated and marking as resolved

.devcontainer/Dockerfile
a0x1ab: Updated and marking as resolved

PR #32130, opened 2025-09-17
{AppConfig} `az appconfig kv set/import`: Add support for JSON comments

albertofori:
**Related command**

`appconfig kv set`
`appconfig kv import`

**Description**

Adds support for stripping comments when writing a kv which is of json content type and might have comments. This is also done for import from a json file.

**Testing Guide**

**History Notes**

{AppConfig} `az appconfig kv set/import`: Add support for JSON comments

---


yonzhan:

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/test_appconfig_json_content_type.py
```diff
@@ -450,3 +469,18 @@ def test_azconfig_json_content_type(self, resource_group, location):
...
+        json_with_comments_file_path = os.path.join(TEST_DIR, 'json_with_comments.json')
+        clean_json_file_path = os.path.join(TEST_DIR, 'json_with_comments_stripped.json')
+
+        with open(json_with_comments_file_path) as json_file:
+            parsed_json = json.loads(strip_json_comments(json_file.read()))
+
+        with open(clean_json_file_path) as json_file:
+            expected_json = json.load(json_file)
+        
+        assert parsed_json == expected_json
```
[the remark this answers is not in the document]
albertofori: Addressed

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/test_appconfig_json_content_type.py
```diff
@@ -272,6 +273,24 @@ def test_azconfig_json_content_type(self, resource_group, location):
         with self.assertRaisesRegex(CLIError, "Set the value again in valid JSON format."):
             self.cmd('appconfig kv set --connection-string {src_connection_string} --key {key} --content-type {content_type} -y')
 
+        # Create a JSON value with both single-line and multiline comments and confirm it can be successfully set though sanitized
+        entry_key = "Key15"
+        
+        json_with_comments = "\"{\\\"key1\\\": \\\"value1\\\",// single-line comment\n\\\"key2\\\": \\\"value2\\\"\n/* multiline \n comment */}\"" # Double escaped as escaped string required in the CLI query
+
+        json_without_comments = "{{\"key1\": \"value1\",\n\"key2\": \"value2\"\n}}"
```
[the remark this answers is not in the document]
albertofori: Updated

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/test_appconfig_json_content_type.py
```diff
@@ -7,6 +7,7 @@
 
 import json
 import os
+from unittest import TestCase
 import yaml
```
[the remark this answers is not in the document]
albertofori: Updated

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -254,3 +261,73 @@ def parse_tags_to_dict(tags):
...
+
+        while idx < length:
+            if json_string[idx] == NEW_LINE:
+                break
+
+            idx += 1
+
+        return idx
+
+    def __find_next_double_quote_index(json_string, start_index):
```
ChristineWanjau: Have you tried using a python library for this? 
I have seen this library - https://pypi.org/project/json-with-comments/

mrm9084: @ChristineWanjau we are trying to avoid taking on a 3rd party dependency. We did a review of them for the Python provider and none of them were exactly great and we can do this with minimal effort.

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -984,3 +983,16 @@ def list_revision(cmd,
...
+
+
+def __normalize_json_input(json_string):
+    try:
+        json.loads(json_string)
+
+        return json_string
+
+    except json.JSONDecodeError:
+        string_without_comments = strip_json_comments(json_string)
```
mrm9084: This seems incorrect for how the CLI should work. As a customer if I have json comments in my local json files, I'll still want them with comments in my app configuration configurations.

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -254,3 +261,73 @@ def parse_tags_to_dict(tags):
                 tags_dict[tag_key] = tag_value
         return tags_dict
     return tags
+
+
+def strip_json_comments(json_string):
```
mrm9084: If we do go with this, see by below comment, is there a reason why we aren't using the same code as the provider? There shouldn't be any difference in what they are doing.

mrm9084 (changes requested):

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
albertofori: I was under the impression that we remove the comments. But you are right. We should only remove in the `import` scenario.

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
albertofori: I would say they are the same thing. This is just moved into smaller functions as I felt the main logic would be more readable. But lemme know if you feel strongly about them being the same code. I don't mind that either

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
mrm9084: It gets a bit more complicated with import. Yes, remove them from things in general, but it would be good to save them if them if they were inside the value part of a json content type, though that could be a bit tricky.

albertofori: That's true. I will say for now, for import, we just remove all comments. Export will keep a value with comments as a string anyways as parsing would fail. So on importing such an exported file, the comments would be in a string literal which should preserve them. Thoughts?

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -254,3 +261,73 @@ def parse_tags_to_dict(tags):
...
+    current_index = 0
+    length = len(json_string)
+    result = []
+
+    def __isEscaped(json_string, char_index):
+        backslash_count = 0
+        idx = char_index - 1
+        while idx >= 0 and json_string[idx] == '\\':
```
mrm9084: It's only used once but `\\` should be a constant.

mrm9084: Actually you already have a constant for it.

albertofori: This has been updated, thanks!

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/test_appconfig_json_content_type.py
```diff
@@ -272,6 +274,27 @@ def test_azconfig_json_content_type(self, resource_group, location):
...
+        # Create a JSON value with both single-line and multiline comments and confirm it can be successfully set though sanitized
+        entry_key = "Key15"
+        
+        escaped_json_with_comments = "\"{\\\"key1\\\": \\\"value1 is of the type \\\\\\\"string\\\\\\\"\\\",// single-line comment\n\\\"key2\\\": \\\"value2\\\"\n/* multiline \n comment */}\"" # Double escaped as escaped string required in the CLI query
+
+        json_response = """{{"key1": "value1 is of the type \\"string\\"",// single-line comment
+"key2": "value2"
+/* multiline 
+ comment */}}"""
+
```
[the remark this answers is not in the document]
albertofori: Done

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
```diff
@@ -1121,3 +1122,11 @@ def __flatten_key_value(key, value, flattened_data, depth, separator):
             )  # Ensure boolean values are properly stringified.
     else:
         flattened_data[key] = value if isinstance(value, str) else json.dumps(value)
+
+
+def __parse_json_file_data(json_string):
```
mrm9084: Should we change this to be a json file instead of just a string. I don't think we have any use case here where a file is better.

Also, does the CLI not use type hinting?

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -254,3 +261,73 @@ def parse_tags_to_dict(tags):
...
         return tags_dict
     return tags
+
+
+def strip_json_comments(json_string):
+    current_index = 0
+    length = len(json_string)
+    result = []
+
+    def __is_escaped(json_string, char_index):
```
mrm9084: If we have these as private is there a good reason to define them in the method, it does make it a bit of a challenge to read.

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -254,3 +261,73 @@ def parse_tags_to_dict(tags):
...
+            index -= 1
+
+        return backslash_count % 2 == 1
+
+    def __find_next_newline_index(json_string, start_index):
+        index = start_index
+
+        while index < length:
+            if json_string[index] == NEW_LINE:
+                break
```
mrm9084: Isn't it cleaner to just return here.

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -254,3 +261,73 @@ def parse_tags_to_dict(tags):
...
+
+            index += 1
+
+        return index
+
+    def __find_next_double_quote_index(json_string, start_index):
```
mrm9084: `start_index` seems like an odd name. And unless I'm missing something it can just be index.

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -984,3 +985,12 @@ def list_revision(cmd,
         return retrieved_revisions
     except HttpResponseError as ex:
         raise CLIErrors.AzureResponseError('List revision operation failed.\n' + str(ex))
+
+
+# Helper function to validate JSON input that may contain comments
+def __validate_json_input(json_string):
```
mrm9084: If you ignore my comment above, couldn't we make this the same method as the one above and move it to utilities.

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -254,3 +261,73 @@ def parse_tags_to_dict(tags):
                 tags_dict[tag_key] = tag_value
         return tags_dict
     return tags
+
+
```
mrm9084: Should the json code just be it's own file. There is one main entry point with a decent amount of just helper functions.

mrm9084 (changes requested):

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
albertofori: The return value is already an index. Defining another parameter called "index", to me, is more confusing

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
albertofori: We always return the current index at the end of the method. Just wanted to avoid duplicating that line. But I can update if you feel strongly about this.

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
albertofori: We can define it outside the method. I just felt passing the length of the string around could be avoided. For readability, I can move them into their own method.

src/azure-cli/azure/cli/command_modules/appconfig/_kv_import_helpers.py
albertofori: > Should we change this to be a json file instead of just a string.

Based on the usage, we can pass the file pointer instead, but not the path.

> Also, does the CLI not use type hinting?

Currently, no

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
mrm9084: See my comment below, where I give another idea.

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
mrm9084: I'm not against it, but given they are methods that are only used in once place seems a bit odd, as it's not like you are making them separate methods so you can easily test them or something. It seems like added complexity, but that's just a personal opinion.

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
mrm9084: Not super strongly

mrm9084 (approved):

ChristineWanjau:
Just curious, in portal we don't allow comments for key-vault reference, AI config and Feature flags do we do the same for CLI?

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
albertofori: Updated to have separate methods outside the main method

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
albertofori: Updated. Moved to `_json` file

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
albertofori: Good call out. This has been updated accordingly

albertofori:
> Just curious, in portal we don't allow comments for key-vault reference, AI config and Feature flags do we do the same for CLI?

Good call out. This has been updated to exclude KeyVault refs, feature flags and AI Config as is in portal. Kindly let me know if I am missing any other content type.

Thanks!

src/azure-cli/azure/cli/command_modules/appconfig/keyvalue.py
```diff
@@ -984,3 +986,12 @@ def list_revision(cmd,
         return retrieved_revisions
     except HttpResponseError as ex:
         raise CLIErrors.AzureResponseError('List revision operation failed.\n' + str(ex))
+
+
+def __validate_json_value(json_string, content_type):
+    # We do not allow comments in keyvault references, feature flags, and AI chat completion configs
+    if content_type in (FeatureFlagConstants.FEATURE_FLAG_CONTENT_TYPE, KeyVaultConstants.KEYVAULT_CONTENT_TYPE, AIConfigConstants.AI_CHAT_COMPLETION_CONTENT_TYPE):
+        json.loads(json_string)
+
```
ChristineWanjau: nit: extra line here

albertofori: Updated. Thanks!

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/test_appconfig_json_content_type.py
```diff
@@ -450,3 +473,18 @@ def test_azconfig_json_content_type(self, resource_group, location):
...
+        json_with_comments_file_path = os.path.join(TEST_DIR, 'json_with_comments.json')
+        clean_json_file_path = os.path.join(TEST_DIR, 'json_with_comments_stripped.json')
+
+        with open(json_with_comments_file_path) as json_file:
+            parsed_json = json.loads(strip_json_comments(json_file.read()))
+
+        with open(clean_json_file_path) as json_file:
+            expected_json = json.load(json_file)
+        
+        self.assertEqual(parsed_json, expected_json)
```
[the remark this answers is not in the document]
albertofori: This has been addressed

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -254,3 +261,73 @@ def parse_tags_to_dict(tags):
...
+    result = []
+
+    def __is_escaped(json_string, char_index):
+        backslash_count = 0
+        index = char_index - 1
+        while index >= 0 and json_string[index] == BACKSLASH:
+            backslash_count += 1
+            index -= 1
+
+        return backslash_count % 2 == 1
```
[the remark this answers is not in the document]
albertofori: Function has been moved to new file. Prioritizing existing pattern since methods are quite self explanatory

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -254,3 +261,73 @@ def parse_tags_to_dict(tags):
...
+
+    def __find_next_newline_index(json_string, start_index):
```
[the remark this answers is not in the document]
albertofori: Function has been moved to new file. Prioritizing existing pattern since methods are quite self-explanatory and are only used by the top-level method.

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -254,3 +261,73 @@ def parse_tags_to_dict(tags):
...
+        index = start_index
+        while index < length:
+            if json_string[index] == DOUBLE_QUOTE and not __is_escaped(json_string, index):
+                return index
+
+            index += 1
+
+        raise ValueError("Unterminated string literal")
+
+    def __find_next_multiline_comment_end(json_string, start_index):
```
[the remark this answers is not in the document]
albertofori: Function has been moved to new file. Prioritizing existing pattern since methods are quite self-explanatory and are only used by the top-level method.

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -254,3 +261,73 @@ def parse_tags_to_dict(tags):
...
+
+            index += 1
+
+        return index
+
+    def __find_next_double_quote_index(json_string, start_index):
+        index = start_index
+        while index < length:
+            if json_string[index] == DOUBLE_QUOTE and not __is_escaped(json_string, index):
+                return index
```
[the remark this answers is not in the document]
albertofori: This has been updated accordingly

src/azure-cli/azure/cli/command_modules/appconfig/_utils.py
```diff
@@ -254,3 +261,73 @@ def parse_tags_to_dict(tags):
...
+
+    def __find_next_multiline_comment_end(json_string, start_index):
+        index = start_index
+        while index < length - 1:
+            if json_string[index:index + 2] == MULTILINE_COMMENT_END:
+                return index + 1
+
+            index += 1
+
+        raise ValueError("Unterminated multi-line comment")
```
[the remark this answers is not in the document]
albertofori: This has been updated accordingly

albertofori:
@zhoxing-ms , @yanzhudd , Could you please take a look at this PR and help merge it. We would want this to go with the current release window

yanzhudd:
License check: @microsoft-github-policy-service rerun

albertofori:
Hi @yanzhudd ,

Please I just pushed the change to add the license text at the beginning of the new file. Thanks!

src/azure-cli/azure/cli/command_modules/appconfig/tests/latest/test_appconfig_json_content_type.py
```diff
@@ -450,3 +470,18 @@ def test_azconfig_json_content_type(self, resource_group, location):
...
+
+        json_with_comments_file_path = os.path.join(TEST_DIR, 'json_with_comments.json')
+        clean_json_file_path = os.path.join(TEST_DIR, 'json_with_comments_stripped.json')
+
+        with open(json_with_comments_file_path) as json_file:
+            parsed_json = json.loads(strip_json_comments(json_file.read()))
+
+        with open(clean_json_file_path) as json_file:
+            expected_json = json.load(json_file)
+
```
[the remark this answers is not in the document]
albertofori: Updated

Issue #31250, opened 2025-04-11
Az acr build parses all files in repo even if excluded in .dockerignore


jzazo:
### Describe the bug

I am trying to build a docker image with `az acr build -t <repo/name:tag> -f Dockerfile -r <registry> .` and I have a `.dockerignore` file with `.venv` in it. However `az acr build` parses all files within the dockerignore file, and it is very slow. It just does not ignore the folder, but parses every file within ignored folders. I don't think this should happen

### Related command

`az acr build -t <repo/name:tag> -f Dockerfile -r <registry> .`

### Errors

After more than 30min, I get an error:
```
Packing source code into tar to upload...
[Errno 2] No such file or directory: '.\\.venv\\bin\\python3.12'
```

### Issue script & Debug output

Long logs of the form `cli.azure.cli.command_modules.acr._archive_utils: .dockerignore: no rule for '.venv/lib/python3.12/site-packages/pyright/dist/dist/typeshed-fallback/stubs/TgCrypto'. parent ignore 'True'`

### Expected behavior

Running az acr build should be fast even if I am using a virtual environment

### Environment Summary

azure-cli                         2.71.0                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              core                              2.71.0
telemetry                          1.1.0                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              Extensions:
bastion                            1.4.0
ssh                                2.0.6                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              Dependencies:
msal                            1.31.2b1
azure-mgmt-resource               23.1.1                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              Python location 'C:\Program Files\Microsoft SDKs\Azure\CLI2\python.exe'
Config directory 'C:\Users\javierzazo\.azure'
Extensions directory 'C:\Users\javierzazo\.azure\cliextensions'                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       Python (Windows) 3.12.8 (tags/v3.12.8:2dc476b, Dec  3 2024, 19:30:04) [MSC v.1942 64 bit (AMD64)]

Legal docs and information: aka.ms/AzureCliLegal


Your CLI is up-to-date.

### Additional context

_No response_

yonzhan:
Thank you for opening this issue, we will look into it.

jzazo:
[...]

PR #32241, opened 2025-10-10
{ACR} `az acr build`: Exclude .venv directories from build

SakoPak:
Add exclusion for .venv directories to improve performance.

**Related command**
az acr build

**Description**
Az acr build parses all files in repo even if excluded in .dockerignore. This PR targets .venv directories from getting processed thus reducing the unnecessarily long build time.

**History Notes**
Issue raised here: https://github.com/Azure/azure-cli/issues/31250

---

- [x ] The PR title and description has followed the guideline in [Submitting Pull Requests](https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules#submitting-pull-requests).

- [x ] I adhere to the [Command Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md).

### DIFF
--- a/src/azure-cli/azure/cli/command_modules/acr/_archive_utils.py
+++ b/src/azure-cli/azure/cli/command_modules/acr/_archive_utils.py
@@ -213,7 +213,7 @@ def _archive_file_recursively(tar, name, arcname, parent_ignored, parent_matchin
 
     # Even if a dir is ignored by .dockerignore, its children might still be included (Docker semantics),
     # so we continue scanning unless it is in our build_ignore_dirs set (performance skip list).
-    if tarinfo.isdir() and not (os.path.basename(tarinfo.name) in build_ignore_dirs):
+    if tarinfo.isdir() and os.path.basename(tarinfo.name) not in build_ignore_dirs:
         for f in os.listdir(name):
             _archive_file_recursively(tar, os.path.join(name, f), os.path.join(arcname, f),
                                       parent_ignored=ignored, parent_matching_rule_index=matching_rule_index,


yonzhan:

src/azure-cli/azure/cli/command_modules/acr/_archive_utils.py
```diff
@@ -66,13 +66,19 @@ def _pack_source_code(source_location, tar_file_path, docker_file_path, docker_f
...
+    build_ignore_dirs = {'.venv'}
 
     def _ignore_check(tarinfo, parent_ignored, parent_matching_rule_index):
         # ignore common vcs dir or file
         if tarinfo.name in common_vcs_ignore_list:
             logger.warning("Excluding '%s' based on default ignore rules", tarinfo.name)
             return True, parent_matching_rule_index
 
+        # Exclude performance-impacting directories that are typically not needed in build context
+        if tarinfo.isdir() and os.path.basename(tarinfo.name) in build_ignore_dirs:
```
northtyphoon, line 80: Can you check if the code will skip all files under the ignored directories? or will it continue to scan all of them?

SakoPak: I added line 214 to exclude only the .venv folder ( in "ignored") from being scanned, since we agreed we are doing a targeted fix to only exclude venv. The code will still scan the child items of .git and .docker_ignore, even though those are in ignored group, according to original comment and code on line 212-213 (  "even the dir is ignored, its child items can still be included, so continue to scan"). This PR change will exclude only .venv from this scan.

src/azure-cli/azure/cli/command_modules/acr/_archive_utils.py
```diff
@@ -66,13 +66,19 @@ def _pack_source_code(source_location, tar_file_path, docker_file_path, docker_f
     original_docker_file_name = os.path.basename(docker_file_path.replace("\\", os.sep))
     ignore_list, ignore_list_size = _load_dockerignore_file(source_location, original_docker_file_name)
     common_vcs_ignore_list = {'.git', '.gitignore', '.bzr', 'bzrignore', '.hg', '.hgignore', '.svn'}
+    build_ignore_dirs = {'.venv'}
 
     def _ignore_check(tarinfo, parent_ignored, parent_matching_rule_index):
         # ignore common vcs dir or file
         if tarinfo.name in common_vcs_ignore_list:
```
northtyphoon: I think it's a latent bug. `tarinfo.name` is the full file path name in the tar. We need to extract the dir or file name like you does `os.path.basename(tarinfo.name)`.

src/azure-cli/azure/cli/command_modules/acr/_archive_utils.py
```diff
@@ -202,11 +210,12 @@ def _archive_file_recursively(tar, name, arcname, parent_ignored, parent_matchin
             tar.addfile(tarinfo)
 
     # even the dir is ignored, its child items can still be included, so continue to scan
-    if tarinfo.isdir():
+    # Exception: Skip scanning performance-impacting directories entirely
+    if tarinfo.isdir() and not (ignored and os.path.basename(tarinfo.name) in build_ignore_dirs):
```
northtyphoon: why do we need to check `ignored` here?

SakoPak: removed since it's redundant.

src/azure-cli/azure/cli/command_modules/acr/_archive_utils.py
SakoPak: updated tarinfo.name

PR #32264, opened 2025-10-15
{App Service} `az functionapp create`: Stack API to take SKU for Flex

khkh-ms:
**Related command**
- az functionapp list-flexconsumption-locations
- az functionapp create --resource-group {RESOURCE_GROUP} --name {APP_NAME} --storage-account {STORAGE_NAME} --flexconsumption-location {REGION} --runtime python --runtime-version 3.11 

**Description**
Updated the stacks API to take SKU as parameter/query string. Also make the Flex specific stacks API class to be generic and able to take SKU as input. 

**Testing Guide**
az functionapp list-flexconsumption-locations

---


yonzhan:

yanzhudd:
please fix the CI issues

yanzhudd:
please note that the code completion date is on 10/28/2025 at 07:00 UTC for the upcoming release. Please address the comments asap, otherwise it has to be postponed to next sprint (11/18)

zhoxing-ms:
> {App Service} az functionapp create: Stack API to take SKU for Flex

May I ask why the title uses `{}`? Is the change in PR imperceptible to customers?

src/azure-cli/azure/cli/command_modules/appservice/custom.py
```diff
@@ -5621,12 +5622,11 @@ def stacks(self):
         self._load_stacks()
         return self._stacks
 
-    def get_flex_raw_function_app_stacks(self, cmd, location, runtime):
-        stacks_api_url = '/providers/Microsoft.Web/locations/{}/functionAppStacks?' \
-                         'api-version=2020-10-01&removeHiddenStacks=true&removeDeprecatedStacks=true&stack={}'
+    def get_raw_function_app_stacks(self, cmd, location, runtime, sku):
+        stacks_api_url = '/providers/Microsoft.Web/locations/{}/functionAppStacks?api-version=2023-01-01&removeHiddenStacks=true&removeDeprecatedStacks=true&stack={}&sku={}'  # pylint: disable=line-too-long
```
kamperiadis: Instead of passing "FC1" from every calling function, why don't you just make it so that this url is

 stacks_api_url = '/providers/Microsoft.Web/locations/{}/functionAppStacks?api-version=2023-01-01&removeHiddenStacks=true&removeDeprecatedStacks=true&stack={}&**sku=FC1**'

This way, we don't have to update all existing functions. I think we should only accept a sku param to the RuntimeHelper class and methods if this class is used for stuff other than flex and this sku value can vary based on those applications. But my understanding is that all of this is for flex and we would have to use the same sku value right?

khkh-ms: Made the class as generic in terms of SKU because the stacks API is not Flex specific anymore.

khkh-ms: Changed as recommended. Will make the class generic in a different PR. thanks!

src/azure-cli/azure/cli/command_modules/appservice/tests/latest/test_functionapp_commands.py
```diff
@@ -2419,6 +2419,19 @@ def test_functionapp_on_linux_version_error(self, resource_group, storage_accoun
         self.cmd('functionapp create -g {} -n {} --plan {} -s {} --runtime python --runtime-version 4.8'
                  .format(resource_group, functionapp, plan, storage_account), expect_failure=True)
 
+    @ResourceGroupPreparer(location=LINUX_ASP_LOCATION_FUNCTIONAPP)
+    @StorageAccountPreparer()
+    def test_functionapp_on_linux_consumption_python_latest(self, resource_group, storage_account):
```
kamperiadis, line 2424: How is this test relevant in testing the changes you made in custom.py if that _FlexFunctionAppStackRuntimeHelper class is for Flex specific apps?

kamperiadis: Resolved offline

khkh-ms: For context: The test was removed by azcli team because it was breaking and blocking the PRs. I am adding the adding the test back after fixing it as part of the PR. thanks!

yanzhudd (approved):

Issue #32259, opened 2025-10-14
`az postgres flexible-server restore` failing with code "NoRegisteredProviderFound"


nnellanspdl:
### Describe the bug

Here is my full command:
```
az postgres flexible-server restore \
  --restore-time <time> \
  --source-server <sourceServerName> \
  --resource-group <rgName> \
  --name <newServerName> \
  --private-dns-zone <resourceIdExistingZone> \
  --subnet <resourceIdExistingSubnet> \
  --debug
```

I am trying to create a new restore server from a backup.

I'm attempting to use an existing Virtual Network, so I am supplying the `--subnet` switch and pointing it at the full resource ID of the existing subnet.

In the debug logs I can see an API call that is attempting to capture information about this existing virtual network:  `DEBUG: urllib3.connectionpool: [https://management.azure.com:443](https://management.azure.com/) "GET /subscriptions/<mySub>/resourcegroups/<myRG>/providers/Microsoft.Network//virtualNetworks/<myVnet>?api-version=2025-05-01 HTTP/1.1" 400 1738`

Observation:  I see a double slash `//` in the path above, could that be a problem?

[...]

PR #32271, opened 2025-10-16
{RDBMS} Return default API version if available in _resolve_api_version function

nachoalonsoportillo:
**Related command**
az postgres flexible-server restore --subnet <resourceIdExistingSubnet>

**Description**
Fixes https://github.com/Azure/azure-cli/issues/32259

Microsoft.Network namespace announced in their most recent ARM manifest a version `2025-05-01` which is theoretically valid for resourceType `virtualNetworks` but is not. Any attempts to call any of the APIs exposed by resources of that type with that API version are erroring out.

If the GET Providers API returns a value for `defaultApiVersion` for that resource type, we should use that instead to avoid problems like this in the future, if they happen to do this again.

**Testing Guide**

**History Notes**
[RDBMS] az postgres flexible-server restore: Fix issue that produces NoRegisteredProviderFound error


yonzhan:

calvinhzy (approved):

PR #32398, opened 2025-11-07
[RDBMS] BREAKING CHANGE `az postgres flexible-server create`: Remove default value to `--version` and remove arguments `--create-default-database` and `--database-name`

nasc17:
**Related command**
`az postgres flexible-server create`

**Description**
PG default version Currently 'PG17'. Remove hardcoded value being set for version, instead utilize capabilities API.
Last breaking change we removed having a database be created as default for every server. Remove the argument '--create-default-database' and stick to having customers use the 'az postgres flexible-server database' group, if they want to create any databases.

**Testing Guide**
Manual

**History Notes**
[RDBMS] BREAKING CHANGE: `az postgres flexible-server create`: Remove default value to `--version` and remove arguments `--create-default-database` and `--database-name`


yonzhan:

nasc17:
@yonzhan @evelyn-ys @calvinhzy please trigger test checks

nasc17:
> LGTM.
> 
> Once we upgrade Python SDK to the one version supporting 2025-08-01 API, we'll have to introduce support for users to override the default database name for Elastic Clusters. At that point, we might want to introduce a new parameter called `--default-database-name` with a default of `postgres`. And the parameter should only be present when `--cluster-option` is set to `ElasticCluster`.

Let's talk more about this for planning SDK update :-)

Issue #30427, opened 2024-11-27
`az ad app create` and `az ad sp create-for-rbac` should not be idempotent


jiasli:
**Related command**
`az ad app create`
`az ad sp create-for-rbac`

**Is your feature request related to a problem? Please describe.**
Derived from IcM 568027592

#9108 made `az ad app create` and `az ad sp create-for-rbac` use `displayName` to uniquely identify an application. If an application with the specified `displayName` already exists, Azure CLI will `PATCH` it with the newly provided properties by calling [Update application](https://learn.microsoft.com/en-us/graph/api/application-update?view=graph-rest-1.0&tabs=http) API. This API requires [`microsoft.directory/applications/basic/update`](https://learn.microsoft.com/en-us/entra/identity/role-based-access-control/custom-available-permissions#microsoftdirectoryapplicationsbasicupdate) permission that a user may not possess. The user may only have [`microsoft.directory/applications/createAsOwner`](https://learn.microsoft.com/en-us/entra/identity/role-based-access-control/custom-available-permissions#microsoftdirectoryapplicationscreateasowner) and expect these commands to work.

This behavior is not aligned with the underlying [Create application](https://learn.microsoft.com/en-us/graph/api/application-post-applications?view=graph-rest-1.0&tabs=http) API where `displayName` is not used as a primary key and multiple applications can have the same `displayName`. Creating applications with the same `displayName` will create multiple applications.

**Describe the solution you'd like**
Change this behavior to align with the underlying [Create application](https://learn.microsoft.com/en-us/graph/api/application-post-applications?view=graph-rest-1.0&tabs=http) API, PowerShell and Entra portal to eliminate possible confusions. In other words, the change will make calling `az ad app create` or `az ad sp create-for-rbac` multiple times create multiple applications with the same `displayName`.

**Describe alternatives you've considered**
Explicitly document this behavior and define it as by-design.

**Additional context**

yonzhan:
Thank you for opening this issue, we will look into it.

keystroke:
@yonzhan this is a crazy security issue and this behavior needs to be removed ASAP!

**Remove any API calls that look objects up by display name. This is fundamentally inappropriate. Display name is not unique, and can be edited at any time, and multiple applications / service principals can have the same display name.**

We observe same repo on version 2.76 and this behavior can cause customers to disrupt an existing application and take it over, and assign permissions to the completely wrong identity!!!

[...]

PR #32437, opened 2025-11-19
{Role} Add warning about using non-unique display names when creating service principal

ReaNAiveD:
**Related command**

`az ad app create`
`az ad sp create-for-rbac`

**Description**

Add a in-tool warning about using non-unique display names to create a service principal.
image
image
image
image

For more details, see  https://github.com/Azure/azure-cli/issues/30427

**Testing Guide**

**History Notes**

{Role} `az ad app create`: Add warning about using non-unique display names in `az ad app create`
{Role} `az ad sp create-for-rbac`: Add warning about using non-unique display names in `az ad sp create-for-rbac`

---


yonzhan:

src/azure-cli/azure/cli/command_modules/role/custom.py
```diff
@@ -1017,6 +1017,10 @@ def app_federated_credential_delete(client, app_identifier, federated_identity_c
 
 
 def create_service_principal(cmd, identifier):
+    logger.warning("The `az ad sp create` command can modify an existing application or service principal "
```
isra-fel: Should we consider displaying the warning only when the command is overwriting an existing SP?

dcaro: That would be preferable.

src/azure-cli/azure/cli/command_modules/role/_help.py
```diff
@@ -508,6 +508,10 @@
 helps['ad sp create'] = """
 type: command
 short-summary: Create a service principal.
+long-summary: >-
+    **IMPORTANT**: The `az ad sp create` command can modify an existing application or service principal if
+    another object shares the same **display name**. Display names aren't unique and can change, which
+    could result in credential loss or incorrect RBAC assignments. Use a **unique object ID or app ID** instead.
```
dcaro: Can we consider a shorter message with a link to our conceptual docs where the user can learn about the issue and have some code to help finding the existing application / sp ?

src/azure-cli/azure/cli/command_modules/role/_help.py
```diff
@@ -508,6 +508,10 @@
 helps['ad sp create'] = """
 type: command
 short-summary: Create a service principal.
+long-summary: >-
+    **IMPORTANT**: The `az ad sp create` command can modify an existing application or service principal if
```
jiasli: Only `az ad app create` and `az ad sp create-for-rbac` can modify an existing application. `az ad sp create` will not.

jiasli: We usually don't use Markdown syntax besides `` `code` ``, as this will make the in-tool help hard to read.

src/azure-cli/azure/cli/command_modules/role/custom.py
```diff
@@ -1145,6 +1149,11 @@ def create_service_principal_for_rbac(
         show_auth_in_json=None, skip_assignment=False, keyvault=None):
     import time
 
+    logger.warning("The `az ad sp create-for-rbac` command can modify an existing application or service principal "
```
jiasli: Same here. I don't really like the idea of showing this warning unconditionally.

src/azure-cli/azure/cli/command_modules/role/custom.py
ReaNAiveD: > Should we consider displaying the warning only when the command is overwriting an existing SP?

We already have a warning when overwriting an existing SP.
image
I have added the new warning before it.

src/azure-cli/azure/cli/command_modules/role/_help.py
ReaNAiveD: Changed

ReaNAiveD: Changed to double quota, i.e. `"az ad app create"`. Also removed the markdown symbol like `*`

src/azure-cli/azure/cli/command_modules/role/custom.py
ReaNAiveD: Move the warning to the `create_appication` method. The warning will only show when there is exact one existing app with the same `display_name` and before the existing warning `Found an existing application instance: (id) %s. We will patch it.`
image

src/azure-cli/azure/cli/command_modules/role/_help.py
```diff
@@ -22,7 +22,14 @@
 helps['ad app create'] = """
 type: command
 short-summary: Create an application.
-long-summary: For more detailed documentation, see https://learn.microsoft.com/graph/api/resources/application
+long-summary: >-
+    IMPORTANT: The "az ad app create" command can modify an existing application or service principal if
```
jiasli: We usually use `WARNING:` instead of `IMPORTANT:`. See #16081

src/azure-cli/azure/cli/command_modules/role/_help.py
```diff
@@ -22,7 +22,14 @@
...
+    another object shares the same display name. Display names aren't unique and can change, which
+    could result in credential loss or incorrect RBAC assignments. Use a unique object ID or app ID instead.
```
jiasli: It is only possible for the user to choose a unique **display name**. The user cannot choose to use a unique **object ID** or **app ID**. The **object ID** or **app ID** is generated by the MS Graph service.

src/azure-cli/azure/cli/command_modules/role/_help.py
```diff
@@ -22,7 +22,14 @@
...
+    For more details, see https://go.microsoft.com/fwlink/?linkid=2342455.
```
jiasli, line 29: This link doesn't give more details than this help message:

image

src/azure-cli/azure/cli/command_modules/role/_help.py
```diff
@@ -22,7 +22,14 @@
...
+
+
+    For more detailed documentation, see https://learn.microsoft.com/graph/api/resources/application
```
jiasli: It's better to describe what documentation - the Microsoft Entra application, not this `az ad app create` command:

```suggestion
    For more detailed documentation on Microsoft Entra application, see https://learn.microsoft.com/graph/api/resources/application
```

src/azure-cli/azure/cli/command_modules/role/_help.py
```diff
@@ -508,6 +515,7 @@
 helps['ad sp create'] = """
 type: command
 short-summary: Create a service principal.
+long-summary:
```
jiasli: The long summary should not be empty.

src/azure-cli/azure/cli/command_modules/role/custom.py
```diff
@@ -615,6 +615,12 @@ def create_application(cmd, client, display_name, identifier_uris=None,
             raise CLIError("More than one application have the same display name '{}': (id) {}, please remove "
                            'them first.'.format(display_name, ', '.join([x[ID] for x in existing_apps])))
         if len(existing_apps) == 1:
+            logger.warning("IMPORTANT: The \"az %s\" command can modify an existing application or service principal "
+                           "if another object shares the same display name. "
+                           "Display names aren't unique and can change, "
+                           "which could result in credential loss or incorrect RBAC assignments. "
+                           "Use a unique object ID or app ID instead. For more details, "
+                           "see https://go.microsoft.com/fwlink/?linkid=2342455.", cmd.name)
```
jiasli: Since `len(existing_apps) == 1` already evaluates to `True`, that means there is already an application that shares the same display name, so using conditional sentence (`can` or `if`) is not unnecessary. 

It's better to combine this warning message with L624 and say:

> Please notice that display names aren't unique, which could result in credential loss or incorrect RBAC assignments. To create a new application, use a unique display name instead. For more details, see https://go.microsoft.com/fwlink/?linkid=2342455.

src/azure-cli/azure/cli/command_modules/role/_help.py
ReaNAiveD: Yes. It might help to reach out to Mike and Alex to see if they can add more context to the documentation.

src/azure-cli/azure/cli/command_modules/role/_help.py
ReaNAiveD: Removed the sentence `Use a unique object ID or app ID instead.`

src/azure-cli/azure/cli/command_modules/role/_help.py
ReaNAiveD: Updated

src/azure-cli/azure/cli/command_modules/role/_help.py
ReaNAiveD: Added

src/azure-cli/azure/cli/command_modules/role/_help.py
ReaNAiveD: Removed

src/azure-cli/azure/cli/command_modules/role/custom.py
ReaNAiveD: I’ve updated the warning as suggested. But I also reserved the warning line indicating the existing application ID to ensure clarity.

PR #32455, opened 2025-11-25
{Core} Add distinct telemetry events for `--version`

DanielMicrosoft:
**Related command**

**Description**
ADO Task: https://msazure.visualstudio.com/One/_workitems/edit/35972904

Previously, `az --version` telemetry events would have no distinct markers to allow for deliberate searching/queries of telemetry records. This PR adds those so that `az --version` telemetry events can be filtered and queried from Kusto. 

**Testing Guide**

Please run `az --version` from this branch and capture the `Correlationid` and use it to query the kusto instance to observe the telemetry. 

Note: an example event initiated from this branch can be found in Kusto using this condition:
```
where CorrelationId == "19692759-ab8b-4c25-b922-1d4ba4b69efa"
```

For comparison purposes, here is an example of a call to `az --version` without the changes in this PR:
```
where CorrelationId == "ac718bcf-0562-478d-beb8-084e32ed385d"
```
**History Notes**

---

### DIFF
--- a/src/azure-cli-core/azure/cli/core/tests/test_telemetry.py
+++ b/src/azure-cli-core/azure/cli/core/tests/test_telemetry.py
@@ -4,6 +4,7 @@
 # --------------------------------------------------------------------------------------------
 
 import unittest
+from unittest import mock
 
 
 class TestCoreTelemetry(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


yonzhan:

src/azure-cli-core/azure/cli/core/__init__.py
```diff
@@ -120,6 +120,10 @@ def get_cli_version(self):
 
     def show_version(self):
         from azure.cli.core.util import get_az_version_string, show_updates
+        from azure.cli.core import telemetry
+
+        telemetry.set_raw_command_name("--version")
+        telemetry.set_command_details("version")
```
DanielMicrosoft: Calling `set_command_details` with only this param is essentially superfluous (nothing additional makes it to telemetry), but we may consider other properties we may want?

yanzhudd: from my understanding, `raw_command` should represent the command name, which is an empty string in this case, and parameters should be `["--version"]`. I think it’s better to stay consistent with the existing telemetry naming conventions
```suggestion
        telemetry.set_raw_command_name("")
        telemetry.set_command_details(command="", parameters=["--version"])
```

DanielMicrosoft: Thank you, good call out. 

I removed the call to `telemetry.set_raw_command_name("")` as it is redundant now.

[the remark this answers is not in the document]
DanielMicrosoft: added test

yanzhudd (approved):

.vscode/launch.json
```diff
@@ -12,33 +12,24 @@
...
-                "WaitOnAbnormalExit",
-                "WaitOnNormalExit",
-                "RedirectOutput"
-            ],
             "justMyCode": false
         },
         {
             "name": "Azure CLI Debug (External Console)",
-            "type": "python",
+            "type": "debugpy",
```
DanielMicrosoft, line 19: updating this property as recommended by linter

.vscode/launch.json
```diff
@@ -12,33 +12,24 @@
                 "--help"
             ],
             "console": "integratedTerminal",
-            "debugOptions": [
```
DanielMicrosoft, line 15: removing values as deprecated

PR #32572, opened 2025-12-23
{Compute} `az vm identity`: Migrate commands to aaz-based implementation

william051200:
**Related command**
`az vm identity assign`
`az vm identity remove`
`az vm identity show`

**Description**

Migration from mgmt.compute to aaz-based

**Testing Guide**

**History Notes**

---


yonzhan:

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -826,8 +825,14 @@ def create_snapshot(cmd, resource_group_name, snapshot_name, location=None, size
...
-    client = _compute_client_factory(cmd.cli_ctx)
-    return client.virtual_machines.get(resource_group_name, vm_name).identity
+    vm = get_vm_by_aaz(cmd, resource_group_name, vm_name)
+
+    identity = vm.get("identity", {}) if vm else None
+
+    if identity and not identity.get('userAssignedIdentities'):
+        identity['userAssignedIdentities'] = None
+
+    return identity or None
```
yanzhudd, line 835: We may need to consider whether this should return an empty dictionary (`{}`) or `None` when there is no identity on the VM. We can refer to the SDK behavior for guidance.

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -837,49 +842,63 @@ def show_vmss_identity(cmd, resource_group_name, vm_name):
...
+        return get_vm_by_aaz(cmd, resource_group_name, vm_name)
 
     def setter(vm, external_identities=external_identities):
-        if vm.identity and vm.identity.type == ResourceIdentityType.system_assigned_user_assigned:
-            identity_types = ResourceIdentityType.system_assigned_user_assigned
-        elif vm.identity and vm.identity.type == ResourceIdentityType.system_assigned and external_identities:
-            identity_types = ResourceIdentityType.system_assigned_user_assigned
-        elif vm.identity and vm.identity.type == ResourceIdentityType.user_assigned and enable_local_identity:
-            identity_types = ResourceIdentityType.system_assigned_user_assigned
+        if vm.get('identity') and vm.get('identity').get('type') == system_assigned_user_assigned:
```
yanzhudd: just for your reference, adding a default value to the `get()` function of `dict` might be more concise in this case
```suggestion
        if vm.get('identity', {}).get('type', None) == system_assigned_user_assigned:
```

src/azure-cli/azure/cli/command_modules/vm/custom.py
william051200: image

Here is the output from SDK behavior, it returns `None` instead of of ( { } ) for empty result.

src/azure-cli/azure/cli/command_modules/vm/custom.py
```diff
@@ -837,49 +842,63 @@ def show_vmss_identity(cmd, resource_group_name, vm_name):
...
-                                                                                        'VirtualMachineUpdate')
-    UserAssignedIdentitiesValue = cmd.get_models('UserAssignedIdentitiesValue')
-    from azure.cli.core.commands.arm import assign_identity as assign_identity_helper
-    client = _compute_client_factory(cmd.cli_ctx)
-    _, _, external_identities, enable_local_identity = _build_identities_info(assign_identity)
+    identity, _, external_identities, enable_local_identity = _build_identities_info(assign_identity)
+
+    system_assigned = "SystemAssigned"
+    user_assigned = "UserAssigned"
+    system_assigned_user_assigned = "SystemAssigned, UserAssigned"
```
yanzhudd: since they are already defined in _vm_utils.py file, it would be better to leverage the definition across all the places that need to read the values

src/azure-cli/azure/cli/command_modules/vm/operations/vm.py
```diff
@@ -155,6 +158,114 @@ def _output(self, *args, **kwargs):
...
+
+        identity = result.get('identity')
+        if not identity:
+            return result
+
+        if not identity.get('principalId'):
+            identity['principalId'] = None
+
+        if not identity.get('tenantId'):
+            identity['tenantId'] = None
```
yanzhudd: may I ask why we need to add the `None` value to the result?

src/azure-cli/azure/cli/command_modules/vm/operations/vm.py
```diff
@@ -155,6 +158,114 @@ def _output(self, *args, **kwargs):
...
+                content['identity'] = {
+                    'userAssignedIdentities': None,
+                    'type': IdentityType.NONE.value
+                }
+                return json.dumps(content)
+
+            identities = content.get('identity', {}).get('userAssignedIdentities')
+            if identities:
+                if 'UserAssigned' in identities.keys():
+                    identities.pop('UserAssigned')
```
yanzhudd, line 236: May I ask what the identities look like in this case?

src/azure-cli/azure/cli/command_modules/vm/operations/vm.py
```diff
@@ -155,6 +158,114 @@ def _output(self, *args, **kwargs):
...
+
+                for key in list(identities.keys()):
+                    identities[key] = None
+
+            if not content.get('identity', {}).get('userAssignedIdentities', {}):
+                content['identity']['userAssignedIdentities'] = None
+
+            return json.dumps(content)
+
+        def __call__(self, *args, **kwargs):
```
yanzhudd, line 246: Why do we need to manually write this \_\_call\_\_() function here, as I remember it is already defined in the parent class

yanzhudd:
since the changes are not visible to users, the title should be started with `{Compute}` instead of `[Compute]`
for more information, please refer to https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules#submitting-pull-requests

yanzhudd:
could you please double check the changes to the recording yaml files? We need to ensure there is no difference for the command behavior after migrating to aaz-based

src/azure-cli/azure/cli/command_modules/vm/operations/vm.py
william051200: I have just validated, I was using them when running some test, it is now redundant. I will be removing those lines. Thank you.

src/azure-cli/azure/cli/command_modules/vm/operations/vm.py
william051200: The reason I have implemented this solution is because:
image
1. VM Patch will always set userAssigned and systemAssigned action flag to `create`.
2. VM Patch only accept a list of userAssigned identities, then `AAZIdentityObject` will always format the identity to create based on the `create` action.
3. Therefore, I will only pass <b>identity to be removed</b> to the list as I did a little twist by inheriting the original VM Patch, then set the <b>identities to be removed</b> into this format `{'identity':None}`. (I found this is the only format to remove userAssigned identity from the list)
4. It works until I need to remove systemAssigned identity from systemAssigned + userAssigned identities, this is because passing userAssigned as empty list (because I am not removing any userAssigned identity) and `AAZIdentityObject` will be converting the identity to `None` instead of `UserAssigned`, cause the removal of systemAssigned + userAssigned identities.
5. The workaround is, I have added a placeholder into the userAssigned identity list and send to customized aaz in this format `{'resource_group': 'william-rg', 'vm_name': 'william-vm1', 'mi_user_assigned': ['UserAssigned']}`. This way, AAZIdentityObject will set the type of identity to `UserAssigned`, hence retain the userAssigned identities.
6. Then I remove the placeholder afterwards in the inherited function.

Result:
image

src/azure-cli/azure/cli/command_modules/vm/operations/vm.py
william051200: Yes, the __call__ function exists but is it not workable for 'removing identity using vm patch class'. Reason as:

This is _update.py, it does support `pre_instance_update`, which modifies the content of request before calling the update API.
image

This is _patch.py, it does not support `pre_instance_update`, and I do not find any other function to perform the same function as `pre_instance_update`
image

And that is why I have created `__call__` function and added one function to modify the content
image

src/azure-cli-core/azure/cli/core/commands/arm.py
```diff
@@ -759,46 +759,50 @@ def _find_property(instance, path):
...
-                if retry_time < retry_times and ' does not exist in the directory ' in ex.message:
-                    time.sleep(5)
-                    logger.warning('Retrying role assignment creation: %s/%s', retry_time + 1,
-                                   retry_times)
-                    continue
-                raise
     return resource
 
 
+def create_role_assignment(cli_ctx, principal_id, identity_role=None, identity_scope=None):
```
yanzhudd, line 774: Hi @zhoxing-ms could you please help review the changes to this file?
It simply separates the logic for calling assignments_client.create() from the assign_identity() function, since we also need to invoke assignments_client.create() in the vm module

zhoxing-ms: Looks good to me

zhoxing-ms (approved):

yanzhudd (approved):

PR #32767, opened 2026-02-10
[App Config] `az appconfig create/update`: Add support to link store with azure front door profile

ChristineWanjau:
**Related command**

az appconfig create/update

**Description**

This PR adds support to link azure front door profile to app configuration store.

**Testing Guide**

**History Notes**

[App config] `az appconfig create/update`: Add support to link azure front door profile

---


yonzhan:

src/azure-cli/azure/cli/command_modules/appconfig/custom.py
```diff
@@ -70,6 +72,10 @@ def create_configstore(cmd,  # pylint: disable=too-many-locals
     if arm_auth_mode is not None:
         arm_authentication_mode = AuthenticationMode.LOCAL if arm_auth_mode == ARMAuthenticationMode.LOCAL else AuthenticationMode.PASS_THROUGH
 
+    azure_front_door = None
+    if azure_front_door_profile is not None:
+        azure_front_door = AzureFrontDoorProperties(resource_id=azure_front_door_profile if azure_front_door_profile else None)
```
albertofori, line 77: Is the inline `if` check needed since line 76 confirms the front door profile is not none?

mrm9084: The `AzureFrontDoorProperties` object has `resource_id` as optional, I'd just remove both if statements. i.e. just have this as line 76. See: https://github.com/Azure/azure-sdk-for-python/blob/2d7c1259d52c5ac41aaedf7581478e2ede32b575/sdk/appconfiguration/azure-mgmt-appconfiguration/azure/mgmt/appconfiguration/models/_models.py#L68. I'd rather have the `None` `resource_id` that is already typed then the extra `None` properties object.

```python
azure_front_door = AzureFrontDoorProperties(resource_id=azure_front_door_profile)
```

src/azure-cli/setup.py
```diff
@@ -65,7 +65,7 @@
     'azure-keyvault-securitydomain==1.0.0b1',
     'azure-mgmt-advisor==9.0.0',
     'azure-mgmt-apimanagement==4.0.0',
-    'azure-mgmt-appconfiguration==5.0.0',
+    'azure-mgmt-appconfiguration==6.0.0b1',
```
mrm9084, line 68: Are we planning on releasing this as a beta from main? or are we expecting the ga of `6.0.0` before then? This will block main from GA releases until this GAs.

src/azure-cli/azure/cli/command_modules/appconfig/custom.py
ChristineWanjau: This makes sense. I have updated accordingly. 
Left one check because when user wants to unlink their afd they pass an empty string. The RP doesn't allow an empty string as a valid property, we should be passing None for such a case.
image

src/azure-cli/setup.py
ChristineWanjau: Do you mean the CLI release? 
I think it's allowed to release using a beta sdk in the CLI. I see multiple beta sdks in the setup.py https://github.com/Azure/azure-cli/blob/d01027d24174a819c801343be8a8404ccff24ef2/src/azure-cli/setup.py#L79

I added is_preview tag in the afd argument.
image
cc: @albertofori thoughts? Is this blocking?

mrm9084 (approved):

src/azure-cli/setup.py
albertofori: @ChristineWanjau just to confirm, this beta release should already have Telemetry in the Configuration Store definition, right?

ChristineWanjau: yes it should

albertofori: Got it, then I don't think there should be anything that we anticipate being GA'd soon so this should be fine. @mrm9084, please correct me if I am wrong or missing something

src/azure-cli/azure/cli/command_modules/appconfig/custom.py
albertofori: Just curious for the update scenario. Assuming user has already linked an AFD resource and wants to update a different property, will the linked AFD resource be cleared since they will not be specifying it? And if so, is this behavior consistent with other properties? I ask since it appears that we will always pass `AzureFrontDoorProperties(None)` when the resource_id is not specified, which might be equivalent payload to unlinking a resource.

Is that the case?

ChristineWanjau: Yeah thats the case. Just updated the logic

albertofori: So I guess the expected behavior is that passing no resource id means no change in update scenario, right? User explicitly passes empty string to unlink.

mrm9084: @ChristineWanjau I see we have a test for adding and removing the profile. Can we add a test for making sure it's not removed if it isn't included?

ChristineWanjau: Yes that makes sense to me. I have added a test for such a case.

yanzhudd (approved):

PR #32831, opened 2026-02-23
[App Config] `az appconfig create/update/feature set`: Add support to enable telemetry

ChristineWanjau:
**Related command**

**Description**

This PR adds support to link an app insight resource to an app configuration store and enable telemetry to a feature flag.

**Testing Guide**

**History Notes**

[AppConfig] `az appconfig create/update`: Enable linking app insights resource to an app configuration store
[AppConfig] `az appconfig feature set`: Enable telemetry for a feature flag

---


yonzhan:

src/azure-cli/azure/cli/command_modules/appconfig/feature.py
```diff
@@ -30,24 +30,54 @@
...
+    try:
+        resource_group_name, _ = resolve_store_metadata(cmd, store_name)
+        configstore_client = cf_configstore(cmd.cli_ctx)
+        store = configstore_client.get(resource_group_name, store_name)
+
+        telemetry = getattr(store, "telemetry", None)
+        is_linked = bool(getattr(telemetry, "resource_id", None)) if telemetry else False
+
+        if not is_linked:
+            logger.warning(
```
mrm9084, line 62: What is the logic behind this being a warning? This is actually something not supported by our portal UI.

Also, the warning is a bit wrong. If we do allow this, assuming the customer setup there application correctly, it will generate telemetry. It will just not be visible via the App Configuration Azure Portal.

src/azure-cli/azure/cli/command_modules/appconfig/feature.py
```diff
@@ -30,24 +30,54 @@
...
+                "App Insights resource for the App Configuration store is not set."
+                "To collect telemetry, connect to an App Insights resource."
+            )
+
+    except Exception as ex:  # pylint: disable=broad-except
+        logger.warning(
+            "Unable to verify App Insights resource for the App Configuration store: %s", str(ex)
```
mrm9084, line 69: This also links to my comment on line 62. What is the actual failure case her in the verify? They have a linked store that was deleted, but still linked? Are we able to give a more helpful error message on our part.

src/azure-cli/azure/cli/command_modules/appconfig/feature.py
```diff
@@ -135,6 +171,16 @@ def set_feature(cmd,
...
 
+                # Update telemetry if telemetry_enabled is specified
+                if telemetry_enabled is not None:
+                    if feature_flag_value.telemetry is None:
+                        feature_flag_value.telemetry = FeatureTelemetry(enabled=telemetry_enabled)
+                    else:
+                        feature_flag_value.telemetry.enabled = telemetry_enabled
+
+                    if telemetry_enabled:
+                        warn_if_app_insights_not_set(cmd, name)
```
mrm9084: I might be missing something, but this check will run twice if this is hit as it will have already been checked on line 121.

src/azure-cli/azure/cli/command_modules/appconfig/feature.py
ChristineWanjau: On Portal, we show this warning and we disable enabling the flag's telemetry if app insights is not set.
image

The idea is just to show a warning if app insight is not linked to the store but allow users to enable telemetry for their flags. Just to confirm we support other avenues to collect telemetry other than app insights?

src/azure-cli/azure/cli/command_modules/appconfig/feature.py
ChristineWanjau: The exception catches failures from `resolve_metadata`, which can fail if the logged-in user lacks sufficient permissions to list stores in the subscription or if the store name doesn't match any existing store. The specific error message from `resolve_metadata` should propagate and be caught by this exception.

src/azure-cli/azure/cli/command_modules/appconfig/feature.py
ChristineWanjau: Yes it will run twice. Updated to just have line 121. Thanks

src/azure-cli/azure/cli/command_modules/appconfig/feature.py
mrm9084: We support other methods, they are just not built into the libraries.

yonzhan:
Please fix CI issues

src/azure-cli/azure/cli/command_modules/appconfig/feature.py
albertofori: Since we still allow enabling telemetry even though an app insights resource is not linked. would it be more accurate to say that telemetry will not be visible from the App Configuration portal if an app insights resource is not linked?

albertofori: I also remember from the email thread Christine had with us and some other folks that Zhenlan suggested we keep the portal warning but allow folks to enable/disable the telemetry toggle. We just won't show a graph if there is no linked app insights. I think this would unblock customers who might be using the other methods @mrm9084 is talking about in the event where they need to enable telemetry on portal/cli

yanzhudd (approved):

PR #32876, opened 2026-02-28
{Branch Management} CredScan Fixes

a0x1ab:
Merge fix #32847 back to dev.


yonzhan:

a0x1ab:
> can you please check if this approach works.. i mean the way you have created this PR from fork to dev.
> 
> I would have thought it will take release commit and push to dev? 

Correct, however the release branch already has these changes and the dev branch is the only one running behind. I see this being pushed in as a seperate commit (created as a fix that is unrelated to the release process itself) allowing us to push changes as if it were identified seperately, which would be from a fork dev anyway.

Attempting to fix the divergence now between dev and release becomes challenging as the dev branch is already ahead and mergeing changes from dev into release would be incorrect as well.

